PR #4490, opened 2015-01-28
Improved Operation.partitionId javadoc

pveentjer:


jerrinot:
:+1:

pveentjer:
This PR could be a bit more trickier than expected because of the PartitionAwareOperation interface

So what happens if an Operation would not implement this interface, but returns a a positive partitionId.

Or

What happens when an Operation does implement PartitionAwareOperation but returns -1 as partitionid.

jerrinot:
I guess the problem here is the base Operation class defines getPartitionId(). 
Perhaps only the Operations implementing the PartitionAwareOperation should have this method?

pveentjer:
Perhaps it would be better to deprecate PartitionAwareOperation and fully rely on the value provided by the getPartitionId. Then you remove the duplication.

jerrinot:
so you would effectively make all operations to be "PartitionAware" + explicitly define "-1" as not-partition-specific. 

That would do the trick as well. I believe that's how it currently works anyway.

pveentjer:
Correct. And you get rid of such logic:

```
  int getPartitionIdForExecution(Operation op) {
        return op instanceof PartitionAwareOperation ? op.getPartitionId() : -1;
    }
```

You call getPartitionId.

@gurbuzali @mdogan what is your opinion about this?

gurbuzali:
:+1: to removing PartitionAware

gurbuzali:
unrelated test failure

pveentjer:
It can't fail unless it is a checkstyle issue since it is only javadoc :)

pveentjer:
https://github.com/hazelcast/hazelcast/pull/4491

gurbuzali:
there is already an issue about the failing test, I've added a comment with the failure link to the issue.
no need to restart the build

PR #4491, opened 2015-01-28
Deprecated PartitionAwareOperation 

pveentjer:
Since it already provided by the Operation.getPartitionId method which is always available.

So people can still use the interface, but it will just be ignored by the system. The only thing the system cares about is the value of the getPartitionid method. If >=0, then partition specific, if <0, then 'generic'.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/operations/PostJoinOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/operations/PostJoinOperation.java
@@ -24,7 +24,6 @@ import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.OperationAccessor;
 import com.hazelcast.spi.OperationService;
-import com.hazelcast.spi.PartitionAwareOperation;
 import com.hazelcast.spi.ResponseHandler;
 import com.hazelcast.spi.UrgentSystemOperation;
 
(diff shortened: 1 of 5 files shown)


gurbuzali:
:+1:  apart from checkstyle errors (unused import)

pveentjer:
Hi Ali, fixing them now.

pveentjer:
verify

mdogan:
I agree that figuring out -if an operation belongs to a partition or not-  should be simple and can be achieved by looking at `partitionId` of an operation, no need for `PartitionAwareOperation`. 

As a historical note, I should add that `PartitionAwareOperation` had extra methods before, one was `getPartitionKey()` method which was used to find key-based RW-lock for an operation. 

But I'm not totally sure about removing `PartitionAwareOperation`, it may be stay as a marker interface for documentation purposes. Without it, it's not very easy to inspect partition-aware-operations in codebase. I know someone can forget to add this marker to an operation, because it's not required anymore... which makes its purpose invalid...

pveentjer:
I'm fine with it being a marker interface, as long as we should clearly document that it has no meaning for Hazelcast.

pveentjer:
I updated the PR. It clearly state that PartitionAwareOperatio is ony meant for documentation purposes, but that the system doesn't care for it any longer.

pveentjer:
verify

pveentjer:
verify

PR #4501, opened 2015-01-29
Review Request: Selector Imbalance, step. no 1

jerrinot:
Please do not merge it as yet. This is a step towards the selector imbalance solution. Read/WriteHandlers can now safely migrate between IOThreads. On top of this various migration strategies can be plugged.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java
@@ -102,7 +102,24 @@ public abstract class AbstractIOSelector extends Thread implements IOSelector {
             if (runnable == null) {
                 return;
             }
+            executeTask(runnable);
+        }
+    }
+
+    private void executeTask(Runnable runnable) {
+        IOSelector target = getTargetIOSelector(runnable);
+        if (target == this) {
             runnable.run();
+        } else {
+            target.addTask(runnable);
+        }
+    }
+
+    private IOSelector getTargetIOSelector(Runnable runnable) {
+        if (runnable instanceof MigratableHandler) {
+            return ((MigratableHandler) runnable).getOwner();
+        } else {
+            return this;
         }
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
@@ -25,16 +25,17 @@ import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.util.logging.Level;
 
-abstract class AbstractSelectionHandler implements SelectionHandler {
+public abstract class AbstractSelectionHandler implements SelectionHandler, MigratableHandler {
 
     protected final ILogger logger;
     protected final SocketChannelWrapper socketChannel;
     protected final TcpIpConnection connection;
     protected final TcpIpConnectionManager connectionManager;
-    protected final Selector selector;
-    protected final IOSelector ioSelector;
-    private final int initialOps;
+    protected Selector selector;
+    protected IOSelector ioSelector;
+
     private SelectionKey selectionKey;
+    private final int initialOps;
 
     public AbstractSelectionHandler(TcpIpConnection connection, IOSelector ioSelector, int initialOps) {
         this.connection = connection;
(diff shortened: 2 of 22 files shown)


jerrinot:
verify

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
```diff
@@ -46,15 +47,43 @@ public AbstractSelectionHandler(TcpIpConnection connection, IOSelector ioSelecto
         this.initialOps = initialOps;
     }
 
+    public void migrate(final IOSelector newIOSelector) {
+        if (ioSelector == newIOSelector) {
+            return;
+        }
+
+        ioSelector.addTask(new Runnable() {
```
pveentjer: Can you pull this into a regular inner class? I find anonymous inner classes always a PITA to read.

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java
```diff
@@ -102,7 +102,16 @@ private void processSelectionQueue() {
             if (runnable == null) {
                 return;
             }
-            runnable.run();
+            if (runnable instanceof WriteHandler) {
```
pveentjer: Can you add a few lines of comment what is being done here for future generations? I'm sure people will get confused.

pveentjer: What I often do, and it certainly is not a hard rule, and perhaps it might not be needed here, is that I pull  the processing of a task out of the loop into a separate method.

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
```diff
@@ -46,15 +47,43 @@ public AbstractSelectionHandler(TcpIpConnection connection, IOSelector ioSelecto
         this.initialOps = initialOps;
     }
 
+    public void migrate(final IOSelector newIOSelector) {
```
pveentjer: Can you add a few lines of Javadoc what the purpose is of this method?

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
```diff
@@ -46,15 +47,43 @@ public AbstractSelectionHandler(TcpIpConnection connection, IOSelector ioSelecto
...
-                selectionKey = socketChannel.register(selector, initialOps, this);
-            } catch (ClosedChannelException e) {
-                handleSocketException(e);
-            }
+            selectionKey = registerSelectionKey();
         }
+        return selectionKey;
+    }
 
+    protected SelectionKey registerSelectionKey() {
```
pveentjer: We have 2 methods to get a selection key; registerSelectionKey and getSelectionKey. I'm sure they are useful in a certain context; can you add some documentation when they should be used because now it is a bit of magic.

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java
pveentjer: Can you add some documentation the the AbstractIOSelector class explaining the mechanism of migrating a WriteHandler? Currently I see loose parts, but it is difficult to understand the whole mechanism. And this is d*mn tricky stuff :)

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
```diff
@@ -46,15 +47,43 @@ public AbstractSelectionHandler(TcpIpConnection connection, IOSelector ioSelecto
...
+    public void migrate(final IOSelector newIOSelector) {
+        if (ioSelector == newIOSelector) {
+            return;
+        }
+
+        ioSelector.addTask(new Runnable() {
+            @Override
+            public void run() {
+                unregisterOp(initialOps);
+                ioSelector = newIOSelector;
```
pveentjer: This is probably where you can null the selectionKey.

pveentjer: I'm missing a selectionKey.cancel:

http://stackoverflow.com/questions/13867133/is-there-a-way-to-de-register-a-selector-on-a-socket-channel

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
```diff
@@ -312,4 +310,8 @@ public void run() {
             EmptyStatement.ignore(e);
         }
     }
+
+    public IOSelector getTargetIOSelector() {
```
pveentjer: Would something with 'owner' not make more sense? You are currently migrating ownership of the WriteHandler from one selector to the next.

pveentjer:
I have added quite a lot of comments; not all need to be applied. But.. my major concern is that I was not able to understand from the code how the threading part works and I failed to quickly create a mental picture (I know how it should be done.. not how it currently is done). So I need to dig deeper in the technical design + we need to have more comprehensive documentation.

PR #4511, opened 2015-01-31
Relaxed Checkstyle constraints for(;;)

pveentjer:
Checkstyle doesn't want to have

```
for(;_;_) ....
```

But wants to have without spaces; and this is not what Intellij autoformatting respects. So this pr relaxes checkstyle checks. I think it is better to fix checkstyle because it causes the least amount of changes to people their environments and personally I don't care if a for loop with or without spaces is used.

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -183,11 +183,11 @@
 
         <!-- Checks for whitespace                               -->
         <!-- See http://checkstyle.sf.net/config_whitespace.html -->
-        <module name="EmptyForIteratorPad"/>
+        <!-- module name="EmptyForIteratorPad"/ -->
         <module name="GenericWhitespace"/>
         <module name="MethodParamPad"/>
-        <module name="NoWhitespaceAfter"/>
-        <module name="NoWhitespaceBefore"/>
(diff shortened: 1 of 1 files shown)


noctarius:
Well I just do `while(true) {}` instead :D

pveentjer:
I know that is an option, but I don't really like it. Perhaps it is a matter of taste or personal preference.

jerrinot:
I'd say most of the checkstyle rules are matter of personal preference. The problem is everyone has a slightly different preference and checkstyle enforces consistency.

pveentjer:
True, but in this case it is about a difference of checkstyle vs intellij. Intellij auto formats code in a different way than checkstyle prefers. With this fix, checkstyle doesn't complain about intellij its auto formatting.

gurbuzali:
I'm in favor of having intellij default auto-formatting as checkstyle

pveentjer:
This PR makes Checkstyle not conflict with Intellij regarding this particular issue.

Donnerbart:
I think it's a good idea to go with IntelliJ default in this case. It's hard enough to have all contributors using our (via checkstyle ensured) coding guidelines. If there is a minor thing where we can get closer to a common IDE default I'm with it.

PR #4512, opened 2015-01-31
Deprecated some OperationService methods.

pveentjer:
Too many methods are exposed in the OperationService which makes us less flexible. By deprecating it now, we can get rid of it earlier because people have the ability to deal with it ASAP.

Some of the things are going to be dropped completely like getting all kinds of statistics, this will be replaced by the blackbox.

Other things are moved to the InternalOperationService because they are internal implementation details not needed to expose to the SPI user. For example sending back a response.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/InternalOperationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/InternalOperationService.java
@@ -17,6 +17,7 @@
 package com.hazelcast.spi.impl;
 
 import com.hazelcast.instance.MemberImpl;
+import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Packet;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.OperationService;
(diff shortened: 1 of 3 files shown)


pveentjer:
verify

Donnerbart:
I'd really like to have a hint in the JavaDoc of deprecated methods, e.g. which method should be used instead or if the feature will be dropped completely etc. An example for this is `Member.getInetSocketAddress()`:
```
    /**
     * Returns the InetSocketAddress of this member.
     *
     * @return InetSocketAddress of this member
     *
     * @deprecated use {@link #getSocketAddress()} instead
     */
    @Deprecated
    InetSocketAddress getInetSocketAddress();
```
I think this will help customers to adapt their code faster.

pveentjer:
The methods deprecated have no alternatives.

The informational methods are going to be replaced by the blackbox.

The send(response) method doesn't belong here, it is an internal implementation detail and therefor is moved to the InternalOperationService. No user should be using this method in the first place.

pveentjer:
I improved the documentation.

mdogan:
Looks good to me. I favor lesser exposed/public methods.

hazelcast/src/main/java/com/hazelcast/spi/impl/InternalOperationService.java
```diff
@@ -62,4 +63,16 @@
      * Shuts down this InternalOperationService.
      */
     void shutdown();
+
+    /**
+     * Sends a response to a remote machine.
+     *
+     * This methods is deprecated since 3.5. It is an implementation detaul
```
Donnerbart: Little typo :)

gurbuzali:
fixed the typo manually

hazelcast/src/main/java/com/hazelcast/spi/impl/InternalOperationService.java
pveentjer: Thanks. I hope your dentist is a bit more careful ;)

PR #4513, opened 2015-02-01
DONT MERGE: Adding caller runs and work-stealing

pveentjer:
Performance in Hazelcast is partly dominated by the threading design: when an operation needs to be processed, it is handed over to the correct partition-queue where the partition-thread picks it up and eventually the response is send back using a wait/notifyAll.

It is a lot cheaper if a thread can execute its own operations, this skips this expensive mechanism and reduces the pressure in the partition-threads so that more remote operations can be executed. 

This is implemented using a different OperationScheduler implementation, so changes to the system are quite minimal. In the current OperationScheduler there is a single queue per operation-thread that is shared between different partitions. 

To steal work from this queue is very difficult since this queue contains an interleaved number of operations from different partitions. The new WorkQueue can be seen as a stack of PartitionQueues. Stealing becomes a lot easier since the whole PartitionQueue can be locked in 1 atomic action. When a PartitionQueue is stolen, it isn't removed from the WorkQueue; the partition thread will just ignore it. 

Some performance numbers

----------------------------------------------
only local calls
----------------------------------------------

class=com.hazelcast.stabilizer.tests.concurrent.atomiclong.AtomicLongTest
countersLength = 1000
threadCount = 40
keyLocality = local
writePercentage = 0

before
3,604,602.61 ops/s

after
17,590,757.57 ops/s

that is a 388 % performance improvement.

----------------------------------------------
random calls
----------------------------------------------

class=com.hazelcast.stabilizer.tests.concurrent.atomiclong.AtomicLongTest
countersLength = 1000
threadCount = 40
keyLocality = random
writePercentage = 0

before
677,200.00 ops/second

after
798,601.00 ops/s

that is a 18% performance improvement. Although 18% doesn't sound like a lot, we put in huge efforts to get a few percent. The nice thing about this optimization is that can speed any short operation in the system, so it isn't bound to a particular data-structure.

I'm sure that we can get at least another 100% performance improvements for local calls since a lot of debris is still on the way; for  example when a response is send back to the thread executing the operation (and executing its own operation) there is no need to do a wait/notify to signal a response.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -67,6 +67,12 @@ public class GroupProperties {
      * partition operations will queue behind other operations of different partitions. The default is 4.
      */
     public static final String PROP_PARTITION_OPERATION_THREAD_COUNT = "hazelcast.operation.thread.count";
+    public static final String PROP_PROGRESSIVE_SCHEDULER_ENABLED = "hazelcast.operation.progressivescheduler.enabled";
+
+    //public static final String PROP_PROGRESSIVE_SCHEDULER_ENABLED = "hazelcast.kickass.scheduler.enabled";
+
+
+
     public static final String PROP_GENERIC_OPERATION_THREAD_COUNT = "hazelcast.operation.generic.thread.count";
     public static final String PROP_EVENT_THREAD_COUNT = "hazelcast.event.thread.count";
     public static final String PROP_EVENT_QUEUE_CAPACITY = "hazelcast.event.queue.capacity";
@@ -210,6 +216,8 @@ public class GroupProperties {
     public static final String PROP_JCACHE_PROVIDER_TYPE = "hazelcast.jcache.provider.type";
 
     public final GroupProperty CLIENT_ENGINE_THREAD_COUNT;
+
+    public final GroupProperty PROGRESSIVE_SCHEDULER_ENABLED;
 
     public final GroupProperty PARTITION_OPERATION_THREAD_COUNT;
 
@@ -399,6 +407,8 @@ public class GroupProperties {
         EVENT_QUEUE_CAPACITY = new GroupProperty(config, PROP_EVENT_QUEUE_CAPACITY, "1000000");
         EVENT_QUEUE_TIMEOUT_MILLIS = new GroupProperty(config, PROP_EVENT_QUEUE_TIMEOUT_MILLIS, "250");
         CLIENT_ENGINE_THREAD_COUNT = new GroupProperty(config, PROP_CLIENT_ENGINE_THREAD_COUNT, "-1");
+
+        PROGRESSIVE_SCHEDULER_ENABLED = new GroupProperty(config, PROP_PROGRESSIVE_SCHEDULER_ENABLED, "true");
 
         CONNECT_ALL_WAIT_SECONDS = new GroupProperty(config, PROP_CONNECT_ALL_WAIT_SECONDS, "120");
         MEMCACHE_ENABLED = new GroupProperty(config, PROP_MEMCACHE_ENABLED, "true");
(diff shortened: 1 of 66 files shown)


pveentjer:
Small update.

The following test is very synthetic, so probably not something you will find in production.

If a single thread on a single member does a read on an IAtomicLong (cheapest structure we have), then without the optimization:

51,036.67 ops/s

With the optimization:

1,316,276.67 ops/s

That is almost 25 times faster. This shows how much impact the threading model has.

bwzhang2011:
@pveentjer , thanks for work-steal brought in.  how about some further test e.g for IQueue or ITopic or IMap (we care about that may be could fork your version for testing)or some test in stabilizer to leave us more message.

pveentjer:
ITopics are not going to benefit from this since they don't rely on operation-threads, but on event threads. We'll provide more tests. I hope this new executor can be included as experimental feature in 3.5 that needs to be enabled explicitly.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -67,6 +67,12 @@
      * partition operations will queue behind other operations of different partitions. The default is 4.
      */
     public static final String PROP_PARTITION_OPERATION_THREAD_COUNT = "hazelcast.operation.thread.count";
+    public static final String PROP_PROGRESSIVE_SCHEDULER_ENABLED = "hazelcast.operation.progressivescheduler.enabled";
+
+    //public static final String PROP_PROGRESSIVE_SCHEDULER_ENABLED = "hazelcast.kickass.scheduler.enabled";
```
jerrinot: this doesn't belong here

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
```diff
@@ -169,6 +161,27 @@
         startCleanupThread();
     }
 
+    private OperationExecutor initOperationExecutor() {
+        return new ProgressiveOperationExecutor(
```
jerrinot: the progressive executor should be used only when the group property is explicitly enabled, shouldn't it?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/progressive/PartitionIdScanThread.java
```diff
@@ -0,0 +1,59 @@
...
+import com.hazelcast.partition.InternalPartition;
+import com.hazelcast.spi.NodeEngine;
+import com.hazelcast.util.executor.HazelcastManagedThread;
+
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+
+public final class PartitionIdScanThread extends HazelcastManagedThread {
```
jerrinot: missing JavaDoc? what's a responsibility of this class?

jerrinot:
there are checkstyle issues.
 
moreover I think it deserves a proper documentation before it's merged - it's relative big architectural change.

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
pveentjer: For the time being it is enabled by default so I can run all the unit tests and see if something breaks. Will be reverted when I make the final PR.

pveentjer:
I'll close it for now. 

The engine is plugable and not enabled by default. Although documentation is needed and perhaps the whole formal process, it isn't used when a user has not explicitly configured it.

bwzhang2011:
@pveentjer, such issue merged into master branch ?

bwzhang2011:
@pveentjer, I noticed that the E-A doesn't include such feature and as you commented it would be run in plugable, so I wonder such progress.

PR #4515, opened 2015-02-02
Deprecated PartitionService.randomPartitionKey method

pveentjer:
Method (which I added a long time ago) has imho questionable value and we should deprecate it so we can reduce the number of methods we expose.


mdogan:
+1 no need for a random string generator in our API.

PR #4521, opened 2015-02-02
Use slf4j as default if found in classpath, fall back to default JDK logger otherwise

dpocock:
Many people already have slf4j in their project classpath so it is a good idea to check for it and just use it if present.

### DIFF
--- a/hazelcast-documentation/src/ConfigurationProperties.md
+++ b/hazelcast-documentation/src/ConfigurationProperties.md
@@ -79,7 +79,7 @@ Property Name | Default Value | Type | Description
 `hazelcast.initial.wait.seconds` | 0 | int  |   Initial time in seconds to wait before node to start completely.
 `hazelcast.map.replica.wait.seconds.for.scheduled.tasks`|10|int|Scheduler delay for the map tasks to be executed on backup members.
 `hazelcast.partition.count` | 271 | int  |   Total partition count.
-`hazelcast.logging.type` | jdk | enum |   Name of [logging](#logging-configuration) framework type to send logging events.
+`hazelcast.logging.type` | slf4j with fallback to jdk | enum |   Name of [logging](#logging-configuration) framework type to send logging events.
 `hazelcast.jmx` | false | bool  |   Enable [JMX](#monitoring-with-jmx) agent.
 `hazelcast.jmx.detailed` | false | bool  |   Enable detailed views on [JMX](#monitoring-with-jmx).
 `hazelcast.mc.max.visible.instance.count` | 100 | int  |   Management Center maximum visible instance count.
--- a/hazelcast-documentation/src/HazelcastConfiguration/AdvConfProperties.md
+++ b/hazelcast-documentation/src/HazelcastConfiguration/AdvConfProperties.md
@@ -68,7 +68,7 @@ Property Name | Default Value | Type | Description
 `hazelcast.jcache.provider.type`||string|Type of the JCache provider. Values can be `client` or `server`.
 `hazelcast.jmx` | false | bool  |   Enable [JMX](#monitoring-with-jmx) agent.
 `hazelcast.jmx.detailed` | false | bool  |   Enable detailed views on [JMX](#monitoring-with-jmx).
-`hazelcast.logging.type` | jdk | enum |   Name of [logging](#logging-configuration) framework type to send logging events.
+`hazelcast.logging.type` | slf4j with fallback to jdk | enum |   Name of [logging](#logging-configuration) framework type to send logging events.
 `hazelcast.map.load.chunk.size` | 1000 | int |   Chunk size for [MapLoader](#persistence) 's map initialization process (MapLoder.loadAllKeys()).
 `hazelcast.map.replica.wait.seconds.for.scheduled.tasks`|10|int|Scheduler delay for map tasks those will be executed on backup members.
 `hazelcast.master.confirmation.interval.seconds` | 30 | int  |   Interval at which nodes send master confirmation.
--- a/hazelcast-documentation/src/HazelcastConfiguration/LoggingConfiguration.md
+++ b/hazelcast-documentation/src/HazelcastConfiguration/LoggingConfiguration.md
@@ -5,11 +5,11 @@ Hazelcast has a flexible logging configuration and does not depend on any loggin
 
 To use built-in adaptors, you should set `hazelcast.logging.type` property to one of predefined types below.
 
--   **jdk**: JDK logging (default)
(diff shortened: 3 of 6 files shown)


pveentjer:
I don't think this PR is going to be merged. If people want to use SL4J, they can just select a system property. WDYT @mesutcelik

dpocock:
I agree that people can already use slf4j - this just helps them get started a little bit faster without having to think about it or research the Hazelcast logging properties, improving the first impression of Hazelcast and also reducing the number of instances where people deploy Hazelcast without remembering to set the property.

mesutcelik:
This PR is normally against fail-fast design choice since we are assuming slf4j is default logging framework in use.

I don't know if there is any trouble making situation where developer wants to use some other logging framework and made a mistake in its logging configuration(typo etc.). Although that is wrong configuration,  Hazelcast picks up slf4j by default from classpath. 

Can this PR mislead developer in such a case?

dpocock:
It is not assuming slf4j is default.  It is checking if slf4j is present.

If slf4j is not present, the Class.forName("org.slf4j.LoggerFactory") throws a ClassNotFoundException (which is being caught and ignored) and it uses the current default, jdk logging.

Therefore, there should not be any case where it can fail.

mesutcelik:
I am pointing the case where slf4j is on classpath, user forgot to configure or did a typo in configuration of its main choice of logging framework(not slf4j) and Hazelcast picked up slf4j directly from classpath...

That means something happened where developer did not notice... 
We had such issues before for some other configurations...

dpocock:
If slf4j is on the classpath it is responsible for finding their "real" logger (e.g. log4j) or sending things through the JDK logger (which is just what Hazelcast does by default anyway)

dpocock:
Maybe this logic could also do

   System.err.println("Hazelcast detected slf4j and will use it.  Explicitly set the property hazelcast.logging.type to stop seeing this warning or consider removing slf4j-api from your classpath if it shouldn't be there.");

jerrinot:
I like the idea of this PR. It makes a lot of sense to me.

I guess if the user makes a typo then Hazelcast should indeed fail to start. We can still preserve this behaviour. SLF4J should be used only and if only:
1. user selects SLF4J explicitly OR
2. user keeps everything as default and SLF4J is on a classpath.

pveentjer:
I'm against switching the SL4J just because it is on the classpath. It makes the mechanism magic and there are enough environments out there with complex jar dependencies (e.g. a application server). If a user wants to use SL4J, just set a system property or a groups-property.

mesutcelik:
After all this discussion, I am ok with the PR. We are already assuming that in case of no configuration, we fall back to JDK Logging. SLF4J will be one step before that assumption...

pveentjer:
So if I understand correctly:

- if SL4FJ is on the classpath and nothing configured explicitly, then SL4J
- if no SL4FJ is on the classpath and nothing configured explictly then java logging.

This sounds like magic to me. I also don't see why we should prefer SLF4J instead of Log4j. What if SLF4J/Logback and log4j is on the classpath.. what are we going to pick?

dpocock:
If slf4j is on the classpath then it is probably already sending somewhere that the user has explicitly chosen by including some other JAR like slf4j-log4j

If logback-classic is in the classpath then this logic will recognise that (because logback-classic includes an org.slf4j.LoggerFactory) and it will be used.

If the user has put these things on the classpath and not configured them properly then that is not Hazelcast's problem.  The aim of this patch is to simply try and respect the aims of users who have not explicitly asked to use JDK logging and to ensure they get going more quickly.

If users are adverse to any kind of magic configuration then they can always use the property to force the configuration to a specific logger.

pveentjer:
Most enterprisy projects I have worked on have huge quantities of jars on their classpath; either pulled in through maven or provided by the application server. So therefore I doubt that most users have full understanding of this new behaviour.

dpocock:
Peter, that is completely true and anticipated: that is why I wrote before, "If the user has put these things on the classpath and not configured them properly then that is not Hazelcast's problem"

Hazelcast will never be able to come up with something that is perfect for all users.  I don't feel my patch is some silver bullet, I just feel it gives some benefit for those who are deliberately adding slf4j to their classpath.

For those enterprise projects the JDK logging may never be seen anyway if they have some wrapper script that redirects stdout or if they are starting their process with some cluster management framework.

Tembrel:
On Tue, Feb 3, 2015 at 6:25 AM, Daniel Pocock <notifications@github.com>
wrote:

> Hazelcast will never be able to come up with something that is perfect for
> all users. I don't feel my patch is some silver bullet, I just feel it
> gives some benefit for those who are deliberately adding slf4j to their
> classpath.
>
But for those who are unintentionally dragging it in, it would be a
surprise.

No, I say the best thing is the status quo: default to j.u.logging unless
user explicitly configures.

For the record, I don't like the java logging, and I do use slf4j. I still
wouldn't want a magic default. I like the fact that you can examine my code
and see where explicitly configure for slf4j.


> For those enterprise projects the JDK logging may never be seen anyway if
> they have some wrapper script that redirects stdout or if they are starting
> their process with some cluster management framework.
>

The point is predictability. As you say, there's no silver bullet.

--tim

pveentjer:
How are we going to proceed with this PR. Some people are in favour of merging it, some people against.

Tembrel:
Hmm, how to respond without just repeating arguments in favor of my
position? :-)

It seems quite possible that there are folks who don't want to use slf4j
but still have it in their classpath because a third-party jar they use
depends on it. Accepting this PR could break their code.

And I say this as an slf4j user, so it's not me dumping on slf4j.

--tim

dpocock:
One alternative is to just make slf4j the default and throw an exception if it is not found at startup.  The exception message could tell people how to set the system property to use JDK logging.

Tembrel:
No, that's making things harder for people who don't want slf4j.

It's not so hard to set up the slf4j system property -- I do it:

        System.setProperty("hazelcast.logging.type", "slf4j");

I see no reason to bend over backwards to avoid having to put this line in.

--tim

enesakar:
Tembrel's argument is quite strong:
"It seems quite possible that there are folks who don't want to use slf4j
but still have it in their classpath because a third-party jar they use
depends on it. Accepting this PR could break their code."
Its possible harm bigger than what it introduces.
I vote against the PR.

gurbuzali:
Hi @dpocock 

First of all thank you for your efforts. We do really appreciate community contribution.
For this particular PR, arguments of developers/users who are against this change are strong.
Therefore I'm closing this PR

Please keep in mind that we value your efforts even though the PR is not merged

PR #4522, opened 2015-02-03
When near cache configured to store OBJECT then more efficient to nearca...

dirkvanrensburg:
When ```readBackupData``` is true and near caching is configured then the value is not placed in the near cache when it is read from backup. On high frequency reads the value will be read from the backup and de-serialised on every request, but if the near cache is configured with ```InMemoryFormat.OBJECT``` then storing it in the near cache is much more efficient. 

It has the downside of using more memory, but that can be controlled by the near cache TTL  and max size settings.

The included test does not test the performance improvement - just that the same object is returned, but local tests showed a 5 times improvement for high frequency gets of the same value when the near cache is configured with ```InMemoryFormat.OBJECT```.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -215,23 +215,25 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
                 return fromNearCache;
             }
         }
+
+        Data result = null;
         // todo action for read-backup true is not well tested.
         if (mapConfig.isReadBackupData()) {
-            final Object fromBackup = readBackupDataOrNull(key);
-            if (fromBackup != null) {
-                return fromBackup;
-            }
+            result = readBackupDataOrNull(key);
         }
-        final GetOperation operation = new GetOperation(name, key);
-        operation.setThreadId(ThreadUtil.getThreadId());
-        final Data value = (Data) invokeOperation(key, operation);
+
+        if (result == null) {
+            final GetOperation operation = new GetOperation(name, key);
+            operation.setThreadId(ThreadUtil.getThreadId());
+            result = (Data) invokeOperation(key, operation);
+        }
 
         if (nearCacheEnabled) {
             if (notOwnerPartitionForKey(key) || cacheKeyAnyway()) {
-                return putNearCache(key, value);
+                return putNearCache(key, result);
             }
         }
-        return value;
+        return result;
     }
 
     private boolean notOwnerPartitionForKey(Data key) {
--- a/hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
@@ -17,6 +17,7 @@
 package com.hazelcast.map.nearcache;
 
 import com.hazelcast.config.Config;
+import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.config.MapConfig;
 import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.core.EntryAdapter;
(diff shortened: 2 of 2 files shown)


bilalyasar:

dirkvanrensburg:
This used to build just fine, I won't have time to fix it right now though.... I'll try and get to it ASAP

jerrinot:
@dirkvanrensburg many thanks for this contribution! It seems as a good improvement to me. We need one more review to get it through. @ahmetmircik promised to review it tomorrow. In mean time, may I kindly ask you to sign our contributor agreement? https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement

Cheers,
Jaromir

dirkvanrensburg:
I have the agreement signed, where to I email it too?

jerrinot:
@dirkvanrensburg: Please send it to info@hazelcast.com 
Thanks!

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
```diff
@@ -521,6 +522,28 @@ public void testNearCacheEntriesNotExpired_afterIMapExpiration() throws Exceptio
...
 
+    @Test
+    public void testNearCacheBackups() throws Exception {
+        final String mapName = randomMapName();
+        final Config config = createNearCachedMapConfig(mapName);
+
+        //Set up backup
+        final MapConfig mapConfig = config.getMapConfig(mapName).setReadBackupData(true);
+        mapConfig.setBackupCount(1);
+        mapConfig.getNearCacheConfig().setInMemoryFormat(InMemoryFormat.BINARY.OBJECT);
```
ahmetmircik: seems a left-over and it should be only `InMemoryFormat.OBJECT ` ?

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -215,23 +215,25 @@ protected Object getInternal(Data key) {
...
+            result = (Data) invokeOperation(key, operation);
         }
-        final GetOperation operation = new GetOperation(name, key);
-        operation.setThreadId(ThreadUtil.getThreadId());
-        final Data value = (Data) invokeOperation(key, operation);
 
         if (nearCacheEnabled) {
             if (notOwnerPartitionForKey(key) || cacheKeyAnyway()) {
-                return putNearCache(key, value);
+                return putNearCache(key, result);
```
ahmetmircik: we do not need to put all `result` here, only put `result` into `near-cache` if the relevant formats are the case as you stated in the PR comment, so if the `IMap` is `BINARY` and `near-cache` is `OBJECT` you can put the value read from backup into `near-cache` otherwise there will be unnecessary memory consumption.

ahmetmircik:
Hi @dirkvanrensburg,  i want to put this PR into sleep for a while, here are some of my reasons that makes me think about like this :

First, `read-backup-data` is not useful for large clusters, the probability of reading from backup is 1+backup-count/100 in a 100 node cluster. So this fix will help much for small clusters.

Secondly, without setting `read-backup-data = true`, `IMap+ near-cache` should perform similar after first remote reads.

Thirdly, current users of this feature may see unexpected increase of memory consumptions if we apply this feature, assuming they are not using near-cache eviction/expiration or they are using heap based eviction policies of IMap.

And last, i am not sure whether we are fixing a general use-case or a specific one.

dirkvanrensburg:
Hi @ahmetmircik 

I am fine with that. It was for a specific usecase. A client was using Hazelcast in a 2-4 node configuration and they used the read-backup and near cache. Objects were stored in serialized form but as OBJECT in near cache. They were seeing slower response times when objects are retrieved from the backup because of repeated deserialisation.

enesakar:
Thanks for contribution.  I see Ahmet (who is module owner of map) has concerns about this development. But also this fix introduces a type of conflict with near-cache's basic definition "caching entries fetched from 'remote' nodes". So I am closing the PR without merging.

PR #4530, opened 2015-02-04
Fixes #4527

### LINKED ISSUE #4527: [TEST-FAILURE] ExecutorServiceTest.testSubmitMultipleNode
The cause is the following. 

When the submit is called, a 'random' partition id is selected. Then the operation is send to the machine owning that partition. Because we don't check the PartitionAwareOperation interface anymore in the BasicOperationService.ensureNoPartitionProblems, we continue with the actual partition checking since the partition-id is set.

Then we run into the actual problem: BasicOperationService.ensureNoPartitionProblems gets a null owner for this call:

  Address owner = internalPartition.getReplicaAddress(op.getReplicaIndex());
          
So the owner for the partition has not yet been set and therefore the call is rejected and that is what you see in the logging:

```
23:02:10,978  WARN [CallableTaskOperation] hz._hzInstance_2_dev.partition-operation.thread-7 - [127.0.0.1]:5002 [dev] [3.5-SNAPSHOT] com.hazelcast.spi.exception.WrongTargetException: WrongTarget! this:Address[127.0.0.1]:5002, target:null, partitionId: 127, replicaIndex: 0, operation: com.hazelcast.executor.impl.operations.CallableTaskOperation, service: hz:impl:executorService
```

So the big questions are: why is the owner not yet set, and if it isn't set yet, why do we fail immediately? Should we wait for an owner? And should not a better exception be thrown like PartionNoOwnerException and be retried?

The problem in the test could be solved by adding a warmup partitions call; but I doubt if that is a correct solution.

pveentjer:
BaseCallableTaskOperation runs into WrongTargetException but it doesn't send a reply
to the Invocation and therefor no retry mechanism.

This is a hack to force sending back a wrongtargetexception.

There are multiple design issues. The BaseCallableTaskOperation is not really partition specific since it only cares about a membe,r not a partition. Also the returns response mechanism is too coarse grained. we need to look at it in the future because it doesn't deal well with control flow logic.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/executor/impl/operations/BaseCallableTaskOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/executor/impl/operations/BaseCallableTaskOperation.java
@@ -39,6 +39,14 @@ abstract class BaseCallableTaskOperation extends Operation implements TraceableO
     protected transient Callable callable;
     private Data callableData;
 
+    // transient.
+    // We are cheating a bit here. The idea is the following. A BaseCallableTaskOperation is always going to be send to a
+    // partition, but the operation doesn't send a response directly and therefor when a WrongTargetException is thronw (e.g.
+    // partition has moved) the operation is not retried. To prevent this from happening, we say that we return a response until
+    // the before-run method is called. Then we know we are going to be offloaded to a different thread and we are not returning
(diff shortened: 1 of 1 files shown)


gurbuzali:
@mdogan any objection ?

mdogan:
nop

PR #4531, opened 2015-02-04
Interceptor id generation de-coupled from adding interceptor, this way w...

eminn:
...e are not adding map interceptors up-front. This enables features like quorum to prevent the operation. Operations should do their job inside their run method, not in the proxy etc.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/AbstractMapServiceContextSupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/AbstractMapServiceContextSupport.java
@@ -152,8 +152,15 @@ abstract class AbstractMapServiceContextSupport implements MapServiceContextSupp
     }
 
     @Override
-    public String addInterceptor(String mapName, MapInterceptor interceptor) {
-        return mapServiceContext.getMapContainer(mapName).addInterceptor(interceptor);
+    public void addInterceptor(String id, String mapName, MapInterceptor interceptor) {
+        MapContainer mapContainer = mapServiceContext.getMapContainer(mapName);
+        mapContainer.addInterceptor(id, interceptor);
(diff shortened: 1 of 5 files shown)


ahmetmircik:
i like PRs with `deletions > additions` :+1:

gurbuzali:
you will do the client changes in another PR ?

gurbuzali:
it's already there, my bad

eminn:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/AbstractMapServiceContextSupport.java
```diff
@@ -152,8 +152,14 @@ public void interceptAfterRemove(String mapName, Object value) {
     }
 
     @Override
-    public String addInterceptor(String mapName, MapInterceptor interceptor) {
-        return mapServiceContext.getMapContainer(mapName).addInterceptor(interceptor);
+    public void addInterceptor(String id, String mapName, MapInterceptor interceptor) {
+        mapServiceContext.getMapContainer(mapName).addInterceptor(id, interceptor);
```
pveentjer: This is called a train wreck. Every word is a wagon and every dot is a connection between the train wagons. One of the problems with train wrecks is that if you get a NPE in production, you have no idea which wagon caused it.

I see that this style is often being used, especially in the map, but it is a well known bad practice.

http://blog.shamess.info/2012/05/31/clean-code-and-train-wrecks/

eminn: Yeah, I know what it called, the purpose of this change was not refactoring but a change of wrong behavior. Will fix it anyway

PR #4532, opened 2015-02-04
Introduced ClusterClock

pveentjer:
Pulled out the logic from ClusterService. 

This PR is a cleanup; extracting out a concept in its own structure.

It also is a step towards reducing the need for NodeEngine in OperationService. This is going to reduce tight coupling and make it easier to test.


Donnerbart:
If I didn't miss anything just internal methods were removed (`ClusterService.getClusterTime()`, `NodeEngine.getClusterTime()`), all public methods have still the same interface (`Cluster.getClusterTime()`). Looks good to me.

pveentjer:
verify

PR #4538, opened 2015-02-04
Added the PacketTransceiver

pveentjer:
The NodeEngine has a lot of responsibilities. Among these was dealing with sending and receiving packets. I have pulled out this logic into the PacketTransceiver which is responsible for transmitting (sending) and receiving packets.

The term transceiver is from the radio terminology:
http://en.wikipedia.org/wiki/Transceiver


ajermakovics:
+1

PR #4540, opened 2015-02-05
Optimization to prevent service name serialization

pveentjer:
Simplified version of:

https://github.com/hazelcast/hazelcast/pull/4468

The idea is that the API's remain the same but that the operation skips serialization of the service name if there is no difference compared to the value returned by the getServiceName method. 

So when an operation overrides this method and later the op.setServiceName with the same name is called, then the serialization (and therefore deserialization) is just skipped. 

I also added to most of the operation in the system the correct getServiceName implementation so that they can benefit from this serialization optimization.


hazelcast/src/test/java/com/hazelcast/spi/OperationSerializationTest.java
```diff
@@ -157,23 +163,73 @@ public void assertSerializationCloneEquals(Operation expected) {
...
+        assertFalse("service name should not be set", copy.isFlagSet(Operation.BITMASK_SERVICE_NAME_SET));
+    }
+
+    @Test
+    public void test_serviceName_whenNotOverridesServiceName_thenSerialized(){
+        DummyOperation op = new DummyOperation();
+        op.setServiceName(DUMMY_SERVICE_NAME);
+        assertSame(DUMMY_SERVICE_NAME, op.getRawServiceName());
+        assertSame(DUMMY_SERVICE_NAME, op.getServiceName());
+        assertTrue("service name should not be set", op.isFlagSet(Operation.BITMASK_SERVICE_NAME_SET));
```
Donnerbart: To be or not to be! You have never experienced Shakespeare until you have read him in the original Klingon.

(I think you want to have to other one in the assert message)

hazelcast/src/test/java/com/hazelcast/spi/OperationSerializationTest.java
```diff
@@ -157,23 +163,73 @@ public void assertSerializationCloneEquals(Operation expected) {
...
+
+        Operation copy = copy(op);
+        assertCopy(DUMMY_SERVICE_NAME, copy.getServiceName());
+        assertCopy(DUMMY_SERVICE_NAME, copy.getRawServiceName());
+        assertTrue("service name should not be set", copy.isFlagSet(Operation.BITMASK_SERVICE_NAME_SET));
```
Donnerbart: Same here :)

Donnerbart:
With the last commit (fix of test) the change looks logical and good to me.

I had a brief look how `Operation.setServiceName()` is called in Hazelcast and the constants seem to be used everywhere. So indeed `==` instead of `equals()` should cause no problems.

pveentjer:
verify

PR #4541, opened 2015-02-05
fix #4533

### LINKED ISSUE #4533: hazelcast.xml variable replacement only works at the front of an xml attribute string
For example, I would like to do:
```xml
<hazelcast>
  <import resource="hazelcast/group-${env}-config.xml"/>
  ...
```
But this tries to import a file called "qa-config.xml" instead of "hazelcast/group-qa-config.xml"

Here is a unit test that demonstrates the same problem.  I am on version 3.4
```java
    @Test
    public void testXmlVariableReplacement() throws Exception {

        String xml = "<hazelcast>\n" +
                "    <properties>\n" +
                "        <property name=\"${env}-with-suffix\">local-with-suffix</property>\n" +
                "        <property name=\"with-prefix-${env}\">with-prefix-local</property>\n" +
                "    </properties>\n" +
                "</hazelcast>";

        // Simulate setenv.sh service configuration
        System.setProperty("env", "local");

        InputStream is = IOUtils.toInputStream(xml, "UTF-8");
        Config config = new XmlConfigBuilder(is).build();
        Properties properties = config.getProperties();
        
        // ok
        assertEquals(properties.getProperty("local-with-suffix"), "local-with-suffix");

        // this fails because "with-prefix-${env}" resolved as "local"
        assertEquals(properties.getProperty("with-prefix-local"), "with-prefix-local");
    }
```

bilalyasar:
only changed XmlConfigPreProcessor.java#replaceVariables.
1) node.value copied to stringBuilder.
2) all replacements are made on stringBuilder
3) finally string builder value is copied to node.value

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
@@ -19,14 +19,15 @@ package com.hazelcast.config;
 import com.hazelcast.core.HazelcastException;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.annotation.QuickTest;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.Properties;
-import org.junit.Test;
-import org.junit.experimental.categories.Category;
-import org.junit.runner.RunWith;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -44,10 +45,7 @@ public class XMLConfigPreProcessorTest {
                 "        <import resource=\"\"/>\n" +
                 "   </network>" +
                 "</hazelcast>";
-
-        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
-        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
-        Config config = configBuilder.build();
+        buildConfig(xml, null);
     }
 
     @Test(expected = HazelcastException.class)
@@ -56,10 +54,7 @@ public class XMLConfigPreProcessorTest {
                 "   <hazelcast>" +
                 "   </hazelcast>" +
                 "</hazelcast>";
-
-        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
-        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
-        Config config = configBuilder.build();
+        buildConfig(xml, null);
     }
 
     @Test
@@ -72,17 +67,13 @@ public class XMLConfigPreProcessorTest {
                         "        <async-backup-count>${notreplaced}</async-backup-count>\n" +
                         "    </semaphore>" +
                         "</hazelcast>";
-        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
-        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
 
         Properties properties = new Properties();
         properties.setProperty("name", "s");
         properties.setProperty("initial.permits", "25");
         properties.setProperty("backupcount.part1", "1");
         properties.setProperty("backupcount.part2", "0");
-        configBuilder.setProperties(properties);
-
-        Config config = configBuilder.build();
+        Config config = buildConfig(xml, properties);
         SemaphoreConfig semaphoreConfig = config.getSemaphoreConfig("s");
         assertEquals(25, semaphoreConfig.getInitialPermits());
         assertEquals(10, semaphoreConfig.getBackupCount());
@@ -107,13 +98,9 @@ public class XMLConfigPreProcessorTest {
                 "    <import resource=\"${config.location}\"/>\n" +
                 "</hazelcast>";
 
-        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
-        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
-
         Properties properties = new Properties();
         properties.setProperty("config.location", file.getAbsolutePath());
-        configBuilder.setProperties(properties);
-        Config config = configBuilder.build();
+        Config config = buildConfig(xml, properties);
         JoinConfig join = config.getNetworkConfig().getJoin();
         assertFalse(join.getMulticastConfig().isEnabled());
         assertTrue(join.getTcpIpConfig().isEnabled());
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
```diff
@@ -500,6 +501,54 @@ public void testXmlDeniesDuplicateMemberAttributesConfig() throws Exception {
...
+    public void testXmlVariableReplacementAsSubstring() throws Exception {
+
+        String xml = "<hazelcast>\n" +
+                "    <properties>\n" +
+                "        <property name=\"${env}-with-suffix\">local-with-suffix</property>\n" +
+                "        <property name=\"with-prefix-${env}\">with-prefix-local</property>\n" +
+                "    </properties>\n" +
+                "</hazelcast>";
+
+        System.setProperty("env", "local");
```
pveentjer: Can you try it without setting a sytem property? The same goes for the other test. The replacement mechanism should work using normal properties as well.

hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
```diff
@@ -500,6 +501,54 @@ public void testXmlDeniesDuplicateMemberAttributesConfig() throws Exception {
...
+        String xml = "<hazelcast>\n" +
+                "    <import resource=\"file://" + "${file}" + "\"/>\n" +
+                "</hazelcast>";
+
+        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
+        System.setProperty("file", file.getAbsolutePath());
+        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
+        Config config = configBuilder.build();
+        JoinConfig join = config.getNetworkConfig().getJoin();
+        assertFalse(join.getMulticastConfig().isEnabled());
```
pveentjer: Personally I would not use the network override to detect if a file is correctly imported. I needed to look at the config to determine what the defaults were. Personally I would go for adding a property in the property section. 

```
<properties>
   <property name="foo">bar</name>
</properties>
```

And then check if the property 'foo' has value 'bar'.

pveentjer:
Minor review comments. Apart from that it looks good to me.

bilalyasar:
@pveentjer, very thanks Peter for your comments, updated.

hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
```diff
@@ -500,6 +501,56 @@ public void testXmlDeniesDuplicateMemberAttributesConfig() throws Exception {
...
+        configBuilder.setProperties(properties);
+        Config config = configBuilder.build();
+
+        assertEquals(config.getProperty("local-with-suffix"), "local-with-suffix");
+        assertEquals(config.getProperty("with-prefix-local"), "with-prefix-local");
+    }
+
+    @Test
+    public void testXmlImportWithVariableReplacementAsSubstring() throws Exception {
+        Properties properties = new Properties();
```
pveentjer: A tip; only declare your variable at the moment you really need it. Now it is spread over the method and makes it more difficult to follow. You do the same thing with the previous test although not that bad.

pveentjer: Also there is some copy paste programming to make a Config. Please pull it out in a method to reduce copy paste + increase readiblity. E.g.

```
 @Test
    public void testXmlVariableReplacementAsSubstring() throws Exception {
        String xml = "<hazelcast>\n" +
                "    <properties>\n" +
                "        <property name=\"${env}-with-suffix\">local-with-suffix</property>\n" +
                "        <property name=\"with-prefix-${env}\">with-prefix-local</property>\n" +
                "    </properties>\n" +
                "</hazelcast>";

 	Properties properties = new Properties();
        properties.setProperty("env", "local");

	Config config = build(xml, properties);

        assertEquals(config.getProperty("local-with-suffix"), "local-with-suffix");
        assertEquals(config.getProperty("with-prefix-local"), "with-prefix-local");
    }

public Config build(String xml, Properties props){
        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
        configBuilder.setProperties(properties);
        return configBuilder.build();
}
```

Please consider cleaning up the rest of the methods to make use of the same build method.

bilalyasar: okey, i will cleanup whole methods in this test class according to your suggestions.. thanks

pveentjer: You could even consider something like this (building on top of a varargs of string):

```
    Config config = build(xml, properties, "env", "local");
```

Much cleaner than that you can't get it. Saves you another few non interesting lines from your test.

```
  public void testXmlVariableReplacementAsSubstring() throws Exception {
        String xml = "<hazelcast>\n" +
                "    <properties>\n" +
                "        <property name=\"${env}-with-suffix\">local-with-suffix</property>\n" +
                "        <property name=\"with-prefix-${env}\">with-prefix-local</property>\n" +
                "    </properties>\n" +
                "</hazelcast>";

      Config config = build(xml, "env", "local");

        assertEquals(config.getProperty("local-with-suffix"), "local-with-suffix");
        assertEquals(config.getProperty("with-prefix-local"), "with-prefix-local");
    }
```

hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
```diff
@@ -500,6 +501,56 @@ public void testXmlDeniesDuplicateMemberAttributesConfig() throws Exception {
...
+                "    <import resource=\"file://" + "${file}" + "\"/>\n" +
+                "</hazelcast>";
+
+        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
+        properties.setProperty("file", file.getAbsolutePath());
+        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
+        configBuilder.setProperties(properties);
+        Config config = configBuilder.build();
+        JoinConfig join = config.getNetworkConfig().getJoin();
+        assertFalse(join.getMulticastConfig().isEnabled());
```
pveentjer: I see you are still relying on network config to determine of the substitution has worked. I really would use the same approach as you do in test testXmlVariableReplacementAsSubstring

hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
pveentjer: Thanks.

Please keep in mind: readability for tests is extremely important because they are an executable specification if your system. If they are more difficult to read than needs to be, the loose their value because people have a hard time figuring out the purpose of a test.

So for normal code readability is important, but for tests imho even more so.

jerrinot: +1

bilalyasar:
@pveentjer, updated.

hazelcast/src/test/java/com/hazelcast/config/XMLConfigPreProcessorTest.java
```diff
@@ -213,10 +186,7 @@ public void testImportEmptyResourceThrowsException() throws Exception {
         String xml = "<hazelcast>\n" +
                 "    <import resource=\"\"/>\n" +
                 "</hazelcast>";
-
-        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
-        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
-        configBuilder.build();
+        buildConfig(xml, null);
```
pveentjer: I'm not going to be very problematic about this; but normally I would create an additional buildConfig method which doesn't need a null. And perhaps it just forwards to buildConfig(xml, null)... but at least you don't see it in the code.

You don't need to change it, but passing in null values to an argument always gives me an strange feeling + decreases readability of the test because the reader is going to think.. what is going on here! And he is getting confused..

pveentjer:
Looks good to me. @gurbuzali can you also have a look?

PR #4529, opened 2015-02-04
Pulled out HazelcastThreadGroup

pveentjer:
Introduced the HazelcastThreadGroup and pulled out that functionality from Node/NodeEngine. 

Also cleaned up the dependencies of the BasicOperationScheduler so it doesn't rely anymore on Node/NodeEngine. This was the intent of this PR so I can start to test it without having a HazelcastInstance up and running. 

This is going to be the first PR of many where I'm going to remove the need of Node/NodeEngine inside components to improve testability and reduce tight coupling.

This PR relies on the following PR to be merged first:
https://github.com/hazelcast/hazelcast/pull/4526

PR #4546, opened 2015-02-05
Introducing the InternalNodeEngine

pveentjer:
Currently a lot of services are directly tied to the NodeEngineImpl and therefore there is a dependency on Node and everything else. This makes it very difficult to test in isolation; you need to have a running cluster. It makes is almost impossible to mock things because we can't control dependencies.

The InternalNodeEngine interface protects us against this. A lot of services like OperationService, EventService etc, get an instance of the InternalNodeEngine injected. For testing we can mock it if needed.

I have currently fixed the map reduce package so it makes use of the InternalNodeEngine interface. Once this is merged, I'll take care of other services.

It follows the same naming/interface approach we use for other services. E.g
OperationService -> public interface
InternalOperationService ->private interface
BasicOperationService -> concrete implementation

The InternalNodeEngine has abilities that are only for internal use and we don't want to expose on the NodeEngine interface.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
@@ -60,6 +60,7 @@ import com.hazelcast.spi.PostJoinAwareService;
 import com.hazelcast.spi.ProxyService;
 import com.hazelcast.spi.impl.InternalOperationService;
 import com.hazelcast.spi.impl.NodeEngineImpl;
+import com.hazelcast.spi.impl.ServiceManager;
 import com.hazelcast.transaction.TransactionManagerService;
 import com.hazelcast.util.executor.ExecutorType;
 
@@ -467,7 +468,8 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
                 return;
             }
 
-            Object service = nodeEngine.getService(serviceName);
+            ServiceManager serviceManager = nodeEngine.getServiceManager();
+            Object service = serviceManager.getService(serviceName);
             if (service == null) {
                 if (nodeEngine.isActive()) {
                     throw new IllegalArgumentException("No service registered with name: " + serviceName);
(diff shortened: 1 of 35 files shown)


noctarius:
Is it possible you created that branch from the wrong basis? It seems to have the unrelated changes for the threading system changes?

pveentjer:
Good one. Well.. I need to mention it; I'm waiting for another PR to be merged.

https://github.com/hazelcast/hazelcast/pull/4529

noctarius:
hehe :D

pveentjer:
The PR this PR was waiting for has been merged.

pveentjer:
verify

gurbuzali:
looks good, can we have another pair of eyes too?

ajermakovics:
In some cases NodeEngine is cast to InternalNodeEngine and in others to still to NodeEngineImpl. Wasn't sure if that's intentional but I guess it's something we can change as we go along. Otherwise it's a +1 from me

ajermakovics:
Since a lot of operations are casting to InternalNodeEngine perhaps we should introduce a:
```java
abstract class AbstractInternalOperation extends AbstractOperation {
    @Override
    public InternalNodeEngine getNodeEngine() {
        return (InternalNodeEngine) super.getNodeEngine();
    }
}
```
?

jerrinot:
isn't it rather an indication the NodeEngine interface is insufficient?

On Tue, Feb 10, 2015 at 10:22 AM, Andrejs Jermakovics <
notifications@github.com> wrote:

> Since a lot of operations are casting to InternalNodeEngine perhaps we
> should introduce a:
>
> abstract class AbstractInternalOperation extends AbstractOperation {
>     @Override
>     public InternalNodeEngine getNodeEngine() {
>         return (InternalNodeEngine) super.getNodeEngine();
>     }
> }
>
> ?
>
> —
> <https://github.com/hazelcast/hazelcast/pull/4546#issuecomment-73676411>.
>



-- 
“Perfection is achieved, not when there is nothing more to add, but when
there is nothing left to take away.”
Antoine de Saint Exupéry

ajermakovics:
it probably is. so this goes back to what we do and don't expose via the SPI.

pveentjer:
I don't think the node engine is insufficient. We don't need to expose every functionality to the end user, some is very specific. In these cases it is better to hide this private API behind an InternalObject cast. Users don't see it and if you cast to it, you are on your own because we can change it as much as we want.

noctarius:
Well but if most Operations have to cast to InternalNodeEngine anyways I would agree with Jaromir. That means users are probably never able to implement what they want (or have to cast to internal API). Haven't had a look why most ops cast but as a SPI user it sounds familiar to me that I miss some things :)

pveentjer:
Most operations don't need to cast to InternalNodeEngine.

gurbuzali:
needs rebase

Issue #4547, opened 2015-02-06
Allow Comparable objects in a collection to be added to an index


dirkvanrensburg:
Some objects contains a collection in them and it will be useful to allow the values in the collection to be added to an index if they are Comparable.

*For example:* Given a Document object with a Collection of categories.

```java
public class Document implements Serializable {
    private String contents;
    private List<String> categories;
}
```
An index can be added for the *`categories`*

```java
map.addIndex( "categories", false);
```
To find all the documents with the category 'fiction'

```java
EntryObject e = new PredicateBuilder().getEntryObject();
Predicate p = e.get("categories").equal("fiction");

Collection<Document> documents = map.values(p);
```

jerrinot:
fixed by #6525

PR #4548, opened 2015-02-06
Allow for collections to be added to the index (issue #4547)

dirkvanrensburg:
Proposed implementation for issue #4547

### DIFF
--- a/hazelcast-documentation/src/QueryIndexing.md
+++ b/hazelcast-documentation/src/QueryIndexing.md
@@ -49,6 +49,36 @@ The following is the Spring declarative configuration for the same sample.
   </hz:indexes>
 </hz:map>
 ```
+
+#### Indexing Collections
+
+All non primitive types should implement *`Comparable`*, but it is possible to add a *`Collection`* of *`Comparable`*
+objects to the index. 
+
+*For example:* Given a Document object with a Collection of categories.
+
+```java
+public class Document implements Serializable {
+    private String contents;
+    private List<String> categories;
+}
+```
+An index can be added for the *`categories`*
+
+```java
+map.addIndex( "categories", false);
+```
+To find all the documents with the category 'fiction'
+
+```java
+EntryObject e = new PredicateBuilder().getEntryObject();
+Predicate p = e.get("categories").contains("fiction");
+
+Collection<Document> documents = map.values(p);
+```
+
+This will then use the index to find all the documents that in their list of categories have the value 'fiction'.
+
 <br></br>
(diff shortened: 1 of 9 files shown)


hazelcast-documentation/src/QueryIndexing.md
```diff
@@ -49,6 +49,36 @@ The following is the Spring declarative configuration for the same sample.
...
+An index can be added for the *`categories`*
+
+```java
+map.addIndex( "categories", false);
+```
+To find all the documents with the category 'fiction'
+
+```java
+EntryObject e = new PredicateBuilder().getEntryObject();
+Predicate p = e.get("categories").equal("fiction");
```
ajermakovics: I really like the idea of this PR but I think we should use `contains` instead of `equal` for checking if an item is in a collection. `equal` will be useful in comparing two collections.

dirkvanrensburg:
Hi @ajermakovics 

I agree that it would sound more natural and I'll add a predicate for contains. I now realise that this change will throw an exception if no index is added for the collection.

Regards,
Dirk

dirkvanrensburg:
Hi @ajermakovics 

I updated the PR to now include a contains Predicate. It assumes that the attribute is a Collection<Comparable> but this is similar to the assumption that the attribute is Comparable

Let me know what you think.

Cheers,
Dirk

ajermakovics:
Hi Dirk, I think the API is more intuitive with `contains`. Will try to have a closer look at the implementation

ajermakovics:
It's really great that there's a test but I believe more cases should be covered like removing entries with collections and updating entries with collections.

We also should make sure that attributes with collections work correctly with other predicates. For example, currently `isNotNull()` and `notEqual()` don't return any results on these attributes. Not sure yet how all the other predicates behave.

I can understand why `IndexImpl.collectionValues` exists but tbh favor storing all records in `IndexImpl.recordValues`.  

While trying out this code I made some changes and ran test cases for the mentioned predicates. The code is in this branch: https://github.com/ajermakovics/hazelcast/commit/b35ce40284907e88b7b1dccaba96a0766d46fa2c

All in all this is looking good and +1 for documentation update.

dirkvanrensburg:
Hi,

Thanks for the feedback, I am a bit caught up at work so I couldn't get to work on any of this. I'll spend some time later in the week.

Cheers,

Dirk

bilalyasar:

pveentjer:
I'm going to close this one due to inactivity. Feel free to re-open it.

wildnez:
Reopening as there is a PRD in place now for this particular case.

jerrinot:
@dirkvanrensburg: I'm closing this PR as it's effectively surpassed by #6525
Thanks for you effort, this PR served as a great inspiration for us!

Issue #4494, opened 2015-01-28
[TEST-FAILURE] ClientInvocationTest.executionCallback_TooLongThrowableStackTrace


bilalyasar:
```
java.lang.AssertionError: Callback should be notified on time!
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.client.spi.impl.ClientInvocationTest.executionCallback_TooLongThrowableStackTrace(ClientInvocationTest.java:85)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.6/com.hazelcast$hazelcast-client/435/testReport/junit/com.hazelcast.client.spi.impl/ClientInvocationTest/executionCallback_TooLongThrowableStackTrace/

sancar:
Closed via #4551

PR #4551, opened 2015-02-06
fix attempt for #4494. Client connection destroy was running twice if re...

sancar:
...ad and writehandler get socket exception at the same time. Changed the implementation to remove the connection once

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
@@ -113,20 +113,9 @@ public class ClientConnectionTest extends HazelcastTestSupport {
         HazelcastClientInstanceImpl clientImpl = ClientTestUtil.getHazelcastClientInstanceImpl(client);
         ClientConnectionManager connectionManager = clientImpl.getConnectionManager();
 
+        final CountingConnectionRemoveListener listener = new CountingConnectionRemoveListener();
 
-        final AtomicInteger count = new AtomicInteger();
-
-        connectionManager.addConnectionListener(new ConnectionListener() {
-            @Override
-            public void connectionAdded(Connection connection) {
-
-            }
-
-            @Override
-            public void connectionRemoved(Connection connection) {
-                count.incrementAndGet();
-            }
-        });
+        connectionManager.addConnectionListener(listener);
 
         final Address serverAddress = new Address(server.getCluster().getLocalMember().getSocketAddress());
         final Connection connectionToServer = connectionManager.getConnection(serverAddress);
@@ -134,6 +123,22 @@ public class ClientConnectionTest extends HazelcastTestSupport {
         connectionManager.destroyConnection(connectionToServer);
         connectionManager.destroyConnection(connectionToServer);
 
-        assertEquals("connection removed should be called only once", 1, count.get());
+        assertEquals("connection removed should be called only once", 1, listener.count.get());
     }
+
+    private class CountingConnectionRemoveListener implements ConnectionListener {
+
+        final AtomicInteger count = new AtomicInteger();
+
+        @Override
+        public void connectionAdded(Connection connection) {
+
+        }
+
+        @Override
+        public void connectionRemoved(Connection connection) {
+            count.incrementAndGet();
+        }
+    }
+
 }
--- a/hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
@@ -18,9 +18,15 @@ package com.hazelcast.client;
 
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.client.config.ClientProperties;
+import com.hazelcast.client.connection.ClientConnectionManager;
+import com.hazelcast.client.impl.ClientTestUtil;
+import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.core.Client;
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.nio.Address;
+import com.hazelcast.nio.Connection;
+import com.hazelcast.nio.ConnectionListener;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.annotation.QuickTest;
@@ -32,6 +38,7 @@ import org.junit.runner.RunWith;
 
 import java.net.InetSocketAddress;
 import java.util.Collection;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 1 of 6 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/impl/ClientTestUtil.java
```diff
@@ -0,0 +1,19 @@
+package com.hazelcast.client.impl;
+
+import com.hazelcast.core.HazelcastInstance;
+import org.junit.Ignore;
+
+@Ignore("not a JUnit test")
```
jerrinot: this will litter test statistics - it will appear as an "ignored test"

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
```diff
@@ -98,4 +105,35 @@ public void testMemberConnectionOrder() {
...
+    public void testConnectionRemovedOnce() {
+        HazelcastInstance server = Hazelcast.newHazelcastInstance();
+        HazelcastInstance client = HazelcastClient.newHazelcastClient();
+        HazelcastClientInstanceImpl clientImpl = ClientTestUtil.getHazelcastClientInstanceImpl(client);
+        ClientConnectionManager connectionManager = clientImpl.getConnectionManager();
+
+
+        final AtomicInteger count = new AtomicInteger();
+
+        connectionManager.addConnectionListener(new ConnectionListener() {
```
jerrinot: I'd probably create a non-anonymous class "CountingConnectionListener" instead. It'd simplify the test.

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
```diff
@@ -98,4 +105,35 @@ public void testMemberConnectionOrder() {
...
+            }
+
+            @Override
+            public void connectionRemoved(Connection connection) {
+                count.incrementAndGet();
+            }
+        });
+
+        final Address serverAddress = new Address(server.getCluster().getLocalMember().getSocketAddress());
+        final Connection connectionToServer = connectionManager.getConnection(serverAddress);
```
jerrinot:
```
getClientConnectionToServer(HazelcastInstance client, HazelcastInstance server);
```
would again make the test-logic easier to follow.

hazelcast-client/src/test/java/com/hazelcast/client/impl/ClientTestUtil.java
sancar: I copied it from our TestUtil class. If it is not necessary I can remove it. But I thought that there is a reason to add to TestUtil and I did the same here.

jerrinot: I assume the classname matches a convention for tests -> Surefire wanted to execute it -> it fails because JUnit compained "No Test Found".

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
sancar: I really don't like creating a method if it is not used more than once especially in the unit tests. I want to see what our tests are doing in one function. I dont want to skip method to method to understand what this test is doing. Of course there are exceptions like making a method to eliminate code duplication etc.

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
sancar: I dont want to change this also because of similar reasons with my other comment. If it is not used more than once, creating another method and class generally leads to less readable code  for me.

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
jerrinot: I understand your point and I agree with most of it! 

My goal is to make the test easier to read. Currently there are two lines, one of them is quite long and not exactly easy to read. Neither line is important in order to follow a logic of the test. It can be treated as "helper" or "utility" code or perhaps even as a "noise". 

I believe factoring this code out into its own method will make the test more concise - the test itself just requires a connection, it's not important how the connection is obtained. 

Anyway, it's not terribly important, it's just 2 lines. We could merge it as it is, but I believe it would make the code better.

jerrinot: a point in case:
this seems more way easier to read:
```java
    @Test
    public void testConnectionRemovedOnce() {
        HazelcastInstance server = Hazelcast.newHazelcastInstance();

        CountingConnectionListener listener = registerCountingConnectionListener(client);

        Connection connection = getConnectionToServer(client, server);
        connectionManager.destroyConnection(connectionToServer);
        connectionManager.destroyConnection(connectionToServer);

        assertNoOfConnectionsRemoved("connection removed should be called only once", listener, 1);
    }
```

I can see what is it doing in a few seconds.

hazelcast-client/src/test/java/com/hazelcast/client/impl/ClientTestUtil.java
sancar: I removing it from both TestUtils then.

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
sancar: But I dont know how we get a connection to server. May be we got the wrong connection :).
 
Anyway, I tried to modify it as you suggested already . Since I need to return connectionManager from getConnectionToServer to server , I could not do it. And moving destroyConnection into getConnectionToServer method does not  make any sense, because I literally dont see what test is doing. 
In short, your sample code is not compiling. :).

sancar: I am adding the CountingConnectionListener by the way. (as CountingConnectionRemoveListener. because the reader can misunderstand since code is elsewhere, I need to define what it clearly doing in the name)

jerrinot: oh yeah, I can't write valid Java code outside of comfort of my IDE anymore. shame on me! :disappointed: 

anyway, I think my point is clear, I won't block merging this so if another reviewer thinks it's good, then I can live with it.

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
```diff
@@ -98,4 +105,40 @@ public void testMemberConnectionOrder() {
         Collection<Client> connectedClients2 = server2.getClientService().getConnectedClients();
         assertEquals(connectedClients2.size(), 0);
     }
+
+    @Test
+    public void testConnectionRemovedOnce() {
```
pveentjer: Can you provide a more descriptive name? The remove is meant for the listener, but it isn't obvious from the test name. testDestroyConnection_connectionRemovedCalledOnce

sancar: done

jerrinot:
:+1:

PR #4553, opened 2015-02-06
Added the InternalPartition.isLocal method

pveentjer:
Instead of doing a thisAddress.equals(internalPartition.getOwnerOrNull). But under the hood the same functionality is used.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/InternalPartition.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/InternalPartition.java
@@ -33,6 +33,18 @@ public interface InternalPartition {
     int MAX_BACKUP_COUNT = MAX_REPLICA_COUNT - 1;
 
     /**
+     * Checks if the partition is local.
+     *
+     * A partition is local if and only if the {@link #getOwnerOrNull()} returns the same address as 'this' address of the
+     * {@link com.hazelcast.cluster.ClusterService#getThisAddress()}. If the address is null or a different address, false
+     * is returned.
+     *
+     * @return true of local, false otherwise.
+     * @since 3.5
+     */
+    boolean isLocal();
+
+    /**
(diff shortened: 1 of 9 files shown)


hazelcast/src/main/java/com/hazelcast/partition/InternalPartition.java
```diff
@@ -33,6 +33,18 @@
     int MAX_BACKUP_COUNT = MAX_REPLICA_COUNT - 1;
 
     /**
+     * Checks if the partition is local.
+     *
+     * A partition is local if and only if the {@link #getOwnerOrNull()} returns the same address as 'this' address of the
+     * {@link com.hazelcast.cluster.ClusterService#getThisAddress()}. If the address is null or a different address, false
+     * is returned.
+     *
+     * @return true of local, false otherwise.
```
jerrinot: typo. true IF local

jerrinot:
there is a typo, otherwise :+1:

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionImpl.java
```diff
@@ -6,7 +6,7 @@
 import java.util.Arrays;
 
 class InternalPartitionImpl implements InternalPartition {
-
+;
```
Donnerbart: This semicolon feels lonely :)

Donnerbart:
Nice little change, another removal of a `NodeEngine` dependency in the code :)

pveentjer:
I'm working currently on the caller runs scheduler (renamed to progressive scheduler). And I need to know which partitions are locally available so the thiefs know which partitions they can rob if they have nothing better to do. So I ran into the lack of the isLocal method

Donnerbart:
verify

jerrinot:
:+1:

PR #4554, opened 2015-02-07
fix JCache Documet example

kazuhira-r:
t is about the size-policy attributes that are written in XML example of JCache but, "ENTRY-COUNT" is wrong, it is "ENTRY_COUNT" as is correct.


serkan-ozal:
@kazuhira-r good catch. Thanks for your fix

noctarius:
Oh you found some more places :) Thx

gurbuzali:
although we have already merged your changes, we need you to sign our contributor-agreement

PR #4555, opened 2015-02-07
fix wrong class name, IList => ISet

kazuhira-r:
In the example of KeyValueSource of MapReduce, where it should be ISet has become the IList.


Donnerbart:
Thank you for that catch and the PR :)

PR #4558, opened 2015-02-09
Resolved findbugs issues in BasicOperationScheduler

pveentjer:
And also got rid (as... part of the fixes) of the ThreadFactories since they don't provide any value anymore.


gurbuzali:
checkstyle - Unused import

pveentjer:
verify

PR #4559, opened 2015-02-09
Fixed so that when client get an exception in authentication or connect ...

sancar:
..., closes the connection.


gurbuzali:
:+1:

PR #4566, opened 2015-02-10
made runnable partition-aware

gurbuzali:
After we've removed PartitionAwareOperation marker, executor service operations are now running in partition-threads. The test was assuming the operations are running on caller threads, fixed it with adding `PartitionAware` interface to the runnables.

Also fixed that runnables which are implementing `PartitionAware` interface were ignored.


gurbuzali:
verify

mdogan:
Personally, I don't like the idea of submitting executor operations to the partition threads (which was not the case before `PartitionAwareOperation` change), this fix is OK for the current situation.

gurbuzali:
please see https://github.com/hazelcast/hazelcast/issues/4573

PR #4569, opened 2015-02-10
if any exception thrown while adding/removing interceptors from clients,...

eminn:
... exceptions are delegated to the client


hazelcast/src/main/java/com/hazelcast/map/impl/client/MapAddInterceptorRequest.java
```diff
@@ -76,6 +76,11 @@ protected OperationFactory createOperationFactory() {
 
     @Override
     protected Object reduce(Map<Address, Object> map) {
+        for (Object result : map.values()) {
+            if (result instanceof Exception) {
```
sancar: It would be better if we check Throwable rather than Exception.

gurbuzali: :+1:

sancar:
:+1:

Issue #4574, opened 2015-02-10
NPE While Getting CacheOperationProvider


serkan-ozal:
Occured while running stabilizer tests on X-Large cluster.

The problem is that `CacheConfig` may not be created yet before any cache operation request is received from node. Since `CacheConfig` is created on demand at first time needed even if it is configured from XML then all members are invoked to use this `CacheConfig`. But before any member take and create this `CacheConfig`, a cache request may access here.

Here are the error logs of failed test from @Danny-Hazelcast.

```
message='Worked ran into an unhandled exception'
   type='Worker exception'
   agentAddress=10.45.5.206
   time=Mon Feb 09 16:54:05 UTC 2015
   workerAddress=client:10.45.5.206
   workerId=worker-10.45.5.206-4-client
   test=TestCase{
          id=icacheMaxMediume
        , class=com.hazelcast.stabilizer.tests.icache.EvictionICacheTest
        , basename=maxCachMediume1
    }
   cause=java.lang.NullPointerException
    at com.hazelcast.cache.impl.client.AbstractCacheAllPartitionsRequest.getOperationProvider(AbstractCacheAllPartitionsRequest.java:30)
    at com.hazelcast.cache.impl.client.CacheSizeRequest.createOperationFactory(CacheSizeRequest.java:60)
    at com.hazelcast.client.impl.client.AllPartitionsClientRequest.process(AllPartitionsClientRequest.java:33)
    at com.hazelcast.client.impl.ClientEngineImpl$ClientPacketProcessor.processRequest(ClientEngineImpl.java:434)
    at com.hazelcast.client.impl.ClientEngineImpl$ClientPacketProcessor.run(ClientEngineImpl.java:353)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:745)
    at com.hazelcast.util.executor.HazelcastManagedThread.executeRun(HazelcastManagedThread.java:76)
    at com.hazelcast.util.executor.HazelcastManagedThread.run(HazelcastManagedThread.java:92)
    at ------ End remote and begin local stack-trace ------.(Unknown Source)
    at com.hazelcast.client.spi.impl.ClientCallFuture.resolveResponse(ClientCallFuture.java:201)
[...]

PR #4577, opened 2015-02-10
If cache config is not exist, throw a retryable exception so that reques...

serkan-ozal:
Fix for issue #4574  
If cache config is not exist, throw a retryable exception so that request can be retried

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/client/AbstractCacheAllPartitionsRequest.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/client/AbstractCacheAllPartitionsRequest.java
@@ -7,6 +7,7 @@ import com.hazelcast.client.impl.client.AllPartitionsClientRequest;
 import com.hazelcast.config.CacheConfig;
 import com.hazelcast.nio.serialization.PortableReader;
 import com.hazelcast.nio.serialization.PortableWriter;
+import com.hazelcast.spi.exception.RetryableHazelcastException;
 
 import java.io.IOException;
 
@@ -27,6 +28,10 @@ abstract class AbstractCacheAllPartitionsRequest extends AllPartitionsClientRequ
     CacheOperationProvider getOperationProvider() {
         ICacheService service = getService();
         CacheConfig cacheConfig = service.getCacheConfig(name);
+        if (cacheConfig == null) {
+            throw new RetryableHazelcastException(
+                    "Cache config for cache " + name + " has not been created yet !");
+        }
         return service.getCacheOperationProvider(name, cacheConfig.getInMemoryFormat());
     }
 


serkan-ozal:
let me run TCK tests also

jerrinot:
do you know a sequence of actions to create such scenario?

serkan-ozal:
Hi @jerrinot,

I will try explain my understanding and I hope I am not missing anything or don't know Hazelcastcast internal incorrect :)

* Each node loads `CacheConfigSimple` (not created `CacheConfig` yet) from XML

* Client requests for cache by `cacheManager.getCache()`

* Client sends `CacheGetConfigRequest` to a member node (Node-X which is partition owner of cache name with prefix).

* `Node-X` gets `CacheGetConfigRequest` and sends `CacheGetConfigOperation` to a node `Node-Y which is the owner of partition for cache name. BTW possibly, `Node-X` and `Node-Y` should the same node since keys of request and operation are same (cache name with prefix)

* `Node-Y` takes `CacheGetConfigOperation` and since `CacheConfig` is not created yet (just  `CacheConfigSimple` created) creates a cache config on its local and sends `CacheCreateConfigOperation` to all other nodes to create cacheconfig on their locals but doesn't wait response from them and immediately returns `CacheConfig` to caller node `Node-X`. Then `Node-X` also returns `CacheConfig` to caller client.

* `cache.getSize()` is called and `CacheSizeRequest` is sent to a random member (`Node-Z`) since its partition id is not set. Then this member sends `CacheSizeOperation` to all members.

* So at any members side, `CacheSizeOperation` may be processed before `CacheCreateConfigOperation`. Since cache config is not created yet, NPE is thrown

BTW, I don't know the exact case how this can happen that `CacheSizeOperation` processed before `CacheCreateConfigOperation`. But I think, since `CacheGetConfigOperation` and `CacheSizeOperation` are sent from different nodes (`Node-X/Y` and `Node-Z`), there is no guarantee at any receiver side that `CacheGetConfigOperation` is processed before `CacheSizeOperation`.

mdogan:
I already discussed this fix with Serkan, 👍. 

Serkan, apart from AbstractCacheAllPartitionsRequest, can you check if there is  any other base request/operation requires the same check.

jerrinot:
great job, detective Serkan! :feelsgood: :fireworks:

serkan-ozal:
@mdogan I had looked and didn't see another request requires cache config but will see one more time to be sure. 

BTW @mdogan @jerrinot could you also see also my explanation about flow of issue and is there any other missing/wrong point ? If there there is I should update myselft about Hazelcast internals :)

Danny-Hazelcast:
so can we get this merged into 3.4.1

mdogan:
This PR is for master branch. @serkan-ozal, can you send PRs for maintenance and 3.4.1 branches too.

serkan-ozal:
@mdogan ok

jerrinot:
@serkan-ozal: the explanation makes sense to me. assuming that:
- either the error happened at the very begin of the test
- or the test is dynamically creating / destroying caches

Is it this correct?

serkan-ozal:
@jerrinot 

As we talked with @Danny-Hazelcast, issue happened 4 minutes later after cluster started.

And also AFAIK (an remember from the actual test code) there is no dynamic cache create/destroy operation.

jerrinot:
4 minutes is quite a long time for such race condition, isn't it?

serkan-ozal:
maybe, but I really wondered the time between first `getCache()` and first `size()` calls.

noctarius:
Wasn't the problem that if someone else destroys the cache this retry would run forever?

serkan-ozal:
good point :), let me look in detail and reproduce case

serkan-ozal:
I have reproduced the case as some nodes takes `CacheCreateConfigOperation` operation after `CacheDestroyOperation` is executed. In this case, operation itself is retried as `retryCount` in `com.hazelcast.client.spi.impl.ClientInvocation`(its default value is `120`) and finished retrying.

noctarius:
This almost feels like infinite if you are waiting for it :D

serkan-ozal:
depends on how important what you wait

noctarius:
If you call `createCacheManager` and you wait for (what is the expected runtime?) :D Even worst you call `getCacheManager` and you won't expect it to block for a long time (the internal create is an implementation detail)

serkan-ozal:
@noctarius right, the problem is that it is hard to understand if the cache config is not created yet, or will not be created (since there is no cache defined), or it is already destroyed. I accept that retrying operation is not good for destroyed/not existing cache. Maybe we should make cache config create operation sync not async

jerrinot:
+1 for sync cache creation

serkan-ozal:
or at least, currently maybe we can make this retry operation and its retry count configurable for jcache (if we won't make it sync). and by default they can be disabled

noctarius:
Well the other option was the workaround we discussed end of last year, keeping track of removed caches for a short time period (like 1 or 2 minutes) but it means additional cleanup logic, etc... still not in favor of this approach ;)

serkan-ozal:
@noctarius that was also one of my fix candidates but I really didn't like it as you said :)

PR #4581, opened 2015-02-11
xml client config-import implementation

bilalyasar:
this is implementation of: hazelcast-client.xml variable Replacement and config import
you can find prd and technical design document in hz-wiki.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
@@ -18,6 +18,7 @@ package com.hazelcast.client.config;
 
 import com.hazelcast.client.util.RandomLB;
 import com.hazelcast.client.util.RoundRobinLB;
+import com.hazelcast.config.AbstractConfigBuilder;
 import com.hazelcast.config.ConfigLoader;
 import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.config.ListenerConfig;
@@ -25,7 +26,6 @@ import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.config.SSLConfig;
 import com.hazelcast.config.SerializationConfig;
 import com.hazelcast.config.SocketInterceptorConfig;
-import com.hazelcast.config.XmlConfigPreProcessor;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 import com.hazelcast.nio.IOUtil;
@@ -47,7 +47,7 @@ import java.util.Properties;
 /**
  * Loads the {@link com.hazelcast.client.config.ClientConfig} using XML.
  */
-public class XmlClientConfigBuilder extends XmlConfigPreProcessor {
+public class XmlClientConfigBuilder extends AbstractConfigBuilder {
 
     private static final int DEFAULT_VALUE = 5;
     private static final ILogger LOGGER = Logger.getLogger(XmlClientConfigBuilder.class);
@@ -103,9 +103,8 @@ public class XmlClientConfigBuilder extends XmlConfigPreProcessor {
     protected Document parse(InputStream inputStream) throws Exception {
 
         final DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
-        Document doc;
         try {
-            doc = builder.parse(inputStream);
+            return builder.parse(inputStream);
         } catch (final Exception e) {
             String msg = "Failed to parse Config Stream"
                     + "\nException: " + e.getMessage()
@@ -115,7 +114,6 @@ public class XmlClientConfigBuilder extends XmlConfigPreProcessor {
         } finally {
             IOUtil.closeResource(inputStream);
         }
-        return doc;
     }
 
     @Override
--- a/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
@@ -18,7 +18,6 @@ package com.hazelcast.client.config;
 
 import com.hazelcast.client.util.RandomLB;
 import com.hazelcast.client.util.RoundRobinLB;
-import com.hazelcast.config.AbstractXmlConfigHelper;
 import com.hazelcast.config.ConfigLoader;
 import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.config.ListenerConfig;
@@ -26,6 +25,9 @@ import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.config.SSLConfig;
 import com.hazelcast.config.SerializationConfig;
 import com.hazelcast.config.SocketInterceptorConfig;
+import com.hazelcast.config.XmlConfigPreProcessor;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.Logger;
 import com.hazelcast.nio.IOUtil;
 import com.hazelcast.util.ExceptionUtil;
 import org.w3c.dom.Document;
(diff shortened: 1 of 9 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -40,13 +42,18 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.util.Properties;
 
 /**
  * Loads the {@link com.hazelcast.client.config.ClientConfig} using XML.
  */
-public class XmlClientConfigBuilder extends AbstractXmlConfigHelper {
+public class XmlClientConfigBuilder extends XmlConfigPreProcessor {
```
pveentjer: Don't extend if you don't need to.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -92,6 +99,39 @@ public XmlClientConfigBuilder() {
...
+                    + "\nException: " + e.getMessage()
+                    + "\nHazelcastClient startup interrupted.";
+            LOGGER.severe(msg);
+            throw e;
+        } finally {
+            IOUtil.closeResource(inputStream);
+        }
+        return doc;
+    }
+
```
pveentjer: No need to create a doc variable, assign to the doc variable and then return it.

Just do 'return builder.parse(inputStream)'. 

Saves you 2 lines of useless code :)

hazelcast-client/src/test/java/com/hazelcast/client/config/XmlClientConfigPreProcessorTest.java
```diff
@@ -0,0 +1,257 @@
...
+                "</socket-interceptor>" +
+                "        </network>" +
+                "</hazelcast-client>";
+        writeStringToStreamAndClose(os, networkConfig);
+
+        String xml = "<hazelcast-client>\n" +
+                "    <import resource=\"${config.location}\"/>\n" +
+                "</hazelcast-client>";
+
+        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
```
pveentjer: Perhaps pull out a method like you did testing the config?

```
 private Config buildConfig(String xml) {
        ByteArrayInputStream bis = new ByteArrayInputStream(xml.getBytes());
        XmlConfigBuilder configBuilder = new XmlConfigBuilder(bis);
        return configBuilder.build();
    }
```

When you are copy pasting code; then your spider senses should start tingling.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
mesutcelik: you might change the name to **AbstractConfigBuilder**

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
mesutcelik: +1

hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
```diff
@@ -193,11 +197,11 @@ private void parseAndBuildConfig(final Config config) throws Exception {
         } catch (final Throwable e) {
             domLevel3 = false;
         }
-        xmlConfigPreProcessor.process(root);
+        process(root);
         handleConfig(root);
     }
 
-    Document parse(InputStream is) throws Exception {
+    protected Document parse(InputStream is) throws Exception {
```
mesutcelik: you could put Override annotation here...

hazelcast/src/main/java/com/hazelcast/config/XmlConfigPreProcessor.java
```diff
@@ -97,11 +108,18 @@ private void loadAndReplaceImportElement(Node root, Node node) throws Exception
         root.removeChild(node);
     }
 
+    protected abstract Document parse(InputStream inputStream) throws Exception;
+
+    protected abstract Properties getProperties();
+
+    protected abstract ConfigType getXmlType();
+
```
mesutcelik: It would be useful to add javadoc for abstract methods. These would help implementor understand the purpose of those methods...

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
Donnerbart: Just if you did this because of the `finally`: A `finally` block will be executed even if you return within the `try-catch`-block (you can even overwrite the return value in a `finally` block, which is kinda creepy). So don't hesitate to return from a `try-catch`, `finally` always wins.

bilalyasar:
@mesutcelik @pveentjer @Donnerbart , thanks for your valuable comments, updated.

mesutcelik:
verify

mesutcelik:
verify

mesutcelik:
looks ok to me.

eminn:
Looks good to merge.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -92,6 +99,37 @@ public XmlClientConfigBuilder() {
...
 
+    @Override
+    protected Document parse(InputStream inputStream) throws Exception {
+
+        final DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+        try {
+            return builder.parse(inputStream);
+        } catch (final Exception e) {
+            String msg = "Failed to parse Config Stream"
+                    + "\nException: " + e.getMessage()
```
jerrinot: this will produce funky output on windows. you should use environment neutral approach like
```java
System.getProperty("line.separator");
```
perhaps we already have a util method for this?

bilalyasar: @jerrinot , good catch, updated.. thanks

pveentjer:
verify

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -92,6 +100,38 @@ public XmlClientConfigBuilder() {
...
+            IOUtil.closeResource(inputStream);
+        }
+    }
+
+    @Override
+    protected Properties getProperties() {
+        return properties;
+    }
+
+    public void setProperties(Properties properties) {
```
pveentjer: I don't know if this exposed to the end user; it if is it needs documentation + a null check. No need to let a user insert a null properties.

hazelcast-client/src/test/java/com/hazelcast/client/config/XmlClientConfigImportVariableReplacementTest.java
```diff
@@ -0,0 +1,247 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+@RunWith(HazelcastSerialClassRunner.class)
+@Category(QuickTest.class)
+public class XmlClientConfigImportVariableReplacementTest {
```
pveentjer: It i a joy reading such a test.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
mesutcelik: I don't think we need setProperties here.

bilalyasar: if user wants to pass properties object other than setting system properties.
like here: https://github.com/hazelcast/hazelcast/pull/4581/files#diff-eaa5dbc388de7caa71a251009b4e8d6cR232
In server side we support this.

mesutcelik:
I see no more open comment so i am merging that one.

PR #4504, opened 2015-01-29
Added type checking unittests to meet the Javadocs exception requirement...

noctarius:
...s on javax.cache.Cache interface, added more unittests in general, especially concurrency

@gurbuzali:
New PR for #4032

PR #4586, opened 2015-02-11
Added unittests for typechecks according to javax.cache.Cache and the up...

noctarius:
...dated ICache Javadocs - related #4504 #4583

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
@@ -25,6 +25,7 @@ import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.nio.serialization.SerializationService;
+import com.hazelcast.spi.OperationResultVerifier;
 import com.hazelcast.spi.exception.TargetDisconnectedException;
 import com.hazelcast.util.Clock;
 import com.hazelcast.util.ExceptionUtil;
@@ -58,6 +59,7 @@ public class ClientInvocationFuture<V> implements ICompletableFuture<V> {
 
     private volatile Object response;
 
+    private volatile OperationResultVerifier<V> operationResultVerifier;
 
     public ClientInvocationFuture(ClientInvocation invocation, HazelcastClientInstanceImpl client,
                                   ClientRequest request, EventHandler handler) {
@@ -164,7 +166,20 @@ public class ClientInvocationFuture<V> implements ICompletableFuture<V> {
         if (response == null) {
             throw new TimeoutException();
         }
-        return (V) response;
+        return (V) verifyResponse(response);
+    }
+
+    private Object verifyResponse(Object value) {
+        if (value instanceof Data) {
+            return value;
+        }
+        if (operationResultVerifier != null) {
+            value = operationResultVerifier.verify((V) value);
+        }
+        if (value instanceof Throwable) {
+            ExceptionUtil.sneakyThrow((Throwable) value);
+        }
+        return value;
     }
 
     @Override
@@ -192,6 +207,10 @@ public class ClientInvocationFuture<V> implements ICompletableFuture<V> {
             }
             callbackNodeList.add(new ExecutionCallbackNode(callback, executor, false));
         }
+    }
+
+    public void setOperationResultVerifier(OperationResultVerifier<V> operationResultVerifier) {
+        this.operationResultVerifier = operationResultVerifier;
     }
 
     public ClientRequest getRequest() {
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
@@ -176,9 +176,6 @@ public class ClientInvocationFuture<V> implements ICompletableFuture<V> {
         if (operationResultVerifier != null) {
             value = operationResultVerifier.verify((V) value);
         }
-        if (value instanceof Throwable) {
-            ExceptionUtil.sneakyThrow((Throwable) value);
-        }
         return value;
     }
 
--- a/hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientJCacheTypeChecks.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientJCacheTypeChecks.java
@@ -1,0 +1,373 @@
+/*
+ * Copyright (c) 2008-2013, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.cache;
+
+import com.hazelcast.cache.ICache;
+import com.hazelcast.cache.impl.HazelcastServerCachingProvider;
+import com.hazelcast.client.HazelcastClient;
+import com.hazelcast.client.cache.impl.HazelcastClientCachingProvider;
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.config.CacheConfig;
+import com.hazelcast.config.Config;
+import com.hazelcast.config.JoinConfig;
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.ICompletableFuture;
+import com.hazelcast.test.AssertTask;
+import com.hazelcast.test.HazelcastSerialClassRunner;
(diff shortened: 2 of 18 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientJCacheTypeChecks.java
```diff
@@ -0,0 +1,373 @@
...
+        CompleteConfiguration<Integer, String> config = new CacheConfig<Integer, String>();
+
+        Cache<Integer, String> cache = cacheManager.createCache(cacheName, config);
+        assertNotNull(cache);
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                CacheManager cm1 = cachingProvider1.getCacheManager();
+                assertNotNull(cm1.getCache(cacheName));
```
bilalyasar: there are many duplications, especially almost every test has this part.
is it possible make shorter it? like making inner class and creating new one every time. or some another solution.

hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientJCacheTypeChecks.java
```diff
@@ -0,0 +1,373 @@
...
+
+        cache.get(1);
+    }
+
+    @Test(expected = ClassCastException.class)
+    public void test_check_types_on_getandremove() throws Exception {
+        final String cacheName = randomMapName();
+
+        CacheManager cacheManager = cachingProvider3.getCacheManager();
+        assertNotNull(cacheManager);
```
bilalyasar: why do you check cacheManager in every test? As far as i learned from @pveentjer , if you don't sure about cacheManager you must create another test for it.

hazelcast/src/test/java/com/hazelcast/cache/JCacheTypeChecks.java
```diff
@@ -0,0 +1,372 @@
...
+        });
+
+        Cache rawCache = (Cache) cache;
+        rawCache.put(1, 1);
+
+        Class<?> clazz = Class.forName("com.hazelcast.cache.impl.AbstractCacheProxyBase");
+        Field configField = clazz.getDeclaredField("cacheConfig");
+        configField.setAccessible(true);
+
+        CacheConfig cacheConfig = (CacheConfig) configField.get(rawCache);
```
bilalyasar: here there are some duplications, it could be good if you write a method like `buildConfig`with some parameters, so readability can be increased. and every test can call that method.

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicInvocationFuture.java
```diff
@@ -8,6 +8,7 @@
 import com.hazelcast.spi.Callback;
 import com.hazelcast.spi.InternalCompletableFuture;
 import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.OperationResultVerifier;
```
bilalyasar: it is not related this pr but you can add headers :)

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
```diff
@@ -164,7 +166,20 @@ private V resolveResponse() throws ExecutionException, TimeoutException, Interru
...
-        return (V) response;
+        return (V) verifyResponse(response);
+    }
+
+    private Object verifyResponse(Object value) {
+        if (value instanceof Data) {
+            return value;
+        }
+        if (operationResultVerifier != null) {
+            value = operationResultVerifier.verify((V) value);
```
serkan-ozal: I think, returning exception instead of throwing it is not a good idea and then checking returned object type. Why result verifier doesn't throw exception ?

hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientJCacheTypeChecks.java
```diff
@@ -0,0 +1,373 @@
...
+import javax.cache.CacheManager;
+import javax.cache.configuration.CompleteConfiguration;
+import java.lang.reflect.Field;
+
+import static junit.framework.Assert.assertNotNull;
+import static junit.framework.Assert.assertNull;
+
+@RunWith(HazelcastSerialClassRunner.class)
+@Category(QuickTest.class)
+public class ClientJCacheTypeChecks
```
serkan-ozal: I think, name of the test should be `ClientJCacheTypeCheckTest`

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
noctarius: For other situations you might want to do something else instead of just rethrowing it. In case of JCache this is correct but probably not for others :)

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxy.java
```diff
@@ -153,7 +154,7 @@ protected AbstractCacheProxy(CacheConfig cacheConfig, NodeEngine nodeEngine, ICa
 
     @Override
     public V get(K key, ExpiryPolicy expiryPolicy) {
-        final Future<V> f = getAsync(key, expiryPolicy);
+        final InternalCompletableFuture<V> f = getAsync(key, expiryPolicy);
```
serkan-ozal: why we use `InternalCompletableFuture` explicitly ?

serkan-ozal: why we use `InternalCompletableFuture` explicitly ?

noctarius: Because this interface has the InternalCompletableFuture::setOperationResultVerifier

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheProxyUtil.java
```diff
@@ -240,7 +254,7 @@ protected static int getPartitionId(NodeEngine nodeEngine, Data key) {
      */
     public static <V> void validateConfiguredValueType(Class<V> valueType, V value)
             throws ClassCastException {
-        if (Object.class != valueType) {
+        if (value != null && valueType != null && Object.class != valueType) {
             //means type checks required
             if (!valueType.isAssignableFrom(value.getClass())) {
                 throw new ClassCastException("Value " + value + "is not assignable to " + valueType);
```
serkan-ozal: I think, type of value should be logged to give more details since `int value  = 1` and `String value = "1"` are same output `1` but their types are different

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
serkan-ozal: I didn't understand why this is true for this case for jcache. Can you explain ?

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxy.java
serkan-ozal: But, in here we just use `f.get()`. So there is no need to explicit `InternalCompletableFuture`.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
noctarius: Read the javax.cache.Cache Javadoc. Throws ClassCastException if value does not match configured type or what do you mean?

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxy.java
noctarius: Oh I see, probably a left over from an intermediate implementation.

hazelcast/src/test/java/com/hazelcast/cache/JCacheTypeChecks.java
serkan-ozal: so many duplicate codes between tests to create config or test something

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
serkan-ozal: I already read the `javax.cache.Cache Javadoc`. I mean, why you just directly don't throw `ClassCastException` while verifying ? Why you return exception as return value then check if it `Throwable` then throw ?

noctarius: Because this OperationResultVerifier is a general addition and nothing JCache specific. Therefore it found it more useful to not throw exceptions directly but yeah maybe we should do it.

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicInvocationFuture.java
noctarius: good catch :) haven't recognized they're missing

hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientJCacheTypeChecks.java
noctarius: oops :D

noctarius:
veify

noctarius:
verify

pveentjer:
@noctarius @serkan-ozal what should we do with this task.

pveentjer:
I'm going to close this one due to inactivity.

PR #4587, opened 2015-02-11
Added the DebugUtil with some methods for debugging

pveentjer:
I added this DebugUtil because often I need to see what happened (e.g. whole called a certain method) in runtime, but can't rely on a debugger.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/DebugUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/DebugUtil.java
@@ -1,0 +1,86 @@
+package com.hazelcast.util;
+
+import com.hazelcast.nio.IOUtil;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+/**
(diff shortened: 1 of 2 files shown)


logs/provisioner.log
```diff
@@ -0,0 +1,11 @@
+INFO  2015-02-11 13:12:05,583 [main] com.hazelcast.stabilizer.provisioner.Provisioner: Hazelcast Stabilizer Provisioner
```
eminn: I think this file is a leftover. Isn't it ?

eminn:
Also there is an empty agents.txt file which is also is not related I think.

logs/provisioner.log
pveentjer: Jup. Thanks for spitting it.

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/util/DebugUtil.java
```diff
@@ -0,0 +1,86 @@
+package com.hazelcast.util;
```
bilalyasar: missing headers

pveentjer:
verify

pveentjer:
Since it only adds code for debugging and doesn't change any existing classes, I'm going to merge this one.

PR #4588, opened 2015-02-11
changed txMap to hold key as Data rather than Object, fixed size calculation

### LINKED ISSUE #4562: TransactionalMap.size() returns inconsistent results during a transaction
Hazelcast version: 3.4
Node type: in-cluster
Configuration: a basic unit-test setup

The problem is probably best described by the following JUnit example:
    
    @Test
    public void transactionalMapCanHaveNegativeSize() throws Exception {

        final TransactionContext transactionContext =
            hazelcastInstance.newTransactionContext(new TransactionOptions().setTransactionType(TransactionOptions.TransactionType.LOCAL));

        transactionContext.beginTransaction();

        final TransactionalMap<String, String> map = transactionContext.getMap("test-map");

        map.put("k", "v");
        map.remove("k");

        assertEquals(0, map.size()); // failure: expected 0 got -1
    }

I think the problem is in `TransactionalMap#remove(Object)`, where it overwrites the "insert" operation in its internal transaction map with a "remove" - so you end up with a remove for an object that never existed. Then when it comes to decide size, it counts the removal with negative size.

I think the right thing to do here is to remove the insertion `TxValueWrapper` from `txMap`.

I'm happy to write up a PR if the above is agreed - the reason I haven't already is that I'm surprised this hasn't been stumbled on before, so I'm wondering if there's some reason it would be intentional.

The work-around we're using is to get the size of the keyset (which returns the expected result).

gurbuzali:
fixes #4562

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
@@ -43,7 +43,7 @@ import java.util.concurrent.TimeUnit;
  */
 public class TransactionalMapProxy extends TransactionalMapProxySupport implements TransactionalMap {
 
-    private final Map<Object, TxnValueWrapper> txMap = new HashMap<Object, TxnValueWrapper>();
+    private final Map<Data, TxnValueWrapper> txMap = new HashMap<Data, TxnValueWrapper>();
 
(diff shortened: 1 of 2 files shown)


gurbuzali:
verify

sancar:
This pr looks good. Although unrelated, I just noticed that we don't have null check for keys and values in methods like put. I think we should have them as we have in IMap.

Issue #4144, opened 2014-11-21
Map replication should mark expirable recordstore 


ahmetmircik:
Otherwise, in some situations, if one does not set map wide expiration or map wide TTL, key based TTL expiration may not work.

Some more details here:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/hazelcast/p6vc9Gt0_sg/LiZ5rzI463IJ

PR #4591, opened 2015-02-11
PutBackupOperation should set TTL before adding enry to recordstore

ahmetmircik:
follow up the fix for https://github.com/hazelcast/hazelcast/issues/4144


jerrinot:
what about other operations? set, putIfAbsent, etc..

enesakar:
looks ok. should be applied to maintenance and master branches

enesakar:
other put operations have same backup operation: PutBackupOperation. Am I right Ahmet?

jerrinot:
ah, that makes sense, thanks!

enesakar:
but I checked it again, PutAllBackupOperation seems need the same fix. Ahmet?

enesakar:
I checked it. PutAllBackupOperation does not need any fix. because we do not support ttl for PutAll. so PR seems ok

jerrinot:
:+1:

PR #4595, opened 2015-02-11
PutBackupOperation should set TTL before adding enry to recordstore

ahmetmircik:
follow up the fix for #4144


gurbuzali:
verify

ahmetmircik:
verify

PR #4596, opened 2015-02-11
Converted jcache config creation to synchronous for callers but still as...

mdogan:
...ync for partition threads.

(cherry picked from commit 69a01ea635ea5c8c857d36f10e0b72fb49e4795f)

This is follow-up PR for https://github.com/hazelcast/hazelcast/pull/4577


serkan-ozal:
@enesakar will be get into 3.4.1 ? I think, we should.

PR #4598, opened 2015-02-11
Converted jcache config creation to synchronous for callers but still as...

mdogan:
...ync for partition threads.

https://github.com/hazelcast/hazelcast/pull/4577


serkan-ozal:
@enesakar @mdogan we should also need this PR merged for 3.4.1 (and maintenance)

serkan-ozal:
This PR makes cache config creation sync from client view. So there is no need to retryable request. :+1:

serkan-ozal:
PR is OK. All unit tests and TCK tests are passed. I will be merge it but let me sure that there is no broken thing on enterprise side.

PR #4602, opened 2015-02-11
OperationHandler cleanup

pveentjer:
Currently we have a single OperationHandler instance of the OperationService.

With this change, each partition will have its own OperationHandler and each generic thread will have its own event handler.

Things like slow operation detection can now be done on the handlers, instead of messing with the operation-threads.

Also future improvements like sampling can be added very easily without being tied to the scheduler

Also the isstillrunning-functionality can be moved out of the scheduler into an external component since we can just access the operation-handlers.

This change is also needed for the caller-runs optimization.


pveentjer:
@mdogan can you have a look? This is the change I talked about this morning

@Donnerbart can you also have a look? This change prevents the need to rely on the thread of the scheduler for the slow operation detector.

hazelcast/src/main/java/com/hazelcast/spi/impl/classicscheduler/ClassicOperationScheduler.java
```diff
@@ -0,0 +1,432 @@
...
+        this.genericOperationHandlers = initGenericOperationHandlers(properties, operationHandlerFactory);
+        this.genericOperationThreads = initGenericThreads();
+
+        this.responseThread = initResponseThread();
+
+        logger.info("Starting with " + genericOperationThreads.length + " generic operation threads and "
+                + partitionOperationThreads.length + " partition operation threads.");
+    }
+
+    private OperationHandler[] initGenericOperationHandlers(GroupProperties properties, OperationHandlerFactory handlerFactory) {
```
Donnerbart: The init methods are in a different order than they are used, so I got a bit confused while reading this on Github (no code navigation/method highlighting). It would be very kind to re-order them.

hazelcast/src/main/java/com/hazelcast/spi/impl/classicscheduler/ClassicOperationScheduler.java
```diff
@@ -0,0 +1,432 @@
...
+        if (currentThread instanceof NIOThread) {
+            return false;
+        }
+
+        //todo: do we want to allow non partition specific tasks to be run on a partitionSpecific operation thread?
+        if (partitionId < 0) {
+            return true;
+        }
+
+        //we are only allowed to execute partition aware actions on an OperationThread.
```
Donnerbart: Doesn't this comment fit better to the previous `if (partitionId < 0)`? The actual if about the thread type, not if the operation is partition aware (and yeah, this is an old issue if it is one).

hazelcast/src/main/java/com/hazelcast/spi/impl/classicscheduler/ClassicOperationScheduler.java
```diff
@@ -0,0 +1,432 @@
...
+
+    @Override
+    public int getRunningOperationCount() {
+        int result = 0;
+        for (PartitionOperationThread thread : partitionOperationThreads) {
+            if (thread.currentOperation != null) {
+                result++;
+            }
+        }
+        for (OperationThread thread : genericOperationThreads) {
```
Donnerbart: I would use `OperationThread` in Line 266, too or `GenericOperationThread` here. The mix confused me, since I thought there was no `GenericOperationThread` at all (from this code snippet).

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
```diff
@@ -583,7 +643,12 @@ private void notifyRemoteCall(Response response) {
     /**
      * Responsible for processing an Operation.
      */
-    private final class OperationHandler implements BasicOperationHandler {
+    private final class BasicOperationHandler implements OperationHandler {
```
Donnerbart: What do you think of ripping this class out of the `BasicOperationService` which is still >1000 lines long? It seems that the `BasicOperationHandler` just uses a few internal members: `node/nodeEngine`, `executedOperationsCount`, `operationBackupHandler` and `backPressureService`. So package private or a constructor shouldn't be that hard to implement.

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
```diff
@@ -583,7 +643,12 @@ private void notifyRemoteCall(Response response) {
...
+
+        @Override
+        public Operation currentOperation() {
+            return null;
```
Donnerbart: // FIXME :D

hazelcast/src/main/java/com/hazelcast/spi/impl/OperationScheduler.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.hazelcast.nio.Packet;
+import com.hazelcast.spi.Operation;
+
+/**
+ * The OperationScheduler is responsible for scheduling operations to be executed locally.
+ */
+public interface OperationScheduler {
+
+    @Deprecated
```
Donnerbart: `@Deprecated` seems to be new and `@PrivateAPI` seems to be gone. First change okay, but shouldn't we keep the `@PrivateAPI` in addition?

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
```diff
@@ -123,7 +124,7 @@
 
     private final long defaultCallTimeoutMillis;
     private final long backupOperationTimeoutMillis;
-    private final OperationHandler operationHandler;
+    private final BasicOperationHandler operationHandler;
     private final OperationBackupHandler operationBackupHandler;
     private final BasicBackPressureService backPressureService;
     private final ResponsePacketHandler responsePacketHandler;
```
Donnerbart: Field can be converted to a local variable.

hazelcast/src/main/java/com/hazelcast/spi/impl/OperationScheduler.java
pveentjer: I don't see any point in @PrivateAPI on impl package; it is all internal so no need to say it is private.

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
pveentjer: Yes. This PR is not ready, but I want to give an indication of the direction we need to move to.

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
pveentjer: We'll rip out more stuff; I'm thinking about a package called basicoperationservice and than moving most of the inner classes as regular class. But not now.. else I won't get it merged.

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
pveentjer: Don't look too much at these details for the time being :) Just see if the API fits your needs and reduces the current thight coupling to the scheduler threads.

Donnerbart:
I like the idea of the new API. Would fit into my 3.5 feature as well, not too much to change (maybe some overlaps with small code cleanups, but that's my problem then).

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
```diff
@@ -680,13 +778,21 @@ private void afterRun(Operation op) {
...
-        private void ensureNoPartitionProblems(Operation op) {
+        protected void ensureNoPartitionProblems(Operation op) {
             int partitionId = op.getPartitionId();
+
             if (partitionId < 0) {
                 return;
             }
 
-            InternalPartition internalPartition = nodeEngine.getPartitionService().getPartition(partitionId);
+            if (partitionId != thisPartitionId) {
```
mdogan: Is this a debug assertion? Operations will be already assigned to the correct operation-handler by the service itself. Or can a mis-assignment happen? Otherwise I think you should set a self-explaining exception message:

```
throw new IllegalStateException("bla bla bla...");
```

hazelcast/src/main/java/com/hazelcast/spi/impl/classicscheduler/PartitionOperationThread.java
```diff
@@ -0,0 +1,30 @@
...
+                                    Queue priorityWorkQueue, ILogger logger,
+                                    HazelcastThreadGroup threadGroup, NodeExtension nodeExtension,
+                                    OperationHandler[] partitionOperationHandlers) {
+        super(name, threadId, workQueue, priorityWorkQueue, logger, threadGroup, nodeExtension);
+        this.partitionOperationHandlers = partitionOperationHandlers;
+    }
+
+    @Override
+    public OperationHandler getOperationHandler(int partitionId) {
+        return partitionOperationHandlers[partitionId];
```
mdogan: I didn't get the logic here. Aren't a `PartitionOperationThread` and a partition `OperationHandler` linked one-to-one? Can `PartitionOperationThread` work with different partition `OperationHandler`s?

pveentjer: Good spitting. That is old code.. Will clean that.

pveentjer: My bad. it is good code.. Each partition will have its own operation-handler. This is needed for the progressive scheduler to allow for workstealing on a partitionlevel.

mdogan: got it. I missed the part operation-handlers created for each partition..

pveentjer:
verify

mdogan:
looks good. :+1:

PR #4604, opened 2015-02-12
Made reading of InMemoryFormat value from XML similar to server nodes - ...

noctarius:
...convert to uppercase first


serkan-ozal:
@noctarius can it be problem for different locales such as TR locale? For example, AFAIK when I made uppercase 'native_memory', possibly, it will be "NATİVE_MEMORY" since Turkish alphabet, "i" => "İ" (not I)

jerrinot:
128 ASCII characters ought to be enough for everyone!

serkan-ozal:
@noctarius Ok, upper case internal already handles this :)
@jerrinot as you know "i" is valid ASCII character and it exist in "native_memory" word. So I don't mention about non-ASCII character

noctarius:
We would have the same problems on server side :D

noctarius:
verify

PR #4621, opened 2015-02-13
changed full scan for set.contains

gurbuzali:
fixes https://github.com/hazelcast/hazelcast/issues/4620

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/collection/CollectionContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/CollectionContainer.java
@@ -114,15 +114,11 @@ public abstract class CollectionContainer implements DataSerializable {
     }
 
     protected boolean contains(Set<Data> valueSet) {
+        Collection<CollectionItem> collection = getCollection();
+        CollectionItem collectionItem = new CollectionItem(-1, null);
         for (Data value : valueSet) {
-            boolean contains = false;
-            for (CollectionItem item : getCollection()) {
-                if (value.equals(item.getValue())) {
-                    contains = true;
-                    break;
-                }
-            }
-            if (!contains) {
+            collectionItem.setValue(value);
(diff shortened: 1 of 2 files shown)


jerrinot:
it relies on the fact CollectionItem::equals doesn't take itemId into consideration. perhaps you should write a test checking this assumption, WDYT?

gurbuzali:
done!

jerrinot:
:+1:

jerrinot:
just a minor note: perhaps and explanation why the test is in place would be good. 
otherwise it might be tempting to just change the test when one changes the equals().

hazelcast/src/main/java/com/hazelcast/collection/CollectionContainer.java
```diff
@@ -114,15 +114,9 @@ protected void clearBackup(Set<Long> itemIdSet) {
...
         for (Data value : valueSet) {
-            boolean contains = false;
-            for (CollectionItem item : getCollection()) {
-                if (value.equals(item.getValue())) {
-                    contains = true;
-                    break;
-                }
-            }
-            if (!contains) {
+            if (!collection.contains(new CollectionItem(-1, value))) {
```
ahmetmircik: In order to prevent the cost caused by extra object creation in every iteration; what about an approach like this:
```
protected boolean contains(Set<Data> valueSet) {
        Collection<CollectionItem> collection = getCollection();
        CollectionItem item = new CollectionItem(-1, null); <--------------------- Create once.
        for (Data value : valueSet) {
            item.setValue(value);<----------------- Only set value
            if (!collection.contains(item)) {
                return false;
            }
        }
        return true;
    }
```

gurbuzali: :+1:

sancar:
Looks good.

PR #4630, opened 2015-02-17
fix exception delegation when client is not configured to use ssl but no...

eminn:
...des. Fixes issue https://github.com/hazelcast/hazelcast-enterprise/issues/128, tests are in the https://github.com/hazelcast/hazelcast-enterprise/pull/130 .This issue only happens in master branch.


sancar:
:+1:

Issue #4114, opened 2014-11-19
[TEST-FAILURE] QueryIndexingTest.testResultsHaveNullFields_whenUsingIndexes


bilalyasar:
```
java.lang.AssertionError: expected:<2500> but was:<2429>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at com.hazelcast.map.query.QueryIndexingTest.testResultsHaveNullFields_whenUsingIndexes(QueryIndexingTest.java:92)
```
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast%20Maintenance/job/Hazelcast-3.maintenance-OpenJDK8/com.hazelcast$hazelcast/59/testReport/junit/com.hazelcast.map.query/QueryIndexingTest/testResultsHaveNullFields_whenUsingIndexes/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.7/com.hazelcast$hazelcast/34/testReport/junit/com.hazelcast.map.query/QueryIndexingTest/testResultsHaveNullFields_whenUsingIndexes/

mdogan:
Failed on master branch: https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast/12093/testReport/junit/com.hazelcast.map.query/QueryIndexingTest/testResultsHaveNullFields_whenUsingIndexes/

```
Error Message

employee.getCity();
Wanted 1 time:
-> at com.hazelcast.map.query.QueryIndexingTest.assertGettersCalledNTimes(QueryIndexingTest.java:108)
But was 2 times. Undesired invocation:
-> at sun.reflect.GeneratedMethodAccessor24.invoke(Unknown Source)
Stacktrace

org.mockito.exceptions.verification.TooManyActualInvocations: 
employee.getCity();
Wanted 1 time:
-> at com.hazelcast.map.query.QueryIndexingTest.assertGettersCalledNTimes(QueryIndexingTest.java:108)
But was 2 times. Undesired invocation:
-> at sun.reflect.GeneratedMethodAccessor24.invoke(Unknown Source)

[...]

PR #4634, opened 2015-02-18
Removed unnecessary serialiser and used serialisable mockito spy instead

ajermakovics:
#4114

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/query/QueryIndexingTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/query/QueryIndexingTest.java
@@ -36,18 +36,28 @@ import org.junit.runner.RunWith;
 public class QueryIndexingTest extends HazelcastTestSupport {
 
     private int count = 2000;
-    private Map<Integer, Employee> employees = newEmployees(count);
-    private Config conf = newConfig(employees);
+    private Map<Integer, Employee> employees;
+    private Config conf;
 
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/map/query/QueryIndexingTest.java
```diff
@@ -134,41 +122,10 @@ private static Config newConfig(final Map<Integer, Employee> employees) {
             } else {
                 val = new Employee(i, "name" + i, "city" + i, 0, true, i);
             }
-            employees.put(i, Mockito.spy(val));
+            Employee spy = MockUtil.serializableSpy(Employee.class, val);
```
pveentjer: Could you schedule a talk about using Mockito on one of our Friday afternoons to educate us a bit? I know the basics and I need to scratch my head with spies. I'm sure that most other developers will be confused as well.

pveentjer: Could you add a few lines of comment about what is happening here? I'm confused.

ajermakovics: It might not be obvious from this diff but if you have a look at the whole file and `assertGettersCalledNTimes` method I think its easier to see what's going on. I'm more used to JMockit than Mockito, learning it myself.

PR #4638, opened 2015-02-18
Fixes and cleanup of client map tests.

Donnerbart:
* Fixed ClientMapBasicTest.testValues() which didn't check for the result size (just the content, which is not the same if elements are duplicated in result).
* Fixed ClientMapBasicTest.testValues_withPredicate() which was copied from testKeySet_withPredicate(), but not adapted to use IMap.values().
* Changed ClientMapBasicTest to use 3 member nodes instead of 1. Adapted testMapStatistics_withClientOperations() to deal with multiple members.
* Created new test class for EntryListener tests.
* Cleanup of ClientMapBasicTest and ClientTxnMapTest.

This PR prepares upcoming changes on IMap.values(), keySet() and entrySet(). Separate PR was made to keep review efforts low.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
@@ -17,15 +17,11 @@
 package com.hazelcast.client.map;
 
 import com.hazelcast.client.HazelcastClient;
-import com.hazelcast.core.EntryEvent;
-import com.hazelcast.core.EntryListener;
 import com.hazelcast.core.EntryView;
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
-import com.hazelcast.core.MapEvent;
 import com.hazelcast.monitor.LocalMapStats;
-import com.hazelcast.query.Predicate;
 import com.hazelcast.query.SqlPredicate;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.annotation.QuickTest;
@@ -55,12 +51,23 @@ import static org.junit.Assert.assertTrue;
 @Category(QuickTest.class)
 public class ClientMapBasicTest {
 
-    static HazelcastInstance client;
-    static HazelcastInstance server;
+    private static final String KEY = "key";
+    private static final String OLD_VALUE = "oldValue";
+    private static final String NEW_VALUE = "newValue";
+
+    private static HazelcastInstance client;
+    private static HazelcastInstance server1;
+    private static HazelcastInstance server2;
+    private static HazelcastInstance server3;
+
+    private String randomMapName = randomString();
+    private IMap<String, String> randomStringMap = client.getMap(randomMapName);
 
     @BeforeClass
     public static void init() {
-        server = Hazelcast.newHazelcastInstance();
+        server1 = Hazelcast.newHazelcastInstance();
+        server2 = Hazelcast.newHazelcastInstance();
+        server3 = Hazelcast.newHazelcastInstance();
         client = HazelcastClient.newHazelcastClient();
     }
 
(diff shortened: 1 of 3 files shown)


Donnerbart:
verify

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -54,13 +54,23 @@
...
+    private static HazelcastInstance server2;
+    private static HazelcastInstance server3;
+
+    private String randomMapName = randomString();
+    private IMap<String, String> randomStringMap = client.getMap(randomMapName);
 
     @BeforeClass
     public static void init() {
-        server = Hazelcast.newHazelcastInstance();
+        server1 = Hazelcast.newHazelcastInstance();
```
pveentjer: Why don't you make use of the HazelcastTestSupport? It has methods for creating clusters.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -54,13 +54,23 @@
 @RunWith(HazelcastParallelClassRunner.class)
 @Category(QuickTest.class)
 public class ClientMapBasicTest {
+    private static final String KEY = "key";
```
pveentjer: I don't like having some global fields to be used as arguments in my methods.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -54,13 +54,23 @@
...
+    private static final String NEW_VALUE = "newValue";
 
-    static HazelcastInstance client;
-    static HazelcastInstance server;
+    private static HazelcastInstance client;
+    private static HazelcastInstance server1;
+    private static HazelcastInstance server2;
+    private static HazelcastInstance server3;
+
+    private String randomMapName = randomString();
```
pveentjer: I don't think this is going to work because the test is executed in parallel. So who is going to guarantee that this field isn't shared between threads? So I think this is broken.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -54,13 +54,23 @@
...
+    private IMap<String, String> randomStringMap = client.getMap(randomMapName);
```
pveentjer: I also think this is broken due to above reasons. These tests are run in parallel, and now you have a shared data-structure between tests. This means that your tests become unpredictable due to races.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -72,765 +82,545 @@ public static void destroy() {
...
     }
 
     @Test
     public void testIsEmpty_afterPutRemove() {
-        final IMap map = client.getMap(randomString());
-        final Object key = "key";
-        map.put(key, "val");
-        map.remove(key);
-        assertTrue(map.isEmpty());
+        randomStringMap.put(KEY, NEW_VALUE);
```
pveentjer: If you really want to be 'lazy', please use

randomStringMap.put("key","newvalue");
randomStringMap.remove("key");
assertTrue(randomStringMap.isEmpty())

There is no need for those global arguments. It makes a test much harder to read.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -1088,72 +828,95 @@ public void testEntrySet_withPredicate() {
...
     @Test(expected = UnsupportedOperationException.class)
-    public void testLocalKeySet_WithPredicate(){
-        final IMap map = client.getMap(randomString());
-        map.localKeySet(new DumPredicate());
+    public void testLocalKeySet_WithPredicate() {
+        randomStringMap.localKeySet(new DummyPredicate());
     }
 
-    static class DumEntryListener implements EntryListener {
+    private static class DummyEntryListener implements EntryListener {
```
pveentjer: Why don't you create a mock using Mockito of you totally don't care about its content?

EntryListener entryListener = mock(EntryListener.class);

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -1088,72 +828,95 @@ public void testEntrySet_withPredicate() {
...
         }
+
+        @Override
         public void mapCleared(MapEvent event) {
         }
     }
 
-    static class DumPredicate implements Predicate {
-        public boolean apply(Map.Entry mapEntry) {
+    private static class DummyPredicate implements Predicate<String, String> {
```
pveentjer: Either create a mock using Mockity, or call it at least FalsePredicate.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
gurbuzali: test-cluster is not usable for client tests

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: I think this makes the whole test much easier to read, because you don't have any variations in the test files like "key", "Key", "KEY" etc. We totally don't care about the content in these cases. In my private projects I would even name those constants ANY_KEY, ANY_NEW_VALUE etc. to express, that it's not important what the content of the variable is. It is the usage/meta info which is the important thing. I can't find any better than a constant to express this (and it's not my idea, I got this from a book about testing).

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: It's not because the tests are all designed to use a unique map and don't change anything in cluster topology. They just change and test data on a map level. This worked before and was not changed by me.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: There is a new instance generated per test, otherwise they couldn't be run successfully, which they do. I'm just using JUnit features here to reduce code duplication and improve readability.

Every time a map is generated in a test, you directly know it's not the default <String, String> map. So you don't have to parse every client.getMap() to see the differences in those tests.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: For me it is the absolute opposite. Especially 'K' and 'k' can be hard to differ. This may be intended (check for different keys), but if you don't use a constant (which is easy to spot because of the UPPERCASE and unique color), you have to parse all strings while reading the tests. You can also easily make a typo in each hand-written string. For constants you have auto-completion.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: Didn't think about mocking. I just fixed the class name without having a deeper look at the test logic at all. But seems to be a good improvement, will change that.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: I didn't know about `FalsePredicate`. Apart from that the same, didn't had a deeper look at that test logic.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: I don't think it is easier to read at all. The problem is that I want to see what the keys are, so I need to go to the top of the file.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: All the methods on the unit tests are processed in parallel. So you have a randomMapName and a randomStringMap which is shared between tests. So if one tests adds an item and he expects to see an increase of 1, it could be that it gets a different value because another test using the same map, just cleared it.

pveentjer: Check the topic of the test:
```
```

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: This is ending in a yes vs no situation. I have given my reasons and you have given yours.

Perhaps another good reason not to do it; we don't do it anywhere else in the system and I would like to keep the tests consistent.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: We use a special type of runner to make sure these tests are run in parallel instead of sequentially. 

So all the test methods are running in parallel and now you have created shared fields. This is racy.

@mdogan can you have a look.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: static vs. non-static. It is perfectly fine to do this, the non-static values are not shared between parallel test executions. I just printed the `randomMapName` and the test name to proof this:
```
04edd77a-f8f8-4e34-a3a8-aa50fa1b311d testPutTTL_AfterExpireWhenKeyExists()
06ff1c91-aa02-44d2-93c0-35f26d43be19 testPutTTL()
0d9c3228-2403-4e6a-b160-5bd4cbf6a731 testPut_whenValueNull()
1700a619-28b0-44a6-af3c-a49aa046dd59 testSize_withMultiKeyPuts()
2881be77-bf12-4032-b8c4-c5aac877297a testPutAsync_withKeyNull()
34747c38-ce57-4fcc-8af2-18a607878ecd testIsEmpty_afterPutRemove()
4f4fd00a-9264-4bda-83f8-28fcfc066841 testPut_whenKeyNull()
50d575d3-3539-4829-a3e7-0e00e62810d3 testSize_whenEmpty()
52993fa8-db22-402d-9420-8313615739ae testPutAsync()
5ce06c98-d082-493a-8a86-aab9307913fe testPutTTL_whenKeyExists()
80095c25-8c65-47b7-b609-39b6fd45fca0 testPutAsync_whenKeyExists()
99d7a8ed-1153-42f1-973a-ec008db6a1c2 testClientGetMap()
9be24525-d9c1-42d8-928c-f1c8c4874df3 testGetName()
a0c14388-a125-4480-b8be-412376b48f0f testPut_whenKeyExists()
a2e3a601-06f2-4beb-8c02-792ad6ffb050 testPutAsyncTTL()
bbb9b9db-5043-49f1-bdf7-b4c731410c70 testIsEmpty_whenNotEmpty()
ca222e9a-4e55-4708-94c9-3044df863ca8 testPutAsync_withValueNull()
d246dfc3-5606-4bd2-9982-ee32db0505e5 testPutTTL_AfterExpire()
e6ae214f-90e5-4987-9ed4-fc42195a4829 testPut()
e94cedb0-3442-455f-86f8-d50b65eef09d testIsEmpty_whenEmpty()
fe6126ce-0c1b-4d14-a264-0e28a338c710 testSize()
```
If it would not be the case the whole test could not become green, which it does.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: Each test runs in its own thread, just static fields should be shared. Otherwise this would not work as it does (tests are all green).

`new Thread(new TestRunner(method, notifier), method.getName()).start();`

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: I just relieved you from that pain:
```
private static final String KEY = randomString();
private static final String OLD_VALUE = randomString();
private static final String NEW_VALUE = randomString();
```
It is completely irrelevant what's the content of those placeholders. They have to be unique, that's the only requirement. It's just the meta info which is relevant for the test case. This is why I consider this as a good practice. It saves you to parse and compare the content of Strings in each test, because this is not relevant for this kind of tests.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: Then we cannot move forward anymore. With the mass of test code we already have there will be no way to avoid inconsistency if we start to improve the test quality. But if we don't do this, there will be no progress. The problem is not the inconsistency, but that we have to define how the consistent state should look like in the end (so new tests will be written directly in that way and old ones can be transformed over time).

I think we should discuss how a good test should look like, but **not** in the line notes of this PR.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: It isn't a matter that you get unique names. But who says that test2 is not going to work on the map  that was created for test1. So unless there is some kind of magic going on, this is broken because of data races. @mdogan please have a look.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: At every test, can you print the System.identityHash for the 'this'? So the test instance. If they are different instances, then it is not racy. If they are the same, it is racy.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: It is your opinion that it an improvement; not mine. I think it is a regression.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
Donnerbart: The `identityHashCode()` is different, just checked that. Everything else would be very surprising, since I know it as best practice to use non-static fields in tests for data which is created identically (but not shared between) tests.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
gurbuzali: I agree with David here, for each test an instance of the test is created. So randomString will make sure that different tests will work on different maps.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: I just did a check check with another test, it appears that a new instance is created for every test method. So there is no shared state between fields. So I was wrong about that this approach is racy. Well.. I learned something new today.

```
 @Test
    @ClientCompatibleTest
    public void testSet(){
        HazelcastInstance hzInstance = createHazelcastInstance();
        IAtomicLong atomicNumber = hzInstance.getAtomicLong(randomString());
        atomicNumber.set(271);
        assertEquals(271,atomicNumber.get());

        System.out.println("testSet:"+System.identityHashCode(this));
    }

    @Test
    @ClientCompatibleTest
    public void testGet(){
        HazelcastInstance hzInstance = createHazelcastInstance();
        IAtomicLong atomicNumber = hzInstance.getAtomicLong(randomString());
        assertEquals(0,atomicNumber.get());
        System.out.println("testGet:"+System.identityHashCode(this));

    }

arted Running Test: AtomicLongTest.testSet
Started Running Test: AtomicLongTest.testGet
ASSERT_TRUE_EVENTUALLY_TIMEOUT = 120
testGet:1332226643
testSet:1068685279
Started Running Test: AtomicLongTest.testSet
Started Running Test: AtomicLongTest.testGet
ASSERT_TRUE_EVENTUALLY_TIMEOUT = 120
testGet:1332226643
testSet:1068685279
Finished Running Test: AtomicLongTest.testSet in 0.836 seconds.
```

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
pveentjer: I didn't know that part. I assumed the same instance was reused for every test method, so therefor you get races.

gurbuzali:
:+1:

Donnerbart:
We can put a merge on this on hold, until Peter and me had time to discuss the wanted style of a test. The fixes are not critical and the improvements should be clear and accepted by everyone.

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

PR #4639, opened 2015-02-18
Introduced the schedulequeue

pveentjer:
I added the ScheduleQueue as an abstraction that merges the BlockingQueue/PriorityQueue thing we have currently in the OperationScheduler. Apart from cleaning up the code, it also makes it a lot easier to integrate the caller runs scheduler since it can reuse all the thread-classes from the classic scheduler.

I also add some tests for the default implementation of this queue.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/classicscheduler/ScheduleQueue.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/classicscheduler/ScheduleQueue.java
@@ -1,0 +1,69 @@
+package com.hazelcast.spi.impl.classicscheduler;
+
+/**
+ * The ScheduleQueue is a kind of priority queue where 'tasks' are queued for scheduling.
+ * <p/>
+ * ScheduleQueue support concurrent producers but only need to support single consumers.
+ * <p/>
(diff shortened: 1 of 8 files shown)


mdogan:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/classicscheduler/ScheduleQueue.java
```diff
@@ -0,0 +1,69 @@
+package com.hazelcast.spi.impl.classicscheduler;
+
+/**
+ * The ScheduleQueue is a kind of priority queue where 'tasks' are queued for scheduling.
+ * <p/>
+ * ScheduleQueue support concurrent producers but only need to support single consumers.
```
jerrinot: why the interface specifies concurrency properties? MPSC? this is a concern of implementation, isn't it?

pveentjer: It isn't an implementation concern in the sense that a promise is made that this queue is never accessed by more than one thread. If your implementation can deal with multiple consumers, that is nice. but not needed.

PR #4643, opened 2015-02-19
Added the tx-executor for the ClientEngine

### LINKED ISSUE #4641: ClientEngineImpl huge number of threads
The ClientEngineImpl has an executor to process packets from clients without a partiton-id. This executor  has a very big poolsize potentially:

e.g.
12 cores per socket 
2 sockets
hyper threading
---> 48 cores 

This is multiplied by 20, so you get 960 threads and this can lead to problems. For example thread creation problems.

The reason this pool can become so big is to deal with a problem in the transactions from clients, but it an cause problems for other components.

The following program will trigger this growth:

```
public static void main(String[] args){

        IExecutorService executorService = client.getExecutorService("foo");
        for(int k=0;k<1000000;k++){
            executorService.submit(new MyRunnable());
        }
    }

    public static class MyRunnable implements Runnable, Serializable{
        @Override
        public void run() {
            try {
                Thread.sleep(1000);
                e.printStackTrace();
            }
        }
    }
```

If you print the size of this executor then you'll see that it grows to its maximum size very quickly.

It doesn't matter that the threadpool removes unused threads, the damage could already have been done.

pveentjer:
Initial work to resolve #4641

The changes to the client tx requests need to be made. And it only reduces the problem, but it doesn't solve it. As long as the system is not making use of tx operation, it will not make use of this potentially dangerously sized threadpool.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
@@ -92,11 +92,14 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
     public static final String SERVICE_NAME = "hz:core:clientEngine";
     private static final int ENDPOINT_REMOVE_DELAY_MS = 10;
     private static final int EXECUTOR_QUEUE_CAPACITY_PER_CORE = 100000;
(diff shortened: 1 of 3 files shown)


gurbuzali:
:+1: 
keep in mind that we need to fix client side requests so that they will have -2 as partitionId

gurbuzali:
verify

PR #4654, opened 2015-02-21
OperationScheduler cleanup

pveentjer:
Renamed OperationScheduler to OperationExecutor to put it more in line with the j.u.c.Executor.

Did some repackaging of OperationExecutor

Renamed OperationHandler to OperationRunner

moved some logic from the invocation into the OperationExecutor

Also added getPartitionId and currentThread information to OperationRunner. This is needed for the slow operation detection.


hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationRunner.java
```diff
@@ -14,38 +14,37 @@
...
  * on and this makes it possible to hook on all kinds of additional functionality like detecting slow operations, sampling which
  * operations are executed most frequently, check if an operation is still running, etc etc.
  *
  * Why is this class an abstract class and not an interface?
  *
- * If this class would be an interface, the {@link #currentTask()} will be more expensive. THis method is going to be called
- * 'frequently'; if there are 5000 partitions, then there will be 5000 calls to this method. Since it is likely that there are
- * multiple implementations of this OperationHandler, there will be virtual method call that can't be inlined if this class
- * would be an interface.
+ * If this class would be an interface, the {@link #currentTask()} will be more expensive. This method is going to be called
```
jerrinot: can you elaborate this statement? do you have any numbers to support it?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/ResponsePacketHandler.java
```diff
@@ -0,0 +1,15 @@
+package com.hazelcast.spi.impl.operationexecutor;
+
+import com.hazelcast.nio.Packet;
+import com.hazelcast.spi.impl.Response;
+
+/**
+ * The {@link com.hazelcast.spi.impl.operationexecutor.ResponsePacketHandler} is responsible for handling
```
jerrinot: I usually like your class-level JavaDoc, sadly this isn't the case. I have no idea what does it involve to "handle" a response packet.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/ResponsePacketHandler.java
```diff
@@ -0,0 +1,15 @@
...
+ * response packets.
+ */
+public interface ResponsePacketHandler {
+
+    Response deserialize(Packet packet) throws Exception;
```
jerrinot: why is response handler responsible for de-serialization? I know there might be a good reason, but it's not clear from this contract.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationRunner.java
pveentjer: Not everything needs to be backed up by benchmarks. Common sense is good enough.

In this case it is fine to have an abstract class. There is no added value of having an interface with the risk of running into megamorphic method calls. With the current approach it can't happen.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
```diff
@@ -95,7 +99,15 @@
      * @throws java.lang.NullPointerException if op is null.
      * @throws IllegalThreadStateException    if the operation is not allowed to be run on the calling thread.
      */
-    void runOperationOnCallingThread(Operation op);
+    void runOnCallingThread(Operation op);
+
+    /**
+     * Tries to run the operation on the calling thread if possible. Otherwise offload it to a different thread.
```
jerrinot: so is it sometimes sync and sometimes async? or does it wait for the operation to finish even when it's offloaded to a different thread?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationRunnerFactory.java
```diff
@@ -0,0 +1,10 @@
+package com.hazelcast.spi.impl.operationexecutor;
+
+public interface OperationRunnerFactory {
+
+    OperationRunner createPartitionRunner(int partitionId);
+
+    OperationRunner createGenericRunner();
+
+    OperationRunner createAdHocRunner();
```
jerrinot: what is AdHodRunner?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/ResponsePacketHandler.java
pveentjer: It is being handled :) So it is being processed if that makes more sense.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/ResponsePacketHandler.java
pveentjer: Not part of this PR. It was already there, so I won't change it.

jerrinot: ok

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
pveentjer: I don't want to use the works async and async here because it is too confusing. 

It means:
- run on calling thread 
- offload to different thread

You want to run on the calling thread as much as possible because it is faster. The logic used to be in the invocation, so the same logic is now moved into the scheduler to make the scheduler smarter and the invocation dumber.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/ResponsePacketHandler.java
jerrinot: so if I pass a Response to this handler, what behaviour can I expect? will it send the response to a caller? what is the caller can't be reached? 

besically a description of expected behaviour for future generations :)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationRunnerFactory.java
pveentjer: Good one. I'll add documentation.

There are 3 types of runners:
- partition runners; a runner responsible for processing all operations for a given partition
- generic runners: a runner responsible for processing generic operations for a given generic thread
- a ad hoc runner: a runner that is used to process generic operations, but it isn't done on a generic thread. For example in the system all kinds of operations are executed on the calling thread (e.g. partition/cluster internals). But.. you need to have an operation runner for them.. but the problem is that operation runners normally are not threadsafe; they have state like .e.g currentTask.

With the ad hoc runner this isn't the case.. it is fully immutable and can be shared between threads.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
jerrinot: sure. I understand performance implications of off-loading a task to a different thread. my understanding is that for some operations this method will block until the operation is finished, but for other (off-loaded) operations it will return immediately. is it so? it should be probably explicitly stated in the contract. 

maybe just a word of warning that in the case of off-loading it's not waiting for the operation to be finished.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/ResponsePacketHandler.java
pveentjer: But I think it is good to simplify it and only add a single method.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
pveentjer: Good point. I'll add it to the documentation.

Waiting here for the result is not desired, the waiting can be done through the invocation future.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/ResponsePacketHandler.java
pveentjer: The idea is that the OperationExecutor should not care what happens because that is part of the ResponsePacketHandler implementation. The Executor provides the threading, the actual handling of the response packet is not the responsibility of the Executor.

pveentjer:
@gurbuzali thanks!

PR #4657, opened 2015-02-22
Cleanup: skipping deregistration when callid 0

pveentjer:
Local calls like Map.get are currently deregistered from the OperationService, even if they are not registered. 

This PR fixes that; so when an operation has call id 0, it isn't registered and therefore doesn't need to be deregistered.

I have no clue of this is going to make any improvement in performance, but I think the logic is a bit cleaner since it doesn't make a lot of sense to do a deregister if there was no register.


jerrinot:
:+1:

Issue #3373, opened 2014-08-24
[TEST-FAILURE] LocalRegionFactoryDefaultTest.testQuery


bilalyasar:
this test fails 3.x problematic tests.
also this test failed maintenance see: https://github.com/hazelcast/hazelcast/issues/3068

```
java.lang.AssertionError: expected:<1> but was:<3>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at com.hazelcast.hibernate.HibernateStatisticsTestSupport.testQuery(HibernateStatisticsTestSupport.java:181)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-problematicTest/com.hazelcast$hazelcast-hibernate3/229/testReport/junit/com.hazelcast.hibernate/LocalRegionFactoryDefaultTest/testQuery/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-nightly/393/com.hazelcast$hazelcast-hibernate3/testReport/com.hazelcast.hibernate/LocalRegionFactoryDefaultTest/testQuery/

mesutcelik:
@kobalski ,

I am closing that right now. Fix attempt at 43966bf and 4f4c09d9cc18638b36bc80c8824ce946cd64408d

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-nightly/com.hazelcast$hazelcast-hibernate3/418/testReport/junit/com.hazelcast.hibernate/LocalRegionFactoryDefaultTest/testQuery/

Donnerbart:
[...]

PR #4674, opened 2015-02-25
Fix for #3373 Changed next time stamp implementation.

kobalski:
As discussed in https://github.com/hazelcast/hazelcast/issues/3068 before, our time stamp implementation used by hibernate query cache is millisecond based which is taken with instance.getCluster().getClusterTime(). It causes test failures in fast machines form time to time. I have modified nextTimestamp(HazelcastInstance instance) so that it waits until a different new timestamp value is generated. This implementation is similar to ehcache implementation which is considered as reference implementation for hibernate second level cache.

### DIFF
--- a/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
+++ b/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
@@ -59,8 +59,9 @@ public final class HazelcastTimestamper {
             for (long current = VALUE.get(), update = Math.max(base, current + 1); update < maxValue;
                  current = VALUE.get(), update = Math.max(base, current + 1)) {
                 if (VALUE.compareAndSet(current, update)) {
-                    if(runs > 1) {
-                        Logger.getLogger(HazelcastTimestamper.class).finest(String.format("Waiting for time to pass. Looped %d times",runs));
+                    if (runs > 1) {
+                        Logger.getLogger(HazelcastTimestamper.class)
+                                .finest(String.format("Waiting for time to pass. Looped %d times" , runs));
                     }
                     return update;
                 }
--- a/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
+++ b/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
@@ -20,6 +20,7 @@ import com.hazelcast.config.Config;
 import com.hazelcast.config.MapConfig;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.instance.GroupProperties;
+import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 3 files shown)


hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
```diff
@@ -29,13 +31,27 @@
...
 
     private static final int SEC_TO_MS = 1000;
+    private static final int BIN_DIGITS = 12;
+    private static final int ONE_MS = 1 << BIN_DIGITS;
+    private static final AtomicLong VALUE  = new AtomicLong();
 
     private HazelcastTimestamper() {
     }
 
     public static long nextTimestamp(HazelcastInstance instance) {
```
mdogan: Can you add a documentation over the method describing the algorithm used here, what it does, what it guarantees, why it's not just a regular timestamp like `System.currentTimeMillis()`, what `BIN_DIGITS` means, what's stored in `VALUE` .. etc?

mdogan:
Is it really 
>it waits until a different new timestamp value is generated

or more like *generates a different timestamp on each call*?

kobalski:
It waits in the while loop part, but what it does is basically as you said, generating different time stamp. VALUE field is for making sure every value method returns is different. I changed BIN_DIGITS field name to BASE_TIME_SHIFT and added some comments and logs to the code.

mesutcelik:
checkstyle failures. execute following to check them locally after the fix

```
mvn clean validate -P checkstyle
```

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
```diff
@@ -29,13 +32,42 @@
...
+     * inner for loop guarantees that unique value is generated.
+     * @see com.hazelcast.hibernate.HazelcastTimestamper#BASE_TIME_SHIFT
+     * @return uniquely & increasing value
+     */
     public static long nextTimestamp(HazelcastInstance instance) {
-        // System time in ms.
-        return instance.getCluster().getClusterTime();
+        int runs = 0;
+        while (true) {
+            long base = instance.getCluster().getClusterTime() << BASE_TIME_SHIFT;
```
ahmetmircik: If negative value for `base` causes problems, this line makes the code fragile. It seems better to me to put a check if a negative `base` is an undesired thing.

hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
```diff
@@ -29,20 +32,49 @@
...
+        }
     }
 
     public static int getTimeout(HazelcastInstance instance, String regionName) {
         try {
             final MapConfig cfg = instance.getConfig().findMapConfig(regionName);
             if (cfg.getTimeToLiveSeconds() > 0) {
-                // TTL in ms
+                // TTL in ms.
                 return cfg.getTimeToLiveSeconds() * SEC_TO_MS;
```
ahmetmircik: This line is not related to this PR but it may suffer from an integer overflow.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
kobalski:
```
 for (long current = VALUE.get(), update = Math.max(base, current + 1); update < maxValue;
        current = VALUE.get(), update = Math.max(base, current + 1)) {
```
In the for loop above max value between base and current + 1 is chosen, if negative value occurs due to bit shifting it will not effect the return value.

kobalski:
Also added integer overflow check for ``` public static int getTimeout(HazelcastInstance instance, String regionName) ``` in order to prevent returning negative value for timeout.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/HazelcastTimestamper.java
ahmetmircik: But what if also `maxValue` is negative, don't you end up an eternal while loop?

kobalski: Yes you are right, it may cause an infinite loop scenario, however inside the for loop, if statement will return the value no matter what,
```
 if (VALUE.compareAndSet(current, update)) {
         if (runs > 1) {
               Logger.getLogger(HazelcastTimestamper.class)
                    .finest(String.format("Waiting for time to pass. Looped %d times", runs));
          }
           return update;
  }
```
because update = Math.max(base, current + 1) will assign update to current + 1 and method will return inside if statement because VALUE.compareAndSet(current, update) will return true because update will be current + 1. Again, you are right the loop can be constructed as an infinite loop because update < maxValue will never be satisfied, however the method will always return in the first run of the loop.

ahmetmircik:
verify

mesutcelik:
@kobalski ,

I am closing that right now. There is another fix attempt at 43966bf and 4f4c09d9cc18638b36bc80c8824ce946cd64408d

Issue #4517, opened 2015-02-02
QueryIndexMigrationTest.testQueryWithIndexDuringJoin


bilalyasar:
```
java.lang.AssertionError: expected:<500> but was:<493>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at com.hazelcast.map.query.QueryIndexMigrationTest.testQueryWithIndexDuringJoin(QueryIndexMigrationTest.java:232)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/429/testReport/junit/com.hazelcast.map.query/QueryIndexMigrationTest/testQueryWithIndexDuringJoin/

mesutcelik:
This hasn't failed lately. Have we fixed such an issue before?

WDYT @ahmetmircik ?

ahmetmircik:
dont remember a fix but it looks like we can close it till the next failure.

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/2746/testReport/junit/com.hazelcast.map.query/QueryIndexMigrationTest/testQueryWithIndexDuringJoin/

mesutcelik:
closed via https://github.com/hazelcast/hazelcast/pull/4686

mesutcelik:
have we fixed that problem in maintenance?

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-OracleJDK1.7/46/com.hazelcast$hazelcast/testReport/com.hazelcast.map.query/QueryIndexMigrationTest/testQueryWithIndexDuringJoin/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3095/testReport/junit/com.hazelcast.map.query/QueryIndexMigrationTest/testQueryWithIndexDuringJoin/

mesutcelik:
no recent failure. closing.

mesutcelik:
[...]

PR #4686, opened 2015-02-25
Using waitAllForSafeState in QueryIndexMigrationTest

ajermakovics:
#4517 
Not really a fix since I couldn't reproduce the test failure. This is just to see if it makes any difference.


ajermakovics:
verify

gurbuzali:
can we have a back-port

jerrinot:
:+1:

PR #4688, opened 2015-02-26
getting mapName from recordStore rather than mapConfig

### LINKED ISSUE #4667: Strange mapName for MapStoreFactory.newMapStore()
Hi.
I used the asterix configuraiton for maps and mapStore configuration using custom MapStoreFactory:
```xml
<hz:map name="query.archive.*"
                    eviction-percentage="25"
                    eviction-policy="LRU"
                    max-size="${judgements.hz.archive.max.size}"
                    max-size-policy="PER_NODE"
                    max-idle-seconds="${judgements.hz.archive.max.idle}"
                    in-memory-format="OBJECT"
                    read-backup-data="true"
                    backup-count="0">
                <hz:map-store enabled="true" factory-implementation="mapStoreProvider"
                              write-delay-seconds="0"/>
            </hz:map>
```
This worked ok for half a year.
Then I tried the more specific configuration for some maps and changed maps configuration to the following:

```xml
<hz:map name="query.archive.RU.WEB"
                   ...>
                <hz:map-store enabled="true" factory-implementation="mapStoreProvider"
                              write-delay-seconds="0"/>
            </hz:map>
            <hz:map name="query.archive.TR.WEB"
                    ...>
                <hz:map-store enabled="true" factory-implementation="mapStoreProvider"
                              write-delay-seconds="0"/>
            </hz:map>
            <hz:map name="query.archive.*.WEB"
                    ....>
                <hz:map-store enabled="true" factory-implementation="mapStoreProvider"
                              write-delay-seconds="0"/>
            </hz:map>

          <hz:map name="query.archive.*"
                    ....>
                <hz:map-store enabled="true" factory-implementation="mapStoreProvider"
                              write-delay-seconds="0"/>
            </hz:map>
```
And now I face the sporadic strange mapName parameter in my MapStoreFactory with asterix: **"query.archive.*"**

Code for MapStoreFactory is not interesting, it tries to create dynamic map store by the name. Name is usually **query.archive.[1].[2]** and 1, 2 are used to create map store. It works but twice the last weeks my mapStoreFactory failed as got the strange map name: "query.archive.*". Usually these sporadic problems are the sign of  concurrency race.

So, independent of my code and the way configuration with asterixes parsed - how can this be that Map StoreFactory gets the sample with asterixes instead of specific map name?

Btw, I use Hz 3.3.2

gurbuzali:
fixes #4667 
since mapName acquired from MapConfig may be a wildcard configured name we should get the actual name from recordStore.

PS: testing this is very hard since issue appears only if an operation runs before the initialization of map-proxy and also triggers eviction. I've manually checked if there are other parts which uses mapConfig.name rather than actual name, there is none.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictionOperator.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictionOperator.java
@@ -75,8 +75,7 @@ public final class EvictionOperator {
         return maxSizeChecker;
     }
 
-    public void removeEvictableRecords(final RecordStore recordStore, int evictableSize, final MapConfig mapConfig,
-                                       boolean backup) {
+    public void removeEvictableRecords(RecordStore recordStore, int evictableSize, MapConfig mapConfig, boolean backup) {
         final MapServiceContext mapServiceContext = this.mapServiceContext;
         final EvictionPolicy evictionPolicy = mapConfig.getEvictionPolicy();
         // criteria is a long value, like last access times or hits,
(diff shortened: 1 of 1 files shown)


ahmetmircik:
:+1:

Issue #2403, opened 2014-05-07
Xml Config: XSD validation


pveentjer:
Currently there is no XSD validation on the client/server-hazelcast xml. 

We need to get this in place so we can do a 

[mop]
[/mop]

Or

[map enoebled="true"]....

(Please replace [ and ] by the bigger/smaller than brackets).

bilalyasar:
closed via #4701

Issue #3256, opened 2014-08-13
Hazelcast client should not just ignore the fact the XML is for server and shouldn't use default xml feature to connect to localhost


amitohale:
when hazelcast client is failed to connect to any server node after parsing the server XML, it used the default client XML configuration present in hazelcast-client.jar which always connects to localhost. 

If this is expected behavior, client should log the name of xml as it's not the same configuration one expects it to be.

Example:

my-client-config.xml is same as server configuration with multicast discovery running with multiple nodes and one node on localhost and when I am creating the hazelcast instance it's connected to localhost. And if I kill the localhost node and run the same stuff again, it fails to connect.

ClientConfig conf = new XmlClientConfigBuilder("my-client-config.xml").build();
HazelcastInstance instance = HazelcastClient.newHazelcastClient(conf);

bilalyasar:
closed via #4701

Issue #2402, opened 2014-05-07
Config: Throw InvalidConfigurationException 


pveentjer:
Currently a HazelcastException is thrown when we run into configuration failures. But we should throw a InvalidConfigurationException.

See XmlConfigBuilder.build:

 Config build(Config config) {
        try {
            parse(config);
        } catch (Exception e) {
            throw new HazelcastException(e);
        }
        config.setConfigurationFile(configurationFile);
        config.setConfigurationUrl(configurationUrl);
        return config;
    }

bilalyasar:
closed via #4701

PR #4701, opened 2015-03-02
fail-fast and schema validation implementation

bilalyasar:
-added schema validation ( server side + client side)
-edited xsd files for fail-fast
-added fail-fast tests.
-edited some existing tests for fail-fast
-added invalidconfig. exception.

member configuration for server side has been changed to:
```               
       <member-list>
                    <member>127.0.0.1</member>          
                    <member>127.0.0.2</member>          
      </member-list>
```
for detailed information see prd and design doc.


noctarius:
Does the schema validation is aware of the import statement? :)

bilalyasar:
@noctarius yes it is :) it imports all other xmls(given in import tag)  than validates the final xml.

pveentjer:
Can you test this on a machine without an internet connection? I have seen in the past that with schema validation that sometimes the system tried to connect to the internet to download the schema's. And this can be a huge problem in production environments.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -170,39 +185,51 @@ private void parseAndBuildConfig(ClientConfig clientConfig) throws Exception {
...
-                final int poolSize = Integer.parseInt(getTextContent(node));
-                clientConfig.setExecutorPoolSize(poolSize);
+            if (occurrenceSet.contains(nodeName)) {
+                throw new InvalidConfigurationException("Duplicate '" + nodeName + "' definition found in XML configuration. ");
+            }
+            handleXmlNode(node, nodeName);
+            if (!canOccurMultipleTimes(nodeName)) {
+                occurrenceSet.add(nodeName);
             }
         }
```
mesutcelik: so I see here that we are disabling to have multiple tags if they are not allowed. isn't there a way to do the same with xsd validation?

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -170,39 +185,51 @@ private void parseAndBuildConfig(ClientConfig clientConfig) throws Exception {
             domLevel3 = false;
         }
         process(root);
+        schemaValidation(root.getOwnerDocument());
         handleConfig(root);
     }
 
     private void handleConfig(final Element docElement) throws Exception {
-        for (Node node : new IterableNodeList(docElement.getChildNodes())) {
+        for (org.w3c.dom.Node node : new IterableNodeList(docElement.getChildNodes())) {
```
mesutcelik: unnecessary full package definition.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -170,39 +185,51 @@ private void parseAndBuildConfig(ClientConfig clientConfig) throws Exception {
...
+        } else if (NETWORK.isEqual(nodeName)) {
+            handleNetwork(node);
+        } else if (LOAD_BALANCER.isEqual(nodeName)) {
+            handleLoadBalancer(node);
+        } else if (NEAR_CACHE.isEqual(nodeName)) {
+            handleNearCache(node);
+        } else if (EXECUTOR_POOL_SIZE.isEqual(nodeName)) {
+            final int poolSize = Integer.parseInt(getTextContent(node));
+            clientConfig.setExecutorPoolSize(poolSize);
+        }
```
mesutcelik: you can add handleExecutorPoolSize method to be aligned with the methods above.

hazelcast-client/src/main/resources/hazelcast-client-config-3.5.xsd
```diff
@@ -23,7 +23,7 @@
 
     <xs:element name="hazelcast-client">
         <xs:complexType>
-            <xs:sequence>
+            <xs:choice minOccurs="1" maxOccurs="unbounded">
```
mesutcelik: can you please expalin in one sentence what is the diff between **<xs:choice minOccurs="1" maxOccurs="unbounded">** and **<xs:all>**

I forget it all the time.

hazelcast-client/src/main/resources/hazelcast-client-config-3.5.xsd
```diff
@@ -51,7 +51,7 @@
         </xs:complexType>
     </xs:element>
     <xs:complexType name="aws">
-        <xs:sequence>
+        <xs:all>
```
mesutcelik: This kind of replacement seems relaxing xmls so enabling user to get rid of order restriction. That is ok if we don't need to have them in order.

hazelcast/src/main/resources/hazelcast-default.xml
```diff
@@ -44,7 +44,9 @@
                 <multicast-port>54327</multicast-port>
             </multicast>
             <tcp-ip enabled="false">
-                <interface>127.0.0.1</interface>
```
mesutcelik: That is deleted so I expect it is not used anymore and including it will not validate the xml.

hazelcast/src/main/resources/hazelcast-default.xml
```diff
@@ -197,11 +199,6 @@
         <value-collection-type>SET</value-collection-type>
     </multimap>
 
-    <multimap name="default">
-        <backup-count>1</backup-count>
-        <value-collection-type>SET</value-collection-type>
-    </multimap>
-
```
mesutcelik: That is deleted so I expect it is not used anymore and including it will not validate the xml.

bilalyasar: i deleted because there was same definition already in xml file.

hazelcast/src/main/resources/hazelcast-default.xml
bilalyasar: i will add again this due to compatibility.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
bilalyasar: unfortunately no.

bilalyasar:
verify

bilalyasar:
verify

mesutcelik:
:+1:

pveentjer:
Can you test it with wifi disabled? I have seen in the past that sometimes the schema is loaded from the internet and then the cluster fails if access is restricted.

bilalyasar:
@pveentjer yes i tried and i don't see any problem.
Also we are using always xsd file inside the jar. This is our design decision. So internet connection is not very important in this respect.

gurbuzali:
I also tried it now, no problem

pveentjer:
Great.

mesutcelik:
Thanks @bilalyasar for the effort!

bilalyasar:
thanks @mesutcelik for reviewing my code.

also this PR closes #2403 #3256 #2402

Issue #4705, opened 2015-03-02
NoSuchMethodError com.hazelcast.core.IMap.addEntryListener(Lcom/hazelcast/core/EntryListener;Z)Ljava/lang/String;


Danny-Hazelcast:
Binary / source-level incompatibility  3.4 VS 3.5

```   Failure[
   message='Worked ran into an unhandled exception'
   type='Worker exception'
   agentAddress=10.168.112.55
   time=Sun Mar 01 02:27:01 EET 2015
   workerAddress=client:10.168.112.55
   workerId=worker-10.168.112.55-4-client
   test=TestCase{
          id=MapEntryListenerTest
        , class=com.hazelcast.stabilizer.tests.map.MapEntryListenerTest
        , basename=MapEntryListenerTest
        , deleteProb=0.2
        , evictProb=0.2
        , keyCount=1000
        , maxEntryListenerDelayMs=0
        , minEntryListenerDelayMs=0
        , randomDistributionUniform=true
        , removeProb=0.2
        , replaceProb=0.25
        , threadCount=3
        , valueCount=1000
        , valueLength=100
        , writeProb=0.4
        , writeUsingPutIfAbsent=0.25
        , writeUsingPutProb=0.5
    }
   cause=java.lang.NoSuchMethodError: com.hazelcast.core.IMap.addEntryListener(Lcom/hazelcast/core/EntryListener;Z)Ljava/lang/String;
	at com.hazelcast.stabilizer.tests.map.MapEntryListenerTest.setup(MapEntryListenerTest.java:88)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at com.hazelcast.stabilizer.utils.ReflectionUtils.invokeMethod(ReflectionUtils.java:29)
	at com.hazelcast.stabilizer.worker.TestContainer.setup(TestContainer.java:168)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
[...]

PR #4708, opened 2015-03-02
Fixes entry listener backward binary-compatibility issue

ahmetmircik:
closes https://github.com/hazelcast/hazelcast/issues/4705

Contains addition of previously removed listener adder method signatures as deprecated


jerrinot:
can you add the original tests as well?

pveentjer:
Can we have this branch run with the stabilizer before merging it?

Issue #4671, opened 2015-02-24
Two issues with local near cache


lukasblu:
Hi,

while stress testing our application, we discovered two issues with using local near caches:

1. Even on a single node, the local near cache is not synchronously invalidated after the map is modified.

2. Also on a single node, it may happen that near cache invalidations are lost.

To 1:
I understand that you do not give full consistency guarantees when using a near cache, however I kind of expected that at least the local node always sees its latest changes.
While adding full consistency guarantees on a distributed near cache might be very expensive (in terms of synchronization and operation latency), providing such guarantees on the local node only should be much cheaper.

To 2:
This is certainly a bug.

Please find below a test case which shows both issues.

If using a ConcurrentHashMap, the test always succeeds. If using a distributed Hazelcast map with local near cache enabled, the test either fails with
  "Assertion violated."
which means that the near cache showed old data, or the test fails with
  "Near cache did *not* become consistent."
which means that even after waiting for some time, the near cache did not show the latest data - hence near cache invalidations were lost.

It would be great to hear your opinion on 1 and a fix for 2 would be amazing! :-)

Thanks and best,
Lukas

```java
package com.nm.test.hazelcast.nearcache;

import com.hazelcast.config.*;
import com.hazelcast.core.Hazelcast;
import com.hazelcast.core.HazelcastInstance;
import com.nm.test.hazelcast.TestHazelcast;
import org.apache.log4j.BasicConfigurator;
import org.apache.log4j.Logger;
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.ConcurrentHashMap;
import java.util.concurrent.ConcurrentMap;
import java.util.concurrent.atomic.AtomicBoolean;
import java.util.concurrent.atomic.AtomicInteger;
import junit.framework.TestCase;

/**
 * A test to ensure the near cache eventually reflects the data from the map.
 */
public class TestLocalNearCache6 extends TestCase {

	private static final Logger logger = Logger.getLogger(TestLocalNearCache6.class);

	private static final int numGetters = 7;

	private static final String mapName = "testMap" + TestLocalNearCache6.class.getSimpleName();

	private static final String key = "key123";

	private AtomicInteger valuePut = new AtomicInteger(0);

	private AtomicBoolean stop = new AtomicBoolean(false);

	private AtomicBoolean failed = new AtomicBoolean(false);

	@Override
	protected void setUp() throws Exception {

		// configure logging
		if (!TestHazelcast.loggingInitialized) {
[...]

PR #4710, opened 2015-03-02
Prevent updates on the near cache if the new value was invalidated in th...

lukasblu:
...e meantime.

Hi @serkan-ozal ,

here is the fix for #4671 which seams to work for me.

Could you have a look and run your complete test suite against it?

Thanks and best,
Lukas

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
@@ -131,7 +131,7 @@ public class NearCache {
     }
 
     // this operation returns the given value in near-cache memory format (data or object)
-    public Object put(Data key, Data data, long invalidateOpCountBefore) {
+    public Object put(Data key, Data data, long invalidateCountBefore) {
         fireTtlCleanup();
         if (evictionPolicy == EvictionPolicy.NONE && cache.size() >= maxSize) {
             // no more space in near-cache -> return given value in near-cache format
@@ -162,7 +162,7 @@ public class NearCache {
             } else {
                 // ignore put since someone else already updated this value
             }
-        } else if (record.invalidateCount < invalidateOpCountBefore) {
+        } else if (record.invalidateCount < invalidateCountBefore) {
             CacheRecord recordNew = new CacheRecord(key, value, invalidateCounter.get());
             if (cache.replace(key, record, recordNew)) {
                 // good, old value updated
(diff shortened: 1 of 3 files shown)


pveentjer:
@serkan-ozal since the JCache is now your baby, can you have a look at this PR?

serkan-ozal:
@pveentjer I had looked at the PR but I think, solution in this PR may creates unnecessary objects even while invalidating. On the other hand, since this near-cache issue is for IMap, maybe @ahmetmircik can also look at this PR.

lukasblu:
Hi @serkan-ozal ,

yes, that is true, my solution above does indeed keep objects to remember which value are no longer valid. However, since the actual value is not retained, the overhead is really small. We are already using the PR in real life solutions and see no overhead what so ever.

Maybe an option to improve this even further would be to keep those invalidation objects only if caching of local entries is enabled and/or to only keep those objects for a very short period of time.

Best,
Lukas

ahmetmircik:
Hi @lukasblu, 

It seems to me this may also be an alternative solution: (btw it needs more tests and not a finished work)
https://github.com/hazelcast/hazelcast/compare/master...ahmetmircik:fix/3.5/nearcacheStalenessMaybeFix?expand=1

_Summary of that work:_
Only keeping an array of counters and increasing values of two random indexes according to pre-calculated hashes and checking whether or not the sum of those indexes has changed, by comparing that sum before and after `put-near-cache` we can easily guess the staleness of the value in near-cache, there should be a very low false positive rate that means we can also remove a not-stale value with a very low probability, like 1 in every 100 put. Also we don't create any object for every invalidation.

bilalyasar:

lukasblu:
Hi @ahmetmircik ,

I had a look at your solution and - without having tested it yet - it feels good.

Did you have a chance to do more testing? And did you already run my two tests from #4671 (TestLocalNearCache6 and TestLocalNearCache7)? I'm not sure if have time to run my tests against your branch today, but I could do more testing on Monday or Tuesday next week.

Thanks and best,
Lukas

Ps. @bilalyasar depending on the test results on Ahmet's solution, my PR would become obsolete...

ahmetmircik:
Hi @lukasblu, 

We have fixed this issue with this PR https://github.com/hazelcast/hazelcast/pull/8091. We will also apply that fix to current 3.6 and 3.7. Now, i am closing this PR. 

Thanks for pointing this issue.

PR #4711, opened 2015-03-03
Introduction of spring-aware tag for hz namespace

emrahkocaman:
`spring-aware` config element added to `hazelcast-spring-3.5.xsd`.

By default hz uses `SpringManagedContext` to scan `SpringAware` annotations. This may cause some performance penalty even if users don't use `SpringAware`. 
Starting with 3.5 `SpringManagedContext`will be disabled by default. To enable it `<hz:spring-aware />` tag is needed to be added to the config.

### DIFF
--- a/hazelcast-documentation/src/SpringIntegration.md
+++ b/hazelcast-documentation/src/SpringIntegration.md
@@ -205,12 +205,12 @@ Hazelcast Distributed Objects could be marked with @SpringAware if the object wa
 - to apply factory callbacks such as `ApplicationContextAware`, `BeanNameAware`,
 - to apply bean post-processing annotations such as `InitializingBean`, `@PostConstruct`.
 
-Hazelcast Distributed `ExecutorService`, or more generally any Hazelcast managed object, can benefit from these features. To enable SpringAware objects, you must first configure `HazelcastInstance` as explained in the [Spring Configuration section](#spring-configuration).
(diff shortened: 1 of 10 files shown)


hazelcast-documentation/src/SpringIntegration.md
```diff
@@ -228,6 +228,7 @@ Hazelcast Distributed `ExecutorService`, or more generally any Hazelcast managed
 
   <hz:hazelcast id="instance">
     <hz:config>
+      <hz:spring-aware />
```
bilalyasar: i am not sure about tradition but adding attribute `enabled` seems more sensible to me.
like `<hz:spring-aware enabled="true"/>` , 
but without attribute still ok for me.

emrahkocaman: @bilalyasar , this is a spring convention. See http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/transaction/annotation/EnableTransactionManagement.html
You enable `Transactional` annotation by adding `<tx:annotation-driven />` for example.

bilalyasar: :+1:

mesutcelik:
@Serdaro ,

Can you please review documentation updates in this PR? There is another one for maintenance-3.x as well. https://github.com/hazelcast/hazelcast/pull/4715

PR #4716, opened 2015-03-03
Collection package structure cleanup

pveentjer:
The implementation classes have now the same package structure as used in com.hazelcast.concurrent. Also the com.hazelcast.queue.impl package has been moved inside of the com.hazelcast.collection.impl package and set up in exactly the same was as list/set etc.

Also there is a separation of txncollections and non transactional structures, so the com.hazelcast.collection.impl root looks like:
```
collection
    client
    operations
list
    client
    operations
set
    client
    operations
queue
    client
    operations
txncollection
    client
    operations
txnlist
    client
    operations
txnset
    client
    operations
````


jerrinot:
:+1:

pveentjer:
In the near future I want to do the same for the collections tests as what I did for the concurrent tests so we can easily test all the local and remote invocations.

mtopolnik:
:+1:

PR #4695, opened 2015-02-27
membership registration made over connection to prevent re registration ...

sancar:
membership registration made over connection to prevent re registration when connection is removed. 
Partition trigger mechanism is changed from client. 
Membershiplistener is converted to initial membership listener to get member list without loss of event. ClientMembershipEvent handler is refactored out from ClientInvocationServiceSupport to make it more understandable.

PR #4694, opened 2015-02-27
fixed client connection retry operations and reconnection problems. fixe...

eminn:
...d the test failure that led us to find those bugs. fixes #4647 , #4692 , #4693

Issue #3525, opened 2014-09-07
[TEST-FAILURE] ClientMapReduceTest failures


bilalyasar:
ClientMapReduceTest.testInProcessCancellation
ClientMapReduceTest.testExceptionDistribution
```
java.lang.Exception: test timed out after 60000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:226)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1033)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1326)
	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:282)
	at com.hazelcast.client.proxy.ClientMapReduceProxy$ClientCompletableFuture.get(ClientMapReduceProxy.java:208)
	at com.hazelcast.spi.impl.AbstractCompletableFuture.get(AbstractCompletableFuture.java:87)
	at com.hazelcast.client.mapreduce.ClientMapReduceTest.testInProcessCancellation(ClientMapReduceTest.java:122)
```

```
java.lang.Exception: test timed out after 60000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:226)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1033)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1326)
	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:282)
	at com.hazelcast.client.proxy.ClientMapReduceProxy$ClientCompletableFuture.get(ClientMapReduceProxy.java:208)
	at com.hazelcast.spi.impl.AbstractCompletableFuture.get(AbstractCompletableFuture.java:87)
	at com.hazelcast.client.mapreduce.ClientMapReduceTest.testExceptionDistribution(ClientMapReduceTest.java:87)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/301/testReport/

mesutcelik:
[...]

PR #4722, opened 2015-03-04
Related to failures in #4695 and #4694 . 

sancar:
Any task that is doing a blocking invocation will work on internal executor. Executor is the place where retry invocations are handled.


eminn:
I merged this pull request locally and this patch also fixes the test failure #3525

gurbuzali:
no need to back-port I guess ?

sancar:
Yes, no need to backport

Issue #3796, opened 2014-10-08
TransactionQueueTest.testIssue859And863


bilalyasar:
```
java.lang.AssertionError: expected:<2000> but was:<1999>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at com.hazelcast.queue.TransactionQueueTest.testIssue859And863(TransactionQueueTest.java:265)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/1803/testReport/junit/com.hazelcast.queue/TransactionQueueTest/testIssue859And863/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/2504/testReport/junit/com.hazelcast.queue/TransactionQueueTest/testIssue859And863/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/418/testReport/junit/com.hazelcast.queue/TransactionQueueTest/testIssue859And863/

sancar:
:(
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast/11673/testReport/junit/com.hazelcast.queue/TransactionQueueTest/testIssue859And863/

sancar:
@jerrinot Are you still working on this? I started to get this failure in all my prs. We need to fix this asap. It seems you already send a fix, but this is still failing.

jerrinot:
@sancar: nope. If ale PRs are failing with this then it's likely a new regression.

sancar:
And I just tried, it passed in my local. Probably randomly failing.

jerrinot:
yeah, according to https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast%24hazelcast/11673/testReport/junit/com.hazelcast.queue/TransactionQueueTest/testIssue859And863/history/ 

it's failing from time to time. I'll have a look.

bilalyasar:
[...]

PR #4724, opened 2015-03-04
Fixed partition backup replica handling for member join and removal cond...

### LINKED ISSUE #4687: Memory Distribution Imbalance (Possible Memory leak?)
Reproduce steps...

(1) Start one hazelcast cluster member. -Xmx3G -Xmx3G -XX:-UseConcMarkSweepGC
(2) Start a client and prime one million entries of a DataSerializable object containing 14 longs, 1 Double, 1 List<String> with ~ 20 entries.  I don't believe however this is the cause, you could maybe try with primitives.  The Key is a Long
(3) Use default map config (so backup = 1)
(4) Now start a second cluster member wait for migration to complete.  Perform full GC and observe heap size. Both should be roughly same.
(5) Now add 3rd member, wait for migration to complete. Perform full GC, memory should be less than first two.
(6) Now add 4th member, wait for migration to complete. Perform full GC, memory will be less than 3rd and considerable less than members 1 & 2.

On my laptop the retained member after full GC at end of the incremental starts was :-
Member 1 : 950mb
Member 2 : 950mb
Member 3 : 750mb
Member 4 : 650mb

I performed heap dumps on Member 1 and Member 4 and noted that Member 1 (oldest) had 1.4 million com.hazelcast.nio.serialization.DefaultData entries and Member 4(youngest) had 989,000 entries. Examining Management Centre shows an even distribution of Map entries.

I tried this test again but with backups=0, started one member and primed, then added 2,3,4 members incrementally.  The memory was evenly balanced, so it appears there might be an issue with setting backups to 1 and then starting with 1 member priming a large dataset and then incrementally adding members.

mdogan:
...itions.

 - When multiple nodes join sequentially after partitions are assigned/distributed, old nodes were failing to clean backup replicas larger than configured backup-count. This was causing a memory leak. See issue #4687.

 - When multiple nodes leave the cluster at the same time (or in a short period), new partition owner was loosing some partition replica versions and this was causing backup nodes for those specific replica-indexes to fail syncing data from owner node, although owner node holds the whole partition data.

 Added MigrationAwareServiceTest to verify all these behaviours.

Fixes #4687

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/impl/PromoteFromBackupOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/impl/PromoteFromBackupOperation.java
@@ -34,6 +34,7 @@ import com.hazelcast.spi.PartitionMigrationEvent;
 import com.hazelcast.spi.impl.NodeEngineImpl;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collection;
 
 import static com.hazelcast.partition.InternalPartitionService.SERVICE_NAME;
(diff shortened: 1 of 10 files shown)


hazelcast/src/main/java/com/hazelcast/partition/impl/PromoteFromBackupOperation.java
```diff
@@ -64,6 +66,42 @@ public void run() throws Exception {
         }
     }
 
+    private void setMissingReplicaVersions() {
```
mdogan: Thanks to @metanet, this is inspired by his commit: https://github.com/metanet/hazelcast/commit/7a1f882c961bb56c7f10c6d20a5e2acf446944e4

hazelcast/src/main/java/com/hazelcast/partition/impl/ResetReplicaVersionOperation.java
```diff
@@ -34,6 +36,10 @@ public void run() throws Exception {
         int partitionId = getPartitionId();
         InternalPartitionService partitionService = getService();
         long[] versions = partitionService.getPartitionReplicaVersions(partitionId);
+        // InternalPartitionService.getPartitionReplicaVersions() returns internal
+        // version array, we need to clone it here
+        versions = Arrays.copyOf(versions, InternalPartition.MAX_BACKUP_COUNT);
```
mdogan: Thanks to @metanet, who found this issue and fixed in his repo:
https://github.com/metanet/hazelcast/commit/7a1f882c961bb56c7f10c6d20a5e2acf446944e4

mdogan:
Failed test: #3796

```
testIssue859And863(com.hazelcast.queue.TransactionQueueTest)  Time elapsed: 123.285 sec  <<< FAILURE!
java.lang.AssertionError: expected:<2000> but was:<1999>
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at com.hazelcast.queue.TransactionQueueTest$2.run(TransactionQueueTest.java:271)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:557)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:573)
	at com.hazelcast.queue.TransactionQueueTest.testIssue859And863(TransactionQueueTest.java:268)
```

verify

PR #4725, opened 2015-03-04
Simplified and documented concurrency behavior of SlowOperationLog

Donnerbart:
* Simplified and documented concurrency behavior of `SlowOperationLog`. There is just a single writer in this scenario, which is already stated in the class documentation.
* Fixed wrong variable name (duration is in milliseconds, not nanoseconds).

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
@@ -20,10 +20,7 @@ import com.eclipsesource.json.JsonArray;
 import com.eclipsesource.json.JsonObject;
 import com.eclipsesource.json.JsonValue;
 import com.hazelcast.management.JsonSerializable;
-import com.hazelcast.util.ConcurrencyUtil;
-import com.hazelcast.util.ConstructorFunction;
 
-import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
@@ -34,23 +31,17 @@ import static com.hazelcast.util.JsonUtil.getLong;
 import static com.hazelcast.util.JsonUtil.getString;
 
 /**
- * Data structure for slow operation logs.
+ * Internal data structure for {@link SlowOperationDetector}.
  * <p/>
- * A single instance of this class is created by {@link SlowOperationDetector} and shared with
- * {@link com.hazelcast.monitor.impl.LocalOperationStatsImpl} to deliver a JSON representation for the Management Center.
+ * A collection of this class is created by {@link SlowOperationDetector} and shared as <code>Collection<JsonSerializable></code>
+ * with {@link com.hazelcast.monitor.impl.LocalOperationStatsImpl} to deliver a JSON representation for the Management Center.
  * <p/>
- * Data is exclusively written by {@link SlowOperationDetector} and just read by other threads.
+ * All data is exclusively written by {@link SlowOperationDetector.SlowOperationDetectorThread} and maybe read by other threads.
+ * Only fields which are exposed via the {@link #toJson()} method need synchronization. All other fields are used single threaded.
  */
-public final class SlowOperationLog implements JsonSerializable {
-    private static final int SHORT_STACKTRACE_LENGTH = 200;
+final class SlowOperationLog implements JsonSerializable {
 
-    private static final ConstructorFunction<Integer, Invocation> INVOCATION_CONSTRUCTOR_FUNCTION
-            = new ConstructorFunction<Integer, Invocation>() {
-        @Override
-        public Invocation createNew(Integer arg) {
-            return new Invocation();
-        }
-    };
+    private static final int SHORT_STACKTRACE_LENGTH = 200;
 
     private final ConcurrentHashMap<Integer, Invocation> invocations = new ConcurrentHashMap<Integer, Invocation>();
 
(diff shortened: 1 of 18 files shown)


Donnerbart:
Maybe Github is the wrong place to have a closer look at this PR (context, highlighting, code navigation is missing).

I hope I got everything right now. I really just need simple read/write barriers to ensure visibility between a single writer and a single reader (at the same time).

Even a stale value or an inconsistent state will not be super critical (e.g. when the `toJSON()` method is called the field `totalInvocations` could already be increased, although the new `Invocation` instance is not in the CHM yet). This feature is just for reporting slow operations, no data will be lost. If we need the data to be 100% consistent, we may need locks, `synchronize` or have to copy the whole data on each change. But I really don't think this is necessary at the moment.

hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
```diff
@@ -171,36 +173,37 @@ public void fromJson(JsonObject json) {
...
         }
 
         @Override
         public JsonObject toJson() {
-            // NOP to read lastAccess to update local thread values
-            if (lastAccessNanos < 0) {
-                assert true;
-            }
+            // read durationMs first to enforce a read barrier
+            int tmpDurationMs = durationMs;
```
mdogan: What's the reason of reading to a tmp variable?

hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
```diff
@@ -208,8 +211,10 @@ public JsonObject toJson() {
         public void fromJson(JsonObject json) {
             id = getInt(json, "id");
             startedAt = getLong(json, "startedAt");
-            durationNanos = getLong(json, "duration");
             lastAccessNanos = System.nanoTime();
+
+            // write durationMs last to enforce a write barrier
+            durationMs = getInt(json, "durationMs");
```
mdogan: I don't understand the reason of mentioning about a memory barrier here. Isn't this just a volatile field initialization? People already know about `volatile` semantics, I guess nothing more special done here...

hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
Donnerbart: I need this value last in the String, but first for the read barrier.

hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
Donnerbart: The write to `durationMs` should piggy back the changes to the other variables, too. And since the order should be important I think it's best to document this, so nobody changes it accidentally.

hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
mdogan: Why not just? 

```java
return "Invocation{" + "id=" + id + ", startedAt=" + startedAt + ", durationMs=" + durationMs + '}';
```

Donnerbart: `id` and `startedAt` are written after the construction of `Invocation` (to could keep the constructor function constant). So they may not be visible for the reader thread without synchronization. This should be done by the volatile `durationMs`.

So I just keep it by the book and always write `durationMs` last and read it first. Maybe this can be optimized further, since `id` and `startedAt` never change, but I'd like to stick with the patterns you can find in the books.

mdogan: To me, visibility of effectively final fields of an object should be guaranteed via safe publication of object reference itself. There should be no need to rely on any read ordering in methods like `toString()` or `toJson()`.

Donnerbart: The values are not set in the constructor, so we can't rely on safe publication. They are set once, but after the construction.

hazelcast/src/main/java/com/hazelcast/spi/impl/SlowOperationLog.java
pveentjer: Just make the field volatile for the reasons mentioned above.

Donnerbart: Which one? `durationMs` **is** already volatile. It is doing the barrier stuff for `id` and `startedAt`.

pveentjer: I'll check the code tomorrow. But I'm talking about lastAccessNanos.

The fact that you are trying to play with fences is not a right move.

pveentjer: Did you read the comments Marko and me have added this morning?

Donnerbart: I had, but I will have to read them again ;)

Regarding `lastAccessNs`: It was originally part of the JSON, this is the reason it was volatile in the first place. It doesn't make sense anymore, I just was too stupid to change the logic accordingly. This should be done in this PR.

Donnerbart:
I just removed the `toString()` methods, which are not used anymore. So a bit less code which we have to care about.

Donnerbart:
While we are arguing about a variable which shouldn't need synchronization (`lastAccessNs`), I fixed a concurrency issue which may be much more important.

The `ConcurrencyUtil.putIfAbsent()` did construct an empty `Invocation` instance, the values were set after it was already in the CHM. If the Management Center would read the data at that point in time, we would get an empty `Invocation` object in the JSON.

Since we have only a single writer we don't need the `putifAbsent()`, so I just used a "check then act" with a normal constructor. This should be fine now, all tests are working locally, including the `TimedMemberStateTest.testSerialization()`.

Donnerbart:
The class `SlowOperationLog` is not exposed to the outside world anymore. Most fields are private now. The only exposed data structure is `Collection<JsonSerializable>`, so it should be very easy to spot which fields need synchronization and which doesn't.

pveentjer:
Make the other long field volatile. Otherwise this PR will not be merged.

pveentjer:
A few more comments after closer inspection of your code.

```
  @Override
        public JsonObject toJson() {
            JsonObject root = new JsonObject();

            // read durationMs first to enforce a read barrier
            root.add("durationMs", durationMs);
            root.add("id", id);
            root.add("startedAt", startedAt);
            return root;
        }
```

There is no reason to add the comment. Even if durationMs would not be volatile this code works fine because the 'invocations' CHM already provides the happens before relation for you. Id/startedAt are effectively immutable because they are set only once and before they are being put in the invocations map, so when this invocation is gotten from this map, its content is correctly published. Therefor there is no reason to rely on durationMs as read barrier to publish these fields. Check 'Memory consistency effects' in the CHM documentation. 

For the same reason the comment about ordering can be removed from this method:

```
 @Override
        public void fromJson(JsonObject json) {
            id = getInt(json, "id");
            startedAt = getLong(json, "startedAt");
            lastAccessNanos = System.nanoTime();

            // write durationMs last to enforce a write barrier
            durationMs = getInt(json, "durationMs");
        }
```
The CHM already provides the happens before relation. This code would work perfectly fine no matter the order of these fields.

But it can even be made simpler:

```
final int id;
final long startedAt;

Invocation(Json json){
    id = getInt(json, "id");
    startedAt = getLong(json, "startedAt");
    lastAccessNanos = System.nanoTime();
   durationMs = getInt(json, "durationMs");
}

```
Now you immediately construct your object correctly. You can make final whatever you can make final.

This code will still work fine without finals and also ordering is irrelevant because of the memory consistency effects of the CHM. As long as you don't touch these fields after you put the object in CHM, you are perfectly fine. 

```
 void purgeInvocations(long nowNanos, long slowOperationLogLifetimeNanos) {
        // method is only called by single SlowOperationDetectorThread, no need for synchronization
        for (Map.Entry<Integer, Invocation> invocationEntry : invocations.entrySet()) {
            if (nowNanos - invocationEntry.getValue().lastAccessNanos > slowOperationLogLifetimeNanos) {
                invocations.remove(invocationEntry.getKey());
            }
        }
    }
```

You say that no synchronization is needed. Perhaps that invocations is only updated from one thread, there could be concurrent readers. The CHM provides this synchronization; so it is fails to claim that no need for synchronization is needed; only it is done for you. So please remove that section.

You also need to document that the 'update' method is only called by the SlowOperationDetectorThread and the same goes for any other method in the SlowOperationLog that is expected to be touched by only this thread.

Another thing I would improve is that you rely on stuff not to be touched by multiple threads, I would try to make it as private as possible. This way you don't need to inspect a surrounding context to see if someone has violated your limitations (either now or in the future).

This being said; the lastAccessNanos can remain non volatile. It isn't being used apart from the SlowOperationThread.

Donnerbart:
Thank you all for the reviews. I really have to get a rubber duck ;)

Your last comment was very helpful, Peter. Thanks for that. The JSON constructor was a very good idea. It was good to step back and have a look at the whole class, instead of changing single lines due to line comments.

I don't see any more methods or fields which I can make private (mostly everything is package private). The only "public" accessible method should be `Collection<JsonSerializable> getSlowOperationLogs()` with its public `fromJSON()` and `toJSON()` (via the `InternalOperationService` interface).

Donnerbart:
verify

PR #4726, opened 2015-03-04
Introduced com.hazelcast.internal package

pveentjer:
Introduced the com.hazelcast.internal package that is going to contain internal API/packages. For example management package is an internal package and nobody should make use of its classes directly.

But hiding it in the internal package
- the root package remains a lot cleaner and easier to understand for the end user. If it is in the root, a user can play with it.
- because we can properly hide it, we don't need to provide any backwards compatibility to these packages since they are not a public API.

I have currently moved com.hazelcast.management and com.hazelcast.osgi into com.hazelcast.internal, but other modules will follow. But don't want to make this PR too big.

Of course we need to make very sure that we don't break anything by doing this!


mdogan:
looks good to me. not sure but may break the management center if it's importing any of classes in `management` package. @emrahkocaman?

pveentjer:
It is using this package e.g. json serializable.

mesutcelik:
I fixed the compile problems as it has been failing since yesterday. https://github.com/hazelcast/management-center/commit/f9342b46a919933987eb1923399df787b0c4bfa4

PR #4691, opened 2015-02-27
[DON'T MERGE] Use separate objects for instant stats and on-demand calculated stats

metanet:
Use separate objects for instant stats and on-demand calculated stats for local map stats

Fixes #4497

PR #4736, opened 2015-03-04
Fix local map stats thread safety problems without breaking the API

### LINKED ISSUE #4497: IMap.getLocalMapStats() not thread safe and returns wrong values of some properties
If more then one threads invoke getLocalMapStats() on the same IMap, then we have a concurrent modification of some fields of one instance of LocalMapStatsImpl object.

call sequence pseudo code is:

```java
IMapProxyImpl(MapProxySupport).getLocalMapStats();
LocalMapStatsProvider.createLocalMapStats(String mapName) {

    final LocalMapStatsImpl localMapStats = getLocalMapStatsImpl(mapName) {
        return ConcurrencyUtil.getOrPutIfAbsent(statsMap, name, constructorFunction);
    }

    localMapStats.init() {
        ownedEntryCount = 0;
        backupEntryCount = 0;
        ownedEntryMemoryCost = 0;
        backupEntryMemoryCost = 0;
        heapCost = 0;
        lockedEntryCount = 0;
        dirtyEntryCount = 0;
        backupCount = 0;
        hits = 0;
    }    

    addOwnerPartitionStats(localMapStats, mapName, partitionId) {
        localMapStats.incrementOwnedEntryMemoryCost(ownedEntryMemoryCost);
        localMapStats.incrementLockedEntryCount(lockedEntryCount);
        localMapStats.incrementHits(hits);
        localMapStats.incrementDirtyEntryCount(recordStore.getMapDataStore().notFinishedOperationsCount());
        localMapStats.incrementHeapCost(recordStore.getHeapCost());
        localMapStats.incrementOwnedEntryCount(recordStore.size());
    }
}
```

metanet:
* This change has an effect on the API.
 * Before this fix, a LocalMapStats object returned to the user is effected from subsequent map operations.
 * After this fix, stats objects are snapshots and not-updated after returned to the user.

 Fixes #4497

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
@@ -1089,7 +1089,6 @@ public class ClientMapBasicTest {
     @Test
     public void testMapStatistics_withClientOperations() {
         final String mapName = randomString();
-        final LocalMapStats serverMapStats = server.getMap(mapName).getLocalMapStats();
 
         final IMap map = client.getMap(mapName);
         final int operationCount = 1123;
@@ -1098,6 +1097,8 @@ public class ClientMapBasicTest {
             map.get(i);
             map.remove(i);
         }
+
+        final LocalMapStats serverMapStats = server.getMap(mapName).getLocalMapStats();
 
         assertEquals("put count", operationCount, serverMapStats.getPutOperationCount());
         assertEquals("get count", operationCount, serverMapStats.getGetOperationCount());
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
@@ -1089,16 +1089,15 @@ public class ClientMapBasicTest {
     @Test
     public void testMapStatistics_withClientOperations() {
         final String mapName = randomString();
-
         final IMap map = client.getMap(mapName);
+        final LocalMapStats serverMapStats = server.getMap(mapName).getLocalMapStats();
+
         final int operationCount = 1123;
         for (int i = 0; i < operationCount; i++) {
             map.put(i, i);
             map.get(i);
             map.remove(i);
         }
-
-        final LocalMapStats serverMapStats = server.getMap(mapName).getLocalMapStats();
 
         assertEquals("put count", operationCount, serverMapStats.getPutOperationCount());
         assertEquals("get count", operationCount, serverMapStats.getGetOperationCount());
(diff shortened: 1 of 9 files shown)


pveentjer:
Isn't this a breakage of the API?

And we are we sending back a new object? Why is it so bad to share the counters directly? With this approach you also get more litter.

metanet:
Yes, it is. For that reason, I stated that in the commit message.
Sharing an updated object is not the main problem. The problem is, stats objects contain fields that is calculated on-demand. A single object is accessed by multiple threads and modified in a non-thread-safe manner. There is no way to pass the same object to multiple threads properly and safely. Therefore, thread-confined objects are created. This PR creates a new object by copying all of the fields.
The proper fix for #4497 is the PR #4691 but it contains too many refactorings. It uses the same single object for thread-safely calculated stats and shares it by putting it into a thread-confined object. Main goal of this PR is providing the simplest solution and the consequences are stated.

pveentjer:
For simple statistics in most cases atomic classes work fine for me. Why can't that be done here?

metanet:
Some stats are not kept as atomic values, they are calculated on-demand by iterating through records etc. for many partitions. I don't know the main reason of this but my first guess is preventing the contention while stat updates.

pveentjer:
Sorry for being a pain in the ass about this; but if the API silently breaks or if we are going to generate litter for the sake of statistics, then I'm on it.

metanet:
The other PR does not break the API and it solves the problem with minimum extra cost.

pveentjer:
It breaks the API in the sense that a user doesn't get an update when he keeps reference to a statistic object. So something that used to work in the past, but not anymore. That is why I call it silent breaking. 

And minimal extra cost ---> creating an object to access statistics I don't consider minimal. In most cases it is one of the main things I want to prevent happening when dealing with statistics.

metanet:
You are right. It breaks it silently. But the previous version was already providing wrong values for some stats. So, i think it is still arguable. 

Most of the statistics are kept in a singleton object (once created, updates are always on the same object). Some other statistics are calculated on demand by design. If you stick to the existing design and  if you want to provide those values to user correctly, your only choice is creating a new object for that on-demand values. For the other singleton object, you can just copy its reference to the new object. Therefore, your cost is creating a new object with only on-demand fields + an extra field that points the other singleton stats object.

For my point of view, I expect a stat object to be a snapshot. Once I get it from some api, I do not expect it to change. But it is not the case for the existing API and I am OK with it. Nevertheless, creating a new object for providing statistics seems OK to me since people usually get stats periodically.

metanet:
https://github.com/metanet/hazelcast/commit/97f3b4260ee77476ac25d99fff3126c1f1bae307 does not break the API

gurbuzali:
:+1:

gurbuzali:
@emrahkocaman can you check also ?

emrahkocaman:
:+1:

ahmetmircik:
:+1:

PR #4747, opened 2015-03-05
backport to generate thread dump (with their locks) on test failure before "tearDown"

serkan-ozal:
Wraps JUnit `Statement` and in case of error (also before `tearDown`of test) dumps threads.


gurbuzali:
forward-port already merged

PR #4749, opened 2015-03-06
Improved collection testing

pveentjer:
Improved testing for collection classes. Tests are repeated for local and distributed setup to make sure that both scenario's work. Also added migration tests to verify that partition migration works. 

Added constructor to configs so you can create them easily with a name new ListConfig("foo")..


hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
```diff
@@ -209,7 +208,7 @@ public static String randomNameOwnedBy(HazelcastInstance instance, String prefix
         for (; ; ) {
             String id = prefix + randomString();
             Partition partition = partitionService.getPartition(id);
-            if (comparePartitionOwnership(false, localMember, partition)) {
+            if (comparePartitionOwnership(true, localMember, partition)) {
```
jerrinot: so the other remote/local tests are not working as expected until this is merged?

jerrinot:
:+1: (once the PR builder is green)

hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
pveentjer: I already fixed that in the master. My test was broken and I thought it was caused by this method being broken.

pveentjer: Nobody else is using this method since I introduced it yesterday. I can now retrieve the name of the executing test:

```
 @Before
    public void setup() {
        Config config = new Config();
        config.addListConfig(new ListConfig("testAdd_whenCapacityReached_thenItemNotAdded*").setMaxSize(10));

        instances = newInstances(config);
        HazelcastInstance local = instances[0];
        HazelcastInstance target = instances[instances.length - 1];
        String methodName = getThreadLocalFrameworkMethod().getMethod().getName();
        String name = randomNameOwnedBy(target, methodName);
        list = local.getList(name);
    }
```

So depending on the test you are executing, you get the right configured element.

jerrinot: why not have something like `boolean isOwnedBy(Partition p, Member m);` instead of the `comparePartitionOwnership` ? 

I found the `comparePartitionOwnership()` a bit confusing -> that's likely a root cause of the wrong boolean flag.

pveentjer: Yes; I also found it confusing. I'll see if I'll clean it up. For the time being it is working

PR #4272, opened 2014-12-10
Enable partition events for partitions that are promoted from backup.

ariordan:
As discussed in issue #2318, enables partition events for partitions that are promoted from backup.

Note: Each events MigrationInfo.source field is set to null.

PR #4752, opened 2015-03-06
enable migration events for promoted-from-backup partitions

### LINKED ISSUE #2318: Migration listener methods (migrationStarted, migrationCompleted, migrationFailed) don't fire when member down
Hi,
I have a distributed application that uses an Hazelcast map and registers a migration listener with Hazelcast's partition service.

When a new application node starts, the set of local map keys on the existing nodes becomes smaller - as expected, because the new node takes over some of the keys - and the migration listener migrationStarted and migrationCompleted methods fire - also as expected.

When a node is stopped, the set of local map keys becomes larger - as expected - however none of the callback methods fires.

Is this by design? I'm using Hazelcast 3.2.

Thank you in advance.

gurbuzali:
back-port of https://github.com/hazelcast/hazelcast/pull/4272
fixes #2318

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
@@ -426,7 +426,7 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
             }
 
             if (promote) {
-                final Operation op = new PromoteFromBackupOperation();
+                final Operation op = new PromoteFromBackupOperation(deadAddress);
                 op.setPartitionId(partition.getPartitionId())
                         .setNodeEngine(nodeEngine)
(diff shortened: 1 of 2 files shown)


jerrinot:
I know this is just a backport, but what if the promotion fail? should we send MigrationStatus.Failed?

gurbuzali:
promotion is just updating the partition table, no data is replicated/moved.
it is just saying the services that "you have some data, they are not backups anymore, they are primary data from now on so do some stuff if necessary"

if you look at the `PromoteFromBackupOperation` all exceptions are just logged as warning.
So the question is: should we send failed event even if we get exception from only one service ?
the other services are promoted successfully.

jerrinot:
I think this ticket can be merged as it is, hence :+1: 

The exact semantic of MigrationStatus (COMPLETED/FAILED) is a very interesting question indeed. Let's discuss it next week!

PR #4761, opened 2015-03-06
Fix/master/issue 4757

serkan-ozal:


pveentjer:
I don't think this PR is needed. The following PR contains a better termination mechanism.

https://github.com/hazelcast/hazelcast/pull/4762

serkan-ozal:
@pveentjer for PR #4762, there is no extra added value with respect to this PR from the ```SlowOperationDetectorThread``` and ```CleanupThread``` termination perspectives. Yes, it includes some refactoring but indeed its first version has worse termination since in case of shutdown, there are unnecessary operations to finish the loop and running check is handled at next iteration so there is unnecessary wait. 

BTW this PR is just about ```SlowOperationDetectorThread``` and ```CleanupThread``` termination thread, not general refactor about operation termination.

pveentjer:
I'll merge it, but I'll apply my cleanups.

gurbuzali:
@pveentjer can you also merge back-port #4760 too and apply your cleanups if necessary

PR #4762, opened 2015-03-06
Improved OperationService termination

### LINKED ISSUE #4757: While executing unit tests, SlowOperationDetectorThread and CleanupThread may not be terminated before next test started
In the ```shutdown``` methods, these threads are not waited to finish their works and so they may continue to running on next test(s)

As seen from the logs below, these threads (from previous tests) may work at next tests

````
"hz._hzInstance_29_dev.migration" 
	java.lang.Thread.State: TIMED_WAITING on lock=java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject@bd5d765
		at sun.misc.Unsafe.park(Native Method)
		at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:196)
		at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(AbstractQueuedSynchronizer.java:2025)
		at java.util.concurrent.LinkedBlockingQueue.poll(LinkedBlockingQueue.java:424)
		at com.hazelcast.partition.impl.InternalPartitionServiceImpl$MigrationThread.doRun(InternalPartitionServiceImpl.java:1778)
		at com.hazelcast.partition.impl.InternalPartitionServiceImpl$MigrationThread.run(InternalPartitionServiceImpl.java:1760)

...

"hz._hzInstance_28_dev.CleanupThread" 
	java.lang.Thread.State: TIMED_WAITING
		at java.lang.Thread.sleep(Native Method)
		at com.hazelcast.spi.impl.BasicOperationService$CleanupThread.sleep(BasicOperationService.java:1114)
		at com.hazelcast.spi.impl.BasicOperationService$CleanupThread.run(BasicOperationService.java:1103)

"hz._hzInstance_28_dev.SlowOperationDetectorThread" 
	java.lang.Thread.State: TIMED_WAITING
		at java.lang.Thread.sleep(Native Method)
		at java.lang.Thread.sleep(Thread.java:302)
		at java.util.concurrent.TimeUnit.sleep(TimeUnit.java:328)
		at com.hazelcast.spi.impl.SlowOperationDetector$SlowOperationDetectorThread.sleepInterval(SlowOperationDetector.java:218)
		at com.hazelcast.spi.impl.SlowOperationDetector$SlowOperationDetectorThread.run(SlowOperationDetector.java:122)
```

There must not be any thread like ```hz._hzInstance_28***```.

pveentjer:
Fixes #4757

The BasicOperationService was termination its threads too slowly, and this can lead to unexpected behavior when terminating and starting instances.

This pr contains a cleanup of the termination of OperationService cleanupthread, slow operation detector and operation executor. Also the SlowOperationDetector was moved to its own package to reduce the amount of noise.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
@@ -1141,12 +1141,17 @@ public final class BasicOperationService implements InternalOperationService {
             }
 
             for (BasicInvocation invocation : invocations.values()) {
+                if(shutdown){
+                    return;
+                }
+
                 try {
                     invocation.handleOperationTimeout();
                 } catch (Throwable t) {
                     inspectOutputMemoryError(t);
                     logger.severe("Failed to handle operation timeout of invocation:" + invocation, t);
                 }
+
                 try {
                     invocation.handleBackupTimeout(backupOperationTimeoutMillis);
                 } catch (Throwable t) {
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
@@ -55,6 +55,8 @@ import com.hazelcast.spi.impl.operationexecutor.OperationRunnerFactory;
 import com.hazelcast.spi.impl.operationexecutor.OperationRunner;
 import com.hazelcast.spi.impl.operationexecutor.ResponsePacketHandler;
 import com.hazelcast.spi.impl.operationexecutor.classic.ClassicOperationExecutor;
+import com.hazelcast.spi.impl.slowoperationdetector.SlowOperationDetector;
+import com.hazelcast.spi.impl.slowoperationdetector.SlowOperationLog;
 import com.hazelcast.util.EmptyStatement;
 import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.executor.ExecutorType;
@@ -72,6 +74,7 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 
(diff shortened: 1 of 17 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
```diff
@@ -1102,13 +1115,17 @@ public void run() {
                     backPressureService.cleanup();
```
serkan-ozal: in case of shutdown, did we really need to wait ```scanHandleOperationTimeout```to finish its work. In case of shutdown, it should break ```for``` loop in itself

pveentjer: good idea

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
```diff
@@ -1090,6 +1101,8 @@ private void assertNoBackupOnPrimaryMember(InternalPartition partition, Address
 
         public static final int DELAY_MILLIS = 1000;
 
+        private volatile boolean shutdown;
```
serkan-ozal: we are interrupting thread, so is it really needed ```shutdown``` field. how about ``ìsInterrupted()``` ?

hazelcast/src/main/java/com/hazelcast/util/ThreadUtil.java
```diff
@@ -16,17 +16,46 @@
...
+     */
+    public static void awaitTermination(long timeoutMs, Thread... threads) throws InterruptedException, TimeoutException {
+        for (Thread thread : threads) {
+            thread.join(timeoutMs);
+
+            if (!thread.isAlive()) {
+                continue;
+            }
+
+            throw new TimeoutException(format("Thread %s failed to complete within %s ms", thread.getName(), timeoutMs));
```
serkan-ozal: for only one non-terminated threads, the other threads also are not waited to finish since method is returned with exception.

pveentjer: Yes. What is the point for waiting for other threads if this one fails?

pveentjer: Because this already is a problematic situation you don't want to have.. so fail fast.. no point in waiting for the rest imho

hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationService.java
pveentjer: Problem with relying on interrupt flag is that the interrupt could have been consumed somewhere and then you don't see it. This is imho more reliable.

gurbuzali:
needs rebase

Issue #3661, opened 2014-09-24
Queues are very very slow


lvca:
Hi guys,
I'm struggling to improve performance of HZ Queues, but with no success. If I've 1 producer + 1 consumer on separate JVMs, but on the same server, I cannot go faster than 1,5k messages per second. I found an old thread (2011) about this topic and it reported similar values.

By a quick profiling using YourKit I see most of the time is spent in KQueueArrayWrapper.kevent0() mehod.

Do you know any trick to speed up queues?

bwzhang2011:
@pveentjer, how does this going on about queue performance once more nodes join ?  as we know, queue is widely used in our project and we do care about its performance and expansibility. hope @pveentjer provides more solution for such and we could not be waiting 3.5 forthcoming as more issue integrated into such version.

RituBathla:
When I implement Hazelcast in my code, it slows down performance significantly.
This is my implementation of Hazelcast:
public class HazelcastComponent
{
    @Value("${cluster.name:DefaultCluster}")
    private String clusterName;

    @PostConstruct
    void init()
    {
        Config config = new Config();
        config.setInstanceName(clusterName);
        config.getNetworkConfig().setPort(NetworkConfig.DEFAULT_PORT);
        config.getNetworkConfig().setPortAutoIncrement(true);

        NetworkConfig network = config.getNetworkConfig();
        JoinConfig join = network.getJoin();
        join.getMulticastConfig().setEnabled(false);
        join.getTcpIpConfig().addMember("localhost").setEnabled(true);

        hazelCast = Hazelcast.newHazelcastInstance(config);
    }

    /**
     * @return The Hazelcast instance.
     */
    public HazelcastInstance getInstance()
    {
        return hazelCast;
    }
}

I'm using Hazelcast 3.2.4 and Websphere 8.5 as my application server.
Please let me know if you need more details regarding the same.

Thanks

pveentjer:
I have no idea which features of HZ you are using to explain what is the cause. So are you using a map/queue etc? Best to create a test which reproduces the issue you are having. Also: slows down considerably... compared to? It slows down over time? It is slow compared to e.g. a concurrent map?

RituBathla:
It slows down considerably as compared to when I remove Hazelcast code.
Using a queue:

BlockingQueue<String> definitionIdQueue = hazelCast.getInstance().getQueue("MODEL_DEFINITION_ID_QUEUE");

[...]

PR #4773, opened 2015-03-09
Ringbuffer

pveentjer:
This is a cleaned up PR. The reliable topic has moved to its own PR.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -142,6 +142,14 @@
     <suppress checks="MethodCount" files="com.hazelcast.concurrent.lock.LockStoreImpl"/>
     <suppress checks="Todo" files="com.hazelcast.concurrent.lock.ConditionImpl"/>
 
+    <!-- ringbuffer -->
+    <suppress checks="JavadocMethod" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocType" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocMethod" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocVariable" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocPackage" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocVariable" files="com.hazelcast.ringbuffer.impl.RingbufferDataSerializerHook"/>
+
     <!-- Storage -->
     <suppress checks="JavadocMethod" files="com.hazelcast.internal.storage[\\/]"/>
     <suppress checks="JavadocType" files="com.hazelcast.internal.storage[\\/]"/>
--- a/hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
@@ -1,0 +1,322 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.config;
+
+import com.hazelcast.spi.annotation.Beta;
+
+import static com.hazelcast.config.InMemoryFormat.NATIVE;
+import static com.hazelcast.util.Preconditions.checkAsyncBackupCount;
+import static com.hazelcast.util.Preconditions.checkBackupCount;
+import static com.hazelcast.util.Preconditions.checkFalse;
+import static com.hazelcast.util.Preconditions.checkNotNegative;
+import static com.hazelcast.util.Preconditions.checkNotNull;
+import static com.hazelcast.util.Preconditions.checkPositive;
+
+/**
+ * Contains the configuration for the {@link com.hazelcast.ringbuffer.Ringbuffer}.
+ * <p/>
+ * The RingBuffer is currently not a distributed data-structure, so its content will be fully stored on a single member
+ * in the cluster and its backup in another member in the cluster.
+ */
+@Beta
+public class RingbufferConfig {
+
+    /**
+     * Default value of capacity of the RingBuffer.
+     */
+    public static final int DEFAULT_CAPACITY = 10 * 1000;
+    /**
+     * Default value of synchronous backup count
+     */
+    public static final int DEFAULT_SYNC_BACKUP_COUNT = 1;
+    /**
+     * Default value of asynchronous backup count
+     */
+    public static final int DEFAULT_ASYNC_BACKUP_COUNT = 0;
+    /**
+     * Default value for the time to live property.
+     */
+    public static final int DEFAULT_TTL_SECONDS = 0;
+    /**
+     * Default value for the InMemoryFormat.
+     */
+    public static final InMemoryFormat DEFAULT_IN_MEMORY_FORMAT = InMemoryFormat.BINARY;
+
+    private String name;
+    private int capacity = DEFAULT_CAPACITY;
+    private int backupCount = DEFAULT_SYNC_BACKUP_COUNT;
+    private int asyncBackupCount = DEFAULT_ASYNC_BACKUP_COUNT;
+    private int timeToLiveSeconds = DEFAULT_TTL_SECONDS;
+    private InMemoryFormat inMemoryFormat = DEFAULT_IN_MEMORY_FORMAT;
+
+    /**
+     * Creates a RingbufferConfig with the provided name.
+     *
+     * @param name the name
+     * @throws java.lang.NullPointerException if name is null
+     */
+    public RingbufferConfig(String name) {
+        this.name = checkNotNull(name, "name can't be null");
+    }
+
+    /**
+     * Clones a RingbufferConfig
+     *
+     * @param config the ringbuffer config to clone
+     * @throws java.lang.NullPointerException if config is null
+     */
+    public RingbufferConfig(RingbufferConfig config) {
+        checkNotNull(config, "config can't be null");
+        this.name = config.name;
+        this.capacity = config.capacity;
+        this.backupCount = config.backupCount;
+        this.asyncBackupCount = config.asyncBackupCount;
+        this.timeToLiveSeconds = config.timeToLiveSeconds;
+        this.inMemoryFormat = config.inMemoryFormat;
+    }
+
+    /**
+     * Creates a new RingbufferConfig by cloning an existing config and overriding the name.
+     *
+     * @param name   the new name
+     * @param config the config.
+     * @throws java.lang.NullPointerException if name or config is null.
+     */
+    public RingbufferConfig(String name, RingbufferConfig config) {
+        this(config);
+        this.name = checkNotNull(name, "name can't be null");
+    }
+
+
+    /**
+     * Returns the name of the ringbuffer.
+     *
+     * @return the name of the ringbuffer.
+     */
+    public String getName() {
+        return name;
+    }
+
+
+    /**
+     * Gets the capacity of the ringbuffer.
+     * <p/>
+     * The capacity is the total number of items in the ringbuffer. The items will remain in the ringbuffer, but the oldest items
+     * will eventually be be overwritten by the newest items.
+     * <p/>
+     * In the future we'll add more advanced policies e.g. based on memory usage or lifespan.
+     *
+     * @return the capacity.
+     */
+    public int getCapacity() {
+        return capacity;
+    }
+
+    /**
+     * Sets the capacity of the ringbuffer.
+     *
+     * @param capacity the capacity.
+     * @return the updated Config.
+     * @throws java.lang.IllegalArgumentException if capacity smaller than 1.
+     * @see #getCapacity()
+     */
+    public RingbufferConfig setCapacity(int capacity) {
+        checkPositive(capacity, "capacity can't be smaller than 1");
+        this.capacity = capacity;
+        return this;
+    }
+
+    /**
+     * Gets the number of synchronous backups.
+     *
+     * @return number of synchronous backups.
+     */
+    public int getBackupCount() {
+        return backupCount;
+    }
+
+    /**
+     * Sets the number of synchronous backups.
+     *
+     * @param backupCount the number of synchronous backups to set
+     * @return the updated SemaphoreConfig
+     * @throws IllegalArgumentException if backupCount smaller than 0,
+     *                                  or larger than the maximum number of backup
+     *                                  or the sum of the backups and async backups is larger than the maximum number of backups
+     * @see #setAsyncBackupCount(int)
+     * @see #getBackupCount()
+     */
+    public RingbufferConfig setBackupCount(int backupCount) {
+        this.backupCount = checkBackupCount(backupCount, asyncBackupCount);
+        return this;
+    }
+
+    /**
+     * Gets the number of asynchronous backups.
+     *
+     * @return the number of asynchronous backups.
+     */
+    public int getAsyncBackupCount() {
+        return asyncBackupCount;
+    }
+
+    /**
+     * Sets the number of asynchronous backups. 0 means no backups
+     *
+     * @param asyncBackupCount the number of asynchronous synchronous backups to set
+     * @return the updated SemaphoreConfig
+     * @throws IllegalArgumentException if asyncBackupCount smaller than 0,
+     *                                  or larger than the maximum number of backup
+     *                                  or the sum of the backups and async backups is larger than the maximum number of backups
+     * @see #setBackupCount(int)
+     * @see #getAsyncBackupCount()
+     */
+    public RingbufferConfig setAsyncBackupCount(int asyncBackupCount) {
+        this.asyncBackupCount = checkAsyncBackupCount(backupCount, asyncBackupCount);
+        return this;
+    }
+
+    /**
+     * Returns the total number of backups: backupCount plus asyncBackupCount.
+     *
+     * @return the total number of backups: backupCount plus asyncBackupCount
+     */
+    public int getTotalBackupCount() {
+        return backupCount + asyncBackupCount;
+    }
+
+    /**
+     * Creates a readonly copy of this RingbufferConfig.
+     *
+     * @return the readonly copy.
+     */
+    public RingbufferConfig getAsReadOnly() {
+        return new RingbufferConfigReadonly(this);
+    }
+
+    /**
+     * Gets the time to live in seconds.
+     *
+     * @return the time to live in seconds. Returns 0 the time to live if the items don't expire.
+     */
+    public int getTimeToLiveSeconds() {
+        return timeToLiveSeconds;
+    }
+
+    /**
+     * Sets the time to live in seconds.
+     *
+     * Time to live is the time the ringbuffer is going to retain items before deleting them.
+     *
+     * Time to live can be disabled by setting timeToLiveSeconds to 0. It means that items won't get removed because they
+     * retire. They will only overwrite. This means that when timeToLiveSeconds is disabled, that after tail did a full
+     * loop in the ring that the size will always be equal to the capacity.
+     *
+     * @param timeToLiveSeconds the time to live period in seconds
+     * @return the updated RingbufferConfig
+     * @throws IllegalArgumentException if timeToLiveSeconds smaller than 0.
+     */
+    public RingbufferConfig setTimeToLiveSeconds(int timeToLiveSeconds) {
+        this.timeToLiveSeconds = checkNotNegative(timeToLiveSeconds, "timeToLiveSeconds can't be smaller than 0");
+        return this;
+    }
+
+    /**
+     * Gets the InMemoryFormat.
+     *
+     * @return the InMemoryFormat.
+     */
+    public InMemoryFormat getInMemoryFormat() {
+        return inMemoryFormat;
+    }
+
+    /**
+     * Sets the InMemoryFormat.
+     *
+     * Setting the InMemoryFormat controls format of storing an item in the ringbuffer:
+     * <ol>
+     *     <li>{@link InMemoryFormat#OBJECT}: the item is stored in deserialized format (so a regular object)</li>
+     *     <li>{@link InMemoryFormat#BINARY}: the item is stored in serialized format (so a is binary blob) </li>
+     * </ol>
+     *
+     * The default is binary. The object InMemoryFormat is useful when:
+     * <ol>
+     *     <li>of the object stored in object format has a smaller footprint than in binary format</li>
+     *     <li>if there are readers using a filter. Since for every filter invocation, the object needs to be available in
+     *     object format.</li>
+     * </ol>
+     *
+     * @param inMemoryFormat the new in memory format.
+     * @return the updated Config.
+     * @throws NullPointerException if inMemoryFormat is null.
+     * @throws IllegalArgumentException if {@link InMemoryFormat#NATIVE} in memory format is selected.
+     */
+    public RingbufferConfig setInMemoryFormat(InMemoryFormat inMemoryFormat) {
+        checkNotNull(inMemoryFormat, "inMemoryFormat can't be null");
+        checkFalse(inMemoryFormat == NATIVE, "InMemoryFormat " + NATIVE + " is not supported");
+        this.inMemoryFormat = inMemoryFormat;
+        return this;
+    }
+
+    @Override
+    public String toString() {
+        return "RingbufferConfig{"
+                + "name='" + name + '\''
+                + ", capacity=" + capacity
+                + ", backupCount=" + backupCount
+                + ", asyncBackupCount=" + asyncBackupCount
+                + ", timeToLiveSeconds=" + timeToLiveSeconds
+                + ", inMemoryFormat=" + inMemoryFormat
+                + '}';
+    }
+
+    /**
+     * A readonly version of the {@link RingbufferConfig}.
+     */
+    @Beta
+    static class RingbufferConfigReadonly extends RingbufferConfig {
+
+        public RingbufferConfigReadonly(RingbufferConfig config) {
+            super(config);
+        }
+
+        @Override
+        public RingbufferConfig setCapacity(int capacity) {
+            throw new UnsupportedOperationException("This config is read-only");
+        }
+
+        @Override
+        public RingbufferConfig setAsyncBackupCount(int asyncBackupCount) {
+            throw new UnsupportedOperationException("This config is read-only");
+        }
+
+        @Override
+        public RingbufferConfig setBackupCount(int backupCount) {
+            throw new UnsupportedOperationException("This config is read-only");
+        }
+
+        @Override
+        public RingbufferConfig setTimeToLiveSeconds(int timeToLiveSeconds) {
+            throw new UnsupportedOperationException("This config is read-only");
+        }
+
+        @Override
+        public RingbufferConfig setInMemoryFormat(InMemoryFormat inMemoryFormat) {
+            throw new UnsupportedOperationException("This config is read-only");
+        }
+    }
+}
--- a/hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
@@ -69,6 +69,7 @@ import static com.hazelcast.config.XmlElements.PROPERTIES;
 import static com.hazelcast.config.XmlElements.QUEUE;
 import static com.hazelcast.config.XmlElements.QUORUM;
 import static com.hazelcast.config.XmlElements.REPLICATED_MAP;
+import static com.hazelcast.config.XmlElements.RINGBUFFER;
 import static com.hazelcast.config.XmlElements.SECURITY;
 import static com.hazelcast.config.XmlElements.SEMAPHORE;
 import static com.hazelcast.config.XmlElements.SERIALIZATION;
@@ -287,6 +288,8 @@ public class XmlConfigBuilder extends AbstractConfigBuilder implements ConfigBui
             handleJobTracker(node);
         } else if (SEMAPHORE.isEqual(nodeName)) {
             handleSemaphore(node);
+        } else if (RINGBUFFER.isEqual(nodeName)) {
+            handleRingbuffer(node);
         } else if (LISTENERS.isEqual(nodeName)) {
             handleListeners(node);
         } else if (PARTITION_GROUP.isEqual(nodeName)) {
@@ -1427,6 +1430,33 @@ public class XmlConfigBuilder extends AbstractConfigBuilder implements ConfigBui
             }
         }
         config.addSemaphoreConfig(sConfig);
+    }
+
+    private void handleRingbuffer(Node node) {
+        Node attName = node.getAttributes().getNamedItem("name");
+        String name = getTextContent(attName);
+        RingbufferConfig rbConfig = new RingbufferConfig(name);
+        for (Node n : new IterableNodeList(node.getChildNodes())) {
+            String nodeName = cleanNodeName(n.getNodeName());
+            String value = getTextContent(n).trim();
+            if ("capacity".equals(nodeName)) {
+                int capacity = getIntegerValue("capacity", value, RingbufferConfig.DEFAULT_CAPACITY);
+                rbConfig.setCapacity(capacity);
+            } else if ("backup-count".equals(nodeName)) {
+                int backupCount = getIntegerValue("backup-count", value, RingbufferConfig.DEFAULT_SYNC_BACKUP_COUNT);
+                rbConfig.setBackupCount(backupCount);
+            } else if ("async-backup-count".equals(nodeName)) {
+                int asyncBackupCount = getIntegerValue("async-backup-count", value, RingbufferConfig.DEFAULT_ASYNC_BACKUP_COUNT);
+                rbConfig.setAsyncBackupCount(asyncBackupCount);
+            } else if ("time-to-live-seconds".equals(nodeName)) {
+                int timeToLiveSeconds = getIntegerValue("time-to-live-seconds", value, RingbufferConfig.DEFAULT_TTL_SECONDS);
+                rbConfig.setTimeToLiveSeconds(timeToLiveSeconds);
+            } else if ("in-memory-format".equals(nodeName)) {
+                InMemoryFormat inMemoryFormat = InMemoryFormat.valueOf(upperCaseInternal(value));
+                rbConfig.setInMemoryFormat(inMemoryFormat);
+            }
+        }
+        config.addRingBufferConfig(rbConfig);
     }
 
     private void handleListeners(final org.w3c.dom.Node node) throws Exception {
--- a/hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
+++ b/hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.ringbuffer;
+
+import com.hazelcast.spi.annotation.Beta;
+
+/**
+ * The result of a {@link Ringbuffer#readManyAsync(long, int, int, com.hazelcast.core.IFunction)} operation.
+ *
+ * Important:
+ * If an item is retrieved multiple times from the result set, a new instance is returned for every invocation. This is done
+ * to prevent unexpected sharing if the {@link com.hazelcast.core.ICompletableFuture} is shared between multiple threads.
+ *
+ * @param <E>
+ */
+@Beta
+public interface ReadResultSet<E> extends Iterable<E> {
+
+    /**
+     * Returns the number of items that have been read before filtering.
+     *
+     * If no filter is set, then the readCount will be the same as size. But if a filter is applied, it could be that items
+     * are read, but are filtered out. So if you are trying to make another read based on the ReadResultSet then you should
+     * increment the sequence by readCount and not by size. Otherwise you will be re-reading the same filtered messages.
+     *
+     * @return the number of items read (including the filtered ones).
+     */
+    int readCount();
+
+    /**
+     * Gets the item at the given index.
+     *
+     * @param index the index
+     * @return the found item.
+     * @throws IllegalArgumentException if index out of bounds.
+     */
+    E get(int index);
+}
--- a/hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
+++ b/hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
@@ -1,0 +1,258 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.ringbuffer;
+
+import com.hazelcast.core.DistributedObject;
+import com.hazelcast.core.ICompletableFuture;
+import com.hazelcast.core.IFunction;
+import com.hazelcast.spi.annotation.Beta;
+
+import java.util.Collection;
+
+/**
+ * A Ringbuffer is a data-structure where the content is stored in a ring like structure. A ringbuffer has a capacity so it
+ * won't grow beyond that capacity and endanger the stability of the system. If that capacity is exceeded, than the oldest
+ * item in the ringbuffer is overwritten.
+ *
+ * The ringbuffer has 2 always incrementing sequences:
+ * <ol>
+ * <li>
+ * tailSequence: this is the side where the youngest item is found. So the tail is the side of the ringbuffer where
+ * items are added to.
+ * </li>
+ * <li>
+ * headSequence: this is the side where the oldest items are found. So the head is the side where items gets
+ * discarded.
+ * </li>
+ * </ol>
+ * The items in the ringbuffer can be found by a sequence that is in between (inclusive) the head and tail sequence.
+ *
+ * If data is read from a ringbuffer with a sequence that is smaller than the headSequence, it means that the data
+ * is not available anymore and a {@link StaleSequenceException} is thrown.
+ *
+ * A Ringbuffer currently is not a distributed data-structure. So all data is stored in a single partition; comparable to the
+ * IQueue implementation. But we'll provide an option to partition the data in the near future.
+ *
+ * A Ringbuffer can be used in a similar way as a queue, but one of the key differences is that a queue.take is destructive,
+ * meaning that only 1 thread is able to take an item. A ringbuffer.read is not destructive, so you can have multiple threads
+ * reading the same item multiple times.
+ *
+ * The Ringbuffer is the backing data-structure for the reliable {@link com.hazelcast.core.ITopic} implementation. See
+ * {@link com.hazelcast.config.ReliableTopicConfig}.
+ *
+ * @param <E>
+ */
+@Beta
+public interface Ringbuffer<E> extends DistributedObject {
+
+    /**
+     * Returns the capacity of this Ringbuffer.
+     *
+     * @return the capacity.
+     */
+    long capacity();
+
+    /**
+     * Returns number of items in the ringbuffer.
+     *
+     * If no ttl is set, the size will always be equal to capacity after the head completed the first loop
+     * around the ring. This is because no items are getting retired.
+     *
+     * @return the size.
+     */
+    long size();
+
+    /**
+     * Returns the sequence of the tail. The tail is the side of the ringbuffer where the items are added to.
+     *
+     * The initial value of the tail is -1.
+     *
+     * @return the sequence of the tail.
+     */
+    long tailSequence();
+
+    /**
+     * Returns the sequence of the head. The head is the side of the ringbuffer where the oldest items in the
+     * ringbuffer are found.
+     *
+     * If the RingBuffer is empty, the head will be one more than the tail.
+     *
+     * The initial value of the head is 0 (1 more than tail).
+     *
+     * @return the sequence of the head.
+     */
+    long headSequence();
+
+    /**
+     * Returns the remaining capacity of the ringbuffer.
+     *
+     * The returned value could be stale as soon as it is returned.
+     *
+     * If ttl is not set, the remaining capacity will always be the capacity.
+     *
+     * @return the remaining capacity.
+     */
+    long remainingCapacity();
+
+    /**
+     * Adds an item to the tail of the Ringbuffer. If there is no space in the Ringbuffer, the add will overwrite the oldest
+     * item in the ringbuffer no matter what the ttl is. For more control on this behavior, check the
+     * {@link #addAsync(Object, OverflowPolicy)} and the {@link OverflowPolicy}.
+     *
+     * The returned value is the sequence of the added item. Using this sequence you can read the added item.
+     *
+     * <h3>Using the sequence as id</h3>
+     * This sequence will always be unique for this Ringbuffer instance so it can be used as a unique id generator if you are
+     * publishing items on this Ringbuffer. However you need to take care of correctly determining an initial id when any node
+     * uses the ringbuffer for the first time. The most reliable way to do that is to write a dummy item into the ringbuffer and
+     * use the returned sequence as initial id. On the reading side, this dummy item should be discard. Please keep in mind that
+     * this id is not the sequence of the item you are about to publish but from a previously published item. So it can't be used
+     * to find that item.
+     *
+     * @param item the item to add.
+     * @return the sequence of the added item.
+     * @throws java.lang.NullPointerException if item is null.
+     * @see #addAsync(Object, OverflowPolicy)
+     */
+    long add(E item);
+
+    /**
+     * Asynchronously writes an item with a configurable {@link OverflowPolicy}.
+     *
+     * If there is space in the ringbuffer, the call will return the sequence of the written item.
+     *
+     * If there is no space, it depends on the overflow policy what happens:
+     * <ol>
+     * <li>{@link OverflowPolicy#OVERWRITE}: we just overwrite the oldest item in the ringbuffer and we violate
+     * the ttl</li>
+     * <li>{@link OverflowPolicy#FAIL}: we return -1 </li>
+     * </ol>
+     *
+     * The reason that FAIL exist is to give the opportunity to obey the ttl. If blocking behavior is required,
+     * this can be implemented using retrying in combination with a exponential backoff. Example:
+     * <code>
+     * long sleepMs = 100;
+     * for (; ; ) {
+     * long result = ringbuffer.addAsync(item, FAIL).get();
+     * if (result != -1) {
+     * break;
+     * }
+     * TimeUnit.MILLISECONDS.sleep(sleepMs);
+     * sleepMs = min(5000, sleepMs * 2);
+     * }
+     * </code>
+     *
+     * @param item           the item to add
+     * @param overflowPolicy the OverflowPolicy to use.
+     * @return the sequenceId of the added item, or -1 if the add failed.
+     * @throws NullPointerException if item or overflowPolicy is null.
+     */
+    ICompletableFuture<Long> addAsync(E item, OverflowPolicy overflowPolicy);
+
+    /**
+     * Reads one item from the Ringbuffer.
+     *
+     * If the sequence is one beyond the current tail, this call blocks until an item is added.
+     *
+     * This means that the ringbuffer can be processed using the following idiom:
+     * <code>
+     * Ringbuffer&lt;String&gt; ringbuffer = hz.getRingbuffer("rb");
+     * long seq = ringbuffer.headSequence();
+     * while(true){
+     * String item = ringbuffer.readOne(seq);
+     * seq++;
+     * ... process item
+     * }
+     * </code>
+     *
+     * This method is not destructive unlike e.g. a queue.take. So the same item can be read by multiple readers or it can be
+     * read multiple times by the same reader.
+     *
+     * Currently it isn't possible to control how long this call is going to block. In the future we could add e.g.
+     * tryReadOne(long sequence, long timeout, TimeUnit unit).
+     *
+     * @param sequence the sequence of the item to read.
+     * @return the read item
+     * @throws StaleSequenceException             if the sequence is smaller then {@link #headSequence()}. Because a
+     *                                            Ringbuffer won't store all event indefinitely, it can be that the data for the
+     *                                            given sequence doesn't exist anymore and the {@link StaleSequenceException}
+     *                                            is thrown. It is up to the caller to deal with this particular situation, e.g.
+     *                                            throw an Exception or restart from the last known head. That is why the
+     *                                            StaleSequenceException contains the last known head.
+     * @throws java.lang.IllegalArgumentException if sequence is smaller than 0 or larger than {@link #tailSequence()}+1.
+     * @throws InterruptedException               todo
+     */
+    E readOne(long sequence) throws InterruptedException;
+
+    /**
+     * Adds all the items of a collection to the tail of the Ringbuffer.
+     *
+     * A addAll is likely to outperform multiple calls to {@link #add(Object)} due to better io utilization and a reduced number
+     * of executed operations.
+     *
+     * If the batch is empty, the call is ignored.
+     *
+     * When the collection is not empty, the content is copied into a different data-structure. This means that:
+     * <ol>
+     * <li>after this call completes, the collection can be re-used.</li>
+     * <li>the collection doesn't need to be serializable</li>
+     * </ol>
+     *
+     * If the collection is larger than the capacity of the ringbuffer, then the items that were written first will
+     * be overwritten. Therefor this call will not block.
+     *
+     * The items are inserted in the order of the Iterator of the collection. If an addAll is executed concurrently with
+     * an add or addAll, no guarantee is given that items are contiguous.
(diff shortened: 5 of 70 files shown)


bwzhang2011:
@pveentjer, as for #3661, ringBuffer distributed brought in to speed up the throughout-put of hazelcast. expect more processing could be benefit from such. thanks a lot for the great work.

bwzhang2011:
@pveentjer, would you mind leaving us more message for ringbuffer feature, as you know, it was brought in from disruptor framework by lmax team. and we're looking forward to the improvement for IQueue or IMap enhancement compared with original implementation through hazelcast simulator testing.

pveentjer:
I don't think I'll add more features for the 3.5 release. The most important part is that we have reliable eventing and the ringbuffer is the first step for that. Perhaps that in the 3.6 release more features are going to be added. PS: This is not a comparison with the LMax ringbuffer. They focus on data processing, I just focus on data storage.

bwzhang2011:
@pveentjer , thanks for view and more features added in the future plan. by the way, hope more document or testing cases added for explanation once it's merged into the master.

pveentjer:
verify

StanAccy:
So based on a review of the PR, have I understood correctly that the RingBuffer will be stored on a single node in the cluster, thus the incrementing counter doesn't need to be distributed across the cluster (and synchronized), and there will be 1 full sync backup of the RingBuffer on another node (by default) with async backups optional, with the buffer data and the counter being synchronized?

I see that you're allowing multiple reads of the same data by multiple readers (unlike queue reads which are destructive).  Will there be a way to prevent buffer overruns and block producers in to the RingBuffer?

Being able to obtain the sequence Id on add() would be very helpful.

ICompletableFuture<E> asyncReadOrWait(long sequenceId); - perhaps readEventStream() ?

Being able to persist the RingBuffer state to disk would be very helpful (or at least a hook to do so)

pveentjer:
Hi @StanAccy 

Comments inline.

```
So based on a review of the PR, have I understood correctly that the RingBuffer will be stored on a single node in the cluster, thus the incrementing counter doesn't need to be distributed across the cluster (and synchronized), and there will be 1 full sync backup of the RingBuffer on another node (by default) with async backups optional, with the buffer data and the counter being synchronized?
```

Correct. But also when we add support for a partitioned ringbuffer, the counter is not going to be a big issue. We'll chop up the content in e.g. 1000 items and only every 1000 items you move to the next partition. So in most cases you have a local counter to deal with. Only when you miss, you need to go to the central maintained counter.. but this centrally maintained counter is only going to be updated once every 1000 calls.

```
I see that you're allowing multiple reads of the same data by multiple readers (unlike queue reads which are destructive). Will there be a way to prevent buffer overruns and block producers in to the RingBuffer?
```

Currently not. But that is something I'm thinking about. The problem is that then reads become a lot more complicated. 

```
Being able to obtain the sequence Id on add() would be very helpful.
```
I removed that this week. The problem with having the add return the sequenceId, is that you now rely on the fact that it is actually being stored. And then it becomes impossible to locally batch a bunch of adds and flush them e.g. every 100ms.

Could you give an example? And perhaps I can add a second add that returns a long.. so no batching can be applied here.

```
ICompletableFuture asyncReadOrWait(long sequenceId); - perhaps readEventStream() 
```

Need to think about that. Because it isn't actually a stream you are reading.. just a single item. 

```
Being able to persist the RingBuffer state to disk would be very helpful (or at least a hook to do so)
```

From our internal wiki:
    * batch support for reading. Currently every item is retrieved individually. But if the ringbuffer tail is at 1000 and a caller is at sequenceId 900 and wants to chew its way to item 1000, then currently he needs to invoke an operation for every item. This causes a lot of overhead. So instead of getting an item one a time, add a method to the ringbuffer where a batchsize can be passed.
    * write batching: either explicit using an addAll. Or implicit by transparently sending all the write after a certain number of writes have happened or when a timeout has expired.
    * expiration policy. Currently items in the Ringbuffer are never expired, only overwritten. In some cases you want to keep data only for a certain amount of time and want to get rid of stale data. With an expiration policy old items can be removed from the ringbuffer to save space.
    * periodic processing. A ringbuffer is very useful to deal with streams of data, e.g. events. Once can schedule a periodic task and process all items e.g. for the last 60 second and do some aggregation; this could also be provided out of the box to reduce overhead.
    * partitioned. Currently the Ringbuffer is fully kept in memory of a single JVM since it is fully stored in a single partition. This can lead to resource problems. The idea is the partition the ringbuffer content where each partition will get a batch of items to store. This way the capacity of the Ringbuffer can be increased by adding more members to the cluster.
    * mechanism to block new items from being added if the capacity is reached. This can be useful for topics where a producer is produces faster than consumers consume. Currently a consumer will get a StaleSequenceException. The problem with this feature is that currently reads are super cheap because a ringbuffer doesn't keep track how many reads for a given item been done. The problem is that a ringbuffer has no idea how long data needs to be kept in memory.
    * client support. Since the client is being rewritten, there is no client support for the Ringbuffer in HZ 3.5.
    * event filtering: when an event enters the ringbuffer a filter could be added to see if the data is valid. If not, it can be dropped
    * event transformation: when an event enter the ringbuffer, it a transformation function could be applied. E.g. fix missing fields etc. Perhaps it can be combined with the filter into a single interface (null means destroyed). Perhaps that a transformer can also take care of the in memory format; e.g. store it as Data or an actual object
    * primitive type array: currently the items in a ringbuffer always are an object-reference even if the actual data is a primitive type. E.g. temperature sensor feed from a thermometer. Supporting primitive types reduces the size of the data to be stored + potentially can reduce object litter.
    * storage: currently we keep everything in memory.
    * logic when item is added. On the same operation that add an item to a ringbuffer, you want to be able to trigger some logic. For example for aggregation when te 10 item in a ringbuffer appears, you want to aggregate all the 10 and write out e.g. to another ringbuffer.  This way you can set up a chain of aggregations, e.g. aggregate per minute, per hour, per day, per week where each aggregation has its own ringbuffer. Also we should look at cheap time based triggering. E.g. if 9 items are there.. and the 10th item doesn't come in.. you perhaps still want to aggregate every x-seconds.
    * resizable array: currently all items in a ringbuffer are stored in an array. Even though arrays have very low storage overhead, if the ringbuffer is configured with 1M items, then with a 64bit pointer, the array will take 8MB. Instead of immediately create an array with the full size up front, the array could grow (and shrink) dynamically.
    * instead of having a fixed capacity, keep storing messages until an expiration policy kicks on. So if someone wants to store 24 hours of data, he creates an expiration policy of 24 hours. We could use a flexible array or a hashmap as storage mechanism
    * consume the head. Currently there is no option to consume an item from the ringbuffer, it is a pure readonly data-structure. But the consume head allow the head to be discarded.
    * other Collection like methods like contains, size etc etc.
    * item tombstone: item's can't be removed from a ringbuffer since this would mess up the sequenceId's. But a tombstone can be put on top of it to prevent the item from being read. This can be useful if for example there is old data that isn't needed anymore because newer data just came in. Then a tombstone can be put on the old item to prevent dealing with this outdated item. If a user wants to read an item with a tombstone, we could send a TombstoneException or something.

StanAccy:
re: add() returning long:  It would be helpful for diagnostic purposes on an async event based platform (like ours is).  We would also use the IDs for external persistence so that any readers can request older ringbuffer items that are no longer in memory (we're aiming for Kafka like functionality, albeit with lower performance but far simpler deployment and manageability - we're not doing millions of ops/s here).  Being able to review the order of events is very important for problem resolution.

For an event based platform, we'd want to ensure any add() was committed to the cluster on a per-add basis  - we wouldn't locally batch adds in a client (or want that) in normal operation because that would break our event platform as being a system-of-record for event notifications.  For the case where we had to batch load a bunch of items it would either be a restoration / startup state or it sounds like we were using the wrong structure for our batch use case.  If the batch function is in place in the final HZ3.5 (or whatever version), then we'd really need to be able to set batch size to 1 and have each and every add() committed to the cluster (including the sync backups).  The client services need to know that an event was sent to the cluster or not - it makes their processing and restoration in the case of failover possible.  Maybe the batch add functionality could be provided as a separate client side wrapper/proxy around the core ringbuffer functionality, and the underlying add() takes a List<> and returns a List<Long>?

bwzhang2011:
nice conversation to learn more about the practice of ringbuffer in hazelcast. the only pity is that the ringbuffer structure for hz client would not be supported for hz3.5. what's more, as there're more than one month before hz3.5 release, maybe the hz3.5 EA for the first one. we hope the final could add the client support as well and there will be more time to make it more robust.

Donnerbart:
What is the state of this PR? Any progress on the XML config part, so I can help you out with the Simulator test?

pveentjer:
Yesterday I added the XML for most of the settings. Only a few tweaks on the reliable topic part needed.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/AbstractRingBufferOperation.java
```diff
@@ -0,0 +1,76 @@
...
+    public AbstractRingBufferOperation(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public String getServiceName() {
+        return SERVICE_NAME;
+    }
+
+    public RingbufferContainer getRingBufferContainer() {
```
jerrinot: why is it public? I understand it's inside the impl package, but still I wouldn't expose it to a world.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,159 @@
...
+ * meaning that only 1 thread is able to do take an item. A ringbuffer.read is not destructive, so you can have multiple threads
+ * reading the same data.
+ *
+ * The Ringbuffer is the backing data-structure for the reliable {@link com.hazelcast.core.ITopic} implementation. See
+ * {@link com.hazelcast.config.ReliableTopicConfig}.
+ *
+ * @param <E>
+ */
+@Beta
+public interface Ringbuffer<E> extends DistributedObject {
```
jerrinot: A very nice JavaDoc!

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,159 @@
...
+     * @param collection the batch of items to add.
+     * @throws java.lang.NullPointerException if batch is null, or if an item in this batch is null.
+     */
+    ICompletableFuture asyncAddAll(Collection<? extends E> collection);
+
+    /**
+     * Reads a batch of items from the Ringbuffer. If the number of available items after the first read item is smaller than
+     * the maxCount, these items are returned. So it could be the number of items read is smaller than the maxCount.
+     *
+     * If there are no items available, this call
```
jerrinot: missing 2nd part of the sentence?

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * The items are added to the result using {@link Collection#add(Object)}
```
jerrinot: why is this important?

jerrinot:
can't `RingBufferProxy` has an internal sequence counter? that would allow to have methods such as `readNext()`

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/AbstractRingBufferOperation.java
pveentjer: Fixed

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: deleted

pveentjer:
Then you would loose isolation. Imagine that there are 2 threads accessing the same proxy.

jerrinot:
fair enough

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,156 @@
...
+     * If the collection is larger than the capacity of the ringbuffer, then the items that were written first will
+     * be overwritten. Therefor this call will not block till until space comes available.
+     *
+     * The items are inserted in the order of the Iterator of the collection. If an addAll is executed concurrently with
+     * an add or addAll, no guarantees are given that items are contiguous.
+     *
+     * @param collection the batch of items to add.
+     * @throws java.lang.NullPointerException if batch is null, or if an item in this batch is null.
+     */
+    ICompletableFuture addAllAsync(Collection<? extends E> collection);
```
jerrinot: does the returned future contain any value?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/ReadAllOperation.java
```diff
@@ -0,0 +1,121 @@
...
+        this.startSequenceId = startSequenceId;
+    }
+
+    @Override
+    public void beforeRun() throws Exception {
+        RingbufferContainer ringbuffer = getRingBufferContainer();
+
+        long tailSequenceId = ringbuffer.tailSequenceId();
+
+        //todo
```
jerrinot: todo what?

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: Nope. I wanted to make the type Void. but that doesn't work currently with our InvocationFutures.

Donnerbart:
Can you rebase, please?

pveentjer:
Quite a lot of stuff has been added:
- batching for read and for write
- retention so you can configure how long items are going to remain in the ringbuffer
- overflow policy for adding; either overwrite or fail. This gives the opportunity to do a retry in combination with an exponential backoff. 
- add returns long so it can be used as unique id generator.

bilalyasar:
can you add invalid configuration test for xsd changes. ( we have InvalidConfigurationTest class), and can you update ConfigXmlGenerator.java class.

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
```diff
@@ -0,0 +1,322 @@
...
+     * </ol>
+     *
+     * @param inMemoryFormat the new in memory format.
+     * @return the updated Config.
+     * @throws NullPointerException if inMemoryFormat is null.
+     * @throws IllegalArgumentException if {@link InMemoryFormat#NATIVE} in memory format is selected.
+     */
+    public RingbufferConfig setInMemoryFormat(InMemoryFormat inMemoryFormat) {
+        checkNotNull(inMemoryFormat, "inMemoryFormat can't be null");
+        checkFalse(inMemoryFormat == NATIVE, "InMemoryFormat " + NATIVE + " is not support");
```
bilalyasar: `is not supported`

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
```diff
@@ -0,0 +1,322 @@
...
+ * The RingBuffer is currently not a distributed data-structure, so its content will be fully stored on a single member
+ * in the cluster and its backup in another member in the cluster.
+ */
+@Beta
+public class RingbufferConfig {
+
+    /**
+     * Default value of capacity of the RingBuffer.
+     */
+    public static final int DEFAULT_CAPACITY = 100 * 1000;
```
bilalyasar: these values are not consistent with full-xml and default-xml configs.

hazelcast/src/main/resources/hazelcast-config-3.5.xsd
```diff
@@ -655,6 +656,20 @@
             </xs:simpleType>
         </xs:attribute>
     </xs:complexType>
+    <xs:complexType name="ringbuffer">
+        <xs:all>
+            <xs:element name="capacity" type="xs:unsignedInt" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="retention-seconds" type="xs:unsignedInt" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="backup-count" type="backup-count" minOccurs="0" maxOccurs="1" default="1"/>
+            <xs:element name="async-backup-count" type="backup-count" minOccurs="0" maxOccurs="1" default="0"/>
+            <xs:element name="in-memory-format" type="in-memory-format" minOccurs="0" maxOccurs="1" default="BINARY"/>
```
bilalyasar: when user enters in memory format as NATIVE then hz will throw invalidArgumentException ( due to setMemoryFormat method calling) , but before that we should throw invalidConfiguration exception.

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
pveentjer: good eyes!

pveentjer: done

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
pveentjer: done.

hazelcast/src/main/resources/hazelcast-config-3.5.xsd
pveentjer: I'm not completely sure what you mean. 

Should the XMLConfigBuilder not catch all exceptions, including IllegalArgumentException, and wrap it with an InvalidConfigurationException? 

Because most of the methods throws IllegalArgument or NPE when the arguments are not correct. 

So I'm a bit lost here. Can you elaborate?

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
pveentjer: It doesn't make sense to add test the the InvalidConfigurationTest.

The RingbufferConfig is validated by the RingbufferConfigTest. There is 100% coverage and all bad inputs are tested. See included source for more info

The InvalidConfigurationTest repeats the same thing. There is no value on duplicating the tests, since the XMLConfigBuilder relies on the RingbufferConfig setters to set the fields.
screenshot from 2015-05-04 10 06 32

```
package com.hazelcast.config;

import com.hazelcast.test.HazelcastParallelClassRunner;
import com.hazelcast.test.annotation.QuickTest;
import org.junit.Test;
import org.junit.experimental.categories.Category;
import org.junit.runner.RunWith;
import org.mockito.internal.matchers.Null;

import static com.hazelcast.config.RingbufferConfig.DEFAULT_ASYNC_BACKUP_COUNT;
import static com.hazelcast.config.RingbufferConfig.DEFAULT_CAPACITY;
import static com.hazelcast.config.RingbufferConfig.DEFAULT_SYNC_BACKUP_COUNT;
import static com.hazelcast.partition.InternalPartition.MAX_BACKUP_COUNT;
import static org.junit.Assert.assertNotNull;
import static org.junit.Assert.assertSame;
import static org.junit.Assert.fail;

@Category(QuickTest.class)
public class RingbufferConfigTest {

    public static final String NAME = "someringbuffer";

    @Test
    public void testDefaultSetting() {
        RingbufferConfig config = new RingbufferConfig(NAME);
        assertEquals(NAME, config.getName());
        assertEquals(DEFAULT_SYNC_BACKUP_COUNT, config.getBackupCount());
        assertEquals(DEFAULT_ASYNC_BACKUP_COUNT, config.getAsyncBackupCount());
        assertEquals(DEFAULT_CAPACITY, config.getCapacity());
    }

    @Test
    public void testCloneConstructor() {
        RingbufferConfig original = new RingbufferConfig(NAME);
        original.setBackupCount(2).setAsyncBackupCount(1).setCapacity(10);

        RingbufferConfig clone = new RingbufferConfig(original);

        assertEquals(original.getName(), clone.getName());
        assertEquals(original.getBackupCount(), clone.getBackupCount());
        assertEquals(original.getAsyncBackupCount(), clone.getAsyncBackupCount());
        assertEquals(original.getCapacity(), clone.getCapacity());
    }

    @Test
    public void testCloneConstructorWithName() {
        RingbufferConfig original = new RingbufferConfig(NAME);
        original.setBackupCount(2).setAsyncBackupCount(1).setCapacity(10);

        RingbufferConfig clone = new RingbufferConfig("foobar", original);

        assertEquals("foobar", clone.getName());
        assertEquals(original.getBackupCount(), clone.getBackupCount());
        assertEquals(original.getAsyncBackupCount(), clone.getAsyncBackupCount());
        assertEquals(original.getCapacity(), clone.getCapacity());
    }

    // =================== set capacity ===========================

    @Test
    public void setCapacity() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setCapacity(1000);

        assertEquals(1000, config.getCapacity());
    }

    @Test(expected = IllegalArgumentException.class)
    public void setCapacity_whenTooSmall() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setCapacity(0);
    }

    // =================== set backups count ===========================

    @Test
    public void setBackupCount() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setBackupCount(4);

        assertEquals(4, config.getBackupCount());
    }

    @Test(expected = IllegalArgumentException.class)
    public void setBackupCount_whenTooSmall() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setBackupCount(-1);
    }

    @Test(expected = IllegalArgumentException.class)
    public void setBackupCount_whenTooLarge() {
        RingbufferConfig ringbufferConfig = new RingbufferConfig(NAME);

        ringbufferConfig.setBackupCount(MAX_BACKUP_COUNT + 1);
    }

    // =================== set async backup count ===========================

    @Test
    public void setAsyncBackupCount() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setAsyncBackupCount(4);

        assertEquals(4, config.getAsyncBackupCount());
    }

    @Test(expected = IllegalArgumentException.class)
    public void setAsyncBackupCount_whenTooSmall() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setAsyncBackupCount(-1);
    }

    @Test(expected = IllegalArgumentException.class)
    public void setAsyncBackupCount_whenTooLarge() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setAsyncBackupCount(MAX_BACKUP_COUNT + 1);
    }

    // ============= get total backup count ====================

    @Test
    public void getTotalBackupCount() {
        RingbufferConfig config = new RingbufferConfig(NAME);
        config.setAsyncBackupCount(2);
        config.setBackupCount(3);

        int result = config.getTotalBackupCount();

        assertEquals(5, result);
    }

    // ================== retention =================================


    @Test(expected = IllegalArgumentException.class)
    public void setRetentionSeconds_whenNegative() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setRetentionSeconds(-1);
    }

    @Test
    public void setRetentionSeconds() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        RingbufferConfig returned = config.setRetentionSeconds(10);

        assertSame(returned, config);
        assertEquals(10, config.getRetentionSeconds());
    }

    // ================== inmemoryformat =================================


    @Test(expected = NullPointerException.class)
    public void setInMemoryFormat_whenNull() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setInMemoryFormat(null);
    }

    @Test(expected = IllegalArgumentException.class)
    public void setInMemoryFormat_whenNative() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        config.setInMemoryFormat(InMemoryFormat.NATIVE);
    }

    @Test
    public void setInMemoryFormat() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        RingbufferConfig returned = config.setInMemoryFormat(InMemoryFormat.OBJECT);

        assertSame(config, returned);
        assertEquals(InMemoryFormat.OBJECT, config.getInMemoryFormat());
    }

    // ==================== toString ================================

    @Test
    public void test_toString() {
        RingbufferConfig config = new RingbufferConfig(NAME);

        String s = config.toString();

        assertEquals("RingbufferConfig{name='someringbuffer', capacity=100000, backupCount=1, " +
                "asyncBackupCount=0, retentionSeconds=30, inMemoryFormat=BINARY}", s);
    }

    // =================== getAsReadOnly ============================

    @Test
    public void getAsReadOnly() {
        RingbufferConfig original = new RingbufferConfig(NAME);
        original.setBackupCount(2).setAsyncBackupCount(1).setCapacity(10).setRetentionSeconds(400);

        RingbufferConfig readonly = original.getAsReadOnly();
        assertNotNull(readonly);

        assertEquals(original.getName(), readonly.getName());
        assertEquals(original.getBackupCount(), readonly.getBackupCount());
        assertEquals(original.getAsyncBackupCount(), readonly.getAsyncBackupCount());
        assertEquals(original.getCapacity(), readonly.getCapacity());
        assertEquals(original.getRetentionSeconds(), readonly.getRetentionSeconds());
        assertEquals(original.getInMemoryFormat(), readonly.getInMemoryFormat());

        try {
            readonly.setCapacity(10);
            fail();
        } catch (UnsupportedOperationException expected) {
        }

        try {
            readonly.setAsyncBackupCount(1);
            fail();
        } catch (UnsupportedOperationException expected) {
        }

        try {
            readonly.setBackupCount(1);
            fail();
        } catch (UnsupportedOperationException expected) {
        }

        try {
            readonly.setRetentionSeconds(1);
            fail();
        } catch (UnsupportedOperationException expected) {
        }

        try {
            readonly.setInMemoryFormat(InMemoryFormat.OBJECT);
            fail();
        } catch (UnsupportedOperationException expected) {
        }
    }
}
```

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -0,0 +1,345 @@
...
+            int index = toIndex(headSequence);
+
+            if (ringExpirationMs[index] > now) {
+                return;
+            }
+
+            //System.out.println("expiring index:" + index);
+
+            // we null the slot and allow the gc to take care of the object.
+            ringItems[index] = null;
```
gurbuzali: shouldn't we clean the `ringExpirationMs[index]` too ? 
I understand that it may not be a problem since you are incrementing headSequence but mentioning just for completeness

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
bilalyasar: actually you write, i want to see some tests because;
when user makes wrong config in xml hazelcast will throw invalidConfigurationException only
when user makes wrong config in Config object( lets say 10 backup-count) hazelcast will throw illegalArgumentException..
so exceptions are different.

gurbuzali:
looks good, just a minor comment

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
pveentjer: Good point. I'll add a line of comment.

pveentjer:
Fixing one bug with ringbuffer container serialization in combination with retention.

```
   // we write the time difference compared to now. Because the clock on the receiving side
            // can be totally different then our time. If there would be a retention of 10 seconds, than
            // the expiration time would be 10 seconds after the insertion time. But the if ringbuffer is
            // migrated to a machine with a time 1 hour earlier, the retention would effectively become
            // 1 hours and 10 seconds.
            if (retentionEnabled) {
                out.writeLong(ringExpirationMs[index] - now);
            }
```

checkstyle/suppressions.xml
```diff
@@ -142,6 +142,14 @@
     <suppress checks="MethodCount" files="com.hazelcast.concurrent.lock.LockStoreImpl"/>
     <suppress checks="Todo" files="com.hazelcast.concurrent.lock.ConditionImpl"/>
 
+    <!-- ringbuffer -->
+    <suppress checks="JavadocMethod" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocType" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocMethod" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocVariable" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
+    <suppress checks="JavadocPackage" files="com.hazelcast.ringbuffer.impl.operations[\\/]"/>
```
mtopolnik: I suggest we fix these regexes, I've already done so on the Enterprise project. It should say `"com/hazelcast/ringbuffer/impl/operations/"`

hazelcast/src/main/resources/hazelcast-config-3.5.xsd
bilalyasar: as default, when user makes wrong config xml we throw InvalidConfigurationException.
because we make [schema validation] (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java#L202) before the parsing Config object.
and in this case when user make format as NATIVE, hazelcast will throw illegalArgumentException not InvalidConfiguration.

pveentjer: I'm still not understanding because the Config objects throw NPE or IllegalArgumentException. 

For example from ExecutorConfig:

```
  /**
     * @param poolSize the poolSize to set
     */
    public ExecutorConfig setPoolSize(final int poolSize) {
        if (poolSize <= 0) {
            throw new IllegalArgumentException("poolSize must be positive");
        }
        this.poolSize = poolSize;
        return this;
    }
```

It should be a concern of the XML parser to trap all IllegalArgument/NullPointerExceptions and convert them to a InvalidConfiguraitonException.

So I'm not seeing what that has to do with my config objects.

pveentjer: Is this related specifically to NATIVE? Which is a valid InMemoryFormat?

checkstyle/suppressions.xml
pveentjer: Good idea. But I would rather do the whole checkstyle file in 1 go in a different PR.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+ * The items in the ringbuffer can be found by a sequence that is in between (inclusive) the head and tail sequence.
+ *
+ * If data is read from a ringbuffer with a sequence that is smaller than the headSequence, it means that the data
+ * is not available anymore and a {@link StaleSequenceException} is thrown.
+ *
+ * A Ringbuffer currently is not a distributed data-structure. So all data is stored in a single partition; comparable to the
+ * IQueue implementation. But we'll provide an option to partition the data in the near future.
+ *
+ * A Ringbuffer can be used in a similar way as a queue, but one of the key differences is that a queue.take is destructive,
+ * meaning that only 1 thread is able to do take an item. A ringbuffer.read is not destructive, so you can have multiple threads
```
mtopolnik: Minor typo: "to do take an item".

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Returns the capacity of this Ringbuffer.
+     *
+     * @return the capacity.
+     */
+    long capacity();
+
+    /**
+     * Returns the size (number of items) in the ringbuffer.
```
jerrinot: What about? `Returns number of items in the ringbuffer.`

pveentjer: done

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: done

mtopolnik:
:+1:

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+     * The returned value could be stale as soon as it is returned.
+     *
+     * If retention is not set, the remaining capacity will always be the capacity.
+     *
+     * @return the remaining capacity.
+     */
+    long remainingCapacity();
+
+    /**
+     * Adds an item to the tail of the Ringbuffer. Of there is no space in the Ringbuffer, the add will overwrite the oldest
```
jerrinot: typo. "Of there is no space"

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+import com.hazelcast.core.ICompletableFuture;
+import com.hazelcast.core.IFunction;
+import com.hazelcast.spi.annotation.Beta;
+
+import java.util.Collection;
+
+/**
+ * A Ringbuffer is a data-structure where the content is stored in a ring like structure. A ringbuffer has a capacity so it
+ * won't grow beyond that capacity and endanger the stability of the system. If that capacity is exceeded, than the oldest
+ * item in the ringbuffer is overwritten.
```
mdogan: As far as I understand, this behaviour depends on `OverflowPolicy`, right? If capacity is exceeded, depending on the overflow-policy, oldest item is overwritten or operation fails...

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
jerrinot: various methods talk about:
- backoff policy
- retention policy
- overflow policy. 

It would be nice to explain these concepts in a class-level JavaDoc.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+    ICompletableFuture<Long> addAsync(E item, OverflowPolicy overflowPolicy);
+
+    /**
+     * Reads one item from the Ringbuffer.
+     *
+     * If the sequence is one beyond the current tail, this call blocks until an item is added.
+     *
+     * This means that the ringbuffer can be processed using the following idiom:
+     * <code>
+     * Ringbuffer&lt;String&gt; ringbuffer = hz.getRingbuffer("rb");
```
jerrinot: I love the example!

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+ * <li>
+ * headSequence: this is the side where the oldest items are found. So the head is the side where items gets
+ * discarded.
+ * </li>
+ * </ol>
+ * The items in the ringbuffer can be found by a sequence that is in between (inclusive) the head and tail sequence.
+ *
+ * If data is read from a ringbuffer with a sequence that is smaller than the headSequence, it means that the data
+ * is not available anymore and a {@link StaleSequenceException} is thrown.
+ *
```
mdogan: A cosmetic change; javadoc will be more readable if you out  `<p/>` between paragraphs. Even IDEA inserts them when you auto-format the javadoc.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
mdogan: Should not this named with capital *B* -> `RingBuffer`? As it is named in Disruptor: https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/RingBuffer.java

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Returns the size (number of items) in the ringbuffer.
+     *
+     * If no retention period is set, the size will always be equal to capacity after the head completed the first loop
+     * around the ring. This is because no items are getting retired.
+     *
+     * @return the size.
+     */
+    long size();
```
mdogan: Why it's `long`? `capacity` is defined as `int` in configuration...

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: Done

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: Correct

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * Returns the remaining capacity of the ringbuffer.
+     *
+     * The returned value could be stale as soon as it is returned.
+     *
+     * If retention is not set, the remaining capacity will always be the capacity.
+     *
+     * @return the remaining capacity.
+     */
+    long remainingCapacity();
```
mdogan: why is it `long`? `capacity` is defined as `int` in config.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: Java 8 doesn't like that! So I disabled to generation of these things in IDEA.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
Donnerbart: This would BTW also help to reduce typo warnings in IDEA (`RingBuffer` seems to be fine, `Ringbuffer` is not known out of the box).

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * If there is no space, it depends on the overflow policy what happens:
+     * <ol>
+     * <li>{@link OverflowPolicy#OVERWRITE}: we just overwrite the oldest item in the ringbuffer and we violate
+     * the retention</li>
+     * <li>{@link OverflowPolicy#FAIL}: we return -1 </li>
+     * </ol>
+     *
+     * The reason that FAIL exist is to give the opportunity to obey the retention policy. A blocking call could have been
+     * created, but this would trap the caller thread and this is not always desirable. If blocking behavior is required,
```
jerrinot: I don't quite understand this part. It's async API so what thread would have been trapped? 

```java
ICompletableFuture<Long> f = ringbuffer.addAsync(item, WAIT);
f.andThen(myHandler);
myThreadIsNotTrapped();
```
?

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: In 3.6 i hope we can add a partitioned version and then we should be able to scale beyond Integer.MAX_VALUE.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
mdogan: Ah, I didn't know that. So how are we gonna start a paragraph then?

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: Originally the call was sync. Will fix jvadoc

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: No idea :)

<p></p>  ??

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
mdogan: then I think you should at least change the `config.getCapacity()` to return `long`... 
```java
public long getCapacity() {
      return capacity;
}
```

pveentjer: Just the getter, or the whole property (and xsd etc).

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @param item the item to add.
+     * @return the sequence of the added item.
+     * @throws java.lang.NullPointerException if item is null.
+     * @see #addAsync(Object, OverflowPolicy)
+     */
+    long add(E item);
+
+    /**
+     * Asynchronously writes an item with a configurable backoff policy.
```
mdogan: with a configurable **overflow** policy

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -0,0 +1,342 @@
...
+ * each other.
+ * The reason why 2 array are created instead of just wrapping the item in a new object containing the expiration is that
+ * we don't want to generate more waste than needed.
+ */
+public class RingbufferContainer implements DataSerializable {
+
+    private static final long RETENTION_DISABLED = 0;
+
+    // contains the actual items
+    Object[] ringItems;
```
jerrinot: why the fields aren't private?

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: done

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
mtopolnik: Does Java 8 complain about an unpaired `<p>`?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -0,0 +1,342 @@
...
+        inMemoryFormat = values()[in.readInt()];
+
+        ringItems = new Object[capacity];
+
+        boolean retentionEnabled = isRetentionEnabled();
+        if (retentionEnabled) {
+            ringExpirationMs = new long[capacity];
+        }
+
+        long now = System.currentTimeMillis();
```
mdogan: If you use `com.hazelcast.util.Clock`, it will be easier to test the time difference you described above.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
mdogan: to not to break compatibility in future, I think just getter is enough. but it's up to you, if you change all...

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
pveentjer: testing

hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param <E>
+ */
+@Beta
+public interface ReadResultSet<E> extends Iterable<E> {
+
+    /**
+     * Returns the number of items that have been read before filtering.
+     *
+     * If no filter is set, then the readCount will be the same as resultCount. But if a filter is applied, it could be that items
```
jerrinot: what is `resultCount`?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
pveentjer: Done

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: I think it is something we can change later one. The Config is marked as beta.

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
pveentjer: Javadoc tool fails

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferProxy.java
```diff
@@ -0,0 +1,213 @@
...
+    static final int MAX_BATCH_SIZE = 1000;
+
+    private final String name;
+    private final int partitionId;
+    private final RingbufferConfig config;
+
+    public RingbufferProxy(NodeEngine nodeEngine, RingbufferService service, String name) {
+        super(nodeEngine, service);
+        this.name = name;
+        this.partitionId = nodeEngine.getPartitionService().getPartitionId(name);
```
mdogan: To be able to handle partition-aware naming, you should use `StringPartitioningStrategy` here. For example in `QueueProxySupport` this is done with:
```java
this.partitionId = nodeEngine.getPartitionService().getPartitionId(getNameAsPartitionAwareData());
```

`AbstractDistributedObject .getNameAsPartitionAwareData()` is:
```java
protected Data getNameAsPartitionAwareData() {
     String name = getName();
     return getNodeEngine().getSerializationService().toData(name, PARTITIONING_STRATEGY);
}
```

pveentjer: done

mdogan: You can add a test to `PartitionControlledIdTest` to ensure this..

hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
pveentjer: should be 'size'

jerrinot:
what's the point of having configurable `InMemoryFormat`? `ReadResultSet` always (de-)serialize stuff anyway.

pveentjer:
Check the javadoc! :)

```
  * The default is binary. The object InMemoryFormat is useful when:
     * <ol>
     *     <li>of the object stored in object format has a smaller footprint than in binary format</li>
     *     <li>if there are readers using a filter. Since for every filter invocation, the object needs to be available in
     *     object format.</li>
     * </ol>
  ```

File: hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/AddAllOperation.java [hunk 26]
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public int getId() {
+        return ADD_ALL_OPERATION;
+    }
+
+    @Override
+    protected void writeInternal(ObjectDataOutput out) throws IOException {
+        super.writeInternal(out);
+
+        out.writeInt(overflowPolicy.ordinal());
```
mdogan: Isn't it better to give an explicit id to each overflow policy? 
```java
public enum OverflowPolicy {

    OVERWRITE(0),
    
    FAIL(1);

    private final int id;

    OverflowPolicy(int id) {
        this.id = id;
    }

    public int getId() {
        return id;
    }
}
```

and then:

```java
out.writeInt(overflowPolicy.getId());
```

jerrinot:
ok

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferProxy.java
pveentjer: Done

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/AddAllOperation.java
```diff
@@ -0,0 +1,133 @@
...
+        for (Data item : items) {
+            out.writeData(item);
+        }
+    }
+
+    @Override
+    protected void readInternal(ObjectDataInput in) throws IOException {
+        super.readInternal(in);
+
+        overflowPolicy = values()[in.readInt()];
```
mdogan: To me, `values()` static import makes this harder to read. I prefer to see `OverflowPolicy.values()`.

pveentjer:
@jerrinot the number of object instances is limited, you can cache them. For example if the ringbuffer would get a temperature feed, you could easily create a TempMeasurement(int value) and cache all the instances. Then storing a single object will cost you 32/64 bits (the reference in the array)... and another 64 for the retention if set.

jerrinot:
:+1: 
A very nice code indeed; pleasure to review it!

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/AddAllOperation.java
pveentjer: Sure. I agree. Don't know how that happened

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/AddAllOperation.java
pveentjer: And the reading?

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
```diff
@@ -0,0 +1,322 @@
...
+     */
+    public static final int DEFAULT_SYNC_BACKUP_COUNT = 1;
+    /**
+     * Default value of asynchronous backup count
+     */
+    public static final int DEFAULT_ASYNC_BACKUP_COUNT = 0;
+    /**
+     * Default value for the retention period.
+     */
+    public static final int DEFAULT_RETENTION_SECONDS = 30;
```
mdogan: Shouldn't default retention time be `0`? so no retention by default...

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/AddAllOperation.java
mdogan:
```java
public enum OverflowPolicy {
...

    public static OverflowPolicy get(int id) {
        switch (id) {
            case 0:
                return OVERWRITE;
            case 1:
                return FAIL;
            default:
                throw new IllegalArgumentException();
        }
    }
}
```

reading:

```java
overflowPolicy = OverflowPolicy.get(in.readInt());
```

pveentjer: Found it. Will update the code

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
pveentjer: The problem with no retention is that the data is not cleared. Only due to overwrite. So it behaves like an array and this can lead to people complaining about memory usage since memory is retained.

mdogan: I see, but if it's documented clearly, it won't be our concern then. But, I, personally, don't expect a retention/TTL on my data-structure unless I explicitly defined. 
If it's default 30 seconds then you should explicitly document this. Otherwise, I guess, it will be a surprise for the user.

pveentjer: I'm open here for suggestion. I'm not bound to retention. I just picked something which made sense to me. @jerrinot @mtopolnik what do you think?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/GenericOperation.java
```diff
@@ -0,0 +1,87 @@
...
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.ringbuffer.impl.RingbufferContainer;
+
+import java.io.IOException;
+
+import static com.hazelcast.ringbuffer.impl.RingbufferDataSerializerHook.GENERIC_OPERATION;
+
+public class GenericOperation extends AbstractRingBufferOperation {
+
+    public static final byte OPERATION_SIZE = 0;
```
metanet: What about using an enum here? Ordinal values can be written to output with DataOutput.writeByte(int).

mdogan: please, no ordinals.. :) this is already explicit/visible.. 
I just don't want to debug an issue caused by enum value inserted in the middle by a huge PR...

metanet: Then, I am OK. In fact, I didn't suggest using an enum only for making it visible. That byte variable is used in run() method of the GenericOperation class to determine a specific behaviour using a switch statement. I think we do that a lot in HZ and over-use switch statements for that purpose. We are transferring a value from someplace to another to pick a specific behaviour at some point. I like the good old strategy pattern where we can move the actual behaviour that is going to be executed.  An enum is very useful for such a use-case. We can do it within a method of the enum. For this code, the only changing thing in run() method is the type of the operation that will be called on ringbuffer an it can be easily moved into methods to instances of an enum. Maybe it is a personal thing but I don't like using switch / if-else blocks instead of a strategy object.

Anyway, I just wanted to state my point and I do not want to go beyond the scope of this PR. Kudos for the good work and clean code.

metanet:
:+1:

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/GenericOperation.java
pveentjer: Thanks.

And everyone thanks for reviewing!

hazelcast/src/main/java/com/hazelcast/ringbuffer/Ringbuffer.java
mtopolnik: Anyway, the correct way is to treat `<p></p>` as a paragraph *container*, not paragraph *break*. So `<p>The paragraph's contents</p>`.

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
mtopolnik: I'm confused about the usage of the word "retention" in your discussion. "To retain" means "to keep", so 0 retention means "immediately toss it away". Did you mean "expiry"? BTW "retention" is a perfectly correct word for me.

pveentjer: Good points. @jerrinot @metanet what do you think? I'm fine with changing the default value for retention or potentially reconsider the name. But keep in mind that this name is also used in other products, for example Kafka.

jerrinot: @mtopolnik raised a very good point indeed! 

@pveentjer: what's the reason use the term "retention" instead of "expiration"? Is it just to be aligned with Kafka? I believe it's more important to be consistent with `IMap` and `JCache` where the term `expiration` is used.

mtopolnik: I agree that it is most useful to have the least amount of synonyms around the API.

pveentjer: @mdogan wdyt?

I'm fine with renaming.

And what about the default. Everyone agrees with 0 ---> no expiraton?

mtopolnik: How about this: make no expiry the default and also reduce the default capacity to, say 1000. This would fight memory leaks in a less surprising manner.

hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
```diff
@@ -1429,6 +1432,33 @@ private void handleSemaphore(final org.w3c.dom.Node node) {
...
+                int capacity = getIntegerValue("capacity", value, RingbufferConfig.DEFAULT_CAPACITY);
+                rbConfig.setCapacity(capacity);
+            } else if ("backup-count".equals(nodeName)) {
+                int backupCount = getIntegerValue("backup-count", value, RingbufferConfig.DEFAULT_SYNC_BACKUP_COUNT);
+                rbConfig.setBackupCount(backupCount);
+            } else if ("async-backup-count".equals(nodeName)) {
+                int asyncBackupCount = getIntegerValue("async-backup-count", value, RingbufferConfig.DEFAULT_ASYNC_BACKUP_COUNT);
+                rbConfig.setAsyncBackupCount(asyncBackupCount);
+            } else if ("retention-seconds".equals(nodeName)) {
+                int retention = getIntegerValue("retention-seconds", value, RingbufferConfig.DEFAULT_ASYNC_BACKUP_COUNT);
```
bilalyasar: i guess, this should be default_retention-seconds (sth. like that)

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
mdogan: I'm ok with *expiry/expiration* term and no expiration by default.

pveentjer: Rename --> ok
default --> ok

Worry --> default capacity... 1000 is very little.

mtopolnik: Very little was sort of my point---a failsafe default with low chance of OOME problems. Achieves the same goal as the default expiry of 30 seconds, but in a less surprising manner. The key message to the user should be that capacity and expiry shoud be carefully considered against the actual use case.

pveentjer: What should be the exact name of the property? In map for example we use ttl.

hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
pveentjer: done

hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
pveentjer: timeToLiveSeconds

bilalyasar:
@pveentjer , I couldn't see spring configuration for RingBuffer in your pull request, what do you think in this respect?

pveentjer:
I'll do them after this PR is merged. Otherwise this things is taking too much time.

pveentjer:
Ok. TTL + default to 0. Default capacity is 10k.

mdogan:
LGTM :+1:

pveentjer:
Then hit the green button :)

sancar:
@pveentjer There are findbugs problems. 

INFO] new com.hazelcast.ringbuffer.impl.operations.AddAllBackupOperation(String, Data[]) may expose internal representation by storing an externally mutable object into AddAllBackupOperation.items ["com.hazelcast.ringbuffer.impl.operations.AddAllBackupOperation"] At AddAllBackupOperation.java:[lines 33-69]
[INFO] new com.hazelcast.ringbuffer.impl.operations.AddAllOperation(String, Data[], OverflowPolicy) may expose internal representation by storing an externally mutable object into AddAllOperation.items ["com.hazelcast.ringbuffer.impl.operations.AddAllOperation"] At AddAllOperation.java:[lines 41-130]

pveentjer:
@Danny-Hazelcast  Testcoverage is above the required minimum

screenshot from 2015-05-05 14 41 40
screenshot from 2015-05-05 14 40 11

pveentjer:
@sancar I'll fix findbugs now.

bilalyasar:
enterprise compilation error:
```
16:27:11 [ERROR] COMPILATION ERROR : 
16:27:11 [INFO] -------------------------------------------------------------

16:27:11 [ERROR] /scratch/jenkins/workspace/Hazelcast-EE-3x/hazelcast-
enterprise/src/main/java/com/hazelcast/security/SecureCallableImpl.java:[135,9] error: 
SecureCallableImpl.HazelcastInstanceDelegate is not abstract and does not override abstract method 
<E>getRingbuffer(String) in HazelcastInstance
```

PR #4791, opened 2015-03-12
Moved ascii to com.hazelcast.internal

pveentjer:


pveentjer:
already agreed with mehmet this package could move to internal.

PR #4800, opened 2015-03-14
Moved com.hazelcast.storage to internal

pveentjer:


pveentjer:
Merging this one. Already talked with Mehmet about it. Eventually this package is going to be removed anyway.

jerrinot:
isn't there a dependency in Enterprise?

pveentjer:
Yes. I'll take care of that.

PR #4802, opened 2015-03-14
Ensure that one millisecond has passed to validate the update of last ac...

### LINKED ISSUE #4796: [TEST-FAILURE] LocalMapStatsTest.testLastAccessTime_updatedConcurrently
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-OracleJDK8/173/com.hazelcast$hazelcast/testReport/com.hazelcast.map/LocalMapStatsTest/testLastAccessTime_updatedConcurrently/

metanet:
...cess time

Fixes #4796

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
@@ -268,11 +268,4 @@ public class LocalMapStatsTest extends HazelcastTestSupport {
         return memberGroupConfig;
     }
 
-    private void sleepAtLeastMillis(long ms) {
-        final long sleepUntil = System.currentTimeMillis() + ms;
-        while (System.currentTimeMillis() < sleepUntil) {
-            sleepMillis(1);
-        }
-    }
-
 }
--- a/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
@@ -132,6 +132,7 @@ public class LocalMapStatsTest extends HazelcastTestSupport {
         final Thread updatingThread = new Thread(new Runnable() {
             @Override
             public void run() {
+                sleepMillis(1);
                 map1.put(key, "value2");
                 map1.getLocalMapStats(); // causes the local stats object to update
             }
--- a/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
@@ -76,10 +76,10 @@ public class LocalMapStatsTest extends HazelcastTestSupport {
             map.put(i, i);
             map.get(i);
         }
-        LocalMapStats localMapStats = map.getLocalMapStats();
+        final LocalMapStats localMapStats = map.getLocalMapStats();
         final long initialHits = localMapStats.getHits();
 
-        final Thread updatingThread = new Thread(new Runnable() {
+        new Thread(new Runnable() {
             @Override
             public void run() {
                 for (int i = 0; i < actionCount; i++) {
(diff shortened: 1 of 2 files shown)


jerrinot:
do you know what's the point of this test? It starts a new thread & calls join() immediately.

pveentjer:
Seems fragile to me. Who says you get a millisecond?

metanet:
Test tries to validate that updates to the local map stats objects from another thread is visible to the current thread. It is why current thread joins to the other one. Since localmapstats object provides last update time as milliseconds, I tried to force it. 

@pveentjer Now I recall that thread.sleep() does not guarantee the duration of sleep. How can I test it properly other than increasing the sleep duration?

jerrinot:
void sleepAtLeastMs(long ms) {

long targetTime = System.getCurrentMillis() + ms;
while (System.getCurrentMillis() < targetTime) {
  sleepMillis(1);
};

}

I still don't see the point of this test as both start() & join() are
happens-before edges. It will be the same as if it was running on a single
thread. (unless I'm missing something)



On Sat, Mar 14, 2015 at 5:52 PM, Ensar Basri Kahveci <
notifications@github.com> wrote:

> Test tries to validate that updates to the local map stats objects from
> another thread is visible to the current thread. It is why current thread
> joins to the other one. Since localmapstats object provides last update
> time as milliseconds, I tried to force it.
>
> @pveentjer <https://github.com/pveentjer> Now I recall that
> thread.sleep() does not guarantee the duration of sleep. How can I test it
> properly other than increasing the sleep duration?
>
> —
> <https://github.com/hazelcast/hazelcast/pull/4802#issuecomment-80548265>.
>



-- 
“Perfection is achieved, not when there is nothing more to add, but when
there is nothing left to take away.”
Antoine de Saint Exupéry

metanet:
Thanks jaromir. You are right about the happens before but the field is
already volatile. I used it to ensure the other operation is ended.

I will fix this test code.

On Saturday, March 14, 2015, Jaromir Hamala <notifications@github.com>
wrote:

> void sleepAtLeastMs(long ms) {
>
> long targetTime = System.getCurrentMillis() + ms;
> while (System.getCurrentMillis() < targetTime) {
> sleepMillis(1);
> };
>
> }
>
> I still don't see the point of this test as both start() & join() are
> happens-before edges. It will be the same as if it was running on a single
> thread. (unless I'm missing something)
>
>
>
> On Sat, Mar 14, 2015 at 5:52 PM, Ensar Basri Kahveci <
> notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
> > Test tries to validate that updates to the local map stats objects from
> > another thread is visible to the current thread. It is why current thread
> > joins to the other one. Since localmapstats object provides last update
> > time as milliseconds, I tried to force it.
> >
> > @pveentjer <https://github.com/pveentjer> Now I recall that
> > thread.sleep() does not guarantee the duration of sleep. How can I test
> it
> > properly other than increasing the sleep duration?
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/hazelcast/hazelcast/pull/4802#issuecomment-80548265
> >.
> >
>
>
>
> --
> “Perfection is achieved, not when there is nothing more to add, but when
> there is nothing left to take away.”
> Antoine de Saint Exupéry
>
> —
> <https://github.com/hazelcast/hazelcast/pull/4802#issuecomment-80553541>.
>


-- 
Ensar Basri Kahveci
Software Engineer

mtopolnik:
@jerrinot Your implementation is off by one: if we read current time as x millis, the next nanosecond we may already read x+1 millis and satisfy the condition to wait "at least one" millisecond.

As for the questioned test code, Ensar has *deleted* the `join` logic.

jerrinot:
@mtopolnik: that's a good catch, thanks!

mtopolnik:
@metanet Since this sleeping logic is a part of this commit, I would advise fixing it.

mtopolnik:
:+1:

serkan-ozal:
:+1:

PR #4806, opened 2015-03-15
Repackaged the ExecutionService

pveentjer:
so that it is in line with other service...

So there is now an InternalExectionService and the classes/interfaces are now in a similar structure as most of our other services.


pveentjer:
verify

mtopolnik:
:+1:

gurbuzali:
needs rebase

PR #4807, opened 2015-03-15
Removing hard dependencies to BasicOperationService

pveentjer:
Preparation work for Dependency Injection of our services instead of the fragile NodeEngineImpl approach we are currently using. 

I have removed some methods like reset/shutdown from the InternalOperationService because the NodeEngineImpl is allowed to know any lifecycle methods or listener methods of the concrete class. This is exactly the same as within a spring application context, where the context knows the implementation details of a class.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/InternalOperationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/InternalOperationService.java
@@ -16,7 +16,6 @@
 
 package com.hazelcast.spi.impl;
 
-import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.internal.management.JsonSerializable;
 import com.hazelcast.nio.Address;
 import com.hazelcast.spi.Operation;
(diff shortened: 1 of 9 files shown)


mtopolnik:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/InternalOperationService.java
```diff
@@ -35,15 +34,13 @@
  */
 public interface InternalOperationService extends OperationService {
 
-    void onMemberLeft(MemberImpl member);
```
jerrinot: why is this no longer a part of InternalOperationService? 
it's not a lifecycle event, it's rather a business event.

pveentjer: It doesn't belong in the Interface. 

It depends on the implementation if and how he wants to listen to member left. If he wants to implement an interface or provide some internal class that does the listening, he should be able to do that any way he sees fit. It isn't part of the interface to enforce that. So you are mixing in an implementation detail into the interface. 

So if the BasicOperationService wants to be registered as a MemberListener, he could e.g. implement the MemberListener interface and register himself as listener to the cluster. This can be done in the BasicOperationService when it is constructed.

jerrinot: fair enough

jerrinot:
:+1:

PR #4813, opened 2015-03-16
Adds "origin" parameter to AbstractCacheService#publishEvent

emrahkocaman:
This parameter will be used to propagate event origin across clusters.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
@@ -277,8 +277,7 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
         }
         records.remove(key);
         if (isEventsEnabled) {
-            publishEvent(CacheEventType.EXPIRED, key, null, toEventData(record), false,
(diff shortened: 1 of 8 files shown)


serkan-ozal:
:+1:

emrahkocaman:
verify

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
```diff
@@ -294,7 +294,7 @@ public R processExpiredEntry(Data key, R record, long expiryTime, long now) {
         records.remove(key);
         if (isEventsEnabled) {
             publishEvent(CacheEventType.EXPIRED, key, null, toEventData(record), false,
-                    IGNORE_COMPLETION, CacheRecord.EXPIRATION_TIME_NOT_AVAILABLE);
+                    IGNORE_COMPLETION, CacheRecord.EXPIRATION_TIME_NOT_AVAILABLE, null);
         }
```
mesutcelik: instead of passing null to the method, you can keep old method signature and inside it you can delegate all existing parameters + null for origin field. That seems more clear to me.

emrahkocaman: Sure, done.

PR #4815, opened 2015-03-17
Repacking of the ProxyServiceImpl

pveentjer:
Put the package structure of the ProxyServiceImpl in line with the other SPI services. So there is now an 'proxyservice' package with an InternalProxyService interface. The implementation of this interface is in the 'impl' package. Also a bunch of inner clases are pulled out to make the class a lot less complicated an easier to test.


gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -384,8 +383,8 @@ public void shutdown(final boolean terminate) {
             setActive(false);
             setMasterAddress(null);
             try {
-            	if (groupProperties.SHUTDOWNHOOK_ENABLED.getBoolean())
-            		Runtime.getRuntime().removeShutdownHook(shutdownHookThread);
+                if (groupProperties.SHUTDOWNHOOK_ENABLED.getBoolean())
```
Donnerbart: Are there still missing `{ }` on this `if`?

pveentjer: I guess it is excluded from checkstyle.

PR #4819, opened 2015-03-17
Fix/3.5/issue 3373

kobalski:
System.currentTimeMillis() has non-monotonic behaviour, query test is failing in Nightly builds. sleepAtLeastSeconds() method added to HazelcastTestSupport class. Increasing sleep value did not solve issue #3373. sleepAtLeastSeconds() will make busy wait until System.currentTimeMillis() guarantees 1 second has passed.


kobalski:
Closed due to rebase problem.

PR #4820, opened 2015-03-17
Fix for issue #3373.

kobalski:
System.currentTimeMillis() has non-monotonic behaviour, query test in hibernate is  failing in Nightly builds. Using sleepAtLeastSeconds() method in HazelcastTestSupport class instead of sleep() between each hibernate query in tests. Increasing sleep value did not solve issue #3373. sleepAtLeastSeconds() will make busy wait until 1 second has passed even though the test is running on a fast machine.

### DIFF
--- a/hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/HibernateStatisticsTestSupport.java
+++ b/hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/HibernateStatisticsTestSupport.java
@@ -163,7 +163,7 @@ public abstract class HibernateStatisticsTestSupport extends HibernateTestSuppor
         for (int i = 0; i < queryCount; i++) {
             list = executeQuery(sf);
             assertEquals(entityCount, list.size());
-            sleep(2);
+            sleepAtLeastSeconds(1);
         }
 
         assertNotNull(list);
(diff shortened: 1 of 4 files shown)


Tembrel:
Looking at the implementation of sleepAtLeastMillis (on which
sleepAtLeastSeconds depends):

https://github.com/hazelcast/hazelcast/blob/e7d3d25abaaedda22678676783d2011dbb173d7f/hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java#L174-L179

Why is this repeated 1ms sleep used instead of just
TimeUnit.MILLISECONDS.sleep(millis)? Concern about spurious wakeups or
interruptions? That could be addressed; see Guava's uninterruptible sleep
utility
<http://docs.guava-libraries.googlecode.com/git-history/v18.0/javadoc/src-html/com/google/common/util/concurrent/Uninterruptibles.html#line.267>.
At the very least, use System.nanoTime instead of System.currentTimeMillis
-- the latter (as was pointed out) is non-monotonic, the former *is*
 monotonic.

--tim

mesutcelik:
verify

metanet:
@Tembrel thanks for your comment. sleepAtLeastMillis() is just a test utility method which is used by a few tests. This test is also one of them. I think it would be better to discuss that method in a different issue. Besides that, the current implementation is only an implementation choice. "+1" is used to avoid off-by-1 error and guarantee of time of sleep to be at least the given amount. You are right about the nanoTime() and we can consider that for the implementation.

mesutcelik:
@kobalski ,

Do we need a forward-port ?

kobalski:
Forward port can be seen here : https://github.com/hazelcast/hazelcast/pull/4836

PR #4822, opened 2015-03-17
Do not increment ReplicatedMap's hit counter with internal operations

### LINKED ISSUE #4281: ReplicatedRecord hit counter is increased by ReplicatedMap internally
ReplicatedMap implementation can access ReplicatedRecord instances internally and increment the hit counter. The cause is the getter which automatically increases the hits counter. We should introduce a getter without the call to access() for internal usage to fix this issue.

metanet:
Fixes #4281


eminn:
Looks good to me.

eminn:
This should be also forward ported to the master branch.

serkan-ozal:
:+1:

PR #4823, opened 2015-03-17
Portable enchancement

sancar:
Changed Portable serialization to include field names and types to simplify Data and Packet structures. Instead of keeping a header storing portable-metadata to be able to lookup class-definitions while preparing Packet and executing queries, now each Portable keeps its own json-like metadata in its binary data.

This change increases Data (which containing Portable) size keep in memory but at the same time reduces Packet size sent over the wire. By simplifying Portable serialization procedure, ClassDefinitions are generated on-demand and registered on local for fast access. By doing this, it's possible to serialize Portables in a json-like format and to access single fields with fast lookups using indexes.

Pushing force behind this change is, it's now possible to easily convert Data to a byte[] (or a native memory pointer). There are at least 2 advantages of using a single byte[] (or a native pointer) for serialized data;
- it removes one more memory indirection from the path
- it makes possible to store/load serialized data in single shot


pveentjer:
I don't know if you can rename methods on Data since it is a public API. You can add a new method with the same name and just deprecate the old method. This way you get the nice name, but you don't break backwards compatibility.

PR #4825, opened 2015-03-17
Refactor executor service tests

mtopolnik:
Executor service tests were all piled up in a single file of >1500 lines. This commit splits them into several classes according to the similarity of their fixtures. That allows the removal of the duplication of the setup/teardown code in each test method.


hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
```diff
@@ -0,0 +1,128 @@
...
+import java.lang.reflect.Field;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.Assert.assertEquals;
+
+public class CompletableFutureTest extends HazelcastTestSupport {
```
pveentjer: Can you add the following on top of your test:

```
@Category(QuickTest.class)
```

This way they are going to be picked up by Maven.

pveentjer:
As soon as I have some serious time, I'll give it a serious look.

hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
```diff
@@ -0,0 +1,134 @@
+package com.hazelcast.executor;
```
bilalyasar: missing headers

hazelcast/src/test/java/com/hazelcast/executor/SpecificSetupTest.java
```diff
@@ -0,0 +1,161 @@
...
+@RunWith(HazelcastSerialClassRunner.class)
+@Category(QuickTest.class)
+public class SpecificSetupTest extends ExecutorServiceTestSupport {
+
+    @Test
+    public void testIssue4667() throws ExecutionException, InterruptedException {
+        try {
+//            TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(2);
+            HazelcastInstance
+                    instance1 =
```
bilalyasar: there are some indentation problems

hazelcast/src/test/java/com/hazelcast/executor/SpecificSetupTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Test
+    public void testLongRunningCallable() throws ExecutionException, InterruptedException, TimeoutException {
+        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(2);
+        Config config = new Config();
+        long callTimeout = 3000;
+        config.setProperty(GroupProperties.PROP_OPERATION_CALL_TIMEOUT_MILLIS, String.valueOf(callTimeout));
+        HazelcastInstance hz1 = factory.newHazelcastInstance(config);
+        HazelcastInstance hz2 = factory.newHazelcastInstance(config);
+        IExecutorService executor = hz1.getExecutorService("test");
```
bilalyasar: i  think, using random string for names more preferable

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
```diff
@@ -0,0 +1,346 @@
...
+        Future future = executor.submit(task);
+        //noinspection EmptyCatchBlock
+        try {
+            future.get(1, TimeUnit.MILLISECONDS);
+            fail("Should throw TimeoutException!");
+        } catch (TimeoutException expected) {
+        }
+        assertFalse(future.isDone());
+        assertTrue(future.cancel(true));
+        assertTrue(future.isCancelled());
```
bilalyasar: i think these lines are never executed, they should be deleted.

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
```diff
@@ -0,0 +1,346 @@
...
+    @Test
+    public void testMultipleFutureGets() throws Exception {
+        Callable<String> task = new BasicTestTask();
+        Future<String> future = executor.submit(task);
+        String s1 = future.get();
+        assertEquals(s1, BasicTestTask.RESULT);
+        assertTrue(future.isDone());
+        String s2 = future.get();
+        assertEquals(s2, BasicTestTask.RESULT);
+        assertTrue(future.isDone());
```
bilalyasar: maybe putting a loop here more readable then duplicating lines.

hazelcast/src/test/java/com/hazelcast/executor/SpecificSetupTest.java
mtopolnik: This is intentional; if there is a standard for indenting multiple-assignment statements, I'll correct it to comply. Line breaks after var names are necessary for quick switching between alternatives by commenting/uncommenting line pairs.

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
mtopolnik: I can't see the reason why they are not executed.

hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category(QuickTest.class)
+public class CompletableFutureTest extends HazelcastTestSupport {
```
pveentjer: The test looks clean to me. 

A personal flavor:
I don't use abbreviations for variables names. I would use 'executionService' as name. This makes code easier to understand since you don't need to lookup the type.

hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
```diff
@@ -0,0 +1,134 @@
...
+    private ExecutorService executorService;
+    private AtomicReference<Object> ref1, ref2;
+
+    @Before
+    public void setUp() throws Exception {
+        HazelcastInstanceProxy proxy = (HazelcastInstanceProxy) createHazelcastInstance();
+        Field originalField = HazelcastInstanceProxy.class.getDeclaredField("original");
+        originalField.setAccessible(true);
+        HazelcastInstanceImpl hz = (HazelcastInstanceImpl) originalField.get(proxy);
+        NodeEngine nodeEngine = hz.node.nodeEngine;
```
pveentjer: We have a method 'getNode(HazelcastInstance)' on the HazelcastTestSupport.

hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
```diff
@@ -0,0 +1,134 @@
...
+        doneLatch = new CountDownLatch(2);
+        final ICompletableFuture<String> f = submitOpenStartLatch();
+        assertOpenEventually(startLatch);
+        f.andThen(setRefAndBumpDoneLatch(ref1));
+        f.andThen(setRefAndBumpDoneLatch(ref2));
+        assertOpenEventually(doneLatch);
+        assertEquals("success", ref1.get());
+        assertEquals("success", ref2.get());
+    }
+
```
pveentjer: Can you also add some method that check exceptional behavior? Is the exception propagated correctly? And what about interruption and cancellation?

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
metanet: You are both expecting an exception and making assertions. When the exception is not thrown, you fail with fail() call. Therefore, assert statements are not reached. 

Even if you don't call fail(), when you do not throw CancellationException, test will fail because of the absence of the exception. If it is also expected to fail with one of the assertions when the exception is not thrown, I think that test method can be divided into multiple test methods.

mtopolnik: Since I'm *expecting* the exception, it is a failure not to receive. When I get the exception, I proceed with further assertions. The test cannot be split further because the later assertions are subject to the precondition that we timed out on `future.get()`

metanet: Ok, I think I am missing the method that is expected to throw CancellationException. I thought future.get() is expected to throw it. Isn't it?

mtopolnik: Yes, the last line of the test is expected to throw it.

bilalyasar: :+1: @mtopolnik yes you are right,

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
```diff
@@ -0,0 +1,346 @@
...
+                    futures.get(i).get();
+                    fail();
+                } catch (CancellationException expected) { }
+            }
+        }
+    }
+
+    @Test
+    public void testInvokeAllTimeoutSuccess() throws Exception {
+        assertFalse(executor.isShutdown());
```
bilalyasar: you have already a test `testShutdownBehaviour()` and you have the same assertion. If you are not sure about this assertion, i guess you have to write new test for this.

hazelcast/src/test/java/com/hazelcast/executor/SpecificSetupTest.java
Donnerbart: What about a (final) boolean field to switch the behavior a bit more convenient? If both variants are important we maybe should even test both (by creating a configurable private method and call it from 2 test methods).

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
Donnerbart: Maybe the `//noinspection EmptyCatchBlock` can be replaced with `EmptyStatement(expected);` I've seen somewhere else.

hazelcast/src/test/java/com/hazelcast/executor/ExecutorServiceTestSupport.java
```diff
@@ -0,0 +1,197 @@
...
+        return createSingleNodeExecutorService(name, ExecutorConfig.DEFAULT_POOL_SIZE);
+    }
+
+    IExecutorService createSingleNodeExecutorService(String name, int poolSize) {
+        final HazelcastInstance instance = createHazelcastInstance(
+                new Config().addExecutorConfig(new ExecutorConfig(name, poolSize)));
+        return instance.getExecutorService(name);
+    }
+
+    static class ScriptRunnable implements Runnable, Serializable, HazelcastInstanceAware {
```
pveentjer: Perhaps an idea to move this inner class out of the ExecutorServiceTestSupport? The same goes for the  ScriptCallable. This makes the test easier to read since you don't need to deal with these details.

pveentjer: Forget what I said.. this is ExecutorServiceTestSupport; right place for such a class.

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
pveentjer: Why do you think these lines are not executed?

hazelcast/src/test/java/com/hazelcast/executor/SpecificSetupTest.java
mtopolnik: No, only one setup reproduces the problem. This whole test case is questionable and will probably have to be removed. It's probably impossible to fix this behavior.

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
```diff
@@ -0,0 +1,346 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category(QuickTest.class)
+public class SingleNodeTest extends ExecutorServiceTestSupport {
```
pveentjer: Would it be possible to set up this test in the same was as the AtomicLongBasicTest, AtomicLongBasicLocalTest/AtomicLongBasicDistributedTest are set up? This way we also get all the serialization/deserialization.

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
```diff
@@ -0,0 +1,346 @@
...
+        HazelcastInstanceAwareRunnable task = new HazelcastInstanceAwareRunnable();
+        try {
+            executor.submit(task).get();
+        } catch (ExecutionException e) {
+            throw (Error)e.getCause();
+        }
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void submitNullTask() throws Exception {
```
pveentjer: Perhaps best to have one style in the method naming. Some methods begin with test and some do not. I would try to keep this in a single test file in the same style. 

I would call it 
submit_whenNullTask

Because often you get a lot of scenario's and using this approach is becomes easier to deal with them:

submit_whenShutdown
submit_whenQueueFull
submit_when...

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
bilalyasar: currently i don't, probably i was confused while examining the code..

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
```diff
@@ -0,0 +1,346 @@
...
+        }
+        assertFalse(future.isDone());
+        assertTrue(future.cancel(true));
+        assertTrue(future.isCancelled());
+        assertTrue(future.isDone());
+        future.get();
+    }
+
+    @Test(expected = CancellationException.class)
+    public void testCancellationAwareTask2() throws ExecutionException, InterruptedException {
```
pveentjer: I don't like naming with a number in them. Apparently it is a different test then testCancellationAwareTask, so perhaps add that in the name. For me it is not clear what the purpose of the test is when I read the name.

hazelcast/src/test/java/com/hazelcast/executor/SpecificSetupTest.java
```diff
@@ -0,0 +1,161 @@
...
+@Category(QuickTest.class)
+public class SpecificSetupTest extends ExecutorServiceTestSupport {
+
+    @Test
+    public void testIssue4667() throws ExecutionException, InterruptedException {
+        try {
+//            TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(2);
+            HazelcastInstance
+                    instance1 =
+                            Hazelcast.newHazelcastInstance(),
```
pveentjer: Probably best to make use of 

HazelcastInstance[] cluster = HazelcastTestSupport.createHazelcastInstanceFactory(2).newInstances();

The HazelcastTestSupport creates 'faster' instances with afaik mocked networking.

hazelcast/src/test/java/com/hazelcast/executor/SmallClusterTest.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category(QuickTest.class)
+public class SmallClusterTest extends ExecutorServiceTestSupport {
```
pveentjer: All these tests can be added to the ExecutorServiceBasicTest and then you can automatically have the same tests run on a cluster and on a pure local setup. 

If you think the test becomes to long, you can set up multiple of such hierarchies. E.g. ExecutorServiceSomeFeatureBasicTest, ExecutorServiceSomeFeatureLocalTest etc.

hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
mtopolnik: Done, also further refactored the test.

hazelcast/src/test/java/com/hazelcast/executor/SpecificSetupTest.java
mtopolnik: This is about a problem which the test factory cannot reproduce. The entire test will be removed, probably, as it's unfixable.

pveentjer: ok.

hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
mtopolnik: Did a round of renaming test methods.

pveentjer:
I think this is a good round of cleanup on the executor tests. 

Can you do a rebase?

If nobody is objecting, I'll hit the merge button after the rebase and build is green.

mtopolnik:
I still plan to implement your idea with doing both single-node and multinode tests, so let's keep it open for a while longer.

pveentjer:
Great thanks. Do you know the current test coverage?

pveentjer:
Sorry for hitting the wrong button.

mtopolnik:
I have no idea about coverage :)

pveentjer:
verify

pveentjer:
Can you do a rebase?

mtopolnik:
verify

mtopolnik:
verify

PR #4832, opened 2015-03-18
Cleanup BasicOperationService

pveentjer:
Pulled out a lot of inner class in the operation service impl
Renamed classes to Impl instead of Basic so that it in line with other components

Pulling out these inner classes makes the code a lot less complicated since you don't need to go to a 1k+ lines file. It is going to make testing a lot easier since you can now test a class directly instead of only indirectly.

I moved classes and or renamed them mostly. Pulled out a few methods as well. A bit shitty that git thinks I have deleted the BasicOperationService and created the OperationServiceImpl.


gurbuzali:
needs rebase

PR #4833, opened 2015-03-18
Forward-port of PR #4822

### LINKED ISSUE #4281: ReplicatedRecord hit counter is increased by ReplicatedMap internally

metanet:
Fixes #4281 

forward-port of #4822


metanet:
verify

pveentjer:
@ahmetmircik can you review this PR since you are the map module owner.

mdogan:
verify

PR #4838, opened 2015-03-19
added async invocation to operation-service and invocation-builder

gurbuzali:
Reasoning behind this enhancement is to be able to make non-blocking invocation inside PartitionThreads.
Currently all invocations which are targeted to different partition-threads (partition aware operations which don't not fall to same partition-thread) throws exception to prevent dead-lock issues and thread-hog issues.
With this async invocation, it will be possible to send an invocation from partition-threads. 
Since it does not return any future to block on, there will not be dead-lock and thread-hog issues.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
@@ -17,6 +17,8 @@
 package com.hazelcast.spi.impl.operationservice;
 
 import com.hazelcast.internal.management.JsonSerializable;
+import com.hazelcast.nio.Address;
+import com.hazelcast.spi.Callback;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.OperationService;
 import com.hazelcast.spi.impl.PartitionSpecificRunnable;
@@ -67,4 +69,8 @@ public interface InternalOperationService extends OperationService {
      * @return collection of long running operation logs.
      */
     Collection<JsonSerializable> getSlowOperations();
+
+    <E> void invokeOnPartition(String serviceName, Operation op, int partitionId, Callback<E> callback);
+
+    <E> void invokeOnTarget(String serviceName, Operation op, Address target, Callback<E> callback);
 }
--- a/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
@@ -229,4 +229,6 @@ public abstract class InvocationBuilder {
     }
 
     public abstract InternalCompletableFuture invoke();
+
+    public abstract void invokeAsync();
 }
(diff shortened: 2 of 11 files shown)


pveentjer:
Hi Ali, I hope I can find some time today to look at this. I sounds like a useful feature to have.

gurbuzali:
yes, I need this for commit/rollback an XA transaction remotely, in case of failure.
the other option is to offload to another thread/executor-service and pass the respond-handler

ajermakovics:
Cool stuff! I think it'll be useful for me as well.
What happens if there is an exception while running the operation in async mode? I can see how a caller might be interested in that as well.

gurbuzali:
the exception will be passed to `notify` method of `Callback` just like a normal response so you have to check if it is throwable or not.

ajermakovics:
Gotcha. If we had pattern matching then that would be easy to handle:). Not sure, feels like ExecutionCallback would fit in nicer.

pveentjer:
I need to think about this one. 

First of all: your feature makes sense. No doubt about that. 

The issue is that I need similar behavior in the future to do all kinds of optimization. I need to know if certain calls are going to be done async or sync. So if a future.andThen or a future.get is done. 

Another pain is that we now get a permutation problem with invocation methods on the OperationService. I would rather have a single method. Here is some pseudo code:

```
        InvocationFuture f = operationService.newInvocationFuture(op);
        f.setPartition(25);
        f.setTimeout(...);
        f.setTryCount(...);
        // for async invocation
        f.setMode(Async)<<---- important
        f.andThen(completionHandler);
        f.invoke();

        InvocationFuture f = operationService.newInvocationFuture(op);
        f.setMember(...)
        f.setTimeout(...);
        f.setTryCount(...);
        f.setMode(Sync)<<---- important
        f.invoke();
       Object result = f.get()
```

The key difference is that currently we do the invoke in the OperationService as part of invocation method. But now the invoke is delayed and you can play with the setting on the invocation future and once you are happy with the settings, you call invoke. 

In the new progressive OperationExecutor I also need to how a future is going to behave. E.g. if it is an async call, the current thread should not steal the partition. Also most of our futures don't leak to the outside, so there will only be a single thread doing .get. So if caller runs optimization was applied to an operation (so the invoking thread processed his own operation) and there is a single getter (the thread self), I can skip the whole notifyAll thing in the invocation future.

I'm fine with merging this if you can integrate it in the InternalOperationService, so that it remains an internal API and we can clean up the interface when we have time for it.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
```diff
@@ -67,4 +69,8 @@
      * @return collection of long running operation logs.
      */
     Collection<JsonSerializable> getSlowOperations();
+
+    <E> void invokeOnPartition(String serviceName, Operation op, int partitionId, Callback<E> callback);
```
pveentjer: Can you rename them to asyncOnvokeOn... This makes it immediately clear that it is async.

pveentjer:
Looks good. Don't you want to control the executor you are going to use to offload the completion handler to?

pveentjer:
Can you rename the methods to async.... so that it is more clear?

PR #4839, opened 2015-03-19
Enhance IDE integration

mathieu-pousse:
Until now, when using the hazelcast spring namespace, the IDEs were not able to know what type of beans were created when using <hz:client/>, <hz:map/>, ...

I have added the requiered information in the xsd to seriously enhance the spring integration with IntelliJ for instance.

Before in IntelliJ: 
image
image

Now: 
image

This is more or less the same with all kind of hz objects.


mdogan:
:+1:

pveentjer:
Thumbs up here as well.

noctarius:
Same from me, pretty cool thing, thanks for the effort! :)

mathieu-pousse:
was a pleasure !

Issue #4753, opened 2015-03-06
regrestion,  can't use *  in xml Jcache config names.  3.5-snapshot


Danny-Hazelcast:
we should add a junit test for this,  

```
Failure[
   message='Worked ran into an unhandled exception'
   type='Worker exception'
   agentAddress=10.0.0.134
   time=Fri Mar 06 09:50:30 UTC 2015
   workerAddress=10.0.0.134:5701
   workerId=worker-10.0.0.134-1-server
   test=TestCase{
          id=loadCache
        , class=com.LoadCache
        , cacheBaseName=j
    }
   cause=java.lang.IllegalArgumentException: CacheConfig is null! /hz/j0
	at com.hazelcast.cache.EnterpriseCacheService.createNewRecordStore(EnterpriseCacheService.java:82)
	at com.hazelcast.cache.impl.CachePartitionSegment$1.createNew(CachePartitionSegment.java:56)
	at com.hazelcast.cache.impl.CachePartitionSegment$1.createNew(CachePartitionSegment.java:53)
	at com.hazelcast.util.ConcurrencyUtil.getOrPutSynchronized(ConcurrencyUtil.java:40)
	at com.hazelcast.cache.impl.CachePartitionSegment.getOrCreateCache(CachePartitionSegment.java:76)
	at com.hazelcast.cache.impl.AbstractCacheService.getOrCreateCache(AbstractCacheService.java:103)
	at com.hazelcast.cache.hidensity.operation.AbstractHiDensityCacheOperation.beforeRun(AbstractHiDensityCacheOperation.java:76)
	at com.hazelcast.spi.impl.BasicOperationService$BasicOperationRunner.run(BasicOperationService.java:732)
	at com.hazelcast.spi.impl.BasicOperationService$BasicOperationRunner.run(BasicOperationService.java:896)
	at com.hazelcast.spi.impl.operationexecutor.classic.OperationThread.processPacket(OperationThread.java:126)
	at com.hazelcast.spi.impl.operationexecutor.classic.OperationThread.process(OperationThread.java:99)
	at com.hazelcast.spi.impl.operationexecutor.classic.OperationThread.doRun(OperationThread.java:85)
	at com.hazelcast.spi.impl.operationexecutor.classic.OperationThread.run(OperationThread.java:60)
	at ------ End remote and begin local stack-trace ------.(Unknown Source)
[...]

PR #4840, opened 2015-03-19
JCache config tests

cherednichenko:
test's for issue #4753

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
@@ -24,9 +24,7 @@ import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.annotation.SlowTest;
-import org.junit.After;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.*;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
@@ -34,82 +32,89 @@ import javax.cache.Cache;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 
 @RunWith(HazelcastSerialClassRunner.class)
 @Category(SlowTest.class)
-public class CacheCreationTest {
+public class CacheCreationTest extends HazelcastTestSupport {
 
-    private final URL configUrl1 = getClass().getClassLoader().getResource("test-hazelcast-real-jcache.xml");
-    Config hzConfig;
+    static Config hzConfig;
+    private static final int THREAD_COUNT = 4;
 
-    @Before
-    public void init() throws Exception{
-        Hazelcast.shutdownAll();
+    @BeforeClass
+    public static void init() throws Exception{
+        final URL configUrl1 = CacheCreationTest.class.getClassLoader().getResource("test-hazelcast-real-jcache.xml");
         XmlConfigBuilder configBuilder = new XmlConfigBuilder(configUrl1.getFile());
         hzConfig = configBuilder.build();
     }
 
+    @Before
     @After
-    public void cleanup() throws Exception{
+    public void killAllHazelcastInstances() throws Exception {
         Hazelcast.shutdownAll();
     }
 
     @Test
     public void creatingASingleCache() throws URISyntaxException {
-
-        HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(hzConfig);
-        HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
-                .createCachingProvider(hazelcastInstance);
-
+        HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
         Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache" + 1);
         cache.get(1);
     }
 
-
     @Test
-    public void creatingASingleCacheFromMultiProviders() throws URISyntaxException {
+    public void creatingASingleCacheFromMultiProviders() throws URISyntaxException, InterruptedException {
 
-        int threadCount=4;
+        ExecutorService execSer = Executors.newFixedThreadPool(THREAD_COUNT);
 
-        final CountDownLatch latch = new CountDownLatch(threadCount);
-        for (int i = 0; i < threadCount; i++) {
-            final int finalI = i;
-            new Thread() {
+        final CountDownLatch latch = new CountDownLatch(THREAD_COUNT);
+        for (int i = 0; i < THREAD_COUNT; i++) {
+            execSer.execute(new Runnable() {
+                @Override
                 public void run() {
-                    HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(hzConfig);
-                    HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
-                            .createCachingProvider(hazelcastInstance);
-
-                    Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache" );
+                    HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
+                    Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache");
                     cache.get(1);
                     latch.countDown();
                 }
-            }.start();
+            });
         }
         HazelcastTestSupport.assertOpenEventually(latch);
+        if (!execSer.awaitTermination(3, TimeUnit.SECONDS)) {
+            execSer.shutdown();
+        }
     }
 
     @Test
-    public void creatingMultiCacheFromMultiProviders() throws URISyntaxException {
+    public void creatingMultiCacheFromMultiProviders() throws URISyntaxException, InterruptedException {
 
-        int threadCount=4;
+        ExecutorService execSer = Executors.newFixedThreadPool(THREAD_COUNT);
 
-        final CountDownLatch latch = new CountDownLatch(threadCount);
-        for (int i = 0; i < threadCount; i++) {
+        final CountDownLatch latch = new CountDownLatch(THREAD_COUNT);
+        for (int i = 0; i < THREAD_COUNT; i++) {
             final int finalI = i;
-            new Thread() {
+            execSer.execute(new Runnable() {
+                @Override
                 public void run() {
-                    HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(hzConfig);
-                    HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
-                            .createCachingProvider(hazelcastInstance);
-
+                    HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
                     Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache"+finalI);
                     cache.get(1);
                     latch.countDown();
                 }
-            }.start();
+            });
         }
         HazelcastTestSupport.assertOpenEventually(latch);
+        if (!execSer.awaitTermination(3, TimeUnit.SECONDS)) {
+            execSer.shutdown();
+        }
+    }
+
+    private HazelcastServerCachingProvider createCachingProvider(Config hzConfig){
+        HazelcastInstance hazelcastInstance = createHazelcastInstance(hzConfig);
+        HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
+                .createCachingProvider(hazelcastInstance);
+        return cachingProvider;
     }
 }
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
@@ -21,6 +21,7 @@ import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.instance.HazelcastInstanceFactory;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.annotation.SlowTest;
@@ -34,8 +35,6 @@ import java.net.URL;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-
 
 @RunWith(HazelcastSerialClassRunner.class)
 @Category(SlowTest.class)
@@ -54,7 +53,7 @@ public class CacheCreationTest extends HazelcastTestSupport {
     @Before
     @After
     public void killAllHazelcastInstances() throws Exception {
-        Hazelcast.shutdownAll();
+        HazelcastInstanceFactory.shutdownAll();
     }
 
     @Test
@@ -66,12 +65,11 @@ public class CacheCreationTest extends HazelcastTestSupport {
 
     @Test
     public void creatingASingleCacheFromMultiProviders() throws URISyntaxException, InterruptedException {
-
-        ExecutorService execSer = Executors.newFixedThreadPool(THREAD_COUNT);
+        ExecutorService executorService = Executors.newFixedThreadPool(THREAD_COUNT);
 
         final CountDownLatch latch = new CountDownLatch(THREAD_COUNT);
         for (int i = 0; i < THREAD_COUNT; i++) {
-            execSer.execute(new Runnable() {
+            executorService.execute(new Runnable() {
                 @Override
                 public void run() {
                     HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
@@ -82,20 +80,17 @@ public class CacheCreationTest extends HazelcastTestSupport {
             });
         }
         HazelcastTestSupport.assertOpenEventually(latch);
-        if (!execSer.awaitTermination(3, TimeUnit.SECONDS)) {
-            execSer.shutdown();
-        }
+        executorService.shutdown();
     }
 
     @Test
-    public void creatingMultiCacheFromMultiProviders() throws URISyntaxException, InterruptedException {
-
-        ExecutorService execSer = Executors.newFixedThreadPool(THREAD_COUNT);
+    public void creatingMultiCacheFromMultiProviders() throws Exception {
+        ExecutorService executorService = Executors.newFixedThreadPool(THREAD_COUNT);
 
         final CountDownLatch latch = new CountDownLatch(THREAD_COUNT);
         for (int i = 0; i < THREAD_COUNT; i++) {
             final int finalI = i;
-            execSer.execute(new Runnable() {
+            executorService.execute(new Runnable() {
                 @Override
                 public void run() {
                     HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
@@ -106,9 +101,7 @@ public class CacheCreationTest extends HazelcastTestSupport {
             });
         }
         HazelcastTestSupport.assertOpenEventually(latch);
-        if (!execSer.awaitTermination(30, TimeUnit.SECONDS)) {
-            execSer.shutdown();
-        }
+        executorService.shutdown();
     }
 
     private HazelcastServerCachingProvider createCachingProvider(Config hzConfig){
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
@@ -19,7 +19,6 @@ package com.hazelcast.cache;
 import com.hazelcast.cache.impl.HazelcastServerCachingProvider;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
-import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.instance.HazelcastInstanceFactory;
 import com.hazelcast.test.HazelcastSerialClassRunner;
@@ -28,7 +27,6 @@ import com.hazelcast.test.annotation.SlowTest;
 import org.junit.*;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
-
 import javax.cache.Cache;
 import java.net.URISyntaxException;
 import java.net.URL;
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
@@ -106,7 +106,7 @@ public class CacheCreationTest extends HazelcastTestSupport {
             });
         }
         HazelcastTestSupport.assertOpenEventually(latch);
-        if (!execSer.awaitTermination(3, TimeUnit.SECONDS)) {
+        if (!execSer.awaitTermination(30, TimeUnit.SECONDS)) {
             execSer.shutdown();
         }
     }
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


bilalyasar:
lets run tests again, because schema validation PR has been merged.
verify

cherednichenko:
Should I run those tests again - I didn't catch?

Actually those tests are passing after Serkan fixed #4753 issue - we re-run with Danny before creating this PR.

bilalyasar:
no, after writing [v]erify, jenkins automatically triggers new build.
and seems it passed. :+1:

cherednichenko:
Ok. :+1:

pveentjer:
@serkan-ozal can you have a look at this PR.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
+                            .createCachingProvider(hazelcastInstance);
+
+                    Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache" );
+                    cache.get(1);
+                    latch.countDown();
+                }
+            }.start();
+        }
+        HazelcastTestSupport.assertOpenEventually(latch);
```
serkan-ozal: @cherednichenko even if this test fails (I know it shouldn't fail :) ) we should be sure that all threads are terminated.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
+                            .createCachingProvider(hazelcastInstance);
+
+                    Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache"+finalI);
+                    cache.get(1);
+                    latch.countDown();
+                }
+            }.start();
+        }
+        HazelcastTestSupport.assertOpenEventually(latch);
```
serkan-ozal: @cherednichenko even if this test fails (I know it shouldn't fail :) ) we should be sure that all threads are terminated.

hazelcast/src/test/resources/test-hazelcast-real-jcache.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
serkan-ozal: Is this `test-hazelcast-real-jcache.xml` used only by this test ? If it is, it may better to clean unnecessary configurations from this xml configuration to make it clean.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
+@Category(SlowTest.class)
+public class CacheCreationTest {
+
+    private final URL configUrl1 = getClass().getClassLoader().getResource("test-hazelcast-real-jcache.xml");
+    Config hzConfig;
+
+    @Before
+    public void init() throws Exception{
+        Hazelcast.shutdownAll();
+        XmlConfigBuilder configBuilder = new XmlConfigBuilder(configUrl1.getFile());
```
serkan-ozal: `hzConfig` may be initialized only once in another method `@BeforeClass` and can be defined static.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @After
+    public void cleanup() throws Exception{
+        Hazelcast.shutdownAll();
+    }
+
+    @Test
+    public void creatingASingleCache() throws URISyntaxException {
+
+        HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(hzConfig);
```
serkan-ozal: Can't we use test Hazelcast instance via `HazelcastTestSupport` ? Do we really need network stuff for this test ?

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+
+    @Test
+    public void creatingASingleCacheFromMultiProviders() throws URISyntaxException {
+
+        int threadCount=4;
+
+        final CountDownLatch latch = new CountDownLatch(threadCount);
+        for (int i = 0; i < threadCount; i++) {
+            final int finalI = i;
```
serkan-ozal: Seems no need to `finalI` for this test

hazelcast/src/test/resources/test-hazelcast-real-jcache.xml
```diff
@@ -0,0 +1,80 @@
...
+    </cache>
+
+    <cache name="testCache">
+        <key-type class-name="java.lang.Integer"/>
+        <value-type class-name="java.lang.String"/>
+        <statistics-enabled>false</statistics-enabled>
+        <management-enabled>false</management-enabled>
+    </cache>
+
+    <cache name="wanRefTestCache">
```
serkan-ozal: I think, `wanRefTestCache` cache config is not needed if this xml configuration file is not used by other tests.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
+            new Thread() {
+                public void run() {
+                    HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(hzConfig);
+                    HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
```
serkan-ozal: Could you make this tests also for client based cache provider ? If you think, it is not scope of this test, could you create another test or test file ?

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
```
serkan-ozal: Could you make this tests also for client based cache provider ? If you think, it is not scope of this test, could you create another test or test file ?

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,115 @@
...
+    @After
+    public void cleanup() throws Exception{
+        Hazelcast.shutdownAll();
+    }
+
+    @Test
+    public void creatingASingleCache() throws URISyntaxException {
+
+        HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(hzConfig);
+        HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider
```
serkan-ozal: Could you make this tests also for client based cache provider ? If you think, it is not scope of this test, could you create another test or test file ?

pveentjer:
@hasancelik can you help @cherednichenko out to resolve the review comments?

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @serkan-ozal Could you consult me how it could be done? Because I can't find suitable form for it in our code. 
I quite understand what exactly should be checked BUT your advise is highly appreciated.

Thanks a lot.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @serkan-ozal Could you consult me how it could be done? Because I can't find suitable form for it in our code. 
I quite understand what exactly should be checked BUT your advise is highly appreciated.

p.s. the same as for previous comment.

hazelcast/src/test/resources/test-hazelcast-real-jcache.xml
cherednichenko: Seems yes - it uses only for this test as as I remember :) 

Do you mean remove all and keep only this part ? 

```
 <cache name="xmlCache*">
    </cache>
```

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: I added:

```
@Category(SlowTest.class)
public class CacheCreationTest extends HazelcastTestSupport {

    static Config hzConfig;

    @BeforeClass
    public void init() throws Exception{
        final URL configUrl1 = getClass().getClassLoader().getResource("test-hazelcast-real-jcache.xml");
        XmlConfigBuilder configBuilder = new XmlConfigBuilder(configUrl1.getFile());
        hzConfig = configBuilder.build();
    }
```

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: Added:

```
@Category(SlowTest.class)
public class CacheCreationTest extends HazelcastTestSupport {
   .......
}
```

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: removed.

hazelcast/src/test/resources/test-hazelcast-real-jcache.xml
cherednichenko: It was merged I believe when we pushed/committed.

Removed.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
serkan-ozal: @cherednichenko `@BeforeClass` annotations must be tagged to static methods. So `init` method must be static

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: I believe that it would be better to create as different commit.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
serkan-ozal: :+1:

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: I believe that it would be better to create as different commit and different test for client.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: I believe that it would be better to create as different commit and different test for client.

cherednichenko:
@pveentjer @hasancelik helped me and we did all changes except one - related to terminated threads.

I'll finish and commit my changes.

Thanks a lot guys.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
serkan-ozal: @cherednichenko you may use `java.util.concurrent.ExecutorService` to wait threads to finish with a timeout with `awaitTimeout` method. If they still don't terminate, you may use `shutdown`

hazelcast/src/test/resources/test-hazelcast-real-jcache.xml
serkan-ozal: yep

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
serkan-ozal: ok

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
serkan-ozal: ok

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @serkan-ozal  done.

hazelcast/src/test/resources/test-hazelcast-real-jcache.xml
cherednichenko: done.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @serkan-ozal  Yeap - you're right :)
Done.

cherednichenko:
I made all changes based on comments to this PR, did some optimisations, checked/run 'checkstyle' and 'findbugs', run all tests -- everything is GREEN, no issues/failures.

Thanks a lot guys ! :+1:

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,120 @@
...
+                public void run() {
+                    HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
+                    Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache"+finalI);
+                    cache.get(1);
+                    latch.countDown();
+                }
+            });
+        }
+        HazelcastTestSupport.assertOpenEventually(latch);
+        if (!execSer.awaitTermination(3, TimeUnit.SECONDS)) {
```
serkan-ozal: can we make it `30` seconds. `3` seconds may not be sufficient since it is also created an hazelcast instance

cherednichenko: Done.

p.s. By the way, this small change fixes issue which I've seen:
```
when I run one test after another (within this class) - there is some situation which is similar to this error occurred before -- "Node factory is already created!". But when I waited +1-2-3 secs more --- all tests are passed.
```

serkan-ozal:
from my side, it is ok. :+1:

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,120 @@
...
+                public void run() {
+                    HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
+                    Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache"+finalI);
+                    cache.get(1);
+                    latch.countDown();
+                }
+            });
+        }
+        HazelcastTestSupport.assertOpenEventually(latch);
+        if (!execSer.awaitTermination(30, TimeUnit.SECONDS)) {
```
serkan-ozal: sorry, I forgot to mention that you may also move wait and shutdown codes to a finally block to be sure that even at failure case, these will be shutdown.

pveentjer: Just put it in the @Before/@After methods of a test. I also see that you are doing 

assertEquals(0, HazelcastClient.getAllHazelcastClients().size());

This is also highly undesirable. Because it make a test difficult to understand and apparently you are not trusting something and you let this lack of confidence leak in your test. If could very well be that there is an issue; if so.. it needs to be addressed. Otherwise you should only rely on before/after methods to set your tests in a valid state.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,120 @@
...
+    public void creatingASingleCache() throws URISyntaxException {
+        HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
+        Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache" + 1);
+        cache.get(1);
+    }
+
+    @Test
+    public void creatingASingleCacheFromMultiProviders() throws URISyntaxException, InterruptedException {
+
+        ExecutorService execSer = Executors.newFixedThreadPool(THREAD_COUNT);
```
pveentjer: Please don't use partial names because it makes code very difficult to understand. Just call it executorService or executor.

pveentjer: Also don't add constants if they have meaning for only a single test. In this case the number of threads you are using is not relevant for the rest of the system. So just declare a local variable ' int threadcount = 10' or something.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @Test
+    public void creatingASingleCache() throws URISyntaxException {
+        HazelcastServerCachingProvider cachingProvider = createCachingProvider(hzConfig);
+        Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache("xmlCache" + 1);
+        cache.get(1);
+    }
+
+    @Test
+    public void creatingASingleCacheFromMultiProviders() throws URISyntaxException, InterruptedException {
+
```
pveentjer: Don't use empty white lines at the beginning of a method.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,120 @@
...
+@RunWith(HazelcastSerialClassRunner.class)
+@Category(SlowTest.class)
+public class CacheCreationTest extends HazelcastTestSupport {
+
+    static Config hzConfig;
+    private static final int THREAD_COUNT = 4;
+
+    @BeforeClass
+    public static void init() throws Exception{
+        final URL configUrl1 = CacheCreationTest.class.getClassLoader().getResource("test-hazelcast-real-jcache.xml");
```
pveentjer: Don't number local variables.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
pveentjer: This code is not correct. You can't wait for termination and then do a shutdown. 

So delete the if...shudown part and just say shutdown. That is sufficient.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
pveentjer: The test name is not really revealing what you are doing. I'm confused about why you need multiple threads to do what you are doing. I think this test can be simplified by removing the threads completely.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+        HazelcastTestSupport.assertOpenEventually(latch);
+        if (!execSer.awaitTermination(3, TimeUnit.SECONDS)) {
+            execSer.shutdown();
+        }
+    }
+
+    @Test
+    public void creatingMultiCacheFromMultiProviders() throws URISyntaxException, InterruptedException {
+
```
pveentjer: No empty lines.

pveentjer: In case of test I just say 'throws Exception' I don't care which exceptions are thrown. They are all going to cause the test to fail.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
serkan-ozal: if a variable's value is not changed, it should be defined as contant. this is common/suggested usecase

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
pveentjer: I think it would be better of this test does not make use of a XML file to get the configuration from; my experience is that having all these test xml files make the system very unmaintainable. Just make a string to stick the configuration in and if you need to have a url, just create a tmp file.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
pveentjer: The problem is that when you get a test with a need for different values for the same 'constant', you run into a problems. Now you are forcing everyone to use 10 thread or whatever number you have defined. If you want to add a test that only need 2 threads, then you add ANOTHER_THREAD_COUNT constant?

serkan-ozal: I don't still understand your objection. This constant is defined in this test method's scope. Every test methods can use their constants in themselves. There is no enforcement to use 10 threads for every one.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: Ok, I'm removing this part:
```
if (!execSer.awaitTermination(3, TimeUnit.SECONDS)) {
            executorService.shutdown();
        }
```

and just putting this one:
```
executorService.shutdown();
```

p.s. and I renamed to 'executorService' as you're right (yesterday I thought about it but hurried up and didn't do it).

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @pveentjer Where did you find this piece of code?  There is no such code at all.

```
assertEquals(0, HazelcastClient.getAllHazelcastClients().size());
```

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @pveentjer
1 - "executorService" was renamed - you're right (I thought about it yesterday too).
2 - before we had "int threadCount=4;" (https://github.com/cherednichenko/hazelcast/commit/108d2c363788624ce4435fbb462b196de4339057#diff-e54d362d7e7f73ac05367fd9784cfe95R73 ) but during discussion/commenting it was removed and rewritten. Should I revert back local 'thread count' ?

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @pveentjer 
1 - Do you mean at line 69, right ?  Removed.
2 - I'm not sure that I understand - if I remove the threads completely - what would it kept here?

We have just threes tests here:
1:
```@Test
    public void creatingASingleCache() throws URISyntaxException {......} ```

2:
``` @Test
    public void creatingASingleCacheFromMultiProviders() throws URISyntaxException, InterruptedException {}```

3:
``` @Test
    public void creatingMultiCacheFromMultiProviders() throws URISyntaxException, InterruptedException {}```

all those tests for a bit different goals: for single cache, for single + thread count from multi providers (```.getCache("xmlCache");```) and third one for multi cache + + thread count from multi providers (```.getCache("xmlCache"+finalI);```). Honestly the initial idea was from @Danny-Hazelcast and maybe he wants to add something but I said how I understand it - correct me if I'm wrong.

Thanks.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @pveentjer 
1 - about local variables - answers/discussions is below but this row ```private static final int THREAD_COUNT = 4;``` was added after reviewing/commenting this PR, initially it was local count ```int threadCount=4;``` .
2 - But these tests were created based on real issues which happened/related to .xml configuration that's why actually the tests are using xml configurations. If we rewrote it - it looses initial sense of it, doesn't it? The issue was mentioned at the top of PR (#4753). But if I didn't understand you and what you suggest is within scope of that bug - just give me a bit more information how to do it.

Thanks.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
cherednichenko: @pveentjer 
1 - removed
2 - Added 'Exception' like: ```public void creatingMultiCacheFromMultiProviders() throws Exception { ```

Thanks.

hazelcast/src/test/java/com/hazelcast/cache/CacheCreationTest.java
pveentjer: I think I got confused with a different PR.

cherednichenko: Ok.

cherednichenko:
@pveentjer @serkan-ozal Guys - I committed new updates - let's define what should I do with 'thread_local' -- should it be local or global ?

Thanks.

serkan-ozal:
@cherednichenko I am OK with both of two ways (means not a big problem from my side) and choosing one of them should not prevent this PR to be merged.

cherednichenko:
@serkan-ozal I'm also OK with the both versions. 
@pveentjer Just convince me about - should I change to local 'threadCount' variable or keep as is, global one ?

pveentjer:
@cherednichenko keep the constant. Not relevant for now.

PR #4843, opened 2015-03-19
Added max result size limit for IMap.values(), IMap.keySet() and IMap.entrySet()

Donnerbart:
Added max result size limit for IMap.values(), IMap.keySet() and IMap.entrySet(). Will also limit the result size of Query operations.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -17,7 +17,11 @@
 package com.hazelcast.instance;
 
 import com.hazelcast.config.Config;
+import com.hazelcast.core.IMap;
 import com.hazelcast.internal.monitors.HealthMonitorLevel;
+import com.hazelcast.map.QueryResultSizeExceededException;
+import com.hazelcast.map.impl.MapQueryResultSizeLimitHelper;
+import com.hazelcast.query.TruePredicate;
 
 /**
  * The GroupProperties contain the Hazelcast properties. They can be set as an environmental variable, or
@@ -276,6 +280,41 @@ public class GroupProperties {
      */
     public static final String PROP_JCACHE_PROVIDER_TYPE = "hazelcast.jcache.provider.type";
 
+    /**
+     * Result size limit for query operations on maps.
+     * <p/>
+     * This value defines the maximum number of returned elements for a single query result. If a query exceeds this number of
+     * elements a {@link QueryResultSizeExceededException} will be thrown.
+     * <p/>
+     * This feature is in place to prevent an OOME if a single node is requesting the whole data set of the cluster, e.g. by
+     * executing a query with {@link TruePredicate}. This applies internally for the {@link IMap#values()}, {@link IMap#keySet()}
+     * and {@link IMap#entrySet()} methods, which are good candidates for OOME in large clusters.
+     * <p/>
+     * This feature depends on an equal distribution of the data on the cluster nodes to calculate the result size limit per node.
+     * Therefore there is a minimum value of {@value MapQueryResultSizeLimitHelper#MINIMUM_MAX_RESULT_LIMIT} defined in
+     * {@link MapQueryResultSizeLimitHelper}. Configured values below the minimum will be increased to the minimum.
+     * <p/>
+     * The feature can be disabled by setting a value of <tt>-1</tt> (which is the default value).
+     */
+    public static final String PROP_QUERY_RESULT_SIZE_LIMIT = "hazelcast.query.result.size.limit";
+
+    /**
+     * Maximum value of local partitions to trigger local pre-check for {@link TruePredicate} query operations on maps.
+     * <p/>
+     * To limit the result size of a query ({@see PROP_QUERY_RESULT_SIZE_LIMIT}) a local pre-check on the requesting node can be
+     * done, before the query is sent to the cluster. Since this may increase the latency the pre-check is limited to a maximum
+     * number of local partitions.
+     * <p/>
+     * By increasing this parameter you can prevent the execution of the query on the cluster by increasing the latency due to the
+     * prolonged local pre-check.
+     * <p/>
+     * The pre-check can be disabled by setting a value of <tt>-1</tt>.
+     *
+     * @see #PROP_QUERY_RESULT_SIZE_LIMIT
+     */
+    public static final String PROP_QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK
+            = "hazelcast.query.max.local.partition.limit.for.precheck";
+
     public final GroupProperty CLIENT_ENGINE_THREAD_COUNT;
 
     public final GroupProperty PARTITION_OPERATION_THREAD_COUNT;
@@ -450,9 +489,10 @@ public class GroupProperties {
     public final GroupProperty BACKPRESSURE_SYNCWINDOW;
     public final GroupProperty BACKPRESSURE_BACKOFF_TIMEOUT_MILLIS;
     public final GroupProperty BACKPRESSURE_MAX_CONCURRENT_INVOCATIONS_PER_PARTITION;
-    /**
-     * @param config
-     */
+
+    public final GroupProperty QUERY_RESULT_SIZE_LIMIT;
+    public final GroupProperty QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK;
+
     public GroupProperties(Config config) {
         HEALTH_MONITORING_LEVEL
                 = new GroupProperty(config, PROP_HEALTH_MONITORING_LEVEL, HealthMonitorLevel.SILENT.toString());
@@ -565,6 +605,10 @@ public class GroupProperties {
                 = new GroupProperty(config, PROP_BACKPRESSURE_MAX_CONCURRENT_INVOCATIONS_PER_PARTITION, "100");
         BACKPRESSURE_BACKOFF_TIMEOUT_MILLIS
                 = new GroupProperty(config, PROP_BACKPRESSURE_BACKOFF_TIMEOUT_MILLIS, "60000");
+
+        QUERY_RESULT_SIZE_LIMIT = new GroupProperty(config, PROP_QUERY_RESULT_SIZE_LIMIT, "-1");
+        QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK
+                = new GroupProperty(config, PROP_QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK, "3");
     }
 
     public static class GroupProperty {
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/BasicMapContextQuerySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/BasicMapContextQuerySupport.java
@@ -17,6 +17,8 @@
 package com.hazelcast.map.impl;
 
 import com.hazelcast.instance.MemberImpl;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.map.QueryResultSizeExceededException;
 import com.hazelcast.map.impl.operation.QueryOperation;
 import com.hazelcast.map.impl.operation.QueryPartitionOperation;
 import com.hazelcast.map.impl.record.Record;
@@ -25,6 +27,7 @@ import com.hazelcast.nio.serialization.SerializationService;
 import com.hazelcast.query.PagingPredicate;
 import com.hazelcast.query.PagingPredicateAccessor;
 import com.hazelcast.query.Predicate;
+import com.hazelcast.query.TruePredicate;
 import com.hazelcast.query.impl.QueryEntry;
 import com.hazelcast.query.impl.QueryResultEntry;
 import com.hazelcast.query.impl.QueryableEntry;
@@ -58,13 +61,16 @@ class BasicMapContextQuerySupport implements MapContextQuerySupport {
 
     private final MapServiceContext mapServiceContext;
     private final NodeEngine nodeEngine;
+    private final MapQueryResultSizeLimitHelper mapQueryResultSizeLimitHelper;
 
     public BasicMapContextQuerySupport(MapServiceContext mapServiceContext) {
         this.mapServiceContext = mapServiceContext;
         this.nodeEngine = mapServiceContext.getNodeEngine();
+        this.mapQueryResultSizeLimitHelper = new MapQueryResultSizeLimitHelper(mapServiceContext, getLogger());
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public Collection<QueryableEntry> queryOnPartition(String mapName, Predicate predicate, int partitionId) {
         SerializationService serializationService = nodeEngine.getSerializationService();
         PagingPredicate pagingPredicate = predicate instanceof PagingPredicate ? (PagingPredicate) predicate : null;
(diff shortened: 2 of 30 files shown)


Donnerbart:
@pveentjer \*poke\*

Donnerbart:
verify

Donnerbart:
@pveentjer \*poke\*

pveentjer:
Finally time.. *making some tea and then getting started*

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -214,6 +214,17 @@
      */
     public static final String PROP_JCACHE_PROVIDER_TYPE = "hazelcast.jcache.provider.type";
 
+    /**
+     * Result size limit of a predicate query operation.
```
pveentjer: Can you provide a bit more information? E.g is this in total? Per member? Per partition? And why do we want this? We don't we get all results in one go? This is going to be asked by people. 

Also explain how it can be disabled so that HZ is working in old (but potentially dangerous) mode.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -214,6 +214,17 @@
...
+     */
+    public static final String PROP_QUERY_RESULT_SIZE_LIMIT = "hazelcast.query.result.size.limit";
+
+    /**
+     * Maximum value of local partitions to trigger local pre-check for <tt>TruePredicate</tt> query operations on maps.
```
pveentjer: Can you provide a bit more information for the mortals like me?

hazelcast/src/test/java/com/hazelcast/map/MapUnboundReturnValues_StringKeyTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.hazelcast.test.HazelcastParallelClassRunner;
+import com.hazelcast.test.annotation.SlowTest;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category(SlowTest.class)
+public class MapUnboundReturnValues_StringKeyTest extends MapUnboundReturnValuesTestSupport {
```
pveentjer: Can't we create a few fast tests? It is good to have nightly tests, but I would like to get some coverage during the day :)

hazelcast/src/main/java/com/hazelcast/map/impl/DataLimitExceededException.java
```diff
@@ -0,0 +1,7 @@
+package com.hazelcast.map.impl;
+
+import com.hazelcast.core.HazelcastException;
+
+public class DataLimitExceededException
```
pveentjer: Isn't it query result size exceeded? Data limit exceeded makes my think of my mobile internet provider.

hazelcast/src/test/java/com/hazelcast/map/MapUnboundReturnValues_StringKeyTest.java
Donnerbart: Hmm, that could be hard, since the feature depends heavily on the distribution of data in the cluster. This is why there is a hard-coded minimum of 100k before the feature triggers at all.

What makes the tests so slow is that I check that the feature doesn't trigger below the configured limit.

If we drop that check we could fill a single node with 120k small elements and see if the exception is thrown:
* filling cluster up to a lower bound which should be safe to not trigger
* call `IMap.values()` to check that lower bound didn't trigger
* repeat:
  * fill a bit more data in
  * call `IMap.values()`

If we drop this check we could just fill a single node with 120k items and run a single `Imap.values()` on it to trigger the exception. This should be much faster (single node, just one query).

How long may that test run? I'll give it a try locally.

Donnerbart: BTW: `MapUnboundReturnValues_PreCheckTest.testMapKeySet_SmallLimit()` takes 16 to 17 seconds on my local machine. I think that is a `SlowTest`.

Donnerbart: With the speedup (and less checks) I could provide a test with ~10 seconds runtime.

hazelcast/src/main/java/com/hazelcast/map/impl/BasicMapContextQuerySupport.java
```diff
@@ -42,13 +44,17 @@
 
     private final MapServiceContext mapServiceContext;
     private final NodeEngine nodeEngine;
+    private final long maxLocalPartitionsLimitForPreCheck;
 
     public BasicMapContextQuerySupport(MapServiceContext mapServiceContext) {
         this.mapServiceContext = mapServiceContext;
         this.nodeEngine = mapServiceContext.getNodeEngine();
+        this.maxLocalPartitionsLimitForPreCheck = Math.abs(
+                nodeEngine.getGroupProperties().QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK.getInteger());
```
ahmetmircik: Do we need to handle special case for `Integer.MIN_VALUE` when using `Math.abs` ?

Donnerbart: That's a good catch! Do we have a common place where I can prepare a `GroupProperty` once, so I don't have to do those checks on every query?

jerrinot:
@Donnerbart: Have you tried to measure impact of on performance?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
```diff
@@ -115,53 +123,56 @@ public void run() throws Exception {
         checkPartitionStateChanges(partitionService, partitionStateVersion);
     }
 
-    private void checkPartitionStateChanges(InternalPartitionService partitionService, int partitionStateVersion) {
-        if (partitionStateVersion != partitionService.getPartitionStateVersion()) {
-            getLogger().info("Partition assignments changed while executing query: " + predicate);
+    public static int getMaxResultLimit(NodeEngine nodeEngine) {
```
jerrinot: it would be nice to have some JavaDoc here. moreover - should this be really a responsibility of QueryOperation?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
```diff
@@ -115,53 +123,56 @@ public void run() throws Exception {
...
+        int maxResultLimit = Math.abs(nodeEngine.getGroupProperties().QUERY_RESULT_SIZE_LIMIT.getInteger());
+        if (maxResultLimit > 0 && maxResultLimit < MINIMUM_MAX_RESULT_LIMIT) {
+            maxResultLimit = MINIMUM_MAX_RESULT_LIMIT;
+            nodeEngine.getLogger(QueryOperation.class)
+                      .warning("Max result limit was set to minimal value of " + MINIMUM_MAX_RESULT_LIMIT);
```
jerrinot: this will be written to a log for each query?

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResultSizeExceededException.java
```diff
@@ -0,0 +1,7 @@
+package com.hazelcast.map.impl;
+
+import com.hazelcast.core.HazelcastException;
+
+public class QueryResultSizeExceededException
```
jerrinot: a public Exception should definitely have a good JavaDoc

hazelcast/src/main/java/com/hazelcast/map/impl/BasicMapContextQuerySupport.java
ahmetmircik: afaik no we don't

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResultSizeExceededException.java
ahmetmircik: also it should not be under `impl` package i think

Donnerbart:
Thank you all for your valuable reviews!

* I encapsulated most of the functionality in the new class `MapQueryResultSizeLimitHelper`, which is used by `BasicMapContextQuerySupport` and `QueryOperation` now
* the `GroupProperties` are just evaluated once when the map context is created
* the `QueryResultSizeExceededException` is now in the package `com.hazelcast.map`, I hope that is a better place
* there is more JavaDoc, especially in the `GroupProperties`

Donnerbart:
@jerrinot No performance test yet, but I hope the impact is minimal in the way I implemented the feature.  The normal result size checks should be very cheap.

The highest risk will be increased latency due to the local pre-check, which is a really new operation within the performance path. But it very easy to disable or maybe should even be disabled by default.

Donnerbart:
verify

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -221,6 +227,37 @@
      */
     public static final String PROP_JCACHE_PROVIDER_TYPE = "hazelcast.jcache.provider.type";
 
+    /**
+     * Result size limit for query operations on maps.
+     * <p/>
+     * This value defines the maximum number of returned elements for a single query result. If a query exceeds this number of
+     * elements a {@link QueryResultSizeExceededException} will be thrown.
+     * <p/>
+     * This feature is in place to prevent an OOME if a single node is requesting the whole data set of the cluster, e.g. by
```
jerrinot: it applies if and only if the TruePredicate is used, doesn't it?

Donnerbart: Just the local pre-check should be limited to the usage of `TruePredicate` (directly or indirectly). The query result limit (checked in `QueryOperation`) should work for all queries.

Donnerbart: The local pre-check doesn't work if the query won't return all values (it's just summing up the partition sizes). So we need that check for `TruePredicate` there. The check in `QueryOperation` checks the real query results, so it will also work for all types of queries.

Donnerbart:
I just made run in the test lab with a map below the result size limit (so all `IMap.values()` were executed without throwing any exception). I compared the actual 3.5 master with the feature branch. I can't see a big difference in the latency:

Imgur

Donnerbart:
I could just get rid of a bit of runtime calculations in `MapQueryResultSizeLimitHelper.getNodeResultLimit()` since most of the values are runtime constant. I also removed the partition count as parameter.

Donnerbart:
Added new test class for the encapsulated helper class. Code coverage is now 100%. Found an of by one error in the group properties parsing with this.

Donnerbart:
verify

Donnerbart:
verify

jerrinot:
I like the current code, I think it can be merged. However I don't understand the latency histogram. Values on the Y-axis looks off. There should be a number of operations for a given latency (as on the X-axis). Hence it always has to be > 1.

Donnerbart:
verify

Donnerbart:
You are right, that looks suspicious. I will have a look with an older Simulator version, before I touched anything in the probes and visualizer. Maybe I messed up something in the time measurement.

Beside that there is still the objection of Peter with the `0` vs. `Integer.MAX_VALUE` to disable the feature (see TDD comments in the wiki). Any opinion about this? It would be quite easy to change, I just don't like the idea to have a special value on two sides of the limit range (<= 0 would be invalid, Integer.MAX_VALUE would be disabled).

Donnerbart:
Updated the feature, so `-1` disabled both configurable values. There are unit tests for all edge cases. The documentation in `GroupProperties` is updated.

Donnerbart:
Rebased, local checkstyle/findbugs was happy, all new tests are still green :)

pveentjer:
I see that we have a different latency distribution graph. I hope the original is still available.

Donnerbart:
What do you mean with this? Do you want to see the probe XML files?

pveentjer:
Normally we get something like this:

http://www.fhwa.dot.gov/publications/research/safety/09049/images/image031.jpg

Donnerbart:
I'm sure this is some kind of bug, introduced by the probes refactoring or the visualizer refactoring. Not a stone was left standing there. To check this is the last Simulator framework thing on my agenda.

Donnerbart:
For the interpretation: I think the scale may be broken, but the relation between both measurings should still be okay. So there was not big shift in latency with my feature. Nevertheless it's hard to read in the current state.

jerrinot:
:+1: 
Good work!

hazelcast/src/main/java/com/hazelcast/map/impl/BasicMapContextQuerySupport.java
```diff
@@ -288,16 +316,19 @@ private Future queryOnLocalMember(String mapName, Predicate predicate, NodeEngin
...
             throws ExecutionException, InterruptedException {
-        for (Future future : futures) {
-            QueryResult queryResult = getQueryResult(future);
+        for (Future<QueryResult> future : futures) {
+            QueryResult queryResult = future.get();
             if (queryResult == null) {
                 continue;
             }
+            if (queryResult.isResultLimitExceeded()) {
+                throw new QueryResultSizeExceededException();
```
jerrinot: you could add an explanation what's going on.

hazelcast/src/main/java/com/hazelcast/map/impl/MapQueryResultSizeLimitHelper.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public long getNodeResultLimit(int ownedPartitions) {
+        return (long) Math.ceil(resultLimitPerPartition * ownedPartitions);
+    }
+
+    boolean isPreCheckEnabled() {
+        return isPreCheckEnabled;
+    }
+
+    void checkMaxResultLimitOnLocalPartitions(String mapName) {
```
jerrinot: could you split this method?

Donnerbart: I don't know a good design pattern to split up those "if (check) return; if (check) return;" methods. I extracted one method, but I don't see any obvious possibilities to refactor something else.

Donnerbart:
I added some JavaDoc to the `IMap` interface to make the users aware of the new feature and the consequences.

Donnerbart:
verify

bilalyasar:
verify

bilalyasar:
verify

bilalyasar:
verify

Donnerbart:
Latency distribution with the actual feature branch vs. actual master:
Imgur

hazelcast/src/main/java/com/hazelcast/map/impl/MapQueryResultSizeLimitHelper.java
jerrinot: I guess it's fine now. it's just a smell when a see a long(-ish) method structured like
```
myMethod() {
//do A
stuff();
stuff();
stuff();

//do B
stuff();
stuff();

//do C
stuff();
stuff();
}
```

but it's alright now.

jerrinot:
I see you added some warnings/comments when the `QueryResultSizeExceededException` is thrown. 

What I meant was more like a message in the exception itself - informing a user what's the current settings and possibly even how to change it.

Donnerbart:
Here are both latency distribution graphs (actual feature branch vs. actual master).

Map filled closely below trigger limit (100k * 1.15 * 0.5 = 103500 elements):
Imgur

Almost empty map (I think there were 4 elements in the map, due to the warmup algorithm):
Imgur

Donnerbart:
verify

Donnerbart:
I updated the JavaDoc in `IMap` to make clear that the exception is just thrown on the server side.

I also added a test for the client side to check that the exception is not thrown.

Donnerbart:
Good that I wrote that client test. Of course the feature kicked in when the client uses predicates, but no exception was thrown. So the result was silently cut. The same was happening for transactional maps.

I wrote tests for the client which check that the normal `IMap.values()`, `IMap.keySet()` and `IMap.entrySet()` throw no exception and return the full result size. I also wrote tests to check that the predicate methods throw the exception.

I still have to write tests for the transactional map. Also the normal, unchanged `IMap.entrySet()` on client side seems to run forever. Maybe there is a big with large data sets in that method.

Donnerbart:
My tests are slow, so I put them in the nightly category. So they are not executed by the PR builder.

There are still failures though, so please do not merge yet.

Donnerbart:
I will split up this PR, since it is already to big and more changes are coming (at least more test code).

Thank you all for the reviews and the comments, nothing is lost just moved to separate commits :)

PR #4848, opened 2015-03-20
Pulled out the IsStillRunning behavior

pveentjer:
nto a seperate service to simplify the invocationfuture

I'm currently working on more behavior in the InvocationFuture for back pressure and this class became too complex (for me and for checkstyle). So I pulled out the IsStillRunning behavior into a seperate structure so the invocationfuture becomes easier to maintain/understand.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
@@ -1,0 +1,146 @@
+package com.hazelcast.spi.impl.operationservice.impl;
+
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.nio.Address;
+import com.hazelcast.spi.ExecutionTracingService;
+import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.TraceableOperation;
+import com.hazelcast.spi.impl.NodeEngineImpl;
+import com.hazelcast.spi.impl.operationexecutor.OperationExecutor;
+import com.hazelcast.spi.impl.operationexecutor.OperationRunner;
+import com.hazelcast.spi.impl.operationservice.impl.operations.IsStillExecutingOperation;
+import com.hazelcast.spi.impl.operationservice.impl.operations.TraceableIsStillExecutingOperation;
+
+import java.util.concurrent.Future;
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
```diff
@@ -0,0 +1,146 @@
...
+            invocation.logger.warning("Asking if operation execution has been started: " + toString());
+            executing = (Boolean) invocation.nodeEngine.toObject(f.get(IS_EXECUTING_CALL_TIMEOUT, TimeUnit.MILLISECONDS));
+        } catch (Exception e) {
+            invocation.logger.warning("While asking 'is-executing': " + toString(), e);
+        }
+        invocation.logger.warning("'is-executing': " + executing + " -> " + toString());
+        return executing;
+    }
+
+    private Operation createCheckOperation(Invocation invocation) {
```
jerrinot: I prefer private methods after public ones. Don't we have a checkstyle rule enforcing this?

pveentjer: Imho it isn't a good style to order methods based on there accessibility. Normally you get a public method (your API) and to implement that public method you get a bunch of private methods that directly follow that public method. This keeps things closely together and makes it easier to understand code.

jerrinot:
now the question is whether the IsStillRunningService should be a public service or whether the OperationService should just delegate into it.

pveentjer:
It isn't a public service because it is in the .impl package. So it is a pure implementation detail not relevant to the rest of the system.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
jerrinot: ok, I don't have too strong opinion here.

jerrinot:
verify

jerrinot:
I haven't checked the actual logic as I assume it's just the original code extracted. 
I like the separation, hence :+1:  (as soon as the PR builder is green)

pveentjer:
Since this only moves existing behaviour, but doesn't change anything, I'm going to merge this one.

PR #4850, opened 2015-03-20
CyclicWriteBehindQueue minor improvements

ahmetmircik:
- Added mutable integer
- Fixed wrong method comment.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/CyclicWriteBehindQueue.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/CyclicWriteBehindQueue.java
@@ -17,6 +17,7 @@
 package com.hazelcast.map.impl.mapstore.writebehind;
 
 import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.util.MutableInteger;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -50,11 +51,11 @@ class CyclicWriteBehindQueue implements WriteBehindQueue<DelayedEntry> {
      *
      * @see WriteBehindStore#loadable(com.hazelcast.nio.serialization.Data)
      */
-    private final Map<Data, Integer> index;
+    private final Map<Data, MutableInteger> index;
 
     public CyclicWriteBehindQueue() {
         this.deque = new ArrayDeque<DelayedEntry>();
-        this.index = new HashMap<Data, Integer>();
+        this.index = new HashMap<Data, MutableInteger>();
     }
 
     /**
@@ -82,8 +83,8 @@ class CyclicWriteBehindQueue implements WriteBehindQueue<DelayedEntry> {
     }
 
     /**
-     * Removes the first occurrence of the specified element in this queue
-     * when searching it by starting from the head of this queue.
+     * Removes the first element of this queue instead of searching for it,
+     * implementation of this method is strongly tied with {@link StoreWorker} implementation.
      *
      * @param entry element to be removed.
      * @return <code>true</code> if removed successfully, <code>false</code> otherwise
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/util/MutableInteger.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.util;
+
+/**
```
pveentjer: Can you also document its thread safety (or lack off). It can be perfectly fine that this class isn't save, but make sure everybody knows it.

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/CyclicWriteBehindQueue.java
```diff
@@ -50,11 +51,11 @@
      *
      * @see WriteBehindStore#loadable(com.hazelcast.nio.serialization.Data)
      */
-    private final Map<Data, Integer> index;
+    private final Map<Data, MutableInteger> index;
 
     public CyclicWriteBehindQueue() {
         this.deque = new ArrayDeque<DelayedEntry>();
-        this.index = new HashMap<Data, Integer>();
+        this.index = new HashMap<Data, MutableInteger>();
```
pveentjer: Why don't you call MultableInteger counter? Makes the code a lot easier to read because the class name now explains its purpose.

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/CyclicWriteBehindQueue.java
```diff
@@ -180,13 +181,13 @@ public void getFrontByNumber(int numberOfElements, Collection<DelayedEntry> coll
...
+        Map<Data, MutableInteger> index = this.index;
 
-        Integer count = index.get(key);
+        MutableInteger count = index.get(key);
         if (count == null) {
-            count = 0;
+            count = new MutableInteger();
         }
-        count++;
+        count.value++;
```
pveentjer: I would add an inc method on count so you get:

counter.inc()

hazelcast/src/main/java/com/hazelcast/util/MutableInteger.java
pveentjer: I would place this class in your map.impl package. No need to share it with the rest of the world.

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/CyclicWriteBehindQueue.java
jerrinot: what about encapsulating the Map + MutableInteger in something like
```java
public class Counter<T> {
  int incAndGet(T data) {
    [...]
  }  

  int decAntGet(T data) {
    [...]
  }

  [ ...other methods... ]
}
```

this would allow you to remove implementation noise such as
```java
        MutableInteger count = index.get(key);
        if (count == null) {
            count = new MutableInteger();
        }
```
from the CyclicWriteBehindQueue class

hazelcast/src/main/java/com/hazelcast/util/MutableInteger.java
ahmetmircik: Added thread-safety comment. I put the class in util because will use it in other modules.

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/CyclicWriteBehindQueue.java
ahmetmircik: will use it in other modules for different purposes.

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/CyclicWriteBehindQueue.java
ahmetmircik: @pveentjer thought it but this seems very easy to use.
@jerrinot thanks. it will be better to change it in another PR. Only wanted to add `MutableInteger` in this PR.

bwzhang2011:
why not use AtomicInteger ?

Donnerbart:
An `AtomicInteger` introduces synchronization costs. If we don't need thread-safety we shouldn't use concurrent classes to avoid this overhead.

bwzhang2011:
@Donnerbart , you're right. as I just deem it to be global variable so as usual it's better to be set as some Atomic way for processing. so if we don't need thread-safety, just the original prototype is enough.

Issue #4846, opened 2015-03-20
[TEST-FAILURE] XmlConfigImportVariableReplacementTest.testThreeResourceCyclicImportThrowsException


bilalyasar:
```
java.lang.Exception: Unexpected exception, expected<com.hazelcast.config.InvalidConfigurationException> but was<com.hazelcast.core.HazelcastException>
	at com.sun.org.apache.xerces.internal.parsers.DOMParser.parse(DOMParser.java:257)
	at com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:347)
	at javax.xml.parsers.DocumentBuilder.parse(DocumentBuilder.java:121)
	at com.hazelcast.config.XmlConfigBuilder.parse(XmlConfigBuilder.java:210)
	at com.hazelcast.config.AbstractConfigBuilder.loadAndReplaceImportElement(AbstractConfigBuilder.java:95)
	at com.hazelcast.config.AbstractConfigBuilder.replaceImportElementsWithActualFileContents(AbstractConfigBuilder.java:80)
	at com.hazelcast.config.AbstractConfigBuilder.process(AbstractConfigBuilder.java:65)
	at com.hazelcast.config.XmlConfigBuilder.parseAndBuildConfig(XmlConfigBuilder.java:200)
	at com.hazelcast.config.XmlConfigBuilder.build(XmlConfigBuilder.java:184)
	at com.hazelcast.config.XmlConfigBuilder.build(XmlConfigBuilder.java:177)
	at com.hazelcast.config.XmlConfigImportVariableReplacementTest.buildConfig(XmlConfigImportVariableReplacementTest.java:489)
	at com.hazelcast.config.XmlConfigImportVariableReplacementTest.testThreeResourceCyclicImportThrowsException(XmlConfigImportVariableReplacementTest.java:170)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/2884/com.hazelcast$hazelcast/testReport/com.hazelcast.config/XmlConfigImportVariableReplacementTest/testThreeResourceCyclicImportThrowsException/

pveentjer:
Should InvalidConfigurationException not be a subclass of HazelcastException?

bilalyasar:
yes, it doesn't have to subclass of HazelcastException.

pveentjer:
Most exception in HZ extend HazelcastException.

mesutcelik:
[...]

PR #4851, opened 2015-03-20
fix for #4846

bilalyasar:


hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -129,7 +129,7 @@ protected Document parse(InputStream inputStream) throws Exception {
                     + lineSeparator + "Exception: " + e.getMessage()
                     + lineSeparator + "HazelcastClient startup interrupted.";
             LOGGER.severe(msg);
-            throw e;
+            throw new InvalidConfigurationException(e.getMessage());
```
pveentjer: This could be problematic because you are throwing the original cause. This makes configuring out the exact cause a problem. In most cases you want to do 'throw new InvalidConfigurationException(e.getMessage(), e);

hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
```diff
@@ -228,7 +228,7 @@ protected Document parse(InputStream is) throws Exception {
                 LOGGER.severe(msg);
 
             }
-            throw e;
+            throw new InvalidConfigurationException(e.getMessage());
```
pveentjer: Also the cause is lost here.

bilalyasar:
@pveentjer, thanks for the feedback, updated.

pveentjer:
I think your build will fails because of some checkstyle issues (spaces).

bilalyasar:
good catch!

pveentjer:
Good work.

PR #4854, opened 2015-03-21
Pulled out InvocationRegistry

pveentjer:
Reducing complexity on invocation due to back pressure. This is one of the side effects I want to get merged before the PR gets too big.

So I pulled out the InvocationRegistry; takes care of registration/deregistration/notification of of invocations. This used to be a responsibility of the OperationServiceImpl. Also the ResponsePacketHandler has been simplified since the registry take care of most of the logic.

Another thing that is introduced is the ErrorResponse. In case of an exception instead of using the NormalResponse and then dealing with the content, the ErrorResponse is now uses so that it is immediately clear that error logic needs to be executed (e.g. retry). This simplifies the logic in registry and invocation.

The idea for the simplified back pressure is that the InvocationRegistry is going to get a capacity of maximum number of remote invocations, e.g. 10k (configurable). This should be more than enough for most users since they won't have more than 1 invocation per thread, so a few hundred at most. As soon as the InvocationRegistry runs out of free invocations, then some exponential backoff can be done. I need this better approach to deal better with the new async api's Ali added.


pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -0,0 +1,365 @@
...
+        } else if (response instanceof CallTimeoutResponse) {
+            notifyCallTimeout((CallTimeoutResponse) response);
+        } else if (response instanceof ErrorResponse) {
+            notifyErrorResponse((ErrorResponse) response);
+        } else {
+            logger.severe("Unrecognized response: " + response);
+        }
+    }
+
+    public void notifyBackupComplete(long callId) {
```
jerrinot: this is an inconsistency in a method signature. all other notifyXY() methods accept a *Response as an argument.

gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
pveentjer: I know. But this method is also being called from the outside (backup operation if I remember correctly). In the future I want to get rid of the responses completely.

jerrinot:
:+1:

jerrinot:
could you squash the 2 commits into one? so we will have nice & clean history.

pveentjer:
Thanks

bilalyasar:
I guess, findbugs will make a complaint :)

PR #4859, opened 2015-03-22
Improved back pressure

pveentjer:


hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -216,6 +210,23 @@
     public static final String PROP_BACKPRESSURE_SYNCWINDOW = "hazelcast.backpressure.syncwindow";
 
     /**
+     * The maximum number of concurrent invocations per partition.
+     *
+     * To prevent the system overloading, HZ can apply a constrain on the number of concurrent invocations. If the maximum
+     * number of concurrent invocations has exceeded and a new invocation comes in, then an exponential back-off is applied
+     * till eventually a timeout happens or there is room for the invocation.
+     *
+     * By default is is configured as 100, so with 271 partitions that would give (271+1)*100=27200 concurrent invocations from a
```
Donnerbart: "By default is is" vs. "it is"

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
```diff
@@ -0,0 +1,200 @@
...
+import com.hazelcast.spi.Operation;
+
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static java.lang.Math.round;
+
+
+/**
+ * The InvocationBackPressureRegulator is responsible for regulating invocation 'pressure' and if sees that the system
```
Donnerbart: Javadoc needs some fixes. Two newlines before Javadoc.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
```diff
@@ -0,0 +1,200 @@
...
+    public CallIdSequence newCallIdSequence() {
+        if (enabled) {
+            return new CallIdSequence.CallIdSequenceWithBackpressure(maxConcurrentInvocations);
+        } else {
+            return new CallIdSequence.CallIdSequenceWithoutBackpressure();
+        }
+    }
+
+    /**
+     * Checks if a sync is forced for
```
Donnerbart: Unfinished sentence, incomplete JavaDoc (parameter, return).

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -0,0 +1,246 @@
...
+        private static final int INDEX_TAIL = 15;
+
+        // instead of using 2 AtomicLongs, we use an array if width of 3 cache lines to prevent any false sharing.
+        private final AtomicLongArray longs = new AtomicLongArray(24);
+
+        private final int maxConcurrentInvocations;
+
+        public CallIdSequenceWithBackpressure(int maxConcurrentInvocations) {
+            if (maxConcurrentInvocations < 1) {
+                throw new IllegalArgumentException("maxConcurrentInvocations can't be smaller than 0");
```
Donnerbart: Code checks for "value < 1", exception message says "value smaller than 0"

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -0,0 +1,246 @@
...
+import com.hazelcast.spi.BackupAwareOperation;
+
+import java.util.concurrent.atomic.AtomicLongArray;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+
+import static com.hazelcast.spi.Operation.CALL_ID_LOCAL_SKIPPED;
+import static com.hazelcast.spi.OperationAccessor.setCallTimeout;
+
+/**
+ * Responsible for generating call id's.
```
Donnerbart: No apostrophe for plural.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -0,0 +1,246 @@
...
+ * <p/>
+ * It is very important that for each {@link #next(Invocation)}, there is a matching {@link #complete(Invocation)}.
+ * If this doesn't match, than either the number of concurrent invocations can grow or shrink ever time. This can lead
```
Donnerbart: ever/over time

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -0,0 +1,246 @@
...
+ * The CallIdSequence provides back pressure if enabled and there are too many concurrent invocations. When this happens,
+ * an exponential backoff policy is applied.
+ * <p/>
+ * Currently a single CallIdSequence is used for all partitions; so there is contention. Also one partition can cause problems
+ * in other partition if a lot of invocations are created for that partition. Then other partitions can't make as many invocations
+ * because a single callIdSequence is being used.
+ * <p/>
+ * In the future we could add a CallIdSequence per partition or using some 'concurrency level' and do a mod based on the
+ * partition-id. The advantage is that you reduce contention and improved isolation, at the expensive of increased complexity
+ * + not always being able to fully utilize the number of invocations.
```
Donnerbart: I think this can be deleted.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -0,0 +1,246 @@
...
+     * - due to system operations
+     * - due to racy nature of checking if space is available and getting the next sequence.
+     *
+     * The last cause is not a problem since the capacity is exceeded temporary and it isn't sustainable. So perhaps
+     * there are a few threads that at the same time see that the there is space and do a next. But any following
+     * invocation needs to wait till there is is capacity.
+     */
+    public static final class CallIdSequenceWithBackpressure extends CallIdSequence {
+        private static final int MAX_DELAY_MS = 500;
+        private static final int INDEX_HEAD = 7;
```
Donnerbart: What's the reason for beginning with 7?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -126,39 +127,52 @@
...
+        if (!(op instanceof WaitSupport)) {
+            return defaultCallTimeout;
+        }
+
+        long waitTimeoutMillis = op.getWaitTimeout();
+        if (waitTimeoutMillis > 0 && waitTimeoutMillis < Long.MAX_VALUE) {
+            /*
+             * final long minTimeout = Math.min(defaultCallTimeout, MIN_TIMEOUT);
+             * long callTimeout = Math.min(waitTimeoutMillis, defaultCallTimeout);
+             * callTimeout = Math.max(a, minTimeout);
```
Donnerbart: I think `a` should be `callTimeout` in your example.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -40,99 +37,80 @@
...
 
     /**
      * Deregisters an invocation.
-     *
-     * If the invocation already already deregisterd or the registration was skipped, the call is ignored.
+     * <p/>
+     * If the invocation registration was skipped, the call is ignored.
      *
      * @param invocation the Invocation to deregister.
      */
```
Donnerbart: `@throws AssertionError if deregister of invocation failed` ?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -356,10 +299,38 @@ private void scanHandleOperationTimeout() {
                 try {
                     invocation.checkBackupTimeout(backupTimeoutMillis);
                 } catch (Throwable t) {
+                    t.printStackTrace();
```
Donnerbart: Superfluous `t.printStackTrace()`? Should be printed via `logger.severe()`.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
```diff
@@ -295,20 +320,20 @@ public void logOperationDeserializationException(Throwable t, long callId) {
         boolean returnsResponse = callId != 0;
 
         if (t instanceof RetryableException) {
-            final Level level = returnsResponse ? Level.FINEST : Level.WARNING;
-            if (operationService.logger.isLoggable(level)) {
-                operationService.logger.log(level, t.getClass().getName() + ": " + t.getMessage());
+            final Level level = returnsResponse ? FINEST : Level.WARNING;
```
Donnerbart: static import for `Level.WARNING`?

pveentjer: static imports.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
pveentjer: Good one.. I'll fix that.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
pveentjer: I only cleaned up the structure of the code. So the wait timeout was already there. It isn't changed by me.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
pveentjer: which part?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
pveentjer: I'm not that crazy about documenting asserts.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -40,99 +37,80 @@
...
+        callIdSequence.complete(invocation);
+
+        setCallId(invocation.op, 0);
+
+        if (callId == 0 || callId == CALL_ID_LOCAL_SKIPPED) {
             return;
         }
 
-        invocations.remove(callId);
+        assert invocations.remove(callId) != null : "failed to deregister callId:" + callId + " " + invocation;
```
mdogan: Are you sure `invocations.remove(callId)` will be called when assertions are disabled (which is default)? AFAIK it will not.

mdogan: try this:

```java
static boolean FOO_CALLED;

        assert foo();
        System.out.println(FOO_CALLED);

    }

    static boolean foo() {
        FOO_CALLED = true;
        return true;
    }
```

pveentjer: That is a good one! 

Initially I didn't have the assert.. but I ran into some test failures and added the assert around it.  So I'll fix that asap.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
```diff
@@ -0,0 +1,204 @@
...
+
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static java.lang.Math.round;
+
+
+/**
+ * The BackpressureRegulator is responsible for regulating invocation 'pressure' and if sees that the system
+ * is getting overloaded, it will apply back pressure, so the the system
```
mdogan: It looks like sentence 
> ... so the the system

is not complete here...

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
```diff
@@ -0,0 +1,204 @@
...
+ * <p/>
+ * Responsible for regulating invocation pressure on the Hazelcast system.cIn most cases the invocations on Hazelcast
```
mdogan: A typo
> on the Hazelcast system.cIn most cases

*on the Hazelcast system. In most cases*

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -0,0 +1,249 @@
...
+        if (invocation.op instanceof BackupAwareOperation) {
+            return false;
+        }
+
+        return true;
+    }
+
+    public static final class CallIdSequenceWithoutBackpressure extends CallIdSequence {
+
+        private static final AtomicLongFieldUpdater<CallIdSequenceWithoutBackpressure> HEAD
```
mdogan: What is the advantage of using `AtomicLongFieldUpdater` here? Why not regular `AtomicLong`?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImplTest.java
```diff
@@ -0,0 +1,300 @@
...
+        };
+        op.setResponseHandler(responseHandler);
+        op.setPartitionId(operationRunner.getPartitionId());
+
+        operationRunner.run(op);
+
+        verify(responseHandler).sendResponse(any(ExpectedRuntimeException.class));
+    }
+
+//    @Test
```
mdogan: Are these tests commented out intentionally or are they left-over?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl_invokeOnTargetTest.java
```diff
@@ -81,6 +81,7 @@ public void test_nonExistingTarget() throws UnknownHostException {
         }
     }
 
+    //todo
```
mdogan: What about adding `@Ignored("TODO: will implement later...")` to make it more visible? Currently this's an empty method.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
pveentjer: One less memory indirection.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl_invokeOnTargetTest.java
pveentjer: Good one.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImplTest.java
pveentjer: I'll check them.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
pveentjer: Strange.. I thought I left a comment here.

There is one less memory indirection. I don't want to make the previous situation slower than the current one. So instead of having an atomiclong with a volatile field, you have a CallSequenceIdWithoutBackPressure with volatile field.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -0,0 +1,249 @@
...
+                    if (remainingTimeoutMs <= 0) {
+                        remainingTimeoutMs = 0;
+                        throw new HazelcastOverloadException("Failed to get a callId for invocation: " + invocation);
+                    }
+
+                    delayMs = nextDelay(remainingTimeoutMs, delayMs);
+                    remainingTimeoutMs -= delayMs;
+                }
+            } finally {
+                setCallTimeout(invocation.op, remainingTimeoutMs);
```
mdogan: I'm not sure if setting remaining time as call-timeout. Because call-timeout is used as timeout of each remote call, it means for each retry not the whole invocation. 

For example normally, using 1min call-timeout, a caller can wait infinitely as long as a retry-response is received in every 1min. But when no response is received in call-timeout (+ some extra time), invocation will throw operation-timeout exception. 

If you set remaining-time as call-timeout, on every retry call-timeout will decrease until it reaches zero. I don't know if this is your intention...

pveentjer: It was my intention. But I know this issue with blocking calls because they interact in a complex way with the timeout. 

What is your suggestion? 

I can introduce a new timeout: wait for call invocation slot timeout 

It will be be deducted from your call timeout. It is just some global constant (through properties) that is the same for every invocation.

mdogan: A new timeout will make it more complex, I think. It's already complex, I've never managed to make it simpler since Hazelcast 1. 

What about storing original call-timeout in invocation and using it for slot timeout and restoring operation call-timeout on every retry?

pveentjer: This new timeout will not be registered on the operation or invocation btw. It is just some kind of timeout used somewhere in the system. The only problem is that a user has no control on this overload timeout apart from a groupsproperty. The time spend on doing waiting will also not be deducted anywhere.. 

That is what I meant with an additional property.

mdogan: ok, fine for me.

pveentjer: Thanks. I'll do  that. We can always come up with a different mechanism later.

pveentjer:
I have updated a PR. I have added a backoff timeut configurable through the GroupsProperties. No more messing around with the timeout of the Operation.

mdogan:
:+1:

pveentjer:
@gurbuzali @mdogan how to proceed? Someone is going to hit the merge button?

gurbuzali:
I'll hit the green button in a minute, reviewing myself already

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
```diff
@@ -0,0 +1,214 @@
...
+        Operation op = (Operation) backupAwareOp;
+
+        // we never apply back-pressure on urgent operations.
+        if (op.isUrgent()) {
+            return false;
+        }
+
+        AtomicInteger syncDelayCounter = syncDelayCounter(op);
+
+        for (; ; ) {
```
Donnerbart: If I'm not totally wrong this loop is not wait-free, isn't it? It does not return after a defined number of tries, theoretically it could loop for ever.

A lot of loops would just happen under extreme load. In this case I would not worry that there is no forced sync operation being created. So a solution for this might be to return with `syncForced = false` if you looped for more tries than the first `syncDelayCounter.get()` was. After so many tries another backup operation did the sync and we could happily retreat with async (and even without decreasing the counter).

Donnerbart: Or maybe better: check if the `prev` is higher than in the last iteration. Then we can be sure another operation was forced to be sync.

pveentjer: I'm not worried about this not being wait free because:
- a thread needs to do so much other stuff to complete a backup, that there is a very small chance that 2 threads at the same moment are going to content with each other
- there are 271 partitions by default and each partition will get its own sync delay counter. This means that there is a very small chance that 2 threads are going to content for the same counter.

Donnerbart: Fair enough. I'm just a bit distrustful whenever I see a `for (; ; )` or `while (true)` :)

pveentjer: That is a typical way to set up a non blocking algorithm. I'm not distrustful about it.

Issue #4758, opened 2015-03-06
Config XmlConfigImportVariableReplacementTest failures in windows environment


bilalyasar:
full story is here:
https://groups.google.com/forum/#!topic/hazelcast/5TCAqJs9BQM

i reproduced errors in windows env.

bilalyasar:
closed via #4867

PR #4867, opened 2015-03-23
fix for #4758

bilalyasar:
file urls should start with 3 slashes.
for more info: http://superuser.com/questions/352133/why-do-file-urls-start-with-3-slashes


jerrinot:
:+1:

PR #4870, opened 2015-03-23
Before join operation is added to NodeExtension interface.

kobalski:
This interface will be implemented by EnterpriseNodeExtension and will be
used for license check of HD Memory.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -648,7 +648,7 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
                     node.getNodeExtension().beforeJoin();
                 } catch (Exception e) {
                     logger.warning(e.getMessage());
-                    sendBeforeJoinCheckFailure(joinRequest.getAddress(),e.getMessage());
+                    sendBeforeJoinCheckFailure(joinRequest.getAddress(), e.getMessage());
                     return;
                 }
             }
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -32,6 +32,7 @@ import com.hazelcast.cluster.impl.operations.MemberInfoUpdateOperation;
 import com.hazelcast.cluster.impl.operations.MemberRemoveOperation;
 import com.hazelcast.cluster.impl.operations.PostJoinOperation;
 import com.hazelcast.cluster.impl.operations.SetMasterOperation;
+import com.hazelcast.cluster.impl.operations.BeforeJoinCheckFailureOperation;
 import com.hazelcast.core.Cluster;
 import com.hazelcast.core.HazelcastException;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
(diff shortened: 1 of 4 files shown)


serkan-ozal:
:+1:

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -186,4 +186,7 @@ public void destroy() {
         logger.info("Destroying node NodeExtension.");
     }
 
+    @Override
+    public void beforeJoin() { logger.info("Before join check."); }
```
pveentjer: We don't do such one liners. Please use 

```
public void beforeJoin() { 
    ogger.info("Before join check."); 
}
```

pveentjer:
I don't see any tests. Can you add them? 

Can you also add documentation. This cluster functionality is complicated and needs to have very good documentation so that we get something other people can use/maintain.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -642,6 +643,16 @@ public void handleJoinRequest(JoinRequest joinRequest, Connection connection) {
                 }
             }
 
+            if (node.isMaster()) {
+                try {
+                    node.getNodeExtension().beforeJoin();
+                } catch (Exception e) {
+                    logger.warning(e.getMessage());
+                    sendBeforeJoinCheckFailure(joinRequest.getAddress(),e.getMessage());
```
pveentjer: I think you have checkstyle issues here. I guess that the cluster.impl package is fully excluded from checkstyle and that is why these issues are not shown.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -186,4 +186,9 @@ public void destroy() {
         logger.info("Destroying node NodeExtension.");
     }
 
+    @Override
+    public void beforeJoin() {
+        logger.info("Before join check.");
```
mdogan: What's the use of this log? If this method is doing nothing then no need to log anything. I think you should remove it or at least make it `FINEST`.

jerrinot: :+1:

kobalski:
@pveentjer I tried to write tests for this PR, but Node,NodeEngine and ClusterServiceImpl classes are impossible to mock, especially implementation of NodeExtension is decided by NodeExtensionFactory which has a static create method. Also I could not find any tests for DefaultNodeExtension class too, for a starting point. If you have any suggestion, how to write tests for DefaultNodeExtension class, help is appreciated. Also this PR will be used for license checking for HD Cache  (https://github.com/hazelcast/hazelcast-enterprise/pull/162),so DefaultNodeExtension does not implement beforeJoin(). As for documentation, I added some java doc comments.

pveentjer:
the NodeEngineImpl and ClusterService/PartitionServiceImpl and Node are quite difficult to deal with. Mocking probably isn't an option at all. But can't you make a test that isn't based on mocking but a true integration tests to see if you code is being triggered at all when the right conditions are met?

kobalski:
Since beforeJoin() method is not implemented by DefaultNodeExtension on open source side, there is no way for verification whether it is called on right condition in an integration test without mocking. beforeJoin() is fully implemented in this PR (https://github.com/hazelcast/hazelcast-enterprise/pull/162/) by EnterpriseNodeExtension, so that I could trigger certain cases (eg: Cluster exceeds licensed HD cache amount) and test if it is called at right condition in an integration test. So basically, there are integration tests written in the related enterprise PR where beforeJoin() method is not empty.

serkan-ozal:
I think, there is a way for customizing ```NodeExtension``` creation by using mocked classloader for ```Config``` object. I think, our custom classloader can behave as there is a definition for  ```NodeExtension``` in ```META-INF/services/com.hazelcast.instance.NodeExtension``` file. So this fake definition points name of our custom ```NodeExtension``` class. 

We will talk this solution with @kobalski tomorrow. However I am not really sure is this mocking stuff really needed and worth to this effort.

kobalski:
verify

mesutcelik:
@serkan-ozal , I don't know if that is in this PR's scope. I think we can create an enhancement issue and merge this PR if all other comments are already considered.

serkan-ozal:
@mesutcelik right :+1: I am ok. if there is no objection, lets merge it

PR #4872, opened 2015-03-23
Map/reduce type variance improvements

Leperous:
You should be able to pass a more variantly-typed reducer, combiner and predicate to a mapping job. The job does not mutate any data so this is a well defined operation.

The same also holds for the KeyValueSource factory methods.


pveentjer:
@noctarius can you have a look since you created the map reduce api.

noctarius:
Since it just adds extends / super for producer and consumers it should be fine. I guess all test are still running tho :)

pveentjer:
@noctarius what are we going to do with this pr?

noctarius:
I guess we can merge it but I'm not module owner anymore, I suppose :)

bilalyasar:

pveentjer:
verify

mtopolnik:
This PR has become stale and was revived as #6693

PR #4873, opened 2015-03-23
[CP] Client protocol message flyweight and base utils

asimarslan:
This PR includes some util classes from Martin Thompson's project, Aeron. Copyright message kept as is.


sancar:
:+1:

PR #4875, opened 2015-03-23
Selector Balancer Implemented

jerrinot:
Motivation:
Hazelcast by default uses 3 threads to read data from TCP connections
and 3 threads to write data to the connections. We have measured significant
fluctuations of performance when the threads are not utilized equally.

This Balancer tries to detect such situations and fix them by moving
com.hazelcast.nio.tcp.ReadHandler and com.hazelcast.nio.tcp.WriteHandler
between threads.

Implementation:
It measures number of events serviced by each handler in a given interval
and if imbalance is detected then it schedules handler migration to improve
the situation. The exact migration strategy can be customized via
com.hazelcast.nio.handlermigration.MigrationStrategy.

Configuration:
Measuring interval can be customized via
com.hazelcast.instance.GroupProperties#HANDLER_MIGRATION_INTERVAL_SECONDS

Implementator's Note:
It doesn't leverage com.hazelcast.nio.ConnectionListener capability
provided by com.hazelcast.nio.ConnectionManager to observe connections as it
has to be notified right after a physical TCP connection is created whilst
ConnectionListener is notified only after a successful (Hazelcast) binding process.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -62,6 +62,18 @@ public class GroupProperties {
     public static final String PROP_VERSION_CHECK_ENABLED = "hazelcast.version.check.enabled";
     public static final String PROP_PREFER_IPv4_STACK = "hazelcast.prefer.ipv4.stack";
     public static final String PROP_IO_THREAD_COUNT = "hazelcast.io.thread.count";
+
+    /**
+     * The interval in seconds between {@link com.hazelcast.nio.tcp.handlermigration.IOBalancer IOBalancer}
+     * executions. The shorter intervals will catch I/O Imbalance faster, but they will cause higher overhead.
+     *
+     * Please see documentation of {@link com.hazelcast.nio.tcp.handlermigration.IOBalancer IOBalancer} for
+     * detailed explanation of the problem.
+     *
+     * Default value is 20 seconds. A negative value disables the balancer.
+     *
+     */
+    public static final String PROP_IO_BALANCER_INTERVAL_SECONDS = "hazelcast.io.balancer.interval.seconds";
     /**
      * The number of partition threads per Member. If this is less than the number of partitions on a Member, then
      * partition operations will queue behind other operations of different partitions. The default is 4.
@@ -239,6 +251,8 @@ public class GroupProperties {
 
     public final GroupProperty IO_THREAD_COUNT;
 
+    public final GroupProperty IO_BALANCER_INTERVAL_SECONDS;
+
     public final GroupProperty EVENT_QUEUE_CAPACITY;
 
     public final GroupProperty EVENT_QUEUE_TIMEOUT_MILLIS;
@@ -406,6 +420,7 @@ public class GroupProperties {
         VERSION_CHECK_ENABLED = new GroupProperty(config, PROP_VERSION_CHECK_ENABLED, "true");
         PREFER_IPv4_STACK = new GroupProperty(config, PROP_PREFER_IPv4_STACK, "true");
         IO_THREAD_COUNT = new GroupProperty(config, PROP_IO_THREAD_COUNT, "3");
+        IO_BALANCER_INTERVAL_SECONDS = new GroupProperty(config, PROP_IO_BALANCER_INTERVAL_SECONDS, "20");
 
         //-1 means that the value is worked out dynamically.
         PARTITION_OPERATION_THREAD_COUNT = new GroupProperty(config, PROP_PARTITION_OPERATION_THREAD_COUNT, "-1");
--- a/hazelcast/src/main/java/com/hazelcast/nio/IOService.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/IOService.java
@@ -95,6 +95,11 @@ public interface IOService {
 
     int getConnectionMonitorMaxFaults();
 
+    /**
+     * @return Time interval between two I/O imbalance checks.
+     */
+    int getBalancerIntervalSeconds();
+
     void onDisconnect(Address endpoint);
 
     boolean isClient();
--- a/hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
@@ -239,6 +239,11 @@ public class NodeIOService implements IOService {
     }
 
     @Override
+    public int getBalancerIntervalSeconds() {
+        return node.groupProperties.IO_BALANCER_INTERVAL_SECONDS.getInteger();
+    }
+
+    @Override
     public void executeAsync(final Runnable runnable) {
         nodeEngine.getExecutionService().execute(ExecutionService.IO_EXECUTOR, runnable);
     }
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
@@ -25,16 +25,17 @@ import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.util.logging.Level;
 
-abstract class AbstractSelectionHandler implements SelectionHandler {
+public abstract class AbstractSelectionHandler implements MigratableHandler {
 
     protected final ILogger logger;
     protected final SocketChannelWrapper socketChannel;
     protected final TcpIpConnection connection;
     protected final TcpIpConnectionManager connectionManager;
-    protected final Selector selector;
-    protected final IOSelector ioSelector;
-    private final int initialOps;
+    protected Selector selector;
+    protected IOSelector ioSelector;
+
     private SelectionKey selectionKey;
+    private final int initialOps;
 
     public AbstractSelectionHandler(TcpIpConnection connection, IOSelector ioSelector, int initialOps) {
         this.connection = connection;
@@ -46,6 +47,25 @@ abstract class AbstractSelectionHandler implements SelectionHandler {
         this.initialOps = initialOps;
     }
 
+    /**
+     * Migrates this handler to a new IOSelector thread.
+     * The migration logic is rather simple:
+     * <p><ul>
+     *     <li>Submit a de-registration task to a current IOSelector thread</li>
+     *     <li>The de-registration task submits a registration task to the new IOSelector thread</li>
+     * </ul></p>
+     *
+     * @param newOwner target IOSelector this handler migrates to
+     */
+    public void migrate(final IOSelector newOwner) {
+        if (ioSelector == newOwner || !socketChannel.isOpen()) {
+            return;
+        }
+
+        ioSelector.addTask(new MigrationTask(newOwner));
+        ioSelector.wakeup();
+    }
+
     protected SelectionKey getSelectionKey() {
         if (selectionKey == null) {
             try {
@@ -54,7 +74,6 @@ abstract class AbstractSelectionHandler implements SelectionHandler {
                 handleSocketException(e);
             }
         }
-
         return selectionKey;
     }
 
@@ -93,4 +112,51 @@ abstract class AbstractSelectionHandler implements SelectionHandler {
         }
     }
 
+    final void unregisterOp(int operation) {
+        SelectionKey selectionKey = getSelectionKey();
+        try {
+            selectionKey.interestOps(selectionKey.interestOps() & ~operation);
+        } catch (Throwable e) {
+            handleSocketException(e);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     */
+    public IOSelector getOwner() {
+        return ioSelector;
+    }
+
+    private class MigrationTask implements Runnable {
+        private final IOSelector newOwner;
+
+        public MigrationTask(IOSelector newOwner) {
+            this.newOwner = newOwner;
+        }
+
+        @Override
+        public void run() {
+            if (!socketChannel.isOpen()) {
+                return;
+            }
+            unregisterOp(initialOps);
+            ioSelector = newOwner;
+            selectionKey.cancel();
+            selectionKey = null;
+            newOwner.addTask(new Runnable() {
+                @Override
+                public void run() {
+                    if (!socketChannel.isOpen()) {
+                        return;
+                    }
+                    selector = newOwner.getSelector();
+                    selectionKey = getSelectionKey();
+                    registerOp(initialOps);
+                }
+            });
+            newOwner.wakeup();
+        }
+    }
 }
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/MigratableHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/MigratableHandler.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.nio.tcp;
+
+/**
+ * Read/Write handlers that supports migration between {@link com.hazelcast.nio.tcp.IOSelector}.
+ *
+ */
+public interface MigratableHandler extends SelectionHandler {
+
+    /**
+     * Migrate to a new IOSelector.
+     *
+     * @param newOwner
+     */
+    void migrate(final IOSelector newOwner);
+
+    /**
+     * Get IOSelector currently owning this handler. Handler owner is a thread running this handler.
+     * {@link com.hazelcast.nio.tcp.handlermigration.IOBalancer IOBalancer} can decide to migrate
+     * a handler to another owner.
+     *
+     * @return current owner
+     */
+    IOSelector getOwner();
+
+    /**
+     * Get number of events recorded by the current handler. It can be used to calculate whether
+     * this handler should be migrated to a different {@link com.hazelcast.nio.tcp.IOSelector}
+     *
+     * @return total number of events recorded by this handler
+     */
+    long getEventCount();
+}
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/ReadHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/ReadHandler.java
@@ -31,16 +31,16 @@ import static com.hazelcast.util.StringUtil.bytesToString;
 /**
  * The reading side of the {@link com.hazelcast.nio.Connection}.
  */
-final class ReadHandler
-        extends AbstractSelectionHandler {
+public final class ReadHandler extends AbstractSelectionHandler {
 
     private final ByteBuffer inputBuffer;
-
-    private final IOSelector ioSelector;
 
     private SocketReader socketReader;
 
     private volatile long lastHandle;
+
+    //This field will be incremented by a single thread. It can be read by multiple threads.
+    private volatile long eventCount;
 
     public ReadHandler(TcpIpConnection connection, IOSelector ioSelector) {
         super(connection, ioSelector, SelectionKey.OP_READ);
@@ -61,7 +61,15 @@ final class ReadHandler
     }
 
     @Override
+    public long getEventCount() {
+        return eventCount;
+    }
+
+    @Override
+    @edu.umd.cs.findbugs.annotations.SuppressWarnings(value = "VO_VOLATILE_INCREMENT",
+            justification = "eventCount is accessed by a single thread only.")
     public void handle() {
+        eventCount++;
         lastHandle = Clock.currentTimeMillis();
         if (!connection.isAlive()) {
             String message = "We are being asked to read, but connection is not live so we won't";
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
@@ -18,11 +18,14 @@ package com.hazelcast.nio.tcp;
 
 import com.hazelcast.cluster.impl.BindMessage;
 import com.hazelcast.config.SocketInterceptorConfig;
+import com.hazelcast.instance.HazelcastThreadGroup;
 import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.LoggingService;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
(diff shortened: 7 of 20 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
```diff
@@ -93,4 +112,47 @@ final void registerOp(int operation) {
...
+            this.newOwner = newOwner;
+        }
+
+        @Override
+        public void run() {
+            if (!socketChannel.isOpen()) {
+                return;
+            }
+            unregisterOp(initialOps);
+            ioSelector = newOwner;
```
ahmetmircik: Does `ioSelector` need to be declared as `volatile`?

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
```diff
@@ -78,9 +79,9 @@ public TcpIpConnectionMonitor createNew(Address endpoint) {
 
     private final boolean socketNoDelay;
 
-    private final ConcurrentMap<Address, Connection> connectionsMap = new ConcurrentHashMap<Address, Connection>(100);
+    private final ConcurrentHashMap<Address, Connection> connectionsMap = new ConcurrentHashMap<Address, Connection>(100);
```
ahmetmircik: Could you please state what is the rationale behind using implementation instead of interface here?

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
mdogan: Just did a quick scan, it seems`MigrationTask` is executed by IO thread itself. So, no need.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/BalancerState.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.hazelcast.nio.tcp.IOSelector;
+import com.hazelcast.nio.tcp.MigratableHandler;
+import com.hazelcast.nio.tcp.SelectionHandler;
+import com.hazelcast.util.ItemCounter;
+
+import java.util.Map;
+import java.util.Set;
+
+class BalancerState<T extends MigratableHandler & SelectionHandler> {
```
ahmetmircik: Maybe you can simply document the responsibility of this class.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOSelectorLoadCalculator.java
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * Calculates a state describing utilization of IOSelector(s) and creates a mapping between IOSelector -> Handler.
+ *
+ * This class is not thread-safe with the exception of
+ * {@link #addHandler(T)}   and
+ * {@link #removeHandler(T)}
+ *
+ */
+class IOSelectorLoadCalculator<T extends MigratableHandler & SelectionHandler> {
```
serkan-ozal: Instead of generic type `and (&)` operation, I would prefer new interface such as `MigratableSelectionHandler` interface that extends from both of `MigratableHandler` and `SelectionHandler` interfaces

hazelcast/src/main/java/com/hazelcast/util/ItemCounter.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Non Thread-Safe Counter of things. It allows to count items without worrying about nulls.
+ *
+ * @param <T>
+ */
+public final class ItemCounter<T> {
+    private final Map<T, Long> map = new HashMap<T, Long>();
```
ahmetmircik: Maybe you can also use something like LongHolder instead of creating a new Long object in every `add` call.
`class LongHolder {
public long value
}`

ahmetmircik: Can you also add some unit tests for `ItemCounter` ?

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
jerrinot: thanks!

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
jerrinot: well spotted! It was a quick hack to Checkstyle stop bitching about "too many dependencies". 
(yes, I'm a terrible person)

Let me find a better solution.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/BalancerState.java
jerrinot: sure. will do. I'm surprised Checkstyle didn't catch it!

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOSelectorLoadCalculator.java
jerrinot: that's a good one! 

"MigratableHandler" was originally named just "Migratable" so it made sense back then. Now, not so much :) 

Thanks!

hazelcast/src/main/java/com/hazelcast/util/ItemCounter.java
jerrinot: oops, I just spotted no unit tests are in this branch. let me send an update!

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
pveentjer: I would just supress the issue in checkstyle for the time being. We need to cleanup the io package anyway.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -62,6 +62,7 @@
     public static final String PROP_VERSION_CHECK_ENABLED = "hazelcast.version.check.enabled";
     public static final String PROP_PREFER_IPv4_STACK = "hazelcast.prefer.ipv4.stack";
     public static final String PROP_IO_THREAD_COUNT = "hazelcast.io.thread.count";
+    public static final String PROP_IO_HANDLER_MIGRATION_INTERVAL_SECONDS = "hazelcast.io.handler.migration.interval.seconds";
```
pveentjer: Can you add more documentation to the property? For external people this property will be magic. And would it not make sense to call it e.g. PROP_IO_IMBALANCE_DETECTOR_INTERVAL_SECONDS. That there is an IO handler is an implementation detail. Also that a migration happens is an implementation detail. 

With this name you focus on what is happening, instead of how it is happening. 

Perhaps there are better names than the one I came up with.

We also need the ability to disable the mechanism when for whatever reason it causes problems.

hazelcast/src/main/java/com/hazelcast/nio/IOService.java
```diff
@@ -95,6 +95,8 @@
 
     int getConnectionMonitorMaxFaults();
 
+    int getHandlerMigrationIntervalSeconds();
```
pveentjer: No Javadoc?

hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
```diff
@@ -239,6 +239,11 @@ public int getConnectionMonitorMaxFaults() {
     }
 
     @Override
+    public int getHandlerMigrationIntervalSeconds() {
+        return node.groupProperties.HANDLER_MIGRATION_INTERVAL_SECONDS.getInteger();
```
pveentjer: I don't know how often you are going to read this value; if it is just once I'm fine with it. Otherwise I suggest pulling it into a final field.

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
```diff
@@ -93,4 +112,47 @@ final void registerOp(int operation) {
...
+    final void unregisterOp(int operation) {
+        SelectionKey selectionKey = getSelectionKey();
+        try {
+            selectionKey.interestOps(selectionKey.interestOps() & ~operation);
+        } catch (Throwable e) {
+            handleSocketException(e);
+        }
+    }
+
+    public IOSelector getOwner() {
```
pveentjer: Can you add javadoc?

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
jerrinot: that's a good point!

hazelcast/src/main/java/com/hazelcast/nio/tcp/MigratableHandler.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.nio.tcp;
+
+/**
+ * Read/Write handlers that supports migration between {@link com.hazelcast.nio.tcp.IOSelector}.
+ *
+ */
+public interface MigratableHandler extends SelectionHandler {
```
pveentjer: Do we have any handlers that are not migratable? If we don't have such handlers, I would suggest merging these methods into the SelectionHandler interface.

hazelcast/src/main/java/com/hazelcast/nio/IOService.java
jerrinot: will do

hazelcast/src/main/java/com/hazelcast/nio/tcp/ReadHandler.java
```diff
@@ -31,16 +31,17 @@
...
     private final ByteBuffer inputBuffer;
 
-    private final IOSelector ioSelector;
-
     private SocketReader socketReader;
 
     private volatile long lastHandle;
 
+    //This field will be incremented by a single thread. It can be read by multiple threads.
+    private volatile long noOfEvents;
```
pveentjer: Bit difficult to read due to no0Of... I would go for eventCount.

hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
jerrinot: yeah, it's read when a member is starting only. I followed the same pattern as other I/O components.

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
jerrinot: sure

hazelcast/src/main/java/com/hazelcast/nio/tcp/ReadHandler.java
```diff
@@ -60,7 +61,15 @@ public void run() {
     }
 
     @Override
+    public long getNoOfEvents() {
```
pveentjer: I would call this method 'getEventCount'.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
```diff
@@ -124,12 +125,14 @@ public TcpIpConnectionMonitor createNew(Address endpoint) {
...
+    private HandlerBalancer handlerBalancer;
     private final Map<String, Integer> selectorIndexPerHostMap;
+    private final LoggingService loggingService;
 
-    public TcpIpConnectionManager(IOService ioService, ServerSocketChannel serverSocketChannel) {
+    public TcpIpConnectionManager(IOService ioService, ServerSocketChannel serverSocketChannel, LoggingService loggingService) {
         this.ioService = ioService;
         this.serverSocketChannel = serverSocketChannel;
-        this.logger = ioService.getLogger(TcpIpConnectionManager.class.getName());
+        this.logger = loggingService.getLogger(TcpIpConnectionManager.class.getName());
```
pveentjer: You should be able to obtain a logger from the IOService.

hazelcast/src/main/java/com/hazelcast/nio/tcp/MigratableHandler.java
jerrinot: We have - the clients handlers for example. In the old implementation I placed in into the SelectorHandler, but then the ClientHandlers had to throw UnsupportedOperationException. 

I like the current way better.

hazelcast/src/main/java/com/hazelcast/nio/tcp/ReadHandler.java
jerrinot: I agree

hazelcast/src/main/java/com/hazelcast/nio/tcp/ReadHandler.java
jerrinot: yes

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
pveentjer: Or

IO_BALANCER_INTERVAL_SECONDS.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
jerrinot: yeah, I don't like it though. LoggingService should be the entry-point when it comes to obtaining loggers.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
```diff
@@ -530,6 +543,7 @@ private void closeServerSocket() {
     private void stop() {
         live = false;
         log(Level.FINEST, "Stopping ConnectionManager");
+        handlerBalancer.stop();
```
pveentjer: Is stopping sufficient or do we need to wait for the thread terminate?

jerrinot: I believe stop is sufficient here

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/HandlerBalancer.java
```diff
@@ -0,0 +1,179 @@
...
+ * Measuring interval can be customized via {@link com.hazelcast.instance.GroupProperties#HANDLER_MIGRATION_INTERVAL_SECONDS}
+ *
+ * It doesn't leverage {@link com.hazelcast.nio.ConnectionListener} capability
+ * provided by {@link com.hazelcast.nio.ConnectionManager} to observe connections as it has to be notified
+ * right after a physical TCP connection is created whilst <code>ConnectionListener</code> is notified only
+ * after a successful (Hazelcast) binding process.
+ *
+ *
+ */
+public class HandlerBalancer implements Runnable {
```
pveentjer: Few suggestions:

Rename this class to IOBalancer.

Don't implement Runnable, but have an internal Thread implementation. Nobody needs to call run on your HandlerBalancer, so no need to expose it. I would totally internalize the thread like we do for a lot of other services. For example the SlowOperationDetector.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/HandlerBalancer.java
```diff
@@ -0,0 +1,179 @@
...
+                TimeUnit.SECONDS.sleep(migrationIntervalSeconds);
+            }
+        } catch (InterruptedException e) {
+            //this thread is about to exit, no reason restoring the interrupt flag
+            EmptyStatement.ignore(e);
+        }
+    }
+
+    public void stop() {
+        isActive = false;
```
pveentjer: Perhaps an idea to do an interrupt as well to make sure the thread breaks free from its sleep. Check the SlowOperationDetector for an example.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/HandlerBalancer.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public void start() {
+        Thread migratorThread = new Thread(this);
```
pveentjer: Have a look at the thread creation. Probably you want to make use of the HazelcastThreadGroup (all threads in this group are notified when the HZ instance terminates)

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/HandlerBalancer.java
jerrinot: good point. 

originally it was run on a thread provided externally. the thread is now internal anyway, so there is no reason to expose it as runnable.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/HandlerBalancer.java
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    private boolean shouldStart(InSelectorImpl[] inSelectors, OutSelectorImpl[] outSelectors) {
+        return inSelectors.length > 1 || outSelectors.length > 1;
+    }
+
+    private void tryToMigrate(BalancerState balancerState) {
+        MigratableHandler handler = strategy.findHandlerToMigrate(balancerState);
+        if (handler != null) {
```
pveentjer: I would go for

```
if(handler == null){
   log...
   return;
}

the cool stuff
```

This way you jump out of a method as soon as you figure out there is no point in staying. It also reduces cyclomatic complexity.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOSelectorLoadCalculator.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private void clearWorkingState() {
+        handlerEventsCounter.reset();
+        selectorEvents.reset();
+        for (Set<MigratableHandler> handlerSet : selectorToHandlers.values()) {
+            handlerSet.clear();
+        }
+    }
+
```
pveentjer: Why these empty lines?

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOSelectorLoadCalculator.java
```diff
@@ -0,0 +1,216 @@
...
+        printDebugTable();
+        return state;
+    }
+
+    private void calculateNewFinalState() {
+        state.minimumEvents = Long.MAX_VALUE;
+        state.maximumEvents = Long.MIN_VALUE;
+        state.sourceSelector = null;
+        state.destinationSelector = null;
+        for (AbstractIOSelector selector : allIOSelectors) {
```
pveentjer: I would rename allIOSelectors to selectors.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOSelectorLoadCalculator.java
```diff
@@ -0,0 +1,216 @@
...
+            long noOfEvents = selectorEvents.get(selector);
```
pveentjer: use eventCount.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOSelectorLoadCalculator.java
```diff
@@ -0,0 +1,216 @@
...
+
+
+    private void calculateNewWorkingState() {
+        for (MigratableHandler handler : handlers) {
+            calculateHandlerState(handler);
+        }
+    }
+
+    private void calculateHandlerState(MigratableHandler handler) {
+        long noOfHandlerEvents = getNoOfEventsSinceLastCheck(handler);
```
pveentjer: Use event count.. the lower and upper case oO is a pain.

hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
pveentjer: That is fine. Ignore my comment.

hazelcast/src/main/java/com/hazelcast/nio/tcp/MigratableHandler.java
pveentjer: In that case it makes sense.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
pveentjer: I know. I would like to get rid of the IOService.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
pveentjer: If you sleep for 20 seconds, then the thread is not going to terminate itself for 20 seconds. With an interrupt you can let him break free.

pveentjer:
Can you rebase?

bwzhang2011:
@jerrinot, how about some performance test brought in just like the improvement comparison like @pveentjer  did for working-steal or else.

pveentjer:
verify

pveentjer:
@jerrinot can you post the benchmark results you created for future reference.

pveentjer:
@gurbuzali @mdogan can you have a last look as well. This good deep into the IO system and you are the most knowledgeable about that part of the system. For me it is a good to go.

pveentjer:
@jerrinot did you provide the option to disable the mechanism? E.g. using the HANDLER_MIGRATION_INTERVAL_SECONDS

gurbuzali:
looks good to go

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -62,6 +62,18 @@
...
     public static final String PROP_IO_THREAD_COUNT = "hazelcast.io.thread.count";
+
+    /**
+     * The interval in seconds between {@link com.hazelcast.nio.tcp.handlermigration.IOBalancer IOBalancer}
+     * executions. The shorter intervals will catch I/O Imbalance faster, but they will cause higher overhead.
+     *
+     * Please see documentation of {@link com.hazelcast.nio.tcp.handlermigration.IOBalancer IOBalancer} for
+     * detailed explanation of the problem.
+     *
+     * Default value is 20 seconds. A negative value disables the balancer.
```
Donnerbart: I don't know how fast an imbalance can occur, be detected and resolved. So should a config value of `0` seconds be a valid interval (from the code it is) or is a minimum of `1` second sufficient?

jerrinot: 0 simply means there will be no sleep between executions. 
It doesn't make much sense, but it's not really too different from e.g. 1.

mdogan:
everything is good, but having a [wip](https://github.com/jerrinot/hazelcast/commit/d54177635fc63b64842318dd488aac115a884329) (work-in-progress) in commit history seems ... unexpected.

Issue #4855, opened 2015-03-22
NPE Slow operation detector


pveentjer:
I think it is caused by a data race. 

* So the OperationRunner.currentTask is seen as slow.
* The operation completes, so current task is null
* The SlowOperationLog is created by accessing the OperationRunner.currentTask. 
which leads to the NPE. 

So make sure you take a 'snapshot' of reality that is valid instead of dealing with something that moves underneath your feet. In most cases storing the operation in a local variable (and passing it around) is the best approach to deal with this.

Could also be that something else is happening :)

```
Exception in thread "hz._hzInstance_39_dev.SlowOperationDetectorThread" java.lang.NullPointerException
	at com.hazelcast.spi.impl.operationexecutor.slowoperationdetector.SlowOperationLog.<init>(SlowOperationLog.java:59)
	at com.hazelcast.spi.impl.operationexecutor.slowoperationdetector.SlowOperationDetector$DetectorThread.getOrCreateSlowOperationLog(SlowOperationDetector.java:206)
	at com.hazelcast.spi.impl.operationexecutor.slowoperationdetector.SlowOperationDetector$DetectorThread.scanOperationRunner(SlowOperationDetector.java:184)
	at com.hazelcast.spi.impl.operationexecutor.slowoperationdetector.SlowOperationDetector$DetectorThread.scan(SlowOperationDetector.java:157)
	at com.hazelcast.spi.impl.operationexecutor.slowoperationdetector.SlowOperationDetector$DetectorThread.run(SlowOperationDetector.java:138)
```

Donnerbart:
Could someone please merge #4827 before I fix this issue?

PR #4881, opened 2015-03-24
Fixed NPE in SlowOperationDetector

Donnerbart:
* fixed NPE in `SlowOperationDetector`
* simplified `totalInvocation` increment (no contention, so `AtomicInteger` doesn't hurt)
* got rid of double stored `id` in `Invocation` (can be fetched from map entry)
* cleanup of custom test methods

Fixes issue #4855.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
@@ -153,17 +153,18 @@ public final class SlowOperationDetector {
         private void scan(long nowNanos, long nowMillis, OperationRunner[] operationRunners,
                           CurrentOperationData[] currentOperationDataArray) {
             for (int i = 0; i < operationRunners.length && running; i++) {
-                if (operationRunners[i].currentTask() == null) {
-                    continue;
-                }
-
                 scanOperationRunner(nowNanos, nowMillis, operationRunners[i], currentOperationDataArray[i]);
             }
         }
 
         private void scanOperationRunner(long nowNanos, long nowMillis, OperationRunner operationRunner,
                                          CurrentOperationData operationData) {
-            int operationHashCode = System.identityHashCode(operationRunner.currentTask());
+            Object operation = operationRunner.currentTask();
+            if (operation == null) {
+                return;
+            }
+
+            int operationHashCode = System.identityHashCode(operation);
             if (operationData.operationHashCode != operationHashCode) {
                 // initialize currentOperationData for newly detected operation
                 operationData.operationHashCode = operationHashCode;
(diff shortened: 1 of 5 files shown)


hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector_JsonTest.java
```diff
@@ -92,7 +93,7 @@ public void testJSON_SlowEntryProcessor() {
         map.executeOnEntries(new SlowEntryProcessor(2));
         waitForAllOperationsToComplete(instance);
 
-        //System.out.println(getTimedMemberState(instance).toJson());
+        //System.out.println(getOperationStats(instance));
```
jerrinot: is this useful? if so, why not `log.finest(stuff)` ?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
```diff
@@ -184,44 +185,51 @@ private void scanOperationRunner(long nowNanos, long nowMillis, OperationRunner
...
+                int totalInvocations = log.totalInvocations.incrementAndGet();
+                operationData.invocation = log.getOrCreateInvocation(operationHashCode, durationNanos, nowNanos, nowMillis);
 
-            logSlowOperation(operationRunner, log);
+                logSlowOperation(operation, log, totalInvocations);
+            }
         }
 
-        private SlowOperationLog getOrCreateSlowOperationLog(final OperationRunner operationRunner) {
+        private SlowOperationLog getOrCreateSlowOperationLogOrNull(OperationRunner operationRunner, Object operation) {
```
ahmetmircik: The name `getOrCreateSlowOperationLogOrNull` implies a dilemma, if i can get or create, how can i see a null-value-returned?

Donnerbart: I prevent creating the stack trace string and the log if the operation has changed after the stack trace creation (I cannot control that). In this case I return `null` to signal that the log creation was aborted.

Let me split the method to make this more clear.

ahmetmircik: thank you

Donnerbart:
The problem was that I always worked with `OperationRunner.currentTask()` which may change or become `null` during the processing of the operation.

The operation is now cached and compared to `OperationRunner.currentTask()` after creating the stack trace (which may take a some milliseconds). If the operation has changed I abort, because the stack trace may already be outdated (new operation or no operation running).

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
```diff
@@ -253,9 +266,8 @@ private void shutdown() {
             detectorThread.interrupt();
             try {
                 detectorThread.join(SLOW_OPERATION_THREAD_MAX_WAIT_TIME_TO_FINISH);
-            } catch (InterruptedException e) {
-                EmptyStatement.ignore(e);
-                // TODO: Interrupt flag is consumed here
+            } catch (InterruptedException ignored) {
+                detectorThread.interrupt();
```
pveentjer: You are restoring the interrupt on the wrong thread. You should restore the interrupt on the calling thread if the join fails, not on the detectorThread. 

And can you split it up in a shut down, that signals to the detector thread to kill himself and add a seperate awaitTermination method? This makes it possible to shutdown all services in 1 go.. and then do the actual waiting.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
```diff
@@ -183,45 +184,57 @@ private void scanOperationRunner(long nowNanos, long nowMillis, OperationRunner
...
+            }
+
+            String stackTrace = stackTraceStringBuilder.toString();
             stackTraceStringBuilder.setLength(0);
+            return stackTrace;
+        }
 
-            Integer hashCode = stackTraceString.hashCode();
-            SlowOperationLog candidate = slowOperationLogs.get(hashCode);
+        private SlowOperationLog getOrCreateSlowOperationLog(String stackTrace, Object operation) {
```
pveentjer: What I often do is to base the naming on the context. Since you are doing stuff with slow operation logs and you know the return type, I often go for a shorter name like 'getOrCreate'. 

This prevents running into those long lines you need to chew through.

pveentjer:
Minor review comments.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
Donnerbart: Good catch. I'll revert my change so we can do this in another PR (this one is already too big).

PR #4884, opened 2015-03-24
fail-fast impl. for new client module

bilalyasar:


hazelcast-client-new/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -170,39 +185,55 @@ private void parseAndBuildConfig(ClientConfig clientConfig) throws Exception {
...
+        if (SECURITY.isEqual(nodeName)) {
+            handleSecurity(node);
+        } else if (PROXY_FACTORIES.isEqual(nodeName)) {
+            handleProxyFactories(node);
+        } else if (PROPERTIES.isEqual(nodeName)) {
+            fillProperties(node, clientConfig.getProperties());
+        } else if (SERIALIZATION.isEqual(nodeName)) {
+            handleSerialization(node);
+        } else if (NATIVE_MEMORY.isEqual(nodeName)) {
+            fillNativeMemoryConfig(node, clientConfig.getNativeMemoryConfig());
```
pveentjer: I would give all these methods the same name. I would go for parse.
so parseHandleSecurity, parseGroup, parseExecutorPoolSize etc.

If you make the method names the same, then there is no confusion for other people reading/maintaining your code.

hazelcast-client-new/src/main/resources/hazelcast-client-config-3.5.xsd
```diff
@@ -401,10 +401,10 @@
         </xs:sequence>
     </xs:complexType>
     <xs:complexType name="ssl">
-        <xs:sequence>
+        <xs:all>
```
pveentjer: My xsd knowledge is extremely rusty, but can it happen that a user now adds multiple similar sections? E.g. multiple times he configures 'size'.

bilalyasar: @pveentjer no,  `all` means 0 or 1 for every item. So user cannot configure `size` more than 1.

Issue #4849, opened 2015-03-20
NullPointerException and IllegalStateException occur when running 'com.hazelcast.simulator.tests.icache.EvictionICacheTest'


cherednichenko:
Version 3.4.
Test: com.hazelcast.simulator.tests.icache.EvictionICacheTest

Logs:
```
INFO  14:22:24 --------------------------------------------------------------
Running Test : test2
TestCase{
      id=test2
    , class=com.hazelcast.simulator.tests.icache.EvictionICacheTest
}
--------------------------------------------------------------
INFO  14:22:24 test2 Starting Test initialization
INFO  14:22:25 test2 Completed Test initialization
INFO  14:22:25 test2 Starting Test setup
WARN  14:22:27 Failure #1  client:10.231.179.195 Worker exception[java.lang.NullPointerException]
WARN  14:22:27 Failure #2  10.231.179.195:5701 Worker exception[java.lang.NullPointerException]
WARN  14:22:27 Failure #3  client:10.231.179.195 Worker exception[java.lang.NullPointerException]
INFO  14:22:27 test2 Completed Test setup
INFO  14:22:27 test2 Starting Test local warmup
INFO  14:22:29 test2 Completed Test local warmup
INFO  14:22:29 test2 Starting Test global warmup
INFO  14:22:31 test2 Completed Test global warmup
INFO  14:22:31 test2 Starting Test start
INFO  14:22:32 test2 Completed Test start
INFO  14:22:32 test2 Test will run for 00d 00h 01m 00s
INFO  14:22:32 test2 Critical Failure detected, aborting execution of test
INFO  14:22:32 test2 Test finished running
INFO  14:22:32 test2 Starting Test stop
WARN  14:22:33 Failure #4  client:10.231.179.195 Worker exception[java.lang.IllegalStateException: Sum of operation probabilities should be exactly 1.0, but is 0.000000]
WARN  14:22:33 Failure #5  client:10.231.179.195 Worker exception[java.lang.IllegalStateException: Sum of operation probabilities should be exactly 1.0, but is 0.000000]
INFO  14:22:34 test2 Completed Test stop
```

from failure file:
```
Failure[
   message='Worked ran into an unhandled exception'
   type='Worker exception'
   agentAddress=10.231.179.195
   time=Fri Mar 20 14:22:26 UTC 2015
   workerAddress=client:10.231.179.195
   workerId=worker-10.231.179.195-15-client
   test=TestCase{
          id=test2
[...]

PR #4901, opened 2015-03-25
Fixing Cache Null issue.

cherednichenko:
During investigation those two issues:
- https://github.com/hazelcast/hazelcast/issues/4849 
- https://github.com/hazelcast/hazelcast-simulator/issues/568 

we found that when we put NULL into the cache - it throws NullPointerException.

This unit test was created to catch this situation. 

This test is PASSED.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheConfigTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheConfigTest.java
@@ -31,6 +31,7 @@ import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.TestHazelcastInstanceFactory;
 import com.hazelcast.test.annotation.QuickTest;
+import com.hazelcast.util.EmptyStatement;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/cache/CacheConfigTest.java
```diff
@@ -203,6 +199,19 @@ public void defaultCacheTest() {
     }
 
     @Test
```
bilalyasar: instead of using try/catch block, using `@Test(expected = NullPointerException.class)` can be more simple and useful.

cherednichenko: No, we discussed it with Peter already and defined that try/catch will be better.

cherednichenko: @bilalyasar 
As the matter of fact NullPointer occurs here:
```
cacheManager.createCache(cacheName, null);
```
and it won't check the second row:
```
assertNotNull(cache);
```
that's why putting ```@Test(expected = NullPointerException.class)``` won't work better than current one, visa versa it's more confusing.

But for others situations using ```@Test ...``` is simpler.

bilalyasar: i couldn't see `assertNotNull(cache);` in your code, am i missing sth?

cherednichenko: it was before - then we reworked.
Don't worry - Peter's already checked this code :)

bilalyasar: :+1:  i am definitely not worry :)  only, i want to learn real reason of this style so i can learn new things and apply in my code in the future, @pveentjer

cherednichenko: :+1: I'm also on the way of learning the different styles - big thanks @pveentjer .

serkan-ozal:
:+1:

gurbuzali:
verify

gurbuzali:
verify

Issue #4677, opened 2015-02-25
IExecutorService future.cancel issues


sancar:
In case of submitting a callable to a member via IExecutorService there is no guarantee that future.cancel will work after callable start to run.(Because they do not run on an assigned partition thread) 
I added new  tests as ClientExecutorServiceCancelTest to demonstrate the failure. Related tests are currently ignored. Although tests is for client there is a high probability that this is an issue for nodes also.(I did not check it).

Holmistr:
Once this issue is fixed, please don't forget to re-enable test in https://github.com/hazelcast/hazelcast/issues/4199

burakcelebi:
@ihsandemir do you think this is fixed with https://github.com/hazelcast/hazelcast/pull/15526?

burakcelebi:
@ihsandemir I checked this with @sancar. The answer is no :)

PR #4909, opened 2015-03-26
Refreshable Request Interface is removed as a preparation for client pro...

sancar:
...tocol start. New tests are added for IExecutorService cancel future(client). 

Client Request in client invocation related classes are converted to client message.


sancar:
A note on  IExecutorService cancel future . It seems that we do not have order guarantee when invocation done via address. And for partition invocations when retry kicks in, order is not guaranteed. Because of all these reasons it is really hard to implement cancel correctly. It seems that we do not have tests for this so far, because of that we never saw the problems with it. 
This is a the issue related to these changes https://github.com/hazelcast/hazelcast/issues/4677.  

We can either leave the tests written in this pr as is (and ignored) , to be fixed in the future. Or I can change the future.cancel return always false, and write tests accordingly. It will basically mean we do not support cancel on client side. 
I can also add sleep to test to make sure test is passes but I really dont want to create another randomly failing test.

These issues also present in the node side. Issue https://github.com/hazelcast/hazelcast/issues/4677 also addresses node side problems.

asimarslan:
It looks good.

PR #4912, opened 2015-03-27
hazelcast-spring module pom.xml cleanup

mesutcelik:
Once this merged, application developers are not required to provide **hazelcast** or **hazelcast-client** dependencies in their pom.xml. **hazelcast-spring** dependency will be sufficient to include hazelcast into spring applications.


mesutcelik:
verify

mesutcelik:
verify

PR #4913, opened 2015-03-27
Implement partition lost listener feature

metanet:
* Partition lost detection is implemented within com.hazecast.partition.impl.ForceUpdateSyncWaitingReplicaVersionsOperation, com.hazelcast.partition.impl.ResetReplicaVersionOperation and com.hazelcast.partition.impl.InternalPartitionServiceImpl.LocalPartitionListener classes.
* New listener interfaces and related event classes are defined in com.hazelcast.partition and com.hazelcast.map packages.
* All partition-related logs under com.hazelcast.partition are converted to the same format for easier grepping. partitionId information is logged with partitionId=* pattern and replicaIndex information is logged with replicaIndex=*
* Changes under com.hazelcast.core are for new listener registration methods on the existing API interfaces.
* Changes under com.hazelcast.config are for registering partition lost listeners via configuration.
* Changes in com.hazelcast.instance.Node are for registering a partition lost listener via configuration.
* Changes under com.hazelcast.map are for handling internal partition lost events and dispatching map partition lost events.
* Changes under com.hazelcast.client.proxy are for registering partition lost listeners from Hazelcast client.


metanet:
verify

sancar:
Everything seems OK for client side.

mdogan:
> All partition-related logs under com.hazelcast.partition are converted to the same format for easier grepping. partitionId information is logged with partitionId=* pattern and replicaIndex information is logged with replicaIndex=*

:100:

hazelcast/src/main/java/com/hazelcast/core/IMap.java
```diff
@@ -898,6 +899,38 @@ boolean tryLock(K key, long time, TimeUnit timeunit)
...
+
+    /**
+     * Adds a MapPartitionLostListener.
+     * <p/>
+     * The addPartitionLostListener returns a register-id. This id is needed to remove the MapPartitionLostListener using the
+     * {@link #removePartitionLostListener(String)} method.
+     * <p/>
+     * There is no check for duplicate registrations, so if you register the listener twice, it will get events twice.
+     * IMPORTANT: Please @see com.hazelcast.partition.PartitionLostListener for weaknesses
+     * IMPORTANT: Listeners registered from HazelcastClient are assigned to nodes randomly. When a node to which
```
mdogan: Isn't this an implementation detail?

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/PartitionServiceProxy.java
```diff
@@ -72,6 +73,16 @@ public boolean removeMigrationListener(String registrationId) {
     }
 
     @Override
+    public String addPartitionLostListener(PartitionLostListener partitionLostListener) {
```
mdogan: Isn't it unexpected you can register a map-partition-lost-listener from client side but not regular partition-lost-listener?

hazelcast/src/main/java/com/hazelcast/partition/impl/ForceUpdateSyncWaitingReplicaVersionsOperation.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    private void sendPartitionLostEvent(int partitionId, int lostReplicaIndex) {
+        final InternalPartitionLostEvent partitionLostEvent =
+                new InternalPartitionLostEvent(partitionId, lostReplicaIndex, getNodeEngine().getThisAddress());
+        NodeEngineImpl nodeEngine = (NodeEngineImpl) getNodeEngine();
+        for (PartitionAwareService service : nodeEngine.getServices(PartitionAwareService.class)) {
+            try {
+                service.onPartitionLost(partitionLostEvent);
```
mdogan: `PartitionAwareService`s are called on partition thread here. Is this intentional? Do you actually need to block partition thread? Or can this be offloaded to another thread?

hazelcast/src/main/java/com/hazelcast/partition/impl/ForceUpdateSyncWaitingReplicaVersionsOperation.java
```diff
@@ -0,0 +1,128 @@
...
+import com.hazelcast.spi.impl.NodeEngineImpl;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+// Runs locally when the node becomes owner of a partition upon a node failure
+// Finds the replica indices that are on the sync-waiting state. Those indices represents the lost backups of the partition.
+// Therefore, it publishes InternalPartitionLostEvent objects to notify related services
+// It also updates the version for lost replicas to the first available version value after the lost backups, or 0 if N/A
+final class ForceUpdateSyncWaitingReplicaVersionsOperation
```
mdogan: This operation and `PromoteFromBackupOperation` are both executed when a node becomes owner of a partition. What about merging both to a single operation? Also  `PromoteFromBackupOperation` contains some logic from this operation, see `setMissingReplicaVersions()`, which I borrowed from your branch before...

hazelcast/src/main/java/com/hazelcast/partition/impl/ForceUpdateSyncWaitingReplicaVersionsOperation.java
metanet: You are right, it should be better to move it to another thread.

hazelcast/src/main/java/com/hazelcast/core/IMap.java
metanet: I just want to inform user that some events may be lost if the listener is registered via client. I will do it with less details.

metanet:
verify

PR #4917, opened 2015-03-27
Added new tests for Executor future cancel. The tests are ignored for now

sancar:
Related issue #4677 . And there are some cleanups in  ExecutorService test .

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketClientDataReaderNew.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketClientDataReaderNew.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2013, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


jerrinot:
I don't understand how it's even possible to make these tests to pass. What if the task is already in progress or even finished at the time when the test thread is calling `cancel()`?

sancar:
It can not finish since there is too long sleep time. And we are interrupting the running tasks to cancel in the remote. See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java#L113

sancar:
I have send another pr for this commit https://github.com/hazelcast/hazelcast/pull/4918
, since this one started to act weird.

jerrinot:
I see, thank you for explanation!

mtopolnik:
@jerrinot The Future API is decoupled from the task which will produce its value. You can cancel a future at any time before it is done, and you'll get the `true` response if your call caused the change of future's status to cancelled.

jerrinot:
I assume if the task is already running then just the interrupt flag is set on a thread and it's responsibility of the task to check it?

mtopolnik:
In the case of JDK's Future, the thread isn't even interrupted if you don't explicitly allow it with fut.cancel(true).

PR #4928, opened 2015-03-28
Prevent running operations on unwanted threads

### LINKED ISSUE #4929: Prevent running operations on unwanted threads
It can happen that on the ResponseThread and the InvocationRegistry.InspectionThread operations are going to be executed. In both cases they can trigger the Invocation retrying behavior. The OperationExecutor is free to execute Operations on the calling thread as long as it isn't a NIOTHread or if there are partition conflicts. 

The potential consequence of having operations run on these threads is:
- bad performance since responses don't come in
- deadlock: imagine that the response thread is getting blocked waiting for a response from a remote system
- system instability because important system messages are slowly getting picked or potentially not picked up
- operation timeout functionality not working correctly because the InspectionThread is hogged to do different things.

pveentjer:
ResponseThread and InvocationRegistry.InspectionThread reset and retry operations. Since these threads didn't implement NIOTHread, the OperationExecutor is free to execute tasks on these threads and that is not something desirable.

Running operations on these threads could lead to performance problems, cluster stability problems and deadlocks.

Fixes #4929

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/NIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/NIOThread.java
@@ -1,7 +1,0 @@
-package com.hazelcast.nio;
-
-/**
- * Marker interface for non-blocking IO threads.
- */
-public interface NIOThread {
-}
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/IOSelector.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/IOSelector.java
@@ -16,11 +16,11 @@
 
 package com.hazelcast.nio.tcp;
 
-import com.hazelcast.nio.NIOThread;
+import com.hazelcast.spi.impl.operationexecutor.OperationHostileThread;
 
 import java.nio.channels.Selector;
 
-public interface IOSelector extends NIOThread {
+public interface IOSelector extends OperationHostileThread {
 
     Selector getSelector();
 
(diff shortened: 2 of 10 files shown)


pveentjer:
verify

jerrinot:
I'm not super excited about marking the Cleanup Thread as NIOThread. It's not NIOThread after all.

pveentjer:
The thing you are not excited about is the name: NIOThread.

jerrinot:
that's quite possible. 
maybe just renaming the marker interface to something more suitable would do the trick.

pveentjer:
My kingdom for a good name.

Donnerbart:
`NOOThread` No Operation Offloading Thread
`NORThread` No Operation Runner Thread

pveentjer:
NOOThread... did a key get stuck on your keyboard? :)

Donnerbart:
Noooooo ;)

pveentjer:
OperationHostileThread.

Donnerbart:
NOPThread

pveentjer:
test... trying to see if I can place a comment

pveentjer:
verify

mdogan:
Since the needed behaviour is very specific for the marker interface, what about something like 
`com.hazelcast.nio.NIOThread` -> `com.hazelcast.spi.imp.InvocationDisallowedThread`?

pveentjer:
Sure. I'll fix it in the master. I'll leave it as it is in the backport.

mdogan:
> What about InvocationHostileThread?

that's fine too.

PR #4930, opened 2015-03-28
Backport: Prevent running operations on unwanted threads

### LINKED ISSUE #4929: Prevent running operations on unwanted threads

pveentjer:
Fixes #4929

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationScheduler.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/BasicOperationScheduler.java
@@ -486,7 +486,7 @@ public final class BasicOperationScheduler {
         }
     }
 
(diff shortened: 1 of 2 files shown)


pveentjer:
verify

pveentjer:
verify

PR #4934, opened 2015-03-28
Performance: less litter in remote response handler

pveentjer:
Inside the RemoteInvocationResponseHandler: Removed the AtomicBoolean and replaced ...it by a AtomicIntegerFieldUpdater. So 1 less AtomicBoolean instance being created per remote invocation.


jerrinot:
:+1:

PR #4938, opened 2015-03-30
Performance: Reduce ClientRequest callback litter

pveentjer:
The PartitionClientRequest and TargetClientRequest create a Callback instance to listen to the operation completion. Instead of creating an instance, the Request implements the ExecutionCallback interface.

This saves prevents 2 object from being created:
- the ExecutionCallback instance. 
- the Callback interface is wrapped in an ExecutionCallbackAdapter to make it compatible. 

By making use of an ExecutionCallback and letting the request implement this interface, we remove 2 object instances per client request.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/client/PartitionClientRequest.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/client/PartitionClientRequest.java
@@ -17,14 +17,15 @@
 package com.hazelcast.client.impl.client;
 
 import com.hazelcast.client.ClientEndpoint;
-import com.hazelcast.spi.Callback;
+import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.spi.InvocationBuilder;
 import com.hazelcast.spi.Operation;
 
 /**
  * Base class for partition based client request.
  */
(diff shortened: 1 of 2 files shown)


pveentjer:
verify

mdogan:
verify

jerrinot:
:+1:

mdogan:
verify

pveentjer:
verify

mdogan:
@pveentjer, this is failing because of following method:

```java
@Override
public void andThen(ExecutionCallback<E> callback, Executor executor) {
    isNotNull(callback, "callback");
    isNotNull(executor, "executor");

    synchronized (this) {
        if (response != null && !(response instanceof InternalResponse)) {
            runAsynchronous(callback, executor);
            return;
        }

        this.callbackHead = new ExecutionCallbackNode<E>(callback, executor, callbackHead);
    }
}
```

When `Invocation::NULL_RESPONSE` is set as response, callback is never executed and response is never received on client side.

pveentjer:
Thanks.

So my code change exposed a bug which was already in the system. I'll create a different PR for that.

pveentjer:
Need to get this PR merged first:

https://github.com/hazelcast/hazelcast/pull/5003

gurbuzali:
verify

PR #4941, opened 2015-03-30
PartitionClientRequest makes use of efficient partitionid method

### LINKED ISSUE #4940: Performance: PartitionClientRequest inefficient determining partition-id
ClientRequests are very inefficient determining the partition id.

E.g. in case of the AbstractAlterRequest:

```
 @Override
    protected int getPartition() {
        Data key = serializationService.toData(name);
        return getClientEngine().getPartitionService().getPartitionId(key);
    }
```

But the partition-id is passed to the member as part of the Packet and with the introduction of the client-protocol, it is found here as well. So we should make use of the pre-calculated partitionid instead of calculating it.

The new client protocol also has the partition-id available, see: ClientMessage 'partitionId' method.

So probably we can get rid of the 'getPartition' method completely.

@asimarslan WDYT.

Are you already passing in the partition-id on the client protocol? Or should this PR be parked?

pveentjer:
Resolves #4940

Once the build is green, I'll remove the getPartitionId code since it doesn't serve any purpose.


sancar:
This build will never be green, because implementation is not ready. We will do related changes very soon, which will be same with what you do here. We can close this one.

PR #4942, opened 2015-03-30
Performance: Add InvocationBuilder.setExecutionCallback

pveentjer:
Before you were forced to use the Callback which gets adapted to an ExecutionCallback. 

So this PR is going to reduce object litter by providing a more performance-friendly method.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.spi;
 
+import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.nio.Address;
 import com.hazelcast.partition.InternalPartition;
 import com.hazelcast.spi.impl.NodeEngineImpl;
@@ -59,6 +60,7 @@ public abstract class InvocationBuilder {
     protected final int partitionId;
     protected final Address target;
     protected Callback<Object> callback;
+    protected ExecutionCallback<Object> executionCallback;
 
     protected long callTimeout = DEFAULT_CALL_TIMEOUT;
     protected int replicaIndex;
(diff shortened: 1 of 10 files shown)


pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
```diff
@@ -224,10 +226,25 @@ public Callback getCallback() {
     }
 
     public InvocationBuilder setCallback(Callback<Object> callback) {
```
mdogan: What about deprecating `Callback` version?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
```diff
@@ -68,7 +67,7 @@
      */
     List<SlowOperationDTO> getSlowOperationDTOs();
 
-    <E> void asyncInvokeOnPartition(String serviceName, Operation op, int partitionId, Callback<E> callback);
+    void asyncInvokeOnPartition(String serviceName, Operation op, int partitionId, Object callback);
```
mdogan: Since these `asyncInvoke*` methods are new additions, why don't you just convert these to use `ExecutionCallback`?
```java
void asyncInvokeOnPartition(String serviceName, Operation op, int partitionId, ExecutionCallback<E> callback)
```

I think there's no need to support `Callback` version...

hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
pveentjer: Good idea.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
pveentjer: Another good idea.

mdogan:
verify

pveentjer:
verify

mdogan:
:+1:

PR #4945, opened 2015-03-30
Improved javadoc on time/timeout methods on callid

pveentjer:


pveentjer:
Since only javadoc + static imports I'll merge this one myself.

Issue #4525, opened 2015-02-03
Index ignores null keys


dayman:
This test will fail.
```
public class HazelcastIndexTest {

    public static class TestClass implements Serializable {
        private String field;

        public TestClass(final String field) {
            this.field = field;
        }

        public String getField() {
            return field;
        }

    }

    @Test
    public void hazelcastIndexDoesntWork() {
        final HazelcastInstance hazelcast = Hazelcast.newHazelcastInstance(null);
        final IMap<String, TestClass> map = hazelcast.getMap("map");
        
        // given
        map.addIndex("field", false); // if we remove this line, test will pass
        map.put("1000", new TestClass(null));
        map.put("1001", new TestClass("not null"));

        // when
        final Predicate emptyFieldPredicate = Predicates.notEqual("field", "value");
        final Collection<TestClass> emptyFieldValues = map.values(emptyFieldPredicate);

        // then
        assertThat(emptyFieldValues, hasSize(2));
    }

}
```
Apparently test should return both of the entries but it ignores one with "null" value.
If we remove index for "field", it will pass.

hasancelik:
Thanks for reporting your issue. Please share with us the following information, to help us resolve your issue quickly and efficiently.
* **Hazelcast version** that you are using  (e.g. 3.4, also specify minor release or latest snapshot)
* **Hazelcast configuration** (hazelcast.xml or programmatic configuration)
* **Cluster topology**
  * Number of Hazelcast member nodes
  * Number of Hazelcast clients
* **Operating System** (for Linux kernel version will be helpful)
* **Logs and Stack Traces** (if available)
* **Integration module versions** if available (e.g. Tomcat, Jetty, Spring, Hibernate)
  ** Detailed configuration information (web.xml, Hibernate configuration, Spring context.xml etc.)

dayman:
**Hazelcast version**: 3.4
**Hazelcast configuration**: I launched embedded instance (with Hazelcast.newHazelcastInstance(null)), so the configuration was default, no xml file.
And as you can see I added index manually from code. 
But also experienced absolutely same problem when launched hazelcast with xml configuration contained
```
<map name="map">
  <backup-count>0</backup-count>
  <async-backup-count>0</async-backup-count>
  <read-backup-data>true</read-backup-data>
  <indexes>
    <index>field</index>
  </indexes>
</map>
```
**Cluster topology**: one node
**Operating system**: Ubuntu 14.04
**Integration module versions**: no other libraries during the test

[...]

Issue #4373, opened 2014-12-31
Predicates with null values throws exception on unordered indexes


alparslanavci:
When a predicate with null value is applied to a map with unordered index, it operates correctly but throws exception because of casting. You can reproduce it with the following test:

```java
@Test(timeout = 1000 * 60)
public void testPredicateEqualityWithUnorderedIndex_whenPredicateValueIsNull() {
        HazelcastInstance instance = createHazelcastInstance();
        IMap<String, SampleObjects.Employee> map = instance.getMap("testPredicateEqualityWithUnorderedIndex");
        map.addIndex("name", false);
        map.addIndex("createDate", false);

        int notNullValueCount = 4;
        for (int i = 0; i < notNullValueCount; i++) {
            Employee v = new Employee();
            v.setName("name" + i);
            v.setCreateDate(new Date());
            map.put("" + i, v);
        }
        map.put(""+ notNullValueCount, new Employee());

        Predicate nullNamePredicate = Predicates.equal("name", null);
        Collection<Employee> nullNameValues = map.values(nullNamePredicate);
        Predicate notNullNamePredicate = Predicates.notEqual("name", null);
        Collection<Employee> notNullNameValues = map.values(notNullNamePredicate);
        Predicate nullCreateDatePredicate = Predicates.equal("createDate", null);
        Collection<Employee> nullCreateDateValues = map.values(nullCreateDatePredicate);
        Predicate notNullCreateDatePredicate = Predicates.notEqual("createDate", null);
        Collection<Employee> notNullCreateDateValues = map.values(notNullCreateDatePredicate);

        assertEquals(1, nullNameValues.size());
        assertEquals(notNullValueCount, notNullNameValues.size());
        assertEquals(1, nullCreateDateValues.size());
        assertEquals(notNullValueCount, notNullCreateDateValues.size());
}
```

pveentjer:
Which HZ version did you use? And can you post the whole stacktrace?

mdogan:
Following patch fixes this issue:

```
Index: ../../hazelcast/hazelcast/src/main/java/com/hazelcast/query/impl/UnsortedIndexStore.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- ../../hazelcast/hazelcast/src/main/java/com/hazelcast/query/impl/UnsortedIndexStore.java	(date 1420453289000)
[...]

PR #4946, opened 2015-03-30
NotEqualPredicate should return false if entry is null (without index) also if index present, it should not throw exception with null values

muratayan:
This pull request solves two related issues #4525 and #4373. 

For #4525, as I commented under the issue, hazelcast replicates the SQL behavior, so when something is searched as `notEqual(someValue)` we should not include null entries to the result set. When indexing is used, since IndexStores store null values in a separate map named `recordsWithNullValue` we achieve this behavior. However, when indexing is not used we do a full scan only with the predicate. 

Thus, in the not equal predicate before calling `super.apply()` (which is EqualPredicate's function), we should check if the entry is null as follows:

```java
@Override
        public boolean apply(Map.Entry entry) {
            Comparable entryValue = readAttribute(entry);
            if (entryValue == null) {
                return false;
            }

            return !super.apply(entry);
        }
```

This way, if the entry is null, it will not be included in the result set no matter what.

For #4373, I've used the @mdogan's proposed patch, also I've added `null` guards to the `TypeConverter` class. Before this `getSubRecords()` function was throwing an exception which was causing us to do full scan on every partition instead of using index.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -121,6 +121,13 @@ public class GroupProperties {
     public static final String PROP_INITIAL_WAIT_SECONDS = "hazelcast.initial.wait.seconds";
     public static final String PROP_MAP_REPLICA_SCHEDULED_TASK_DELAY_SECONDS
             = "hazelcast.map.replica.scheduled.task.delay.seconds";
+    /**
+     * PROP_MAP_EXPIRY_DELAY_SECONDS is useful to deal with some possible edge cases e.g. when using EntryProcessor,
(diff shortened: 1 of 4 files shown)


gurbuzali:
verify

pveentjer:
verify

gurbuzali:
looks good, we need a 'forward-port' to master too

mdogan:
verify

mdogan:
@muratayan, can you create a forward-port to master branch?

muratayan:
@mdogan working on it

dimas:
(I added it to 4525 but copying here as well)

Is it by design that 
```java
Predicates.not(Predicates.equal("field", "value"));
```
still returns null values while

```java
Predicates.notEquals("field", "value");
```
does not?
(It is Hazelcast 3.4.4)

PR #4947, opened 2015-03-30
ClientMessage header fields are set and made socket writable

sancar:
ClientMessage header fields are set. 
ClientMessage made  socketWritable.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
@@ -33,6 +33,7 @@ import com.hazelcast.client.spi.impl.DefaultAddressTranslator;
 import com.hazelcast.client.txn.ClientTransactionManager;
 import com.hazelcast.client.util.RoundRobinLB;
 import com.hazelcast.collection.impl.list.ListService;
+import com.hazelcast.collection.impl.queue.QueueService;
 import com.hazelcast.collection.impl.set.SetService;
 import com.hazelcast.concurrent.atomiclong.AtomicLongService;
 import com.hazelcast.concurrent.atomicreference.AtomicReferenceService;
@@ -76,7 +77,6 @@ import com.hazelcast.multimap.impl.MultiMapService;
 import com.hazelcast.nio.ClassLoaderUtil;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.nio.serialization.SerializationService;
-import com.hazelcast.collection.impl.queue.QueueService;
 import com.hazelcast.replicatedmap.impl.ReplicatedMapService;
 import com.hazelcast.security.Credentials;
 import com.hazelcast.security.UsernamePasswordCredentials;
@@ -101,6 +101,7 @@ public class HazelcastClientInstanceImpl implements HazelcastInstance {
 
     private static final AtomicInteger CLIENT_ID = new AtomicInteger();
     private static final ILogger LOGGER = Logger.getLogger(HazelcastClient.class);
+    private static final short protocolVersion = 1;
 
     private final ClientProperties clientProperties;
     private final int id = CLIENT_ID.getAndIncrement();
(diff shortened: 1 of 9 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessageFramingHelper.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package com.hazelcast.client.impl.protocol;
+
+import java.util.List;
+
+/**
+ * Chunks given client message to frames
+ */
+public interface ClientMessageFramingHelper {
```
mdogan: A helper class as an interface? are you sure?

mdogan:
I think you need to describe what's *framing* in the first place?

sancar:
I am adding following extended javadoc .
```
/**
 * Chunks given client message to frames(smaller messages) to let the client messages to be interleaved.
 * With the help of the framing, some important client messages will not be blocked because
 * of a huge client message trying to get through the socket.
 */
```

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessageFramingHelper.java
sancar: Now that you asked, I am not sure. :) . 
It was that way because we will implement framing/chunking later. 
I have a second idea. I will implement one complete helper class(without interface), which will get frame size. I will pass int max as default and later when we can implement the details of that class and pass different frame sizes. It will be more clear, ones I update the pr.

sancar:
We decided to implement chunking/framing after we had the mechanism to put the framed messages together.

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
```diff
@@ -121,6 +121,7 @@
     private final LoadBalancer loadBalancer;
     private final ClientExtension clientExtension;
     private final Credentials credentials;
+    private final short protocolVersion = 1;
```
mdogan: This is a compile time constant value, why not a static field?

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientNonSmartInvocationServiceImpl.java
```diff
@@ -29,6 +29,7 @@ public void invokeOnConnection(ClientInvocation invocation, ClientConnection con
 
     @Override
     public void invokeOnPartitionOwner(ClientInvocation invocation, int partitionId) throws IOException {
+        invocation.getClientMessage().partitionId(partitionId);
```
mdogan: We generally use `set*` prefix for setter methods in Hazelcast. I think choosing another naming convention is confusing.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
```diff
@@ -257,4 +261,40 @@ public ClientMessage getPayloadData(byte[] payload) {
         return this;
     }
 
+    @Override
+    public boolean writeTo(ByteBuffer destination) {
+        byte[] byteArray = buffer().byteArray();
+        int size = byteArray.length;
+        if (valueOffset >= size) {
```
mdogan: Can `valueOffset` be greater than `size`?

sancar: I took a second look at the code. It can be at most size. It can not be greater. And even equal check seems unnecessary. I will update

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientNonSmartInvocationServiceImpl.java
sancar: Yes, it looks odd to me too. I think we can change that.

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
sancar: Changing it as static.

mdogan:
:+1:

PR #4949, opened 2015-03-30
Single node performs map loading from a MapLoader

ajermakovics:
**Summary of functionality** 

Initial loading:
- When a RecordStore is created it checks whether its partition is the owner of map's name
- If it is, it becomes the Loader RecordStore (see below)
- If it isn't, it becomes one of Receiver RecordStores
 
Loader RecordStore
- Iterates all MapLoader's keys in a background thread and sends to Receivers in batches. Size of batches is determined using the group property 
hazelcast.map.load.chunk.size.
- It will stop iterating when all the keys are exhausted or the maximum size is reached as determined by MaxSize config.
- After finishing the iteration it will send an operation to all partitions telling that loading of keys is finished
 
Receiver RecordStore
- Upon creation sends a PartitionCheckIfLoaded operation to the Loader RecordStore. That will create the LoaderRecordStore and initiated loading in the case where it hasn't started already.
- When it receives batches of keys it loads values in a background thread and keeps a completion Future of that loading
- If waits for the operation from the Loader indicating that all keys have been sent out
- The record store is considered loaded when all of its completion Future's are done all keys have been sent out.

There are probably some details I've missed so feel free to ask. Also this doesn't yet include failover which I plan for another PR as this one is big already.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoader.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoader.java
@@ -1,0 +1,297 @@
+package com.hazelcast.map.impl;
+
+import com.hazelcast.config.MaxSizeConfig;
+import com.hazelcast.config.MaxSizeConfig.MaxSizePolicy;
+import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.core.IFunction;
+import com.hazelcast.core.MapLoader;
+import com.hazelcast.map.impl.mapstore.MapStoreContext;
+import com.hazelcast.map.impl.operation.LoadAllOperation;
+import com.hazelcast.map.impl.operation.PartitionCheckIfLoadedOperation;
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.partition.InternalPartitionService;
+import com.hazelcast.spi.ExecutionService;
+import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.OperationService;
+import com.hazelcast.spi.impl.AbstractCompletableFuture;
+import com.hazelcast.util.CollectionUtil;
+import com.hazelcast.util.UnmodifiableIterator;
+import com.hazelcast.util.ValidationUtil;
+
+import java.io.Closeable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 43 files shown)


bwzhang2011:
mapstore or loader has many issues related. @ajermakovics thanks for such for improvement and also hope for such could be merged into other branch once is confirmed.

ajermakovics:
verify

gurbuzali:
We have a test failure, might be related to the changes made in this PR:
`com.hazelcast.client.map.ClientMapLoaderExceptionHandlingTest.test_initial_map_load_propagates_exception_to_client`

ajermakovics:
Ah yes, it is probably is related. Didn't run client tests before

mdogan:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoader.java
```diff
@@ -0,0 +1,297 @@
...
+
+        for (int partitionId = 0; partitionId < partitions; partitionId++) {
+            LoadAllOperation op = new LoadAllOperation(mapName, Collections.<Data>emptyList(), replaceExistingValues, lastBatch);
+            futures.add(opService.invokeOnPartition(SERVICE_NAME, op, partitionId));
+        }
+
+        return futures;
+    }
+
+    public static int getMaxSize(int clusterSize, MaxSizeConfig maxSizeConfig) {
```
ahmetmircik: is it possible that this method can return a negative value? For example, in case of policy is not `MaxSizePolicy.PER_NODE`.

ajermakovics: Yes, it can return a negative value. In that case keys will not be limited inside the `sendKeysInBatches` method.

ahmetmircik: So if i have a, for example, max size PER_PARTITION, is it possible to get OOM exception?

ahmetmircik: Maybe OOM is not a concern since you are doing batching but it seems load times may increase when using PER_PARTITION max size policy due to that negative return value.

ajermakovics: Why would load time increase? The negative value won't have any additional effect

        if (maxSize > 0) {
            dataKeys = limit(dataKeys, maxSize);

ahmetmircik: i am assuming if `maxSize < 0`, we will not limit the number of keys to be loaded on owner partitions.

ajermakovics: correct

gurbuzali:
needs rebase

ajermakovics:
PR #5068 contains all these changes but came later.

PR #4950, opened 2015-03-30
Expose isClusterSafe and isLocalMemberSafe through MBeans

henri-tremblay:
Adding those two attributes will be really helpful to know the state of the cluster. Using the PartitionService from HazelcastClient throws an UnsupportedOperationException.

Using an MBean seems to be the easiest way.

I've exposed it as an attribute. I don't know if getting these information is costly. If it is, you might want to expose an operation instead.


mesutcelik:
verify

mesutcelik:
Can you please sign contibutor agrement?

henri-tremblay:
Sure... Dumb me... made the paperwork but never sent it. Now sent.

mesutcelik:
Thanks @henri-tremblay for the contribution.

@Serdaro ,
I have added new attributes into documentation. Do we need some other update?
e623287

Issue #4790, opened 2015-03-11
unpredictable warning when initializing with Config    


bwzhang2011:
when upgraded to hz3.5 snapshot for testing, my project has got some warning message like this:
No configuration found for hazelcastmq.txn-topic, using default config!
while tracing, I noticed that such originated from Config class for the method of  'private <T> T lookupByPattern(Map<String, T> configPatterns, String itemName)' at line 901.

But I could not know that why it would induce the warning message (both would be ok under 3.4.x version).
so I dare say it should do some enhancement.

pveentjer:
It is a warning that no concrete configuration was found, so the system is using a default. This log message should not have been logged under the warning level imho but on info or even debug. This 'warning' is new since HZ 3.5

bwzhang2011:
@pveentjer, thanks for review. I set my hazelcast cluster configuration through hazelcast-spring but it is ok since it swithced to hazelcast 3.4+ but the warning as you described is in hz3.5. and I don't quite understand why concrete configuration is needed for wild card is support in hazelcast configuration

pveentjer:
I think you can ignore it safely. Imho warning is not a warning, but just a debugging message.

Donnerbart:
Issue is resolved and will not be part of the 3.5 release.

PR #4959, opened 2015-03-31
Reduce log level of "No configuration found for ..., using default config!"

Donnerbart:
Set log level of "No configuration found for ..., using default config!" from `WARNING` to `FINEST`.

Fixes issue #4790.


Donnerbart:
verify

pveentjer:
Tail failure unrelated to this logging change. So therefore I'll merge it.

PR #4962, opened 2015-03-31
Fix for Issue 4960 [Cache Config Is Not Created on Cluster If Executer of `CacheCreateConfigOperation` Has Already Cache Config]

### LINKED ISSUE #4960: Cache Config Is Not Created on Cluster If Executer of `CacheCreateConfigOperation` Has Already Cache Config
Currently, when a client gets a cache that is not created yet, it gets its config from the node that owner of cache. After gets the config, it sends a request (`CacheCreateConfigRequest`) to this node create cache config also on other nodes. But at current code, since that cache config is already created at that node (since its config has already been taken from it via `CacheGetConfigRequest`), it doesn't create also on other nodes as seen here https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheCreateConfigOperation.java#L89

serkan-ozal:
Fixes #4960


pveentjer:
There are a bunch of test failures:

    com.hazelcast.client.cache.ClientCacheConfigTest.cacheManagerByLocationFileTest
    com.hazelcast.client.cache.ClientCacheConfigTest.cacheManagerByInstanceNameTest
    com.hazelcast.client.cache.ClientCacheConfigTest.testGetPreConfiguredCache
    com.hazelcast.client.cache.ClientCacheConfigTest.cacheManagerByLocationClasspathTest
    com.hazelcast.client.cache.ClientCacheConfigTest.createCacheConfigOnAllNodes

pveentjer:
Can you squash your commits to create a clean commit history.

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/ClientCacheConfigTest.java
```diff
@@ -170,4 +168,39 @@ public void testGetPreConfiguredCache() {
...
+        assertNotNull(cache);
+
+        // First attempt to use the cache will trigger to create its record store
+        for (int i = 0; i < 10; i++) {
+            cache.put(String.valueOf(i), String.valueOf(i));
+        }
+
+        client.shutdown();
+
+        server1.shutdown();
```
pveentjer: Where do the client en server need to be shut down? They are normally shut down when the test completes.

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/ClientCacheConfigTest.java
```diff
@@ -170,4 +168,39 @@ public void testGetPreConfiguredCache() {
...
+        // Create the client cache manager
+        HazelcastClientCachingProvider cachingProvider = HazelcastClientCachingProvider.createCachingProvider(client);
+        CacheManager cacheManager = cachingProvider.getCacheManager();
+
+        Cache<String, String> cache = cacheManager.getCache(CACHE_NAME);
+
+        assertNotNull(cache);
+
+        // First attempt to use the cache will trigger to create its record store
+        for (int i = 0; i < 10; i++) {
```
pveentjer: This test looks strange to me. You do some logic, but there is not a single assert to validate that the logic actually did something.

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheCreateConfigOperation.java
```diff
@@ -86,7 +86,7 @@ public void run() throws Exception {
         if (!ignoreLocal) {
             response = service.createCacheConfigIfAbsent(config);
         }
-        if (createAlsoOnOthers && response == null) {
+        if (createAlsoOnOthers) {
```
pveentjer: Can you simplify this code. It contains a lot of nested control structures. This can easily be fixed by pulling out some methods.

pveentjer: Example cleanup:

```
 @Override
        AbstractCacheService service = getService();
        if (!ignoreLocal) {
            response = service.createCacheConfigIfAbsent(config);
        }
        
        if(!createAlsoOnOthers){
            return;
        }

        NodeEngine nodeEngine = getNodeEngine();
        Collection<MemberImpl> members = nodeEngine.getClusterService().getMemberList();
        int remoteNodeCount = members.size() - 1;
        if (remoteNodeCount <= 0) {
            return;
        }
        
        postponeReturnResponse();
        Callback<Object> callback = new CacheConfigCreateCallback(getResponseHandler(), remoteNodeCount);
        for (MemberImpl member : members) {
            invokeOnMember(callback, operationService, member);
        }
    }

    private void invokeOnMember(Callback<Object> callback, OperationService operationService, MemberImpl member) {
        if (!member.localMember()) {
            CacheCreateConfigOperation op = new CacheCreateConfigOperation(config, false);
            operationService
                    .createInvocationBuilder(AbstractCacheService.SERVICE_NAME, op, member.getAddress())
                    .setCallback(callback)
                    .invoke();
        }
    }
```

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/ClientCacheConfigTest.java
serkan-ozal: right, they need to be shutdown at all cases even failure case

serkan-ozal:
I have intentionally splitted them into multiple commits since changes are different (such as one for fix and another for test). So merging them into a single commit doesn't make sense to me even they are related with same issue

serkan-ozal:
btw, will look test failures.

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/ClientCacheConfigTest.java
serkan-ozal: there is no need to assert. in this test, I expect no exception after first cache operation since cache record stores are created as lazy. But ten times may be unnecessary

serkan-ozal:
weird logs from the test failure from `hazelcast-client-new`

@sancar maybe you can look also to logs

```
19:08:09,486  INFO [TcpIpConnection] hz._hzInstance_1_cluster1.IO.thread-in-0 - [127.0.0.1]:5701 [cluster1] [3.5-SNAPSHOT] Connection [localhost.localdomain/127.0.0.1:51333] lost. Reason: java.io.EOFException[Remote socket closed!]
19:08:09,486  WARN [ClientConnection] main - Connection [/127.0.0.1:5701] lost. Reason: java.lang.IndexOutOfBoundsException[index=8, length=4, capacity=0]
19:08:09,486  WARN [ReadHandler] hz._hzInstance_1_cluster1.IO.thread-in-0 - [127.0.0.1]:5701 [cluster1] [3.5-SNAPSHOT] hz._hzInstance_1_cluster1.IO.thread-in-0 Closing socket to endpoint null, Cause:java.io.EOFException: Remote socket closed!
19:08:09,487  WARN [ClusterListenerSupport] main - Unable to get alive cluster connection, try in 2975 ms later, attempt 1 of 2.
19:08:12,463  INFO [SocketAcceptor] hz._hzInstance_1_cluster1.IO.thread-Acceptor - [127.0.0.1]:5701 [cluster1] [3.5-SNAPSHOT] Accepting socket connection from /127.0.0.1:51355
19:08:12,464  INFO [TcpIpConnectionManager] cached3 - [127.0.0.1]:5701 [cluster1] [3.5-SNAPSHOT] Established socket connection between /127.0.0.1:5701 and localhost.localdomain/127.0.0.1:51355
19:08:12,465  WARN [ClientConnection] main - Connection [/127.0.0.1:5701] lost. Reason: java.lang.IndexOutOfBoundsException[index=8, length=4, capacity=0]
19:08:12,465  INFO [TcpIpConnection] hz._hzInstance_1_cluster1.IO.thread-in-1 - [127.0.0.1]:5701 [cluster1] [3.5-SNAPSHOT] Connection [localhost.localdomain/127.0.0.1:51355] lost. Reason: java.io.EOFException[Remote socket closed!]
19:08:12,465  WARN [ReadHandler] hz._hzInstance_1_cluster1.IO.thread-in-1 - [127.0.0.1]:5701 [cluster1] [3.5-SNAPSHOT] hz._hzInstance_1_cluster1.IO.thread-in-1 Closing socket to endpoint null, Cause:java.io.EOFException: Remote socket closed!
19:08:12,465  WARN [ClusterListenerSupport] main - Unable to get alive cluster connection, try in 2997 ms later, attempt 2 of 2.
```

serkan-ozal:
verify

serkan-ozal:
just saw it. they had been marked as ignored before. so I will add test but reverted back as ignored

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/ClientCacheConfigTest.java
pveentjer: Yes. That is done for you.

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/ClientCacheConfigTest.java
pveentjer: I think we should do better validation that the caches are really created. You can create an operator/runnable and check if the caches exist. 

```
class CheckCacheExistOperator extends AbstractOperation{
	String cachename;

	void run(){
		CacheService service = ...;	
		result = service.exist(cacheName)	
	}
}
```

And then you do something like this:

```
void assertCacheExistsOnAllNoded(String cacheName){
	for(Member member: cluster.getMembers()){
		Future<Boolean> future = operationService.invokeOnMember(null, new CheckCacheExistsOperator(cacheName), getAddress(member));
		assertTrue(future.get());
	}
}
```

And you test would look like this

```
@Test
public void createCacheConfigOnAllNodes() {
	final String CACHE_NAME = "myCache";
	Config config = new Config();
	CacheSimpleConfig cacheConfig = new CacheSimpleConfig().setName(CACHE_NAME);
	config.addCacheConfig(cacheConfig);
	// Create servers with configured caches
	HazelcastInstance server1 = Hazelcast.newHazelcastInstance(config);
	HazelcastInstance server2 = Hazelcast.newHazelcastInstance(config);
	// Create the hazelcast client instance
	// Create the client cache manager
	HazelcastClientCachingProvider cachingProvider = HazelcastClientCachingProvider.createCachingProvider(client);
	CacheManager cacheManager = cachingProvider.getCacheManager();
	Cache<String, String> cache = cacheManager.getCache(CACHE_NAME);
	
	assertNotNull(cache);
	assertCacheExistsOnAllNoded(cacheName);
}
```

This way you have really checked that your behaviour works correctly. The current tests doesn't test anything since the cacheManager.getCache(CACHE_NAME) could also fail to create the cache on all members without this test detecting it.

serkan-ozal: @pveentjer 

```
This way you have really checked that your behaviour works correctly. The current test doesn't test anything since the cacheManager.getCache(CACHE_NAME) could also fail to create the cache on all members without this test detecting it.
```

This test also checks ```cacheManager.getCache(CACHE_NAME)``` since cache configs are created on nodes as lazy when client triggers this call.

Please be sure that you understand the behaviour of the test before making comment or contact me

pveentjer: @serkan-ozal 

if I see a test with questionable quality than I'll add that in a github review comment. That is the purpose of a code review.

serkan-ozal: @pveentjer right, but as long as you understand the behaviour correctly. You can ask the behaviour of the test over here if you see it questionable before saying "it does not testing anything".

pveentjer: I understand the behaviour. The test is called 'createCacheConfigOnAllNodes'. 

I think it makes sense to validate explicitly that the config has been created on all nodes using an operation. This way you don't rely on side effects to see your system really behaves like expected. Please have a look at the test example I provided; isn't that very explicit?

mdogan:
verify

gurbuzali:
verify

mdogan:
:+1:

kobalski:
:+1:

PR #4963, opened 2015-03-31
Backport fix for Issue 4960 [Cache Config Is Not Created on Cluster If Executer of `CacheCreateConfigOperation` Has Already Cache Config] 

### LINKED ISSUE #4960: Cache Config Is Not Created on Cluster If Executer of `CacheCreateConfigOperation` Has Already Cache Config

serkan-ozal:
Fixes #4960 

Backport of https://github.com/hazelcast/hazelcast/pull/4962


gurbuzali:
verify

mdogan:
:+1:

kobalski:
:+1:

PR #4970, opened 2015-04-01
Client Protocol serverside base classes

asimarslan:
ClientMessage improvement and Parameters base
Authentication impls
MessageTask Factory


noctarius:
There is one thing from having a quick look over it that hurts my eyes: I see a lot of public fields, do we really need them? :D

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/AuthenticationBaseMessageTask.java
```diff
@@ -0,0 +1,95 @@
...
+import com.hazelcast.util.UuidUtil;
+
+import java.security.Permission;
+import java.util.Collection;
+import java.util.Set;
+
+/**
+ * Base authentication task
+ */
+public abstract class AuthenticationBaseMessageTask<CM extends ClientMessage>
```
noctarius: Wouldn't it be more like BaseAuthenticationMessageTask or AbstractAuthenticationMessageTask?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ClientMessageBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+            return;
+        }
+
+        //MESSAGE IS COMPLETE HERE
+        final byte flags = (byte) message.getFlags();
+
+        if ((flags & BEGIN_AND_END_FLAGS) == BEGIN_AND_END_FLAGS) {
+            //HANDLE-MESSAGE
+            ClientMessage cm = new ClientMessage();
+            cm.wrapForDecode(cm.buffer().byteBuffer(), 0);
```
noctarius: What is this buffer? Where is it coming from? Maybe it just misses a comment but it doesn't look to me as it make sense :)

asimarslan:
@noctarius ,
Classes with name XParameters are for message encode decode. We will have one for each old ClientRequests and they are gone be code generated from a list of parameters. The ones here are for quick placeholders until we generate them. I may remove the public fields in generated code . I don't want to start a discussion about "getter setters being evil"

asimarslan:
verify

pveentjer:
I agree with Asim. In some cases, like dto's, there is nothing wrong with public fields. Using a getter/setter gives the impression of an object.. but it isn't..  so why give the impression at all.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/Flyweight.java
```diff
@@ -97,4 +83,22 @@ protected void uint32Put(final int offset, final long value, final ByteOrder byt
...
+            newBuffer.put(buffer.byteBuffer());
+            buffer.wrap(newBuffer);
+        }
+    }
+    private static int findSuitableCapacity(int capacity, final int requiredCapacity) {
+        do {
+            capacity <<= 1;
+        } while (capacity < requiredCapacity);
+
+        return capacity;
```
ahmetmircik: if we do not put the line with stars below, we always shift the capacity even it is not necessary. Is this intentional?

```
private static int findSuitableCapacity(int capacity, final int requiredCapacity) {
        do {
            if(capacity > requiredCapacity) { ******************** 
                break;
            }

            capacity <<= 1;
        } while (capacity < requiredCapacity);

        return capacity;
    }
```

mdogan: @ahmetmircik, see method `ensureCapacity()`, it checks for the capacity first and then calls `findSuitableCapacity()` if required.

```java
if (requiredCapacity > capacity) {
   final int newCapacity = findSuitableCapacity(capacity, requiredCapacity);
   ....
}
```

ahmetmircik: i see thanks, but it makes code fragile and it may also fail under unforeseen conditions.

checkstyle/suppressions.xml
```diff
@@ -190,6 +190,8 @@
     <suppress checks="" files="com.hazelcast.client.impl.protocol.util[\\/]"/>
```
mdogan: Why all `protocol` packages are suppressed? I think it's easier to make them checkstyle compatible during development instead of making them at the end in bulk. Even you can forget to remove this suppressions and they'll never get checked until someone notices. 

Writing this comment because I saw some checkstyle issues in the code and noticed that they all suppressed.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/map/MapMessageType.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Map message types
+ */
+public enum MapMessageType {
+
+    /**
+     * Map put
+     */
+    MAP_PUT(9),
```
ahmetmircik: Can i assume some enum instances with `id` smaller than `9` ? If that is the case, it seems this `MapMessageType` class is following an id-sequence but this class has no information/relation of those classes which have previous id-sequences. This is increasing fragility.

asimarslan: MessageType id is (will be) documented in ClientProtocol . So those ids shouldn't be modified. As there will be around 500 messages I'm planing to organise these by distributed object types like map.
Fragility shouldn't  be an issue as they are validated against our client protocol.

PR #4980, opened 2015-04-02
removed compile time dependency with hazelcast-hibernate module

mesutcelik:
* upgraded latest hibernate versions
* removed compile time dependency with hazelcast-hibernate module


mdogan:
Currently, Spring module depends on Hibernate only in test scope. Should we add a provided/optional Hibernate dependency too, to make it visible in dependency graph?

mesutcelik:
Actually, I just removed it because we don't need hazelcast-hibernateX module at compile time any more. I don't really know what can be added-value in terms of tooling or dependency graph if i put it back. 

As far as I see from the link below, hazelcast-hibernateX won't be visible in the dependency graph of projects which depend on hazelcast-spring. That is how optional tag works.

check the section "How do optional dependencies work?"
https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

PR #4982, opened 2015-04-02
[CP] ClientMessage  and parameter encode-decode improved, unit tests added

asimarslan:
improves and add unit test for #4970 
Authentication test
util tests
Checkstyle suppressions removed with exception to external utils and parameter dto classes


asimarslan:
@sancar,  authentication works see RawProtocolTest

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ClientMessageAccumulator.java
```diff
@@ -0,0 +1,63 @@
+package com.hazelcast.client.impl.protocol.util;
+
+import com.hazelcast.client.impl.protocol.ClientMessage;
+
+import java.nio.ByteBuffer;
+
+/**
+ * ClientMessage builder from incoming bytebuffer parts
+ */
+public class ClientMessageAccumulator
```
sancar: This one extending ClientMessage seems odd. I think ClientMessageAccumulator is not a CLientMessage. Should we rather pass ClientMessage as parameter or get as return value?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ClientMessageAccumulator.java
```diff
@@ -0,0 +1,63 @@
...
+            return readLength;
+        }
+        return 0;
+    }
+
+    public boolean isComplete() {
+        return (index() > HEADER_SIZE) && (index() == getFrameLength());
+    }
+
+    public int remainingSize() {
```
sancar: Should this be public?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/Flyweight.java
```diff
@@ -28,12 +28,12 @@
  */
 public class Flyweight {
 
-    private static final int INITIAL_BUFFER_CAPACITY = 4096;
-    private static final ByteBuffer INITIAL_BUFFER = ByteBuffer.allocate(INITIAL_BUFFER_CAPACITY);
-
-    protected final MutableDirectBuffer buffer = new UnsafeBuffer(INITIAL_BUFFER);
+    public static final int INITIAL_BUFFER_CAPACITY = 4096;
```
sancar: Can this field be protected?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/AbstractCallableMessageTask.java
```diff
@@ -31,9 +31,16 @@ protected AbstractCallableMessageTask(ClientMessage clientMessage, Node node, Co
 
     @Override
     public final void processMessage() {
-        Object result = call();
-        endpoint.sendResponse(result, parameters.getCorrelationId());
+        try {
+            final ClientMessage result = call();
+            endpoint.sendClientMessage(result);
```
sancar: Who sets the correlation id of this client message? How about if we make sendClientMessage getting a correlation id as parameter, and "sendClientMessage" sets the value inside the method so that we do not accidentally forget to set the correlation id.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/AbstractCallableMessageTask.java
```diff
@@ -22,8 +22,8 @@
 /**
  * Base callable Message task.
  */
-public abstract class AbstractCallableMessageTask<CM extends ClientMessage>
-        extends AbstractMessageTask<CM> {
+public abstract class AbstractCallableMessageTask<P>
+        extends AbstractMessageTask<P> {
```
sancar: Can we make Parameters extending a common interface so that we use that interface everywhere instead of generics? Because this generic is not guarding us against anything in compile time.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ClientMessageAccumulator.java
asimarslan: ClientMessageAccumulator is a ClientMessage with accumulation feature. Maybe we can update its name.

sancar:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/Flyweight.java
asimarslan: No it is a shared constant that can be used to initialize flyweight

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/AbstractCallableMessageTask.java
asimarslan: Parameters are used by task implementers and they have to implement some abstract methods that are specific to that parameter impl so the actual guarding is in the task implementation. If we need to group these Parameters for some other reason we may define an interface in future. 

Side note: These Parameters actually STRUCT's but we don't have them in java yet. They are not actual objects but dummy parameter wrappers .

PR #4431, opened 2015-01-18
NEAR CACHE IMPL

serkan-ozal:
This PR contains (will contain) Near Cache infrastructure for general usage instead of only client.
Also, this PR doesn't contain (won't contain) eviction implementation for Near Cache since eviction support will be handled another PR to minimize scope of this PR for easier review.

PR #4984, opened 2015-04-03
Near Cache Implementation

serkan-ozal:
This PR contains Near Cache infrastructure for general usage instead of only client.
Also, this PR doesn't contain eviction implementation for Near Cache since eviction support will be handled another PR.

**NOTE:** All commits merged version of PR https://github.com/hazelcast/hazelcast/pull/4431


hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -89,22 +113,68 @@
...
+            nearCache = nearCacheManager
+                    .getOrCreateNearCache(nameWithPrefix, nearCacheConfig, nearCacheContext);
+            registerInvalidationListener();
         } else {
             nearCache = null;
             cacheOnUpdate = false;
         }
     }
 
+    protected class ClientNearCacheExecutor implements Executor {
```
gurbuzali: instead of creating a wrapper class you could've just extend `ClientExecutionService` with `Executor`
it has the the method (`void execute(Runnable command)`) already

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/NearCacheContext.java
```diff
@@ -0,0 +1,60 @@
...
+    public SerializationService getSerializationService() {
+        return serializationService;
+    }
+
+    public NearCacheContext setSerializationService(SerializationService serializationService) {
+        this.serializationService = serializationService;
+        return this;
+    }
+
+    public Executor getNearCacheExecutor() {
```
gurbuzali: this executor is never used, will it be used in enterprise ?
if not we can get rid of `NearCacheContext` and just use `SerializationService`

serkan-ozal: yes, it will be used for expiration task while implementing eviction. but not used right now

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
mdogan: `ClientNearCacheExecutor ` can be `private static`.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -89,22 +113,68 @@
...
+        }
+
+        @Override
+        public void execute(Runnable runnable) {
+            clientExecutionService.execute(runnable);
+        }
+
+    }
+
+    protected Executor createNearCacheExecutor(ClientExecutionService clientExecutionService) {
```
mdogan: this method can be `private static`.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -89,22 +113,68 @@
...
+        return new ClientNearCacheExecutor(clientExecutionService);
+    }
+
+    @Override
+    public void close() {
+        if (nearCache != null) {
+            removeInvalidationListener();
+            nearCacheManager.clearNearCache(nearCache.getName());
```
mdogan: why not just `nearCache.clear()`? It seems `NearCacheManager.clearNearCache()` is redundant, doing nothing extra.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -80,7 +95,16 @@
     private static final long MAX_COMPLETION_LATCH_WAIT_TIME = TimeUnit.MINUTES.toMillis(5);
     private static final long COMPLETION_LATCH_WAIT_TIME_STEP = TimeUnit.SECONDS.toMillis(1);
 
-    protected final ClientNearCache<Data, Object> nearCache;
+    protected final ILogger logger = Logger.getLogger(getClass());
+
+    protected final HazelcastClientCacheManager cacheManager;
+    protected final NearCacheManager nearCacheManager;
+    // Object => Data or <V>
+    protected NearCache<Data, Object> nearCache;
```
mdogan: `nearCache` field can be `final`. I guess there's no need to null it when destroyed. Otherwise it should be marked as `volatile`.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
serkan-ozal: currently, there is no specific stuff on ```NearCacheManager``` for clearing cache. But I prefer giving responsibility for such kind of operations like clear or destroy to manager

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -464,4 +529,124 @@ protected ICompletableFuture createCompletedFuture(Object value) {
...
+            removeInvalidationListener(member, false);
+        }
+
+        public void memberAttributeChanged(MemberAttributeEvent memberAttributeEvent) {
+
+        }
+
+    }
+
+    protected final class NearCacheInvalidationHandler implements EventHandler<CacheInvalidationMessage> {
```
mdogan: cna be `private` instead of `protected final`.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -464,4 +529,124 @@ protected ICompletableFuture createCompletedFuture(Object value) {
                 clientContext.getExecutionService().getAsyncExecutor());
     }
 
+    protected final class NearCacheMembershipListener implements MembershipListener {
```
mdogan: can be private instead of protected final.

mdogan:
Added some minor comments.

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheDataRecordStore.java
```diff
@@ -0,0 +1,160 @@
...
+                // Give priority to Data typed candidate.
+                // So there will be no extra convertion from Object to Data.
+                if (candidate instanceof Data) {
+                    selectedCandidate = candidate;
+                    break;
+                }
+            }
+            if (selectedCandidate != null) {
+                return selectedCandidate;
+            } else {
```
gurbuzali: `else` is redundant

hazelcast/src/main/java/com/hazelcast/cache/impl/CachePortableHook.java
```diff
@@ -148,27 +151,26 @@ public Portable createNew(Integer arg) {
...
+                    public Portable createNew(Integer arg) {
+                        return new CacheInvalidationMessage();
+                    }
+                };
+                constructors[REMOVE_INVALIDATION_LISTENER] = new ConstructorFunction<Integer, Portable>() {
+                    public Portable createNew(Integer arg) {
+                        return new CacheRemoveInvalidationListenerRequest();
+                    }
+                };
+                //  constructors[SEND_STATS] = new ConstructorFunction<Integer, Portable>() {
```
gurbuzali: leftover ?

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheObjectRecordStore.java
```diff
@@ -0,0 +1,140 @@
...
+                // Give priority to non Data typed candidate.
+                // So there will be no extra convertion from Data to Object.
+                if (!(candidate instanceof Data)) {
+                    selectedCandidate = candidate;
+                    break;
+                }
+            }
+            if (selectedCandidate != null) {
+                return selectedCandidate;
+            } else {
```
gurbuzali: same here no need for `else`

hazelcast/src/main/java/com/hazelcast/cache/impl/CachePortableHook.java
serkan-ozal: yes, will remove it

PR #4985, opened 2015-04-03
[Possible fix of #4911] Assert local executor stats eventually

### LINKED ISSUE #4911: Test of getLocalExecutorStats() is failing spuriously
The commit bd9bb627bb9dceae1c2971ea0e4f0abae097d893 has commented out a test of `getLocalExecutorStats()` because it is failing spuriously. After fixing this issue the test should be enabled.

This is probably related to issue #4677.

metanet:
* As far as I understand, this issue is not related to #4677, since the test fails before asserting the task cancellation. The cause of the rare fails is scheduling the task to another thread. I have run the test 100K times and it hasn't failed.

Fixes #4911

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/executor/SingleNodeTest.java
@@ -20,6 +20,7 @@ import com.hazelcast.core.IExecutorService;
 import com.hazelcast.core.Member;
 import com.hazelcast.core.MemberSelector;
 import com.hazelcast.monitor.LocalExecutorStats;
+import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.annotation.QuickTest;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


serkan-ozal:
:+1:

mdogan:
> The cause of the rare fails is scheduling the task to another thread.

I did not understand, which task is scheduled to another thread?

metanet:
SingleNodeTest:347 submits the task to another thread. Therefore, SingleNodeTest:338 fails occasionally.

mdogan:
Still, not understand. 

First `k` executions are guaranteed to be completed, after `assertOpenEventually(LatchRunnable.latch)`.
```java
final LatchRunnable r = new LatchRunnable();
        for (int i = 0; i < k; i++) {
            executor.execute(r);
        }
assertOpenEventually(LatchRunnable.latch);
```
And `executor.submit(new SleepingTask(10))` is guaranteed to be completed after `f.get();`. 

So, where is the race window here?

metanet:
Sorry for not telling it clearly. 

As you say, first k operations are guaranteed to be executed. 
The next task is the SleepingTask which is k + 1 and it is submitted and eventually started. After that, cancel() invocation cancels it before completion. Therefore, f.get() throws a CancellationException.

mdogan:
ok, I got your point now. thanks.

mdogan:
:+1:

PR #4992, opened 2015-04-05
PortableTest factory and classIds gathered in one place. Test added for ...

sancar:
...rawDataInput/Output read/writeObject methods . All three serialization methods are tested(Portable,IdentifiedDataSerializable and CustomSerializable)


mdogan:
:+1:

PR #4997, opened 2015-04-06
if invocation is asyc do not try to run it on calling thread

gurbuzali:
running async invocation on calling thread actually makes it sync


jerrinot:
:+1: 


A side-note: I think `Invocation` should be refactored. All the methods `doInvoke()`, `invokeInternal()`, `invokeXY()` are quite confusing.

gurbuzali:
I agree, afaik it will be refactored with the black-box design
@pveentjer knows better

pveentjer:
The thing I have in mind is that we expose on the API what kind of behavior we want to have on the invocation. For example if I do an async call, I'm totally unhappy if the system is going to run my operation on my thread. Also we need to know how many threads are going to get the future. 99% of the invocations, only a single thread will get the future and call get. If a caller runs optimization is applied, then there is no need to do the lock and call notifyAll.

The blackbox is purely to store all kinds of statistics/events of the system. It won't affect the invocation API.

pveentjer:
verify

gurbuzali:
verify

gurbuzali:
verify

PR #4999, opened 2015-04-06
Disabling replication in QueryIndexingTest

ajermakovics:
Fix for test failure #4114 
It appears there is a race condition between:
 - map is filled
 - assert cluster is safe (no replication going on)
 - test starts to run queries
 - replication starts and queries don't use index.
 
This change disables replication since its not used in this test.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/query/QueryIndexingTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/query/QueryIndexingTest.java
@@ -4,6 +4,7 @@ import com.hazelcast.config.Config;
 import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
+import com.hazelcast.instance.GroupProperties;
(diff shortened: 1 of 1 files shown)


ajermakovics:
Was failing on maintenance as well so will backport

ajermakovics:
verify

mesutcelik:
verify

mesutcelik:
verify

mesutcelik:
@gurbuzali ,
#4114 is failing very often. Can you have a look at this PR?

gurbuzali:
backport ?

jerrinot:
I'm having similar issues with `MapLoaderTest.testMapLoaderLoadUpdatingIndex()` - the test is sometimes failing on Jenkins. 

I can reproduce it on my local by introducing artificial sleeping on certain places. When I disable replication then the test is always passing.

ajermakovics:
@jerrinot If that's the case then more index tests could be affected. Disabling replication in all of them is an option but a more general solution would be better. Not sure what it is yet

PR #5009, opened 2015-04-06
BackPressureRegulator simplification

pveentjer:
No threadsafety is required for the sync delays since they will only be accessed by partition threads and at any given moment only 1 partition thread will touch a sync delay.

There is some additional cleanup:
- no more logging on warning if a backup replica doesn't find an invocation. This is something which can happen + it caused huge log files in the nightly build with the back pressure stress test
- the invocation registry will log the number of operations timeout or backup timeout it has found on every run (if something is found.. else no logging)
- sync window has 1 as lowest value.. meaning on every run.
- minor cleanup in invocation to be checkstyle complient
- solved the async invocation async backup nightly failure. Caused by too small backup timeout

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
@@ -6,8 +6,10 @@ import com.hazelcast.spi.BackupAwareOperation;
 import com.hazelcast.spi.Operation;
 
 import java.util.Random;
-import java.util.concurrent.atomic.AtomicInteger;
 
+import static com.hazelcast.nio.Bits.CACHE_LINE_LENGTH;
+import static com.hazelcast.nio.Bits.INT_SIZE_IN_BYTES;
+import static java.lang.Math.max;
 import static java.lang.Math.round;
 
 
@@ -52,8 +54,13 @@ public class BackpressureRegulator {
         }
     };
 
-    // we have a counter per partition + we have an additional counter for generic operations.
-    private final AtomicInteger[] syncDelays;
+    // number of ints in a single cache-line.
+    private static final int INTS_PER_CACHE_LINE = CACHE_LINE_LENGTH / INT_SIZE_IN_BYTES;
+
+    // There is syncDelay counter per partition. To prevent false sharing, the syncDelay for each partition is padded additional
+    // int's so that each sync delay is on its own cache-line.
+    // Each partition will always be handled by 1 thread at any given moment, so thread-safety is not required.
+    private final int[] syncDelays;
 
     private final boolean enabled;
     private final boolean disabled;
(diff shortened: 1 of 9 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
```diff
@@ -6,8 +6,9 @@
 import com.hazelcast.spi.Operation;
 
 import java.util.Random;
-import java.util.concurrent.atomic.AtomicInteger;
 
+import static com.hazelcast.client.impl.protocol.util.BitUtil.CACHE_LINE_LENGTH;
```
mdogan: Importing something from `com.hazelcast.client.impl.protocol....` seems not right to me. Even I find some methods and constants in `com.hazelcast.client.impl.protocol.util.BitUtil` redundant, we already have `com.hazelcast.nio.Bits`.

pveentjer: True.. I wanted to move it..but then you get a big code change. I want to move it to com.hazelcast.utiil.

And not everything is found. Personally I would like to merge all these BitUtil classes into a single class.

mdogan: We already have two distinct classes; `com.hazelcast.nio.Bits` for bit/byte/primitive manipulation and `com.hazelcast.util.QuickMath` for shorthand math calculations. I think we should request client team to remove `com.hazelcast.client.impl.protocol.util.BitUtil` and add required methods/constants to these two classes.

pveentjer: Sure. But should that be part of this PR?

mdogan: I think you can put `CACHE_LINE_LENGTH` into `com.hazelcast.nio.Bits` (leaving client one as is). It has already a field for integer size, `INT_SIZE_IN_BYTES`. I'll request client team to do additional changes.

pveentjer: I'll do that.

asimarslan: Guys,
Client Protocol is being implemented as partial PR's and we will make the code clean up including  merging the redundant utils to hazelcast util as a last step of this process so please do not use anything from protocol classes until we finalise them.

pveentjer: I already removed the dependency. 

Can you make sure that after your cleanup, there is no duplication?

mdogan:
looks good.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -289,25 +295,40 @@ private void scanHandleOperationTimeout() {
...
                 } catch (Throwable t) {
                     inspectOutputMemoryError(t);
                     logger.severe("Failed to handle backup timeout of invocation:" + invocation, t);
                 }
             }
+
+            log(backupTimeouts, invocationTimeouts);
+        }
+
+        private void log(int backupTimeouts, int invocationTimeouts) {
```
mdogan: This will be logged at most once a second, right?

pveentjer: Yes. And only when it has something sensible to say.. so if nothing got timeout --> nologging

pveentjer:
verify

pveentjer:
verify

pveentjer:
verify

pveentjer:
verify

pveentjer:
verify

mesutcelik:
Can we merge that one? @pveentjer @gurbuzali @mdogan 
This is required to make nightly-build pass.

gurbuzali:
:+1:

PR #5012, opened 2015-04-07
allows setting zero to connection-attempt-limit, fixes #4967

### LINKED ISSUE #4967: Hazelcast spring client connection-attempt-limit does not allow zero
The hazelcast client code allows to use the number zero for connectionAttemptLimit which internally results in int.maxValue.
Problem is the XSD of the hazelcast spring config requires it to be at least 1, so you cannot set it to infinite from spring config.
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-spring/src/main/resources/hazelcast-spring-3.4.xsd#L1611

mesutcelik:
hazelcast-client,hazelcast-spring and hazelcast-client-new have been fixed all together.


bilalyasar:
:+1: seems ok to me.

PR #5013, opened 2015-04-07
Distributed Cache.load optimisation

ajermakovics:
Now filtering keys to send to each node for loading. Each node should only receive keys that it owns. Previously all keys were sent to every node.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxyBase.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxyBase.java
@@ -233,13 +233,13 @@ abstract class AbstractCacheProxyBase<K, V> {
                 OperationService operationService = getNodeEngine().getOperationService();
                 OperationFactory operationFactory;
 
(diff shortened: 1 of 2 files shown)


serkan-ozal:
@ajermakovics While in `run` (in `for` loop) method of `LoadAllTask`, doesn't it cause some keys are not sent to any partition if there is a partition re-assignment at this time ?

ajermakovics:
Good question. `PartitionService.getMemberPartitionsMap` returns a copy of the partition table so that won't change. If partition assignments change while running that method then we might have a problem.  Can they change? Not sure yet...
On the other hand OperationService uses the same method when doing invokeOnAllPartitions.  
verify

serkan-ozal:
Yes, `OperationService` uses similar way. In this case, `CacheLoadAll` operation (and also as well as for other operation in `OperationService`) eventually is sent to right node (partition), but I think it is possible with wrong keys data in this case.

gurbuzali:
verify

ajermakovics:
verify

PR #5020, opened 2015-04-08
MessageListener concurrency documentation

### LINKED ISSUE #5019: [DOCUMENTATION] ITopic MessageListener concurrency
Currently no documentation is available for the need for concurrency control on MessageListener.

MessageListeners don't need to be threadsafe because they will be called by at most 1 thread at any given moment.

Also we need to document the memory consistency effects; the user should rely that we provide a happens before relation and that he doesn't need to make things synchronized or volatile for the sake of safe publication.

pveentjer:
Added section MessageListener regarding threadsafety and memory consistency effects

Fixes #5019

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/MessageListener.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/MessageListener.java
@@ -21,6 +21,11 @@ import java.util.EventListener;
 /**
  * Message listener for {@link ITopic}.
  *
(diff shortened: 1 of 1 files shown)


Donnerbart:
verify

pveentjer:
Since it is just a javadoc change I'll merge this one.

Serdaro:
ok.

Issue #4915, opened 2015-03-27
[TEST-FAILURE] SingleNodeTest.cancelWhileQueued


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8/com.hazelcast$hazelcast/511/testReport/junit/com.hazelcast.executor/SingleNodeTest/cancelWhileQueued/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK9-Quality-Outreach/17/com.hazelcast$hazelcast/testReport/com.hazelcast.executor/SingleNodeTest/cancelWhileQueued/

muratayan:
I am working on this.

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK9-Quality-Outreach/com.hazelcast$hazelcast/33/testReport/junit/com.hazelcast.executor/SingleNodeTest/cancelWhileQueued/

gurbuzali:
closed via https://github.com/hazelcast/hazelcast/pull/5022

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/com.hazelcast$hazelcast/405/testReport/junit/com.hazelcast.executor/SingleNodeTest/cancelWhileQueued/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-sonar/com.hazelcast$hazelcast/594/testReport/junit/com.hazelcast.executor/SingleNodeTest/cancelWhileQueued/

muratayan:
```java
assertFalse(queuedFuture.isDone());
assertTrue(queuedFuture.cancel(true));
assertTrue(queuedFuture.isCancelled());
assertTrue(queuedFuture.isDone());
```
Sometimes queuedFuture is mysteriously finishing before `cancel(true)` operation.

In theory it should remain in the queue while first task is still running which is:
```java
Callable task1 = new SleepingTask(100);
Future inProgFuture = executor.submit(task1);
```

bilalyasar:
[...]

PR #5022, opened 2015-04-08
SingleNodeTest.cancelWhileQueued is fixed

muratayan:
This pull request resolves the test failure issue #4915. For this test' we have an `ExecutorService` with only one thread; we are first submitting a sleep task to it, so the second task we submit consecutively is queued. However, problem was, sometimes, by the time the second task can be submitted, first task finishes. Thus, while we ask for a `CancellationException`, we get an `AssertionError` because the task we expect to be queued is already finished as well.

There is one thing to be considered in my solution, when we repeat the test more than 100 times with `@Repeat()` annotation, `ExecutorService` starts to return `CompletedFuture`  instances, so the test definitely fails. Maybe we could use the submit function with true `preventSync` variable.


bilalyasar:

ahmetmircik:
Seems mergeable to me.

mesutcelik:
verify

muratayan:
@bilalyasar can you verify please?

bilalyasar:
verify

PR #5025, opened 2015-04-08
Fix for issue #4631 (CacheManager.getCache() does not re-open closed cache)

### LINKED ISSUE #4631: CacheManager.getCache() does not re-open closed cache
According to the spec;

> Depending on the implementation and Cache topology, e.g., a storage-backed or distributed caches, the contents of a closed Cache may still be available and accessible by other applications or in fact via the Cache Manager that previously owned the Cache if an application calls getCache at some point in the future.

CacheManager.getCache() method should let access to the closed cache and re-open it. However, cache can be accessed by getCache but is still closed.

```java
    @org.junit.Test
    public void testGetCacheAfterClose() {
        CachingProvider provider = HazelcastServerCachingProvider
                .createCachingProvider(Hazelcast.newHazelcastInstance());

        CacheManager cacheManager = provider.getCacheManager();

        String cacheName = "test";
        Cache<Object, Object> cache = cacheManager.createCache(cacheName, new CacheConfig());
        cache.close();
        Assert.assertTrue(cache.isClosed());

        Cache<Object, Object> cacheAfterClose = cacheManager.getCache(cacheName);
        Assert.assertNotNull(cacheAfterClose);
        Assert.assertFalse("Cache should be opened after CacheManager.getCache() is called", cacheAfterClose.isClosed());
    }
```

serkan-ozal:
Fixes #4631

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/ICache.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/ICache.java
@@ -944,4 +944,12 @@ public interface ICache<K, V>
      */
     CacheStatistics getLocalCacheStatistics();
 
+    /**
+     * Opens cache if available (not destroyed).
+     * If cache is already open, does nothing.
+     *
+     * @throws java.lang.IllegalStateException if cache is already destroyed
+     */
+    void open();
+
 }
--- a/hazelcast/src/main/java/com/hazelcast/cache/ICache.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/ICache.java
@@ -944,12 +944,4 @@ public interface ICache<K, V>
      */
     CacheStatistics getLocalCacheStatistics();
 
-    /**
-     * Opens cache if available (not destroyed).
-     * If cache is already open, does nothing.
-     *
-     * @throws java.lang.IllegalStateException if cache is already destroyed
-     */
-    void open();
-
 }
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -72,6 +72,7 @@
     <suppress checks="JavadocType" files="com.hazelcast.cache.impl[\\/]"/>
     <suppress checks="JavadocMethod" files="com.hazelcast.cache.impl[\\/]"/>
     <suppress checks="JavadocVariable" files="com.hazelcast.cache.impl[\\/]"/>
+    <suppress checks="MethodCount" files="com.hazelcast.cache.ICache"/>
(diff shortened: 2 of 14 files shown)


hazelcast/src/main/java/com/hazelcast/cache/ICache.java
```diff
@@ -944,4 +944,12 @@
      */
     CacheStatistics getLocalCacheStatistics();
 
+    /**
+     * Opens cache if available (not destroyed).
+     * If cache is already open, does nothing.
+     *
+     * @throws java.lang.IllegalStateException if cache is already destroyed
+     */
+    void open();
```
mdogan: Do we really need to add `open()` to public API?

mdogan: I prefer not to introduce a new public method.

serkan-ozal: Via a new internal API like `ICacheInternal` (`ICache` will extends from this), we can handle some features via this interface. Reason for adding new method to public API is handling case from only `AbstractHazelcastCacheManager` instead of handling in both of `HazelcastClientCacheManager` and `HazelcastServerCacheManager` since there is no connection between `ClientCacheProxy` and `CacheProxy` except `ICache` interface.

mdogan: No, `ICache` should not extend internal interfaces. Only proxies can implement/extend internal interfaces/classes. You can make `ClientProxy`/`ClientCacheProxy` extend/implement internal one.

serkan-ozal: But in that case, since there is no relation between `ICache` and `ICacheInternal` and there will be a cast (in `AbstractHazelcastCacheManager`) to `ICacheInternal` via `instanceof` or forcefully which doesn't not seems good to me.

serkan-ozal: Another way, is extending `ICacheInternal` from `ICache` and `AbstractHazelcastCacheManager` interact with `ICacheInternal` but gives `ICache` to API user since in this case `ICacheInternal` is a kind of `ICache`

serkan-ozal: @mdogan can you look at the updated version of this PR with `ICacheInternal` that extends from `ICache` but servies `ICache` to API users. WDYT about this one ?

serkan-ozal:
verify

PR #5032, opened 2015-04-09
[CP] put and entry listener tests are passing for new client protocol

sancar:
Getting Partition and Member list, Authentication, Map put and add entry listener is implemented to make sure SimpleProtocolTest is passing. This test will probably removed in later client protocol since there is already similar tests.


sancar:
verify

asimarslan:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/MessageTaskFactoryImpl.java
```diff
@@ -22,21 +44,41 @@ public MessageTaskFactoryImpl(Node node) {
...
                 return new AuthenticationCustomCredentialsMessageTask(clientMessage, node, connection);
             }
         });
-        tasks.put(MapMessageType.MAP_PUT.id(), new MessageTaskFactory() {
+        tasks.put(ClientMessageType.MAP_PUT_REQUEST.id(), new MessageTaskFactory() {
             public PartitionSpecificRunnable create(ClientMessage clientMessage, Node node, Connection connection) {
                 return new MapPutMessageTask(clientMessage, node, connection);
             }
         });
+        tasks.put(ClientMessageType.ADD_ENTRY_LISTENER_REQUEST.id(), new MessageTaskFactory() {
```
mdogan: Is there any interface/class named `MessageTask`? Otherwise `MessageTaskFactory` name doesn't name sense to me. It's creating a type that does not exist.

mdogan: Btw, this class is called `MessageTaskFactoryImpl` but it doesn't implement `MessageTaskFactory` or any other interface. another weird naming...

Issue #4617, opened 2015-02-12
close() method of a Closeable CacheLoader is called without explicitly calling Cache.close()


mdogan:
`close()` method of a singleton `Closeable` `CacheLoader` is called when some of the partitions are moved to a newly started `HazelcastInstance`... which is an unexpected situation.. I even call this violation. 

```java
@org.junit.Test
public void testCloseableCacheLoader() throws InterruptedException {
    CachingProvider provider = HazelcastServerCachingProvider
            .createCachingProvider(Hazelcast.newHazelcastInstance());

    CacheManager cacheManager = provider.getCacheManager();

    CloseableCacheLoader loader = new CloseableCacheLoader();

    Factory<CloseableCacheLoader> loaderFactory = FactoryBuilder.factoryOf(loader);
    CompleteConfiguration<Object, Object> configuration = new CacheConfig()
            .setCacheLoaderFactory(loaderFactory).setReadThrough(true);

    Cache<Object, Object> cache = cacheManager.createCache("test", configuration);

    // trigger partition assignment
    cache.get("key");

    Hazelcast.newHazelcastInstance();

    for (int i = 0; i < 1000; i++) {
        cache.get(i);
        LockSupport.parkNanos(1000);
    }

    Assert.assertFalse("CacheLoader should not be closed!", loader.closed);
}

static class CloseableCacheLoader implements CacheLoader, Closeable, Serializable {

    volatile boolean closed = false;

    @Override
    public Object load(Object key) throws CacheLoaderException {
        if (closed) {
            throw new IllegalStateException();
        }
        return null;
    }

    @Override
    public Map loadAll(Iterable keys) throws CacheLoaderException {
        if (closed) {
            throw new IllegalStateException();
        }
        return null;
    }

    @Override
    public void close() throws IOException {
        closed = true;
    }
}
```

serkan-ozal:
@mdogan taking this one

Issue #4616, opened 2015-02-12
Cache.close() doesn't call close() method of registered Closeable CacheEntryListener


mdogan:
Cache.close() doesn't call close() method of registered Closeable CacheEntryListener. It seems there's no TCK test to verify but this is a spec violation. `Cache.close()` javadoc explicitly says:

```
must close and release all resources being coordinated on behalf of the Cache by the 
{@link CacheManager}. This includes calling the <code>close</code> method on configured 
{@link CacheLoader}, {@link CacheWriter}, registered {@link CacheEntryListener}s and
{@link ExpiryPolicy} instances that implement the java.io.Closeable interface.
```

```java
@org.junit.Test
public void testCloseableCacheListener() {
    CachingProvider provider = HazelcastServerCachingProvider
            .createCachingProvider(Hazelcast.newHazelcastInstance());

    CacheManager cacheManager = provider.getCacheManager();

    CloseableListener listener = new CloseableListener();

    Factory<CloseableListener> listenerFactory = FactoryBuilder.factoryOf(listener);
    CompleteConfiguration<Object, Object> configuration = new CacheConfig()
            .addCacheEntryListenerConfiguration(
                    new MutableCacheEntryListenerConfiguration(listenerFactory, null, true, false));

    Cache<Object, Object> cache = cacheManager.createCache("test", configuration);
    cache.close();

    Assert.assertTrue("CloseableListener.close() should be called when cache is closed!", 
                listener.closed);
}

static class CloseableListener implements CacheEntryCreatedListener, Closeable, Serializable {

    volatile boolean closed = false;

    @Override
    public void onCreated(Iterable iterable) throws CacheEntryListenerException {
        if (closed) {
            throw new IllegalStateException();
        }
    }

    @Override
    public void close() throws IOException {
        closed = true;
    }
}
```

serkan-ozal:
@mdogan taking this one

PR #5033, opened 2015-04-09
Fix for issue #4616 [close() method of a Closeable CacheLoader is called without explicitly calling Cache.close()] and #4617 [Cache.close() doesn't call close() method of registered Closeable CacheEntryListener]

serkan-ozal:
Fix for both issues #4616 and #4617


gurbuzali:
:+1: 
waiting for the back-port https://github.com/hazelcast/hazelcast/pull/5039

PR #5037, opened 2015-04-10
Renamed containers for IAtomicLong and IAtomicReference

pveentjer:
to be more in...line with the rest of the system. The container for IAtomicLong ==> AtomicLongContainer. And the container for IAtomicReference ==> AtomicReferenceContainer.


jerrinot:
:+1:

PR #5049, opened 2015-04-13
Running Query Operation on a GenericPool

jerrinot:
According to our measurement it yields significant performance improvement up until high 100,000s of entries per member.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -316,6 +316,19 @@ public class GroupProperties {
             = "hazelcast.backpressure.max.concurrent.invocations.per.partition";
 
 
+    /**
+     * Run Query Evaluations for multiple partitions in parallel.
+     *
+     * Each Hazelcast member evaluates query predicates using a single thread by default. In most cases the overhead of
+     * inter-thread communication overweight benefit of parallel execution.
+     *
+     * When you have a large dataset and/or slow predicate you may benefit from parallel predicate evaluations.
+     * Set to true if you are using slow predicates or have > 100,000s entries per member.
+     *
+     * Default: false
+     *
+     */
+    public static final String PROP_QUERY_PREDICATE_PARALLEL_EVALUATION = "hazelcast.query.predicate.parallel.evaluation";
 
 
     /**
@@ -542,6 +555,8 @@ public class GroupProperties {
     public final GroupProperty QUERY_RESULT_SIZE_LIMIT;
     public final GroupProperty QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK;
 
+    public final GroupProperty QUERY_PREDICATE_PARALLEL_EVALUATION;
+
     public GroupProperties(Config config) {
         HEALTH_MONITORING_LEVEL
                 = new GroupProperty(config, PROP_HEALTH_MONITORING_LEVEL, HealthMonitorLevel.SILENT.toString());
@@ -664,6 +679,9 @@ public class GroupProperties {
         QUERY_RESULT_SIZE_LIMIT = new GroupProperty(config, PROP_QUERY_RESULT_SIZE_LIMIT, "-1");
         QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK
                 = new GroupProperty(config, PROP_QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK, "3");
+
+        QUERY_PREDICATE_PARALLEL_EVALUATION
+                = new GroupProperty(config, PROP_QUERY_PREDICATE_PARALLEL_EVALUATION, "false");
     }
 
     public static class GroupProperty {
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
@@ -17,6 +17,7 @@
 package com.hazelcast.map.impl.operation;
 
 import com.hazelcast.core.MemberLeftException;
+import com.hazelcast.instance.GroupProperties;
 import com.hazelcast.map.impl.MapContextQuerySupport;
 import com.hazelcast.map.impl.MapServiceContext;
 import com.hazelcast.map.impl.QueryResult;
@@ -31,6 +32,7 @@ import com.hazelcast.spi.ExceptionAction;
 import com.hazelcast.spi.ExecutionService;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.ReadonlyOperation;
+import com.hazelcast.spi.exception.RetryableHazelcastException;
 import com.hazelcast.spi.exception.TargetNotMemberException;
 import com.hazelcast.util.FutureUtil;
 import java.io.IOException;
@@ -74,6 +76,7 @@ public class QueryOperation extends AbstractMapOperation implements ReadonlyOper
     @Override
     public void run() throws Exception {
         InternalPartitionService partitionService = getNodeEngine().getPartitionService();
+        NodeEngine nodeEngine = getNodeEngine();
         MapServiceContext mapServiceContext = mapService.getMapServiceContext();
         MapContextQuerySupport mapQuerySupport = mapServiceContext.getMapContextQuerySupport();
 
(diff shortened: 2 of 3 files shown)


Donnerbart:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
```diff
@@ -115,12 +113,41 @@ public void run() throws Exception {
...
 
+    private void fullTableScan(Collection<Integer> initialPartitions, GroupProperties groupProperties)
+            throws InterruptedException, ExecutionException {
+        if (pagingPredicate != null) {
+            runParallelForPaging(initialPartitions);
+        } else {
+            boolean parallelEvaluation = groupProperties.QUERY_PREDICATE_PARALLEL_EVALUATION.getBoolean();
+            if (parallelEvaluation) {
+                runParallel(initialPartitions);
+            } else{
```
Donnerbart: `else{` -> `else {`

hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
```diff
@@ -115,12 +113,41 @@ public void run() throws Exception {
...
 
+    protected void runSingleThreaded(final Collection<Integer> initialPartitions) throws InterruptedException, ExecutionException {
+        final NodeEngine nodeEngine = getNodeEngine();
+        final ExecutorService executor
+                = nodeEngine.getExecutionService().getExecutor(ExecutionService.QUERY_EXECUTOR);
+
+        for (Integer partitionId : initialPartitions) {
+            MapContextQuerySupport mapContextQuerySupport = mapService.getMapServiceContext()
+                    .getMapContextQuerySupport();
+            Collection<QueryableEntry> queryableEntries = mapContextQuerySupport.queryOnPartition(name, predicate, partitionId);
```
Donnerbart: Note to myself: We may have to modify this like `runParallel()` if the unbounded return values feature get's merged.

Will be something like this (not tested):
```
        Collection<Collection<QueryableEntry>> returnedResults = mapContextQuerySupport.queryOnPartition(name, predicate, partitionId);
        for (Collection<QueryableEntry> returnedResult : returnedResults) {
            if (returnedResult == null) {
                continue;
            }
            if (!addQueryEntriesToResult(returnedResult)) {
                break;
            }
        }
```

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -260,7 +260,19 @@
     public static final String PROP_BACKPRESSURE_MAX_CONCURRENT_INVOCATIONS_PER_PARTITION
             = "hazelcast.backpressure.max.concurrent.invocations.per.partition";
 
-
+    /**
+     * Run Query Evaluations for multiple partitions in parallel.
+     *
+     * Each Hazelcast member evaluates query predicates using a single thread by default. It most cases the overhead of
```
Donnerbart: `It most cases` -> `In most cases`

Donnerbart:
Beside checkstyle issues and minor comments :+1:

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
jerrinot: thanks!

jerrinot:
verify

jerrinot:
some test failure as suspicious:
```
11:00:43 
11:00:43 Tests in error: 
11:00:43   QueryAdvancedTest.testQueryAfterInitialLoad:410->HazelcastTestSupport.assertTrueEventually:639->HazelcastTestSupport.assertTrueEventually:625->Object.wait:-2 » TestTimedOut
11:00:43   MapStoreTest.testKeysWithPredicateShouldLoadMapStore:1189->HazelcastTestSupport.assertTrueEventually:639->HazelcastTestSupport.assertTrueEventually:625->Object.wait:-2 » TestTimedOut
11:00:43   MapStoreTest.issue614:332->Object.wait:-2 » TestTimedOut test timed out after ...
11:00:43 
11:00:43 Tests run: 2450, Failures: 0, Errors: 3, Skipped: 16
```

jerrinot:
verify

pveentjer:
How is this task going?

jerrinot:
verify

jerrinot:
verify

jerrinot:
Good. All tests are passing now. Now I can finally run the benchmark @ajermakovics requested. 

There has been a spurious test failure, but it was caused by a racy test. I'll send a new PR with a fix.

jerrinot:
verify

Donnerbart:
Looks fine to me. The `addQueryEntriesToResult()` should set the exceeded flag as soon as my PR get's merged, so there should be no need for further sync between our features.

I'm just not sure if we should make this the default behavior, since the performance numbers were not 100% clear in the TDD. But I think that's what Jaromir is going to do in the lab later today.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
```diff
@@ -114,6 +111,41 @@ public void run() throws Exception {
...
+        MapServiceContext mapServiceContext = mapService.getMapServiceContext();
+        MapContextQuerySupport querySupport = mapServiceContext.getMapContextQuerySupport();
+        for (Integer partitionId : initialPartitions) {
+            try {
+                Collection<QueryableEntry> entries = querySupport.queryOnPartition(name, predicate, partitionId);
+                if (!addQueryEntriesToResult(entries)) {
+                    break;
+                }
+            } catch (RuntimeException e) {
+                if (storedException == null) {
```
Donnerbart: I'm not sure if you have to safe the exception.

The `getResult()` from the parallel execution has the `FutureUtil.RETHROW_EVERYTHING` exception handling set. If I understood the code correctly an exception is re-thrown immediatly on each `future.get()`. This would be each `querySupport.queryOnPartition()` call in your method. In the parallel execution this would also happen before even a single query entry is added to the result. So it should be safe to do this directly in the for loop.

jerrinot: It's actually needed - as otherwise some tests were failing. 

Here is why:
- Let's assume I have a map with a (slow) MapLoader. The map has been just created and no operation has been executed against this map.
- Someone sends a new `QueryOperation`.
- The parallel execution sends tasks to every partition and waits for results
- Each task will trigger partition map creation. This trigger content to by loaded from a MapLoader 
- the `future.get()` throws an RetryableException because it's still loading from a MapLoader -> The Exception is propagated to a caller. 
- The important thing is the tasks for all partitions had been already send -> they will trigger initialization MapLoader for all map partitions even when the very first `future.get()` throws an  Exception
- The caller eventually retry the QueryOperation -> at this point all partitions are ready with a content loaded

Things are quite different when the QueryOperation is not offloading tasks to Query Executor:
- It calls `querySupport.queryOnPartition(name, predicate, partitionId);` in a loop for all partitions. - Each call will initialization a given partition -> it triggers mapLoader -> the call throws the RetryableException (because it's loading)
- Now unlike the parallel scenario the Exception is propagated back to a caller and no additional map partition initialization is trigged!
- The caller will eventually retry the QueryOperation -> the 1st iteration will succeed (because the 1st map partition is already loaded at this point), but the 2nd iteration will fail again.

I hope you made it a bit clearer :)

Donnerbart: Yes, thanks! I still have to look into the (slow) MapLoader thing, but your explanation was very helpful.

So I have no objection to the exception approach of your method anymore :)

pveentjer:
+1 Can you do a rebase and give me a ping?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
```diff
@@ -114,6 +111,41 @@ public void run() throws Exception {
...
+    }
+
+    protected void runSingleThreaded(final Collection<Integer> initialPartitions) {
+        RuntimeException storedException = null;
+        MapServiceContext mapServiceContext = mapService.getMapServiceContext();
+        MapContextQuerySupport querySupport = mapServiceContext.getMapContextQuerySupport();
+        for (Integer partitionId : initialPartitions) {
+            try {
+                Collection<QueryableEntry> entries = querySupport.queryOnPartition(name, predicate, partitionId);
+                if (!addQueryEntriesToResult(entries)) {
```
Donnerbart: For the rebase: This should be just `result.addAll(entries);` now, everything else should still be the same :)

pveentjer:
Do we have new performance info when the query is complex? I guess that with complex queries the overhead of reflection starts to outweighs the benefits you have introduced and therfore the effects will be less.

pveentjer:
Can you add some tests? I really want to have these branches (both of them) covered.

Donnerbart:
With all the actual changes, JavaDoc and tests: :+1:

pveentjer:
+1 from me.

@ahmetmircik can you have a look as well.

pveentjer:
Can be merged once green.

jerrinot:
verify

pveentjer:
wooohoooooo

PR #5050, opened 2015-04-13
[CP] Code generator for client protocol dto objects, xParameters classes

asimarslan:
Client Protocol has DTO's to encode/decode from/to byteBuffer. Each has an encode and decode method where their impls are exactly same but the parameters differ. So we introduce an annotation processor based code generator. Interfaces annotated with **GenerateParameters** are the parameter DTO templates to be generated.

Checkstyle/findbugs handles these generated files they are still suppressed and we may not solve all the warnings.

By this PR, we will control the client protocol serialization from a single point namely the template interfaces.


hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
```diff
@@ -0,0 +1,15 @@
...
+
+    private CodeGenerationUtils() {
+    }
+
+    public static String capitalizeFirstLetter(String input) {
+        return input.substring(0, 1).toUpperCase() + input.substring(1);
+    }
+
+    public static String getPackageNameFromQualifiedName(String qualifiedClassName) {
+        return qualifiedClassName.substring(0, qualifiedClassName.lastIndexOf("."));
```
jerrinot: this fails for classes from a default package.

hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
```diff
@@ -0,0 +1,15 @@
+package com.hazelcast.client.protocol.generator;
+
+public final class CodeGenerationUtils {
+
+    private CodeGenerationUtils() {
+    }
+
+    public static String capitalizeFirstLetter(String input) {
+        return input.substring(0, 1).toUpperCase() + input.substring(1);
```
jerrinot: this depends on a default locale of JVM. 
is it what you want? perhaps you want to use a well-defined (fixed) locale.

hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerator.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private void saveClass(TypeElement classElement, String packageName, String className, String content) {
+        JavaFileObject file;
+        try {
+            final String fullClassName = packageName + "." + className;
+            file = filer.createSourceFile(fullClassName, classElement);
+            file.openWriter().append(content).close();
+        } catch (IOException e) {
+            messager.printMessage(Diagnostic.Kind.ERROR, e.getMessage());
```
jerrinot: shouldn't you throw an exception (and stop the build) instead?

asimarslan:
verify

asimarslan:
verify

sancar:
:+1:

asimarslan:
@jerrinot, I will have more improvement on this generator waiting on another branch. I will add the fixes related to your comments to the following PR.

PR #5052, opened 2015-04-13
added comma support for client network config

### LINKED ISSUE #4933: Comma Separated network members for hazelcast client + spring integration
When we are configuring a hazelcast client using Spring XML tags, we  use the below spring configuration XML. It would be a nice feature to allow a comma-separated hazelcast nodes for client configuration, like hz:members tag that is currently present in the Hazelcast Server tag. In this way, we dont have to modify the spring configuration XML and simply change the property file that is loaded for spring.

<hz:client id="hazelcastClient">
	<hz:group name="${hazelcast.group.name}" password="${hazelcast.group.password}" />
	<hz:properties>
		<hz:property name="hazelcast.logging.type">${hazelcast.logging.type}</hz:property>
	</hz:properties>
	<hz:network smart-routing="true" redo-operation="true"
		connection-attempt-limit="${java.int.max.value}">
		<hz:member>${hazelcast.network.member1}</hz:member>
		<hz:member>${hazelcast.network.member2}</hz:member>
	</hz:network>
</hz:client>

bilalyasar:
this pr closes #4933

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientNetworkConfig.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientNetworkConfig.java
@@ -20,8 +20,10 @@ import com.hazelcast.config.SSLConfig;
 import com.hazelcast.config.SocketInterceptorConfig;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
+import java.util.StringTokenizer;
+
+import static com.hazelcast.util.ValidationUtil.hasText;
 
 /**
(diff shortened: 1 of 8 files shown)


mesutcelik:
seems ok but requires documentation update. 
I also don't know if that is a good idea to use singular ```member``` tag for comma seperated values.

pveentjer:
I'm also not happy with using member for allow for multiple items. It is a bit of a surprise and a violation of the 'The Principle of Least Astonishment'

mesutcelik:
I think we should introduce `members` tag for this issue with another PR

PR #5057, opened 2015-04-14
spring schema supporting aws disabled

### LINKED ISSUE #4310: XML schema and XML config validation is not compliant for AWS configuration
When aws is disabled in spring xml configuration file, Hazelcast 3.3.3 is complaining about missing properties in the configuration file via message like `accessKey and secretKey can not be null`
__This does not comply to the XML schema__. For both, hazelcast and hazelcast-spring those 2 __properties are both _optional_ according to the schema__. 

This does not seem to be right, because those values do not make sense, if aws is disabled. 

spring schema:
```
<xs:complexType name="aws">
        <xs:sequence/>
        <xs:attribute name="enabled" type="xs:string" use="optional" default="true"/>
        <xs:attribute name="access-key" type="xs:string" use="optional" default=""/>
        <xs:attribute name="secret-key" type="xs:string" use="optional" default=""/>
 </xs:complexType>    
```
This gives error:
`<hz:aws enabled="false">`

bilalyasar:
default values of secret-key and access-key of aws config were "" ( empty string) in spring xsd.
But in hazelcast aws config we have empty string checking, see:  https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/config/AwsConfig.java#L74

so in spring xsd we have to remove default values.

this pr closes #4310

### DIFF
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.5.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.5.xsd
@@ -1273,8 +1273,8 @@
     <xs:complexType name="aws">
         <xs:sequence/>
         <xs:attribute name="enabled" type="xs:string" use="optional" default="true"/>
-        <xs:attribute name="access-key" type="xs:string" use="optional" default=""/>
-        <xs:attribute name="secret-key" type="xs:string" use="optional" default=""/>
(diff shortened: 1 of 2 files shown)


mesutcelik:
can you please send the backport too?

Issue #5043, opened 2015-04-12
SlowOperationDetector disable logging option


pveentjer:
There should be an option to disable the SlowOperatioDetector logging the stacktraces to the log file. It is very noisy and if you already using the management center or the performance monitor, there is no need to have the stacktraces written to the normal log file.

pveentjer:
There should be a hint logging to indicate to the end user that this feature is available. So the first time you log something on the logger, we should hint about the exist

Donnerbart:
Do you want an option just for the stack trace part or the whole logging of the `SlowOperationDetector`? I could create two properties. We could also differ between the first occurence of an operation and the follow ups:
```
SLOW_OPERATION_DETECTOR_LOGGING_LEVEL = {"none", "first", "always"};
SLOW_OPERATION_DETECTOR_LOGGING_STACKTRACE = {"none", "first", "always"};
```

or just:
```
SLOW_OPERATION_DETECTOR_LOGGING_STACKTRACE = {"false", "true"};
```

PR #5059, opened 2015-04-14
Minor changes to logging of slow operations

### LINKED ISSUE #5041: SlowInvocation should show full classname
SlowOperationDTO.operation shows only simple classname, not the package. This is no good because this can lead to ambiguity issues and we can't track the actual class.

Donnerbart:
* Changed the content of operation in slow operation logs to contain the fully qualified classname instead of `operation.toString()`.
* Added `operation.toString()` as details to each invocation and `SlowOperationInvocationDTO`.
* Added a property to switch off stack trace logging.

Fixes #5041 and #5043.


Donnerbart:
BTW, this PR should not break the actual implementation in Management Center, because I just added a new field in the DTO (and changed the content of another one, but it's still a String).

PR #5060, opened 2015-04-14
hazelcast-client-full xml revision

bilalyasar:
i deleted some configs that we don't use anymore. ( as i talked with @sancar )
also there are some missing configs that i added to full-xml.

now client-xml includes all things from hazelcast-client-xsd.

i didn't write validation test because we have already a test, see: https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/test/java/com/hazelcast/client/config/XmlClientConfigBuilderTest.java#L255

### DIFF
--- a/hazelcast-client-new/src/main/resources/hazelcast-client-full.xml
+++ b/hazelcast-client-new/src/main/resources/hazelcast-client-full.xml
@@ -23,20 +23,34 @@
         <name>dev</name>
         <password>dev-pass</password>
     </group>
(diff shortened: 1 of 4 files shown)


mesutcelik:
there are some possibly related test failures.
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/12820/

bilalyasar:
@mesutcelik i have rebased and fixed failing tests, i think it can be merged.

PR #5064, opened 2015-04-14
Near Cache Expiration Implementation

serkan-ozal:
This PR contains expiration implementation for Near Cache but not contains eviction implementation.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheDataRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheDataRecordStore.java
@@ -23,6 +23,7 @@ import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.util.Clock;
 
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
@@ -45,10 +46,10 @@ public class NearCacheDataRecordStore<K, V>
     protected long getKeyStorageMemoryCost(K key) {
         if (key instanceof Data) {
             return
-                    // Reference to this key data inside map ("store" field)
-                    REFERENCE_SIZE
-                            // Heap cost of this key data
-                            + ((Data) key).getHeapCost();
+                // Reference to this key data inside map ("store" field)
+                REFERENCE_SIZE
+                // Heap cost of this key data
+                + ((Data) key).getHeapCost();
         } else {
             // Memory cost for non-data typed instance is not supported.
             return 0L;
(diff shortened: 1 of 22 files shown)


serkan-ozal:
verify

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheDataRecordStore.java
```diff
@@ -157,4 +167,18 @@ public int size() {
         return store.size();
     }
 
+    @Override
+    public void doExpiration() {
+        for (Map.Entry<K, NearCacheDataRecord> entry : store.entrySet()) {
+            if (Thread.currentThread().isInterrupted()) {
```
gurbuzali: We are checking interrupted status for every iteration, is this necessary ?
When do we need to interrupt this thread ? while destroying ?

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheObjectRecordStore.java
```diff
@@ -137,4 +144,18 @@ public int size() {
         return store.size();
     }
 
+    @Override
+    public void doExpiration() {
+        for (Map.Entry<K, NearCacheObjectRecord> entry : store.entrySet()) {
+            if (Thread.currentThread().isInterrupted()) {
```
gurbuzali: same

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheObjectRecordStore.java
```diff
@@ -137,4 +144,18 @@ public int size() {
         return store.size();
     }
 
+    @Override
```
gurbuzali: data record store and object record store classes looks very similar, can't you use a base class to eliminate code duplication ?

serkan-ozal: ok

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheDataRecordStore.java
serkan-ozal: Yes, while destroying near-cache, expiration task is shutdown via `shutdown(true)` and this also interrupts `ExpirationTask` runner thread. So via `Thread.currentThread().isInterrupted()`, we don't wait until expiration finished since near-cache is destroyed and no need to wait for finishing of expiration.

metanet:
:+1:

PR #5077, opened 2015-04-16
[CP] Templates for code generation are implemented for some small data structures.

sancar:
Templates for code generation are implemented for following data structures: AtomicLong, AtomicReference, Condition, CountDownLatch, ExecutorService, Lockm Semaphore, Topic


sancar:
verify

PR #5079, opened 2015-04-17
New group properties for EWR

emrahkocaman:


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -293,7 +294,7 @@ boolean initInvocationTarget() {
         }
 
         targetMember = nodeEngine.getClusterService().getMember(invTarget);
-        if (!isJoinOperation(op) && targetMember == null) {
+        if (!(isJoinOperation(op) || isWanReplicationOperation(op)) && targetMember == null) {
```
jerrinot: what's the reason for the change? how is it related to new GroupProperties?

jerrinot: + perhaps you might get a minor performance benefit by changing order of operands to
```java
targetMember == null && !(isJoinOperation(op) || isWanReplicationOperation(op))
```

as targetMember is not null in vast majority of cases - neither `isJoinOperation()` or `isWantReplicationOperation()` will be called thanks to short-circuiting.

emrahkocaman: @jerrinot , For better delivery, we're changing the wan event delivery mechanism (for Batch and no-delay implementations) and new one depends on operations. Since target wan members are not members of the source cluster, we needed to introduce this change to allow WAN operations run on target cluster members.

jerrinot: it would be great to include this explanation in a commit message. just for a future reference. 
currently the commit message doesn't say anything about it.

emrahkocaman:
Closing this PR, I'll create two different PR's for group properties and operation check.

jerrinot:
you could do that even with this PR - git interactive rebase allows you to mess with commits. 

I think best what you can is to have a single branch named let's say "Preparation for Enterprise WAN Replication" and commit two changes:

Add Group Properties for Enterprise WAN Replication
//Explanation

WAN Replication Operation Check (or something like this)
//Explanation

You could push this branch to GitHub and open a single PR. The two distinct commits would be maintained in a history, but the merge commit would contain a larger logical block.

PR #5080, opened 2015-04-17
Cleanup of query related classes

Donnerbart:
Cleanup of query related classes (as preparation for unbounded return values PR).

This PR has been pulled out of my "unbounded return values" PR to easy the review process and to keep the PRs smaller. Maybe not everything in this PR makes sense, since the context of the upcoming feature PR is missing.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
@@ -376,6 +376,7 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
                     handleAuthenticationFailure(endpoint, request);
                 }
             } catch (Throwable e) {
+                logProcessingFailure(request, e);
                 handleProcessingFailure(endpoint, request, packet.getData(), e);
             }
         }
@@ -404,7 +405,7 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
             }
         }
 
-        private void handleProcessingFailure(ClientEndpointImpl endpoint, ClientRequest request, Data data, Throwable e) {
+        private void logProcessingFailure(ClientRequest request, Throwable e) {
             Level level = nodeEngine.isActive() ? Level.SEVERE : Level.FINEST;
             if (logger.isLoggable(level)) {
                 if (request == null) {
@@ -413,7 +414,9 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
                     logger.log(level, "While executing request: " + request + " -> " + e.getMessage(), e);
                 }
             }
+        }
 
+        private void handleProcessingFailure(ClientEndpointImpl endpoint, ClientRequest request, Data data, Throwable e) {
             if (request != null && endpoint != null) {
                 endpoint.sendResponse(e, request.getCallId());
             } else if (data != null && endpoint != null) {
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/client/MapEntrySetRequest.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/client/MapEntrySetRequest.java
@@ -32,6 +32,7 @@ import com.hazelcast.security.permission.MapPermission;
 import com.hazelcast.spi.OperationFactory;
 import java.io.IOException;
 import java.security.Permission;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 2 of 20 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -376,6 +376,7 @@ public void run() {
                     handleAuthenticationFailure(endpoint, request);
                 }
             } catch (Throwable e) {
+                logProcessingFailure(request, e);
```
Donnerbart: I pulled out the logging because I most probably need the `handleProcessingFailure()` method on its own in the next PR. Even if not I don't think this will hurt if I don't use it in the future.

hazelcast/src/main/java/com/hazelcast/map/impl/client/MapEntrySetRequest.java
```diff
@@ -53,15 +54,12 @@ protected OperationFactory createOperationFactory() {
 
     @Override
     protected Object reduce(Map<Integer, Object> map) {
-        MapEntrySet entrySet = new MapEntrySet();
+        Set<Map.Entry<Data, Data>> entrySet = new HashSet<Map.Entry<Data, Data>>();
```
Donnerbart: Please have a look at this change! It's done the same way in `MapKeySetRequest` and `MapValuesRequest`, so we can make use of the built-in `addAll()` method.

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
```diff
@@ -280,25 +281,26 @@ private VersionedValue lockAndGet(Data key, long timeout) {
...
+            invokeQueryOperation(predicate, operationService, members, futures);
             collectResults(partitions, result, futures);
             if (partitions.size() == partitionCount) {
                 return result;
             }
-            findMissingPartitions(partitionCount, partitions, missingList);
         } catch (Throwable t) {
-            missingList.clear();
-            findMissingPartitions(partitionCount, partitions, missingList);
+            EmptyStatement.ignore(t);
```
Donnerbart: Please have a look at this! I saw no sense in calling `findMissingPartitions()` as last statement in the `try` block and calling it again in the `catch` clause (after clearing the list). I think this can be safely moved out of that block.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/QueryOperation.java
```diff
@@ -171,9 +158,24 @@ protected void runParallelForPaging(Collection<Integer> initialPartitions) throw
...
         }
-        for (QueryableEntry entry : toMerge) {
+        addQueryEntriesToResult(toMerge);
+    }
+
+    private static Collection<Collection<QueryableEntry>> getResult(List<Future<Collection<QueryableEntry>>> lsFutures) {
+        return returnWithDeadline(lsFutures, QUERY_EXECUTION_TIMEOUT_MINUTES, TimeUnit.MINUTES, FutureUtil.RETHROW_EVERYTHING);
+    }
+
+    private boolean addQueryEntriesToResult(Collection<QueryableEntry> queryableEntries) {
```
Donnerbart: The `boolean` return value doesn't make sense at the moment, but I will definitely use this in the next PR. That upcoming change will have a scenario for the `false` return statement.

hazelcast/src/main/java/com/hazelcast/map/impl/client/MapEntrySetRequest.java
jerrinot: +1

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
jerrinot: I'm not sure here. @gurbuzali: could you please have a look on this?

Donnerbart: `missingList` is just used as parameter of `findMissingPartitions()`. So before `findMissingPartitions()` is called the list is always empty.

Since `findMissingPartitions()` is the last statement in the `try` block there can be nothing in `missingList` if an exception occurs from the other statements. So the `misslingList.clear();` seems pointless.

If `findMissingPartitions()` is causing the exception it will happen again, since we call it with the same parameters, which cannot change in between. And since there is no other `try`-`catch` the exception of the second call would be propagated.

So it should be the same behavior (just easier to read) to create an fill the list after the `try` block.

Donnerbart: BTW, the reason why I touched this is that I have to insert a new `catch` for the specific exception which will be thrown by the "unbounded return values" feature.

jerrinot:
:+1:

Donnerbart:
It's green \o/ Will someone merge or at least have another look?

jerrinot:
it's up to a module owner to press the merge button. 
@ajermakovics: can you please have a look?

pveentjer:
It looks good to me. I would like to have @gurbuzali  and @asimarslan  to have a look as well since they are the module owners of the code being touched.

Donnerbart:
Everything I considered to be questionable I commented by myself. Most of these changes should be no-brainers. I just followed the code paths of queries, so a lot of classes and modules are involved. Sorry for that :( Also simple things like newline corrects and removal of `final` look horrible on GitHub, but they are super easy and quite safe to do in IDEA.

PR #5082, opened 2015-04-17
Small fixes and improvements on client protocol impl

mdogan:
- Fixed memory-leak in ClientMessageBuilder and rearranged a bit.
- Removed duplicate ByteBuffer wrapping and replaced all with direct byte[] allocations.
- Renamed client protocol socket reader and writer.
- Enabled protocol tests, they were not running before.


sancar:
:+1:

PR #5084, opened 2015-04-17
Fixed client membership listener. Enabled more tests.

mdogan:
Follow up https://github.com/hazelcast/hazelcast/pull/5083


mdogan:
verify

sancar:
:+1:

PR #5090, opened 2015-04-18
added async versions of commit and rollback methods to TransactionLog interface

gurbuzali:
This PR is needed for further XA implementation
PS: updated atomikos to latest version


hazelcast/src/main/java/com/hazelcast/collection/impl/txncollection/CollectionTransactionLog.java
```diff
@@ -91,6 +99,14 @@ public Future rollback(NodeEngine nodeEngine) {
     }
 
     @Override
+    public void rollbackAsync(NodeEngine nodeEngine, ExecutionCallback callback) {
```
pveentjer: Probably we should do argument checking.

sancar:
:+1:

PR #5091, opened 2015-04-18
Condition , ExecutorService, Topic , CountdownLatch tasks are implemented

sancar:


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/replicatedmap/AbstractReplicatedMapAddEntryListenerMessageTask.java
```diff
@@ -0,0 +1,127 @@
...
+import com.hazelcast.query.Predicate;
+import com.hazelcast.replicatedmap.impl.ReplicatedMapService;
+import com.hazelcast.replicatedmap.impl.record.ReplicatedRecordStore;
+import com.hazelcast.security.permission.ActionConstants;
+import com.hazelcast.security.permission.MapPermission;
+
+import java.security.Permission;
+
+public abstract class AbstractReplicatedMapAddEntryListenerMessageTask<Parameter>
+        extends AbstractCallableMessageTask<Parameter>
```
asimarslan: Lets use just P or Parameter everywhere

Issue #5072, opened 2015-04-15
Socket connection timeout caused by reverse name lookup delay


rofsdahl:
**With one member running, when starting a second member it takes 5 minutes before it joins the cluster.**

* Hazelcast version 3.4.2, java version 1.8
* Using 2 members (one Windows 7, one Oracle Linux 7 running in VM)
* Not using multicast
* Members manually configured with IP addresses
* **Host names not registered in DNS (nor in /etc/hosts)**

When TcpIpConnectionManager calls InetAddress.getHostName() from assignSocketChannel(), then a reverse name lookup is initiated (since the host names are not registered in DNS, nor in /etc/hosts). getHostName blocks for ~5 seconds (before it falls back to returning the IP address as host name). This causes socket connection timeout in the main thread (TcpIpJoiner gives up connecting after 5 seconds by default). Merging with the other node happens after split brain recovery is done (default after 300 seconds). See log A.

When the host name is registered in DNS or /etc/hosts, then getHostName() returns immediately, so that no timeout occurs for the socket connection in the main thread. See log B.

Is it possible to avoid calling getHostName if IP addresses are configured, or at least increase the default socket connection timeout value to >5 seconds?

Log A (host names NOT registered in DNS): http://pastebin.com/StGfVL8q
Log B (host names registered in DNS): http://pastebin.com/VzExp9NF

jerrinot:
Many thanks for a really nice bug report! 
We will fix in the upcoming Hazelcast 3.5 and the fix will likely be ported to a 3.4.x line as well.

rofsdahl:
Thanks for fixing this issue so quickly!

jerrinot:
Fixed in both maintanance-3.x and current master.
Thanks again for reporting the problem!

PR #5094, opened 2015-04-19
Calling InetSocketAddress::getHostName only when the selector hack is enabled.

jerrinot:
Fixing #5072

The other alternative is to remove the selector hack altogether. There is less need for it given the existence of IOBalancer.


pveentjer:
+1 for the hack to be removed. It has served its purpose, but the balancer is the way forward.

jerrinot:
ok, this PR can be closed.

I'll send a PR to remove the hack from a master and another PR to apply a fix in a maintenance branch.

PR #5095, opened 2015-04-20
WanOperation for delivery control in enterprise wan replication

emrahkocaman:
Also changes `OperationRunnerImpl` and `OperationAccessor` to allow WanOperation
to run on non-member targets

2nd part of https://github.com/hazelcast/hazelcast/pull/5079


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
```diff
@@ -311,7 +312,9 @@ public void run(Packet packet) throws Exception {
     }
 
     private boolean ensureValidMember(Operation op) {
-        if (isJoinOperation(op) || node.clusterService.getMember(op.getCallerAddress()) != null) {
+        if (isJoinOperation(op)
+                || isWanReplicationOperation(op)
+                || node.clusterService.getMember(op.getCallerAddress()) != null) {
```
jerrinot: this can be short-circuited as well - as majority of operations have a valid caller set.

emrahkocaman:
verify

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
emrahkocaman: Done.

hazelcast/src/main/java/com/hazelcast/spi/OperationAccessor.java
```diff
@@ -43,6 +44,11 @@ public static boolean isMigrationOperation(Operation op) {
                 && op.getClass().getClassLoader() == THIS_CLASS_LOADER;
     }
 
+    public static boolean isWanReplicationOperation(Operation op) {
+        return op instanceof WanReplicationOperation
+                && op.getClass().getClassLoader() == THIS_CLASS_LOADER;
```
pveentjer: Why do you add this to the OperationAccessor? And why does it matter which classloaded loaded it?

pveentjer: Ok. I see that already is being done on the OperationAccessor by other functions. Not very nice. Can you move all of them to a utility class called e..g Operations?

emrahkocaman: Ok. Moved to Operations.

hazelcast/src/main/java/com/hazelcast/spi/Operations.java
```diff
@@ -0,0 +1,47 @@
...
+        return op instanceof JoinOperation
+                && op.getClass().getClassLoader() == THIS_CLASS_LOADER;
+    }
+
+    public static boolean isMigrationOperation(Operation op) {
+        return op instanceof MigrationCycleOperation
+                && op.getClass().getClassLoader() == THIS_CLASS_LOADER;
+    }
+
+    public static boolean isWanReplicationOperation(Operation op) {
```
pveentjer: Can you add Javadoc.

Can you ask @mdogan why the classloader is added to deal with marker interfaces? Because we are using marker interfaces at other places without class loader check.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
```diff
@@ -239,7 +240,7 @@ protected void ensureNoPartitionProblems(Operation op) {
     }
 
     private boolean retryDuringMigration(Operation op) {
-        return !(op instanceof ReadonlyOperation || OperationAccessor.isMigrationOperation(op));
+        return !(op instanceof ReadonlyOperation || Operations.isMigrationOperation(op));
```
pveentjer: Can you use a static import. So the Operations. stuff isn't needed anymore. I'm trying to apply them where possible to make code easier to read (less long lines).

hazelcast/src/main/java/com/hazelcast/spi/Operations.java
```diff
@@ -0,0 +1,47 @@
...
+package com.hazelcast.spi;
+
+import com.hazelcast.cluster.impl.operations.JoinOperation;
+import com.hazelcast.cluster.impl.operations.WanReplicationOperation;
+import com.hazelcast.partition.MigrationCycleOperation;
+
+/**
+ * Utility class that contains helper methods related to {@link Operation}
+ */
+public final class Operations {
```
pveentjer: Can you move this whole class into the com.hazelcast.spi.impl.operations package? Afaik user doesn't need to see this and we need to prevent exposing new API's if not needed

hazelcast/src/main/java/com/hazelcast/spi/Operations.java
emrahkocaman: Javadoc added. I don't think we can add tests for `isWanReplicationOperation` in OSS as there is no concrete classes implementing `WanReplicationOperation`.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
emrahkocaman: Done, even this is out of scope of this PR.

hazelcast/src/main/java/com/hazelcast/spi/Operations.java
emrahkocaman: Moved to "com.hazelcast.spi.impl.operationutil" package.

mesutcelik:
verify

mdogan:
LGTM :+1:

PR #5110, opened 2015-04-20
Added QueryCache and QueryCacheConfig

ahmetmircik:
Please read docs of `QueryCache`in this PR to have some more idea.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/QueryCacheConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/QueryCacheConfig.java
@@ -1,0 +1,423 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.config;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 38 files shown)


pveentjer:
Query caching is afaik not part of the 3.5 release. There are a few things I'm worried about:

- how long are these results going to be stored? Currently query results can be huge, so there is risk at retaining a lot of data in memory.

- you are storing based on query. But what if a user generates a query dynamically e.g. 
```
for(int k=0;k<1000;k++){
    map.query("age="+k)
}
```
How many query results are going to be cached?

I think it is better that we create an official Design Document and give you some serious time to implement and test this feature. I think the feature is good to have; we don't have it.. some of our competition has it, but there are too many questions for me to merge it currently.

pveentjer:
Ignore what I said above.. I thought it was not an official feature for 3.5

hazelcast/src/main/java/com/hazelcast/core/IMap.java
```diff
@@ -1329,4 +1330,28 @@ boolean tryLock(K key, long time, TimeUnit timeunit)
...
+    /**
+     * Creates an always up to date snapshot of this {@code IMap}.
+     *
+     * @param name         the name of {@code QueryCache}
+     * @param predicate    the predicate for filtering entries.
+     * @param includeValue {@code true} if this {@code QueryCache} is allowed to cache values of entries, otherwise {@code false}
+     * @return the {@code QueryCache} instance with the supplied {@code name}
+     * @see QueryCache
+     */
+    QueryCache<K, V> getQueryCache(String name, Predicate<K, V> predicate, boolean includeValue);
```
pveentjer: Please add documentation regarding null argument behaviour. The same goes for the method bellow.

hazelcast/src/main/java/com/hazelcast/config/QueryCacheConfig.java
```diff
@@ -0,0 +1,302 @@
...
+        return populateInitially;
+    }
+
+    /**
+     * Set {@code true} to enable initial population, otherwise set {@code false}.
+     *
+     * @param populateInitially set {@code true} to enable initial population, otherwise set {@code false}
+     * @return this {@code QueryCacheConfig} instance.
+     */
+    public QueryCacheConfig setPopulateInitially(boolean populateInitially) {
```
pveentjer: Would not be better to call it just populate?

pveentjer:
Can you rebase?

PR #5111, opened 2015-04-20
Added no argument constructor to QueryEntry

ahmetmircik:
It may be useful in some cases to use same instance instead of creating a new one for every iteration when scanning large data sets for example.


hazelcast/src/main/java/com/hazelcast/query/impl/QueryEntry.java
```diff
@@ -31,14 +31,21 @@
...
+    private SerializationService serializationService;
+
+    public QueryEntry() {
+    }
 
     public QueryEntry(SerializationService serializationService, Data indexKey, Object key, Object value) {
+        setFields(serializationService, indexKey, key, value);
+    }
+
+    public void setFields(SerializationService serializationService, Data indexKey, Object key, Object value) {
```
pveentjer: I would call this method 'init'.

pveentjer:
Can be merged once green.

PR #5113, opened 2015-04-20
Added ClientAwareService behavior to MapService

ahmetmircik:
Introduced `AbstractMapService` to fix checkstyle issue.


pveentjer:
We should solve the actual problem of the complexity and not keep adding layers of unnecessary complexity. 

I don't see the added value for:
- AbstractMapServiceContextAwareFactory
- DefaultMapServiceContextAwareFactory
- MapServiceConstructor

The MapService can be created like any other service e.g. the ExecutionServiceImpl. On the init method the right dependencies can be pulled out of the NodeEngineImpl. Then you can also get rid of all the setters of MapService.

ahmetmircik:
These all for to make extendibility easier. For now we can keep them as is.

pveentjer:
The question is if all this extensability is needed and if there are no better ways. I have a very hard time understanding the code of the Map.

ahmetmircik:
Extendibility is needed for enterprise and it is in use atm. The whole point is:
 `MapService` is created by `MapServiceContextAwareFactory`. `MapServiceContextAwareFactory` also knows how to create auxiliary services used by `MapService`, like `ClientAwareService`. To extend `MapService` behavior, you only need to give the desired auxiliary service's new implementation to `MapServiceContextAwareFactory`.

pveentjer:
My observation is that you are introducing the AbstractMapService so you can reduce the number of methods. Imho you are causing more complexity than is being solved. I would rather have a suppression on the method count than a fake class hierarchy. 

MapService = just a class with a lot of needs. So it will probably get a lot of dependencies. If the class remains dumb, then having those setter methods should not be an issue.

ahmetmircik:
liked it. done.

hazelcast/src/main/java/com/hazelcast/map/impl/MapClientAwareService.java
```diff
@@ -0,0 +1,38 @@
...
+package com.hazelcast.map.impl;
+
+import com.hazelcast.spi.ClientAwareService;
+
+/**
+ * Defines {@link ClientAwareService} behavior of map service.
+ *
+ * @see ClientAwareService
+ */
+class MapClientAwareService implements ClientAwareService {
```
pveentjer: what is the purpose of the MapClientAwareService? It gets a dependency which isn't used and it has a method which is a NOP.

ahmetmircik: used in enterprise. to clean up resources.

pveentjer: Ok. So you override this structure in enterprise by something sensible. 

I'm fine with adding a No-implementation. Can you add it to the documentation + Can you remove the constructor and the field since it isn't used?

pveentjer:
ping.

PR #5119, opened 2015-04-20
ClientMapProxy is implemented with parameters related to protocol

sancar:


hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -441,35 +464,37 @@ public void set(K key, V value, long ttl, TimeUnit timeunit) {
...
     }
 
     @Override
     public boolean isLocked(K key) {
         checkNotNull(key, NULL_KEY_IS_NOT_ALLOWED);
         final Data keyData = toData(key);
-        MapIsLockedRequest request = new MapIsLockedRequest(name, keyData);
-        Boolean result = invoke(request, keyData);
-        return result;
+        ClientMessage clientMessage = MapIsLockedParameters.encode(name, keyData, ThreadUtil.getThreadId());
```
pveentjer: Isn't it easier to call it request and response instead of 'clientMessage' and 'resultMessage'. This provides more information which message is used for the request and which message contains the response.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapGetEntryViewMessageTask.java
```diff
@@ -47,6 +50,18 @@ protected MapGetEntryViewParameters decodeClientMessage(ClientMessage clientMess
         return MapGetEntryViewParameters.decode(clientMessage);
     }
 
+    @Override
+    protected ClientMessage encodeResponse(Object response) {
+        EntryView<Data, Data> dataEntryView = (EntryView<Data, Data>) response;
```
pveentjer: Perhaps best to use single line per argument.

sancar:
verify

pveentjer:
Will merge once green.

PR #5123, opened 2015-04-21
Exception throwing prevented in case of null value addition to index. 

ahmetmircik:
This is for the cases which we want to query on keys. If there is no index for keys, querying on keys should silently return null to indicate no result found

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/IndexService.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/IndexService.java
@@ -30,8 +30,9 @@ import java.util.concurrent.atomic.AtomicReference;
  * This class contains methods which manipulate and access index.
  */
 public class IndexService {
+    private static final Index[] EMPTY_INDEX = {};
     private final ConcurrentMap<String, Index> mapIndexes = new ConcurrentHashMap<String, Index>(3);
-    private final AtomicReference<Index[]> indexes = new AtomicReference<Index[]>();
+    private final AtomicReference<Index[]> indexes = new AtomicReference<Index[]>(EMPTY_INDEX);
     private volatile boolean hasIndex;
 
     public synchronized Index destroyIndex(String attribute) {
@@ -60,7 +61,8 @@ public class IndexService {
     }
 
     public void removeEntryIndex(Data indexKey) throws QueryException {
-        for (Index index : indexes.get()) {
+        Index[] indexes = getIndexes();
+        for (Index index : indexes) {
             index.removeEntryIndex(indexKey);
         }
     }
@@ -70,7 +72,8 @@ public class IndexService {
     }
 
     public void saveEntryIndex(QueryableEntry queryableEntry) throws QueryException {
-        for (Index index : indexes.get()) {
+        Index[] indexes = getIndexes();
+        for (Index index : indexes) {
             index.saveEntryIndex(queryableEntry);
         }
     }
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/query/impl/IndexService.java
```diff
@@ -56,11 +57,15 @@ public synchronized Index addOrGetIndex(String attribute, boolean ordered) {
     }
 
     public Index[] getIndexes() {
+        if (indexes.get() == null) {
```
jerrinot: wouldn't be better to set the reference to EMPTY_INDEX in a constructor? you wouldn't have to deal with a special case.

ahmetmircik: done.

hazelcast/src/test/java/com/hazelcast/query/impl/IndexServiceTest.java
```diff
@@ -96,4 +97,52 @@ public void testIndex2() throws Exception {
...
+
+        shouldReturnNull_whenQueryingOnKeys(indexService);
+    }
+
+
+    @Test
+    public void shouldNotThrowException_withNullValues_whenNoIndexAdded() throws Exception {
+        IndexService indexService = new IndexService();
+
+        shouldReturnNull_whenQueryingOnKeys(indexService);
```
jerrinot: what's the point of calling another method?

gurbuzali: to get rid off code duplication ? (see above test too)

jerrinot: ops, I missed that! thanks!

jerrinot:
(probably) unrelated test failures:
```
com.hazelcast.spring.cache.TestCacheManager.testCacheNames
com.hazelcast.spring.cache.TestCacheManager.test
```

jerrinot:
verify

hazelcast/src/main/java/com/hazelcast/query/impl/getters/NullGetter.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.query.impl.getters;
+
+/**
+ * Null getter provides neutral-null behaviour for a {@link Getter}
+ */
+public final class NullGetter extends Getter {
```
pveentjer: Can you add some tests for NullGetter.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/ReflectionHelper.java
```diff
@@ -62,6 +64,10 @@ private ReflectionHelper() {
     }
```
pveentjer: Can you add tests for the changes made in the ReflectionHelper.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/NullGetter.java
```diff
@@ -0,0 +1,47 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.query.impl.getters;
+
+/**
+ * Null getter provides neutral-null behaviour for a {@link Getter}
```
pveentjer: What is null neutral behaviour. Can you elaborate.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/NullGetter.java
ahmetmircik: done

PR #5124, opened 2015-04-21
Added max result size limit for query operations.

Donnerbart:
This is the actual implementation of the "unbounded return values" feature for HZ 3.5.

Most of the code is new test code. I hope I covered all needed classes and scenarios. I have no idea how to test the new client. Maybe we have to do a follow-up PR with some changes if the new client is fully wired and in place.

There is also a big issue with `ClientMapProxy.entrySet()` which seems to have a O(n²) complexity, so it takes too long to test it at the moment. I will create a separate issue for that, please let's not discuss that in this PR. It is not an issue of this feature or these changes. It's most probably a very old problem. Issue: #5130

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -17,7 +17,11 @@
 package com.hazelcast.instance;
 
 import com.hazelcast.config.Config;
+import com.hazelcast.core.IMap;
 import com.hazelcast.internal.monitors.HealthMonitorLevel;
+import com.hazelcast.map.QueryResultSizeExceededException;
+import com.hazelcast.map.impl.QueryResultSizeLimiter;
+import com.hazelcast.query.TruePredicate;
 
 /**
  * The GroupProperties contain the Hazelcast properties. They can be set as an environmental variable, or
@@ -276,6 +280,41 @@ public class GroupProperties {
      */
     public static final String PROP_JCACHE_PROVIDER_TYPE = "hazelcast.jcache.provider.type";
 
+    /**
+     * Result size limit for query operations on maps.
+     * <p/>
+     * This value defines the maximum number of returned elements for a single query result. If a query exceeds this number of
+     * elements a {@link QueryResultSizeExceededException} will be thrown.
+     * <p/>
+     * This feature is in place to prevent an OOME if a single node is requesting the whole data set of the cluster, e.g. by
+     * executing a query with {@link TruePredicate}. This applies internally for the {@link IMap#values()}, {@link IMap#keySet()}
+     * and {@link IMap#entrySet()} methods, which are good candidates for OOME in large clusters.
+     * <p/>
+     * This feature depends on an equal distribution of the data on the cluster nodes to calculate the result size limit per node.
+     * Therefore there is a minimum value of {@value QueryResultSizeLimiter#MINIMUM_MAX_RESULT_LIMIT} defined in
+     * {@link QueryResultSizeLimiter}. Configured values below the minimum will be increased to the minimum.
+     * <p/>
+     * The feature can be disabled by setting a value of <tt>-1</tt> (which is the default value).
+     */
+    public static final String PROP_QUERY_RESULT_SIZE_LIMIT = "hazelcast.query.result.size.limit";
+
+    /**
+     * Maximum value of local partitions to trigger local pre-check for {@link TruePredicate} query operations on maps.
+     * <p/>
+     * To limit the result size of a query ({@see PROP_QUERY_RESULT_SIZE_LIMIT}) a local pre-check on the requesting node can be
+     * done, before the query is sent to the cluster. Since this may increase the latency the pre-check is limited to a maximum
+     * number of local partitions.
+     * <p/>
+     * By increasing this parameter you can prevent the execution of the query on the cluster by increasing the latency due to the
+     * prolonged local pre-check.
+     * <p/>
+     * The pre-check can be disabled by setting a value of <tt>-1</tt>.
+     *
+     * @see #PROP_QUERY_RESULT_SIZE_LIMIT
+     */
+    public static final String PROP_QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK
+            = "hazelcast.query.max.local.partition.limit.for.precheck";
+
     public final GroupProperty CLIENT_ENGINE_THREAD_COUNT;
 
     public final GroupProperty PARTITION_OPERATION_THREAD_COUNT;
@@ -450,9 +489,10 @@ public class GroupProperties {
     public final GroupProperty BACKPRESSURE_SYNCWINDOW;
     public final GroupProperty BACKPRESSURE_BACKOFF_TIMEOUT_MILLIS;
     public final GroupProperty BACKPRESSURE_MAX_CONCURRENT_INVOCATIONS_PER_PARTITION;
-    /**
-     * @param config
-     */
+
+    public final GroupProperty QUERY_RESULT_SIZE_LIMIT;
+    public final GroupProperty QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK;
+
     public GroupProperties(Config config) {
         HEALTH_MONITORING_LEVEL
                 = new GroupProperty(config, PROP_HEALTH_MONITORING_LEVEL, HealthMonitorLevel.SILENT.toString());
@@ -565,6 +605,10 @@ public class GroupProperties {
                 = new GroupProperty(config, PROP_BACKPRESSURE_MAX_CONCURRENT_INVOCATIONS_PER_PARTITION, "100");
         BACKPRESSURE_BACKOFF_TIMEOUT_MILLIS
                 = new GroupProperty(config, PROP_BACKPRESSURE_BACKOFF_TIMEOUT_MILLIS, "60000");
+
+        QUERY_RESULT_SIZE_LIMIT = new GroupProperty(config, PROP_QUERY_RESULT_SIZE_LIMIT, "-1");
+        QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK
+                = new GroupProperty(config, PROP_QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK, "3");
     }
 
     public static class GroupProperty {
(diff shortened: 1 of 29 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -375,6 +376,8 @@ public void run() {
                 } else {
                     handleAuthenticationFailure(endpoint, request);
                 }
+            } catch (QueryResultSizeExceededException e) {
```
Donnerbart: I'm not 100% sure about this part. I'm just leaving out to log the exception here, which I think should be fine. We don't log this exception at any other proxy. It is meant for the customer who called the method, like a back pressure exception. We should not put a whole stack trace in the log file IMHO.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
```diff
@@ -102,6 +103,11 @@ private void collectResults(Collection<QueryResultEntry> result, List<Future> fu
         for (Future future : futures) {
             QueryResult queryResult = (QueryResult) future.get();
             if (queryResult != null) {
+
+                if (queryResult.isResultLimitExceeded()) {
```
Donnerbart: I could not test this yet since as far as I saw these tasks are not yet wired in the new client module. But this should hopefully be the right place to fire the exception.

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainerSupport.java
```diff
@@ -26,7 +26,7 @@
  *
  * @see MapContainer
  */
-abstract class MapContainerSupport {
+public abstract class MapContainerSupport {
```
Donnerbart: I had to make this class `public` since I was forbidden to mock a method from it otherwise.

pveentjer: No problem. 

I don't care of stuff in .impl package is public or not.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
pveentjer: The ClientEngineImpl should not deal with this responsibility. ClientEngineImpl deals with client invocations/responses.. it doesn't deal with maps or queries.

hazelcast/src/main/java/com/hazelcast/map/QueryResultSizeExceededException.java
```diff
@@ -0,0 +1,54 @@
...
+import static java.lang.String.format;
+
+/**
+ * This exception is thrown when a query exceeds a configurable result size limit.
+ *
+ * @see com.hazelcast.instance.GroupProperties#PROP_QUERY_RESULT_SIZE_LIMIT
+ */
+public class QueryResultSizeExceededException extends HazelcastException {
+
+    public QueryResultSizeExceededException() {
```
pveentjer: Is this constructing being called?

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
Donnerbart: In one of my tests I saw the `QueryResultSizeExceededException` being logged by that catch. I'll have a another look which test that was.

hazelcast/src/main/java/com/hazelcast/map/QueryResultSizeExceededException.java
Donnerbart: Yes, all of them are used.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -17,7 +17,11 @@
 package com.hazelcast.instance;
 
 import com.hazelcast.config.Config;
+import com.hazelcast.core.IMap;
 import com.hazelcast.internal.monitors.HealthMonitorLevel;
+import com.hazelcast.map.QueryResultSizeExceededException;
+import com.hazelcast.map.impl.MapQueryResultSizeLimitHelper;
```
pveentjer: Why don't you call this class:

MapQueryResultSizeLimitHelper ----> MapQueryResultSizeLimiter.

Every object helps out (or should be deleted :) ).

Donnerbart: Good name, thanks!

jerrinot: +1

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResult.java
```diff
@@ -30,29 +32,36 @@
...
 
-    private final Collection<QueryResultEntry> result;
+    private final Collection<QueryResultEntry> result = new LinkedHashSet<QueryResultEntry>();
 
     private Collection<Integer> partitionIds;
 
+    private transient long resultLimit;
+    private transient long resultSize;
+
     public QueryResult() {
```
pveentjer: I try to go through 1 constructor. So I would call this(Long.MAX_VALUE). Having a single true constructor (and the rest convenience calls) reduces duplication (and bugs)

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResult.java
```diff
@@ -30,29 +32,36 @@
...
     }
 
-    public Collection<Integer> getPartitionIds() {
-        return partitionIds;
+    public void addAll(Collection<QueryableEntry> queryableEntries) {
+        for (QueryableEntry entry : queryableEntries) {
+            if (++resultSize > resultLimit) {
+                throw new QueryResultSizeExceededException();
+            }
+            result.add(new QueryResultEntryImpl(entry.getKeyData(), entry.getIndexKey(), entry.getValueData()));
```
pveentjer: I'm always careful with adding complex statements with deeper nesting. I personally prefer something like this:

```
QueryResultEntryImpl queryEntry = new QueryResultEntryImpl(
      entry.getKeyData(), entry.getIndexKey(), entry.getValueData())
result.add(queryEntry);
```

A bit more lines but less complex statements to deal with.

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResultSizeLimiter.java
```diff
@@ -0,0 +1,152 @@
...
+
+import com.hazelcast.instance.GroupProperties;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.map.QueryResultSizeExceededException;
+import com.hazelcast.spi.NodeEngine;
+
+import java.util.Collection;
+
+/**
+ * This class contains the main functionality to limit the result size of queries.
```
pveentjer: I would not call it 'this class'. Because then you would see that at every interface/class.

"Responsible for limiting result size of queries."

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResultSizeLimiter.java
```diff
@@ -0,0 +1,152 @@
...
+ */
+public class QueryResultSizeLimiter {
+
+    // due to the unequal data distribution we set a minimum value for the trigger to ensure a better distribution
+    public static final int MINIMUM_MAX_RESULT_LIMIT = 100000;
+
+    // due to the unequal data distribution we add a security margin to prevent false positives
```
pveentjer: few more lines of comments for future readers why this matters.

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResultSizeLimiter.java
```diff
@@ -0,0 +1,152 @@
...
+
+    void checkMaxResultLimitOnLocalPartitions(String mapName) {
+        // check if feature is enabled
+        if (!isPreCheckEnabled) {
+            return;
+        }
+
+        // limit number of local partitions to check to keep runtime constant
+        Collection<Integer> localPartitions = mapServiceContext.getOwnedPartitions();
+        int partitionsToCheck = Math.min(localPartitions.size(), maxLocalPartitionsLimitForPreCheck);
```
pveentjer: static import

hazelcast/src/test/java/com/hazelcast/map/impl/operation/QueryPartitionOperationTest.java
```diff
@@ -0,0 +1,52 @@
...
+public class QueryPartitionOperationTest extends QueryOperationTestSupport {
+
+    @Override
+    protected AbstractMapOperation createQueryOperation() {
+        return new QueryPartitionOperation(MAP_NAME, TruePredicate.INSTANCE);
+    }
+
+    @Test
+    public void testConstructor() {
+        new QueryPartitionOperation();
```
pveentjer: This doesn't offer a lot of value. You not doing any validation on the fields. This can be relevant in some cases. So either validate the fields or remote the test method

Donnerbart: I see that very differently:
* It provided the only usage in code for the default constructor, so there is no need for a `@SuppressWarnings("unused")` in the class.
* It ensures that the default constructor doesn't get deleted, because it looks unused, which will cause runtime exceptions, not compile time exceptions.
* It increases code coverage with a unit test (not all tests are executed in all builds, especially the slow nightly integration tests which will use the constructor).

hazelcast/src/main/java/com/hazelcast/map/impl/QueryResult.java
Donnerbart: Resolved but won't disappear since unchanged line was commented.

hazelcast/src/test/java/com/hazelcast/map/impl/operation/QueryPartitionOperationTest.java
pveentjer: In case of an operation, I always make sure that the operation gets deserialized. This triggers the serialization/deserialization + default constructor.

pveentjer: Anyhow.. I'm not going to make a fuss about it.

pveentjer:
+1 I'm fine with merging this feature.

PR #5135, opened 2015-04-22
Small improvements for SlowOperationDetector

### LINKED ISSUE #5043: SlowOperationDetector disable logging option

Donnerbart:
Added hint in first log of SlowOperationDetector.
Added a missing test for the property to disable the feature.
Added the stack trace config property to some tests to increase code coverage.

Fixes #5043.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
@@ -62,6 +62,8 @@ public final class SlowOperationDetector {
 
     private final DetectorThread detectorThread;
 
+    private boolean isFirstLog = true;
+
     @edu.umd.cs.findbugs.annotations.SuppressWarnings({"EI_EXPOSE_REP2" })
     public SlowOperationDetector(LoggingService loggingServices,
                                  OperationRunner[] genericOperationRunners,
@@ -133,6 +135,7 @@ public final class SlowOperationDetector {
             super(threadGroup.getInternalThreadGroup(), threadGroup.getThreadNamePrefix("SlowOperationDetectorThread"));
         }
 
+        @Override
         public void run() {
             long lastLogPurge = System.nanoTime();
             while (running) {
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
```diff
@@ -43,6 +43,8 @@
     private static final long ONE_SECOND_IN_NANOS = TimeUnit.SECONDS.toNanos(1);
     private static final long SLOW_OPERATION_THREAD_MAX_WAIT_TIME_TO_FINISH = TimeUnit.SECONDS.toMillis(10);
 
+    private static boolean isFirstLog = true;
```
pveentjer: Perhaps better not to make this static. Let each HZ instance tell it once. Also if it is static, you have threading issues here (no happens before relation).

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetectorBasicTest.java
```diff
@@ -46,6 +46,21 @@ public void teardown() {
     }
 
     @Test
+    public void testDisabled() {
+        Config config = new Config();
+        config.setProperty(GroupProperties.PROP_SLOW_OPERATION_DETECTOR_ENABLED, "false");
+        config.setProperty(GroupProperties.PROP_SLOW_OPERATION_DETECTOR_THRESHOLD_MILLIS, "1000");
+
+        instance = getSingleNodeCluster(config);
```
pveentjer: Why don't you make use of the HazelcastTestSupport.createHazelcastInstance method?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetectorBasicTest.java
```diff
@@ -46,6 +46,21 @@ public void teardown() {
...
+
+        getInternalOperationService(instance).execute(new SlowRunnable(3, Operation.GENERIC_PARTITION_ID));
```
pveentjer: Make use of the HazelcastTestSupport.getOperationService method which returns the InternalOperationService.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetectorBasicTest.java
Donnerbart: I am, you are just missing context:

```
    HazelcastInstance getSingleNodeCluster(Config config) {
        MapConfig mapConfig = new MapConfig();
        mapConfig.setName("slowOperation*");
        config.addMapConfig(mapConfig);

        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(1);
        return factory.newHazelcastInstance(config);
    }
```

pveentjer:
Also add @Override on your Thread.run method.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetectorBasicTest.java
pveentjer: Damn github reviews

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetectorAbstractTest.java
```diff
@@ -113,10 +101,6 @@ static void shutdownOperationService(HazelcastInstance instance) {
         return invocationMap.values();
     }
 
-    static InternalOperationService getInternalOperationService(HazelcastInstance instance) {
-        return TestUtil.getNode(instance).nodeEngine.getOperationService();
-    }
-
     static int getDefaultPartitionId(HazelcastInstance instance) {
```
pveentjer: I find default partition id extremely confusing. Also 'default key' is very confusing.

Donnerbart: Please have a look in IDEA, that method is already in master. It may be confusing because it is not 100% easy what I have to do there. I totally don't care about keys and values in my tests. I just need the partitionId of the one entry which is in the map for a nested operation with different partitionIds.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/slowoperationdetector/SlowOperationDetector.java
```diff
@@ -231,22 +234,39 @@ private SlowOperationLog getOrCreate(String stackTrace, Object operation) {
...
+                logger.warning(format("Slow operation detected: %s (%d invocations)", log.operation, totalInvocations));
             }
-            logger.warning(format("Slow operation detected: %s (%d invocations)", log.operation, totalInvocations));
+        }
+
+        private void logWithConfigHint(SlowOperationLog log, int totalInvocations) {
+            // print a hint once how to enable logging of stack traces
+            logger.warning(format("Slow operation detected: %s"
+                            + "%nHint: You can enable the logging of stack traces with the following config property:"
+                            + " hazelcast.slow.operation.detector.stacktrace.logging.enabled",
```
pveentjer: I would use the GroupProperty for it. This way you don't run into problems when it is renamed.

Donnerbart:
verify

PR #5137, opened 2015-04-22
Call Home Enhancements

kobalski:
With this PR hazelcast nodes will send HDMemory configuration of the
cluster and also will send client information connected cluster.

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/ConnectedClientOperationTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/ConnectedClientOperationTest.java
@@ -1,0 +1,63 @@
+package com.hazelcast.client;
+
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.config.Config;
+import com.hazelcast.core.ClientType;
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.instance.Node;
+import com.hazelcast.instance.TestUtil;
+import com.hazelcast.test.HazelcastSerialClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(HazelcastSerialClassRunner.class)
+public class ConnectedClientOperationTest extends HazelcastTestSupport {
+
+    @BeforeClass
+    @AfterClass
+    public static void cleanupClass() {
+        Hazelcast.shutdownAll();
+    }
+
+    @Before
+    @After
+    public void cleanup() {
+        Hazelcast.shutdownAll();
+    }
+
+    @Test
+    public void testNumberOfConnectedClients() throws Exception{
+
+        Config config = new Config();
+        HazelcastInstance h1 = Hazelcast.newHazelcastInstance(config);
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1287,6 +1292,59 @@ private void setMembersRef(Map<Address, MemberImpl> memberMap) {
...
+        for (MemberImpl member : node.getClusterService().getMemberList()) {
+            Address target = member.getAddress();
+            InvocationBuilder invocationBuilder = operationService1.createInvocationBuilder(serviceName,
+                    clientInfoOperation, target);
+            Future<Object> future = invocationBuilder.setTryCount(1).invoke();
+            HashMap<String, ClientType> endpoints = new HashMap<String, ClientType>();
+            try {
+                endpoints = (HashMap<String, ClientType>)future.get();
+            } catch (InterruptedException e) {
+                logger.warning("Cannot get client information from : " + target.toString());
```
mesutcelik: so we swallow the exception if we don't get any info from some of the members. @sancar any suggestion here?

hazelcast/src/main/java/com/hazelcast/util/VersionCheck.java
```diff
@@ -154,13 +145,9 @@ private Document fetchWebService(String urlStr) {
...
             conn.setReadTimeout(TIMEOUT * 2);
             in = new BufferedInputStream(conn.getInputStream());
-            final DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
-            return builder.parse(in);
-        } catch (SAXException ignored) {
-            EmptyStatement.ignore(ignored);
-        } catch (IOException ignored) {
-            EmptyStatement.ignore(ignored);
-        } catch (ParserConfigurationException ignored) {
+            IOUtil.closeResource(in);
```
mesutcelik: that is already called in finally block.

hazelcast/src/main/java/com/hazelcast/util/VersionCheck.java
```diff
@@ -154,13 +145,9 @@ private Document fetchWebService(String urlStr) {
...
+            return null;
```
mesutcelik: I don't think return value is used by anyone so we can change the return type of the method to ```void```

hazelcast/src/main/java/com/hazelcast/util/VersionCheck.java
```diff
@@ -168,4 +155,29 @@ private Document fetchWebService(String urlStr) {
 
         return null;
```
mesutcelik: we can remove this.

hazelcast-client-new/src/test/java/com/hazelcast/client/ConnectedClientOperationTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(HazelcastSerialClassRunner.class)
+public class ConnectedClientOperationTest extends HazelcastTestSupport {
```
bilalyasar: minor comment, missing category again

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
kobalski: Increased tryCount to 50.

serkan-ozal:
:+1:

PR #5145, opened 2015-04-22
Renamed MapServiceContextAwareFactory to MapServiceFactory

pveentjer:
MapServiceContextAwareFactory is a confusing name. It seems it is a Factory to create MapServiceContextAware instances. It was really putting me on the wrong foot.  The only thing this factory produces are MapService instances.

Factories often get dependency injected; but this is not part of the name. The main argument to determine the name of a factory is what they produce, not how they produce it. 

Therefore 'ContextAware' doesn't belong in the name of the factory.


ahmetmircik:
Seems mergeable.

PR #5146, opened 2015-04-22
MapServiceContext simplifications

pveentjer:
Folded MapServiceContextSupport into MapServiceContext
Folded AbstractMapServiceContextSupport into MapServiceContextImpl

I folded super interface and super classes. The reason these classes were introduced was to reduce the number of methods I and data coupling issues. The actual problem is that MapServiceContext has too many responsibilities, that is why it has many methods and many dependencies.

So this is a first step: make the problem visible and the code way less confusing. Checkstyle supression for method count and data coupling have been added.
Next step is to truly fix the problem: analyse which methods truly belong in MapServiceContext, and methods that don't (e.g. toObject,toData etc)


checkstyle/suppressions.xml
```diff
@@ -540,6 +540,8 @@
     <suppress checks="ClassFanOutComplexityCheck" files="com.hazelcast.map.impl.proxy.MapProxySupport"/>
     <suppress checks="ClassDataAbstractionCoupling" files="com.hazelcast.map.impl.proxy.MapProxySupport"/>
     <suppress checks="CyclomaticComplexity" files="com.hazelcast.map.impl.client.AbstractTxnMapRequest"/>
+    <suppress checks="MethodCount|ClassDataAbstractionCoupling" files="com.hazelcast.map.impl.MapServiceContextImpl"/>
+    <suppress checks="MethodCount" files="com.hazelcast.map.impl.MapServiceContext"/>
```
ajermakovics: I guess what's confusing is that the title says simplifications but then there's checkstyle suppressions

pveentjer: The problem is that the complexity was there, but it was hidden. To deal with too many methods and data coupling issues, super classes and super interfaces were introduced. 

In case of the MapServiceContext is only made the whole thing a lot more complex because you now need to deal with a lot more twice the number of structures and half of them they can't carry their weight.

So the simplification is that these unwanted interface/abstract-class folded into the place where they should have been. And yes.. now you get the checkstyle issue back.

The next step is to figure out how to correctly reduce complexity.. and not by applying just some cosmetic changes.

pveentjer:
verify

pveentjer:
verify

checkstyle/suppressions.xml
pveentjer: Probably the MapServiceContext interface can be collapsed in the MapServiceContextImpl and all the getters can be removed and just allow for public field access. Most of the methods on the MapServiceContext are just dependency getters.

ahmetmircik:
Seems mergeable but also check enterprise before merge.

PR #5147, opened 2015-04-22
MapContainerSupport Folded into MapContainer

pveentjer:
MapContainerSupport can't carry its weight. It increases code complexity and doesn't serve a real purpose from a reusability or abstraction point of view.


pveentjer:
verify

pveentjer:
verify

Donnerbart:
:+1:

Makes it much more easy to mock the `MapContainer`.

ahmetmircik:
seems ok.

PR #5149, opened 2015-04-22
Preconditions cleanup: round 1

pveentjer:
Renamed ValidationsUtils to Preconditions. 
Renamed most of the methods to begin with check.

This is in line with how Preconditions are being used in Guava:

http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/base/Preconditions.html

More cleanup is required. We should take a choice if we want to pass in the argument name or if we want to pass in the full message. 

We also need to see what we are going to do with :
* isNotNull --> throws IllegalArgumentException + gets argument name
* checkNotNull --> throws NullPointerException + full message

At some placed isNotNull is being used with an argname and not a message. 

So having these similar methods in the system is confusing and error prone.


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
```diff
@@ -734,7 +647,7 @@ public ICompletableFuture submitToKey(K key, EntryProcessor entryProcessor) {
                                                     JobTracker jobTracker) {
 
         try {
-            ValidationUtil.isNotNull(jobTracker, "jobTracker");
+            Preconditions.isNotNull(jobTracker, "jobTracker");
```
Donnerbart: Could be a static import like the other calls.

pveentjer: I guess that one slipped through.

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
```diff
@@ -315,12 +319,9 @@ public void delete(Object key) {
...
     public Set keySet(Predicate predicate) {
         checkTransactionState();
-        if (predicate == null) {
-            throw new NullPointerException("Predicate should not be null!");
-        }
-        if (predicate instanceof PagingPredicate) {
-            throw new NullPointerException("Paging is not supported for Transactional queries!");
-        }
+        checkNotNull(predicate, "Predicate should not be null!");
+        checkNotInstanceOf(PagingPredicate.class, predicate, "Paging is not supported for Transactional queries!");
```
Donnerbart: This method throws an `IllegalArgumentException` instead of `NullPointerException` before. But this is not considered as API change, isn't it? Maybe a map test will fail, but we'll see :)

hazelcast/src/main/java/com/hazelcast/memory/MemorySize.java
```diff
@@ -37,15 +39,14 @@ public MemorySize(long value, MemoryUnit unit) {
         if (value < 0) {
```
Donnerbart: You may replace this with a `checkNotNegative()` method.

hazelcast/src/main/java/com/hazelcast/multimap/impl/ObjectMultiMapProxy.java
```diff
@@ -308,7 +308,7 @@ public LocalMultiMapStats getLocalMultiMapStats() {
                                                     JobTracker jobTracker) {
 
         try {
-            ValidationUtil.isNotNull(jobTracker, "jobTracker");
+            Preconditions.isNotNull(jobTracker, "jobTracker");
```
Donnerbart: Should be a static import.

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
jerrinot: it definitely *is* API change. You can see an Exception as an alternative return type.

hazelcast/src/main/java/com/hazelcast/spi/impl/proxyservice/impl/ProxyServiceImpl.java
```diff
@@ -151,7 +143,7 @@ public void destroyDistributedObject(String serviceName, String name) {
         waitWithDeadline(calls, TIME, TimeUnit.SECONDS, DESTROY_PROXY_EXCEPTION_HANDLER);
     }
 
-    public void destroyLocalDistributedObject(String serviceName, String name, boolean fireEvent) {
+     public void destroyLocalDistributedObject(String serviceName, String name, boolean fireEvent) {
```
Donnerbart: It looks like this leading space character is superfluous.

hazelcast/src/main/java/com/hazelcast/util/Preconditions.java
```diff
@@ -82,40 +82,69 @@ public static String hasText(String argument, String argName) {
...
     /**
-     * Tests if a long value is positive.
+     * Tests if a long value is not negative.
      *
-     * @param value        the long value tested to see if it is positive.
+     * @param value   the long value tested to see if it is not negative.
+     * @param message the message
+     * @throws java.lang.IllegalArgumentException if the value is negative.
+     */
+    public static void checkNegative(long value, String message) {
```
Donnerbart: This method is new, isn't it? Can't see a test for on GitHub.

Donnerbart:
It's not a new issue due to this PR, but I find the coexistence of `checkNotNull()` and `isNotNull()` a bit confusing. When to use which method? One produces a NPE the other an IllegalArgumentException. `isNotNull()` also creates an error message for you. Nevertheless the method names don't help me much to choose the right one.

Also the JavaDoc for `isNotNull()` is wrong, the mentioned `string` should be called `argument`, since it uses generics.

Donnerbart:
My eyes are a bit blurry after scrolling through this PR, but beside those minor issues everything looks fine :+1:

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
pveentjer: Yes.. I have not touched any of the behaviour. What is NPE --> NPE.. what is IllegalArgumentException --> remains IllegalArgumentException. 

That is why this is phase 1..

pveentjer:
The checkNotNull and the isNotNull thing sucks.. i want to move that to phase2.. we need to decide what we are going to do with isNotNull

hazelcast/src/main/java/com/hazelcast/spi/impl/proxyservice/impl/ProxyServiceImpl.java
pveentjer: I guess my ide messed up. I also don't like noise space.

hazelcast/src/main/java/com/hazelcast/memory/MemorySize.java
pveentjer: I don't want to do that yet because of the string creation issue.

hazelcast/src/main/java/com/hazelcast/multimap/impl/ObjectMultiMapProxy.java
pveentjer: Done

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
jerrinot: actually as @Donnerbart spotted here is a behaviour change: `checkNotInstanceOf` throws IllegalArgumentEx, but for some reason the original code is
```java
if (predicate instanceof PagingPredicate) {
            throw new NullPointerException("Paging is not supported for Transactional queries!");
}
```

Donnerbart: Not in this case. It was NPE, but the method `checkNotInstanceOf()` seems to throw `IllegalArgumentException`.

pveentjer: I think the original code can be seen as a bug since the throw exception is not correct. So a bug fix would be to make it an IllegalArgumentException

jerrinot: you are right it's probably a bug. 
I would still consider creating an extra PR to avoid mixing pure cleanups and behaviour changes.

hazelcast/src/main/java/com/hazelcast/memory/MemorySize.java
Donnerbart: Good point!

GitHub should have a 'resolved' feature for line comments, where no change is done eventually.

Donnerbart:
I think you created some new methods, but I can't see tests for them. Are the other methods tested?

pveentjer:
Updates; now with everything being tested.

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
pveentjer: I think that creating a PR for something I would hit the merge button for immediately, is too much overhead.

pveentjer:
verify

pveentjer:
Can someone have another look at it? Would like to get it merged.

jerrinot:
I'd give it :+1: if there wasn't the behaviour change @Donnerbart discovered.

pveentjer:
If you throw a NPE if case of a type violation issue than that is a bug. The original behaviour was broken in the first place. I have fixed that.

pveentjer:
I have reintroduced the bug to keep people happy:
https://github.com/hazelcast/hazelcast/commit/cd6ea02541dca1cf164efa4dc6c4d63396e27ee2

PR #5150, opened 2015-04-22
Minor MapmanagedService cleanup

pveentjer:
Pulled out ObjectNamespaceLockStoreInfoConstructorFunction.

This class was an anonymous inner class with had inside an anonymous inner class. Was very hard to see what was happening. So the outer anonymous inner class is pulled out.


Donnerbart:
:+1:

pveentjer:
verify

ahmetmircik:
:+1:

PR #5152, opened 2015-04-23
Renamed MapEventPublisherSupport to MapEventPublisherImpl

pveentjer:
Renamed MapEventPublisherSupport to MapEventPublisherImpl. 

Calling an implementation 'Support' is not a convention in Hazelcast. It is either Default... or ...Impl. 

If there is a support class then it implement some of the methods.. but it isn't a full implementation. Or it contains some reusable methods.

E.g.:
ClientInvocationServiceSupport
ClusterListenerSupport

Within the com.hazelcast.map package support is being used for unexpected reasons. There are also interfaces that contain Support e.g. 

MapServiceContextEventListenerSupport
MapServiceContextInterceptorSupport
MapServiceContextSupport
MapContextQuerySupport

Having a name that doesn't clearly explain its intent makes code hard to understand because you can't guess what it does.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapEventPublisher.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapEventPublisher.java
@@ -24,7 +24,7 @@ import com.hazelcast.nio.serialization.Data;
 /**
  * Helper methods for publishing events.
  *
- * @see MapEventPublisherSupport
+ * @see MapEventPublisherImpl
  */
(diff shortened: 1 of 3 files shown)


ahmetmircik:
Please check out enterprise for possible implications of this PR other than that seems mergeable.

pveentjer:
Thanks for reviewing. Appreciate it. I have dealt with all the PR's and have updated enterprise accordingly.

PR #5155, opened 2015-04-23
Selector Hack Removed from master

jerrinot:
Fixing #5072

Selector Hack is no longer needed as IOBalancer deals with
selector imbalance situations.


pveentjer:
Will merge once green.

PR #5160, opened 2015-04-24
XA Improvements

gurbuzali:
XA Transactions decoupled from normal transactions by adding new implementations of TransactionContext and Transaction interfaces
Fixed/improved recovery mechanism of XA Transactions to be XA specification compliant
Added Specification Compatibility tests both for server and client side


gurbuzali:
some test failures
```
    com.hazelcast.client.protocol.RawProtocolTest.testAuthenticateWithUsernameAndPassword
    com.hazelcast.cluster.TcpIpJoinTest.test_whenIncompatibleJoiners
```

gurbuzali:
verify

gurbuzali:
need to fix enterprise after merging this PR, due to newly added method to the `HazelcastInstance` interface

noctarius:
This thing is huge, it will be a hard part to review this and follow what is going on.

hazelcast-ra/hazelcast-jca/src/main/java/com/hazelcast/jca/HazelcastConnectionImpl.java
```diff
@@ -336,6 +337,13 @@ public Endpoint getLocalEndpoint() {
 
     // unsupported operations
 
+
+    @Override
+    public HazelcastXAResource getXAResource() {
+        //TODO
```
gurbuzali: todo

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/transaction/TransactionCommitMessageTask.java
```diff
@@ -39,12 +37,12 @@ public TransactionCommitMessageTask(ClientMessage clientMessage, Node node, Conn
     @Override
     protected ClientMessage innerCall() throws Exception {
         TransactionContext transactionContext = endpoint.getTransactionContext(parameters.transactionId);
-        if (parameters.prepareAndCommit) {
-            transactionContext.commitTransaction();
-        } else {
-            Transaction transaction = TransactionAccessor.getTransaction(transactionContext);
-            transaction.commit();
-        }
+//        if (parameters.prepareAndCommit) {
```
gurbuzali: need to remove commented out code

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/transaction/TransactionPrepareMessageTask.java
```diff
@@ -40,8 +38,8 @@ public TransactionPrepareMessageTask(ClientMessage clientMessage, Node node, Con
     protected ClientMessage innerCall() throws Exception {
 
         TransactionContext transactionContext = endpoint.getTransactionContext(parameters.transactionId);
-        Transaction transaction = TransactionAccessor.getTransaction(transactionContext);
-        transaction.prepare();
+//        Transaction transaction = TransactionAccessor.getTransaction(transactionContext);
```
gurbuzali: remove

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XATransactionContextImpl.java
```diff
@@ -0,0 +1,144 @@
...
+            = new HashMap<TransactionalObjectKey, TransactionalObject>(2);
+
+    public XATransactionContextImpl(NodeEngineImpl nodeEngine, Xid xid, String txOwnerUuid, int timeout) {
+        this.nodeEngine = nodeEngine;
+        this.transaction = new XATransactionImpl(nodeEngine, xid, txOwnerUuid, timeout);
+    }
+
+    @Override
+    public void beginTransaction() {
+        throw new UnsupportedOperationException("XA Transaction cannot be started manually!!!");
```
gurbuzali: add tests for manually beging/commit/rollback

gurbuzali: don't forget client-side

hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
```diff
@@ -97,7 +97,7 @@ public AbstractHazelcastClassRunner(Class<?> klass, Object[] parameters,
     protected Statement withAfters(FrameworkMethod method, Object target,
                                    Statement statement) {
         List<FrameworkMethod> afters = getTestClass().getAnnotatedMethods(After.class);
-        if (!DISABLE_THREAD_DUMP_ON_FAILURE) {
```
gurbuzali: leftover, should be removed

hazelcast/src/main/java/com/hazelcast/transaction/client/RollbackXATransactionRequest.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public RollbackXATransactionRequest(String txnId) {
+        super(txnId);
+    }
+
+    @Override
+    public Object call() throws Exception {
+        Transaction transaction = getTransaction();
+        transaction.rollback();
```
gurbuzali: need to remove context `endpoint.removeTransactionContext(txnId);`

juanavelez:
In hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XAResourceImpl.java
 I think there is something wrong with start() when resume/join:

1. getTransactionContext() never returns a null but throws an exception when the thread is not associated to an existing tx context, so the code after it will never create a new tx context.

2. Also, end() assumes that one or more threads can start different contexts, but start() resume/join does not assume this, it only allows the thread that started the transaction to suspend or join. One of the issues I have noticed while running HZ 3.4 TXs with JBoss EAP 6.4.0 and CMT is that JBoss TS can use multiple threads to handle the same transaction

gurbuzali:
Hi @juanavelez thank you for pointing out, I've fixed the `start` part which should solve the issue

gurbuzali:
verify

gurbuzali:
verify

juanavelez:
@gurbuzali: Are there going to be any changes to the JCA XAResouce side as well? hazelcast-ra/hazelcast-jca/src/main/java/com/hazelcast/jca/XAResourceWrapper.java. I think this also needs to be re-written based on your latest design. Thanks.

hazelcast-client-new/src/main/java/com/hazelcast/client/txn/proxy/xa/XAResourceProxy.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Override
+    public TransactionContext getTransactionContext() {
+        long threadId = Thread.currentThread().getId();
+        return threadContextMap.get(threadId);
+    }
+
+    @Override
+    public synchronized void start(Xid xid, int flags) throws XAException {
```
mdogan: I didn't quite understand the internals here, but is it really required to make all these methods `synchronized`?

gurbuzali: actually no, they are leftovers of the first implementation (not even the previous implementation)

hazelcast-client-new/src/main/java/com/hazelcast/client/txn/proxy/xa/XAResourceProxy.java
```diff
@@ -0,0 +1,239 @@
...
+            return true;
+        }
+        String otherGroupName = null;
+        if (xaResource instanceof XAResourceProxy) {
+            otherGroupName = ((XAResourceProxy) xaResource).getGroupName();
+        }
+        if (xaResource instanceof XAResourceImpl) {
+            otherGroupName = ((XAResourceImpl) xaResource).getGroupName();
+        }
+        return getGroupName().equals(otherGroupName);
```
mdogan: As far as I remember, client doesn't have to use `group-name` as its login credentials.

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/ClientTransactionManagerService.java
```diff
@@ -0,0 +1,41 @@
...
+
+    <T> T executeTransaction(TransactionalTask<T> task);
+
+    <T> T executeTransaction(TransactionOptions options, TransactionalTask<T> task);
+
+    TransactionContext newTransactionContext();
+
+    TransactionContext newTransactionContext(TransactionOptions options);
+
+    TransactionContext newXaTransactionContext(Xid xid, int defaultTimeout);
```
mdogan: What's the unit of `defaultTimeout`?

gurbuzali: it is `seconds` (XA Specification), we can change the variable to `timeoutInSeconds`

hazelcast-client-new/src/main/java/com/hazelcast/client/txn/proxy/xa/XAResourceProxy.java
gurbuzali: There is a problem then, we use `groupName` as a unique id of cluster.
what can we use instead of `groupName` ?

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/ClientTransactionManagerService.java
mdogan: You should also document this in method javadoc...

gurbuzali: will do

hazelcast-client-new/src/main/java/com/hazelcast/client/txn/proxy/xa/XATransactionProxy.java
```diff
@@ -0,0 +1,151 @@
...
+    public XATransactionProxy(HazelcastClientInstanceImpl client, ClientConnection connection, Xid xid, int timeout) {
+        this.client = client;
+        this.connection = connection;
+        this.timeout = timeout;
+        this.xid = new SerializableXID(xid.getFormatId(), xid.getGlobalTransactionId(), xid.getBranchQualifier());
+    }
+
+    void begin() {
+        try {
+            if (state == ACTIVE) {
```
mdogan: Is `XATransactionProxy` used by a single thread only?

gurbuzali: no, can be commit/rollbacked from a different thread, same for `XATransactionImpl`.
They both needs concurrency controls.

hazelcast-client-new/src/main/java/com/hazelcast/client/txn/proxy/xa/XAResourceProxy.java
mdogan: Even if client uses `group-name`, it doesn't have to be unique. One can create different clusters with different passwords. Even one can create different clusters with same group/password using different member lists.

hazelcast-client-new/src/main/java/com/hazelcast/client/txn/proxy/xa/XATransactionProxy.java
mdogan: so, what will happen when multiple threads call `begin()/end()/commit()` concurrently?

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XAResourceImpl.java
```diff
@@ -0,0 +1,303 @@
...
+    private final String groupName;
+
+    public XAResourceImpl(NodeEngine nodeEngine, XAService service) {
+        super(nodeEngine, service);
+        GroupConfig groupConfig = nodeEngine.getConfig().getGroupConfig();
+        groupName = groupConfig.getName();
+    }
+
+    @Override
+    public synchronized void start(Xid xid, int flags) throws XAException {
```
mdogan: same here; do all methods have to be `synchronized`? At least you should use an internal mutex, because instance of this class is exposed to the user, one can use it as his/her own monitor object... which can lead to an unexpected deadlock.

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XAResourceImpl.java
```diff
@@ -0,0 +1,303 @@
...
+        OperationService operationService = nodeEngine.getOperationService();
+        SerializableXID serializableXID =
+                new SerializableXID(xid.getFormatId(), xid.getGlobalTransactionId(), xid.getBranchQualifier());
+        Data xidData = nodeEngine.toData(serializableXID);
+        int partitionId = partitionService.getPartitionId(xidData);
+        FinalizeRemoteTransactionOperation operation = new FinalizeRemoteTransactionOperation(xidData, isCommit);
+        InternalCompletableFuture<Integer> future = operationService.invokeOnPartition(SERVICE_NAME, operation, partitionId);
+        Integer errorCode;
+        try {
+            errorCode = future.get();
```
mdogan: You're doing a blocking call under a `synchronized`block. Is it definitely required?

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XAResourceImpl.java
gurbuzali: will remove `synchronized`, not needed

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XAResourceImpl.java
gurbuzali: will remove `synchronized`

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XAResourceImpl.java
mdogan: then how will you guarantee thread-safety? As far as I understand these methods can be called concurrently..

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XATransactionContextImpl.java
```diff
@@ -0,0 +1,144 @@
...
+import com.hazelcast.transaction.TransactionalObject;
+import com.hazelcast.transaction.impl.Transaction;
+import com.hazelcast.transaction.impl.TransactionalObjectKey;
+
+import javax.transaction.xa.XAResource;
+import javax.transaction.xa.Xid;
+import java.util.HashMap;
+import java.util.Map;
+
+public class XATransactionContextImpl implements TransactionContext {
```
mdogan: Is this class used by only a single thread?

hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XATransactionImpl.java
```diff
@@ -0,0 +1,293 @@
...
+import static com.hazelcast.transaction.impl.Transaction.State.PREPARED;
+import static com.hazelcast.transaction.impl.Transaction.State.PREPARING;
+import static com.hazelcast.transaction.impl.Transaction.State.ROLLED_BACK;
+import static com.hazelcast.transaction.impl.Transaction.State.ROLLING_BACK;
+import static com.hazelcast.transaction.impl.xa.XAService.SERVICE_NAME;
+import static com.hazelcast.util.FutureUtil.RETHROW_TRANSACTION_EXCEPTION;
+import static com.hazelcast.util.FutureUtil.logAllExceptions;
+import static com.hazelcast.util.FutureUtil.waitWithDeadline;
+
+final class XATransactionImpl implements Transaction, TransactionSupport {
```
mdogan: Is this class used by only a single thread?

mdogan:
I think you should clearly document thread-safety requirements/features for each class like; `XATransactionImpl`, `XATransactionContextImpl`, `XAResourceImpl` etc.

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/ClientTransactionManagerService.java
pveentjer: Should it not be 'newXATransactionContext' ?

gurbuzali: yes

hazelcast-client-new/src/main/java/com/hazelcast/client/txn/TransactionProxy.java
```diff
@@ -97,53 +79,33 @@ void begin() {
...
-            ClientMessage request = TransactionCommitParameters.encode(txnId, threadId, prepareAndCommit);
+            ClientMessage request = TransactionCommitParameters.encode(txnId, threadId, true);
             invoke(request);
             state = COMMITTED;
         } catch (Exception e) {
             state = ROLLING_BACK;
             throw ExceptionUtil.rethrow(e);
         } finally {
-            closeConnection();
+            THREAD_FLAG.set(null);
```
metanet: Can we use a more explanatory name like TRANSACTION_EXISTS?

gurbuzali: makes sense

gurbuzali:
verify

metanet:
I also think that thread-safety guarantees should be documented for implementation classes that are provided to user.

Besides that, :+1:

juanavelez:
I am testing the latest build (3.5-EA2-SNAPSHOT) in a JBoss EAP 6.4.0 environment using the JCA-RA and Container-Managed Transactions and the example given by the reference documentation (section 10.3.5 Sample JBoss AS 7 / EAP 6 Web Application Configuration)  is broken. It seems the ManagedConnectionImpl.getTx() is returning the Tx from the instance itself which is not the same the as the Tx instance the xaResource instance variable and therefore it fails with "Transaction is not Active" coming from HazelcastConnectionImpl's getTransactionalMap.

    @Override
    public <K, V> TransactionalMap<K, V> getTransactionalMap(String name) {
        final TransactionContext txContext = this.managedConnection.getTx().getTxContext();
        if (txContext == null) {
            throw new IllegalStateException("Transaction is not active");
        }
        return txContext.getMap(name);
    }

 I think with the redesign the TxContext should come from the XAResource within the ManagedConnectionImpl and not its own Tx, which it leads to question if that Tx instance variable is still necessary.

    public ManagedConnectionImpl(ConnectionRequestInfo cxRequestInfo, ManagedConnectionFactoryImpl factory) {
        this.setLogWriter(factory.getLogWriter());
        log(Level.FINEST, "ManagedConnectionImpl");

        this.factory = factory;
        this.cxRequestInfo = cxRequestInfo;

        this.id = ID_GEN.incrementAndGet();
        this.tx = new HazelcastTransactionImpl(factory, this);
        HazelcastInstance instance = factory.getResourceAdapter().getHazelcast();
        this.xaResource = instance.getXAResource();

        factory.logHzConnectionEvent(this, HzConnectionEvent.CREATE);
    }

Most likely this means adding a new method getTx() to the HazelcastXAResource?

PR #5163, opened 2015-04-24
TransactionalMapProxy wrong exception when paging predicate

pveentjer:
throws NPE instead of IllegalArgumentException in case of paging predicate.


jerrinot:
It's API change, however the original behaviour was buggy. 
Hence :+1: for this.

Donnerbart:
:+1:

PR #5165, opened 2015-04-24
Moved 'getConnectedClientStats()' method to ClientEngine from ClusterSer...

kobalski:
...vice and simplified method. This pr contains simplifications and refactorings of this PR : https://github.com/hazelcast/hazelcast/pull/5161

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
@@ -642,8 +642,7 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
         }
 
         //Now we are regrouping according to the client type
-        for (Map.Entry<String, ClientType> entry : clientsMap.entrySet()) {
-            ClientType clientType = entry.getValue();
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -605,4 +609,61 @@ void removeMappings() {
...
+                for (Map.Entry<String, ClientType> entry : endpoints.entrySet()) {
+                    clientsMap.put(entry.getKey(), entry.getValue());
+                }
+            } catch (Exception e) {
+                logger.warning("Cannot get client information from: " + target.toString(), e);
+            }
+        }
+
+        //Now we are regrouping according to the client type
+        for (Map.Entry<String, ClientType> entry : clientsMap.entrySet()) {
```
mdogan: You can iterate just `values()` here instead of entries. So, you'll avoid created garbage objects.

```java
for (ClientType clientType : clientsMap.values()) {
...
```

kobalski: You are right! No need for extra variable.

mdogan:
:+1:

PR #5166, opened 2015-04-24
Config backup-count validation

pveentjer:
Adds backup/async-backup validation on configs where missing + removes copy/paste code on places where implemented.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/Preconditions.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/Preconditions.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.util;
 
+import static com.hazelcast.partition.InternalPartition.MAX_BACKUP_COUNT;
 import static java.lang.String.format;
 
 /**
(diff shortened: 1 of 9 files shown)


hazelcast/src/main/java/com/hazelcast/util/Preconditions.java
```diff
@@ -148,12 +150,71 @@ public static int checkPositive(int value, String errorMessage) {
...
+        if (newBackupCount < 0) {
+            throw new IllegalArgumentException("backup-count can't be smaller than 0");
+        }
+
+        if (currentAsyncBackupCount < 0) {
+            throw new IllegalArgumentException("async-backup-count can't be smaller than 0");
+        }
+
+        if (newBackupCount > MAX_BACKUP_COUNT) {
+            throw new IllegalArgumentException("backup-count can't be larger than than " + MAX_BACKUP_COUNT);
```
jerrinot: is this necessary? the subsequent check will take care of it anyway. I can imagine it produces better error message, but I'm not sure if that's worth having an extra check.

pveentjer: You detect any potential mess up. So you don't rely on the other method being called..

jerrinot: what about including passed values in a error message?

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/util/Preconditions.java
pveentjer: Good idea. Currently that isn't done in the existing code.

I would like to do it in one go for the whole preconditions library

PR #5040, opened 2015-04-11
Introduced Blackbox

pveentjer:
This PR contains the integration of the blackbox sensors.

First the following pr, which contains the actual blackbox, needs to be merged:
https://github.com/hazelcast/hazelcast/pull/5167

The Blackbox provides insight what is happening inside of the Hazelcast member. It can currently be read using the Performance monitor which was completely rewritten for this purpose. In the future we'll add a real storage solution for the blackbox itself.

Lots of sensor points were added into the system. The SlowOperationDetector also dumps to the Performance log.

The blackbox is always on; it can't be disabled. The overhead of the blackbox is very low since it will periodically sample fields/methods of objects to get internal state. These fields can be simple volatile long and can even be written using a lazySet to make it even cheaper than a volatile write.

The blackbox will be the central hub to record/access all kinds of sensor data. Before it was done ad hoc and all over the place. But now all this information is accessible through the blackbox.

Sensor can easily be created by:
```
@SensorInput
private int field;

@SensorInput
int size(){return ....}
```
And then registering an object which such annotations at the blackbox.

The Performance log will not log indefinitely. It uses a rolling file approach and by default it allows for 10 files of 10mb to be kept on disk. This is configurable using properties (e.g. system properties).

An example log:
```
====================[ Build Info ]======================================
property                        value
------------------------------------------------------------------------
version                         3.5-SNAPSHOT
build                           20150411
buildNumber                     20150411
revision                        3.5-SNAPSHOT
enterprise                      false


================[ System Properties ]===================================
property                        value
------------------------------------------------------------------------
[...]

PR #5167, opened 2015-04-24
Adds Blackbox without integration

pveentjer:
This is a cleaned up PR. It contains the actual blackbox and the sensor packs. But it isn't registered into the system.

The registration is part of this PR:
https://github.com/hazelcast/hazelcast/pull/5040

The reason for separating the blackbox from the sensor integration is to reduce the risk of getting the blackbox integrated.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/blackbox/Blackbox.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/blackbox/Blackbox.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.blackbox;
+
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The Blackbox is responsible for recording all kinds of Hazelcast/JVM specific information to
+ * help out with all kinds of issues.
+ *
+ * Each HazelcastInstance has its own Blackbox.
+ *
+ * A Blackbox can contain many {@link Sensor}. Each sensor has an input and each sensor can be identified using a
+ * parameter; a String. This parameter can be any string, the general structure is something like:
+ * <ol>
+ *     <li>proxy.count</li>
+ *     <li>operation.completed.count</li>
+ *     <li>operation.partition[14].count</li>
+ * </ol>
+ * For the time being there the Blackbox doesn't require any syntax for the parameter content; so any String is fine.
+ *
+ * <h1>Duplicate Registrations</h1>
+ * The Blackbox is also lenient regarding duplicate registrations of sensors. So if a Sensor is created and
+ * an SensorInput is set and a new registration for the same Sensor is done, the old SensorInput is overwritten.
+ * The reason to be lenient is that the Blackbox should not throw exception. Of course there will be a log
+ * warning.
+ *
+ * <h1>Performance</h1>
+ * The Blackbox is designed for low overhead sensors. So once a sensor is registered, there is no overhead
+ * for the provider of the sensor data. The provider could have for example a volatile long field and increment
+ * this using a lazy-set. As long as the Blackbox can frequently read out this field, the Blackbox is perfectly
+ * happy with such low overhead sensor-inputs. So it is up to the provider of the sensor-input how much overhead
+ * is required.
+ */
+public interface Blackbox {
+
+    /**
+     * Scans the source object for any fields/methods that have been annotated with {@link SensorInput} annotation, and
+     * registering these fields/methods as sensors.
+     *
+     * If sensors with the same parameter already exist, there source/inputs will be updated. So multiple registrations
+     * if the same object are ignored.
+     *
+     * If an object has no sensor annotations, the call is ignored.
+     *
+     * @param source          the object to scan.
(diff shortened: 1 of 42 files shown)


pveentjer:
verify

mtopolnik:
:+1:

jerrinot:
there is a failing test

hazelcast/src/main/java/com/hazelcast/internal/blackbox/impl/BlackboxImpl.java
```diff
@@ -0,0 +1,207 @@
...
+import static com.hazelcast.util.Preconditions.checkNotNull;
+import static java.lang.String.format;
+
+/**
+ * The {@link Blackbox} implementation.
+ *
+ * Interesting read:
+ * https://source.android.com/devices/sensors/sensor-types.html
+ */
+public class BlackboxImpl implements Blackbox {
```
jerrinot: do you expect to have multiple implementations?

hazelcast/src/main/java/com/hazelcast/internal/blackbox/Blackbox.java
```diff
@@ -0,0 +1,134 @@
...
+ * warning.
+ *
+ * <h1>Performance</h1>
+ * The Blackbox is designed for low overhead sensors. So once a sensor is registered, there is no overhead
+ * for the provider of the sensor data. The provider could have for example a volatile long field and increment
+ * this using a lazy-set. As long as the Blackbox can frequently read out this field, the Blackbox is perfectly
+ * happy with such low overhead sensor-inputs. So it is up to the provider of the sensor-input how much overhead
+ * is required.
+ */
+public interface Blackbox {
```
jerrinot: could you explain what is a parameter?

hazelcast/src/main/java/com/hazelcast/internal/blackbox/sensorpacks/OperatingSystemSensorPack.java
```diff
@@ -0,0 +1,108 @@
...
+
+        if (long.class.equals(method.getReturnType())) {
+            blackBox.register(osBean, parameter,
+                    new LongSensorInput() {
+                        @Override
+                        public long get(Object bean) throws Exception {
+                            return (Long) method.invoke(bean);
+                        }
+                    });
+        } else {
```
jerrinot: I'd be defensive here and use `else if (double.class.equals(method.getReturnedType()))`;

hazelcast/src/main/java/com/hazelcast/internal/blackbox/impl/BlackboxImpl.java
```diff
@@ -0,0 +1,207 @@
...
+        return sensor;
+    }
+
+    @Override
+    public synchronized <S> void deregister(S source) {
+        checkNotNull(source, "source can't be null");
+
+        boolean changed = false;
+        for (Map.Entry<String, SensorImpl> entry : sensors.entrySet()) {
+            String parameter = entry.getKey();
```
jerrinot: just a minor one: you can move this one down below the `continue` short-circuit

hazelcast/src/main/java/com/hazelcast/internal/blackbox/impl/AccessibleObjectInput.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Provides an abstraction for {@link FieldSensorInput} and {@link MethodSensorInput}.
```
jerrinot: I'm not quite sure what does it mean

jerrinot:
there are some minor things. I have to look at the AccessibleObjectInput.java more in details tomorrow. Overall I like both the idea and the implementation!

hazelcast/src/main/java/com/hazelcast/internal/blackbox/impl/BlackboxImpl.java
pveentjer: I do not expect multiple implementation. I do think that for larger 'systems', that they are set up using interfaces so you have a nice wall that prevents tight coupling. We do it for all the larger components in the system e.g. OperationService, ProxyService etc.

Setting up interfaces helps to make a very clear separation between API and implementation.

hazelcast/src/main/java/com/hazelcast/internal/blackbox/Blackbox.java
pveentjer: Sure. I'll it to the documentation.

I got the term from the aeronautical industry when they are explaining the blackbox.

hazelcast/src/main/java/com/hazelcast/internal/blackbox/sensorpacks/OperatingSystemSensorPack.java
pveentjer: sure. Will do.

hazelcast/src/main/java/com/hazelcast/internal/blackbox/impl/BlackboxImpl.java
jerrinot: ok. I'm fine with that. 
I thought you weren't a fan of interfaces when only a single implemented is expected.

hazelcast/src/main/java/com/hazelcast/internal/blackbox/Blackbox.java
jerrinot: it would be nice to have it in-place. as a JavaDoc of this class.

hazelcast/src/test/java/com/hazelcast/internal/blackbox/impl/MethodSensorInputTest.java
```diff
@@ -0,0 +1,280 @@
...
+
+        try {
+            fieldSensorInput.getDouble(source);
+            fail();
+        } catch (IllegalStateException expected) {
+
+        }
+    }
+
+    private class SomeSource {
```
jerrinot: just a tip: testing extreme values such as `Integer.MAX_VALUE` and `Integer.MIN_VALUE` would make it a better test.

jerrinot:
:+1:

Donnerbart:
Should this one be closed in favor of #5195?

PR #5170, opened 2015-04-26
Added preconditions.checkTrue

pveentjer:


pveentjer:
Since 100% tested and minor addition (nobody is using this method for now) I'll merge this one myself.

PR #5175, opened 2015-04-27
Exposed TcpIpJoiner.MAX_PORT_TRIES as a configurable system property

### LINKED ISSUE #5062: Enhancement: Expose TcpIpJoiner.MAX_PORT_TRIES as a configurable property
Currently Hazelcast tries at most 3 incremental ports when tcp-ip joiner is used and no explicit port is defined in member-list.  See https://github.com/hazelcast/hazelcast/blob/68811925d185aadcf0eaadd84ea7b45ec8b4ce30/hazelcast/src/main/java/com/hazelcast/cluster/impl/TcpIpJoiner.java#L396

This `TcpIpJoiner.MAX_PORT_TRIES` compile time constant can be exposed to allow searching more ports.

See Zendest ticket: 838

mdogan:
Fixes #5062

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/TcpIpJoiner.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/TcpIpJoiner.java
@@ -22,6 +22,7 @@ import com.hazelcast.config.InterfacesConfig;
 import com.hazelcast.config.NetworkConfig;
 import com.hazelcast.config.TcpIpConfig;
 import com.hazelcast.core.Member;
+import com.hazelcast.instance.GroupProperties;
 import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.instance.Node;
 import com.hazelcast.nio.Address;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/cluster/impl/TcpIpJoiner.java
```diff
@@ -45,13 +45,15 @@
...
-    private static final int MAX_PORT_TRIES = 3;
     private static final long JOIN_RETRY_WAIT_TIME = 1000L;
 
-    private volatile boolean claimingMaster = false;
+    private final int maxPortTryCount;
+    private volatile boolean claimingMaster;
 
     public TcpIpJoiner(Node node) {
         super(node);
+        int tryCount = node.groupProperties.TCP_JOIN_PORT_TRY_COUNT.getInteger();
```
pveentjer: Probably better to throw an exception here (fail fast).

I think we should give the group properties a good cleanup so that its constrains can be defined in the group property itself instead of dealing with it when you get the property.

```
GroupProperty TRY_COUNT = new GroupProperty("com.hazelcast.tcp.tryCount").asInt(..).isPositive(...).build(); 
```

pveentjer:
One comment. Apart from that it looks good.

hazelcast/src/main/java/com/hazelcast/cluster/impl/TcpIpJoiner.java
mdogan: done

pveentjer:
Can be merged once green.

PR #5177, opened 2015-04-27
Moved version check after node starts.

kobalski:
Version checking operation is done after node start completes and active, because we are collecting connected client information and node has to be active first.


mesutcelik:
do we have any test for the change?

kobalski:
I wrote test for this change by mocking VersionCheck class and verifying check method is called inside Node.start method, but in order to achieve that I have to make VersionCheck class non-final and also inside Node class I needed to add a setter for versionCheck. https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/instance/Node.java#L128 Also we do not have any tests for VersionCheck class, If you want I can add tests to this PR, with the changes I have described. I only changed when VersionCheck.check with this PR.

PR #5179, opened 2015-04-27
NodeEngineImpl.getPortableContext removed since unused

pveentjer:


pveentjer:
Merging this one myself since it is just the removal of an unused method.

PR #5187, opened 2015-04-27
I/O scheduling bug fixed

jerrinot:
When a last packet doesn't fully fit the outputBuffer then we have to re-schedule writeHandler again
otherwise there is a risk of a dangling packet - a packet which is partially written.

This is very visible in queries - a packet with QueryResult can easily exceed a size of the outputBuffer.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
@@ -157,7 +157,7 @@ public final class WriteHandler extends AbstractSelectionHandler implements Runn
      * This call is only made by the IO thread.
      */
     private void unschedule() {
-        if (dirtyOutputBuffer()) {
+        if (dirtyOutputBuffer() || currentPacket != null) {
             // Because not all data was written to the socket, we need to register for OP_WRITE so we get
             // notified when the socketChannel is ready for more data.
             registerOp(SelectionKey.OP_WRITE);


jerrinot:
@mdogan @pveentjer: guys, may I kindly ask you to review this change? it's deep down in the I/O so I'd like to hear your opinion here. thanks!

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
```diff
@@ -172,7 +172,7 @@ private void unschedule() {
         // So the outputBuffer is empty, so we are going to unschedule ourselves.
         scheduled.set(false);
 
-        if (writeQueue.isEmpty() && urgentWriteQueue.isEmpty()) {
+        if (writeQueue.isEmpty() && urgentWriteQueue.isEmpty() && currentPacket == null) {
```
mdogan: So, problem happens when a packet doesn't fit into buffer and there are no other pending packets in the queue. When both packets queues are empty, we are removing write interest immediately, ignoring partially written packet. And that packet's remaining part is written only when another packet is enqueued and write interest is set. 

Right? Anything wrong in my understanding?

jerrinot:
verify

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
jerrinot: precisely! 

If you don't mind I'll steal your explanation and use it as a commit message.

sancar: A concern about our new WriteHandler. Old way of handling read and write was there since very early days of hazelcast and it is tested and proven over these years that it is stable. And as far as I know, we changed it so that it becomes more readable. I personally found two major bugs in new one. And there is another critical bug found by @jerrinot . My question is that are we confident enough release with this one? I might be missing something. Does this change has any other proven advantages  like performance other than readability? If there is performance improvement, It make sense to insist on continue with new WriteHandler implementation. Note: I did not change the client side WrieHandler to new one, because I think it is not mature enough.

pveentjer: Good points. 

These changes were made by me in HZ 3.4 at 12/16/2014. Reverting to the old write handlers could very well introduce new bugs since changes have been made in the IO system after I did the cleanup.

I propose that we give the IO system a complete cleanup.

The current code is complicated because of:
- the inherent nature of NIO
- accidental complexity caused by complex code.

jerrinot: +1 to @pveentjer

sancar: I believe your changes in WriteHandler starts here 
https://github.com/hazelcast/hazelcast/commit/de557b4b1700a72f7a55f7a894cad95af689f24a
And version seems to be 3.5-SNAPSHOT. 
https://github.com/hazelcast/hazelcast/blob/de557b4b1700a72f7a55f7a894cad95af689f24a/pom.xml#L26

mdogan: @jerrinot, of course!

@sancar, @pveentjer; 
I find the new WriteHandler impl more readable and understandable. Old one was working good but it was hard to understand the code. Java NIO API is already very complex, as Peter said. Actually it's more complex than it should be, I think. 

I personally don't mind minor bugs which can be easily fixed in a patch release. Since we are testing this change since 3.4, it seems working. This one, for instance, is a critical one, which occurs when system is under very low load and causes big latency jump. It's hidden from our eyes thanks to heartbeat messages sent periodically :). 

 @jerrinot: can you also add a test to show the issue and verify the problem? You can increase the heartbeat interval to a very large value in test to prevent it to interleave...

pveentjer: @sancar I'm not understanding the point you try to make.

@general
We should have a a look at cleaning up the IO system. But probably best to make the whole thing pluggable so we can keep using the old and swap in the new if needed. This way we can fall back upon something battle hardened.

sancar: @pveentjer All I wanted to make sure is that if we are confident on the new changes. If we are not, it is not too late to revert. If we think this is much more readable and easily we can easily fix the bugs, then I am OK with the changes and we can leave as is.

enesakar: Peter, refactoring code; making it more simpler and more readable is good.
But what I get from Sancar's statement; this caused several critical bugs and time spent to solve them.

Possible reasons? 
- Mistake of implementor or reviewers (this is always possible)
- Low coverage of tests that touches such core parts (how possible is it to test such low level functionality)

I know you are planning to refactor/cleanup other parts of the code:
So what should be done to prevent such problems in future?

pveentjer:
The following code is a bit cleaner since there is no need to unschedule; since that is what already has happened when your code addition is executed.

```
 private void unschedule() {
        if (dirtyOutputBuffer() || currentPacket != null) {
            // Because not all data was written to the socket, we need to register for OP_WRITE so we get
            // notified when the socketChannel is ready for more data.
            registerOp(SelectionKey.OP_WRITE);

            // If the outputBuffer is not empty, we don't need to unschedule ourselves. This is because the
            // WriteHandler will be triggered by a nio write event to continue sending data.
            return;
        }
```

pveentjer:
```
 public void handle() {
        eventCount++;
        lastHandle = Clock.currentTimeMillis();
        if (!connection.isAlive()) {
            return;
        }

        if (socketWriter == null) {
            logger.log(Level.WARNING, "SocketWriter is not set, creating SocketWriter with CLUSTER protocol!");
            createWriter(Protocols.CLUSTER);
        }

        try {
            do {
                fillOutputBuffer();
            } while (writeOutputBufferToSocket());
        } catch (Throwable t) {
            logger.severe("Fatal Error at WriteHandler for endPoint: " + connection.getEndPoint(), t);
        }
        unschedule();
    }

    /**
     * Writes to content of the outputBuffer to the socket.
     *
     * @throws Exception
     * @returns true if everything was written
     */
    private boolean writeOutputBufferToSocket() throws Exception {
        if (!dirtyOutputBuffer()) {
            return true;
        }

        // So there is data for writing, so lets prepare the buffer for writing and then write it to the socketChannel.
        outputBuffer.flip();
        try {
            socketChannel.write(outputBuffer);
        } catch (Exception e) {
            currentPacket = null;
            handleSocketException(e);
            return true;
        }
        // Now we verify if all data is written.
        if (!outputBuffer.hasRemaining()) {
            // We managed to fully write the outputBuffer to the socket, so we are done.
            outputBuffer.clear();
            return true;
        }
        // We did not manage to write all data to the socket. So lets compact the buffer so new data
        // can be added at the end.
        outputBuffer.compact();
        return false;
    }
```

pveentjer:
I think just adding implementing the following:

```
private void unschedule() {
        if (dirtyOutputBuffer() || currentPacket != null) {
```

Is the simplest solution that doesn't introduce any new issues.

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
pveentjer: It is difficult one. The problem is that if there would not be some kind of regular packet that released the stuck packet, you would get a testing failure. But because e.g. a heart beat and other potential operations (responses etc) the error gets hidden.

A general purpose performance test where we test with small packets and many concurrently won't expose the problem. Only with large packets and little concurrent calls, this issue would have showed up. But.. probably only a performance degradation issue.. not from a correctness point of view (due to e..g heart beat).

My crucial mistake here is that I did refactoring without sufficient tests being available for the IO system. The IO System relies on implicit tests; so it gets tested because other components get tested that rely on the io system (e.g. the IMap).

So in this case we should have provided a lot of tests before attempting to refactoring.  But to make this happen, we need to make sure that we get better testable code. And this by its very nature, requires changes in the actual code to make it happen.

jerrinot:
implemented. 

I'll look at a possible test scenario, but it's tricky. A unit test is virtually impossibly here and integration test would have to rely on some progress indicators - this introduces a risk of spurious failures / false negatives.

pveentjer:
+1 (for merging it)

mdogan:
@jerrinot, have you tested/verified that if maintenance branch also suffers from the same issue or not? It's using the old version of `WriteHandler` but maybe it has the same defect...

jerrinot:
@mdogan: the latest maintanance-3.x (2917dc796) is fine

mdogan:
@jerrinot: thx

:+1: to merge

PR #5193, opened 2015-04-28
[CP] ClientBasicMapTest enabled

eminn:


eminn:
verify

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -675,11 +675,11 @@ public String addEntryListener(EntryListener listener, Predicate<K, V> predicate
         final Data keyData = toData(key);
         ClientMessage request = MapGetEntryViewParameters.encode(name, keyData, ThreadUtil.getThreadId());
         ClientMessage response = invoke(request, keyData);
-        EntryViewParameters parameters = EntryViewParameters.decode(response);
-
-        if (parameters == null) {
+        // TODO : Should entryview parameters handle null cases ?
+        if (response.getMessageType() == VoidResultParameters.TYPE.id()) {
```
sancar: Changing the return type according to result does not sound good. As you state in your comment, it could be better to handle this in EntryViewParameters with an extra boolean value or checking if key is NULL . Because it is not something that does not return anything. Void is used for such cases.  Passing DefaultData.NULL_DATA as key and and value for null when sending could be other option rather than extra boolean.

asimarslan: @sancar : Can you refactor entryviewParameter in a seperate PR ?

sancar: Yes, I can.

PR #5194, opened 2015-04-28
Improved logging in tests for SlowOperationDetector.

Donnerbart:
* Improved logging in tests for `SlowOperationDetector`
* Made use of `assertTrueEventually()` to give `SlowOperationDetector` enough time to scan for slow operations
* Implemented `toString()` methods for `PartitionWideEntryWithPredicateOperation` and `TruePredicate`

There are still tests failing and I have no idea why exactly. So I improved the logging to gather more information if tests fail again.


Donnerbart:
The tests are not failing locally on my machine, even with `@Repeat(value=500)`.

PR #5195, opened 2015-04-29
Hazelcast Metrics Core

pveentjer:
This is formally known as the Blackbox.

This PR only adds some Metrics classes; but it doesn't integrate them in HZ at all. They are just a bunch if unused class file. So the risk of merging them is very low

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -447,6 +447,8 @@
 
     <suppress checks="" files="com.hazelcast.util.DebugUtil"/>
 
+    <suppress checks="IllegalImport" files="com.hazelcast.util.counters.SwCounter"/>
+    <suppress checks="InnerAssignment" files="com.hazelcast.util.counters.SwCounter"/>
 
     <suppress checks="" files="com.hazelcast.util.HashUtil"/>
     <suppress checks="MagicNumber" files="com.hazelcast.util.QuickMath"/>
--- a/hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/GaugeImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/GaugeImpl.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.metrics.impl;
+
+import com.hazelcast.internal.metrics.DoubleProbe;
+import com.hazelcast.internal.metrics.Gauge;
+import com.hazelcast.internal.metrics.LongProbe;
+import com.hazelcast.internal.metrics.Metric;
+import com.hazelcast.logging.ILogger;
+
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
+/**
+ * Default {@link Metric} implementation.
+ *
+ * @param <S>
+ */
+public class GaugeImpl<S> implements Gauge {
+
+    volatile Object input;
+    volatile S source;
+
+    private final ILogger logger;
+    private final String name;
+
+    public GaugeImpl(String name, ILogger logger) {
+        this.name = name;
+        this.logger = logger;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public void render(StringBuilder sb) {
+        checkNotNull(sb, "sb can't be null");
+
+        sb.append(name).append('=');
+
+        Object input = this.input;
+        S source = this.source;
+
+        if (input == null || source == null) {
+            sb.append("NA");
+            return;
+        }
+
+        try {
+            if (input instanceof LongProbe) {
+                LongProbe<S> function = (LongProbe) input;
+                sb.append(function.get(source));
+            } else {
+                DoubleProbe<S> function = (DoubleProbe) input;
+                sb.append(function.get(source));
+            }
+        } catch (Exception e) {
+            logger.warning("Failed to update metric:" + name, e);
+            sb.append("NA");
+        }
+    }
+
+    @Override
+    public long readLong() {
+        Object input = this.input;
+        S source = this.source;
+        long result = 0;
+
+        if (input == null || source == null) {
+            return result;
+        }
+
+        try {
+            if (input instanceof LongProbe) {
+                LongProbe<S> longInput = (LongProbe<S>) input;
+                result = longInput.get(source);
+            } else {
+                DoubleProbe<S> doubleInput = (DoubleProbe<S>) input;
+                double doubleResult = doubleInput.get(source);
+                result = Math.round(doubleResult);
+            }
+            return result;
+        } catch (Exception e) {
+            logger.warning("Failed to update metric:" + name, e);
+            return result;
+        }
+    }
+
+    @Override
+    public double readDouble() {
+        Object input = this.input;
+        S source = this.source;
+        double result = 0;
+
+        if (input == null || source == null) {
+            return result;
+        }
+
+        try {
+            if (input instanceof LongProbe) {
+                LongProbe<S> longInput = (LongProbe<S>) input;
+                result = longInput.get(source);
+            } else {
+                DoubleProbe<S> doubleInput = (DoubleProbe<S>) input;
+                result = doubleInput.get(source);
+            }
+            return result;
+        } catch (Exception e) {
+            logger.warning("Failed to update metric:" + name, e);
+            return result;
+        }
+    }
+}
--- a/hazelcast/src/test/java/com/hazelcast/internal/metrics/impl/FieldProbeTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/internal/metrics/impl/FieldProbeTest.java
@@ -1,0 +1,172 @@
+package com.hazelcast.internal.metrics.impl;
+
+import com.hazelcast.internal.metrics.Probe;
+import com.hazelcast.internal.metrics.impl.FieldProbe.DoubleFieldProbe;
+import com.hazelcast.internal.metrics.impl.FieldProbe.LongFieldProbe;
+import com.hazelcast.test.HazelcastParallelClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
+import com.hazelcast.test.annotation.QuickTest;
+import com.hazelcast.util.counters.Counter;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static com.hazelcast.internal.metrics.impl.FieldProbe.createFieldProbe;
+import static com.hazelcast.util.counters.SwCounter.newSwCounter;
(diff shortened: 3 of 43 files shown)


Donnerbart:
verify

checkstyle/suppressions.xml
```diff
@@ -148,6 +148,10 @@
     <suppress checks="JavadocMethod" files="com.hazelcast.internal.storage[\\/]"/>
     <suppress checks="JavadocVariable" files="com.hazelcast.internal.storage[\\/]"/>
 
+    <!-- we don't care about return count here since there are just many types to deal with -->
+    <suppress checks="ReturnCount" files="com.hazelcast.internal.metrics.impl.FieldSensor"/>
```
metanet: I suppose these lines must be fixed to FieldGauge and MethodGauge

pveentjer: I think they can be removed completely since I simplified the methods a lot.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/GaugeImpl.java
```diff
@@ -0,0 +1,150 @@
...
+            logger.warning("Failed to update metric:" + name, e);
+            return result;
+        }
+    }
+
+    @Override
+    public double readDouble() {
+        Object input = this.input;
+        S source = this.source;
+        double result = 0;
```
metanet: There are too many type checks and if-else statements here. I think it may be a code smell about the type hierarchy.
What I would probably do here is something like this:


    interface ProbeReader { // a new type for reading values

        long getAsLong(Object source);

        double getAsDouble(Object source);

    }

    class LongProbeReader {

        private LongProbe probe;

        public long getAsLong(Object source)
                throws Exception {
            return probe.get(source);
        }

        public double getAsDouble(Object source)
                throws Exception {
            return probe.get(source);
        }

    }

    private static class DoubleProbeReader {

        private DoubleProbe probe;

        public long getAsLong(Object source)
                throws Exception {
            double doubleResult = probe.get(source);
            return Math.round(doubleResult);
        }

        public double getAsDouble(Object source)
                throws Exception {
            return probe.get(source);
        }

    }


    public long readLong() {
        ProbeReader reader = this.input;
        return reader.getAsLong(this.source);
    }

    public double readDouble() {
        ProbeReader reader = this.input;
        return reader.getAsDouble(this.source);
    }

I would also get rid of AccessibleObjectGauge and defined finer-grained probes such as MapSizeProbe, CollectionSizeProbe, CounterValueProbe etc. I think this will make a better type hierarchy here and result in simpler code.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/MethodGauge.java
```diff
@@ -0,0 +1,114 @@
...
+        return namePrefix + "." + name;
+    }
+
+    @Override
+    public boolean isDouble() {
+        return isDouble(type);
+    }
+
+    @Override
+    public long getLong(Object source) throws Exception {
```
metanet: My comment above also applies here.

metanet: My comment above also applies here.

jerrinot:
:+1:

Donnerbart:
verify

hazelcast/src/test/java/com/hazelcast/internal/metrics/impl/FieldProbeTest.java
```diff
@@ -0,0 +1,213 @@
...
+import static com.hazelcast.util.counters.SwCounter.newSwCounter;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+
+public class FieldProbeTest {
+
+    @Test
```
metanet: Why don't you use "@Test(expected = IllegalArgumentException.class)" here?

pveentjer: Sure. I'll change that.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/GaugeImpl.java
pveentjer: I'm against this. Because now you repeat the logic for every probe.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/MethodGauge.java
pveentjer: Thanks for the suggestion, but in my book it is not an improvement.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/GaugeImpl.java
metanet: No. There can be 2 kind of probes. Long probes and double probes. CollectionSizeProbe is a long probe, MapSizeProbe is also a long probe. If a field or method has a double return value, it can be a DoubleFieldProbe or DoubleMethodProbe. You can wrap these probes with corresponding LongProbeReader or DoubleProbeReader objects and there wont be a duplication. In the current code, there are if-else / switch statements everywhere.

pveentjer: Let me think about that.

pveentjer: Assuming we leave the current LongProbe and DoubleProbe interface as they are, then the following code could be simplified to:

before
```
 @Override
    public long readLong() {
        Object input = this.input;
        S source = this.source;
        long result = 0;

        if (input == null || source == null) {
            return result;
        }

        try {
            if (input instanceof AccessibleObjectProbe) {
                AccessibleObjectProbe accessibleObjectInput = (AccessibleObjectProbe) input;
                if (accessibleObjectInput.isDouble()) {
                    double doubleResult = accessibleObjectInput.getDouble(source);
                    result = Math.round(doubleResult);
                } else {
                    result = accessibleObjectInput.getLong(source);
                }
            } else if (input instanceof LongProbe) {
                LongProbe<S> longInput = (LongProbe<S>) input;
                result = longInput.get(source);
            } else if (input instanceof DoubleProbe) {
                DoubleProbe<S> doubleInput = (DoubleProbe<S>) input;
                double doubleResult = doubleInput.get(source);
                result = Math.round(doubleResult);
            }
            return result;
        } catch (Exception e) {
            logger.warning("Failed to update metric:" + name, e);
            return result;
        }
    }
```

after
```
    @Override
    public long readLong() {
        Object input = this.input;
        S source = this.source;
        long result = 0;

        if (input == null || source == null) {
            return result;
        }

        try {
           if (input instanceof LongProbe) {
                LongProbe<S> longInput = (LongProbe<S>) input;
                result = longInput.get(source);
            } else if (input instanceof DoubleProbe) {
                DoubleProbe<S> doubleInput = (DoubleProbe<S>) input;
                double doubleResult = doubleInput.get(source);
                result = Math.round(doubleResult);
            }
            return result;
        } catch (Exception e) {
            logger.warning("Failed to update metric:" + name, e);
            return result;
        }
    }
```

metanet: I would prefer to wrap these probes into LongProbeReader and DoubleProbeReader interfaces I put above. Then I won't check type of the probe for reading the value. Since it is wrapped with the correct reader, reader will do the job for me without any checks.

pveentjer: Problem with wrapping ----> it requires a wrapper object for explicit registrations of LongProbe/DoubleProbe interfaces

metanet: Yes, every probe must be wrapped. Do you refer to memory consumption of extra objects? Is it probable that there will be millions of probes. I don't think that.

pveentjer: This is a cleanup with the FieldProbe either dealing with Long or dealing with Double.

```
abstract class FieldProbe<S> {

    static <S> FieldProbe<S> create(Field field, Probe  probe) {
        int type = AccessibleObjectProbe.getType(field.getType());
        if (type == -1) {
            throw new IllegalArgumentException(format("@Gauge field '%s' is of an unhandled type", field));
        }

        if (AccessibleObjectProbe.isDouble(type)) {
            return new DoubleFieldProbe<S>(field, probe, type);
        } else {
            return new LongFieldProbe<S>(field, probe, type);
        }
    }

    final Probe probe;
    final Field field;
    final int type;

    FieldProbe(Field field, Probe probe, int type) {
        this.field = field;
        this.probe = probe;
        this.type = type;
        field.setAccessible(true);
    }

     void register(MetricsRegistryImpl metricsRegistry, Object source, String namePrefix) {
        String name = getName(namePrefix);
        metricsRegistry.registerInternal(source, name, this);
    }

    private String getName(String namePrefix) {
        String name = field.getName();
        if (!probe.name().equals("")) {
            name = probe.name();
        }

        return namePrefix + "." + name;
    }

    static class LongFieldProbe<S> extends FieldProbe implements LongProbe<S> {

        public LongFieldProbe(Field field, Probe probe, int type) {
            super(field, probe, type);
        }

        @Override
        public long get(S source) throws Exception {
            switch (type) {
                case TYPE_PRIMITIVE_LONG:
                    return field.getLong(source);
                case TYPE_LONG_NUMBER:
                    Number longNumber = (Number) field.get(source);
                    return longNumber == null ? 0 : longNumber.longValue();
                case TYPE_MAP:
                    Map<?, ?> map = (Map<?, ?>) field.get(source);
                    return map == null ? 0 : map.size();
                case TYPE_COLLECTION:
                    Collection<?> collection = (Collection<?>) field.get(source);
                    return collection == null ? 0 : collection.size();
                case TYPE_COUNTER:
                    Counter counter = (Counter) field.get(source);
                    return counter == null ? 0 : counter.get();
                default:
                    throw new IllegalStateException("Unhandled type:" + type);
            }
        }
    }

   static class DoubleFieldProbe<S> extends FieldProbe implements DoubleProbe<S> {

        public DoubleFieldProbe(Field field, Probe probe, int type) {
            super(field, probe, type);
        }

        @Override
        public double get(Object source) throws Exception {
            switch (type) {
                case TYPE_DOUBLE_PRIMITIVE:
                    return field.getDouble(source);
                case TYPE_DOUBLE_NUMBER:
                    Number doubleNumber = (Number) field.get(source);
                    return doubleNumber == null ? 0 : doubleNumber.doubleValue();
                default:
                    throw new IllegalStateException("Unhandled type:" + type);
            }
        }
    }
}
```

This allows the following cleanup:

```
  @Override
    public long readLong() {
        Object input = this.input;
        S source = this.source;
        long result = 0;

        if (input == null || source == null) {
            return result;
        }

        try {
           if (input instanceof LongProbe) {
                LongProbe<S> longInput = (LongProbe<S>) input;
                result = longInput.get(source);
            } else if (input instanceof DoubleProbe) {
                DoubleProbe<S> doubleInput = (DoubleProbe<S>) input;
                double doubleResult = doubleInput.get(source);
                result = Math.round(doubleResult);
            }
            return result;
        } catch (Exception e) {
            logger.warning("Failed to update metric:" + name, e);
            return result;
        }
    }
```

Since we don't need to deal with AccessibleObjectProbe anymore

pveentjer: The original memory constraints are 10KB. Currently it is increased to 250KB, but Greg is not happy with that. So I'll refrain from adding a wrapper object for the sake of OO.

metanet: I would even go for creating a probe class for each one of the "case" blocks above.

pveentjer: I know you want to do this :)

I find the complexity caused by this approach not worth the value it provides.

pveentjer:
verify

gurbuzali:
verify

pveentjer:
Can someone have a last look at this pr? Test failures are not related to metrics, since metrics are not integrated at all. So they are just dormant class files for a HZ instance.

gurbuzali:
verify

gurbuzali:
verify

PR #5196, opened 2015-04-29
Hazelcast Metrics Integration

pveentjer:


Donnerbart:
Seems to depend on #5195. I think we should review/merge that first and let Peter rebase this one. This should reduce the review work of this one, since half of this PR should vanish.

pveentjer:
verify

gurbuzali:
needs rebase and it would be good to squish the commits

gurbuzali:
checkstyle error
```
09:16:17 /scratch/jenkins/workspace/Hazelcast-3.x-pr-builder/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java:103:28: '{' is not preceded with whitespace.
```

gurbuzali:
I see that this is disabled by default. 
What is the impact to performance when a user enabled this ?
Will it affect the performance even if it is disabled ?

pveentjer:
The metrics can't be disabled; they are always on. Only dumping the content to file can be enabled/disabled. The performance impact of the metrics were not measurable. See the wiki. 10 runs with master and 10 runs with metrics branch.. performance difference was 0.56% positive (so metrics was faster than master).

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java
```diff
@@ -95,6 +99,11 @@ public final void addTask(Runnable runnable) {
         selectorQueue.add(runnable);
     }
 
+    @Probe
+    private long idleTime() {
+        return Math.abs(System.currentTimeMillis() - lastSelectTimeMs);
```
jerrinot: shouldn't this be `Math.max(System.currentTimeMillis() - lastSelectTimeMs, 0);`

as currently it can give you a false impression of staleness. 

let's assume there is a long GC pause right after the `System.currentTimeMillis()`. Then the `lastSelectTimeMs` might be greater than the value returned by `currentTimeMillis()` -> it will be reported as positive idle time which is simply not true.

metanet: Peter, how is 0.56% improvement possible while you are introducing additional load for counting etc?

pveentjer: Performance variations between runs

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -57,11 +57,32 @@
      * <p/>
      * The performance monitor logs all metrics into the log file.
      */
-    public static final String PROP_PERFORMANCE_MONITORING_ENABLED = "hazelcast.performance.monitoring.enabled";
+    public static final String PROP_PERFORMANCE_MONITOR_ENABLED = "hazelcast.performance.monitoring.enabled";
+
+    /**
+     * The delay in seconds between monitor of the performance.
```
jerrinot: is it a delay or an interval of measurement cycles? perhaps a native English speaker could have look here. @Danny-Hazelcast  ?

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -57,11 +57,32 @@
...
+            = "hazelcast.performance.monitor.max.rolled.file.size";
+
     /**
-     * The delay in seconds between monitoring of the performance.
+     * The PerformanceMonitor uses a rolling file approach to prevent eating too much disk space.
+     *
+     * This property sets the maximum number of rolling files to keep on disk.
      */
-    public static final String PROP_PERFORMANCE_MONITORING_DELAY_SECONDS = "hazelcast.performance.monitoring.delay.seconds";
+    public static final String PROP_PERFORMANCE_MONITOR_MAX_ROLLED_FILE_COUNT
```
jerrinot: can document the default value + describe a way how to disable it? (if there is any)

hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java
pveentjer: @jerrinot fixed and for other occurrances as well (read/write-handler)

metanet: So what we should say is this is not introducing a significant overhead

pveentjer: yes

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceMonitor.java
```diff
@@ -16,61 +16,107 @@
...
-            sb.append("ConnectionManager metrics\n");
-            connectionManager.dumpPerformanceMetrics(sb);
-            sb.append("OperationService metrics\n");
-            operationService.dumpPerformanceMetrics(sb);
+        logger.info("PerformanceMonitor started");
 
-            logger.info(sb.toString());
+        GroupProperties.GroupProperty slowOperationDetectorEnabled = node.getGroupProperties().SLOW_OPERATION_DETECTOR_ENABLED;
+        if (!slowOperationDetectorEnabled.getBoolean()) {
+            logger.info("To enable the SlowOperationDetector in the Performance log, set the following property:\n"
```
jerrinot: 1. should this go into INFO level? I don't think so. I understand the urge to give users hints however in this case it seems spammy. 
2. it should use platform neutral new-line delimiter

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
pveentjer: If you disable the performance monitor, then there is no logging.

jerrinot: can I enable the monitor, but disable the rolling thing?

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceMonitor.java
pveentjer: I think it should be put under info else a user will never see it. Debug is for debugging.

hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
```diff
@@ -69,227 +58,415 @@
...
+            } catch (OutOfMemoryError e) {
+                OutOfMemoryErrorDispatcher.onOutOfMemory(e);
+            } catch (Throwable t) {
+                logger.warning("Health Monitor failed", t);
             }
-        } catch (OutOfMemoryError e) {
-            OutOfMemoryErrorDispatcher.onOutOfMemory(e);
+        }
+
+        private void logPerformanceMonitorHint() {
```
jerrinot: I guess the hint is appropriate here, but why it's called only when the monitor level is set to SILENT?

hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
```diff
@@ -69,227 +58,415 @@
...
+            if (!performanceLogHint) {
+                return;
+            }
+
+            // we only log the hint once.
+            performanceLogHint = false;
+
+            logger.info(
+                    "The HealthMonitor has detected a high load on the system. For more detailed information, \n"
```
jerrinot: platform-specific line delimiter

pveentjer: fixed

hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
pveentjer: Fixed

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceMonitor.java
```diff
@@ -16,61 +16,107 @@
...
+            return this;
+        }
 
-        while (node.isActive()) {
-            sb.append("\n");
-            sb.append("ConnectionManager metrics\n");
-            connectionManager.dumpPerformanceMetrics(sb);
-            sb.append("OperationService metrics\n");
-            operationService.dumpPerformanceMetrics(sb);
+        logger.info("PerformanceMonitor started");
```
jerrinot: this feels spammy to me too - we don't log every starting component on INFO level.

pveentjer: By default it isn't started. So since the user explicitly enabled it, he should be informed that his explicit action is doing something.

jerrinot: fair enough.

pveentjer: If it would be on by default ---> I agree with you.

hazelcast/src/test/java/com/hazelcast/internal/monitors/PerformanceMonitorTest.java
```diff
@@ -0,0 +1,200 @@
...
+                assertNotNull(content);
+
+                assertTrue(content.contains("Build Info"));
+                assertTrue(content.contains("BuildNumber"));
+            }
+        });
+    }
+
+//    @Test
+//    public void testRollover() {
```
jerrinot: it should have an explanation why it's commented out

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
```diff
@@ -0,0 +1,322 @@
...
+
+            for (Object key : keys) {
+                String keyString = (String) key;
+                String value = properties.getProperty(keyString);
+                writer.append(format("%-60s  %4s\n", keyString, value));
+            }
+        }
+
+        private void renderBuildInfo(BufferedWriter writer) throws IOException {
+            writer.append("====================[ Build Info ]======================================\n");
```
jerrinot: platform-specific

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -56,12 +56,41 @@
...
+     * The PerformanceMonitor uses a rolling file approach to prevent eating too much disk space.
+     *
+     * This property sets the maximum size in MB for a single file.
+     *
+     * Every HazelcastInstance will get its own history of log files.
+     *
+     * The default is 10.
      */
-    public static final String PROP_PERFORMANCE_MONITORING_ENABLED = "hazelcast.performance.monitoring.enabled";
+    public static final String PROP_PERFORMANCE_MONITOR_MAX_ROLLED_FILE_SIZE
```
jerrinot: what about `_MB` suffix?

pveentjer: sure.

pveentjer: I guess the property as well

hazelcast/src/test/java/com/hazelcast/internal/monitors/PerformanceMonitorTest.java
pveentjer: Didn't have time to implement that test.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
jerrinot: yep.

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceMonitor.java
```diff
@@ -16,61 +16,107 @@
...
+                logger.warning(t.getMessage(), t);
+            }
+        }
+
+        private void sleep() {
+            for (int k = 0; k < delaySeconds; k++) {
+                try {
+                    Thread.sleep(DELAY_MILLIS);
+                } catch (InterruptedException e) {
+                    // we can eat the interrupt since
```
jerrinot: missing 2nd part of the comment?

pveentjer: done

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
```diff
@@ -0,0 +1,322 @@
...
+                    + logFile.getAbsolutePath() + " cause:" + e.getMessage());
+        }
+    }
+
+    private void deleteOld() {
+        File file = new File(format(pathname, index - maxRollingFileCount));
+        file.delete();
+    }
+
+    public boolean forceRender() {
```
jerrinot: it's a public method (even just for internal usage), could document it? I'm not quite sure what is it supposed to do.

pveentjer: fixed

pveentjer:
```
  /**
     * Checks if log file is forced to be rendered.
     */
    boolean isRenderingForced() {
        if (logFile == null) {
            return false;
        }

        return logFile.lastModified() != lastAccessed;
    }
```

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
pveentjer: fixed

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
```diff
@@ -0,0 +1,322 @@
...
+            renderMetricsRegistry(writer);
+            writer.append("\n\n");
+            if (renderSlowOperation(writer)) {
+                writer.append("\n\n");
+            }
+        }
+
+        private void renderHeader(BufferedWriter writer) throws IOException {
+            writer.append("================[ Metrics ]============================================\n");
+            Calendar calendar = Calendar.getInstance();
```
jerrinot: this can be quite slow and you need is a `new Date()`.

pveentjer: I'm not sure what you mean

jerrinot: you construct a calendar instance only to print a timestamp, right?
```
Date now = new Date();
writer.append(sdf.format(now)).append('\n');
```

Check source code of `Calendar getInstance()`. I believe it's unnecessary in your case. (unless I misunderstood this part badly)

pveentjer: I always get a terrible headache from the date/calendar api

jerrinot: it sucks because it closely reminds real world and real world sucks! :)))

gurbuzali:
verify

jerrinot:
this would be super-nice to have in 3.5.

Donnerbart:
Hmm, there seems to be one related test failure: `com.hazelcast.internal.monitors.PerformanceMonitorTest.testMetricsRegistry`
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast/13486/testReport/junit/com.hazelcast.internal.monitors/PerformanceMonitorTest/testMetricsRegistry/

Donnerbart:
verify

PR #5202, opened 2015-04-29
Enabled more new client tests

mdogan:


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/GetPartitionsMessageTask.java
```diff
@@ -61,7 +62,7 @@ protected ClientMessage call() {
             Address owner = partitions[i].getOwnerOrNull();
             int index = -1;
             if (owner == null) {
-                return null;
+                return VoidResultParameters.encode();
```
sancar: This does not look like it is going to work correctly. We will get exception when trying to read the result here:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java#L103

I solved similar problem in map.getEntryView by adding a boolean to parameters to differentiate null or not.

mdogan: This is what you suggested when returning null, I've no idea. just show me the correct version here.

sancar: I am in favor of returning always same type of message for simplicity. I am using VoidResultParameters when client side not expecting to read what is returned. 

To return null, I added a boolean to EntryView to  differentiate if it is null or not.
But for this case, it  does not seem a good fit to me. How about we throw an exception when null ?
Or returning an empty list for addresses and indices ? The later seems like a better choice to me. 
What do you think?

asimarslan: We may return VoidResult if client do not decode.

asimarslan: @sancar the code you send the link never called with non-owner adress

mdogan: @sancar: updated.

sancar:
:+1:

sancar:
It seems test is failed because there is a bug in GetPartitionsResultParameters.calculateDataSize
Following method should fix the bug

```
 public static int calculateDataSize(Address[] addresses, int[] ownerIndexes) {
        int dataSize = ClientMessage.HEADER_SIZE;
        dataSize += BitUtil.SIZE_OF_INT;
        for (Address address : addresses) {
            dataSize += ParameterUtil.calculateAddressDataSize(address);
        }
        dataSize += BitUtil.SIZE_OF_INT;
        dataSize += ownerIndexes.length * BitUtil.SIZE_OF_INT;
        return dataSize;
    }
```

mdogan:
@sancar: you're right `calculateDataSize()` was wrong, fixed it as you suggested. now it seems ok.

PR #5207, opened 2015-04-29
[CP] Fixed lock and condition tasks and enabled tests

eminn:


hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientLockProxy.java
```diff
@@ -87,7 +87,7 @@ public void lock(long leaseTime, TimeUnit timeUnit) {
     }
 
     public void forceUnlock() {
-        ClientMessage request = LockForceUnlockParameters.encode(getName());
+        ClientMessage request = LockForceUnlockParameters.encode(getName(), ThreadUtil.getThreadId());
```
sancar: I removed this threadId intentionally before. Is it necessary to pass threadId ? If yes, can you show me where it is used.

gurbuzali: force unlock should not need any threadId afaik

eminn: You guys are right about that, my bad.

sancar:
:+1:

PR #5208, opened 2015-04-29
Transactional Map , Queue, Set tests and  ClientTxnTest are opened .

sancar:


sancar:
I will send Transactional MultiMap after https://github.com/hazelcast/hazelcast/pull/5201 is merged.

PR #5213, opened 2015-04-30
TCP Exception handling tweaks

pveentjer:
Show stracktrace if debug enabled and an exception is thrown
WriteHandler now fails if an unknown protocol is offered instead of defaulting to text


pveentjer:
verify

jerrinot:
:+1:

pveentjer:
@mdogan can you have a look at this one as well?

PR #5214, opened 2015-04-30
MultiMapTxn  , ExecutorService & Quorum tests are opened. 

sancar:
causeClassName is added to ExceptionResultParameters

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/quorum/ClientMapWriteQuorumTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/quorum/ClientMapWriteQuorumTest.java
@@ -19,22 +19,16 @@ package com.hazelcast.client.quorum;
 import com.hazelcast.client.HazelcastClient;
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.config.MapConfig;
+import com.hazelcast.config.QuorumConfig;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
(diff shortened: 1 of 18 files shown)


hazelcast-client-new/src/test/java/com/hazelcast/client/quorum/ClientMapWriteQuorumTest.java
```diff
@@ -361,6 +362,7 @@ public void testExecuteOnEntriesOperationSuccessfulWhenQuorumSizeMet() throws Ex
     }
 
     @Test(expected = QuorumException.class)
+    @Ignore
     public void testExecuteOnEntriesOperationThrowsExceptionWhenQuorumSizeNotMet() throws Exception {
```
sancar: @eminn Can you take look at this one and the one in ClientMapReadWriteQuorumTest with same name ? These two are still failing with my fixes. I could not figure out the problem.

eminn: I found the problem, will add as line note.

eminn:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java#L1063-L1065 

and 

https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java#L1081-L1083

methods does not makes any invocation to the server side. So that is the reason why you have a test failure. `ClientMessage response = invoke(request);` should be added between request encoding and response decoding.

Issue #254, opened 2012-08-29
IMap.put with TTL over WAN not working


psurma:
When updating a map entry which is replicated over WAN, the TTL (TimeToLive) is not honoured in the remote cluster map.

I am using the put method of the IMap interface to update/overwrite a cache entry.

V put(K key,
      V value,
      long ttl,
      TimeUnit timeunit)

When the time-out expires, the entry disappears from the cluster in which the key is owned, however it remains in the remote cluster.

PR #5221, opened 2015-05-02
WAN replication bug for map entries with a specific TTL

matjohns:
Related to issue #254, but appears that entries with individualised TTLs are not correctly replicated to a remote cluster (hence do not expire, or do so at the default time). Looks like the TTL and ExpirationTime for the merged Record are not being set.

Fix within {Default,AbstractEvictable}RecordStore to set appropriate fields on a synced Record and excluding comparable objects with differing expiry times from being considered equivalent (hence sync skipped).

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
@@ -18,6 +18,7 @@ package com.hazelcast.map.impl;
 
 import com.hazelcast.config.EvictionPolicy;
 import com.hazelcast.config.MapConfig;
+import com.hazelcast.core.EntryView;
 import com.hazelcast.instance.GroupProperties;
 import com.hazelcast.map.impl.eviction.EvictionOperator;
 import com.hazelcast.map.impl.eviction.MaxSizeChecker;
@@ -276,7 +277,7 @@ abstract class AbstractEvictableRecordStore extends AbstractRecordStore {
     }
 
     protected void markRecordStoreExpirable(long ttl) {
-        if (ttl > 0L) {
+        if (ttl > 0L && ttl < Long.MAX_VALUE) {
             hasEntryWithCustomTTL = true;
         }
     }
@@ -383,6 +384,22 @@ abstract class AbstractEvictableRecordStore extends AbstractRecordStore {
 
         final long maxIdleMillis = calculateMaxIdleMillis(mapContainer.getMapConfig());
         setExpirationTime(record, maxIdleMillis);
+    }
+
+    protected void mergeRecordExpiration(Record record, EntryView mergingEntry) {
+        final long ttlMillis = mergingEntry.getTtl();
+        record.setTtl(ttlMillis);
+
+        final long lastAccessTime = mergingEntry.getLastAccessTime();
+        record.setLastAccessTime(lastAccessTime);
+
+        final long lastUpdateTime = mergingEntry.getLastUpdateTime();
+        record.setLastUpdateTime(lastUpdateTime);
+
+        final long maxIdleMillis = mapContainer.getMaxIdleMillis();
+        setExpirationTime(record, maxIdleMillis);
+
+        markRecordStoreExpirable(record.getTtl());
     }
 
 
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
@@ -396,7 +396,7 @@ abstract class AbstractEvictableRecordStore extends AbstractRecordStore {
         final long lastUpdateTime = mergingEntry.getLastUpdateTime();
         record.setLastUpdateTime(lastUpdateTime);
 
-        final long maxIdleMillis = mapContainer.getMaxIdleMillis();
+        final long maxIdleMillis = calculateMaxIdleMillis(mapContainer.getMapConfig());
         setExpirationTime(record, maxIdleMillis);
 
         markRecordStoreExpirable(record.getTtl());
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
@@ -837,6 +837,7 @@ public class DefaultRecordStore extends AbstractEvictableRecordStore implements 
             }
             newValue = mapDataStore.add(key, newValue, now);
             record = createRecord(key, newValue, now);
+            mergeRecordExpiration(record, mergingEntry);
             records.put(key, record);
             updateSizeEstimator(calculateRecordHeapCost(record));
         } else {
@@ -854,6 +855,9 @@ public class DefaultRecordStore extends AbstractEvictableRecordStore implements 
                 //remove from map & invalidate.
                 deleteRecord(key);
                 return true;
+            }
+            if (newValue == mergingEntry.getValue()) {
+                mergeRecordExpiration(record, mergingEntry);
             }
             // same with the existing entry so no need to map-store etc operations.
             if (mapServiceContext.compare(name, newValue, oldValue)) {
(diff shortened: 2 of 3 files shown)


mesutcelik:
Thanks @matjohns for the contribution. 

@ahmetmircik @emrahkocaman , can you please have a look?

emrahkocaman:
LGTBM

hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
```diff
@@ -382,6 +383,16 @@ protected void accessRecord(Record record, long now) {
         setExpirationTime(record, maxIdleMillis);
     }
 
+    protected void mergeRecordExpiration(Record record, EntryView mergingEntry) {
+        if (mergingEntry.getTtl() > 0 && mergingEntry.getTtl() != Long.MAX_VALUE) {
```
ahmetmircik: If `TTL = 0`, entry should live eternally in `IMap`. Can you please state your intention by not setting zero TTL from merging entry?

hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
```diff
@@ -808,13 +809,17 @@ public boolean merge(Data key, EntryView mergingEntry, MapMergePolicy mergePolic
                 return true;
             }
             // same with the existing entry so no need to map-store etc operations.
-            if (mapServiceContext.compare(name, newValue, oldValue)) {
+            if (mapServiceContext.compare(name, newValue, oldValue)
+                    && mergingEntry.getExpirationTime() == existingEntry.getExpirationTime()) {
```
ahmetmircik: comparing expiration times adding no value here. just updating expiration time of existing record seems sufficient.
WDYT?

hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
matjohns: As you say, those values represent eternal entries, however when an entry is created on cluster ```A``` without a TTL using ```IMap.put(k,v)```, the TTL is received by cluster ```B``` as ```Long.MAX_VALUE``` rather than ```zero```.

Currently the TTL on the record is not being set and the replicated entries do live eternally (as this is the bug I am seeking to address), but if we to pass through ```Long.MAX_VALUE``` this would inappropriately enable expiration scanning on the ```DistributedObject``` as a ```TTL > 0``` is passed to ```markRecordStoreExpirable(long ttl)```.

However if preferred, we could simply set the TTL from the merging entry here, and update ```AbstractEvictableRecordStore.markRecordStoreExpirable(long ttl)``` like so...

```
    protected void markRecordStoreExpirable(long ttl) {
        if (ttl > 0L && ttl < Long.MAX_VALUE) {
            expirable = true;
        }
    }
```

In writing this, that sounds like a more elegant solution, and will update the request as such :smile:

hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
```diff
@@ -808,13 +809,17 @@ public boolean merge(Data key, EntryView mergingEntry, MapMergePolicy mergePolic
...
                 return true;
             }
             newValue = mapDataStore.add(key, newValue, now);
             onStore(record);
             updateSizeEstimator(-calculateRecordHeapCost(record));
             recordFactory.setValue(record, newValue);
+            if (newValue == mergingEntry.getValue()) {
+                mergeRecordExpiration(record, mergingEntry);
+            }
```
matjohns: Agreed, just need to move this code above the comparison to ensure that the TTL and expiration time of the record is updated even if the entries object is the same (the overwriting to outlast entries within ```Entry_TTL_Replication_Issue```test, is to verify this exact case).

hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
```diff
@@ -382,6 +383,16 @@ protected void accessRecord(Record record, long now) {
...
+            record.setTtl(mergingEntry.getTtl());
+        }
+        if (mergingEntry.getExpirationTime() > 0 && mergingEntry.getExpirationTime() != Long.MAX_VALUE) {
+            record.getStatistics().setExpirationTime(mergingEntry.getExpirationTime());
```
ahmetmircik: Here `record.getStatistics()` may throw NPE.
 
To set expiration time correctly you should call this method `ExpirationTimeSetter#setExpirationTime` that method takes into account `max-idle-seconds` and also nullability of `record.getStatistics()`.

hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
ahmetmircik: Yes to fix `markRecordStoreExpirable ` makes sense.

hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
matjohns: Goodness, yes massive NPE problem; thank you!

I was thinking that ```ExpirationTimeSetter#setExpirationTime``` would not be appropriate because it regenerates the expiration time from the entries TTL at the point of WAN replication receipt, rather than synchronising the appropriate value that was set at the point of remote creation.

However this raises an interesting dilemma in terms of if the ```max-idle-seconds``` is different between remote clusters this wouldn't be honoured as the calculation of the expiration time was based on the configuration of source cluster... perhaps unlikely, but feasible.

Basically either I fix up the NPE issue, as such...
```
        if (record.getStatistics() != null) {
            record.getStatistics().setExpirationTime(mergingEntry.getExpirationTime());
        }
```

or use ```ExpirationTimeSetter#updateExpiryTime``` to both set the TTL and expiration time (admittedly having to source the appropriate ```MapConfig``` in order to get ```max-idle-seconds```), but the entry will effectively have a later expiry that the source cluster.

Thoughts? I prefer the former as it is less vulnerable to any async WAN replication delays.

hazelcast/src/main/java/com/hazelcast/map/impl/AbstractEvictableRecordStore.java
```diff
@@ -382,6 +383,12 @@ protected void accessRecord(Record record, long now) {
         setExpirationTime(record, maxIdleMillis);
     }
 
+    protected void mergeRecordExpiration(Record record, EntryView mergingEntry) {
+        record.setTtl(mergingEntry.getTtl());
```
ahmetmircik: I think setting only TTL can still be problematic here. We should also replicate `mergingEntry.getLastUpdateTime()` for TTL based eviction and `mergingEntry.getLastAccessTime()` for idle eviction. By replicating those values, entries life spans will be much closer to the expected ones.

For example, by making this change; if merging entries TTL equals to 100 seconds and remaining life span is 10 seconds, we can expect that entry will be expired after 10 seconds later in replica cluster instead of 100 seconds.

ajermakovics:
:+1:

gurbuzali:
verify

ahmetmircik:
Hi @matjohns, can you please rebase on current master? I want to merge your PR.
And also can you send a backport PR for maintenance branch?

matjohns:
@ahmetmircik there appears compilation issue on both upstream/master and upstream/maintenance-3.x at the moment preventing me from running the test suites to confirm a clean merge. I have created matjohns:maintenance-3.x-backport-pr5221 with the backport, but haven't yet sent the PR

mesutcelik:
@matjohns ,

I checked master already and don't see any compile problem. Can you attach what you see on your terminal?

matjohns:
@mesutcelik 

My bad; stale classes from switching branches too many times :) there is one minor issue interface change issue I need to address here; I'm just running the test suite before I commit/push the fix.

matjohns:
@ahmetmircik @mesutcelik should be good to go

mesutcelik:
verify

mesutcelik:
Thanks @matjohns for the contribution :+1:

Issue #5172, opened 2015-04-27
[TEST-FAILURE] ClientMapUnboundReturnValues_BasicTest failures


bilalyasar:
```
com.hazelcast.spi.exception.TargetDisconnectedException: null
	at com.hazelcast.client.spi.impl.ClientInvocationFuture.get(ClientInvocationFuture.java:111)
	at com.hazelcast.client.spi.impl.ClientInvocationFuture.get(ClientInvocationFuture.java:92)
	at com.hazelcast.client.spi.ClientProxy.invoke(ClientProxy.java:147)
	at com.hazelcast.client.proxy.ClientMapProxy.values(ClientMapProxy.java:811)
	at com.hazelcast.client.map.ClientMapUnboundReturnValuesTestSupport.internalRunWithoutException(ClientMapUnboundReturnValuesTestSupport.java:228)
	at 
```
ClientMapUnboundReturnValues_BasicTest.testClientMap_withoutException_NoPreCheck
ClientMapUnboundReturnValues_BasicTest.testClientMap_withoutException_PreCheck
ClientMapUnboundReturnValues_BasicTest.testClientMap_withException_PreCheck
ClientMapUnboundReturnValues_BasicTest.testClientMap_withException_NoPreCheck
ClientMapUnboundReturnValues_BasicTest.com.hazelcast.client.map.ClientMapUnboundReturnValues_BasicTest
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/546/

bilalyasar:
similar failures:
ClientMapUnboundReturnValues_TxnTest.testClientTxnMap_withException_PreCheck
ClientMapUnboundReturnValues_TxnTest.testClientTxnMap_withException_NoPreCheck
ClientMapUnboundReturnValues_TxnTest.testClientTxnMap_withoutException_NoPreCheck
ClientMapUnboundReturnValues_TxnTest.testClientTxnMap_withoutException_PreCheck
ClientMapUnboundReturnValues_TxnTest.com.hazelcast.client.map.ClientMapUnboundReturnValues_TxnTest

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/546/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/547/console
and nightly build goes to timeout due to this test.

bilalyasar:
[...]

PR #5231, opened 2015-05-04
Fix timeout issue of nightly build

Donnerbart:
* Fixed issue that a `QueryResultSizeExceededException` from a `QueryOperation` was not correctly catched, so the `QueryPartitionOperation` fallback was executed
* Fixed some issues with new tests for unbound return values feature and new client (missing constructor in `QueryResultSizeExceededException`)

There are still tests failing, but they shouldn't timeout anymore. So with this PR the nightly build should finish again. We still have to investigate why some of the tests end up with a `TargetDisconnectedException`.

At least some of the tests are working properly now:
new-client-tests

Should fix timeout due to https://github.com/hazelcast/hazelcast/issues/5172.

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapUnboundReturnValuesTestSupport.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapUnboundReturnValuesTestSupport.java
@@ -13,6 +13,8 @@ import com.hazelcast.map.impl.QueryResultSizeLimiter;
 import com.hazelcast.query.TruePredicate;
 import com.hazelcast.transaction.TransactionContext;
 import com.hazelcast.util.EmptyStatement;
+import com.hazelcast.util.ExceptionUtil;
+import org.junit.After;
 
 import java.util.UUID;
 
@@ -20,7 +22,7 @@ import static java.lang.String.format;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
-abstract class ClientMapUnboundReturnValuesTestSupport {
+public abstract class ClientMapUnboundReturnValuesTestSupport {
 
     protected static final int PARTITION_COUNT = 271;
 
@@ -35,6 +37,15 @@ abstract class ClientMapUnboundReturnValuesTestSupport {
 
     private int configLimit;
     private int upperLimit;
+
+    @After
+    public void tearDown() {
+        if (serverMap != null) {
+            serverMap.destroy();
+        }
+        HazelcastClient.shutdownAll();
+        Hazelcast.shutdownAll();
+    }
 
     /**
      * This test calls {@link IMap} methods once which are expected to throw {@link QueryResultSizeExceededException}.
@@ -53,8 +64,6 @@ abstract class ClientMapUnboundReturnValuesTestSupport {
 
         fillToUpperLimit(serverMap, clientMap);
         internalRunWithException(clientMap);
-
-        shutdown(serverMap);
     }
 
     /**
@@ -74,8 +83,6 @@ abstract class ClientMapUnboundReturnValuesTestSupport {
 
         fillToUpperLimit(serverMap, clientMap);
         internalRunWithoutException(clientMap);
-
-        shutdown(serverMap);
     }
 
     /**
@@ -87,8 +94,6 @@ abstract class ClientMapUnboundReturnValuesTestSupport {
         internalSetUpClient(PARTITION_COUNT, 1, 1, PRE_CHECK_TRIGGER_LIMIT_INACTIVE);
 
         internalRunCheckUnsupported(clientMap);
-
-        shutdown(serverMap);
     }
 
     /**
@@ -108,8 +113,6 @@ abstract class ClientMapUnboundReturnValuesTestSupport {
 
         fillToUpperLimit(serverMap, clientMap);
         internalRunTxnWithException(clientMap.getName());
-
-        shutdown(serverMap);
     }
 
     /**
@@ -129,8 +132,6 @@ abstract class ClientMapUnboundReturnValuesTestSupport {
 
         fillToUpperLimit(serverMap, clientMap);
         internalRunTxnWithoutException(clientMap.getName());
-
-        shutdown(serverMap);
     }
 
     private void internalSetUpClient(int partitionCount, int clusterSize, int limit, int preCheckTrigger) {
@@ -178,6 +179,13 @@ abstract class ClientMapUnboundReturnValuesTestSupport {
         assertEquals("Expected map size of client map to match upperLimit", upperLimit, queryMap.size());
     }
 
+    private void checkException(QueryResultSizeExceededException e) {
+        String exception = ExceptionUtil.toString(e);
+        if (exception.contains("QueryPartitionOperation")) {
+            fail("QueryResultSizeExceededException was thrown by QueryPartitionOperation:\n" + exception);
+        }
+    }
+
     private void failExpectedException(String methodName) {
         fail(format("Expected QueryResultSizeExceededException while calling %s with limit %d and upperLimit %d",
                 methodName, configLimit, upperLimit));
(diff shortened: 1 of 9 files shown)


hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapUnboundReturnValuesTestSupport.java
```diff
@@ -178,6 +179,13 @@ private void fillToUpperLimit(IMap<Integer, Integer> fillMap, IMap<Integer, Inte
         assertEquals("Expected map size of client map to match upperLimit", upperLimit, queryMap.size());
     }
 
+    private void checkException(QueryResultSizeExceededException e) {
```
Donnerbart: Checks the exception if it has been thrown by a `QueryPartitionOperation`, which would mean that the query execution happened twice. This happened in some cases because the exception of the `QueryOperation` has not been caught correctly.

hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapUnboundReturnValuesTestSupport.java
```diff
@@ -282,24 +291,26 @@ private void internalRunCheckUnsupported(IMap<Integer, Integer> queryMap) {
...
+            }
+
+            try {
+                txnMap.keySet(TruePredicate.INSTANCE);
+                failExpectedException("TransactionalMap.keySet(predicate)");
+            } catch (QueryResultSizeExceededException e) {
+                checkException(e);
+            }
+        } finally {
+            transactionContext.rollbackTransaction();
```
Donnerbart: Transaction flag for the test thread was not released on test failures.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -90,17 +91,17 @@ public void run() {
...
             }
-            final Credentials credentials = endpoint.getCredentials();
+            Credentials credentials = endpoint.getCredentials();
             interceptBefore(credentials);
             checkPermissions(endpoint);
             processMessage();
             interceptAfter(credentials);
-
         } catch (Throwable e) {
+            logProcessingFailure(e);
```
Donnerbart: We may catch the `QueryResultSizeExceededException` here and suppress the logging of the stack trace.

hazelcast/src/main/java/com/hazelcast/map/QueryResultSizeExceededException.java
```diff
@@ -27,6 +27,10 @@
  */
 public class QueryResultSizeExceededException extends HazelcastException {
 
+    public QueryResultSizeExceededException(String message) {
```
Donnerbart: This missing constructor was the main cause for the timeout issue of the nightly build.

hazelcast/src/main/java/com/hazelcast/map/impl/BasicMapContextQuerySupport.java
```diff
@@ -142,9 +142,10 @@ public Set queryLocalMember(String mapName, Predicate predicate, IterationType i
             if (partitionIds.isEmpty()) {
                 return result;
             }
-        } catch (QueryResultSizeExceededException e) {
-            throw ExceptionUtil.rethrow(e);
         } catch (Throwable t) {
+            if (t.getCause() instanceof QueryResultSizeExceededException) {
```
Donnerbart: The `QueryResultSizeExceededException` is always encapsulated in an `ExecutionException`, so we have to catch it in the right way. See `InvocationFuture.resolveApplicationResponseOrThrowException()`:
```java
        // To obey Future contract, we should wrap unchecked exceptions with ExecutionExceptions.
        throw new ExecutionException((Throwable) response);
```

sancar: @Donnerbart ExceptionUtil.rethrow(t); already calls getCause and throws the cause (if there is one) encapsulated in HazelastException(Runtime Exception). Can you describe more why do we need to do this change ?  Because I have a feeling you are doing the same thing with more code, which is not good.

Donnerbart: It's not the problem that I'm not encapsulating my custom exception. It's the other way around. My custom exception got wrapped with a `ExecutionException`, so I didn't catch it properly.

Originally there was no `ExceptionUtil.rethrow(t);` at all. It was part of the fallback mechanism of the query logic to ignore the first exceptions at all. This is how the code looked in the (mostly duplicated) query logic through the system:
```java
try {
  queryMembers();
} catch (Throwable t) {
  // ignore or at most log
}
try {
  queryPartitions()
} catch (Throwable t) {
  ExceptionUtil.rethrow(t);
}
```
The problem with my `QueryResultSizeExceededException` is that we have to catch it after the `queryMember()` part. Otherwise we will do the whole query work again for each partition (and with the same exception being thrown again).

So what is actually implemented is a catch of the `QueryResultSizeExceededException` with an immediate rethrow:
```java
try {
  queryMembers();
} catch (QueryResultSizeExceededException e) {
  ExceptionUtil.rethrow(e);
} catch (Throwable t) {
  // ignore or at most log
}
try {
  queryPartitions()
} catch (Throwable t) {
  ExceptionUtil.rethrow(t);
}
```
The problem is, that this didn't work, because the `QueryResultSizeExceededException` was always encapsulated in an `ExecutionException`. Because of this I had to move the rethrow in the `catch (Throwable)` and look for the cause.

Donnerbart: So it's not about throwing the right exception (this was already working), but about catching the `QueryResultSizeExceededException` in the correct way.

sancar: I see. I misunderstood it at first.

Donnerbart: You're very welcome. The code looks strange and ugly, so it's good you made a note about that!

sancar: How do you feel about modifying addResultsOfPagingPredicate and addResultsOfPredicate so that they throw only Runtime exceptions like queryOnPartitions? This way we can leave the catching part as is, which is , I think , more readable. (Also same for the collectResults method in TransactionalMapProxySupport)

Donnerbart: The whole code is about preventing the second block (retry on all partitions) if, and only if, the `QueryResultSizeExceededException` was thrown in the first block (query on members).

The problem is, that we don't want to catch and rethrow all exceptions, just my custom exception. I don't know this by heart, but there are situations where the first part will fail (I think some slow `MapLoader` scenarios). So we cannot just catch all exceptions and stop. So we won't get rid of catching my custom exception in all these parts, even if we convert it to another exception type.

Or did I get your idea wrong?

sancar: My suggestion does not change the behavior but only changes where we handle to getCause thingy. My suggestion is that let the getCause done in `BasicMapContextQuerySupport.addResultsOfPagingPredicate ` , `BasicMapContextQuerySupport.addResultsOfPredicate ` and `TransactionalMapProxySupport . collectResults ` as it is done in `BasicMapContextQuerySupport.queryOnPartitions ` via ExceptionUtil.rethrow . Then  you can revert the  catch part that you changed in this pr. This way, these methods will either throw QueryResultSizeExceededException(which will be catched and rethrown) or any other runtime exception which will be logged.

Donnerbart: For me the biggest pain is that we have to catch the exception in so many places, not that the catch itself is not in the most elegant way right now.

So I would prefer to fix this in separate cleanup PR (maybe for 3.6). I'd like to de-duplicate big parts of the shared query logic and find a good solution to deal with this exception then.

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapUnboundReturnValuesTestSupport.java
```diff
@@ -178,6 +179,13 @@ private void fillToUpperLimit(IMap<Integer, Integer> fillMap, IMap<Integer, Inte
         assertEquals("Expected map size of client map to match upperLimit", upperLimit, queryMap.size());
     }
 
+    private void checkException(QueryResultSizeExceededException e) {
```
jerrinot: wat?

Donnerbart: What wat?

This is done to ensure, that we don't execute the query a second time on partitions, after it already failed with `QueryResultSizeExceededException` on members (which may have a serious performance impact). This is one of the bugs which is fixed in this PR.

gurbuzali:
verify

ahmetmircik:
seems mergeable.

Donnerbart:
PR Builder is green. Can anyone else have a look and maybe merge? We need to get the nightly build running again :(

PR #5234, opened 2015-05-05
Ad-Hoc Optimization of the AndPredicate

jerrinot:
Predicates in a format 'attribute >= from AND attribute <= to'
can be safely rewritten as 'attribute BETWEEN from to'.
This yields a dramatic performance improvement when the 'attribute'
has an ordered index.

*Test Scenario*
4 members running on 4 boxes on 1gbit network.
```
com.hazelcast.simulator.tests.map.SqlPredicateTest
keyCount=800000
maxAge=20000
sql=age >= 5000 and age <= 5005
threadCount=1
```

Throughput *before*: 28 ops / s
Throughput *after*:  4,612 ops / s (=saturated network)

This is a demo or teaser showing how much of impact a simple optimization can have.
We should write a proper Optimizer / Planner for next versions of Hazelcast
as the current approach is sub-optimal.

I believe even a simple rule-based optimizer can yield very interesting results.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/Predicates.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/Predicates.java
@@ -30,6 +30,7 @@ import com.hazelcast.query.impl.QueryContext;
 import com.hazelcast.query.impl.QueryableEntry;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
@@ -465,6 +466,11 @@ public final class Predicates {
 
         @Override
         public Set<QueryableEntry> filter(QueryContext queryContext) {
+            Set<QueryableEntry> optoResult = tryOptimizedFiltering(queryContext);
+            if (optoResult != null) {
+                return optoResult;
+            }
+
             Set<QueryableEntry> smallestIndexedResult = null;
             List<Set<QueryableEntry>> otherIndexedResults = new LinkedList<Set<QueryableEntry>>();
             List<Predicate> lsNoIndexPredicates = null;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/query/Predicates.java
```diff
@@ -465,6 +466,11 @@ public AndPredicate(Predicate... predicates) {
 
         @Override
         public Set<QueryableEntry> filter(QueryContext queryContext) {
+            Set<QueryableEntry> optoResult = tryOptimizedFiltering(queryContext);
+            if (optoResult != null) {
+                return optoResult;
+            }
+
             Set<QueryableEntry> smallestIndexedResult = null;
```
pveentjer: It it simpler to just optimize the query? 

Predicate optimizedPredicate = optimize(predicate);

In the optimize you do a tree traversal and replace the expression. Once the predicate is optimized you execute it as it it was written using an in between instead of an or.

jerrinot: that's the way to go; however it's not without consequences - e.g. in the case of an error toString() would display that  BETWEEN predicate was used instead of AND. for now I wanted the least risky change and simplest thing possible demonstrating the effect.

hazelcast/src/main/java/com/hazelcast/query/Predicates.java
```diff
@@ -498,6 +497,82 @@ public AndPredicate(Predicate... predicates) {
...
+            if (!shouldOptimize()) {
+                return null;
+            }
+            return filterOptimized(queryContext);
+        }
+
+        private boolean shouldOptimize() {
+            //currently we have only a very simple optimization for the case when:
+            //attribute >= from AND attribute <= to
+            if (predicates.length != 2) {
```
ajermakovics: Nice. Probably not necessary to be 2 here. There could be more predicates and two of them could get converted to a single between predicate.

jerrinot: right, I'll have a look

hazelcast/src/main/java/com/hazelcast/query/Predicates.java
pveentjer: This rewrite is done during execution. So a user has no idea that his query was rewritten. And even if it was.. I don't see a problem with that.

pveentjer: For demonstrating it is fine.

pveentjer:
@jerrinot what are we going to do with this one? A rule based optimizer is planned for 3.6 and it provide the same functionality as provided in this pr.

jerrinot:
indeed. we can close this.

PR #5239, opened 2015-05-05
Massive speed improvement when using onKeys or KeyPredicate

noctarius:
The way the predicate and keys were handled was more than optimistic :)

edit: Added a key-partition mapping (cache)


noctarius:
verify

noctarius:
verify

noctarius:
verify

PR #5241, opened 2015-05-05
IOBalancer Tests Added and minor change in Connection Registration in IOBalancer

jerrinot:
1. IOBalancer Tests Added
2. Refactoring: - IOBalancer is now a top-level class
3. Different integration point with TcpIpConnectionManager

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
@@ -28,9 +28,6 @@ import com.hazelcast.nio.tcp.OutSelectorImpl;
 import com.hazelcast.nio.tcp.ReadHandler;
 import com.hazelcast.nio.tcp.TcpIpConnection;
 import com.hazelcast.nio.tcp.WriteHandler;
-import com.hazelcast.util.EmptyStatement;
-
-import java.util.concurrent.TimeUnit;
 
 /**
  * It attempts to detect and fix a selector imbalance problem.
@@ -65,7 +62,8 @@ public class IOBalancer {
     private final LoadTracker outLoadTracker;
 
     private final HazelcastThreadGroup threadGroup;
-    private volatile boolean isActive;
+    private volatile boolean shouldStart;
+    private IOBalancerThread ioBalancerThread;
 
     public IOBalancer(InSelectorImpl[] inSelectors, OutSelectorImpl[] outSelectors, HazelcastThreadGroup threadGroup,
                       int migrationIntervalSeconds, LoggingService loggingService) {
@@ -77,7 +75,7 @@ public class IOBalancer {
         this.inLoadTracker = new LoadTracker(inSelectors, loggingService);
         this.outLoadTracker = new LoadTracker(outSelectors, loggingService);
 
-        this.isActive = shouldStart(inSelectors, outSelectors);
+        this.shouldStart = shouldStart(inSelectors, outSelectors);
     }
 
     public void connectionAdded(Connection connection) {
@@ -86,15 +84,14 @@ public class IOBalancer {
         }
 
         ReadHandler readHandler = ((TcpIpConnection) connection).getReadHandler();
-        if (log.isFinestEnabled()) {
-            log.finest("Adding a read handler " + readHandler);
-        }
-        inLoadTracker.addHandler(readHandler);
-
         WriteHandler writeHandler = ((TcpIpConnection) connection).getWriteHandler();
+
         if (log.isFinestEnabled()) {
-            log.info("Adding a write handler " + writeHandler);
+            log.finest("Connection " + connection + " uses read handler " + readHandler + " and write handler "
+                    + writeHandler);
         }
+
+        inLoadTracker.addHandler(readHandler);
         outLoadTracker.addHandler(writeHandler);
     }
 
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
```diff
@@ -77,7 +75,7 @@ public IOBalancer(InSelectorImpl[] inSelectors, OutSelectorImpl[] outSelectors,
         this.inLoadTracker = new LoadTracker(inSelectors, loggingService);
         this.outLoadTracker = new LoadTracker(outSelectors, loggingService);
 
-        this.isActive = shouldStart(inSelectors, outSelectors);
+        this.shouldStart = shouldStart(inSelectors, outSelectors);
```
pveentjer: I like the name active a lot more.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
```diff
@@ -178,26 +179,4 @@ private void tryMigrate(LoadImbalance loadImbalance) {
         }
         handler.migrate(destinationSelector);
     }
-
-    private final class IOBalancerThread extends Thread {
```
pveentjer: Why did you move the thread to the outside? It does almost nothing and it relies on the internals of the balancer.

hazelcast/src/test/java/com/hazelcast/nio/tcp/handlermigration/LoadTrackerTest.java
```diff
@@ -0,0 +1,106 @@
...
+        loadTracker.updateImbalance();
+        assertEquals(100, loadImbalance.minimumEvents);
+        assertEquals(300, loadImbalance.maximumEvents);
+        assertEquals(selector1, loadImbalance.destinationSelector);
+        assertEquals(selector2, loadImbalance.sourceSelector);
+    }
+
+    @Test
+    public void testAddHandler() throws Exception {
+
```
pveentjer: Ahh.. the famous empty line test :)

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
jerrinot: me too, but it's misleading. 
it's set to `true` in a constructor even when it's not active yet. it's active only after calling the `start()` method.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
jerrinot: the original idea was to improve testability. I didn't help great deal, but I like it outside anyway.

hazelcast/src/test/java/com/hazelcast/nio/tcp/handlermigration/LoadTrackerTest.java
jerrinot: ops, good catch, thanks!

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
noctarius: Isn't it enabled and running, started or active?

jerrinot: `enabled` is a good one!

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
```diff
@@ -77,7 +75,7 @@ public IOBalancer(InSelectorImpl[] inSelectors, OutSelectorImpl[] outSelectors,
         this.inLoadTracker = new LoadTracker(inSelectors, loggingService);
         this.outLoadTracker = new LoadTracker(outSelectors, loggingService);
 
-        this.isActive = shouldStart(inSelectors, outSelectors);
+        this.enabled = shouldStart(inSelectors, outSelectors);
```
pveentjer: I would call the method: isEnabled.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancer.java
pveentjer: I'm fine with leaving since it a flavor thing. But the thread is small and pokes in the internal of that class. The thread is tied to the implementation, so I make it internal. Unless it becomes big, then I rip it apart.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/IOBalancerThread.java
```diff
@@ -0,0 +1,55 @@
...
+        this.ioBalancer = ioBalancer;
+        this.log = log;
+        this.migrationIntervalSeconds = migrationIntervalSeconds;
+    }
+
+    @Override
+    public void run() {
+        try {
+            log.finest("Starting IOBalancer thread");
+            while (!Thread.interrupted()) {
```
pveentjer: I never rely on interrupted as flag for a thread to stop. In your case it is ok since you know the InterruptedException isn't swallowed. But it can turn unpredictable quickly.

jerrinot:
verify

Issue #4252, opened 2014-12-08
Replicated Map test failures


enesakar:
See:
https://github.com/hazelcast/hazelcast/issues/3127
https://github.com/hazelcast/hazelcast/issues/3162
https://github.com/hazelcast/hazelcast/issues/4140
https://github.com/hazelcast/hazelcast/issues/4181
https://github.com/hazelcast/hazelcast/issues/4138

mesutcelik:
I don't see those ReplicatedMap Tests failing anymore. any idea?

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/197/testReport/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/198/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/com.hazelcast$hazelcast/200/testReport/junit/com.hazelcast.replicatedmap/ReplicatedMapTest/testEntrySetObjectDelay0/

metanet:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/202/testReport/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/com.hazelcast$hazelcast/207/testReport/junit/com.hazelcast.replicatedmap/ReplicatedMapTest/testSizeObjectDelay0/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/com.hazelcast$hazelcast/208/testReport/junit/com.hazelcast.replicatedmap/ReplicatedMapTest/testSizeBinaryDelay0/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/210/testReport/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/217/

bilalyasar:
[...]

Issue #3631, opened 2014-09-22
replicated map problems


bilalyasar:
for a while, i am examining replicatedMap codes, logic , working principles , here are my findings:

- we couldn't add entryListener for specific key.. EntryEvents are called everytime. see https://gist.github.com/bilalyasar/0e50fbca77d4b5b70a02

- when we add entry with timeout , and when we add entryListener, after timeout, entry.evicted doesn't called. see: https://gist.github.com/bilalyasar/3729a9864d240215205a

- when we add entry with timeout, and after timeout, actually entry is not removed from map. (memory leak)  i dont have test for this, i am tracking this bug with debug mode. 
problem is at: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/AbstractReplicatedRecordStore.java#L114

- javadoc exceptions problems: there are some cases map doesn't throw exceptions. For example: adding null entryListener, this should be revised.

- also i see that, map.clearAll, evictAll events are not implemented. This could be nice feature.

- when we start instance, instance automatically get distributed proxies with FinalizeJoinOperation, and after instance get proxy calls replicatedMap.initialize() method for synchronize replicatedMap. Also this method is called again when we get hazelcast.getReplicatedMap(). maybe some optimizations could be done in this respect.

eminn:
Fixed by #6176

PR #5244, opened 2015-05-05
Replicatedmap speed improvements

noctarius:
Added all lazy views to ReplicatedMap to prevent copying and re-hashing (HashSet) while retrieving values - speed improvement ~100x

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazyCollection.java
+++ b/hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazyCollection.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.replicatedmap.impl.record;
+
+import com.hazelcast.replicatedmap.impl.record.LazySet.IteratorFactory;
+
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map.Entry;
+
+class LazyCollection<K, V>
+        implements Collection<V> {
+
+    private final InternalReplicatedMapStorage<K, V> storage;
+    private final IteratorFactory<K, V, V> iteratorFactory;
+    private final Collection<ReplicatedRecord<K, V>> values;
+
+    public LazyCollection(IteratorFactory<K, V, V> iteratorFactory, InternalReplicatedMapStorage<K, V> storage) {
+        this.iteratorFactory = iteratorFactory;
+        this.values = storage.values();
+        this.storage = storage;
+    }
+
+    @Override
+    public int size() {
+        return values.size();
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return values.isEmpty();
+    }
+
+    @Override
+    public boolean contains(Object value) {
+        throw new UnsupportedOperationException("LazySet does not support contains requests");
+    }
+
+    @Override
+    public Iterator<V> iterator() {
+        Iterator<Entry<K, ReplicatedRecord<K, V>>> iterator = storage.entrySet().iterator();
+        return iteratorFactory.create(iterator);
+    }
+
+    @Override
+    public Object[] toArray() {
+        List<Object> result = new ArrayList<Object>(storage.values().size());
+        Iterator<V> iterator = iterator();
+        while (iterator.hasNext()) {
+            result.add(iterator.next());
+        }
+        return result.toArray(new Object[result.size()]);
+    }
+
+    @Override
+    public <T> T[] toArray(T[] a) {
+        List<Object> result = new ArrayList<Object>(storage.values().size());
+        Iterator<V> iterator = iterator();
+        while (iterator.hasNext()) {
+            result.add(iterator.next());
+        }
+        if (a.length != result.size()) {
+            a = (T[]) Array.newInstance(a.getClass().getComponentType(), result.size());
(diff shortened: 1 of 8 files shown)


noctarius:
verify

dbrimley:
This is an important improvement for us.

metanet:
Isn't this PR changing the public API? values(), keySet(), entrySet() methods return modifiable collections in the master branch. This PR makes them return immutable collections.

noctarius:
The documentation already states it is a non-backed collection / set and will not reflect back to the storage but yeah old implementation was mutable in the old version but since API is still marked as @BETA this is ok but I guess we can remove this beta flag in 3.5.

metanet:
I think being not-backed has nothing to do with changing the API since it is still same with the master branch. A user can still modify the collection while he knows that it is not reflected to the internal map.

noctarius:
As I mentioned, ReplicatedMap is still marked as beta which means exactly that changes to the public API might happen in the near future. But yes we should mention it in the documentation.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazyCollection.java
```diff
@@ -0,0 +1,110 @@
...
+        List<Object> result = new ArrayList<Object>(storage.values().size());
+        Iterator<V> iterator = iterator();
+        while (iterator.hasNext()) {
+            result.add(iterator.next());
+        }
+        return result.toArray(new Object[result.size()]);
+    }
+
+    @Override
+    public <T> T[] toArray(T[] a) {
```
jerrinot: this smells like a ClassCastException to me

metanet:
I am not sure about removing the BETA annotation. There are a few chronical test failures of ReplicatedMap such as https://github.com/hazelcast/hazelcast/issues/3631 https://github.com/hazelcast/hazelcast/issues/4252 
I am not an authority here but still I think we should solve the problems first before making it non beta.

mesutcelik:
I think Basri is right. We shouldn't remove **beta** before fixing test failures. They are there since it is released and seems very hard to fix without big architectural changes. We have a PRD for that in 3.6

noctarius:
Fine with that! :)

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazySet.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return storage.isEmpty();
+    }
+
+    @Override
+    public boolean contains(Object o) {
+        throw new UnsupportedOperationException("LazySet does not support contains requests");
```
ahmetmircik: Can you please point out why `contains` is unsupported? Can't we provide at least read-only method impls ?

noctarius: Sure by iterating over all values since the wrapped iterator returns a ReplicatedRecord and not the real value.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazyCollection.java
noctarius: good point :D

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazySet.java
noctarius: I looked into it and I don't think we should implement it, it is O(n) if you call this method, if a user calls it more often it will be horrible for the applications runtime. It is better to throw the exception and let the user copy the result set.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazyCollection.java
noctarius: fixed

noctarius:
verify

noctarius:
verify

metanet:
PR note says that a speed improvement ~100x is provided. This is great. What it does is basically preventing a collection copy operation. So when I have 10 items in my map, I won't have a significant speed improvement. When I have 1M items, I will probably have an improvement much more than 100x, maybe 1000 or 10000x. Maybe ~100x is provided as an average value, but it is very fuzzy. Still, I am OK with that since it obviously provides a performance improvement. 

In fact, my concern is as follows: This PR causes a significant change in the behaviour. It returns implementations of Collection interface from the methods keySet(), entrySet() and values(). But those implementations are basically iterators. So what it does is returning iterators instead of collections. It is returned as collection only because the API enforces it. It seems to me as a design issue. Can we consider adding new methods to the ReplicatedMap interface and returning iterators from these new methods?

noctarius:
No we can't since existing frameworks like Google Guava expect those methods.

noctarius:
verify

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/LazySet.java
Tembrel: Since LazySet isn't a public type, it'll be important to document, where public methods __do__ return lazy views, that users will have to copy these views whenever they want to perform any access other than simple iteration. This would be true whether or not `contains` is supported, but I think it's even more important when a containment test on a returned set can throw a runtime exception.

noctarius: it is in the javadoc of the replicated map

noctarius: Oh you mean to document that they have to do a copy, isn't it obvious if you try to call contains and it throws an exception? :) That's the reason I chose exception over "just do it".

Tembrel: It is? I don't see anything in the replicated map javadoc about the behavior of the returned lazy sets, other than that they are not backed by the map and don't promise any special ordering. It doesn't mention performance issues or runtime exceptions due to unsupported operations. I don't think the word "lazy" adequately conveys these extra restrictions.

metanet: I think @Tembrel has a point. Users should understand it from the documentation that they must copy the collection before doing any action except iterating. I think it is easier than running the code and getting an exception.

Tembrel: I know the ideal is to test every code path, but it's easy to see how in practice a stray call to `contains` might get missed. Adding a note in the public API about it being unsupported could help avoid some unpleasantness.

PR #5245, opened 2015-05-05
Use spaces everywhere in the file and fix indentation

henri-tremblay:
Just a little fix because it was annoying me. You probably have the same kind of issue in other files. For instance, server.bat have two tabs, but there are no lines indented with spaces, the file is still indented correctly.


pveentjer:
Since this is just a change in a bash file; I don't think we need to rely on the CI.

@gurbuzali @mesutcelik I think we can just merge this one.

PR #5246, opened 2015-05-06
Fixes regarding lock cleanup and lease expiration during member-left (network split)

mdogan:
- Lock cleanup task submitted when a member is left should be executed on related partition threads, otherwise it can miss previously enqueued or being executed lock operations. So, this race can cause infinitely hanging locks.

- Lock cleanup operation should check owner-uuid during its execution in partition thread to avoid race.

- Added "reference-call-id" to lock operation to provide operation retry idempotence. When a member crashes after processing a lock request and sending its backup, caller will retry the lock operation. If it happens that backup member already processed the lock request, then caller will acquire lock resource 2nd time (because it's reentrant) unintentionally. When it releases the lock only once, lock will hang infinitely. To avoid that, now lock operation and lock service always keep the original/reference call-id for a specific lock resource and if it's tried to be acquired multiple times with the same reference-call-id, lock request will just return success instead of processing it again.

- Introduced max-lease-time (default is Integer.MAX seconds) for locks. When lock is acquired without an explicit lease-time then max-lease-time setting will be used. Caller can use a lower value but not a greater one. When a greater lease-time is given, invocation/operation will fail immediately.

- Fixed a bug which lease-time is not set correctly on backups and when backup is promoted, lease handling process doesn't kicks in.

- Added a safety unlock call when a lock invocation timeouts to decrease chance of hanging locks.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -17,6 +17,7 @@
 package com.hazelcast.map.impl.proxy;
 
 import com.hazelcast.concurrent.lock.LockProxySupport;
+import com.hazelcast.concurrent.lock.LockServiceImpl;
 import com.hazelcast.config.EntryListenerConfig;
 import com.hazelcast.config.ListenerConfig;
 import com.hazelcast.config.MapConfig;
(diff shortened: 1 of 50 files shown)


mdogan:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -156,7 +157,14 @@ protected MapProxySupport(final String name, final MapService service, NodeEngin
...
         localMapStats = mapServiceContext.getLocalMapStatsProvider().getLocalMapStatsImpl(name);
         this.partitionService = getNodeEngine().getPartitionService();
-        lockSupport = new LockProxySupport(new DefaultObjectNamespace(MapService.SERVICE_NAME, name));
+
+        LockService lockService = nodeEngine.getSharedService(LockService.SERVICE_NAME);
+        if (lockService != null) {
+            lockSupport = new LockProxySupport(new DefaultObjectNamespace(MapService.SERVICE_NAME, name),
+                    lockService.getMaxLeaseTimeInMillis());
+        } else {
+            lockSupport = null;
```
ahmetmircik: Afais `IMap#lock` may throw NPE after this change because of `lockSupport  = null` case.  That case is a new thing introduced with this PR? intentional?

mdogan: If `LockService ` is not enabled, either way call will fail. I don't think this is a new thing. I can make failure message more friendly.

pveentjer:
What are we going to do with this PR?

pveentjer:
I don't see a single test add. Also it is better to split this task up in 6 pr's (for each bullet point one). There is just too much bug fixing and cleanup in a single pr; especially regarding something as complicated as distributed locks in combination with split brain.

mdogan:
Closed in favour of #5601, #5602 and #5603

PR #5247, opened 2015-05-06
Fixes & enhancements regarding heartbeat mechanims and split brain handling

mdogan:
- During split-brain merge process "ClusterServiceImpl.merge()", cluster, partitions should be reset first, then all connections should be closed and all invocations from current member should be notified to retry. Then after gathering merge-tasks from related services, all services also should be reset. After that connection manager will be started and node will join target cluster. This will make invocations race free during merge process.

- Only master node should answer to the split-brain merge requests. Other members should silently ignore the request and optionally log it.

- Non-master members should send master confirmation as soon as possible after they detect a master change. This is just to speed up and help detection of possible network split. Since if a master member receives a master confirmation from a non-member endpoint, it sends a "remove-me-from-your-cluster" request to the caller.

- Master confirmations should be reset only when a node becomes master after old master disconnects.

- Member UUID should not change when it's merging to a cluster, since this makes very hard to detect a member's uuid for uuid-sensitive operations like locks.

- Split brain handling process improved and changed significantly.  Instead of just comparing cluster member count between clusters, complete member-lists will be compared, if at least one member is common between two clusters, clusters will reject to merge. They will not try to merge until they have completely different members in their cluster.  Eventually both cluster will diverge because of heartbeat and/or master-confirmation timeouts.

- Heartbeat mechanism is changed to explicitly send heartbeats instead of relying read and write timestamps of other members. Since, it can be that other node is replying with an error response (like caller-not-member-exception), but currently caller node assumes it's heartbeating implicitly. And this causes one side to assume they are in the same cluster but practically it cannot execute operations on other member.

- Heartbeat and master-confirmation messages, now, include a timestamp to prevent a very old message (older than timeout/2) to be processed.

- InternalPartitionService should not process a partition state which is sent from a non-master member, as it's doing for non-member right now.


mdogan:
depends on fix: https://github.com/hazelcast/hazelcast/pull/5259

hazelcast/src/main/java/com/hazelcast/spi/impl/transceiver/impl/PacketTransceiverImpl.java
```diff
@@ -66,10 +65,6 @@ public boolean transmit(Packet packet, Connection connection) {
         if (connection == null || !connection.isAlive()) {
             return false;
         }
```
pveentjer: I like the following section!!

You can even simplify it to

```
If(connection = null) return false;
return connection.write(packet)
```

Since the write will do an alive check anyway.

mdogan:
Closed in favour of #5606, #5607, #5609 and #5610

PR #5248, opened 2015-05-06
Fixes and enhancements regarding invocation system

### LINKED ISSUE #5261: ILock acquired although lock() throws OperationTimeoutException
Hi,

we encountered an issue where calling the lock method on an ILock threw an OperationTimeoutException but the lock was still acquired, which is bad because the usual locking idiom is not correct:

```java
lock.lock() // If this throws an OperationTimeoutException, the lock will never be unlocked
try {
  // Critical section
} finally {
  lock.unlock();
}
```

It seems that this can happen in Hazelcast 3.3 if a lock operation is queued on the callee side, but not executed for 2*operation timeout. We don't know yet what caused this delay (many operations enqueued, operation execution slow?) in our case, but it seemed to happen. 

Since the caller does not receive a WAIT_RESPONSE in this case, he will eventually check if the operation is executing, which is not the case, and fail with an OperationTimeoutException.

The lock operation is still enqueued on the callee side and will eventually be executed, resulting in the lock being acquired without the caller knowing.

It seems that this behaviour has been changed in the meantime, but I implemented a unit test and it also has problems in Hazelcast 3.4.1 (I couldn't test with Hazelcast 3.4.2 yet):

The test case reproduces the problem using a blocking store implementation to artificially delay the execution of the lock operation and provoke the OperationTimeoutException. It fails for Hazelcast 3.3, but not for Hazelcast 3.4.1 if the lock operation is executed on a remote node. Interestingly, if you run only a single node, and the lock operation is thus local, the test is successful with Hazelcast 3.3, but fails with Hazelcast 3.4.1. And in the case where it is successful for Hazelcast 3.4.1, there is still a FATAL log message:

```
40252 [hz._hzInstance_2_dev.response] FATAL com.hazelcast.spi.OperationService  - [10.0.1.49]:5701 [dev] [3.4.1] While processing response...
com.hazelcast.core.HazelcastException: No invocation for response: com.hazelcast.spi.impl.CallTimeoutResponse@2de3e7d2
	at com.hazelcast.spi.impl.BasicOperationService$ResponsePacketHandler.notifyRemoteCall(BasicOperationService.java:602)
	at com.hazelcast.spi.impl.BasicOperationService$ResponsePacketHandler.handle(BasicOperationService.java:586)
	at com.hazelcast.spi.impl.BasicOperationService$ResponsePacketHandler.access$1300(BasicOperationService.java:579)
	at com.hazelcast.spi.impl.BasicOperationService$BasicDispatcherImpl.dispatch(BasicOperationService.java:564)
	at com.hazelcast.spi.impl.BasicOperationScheduler$ResponseThread.process(BasicOperationScheduler.java:532)
	at com.hazelcast.spi.impl.BasicOperationScheduler$ResponseThread.doRun(BasicOperationScheduler.java:524)
	at com.hazelcast.spi.impl.BasicOperationScheduler$ResponseThread.run(BasicOperationScheduler.java:501)
```

The test (actually two, one for local and one for remote execution) is the following:

```java
import com.hazelcast.config.MapConfig;
import com.hazelcast.config.MapStoreConfig;
import com.hazelcast.core.*;
import java.util.*;

/**
 * A test to provoke an operation timeout exception when trying to acquire a lock, and see if the
 * lock is actually not acquired if a timeout exception happens.
 */
public class TestLock2 extends TestCase {

	private static final Logger logger = Logger.getLogger(TestLock2.class);

	private static final String MAP_NAME = "DummyMap";
	private static final long OPERATION_CALL_TIMEOUT = 3000;

	private boolean lockException;

	@Override
	public void setUp() throws Exception {

		// configure logging
		}
		
		lockException = false;

	}

	public void testLock2Local() throws InterruptedException {
		testLock2(false);
	}
	
	public void testLock2Remote() throws InterruptedException {
		testLock2(true);
	}
	
	private void testLock2(boolean remote) throws InterruptedException {

		// create hazelcast config
		config.setProperty("hazelcast.operation.call.timeout.millis", String.valueOf(OPERATION_CALL_TIMEOUT));

		// create shared map store implementation
		BlockingMapStore store = new BlockingMapStore();
		// configure map store
		MapStoreConfig mapStoreConfig = new MapStoreConfig();
		mapStoreConfig.setEnabled(true);
		mapStoreConfig.setClassName(null);
		mapStoreConfig.setImplementation(store);
		MapConfig mapConfig = config.getMapConfig(MAP_NAME);
		mapConfig.setMapStoreConfig(mapStoreConfig);

		// start a hazelcast instances
		HazelcastInstance hcInstance2 = null;
		if (remote) {
			hcInstance2 = Hazelcast.newHazelcastInstance(config);
			while (hcInstance1.getCluster().getMembers().size() < 2) {
				Thread.sleep(250);
			}
		}

		logger.info("Hazelcast instances started.");
		try {
			
			String curKey = Double.toString(Math.random());
			if (remote) {
				// find a remote key to ensure remote execution
				while (hcInstance1.getPartitionService().getPartition(curKey).getOwner().localMember()) {
					curKey = Double.toString(Math.random());
				}
			}
			final String key = curKey;
			
			final IMap<String, String> map = hcInstance1.getMap(MAP_NAME);
			final ILock lock = hcInstance1.getLock(key);

			// block partition operation thread for key with a put operation
			new Thread(new Runnable() {

				@Override
				public void run() {
					map.put(key, "value");
				}

			}).start();

			try {
				Thread.sleep(1000);
			} catch (InterruptedException e1) {
				e1.printStackTrace();
			}

			// start lock operation
			new Thread(new Runnable() {

				@Override
				public void run() {
					logger.info("before acquire lock");
					try {
						lock.lock();
					} catch (OperationTimeoutException e) {
						lockException = true;
						logger.info("exception in lock", e);
					}
					if (!lockException) {
						logger.info("acquired lock");
					}
				}

			}).start();

			// wait for operation timeout exception to occur
			try {
				Thread.sleep(2 * OPERATION_CALL_TIMEOUT + 1000);
				e.printStackTrace();
			}

			logger.info("all operations started");
			logger.info("release blocking store");
			// let the put operation proceed, such that the lock operation will be executed
			store.release();

			try {
				Thread.sleep(1000);
				e.printStackTrace();
			}

			boolean locked = lock.isLocked();
			System.out.println("Lock locked: " + locked);
			assertTrue("Lock is locked although lock operation failed with an OperationTimeoutException", !(lockException && locked));

		} finally {
			hcInstance1.shutdown();
			if (remote) {
				hcInstance2.shutdown();
			}
		}

	}
	
	/**
	 * Map store that blocks put operations until {@link #release()} is called.
	 */
	private static class BlockingMapStore implements MapStore<String, String> {

		private boolean block = true;
		
		public synchronized void release() {
			block = false;
			this.notify();
		}
		
		@Override
		public String load(String key) {
			return null;
		}

		@Override
		public Map<String, String> loadAll(Collection<String> keys) {
			return new HashMap<String, String>();
		}

		@Override
		public Set<String> loadAllKeys() {
			return new HashSet<String>();
		}

		@Override
		public synchronized void store(String key, String value) {
			while (block) {
				try {
					this.wait();
					throw new RuntimeException(e);
				}
			}
		}

		@Override
		public void storeAll(Map<String, String> map) {
			// do nothing
		}

		@Override
		public void delete(String key) {
			// do nothing
		}

		@Override
		public void deleteAll(Collection<String> keys) {
			// do nothing
		}
		
	}

}
```

Thanks for having a look!

Cheers,
Andreas

mdogan:
- When an invocation timeout is detected (original invocation or its backup), it should be removed from registry. Normally this is done when response is set, but if response is set (either original response just arrives or member left happens) before operation timeout response, invocation can be left in registry.

- Added `response != null` check before throwing operation-timeout-exception to prevent getting a timeout when invocation is just retried.

- `ResponseAlreadySentException` cannot be propagated to the caller, it should be logged on remote side, otherwise this causes error loop.

- Optimized multiple backups to serialize backup operation only once instead of multiple times for all backup replicas.

- Prevent invocation retry if a response is already set to future. This fixes #5261.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
@@ -534,7 +534,7 @@ public abstract class Operation implements DataSerializable {
     @Override
     public String toString() {
         final StringBuilder sb = new StringBuilder(getClass().getName()).append('{');
-        sb.append("serviceName='").append(serviceName).append('\'');
+        sb.append("serviceName='").append(getServiceName()).append('\'');
         sb.append(", partitionId=").append(partitionId);
         sb.append(", callId=").append(callId);
         sb.append(", invocationTime=").append(invocationTime);
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
@@ -144,8 +144,11 @@ final class InvocationFuture<E> implements InternalCompletableFuture<E> {
      * 'set' calls are ignored.
      *
      * @param offeredResponse The type of response to offer.
+     * @return <tt>true</tt> if offered response, either a final response or an internal response,
+     * is set/applied, <tt>false</tt> otherwise. If <tt>false</tt> is returned, that means offered response is ignored
+     * because a final response is already set to this future.
      */
-    public void set(Object offeredResponse) {
+    public boolean set(Object offeredResponse) {
         assert !(offeredResponse instanceof Response) : "unexpected response found: " + offeredResponse;
 
         if (offeredResponse == null) {
@@ -167,12 +170,12 @@ final class InvocationFuture<E> implements InternalCompletableFuture<E> {
                 }
 
                 operationService.invocationsRegistry.deregister(invocation);
-                return;
+                return false;
             }
 
             response = offeredResponse;
             if (offeredResponse == WAIT_RESPONSE) {
-                return;
+                return true;
             }
             callbackChain = callbackHead;
             callbackHead = null;
@@ -183,6 +186,7 @@ final class InvocationFuture<E> implements InternalCompletableFuture<E> {
 
 
         notifyCallbacks(callbackChain);
+        return true;
     }
 
     private void notifyCallbacks(ExecutionCallbackNode<E> callbackChain) {
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
@@ -165,6 +165,8 @@ final class InvocationFuture<E> implements InternalCompletableFuture<E> {
                     logger.finest("Future response is already set! Current response: "
                             + response + ", Offered response: " + offeredResponse + ", Invocation: " + invocation);
                 }
+
+                operationService.invocationsRegistry.deregister(invocation);
                 return;
             }
 
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
@@ -267,6 +267,9 @@ final class InvocationFuture<E> implements InternalCompletableFuture<E> {
                     boolean executing = operationService.getIsStillRunningService().isOperationExecuting(invocation);
                     if (!executing) {
                         Object operationTimeoutException = invocation.newOperationTimeoutException(pollCount * pollTimeoutMs);
+                        if (response != null) {
+                            continue;
+                        }
                         // tries to set an OperationTimeoutException response if response is not set yet
                         set(operationTimeoutException);
                     }
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationTimeoutTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationTimeoutTest.java
@@ -1,32 +1,45 @@
 package com.hazelcast.spi.impl.operationservice.impl;
 
+import com.hazelcast.concurrent.lock.InternalLockNamespace;
+import com.hazelcast.concurrent.lock.operations.IsLockedOperation;
+import com.hazelcast.concurrent.lock.operations.LockOperation;
 import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.ILock;
 import com.hazelcast.core.IQueue;
+import com.hazelcast.core.OperationTimeoutException;
 import com.hazelcast.core.Partition;
 import com.hazelcast.instance.GroupProperties;
 import com.hazelcast.map.impl.MapService;
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.spi.AbstractOperation;
+import com.hazelcast.spi.InternalCompletableFuture;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.OperationService;
 import com.hazelcast.spi.PartitionAwareOperation;
+import com.hazelcast.spi.impl.NodeEngineImpl;
+import com.hazelcast.spi.impl.operationservice.InternalOperationService;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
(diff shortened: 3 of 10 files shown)


mdogan:
verify

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -458,7 +458,7 @@ public final void readData(ObjectDataInput in) throws IOException {
     @Override
     public String toString() {
         final StringBuilder sb = new StringBuilder(getClass().getName()).append('{');
-        sb.append("serviceName='").append(serviceName).append('\'');
+        sb.append("serviceName='").append(getServiceName()).append('\'');
```
pveentjer: What is the added value of calling the getter? It is a field in the Operation without any form of logic. Also it isn't in line with the other fields being used in this toString method.

mdogan: `serviceName` is null by default and it's useless to print null here. Most of the operations are overriding `getServiceName()` to return a hardcoded service-name.

pveentjer: homer-simpson-quotes-doh-i11

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
```diff
@@ -144,8 +144,9 @@ public void run() {
      * 'set' calls are ignored.
      *
      * @param offeredResponse The type of response to offer.
+     * @return <tt>true</tt> is offered response is set, <tt>false</tt> otherwise.
```
pveentjer: Can you add a bit more explanation of what 'set' means. The issue here is the 'WAIT_RESPONSE'; it is set.. but it isn't the final response for the invocation.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationTimeoutTest.java
```diff
@@ -217,4 +231,73 @@ private boolean isInterruptionCaught() {
...
+        long callTimeout = 3000L;
+        LongRunningOperation longRunningOperation = new LongRunningOperation(2 * callTimeout + 1000);
+        InternalOperationService operationService = nodeEngine.getOperationService();
+        operationService.invokeOnPartition(null, longRunningOperation, partitionId);
+
+        InternalCompletableFuture<Object> future = operationService.createInvocationBuilder(null,
+                new LockOperation(new InternalLockNamespace(key), nodeEngine.toData(key), 1, -1), partitionId)
+                .setCallTimeout(callTimeout).invoke();
+
+        boolean timedOut;
```
pveentjer: What is the path you expect to get executed? Afaik it should be the OperationTimeoutException since that is the name of the test method. So why the logic for the 'timeOut=false'?

pveentjer:
Minor review comments.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
mdogan: New version:
> @return <tt>true</tt> if offered response, either a final response or an internal response, is set/applied, <tt>false</tt> otherwise. If <tt>false</tt> is returned, that means offered response is ignored because a final response is already set to this future.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationTimeoutTest.java
mdogan: It should be `OperationTimeoutException` but I don't remember why other path is there :) Probably  a left over from when I was playing with different timeouts scenarios. Removed the other path.

pveentjer:
+1

mdogan:
verify

gurbuzali:
do we need to backport fixes ?

PR #5250, opened 2015-05-06
Cleanup of doc build script

Donnerbart:
Cleanup of build script:
* re-ordering of functions
* format of functions
* usage of ${variable}

The functionality should not have changed. The file should just be more readable and all functions should be in the right order. Also the warnings in IDEA were reduced.


Donnerbart:
verify

PR #5252, opened 2015-05-06
Make sure hazelcast can be launched from any directory

henri-tremblay:
Currently, server.sh and server.bat can only be called from HAZELCAST_HOME. This fix allows it to be called from anywhere.


pveentjer:
Thanks Henri.

PR #5254, opened 2015-05-07
Better LazySet/Collection documentation

noctarius:
Updated the documentation for ReplicatedMap::entrySet, ReplicatedMap::values and ReplicatedMap::keySet as discussed yesterday (in PR #5244).


noctarius:
@Tembrel @metanet do you like this or do you think something should still be added? :)
PS: thanks for the comments yesterday, you're totally right, it was missing!

pveentjer:
This will not be added in the EA release. Perhaps in the next EA; we need to see.

Tembrel:
This is great, thanks for doing this.

Minor suggestion: `would result in extremely poor runtime if it is called multiple times in a row` could be rephrased as `would result in very poor performance if called repeatedly`.

noctarius:
like it!

noctarius:
verify

gurbuzali:
committed manually since there was a conflict

PR #5255, opened 2015-05-07
Cluster uptime and id is sent to call home.

kobalski:
This PR adds cluster start time into ClusterClock. Start time of the cluster is set when a master node is assigned in cluster and when a new node makes join request masters start time value is sent to joined node and saved in it ClusterClock. Start time value will be used to calculate uptime of the cluster and will be sent to call home. Also, clusterId sent to call home before was GroupName property and it was not unique value. With this PR clusterId will be concatenation of GroupName and start time. It will provide a pseudo uniqueness since possibility of two clusters starting at the same milliseconds is really low.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/ClusterClock.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/ClusterClock.java
@@ -20,12 +20,20 @@ public interface ClusterClock {
 
     /**
      * Returns the cluster-time in milliseconds.
-     *
      * The cluster-time measures elapsed time since the cluster was created. Comparable to the {@link System#nanoTime()}.
-     *
      * @return the cluster-time (elapsed milliseconds since the cluster was created).
      */
     long getClusterTime();
 
     long getClusterTimeDiff();
+
+    /**
+     * Returns the cluster  up-time in milliseconds.
+     * When first node in cluster becomes master, its clusterTime value is saved as
+     * clusterStartTime. Up-time of cluster is calculated as
+     * ClusterClock#getClusterTime - clusterStartTime.
+     *
+     * @return the cluster-up-time
+     */
+     long getClusterUpTime();
 }
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
@@ -20,10 +20,12 @@ import com.hazelcast.cluster.ClusterClock;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.util.Clock;
 
+
 public class ClusterClockImpl implements ClusterClock {
 
     private final ILogger logger;
     private volatile long clusterTimeDiff = Long.MAX_VALUE;
+    private volatile long clusterStartTime = Long.MIN_VALUE;
 
     public ClusterClockImpl(ILogger logger) {
         this.logger = logger;
@@ -46,4 +48,20 @@ public class ClusterClockImpl implements ClusterClock {
     public long getClusterTimeDiff() {
         return (clusterTimeDiff == Long.MAX_VALUE) ? 0 : clusterTimeDiff;
     }
+
+    @Override
+    public long getClusterUpTime() {
+        return Clock.currentTimeMillis() - clusterStartTime;
+    }
+
+    public void setClusterStartTime(long startTime) {
+        if (this.clusterStartTime == Long.MIN_VALUE) {
+            this.clusterStartTime = startTime;
+        }
+    }
+
+    public long getClusterStartTime() {
+       return clusterStartTime;
+    }
+
 }
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -172,6 +172,8 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
 
     private final ClusterClockImpl clusterClock;
 
+    private String clusterId = null;
+
     public ClusterServiceImpl(final Node node) {
         this.node = node;
         nodeEngine = node.nodeEngine;
@@ -198,6 +200,17 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
     @Override
     public ClusterClockImpl getClusterClock() {
         return clusterClock;
+    }
+
+    @Override
+    public String getClusterId() {
+        return clusterId;
+    }
+
+    public void setClusterId(String clusterId) {
+        if (this.clusterId == null) {
+            this.clusterId = clusterId;
+        }
     }
 
     @Override
@@ -976,7 +989,9 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
                 final int count = members.size() - 1 + setJoins.size();
                 final List<Future> calls = new ArrayList<Future>(count);
                 for (MemberInfo member : setJoins) {
-                    calls.add(invokeClusterOperation(new FinalizeJoinOperation(memberInfos, postJoinOp, time), member.getAddress()));
+                    final long startTime = clusterClock.getClusterStartTime();
+                    calls.add(invokeClusterOperation(new FinalizeJoinOperation(memberInfos, postJoinOp, time,
+                            clusterId, startTime), member.getAddress()));
                 }
                 for (MemberImpl member : members) {
                     if (!member.getAddress().equals(thisAddress)) {
--- a/hazelcast/src/main/java/com/hazelcast/util/UuidUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/UuidUtil.java
@@ -55,6 +55,10 @@ public final class UuidUtil {
         }
     }
 
+    public static String createClusterUuid() {
+        return  buildRandomUuidString();
+    }
+
     public static String createMemberUuid(Address endpoint) {
         return buildRandomUUID().toString();
     }
--- a/hazelcast/src/main/java/com/hazelcast/util/VersionCheck.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/VersionCheck.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.util;
 
+import com.hazelcast.cluster.impl.ClusterServiceImpl;
 import com.hazelcast.config.NativeMemoryConfig;
 import com.hazelcast.core.ClientType;
 import com.hazelcast.instance.Node;
@@ -24,8 +25,11 @@ import com.hazelcast.nio.IOUtil;
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
 import java.net.URL;
 import java.net.URLConnection;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
@@ -92,7 +96,8 @@ public final class VersionCheck {
 
     }
 
-    private void doCheck(Node hazelcastNode, String version, boolean isEnterprise) {
+    public Map<String, String> doCheck(Node hazelcastNode, String version, boolean isEnterprise) {
+
         String downloadId = "source";
         InputStream is = null;
         try {
@@ -110,29 +115,36 @@ public final class VersionCheck {
 
         //Calculate native memory usage from native memory config
         NativeMemoryConfig memoryConfig = hazelcastNode.getConfig().getNativeMemoryConfig();
-        long totalNativeMemorySize = hazelcastNode.getClusterService().getSize()
+        final ClusterServiceImpl clusterService = hazelcastNode.getClusterService();
+        long totalNativeMemorySize = clusterService.getSize()
                 * memoryConfig.getSize().bytes();
         String nativeMemoryParameter = (isEnterprise)
                 ? Long.toString(MemoryUnit.BYTES.toGigaBytes(totalNativeMemorySize)) : "0";
         //Calculate connected clients to the cluster.
         Map<ClientType, Integer> clusterClientStats = hazelcastNode.clientEngine.getConnectedClientStats();
+        RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
 
-        UrlActionParameterCreator parameterCreator = new UrlActionParameterCreator();
+        Long clusterUpTime = clusterService.getClusterClock().getClusterUpTime();
+
+        VersionCheckParameterCreator parameterCreator = new VersionCheckParameterCreator();
         parameterCreator.addParam("version", version);
         parameterCreator.addParam("m", hazelcastNode.getLocalMember().getUuid());
         parameterCreator.addParam("e", Boolean.toString(isEnterprise));
         parameterCreator.addParam("l", MD5Util.toMD5String(hazelcastNode.getConfig().getLicenseKey()));
         parameterCreator.addParam("p", downloadId);
-        parameterCreator.addParam("c", MD5Util.toMD5String(hazelcastNode.getConfig().getGroupConfig().getName()));
-        parameterCreator.addParam("crsz", convertToLetter(hazelcastNode.getClusterService().getMembers().size()));
+        parameterCreator.addParam("c", clusterService.getClusterId());
+        parameterCreator.addParam("crsz", convertToLetter(clusterService.getMembers().size()));
         parameterCreator.addParam("cssz", convertToLetter(hazelcastNode.clientEngine.getClientEndpointCount()));
         parameterCreator.addParam("hdgb", nativeMemoryParameter);
         parameterCreator.addParam("ccpp", Integer.toString(clusterClientStats.get(ClientType.CPP)));
         parameterCreator.addParam("cdn", Integer.toString(clusterClientStats.get(ClientType.CSHARP)));
         parameterCreator.addParam("cjv", Integer.toString(clusterClientStats.get(ClientType.JAVA)));
-
+        parameterCreator.addParam("cuptm", Long.toString(clusterUpTime));
+        parameterCreator.addParam("nuptm", Long.toString(runtimeMxBean.getUptime()));
         String urlStr = BASE_VERSION_CHECK_URL + parameterCreator.build();
         fetchWebService(urlStr);
+
+        return parameterCreator.getParameters();
     }
 
     private void fetchWebService(String urlStr) {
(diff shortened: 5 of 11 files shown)


kobalski:
verify

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
```diff
@@ -24,6 +24,7 @@
 
     private final ILogger logger;
     private volatile long clusterTimeDiff = Long.MAX_VALUE;
+    private volatile long clusterStartTime = Long.MIN_VALUE;
```
mesutcelik: what is the reason to set it to `MIN_VALUE` ?

hazelcast/src/main/java/com/hazelcast/cluster/ClusterClock.java
```diff
@@ -28,4 +28,13 @@
     long getClusterTime();
 
     long getClusterTimeDiff();
+
+    /**
+     * Returns the cluster start time in milliseconds.
+     *
```
mesutcelik: There are spaces...

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -667,6 +668,11 @@ public void handleJoinRequest(JoinRequest joinRequest, Connection connection) {
             if (now > timeToStartJoin) {
                 startJoin();
             }
+            if (node.isMaster()) {
+                long startTime = clusterClock.getClusterStartTime();
+                nodeEngine.getOperationService().send(new ClusterStartTimeUpdateOperation(startTime),
```
mesutcelik: what happens if send operation returns `false` ?

mdogan: Why do you need another operation to set cluster start time, instead of putting this information into finalize join operation?

hazelcast/src/test/java/com/hazelcast/cluster/ClusterClockTest.java
```diff
@@ -0,0 +1,99 @@
...
+        Hazelcast.shutdownAll();
+    }
+
+    @Test
+    public void test_start_time_single_node_cluster() throws Exception {
+        Config config = new Config();
+        HazelcastInstance h1 = Hazelcast.newHazelcastInstance(config);
+        Node node1 = TestUtil.getNode(h1);
+        System.out.println(node1.isMaster());
+        assertNotEquals(Long.MIN_VALUE , node1.getClusterService().getClusterClock().getClusterStartTime());System.out.println();
```
mesutcelik: there is leftover at the end of the line `System.out.println();`

hazelcast/src/main/java/com/hazelcast/util/VersionCheck.java
```diff
@@ -116,23 +120,32 @@ private void doCheck(Node hazelcastNode, String version, boolean isEnterprise) {
                 ? Long.toString(MemoryUnit.BYTES.toGigaBytes(totalNativeMemorySize)) : "0";
         //Calculate connected clients to the cluster.
         Map<ClientType, Integer> clusterClientStats = hazelcastNode.clientEngine.getConnectedClientStats();
+        RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
+
+        Long clusterUpTime = hazelcastNode.getClusterService().getClusterClock().getClusterTime()
+                - hazelcastNode.getClusterService().getClusterClock().getClusterStartTime();
+        String clusterId = hazelcastNode.getConfig().getGroupConfig().getName()
```
mdogan: Is this a good cluster-id? I guess you want it to be unique (sort of), why don't you create a UUID for cluster?

hazelcast/src/test/java/com/hazelcast/cluster/ClusterClockTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Before
+    @After
+    public void cleanup() {
+        Hazelcast.shutdownAll();
+    }
+
+    @Test
+    public void test_start_time_single_node_cluster() throws Exception {
+        Config config = new Config();
+        HazelcastInstance h1 = Hazelcast.newHazelcastInstance(config);
```
mdogan: Do you really need to test with network?

hazelcast/src/test/java/com/hazelcast/util/VersionCheckTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @After
+    public void cleanup() {
+        Hazelcast.shutdownAll();
+    }
+
+    @Test
+    public void testVersionCheckParameters() throws Exception {
+
+        Config config = new Config();
+        HazelcastInstance h1 = Hazelcast.newHazelcastInstance(config);
```
mdogan: Do you really need to test with network?

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
```diff
@@ -46,4 +47,16 @@ public void setMasterTime(long masterTime) {
     public long getClusterTimeDiff() {
         return (clusterTimeDiff == Long.MAX_VALUE) ? 0 : clusterTimeDiff;
     }
+
+    @Override
+    public long getClusterStartTime() {
```
mdogan: Is this local time? What if clocks of nodes are not sync? What if some of the nodes are in different time-zone?

hazelcast/src/test/java/com/hazelcast/cluster/ClusterClockTest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Test
+    public void all_nodes_should_have_the_same_cluster_start_time() throws Exception {
+        Config config = new Config();
+        HazelcastInstance h1 = Hazelcast.newHazelcastInstance(config);
+        HazelcastInstance h2 = Hazelcast.newHazelcastInstance(config);
+        HazelcastInstance h3 = Hazelcast.newHazelcastInstance(config);
+        HazelcastInstance h4 = Hazelcast.newHazelcastInstance(config);
+
```
mesutcelik: Can't we use `TestHazelcastInstanceFactory.createHazelcastInstanceFactory` here? then cleanup is done automatically...

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
ahmetmircik: Also what will happen if i set system clock back or forth one hour? How will this affect calculation? Should that cluster time be increased monotonically?

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
kobalski: It is used for controlling whether the value is set before or not. If it is not MIN_VALUE, I do not update the value.

hazelcast/src/main/java/com/hazelcast/cluster/ClusterClock.java
kobalski: Fixed.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
kobalski: I have updated send with invokeOnTarget, so this operation will be tried DEFAULT_TRY_COUNT times. Even if it fails after all tries, that particular node will not set its clusterStartTime and clusterId correctly. This process will be tried again when a new node joins cluster. Also I created a new operation instead of using FinalizeJoinOperation because FinalizeJoinOperation is used more critical sections of cluster forming. I did not understand the mechanisms where it is used. So, I created another independent operation.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterClockTest.java
kobalski: Fixed.

hazelcast/src/main/java/com/hazelcast/util/VersionCheck.java
kobalski: Yes, you are right. I have added UUID based id for the cluster and getting it from ClusterService.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterClockTest.java
kobalski: For this case yes, I have tried with TestHazelcastInstanceFactory did not work, because handleJoinRequest is not called without network testing, and it is where I set clusterStartTime and clusterId.

hazelcast/src/test/java/com/hazelcast/util/VersionCheckTest.java
kobalski: No, I fixed it.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
kobalski: Right now we do not need to handle these cases. If the user sets system clock to a previous time of the clusterStartTime, upTime value will be negative. However this case is not even handled by the JVM itself. if you try RuntimeMXBean's uptime() method, while setting system time back and forth, you will see it is not monotonic too. Also if the nodes clocks are not sync, since all the nodes will send its uptime values to call home server, we can group nodes with unique clusterId, try to calculate uptime with the other data provided in call home. Basically, you are right these problems are perfectly valid, but this I think this implementation will be enough for now.

ahmetmircik: Thanks but please see this bug report and test the provided reproducer code there with the older java versions than 8 https://bugs.openjdk.java.net/browse/JDK-6523160, according to my tests and bug report page, it is already fixed in java 8. If any java version returns a negative value for uptime, it will be a bug. So, as far as i understand, there is a possibility that all nodes may return negative value for node uptimes in this PR. To me this should be fixed, but at least possible anomalies should be documented if it will not be fixed. 
Because a negative uptime value may be surprising.

kobalski: @ahmetmircik , You are totally right it is fixed with java 8. However, the fix they made requires some uptime counter which monotonically increases and it is implemented with native code.(You can see it here :  https://bugs.openjdk.java.net/browse/JDK-8027294) 

In java version 8, RuntimeMXBean uses value of the monotonic counter.(http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/sun/management/VMManagementImpl.java#183) However versions before 8, they are subtracting System.currentTimeMillis() value from vmStartTime value which causes the negative values for uptime of the jvm. (http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7-b147/sun/management/RuntimeImpl.java#118). So the perfect solution for this case, would require a cluster wide monotonic counter. Apparently, it took 18 years for JVM guys to write a uptime counter for a single JVM :), writing it for multiple jvm clusters would be challenging task.

I am using System.nanoTime() for now, because java doc says : "This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time."  It seems to be not affected by system clock changes and preserves its monotonicity only if the underlying OS does. There is bug ticket for such case (https://bugs.openjdk.java.net/browse/JDK-6458294).  At least using nano time seems to solve negative values for uptime in most OS.

I also updated java doc and stated that ClusterClock.getClusterStartTime() should be used with care and it is only used for statistical purposes.

hazelcast/src/main/java/com/hazelcast/cluster/ClusterClock.java
```diff
@@ -20,12 +20,21 @@
...
+    /**
+     * Returns the cluster start time in nanoseconds.
+     * When first node in cluster becomes master, its System#nanoTime() value is saved as
+     * clusterStartTime. Although in most JVM, System#nanoTime() is not related to
+     * any other notion of system or wall-clock time, be careful about using this value
+     * in a critical place, this value is only used for statistics purpose for calculating
+     * estimated up-time of  cluster.
+     * @return the cluster-start-time
+     */
+     long getClusterStartTime();
```
ahmetmircik: This may return millis as in your first commit. It may be confusing to return nanos since `long getClusterTime();` is returning millis but `getClusterStartTime ` is returning nanos. And also `RuntimeMXBean#getStartTime` returns result in millis.

hazelcast/src/main/java/com/hazelcast/cluster/impl/operations/ClusterStartInfoUpdateOperation.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * When a node wants to join the cluster, cluster start time is sent
+ * with joining node and saved in {@link com.hazelcast.cluster.impl.ClusterClockImpl}
+ * of {@link com.hazelcast.cluster.impl.ClusterServiceImpl} This information can
+ * be used to calculate the uptime of the cluster.
+ */
+public class ClusterStartInfoUpdateOperation extends AbstractClusterOperation implements JoinOperation {
+
+    private long startTime = Clock.currentTimeMillis();
```
ahmetmircik: What about `long startTime = System.nanoTime()/1000000;`?

Donnerbart: I have no idea what nanoseconds divided by a long number is. Please use `TimeUnit.NANOSECONDS(System.nanoTime().toWhatEverYouNeed()` instead :)

hazelcast/src/main/java/com/hazelcast/cluster/ClusterClock.java
kobalski: Fixed it, you are right about consistent behaviour.

ahmetmircik:
Can you please also wrap up your commits to one single commit, if it is possible? Other than that this PR seems mergeable to me. Thanks.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -667,6 +681,12 @@ public void handleJoinRequest(JoinRequest joinRequest, Connection connection) {
             if (now > timeToStartJoin) {
                 startJoin();
             }
+            if (node.isMaster()) {
+                long startTime = clusterClock.getClusterStartTime();
+                nodeEngine.getOperationService().invokeOnTarget(SERVICE_NAME,
```
mdogan: Shouldn't this be part of join process? Instead of sending a separate operation...

hazelcast/src/main/java/com/hazelcast/util/UuidUtil.java
```diff
@@ -55,6 +55,10 @@ private static long seedUniquifier() {
         }
     }
 
+    public static String createClusterUuid() {
+        return  buildRandomUuidString().toString();
```
mdogan: No need for `.toString()` call at the end.

kobalski: Fixed.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
kobalski: During join process, I tried to use FinalizeJoinOperation, MemberInfoUpdateOperation for setting clusterId or startTime, but these operations are sent to all members. What I want is, master node  sending its clusterId and startTime to new node making the join request. Besides, join process is so complicated, I did not want to touch its codebase.

mdogan: `FinalizeJoinOperation` is sent to only newly joining members.

kobalski: Yes, but ```FinalizeJoinOperation``` is sent by every node to newly joining member, what I want is only master node sending its clusterId and startTime to the newly joining node. I tried to implement using only ```FinalizeJoinOperation```  and removed the extra operation from the PR. During join process master node, adds an extra ```FinalizeJoinOperation```  like this in ```startJoin``` method.  

```
for (MemberInfo member : setJoins) {
    if (node.isMaster()) {
        //Only master node should send clusterId and startTime
        long startTime = clusterClock.getClusterStartTime();
        calls.add(invokeClusterOperation(new FinalizeJoinOperation(memberInfos, postJoinOp, time,
                clusterId, startTime), member.getAddress()));
    }
    calls.add(invokeClusterOperation(new FinalizeJoinOperation(memberInfos, postJoinOp, time), member.getAddress()));
}
```  
Also inside run method of  ```FinalizeJoinOperation```  newly joined sets clusterId and startTime like this,

```
if (isClusterInfoUpdateOperation()) {
    // Master node sends extra finalize join operation
    // to set clusterId and startTime.
    clusterService.setClusterId(clusterId);
    clusterService.getClusterClock().setClusterStartTime(clusterStartTime);
    return;
}
```

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -976,6 +989,12 @@ private void startJoin() {
                 final int count = members.size() - 1 + setJoins.size();
                 final List<Future> calls = new ArrayList<Future>(count);
                 for (MemberInfo member : setJoins) {
+                    if (node.isMaster()) {
+                        //Only master node should send clusterId and startTime
+                        long startTime = clusterClock.getClusterStartTime();
+                        calls.add(invokeClusterOperation(new FinalizeJoinOperation(memberInfos, postJoinOp, time,
```
mdogan: `FinalizeJoinOperation` is sent from only master node and to only newly joined nodes. I don't know how you decided that it's sent by every node to newly joining member. `startJoin()` method is also called by master node only. 

So no need to check master here, no need for `FinalizeJoinOperation.isClusterInfoUpdateOperation()` etc...

kobalski: Yes, you are right ```startJoin()``` is called only by master node, I misunderstood the logic about time in this code, 
```
if (firstJoinRequest == 0) {
	firstJoinRequest = now;
}

if (setJoins.add(memberInfo)) {
	sendMasterAnswer(joinRequest.getAddress());
	if (now - firstJoinRequest < maxWaitMillisBeforeJoin) {
    	timeToStartJoin = now + waitMillisBeforeJoin;
	}
}
if (now > timeToStartJoin) {
	startJoin();
}
``` 
I removed unnecessary parts.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterClockTest.java
mdogan: Are you sure? As far as I know/see, `handleJoinRequest` is not related to network, it's called by joiner implementation.

kobalski: The problem was our ```MockJoiner``` used for non-network tests does not call ```node.setAsMaster``` method. I changed the test with non-network version, but I needed add ```node.setAsMaster()``` call to our ```MockJoiner```' s ```doJoin()``` method inside ```TestNodeRegistry``` class. I hope it is ok, if it is not ok, I need the test with network, because I am setting the clusterId and clusterStartTime inside of ```node.setAsMaster()```.

kobalski:
@mdogan I have fixed all the comments. Is final version ok?

mdogan:
yes, ok.

PR #5259, opened 2015-05-07
Client partition refresh should only done over owner connection and s…

mdogan:
…hould not block/retry if connection fails. Otherwise a newly submitted/scheduled partition table refresh task cannot be executed and invocations fails after timeout.

see failure: https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/13477/


mdogan:
verify

sancar:
:+1:

PR #5279, opened 2015-05-11
FindBugs fix, Iterator.next should throw NoSuchElementException

ajermakovics:
checkHasNext does throw NoSuchElementException but that's not picked up by FindBugs. Need to throw explicitly.


Donnerbart:
Hmm, strange bug. Fix looks okay for me, since it does the same as the `ValidationUtil`. Did you run FindBugs locally and it stops complaining now?

jerrinot:
technically it's not the same as now there is no message passed into the `NoSuchElementException` ;-)

ajermakovics:
Indeed, the findbugs rule could be enhanced. It passes now.

ajermakovics:
The rule http://findbugs.sourceforge.net/bugDescriptions.html#IT_NO_SUCH_ELEMENT

jerrinot:
:+1:

PR #5284, opened 2015-05-12
while replicating multimap, idGen updated accordingly

### LINKED ISSUE #5220: Multimap becomes inconsistent when remove before shutdown
Hello folks,

We're experiencing a rather serious issue in Vert.x which results in inconsistent/corrupt multimap entries in the hazelcast cluster.

I have managed to reproduce this using pure Hazelcast APIs (no Vert.x involved).

Reproducer is here https://github.com/purplefox/hz-bug

The README contains a full explanation and instructions for reproducing.

There is also a thread from the Vert.x google group discussing this here:

https://groups.google.com/forum/?fromgroups#!topic/vertx/Im4UumOG8Ts

This issue is currently affecting Vert.x users in production, so if you could take a look / provide a workaround, that would be most helpful.

gurbuzali:
fixes #5220

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/multimap/MultiMapTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/multimap/MultiMapTest.java
@@ -61,6 +61,47 @@ import static org.junit.Assert.assertTrue;
 public class MultiMapTest extends HazelcastTestSupport {
 
     /**
+     * idGen is not set while replicating
+     */
+    @Test
+    public void testIssue5220() {
+        String name = randomString();
+        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(3);
+        HazelcastInstance instance1 = factory.newHazelcastInstance();
+        MultiMap<Object, Object> mm1 = instance1.getMultiMap(name);
+        // populate multimap while instance1 is owner
+        // records will have ids from 0 to 10
+        for (int i = 0; i < 10; i++) {
+            mm1.put("ping-address", "instance1-" + i);
+        }
+        HazelcastInstance instance2 = factory.newHazelcastInstance();
+        MultiMap<Object, Object> mm2 = instance2.getMultiMap(name);
+        // now the second instance is the owner
+        // if idGen is not set while replicating
+        // these entries will have ids from 0 to 10 too
+        for (int i = 0; i < 10; i++) {
+            mm2.put("ping-address", "instance2-" + i);
+        }
+        HazelcastInstance instance3 = factory.newHazelcastInstance();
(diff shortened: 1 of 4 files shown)


hazelcast/src/test/java/com/hazelcast/multimap/MultiMapTest.java
```diff
@@ -61,6 +61,47 @@
...
+     * idGen is not set while replicating
+     */
+    @Test
+    public void testIssue5220() {
+        String name = randomString();
+        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(3);
+        HazelcastInstance instance1 = factory.newHazelcastInstance();
+        MultiMap<Object, Object> mm1 = instance1.getMultiMap(name);
+        // populate multimap while instance1 is owner
+        // records will have ids from 0 to 10
```
jerrinot: records will have ids from 0 to 9?

bwzhang2011:
@purplefox, maybe you could build the 3.4.3 snapshot with such p.r to make sure whether such is resovled or not.

purplefox:
Yes I will test it :)

But if you could explain the rationale of the 1000000 offset that would help my confidence. If ids do (by some small chance) clash would there still be an issue?

hazelcast/src/test/java/com/hazelcast/multimap/MultiMapTest.java
gurbuzali: yes, I've meant 10 as exclusive :)

gurbuzali:
Hi @purplefox before this fix when the owner of the entry changed (may happen when you've started new nodes) there may be same recordIds for the records. For example first you start a node and populate it, recordId starts from 0. then another node comes in and takes the ownership which results a migration. after migration if you put more entries they will start from 0 too, and it means overlapping.

The offset is just to be on the safe side, it is not the actual solution. if you take a look at the changes at `MigrationService` you will see that before inserting migrationData I'm calculating the maxRecordId by iterating all records. I'm just adding an offset to that maxRecordId to be on the safe side. it would be enough to add just 1.

purplefox:
Can you tell me what release this will be in?

We need this fix along with the workaround I produced for https://github.com/hazelcast/hazelcast/issues/5220 in order to give a reasonably stable solution for our customers ASAP.

Any chance of an emergency patch release or similar?

gurbuzali:
We are planning to release 3.4.3 next week. I don't have an exact date since the internal testing time can vary.

purplefox:
Ok thanks, the sooner the better from our perspective. Thanks for your efforts on this :)

purplefox:
Any news on the 3.4.3 release. This is getting critical for us now.

PR #5288, opened 2015-05-12
Fixed NullPointerException when Combiner::finalizeCombine is used

### LINKED ISSUE #5283: Combiner's finalizeCombine() is not final operation on a combiner
I tried to use this method to null-out all variables as per contract (...intended to cleanup internal state...) but got null-pointers later.

Once the mapping is done, the *DefaultContext#finish()* first calls the *finalizeCombine()* on all combiners and then calls (suspicious) *requestChunk()* which in turn calls *finalizeChunk() + reset()* that most probably rely on internal state which is now gone.

Maybe I got the usage wrong, would then somebody be able to explain a proper use-case?

noctarius:
fixed #5283


gurbuzali:
verify

PR #5290, opened 2015-05-12
Partition count should be equal or greater than member count to use g…

ahmetmircik:
…enerateKeyOwnedBy method. Otherwise that method may enter endless loop


hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
```diff
@@ -386,6 +386,18 @@ protected static String generateKeyOwnedBy(HazelcastInstance instance, boolean g
         }
     }
 
+    private static void checkPartitionCount(HazelcastInstance instance) {
```
jerrinot: what about a better name? `checkPartitionCountGreaterOrEqualMemberCount()` ?

ahmetmircik: fine, updated https://github.com/hazelcast/hazelcast/pull/5291

PR #5294, opened 2015-05-12
[DONT MERGE] Feature/handling cluster exception in web filter

kobalski:
Introducing a wrapper class which delegates map operations to hazelcast instance in WebFilter. By doing this, hazelcast based exceptions are caught and handled in a more centralized way.


jerrinot:
you are eating the Exception. 

All you will see in the log will be the error message. No stacktrace, not even a message text from the original Exception. This sounds like a recipe for a nightmare troubleshooting session to me.

hazelcast-wm/src/main/java/com/hazelcast/web/SessionCleanUpService.java
```diff
@@ -0,0 +1,136 @@
...
+        this.name = name;
+        executor = Executors.newSingleThreadScheduledExecutor(new CleanupThreadFactory());
+    }
+
+
+    public void registerHazelcastInstanceDelegate(final HazelcastInstanceDelegate delegate) {
+        executor.scheduleWithFixedDelay(new Runnable() {
+            public void run() {
+                while (failQueue.isEmpty()) {
+                    KeyEntryProcessorPair pair = failQueue.poll();
```
bilalyasar: rather than `poll`, i prefer `peek`  then `remove`.. because in the case of `poll` then `failure` then `offer`  you just change order of the events actually.. and it can be problematic in some cases.

Issue #5282, opened 2015-05-11
[TEST-FAILURE] MapLoaderMultiNodeTest/testLoadsAll_whenLazyModeAndLoadAll


ahmetmircik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.7/com.hazelcast$hazelcast/548/testReport/junit/com.hazelcast.map.mapstore/MapLoaderMultiNodeTest/testLoadsAll_whenLazyModeAndLoadAll/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/535/testReport/junit/com.hazelcast.map.mapstore/MapLoaderMultiNodeTest/testLoadsAll_whenLazyModeAndLoadAll/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK7/com.hazelcast$hazelcast/535/testReport/junit/com.hazelcast.map.mapstore/MapLoaderMultiNodeTest/testLoadsAll_whenLazyModeAndLoadAll/

gurbuzali:
should be fixed via https://github.com/hazelcast/hazelcast/pull/5295

PR #5295, opened 2015-05-12
Wait for LoadMapOperation to return during IMap.loadAll

ajermakovics:
Otherwise seeing other operations sneaking in before LoadMapOperation and causing the map to load twice as in: #5282


ahmetmircik:
verify

gurbuzali:
so this will fix #5282 ?

ajermakovics:
Should do. unless there's another reason for it to fail

ajermakovics:
verify

ahmetmircik:
@gurbuzali seems mergeable.

PR #5309, opened 2015-05-15
Document behavior of ICompletableFuture methods

metanet:
This interface provides useful methods to register callbacks for executing after a future is completed but it does not contain any documentation about behaviour or order of the execution. Moreover, all the existing implementations un-intuitively executes subsequently registered callbacks with a random order.

We should start with documenting the methods. We can discuss the execution order and maybe provide an ordering-guarantee afterwards.


serkan-ozal:
:+1:

PR #5312, opened 2015-05-15
Fix for IBM JDK 6

noctarius:
Fixed IBM Java 6 unittest failures. The problems are caused by the ArrayList implementation in that very version of J9. We might want to consider updating it, it is fixed in the latest version, probably the reason I couldn't reproduce it locally.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/EntrySetIteratorFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/EntrySetIteratorFactory.java
@@ -21,6 +21,7 @@ import com.hazelcast.replicatedmap.impl.record.LazySet.IteratorFactory;
 import java.util.AbstractMap;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.NoSuchElementException;
 
 class EntrySetIteratorFactory<K, V>
         implements IteratorFactory<K, V, Map.Entry<K, V>> {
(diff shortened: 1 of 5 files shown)


pveentjer:
We should not add possible fixes. Either it is a fix or it isn't.

noctarius:
I can't reproduce, I can't fix it surely.

pveentjer:
I understand it is difficult to fix a problem that is difficult to reproduce. But who says that this fixes the problem? Perhaps it causes new problems.

noctarius:
Well at least it fixes the Iterator contract which doesn't require to call Iterator::hasNext before Iterator::next and I guess on IBM that might sometimes happen (depending on the JIT but I can't make it happen locally).

pveentjer:
I don't see any tests added. And afaik the nature of next has changed since it now throws a NoSuchElementException. So we need to have additional tests for these IteratorFactories.

noctarius:
http://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html#next() this is what the spec defines

pveentjer:
I know how the next works :) But we need to make sure we follow good practices when making code changes, and adding tests is one of them.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/EntrySetIteratorFactory.java
```diff
@@ -49,6 +50,18 @@ public boolean hasNext() {
 
             @Override
             public Map.Entry<K, V> next() {
+                if (entry == null) {
+                    while (iterator.hasNext()) {
```
metanet: - Isn't this loop skipping entries? In hasNext() method, it calls .next(), checks the entry and returns true. After that, it does not use that entry, instead of it, it makes a new .next() call. Therefore it skips an entry. Is my understanding wrong? 
- Isn't if(!testEntry(entry)) in next() method redundant? hasNext() method already validates an entry before returning.
- Shouldn't we set null to entry after next() method returns a value?

metanet:
I think this PR should have corresponding unit tests for the iterators.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/KeySetIteratorFactory.java
```diff
@@ -48,6 +49,18 @@ public boolean hasNext() {
 
             @Override
             public K next() {
+                if (entry == null) {
+                    while (iterator.hasNext()) {
```
metanet: I couldn't understand how this while loop works. Shouldn't it stop when a valid entry is found?

pom.xml
```diff
@@ -240,7 +240,7 @@
                         -Dhazelcast.test.use.network=false
                     </argLine>
                     <includes>
-                        <include>**/**.java</include>
+                        <include>com/hazelcast/replicatedmap/**.java</include>
```
eminn: This change should be reverted when it is about to merge.

noctarius: Yes it will be reverted, currently I'm just running the repmap tests in my own jenkins build and running all takes too much time :)

eminn: I see the intention, I added note for in case of it is forgotten.

noctarius: k :P

pveentjer:
@Danny-Hazelcast can you verify the modified classes have 87.5% test coverage?

gurbuzali:
verify

bilalyasar:
seems this PR didn't fix the failures. ( at least for client side)
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.6/556/

noctarius:
oops I guess I haven't done anything about client side

noctarius:
As I already mentioned above the bug is actually in the JRE implementation and fixed in newer version, therefore this is not a real bugfix but a workaround (in the test only!) for an old JRE bug.

PR #5313, opened 2015-05-15
Fix for IBM JDK 6 (backport)

noctarius:
Fixed IBM Java 6 unittest failures. The problems are caused by the ArrayList implementation in that very version of J9. We might want to consider updating it, it is fixed in the latest version, probably the reason I couldn't reproduce it locally.

edit: added the missing client fix as well


mesutcelik:
any update on that? I think we need to merge it to make CI back to green...

noctarius:
As the initial comment states: added the client part to the commit as it is not yet merged, for master there is an additional PR

bilalyasar:
this test fails last 20 days, https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.6/ 
fyi.

PR #5316, opened 2015-05-15
Not sending operations to replicas if map has no backups

ajermakovics:
Otherwise WrongTarget exceptions appear in the logs.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoader.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoader.java
@@ -63,9 +63,10 @@ public class MapKeyLoader {
     private IFunction<Object, Data> toData;
     private ExecutionService execService;
 
-    private int maxSize;
+    private int maxSizePerNode;
     private int maxBatch;
     private int mapNamePartition;
+    private boolean hasBackup;
 
     private LoadFinishedFuture loadFinished = new LoadFinishedFuture(true);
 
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoader.java
```diff
@@ -250,10 +252,12 @@ private void sendBatch(Map<Integer, List<Data>> batch, boolean replaceExistingVa
             futures.add(opService.invokeOnPartition(SERVICE_NAME, op, partitionId));
         }
 
-        LoadAllOperation op = new LoadAllOperation(mapName, Collections.<Data>emptyList(), replaceExistingValues, lastBatch);
-
         // notify SENDER_BACKUP
-        futures.add(opService.createInvocationBuilder(SERVICE_NAME, op, mapNamePartition).setReplicaIndex(1).invoke());
+        InternalPartition partition = partitionService.getPartition(mapNamePartition);
+        if (hasBackup && partition.getReplicaAddress(1) != null) {
```
ahmetmircik: Afaik `partition.getReplicaAddress(1)` may return `null` even if you have backups, how does this possible case affect your fix?

ajermakovics: The fix will solve the original problem which was unnecessary warning messages in the logs due to sending to non-existing replica.
Need to have another look how it affects failover...

ajermakovics: Found a better way. Now instead of checking replica address, checking for cluster size. In the original issue there was a single node and sending to replicas was generating an exception.

ahmetmircik: :+1:

ahmetmircik:
This PR seems mergeable to me @gurbuzali ?

PR #5322, opened 2015-05-18
Hazelcast Metrics Integration

pveentjer:


hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/LockStripe.java
```diff
@@ -0,0 +1,52 @@
...
+            stripe[k] = new Object();
+        }
+    }
+
+    Object getLock(Object source) {
+        int hash = identityHashCode(source);
+
+        // the following code first converts the hash to a positive value for all values except Integer.MIN_VALUE
+        // and then calls min deal with Integer.MIN_VALUE; which is safely converted to 0.
+        hash = min(0, abs(hash));
```
jerrinot: perhaps I'm missing something, but I assume expression is always equals to 0?

pveentjer: Good spotting. I'll fix that one.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/FieldProbe.java
```diff
@@ -71,14 +75,14 @@ private String getName(String namePrefix) {
             throw new IllegalArgumentException(format("@Probe field '%s' is of an unhandled type", field));
         }
 
-        if (AccessibleObjectProbe.isDouble(type)) {
+        if (isDouble(type)) {
             return new DoubleFieldProbe<S>(field, probe, type);
```
jerrinot: this smells a bit - a super class has knowledge about its subclasses.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/SortedProbesInstances.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.hazelcast.internal.metrics.impl;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+
+class SortedProbesInstances extends ArrayList<ProbeInstance> {
```
jerrinot: can you document a contract of this class? it's not obvious to me. why does it extend `ArrayList`?

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/FieldProbe.java
pveentjer: The factory has knowledge about the subclasses. Why do you think that is a problem?

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/SortedProbesInstances.java
pveentjer: I'll add.. It contains a sorted list of the probe instances. Needed for rendering.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/FieldProbe.java
jerrinot: it seems it's both a factory and superclass of implementations produced by the factory. 
as I said - I'm not sure if that's really a problem, but it smells, to me anyway.

pveentjer: In my book there is nothing smelly about it.

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/SortedProbesInstances.java
jerrinot: the confusing thing is this: it's a mutable structure, but the order is not maintained on a mutation.

pveentjer: It is effectively immutable. Once created it isn't changed.

jerrinot:
can you squeeze the commits?

Donnerbart:
Could you please check this build failures?
```
/com/hazelcast/nio/tcp/WriteHandler.java:28:8: Unused import - com.sun.javafx.font.Metrics.
/com/hazelcast/nio/tcp/ReadHandler.java:27:8: Unused import - com.sun.javafx.font.Metrics.
/com/hazelcast/nio/tcp/AbstractIOSelector.java:23:8: Unused import - com.sun.javafx.font.Metrics.
/com/hazelcast/nio/tcp/AbstractIOSelector.java:40:5: Variable access definition in wrong order.
/com/hazelcast/nio/tcp/AbstractSelectionHandler.java:21:8: Unused import - com.sun.javafx.font.Metrics.
```

pveentjer:
verfify

pveentjer:
verify

pveentjer:
verify

jerrinot:
verify

jerrinot:
one of my cluster members is reporting `cluster.clock.clusterTimeDiff=9223372036854775807`. I guess it's a `Long.MAX_VALUE` being reported on a master member?

pveentjer:
@jerrinot what do you mean? I'm not fully understanding what you have in mind.

jerrinot:
I have a cluster of two members. Performance monitor is reporting `clusterTimeDiff=11` for one member and  `9223372036854775807` for the other member. 

I suppose the `9223372036854775807` =Long.MAX_VALUE(?) is a bogus number reported on a master. As obviously the master has no `timeDiff`.

pveentjer:
sure. Makes sense. I just exposes the property; I didn't implement the logic.

mdogan:
@jerrinot: as far as I see, it's the initial value, but `getClusterTimeDiff() ` method seems to return the correct value `0`:

```java
private volatile long clusterTimeDiff = Long.MAX_VALUE;

@Override
public long getClusterTimeDiff() {
     return (clusterTimeDiff == Long.MAX_VALUE) ? 0 : clusterTimeDiff;
}
```

pveentjer:
It would be nicer then if I put the @Probe on the method instead of the field. Or if the discrepancy between the field and its getter gets resolved.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
```diff
@@ -24,6 +25,8 @@
 public class ClusterClockImpl implements ClusterClock {
 
     private final ILogger logger;
+
+    @Probe(name = "clusterTimeDiff")
```
mdogan: This probe, I guess, should be places on `getClusterTimeDiff()` method instead of field.

jerrinot:
@mdogan: thanks!
@pveentjer: sounds like a good idea!

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterClockImpl.java
jerrinot: +1

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
```diff
@@ -112,12 +181,15 @@ public void offer(SocketWritable packet) {
         schedule();
     }
 
+
     private SocketWritable poll() {
         for (; ; ) {
+            boolean urgent = false;
```
jerrinot: this logic is vice-versa. it should be `true` here

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
```diff
@@ -112,12 +181,15 @@ public void offer(SocketWritable packet) {
...
             SocketWritable packet = urgentWriteQueue.poll();
 
             if (packet == null) {
                 packet = writeQueue.poll();
+                urgent = true;
```
serkan-ozal: As I already mentioned at `apple-jcache` slack group, since it is taken from `writeQueue` it is not urgent. It should be the opposite.

jerrinot: and `false` here

jerrinot: a race-condition between me and @serkan-ozal :)

jerrinot:
floating point numbers should be formatted nicely. currently I can see this output: `operation.invocations.usedPercentage=5.5147058823529414E-5`

pveentjer:
I don't see anything wrong with that. That is how floating point numbers work.

jerrinot:
It's one of possible representations. As far as I know we don't use this
representation anywhere.

Anyway, if you find it a convenient format it's not a showstopper for me.

pveentjer:
The problem is that when the exponent is very negative you can run into problems. Either you need to give up on precision, or you need to print a lot of trailing zeor's which make the output very hard to understand

jerrinot:
Fair enough

jerrinot:
:+1:

pveentjer:
PS: We need to tweak the performance report so that it becomes better. So this isn't the final metrics pr. Also the optimization so that we reduce overhead per gauge, needs to be added.

pveentjer:
@gurbuzali @mdogan can you have a look at this pr?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
```diff
@@ -108,6 +112,10 @@ public ClassicOperationExecutor(GroupProperties properties,
                 + partitionOperationThreads.length + " partition operation threads.");
     }
 
+    private void registerThreadSensors(OperationThread operationThread) {
```
mdogan: an empty private method? also never called...

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
```diff
@@ -0,0 +1,301 @@
...
+     */
+    boolean isRenderingForced() {
+        if (logFile == null) {
+            return false;
+        }
+
+        return logFile.lastModified() != lastModified;
+    }
+
+    private final class BodyRendered {
```
mdogan: typo: `BodyRenderer`

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
```diff
@@ -0,0 +1,301 @@
...
+    private int maxRollingFileSizeBytes;
+
+    PerformanceLogFile(PerformanceMonitor performanceMonitor) {
+        this.bodyRendered = new BodyRendered(performanceMonitor.humanFriendlyFormat);
+        this.logger = performanceMonitor.logger;
+        this.hazelcastInstance = performanceMonitor.hazelcastInstance;
+        Node node = hazelcastInstance.node;
+        this.metricRegistry = node.nodeEngine.getMetricsRegistry();
+        this.operationService = node.nodeEngine.getOperationService();
+        this.pathname = "performance-" + hazelcastInstance.getName() + "-" + currentTimeMillis() + "-%03d.log";
```
mdogan: What about using member address (host + port) instead of `hazelcastInstance.getName()`? Because instance name probably will be the same for all members started in a separate process/machine. When you gather/pull all logs to a single directory, you cannot tell which log belongs to which member.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
pveentjer: Good spotting. That is old code before the probe annotation scanning was added I think.

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
pveentjer: Fixed.

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
pveentjer: Do you want to include the timestamp or not? I'm fine with using the hostname + port.. but the timestamp is needed if you have multiple runs in the same directory.

mdogan: timestamp needed, yes. only hostname+port instead of instance-name.

mdogan: `"performance-" + (host + port) + "-" + currentTimeMillis() + "-%03d.log"`

But you may need to escape `host`, it can include `:` for ipv6.

mdogan:
There's a recursion in `com.hazelcast.test.TestNodeRegistry$MockConnection.toString()` method. 

```java
@Override
public String toString() {
    return "MockConnection{" +
            "localEndpoint=" + localEndpoint +
            ", localConnection=" + localConnection +
            ", remoteEndpoint=" + remoteEndpoint +
            '}';
}
```

And because of that your some tests fail with stack-overflow.

```
 java.lang.StackOverflowError: null
13:02:27 	at java.lang.AbstractStringBuilder.expandCapacity(AbstractStringBuilder.java:130)
13:02:27 	at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:114)
13:02:27 	at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:415)
13:02:27 	at java.lang.StringBuilder.append(StringBuilder.java:132)
13:02:27 	at com.hazelcast.test.TestNodeRegistry$MockConnection.toString(TestNodeRegistry.java:463)
13:02:27 	at java.lang.String.valueOf(String.java:2849)
13:02:27 	at java.lang.StringBuilder.append(StringBuilder.java:128)
13:02:27 	at com.hazelcast.test.TestNodeRegistry$MockConnection.toString(TestNodeRegistry.java:463)
13:02:27 	at java.lang.String.valueOf(String.java:2849)
13:02:27 	at java.lang.StringBuilder.append(StringBuilder.java:128)
13:02:27 	at com.hazelcast.test.TestNodeRegistry$MockConnection.toString(TestNodeRegistry.java:463)
13:02:27 	at java.lang.String.valueOf(String.java:2849)
13:02:27 	at java.lang.StringBuilder.append(StringBuilder.java:128)
13:02:27 	at com.hazelcast.test.TestNodeRegistry$MockConnection.toString(TestNodeRegistry.java:463)
13:02:27 	at java.lang.String.valueOf(String.java:2849)
13:02:27 	at java.lang.StringBuilder.append(StringBuilder.java:128)
13:02:27 	at com.hazelcast.test.TestNodeRegistry$MockConnection.toString(TestNodeRegistry.java:463)
13:02:27 	at java.lang.String.valueOf(String.java:2849)
13:02:27 	at java.lang.StringBuilder.append(StringBuilder.java:128)
13:02:27 	at com.hazelcast.test.TestNodeRegistry$MockConnection.toString(TestNodeRegistry.java:463)
13:02:27 	at java.lang.String.valueOf(String.java:2849)
13:02:27 	at java.lang.StringBuilder.append(StringBuilder.java:128)
13:02:27 	at com.hazelcast.test.TestNodeRegistry$MockConnection.toString(TestNodeRegistry.java:463)
```

pveentjer:
Hmm.. strange. Perhaps it is caused by changes today/yesterday in core

mdogan:
yes. `com.hazelcast.test.TestNodeRegistry$MockConnection.toString()` method is new addition by @sancar's client test framework. And problem happens when  `MockConnection.localConnection == MockConnection.this`.

sancar:
@pveentjer @mdogan I will remove localConnection from toString method via direct commit.

pveentjer:
@sancar thanks. Can you give a ping when done? I'll do a rebase.

mdogan:
@sancar: fine for me. but strange thing is how can a connection have a connection to itself?

sancar:
I send the fix to master and maintenance. Sorry for the inconvenience.
https://github.com/hazelcast/hazelcast/commit/51678a21f014f83d682756ed6b48fa7cafc01d1c
https://github.com/hazelcast/hazelcast/commit/3a85d3a65a932c5a55243abdafc12de2b7ef8d5e

sancar:
@mdogan That I am not sure. I think I did not touch these parts, but I will have a look.

sancar:
@mdogan Only code that setting localConnection is as follows
```
 NodeEngineImpl nodeEngine = nodes.get(address);
                MockConnection thisConnection = new MockConnection(address, node.getThisAddress(), node.nodeEngine);
                conn = new MockConnection(node.getThisAddress(), address, nodeEngine);
                conn.localConnection = thisConnection;
                thisConnection.localConnection = conn;
```

These two references can not be equal. 

My theory, there was recursion with two classes. 
ConnectionA calls toString of ConnectionB
ConnectionB calls toString of ConnectionA

mdogan:
@sancar: ah right. That makes sense now.

pveentjer:
Is this caused by the metrics branch?

mdogan:
@pveentjer: no. Just `MockConnection.toString()` method was flawed and @sancar fixed it back.

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceLogFile.java
pveentjer: performance-127.0.0.1-5001-1436968195180-000.log

Donnerbart: On an IPv6 only machine the host may be something like `2001:0db8:85a3:08d3:1319:8a2e:0370:7347` which may produce an illegal filename.

pveentjer:
```
  private String getPathName() {
        Member localMember = hazelcastInstance.getCluster().getLocalMember();
        InetSocketAddress address = localMember.getSocketAddress();
        String addressString = address.getHostString().replace(":", "_")+"#"+address.getPort();
        return "performance-" + addressString + "-" + currentTimeMillis() + "-%03d.log";
    }
```

gurbuzali:
:+1:

Issue #4632, opened 2015-02-17
Unnecessary de-serialization at server-side when using Cache.get()


alparslanavci:
It seems objects are unnecessarily de-serialized on server-side when Cache.get() is called.

```java
    @org.junit.Test
    public void testCacheGet_withCustomSerialization() {
        Hazelcast.newHazelcastInstance();

        ClientConfig config = new ClientConfig();
        SerializerConfig serializerConfig = new SerializerConfig()
                .setImplementation(new FooXmlSerializer())
                .setTypeClass(Foo.class);
        config.getSerializationConfig().addSerializerConfig(serializerConfig);
        HazelcastInstance hazelcastClient = HazelcastClient.newHazelcastClient(config);

        CachingProvider provider = HazelcastClientCachingProvider.createCachingProvider(hazelcastClient);
        CacheManager cacheManager = provider.getCacheManager();

        String cacheName = "test";
        Cache<Object, Foo> cache = cacheManager.createCache(cacheName, new CacheConfig());
        String key = "foo";
        Foo valueBefore = new Foo();
        cache.put(key, valueBefore);
        Foo valueAfter = cache.get(key);
        assertEquals(valueBefore, valueAfter);
    }

    private class Foo {
        private String foo;

        public String getFoo() {
            return foo;
        }

        public void setFoo(String foo) {
            this.foo = foo;
        }

        @Override
        public boolean equals(Object o) {
            if (this == o) return true;
            if (o == null || getClass() != o.getClass()) return false;

            Foo foo1 = (Foo) o;

            if (foo != null ? !foo.equals(foo1.foo) : foo1.foo != null) return false;

            return true;
        }

        @Override
        public int hashCode() {
            return foo != null ? foo.hashCode() : 0;
        }
    }

    public class FooXmlSerializer implements StreamSerializer<Foo> {

        @Override
        public int getTypeId() {
            return 10;
        }

        @Override
        public void write(ObjectDataOutput out, Foo object) throws IOException {
            out.writeUTF(object.getFoo());
        }

        @Override
        public Foo read(ObjectDataInput in) throws IOException {
            Foo foo = new Foo();
            foo.setFoo(in.readUTF());
            return foo;
        }

        @Override
        public void destroy() {
        }
    }
```

pveentjer:
it isn't completely unwanted because in your case, the Foo instance, is potentially shared between multiple threads. That is why a defensive copy is made to prevent unexpected sharing of instances.

[...]

Issue #4775, opened 2015-03-09
ClassNotFoundException when using war classes with JCache api


smillidge:
When Hazelcast is installed as an OSGI bundle on GlassFish (i.e. on the System classpath) and you deploy a war file that uses Hazelcast. A ClassNotFoundException is thrown when trying to put an object in the cache using the JCache api when that object is an instance of a class contained in the war.

When using the Hazelcast api raw it is possible to place an object into the Hazelcast cache. It is only the JCache api that fails.

JCache fails even if you specify the ClassLoader when creating the CacheManager e.g.

```java
        ClassLoader tccl = Thread.currentThread().getContextClassLoader();
        Thread.currentThread().setContextClassLoader(HazelcastInstance.class.getClassLoader());
        HazelcastInstance instance = Hazelcast.newHazelcastInstance();
        Thread.currentThread().setContextClassLoader(tccl);
        cp = HazelcastServerCachingProvider.createCachingProvider(instance);
        CacheManager cm = cp.getCacheManager(cp.getDefaultURI(), Thread.currentThread().getContextClassLoader());
```

Exception is;
```
  StandardWrapperValve[TestServlet]: Servlet.service() for servlet TestServlet threw exception
com.hazelcast.nio.serialization.HazelcastSerializationException: java.lang.ClassNotFoundException: fish.payara.jsr107.test.TestPojo not found by com.hazelcast [304]
	at com.hazelcast.nio.serialization.DefaultSerializers$ObjectSerializer.read(DefaultSerializers.java:201)
	at com.hazelcast.nio.serialization.StreamSerializerAdapter.toObject(StreamSerializerAdapter.java:65)
	at com.hazelcast.nio.serialization.SerializationServiceImpl.toObject(SerializationServiceImpl.java:260)
	at com.hazelcast.cache.impl.CacheRecordStore.dataToValue(CacheRecordStore.java:112)
	at com.hazelcast.cache.impl.CacheRecordStore.recordToValue(CacheRecordStore.java:124)
	at com.hazelcast.cache.impl.AbstractCacheRecordStore.get(AbstractCacheRecordStore.java:766)
	at com.hazelcast.cache.impl.operation.CacheGetOperation.run(CacheGetOperation.java:50)
	at com.hazelcast.spi.impl.BasicOperationService$OperationHandler.handle(BasicOperationService.java:733)
	at com.hazelcast.spi.impl.BasicOperationService$OperationHandler.access$500(BasicOperationService.java:709)
[...]

PR #5326, opened 2015-05-18
Return value as storage format in Cache

serkan-ozal:
Fixes https://github.com/hazelcast/hazelcast/issues/4632. There is no need for converting stored data to object for `BINARY` storage format. This behaviour is unnecessary and may cause `ClassNotFoundException` if required class is not exist at server but just caller (or client) as seen in this issue https://github.com/hazelcast/hazelcast/issues/4775


metanet:
:+1:

kobalski:
:+1:

serkan-ozal:
Will send also its backport PR after this one is merged

PR #5327, opened 2015-05-18
Cleanup of AbstractCompletableFuture (non blocking again) + added tests

pveentjer:
This is a cleaned up version of the fix from this PR:

https://github.com/hazelcast/hazelcast/pull/5306

So the lock is dropped again and non blocking part is simplified by using a single cas that prevents ending up with a pending callback when a result is set concurrently with an andThen.

This PR also contains tests for the changes.


metanet:
I liked the single-cas approach. :+1:

jerrinot:
:+1:

PR #5328, opened 2015-05-20
Fixed display of system load average in HealthMonitor

Donnerbart:
Fixed display of system load average in `HealthMonitor`, which is not a percentaged value. Fixed display of negative values, which occurs on OS which don't provide this metric (e.g. Windows).

The load average value cannot be converted to a percentaged value by multiplying with 100. You have to calculate in at least the CPU core count. We should better print the original value.

Before -> After
`load.systemAverage=73.00%` -> `load.systemAverage=0.73`
`load.systemAverage=-100.00%` -> `load.systemAverage=n/a`

For the `SystemCpuLoad` and `ProcessCpuLoad` values the actual approach should be fine:
> The interface `com.sun.management.OperatingSystemMXBean`, has two new methods that easily get this information; `getSystemCpuLoad()` and `getProcessCpuLoad()`. Both methods return the load as a double value in the [0.0,1.0] interval, with 0.0 representing no load to 1.0 representing 100% CPU load for the whole system or the current JVM process respectively.

We can't use those methods since it's Java 7, but both values are percentaged.

Reported via Zendesk ticket 925.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
@@ -39,6 +39,7 @@ import static com.hazelcast.memory.MemoryStatsSupport.freePhysicalMemory;
 import static com.hazelcast.memory.MemoryStatsSupport.freeSwapSpace;
 import static com.hazelcast.memory.MemoryStatsSupport.totalPhysicalMemory;
 import static com.hazelcast.memory.MemoryStatsSupport.totalSwapSpace;
+import static com.hazelcast.util.OperatingSystemMXBeanSupport.getSystemLoadAverage;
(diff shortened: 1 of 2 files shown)


jerrinot:
This should merged only after #5322 is merged.

Also the change should be highlighted in Release Notes as it might break log scrapers.

Donnerbart:
I agree, please merge #5322 first (if there are conflicts this PR will be very easy to adapt).

About the load value itself: There are many myths about it, a lot of interpretations, stackoverflow tickets etc. My opinion is very simple: We should not modify (aka interpret) the value, just pass it through. This is the easiest thing we can do to avoid confusion. Our current approach is an interpretation which might fit for some customers, but not all.

Jaromir is right that it might break log parsers. If this get's merged I'll talk to Serdar about the release notes.

pveentjer:
I'll merge it. I'll deal with the metrics integration later.

PR #5329, opened 2015-05-20
Server should reject authentication requests from a client having an unknown owner member

sancar:
...Otherwise clients can execute illegal operations on cluster and cause infinitely hanging resources (for ex: locks).

When owner connection is lost, client invocations should retry until owner-connection is established or client shutdowns or a timeout happens.

Add  back the missing authenticated check to server for client requests

Changes made both old and new client.

Some of the fixes are ported from @mdogan fixes on a customer branch.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -237,6 +237,10 @@ abstract class ClientInvocationServiceSupport implements ClientInvocationService
         }
     }
 
+    public boolean isShutdown() {
+        return isShutdown;
+    }
+
     public void shutdown() {
         isShutdown = true;
         responseThread.interrupt();
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
@@ -21,23 +21,23 @@ import com.hazelcast.client.connection.nio.ClientConnection;
 import com.hazelcast.client.impl.ClusterAuthenticator;
 import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.client.impl.MemberImpl;
+import com.hazelcast.client.spi.ClientClusterService;
+import com.hazelcast.core.HazelcastException;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
-import com.hazelcast.security.Credentials;
 
 import java.io.IOException;
 
 public final class ClientSmartInvocationServiceImpl extends ClientInvocationServiceSupport {
 
     private final LoadBalancer loadBalancer;
-    private final Credentials credentials;
     private final ClusterAuthenticator authenticator;
 
     public ClientSmartInvocationServiceImpl(HazelcastClientInstanceImpl client, LoadBalancer loadBalancer) {
         super(client);
         this.loadBalancer = loadBalancer;
-        credentials = client.getCredentials();
-        authenticator = new ClusterAuthenticator(client, credentials);
+        authenticator = new ClusterAuthenticator(client, client.getCredentials());
+
     }
 
     public void invokeOnPartitionOwner(ClientInvocation invocation, int partitionId) throws IOException {
@@ -45,10 +45,9 @@ public final class ClientSmartInvocationServiceImpl extends ClientInvocationServ
         if (owner == null) {
             throw new IOException("Partition does not have owner. partitionId : " + partitionId);
         }
-        ClientConnection connection =
-                (ClientConnection) connectionManager.getOrConnect(owner, authenticator);
         invocation.getClientMessage().setPartitionId(partitionId);
-        send(invocation, connection);
+        Connection connection = getConnection(owner);
+        send(invocation, (ClientConnection) connection);
     }
 
     @Override
@@ -57,7 +56,7 @@ public final class ClientSmartInvocationServiceImpl extends ClientInvocationServ
         if (randomAddress == null) {
             throw new IOException("Not address found to invoke ");
         }
-        final Connection connection = connectionManager.getOrConnect(randomAddress, authenticator);
+        final Connection connection = getConnection(randomAddress);
         send(invocation, (ClientConnection) connection);
     }
 
@@ -70,10 +69,29 @@ public final class ClientSmartInvocationServiceImpl extends ClientInvocationServ
         if (!isMember(target)) {
             throw new IOException("Target :  " + target + " is not member. ");
         }
-        final Connection connection = connectionManager.getOrConnect(target, authenticator);
+        final Connection connection = getConnection(target);
         invokeOnConnection(invocation, (ClientConnection) connection);
     }
 
+    private Connection getConnection(Address target) throws IOException {
+        ensureOwnerConnectionAvailable();
+        return connectionManager.getOrConnect(target, authenticator);
+    }
+
+    private void ensureOwnerConnectionAvailable() throws IOException {
+        ClientClusterService clientClusterService = client.getClientClusterService();
+        Address ownerConnectionAddress = clientClusterService.getOwnerConnectionAddress();
+
+        boolean isOwnerConnectionAvailable = ownerConnectionAddress != null
+                && connectionManager.getConnection(ownerConnectionAddress) != null;
+
+        if (!isOwnerConnectionAvailable) {
+            if (isShutdown()) {
+                throw new HazelcastException("ConnectionManager is not active!");
+            }
+            throw new IOException("Not able to setup owner connection!");
+        }
+    }
 
     @Override
     public void invokeOnConnection(ClientInvocation invocation, ClientConnection connection) throws IOException {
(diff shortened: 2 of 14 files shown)


mdogan:
:+1:

pveentjer:
Will every client request contain a member list?

And afaik these are just fixes for symptoms, but it doesn't resolve the underlying replication issues. So I don't see a lot of value in these changes since it doesn't solve the root cause.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -89,23 +90,47 @@ public void run() {
...
-            processMessage();
-            interceptAfter(credentials);
+
         } catch (Throwable e) {
             logProcessingFailure(e);
             handleProcessingFailure(e);
         }
     }
 
+    protected boolean isAuthenticatorMessage() {
```
pveentjer: should this not be 'isAuthenticationMessage' ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -89,23 +90,47 @@ public void run() {
...
+        return false;
+    }
+
+    private void initializeAndRunMessage() {
```
pveentjer: Perhaps a more obvious name would be 'initialize' or 'initializeAndProcessMessage'. Otherwise you have run and process as description what you are doing.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -89,23 +90,47 @@ public void run() {
...
+        Credentials credentials = endpoint.getCredentials();
+        interceptBefore(credentials);
+        checkPermissions(endpoint);
+        processMessage();
+    }
+
+    private void handleAuthenticationFailure() {
+        Exception exception;
+        if (nodeEngine.isActive()) {
+            String message = "Client " + endpoint + " must authenticate before any operation.";
```
pveentjer: Would 'client ... failed to authenticate' not be sufficient?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -95,6 +100,11 @@ private ClientMessage handleAuthenticated() {
             }
         }
 
+        if (clientEngine.getClusterService().getMember(principal.getOwnerUuid()) == null) {
```
pveentjer: I try to prevent 'complex' expression in if statements because it isn't obvious what you are doing.

```
boolean isMember = clientEngine.getClusterService().getMember(principal.getOwnerUuid());
if(!isMember){
}
```

This is a lot easier to understand since it explain what you are doing.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
```diff
@@ -69,10 +68,30 @@ public void invokeOnTarget(ClientInvocation invocation, Address target)
...
+    private Connection getConnection(Address target) throws IOException {
+        ensureOwnerConnection();
+        return connectionManager.getOrConnect(target, authenticator);
+    }
+
+    private void ensureOwnerConnection() throws IOException {
+        ClientClusterService clientClusterService = client.getClientClusterService();
+        Address ownerConnectionAddress = clientClusterService.getOwnerConnectionAddress();
+
+        if (ownerConnectionAddress == null
```
pveentjer: I would introduce a local variable explaining the intent.

```
boolean isOwner = ownerConnectionAddress!=null 
     && connectionManager.getConnection(ownerConnectionAddress) !=null;
if(!isOwner){
     ....
}
```

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -237,6 +237,10 @@ public Throwable createNew(Object arg) {
         }
     }
 
+    public boolean isShutdown() {
```
pveentjer: Does this method override a method? If so, please add @Override

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
```diff
@@ -45,10 +45,9 @@ public void invokeOnPartitionOwner(ClientInvocation invocation, int partitionId)
         if (owner == null) {
             throw new IOException("Partition does not have owner. partitionId : " + partitionId);
         }
-        ClientConnection connection =
-                (ClientConnection) connectionManager.getOrConnect(owner, authenticator);
         invocation.getClientMessage().setPartitionId(partitionId);
-        send(invocation, connection);
+        Connection connection = getConnection(owner);
```
pveentjer: Wouldn't it be better if getConnection(owner) returns a client connection? Can a client have anything else than client connections?

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
```diff
@@ -70,10 +69,29 @@ public void invokeOnTarget(ClientInvocation invocation, Address target)
...
+        final Connection connection = getConnection(target);
         invokeOnConnection(invocation, (ClientConnection) connection);
     }
 
+    private Connection getConnection(Address target) throws IOException {
+        ensureOwnerConnection();
+        return connectionManager.getOrConnect(target, authenticator);
+    }
+
+    private void ensureOwnerConnection() throws IOException {
```
pveentjer: The name is a obscure. What does it mean 'ensureOwnerConnection' ?? Should that no be e.g. 'ensureOwnerConnectionAvailable' ?

hazelcast/src/main/java/com/hazelcast/client/ClientEndpoint.java
```diff
@@ -72,4 +72,5 @@
 
     ClientPrincipal getPrincipal();
 
+    boolean isAuthenticated();
```
pveentjer: Probaly Javadoc should be added

pveentjer:
I don't see a single test.

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: Not it does not override.

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
sancar: I am mostly trying to use Connection interface to be able to change the underlying implementation easily. Since it is a change in process, there are some places uses ClientConnection and some places that uses Connection.

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
sancar: Changing the name to ensureOwnerConnectionAvailable

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
sancar: Changing with appropriate name.

hazelcast/src/main/java/com/hazelcast/client/ClientEndpoint.java
sancar: I will add the javadoc, even though I think it will just state the obvious.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
sancar: Changing it.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
sancar: Changing to "initializeAndProcessMessage"

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
sancar: Honestly I dont think there is much difference. But I dont want to change the exception message, since it kind of a legacy. And used this way before.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
sancar: Changing as suggested.

sancar:
Fixes are related to heartbeat failure. It is tested by Mehmet and me in our local by dropping packages in network level. Unfortunately, we dont have a way to test this kind of failures.

sancar:
@pveentjer Answers are inlined. 
And sorry for responding late. I was busy last two weeks. 

Will every client request contain a member list? Or is this done once during authentication?
>>> Of course every request will not contain member list.

And afaik these are just fixes for symptoms, but it doesn't resolve the underlying replication issues.
So I don't see a lot of value in these changes since it doesn't solve the root cause. It will increase the complexity of the system due to work arounds.
>> It is not for symtomps. It is real a fix mostly for split brain scnerio. Client requests should be rejected >> by the nodes that are split from cluster. In split brain scenerio, Only nodes that should accept the 
>> client requests are the one with the node that is the owner connection of the cllient.

PR #5330, opened 2015-05-20
[Backport] Return value as storage format in Cache

serkan-ozal:
Backport of PR #5326


serkan-ozal:
Since its forward-port PR #5326 has been already merged, merging also this one

Issue #4335, opened 2014-12-19
JCache sync listener completion latch problems


mdogan:
This is a follow-up issue for https://github.com/hazelcast/hazelcast/pull/4333. See PR comments for more info.

Sync listener completion latch has some problems:

- It's not defined what will happen a completion event doesn't arrive; throw error or silently ignore
- There's a hardcoded 5min upper bound limit for completion latch await
- There's no way of knowing status of invocation during latch wait phase. What if operation is already timed-out? What if invocation has failed? When operation/invocation fails, latch.await() should be notified and caller should get failure result immediately.

serkan-ozal:
closed via #5332

PR #5332, opened 2015-05-20
Check status of "ICompletableFuture" while waiting for completion latch

serkan-ozal:
Fix for #4335


jerrinot:
Could you please elaborate mechanics of the fix? What was wrong and how you fixed that? 
The #4335 raises a few points, is this fix addressing all of them?

serkan-ozal:
verify

serkan-ozal:
Hi @jerrinot,

Currently if invoked cache operation fails (such as operation timeout) before completion latch, completion latch waits as redundant. 

https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java#L194

With this PR, `awaitLatch` checks status of `InvocationFuture`. If it finishes with error (`isDone` is `true` and response is a kind of error), caller doesn't wait completion latches and returns.

As we talked with @mdogan, this cover 2nd and 3rd issues are covered. However, for 1st issue, we keep the current implementation.

metanet:
:+1:

kobalski:
:+1:

Issue #5257, opened 2015-05-07
MapLoaderFailoverTest.testLoadsAll_whenInitialLoaderNodeRemovedWhileLoading


bilalyasar:
```
org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
    at java.lang.Object.wait(Native Method)
    at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.pollResponse(InvocationFuture.java:299)
    at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.waitForResponse(InvocationFuture.java:247)
    at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.get(InvocationFuture.java:224)
    at com.hazelcast.util.executor.DelegatingFuture.get(DelegatingFuture.java:72)
    at com.hazelcast.util.executor.DelegatingFuture.get(DelegatingFuture.java:57)
    at com.hazelcast.map.mapstore.MapLoaderFailoverTest.testLoadsAll_whenInitialLoaderNodeRemovedWhileLoading(MapLoaderFailoverTest.java:139)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/531/testReport/junit/com.hazelcast.map.mapstore/MapLoaderFailoverTest/testLoadsAll_whenInitialLoaderNodeRemovedWhileLoading/

bilalyasar:
it is map loader failover test, updated title.

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-sonar/com.hazelcast$hazelcast/622/testReport/junit/com.hazelcast.map.mapstore/MapLoaderFailoverTest/testLoadsAll_whenInitialLoaderNodeRemovedAfterLoading/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8/com.hazelcast$hazelcast/561/testReport/junit/com.hazelcast.map.mapstore/MapLoaderFailoverTest/testLoadsAll_whenInitialLoaderNodeRemovedAfterLoading/

bilalyasar:
[...]

PR #5336, opened 2015-05-20
Fix for test failure 5257

ajermakovics:
Fix for #5257
If map is still loading ensure that the loader partition is created

Depends on pr #5316


mesutcelik:
rebase?

ajermakovics:
Rebased

PR #5342, opened 2015-05-21
Fix race in IOBalancer

pveentjer:
Fixes an issue with the IOBalancer which can cause OperationTImeoutException.

In short: it can happen that an write handler is migrating and it is going to process a packet on the new IOSelector before it is fully migrated.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractSelectionHandler.java
@@ -47,24 +47,6 @@ public abstract class AbstractSelectionHandler implements MigratableHandler {
         this.initialOps = initialOps;
     }
 
-    /**
-     * Migrates this handler to a new IOSelector thread.
-     * The migration logic is rather simple:
-     * <p><ul>
-     *     <li>Submit a de-registration task to a current IOSelector thread</li>
-     *     <li>The de-registration task submits a registration task to the new IOSelector thread</li>
-     * </ul></p>
-     *
-     * @param newOwner target IOSelector this handler migrates to
-     */
-    public void requestMigration(final IOSelector newOwner) {
-        if (ioSelector == newOwner || !socketChannel.isOpen()) {
-            return;
-        }
-
-        ioSelector.addTaskAndWakeup(new MigrationTask(newOwner));
-    }
-
     protected SelectionKey getSelectionKey() {
         if (selectionKey == null) {
             try {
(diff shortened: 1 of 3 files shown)


pveentjer:
verify

pveentjer:
verify

pveentjer:
Bit of background:

We first made the failure happen in a short amount of time using the MonekBalancerStrategy and setting the delay to 1 second. So every second you get some kind of nonsense migration. With the master branch this normally fails in a few minutes. If all the test runs we did, the longest it took was 22 minutes. 

After the fix is applied and still with running the monkey balancer with a delay of 1 second. 3 clusters of 10 machines each have been running the test for 30 minutes for 12 hours. Also we had a 4th cluster that runs a very large test suite in parallel (again 10 machines). There was not a single failure. 

So we are confident that this fix solves the race problem in the writehandler.

jerrinot:
:+1:

mdogan:
Can you describe the exact race path? 

This fix introduces a secondary task scheduling mechanism to `WriteHandler` (`IOSelector.addTask()` and `WriteHandler.offer(TaskPacket)`) and that's a confusion to me.

pveentjer:
One of the races is the following

initial state: write handler is not scheduled

1: balancer thread: starts the migration and sets the ioselector field to the new owner. STOP
2: user thread: schedules a packet. Since unscheduled, the scheduled flag is set, and the write handler is scheduled at the new owner.
3: balancer thread: continues and schedules a migration complete task at the new owner.

So the new ioselector first process the write handled due to the scheduling of the packet. And after that completes, the completemigration task is executed.

So the write handler is being used before it is fully migrated.

The solution: we keep the scheduled flag set during migration. Therefor we have the guarantee that the write handler can't be registered at the new io selector before the migration completes.

mdogan:
> 1: balancer thread: starts the migration and sets the ioselector field to the new owner. STOP
I guess you mean, IOSelector thread, since migration task runs on selector thread.

Do we really need to set `ioselector` to the new owner at the beginning? What about making it null at the beginning and set it to the new owner after it's set? So, when user thread sees it null, it doesn't try to schedule at all. 

Another thing I see is, `ioselector` field is not defined as volatile, I'm not sure how it's visibility to user threads is guaranteed.

pveentjer:
I need to think about it. 

I think the solution your are suggestion has its own set of complexities because it complicates scheduling for a normal packet since it needs to check for another field. So I need to think about it.

The big advantage of the current approach is that with the scheduled flag, we know that an write handler will only be registered at one IOSelector at most.

IOSelector doesn't need to be volatile since it is either used by the IOSelector-thread, or used when the scheduled flag is just set; so there is a happens before relation between completion of the migration (where it is unset) and using the selector by a user thread.

mdogan:
got it, thx. 

:+1:

pveentjer:
Thanks.

PR #5346, opened 2015-05-21
Integrated addTask and wakeup so that there are no races

pveentjer:
Due to migration (a handler moving from one IOSelector to another) it can happen that addTask is called on the old Selector and wakeup called on the new.

This fix prevents this problem.

This is done by replacing:

```
ioSelector.addTask(task)
ioSelector.wakeup
```

with:
```
ioSelector.addTaskAndWakeup(task)
```

I have removed the wakeup method since it isn't used any more. 

We are working on another fix for the IOSelector that prevent adding a task to the wrong selector or adding a task to the new selector when not needed.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/AbstractClientSelectionHandler.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/AbstractClientSelectionHandler.java
@@ -80,8 +80,7 @@ public abstract class AbstractClientSelectionHandler implements SelectionHandler
     }
 
     public void register() {
-        ioSelector.addTask(this);
-        ioSelector.wakeup();
+        ioSelector.addTaskAndWakeup(this);
     }
 
 }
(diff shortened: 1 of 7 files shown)


jerrinot:
:+1:

Donnerbart:
Did I get it right that the fix is not the replacement of the methods (since `addTaskAndWakeup(task)` is still not atomic), but the use of the queue instead of adding the task to the selector itself?

jerrinot:
I see this as a simplification = good.

Donnerbart:
Of course, I see a good cleanup. I just don't see a fix (as the PR title suggests). If this is just a cleanup you get an instant +1 from me, looks good for me as well.

pveentjer:
'''Did I get it right that the fix is not the replacement of the methods (since addTaskAndWakeup(task) is still not atomic), but the use of the queue instead of adding the task to the selector itself?'''

What is not atomic here is that an add task can be done on one IOSelector, then a migration happens, and then a IOSelector.wakeup is done on the new selector.

```
ioSelector.addTask(task)
<--------- migration happened here.
ioSelector.wakeup
```

Donnerbart:
Thanks a lot for that explanation! So the problem is the ` AbstractSelectionHandler$MigrationTask.run()` which does `ioSelector = newOwner;`. I think I got it. This was the racy part I couldn't see just from the PR. Nice catch!

:+1:

gurbuzali:
:+1:

pveentjer:
There are more races related to this migration. This is just one of the fixes that needs to be applied. More will follow.

Donnerbart:
I think it's hard to impossible to provide a unit test for this? If so this one should be ready to merge (3 reviews).

Issue #5260, opened 2015-05-07
XA Transactions throwing TransactionException instead of an XAException on timeout


sertugkaya:
When using XA Transactions in Hazelcast 3.4;
Do some transactional operation, sleep until timeout (or possibly changing timeout value to a small value) and then call commit on `XAResource` and let it timeout. 
Eventually Hazelcast is throwing a `TransactionException` instead of an `XAException` with code `XAException.XA_RBTIMEOUT`.

PR #5350, opened 2015-05-22
fixed setting timeout for xa transactions and throwing correct exception upon timeout

gurbuzali:
fixes https://github.com/hazelcast/hazelcast/issues/5260


metanet:
:+1:

PR #5352, opened 2015-05-22
Send cache configs to the new node at join 

### LINKED ISSUE #4841: On Dynamic Cluster, New Member Doesn't Know Cache Configs At Join
On Dynamic Cluster, new member can take and execute a cache operation before it gets cache config via cache replication operation. In this case, there will be error about unknown cache/cache config. After a cache replication event is received, these configs will be taken but at this time it does'n execute some cache operations (such as cache put backup) due to unknown cache config. And this may cause data lost.

serkan-ozal:
Fixes #4841

Sends cache configurations to the new node at join via Post-Join operation


serkan-ozal:
verify

pveentjer:
What happens when there is a split brain, different cache configs are made under the same name and the partitioned clusters join again?

serkan-ozal:
The old cache configs are not replaced with the new ones (so continues to work with old ones) and there is no extra logic for handling split-brain case at JCache at the moment

kobalski:
:+1:

emrahkocaman:
:+1:

PR #5357, opened 2015-05-22
clear indexes before merging the clusters since it can lead to a stal…

### LINKED ISSUE #5148: Records with an index on an attribute that is also part of the key of an IMap persists after removal
We are seeing some weird behavior from Hazelcast after following these steps:

1. Start a cluster with 2 members
2. Add an object to an IMap where the key of the IMap is a PartitionAwareKey with the key is the same as an attribute on the object
3. Segment the cluster
4. Remove the object from one of the nodes
5. Re-join the cluster (object will be merged back)
6. Remove the object again

At this point, a call to retrieve the object using the PartitionAwareKey will correctly return null; however, using the Predicates API will return the previously deleted object!

As a workaround, removing the index for the attribute that is a part of the key seems to resolve the issue but for my company this attribute is what is queried most often thus we need it fixed.

I tried debugging the issue and it appears IndexService is correctly removing the entry but still the predicate queries return deleted objects.  I have a test case that reproduces the issue I'll be happy to share.

eminn:
…e results, the index information will be brought by post join operations. Fixes #5148.

Thanks @markathomas for reporting the bug and providing a great test case.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
@@ -33,7 +33,6 @@ import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.exception.RetryableHazelcastException;
 import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.FutureUtil;
-
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


markathomas:
You're welcome. Thanks for fixing the issue!

pveentjer:
I would suggest adding some tests that directly call the DefaultRecordStore with true/false.

hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
```diff
@@ -438,13 +439,23 @@ public int clear() {
     }
 
     @Override
-    public void reset() {
+    public void reset(boolean clearIndexes) {
```
pveentjer: Can you add some javadoc?

hazelcast/src/main/java/com/hazelcast/map/impl/RecordStore.java
```diff
@@ -227,7 +227,7 @@
 
     void clearPartition();
 
-    void reset();
+    void reset(boolean clearIndexes);
```
pveentjer: Javadoc.

hazelcast/src/main/java/com/hazelcast/map/impl/DefaultRecordStore.java
```diff
@@ -438,13 +439,23 @@ public int clear() {
...
         clearRecordsMap(Collections.<Data, Record>emptyMap());
         resetSizeEstimator();
         resetAccessSequenceNumber();
+        if (clearIndexes) {
+            clearIndexes();
+        }
         mapDataStore.clear();
     }
 
+    private void clearIndexes() {
```
ahmetmircik: maybe you can add this method to index service. seems much belong to there.

Btw are you removing all indexes of this map in this partition ? Shouldn't it be only this partitions indexes to be removed?

eminn:
done

PR #5359, opened 2015-05-22
Monkey IOBalancer Strategy Added

jerrinot:
It always tries to shuffle read/write handlers.
It's intended to be used for Hazelcast stress tests only.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/StaticMigrationStrategy.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/StaticMigrationStrategy.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.nio.tcp.handlermigration;
+
+import com.hazelcast.nio.tcp.MigratableHandler;
+import java.util.Set;
+
+/**
+ * Simple Migration Strategy for Handlers.
+ *
+ * It attempts to trigger a migration if a ratio between least busy and most busy selectors
+ * exceeds {@link #MIN_MAX_RATIO_MIGRATION_THRESHOLD}.
+ *
+ * Once a migration is triggered it tries to find the busiest handler registered in
+ * {@link LoadImbalance#sourceSelector} which wouldn't cause
+ * overload of the {@link LoadImbalance#destinationSelector} after a migration.
+ *
+ */
+class StaticMigrationStrategy implements MigrationStrategy {
+
+    /**
+     * You can use this property to tune whether the migration will be attempted at all. The higher the number is
+     * the more likely the migration will be attempted. Too higher number will result in unnecessary overhead, too
+     * low number will cause performance degradation due selector imbalance.
+     *
+     * Try to schedule a migration if the least busy IOSelector receives less events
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/MonkeyStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Random;
+import java.util.Set;
+
+/**
+ * IOBalancer Migration Strategy intended to be used by stress tests only.
+ * It always tries to select a random {@link MigratableHandler handler} to be migrated.
+ */
+public class MonkeyStrategy extends MigrationStrategy {
+
+    private Random random = new Random();
```
pveentjer: make final :)

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/MonkeyStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Iterator;
+import java.util.Random;
+import java.util.Set;
+
+/**
+ * IOBalancer Migration Strategy intended to be used by stress tests only.
+ * It always tries to select a random {@link MigratableHandler handler} to be migrated.
+ */
+public class MonkeyStrategy extends MigrationStrategy {
```
pveentjer: If you really want to keep it in production, I would pull out an interface.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/MonkeyStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+import com.hazelcast.nio.tcp.MigratableHandler;
+
+import java.util.Iterator;
+import java.util.Random;
+import java.util.Set;
+
+/**
+ * IOBalancer Migration Strategy intended to be used by stress tests only.
+ * It always tries to select a random {@link MigratableHandler handler} to be migrated.
+ */
```
pveentjer: Can you add a few lines of comment why this is useful. This prevent future confusion.

pveentjer:
Minor review comments. Looks good.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/MonkeyStrategy.java
jerrinot: will do. I was considering it and couldn't decide. if you prefer an interface, I have no problem with it.

pveentjer:
I didn't check the test but:
- can you make sure that you verified that the monkey balancer is selected when configured
- can you make sure you remove the system properties are unset when the test completes

jerrinot:
I could do that, but it would require reflection and relying on a field name as there is no API to check current migration strategy. it's an implementation detail of IOBalancer.

pveentjer:
It needs to be tested so we know sure that with a default configuration the monkey balancer is not active.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/StaticMigrationStrategy.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * It attempts to trigger a migration if a ratio between least busy and most busy selectors
+ * exceeds {@link #MIN_MAX_RATIO_MIGRATION_THRESHOLD}.
+ *
+ * Once a migration is triggered it tries to find the busiest handler registered in
+ * {@link LoadImbalance#sourceSelector} which wouldn't cause
+ * overload of the {@link LoadImbalance#destinationSelector} after a migration.
+ *
+ */
+class StaticMigrationStrategy implements MigrationStrategy {
```
pveentjer: I would not call this 'static' because it is everything but static. Perhaps just call it DefaultMigrationStrategy.

pveentjer:
I'll merge it since the change are just cosmetic.

hazelcast/src/main/java/com/hazelcast/nio/tcp/handlermigration/StaticMigrationStrategy.java
jerrinot: It's `static` as there are static thresholds configured. as opposite to some sort `heuristic`.

pveentjer: To me statics looks like there is a static list of migrations or something. And that isn't the case. 

I find static not informative

PR #5369, opened 2015-05-26
introduced clearIndexes() on IndexService and moved index deletion lo…

eminn:
…gic from RecordStore to IndexService


ahmetmircik:
:+1:

PR #5373, opened 2015-05-26
[hazelcast] [backport] Web filter enhancements and bug fixes

kobalski:
This PR contains 3 enhancements for our web filter implementation.
1)  Class Dependency Removal to Domain Objects
2)  Fallback to Local Mode when Hazelcast Cluster is not available
3)  Ignored list of attributes in distributed mode.


hazelcast-wm/src/main/java/com/hazelcast/web/ClusteredSessionService.java
```diff
@@ -0,0 +1,567 @@
+/*
+ * Copyright (c) 2008-2014, Hazelcast, Inc. All Rights Reserved.
```
mesutcelik: can you please update header files?

hazelcast-wm/src/main/java/com/hazelcast/web/AbstractIMapWrapper.java
```diff
@@ -0,0 +1,175 @@
...
+import com.hazelcast.core.IMap;
+import com.hazelcast.map.EntryProcessor;
+import com.hazelcast.query.Predicate;
+import com.hazelcast.util.EmptyStatement;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+
+public class AbstractIMapWrapper<K, V> {
```
bilalyasar: seems you don't use this class in the code. Did you use this class in somewhere or i am missing something?

hazelcast-wm/src/main/java/com/hazelcast/web/ClusteredSessionService.java
```diff
@@ -0,0 +1,270 @@
...
+     * @param sessionId  sessionId
+     * @param invalidate if true remove the distributed session, otherwise just
+     *                   remove the jvm reference
+     */
+    public boolean deleteSession(String sessionId, boolean invalidate) {
+        try {
+            doDeleteSession(sessionId, invalidate);
+            return true;
+        } catch (Exception e) {
+            orphanSessions.add(sessionId);
```
bilalyasar: here we loss `invalidate` variable...

hazelcast-wm/src/main/java/com/hazelcast/web/ClusteredSessionService.java
```diff
@@ -0,0 +1,270 @@
...
+                                hzConfig.addMapConfig(mapConfig);
+                            }
+                        } catch (UnsupportedOperationException ignored) {
+                            LOGGER.info("client cannot access Config.");
+                        }
+                        failedConnection = false;
+                        LOGGER.info("Successfully Connected!");
+                        String sessionId = orphanSessions.poll();
+                        while (sessionId != null) {
+                            if (!deleteSession(sessionId, false)) {
```
bilalyasar: so we assume that `invalidate` variable is false.

hazelcast-wm/src/main/java/com/hazelcast/web/HazelcastHttpSession.java
```diff
@@ -0,0 +1,368 @@
...
+
+    public void putValue(final String name, final Object value) {
+        setAttribute(name, value);
+    }
+
+    public void removeAttribute(final String name) {
+        LocalCacheEntry entry = localCache.get(name);
+        if (entry != null && entry != WebFilter.NULL_ENTRY) {
+            entry.setValue(null);
+            entry.setRemoved(true);
```
bilalyasar: here we are setting `entry.setRemoved(true)` ..
in setAttribute method i guess we have to make `entry.setRemoved(false)`
otherwise in `getAttribute` method, it can be problematic.

hazelcast-wm/src/main/java/com/hazelcast/web/entryprocessor/GetAttributeNamesEntryProcessor.java
```diff
@@ -0,0 +1,82 @@
...
+        return WebDataSerializerHook.GET_ATTRIBUTE_NAMES;
+    }
+
+    @Override
+    public Object process(Map.Entry<String, SessionState> entry) {
+        SessionState sessionState = entry.getValue();
+        if (sessionState == null) {
+            return null;
+        }
+        sessionState.addJvmId(jvmId);
```
bilalyasar: missing `entry.setValue` method

kobalski: Fixed.

hazelcast-wm/src/main/java/com/hazelcast/web/ClusteredSessionService.java
kobalski: Good catch, I fixed it `orphanSessions` now stores `SimpleEntry<String,Boolean>` instead of `String` and `deleteSession` is called with actual value of `invalidate`.

hazelcast-wm/src/main/java/com/hazelcast/web/HazelcastHttpSession.java
kobalski: `LocalCacheEntry.removed` is used only here inside `getAttribute `  https://github.com/hazelcast/hazelcast/pull/5373/files#diff-d770d63698a9e0e449d0d8ad43f24943R120 even if you set it to false it will return null. So lets leave this with true for now.

serkan-ozal:
:+1:

hazelcast-wm/src/main/java/com/hazelcast/web/HazelcastHttpSession.java
bilalyasar: if we set it to false it will return `cacheEntry.getValue();`
or i am missing sth.?

Issue #4739, opened 2015-03-05
[TEST-FAILURE] ExecutorServiceTest.testCancellationAwareTask2


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/361/com.hazelcast$hazelcast/testReport/com.hazelcast.executor/ExecutorServiceTest/testCancellationAwareTask2/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.7/501/com.hazelcast$hazelcast/testReport/com.hazelcast.executor/ExecutorServiceTest/testCancellationAwareTask2/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8/531/com.hazelcast$hazelcast/testReport/com.hazelcast.executor/ExecutorServiceTest/testCancellationAwareTask2/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/31/com.hazelcast$hazelcast/testReport/com.hazelcast.executor/ExecutorServiceTest/testCancellationAwareTask2/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/com.hazelcast$hazelcast/405/testReport/junit/com.hazelcast.executor/ExecutorServiceTest/testCancellationAwareTask2/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/com.hazelcast$hazelcast/33/testReport/junit/com.hazelcast.executor/ExecutorServiceTest/testCancellationAwareTask2/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/com.hazelcast$hazelcast/408/testReport/junit/com.hazelcast.executor/ExecutorServiceTest/testCancellationAwareTask2/

bilalyasar:
[...]

PR #5376, opened 2015-05-26
Ensure that the task-to-cancel is not started within ExecutorServiceTest

metanet:
This change narrows down scope of the test failure #4739.


serkan-ozal:
:+1:

PR #5378, opened 2015-05-27
intercepted record values sent to the backup and afterPut methods

### LINKED ISSUE #5211: Changes made by interceptor do not appear in backup
Hazelcast version : 3.2.3
I have the following setup - 2 applications, each of them running a hazelcast node, with backup count = 2. Each of them also has a hazelcast client, which is used to write some data into a distributed map. Finally, there is an interceptor attached to this map, which slightly modifies any incoming entries in its interceptPut method.

The problem is, that if I kill one of the nodes, entries restored from backup are in their state before the interceptor modified them. 
Sample test application here - http://pastebin.com/nHRnqYHY

NB1: This only happens if I write the new entries using the Hazelcast client. If I write directly using the HZ node instance, everything is OK.
NB2: Another thing, which is slightly inconvenient - in the afterPut method, I also get the unmodified value.

eminn:
Fixes #5211

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/BasePutOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/BasePutOperation.java
@@ -94,14 +94,12 @@ public abstract class BasePutOperation extends LockAwareOperation implements Bac
     public Operation getBackupOperation() {
         final Record record = recordStore.getRecord(dataKey);
         final RecordInfo replicationInfo = buildRecordInfo(record);
-        MapDataStore<Data, Object> mapDataStore = recordStore.getMapDataStore();
-        Data dataValueForBackup = dataValue;
-        // if data-store is post processing, then we need to retrieve the 'processed' value from record
-        // not the value initially provided
-        if (mapDataStore.isPostProcessingMapStore()) {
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/operation/BasePutOperation.java
```diff
@@ -94,14 +93,8 @@ public boolean shouldBackup() {
...
         final RecordInfo replicationInfo = buildRecordInfo(record);
-        MapDataStore<Data, Object> mapDataStore = recordStore.getMapDataStore();
-        Data dataValueForBackup = dataValue;
-        // if data-store is post processing, then we need to retrieve the 'processed' value from record
-        // not the value initially provided
-        if (mapDataStore.isPostProcessingMapStore()) {
-            dataValueForBackup = mapService.getMapServiceContext().toData(record.getValue());
-        }
-        return new PutBackupOperation(name, dataKey, dataValueForBackup, replicationInfo, putTransient);
+        dataValue = mapService.getMapServiceContext().toData(record.getValue());
```
ahmetmircik: Can we check if interceptor is defined and then do this serialization ?

PR #5380, opened 2015-05-27
ICondition.awaitNanos() should return remaining time not elapsed time…

### LINKED ISSUE #4603: awaitNanos() method returns the wrong value for both the ClientConditionProxy and ConditionImpl classes
Both of the ClientConditionProxy and ConditionImpl classes implement the Condition interface, however they incorrectly return the elapsed time, instead of the remaining time (which is one of the primary purposes of that method).

For example:

https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/concurrent/lock/ConditionImpl.java#L69-74

```
    public long awaitNanos(long nanosTimeout) throws InterruptedException {
        long start = System.nanoTime();
        await(nanosTimeout, TimeUnit.NANOSECONDS);
        long end = System.nanoTime();
        return (end - start);
    }
```

mdogan:
…. Fixes #4603

(cherry picked from commit 218ab9353bd68696ed6021fa1f7bf188ed0da081)


pveentjer:
:+1:

PR #5381, opened 2015-05-27
ICondition.awaitNanos() should return remaining time not elapsed time…

### LINKED ISSUE #4603: awaitNanos() method returns the wrong value for both the ClientConditionProxy and ConditionImpl classes

mdogan:
…. Fixes #4603

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/lock/ClientConditionTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/lock/ClientConditionTest.java
@@ -9,10 +9,6 @@ import com.hazelcast.spi.exception.DistributedObjectDestroyedException;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.annotation.QuickTest;
-import java.io.IOException;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Before;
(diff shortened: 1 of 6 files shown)


hazelcast-client-new/src/test/java/com/hazelcast/client/lock/ClientConditionTest.java
```diff
@@ -177,4 +179,19 @@ public void testConditionUsage() throws InterruptedException {
...
+        ICondition condition = lock.newCondition("condition");
+
+        lock.lock();
+        try {
+            long timeout = 1000L;
+            long remainingTimeout = condition.awaitNanos(timeout);
+            assertTrue("Remaining timeout should be <= 0, but it's = " + remainingTimeout,
+                    remainingTimeout <= 0);
+        } finally {
+            lock.unlock();
```
pveentjer: For my curiosity, why do we need to unlock?

pveentjer:
Minor comment. 

:+1:

hazelcast-client-new/src/test/java/com/hazelcast/client/lock/ClientConditionTest.java
mdogan: no need, just habit..

Issue #4143, opened 2014-11-21
queue.take() doesnt get interrupted on shutdown


StanAccy:
When we shutdown Hazelcast with threads listening on a queue, we assumed we would receive interruptions on the waiting threads (its the only exception mentioned in the queue.take() javadoc) - instead we dont, we get HazelcastInstanceNotActiveException thrown.

Please could you clarify in the Javadocs what should occur in this situation (HZ 3.3.2).  We're fine catching the additional exception as long as its declared in the API.

```java
		Config cfg = new Config();

		NetworkConfig networkConfig = new NetworkConfig();
		networkConfig.setPort( 10500 );
		cfg.setNetworkConfig( networkConfig );

		GroupConfig groupConfig = new GroupConfig();
		groupConfig.setName( "test" );
		groupConfig.setPassword( "test" );
		cfg.setGroupConfig( groupConfig );

		HazelcastInstance node = Hazelcast.newHazelcastInstance( cfg );

		Thread t = new Thread( () -> {
			IQueue<Object> testQ = node.getQueue( "test" );

			boolean exit = false;
			while( !exit )
			{
				try
				{
					Object obj = testQ.take();
				}
				catch( InterruptedException e )
				{
					System.err.println( "InterruptedException, exiting..." );
					e.printStackTrace();
					exit = true;
				}
				catch( Exception e )
				{
					System.err.println( "Exception Caught: " + e.getClass().getName() + " : " + e.getMessage() );
					e.printStackTrace();
					exit = true;
				}
			}
		} );
		t.start();

		Thread.sleep( 2000 );
		node.shutdown();
```

Serdaro:
@gurbuzali, do you need any help on this, documentation-wise?

ashutosh0205:
I have also found that not just for shutdown() but when I am actually interrupting thread using thread. interrupt(), I expect it to throw right exception i. e. InterruptedException() and this is the exception which is mentioned in throws clause of take() method but what I am getting is hazelcastException() which has InterruptedException() wrapped inside it.

baptistemesta:
Hi,
Just to add that there is the same issue on Thread.interrupt(). but it's working fine if you use HazelcastClient instead of Halzelcast (see commented code below)
```java
    @Test
    public void testInterrupt() throws InterruptedException {
        HazelcastInstance instance = Hazelcast.newHazelcastInstance();
        HazelcastInstance client = HazelcastClient.newHazelcastClient();
        //KO
[...]

PR #5395, opened 2015-05-28
Queue.take/poll/put/offer should throw InterruptedException directly …

mdogan:
…instead of wrapping inside HazelcastException. Partially fixes issue #4143.


mdogan:
It seems PR builder is not triggered. 

verify

PR #5396, opened 2015-05-28
Queue.take/poll/put/offer should throw InterruptedException directly …

mdogan:
…instead of wrapping inside HazelcastException. Partially fixes issue #4143.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueProxySupport.java
@@ -91,11 +91,7 @@ abstract class QueueProxySupport extends AbstractDistributedObject<QueueService>
         checkObjectNotNull(data);
 
         OfferOperation operation = new OfferOperation(name, timeout, data);
-        try {
-            return (Boolean) invokeAndGet(operation);
-        } catch (Throwable throwable) {
-            throw ExceptionUtil.rethrowAllowInterrupted(throwable);
-        }
+        return (Boolean) invokeAndGet(operation, InterruptedException.class);
     }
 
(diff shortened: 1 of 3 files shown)


jerrinot:
I understood it covers this case: https://github.com/hazelcast/hazelcast/issues/4143#issuecomment-78520852

and not the case when Hazelcast instance is shutting down. Is this correct?

mdogan:
@jerrinot: exactly.

Currently all api calls throw `HazelcastInstanceNotActiveException` during shutdown. I think it's best to add that to documentation.

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueProxySupport.java
```diff
@@ -175,12 +167,16 @@ protected void checkObjectNotNull(Object o) {
     }
 
     private <T> T invokeAndGet(QueueOperation operation) {
+        return invokeAndGet(operation, RuntimeException.class);
```
jerrinot: I can't tell I particularly like this. From the other hand I'm not sure how to solve it more elegantly.

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueProxySupport.java
mdogan: Another ways is to duplicate `invokeAndGet()` method...

jerrinot: yes and I don't think it would be any better.

mdogan:
It seems PR builder is not triggered. 

verify

Issue #4957, opened 2015-03-31
Updates from EntryProcessor doesn't take "write-coalescing" into account


noctarius:
See corresponding todo in the code: https://github.com/hazelcast/hazelcast/blob/v3.4.1/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/WriteBehindStore.java#L85

PR #5403, opened 2015-05-28
Fixed all partial updates on an entry via entry-processor should be r…

ahmetmircik:
…eflected to map-store when in memory format is OBJECT

closes https://github.com/hazelcast/hazelcast/issues/4957


mdogan:
:+1:

PR #5406, opened 2015-05-29
behaviour changed, when entry removed event is fired event.getOldValu…

### LINKED ISSUE #4644: Multimap listener bug
Hazelcast 3.4.

When removing keys from a multimap with a listener, the entryRemoved method is called but in order to get the removed value one must call to event.getValue() instead of event.getOldValue().

On the other hand, this works fine with IMaps.

eminn:
Behavior changed, when entry removed event is fired event.getOldValue() returns the value and event.getValue() returns null. 

Fixes #4644

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/multimap/AbstractMultiMapAddEntryListenerMessageTask.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/multimap/AbstractMultiMapAddEntryListenerMessageTask.java
@@ -30,7 +30,6 @@ import com.hazelcast.nio.Connection;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.security.permission.ActionConstants;
 import com.hazelcast.security.permission.MultiMapPermission;
-
 import java.security.Permission;
 
 public abstract class AbstractMultiMapAddEntryListenerMessageTask<P> extends AbstractCallableMessageTask<P> {
@@ -86,13 +85,13 @@ public abstract class AbstractMultiMapAddEntryListenerMessageTask<P> extends Abs
                 }
                 DataAwareEntryEvent dataAwareEntryEvent = (DataAwareEntryEvent) event;
                 Data key = dataAwareEntryEvent.getKeyData();
-
                 Data value = dataAwareEntryEvent.getNewValueData();
+                Data oldValue = dataAwareEntryEvent.getOldValueData();
 
                 final EntryEventType type = event.getEventType();
                 final String uuid = event.getMember().getUuid();
 
-                sendClientMessage(key, encodeEvent(key, value, type.getType(), uuid, 1));
+                sendClientMessage(key, encodeEvent(key, value, oldValue, type.getType(), uuid, 1));
             }
         }
 
@@ -102,12 +101,12 @@ public abstract class AbstractMultiMapAddEntryListenerMessageTask<P> extends Abs
                 final EntryEventType type = event.getEventType();
                 final String uuid = event.getMember().getUuid();
                 sendClientMessage(null, encodeEvent(null,
-                        null, type.getType(),
+                        null, null, type.getType(),
                         uuid, event.getNumberOfEntriesAffected()));
             }
         }
     }
 
-    protected abstract ClientMessage encodeEvent(Data key, Data value,
+    protected abstract ClientMessage encodeEvent(Data key, Data value, Data oldValue,
                                                  int type, String uuid, int numberOfEntriesAffected);
 }
--- a/hazelcast/src/main/java/com/hazelcast/multimap/impl/client/AddEntryListenerRequest.java
+++ b/hazelcast/src/main/java/com/hazelcast/multimap/impl/client/AddEntryListenerRequest.java
@@ -67,10 +67,11 @@ public class AddEntryListenerRequest extends CallableClientRequest implements Re
                     DataAwareEntryEvent dataAwareEntryEvent = (DataAwareEntryEvent) event;
                     Data key = dataAwareEntryEvent.getKeyData();
                     Data value = dataAwareEntryEvent.getNewValueData();
+                    Data oldValue = dataAwareEntryEvent.getOldValueData();
                     Data mergingValue = dataAwareEntryEvent.getMeringValueData();
                     final EntryEventType type = event.getEventType();
                     final String uuid = event.getMember().getUuid();
-                    PortableEntryEvent portableEntryEvent = new PortableEntryEvent(key, value, null, mergingValue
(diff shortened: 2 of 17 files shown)


hazelcast/src/test/java/com/hazelcast/multimap/MultiMapListenerTest.java
```diff
@@ -0,0 +1,552 @@
...
+import static org.junit.Assert.assertTrue;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category(QuickTest.class)
+public class MultiMapListenerTest extends HazelcastTestSupport {
+
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testAddLocalEntryListener_whenNull() {
+        final MultiMap mm = createHazelcastInstance().getMultiMap(randomString());
```
pveentjer: I would suggest pulling out he multimap creation into the setup method.

hazelcast/src/test/java/com/hazelcast/multimap/MultiMapListenerTest.java
```diff
@@ -0,0 +1,552 @@
...
+        final HazelcastInstance instance = createHazelcastInstance();
+        MultiMap<String, String> map = instance.getMultiMap("testMultiMapEntryListener");
+        final CountDownLatch latchAdded = new CountDownLatch(3);
+        final CountDownLatch latchRemoved = new CountDownLatch(1);
+        final CountDownLatch latchCleared = new CountDownLatch(1);
+        final Set<String> expectedValues = new CopyOnWriteArraySet<String>();
+        expectedValues.add("hello");
+        expectedValues.add("world");
+        expectedValues.add("again");
+        map.addEntryListener(new EntryAdapter<String, String>() {
```
pveentjer: This test is too big due to the anonymous inner class.

hazelcast/src/test/java/com/hazelcast/multimap/MultiMapListenerTest.java
```diff
@@ -0,0 +1,552 @@
...
+            @Override
+            public void mapEvicted(MapEvent event) {
+            }
+
+            @Override
+            public void mapCleared(MapEvent event) {
+                latchCleared.countDown();
+            }
+        }, true);
+        map.put("1", "hello");
```
pveentjer: This test does to much. When you do an action, assert, action assert then there are multiple test methods hidden in a single test.

hazelcast/src/test/java/com/hazelcast/multimap/MultiMapListenerTest.java
```diff
@@ -0,0 +1,552 @@
...
+        assertEquals(1, map.get("2").size());
+        assertEquals(3, map.size());
+        map.remove("2");
+        assertEquals(2, map.size());
+        map.clear();
+        try {
+            assertTrue(latchAdded.await(5, TimeUnit.SECONDS));
+            assertTrue(latchRemoved.await(5, TimeUnit.SECONDS));
+            assertTrue(latchCleared.await(5, TimeUnit.SECONDS));
+        } catch (InterruptedException e) {
```
pveentjer: Why do you catch this exception? Are you expecting it to be thrown? In most cases I just let my test 'throws Exception' so I get rid of the logging noise.

hazelcast/src/test/java/com/hazelcast/multimap/MultiMapListenerTest.java
```diff
@@ -0,0 +1,552 @@
...
+            e.printStackTrace();
+            assertFalse(e.getMessage(), true);
+        }
+    }
+
+    @Test
+    public void testListeners() throws Exception {
```
pveentjer: Test also too big.

hazelcast/src/test/java/com/hazelcast/multimap/MultiMapListenerTest.java
```diff
@@ -0,0 +1,552 @@
...
+        getMultiMap(instances, name).put("key2", "val2");
+        getMultiMap(instances, name).put("key3", "val3");
+        getMultiMap(instances, name).put("key7", "val7");
+
+        assertSizeEventually(1, keys);
+    }
+
+    private MultiMap getMultiMap(HazelcastInstance[] instances, String name) {
+        final Random rnd = new Random();
+        return instances[rnd.nextInt(instances.length)].getMultiMap(name);
```
pveentjer: Adding randomization to a integration test makes it hard to debug when there is an issue. In this case I don't understand why you need to randomly pick an hz instance. Why not pick the item at position 0?

eminn:
The purpose of this PR is not to refactor/rewrite the multimap tests, which I just moved to the a new class. I won't going to put an effort regarding to that.

pveentjer:
I can live with that. Can you update the title to something shorter? I'll have a closer look at the change.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/multimap/AbstractMultiMapAddEntryListenerMessageTask.java
```diff
@@ -102,12 +101,12 @@ public void onMapEvent(MapEvent event) {
...
                 sendClientMessage(null, encodeEvent(null,
-                        null, type.getType(),
+                        null, null, type.getType(),
                         uuid, event.getNumberOfEntriesAffected()));
             }
         }
     }
 
-    protected abstract ClientMessage encodeEvent(Data key, Data value,
+    protected abstract ClientMessage encodeEvent(Data key, Data value, Data oldValue,
```
pveentjer: It makes more sense to do the old value and then the value. E.g. if you look at the AtomicLong.compareAndSwap(old, new)

Should be very easy to change with Intellij signature change refactoring.

hazelcast/src/main/java/com/hazelcast/multimap/impl/client/AddEntryListenerRequest.java
```diff
@@ -91,6 +89,12 @@ public void onMapEvent(MapEvent event) {
         return registrationId;
     }
 
+    private PortableEntryEvent createPortableEntryEventFromMapEvent(MapEvent event) {
```
pveentjer: What about 'toPotableEntryEvent'.

pveentjer:
minor review comments

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/multimap/AbstractMultiMapAddEntryListenerMessageTask.java
eminn: Well, your proposed ordering is more convenient to me but the convention on event handling for map and multimap uses (value, oldValue) order, so I leave this as is since it can lead to mistakes if I just change this.

PR #5409, opened 2015-05-29
Fixes Member Update Ordering problem

### LINKED ISSUE #5393: MigrationAwareServiceTest Failures
```
java.lang.AssertionError: expected:<542> but was:<506>
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/com.hazelcast$hazelcast/72/testReport/junit/com.hazelcast.spi/MigrationAwareServiceTest/testPartitionDataSize_whenNodesStartedParallel_withSingleBackup/

eminn:
Member update ordering problem solved when receiving member list updates from master node that leads to an inconsistent member-list views across cluster members by ignoring and requesting new member-list from master when an inconsistency happens. Fixes #5393.

Thanks @mdogan for his time on providing a base fix for this pull request.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -33,6 +33,7 @@ import com.hazelcast.cluster.impl.operations.MemberInfoUpdateOperation;
 import com.hazelcast.cluster.impl.operations.MemberRemoveOperation;
 import com.hazelcast.cluster.impl.operations.PostJoinOperation;
 import com.hazelcast.cluster.impl.operations.SetMasterOperation;
+import com.hazelcast.cluster.impl.operations.TriggerMemberListPublishOperation;
 import com.hazelcast.core.Cluster;
 import com.hazelcast.core.HazelcastException;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
@@ -79,6 +80,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -502,13 +504,27 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
         }
     }
 
+
+    public void sendMemberListToMember(Address target) {
+        if (!isMaster()) {
+            return;
+        }
+        if (thisAddress.equals(target)) {
+            return;
+        }
+        final Collection<MemberImpl> members = getMemberList();
+        MemberInfoUpdateOperation op = new MemberInfoUpdateOperation(
+                createMemberInfoList(members), clusterClock.getClusterTime(), false);
+        nodeEngine.getOperationService().send(op, target);
+    }
+
     private void sendMemberListToOthers() {
         if (!isMaster()) {
             return;
         }
         final Collection<MemberImpl> members = getMemberList();
         MemberInfoUpdateOperation op = new MemberInfoUpdateOperation(
-                createMemberInfos(members), clusterClock.getClusterTime(), false);
+                createMemberInfoList(members), clusterClock.getClusterTime(), false);
         for (MemberImpl member : members) {
             if (member.equals(thisMember)) {
                 continue;
@@ -757,7 +773,7 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
                 final PostJoinOperation postJoinOp = postJoinOps != null && postJoinOps.length > 0
                         ? new PostJoinOperation(postJoinOps) : null;
 
-                Operation op = new FinalizeJoinOperation(createMemberInfos(getMemberList()), postJoinOp,
+                Operation op = new FinalizeJoinOperation(createMemberInfoList(getMemberList()), postJoinOp,
                         clusterClock.getClusterTime(), false);
                 nodeEngine.getOperationService().send(op, target);
             } else {
@@ -976,7 +992,7 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
                 // pause migrations until join, member-update and post-join operations are completed.
                 node.getPartitionService().pauseMigration();
                 final Collection<MemberImpl> members = getMemberList();
-                final Collection<MemberInfo> memberInfos = createMemberInfos(members);
+                final Collection<MemberInfo> memberInfos = createMemberInfoList(members);
                 for (MemberInfo memberJoining : setJoins) {
                     memberInfos.add(memberJoining);
                 }
@@ -1009,8 +1025,16 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
         }
     }
 
-    private static Collection<MemberInfo> createMemberInfos(Collection<MemberImpl> members) {
-        final Collection<MemberInfo> memberInfos = new LinkedList<MemberInfo>();
+    private static List<MemberInfo> createMemberInfoList(Collection<MemberImpl> members) {
+        final List<MemberInfo> memberInfos = new LinkedList<MemberInfo>();
+        for (MemberImpl member : members) {
+            memberInfos.add(new MemberInfo(member));
+        }
+        return memberInfos;
+    }
+
+    private static Set<MemberInfo> createMemberInfoSet(Collection<MemberImpl> members) {
+        final Set<MemberInfo> memberInfos = new HashSet<MemberInfo>();
         for (MemberImpl member : members) {
             memberInfos.add(new MemberInfo(member));
         }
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1060,6 +1074,44 @@ public void updateMembers(Collection<MemberInfo> members) {
...
+        // member-update process only accepts new member updates
+        if (currentMembersSize == newMembersSize) {
+            Set<MemberInfo> currentMemberInfos = createMemberInfoSet(currentMembers.values());
+            if (currentMemberInfos.containsAll(newMemberInfos)) {
+                logger.finest("Received a periodic member update, no need to process...");
+            } else {
+                logger.warning("Received an inconsistent member update "
+                        + "which contains new members and removes some of the current members! "
+                        + "Ignoring and requesting a new member update...");
+                nodeEngine.getOperationService().send(new TriggerMemberListPublishOperation(), getMasterAddress());
```
jerrinot: can't this lead to a "MemberInfoUpdateOperation-storm" when:
1. a master is first to detect a memberX is dead
2. a new member joins
?

then a master will send

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1060,6 +1074,44 @@ public void updateMembers(Collection<MemberInfo> members) {
         }
     }
 
+    private boolean shouldProcessMemberUpdate(Map<Address, MemberImpl> currentMembers,
+                                              Collection<MemberInfo> newMemberInfos) {
+
+        int currentMembersSize = currentMembers.size();
+        int newMembersSize = newMemberInfos.size();
+
+        if (currentMembersSize > newMembersSize) {
```
jerrinot: I don't understand the reasoning behind this. 
how `currentMembersSize > newMembersSize` implies the update is old? perhaps a master was just faster when detecting dead members?

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
eminn: Until the member removal event is delivered to this node.

jerrinot: it can hog network for quite a few seconds and cause additional troubles

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
eminn: Master is the only one detects the dead members, and when it is detected a dead member it fires a member removal event. And this is not the place where we handle member removals.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
eminn: Your scenario cannot happen at the same since the operations done under the cluster service lock. If the member is dead, we remove that and sent the event to other members then join starts. So I assume that event will arrive first then if the inconsistency happens we request member-list again. I cannot foresee the case that it will hog the network.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
eminn: Image cluster size is increasing consequently, which is most likely when the cluster is just forming, in a 10 node cluster if we just receive 9 node member-list we can assume that is old. Even though the cluster size decreased this is not the place we update the member-list for dead member since it just updates the member-list. We have to wait for event to properly declare the node as dead.

pveentjer:
I don't see a single test being added.

pveentjer:
Can you give a clear explanation how the ordering problem has happened? I hear from jaromir that it has to do with the operations being executed on the generic thread and because of this, operations can be executed out of order and you run in this particular problem. Is this understanding correct? If not, can you add a clear explanation of what is happening?

eminn:
@pveentjer your understanding is correct.

I like the testing idea with custom deserializer, it is definitely worth trying. Currently, when we add a delay with some randomization directly on the codebase, the issue appears more quickly.

I will update the PR with the tests, thanks for the idea.

pveentjer:
Hi Emin; normally that is also how I would get started; just add a few hacks to see if I can quickly trigger the problem. But eventually a testing harness is needed to make sure the solution really works.

Can you add a lit of explanation of what the solution is? 

Perhaps I'm missing something, but AFAIK the issue is that you are processing an old event even though a new event already has been processed. And this leads to the out of order execution. Would it not make more sense to reject any out of old event? So you keep track of the master/time-stamp and if you receive an event from the same master, but an older timestamp, then you just discard it.

Again.. I'm not that familiar with this section of the codebase.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
jerrinot: as discussed - there is still the problem when 2 members disappear and 1 new member joins.

gurbuzali:
verify

pveentjer:
Since there is no test added (also the original problem has not been reproduced) this PR can't be merged.

I think it is better to close this PR and increase the frequency of member updates. So it can be that you get out of sync temporarily due to a out of order execution, but a few seconds later you will be in sync again.

I think this is a lot less risky than having a lot of untested changes in the system. Especially at a brittle system like the cluster service.

eminn:
Pull request is updated with a test case, I had to introduce a flag to override default serialization hierarchy since Operation class implements DataSerializable and serialization service picks up DataSerializer for every operation regardless of custom serializer registration.

mesutcelik:
verify

eminn:
verify

pveentjer:
Did you run the test case without your changes and did you see the bar turn red? I'll run it locally as well to verify that the test cases shows the system is broken.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -276,26 +277,26 @@
     /**
      * Controls the frequency of a BackupAwareOperation getting its async backups converted to a sync backups. This is needed
      * to prevent an accumulation of asynchronous backups and eventually running into stability issues.
-     *
+     * <p/>
```
pveentjer: Please remove all these <p/>. They are illegal for Java 8. So please update your intellij to stop inserting it.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -375,6 +375,14 @@
     public static final String PROP_QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK
             = "hazelcast.query.max.local.partition.limit.for.precheck";
 
+    /**
+     * If enabled, the default serialization hierarchy (DataSerializable -> Portable -> Constant -> Custom )
+     * will be overrided and custom serialization will be checked first to serialize objects.
+     */
+    public static final String PROP_SERIALIZATION_CUSTOM_OVERRIDE
```
pveentjer: No need for a public property imho. This is a hack to test certain behavior; we don't want to expose it to the outside world. Therefor it should not be added to GroupProperties; so this whole class can be reverted.

pveentjer:
More untested changes. So this PR is not in a state to get merged.

hazelcast/src/test/java/com/hazelcast/cluster/JoinStressTest.java
```diff
@@ -199,4 +240,46 @@ private void initNetworkConfig(NetworkConfig networkConfig, int basePort, int po
...
+
+        }
+    }
+
+    public static class DelayedMemberInfoUpdateOperation extends MemberInfoUpdateOperation {
+
+        public DelayedMemberInfoUpdateOperation() {
+        }
+
+        @Override
```
pveentjer: Why do you copy paste all this logic? You can just add a sleep and then call super.run().

pveentjer:
I believe you are using this test to shows that there is a bug in the system:

testJoincompletesCorrectlyWhenMultipleNodesStartedParallel

When I run this test in isolation, the test doesn't fail when  remove the bugfixing-code. When I run the test together with all the test methods it sometimes fails. 

So this test is not yet good enough to reproduce the problem. If you put a @Repeat(value = 100) above the test, I was able to reproduce the bug on every run. So can you add this annotation to make sure the test becomes a reliable mechanism to reproduce the problem (and therefor to show that the fix works).

pveentjer:
The simplest thing you can do to reproduce the problem is to let the MemberInfoUpdateOperation read out a system property and based on that you do a delay. Currently you have made a lot of changes in production code just to trigger this behavior. Just adding a system property check in the MemberInfoUpdate is the lesser of 2 evils to trigger the behavior because the code changes are way smaller.

pveentjer:
Looks good.

@jerrinot can you have one more look at it? Afaik you also know a few things about the reordering issue.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1060,6 +1074,44 @@ public void updateMembers(Collection<MemberInfo> members) {
...
+            logger.severe("Received an older member update, no need to process...");
```
jerrinot: Don't you want to send the trigger message here? I think that's what we discussed.

Also I wouldn't use the severe log-level. I don't see anything severe on this - given the current design it's something we can totally expect.

hazelcast/src/main/java/com/hazelcast/cluster/impl/operations/TriggerMemberListPublishOperation.java
```diff
@@ -0,0 +1,35 @@
...
+ * Requests member list publish from master node
+ */
+public class TriggerMemberListPublishOperation extends AbstractClusterOperation {
+
+    public TriggerMemberListPublishOperation() {
+    }
+
+    @Override
+    public void run() throws Exception {
+        final ClusterServiceImpl clusterService = getService();
```
jerrinot: what if we are no longer a master at this point?

eminn: it is already handled inside the sendMemberListToMember() method, which we just do nothing if we are not master.

eminn:
verify

hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializationServiceImpl.java
```diff
@@ -115,6 +119,7 @@ public Object getPartitionKey(Object key) {
         this.managedContext = managedContext;
         this.globalPartitioningStrategy = partitionStrategy;
         this.outputBufferSize = initialOutputBufferSize;
+        this.overrideCustomSerialization = parseBoolean(System.getProperty(SERIALIZATION_CUSTOM_OVERRIDE, "false"));
```
jerrinot: just a trivial readability thing: you can use `Boolean.getBoolean()` method for this.

jerrinot:
looks good to me.

metanet:
:+1:

PR #5410, opened 2015-05-29
[backport] Fixes Member Update Ordering problem

### LINKED ISSUE #5393: MigrationAwareServiceTest Failures

eminn:
Ordering problem solved when receiving member list updates from master node that leads to an inconsistent member-list views across cluster members by ignoring and requesting new member-list from master when an inconsistency happens. Fixes #5393

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -32,6 +32,7 @@ import com.hazelcast.cluster.impl.operations.MemberInfoUpdateOperation;
 import com.hazelcast.cluster.impl.operations.MemberRemoveOperation;
 import com.hazelcast.cluster.impl.operations.PostJoinOperation;
 import com.hazelcast.cluster.impl.operations.SetMasterOperation;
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -940,8 +956,16 @@ private void startJoin() {
         }
     }
 
-    private static Collection<MemberInfo> createMemberInfos(Collection<MemberImpl> members) {
-        final Collection<MemberInfo> memberInfos = new LinkedList<MemberInfo>();
+    private static List<MemberInfo> createMemberInfoList(Collection<MemberImpl> members) {
```
pveentjer: I would rename it to 'toMemberInfoList'.

pveentjer:
I don't see any testing being added which demonstrates the original behavior was broken and the new behavior solves the problem.

PR #5418, opened 2015-05-31
Some of the passing ignored tests during protocol development are ope…

sancar:
Some of the passing ignored tests during protocol development are opened. Unused parameter templates are removed. Made adding listeners retyable to preserve old behaviour.


sancar:
verify

PR #5425, opened 2015-06-02
Testing improvements for com.hazelcast.monitor package

Donnerbart:
I had to add some `@equals()` and `@hashCode()` methods to write a successful serialization test. I added some missing `@Override` annotations and re-ordered some methods to equalize the classes.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -237,18 +237,23 @@
 
     <suppress checks="ClassFanOutComplexity" files="com.hazelcast.client.cache.impl.AbstractClientInternalCacheProxy"/>
 
-
     <!-- Monitor -->
     <suppress checks="JavadocMethod" files="com.hazelcast.monitor[\\/]"/>
     <suppress checks="JavadocType" files="com.hazelcast.monitor[\\/]"/>
-    <suppress checks="JavadocMethod" files="com.hazelcast.monitor[\\/]"/>
     <suppress checks="JavadocVariable" files="com.hazelcast.monitor[\\/]"/>
-    <suppress checks="NPathComplexity" files="com.hazelcast.monitor.impl.MemberStateImpl"/>
-    <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.impl.MemberStateImpl"/>
-    <suppress checks="MethodCount" files="com.hazelcast.monitor.impl.LocalMapStatsImpl"/>
     <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.TimedMemberState"/>
     <suppress checks="NPathComplexity" files="com.hazelcast.monitor.TimedMemberState"/>
-
+    <suppress checks="NPathComplexity" files="com.hazelcast.monitor.impl.LocalExecutorStatsImpl"/>
+    <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.impl.LocalCacheStatsImpl"/>
+    <suppress checks="NPathComplexity" files="com.hazelcast.monitor.impl.LocalCacheStatsImpl"/>
+    <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.impl.MemberStateImpl"/>
+    <suppress checks="NPathComplexity" files="com.hazelcast.monitor.impl.MemberStateImpl"/>
+    <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.impl.LocalMapStatsImpl"/>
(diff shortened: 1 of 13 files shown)


pveentjer:
:+1:

checkstyle/suppressions.xml
```diff
@@ -237,18 +237,23 @@
...
     <suppress checks="JavadocVariable" files="com.hazelcast.monitor[\\/]"/>
-    <suppress checks="NPathComplexity" files="com.hazelcast.monitor.impl.MemberStateImpl"/>
-    <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.impl.MemberStateImpl"/>
-    <suppress checks="MethodCount" files="com.hazelcast.monitor.impl.LocalMapStatsImpl"/>
     <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.TimedMemberState"/>
     <suppress checks="NPathComplexity" files="com.hazelcast.monitor.TimedMemberState"/>
-
+    <suppress checks="NPathComplexity" files="com.hazelcast.monitor.impl.LocalExecutorStatsImpl"/>
+    <suppress checks="CyclomaticComplexity" files="com.hazelcast.monitor.impl.LocalCacheStatsImpl"/>
+    <suppress checks="NPathComplexity" files="com.hazelcast.monitor.impl.LocalCacheStatsImpl"/>
```
jerrinot: are these suppressions new?

hazelcast/src/main/java/com/hazelcast/monitor/TimedMemberState.java
```diff
@@ -190,12 +190,9 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder("TimedMemberState{\n");
-        sb.append("\t");
+        StringBuilder sb = new StringBuilder("TimedMemberState{\n\t");
```
jerrinot: if you are changing the `toString()` method then I'd suggest to make it platform neutral.

checkstyle/suppressions.xml
Donnerbart: Yes, because I added a new `equals()` and `hashCode()` methods (to make the serialization assert work), which are the reasons for the already existing suppressions.

hazelcast/src/main/java/com/hazelcast/monitor/TimedMemberState.java
Donnerbart: Do we have a newline constant/variable/getter somewhere in Hazelcast? What do you prefer for this? I don't like to put a lot of `System.getProperty("line.separator")` calls into the code.

jerrinot: it's in the `StringUtil` class

jerrinot:
are you sure the `equals` / `hashcode` methods make sense semantically? I don't have a strong opinion as I don't know how it's used.

hazelcast/src/main/java/com/hazelcast/monitor/TimedMemberState.java
Donnerbart: Thanks, found it.

checkstyle/suppressions.xml
jerrinot: ok. RESOVED.

Donnerbart:
Without it the `assertEquals(memberState, deserialized);` doesn't work. Since those classes are just there to be serialized (and deserialized in the MC), I don't know what else to test than the serialization via `toJson()`/`fromJson()` and check if everything is still the same.

jerrinot:
I can see one possible issue: two instances of stats from a different caches happen to contain the same data. they are put into a set -> boom. You just lost one. For extra bad luck add a subsequent mutation (it's not immutable) 

I'm not sure how realistic this scenario is. It's up to you, I trust your sense here.

:+1:  from me.

Donnerbart:
That's indeed a good point I didn't thought of. I just started to follow `MemberStateImpl` (which had no sufficient code coverage to get my feature passing QA). And that class has a `hashCode()` and `equals()` implementation, which relies on all members to be comparable. I just followed this -maybe bad design- down to the bottom.

I don't see a realistic chance for any trouble here. The instances of the statistic classes are all stored in maps with the name of the data structure as key (name of map, multimap, queue etc.).

Donnerbart:
verify

Donnerbart:
Unrelated test failure `MapStoreTest.testGetAllKeys:490 expected:<1000> but was:<884>` (https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/13807/console)

PR #5428, opened 2015-06-02
Near-cache invalidation event batching implementation

serkan-ozal:
- Sends near-cache invalidation events as batch to client if event batching is enabled (`hazelcast.cache.invalidation.batch.enabled`). 
- Events are sent to client when the event count is reached to max size (`hazelcast.cache.invalidation.batch.size`) or batch event sender tasks flushes the invalidation event queue in its period (`hazelcast.cache.invalidation.batchfrequency.seconds`)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheService.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheService.java
@@ -16,17 +16,28 @@
 
 package com.hazelcast.cache.impl;
 
+import com.hazelcast.cache.impl.client.CacheBatchInvalidationMessage;
 import com.hazelcast.cache.impl.client.CacheInvalidationListener;
-import com.hazelcast.cache.impl.client.CacheInvalidationMessage;
+import com.hazelcast.cache.impl.client.CacheSingleInvalidationMessage;
 import com.hazelcast.cache.impl.operation.CacheReplicationOperation;
 import com.hazelcast.nio.serialization.Data;
 
 import com.hazelcast.spi.EventRegistration;
 import com.hazelcast.spi.EventService;
+import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.PartitionReplicationEvent;
 
 import java.util.Collection;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Cache Service is the main access point of JCache implementation.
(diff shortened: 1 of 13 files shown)


gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/CacheCodecTemplate.java
```diff
@@ -32,8 +32,8 @@
     @Request(id = 1, retryable = true, response = ResponseMessageConst.STRING, event = {EventMessageConst.EVENT_CACHE})
     void addEntryListener(String name);
 
-    @Request(id = 2, retryable = true, response = ResponseMessageConst.STRING,
-            event = {EventMessageConst.EVENT_CACHEINVALIDATION})
+    @Request(id = 2, retryable = false, response = ResponseMessageConst.STRING,
+            event = {EventMessageConst.EVENT_CACHEINVALIDATION, EventMessageConst.EVENT_CACHEBATCHINVALIDATION})
     void addInvalidationListener(String name);
```
sancar: Listener requests should be retryable in client .

serkan-ozal: ok

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheService.java
```diff
@@ -111,14 +152,93 @@ public String addInvalidationListener(String name, CacheEventListener listener)
...
+                sendSingleInvalidationEvent(name, key, sourceUuid);
+            }
+        }
+    }
+
+    protected void sendSingleInvalidationEvent(String name, Data key, String sourceUuid) {
+        EventService eventService = nodeEngine.getEventService();
+        Collection<EventRegistration> registrations = eventService.getRegistrations(SERVICE_NAME, name);
+        if (!registrations.isEmpty()) {
+            //TODO Fix below for client protocol
```
sancar: We can remove this comment. There nothing to fix related to protocol anymore

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheAddInvalidationListenerTask.java
```diff
@@ -46,13 +51,24 @@ public void handleEvent(Object eventObject) {
...
-                    ClientMessage eventMessage
-                            = CacheAddInvalidationListenerCodec.encodeCacheInvalidationEvent(message.getName(),
-                            message.getKey(), message.getSourceUuid());
-                    sendClientMessage(message.getKey(), eventMessage);
-
+                    if (eventObject instanceof CacheSingleInvalidationMessage) {
+                        CacheSingleInvalidationMessage message = (CacheSingleInvalidationMessage) eventObject;
+                        ClientMessage eventMessage = CacheAddInvalidationListenerCodec.
+                                    encodeCacheInvalidationEvent(message.getName(), message.getKey(), message.getSourceUuid());
+                        sendClientMessage(message.getName(), eventMessage);
```
sancar: I see that you changed the orderKey from key to distributed object name. Even Though it does not break the current guarantee (in fact, it causes more strict guarantee), I was wondering why did you decided to change that?

sancar:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheAddInvalidationListenerTask.java
serkan-ozal: - to be consistent with batch event
- `key` may be `null` sometimes such as invalidates all entries (clear near-cache) when cache clear
- ordering events as their keys doesn't make sense to me

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheService.java
serkan-ozal: right, just leftover

Issue #5035, opened 2015-04-10
[TEST-FAILURE] MapLoaderTest.testMapLoaderLoadUpdatingIndex


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK6/499/com.hazelcast$hazelcast/testReport/com.hazelcast.map.mapstore/MapLoaderTest/testMapLoaderLoadUpdatingIndex/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-OracleJDK1.6/com.hazelcast$hazelcast/247/testReport/junit/com.hazelcast.map.mapstore/MapLoaderTest/testMapLoaderLoadUpdatingIndex/

jerrinot:
The issue has the same root cause as #5393

jerrinot:
There is another possible root cause - there is a fixed 3s timeout in `map.destroy()` 
I can reproduce it in 100% of cases using 8dad05f3dd4d8e539d606809acdda15c2e874007

gurbuzali:
closed via https://github.com/hazelcast/hazelcast/pull/5433

hasancelik:
```
java.lang.AssertionError: call-count of loadAllKeys method is problematic expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.map.mapstore.MapLoaderTest$4.run(MapLoaderTest.java:214)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:828)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:842)
	at com.hazelcast.map.mapstore.MapLoaderTest.assertLoadAllKeysCount(MapLoaderTest.java:211)
	at com.hazelcast.map.mapstore.MapLoaderTest.testMapLoaderLoadUpdatingIndex(MapLoaderTest.java:124)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/698/testReport/junit/com.hazelcast.map.mapstore/MapLoaderTest/testMapLoaderLoadUpdatingIndex/

[...]

PR #5433, opened 2015-06-02
Longer timeout of DistributedObjectDestroyOperation

jerrinot:
Forward fix for #5035
See 8dad05f3dd4d8e539d606809acdda15c2e874007 for a reproducer.

In a nutshell:
The test destroys a map, creates it again an expects each member will call loadAllKeys on a MapLoader.

However it waits just 3 seconds for a response of DistributedObjectDestroyOperation. If the response is slower
then it carries on -> it means the test will work with the old map which has not been destroyed yet
-> the getAllKeys won't be called -> assertion failure.

The  3 seconds timeout seems to really short to me. I'm increasing it to 30 seconds.
Moreover I'm increasing a log level as I don't think a missing response is a business as usual.


jerrinot:
verify

jerrinot:
verify

jerrinot:
verify

PR #5441, opened 2015-06-03
Check the invocation on remote node before issuing OperationTimeOutEx…

### LINKED ISSUE #4398: OperationTimeoutException when calling get on task future after hazelcast.operation.call.timeout.millis
I am executing some tests with a build of Hazelcast 3.4.1-dev from December 22 2014. I discovered a problem with long-running tasks similar to issue #1949. It seems to be new in 3.4/3.4.1.

The problem is the following: If you execute a task that takes at least two times as long as the configured hazelcast.operation.call.timeout.millis, and you call get on a remote future sometime after two times the timeout, an OperationTimeoutException (message "No response for x ms. Aborting invocation!...") is thrown. This can easily happen if you execute a long-running task and call get on all futures, since the first get will block until the task has finished on the corresponding node.

The following unit test can be used to reproduce the problem. It fails with the beforementioned Hazelcast 3.4.1 build, but not with e.g., 3.3.4:

```java
import com.hazelcast.core.IExecutorService;
import com.hazelcast.core.Member;
import java.io.Serializable;
import java.util.Map;
import java.util.concurrent.Future;

public class TestTask2 extends TestCase {

	private static final Logger logger = Logger.getLogger(TestTask1.class);

	@Override
	protected void setUp() throws Exception {

		// configure logging
		}
	}

	public void testTask2() throws Exception {

		// create hazelcast config

		// set call timeout to 1 second to make the problem appear quicker
		config.setProperty("hazelcast.operation.call.timeout.millis", "2000");

		// create two Hazelcast instances
		HazelcastInstance hcInstance2 = Hazelcast.newHazelcastInstance(config);
		try {

			// submit the long-running task
			IExecutorService executorService = hcInstance1.getExecutorService("default");
			Map<Member, Future<Object>> futures = executorService.submitToAllMembers(new SleepingTask());

			// call get on the "remote" member
			Member remoteMember = hcInstance2.getCluster().getLocalMember();
			// wait for 5 seconds
			// when we call get on the future after sleeping, the OperationTimeoutException will be raised in Hazelcast 3.4.1-20141222 dev
			Thread.sleep(5000);
			String result = (String) futures.get(remoteMember).get();

			assertEquals("Success", result);

		} finally {
			// shutdown Hazelcast instances
			hcInstance1.getLifecycleService().shutdown();
			hcInstance2.getLifecycleService().shutdown();
		}

		logger.info("Test done.");
	}

	private static class SleepingTask implements Callable<Object>, Serializable {

		private static final long serialVersionUID = 1L;

		@Override
		public Object call() throws Exception {
			Thread.sleep(6000);
			return "Success";
		}

	}

}
```

Thanks for taking a look and cheers,
Andreas

metanet:
Check the invocation on remote node before issuing OperationTimeOutException

* Check if the invocation is still running on remote node before issuing OperationTimeoutException

This change is originally taken from mdogan's own repo: https://github.com/mdogan/hazelcast/commit/a5db746d227b5ea0206e4f7f35025c0bd6c66cca
He has already fixed the issue. Since he is away this week and I can't contact him, I just applied his changes. 

Fixes #4398

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.spi.impl.operationservice.impl;
 
+import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Address;
 import com.hazelcast.spi.ExecutionTracingService;
@@ -29,6 +30,8 @@ import com.hazelcast.spi.impl.operationservice.impl.operations.TraceableIsStillE
 
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+
+import static com.hazelcast.spi.ExecutionService.SYSTEM_EXECUTOR;
 
 /**
  * Checks if an Operation is still running. This is useful when an operation isn't executed (e.g. got lost due to an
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
```diff
@@ -73,15 +76,56 @@ public boolean isOperationExecuting(Invocation invocation) {
...
         } catch (Exception e) {
-            invocation.logger.warning("While asking 'is-executing': " + toString(), e);
+            invocation.logger.warning("While asking 'is-executing': " + invocation, e);
         }
-        invocation.logger.warning("'is-executing': " + executing + " -> " + toString());
+        invocation.logger.warning("'is-executing': " + executing + " -> " + invocation);
         return executing;
     }
 
+    public void timeoutInvocationIfNotExecuting(final Invocation invocation) {
```
pveentjer: Can you add some public javadoc?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
```diff
@@ -73,15 +76,56 @@ public boolean isOperationExecuting(Invocation invocation) {
...
+        try {
+            final Operation isStillExecuting = createCheckOperation(invocation);
+            final ExecutionCallback callback = new ExecutionCallback() {
```
pveentjer: Can you pull out this anonymous inner class. It makes the method very hard to understand.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
```diff
@@ -73,15 +76,56 @@ public boolean isOperationExecuting(Invocation invocation) {
...
+                    setOperationTimeout();
+                }
+
+                private void setOperationTimeout() {
+                    InvocationFuture future = invocation.invocationFuture;
+                    future.set(invocation.newOperationTimeoutException(future.getMaxCallTimeout()));
+                }
+            };
+
+            nodeEngine.getExecutionService().execute(SYSTEM_EXECUTOR, new Runnable() {
```
pveentjer: Also pull out this inner class.

gurbuzali:
do we need back-port here ?

PR #5442, opened 2015-06-03
Fixed a bug in outbound ports parser

jerrinot:
- Fixed a parser bug - when multiple outbound ports are separated by a comma followed by a space. 
- NodeIOService accepts NodeEngineImpl in a constructor instead of touching a field in the Node class.
- Tests for NodeIOServiceTest::getOutboundPorts added

This should be ported to a maintenance branch too.


Donnerbart:
:+1: (and thanks for the elaboration in our team channel)

jerrinot:
verify

Issue #5437, opened 2015-06-03
[TEST-FAILURE]  ClientNearCacheTest.putToCacheAndUpdateFromOtherNodeThenGetUpdatedFromClientNearCacheWithBinaryInMemoryFormat


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/3268/com.hazelcast$hazelcast-client-new/testReport/com.hazelcast.client.cache.nearcache/ClientNearCacheTest/putToCacheAndUpdateFromOtherNodeThenGetUpdatedFromClientNearCacheWithBinaryInMemoryFormat/

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/451/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/3269/testReport/

serkan-ozal:
closed via https://github.com/hazelcast/hazelcast/pull/5443

PR #5443, opened 2015-06-03
Be sure that value data is deserialized before returning to caller in "DelegatingExecutionCallback" [CLIENT-NEW]

serkan-ozal:
- Fixes issue #5437
- Deserializes value data before retuning to caller if there is no "shouldDeserializeData" flagged "DelegatingExecutionCallback" registered


sancar:
:+1:

metanet:
:+1:

Issue #4331, opened 2014-12-19
NullPointerException during map.put() and map.destroy() are called concurrently


mdogan:
When map.destroy() is called, it clears map data out of partition threads and it makes partition data inconsistent for operations running concurrently in partition threads.

```
java.lang.NullPointerException
        at com.hazelcast.map.impl.record.Records.buildRecordInfo(Records.java:22)
        at com.hazelcast.map.impl.operation.BasePutOperation.getBackupOperation(BasePutOperation.java:89)
        at com.hazelcast.spi.impl.BasicOperationService$OperationBackupHandler.initBackupOperation(BasicOperationService.java:973)
        at com.hazelcast.spi.impl.BasicOperationService$OperationBackupHandler.newBackup(BasicOperationService.java:961)
        at com.hazelcast.spi.impl.BasicOperationService$OperationBackupHandler.makeBackups(BasicOperationService.java:948)
        at com.hazelcast.spi.impl.BasicOperationService$OperationBackupHandler.backup(BasicOperationService.java:915)
        at com.hazelcast.spi.impl.BasicOperationService$OperationHandler.handleResponse(BasicOperationService.java:769)
        at com.hazelcast.spi.impl.BasicOperationService$OperationHandler.handle(BasicOperationService.java:734)
        at com.hazelcast.spi.impl.BasicOperationService$OperationHandler.access$500(BasicOperationService.java:709)
        at com.hazelcast.spi.impl.BasicOperationService.runOperationOnCallingThread(BasicOperationService.java:253)
        at com.hazelcast.spi.impl.BasicInvocation.doInvokeLocal(BasicInvocation.java:280)
        at com.hazelcast.spi.impl.BasicInvocation.doInvoke(BasicInvocation.java:262)
        at com.hazelcast.spi.impl.BasicInvocation.invoke(BasicInvocation.java:229)
        at com.hazelcast.spi.impl.BasicInvocationBuilder.invoke(BasicInvocationBuilder.java:45)
        at com.hazelcast.client.impl.client.PartitionClientRequest.process(PartitionClientRequest.java:60)
        at com.hazelcast.client.impl.ClientEngineImpl$ClientPacketProcessor.processRequest(ClientEngineImpl.java:434)
        at com.hazelcast.client.impl.ClientEngineImpl$ClientPacketProcessor.run(ClientEngineImpl.java:353)
[...]

PR #5449, opened 2015-06-04
Null checking for data has been added into BasePutOperation

toktarev:
Null checking for data has been added into all critical places where we work with RecordStore


serkan-ozal:
also, in `run` methods at implementations (sub-types) of `BasePutOperation`, `null` `dataValue` should be handled and ignored.

toktarev:
if (dataValue!=null) {
            final Record record = recordStore.putBackup(dataKey, dataValue, ttl);
            if (recordInfo != null) {
                Records.applyRecordInfo(record, recordInfo);
            }
        }

like this ?

toktarev:
What if reall null was in the map ?

toktarev:
ttl = recordInfo != null ? recordInfo.getTtl() : ttl;       
        if (recordInfo != null) {
            final Record record = recordStore.putBackup(dataKey, dataValue, ttl);
            Records.applyRecordInfo(record, recordInfo);
        }
I suppose it would be correct 
isn't it ?

serkan-ozal:
@toktarev AFAIK, `null` values (and also keys) are not supported at put.

And also at `afterRun` method in `PutBackupOperation`, there is no need eviction if value is `null`.

``` java
@Override
public void afterRun() throws Exception {
    if (dataValue != null) {
        evict(true);
    }
}
```

serkan-ozal:
`Null` value is not allowed to put https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java#L93

toktarev:
I'd better check nullability of recordInfo

toktarev:
This is a marker of if it is fake PutBackupOperation

toktarev:
As I see 
recordInfo!=null if and only if datavalue!=null
so it is equal what to check

recordInfo I have to check anyway - we don't need 2 checks here

serkan-ozal:
yes, as I see also, `recordInfo` is always not `null` if `dataValue` is not `null`. 

But I still don't know why the old code checks this case 
``` java 
if (recordInfo != null) {
            Records.applyRecordInfo(record, recordInfo);
}
```

Because at the current implementation, as I tracked, `recordInfo` is always not `null` (I couldn't find a caller that may call this  `recordInfo` value with `null`) and built like this

``` java
final RecordInfo recordInfo = buildRecordInfo(record);

...

public static RecordInfo buildRecordInfo(Record record) {
        final RecordInfo info = new RecordInfo();
        info.setStatistics(record.getStatistics());
        info.setVersion(record.getVersion());
        info.setEvictionCriteriaNumber(record.getEvictionCriteriaNumber());
        info.setLastAccessTime(record.getLastAccessTime());
        info.setLastUpdateTime(record.getLastUpdateTime());
        info.setTtl(record.getTtl());
        return info;
}
```

Anyway, this workaround seems ok to me but the better one should be implemented by destroying the map itself as thread safe with partition aware operations in partition threads.

toktarev:
I am agree.

Will we implement "true" implementation during this 3.5 bug fixing process ?

Or simple workaround will be enough ?

serkan-ozal:
In my opinion, it is enough but not %100 sure about that are we missing some other cases for null `recordInfo` or `dataValue`. 

So, I think also @gurbuzali should have a look this PR. @gurbuzali WDYT?

toktarev:
For recordInfo I fixed all cases because checked all invocations of getRecord

gurbuzali:
@ahmetmircik can you please review ?

ahmetmircik:
__About the issue #4331__:
Afais the main issue is `map#destroy` may run concurrently with a partition operation and as a result of that a partition operation may see some null data. So to me a right fix should handle that concurrent access case. 

__About this PR__:
I'm in favor of to see this work as a refactoring PR. But normally, we should only send processed keys/entries on backups, that means if something is processed on owner, it should be send to backup nodes so the null-ness of a record should not be an issue when preparing backup operations. In this PR all operations can be refactored against this expected behavior. 

For now we can change the milestone of the issue #4331 and this PR to 3.5.1. @gurbuzali ?

PR #5452, opened 2015-06-05
Debug logging for slow invocations to deal with OperationTimeout

pveentjer:
Some logging to detect slow invocations for query op timeout (and anything else that takes time). It does this by lifting on top of the existing periodic scan of all pending invocations and adding some logging if an invocation takes more than a configurable amount of time.

By default it is disabled, but can be enabled by setting a group property. I added it as public property because it is useful for all kinds of problem detection.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
@@ -31,6 +31,7 @@ import com.hazelcast.spi.impl.operationservice.impl.responses.CallTimeoutRespons
 import com.hazelcast.spi.impl.operationservice.impl.responses.ErrorResponse;
 import com.hazelcast.spi.impl.operationservice.impl.responses.NormalResponse;
 import com.hazelcast.spi.impl.operationservice.impl.responses.Response;
+import com.hazelcast.util.Clock;
 import com.hazelcast.util.EmptyStatement;
 
 import java.util.concurrent.ConcurrentHashMap;
@@ -69,6 +70,7 @@ public class InvocationRegistry {
     private final ILogger logger;
     private final InspectionThread inspectionThread;
     private final CallIdSequence callIdSequence;
+    private final long slowInvocationThresholdMs;
 
     public InvocationRegistry(OperationServiceImpl operationService, int concurrencyLevel) {
         this.operationService = operationService;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -76,13 +78,23 @@ public InvocationRegistry(OperationServiceImpl operationService, int concurrency
...
         this.backupTimeoutMillis = props.OPERATION_BACKUP_TIMEOUT_MILLIS.getLong();
         this.invocations = new ConcurrentHashMap<Long, Invocation>(INITIAL_CAPACITY, LOAD_FACTOR, concurrencyLevel);
         this.inspectionThread = new InspectionThread();
         inspectionThread.start();
     }
 
+    private long initSlowInvocationThresholdMs(GroupProperties props) {
+        long thresholdMs = props.SLOW_INVOCATION_DETECTOR_THRESHOLD_MILLIS.getLong();
+        if (thresholdMs > -1) {
+            logger.warning("Slow invocation detector enabled, using threshold: " + thresholdMs + " ms");
```
jerrinot: warning? why not `info` ?

pveentjer: sure.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -319,28 +333,50 @@ private void scanHandleOperationTimeout() {
...
+            } catch (Throwable t) {
+                inspectOutputMemoryError(t);
+                logger.severe("Failed to handle operation timeout of invocation:" + invocation, t);
+                return false;
+            }
+        }
+
+        private void logSlowInvocation(long now, Invocation invocation) {
+            if (slowInvocationThresholdMs > 0) {
+                long durationMs = now - invocation.op.getInvocationTime();
```
jerrinot: does it relies on clock being in sync between members?

pveentjer: No..  it is always the local clock being used here. So you are scanning on the invoking side. There is no relation with a clock on the remote side that is relevant for this detection.

jerrinot: right

pveentjer: However.. There is one thing that might need some reconsideration. And that the invocation time is set using cluster lock, but timeout checks are done using Clock... and Cluster clock can skew.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -319,28 +333,50 @@ private void scanHandleOperationTimeout() {
...
+                return false;
+            }
+        }
+
+        private void detectSlowInvocation(long now, Invocation invocation) {
+            if (slowInvocationThresholdMs > 0) {
+                long durationMs = now - invocation.op.getInvocationTime();
+                if (durationMs > slowInvocationThresholdMs) {
+                    logger.info("Slow invocation: duration=" + durationMs + " ms, operation="
+                            + invocation.op.getClass().getName() + " inv:" + invocation + "\n");
```
jerrinot: why the \n char?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
jerrinot: that's a good finding!

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
pveentjer: Remainder from System.out... was consequence of my update fight.. Will removeit.. good spotting

Donnerbart:
I just dislike the order of method definition being reverse to their usage (which often struggles me when reading code, especially on GitHub without code navigation). But that's just cosmetic.

:+1:

pveentjer:
Good spotting. Will fix that. I didn't see it, I also like it that way,

jerrinot:
:+1:

PR #5454, opened 2015-06-05
Invocation timeout methods should use ClusterClock

pveentjer:
An operation.invocationTime is set using the ClusterClock. But the timeout check methods make use of a regular Clock.

Afaik this can lead to problems when the clocks are not in sync. And can lead to detecting OperationTimeouts to early or too late. I don't know if this is actual a fix or the whole ClusterClock is just a broken approach, but at least the set of invcationTime and the usage of it, make use of the same underlying ClusterClock.

Another related issue is that when the Cluster clock goes back in time because a new master is elected with a clock very far behind, then timeout won't work because the timestamps used for the invocation are based on the old master; not the new. So if the new master is an hour behind, all invocation won't be timing out for an hour.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -466,7 +466,7 @@ abstract class Invocation implements ResponseHandler, Runnable {
         boolean hasWaitingThreads = invocationFuture.getWaitingThreadsCount() > 0;
         boolean notExpired = maxCallTimeout == Long.MAX_VALUE
                 || expirationTime < 0
-                || expirationTime >= Clock.currentTimeMillis();
+                || expirationTime >= nodeEngine.getClusterService().getClusterClock().getClusterTime();
 
         if (hasResponse || hasWaitingThreads || notExpired) {
             return false;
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -533,7 +533,7 @@ boolean checkBackupTimeout(long timeoutMillis) {
 
         // If this has not yet expired (so has not been in the system for a too long period) we ignore it.
         long expirationTime = responseReceivedMillis + timeoutMillis;
-        boolean timeout = expirationTime > 0 && expirationTime < Clock.currentTimeMillis();
+        boolean timeout = expirationTime > 0 && expirationTime < nodeEngine.getClusterService().getClusterClock().getClusterTime();
```
jerrinot: that's a terrible train-wreck!

pveentjer: Ah yes.. a direct consequence of not having DI :)

jerrinot:
verify

gurbuzali:
verify

mtopolnik:
verify

pveentjer:
Probably the clusterclock will be removed in Hazelcast 3.7. 

https://hazelcast.atlassian.net/wiki/display/EN/Operation+timeout+improvements

PR #5459, opened 2015-06-05
Fix update/invalidate race in NearCache

mtopolnik:
Fix for #4671. Also contributes Agrona collections backport, which is used in the fix.

Statement of issue:

Currently `NearCache` has a race condition between the methods `get`, `put` and `invalidate`. In outline, this interleaving is possible (all operations on the same key):

1. *threadA*: `get` from near cache -> cache miss;
2. *threadA*: fetch from source of truth;
3. *threadB*: update source of truth;
4. *threadB*: `invalidate` in near cache;
5. *threadA*: `put` to near cache.

The value that was put in step 5 is stale, having been replaced in step 3 by a newer value. Cache invalidation in step 4 is overridden by the put in step 5. Now the cache is permanently inconsistent.

Fix: introduce the methods `announcePut` and `putAnnounced` to allow the following workflow:

1. `get` from near cache -> miss;
2. announce to update a cache key, obtain a _permit_ (an integer from an atomic sequence);
3. fetch from source of truth;
4. execute announced put, providing the permit;
5. if the permit checks out, near cache accepts the value and dismisses the permit.

In step 2, near cache remembers your permit and associates it with the key. In step 4 it checks that your permit is not stale.

The `invalidate` operation revokes any permit it finds.

Permit is accepted by `putAnnounced` if it is greater than or equal to the stored permit.

Synchronization had to be introduced because `announcePut`, `putAnnounced`, and `invalidate` must be mutually linearizable and they all maintain a complex invariant. Several update operations must be kept atomic by the lock.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.map.impl;
 
+import com.hazelcast.agrona.collections.Long2LongHashMap;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.EvictionPolicy;
 import com.hazelcast.config.InMemoryFormat;
@@ -47,9 +48,11 @@ public class NearCache {
      * Used when caching nonexistent values.
      */
     public static final Object NULL_OBJECT = new Object();
+    public static final Object NOT_ANNOUNCED = new Object();
     public static final String NEAR_CACHE_EXECUTOR_NAME = "hz:near-cache";
     private static final double EVICTION_FACTOR = 0.2;
     private static final int CLEANUP_INTERVAL = 5000;
+    private static final long NO_PERMIT = 0;
     private final int maxSize;
     private volatile long lastCleanup;
     private final long maxIdleMillis;
@@ -63,6 +66,9 @@ public class NearCache {
     private final NearCacheStatsImpl nearCacheStats;
     private final SerializationService serializationService;
     private final Comparator<NearCacheRecord> selectedComparator;
+    private final Object announcementLock = new Object();
+    private long announcementPermitSeq;
+    private final Long2LongHashMap announcements = new Long2LongHashMap(NO_PERMIT);
 
     private SizeEstimator nearCacheSizeEstimator;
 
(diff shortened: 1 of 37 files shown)


mtopolnik:
verify

serkan-ozal:
@mtopolnik 

Why did you changed `NearCacheTestSupport` and created a unit test (`NearCacheInvalidationTest`) for Map's near-cache under `cache.nearcache` package. Near-Cache implementations of `IMap` and `ICache` are different. So you should move `NearCacheInvalidationTest` (which is for `IMap`) under `map` package

serkan-ozal:
What is the added value of Agrona collections here?
Do we really need them for fixing this issue?

mtopolnik:
Yes, I used the `Long2LongHashMap` to improve memory footprint, avoid garbage emission, and improve performance.

>Why did you changed NearCacheTestSupport

What changes do you have in mind? I see only some trivial changes.

serkan-ozal:
Do we really need all of them for just `Long2LongHashMap`?

serkan-ozal:
And yes, seems that `Long2LongHashMap` dependents to others

mtopolnik:
I have prepared this backport a while ago so, since I needed this one class, I figured to throw them in as a complete package. There are many good use cases for them, another one has already been backported by Asim earlier on. This can now be a unified place where to grab them when nedeed.

hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
```diff
@@ -63,6 +66,9 @@
     private final NearCacheStatsImpl nearCacheStats;
     private final SerializationService serializationService;
     private final Comparator<NearCacheRecord> selectedComparator;
+    private final Object announcementLock = new Object();
+    private volatile long announcementPermitSeq;
```
serkan-ozal: `announcementPermitSeq` is only used (increased and get) in a `synchronized` block in `announcePut` method. So do we really need to define it as `volatile`?

mtopolnik: Yes, forgot to remove that.

mtopolnik:
Closed this PR; when I'm done with the lock-free implementation, I'll reopen.

PR #5481, opened 2015-06-09
Cleanup of com.hazelcast.monitor package

Donnerbart:
* Removed unnecessary `equals()` and `hashCode()` methods from statistic classes.
* Increased line coverage of all classes to 100%.

The tests for `MemberStateImpl` and `TimedMemberState` have been changed to not use `equals()` anymore, but just check for their member fields. The serialization of each class is tested separately.

All tests increase the according statistic class to 100% line coverage.


jerrinot:
:+1:

PR #5495, opened 2015-06-14
Moved DistributedObjectEventPacket to correct package

pveentjer:
This class is an implementation detail of the ProxyServiceImpl, so is
now moved into the proxyservice.impl package.


pveentjer:
ping

PR #5496, opened 2015-06-14
DefaultScheduledQueue supports Multiple Consumers

jerrinot:
This fixes OperationTimeoutException caused by a lost item when multiple generic operation threads tried to take an item from the queue.

This implementation provides the same guarantee when it comes to priority of urgent tasks as the original MPSC queue.

The stress test was kindly contributed by @pveentjer 

I'll send a PR with a forward port to a master branch once it's reviewed and merged.


pveentjer:
+1 Thanks Jaromir!

ajermakovics:
+1

PR #5497, opened 2015-06-14
DefaultScheduledQueue supports Multiple Consumers

jerrinot:
Forward port of #5496

This fixes OperationTimeoutException caused by a lost item when multiple generic operation threads tried to take an item from the queue.

This implementation provides the same guarantee when it comes to priority of urgent tasks as the original MPSC queue.

The stress test was kindly contributed by @pveentjer
(cherry picked from commit 9463460)


pveentjer:
:+1:

ahmetmircik:
Can you also provide a fix for maintenance branch @jerrinot ?

jerrinot:
@ahmetmircik: sure thing! I didn't know the `maintanance-3.x` has been already rebased on the `3.5`

jerrinot:
@ahmetmircik: here you go: #5504 
and thanks for a reminder!

PR #5502, opened 2015-06-15
increase cluster-quorum test coverage

eminn:


metanet:
:+1:

hazelcast/src/test/java/com/hazelcast/quorum/QuorumListenerTest.java
```diff
@@ -229,4 +257,16 @@ public void onChange(QuorumEvent quorumEvent) {
         }
         assertOpenEventually(belowLatch);
     }
+
+    static class QuorumListener implements com.hazelcast.quorum.QuorumListener {
+
+        static AtomicInteger failCount = new AtomicInteger();
+        @Override
+        public void onChange(QuorumEvent quorumEvent) {
+            assertEquals(3, quorumEvent.getThreshold());
```
pveentjer: I don't know if this is going to work. Because it will throw an Assertion error on the thread that calls the listener; not the thread that calls the test.

The if(.. isPresent) is the right approach to deal with the asynchronous behavior.

hazelcast/src/test/java/com/hazelcast/quorum/QuorumListenerTest.java
```diff
@@ -69,6 +71,32 @@ public void onChange(QuorumEvent quorumEvent) {
...
+        QuorumConfig quorumConfig = new QuorumConfig(quorumName, true, 3);
+        quorumConfig.addListenerConfig(listenerConfig);
+        config.getMapConfig(mapName).setQuorumName(quorumName);
+        config.addQuorumConfig(quorumConfig);
+        HazelcastInstance instance = createHazelcastInstance(config);
+        IMap<Object, Object> map = instance.getMap(mapName);
+        try {
+            map.put(generateKeyOwnedBy(instance), 1);
+        } catch (Exception e) {
+            e.printStackTrace();
```
pveentjer: I'm a bit unclear. Are you expecting this exception or not? If you are not expecting it, then I would just propagate it and let the test runner deal with the exception.

If you are expecting it, then I would expect something like this:

```
try{
   map.put(...)
   fail();
}catch(TypedException expected){
     ... in most cases you don't want to print the stacktrace.. just noise.
}
```

pveentjer:
@eminn ping. Any update on this pr?

PR #5503, opened 2015-06-15
[2.x] Fixes issue #3308: FactoryImpl.lsInstanceListeners leaks after client is disconnected

### LINKED ISSUE #3308: FactoryImpl maintaining references to ClientEndpoint objects after connection is closed leading to memory leak
In 2.x, FactoryImpl maintains references to ClientEndpoint objects even after the client connection has been closed.  Those lingering ClientEndpoint instances, in turn, prevent their CallContext objects from being garbage-collected (we have reported another issue that causes a leak of CallContext objects here: https://github.com/hazelcast/hazelcast/issues/2829).  This causes a memory leak as our client services are restarted.

See the heap dump below.  Note that the connection is not live, and that the ClientEndpoint objects are successfully removed from the ClientHandlerService's references.  The first image is an instance of ClientEndpoint with a connection that is still live, for comparison.

image

image

mdogan:
Fixes #3308

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/impl/ClientHandlerService.java
+++ b/hazelcast/src/main/java/com/hazelcast/impl/ClientHandlerService.java
@@ -240,6 +240,7 @@ public class ClientHandlerService implements ConnectionListener {
         return mapClientEndpoints.get(conn);
     }
 
+    // package-private for testing
     Collection<ClientEndpoint> getClientEndpoints() {
         return mapClientEndpoints.values();
     }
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/impl/ClientHandlerService.java
```diff
@@ -240,6 +240,10 @@ public ClientEndpoint getClientEndpoint(Connection conn) {
         return mapClientEndpoints.get(conn);
     }
 
+    Collection<ClientEndpoint> getClientEndpoints() {
```
jerrinot: is this added just for sake of tests?

mdogan: yes

PR #5520, opened 2015-06-17
Added tests for vector clock implementation

eminn:


hazelcast/src/test/java/com/hazelcast/replicatedmap/VectorClockTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+        clock2 = clock2.incrementClock(member1);
+        clock2 = clock2.incrementClock(member2);
+        clock2 = clock2.incrementClock(member2);
+        clock2 = clock2.incrementClock(member3);
+
+        // we need to represent happens before with an enum not with boolean
+        assertTrue(VectorClockTimestamp.happenedBefore(clock1, clock2));
+        assertFalse(VectorClockTimestamp.happenedBefore(clock2, clock1));
+
```
pveentjer: Why these emptylines?

hazelcast/src/test/java/com/hazelcast/replicatedmap/VectorClockTest.java
```diff
@@ -0,0 +1,267 @@
...
+    @Test
+    public void testConcurrentEventsShouldNotHaveHappensBeforeRelation() throws Exception {
+        VectorClockTimestamp clock1 = new VectorClockTimestamp();
+        VectorClockTimestamp clock2 = new VectorClockTimestamp();
+        MemberImpl member1 = new MemberImpl(new Address("127.0.0.1", 0));
+        MemberImpl member2 = new MemberImpl(new Address("127.0.0.2", 0));
+        clock1 = clock1.incrementClock(member1);
+        clock2 = clock2.incrementClock(member2);
+        // we need to represent happens before with an enum not with boolean
+        assertFalse(VectorClockTimestamp.happenedBefore(clock1, clock2));
```
pveentjer: I would introduce a custom assert assertHappensBefore and assertNotHappensBefore.

pveentjer:
Looks good. Minor comments.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/VectorClockTimestamp.java
```diff
@@ -129,7 +129,7 @@ static VectorClockTimestamp copyVector(VectorClockTimestamp vectorClockTimestamp
         return new VectorClockTimestamp(clocks);
     }
 
-    static boolean happenedBefore(VectorClockTimestamp x, VectorClockTimestamp y) {
+    public static boolean happenedBefore(VectorClockTimestamp x, VectorClockTimestamp y) {
```
metanet: Can we make this method non-static? Then we can use it like clock1.happenedBefore(clock2)

pveentjer: What do you mean? I personally prefer my asserts statics so there is no hidden context slipping in.

pveentjer: Ahhh... you are not talking about asserts... I like your suggestion!

hazelcast/src/test/java/com/hazelcast/replicatedmap/VectorClockTest.java
eminn: I find it much more easier to read this way.

hazelcast/src/test/java/com/hazelcast/replicatedmap/VectorClockTest.java
eminn: Good idea !

hazelcast/src/test/java/com/hazelcast/replicatedmap/VectorClockTest.java
Donnerbart: I think Peter refers to the empty lines after the asserts. There is nothing after that, so they are not really improve the readability. Same goes for the empty lines between tests and after the last test.

eminn: If so, then it is OK, they can be removed

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/VectorClockTimestamp.java
eminn: Well, that piece of code needs a refactor, since boolean return value makes it hard to use, which I'm not going to do in this PR. But I'll bear in my mind your suggestion when I'm doing those changes.

metanet: :+1:

PR #5521, opened 2015-06-17
[backport] Added tests for vector clock implementation

eminn:


pveentjer:
Although I don't mind having additional tests, is there a need to backport these tests?

eminn:
Since I'm backporting almost everything, I didn't give it a thought when I was raising this pull request. There is no need to backport the tests.

PR #5522, opened 2015-06-17
Seperated public/internal part of partition api

pveentjer:
Looks like a big PR, but actually is small. Most of the changes are due to moving classes. No other refactoring were done.


pveentjer:
Arrghh.. InternalPartition is part of public API in the SPI...

PR #5523, opened 2015-06-17
Moved ServiceInfo from spi to spi.impl.servicemanager

pveentjer:
Although this class was located in the com.hazelcast.spi package, there is no public API where this is being used. So it isn't exposed in any way apart from being in the spi module. So I consider it safe to
move inside the implementation.


gurbuzali:
:+1:

Donnerbart:
Provided that the assumption that `ServiceInfo` is no public API is correct: :+1:

PR #5533, opened 2015-06-18
[DON'T MERGE] transitive dependency fixes

mesutcelik:
* changed scope to *provided* for some dependencies
* excluded hazelcast-code-generator from hazelcast-all.jar
* skiped maven deploy for *hazelcast-code-generator*

### DIFF
--- a/hazelcast/pom.xml
+++ b/hazelcast/pom.xml
@@ -322,14 +322,16 @@
             <groupId>net.sourceforge.findbugs</groupId>
             <artifactId>annotations</artifactId>
             <version>1.3.2</version>
-            <scope>provided</scope>
+            <!--changing scope to provided gives problems with hazelcast-client module-->
+            <scope>compile</scope>
         </dependency>
 
         <dependency>
             <groupId>com.eclipsesource.minimal-json</groupId>
             <artifactId>minimal-json</artifactId>
             <version>0.9.2</version>
-            <scope>provided</scope>
+            <!-- shade plugin does not pick up this jar when changing scope to provided-->
+            <scope>compile</scope>
         </dependency>
 
         <dependency>
--- a/hazelcast/pom.xml
+++ b/hazelcast/pom.xml
@@ -322,20 +322,21 @@
             <groupId>net.sourceforge.findbugs</groupId>
             <artifactId>annotations</artifactId>
             <version>1.3.2</version>
-            <scope>compile</scope>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
             <groupId>com.eclipsesource.minimal-json</groupId>
             <artifactId>minimal-json</artifactId>
             <version>0.9.2</version>
-            <scope>compile</scope>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast-code-generator</artifactId>
             <version>${version}</version>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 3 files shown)


hazelcast/pom.xml
```diff
@@ -322,20 +322,21 @@
...
-            <scope>compile</scope>
+            <scope>provided</scope>
         </dependency>
 
         <dependency>
             <groupId>com.eclipsesource.minimal-json</groupId>
             <artifactId>minimal-json</artifactId>
             <version>0.9.2</version>
-            <scope>compile</scope>
+            <scope>provided</scope>
```
jerrinot: will this be still picked by the shade plugin?

mesutcelik: let me check

hazelcast/pom.xml
```diff
@@ -322,20 +322,21 @@
...
         </dependency>
 
         <dependency>
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast-code-generator</artifactId>
             <version>${version}</version>
+            <scope>provided</scope>
```
jerrinot: the generated class will still be part of the com.hazelcast:hazelcast artefact?

hazelcast/pom.xml
mesutcelik: seems it doesn't include if we change the scope so we should change it back to compile.
Do you know any other solution?

jerrinot: nope, sorry.
I'm no Maven wizard tho. Perhaps @Donnerbart might know the answer.

jerrinot: can't you explicitly specify artefacts you want to shade in regardless of theirs scope?

jerrinot:
context: https://groups.google.com/forum/#!topic/hazelcast/IVqFmbgxlao

hazelcast/pom.xml
jerrinot: https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#createDependencyReducedPom

we just need to upload the right pom.xml to Maven central.

jerrinot:
thinking about it more I believe @ajermakovics has to be our best expert when it comes to build systems! :)

hazelcast/pom.xml
mesutcelik: yes

mesutcelik:
I don't see why cloud compilation failed with this PR.

```14:36:48 [INFO] -------------------------------------------------------------
14:36:48 [ERROR] COMPILATION ERROR : 
14:36:48 [INFO] -------------------------------------------------------------
14:36:48 [ERROR] error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider com.hazelcast.client.protocol.generator.CodecCodeGenerator not found```

**Comment by mesutcelik** [comment]
verify

**Comment by mesutcelik** [comment]
another failure

```
QueryBasicTest.testQueryUsingNestedPortableObjectWithIndex
```

**Comment by mesutcelik** [comment]
verify

**Comment by jerrinot** [comment]
is it really fixing the problem? I don't see why Findbugs and the JSON lib should be visible at all as a dependency of the final artefact.

**Comment by mesutcelik** [comment]
This PR seems mostly like experimenting stuff. I see that **minimal-json**(shade plugin ignores compile scoped dependencies), **findbugs**(client module is using it) and **hazelcast-code-generator**( hazelcast-cloud does not compile with provided scope) has their own problems and different solutions. 

I will probably send another complete PR fixing all those problems but i still don't get why this PR is not compiling...

```
20:39:11 [ERROR] COMPILATION ERROR : 
20:39:11 [INFO] -------------------------------------------------------------
20:39:11 [ERROR] error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider com.hazelcast.client.protocol.generator.CodecCodeGenerator not found
```

jerrinot:
I see. thanks for clarification!

mesutcelik:
please check : https://github.com/hazelcast/hazelcast/pull/5542

PR #5535, opened 2015-06-18
[Backport] PagingPredicate Improvements

gurbuzali:
Removed anchor mechanism from `PagingPredicate`
added `setPage` method to `PagingPredicate`
refactored/enhanced internals

backport of https://github.com/hazelcast/hazelcast/pull/5534


jerrinot:
what are side-effects of this change?

gurbuzali:
the only and biggest side-effect is the increased data pulled from server to caller-side

jerrinot:
I don't quite understand why you removed the anchor. It's still useful when you are iterating page after page, isn't it?

hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java
```diff
@@ -298,14 +274,46 @@ public Predicate getPredicate() {
...
+
+    Map.Entry<Integer, Map.Entry> getNearestAnchorEntry() {
+        int size = anchorList.size();
+        if (page == 0 || size == 0) {
+            return NULL_ANCHOR;
+        }
+
+        Map.Entry anchoredEntry;
+        if (page <= size) {
+            anchoredEntry = anchorList.get(page - 1);
```
jerrinot: this is called for each entry, isn't it? if so then the `LinkedList` might not be the best idea.

hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java
```diff
@@ -298,14 +274,46 @@ public Predicate getPredicate() {
...
+        SimpleImmutableEntry anchorEntry = new SimpleImmutableEntry(page, anchor);
+        if (page < anchorList.size()) {
+            anchorList.set(page, anchorEntry);
+        } else {
+            anchorList.add(anchorEntry);
+        }
+    }
+
+    Map.Entry<Integer, Map.Entry> getNearestAnchorEntry() {
+        int size = anchorList.size();
```
jerrinot: `anchorCount` or something like this would be better. otherwise it might be confused with `pageSize`

hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java
```diff
@@ -71,16 +70,13 @@
  */
 public class PagingPredicate implements IndexAwarePredicate, DataSerializable {
 
-    private Predicate predicate;
+    private static final Map.Entry<Integer, Map.Entry> NULL_ANCHOR = new SimpleImmutableEntry(-1, null);
 
+    private final LinkedList<Map.Entry<Integer, Map.Entry>> anchorList = new LinkedList<Map.Entry<Integer, Map.Entry>>();
```
jerrinot: there is an assumption that on Xth position is an anchor for page X. Is this correct? Are you sure this assumption is always valid? the code is quite complex, it's not obvious too me.

jerrinot:
we could write a simulator test to compare performance against the current master

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1115,95 +1072,6 @@ protected long getTimeInMillis(final long time, final TimeUnit timeunit) {
         return new ClientMapEventHandler(listenerAdaptor, includeValue);
     }
 
-    private class ClientMapEventHandler implements EventHandler<PortableEntryEvent> {
```
jerrinot: is this just moved down in the file?

gurbuzali: yes, idea refactored it automatically

hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java
gurbuzali: Yes, that's what we assume (should be always valid)
we can add some asserts for that

hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java
gurbuzali: I can do that

hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java
gurbuzali: changed it to `ArrayList`

gurbuzali: although refactored the code so that it will be called once per partition not per entry
I've kept `ArrayList` change because we use `set` method of list a lot in order to set the anchors
it's better to have a random access list

jerrinot:
+1 once it's green. 
I'll write Simulator tests for this tomorrow.

hazelcast/src/main/java/com/hazelcast/query/PagingPredicate.java
```diff
@@ -298,14 +275,49 @@ public Predicate getPredicate() {
...
+        if (page < anchorCount) {
+            anchorList.set(page, anchorEntry);
+        } else if (page == anchorCount) {
+            anchorList.add(anchorEntry);
+        } else {
+            throw new IllegalArgumentException("Anchor index is not correct, expected: " + page + " found: " + anchorCount);
+        }
+    }
+
+    Map.Entry<Integer, Map.Entry> getNearestAnchorEntry() {
```
ahmetmircik: can you please document what nearest-anchor is?

ahmetmircik:
i have one minor comment other than PR seems mergeable to me.

gurbuzali:
will add JavaDoc after merging

PR #5542, opened 2015-06-19
findbugs and hazelcast-code-generator scope made provided

mesutcelik:
**findbugs:** I have moved up the findbugs dependency and made it provided. That way all dependent modules will get the dependency. Previously it was through hazelcast module.
**hazelcast-code-generator:** it is provided now. 
**minimal-json:** I could not find any solution for minimal-json.

### DIFF
--- a/hazelcast/pom.xml
+++ b/hazelcast/pom.xml
@@ -319,13 +319,6 @@
         </dependency>
 
         <dependency>
-            <groupId>net.sourceforge.findbugs</groupId>
-            <artifactId>annotations</artifactId>
-            <version>1.3.2</version>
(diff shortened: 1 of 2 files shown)


mesutcelik:
here is the issue created for minimal-json : https://github.com/hazelcast/hazelcast/issues/5543

mesutcelik:
when we switch **hazelcast-code-generator** dependency on **hazelcast** to *provided*, all dependent projects on **hazelcast** module is failing when compiled with Java 1.6.

It seems annotation processing with Java 1.6 sucks. 

thinking about solutions with @mdogan

mesutcelik:
verify

PR #5545, opened 2015-06-21
Minor IOBalancer tweaks

pveentjer:
Improved logging 
Removed logging bug where the wrong property was mention
Minor code cleanup

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/iobalancer/IOBalancer.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/iobalancer/IOBalancer.java
@@ -16,7 +16,6 @@
 
 package com.hazelcast.nio.tcp.iobalancer;
 
-import com.hazelcast.instance.GroupProperties;
 import com.hazelcast.instance.HazelcastThreadGroup;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.LoggingService;
@@ -29,13 +28,16 @@ import com.hazelcast.nio.tcp.ReadHandler;
 import com.hazelcast.nio.tcp.TcpIpConnection;
 import com.hazelcast.nio.tcp.WriteHandler;
 
+import static com.hazelcast.instance.GroupProperties.PROP_IO_BALANCER_INTERVAL_SECONDS;
+import static com.hazelcast.instance.GroupProperties.PROP_IO_THREAD_COUNT;
+
 /**
  * It attempts to detect and fix a selector imbalance problem.
  *
  * By default Hazelcast uses 3 threads to read data from TCP connections and 3 threads to write data to connections.
  * We have measured significant fluctuations of performance when the threads are not utilized equally.
  *
- * <code>com.hazelcast.nio.tcp.handlermigration.HandlerBalancer</code> tries to detect such situations and fix
+ * <code>com.hazelcast.nio.tcp.iobalancer.HandlerBalancer</code> tries to detect such situations and fix
  * them by moving {@link com.hazelcast.nio.tcp.ReadHandler} and {@link com.hazelcast.nio.tcp.WriteHandler} between
  * threads.
  *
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/iobalancer/IOBalancer.java
```diff
@@ -157,39 +158,43 @@ private MigrationStrategy createMigrationStrategy() {
...
+        }
+
         return true;
     }
 
     private void tryMigrate(LoadImbalance loadImbalance) {
         MigratableHandler handler = strategy.findHandlerToMigrate(loadImbalance);
         if (handler == null) {
-            log.finest("There had been I/O imbalance detected, but no suitable migration candidate was found.");
+            log.finest("I/O imbalance are detected, but no suitable migration candidate is found.");
```
Donnerbart: Active vs. passive?
`log.finest("Deteced I/O imbalance, but found no suitable migration candidate.");`

hazelcast/src/main/java/com/hazelcast/nio/tcp/iobalancer/IOBalancer.java
```diff
@@ -157,39 +158,43 @@ private MigrationStrategy createMigrationStrategy() {
...
             return;
         }
 
         IOSelector destinationSelector = loadImbalance.destinationSelector;
         if (log.isFinestEnabled()) {
             IOSelector sourceSelector = loadImbalance.sourceSelector;
-            log.finest("Scheduling a migration of a handler " + handler
+            log.finest("Scheduling a migration of handler " + handler
                     + " from a selector thread " + sourceSelector + " to " + destinationSelector);
```
Donnerbart: Would remove the "a" in "from a selector thread" as well. And maybe "the migration" instead of "a migration", but I'm not 100% sure about that.

```
log.finest("Scheduling the migration of handler " + handler
        + " from selector thread " + sourceSelector + " to " + destinationSelector);
```

pveentjer: Didn't see that one.

hazelcast/src/main/java/com/hazelcast/nio/tcp/iobalancer/IOBalancerThread.java
```diff
@@ -27,29 +27,37 @@
...
                 ioBalancer.checkWriteHandlers();
-                TimeUnit.SECONDS.sleep(migrationIntervalSeconds);
+                TimeUnit.SECONDS.sleep(balancerIntervalSeconds);
             }
         } catch (InterruptedException e) {
             log.finest("IOBalancer thread stopped");
             //this thread is about to exit, no reason restoring the interrupt flag
             EmptyStatement.ignore(e);
+        } catch (Throwable e) {
+            log.severe("IOBalancer failed", e);
```
Donnerbart: This seems to be the only change in behavior of this PR, if I didn't miss anything.

What's the reason to catch all throwables here? Prevent the I/O-Balancer from crashing completely if there is an error in a single `run()`?

pveentjer: Why do you think it is a change in behavior?

Donnerbart: `Throwable` covers `Error`, so beside all other exceptions you are consuming quite severe errors like OOME here, don't you? I'm just asking if there is a reason for doing this, since I'm not sure if this is a good practice.

metanet:
:+1:

hazelcast/src/main/java/com/hazelcast/nio/tcp/iobalancer/IOBalancerThread.java
pveentjer: I don't consume anything more or less than already is being ignored :)

Donnerbart: You added a `catch (Throwable e)` and just log it. If this is an OOME it won't be propagated anymore. Or do I miss anything?

As I said, it's just suspicious for me and I have to improve my knowledge about good practices in throwing and catching exceptions. From what I've read so far catching `Throwable` is not the best thing to do, as it includes `Error` which may need a proper handling (and not just logging).

pveentjer: Where should a thread propagate to :)

Donnerbart: Hmm, let me sleep about this (and read a bit more later). This might also be a good topic for an internal training (exceptions, errors, throwable, exception propagation, global exception handler, checked/unchecked, when to use which one, when to catch which one etc.).

Donnerbart:
:+1:

PR #5546, opened 2015-06-21
Minor Logging fix in IOBalancer

pveentjer:
It logs the wrong property when migrationIntervalSeconds < 0. It logs
PERFORMANCE_MONITOR_ENABLED but should log PROP_IO_BALANCER_INTERVAL_SECONDS.


Donnerbart:
Change seems to make sense, especially since `migrationIntervalSeconds` will be renamed to `balancerIntervalSeconds` in #5545.

:+1:

metanet:
:+1:

Issue #592, opened 2013-07-25
On the fly configuration of DistributedObjects


pveentjer:
As a user I would like to be able to do on the fly creation of e.g. a Map (also all the other structures).

MapConfig mapConfig = new MapConfig()
mapConfig.set....
IMap map = hz.getMap(mapConfig)

Sometimes it isn't possible to provide the configuration up front. With this mechanism one can define new configurations on the fly.

In a multi node setup, the configuration also needs to be distributed to all the nodes so that each node has the same configuration.

FlavioF:
+1

zeitos:
+1

pveentjer:
+1

donovankim:
+1

magiccrafter:
+1

bitsofinfo:
+1

bitsofinfo:
Please add functionality similar to the discovery strategy SPI for the WAN configuration...

abourdon:
Hi,

Sorry to re-up this issue but I'm facing a related need and I would like to know if there is a way to do it.

I would like to dynamically set an **existing** configuration.

For instance, based on the @pveentjer's sample, I would like to write something like:

```java
public static void main(String[] argv) throws Exception {
        HazelcastInstance hz = Hazelcast.newHazelcastInstance();

        hz.getConfig().addMapConfig(new MapConfig().setName("fly:peter").setTimeToLiveSeconds(1));
        IMap map = hz.getMap("fly:peter");
        map.put("foo", "bar");

        System.out.println("get:" + map.get("foo"));
        Thread.sleep(10000);
        System.out.println("get:" + map.get("foo"));

        // Setting the existing configuration
        hz.getConfig().getMapConfig("fly:peter").setTimeToLiveSeconds(100);
        map.put("foo", "bar");

        System.out.println("get:" + map.get("foo"));
        Thread.sleep(10000);
        System.out.println("get:" + map.get("foo"));

        System.out.println("Done");
    }
```

And be able to see the following output:

```
get:bar
get:null
get:bar
get:bar
Done
```

But re-configuration seems to be not applied to the map because real output is the following:

```
get:bar
get:null
get:bar
get:null
Done
```

Is there a way to dynamically set an existing configuration and apply changes to related objects?

Regards,
Aurélien

jerrinot:
[...]

Issue #380, opened 2012-12-17
Support for adding configuration entries in runtime


erljef:
This is a feature request for the possibility to add map configurations in runtime.

Our usage of Hazelcast assumed (incorrectly) that MapConfigs can be added in runtime. This seems to only partly be the case as configurations are only available on subsequently started nodes if the actual map has not been created yet anywhere in the cluster.

``` Java
HazelcastInstance instance1 = Hazelcast.newHazelcastInstance(new Config().setGroupConfig(new GroupConfig("test")));
HazelcastInstance instance2 = Hazelcast.newHazelcastInstance(new Config().setGroupConfig(new GroupConfig("test")));
HazelcastInstance instance3 = Hazelcast.newHazelcastInstance(new Config().setGroupConfig(new GroupConfig("test")));

instance1.getConfig().addMapConfig(new MapConfig("testMap").setBackupCount(3));
instance2.getConfig().addMapConfig(new MapConfig("testMap").setBackupCount(3));
instance3.getConfig().addMapConfig(new MapConfig("testMap").setBackupCount(3));

instance1.getMap("testMap");
instance2.getMap("testMap");
instance3.getMap("testMap");
```

The previous sample works; the IMap instances for map "testMap" are all backed by CMaps using the correct configuration (name='testMap', backup-count=3).

The following, however, does not work.

``` Java
HazelcastInstance instance1 = Hazelcast.newHazelcastInstance(new Config().setGroupConfig(new GroupConfig("test")));
instance1.getConfig().addMapConfig(new MapConfig("testMap").setBackupCount(3));
instance1.getMap("testMap");

HazelcastInstance instance2 = Hazelcast.newHazelcastInstance(new Config().setGroupConfig(new GroupConfig("test")));
instance2.getConfig().addMapConfig(new MapConfig("testMap").setBackupCount(3));
instance2.getMap("1");

HazelcastInstance instance3 = Hazelcast.newHazelcastInstance(new Config().setGroupConfig(new GroupConfig("test")));
instance3.getConfig().addMapConfig(new MapConfig("testMap").setBackupCount(3));
instance3.getMap("1");
```

What happens is that when instance2 and -3 are created and join the cluster they receive the map names from instance1 and creates their own CMaps for the received names. This happens before the MapConfig for the map has been added and the default configuration is used for the CMap instead.

Failing test case:

``` Java
import org.junit.Test;

import com.hazelcast.config.Config;
import com.hazelcast.config.GroupConfig;
[...]

PR #5547, opened 2015-06-21
5485 - fix to keep near cache config to be in sync with map container

### LINKED ISSUE #5485: Memory leak in hazelcast cluster caused by race condition on node join
The way how issue is reproducible in _our_ environment
* Start _node1_
* Load _node1_ with large number of http requests (each request will result into hundreds of hazelcast operations)
* Starting _node2_ when load is running
* You'll see map size node _node2_ growing above configured max size. Issue is _not_ reproducible if load is applied after node joined.

I was unable to reproduce issue using unit test, however when debugging it looks like check in 
MaxSizeChecker#isEvictablePerNode doesn't succeed (hz. v3.4.2) and number of items returned is less than number if item returned by map.size.

Issue is reproducible on hz 3.3.1 as well.

Would be awesome to know if this is known issue or/and if any workaround exist.

PetroSemeniuk:
Hi guys,

  This PR addresses next issues:
* Fixes memory leak which happens if near cache invalidation event was received, but map container for certain map wasn't yet configured (ie. was added later when hazelcast instance was already started). In that case near cache max will have size of max integer.
* Fixes problem when map configuration can't be updated at runtime (I believe this functionality is part of Hazecast Management Console). While map container can be successfully reconfigured, corresponding near cache will container stale value of max size.

On my side all tests are passing. Please let me know whether anything else need to be done.

Cheers
Petro

Fixes #5485

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
@@ -50,13 +50,12 @@ public class NearCache {
     public static final String NEAR_CACHE_EXECUTOR_NAME = "hz:near-cache";
     private static final double EVICTION_FACTOR = 0.2;
     private static final int CLEANUP_INTERVAL = 5000;
-    private final int maxSize;
     private volatile long lastCleanup;
-    private final long maxIdleMillis;
-    private final long timeToLiveMillis;
     private final EvictionPolicy evictionPolicy;
     private final InMemoryFormat inMemoryFormat;
     private final NodeEngine nodeEngine;
+    private final MapContainer mapContainer;
+    private final SizeEstimator nearCacheSizeEstimator;
     private final AtomicBoolean canCleanUp;
     private final AtomicBoolean canEvict;
     private final ConcurrentMap<Data, NearCacheRecord> cache;
@@ -64,20 +63,16 @@ public class NearCache {
     private final SerializationService serializationService;
     private final Comparator<NearCacheRecord> selectedComparator;
 
-    private SizeEstimator nearCacheSizeEstimator;
-
     /**
-     * @param mapName    name of map which owns near cache.
+     * @param mapContainer mapContainer which owns near cache.
      * @param nodeEngine node engine.
      */
-    public NearCache(String mapName, NodeEngine nodeEngine) {
+    public NearCache(NodeEngine nodeEngine, MapContainer mapContainer) {
         this.nodeEngine = nodeEngine;
+        this.mapContainer = mapContainer;
         Config config = nodeEngine.getConfig();
-        NearCacheConfig nearCacheConfig = config.findMapConfig(mapName).getNearCacheConfig();
-        maxSize = nearCacheConfig.getMaxSize() <= 0 ? Integer.MAX_VALUE : nearCacheConfig.getMaxSize();
-        maxIdleMillis = TimeUnit.SECONDS.toMillis(nearCacheConfig.getMaxIdleSeconds());
+        NearCacheConfig nearCacheConfig = config.findMapConfig(mapContainer.getName()).getNearCacheConfig();
         inMemoryFormat = nearCacheConfig.getInMemoryFormat();
-        timeToLiveMillis = TimeUnit.SECONDS.toMillis(nearCacheConfig.getTimeToLiveSeconds());
         evictionPolicy = EvictionPolicy.valueOf(nearCacheConfig.getEvictionPolicy());
         selectedComparator = NearCacheRecord.getComparator(evictionPolicy);
         cache = new ConcurrentHashMap<Data, NearCacheRecord>();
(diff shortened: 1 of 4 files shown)


PetroSemeniuk:
@devOpsHazelcast hey, al tests were 100% green on my side.

I don't ave access to your cloudbees setup, can you post it here what failures were?

Donnerbart:
It was not a test failure, but CheckStyle issue. Please use the Checkstyle and Findbugs profile locally to find these :)

```
00:00:29.422 [INFO] --- maven-checkstyle-plugin:2.12:checkstyle (default) @ hazelcast ---
00:00:44.697 [INFO] Starting audit...
00:00:44.697 /scratch/jenkins/workspace/Hazelcast-3.x-pr-builder/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java:87:28: '{' is not preceded with whitespace.
00:00:44.697 /scratch/jenkins/workspace/Hazelcast-3.x-pr-builder/hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java:89:48: ':' is not preceded with whitespace.
```

hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
```diff
@@ -86,12 +81,26 @@ public NearCache(String mapName, NodeEngine nodeEngine) {
         nearCacheStats = new NearCacheStatsImpl();
         lastCleanup = Clock.currentTimeMillis();
         serializationService = nodeEngine.getSerializationService();
+        nearCacheSizeEstimator = mapContainer.getNearCacheSizeEstimator();
+    }
+
+    public int getMaxSize(){
```
Donnerbart: Missing whitespace `){`

hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
```diff
@@ -86,12 +81,26 @@ public NearCache(String mapName, NodeEngine nodeEngine) {
...
+        final int maxSize = mapContainer.getMapConfig().getNearCacheConfig().getMaxSize();
+        return maxSize <= 0 ? Integer.MAX_VALUE: maxSize;
```
Donnerbart: Here is the other one.

hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
PetroSemeniuk: Fixed. Thanks!

hazelcast/src/main/java/com/hazelcast/map/impl/NearCache.java
PetroSemeniuk: Fixed. Thanks!

ahmetmircik:
Hi @PetroSemeniuk, can you also provide separate tests for the fixed issues?

gurbuzali:
https://github.com/hazelcast/hazelcast/issues/5485

PetroSemeniuk:
@ahmetmircik I added test for max size check. Would that be enough?

ahmetmircik:
Hi @PetroSemeniuk, can you please also write a test for the other case? Btw, we are not fully  supporting runtime configuration for `IMap` atm only some sub-set of  `IMap` config can be updateable over management center.

PetroSemeniuk:
Hi @ahmetmircik , I added unit test for another use case as well. 

I'm aware that runtime configuration of IMap officially is not fully supported (few related links: http://stackoverflow.com/questions/21379171/hazelcast-mapstoreconfig-ignored, https://github.com/hazelcast/hazelcast/issues/592 and https://github.com/hazelcast/hazelcast/issues/380) however in our case almost all of our imaps are configured dynamically (we have more than hundred of imaps, holding tens of thousand of items and getting zillions of hits. For us updating max size is the most crucial (I believe there should be at least few more customers with this need) and this PR addresses this.

It doesn't compromise near cache performance or architecture in any way and simply makes reasoning about config state of near cache more straightforward.

hazelcast/src/test/java/com/hazelcast/map/EvictionTest.java
```diff
@@ -107,6 +116,62 @@ public void testGetEntryView_withTTL() throws Exception {
...
+                .setNearCacheConfig(nearCacheConfig);
+
+        final TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(1);
+        final HazelcastInstance h = factory.newHazelcastInstance(cfg);
+        final IMap<Object, Object> map = h.getMap(mapName);
+        for (Integer i = 0; i < size * 2; i++) {
+            map.put(i.toString(), i.toString());
+            map.get(i.toString());//trigger put into near cache
+        }
+        sleepSeconds(3); //wait for eviction
```
ahmetmircik: Can you change `sleepSeconds` with `assertTrueEventually `? 

```
                @Override
                     NearCacheStats originalNearCacheStats = map.getLocalMapStats().getNearCacheStats();
                     assertThat(originalNearCacheStats.getOwnedEntryCount(), lessThan((long)size));
                     assertThat(originalNearCacheStats.getOwnedEntryCount(), greaterThan((long) (size/2)));                
              }
            });

```

hazelcast/src/test/java/com/hazelcast/map/EvictionTest.java
```diff
@@ -107,6 +116,62 @@ public void testGetEntryView_withTTL() throws Exception {
...
+        final MapConfig halvedMapConfig = mapConfig
+                .setMaxSizeConfig(halvedMaxSizeConfig)
+                .setNearCacheConfig(halvedNearCacheConfig);
+        mapContainer.setMapConfig(halvedMapConfig);
+
+        for (Integer i = 0; i < size * 2; i++) {
+            map.put(i.toString(), i.toString());
+            map.get(i.toString());//trigger put into near cache
+        }
+        sleepSeconds(3); //wait for eviction
```
ahmetmircik: same here.

hazelcast/src/test/java/com/hazelcast/map/EvictionTest.java
PetroSemeniuk: Done.

hazelcast/src/test/java/com/hazelcast/map/EvictionTest.java
PetroSemeniuk: Done.

PetroSemeniuk:
@ahmetmircik all done. Thanks for looking into that!

P.S. Sorry for late reply, was busy optimising memory in our own product and missed your message.

ahmetmircik:
Thanks for this PR but since we don't support dynamic configuration, i tend to close it and will revisit it later if we decide to have dynamic configuration.

PR #5549, opened 2015-06-22
Introduced OperationResponseHandler

pveentjer:
Replaced the ResponseHandler by the OperationResponseHandler. The old ResponseHandler is still working; so we are backwards compatible. But all interactions with the ResponseHandler in HZ have been replaced with the OperationResponseHandler.

In this PR the main advantage is that the OperationResponseHandler is not bound to a particular operation like the ResponseHandler. So this saves 1 object for every remote operation.

In itself this isn't very exciting, but the OperationResponseHandler is also going to receive specific methods for all the different types of responses we are sending; instead of relying on the Response objects. This way we can get rid of all the response objects and potentially get rid of the whole response thread. Currently it can't be removed because the response-thread does quite a lot of heavy lifting, like response deserialization, but in the future a lot of this work can be removed. This PR is the first phase to get there.

The PR looks big, but it is mostly caused by the simplification of operations that access the ResponseHandler.


pveentjer:
veriy

pveentjer:
verify

gurbuzali:
:+1:

mdogan:
looks good. I liked the idea of using the same response-handler for all (nearly) kind of operations instead of creating one for each.

PR #5550, opened 2015-06-22
Configurable InvocationFuture's callback-executor 

ahmetmircik:
Made InvocationFuture's callback-executor settableso a custom executor can be given to run a callback.

This is first part of the fix of https://github.com/hazelcast/hazelcast/issues/4671

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
@@ -21,6 +21,8 @@ import com.hazelcast.nio.Address;
 import com.hazelcast.partition.InternalPartition;
 import com.hazelcast.spi.impl.NodeEngineImpl;
 
+import java.util.concurrent.Executor;
+
 /**
  * The InvocationBuilder is responsible for building an invocation of an operation and invoking it.
  * <p/>
@@ -61,6 +63,7 @@ public abstract class InvocationBuilder {
     protected final Address target;
     protected Callback<Object> callback;
     protected ExecutionCallback<Object> executionCallback;
+    protected Executor callbackExecutor;
 
     protected long callTimeout = DEFAULT_CALL_TIMEOUT;
     protected int replicaIndex;
@@ -86,7 +89,7 @@ public abstract class InvocationBuilder {
         this.target = target;
     }
 
-     /**
+    /**
      * Sets the replicaIndex.
      *
      * @param replicaIndex the replica index
@@ -165,6 +168,7 @@ public abstract class InvocationBuilder {
 
     /**
      * Gets the operation to execute.
+     *
      * @return the operation to execute
      */
     public Operation getOp() {
@@ -175,7 +179,7 @@ public abstract class InvocationBuilder {
      * Gets the replicaIndex.
      *
      * @return the replicaIndex
-      */
+     */
 
     public int getReplicaIndex() {
         return replicaIndex;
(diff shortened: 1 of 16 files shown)


gurbuzali:
:+1: 
will you backport this to maintenance ?

hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
```diff
@@ -265,5 +268,20 @@ public InvocationBuilder setExecutionCallback(ExecutionCallback<Object> executio
...
 
+    /**
+     * Sets the {@code callbackExecutor} in which the {@link #executionCallback} will be run
+     *
+     * @param callbackExecutor the callbackExecutor in which the callback will be run
+     * @return the updated InvocationBuilder.
+     * @throws java.lang.IllegalStateException if an callbackExecutor has already been set.
+     */
+    public InvocationBuilder setCallbackExecutor(Executor callbackExecutor) {
+        if (this.callbackExecutor != null) {
```
pveentjer: I would use the preconditions for this.

pveentjer: Should there not be a check if the calllbackExecutor is null? I don't think that is a desirable situation. Or otherwise it needs to be documented.

pveentjer:
Dumb question: The InvocationFuture.andThen method allows for passing in an Executor that can be used for the callback.

So if you would change the signature of your new method from:
```
Map<Integer, Object> invokeOnPartitions(String serviceName, OperationFactory operationFactory,
                                           Collection<Integer> partitions, Executor callbackExecutor,
                                            ExecutionCallback callback) throws Exception;
```

to:

```
Map<Integer, ICompletableFuture> asyncInvokeOnPartitions(String serviceName, OperationFactory operationFactory,
                                           Collection<Integer> partitions) throws Exception;
```

Then you can just do:
```
for(ICompletableFuture f: asyncInvokeOnPartition...){
    f.andThen(callback, executor);
}
```
This way you dramatically reduce the number of code changes.

ahmetmircik:
I want to be sure that my callback is given to the supplied executor in the response setter thread.  Otherwise there is a possibility that the `andThen` method can be called from a user thread. So i need to  set callback & executor before creating invocation future.

hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
ahmetmircik: documented.

ahmetmircik:
verify

pveentjer:
```
I want to be sure that my callback is given to the supplied executor in the response setter thread. Otherwise there is a possibility that the andThen method can be called from a user thread.
As a result, i need to set callback & executor before creating invocation future.
```

There still is no need for all these changes. You can still make use of the andThen with the customizable executor. But then you move the logic inside of the OperationService.

The Invocation/Future is already is super complex beast and I don't want add more logic to it for no good reason. The andThen provides the behavior your are looking for.

pveentjer:
I think we should close this request. There is an assumption that HZ will provide an ordering on invocations and their responses. This is an assumption that can lead to problems in the future:
- what if we add multiple response threads on the same queue to reduce load
- what if we add multiple connections between members to reduce load
- what if connections become more flexible with the order of packets being send

We should refrain from such unwanted ordering contraints so that we have a lot of design flexibility. Otherwise we'll run into very vague race problems when we start to change the hz internals.

ahmetmircik:
closing for now.

PR #5553, opened 2015-06-22
Allow to use JDK8 JavaScript scripting engine in OSGi platforms

adessaigne:
This patch allows to use the JavaScript scripting engine of JDK8 (named _Nashorn_). It is still capable of using the one of JDK7 (named _Rhino_).

Without this patch, the following log is displayed when running in JDK8
```
2015-05-12 16:48:06,805 [FelixStartLevel] WARN com.hazelcast.osgi.OSGiScriptEngineManager - Found ScriptEngineFactory candidate for com.sun.script.javascript.RhinoScriptEngineFactory, but cannot load class! -> java.lang.ClassNotFoundException: com.sun.script.javascript.RhinoScriptEngineFactory not found by com.hazelcast [156]
```

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/osgi/OSGiScriptEngineManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/osgi/OSGiScriptEngineManager.java
@@ -18,6 +18,7 @@ package com.hazelcast.internal.osgi;
 
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
+import com.hazelcast.nio.ClassLoaderUtil;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 
@@ -245,7 +246,7 @@ public class OSGiScriptEngineManager extends ScriptEngineManager {
     private ClassLoader loadScriptEngineFactoryClassLoader(String factoryName) {
         //We do not really need the class, but we need the classloader
         try {
-            return tryLoadClass(factoryName).getClassLoader();
+            return ClassLoaderUtil.tryLoadClass(factoryName).getClassLoader();
         } catch (ClassNotFoundException cnfe) {
             // may fail if script implementation is not in environment
             logger.warning("Found ScriptEngineFactory candidate for "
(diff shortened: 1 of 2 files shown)


adessaigne:
verify

bilalyasar:
verify

adessaigne:
Do this test failure related to my pull-request ? I failing at finding a connection between the error and my pull-request.

serkan-ozal:
verify

serkan-ozal:
@adessaigne Thanks for your work. Could you please sign [Hazelcast Contributor Agreement](https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement)

adessaigne:
@serkan-ozal I've just sent the email with a scan of my signed Hazelcast Contributor Agreement.

pveentjer:
verify

ajermakovics:
Seems like ClassLoaderUtil could be used here instead of custom methods

adessaigne:
Yes, I thought that too, but the ```ClassLoaderUtil``` is located in ```com.hazelcast.nio``` and not in ```com.hazelcast.util```. Also its ```loadClass``` method doesn't have the fallback mechanism from one class loader to another when there is a ```ClassNotFoundException```.

ajermakovics:
If the existing method isn't suitable then I think a new one can be added to ClassLoaderUtil. Having these methods in one place is makes it easier to discover them. The existing one seems to have some fallback though I don't know it works in this case

verify

adessaigne:
If I update the code, the ```com.hazelcast.internal.osgi``` package will have a dependency towards ```com.hazelcast.nio```. I found it a bit odd, are you ok with that ?

As of today, the dependencies of ```com.hazelcast.internal.osgi``` are : ```com.hazelcast.core.```, ```com.hazelcast.logging``` and ```com.hazelcast.internal.management```

ajermakovics:
nio may not be the best place for ClassLoaderUtil anymore and eventually we should move it to util but that's probably out of scope of this pr. Other than that it's a +1 from me, not sure what others think

mesutcelik:
:+1:  for moving methods to `ClassLoaderUtil` only.

adessaigne:
OK then, I'll move ```tryLoadClass``` and ```isClassDefined``` into ```ClassLoaderUtil```.
How should I rename my ```tryLoadClass``` method because there already is a method named that way in ```ClassLoaderUtil``` ?

adessaigne:
What should I do in order to help you integrate this PR into the code base ?

ajermakovics:
Any reason the existing loadClass method can't be used? It also does fallback. It'd be better if we didn't have two methods doing very similar things. 
If it can't be used then moving the methods should be enough. Doesn't need renaming since signatures are different.

adessaigne:
The ```loadClass``` method is indeed different from my ```tryLoadClass``` method. My first attempt is with ```Class.loadClass(className)``` which is the last statement of the ```loadClass``` method but it's never called due to logic above.

ajermakovics:
verify

mesutcelik:
verify

adessaigne:
Hi everyone ? Is everything ok ? Did I forget to do something to help you integrate this PR ?

pveentjer:
@ajermakovics @mesutcelik what should we do with this one?

adessaigne:
Thanks a lot !

mesutcelik:
Thanks goes to you for the contribution @adessaigne !

PR #5558, opened 2015-06-22
Removed deprecated method distributedobject getid

pveentjer:


gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/core/DistributedObjectEvent.java
```diff
@@ -68,7 +68,7 @@ public EventType getEventType() {
      * @return the identifier of DistributedObject
      */
     public Object getObjectId() {
```
jerrinot: shouldn't this method be renamed/deprecated as well? 
as now there is no obvious connection between `eventObjectId` and `distributedObjectName`.

pveentjer: Let me see.

pveentjer: Good spotting. I have deprecated that method and added a getObjectName method.

jerrinot:
:+1: (once it's green)

PR #5571, opened 2015-06-23
Serialization ByteArrayObjectDataInput pooling

pveentjer:
Currently the ByteArrayObjectDataInput is not pooled in the SerializationServiceImpl; unlike the ByteArrayObjectDataOutput. This PR fixes this problem, this means that for every deserialization one object less is created.

This means that for a remote executed operation, there are at least 2 objects less being created. A deserialization of the operation and a deserialization of the response. In practice the number could be even higher, if there is nested deserialization. 

Another minor improvement are:
* for every deserialization 2 ConcurrentHashMap accessed were done; now only 1.
* with the old approach there we no bound on the size of the pooled inputs due to a programming mistake. With the new approach there is a cap of 3 (which was the original intent.. as discussed with Mehmet).

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/serialization/ByteArrayObjectDataInput.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/serialization/ByteArrayObjectDataInput.java
@@ -34,7 +34,7 @@ class ByteArrayObjectDataInput extends InputStream implements BufferObjectDataIn
 
     byte[] data;
 
-    final int size;
+    int size;
 
     int pos;
 
@@ -51,12 +51,25 @@ class ByteArrayObjectDataInput extends InputStream implements BufferObjectDataIn
     }
 
     ByteArrayObjectDataInput(byte[] data, int offset, SerializationService service, ByteOrder byteOrder) {
-        super();
         this.data = data;
         this.size = data != null ? data.length : 0;
         this.service = service;
-        pos = offset;
-        bigEndian = byteOrder == ByteOrder.BIG_ENDIAN;
+        this.pos = offset;
+        this.bigEndian = byteOrder == ByteOrder.BIG_ENDIAN;
+    }
+   @Override
+    public void init(byte[] data, int offset) {
+        this.data = data;
+        this.size = data != null ? data.length : 0;
+        this.pos = offset;
+    }
+
+    @Override
+    public void clear() {
+        this.data = null;
+        this.pos = 0;
+        this.size = 0;
+        this.mark = 0;
     }
 
     public int read() throws IOException {
--- a/hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
@@ -23,12 +23,11 @@ import java.util.Arrays;
 
 @edu.umd.cs.findbugs.annotations.SuppressWarnings("EI_EXPOSE_REP")
 public final class DefaultData implements Data {
-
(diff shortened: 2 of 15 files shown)


hazelcast/src/main/java/com/hazelcast/nio/serialization/ByteArrayObjectDataInput.java
```diff
@@ -51,12 +51,23 @@
...
+    public void init(byte[] data, int offset) {
+        this.data = data;
+        this.size = data != null ? data.length : 0;
+        this.pos = offset;
+        this.mark = 0;
+    }
+
+    @Override
+    public void clear() {
+        this.data = null;
```
mdogan: What about zeroing `size` and `pos`? Otherwise methods like `available()` or `position()` can report incorrect values after clear...

hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializationServiceImpl.java
```diff
@@ -252,7 +251,7 @@ protected final int calculatePartitionHash(Object obj, PartitioningStrategy stra
         return partitionHash;
     }
 
-    public final <T> T toObject(final Object object) {
+    public final <T> T toObject2(final Object object) {
```
mdogan: `toObject2()` ??

hazelcast/src/main/java/com/hazelcast/nio/serialization/ByteArrayObjectDataInput.java
pveentjer: Sure.. Will do.

hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializationServiceImpl.java
pveentjer: Ahhh... ok.. Need to remove that..  I was switching between the 2 implementations.

pveentjer:
As discussed with Mehmet. Hz enterprise needs special types of streams based. I have modified the design so that in HZ enterprise the pool implementation can be swapped out by a different one that allows for the HZ enterprise features.

pveentjer:
verify

pveentjer:
verify

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
```diff
@@ -24,11 +24,12 @@
 @edu.umd.cs.findbugs.annotations.SuppressWarnings("EI_EXPOSE_REP")
 public final class DefaultData implements Data {
 
+    public static final int DATA_OFFSET = 5;
```
mdogan: Can you also make `TYPE_OFFSET` and `PARTITION_HASH_BIT_OFFSET` public and keep them in order again? To me, seeing them in order makes easier to read/understand. 

```java
public static final int TYPE_OFFSET = 0;
public static final int PARTITION_HASH_BIT_OFFSET = 4;
public static final int DATA_OFFSET = 5;
```

mdogan:
just a minor comment. 

:+1:

mdogan:
Can you also prepare a PR for enterprise too? otherwise it will fail when this is merged.

hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
pveentjer: Will do.

pveentjer:
The PR is pending.

gurbuzali:
:+1:

PR #4110, opened 2014-11-19
Reimplement hazelcast hibernate caching

cszmajda:
# tl;dr

The previous implementation of Hibernate L2 caching had incorrect transactional behavior and would cache stale values and/or fail with "IllegalMonitorStateException: current thread not owner of the lock" under heavy load.

This PR re-implements the Hibernate region cache to implement Hibernate "soft locking" correctly. The implementation is loosely based on Ehcache (Apache 2 licensed), which should be considered the "reference" implementation of Hibernate's L2 caching SPI. Calls to `IMap.lock` have been replaced wherever possible with CAS or entry processors, resulting in more efficient and better transactional semantics, even in edge cases such as server crashing and network partitions.

We've been running this implementation in production for months, and have subjected it to extremely heavy stress testing in our cluster lab (up to 400 threads spread over 4 high-spec machines subjecting a cluster of 1-4 nodes to continuous hibernate-intensive load). Problems we had in the previous implementation due to stale data and "IllegalMonitorStateException: current thread not owner of the lock" errors have been completely eliminated. 

# Soft locking

The main difference in this implementation over the old one is that this one correctly implements how Hibernate's soft locks are suppose to behave. There is some documentation on this behavior in the [JavaDoc for ExpiryMarker](https://github.com/hazelcast/hazelcast/pull/4110/files#diff-75a485177b9e1e7dcf639fda2844cd5fR27). Soft-locking requires the following characteristics:

* Multiple transactions (i.e. threads) can soft lock an entry concurrently
* An entry isn't unlocked until all transactions have completed and unlocked the entry
* Soft locks have an expiry/timeout timestamp to handle the case where an transaction fails to unlock after locking (e.g server crash, network partition)
* The entry is considered null (i.e. a cache miss) for the duration of it being locked

The previous implementation did not meet these requirements:

* It attempted to lock an entry exclusively for a set timeout, if it failed to lock the entry it did not attempt to unlock it
* It did not report the cache entry as empty when it is soft locked, causing stale cache values to be used by hibernate when determining what it should do
* Entries were not unlocked in the case of server crashing. 
* Locks were silently released in the event of a network partition that later heals.  Notice that `ClusterServiceImpl.merge` resets all managedServices including every lockStore. This means that all locks held are effectively released, invariably leading to `java.lang.IllegalMonitorStateException: Current thread is not owner of the lock! -> <not-locked>`

PR #5572, opened 2015-06-23
 Hibernate 3 port PR #4110

kobalski:
Soft locking improvements, use of EntryProcessors for cache operations.
Hibernate 3 port for #4110


kobalski:
verify

PR #5574, opened 2015-06-23
Removed object litter in FastStringCreator

pveentjer:
The FastStringCreator makes use of reflection and therefor a varargs is required.
This varargs array is just litter and is now pooled by putting it in a threadlocal.

Main reason to target litter is to reduce gc's and therefor get a shorter tail on the latency distribution.


Donnerbart:
:+1:

Issue #5516, opened 2015-06-17
MaxSizeChecker.getApproximateMaxSize: implicit cast to int breaks isEvictableHeapPercentage, possibly others


LoneRifle:
When the eviction policy of a distributed IMap is set to heap percentage with a threshold of 1%, MaxSizeChecker will invoke isEvictableHeapPercentage when determining whether to evict entries. getApproximateMaxSize will then be invoked with a maxSizeFromConfig argument of 1.

Because the return type of getApproximateMaxSize is an int, the expression:
`maxSizeFromConfig * EVICTION_START_THRESHOLD_PERCENTAGE / ONE_HUNDRED_PERCENT `
will be cast from 0.95 to 0, such that:
`maxSize < (1D * ONE_HUNDRED_PERCENT * usedHeapSize / total)`
will always evaluate to true, and entries will always be evicted.

PR #5579, opened 2015-06-24
Fixes getApproximateMaxSize calculation bug

ahmetmircik:
fixes https://github.com/hazelcast/hazelcast/issues/5516

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoaderUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoaderUtil.java
@@ -51,7 +51,7 @@ public final class MapKeyLoaderUtil {
     }
 
     static Iterator<Map<Integer, List<Data>>> toBatches(final Iterator<Entry<Integer, Data>> entries,
-            final int maxBatch) {
+                                                        final int maxBatch) {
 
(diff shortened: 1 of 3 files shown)


eminn:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoaderUtil.java
```diff
@@ -86,15 +86,14 @@ public boolean hasNext() {
...
-        int maxSizePerNode = getApproximateMaxSize(maxSizeConfig, MaxSizePolicy.PER_NODE);
+        double maxSizePerNode = getApproximateMaxSize(maxSizeConfig, MaxSizePolicy.PER_NODE);
 
         if (maxSizePerNode == MaxSizeConfig.DEFAULT_MAX_SIZE) {
             // unlimited
             return -1;
         }
 
-        return maxSizePerNode;
+        return (int) maxSizePerNode;
```
gurbuzali: don't we have the same effect here ? 
casting 0.95 to 0 ?

ahmetmircik: That narrowing should be fine, at some point we should do that. Main issue we are trying to fix here appears in percentage comparisons.

PR #5580, opened 2015-06-24
Fixes getApproximateMaxSize calculation bug

### LINKED ISSUE #5516: MaxSizeChecker.getApproximateMaxSize: implicit cast to int breaks isEvictableHeapPercentage, possibly others

ahmetmircik:
fixes #5516


eminn:
:+1:

PR #5582, opened 2015-06-24
[DONT MERGE] Reduce Object litter Selector

pveentjer:
The Selector internally uses a Hashset to store the SelectionKeys. The problem is that for every SelectionKey being added, litter is being created. 

To resolve this problem; the FastSelectionKeys modifies the internally stored HashSet and replaces it by an optimized version that doesn't generate litter. 

This trick comes I saw first in Netty. (https://github.com/netty/netty/commit/3ce9ab2e72235bf2ae45f57de11803249419cb69)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/FastSelectionKeys.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/FastSelectionKeys.java
@@ -1,0 +1,210 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/FastSelectionKeys.java
```diff
@@ -0,0 +1,210 @@
...
+        SelectionKeys tmp = activeKeys;
+        activeKeys = passiveKeys;
+        passiveKeys = tmp;
+
+        activeKeys.size = 0;
+        return passiveKeys.keys;
+    }
+
+    @Override
+    public boolean remove(Object o) {
```
mdogan: Is this not-implemented because it's not called? In that case, I think it's better to throw unsupported exception instead of no-op.

mdogan: It's used in `AbstractPollSelectorImpl.implDereg()` and `KQueueSelectorImpl.implDereg()`.

hazelcast/src/main/java/com/hazelcast/nio/tcp/FastSelectionKeys.java
```diff
@@ -0,0 +1,210 @@
...
+        return false;
+    }
+
+    @Override
+    public boolean contains(Object o) {
```
mdogan: `contains` is used in `KQueueSelectorImpl.updateSelectedKeys()` and `AbstractPollSelectorImpl.updateSelectedKeys()`.

Issue #5578, opened 2015-06-24
3.5 performance improvement: submitToKeyOwner deserializes request twice.


oobles:
While doing some investigation into an issue we have with IExecutor.submitToKeyOwner I notice that the request is deserialized twice for a single call (we are using custom serialization which could be the cause).  These a two stack traces showing the call stacks where deserialization occurs of the request twice.  This has the effect of not allowing ObjectPool patterns for the object we deserialize because the first object is discarded and can't be returned to the ObjectPool.  The second is that it performs work that is thrown away.  In instances where the object is large or deep, this could improve performance considerably.

Here's the two stack traces where the object is read from the same call.

First time object is deserialized:
	at com.hazelcast.nio.serialization.StreamSerializerAdapter.read(StreamSerializerAdapter.java:41)
	at com.hazelcast.nio.serialization.SerializationServiceImpl.readObject(SerializationServiceImpl.java:325)
	at com.hazelcast.nio.serialization.ByteArrayObjectDataInput.readObject(ByteArrayObjectDataInput.java:489)
	at com.hazelcast.executor.impl.client.PartitionTargetCallableRequest.read(PartitionTargetCallableRequest.java:120)
	at com.hazelcast.client.impl.client.ClientRequest.readPortable(ClientRequest.java:97)
	at com.hazelcast.nio.serialization.PortableSerializer.read(PortableSerializer.java:84)
	at com.hazelcast.nio.serialization.PortableSerializer.read(PortableSerializer.java:73)
	at com.hazelcast.nio.serialization.PortableSerializer.read(PortableSerializer.java:29)
	at com.hazelcast.nio.serialization.StreamSerializerAdapter.read(StreamSerializerAdapter.java:41)
	at com.hazelcast.nio.serialization.SerializationServiceImpl.toObject(SerializationServiceImpl.java:276)
	at com.hazelcast.client.impl.ClientEngineImpl$ClientPacketProcessor.loadRequest(ClientEngineImpl.java:391)
	at com.hazelcast.client.impl.ClientEngineImpl$ClientPacketProcessor.run(ClientEngineImpl.java:366)
	at com.hazelcast.spi.impl.operationservice.impl.OperationRunnerImpl.run(OperationRunnerImpl.java:100)
	at com.hazelcast.spi.impl.operationexecutor.classic.OperationThread.processPartitionSpecificRunnable(OperationThread.java:130)
[...]

PR #5583, opened 2015-06-24
Double serialization fix for new client executor-service

gurbuzali:
moved deserialization/serialization into securityContext check for executor-service message-tasks
fixes for new-client https://github.com/hazelcast/hazelcast/issues/5578


Donnerbart:
I have no clue about the `SecurityContext`, but the change looks reasonable and tested. Nice catch! If there are no objections from people who now more about the details I would give it a thumbs up :)

sancar:
:+1:

PR #5584, opened 2015-06-24
code generator dependency moved to hazelcast.pom and annotation processer listed in mvn.compiler.plugin

sancar:
This fix uses maven-compiler plugin configuration for annotation-processor instead of java service discovery mechanism (via META-INF/services). That way, only hazelcast module will be effected by annotation processor. As a result, hazelcast module will only have provided dependency on code-generator instead of compile scope.

Complements work done by https://github.com/hazelcast/hazelcast/pull/5542

### DIFF
--- a/hazelcast/pom.xml
+++ b/hazelcast/pom.xml
@@ -39,6 +39,28 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


jerrinot:
I can confirm it's working. it's build-able again. 
:+1:  (I'm still not sure about all possible side-effects (will this work in various IDE properly?), but at least it's buildable from a command-line again.

PR #5587, opened 2015-06-25
Removed not needed super() calls.

pveentjer:
This is automatically generated by the javac anyway. No need to do it manually.


pveentjer:
verify

pveentjer:
Since it isn't a functional code change; but just removal of redundant code, I'll merge this one myself.

Issue #5301, opened 2015-05-13
Lazy deserialize while executing entry-processor


gurbuzali:
Currently key/value are deserialized prior to execution of entry-processor by default
This leads to need to the domain object at server-side, even if entry-processor never uses it

mesutcelik:
Can we also check that if following methods are requiring domain objects in the classpath?

* ClienMapProxy.keySet(Predicate predicate)
* ClienMapProxy.entrySet(Predicate predicate)
* ClientMapProxy. executeOnEntries(EntryProcessor entryProcessor)

@sancar , 
I think you can comment on this as well. I know you already checked and have some findings.

PR #5589, opened 2015-06-25
Introduces LazyMapEntry. 

ahmetmircik:
It provides on-demand deserialization of key-value pairs for map entry which is passed in `EntryProcessor#process` & `EntryBackupProcessor#processBackup` methods.

closes https://github.com/hazelcast/hazelcast/issues/5301
closes https://github.com/hazelcast/hazelcast/issues/5611

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.map.impl;
+
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.nio.serialization.SerializationService;
+
+import java.util.Map;
+
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
+/**
+ * A {@link java.util.Map.Entry Map.Entry} implementation which serializes/de-serializes key and value objects on demand.
+ * It is beneficial when you need to prevent unneeded serialization/de-serialization
+ * when creating a {@link java.util.Map.Entry Map.Entry}. Mainly targeted to supply a lazy entry to
+ * {@link com.hazelcast.map.EntryProcessor#process(Map.Entry)} and
+ * {@link com.hazelcast.map.EntryBackupProcessor#processBackup(Map.Entry)}} methods.
+ * <p/>
+ * <STRONG>Note that this implementation is not synchronized and is not thread-safe.</STRONG>
+ *
+ * @see com.hazelcast.map.impl.operation.EntryOperation#createMapEntry(Data, Object)
+ */
+public class LazyMapEntry implements Map.Entry {
+
+    private Object keyObject;
+    private Object valueObject;
+    private transient Data keyData;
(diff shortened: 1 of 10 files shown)


gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
```diff
@@ -0,0 +1,170 @@
...
+ * A {@link java.util.Map.Entry Map.Entry} implementation which serializes/de-serializes key and value objects on demand.
+ * It is beneficial when you need to prevent unneeded serialization/de-serialization
+ * when creating a {@link java.util.Map.Entry Map.Entry}. Mainly targeted to supply a lazy entry to
+ * {@link com.hazelcast.map.EntryProcessor#process(Map.Entry)} method.
+ * <p/>
+ * <STRONG>Note that this implementation is not synchronized and is not thread-safe.</STRONG>
+ *
+ * @see com.hazelcast.map.impl.operation.EntryOperation#createMapEntry(Data, Object)
+ */
+public class LazyMapEntry implements Map.Entry, Serializable {
```
mdogan: Why `Serializable`?

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
```diff
@@ -0,0 +1,170 @@
...
+        return (getKey() == null ? 0 : getKey().hashCode())
+                ^ (getValue() == null ? 0 : getValue().hashCode());
+    }
+
+    @Override
+    public String toString() {
+        return getKey() + "=" + getValue();
+    }
+
+    private void writeObject(java.io.ObjectOutputStream out)
```
mdogan: Why do you need to override serialization methods?

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return keyObject;
+    }
+
+
+    @Override
+    public Object getValue() {
+        // It is required to check null serializationService because
+        // after subsequent serialization and deserialization of a LazyMapEntry instance
+        // serializationService will be null.
```
mdogan: I didn't understand the comment here, how `serializationService`will be null after serialization and deserialization?

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
ahmetmircik: Previous class was `MapEntrySimple` and it was `Serializable`, preserved same behaviour.

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
ahmetmircik: `EntryProcessor#process(entry)` may return `LazyMapEntry` instance as a response to user but there is a possibility that he did not trigger on-demand deserialization, means not called `LazyMapEntry#getKey` or `LazyMapEntry#getValue`, in that case, `LazyMapEntry#keyObject` will be null but actually it may not be null. Trying to prevent that anomaly by overriding them.

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
ahmetmircik: We are not injecting `serializationService` after deserialization of `LazyMapEntry`, this entry can be used by user. Specifically this is required when one explicitly sets `LazyMapEntry#valueObject` to `null` and return `LazyMapEntry` as a response from `Entryprocessor#process` method. 
Or can we inject `serializationService` out-of-the-box?

mdogan: ah, got it.

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
mdogan: Since we don't make any guarantee/promise about serializability of `Map.Entry` given to the `EntryProcessor`, I think we can just drop `Serializable` to make this class simpler.

ahmetmircik: done and created an issue to document it https://github.com/hazelcast/hazelcast/issues/5611

PR #5599, opened 2015-06-26
subList method now returns an unmodifiable list

### LINKED ISSUE #5508: IList.iterator() / listIterator() does not support iterator.remove()
Whereas IQueue.iterator() returns an iterator which explicitly throws an UnsupportedOperationException on iterator.remove(), in case of IList this operation is simply a NOOP.
One can call iterator.remove(), but it doesn't do anything.
A look in the code (ListProxyImpl) reveals, that the iterator is retrieved from a copy of the list, which explains the situation. Nevertheless here an UnsupportedOperationException would be more clear than simpy to ignore the call silently.
Of course, the "real" remove() functionality would be much better :)

My Hazelcast version: 3.4.2
Example code to reproduce:
		HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance();
		IList<String> list = hazelcastInstance.getList("iterate");
		list.add("the entry");
		Iterator<String> iterator = list.iterator();
		iterator.next();
		iterator.remove();
		System.out.println("list size should be 0, actual size:" + list.size());
		hazelcastInstance.shutdown();

Relevant result on console:
list size should be 0, actual size:1

Since iterator() calls listIterator() and that one calls listIterator(index), the issue is for all iterator-retrieving methods.

gurbuzali:
fixes #5508

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListProxyImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListProxyImpl.java
@@ -29,8 +29,10 @@ import com.hazelcast.core.IList;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.impl.SerializableCollection;
+
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListProxyImpl.java
```diff
@@ -130,15 +132,15 @@ public boolean addAll(int index, Collection<? extends E> c) {
 
     @Override
     public List<E> subList(int fromIndex, int toIndex) {
-        final ListSubOperation operation = new ListSubOperation(name, fromIndex, toIndex);
-        final SerializableCollection result = invoke(operation);
-        final Collection<Data> collection = result.getCollection();
-        final List<E> list = new ArrayList<E>(collection.size());
+        ListSubOperation operation = new ListSubOperation(name, fromIndex, toIndex);
+        SerializableCollection result = invoke(operation);
+        Collection<Data> collection = result.getCollection();
```
pveentjer: Isn't this a bit wastful? For you get back a collection, then you copied it into an ArrayList.

This generates a lot of waste.

pveentjer:
Minor review comment.

hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListProxyImpl.java
noctarius: This is what I removed for the last performance improvements. We might just want to create a view wrapper without copying the content whereas in this case it is not as bad since we don't copy hashmaps / hashsets which are pretty bad due to re-hashing.

metanet:
:+1:

PR #5601, opened 2015-06-26
Lock resource cleanup on member left

mdogan:
- Lock cleanup task submitted when a member is left should be executed on related partition threads, otherwise it can miss previously enqueued or being executed lock operations. So, this race can cause infinitely hanging locks.

- Lock cleanup operation should check owner-uuid during its execution in partition thread to avoid race.

Extracted from #5246

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java
@@ -179,14 +179,14 @@ public final class LockServiceImpl implements LockService, ManagedService, Remot
     public void memberRemoved(MembershipServiceEvent event) {
         final MemberImpl member = event.getMember();
         final String uuid = member.getUuid();
-        releaseLocksOf(uuid);
+        releaseLocksOwnedBy(uuid);
     }
 
     @Override
     public void memberAttributeChanged(MemberAttributeServiceEvent event) {
     }
 
-    private void releaseLocksOf(final String uuid) {
+    private void releaseLocksOwnedBy(final String uuid) {
         final InternalOperationService operationService = (InternalOperationService) nodeEngine.getOperationService();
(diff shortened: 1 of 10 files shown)


hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java
```diff
@@ -183,34 +186,47 @@ public void memberRemoved(MembershipServiceEvent event) {
     public void memberAttributeChanged(MemberAttributeServiceEvent event) {
     }
 
-    private void releaseLocksOf(String uuid) {
-        for (LockStoreContainer container : containers) {
-            for (LockStoreImpl lockStore : container.getLockStores()) {
-                releaseLock(uuid, container, lockStore);
-            }
+    private void releaseLocksOf(final String uuid) {
```
pveentjer: perhaps releaseLocksFrom? or releaseLockedOwnedBy?

pveentjer:
I don't see any tests added. We need to have some in place to verify if the behavior works correctly.

pveentjer:
If you can add some tests, I'm fine with merging it.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java
mdogan: ok, renaming to `releaseLocksOwnedBy`.

mdogan:
> I don't see any tests added. We need to have some in place to verify if the behavior works correctly.

Problem is, this is to fix a race between cleanup operation and lock operation. I'll try to write a stress test to create the issue but I'm not sure if I ever will be able to.

mdogan:
Finally I'm able to reproduce the race with a custom lock operation: dfdc4af45f7558096839a1b666f35fba5c0c49c4

mdogan:
verify

mdogan:
verify

Issue #5570, opened 2015-06-23
[TEST-FAILURE] ClientServiceTest.testConnectedClientsWithReAuth


bilalyasar:
```
com.hazelcast.core.HazelcastInstanceNotActiveException: Hazelcast instance is not active!
	at com.hazelcast.client.impl.HazelcastClientProxy.getClient(HazelcastClientProxy.java:280)
	at com.hazelcast.client.impl.HazelcastClientProxy.getMap(HazelcastClientProxy.java:113)
	at com.hazelcast.client.ClientServiceTest.testConnectedClientsWithReAuth(ClientServiceTest.java:258)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast-client/578/testReport/junit/com.hazelcast.client/ClientServiceTest/testConnectedClientsWithReAuth/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK6/com.hazelcast$hazelcast-client-new/582/testReport/junit/com.hazelcast.client/ClientServiceTest/testConnectedClientsWithReAuth/

mdogan:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast-client/14222/testReport/junit/com.hazelcast.client/ClientServiceTest/testConnectedClientsWithReAuth/

mdogan:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/14220/com.hazelcast$hazelcast-client/testReport/com.hazelcast.client/ClientServiceTest/testConnectedClientsWithReAuth/

mdogan:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast-client/14219/testReport/junit/com.hazelcast.client/ClientServiceTest/testConnectedClientsWithReAuth/

PR #5602, opened 2015-06-26
Makes lock operation retries idempotent

mdogan:
Added "reference-call-id" to lock operation to provide operation retry idempotence. When a member crashes after processing a lock request and sending its backup, caller will retry the lock operation. If it happens that backup member already processed the lock request, then caller will acquire lock resource 2nd time (because it's reentrant) unintentionally. When it releases the lock only once, lock will hang infinitely. To avoid that, now lock operation and lock service always keep the original/reference call-id for a specific lock resource and if it's tried to be acquired multiple times with the same reference-call-id, lock request will just return success instead of processing it again.

PS: _note that client side requests are not covered here, clients have their own retry logic._

Depends on #5601
Extracted from #5246

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
@@ -182,7 +182,11 @@ public abstract class Operation implements DataSerializable {
     // Accessed using OperationAccessor
     final Operation setCallId(long callId) {
         this.callId = callId;
+        onSetCallId();
         return this;
+    }
+
+    protected void onSetCallId() {
     }
 
     public boolean validatesTarget() {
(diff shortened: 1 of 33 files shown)


mdogan:
failed https://github.com/hazelcast/hazelcast/issues/5570

verify

mdogan:
verify

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -182,9 +182,13 @@ public final long getCallId() {
     // Accessed using OperationAccessor
     final Operation setCallId(long callId) {
         this.callId = callId;
+        onSetCallId();
         return this;
     }
 
+    protected void onSetCallId() {
```
pveentjer: Can you add a line of comment what the purpose is of this method and when it is expected to be called?

pveentjer:
Apart from that:

:+1:

jerrinot:
I believe this PR introduced a subtle bug: It does not distinguish between lockReferenceId and unlockReferenceId. 

When using transactionalMap both lock/unlock operations might have the same callId/referenceId -> the unlock on commit will fail.

PR #5603, opened 2015-06-26
Introduced lock max lease time

mdogan:
- Introduced max-lease-time (default is Long.MAX_VALUE seconds) for locks. When lock is acquired without an explicit lease-time then max-lease-time setting will be used. Caller can use a lower value but not a greater one. When a greater lease-time is given, invocation/operation will fail immediately.

- Fixed a bug which lease-time is not set correctly on backups and when backup is promoted, lease handling process doesn't kicks in.

Depends on #5601 and #5602
Extracted from #5246


gurbuzali:
verify

gurbuzali:
:+1:

metanet:
:+1:

PR #5606, opened 2015-06-26
InternalPartitionService reject non master partition-state

mdogan:
…s sent from a non-master member, as it's doing for non-member right now.

Extracted from #5247


metanet:
:+1:

pveentjer:
:+1:

PR #5607, opened 2015-06-26
Fixes/Improvements in ClusterServiceImpl.merge()

mdogan:
- During split-brain merge process "ClusterServiceImpl.merge()", cluster state should be reset, all connections should be closed and all invocations from current member should be notified to retry. Then after gathering merge-tasks from related services, all services also should be reset. After that connection manager will be started and node will join target cluster. This will make invocations race free during merge process.

- Improved (TCPIP)ConnectionManager to support start and stop instead of one-shot restart and added some javadocs.

Extracted from #5247

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -902,48 +902,8 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
     public void merge(Address newTargetAddress) {
         if (preparingToMerge.compareAndSet(true, false)) {
             node.getJoiner().setTargetAddress(newTargetAddress);
-            final LifecycleServiceImpl lifecycleService = node.hazelcastInstance.getLifecycleService();
-            lifecycleService.runUnderLifecycleLock(new Runnable() {
-                public void run() {
-                    lifecycleService.fireLifecycleEvent(MERGING);
-                    final NodeEngineImpl nodeEngine = node.nodeEngine;
-                    final Collection<SplitBrainHandlerService> services = nodeEngine
-                            .getServices(SplitBrainHandlerService.class);
-                    final Collection<Runnable> tasks = new LinkedList<Runnable>();
-                    for (SplitBrainHandlerService service : services) {
-                        final Runnable runnable = service.prepareMergeRunnable();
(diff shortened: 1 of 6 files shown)


pveentjer:
There are not tests added so this can't be merged.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -420,7 +420,8 @@ public void shutdown(final boolean terminate) {
         }
     }
 
-    public void onRestart() {
+    public void reset() {
```
pveentjer: Can you add javadoc what the purpose is of 'reset'. Else it will not be maintainable by other engineers.

hazelcast/src/main/java/com/hazelcast/nio/ConnectionManager.java
```diff
@@ -39,11 +39,32 @@
...
+     * @throws IllegalStateException if ConnectionManager is shutdown
+     */
     void start();
 
-    void restart();
+    /**
+     * Stops ConnectionManager, releases its resources, stops threads etc.
+     * When stopped ConnectionManager can be started again using {@link #start()}.
+     *
+     * This method has no effect if it's already stopped or shutdown.
```
pveentjer: Can you add a few lines of comments what the purpose is of the stop. Apparently it is needed for the merging process.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -894,25 +894,45 @@ public void merge(Address newTargetAddress) {
             lifecycleService.runUnderLifecycleLock(new Runnable() {
```
pveentjer: Can you make this a normal inner class. This method has gotten too long to deal with.

hazelcast/src/main/java/com/hazelcast/nio/ConnectionManager.java
```diff
@@ -39,11 +39,32 @@
...
+     *
+     */
+    void stop();
+
+    /**
+     * Shutdowns ConnectionManager completely. ConnectionManager won't be operational anymore and
+     * cannot be started back.
```
pveentjer: cannot be restarted.

mdogan:
verify

mdogan:
verify

PR #5609, opened 2015-06-26
Split brain fixes/enhancements

mdogan:
- Only master node should answer to the split-brain merge requests. Other members should silently ignore the request and optionally log it.

- Non-master members should send master confirmation as soon as possible after they detect a master change. This is just to speed up and help detection of possible network split. Since if a master member receives a master confirmation from a non-member endpoint, it sends a "remove-me-from-your-cluster" request to the caller.

- Master confirmations should be reset only when a node becomes master after old master disconnects.

- Member UUID should not change when it's merging to a cluster, since this makes very hard to detect a member's uuid for uuid-sensitive operations like locks.

- Split brain handling process improved and changed significantly.  Instead of just comparing cluster member count between clusters, complete member-lists will be compared, if at least one member is common between two clusters, clusters will reject to merge. They will not try to merge until they have completely different members in their cluster.  Eventually both cluster will diverge because of heartbeat and/or master-confirmation timeouts.

- Heartbeat mechanism is changed to explicitly send heartbeats instead of relying read and write timestamps of other members. Since, it can be that other node is replying with an error response (like caller-not-member-exception), but currently caller node assumes it's heartbeating implicitly. And this causes one side to assume they are in the same cluster but practically it cannot execute operations on other member.

**Depends on #5607, should be merged after #5607**
Extracted from #5247


mdogan:
failed #5570

verify

mdogan:
verify

pveentjer:
Can you rebase?

PR #5610, opened 2015-06-26
Heartbeat improvements

mdogan:
- Heartbeat and master-confirmation messages, now include a timestamp to prevent a very old message (older than timeout/2) to be processed.

- Master time is updated with every heartbeat message, that provides a better syncing cluster clock.

**Depends on both #5607 and #5609, should be merged after them**
Extracted from #5247


eminn:
:+1:

pveentjer:
The problem with relying on a simple-clock for consensus is that it remains fragile.

mdogan:
Of course. No everybody is claiming otherwise. This is an improvement version of current system, makes it less fragile. But repeating the already known truth every time is just a noise. It doesn't provide any value.

On the other side, my task here is preparing/submitting a PR for the result of my previous work. If people don't like this, it's not merged and closed. I do not care that much.

pveentjer:
Can you explain what the improvement is that makes it less fragile. There are 27 files changed in a single PR without a clear explanation what the actual improvement is. You mention that a timestamp is added to deal with old messages, but that isn't sufficient for me to determine:
- what the actual problem is
- how changes resolve the problem
- the consequences of this fix (could it be that new problems are introduced)

mdogan:
If you read the description, you'll see there are two more PRs which this one depends on. Each PR has clear explanation of what it changes. There's also a wiki page that explains the problems and suggested solutions. I'll add no more.

mdogan:
Because this PR depends on previous two, it includes their changes too. So, you should read then in order.

metanet:
:+1:

mdogan:
failed #5570

verify

mdogan:
verify

PR #5612, opened 2015-06-26
hazelcast web manager doc updated

bilalyasar:


bilalyasar:
@Serdaro , these changes also should be in maintenance branch, do i need to send new PR? or can you add them?

Serdaro:
No problem, I will be adding it to the maintenance branch.

PR #5615, opened 2015-06-26
Improved ringbuffer docs

pveentjer:


hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
 ## RingBuffer
 
+The RingBuffer is a data structure where the data is stored in a ring-like structure. You can think of it as a circular array with a
+certain capacity. Each element in a RingBuffer can be accessed using a sequence ID. This ID is between the head and tail (inclusive) 
+of the RingBuffer. Each ringbuffer has a tail and a head. The tail is where the items are added and the head is where items get over
+written or get expired.
```
noctarius: You might want to add a picture how a ringbuffer works, the current explanation may not work for people with no idea what a RB is at all.

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
 ## RingBuffer
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
 
+The RingBuffer is a data structure where the data is stored in a ring-like structure. You can think of it as a circular array with a
+certain capacity. Each element in a RingBuffer can be accessed using a sequence ID. This ID is between the head and tail (inclusive)
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+```
 
-The RingBuffer can sometimes be a better alternative than an IQueue.  Unlike IQueue, the RingBuffer does not remove the items, it only reads the items using a certain position. There are many advantages using this approach:
+The add method returns the sequence of the inserted item and this value will always be unique. This can sometimes be used for a 
+very cheap way of generating unique id's if you are already making use of the Ringbuffer.
 
-* The same item can be read multiple times by the same thread; this is useful for realizing semantics of read at least once or read at most once.
-* The same item can be read by multiple threads. Normally you could use a IQueue per thread for the same semantic, but this is way less efficient.
+### IQueue vs Ringbuffer.
+The RingBuffer can sometimes be a better alternative than an IQueue. Unlike IQueue, the RingBuffer does not remove the items, it only
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+reads the items using a certain position. There are many advantages using this approach:
+* The same item can be read multiple times by the same thread; this is useful for realizing semantics of read-at-least-once or 
+read-at-most-once.
+* The same item can be read by multiple threads. Normally you could use a IQueue per thread for the same semantic, but this is a 
+lot less efficient because of increased remoting. A take from an IQueue is destructive, so the change needs to be replicated.
 * Reads are extremely cheap since there is no change in the RingBuffer, therefore no replication is required.
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+The RingBuffer can sometimes be a better alternative than an IQueue. Unlike IQueue, the RingBuffer does not remove the items, it only
+reads the items using a certain position. There are many advantages using this approach:
+* The same item can be read multiple times by the same thread; this is useful for realizing semantics of read-at-least-once or 
+read-at-most-once.
+* The same item can be read by multiple threads. Normally you could use a IQueue per thread for the same semantic, but this is a 
+lot less efficient because of increased remoting. A take from an IQueue is destructive, so the change needs to be replicated.
 * Reads are extremely cheap since there is no change in the RingBuffer, therefore no replication is required. 
-* Reads can be batched to speed up performance. Using read (and write) batching can dramatically improve performance of the RingBuffer.
+* Reads and writes can be batched to speed up performance. Using read (and write) batching can dramatically improve performance 
+of the RingBuffer.
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
 
+### Capacity
+A Ringbuffer is by default configured with a capacity of 10000 items. In a lot of cases you want to change this number to something
```
noctarius: "By default a Ringbuffer is configured ..."

hazelcast-documentation/src/RingBuffer.md
pveentjer: That is a good idea but currently don't have the time for that.

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+that fits your needs better. 
 
-The following are the methods included in the RingBuffer interface.
+Below is a code example of a ringbuffer with a capacity of 2000 items.
+```
+<ringbuffer name="rb">
+    <capacity>2000</capacity>
+</ringbuffer>
+```
+For now, the RingBuffer is not a partitioned data structure; its data is stored in a single partition and the replicas are
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
 
+The RingBuffer is a data structure where the data is stored in a ring-like structure. You can think of it as a circular array with a
+certain capacity. Each element in a RingBuffer can be accessed using a sequence ID. This ID is between the head and tail (inclusive) 
+of the RingBuffer. Each ringbuffer has a tail and a head. The tail is where the items are added and the head is where items get over
+written or get expired.
 
-The RingBuffer is a data structure where the data is stored in a ring-like structure. You can think of it as a circular array with a certain capacity. In this circular array, the oldest item gets overwritten in case a new item is written when the maximum capacity is reached. For now, the RingBuffer is not a partitioned data structure; its data is stored in a single partition and the replicas are stored in another partition.
+Reading from the ringbuffer is very simple. Just get the current head and start reading. The 'readOne' returns the item at the 
+given sequence or blocks if no item is available. To read the next item, the sequence is incremented by one.
+```
```
noctarius:
```java

File: hazelcast-documentation/src/RingBuffer.md [hunk 10]
```diff
@@ -1,47 +1,233 @@
...
+    sequence++;
+    ... process item
+}  
+```
+By exposing the sequence you can now move any item from the ringbuffer as long as the item is still available. If it isn't available
+anymore, a StaleSequenceException is thrown.
 
-Each element in a RingBuffer can be accessed using a sequence ID. This ID is between the head and tail (inclusive) of the RingBuffer. Head is the side where items are discarded and tail is the side where items are added to.
+Adding an item to the ringbuffer is also very easy:
+```
```
noctarius:
```java

File: hazelcast-documentation/src/RingBuffer.md [hunk 11]
```diff
@@ -1,47 +1,233 @@
...
+of the RingBuffer.
 
+### Capacity
+A Ringbuffer is by default configured with a capacity of 10000 items. In a lot of cases you want to change this number to something
+that fits your needs better. 
 
-The following are the methods included in the RingBuffer interface.
+Below is a code example of a ringbuffer with a capacity of 2000 items.
+```
+<ringbuffer name="rb">
```
noctarius:
```java

File: hazelcast-documentation/src/RingBuffer.md [hunk 12]
```diff
@@ -1,47 +1,233 @@
...
-  ICompletableFuture<ReadResultSet<E>> readManyAsync(long startSequence, 
-                         int minCount, int maxCount, 
-                         IFunction<E, Boolean> filter);
+### Synchronous and Asynchronous backups.
+A Ringbuffer by default has a single synchronous backup. But this can be controlled, just like most of the other Hazelcast 
+distributed data-structures by setting the sync and async backups. In the example below a ringbuffer is configured with 0
+sync backups and 1 async backup:
+
+```
+<ringbuffer name="rb">
```
noctarius:
```xml

File: hazelcast-documentation/src/RingBuffer.md [hunk 13]
```diff
@@ -1,47 +1,233 @@
...
+### Time to live.
+The RingBuffer can be configured with a time to live in seconds. Using this setting, you can control how long the items remain in 
+the RingBuffer before they are expired. By default, the time to live is set to 0, meaning that unless the item is overwritten, 
+it will remain in the RingBuffer indefinitely. If a time to live is set and an item is added, then depending on the OverwritePolicy, 
+either the oldest item is overwritten, or the call is rejected. 
 
-The RingBuffer can be configured with a time to live in seconds. Using this setting, you can control how long the items remain in the RingBuffer before they are deleted. By default, the time to live is set to 0, meaning that unless the item is overwritten, it will remain in the RingBuffer indefinitely. If a time to live is set and an item is added, then depending on the OverwritePolicy, either the oldest item is overwritten, or the call is rejected. 
+In the code fragment below a ringbuffer is configured with a time to live of 180 seconds.
+```
+<ringbuffer name="rb">
```
noctarius:
```xml

File: hazelcast-documentation/src/RingBuffer.md [hunk 14]
```diff
@@ -1,47 +1,233 @@
...
+distributed data-structures by setting the sync and async backups. In the example below a ringbuffer is configured with 0
+sync backups and 1 async backup:
+
+```
+<ringbuffer name="rb">
+    <backup-count>0</backup-count>
+    <async-backup-count>1</async-backup-count>
+</ringbuffer>
 ```
+An async backup will probably give you better performance, but there is a chance that the chance that the item added is lost
```
noctarius: Remove 2nd "the chance" :)

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+    <backup-count>0</backup-count>
+    <async-backup-count>1</async-backup-count>
+</ringbuffer>
 ```
+An async backup will probably give you better performance, but there is a chance that the chance that the item added is lost 
+when the member owning the primary crashed before the replication has completed. Probably you want to have a look at the batching
+methods if you need high performance but don't want to give up on consistency.
 
+### Time to live.
+The RingBuffer can be configured with a time to live in seconds. Using this setting, you can control how long the items remain in
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+    <async-backup-count>1</async-backup-count>
+</ringbuffer>
 ```
+An async backup will probably give you better performance, but there is a chance that the chance that the item added is lost 
+when the member owning the primary crashed before the replication has completed. Probably you want to have a look at the batching
+methods if you need high performance but don't want to give up on consistency.
 
+### Time to live.
+The RingBuffer can be configured with a time to live in seconds. Using this setting, you can control how long the items remain in 
+the RingBuffer before they are expired. By default, the time to live is set to 0, meaning that unless the item is overwritten,
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+</ringbuffer>
 ```
+An async backup will probably give you better performance, but there is a chance that the chance that the item added is lost 
+when the member owning the primary crashed before the replication has completed. Probably you want to have a look at the batching
+methods if you need high performance but don't want to give up on consistency.
 
+### Time to live.
+The RingBuffer can be configured with a time to live in seconds. Using this setting, you can control how long the items remain in 
+the RingBuffer before they are expired. By default, the time to live is set to 0, meaning that unless the item is overwritten, 
+it will remain in the RingBuffer indefinitely. If a time to live is set and an item is added, then depending on the OverwritePolicy,
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+<ringbuffer name="rb">
+    <time-to-live-seconds>180</time-to-live-seconds>
+</ringbuffer>
+```
 
-The RingBuffer can also be configured with an InMemoryFormat which controls the format of stored items. By default, `BINARY` is used, meaning that the object is stored in a serialized form. You can select the `OBJECT` InMemoryFormat, which is useful when filtering is applied or when the `OBJECT` InMemoryFormat can lead to a smaller memory footprint than a `BINARY`. 
+### Overflow policy
+using the overflow policy one can determine what to do if the oldest item in the ringbuffer is not old enough to expire. There
+are currently 2 options available:
+* OverflowPolicy.OVERWRITE: in this case the oldest item is overwritten. You might wonder why not set the time to live to 0, since
```
noctarius: I don't understand that point - non of the sentences. Sounds like you got some changes into that don't fit together :)

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+if there is enough capacity in the ringbuffer.
+* OverflowPolicy.FAIL: the call is aborted. The methods that make use of the OverflowPolicy return -1 to indicate that adding
+the item has failed. 
 
-The RingBuffer supports filtered reads. For example, when one thread only wants to see certain messages, one can filter the items after they are received from the RingBuffer. The problem is that this approach can be very inefficient since a lot of useless data needs to be sent over the line. When a filter is used, then the filtering happens at the source, which is a lot more efficient.
+Using the the overflow policy gives fine control on what to do if the Ringbuffer is full. The policy can also be used to making 
+a back pressure mechanism. Below a code example can be found where an exponential backoff is used.
 
-The RingBuffer provides asynchronous methods for the more powerful methods like batched reading with filtering or batch writing. To make these methods synchronous, just call `get()` on the returned future.
+```
```
noctarius:
```java

File: hazelcast-documentation/src/RingBuffer.md [hunk 20]
```diff
@@ -1,47 +1,233 @@
...
+        break;
+    }
+    
+    TimeUnit.MILLISECONDS.sleep(sleepMs);
+    sleepMs = min(5000, sleepMs * 2);
+}
+```
+
+### In Memory Format
+The RingBuffer can also be configured with an InMemoryFormat which controls the format of stored items. By default, `BINARY` is used,
```
noctarius: Ringbuffer instead of RingBuffer

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+```
+
+### In Memory Format
+The RingBuffer can also be configured with an InMemoryFormat which controls the format of stored items. By default, `BINARY` is used, 
+meaning that the object is stored in a serialized form. You can select the `OBJECT` InMemoryFormat, which is useful when filtering is 
+applied or when the `OBJECT` InMemoryFormat can lead to a smaller memory footprint than a `BINARY`. 
+
+In the example below a ringbuffer is configured with a OBJECT In Memory Format:
+
+```
```
noctarius:
```xml

File: hazelcast-documentation/src/RingBuffer.md [hunk 22]
```diff
@@ -1,47 +1,233 @@
...
+<ringbuffer name="rb">
+    <in-memory-format>BINARY</in-memory-format>
+</ringbuffer>
+```
+
+### addAllAsync
+
+In the previous examples the Ringbuffer.add method was being used to add an item to the ringbuffer. The problem with the add
```
noctarius: `Ringbuffer::add`

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+is that it always overwrites and that it doesn't support batching. Batching can have a huge impact on performance. That is why the
+addAllAsync method was added.
```
noctarius: `addAllAsync`

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,47 +1,233 @@
...
+### addAllAsync
+
+In the previous examples the Ringbuffer.add method was being used to add an item to the ringbuffer. The problem with the add
+is that it always overwrites and that it doesn't support batching. Batching can have a huge impact on performance. That is why the
+addAllAsync method was added. 
 
-For more details about RingBuffer configuration, check the `RingbufferConfig` class.
+Example:
+
+```
```
noctarius:
```java

File: hazelcast-documentation/src/RingBuffer.md [hunk 25]
```diff
@@ -1,49 +1,236 @@
...
+List<String> items = Arrays.asList("1","2","3")
+ICompletableFuture<Long> f = rb.addAllAsync(items, OverflowPolicy.OVERWRITE);
+f.get()
+```        
+In this case the 3 strings are being added to the ringbuffer using the OverflowPolicy.OVERWRITE policy. Check the OverflowPolicy
+for more details.
+
+### readManyAsync
 
+In the previous example the 'readOne' was being used. It is simple, but it isn't very efficient:
```
noctarius: `readOne`

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,49 +1,236 @@
...
+The Ringbuffer provides asynchronous methods for the more powerful methods like batched reading with filtering or batch writing. 
+To make these methods synchronous, just call `get()` on the returned future.
+
+Example:
+```java
+ICompletableFuture f = ringbuffer.addAsync(item, OverflowPolicy.FAIL);
+f.get();
 ```
 
+But the ICompletable future can also be used to get notified when the operation has completed. For example when you want to
```
noctarius: ICompletableFuture

hazelcast-documentation/src/RingBuffer.md
```diff
@@ -1,49 +1,236 @@
...
+   }
+});
+```
+THe advantage of this approach is that the thread that does the call isn't block till the response is returned.
 
-The RingBuffer can be configured with a time to live in seconds. Using this setting, you can control how long the items remain in the RingBuffer before they are deleted. By default, the time to live is set to 0, meaning that unless the item is overwritten, it will remain in the RingBuffer indefinitely. If a time to live is set and an item is added, then depending on the OverwritePolicy, either the oldest item is overwritten, or the call is rejected. 
+### Full Configuration examples
 
-The RingBuffer can also be configured with an InMemoryFormat which controls the format of stored items. By default, `BINARY` is used, meaning that the object is stored in a serialized form. You can select the `OBJECT` InMemoryFormat, which is useful when filtering is applied or when the `OBJECT` InMemoryFormat can lead to a smaller memory footprint than a `BINARY`. 
+A Ringbuffer can be configured in the XML. Below a full example of a ringbuffer called 'rb' can be found. It is configured
```
noctarius: I would add to sub-sections, "declarative configuration" and "programatic configuration"

noctarius: The second sentence should more be like "The following snippet shows the configuration of a ringbuffer called 'rb'. The configuration is modeled after the ringbuffers defaults.

Donnerbart:
In our documentation style guide it says [RingBuffer](https://hazelcast.atlassian.net/wiki/display/CM/Hazelcast+Documentation+Style+Guide#HazelcastDocumentationStyleGuide-StandardizingWordsandTerms), but I think even the class name violates this :D We should have one or the other corrected.

pveentjer:
It is Ringbuffer.

PR #5622, opened 2015-06-29
Introduced UnmodifiableLazyList

gurbuzali:
refactored SerailizableCollection accordingly

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientListProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientListProxy.java
@@ -50,10 +50,10 @@ import com.hazelcast.core.ItemListener;
 import com.hazelcast.core.Member;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.nio.serialization.SerializationService;
+import com.hazelcast.spi.impl.UnmodifiableLazyList;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -142,12 +142,9 @@ public class ClientListProxy<E> extends ClientProxy implements IList<E> {
         ClientMessage request = ListIteratorCodec.encodeRequest(name);
         ClientMessage response = invoke(request);
         ListIteratorCodec.ResponseParameters resultParameters = ListIteratorCodec.decodeResponse(response);
-        Collection<Data> resultCollection = resultParameters.list;
-        final ArrayList<E> list = new ArrayList<E>(resultCollection.size());
-        for (Data value : resultCollection) {
-            list.add((E) toObject(value));
-        }
-        return Collections.unmodifiableCollection(list).iterator();
+        List<Data> resultCollection = (List<Data>) resultParameters.list;
+        SerializationService serializationService = getContext().getSerializationService();
+        return new UnmodifiableLazyList<E>(resultCollection, serializationService).iterator();
     }
 
     public Object[] toArray() {
@@ -246,43 +243,6 @@ public class ClientListProxy<E> extends ClientProxy implements IList<E> {
         return stopListening(request, registrationId);
     }
 
-    private class ItemEventHandler extends ListAddListenerCodec.AbstractEventHandler
-            implements EventHandler<ClientMessage> {
-
-        private final boolean includeValue;
-        private final ItemListener<E> listener;
-
-        public ItemEventHandler(boolean includeValue, ItemListener<E> listener) {
-            this.includeValue = includeValue;
-            this.listener = listener;
-        }
-
-        @Override
-        public void handle(Data dataItem, String uuid, int eventType) {
-            SerializationService serializationService = getContext().getSerializationService();
-            ClientClusterService clusterService = getContext().getClusterService();
-
-            E item = includeValue ? (E) serializationService.toObject(dataItem) : null;
-            Member member = clusterService.getMember(uuid);
-            ItemEvent<E> itemEvent = new ItemEvent<E>(name, ItemEventType.getByType(eventType), item, member);
-            if (eventType == ItemEventType.ADDED.getType()) {
-                listener.itemAdded(itemEvent);
-            } else {
-                listener.itemRemoved(itemEvent);
-            }
-        }
-
-        @Override
-        public void beforeListenerRegister() {
-
-        }
-
-        @Override
-        public void onListenerRegister() {
-
-        }
-    }
-
     protected <T> T invoke(ClientMessage req) {
         return super.invoke(req, getPartitionKey());
     }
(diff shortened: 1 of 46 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
```diff
@@ -0,0 +1,235 @@
...
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+
+/**
+ * TODO add a proper JavaDoc
+ */
+public class UnmodifiableLazyList<E> implements List<E> {
```
pveentjer: Isn't it easier to implement the AbstractList?

pveentjer:
Can you add some tests; especially for the UnmodifiableLazyList. 

And of course; the comment. Perhaps just remove it since it is in the .impl package and the class isn't complicated.

hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
```diff
@@ -0,0 +1,235 @@
...
+    @Override
+    public Object[] toArray() {
+        convertAll();
+        return list.toArray();
+    }
+
+    @Override
+    public <T> T[] toArray(T[] a) {
+        convertAll();
+        return (T[]) list.toArray(a);
```
ahmetmircik: seems a `ClassCastException`?

hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
pveentjer: One addition for documentation: subsequent calls for the same object returns different instances. Every time the value is deserialized again. This is important to document.

hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListProxyImpl.java
```diff
@@ -134,13 +135,9 @@ public boolean addAll(int index, Collection<? extends E> c) {
     public List<E> subList(int fromIndex, int toIndex) {
```
ahmetmircik: maybe it should be better to add this new lazy behavior in javaDocs also.

hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
ahmetmircik: can you add some tests for this class?

hazelcast/src/main/java/com/hazelcast/spi/impl/SerializableCollection.java
```diff
@@ -23,36 +23,26 @@
...
 public final class SerializableCollection implements IdentifiedDataSerializable, Iterable<Data> {
 
-    private Collection<Data> collection;
+    private List<Data> collection;
 
     public SerializableCollection() {
     }
 
-    public SerializableCollection(Collection<Data> collection) {
+    public SerializableCollection(List<Data> collection) {
```
ahmetmircik: can it be named to `SerializableList` instead of `SerializableCollection`?

hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
gurbuzali: actually it is the opposite, when deserilaized first, I replace the serialized version. This way you don't need to deserialize again and you don't get different instances.

PS: I will add more tests and javadoc

hazelcast/src/main/java/com/hazelcast/spi/impl/SerializableCollection.java
gurbuzali: I've thought the same thing but reluctant to do so

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientListProxy.java
```diff
@@ -325,28 +285,59 @@ public int indexOf(Object o) {
...
         ListSubCodec.ResponseParameters resultParameters = ListSubCodec.decodeResponse(response);
-        Collection<Data> resultCollection = resultParameters.list;
-        final List<E> list = new ArrayList<E>(resultCollection.size());
-        for (Data value : resultCollection) {
-            list.add((E) toObject(value));
-        }
-        return list;
+        List<Data> resultCollection = (List<Data>) resultParameters.list;
+        SerializationService serializationService = getContext().getSerializationService();
+        return new UnmodifiableLazyList<E>(resultCollection, serializationService);
```
ahmetmircik: Previous return type `ArrayList` was `Serializable` but new `UnmodifiableLazyList ` is not. May this be a problem @gurbuzali WDYT?

hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
pveentjer: Ok. But than this means that this collection can't be used as a returned value of a future that is exposed to the outside. E.g. map.getAll = fine.. ,map.getAllAsync is not fine.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientListProxy.java
gurbuzali: Yes this is definitely a problem for 3.5.x releases, it shouldn't be a problem for 3.6 with a proper documentation.

hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
gurbuzali: I didn't get the point with async here, can you elaborate ?

hazelcast/src/main/java/com/hazelcast/spi/impl/SerializableCollection.java
```diff
@@ -83,11 +70,11 @@ public int getId() {
 
     @Override
     public Iterator<Data> iterator() {
-        final Iterator<Data> iterator = collection == null ? null : collection.iterator();
+        final Iterator<Data> iterator = collection.iterator();
         return new Iterator<Data>() {
```
ajermakovics: `new UnmodifiableIterator()` from util could be handy here

hazelcast/src/main/java/com/hazelcast/spi/impl/UnmodifiableLazyList.java
pveentjer: pseudo code:

ICompletableFuture<List> f = map.getAllAsync(..)

If this future would be handed over to multiple threads, then they can get a shared instance of your UnmodifiableLazyList. And then you have a race problem on your hands.

gurbuzali: List list = map.getAll(...)
if this list would handed over to multiple threads then this cad lead concurrency problems too.
What's the difference between sync/async ?

pveentjer: Perhaps that is broken too :)

gurbuzali: :)
then a simple `returned implementation is not thread-safe` statement in the JavaDoc/Documentation should be suffice ?

pveentjer: Did you manage to have a look at the AbstractList? Perhaps it can simplify your list implementation dramatically (and reduce your testing efforts).

gurbuzali: yes, I'll use AbstractList. I'm just collecting all review comments

pveentjer: Great, thanks.

gurbuzali:
verify

gurbuzali:
@pveentjer can you take another look please ? I've tried to address all comment-reviews
PS: changed file count is high because of a renaming

pveentjer:
Merging

PR #5623, opened 2015-06-29
ResponseThread should not eat exception

### LINKED ISSUE #5619: ResponseThread should log exception
(Hazelcast 3.5.0)

The com.hazelcast.spi.impl.classicscheduler.ResponseThread::process method may catch throwables. When this occurs, it logs an unhelpful message, and ignores the actual exception. 

This method should be improved so as to additionally log the cause, or at least the exception class and message.

In my case, it was catching a HazelcastSerializationException to do with an overridden DataSerializable class. I could only figure this out from debugging the application.

pveentjer:
but should log the actual cause

Fixes #5619


Donnerbart:
:+1:

pveentjer:
verify

pveentjer:
verify

metanet:
:+1:

PR #5625, opened 2015-06-29
Seperated serialization API from implementation

pveentjer:
All the classes required for implementation have been moved to com.hazelcast.nio.serialization.impl.

This is in line with other packages.

It also makes the Hazelcast serialization API a lot easier to use for end users since all the
implementation details have been moved out of sight.

Remark for the reviewers: no logical changes have been made. Only classes have been moved to the right package + @Override added in some cases.


Donnerbart:
Could you add a screenshot with the file list of the new impl package? Should be easier to see what has been changed by just looking at the final folder structure.

You can directly drag&drop an image into a comment field on GitHub to upload.

pveentjer:
changesserialization

pveentjer:
And the API:

screenshot_2015-06-29_17-37-27

mdogan:
`ClassDefinitionBuilder` may be part of public API. It's even mentioned in docs: http://docs.hazelcast.org/docs/3.5/manual/html-single/hazelcast-documentation.html#null-portable-serialization

Other than that looks good.

mdogan:
Btw, please do not forget to align enterprise repo with these changes.

pveentjer:
Thanks! I'm currently looking at a performance related issue; once that is done I'll update this pr.

pveentjer:
I created an enterprise PR as well:

https://github.com/hazelcast/hazelcast-enterprise/pull/268

mtopolnik:
+1

PR #5626, opened 2015-06-29
Deleted PortableAdapter

pveentjer:
Class is not used.


mdogan:
merging directly since has no side effects.

PR #5627, opened 2015-06-29
AbstractIoSelector + RWHandlers toString

pveentjer:
The toString is needed for logging purposes in the iobalancer. Currently
they show the object.toString, which is not very useful.


hazelcast/src/main/java/com/hazelcast/nio/tcp/AbstractIOSelector.java
```diff
@@ -195,4 +195,9 @@ private void handleSelectFailure(Throwable e) {
             Thread.currentThread().interrupt();
         }
     }
+
+    @Override
+    public String toString(){
```
Donnerbart: whitespace missing between `){`

Donnerbart:
Beside the minor typo which lead to the build failure: :+1:

metanet:
:+1:

PR #5637, opened 2015-07-01
Fixes #5636.

### LINKED ISSUE #5636: DoubleSumCombiner sums on long
Loses precision.

nilskp:
[Review on Reviewable](https://reviewable.io/reviews/hazelcast/hazelcast/5637)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/mapreduce/aggregation/impl/DoubleSumAggregation.java
+++ b/hazelcast/src/main/java/com/hazelcast/mapreduce/aggregation/impl/DoubleSumAggregation.java
@@ -108,7 +108,7 @@ public class DoubleSumAggregation<Key, Value>
     private static final class DoubleSumCombiner
             extends Combiner<Double, Double> {
 
-        private long chunkSum;
+        private double chunkSum;
 
         @Override
         public void combine(Double value) {
--- a/hazelcast/src/main/java/com/hazelcast/mapreduce/impl/task/DefaultContext.java
+++ b/hazelcast/src/main/java/com/hazelcast/mapreduce/impl/task/DefaultContext.java
@@ -102,11 +102,12 @@ public class DefaultContext<KeyIn, ValueIn>
         Combiner<ValueIn, ?> combiner = combiners.get(key);
         if (combiner == null) {
             combiner = combinerFactory.newCombiner(key);
-            combiner.beginCombine();
 
             Combiner<ValueIn, ?> temp = combiners.putIfAbsent(key, combiner);
             if (temp != null) {
                 combiner = temp;
+            } else {
+                combiner.beginCombine();
             }
         }
         return combiner;


nilskp:
Committed another fix, unrelated to the issue. Caused by my poor git-fu skills, but now I know better.

mesutcelik:
Thanks @nilskp ,
Do we have a test for the fix? Can you please also send the backport?

mesutcelik:
@nilskp , any update?

hazelcast/src/main/java/com/hazelcast/mapreduce/impl/task/DefaultContext.java
```diff
@@ -102,11 +102,12 @@ public void finalizeCombiners() {
         Combiner<ValueIn, ?> combiner = combiners.get(key);
         if (combiner == null) {
             combiner = combinerFactory.newCombiner(key);
```
toktarev: It is a good fix - but it would be more efficient to omit combinerFactory.newCombiner(key); it case if key exists.
We can add computeIfAbsent into out HazelCast ConcurrentHashMap SampleableConcurrentHashMap.
And use it in a lot of similar places. It would reduce unnecessary objects creation.

hazelcast/src/main/java/com/hazelcast/mapreduce/aggregation/impl/DoubleSumAggregation.java
```diff
@@ -108,7 +108,7 @@ public int getId() {
     private static final class DoubleSumCombiner
             extends Combiner<Double, Double> {
 
-        private long chunkSum;
```
toktarev: It is good. I am agree.

hazelcast/src/main/java/com/hazelcast/mapreduce/impl/task/DefaultContext.java
toktarev: private Function<KeyIn, Combiner<ValueIn, ?>> combinerFunction = new Function<KeyIn, Combiner<ValueIn, ?>>() {
        @Override
        public Combiner<ValueIn, ?> apply(KeyIn keyIn) {
            Combiner<ValueIn, ?> combiner = combinerFactory.newCombiner(keyIn);
            combiner.beginCombine();
            return combiner;
        }
    };

    public Combiner<ValueIn, ?> getOrCreateCombiner(KeyIn key) {
        return combiners.computeIfAbsent(key,combinerFunction);
    }
It will be like this.

nilskp: That'll work too, provided the implementation is actually able to do this inside the lock.

toktarev: It will work, but computeIfAbsent would let to use it in another places.

Actually I think that if we can write a code without lock it is better to write it without lock.

mesutcelik:
fixed by https://github.com/hazelcast/hazelcast/pull/5921

Issue #5552, opened 2015-06-22
[TEST-FAILURE] IOBalancerStressTest.testEachConnectionUseDifferentSelectorEventually


bilalyasar:
```
java.lang.AssertionError: Values should be different. Actual: Thread[hz._hzInstance_28_dev.IO.thread-in-1,5,_hzInstance_28_dev]
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failEquals(Assert.java:185)
	at org.junit.Assert.assertNotEquals(Assert.java:161)
	at org.junit.Assert.assertNotEquals(Assert.java:175)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast/609/testReport/junit/com.hazelcast.nio.tcp.iobalancer/IOBalancerStressTest/testEachConnectionUseDifferentSelectorEventually/

gurbuzali:
I've manged to reproduce it (just once) with finest logging enabled.
here is a log snippet looks suspicious to me
```
11:19:34,038 DEBUG [IOBalancer] hz._hzInstance_3_dev.IOBalancerThread - [127.0.0.1]:5703 [dev] [3.6-SNAPSHOT] I/O imbalance is detected, but no suitable migration candidate is found.
```

pveentjer:
This test was generously provided by Ali to reproduce the problem:

```
public void testEachConnectionUseDifferentSelectorEventually() {
        Config config = new Config();
        config.setProperty(GroupProperties.PROP_IO_BALANCER_INTERVAL_SECONDS, "1");
        config.setProperty(GroupProperties.PROP_IO_THREAD_COUNT, "2");

        HazelcastInstance instance1 = Hazelcast.newHazelcastInstance(config);
        HazelcastInstance instance2 = Hazelcast.newHazelcastInstance(config);
        HazelcastInstance instance3 = Hazelcast.newHazelcastInstance(config);

        instance2.shutdown();
        instance2 = Hazelcast.newHazelcastInstance(config);

        IMap<Integer, Integer> map = instance1.getMap(randomMapName());

        long deadLine = System.currentTimeMillis() + TEST_DURATION_SECONDS * 1000;
        for (int i = 0; System.currentTimeMillis() < deadLine; i++) {
            map.put(i % 1000, i);
        }

        TcpIpConnectionManager manager1 = (TcpIpConnectionManager) getConnectionManager(instance1);
[...]

PR #5639, opened 2015-07-01
Balancer improvements

pveentjer:
The performance with the io balancer included performance still had dips of 15% once and a while; but performance with these changes is a lot more stable. The cause of the biggest dips is imbalance in the selector load. This is caused by:

1: not all connections were seen by the  IO-Balancer. Some active connections
were not seen, but receive load. Because of these connections being invisible, 
the IO-Balancer doesn't deal with balancing load optimally.

2: if there is 1 selector with a huge number of events, but only a single
handler, this was accepted as a good source selector to move a handler from.
In case of a single handler it doesn't make any sense since you don't want
to move a single handler from an otherwise empty selector.

This is fixed by

1: registering every active connection to the IO-Balancer instead of only the 'registered' ones

2: ignoring selector with a single handler; even though it has a huge number of events

Once merged I'll backport the changes to 3.5 maintenance.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
@@ -144,6 +144,7 @@ public class TcpIpConnectionManager implements ConnectionManager {
         this.loggingService = loggingService;
     }
 
+
     public void interceptSocket(Socket socket, boolean onAccept) throws IOException {
         if (!isSocketInterceptorEnabled()) {
             return;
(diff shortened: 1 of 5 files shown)


gurbuzali:
does this fixes https://github.com/hazelcast/hazelcast/issues/5552 ?

pveentjer:
It could very well be that the issue is fixed since before the fix it can happen that work doesn't get balanced correctly.

pveentjer:
I still got a test failure after may iterations.

pveentjer:
The test is broken :)

I just checked the balancing of the connections over the selectors:

```
in selectors
hz._hzInstance_32_dev.IO.thread-in-0 :84473
	hz._hzInstance_32_dev.IO.thread-in-0 eventCount:84472
	hz._hzInstance_32_dev.IO.thread-in-0 eventCount:1
hz._hzInstance_32_dev.IO.thread-in-1 :23000
	hz._hzInstance_32_dev.IO.thread-in-1 eventCount:23000
out selectors
hz._hzInstance_32_dev.IO.thread-out-0 :2
	hz._hzInstance_32_dev.IO.thread-out-0 eventCount:107237
hz._hzInstance_32_dev.IO.thread-out-1 :2
	hz._hzInstance_32_dev.IO.thread-out-1 eventCount:2
	hz._hzInstance_32_dev.IO.thread-out-1 eventCount:36758
```

So the balancing is fine.. only the wrong connections are being used (due to connection duplication). So I need to have a closer look at the test.

pveentjer:
The test is fixed. @gurbuzali can you have another look at it.

gurbuzali:
:+1:

PR #5658, opened 2015-07-05
Introduced tryLock with lease

gurbuzali:
both new and old client implemented
@sancar can you please check especially the new-client implementation


sancar:
:+1:

eminn:
:+1:

Issue #5551, opened 2015-06-22
Hazelcat 3.5 - InvalidConfigurationException: cvc-complex-type.2.4.a: 


ssozonoff:
I have a configuration working fine with Hazelcast 3.4.3, upon upgrading to 3.5 I get the following XML validation error. The XML itself validates just fine using an external validator so seems something is messed up with your internal validation. Can the validation be disabled ?

Thanks

Caused by: com.hazelcast.config.InvalidConfigurationException: cvc-complex-type.2.4.a: Invalid content was found starting with element 'group'. One of '{"http://www.hazelcast.com/schema/client-config":import, "http://www.hazelcast.com/schema/client-config":group, "http://www.hazelcast.com/schema/client-config":license-key, "http://www.hazelcast.com/schema/client-config":properties, "http://www.hazelcast.com/schema/client-config":network, "http://www.hazelcast.com/schema/client-config":executor-pool-size, "http://www.hazelcast.com/schema/client-config":security, "http://www.hazelcast.com/schema/client-config":listeners, "http://www.hazelcast.com/schema/client-config":serialization, "http://www.hazelcast.com/schema/client-config":native-memory, "http://www.hazelcast.com/schema/client-config":proxy-factories, "http://www.hazelcast.com/schema/client-config":load-balancer, "http://www.hazelcast.com/schema/client-config":near-cache, "http://www.hazelcast.com/schema/client-config":query-caches}' is expected.
	at com.hazelcast.config.AbstractXmlConfigHelper.schemaValidation(AbstractXmlConfigHelper.java:193)
	at com.hazelcast.client.config.XmlClientConfigBuilder.parseAndBuildConfig(XmlClientConfigBuilder.java:183)
	at com.hazelcast.client.config.XmlClientConfigBuilder.build(XmlClientConfigBuilder.java:164)
	at com.hazelcast.client.config.XmlClientConfigBuilder.build(XmlClientConfigBuilder.java:157)
	at com.hazelcast.client.HazelcastClient.newHazelcastClient(HazelcastClient.java:69)
	at com.hazelcast.client.cache.impl.HazelcastClientCachingProvider.instanceFromProperties(HazelcastClientCachingProvider.java:114)
	at com.hazelcast.client.cache.impl.HazelcastClientCachingProvider.createHazelcastCacheManager(HazelcastClientCachingProvider.java:62)
[...]

PR #5659, opened 2015-07-06
fix for #5551

bilalyasar:
closes the issue: https://github.com/hazelcast/hazelcast/issues/5551

after this PR has been merged,
users should add hazelcast namespace to `hazelcast` tag.

`<hazelcast xmlns="http://www.hazelcast.com/schema/config">`
...
...
`</hazelcast>`

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
@@ -125,7 +125,9 @@ public class XmlClientConfigBuilder extends AbstractConfigBuilder {
 
     @Override
     protected Document parse(InputStream inputStream) throws Exception {
-        final DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+        DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+        dbf.setNamespaceAware(true);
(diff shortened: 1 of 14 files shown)


bilalyasar:
verify

bilalyasar:
seems failure doesn't related to this PR.
verify

bilalyasar:
@devOpsHazelcast verify please

bilalyasar:
verify

mesutcelik:
I don't know if that is common practice to include NS in the top level xml tag.

bilalyasar:
i think normally, users should include namespace and uri in their xml files.

mesutcelik:
Do we need to update any documentation regarding this change?

mdogan:
@bilalyasar:
> i think normally, users should include namespace and uri in their xml files.

In our case, is it *should* or a *must*?

bilalyasar:
@mesutcelik : maybe we can give info in documentation config section
@mdogan : they must, it will be obligation..

mesutcelik:
@bilalyasar ,
Can you please check what is the common practice in here on the market? I think we should follow the same path.

bilalyasar:
@mesutcelik , i looked spring and hibernate.. when i remove the namespace from their xml, they don't start and throw exceptions like `invalidConfigurationException`..

bilalyasar:
verify

mesutcelik:
:+1:

PR #5672, opened 2015-07-09
Backport of Agrona collections

mtopolnik:
Contributes a subset of Agrona collections to Hazelcast codebase:

1. `BiInt2ObjectMap`
2. `Int2ObjectHashMap`
3. `IntHashSet`
4. `Long2LongHashMap`
5. `Long2ObjectHashMap`
6. `LongHashSet`

`Int2ObjectHashMap` is already used by the `client` code and `Long2ObjectHashMap` will be used in a feature planned for 3.6. There are probably more target usages around the current codebase where their application would save memory and increase performance.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/Bits.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/Bits.java
@@ -16,6 +16,8 @@
 
 package com.hazelcast.nio;
 
+import java.nio.charset.Charset;
+
 /**
  * Access and manipulate bits, bytes, primitives ...
  */
@@ -57,6 +59,15 @@ public final class Bits {
      * Length of the data blocks used by the CPU cache sub-system in bytes.
      */
     public static final int CACHE_LINE_LENGTH = 64;
+
+    /**
+     * A reusable instance of the UTF-8 charset
+     * */
+    public static final Charset UTF_8 = Charset.forName("UTF-8");
+    /**
+     * A reusable instance of the ISO Latin-1 charset
+     * */
+    public static final Charset ISO_8859_1 = Charset.forName("ISO-8859-1");
 
     private Bits() {
     }
--- a/pom.xml
+++ b/pom.xml
@@ -221,7 +221,7 @@
                 <version>${maven.surefire.plugin.version}</version>
                 <configuration combine.self="override">
                     <redirectTestOutputToFile>true</redirectTestOutputToFile>
-
+                    <runOrder>failedfirst</runOrder>
 
                     <!-- 1C means 1 process per cpu core -->
                     <!-- forkCount>1C</forkCount>
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,6 @@
+The package com.hazelcast.util.collection and the classes:
+com.hazelcast.util.QuickMath
+com.hazelcast.client.impl.protocol.util.UnsafeBuffer
+com.hazelcast.client.impl.protocol.util.BufferBuilder
+contain code originating from the Agrona project
+(https://github.com/real-logic/Agrona).
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -213,10 +213,8 @@
     <suppress checks="JavadocMethod" files="com.hazelcast.client.impl[\\/]"/>
 
     <!--CLIENT PROTOCOL-->
-    <!--TODO: we will refactor these External utils-->
-    <suppress checks="" files="com.hazelcast.client.impl.protocol.util[\\/]"/>
-    <!--TODO we will refactor these encode-decode code-->
-    <!--Parameters class are gonna be auto generated-->
+    <suppress checks="IllegalImport" files="com/hazelcast/client/impl/protocol/util/UnsafeBuffer"/>
+    <!--Parameters classes are auto-generated-->
     <suppress checks="" files="com.hazelcast.client.impl.protocol.map.*Parameters"/>
     <suppress checks="" files="com.hazelcast.client.impl.protocol.*Parameters"/>
     <suppress checks="" files="com.hazelcast.client.impl.protocol.*Codec"/>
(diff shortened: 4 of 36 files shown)


hazelcast/src/main/java/com/hazelcast/agrona/collections/Int2ObjectHashMap.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Real Logic Ltd.
```
asimarslan: I think we should keep the copyright of the original owner.

asimarslan:
I think the **agrona** name in the package has no value as you can put this info into javadoc. What about just removing? Also We already have two container impl in `com.hazelcast.util` you can put these there or make a *container* package and put them all in one place.

We previously backported `BitUtil` during client protocol dev but decided to merge it with the same purpose class `com.hazelcast.nio.Bits`

asimarslan:
We already have `Int2ObjectHashMapTest`in `com.hazelcast.client.protocol` we may remove one of them.

hazelcast/src/main/java/com/hazelcast/agrona/collections/Int2ObjectHashMap.java
mtopolnik: Not sure about this... as far as I could study it, the point is to make the attribution in `NOTICE.TXT`. What we have is derivative work.

mtopolnik:
>We already have Int2ObjectHashMapTestin com.hazelcast.client.protocol we may remove one of them.

This is already done: https://github.com/hazelcast/hazelcast/pull/5672/files#diff-de46e701b778043002544ee5681ff89e

>I think the agrona name in the package has no value as you can put this info into javadoc.

Attribution wasn't the motivation anyway (that's covered in `NOTICE.TXT`), it's just a lack of a better idea on my part. How about `com.hazelcast.util.collection` and put `CacheConcurrentHashMap` and `ConcurrentReferenceHashMap` there?

hazelcast/src/main/java/com/hazelcast/agrona/collections/Int2ObjectHashMap.java
asimarslan: Licensing issue always confuse me. But I think you are right we need a NOTICE file . I saw this example in an apache project:
https://github.com/apache/karaf/blob/master/NOTICE

mtopolnik: Yes, whenever derivative work is added to a project, a `NOTICE` file is needed; this is certain. The only point I wasn't completely sure about is that copyright line. However, since this is truly _derivative_ work and not pure copy-paste (it backports Java 8 code to Java 6), it makes sense to me to have our copyright.

mtopolnik:
>We previously backported BitUtil during client protocol dev but decided to merge it with the same purpose class com.hazelcast.nio.Bits

There is currently no overlap between `BitUtil` and `Bits`, but it makes sense to merge into a single place. However, perhaps we should reduce the scope of `BitUtil` as currently propesed? Do we need the hex string <-> byte array conversions? It's great for debugging purposes but we don't necessarily need to add it to the main project.

asimarslan:
For the Int2ObjectHashMapTest comment I wanted to say that the test files are duplicate.
 
https://github.com/hazelcast/hazelcast/blob/2365708642f19863eaa18277261b1c01e7b25548/hazelcast/src/test/java/com/hazelcast/client/protocol/Int2ObjectHashMapTest.java

mtopolnik:
Now the code is in `com.hazelcast.util.collection`; `BitUtil` has been discontinued and just two methods were scavenged into `...nio.Bits`. Checkstyle warnings have been fixed both for this contribution and for the earlier backport of `UnsafeBuffer`.

hazelcast/src/main/java/com/hazelcast/nio/Bits.java
```diff
@@ -337,4 +348,34 @@ public static long combineToLong(int x, int y) {
...
+     * Fast method of finding the next power of 2 greater than or equal to the supplied value.
+     * <p/>
+     * If the value is &lt;= 0 then 1 will be returned.
+     * <p/>
+     * This method is not suitable for {@link Integer#MIN_VALUE} or numbers greater than 2^30.
+     *
+     * @param value from which to search for next power of 2
+     * @return The next power of 2 or the value itself if it is a power of 2
+     */
+    public static int findNextPositivePowerOfTwo(final int value) {
```
mdogan: We have also `QuickMath.nextPowerOfTwo()`, doing the same job.

mtopolnik: Inconveniently, `nextPowerOfTwo` is `long`-based and my version was `int`-based. This will cause trouble at the highest bit (that's why it was  called "find next _positive_ power of two".

mtopolnik: Also, the current `nextPowerOfTwo` actually returns a negative value for `nextLongPowerOfTwo(1L << 63)`. This is probably broken.

mdogan: There two versions of `nextPowerOfTwo()` one is int-based, other is long. `public static long nextPowerOfTwo(long value)` and `public static int nextPowerOfTwo(int value)`. I don't say this one is better, I just say we should have only a single version of next-power-of-two.

Also what do you expect `nextPowerOfTwo(1L << 63)` to return? A positive value?  `1L << 63` itself is a negative value.

mtopolnik: Indeed, I was just writing the same. My point was intended to apply to `(1L << 62) + 1`, which is a positive value.

However, both methods actually do the same in this respect and `findNextPositivePowerOfTwo` documents it is not suitable for arguments above `1 << 30`.

What "positive" refers to, and where these methods differ, is the behavior against nonpositive arguments. `findNextPositivePowerOfTwo` returns 1 and `nextPowerOfTwo` returns 0. Here `nextPowerOfTwo` actually breaks its contract by returning 0 for 0.

mtopolnik: `nextPowerOfTwo` also breaks its contract by returning `i` for any `i` which is already a power of two. In this case it is probably the stated contract which is broken and the actual behavior is correct.

mtopolnik: Now I notice that `QuickMath` also has `bytesToHex`, but less optimal than the one from Agrona (allocates a temp `char[]`on each call).

mtopolnik:
verify

pom.xml
```diff
@@ -221,7 +221,7 @@
                 <version>${maven.surefire.plugin.version}</version>
                 <configuration combine.self="override">
                     <redirectTestOutputToFile>true</redirectTestOutputToFile>
-
+                    <runOrder>failedfirst</runOrder>
```
mdogan: Is this intentional?

mdogan:
:+1:

pom.xml
mtopolnik: It is. What are the downsides? I added it so I don't have to manually specify which test to run when I'm fixing a failed test.

asimarslan:
:+1:

PR #5674, opened 2015-07-10
Fixes #4888 - LockStore memory leak on map/multimap destroy 

### LINKED ISSUE #4888: LockStore not cleared when map is destroyed causing OutOfMemory
Hazelcast version:3.4.1
Cluster size:1
Number of the clients:1
Version of Java:1.6_45
OS: Linux version 2.6.32-358.el6.x86_64

In the com.hazelcast.concurrent.lock.LockStoreContainer class when clearLockStore is called the lockStores map entry is not removed.

I have attached screenshot of the heap dump from JVisualVM
lock_store_heap_dump
. Notice how com.hazelcast.spi.DefaultObjectNamespace and  com.hazelcast.concurrent.lock.LockStoreImpl classes hang around even after the map has been removed.

The lockstore objects pile up till the server finally runs out of memory

mdogan:
LockStore should be removed from registry when map/multimap is destroyed. Otherwise it will be a memory leak. 

Fixes #4888

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/MapLockTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/MapLockTest.java
@@ -16,9 +16,13 @@
 
 package com.hazelcast.map;
 
+import com.hazelcast.concurrent.lock.LockService;
+import com.hazelcast.concurrent.lock.LockServiceImpl;
+import com.hazelcast.concurrent.lock.LockStoreContainer;
 import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
+import com.hazelcast.spi.impl.NodeEngineImpl;
 import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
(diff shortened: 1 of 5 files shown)


hazelcast/src/test/java/com/hazelcast/map/MapLockTest.java
```diff
@@ -386,6 +390,28 @@ public void run() throws Exception {
...
 
+    /**
+     * See issue #4888
+     */
+    @Test
+    public void lockStoreShouldBeRemoved_whenMapIsDestroyed() {
+        HazelcastInstance hz = createHazelcastInstance();
+        IMap map = hz.getMap(randomName());
+        for (int i = 0; i < 1000; i++) {
+            map.lock(i);
```
pveentjer: I guess this is my lack of understanding, but if the key is unlocked after it is locked, it the key not automatically removed? If that is the case, then there is no purpose of doing lock/unlock. So I guess my knowledge is lacking here.

pveentjer:
Looks good. Minor comment.

hazelcast/src/test/java/com/hazelcast/map/MapLockTest.java
mdogan: Original issue is, there are leaking empty `LockStore`s after map is destroyed. During destroy, `LockStore`s are already purged, all entries are removed. But `LockStore` itself remains in the lock registry. 

So, it doesn't matter, unlock is called or not. If it makes the test easier to understand, I can remove the unlocks.

mdogan: Removed unlock.

mdogan: done.

pveentjer:
:+1:

hazelcast/src/test/java/com/hazelcast/map/MapLockTest.java
pveentjer: Thanks for the explanation. Will merge once green

Issue #5349, opened 2015-05-22
Thread.sleep in MapProxySupport slows instance start up


mahileeb:
I've noticed that our automated tests have slowed down since upgrading to 3.5EA due to the Thread.sleep(1000) at MapProxySupport:664 causing the load of each map to take at least one second.

Is there someway this can be changed to a wait/notify instead of a fixed sleep?

mahileeb:
It's actually worse than 1sec per map, seems more like 4sec. All these maps are empty on start so loading should be instant

bwzhang2011:
@gurbuzali, would you mind introducing the sleep use case in MapProxySupport and how to adjust such parameter just as @mahileeb pointed out.

PR #5676, opened 2015-07-10
Avoiding sleep when loading empty maps in eager mode

ajermakovics:
Issue #5349


ajermakovics:
verify

jerrinot:
:+1:

gurbuzali:
:+1:

PR #5677, opened 2015-07-10
Avoiding sleep when loading empty maps in eager mode (forward port)

ajermakovics:
Issue #5349


eminn:
Closing this one, rebased version is #6689

PR #5680, opened 2015-07-11
Reduce transmitted data for cache invalidation message

serkan-ozal:
Filters cache invalidation messages by checking/comparing source uuid and target uuid before sending them. So, 36 length source uuid values are neither serialized to data nor transmitted to client.

P.S: Breaks backward compatibility for new client (not old client). As we talked with @enesakar since it is experimental this is not problem and acceptable.


mesutcelik:
:+1:

kobalski:
:+1:

PR #5689, opened 2015-07-13
Minor AbstractIOSelector logging cleanup

pveentjer:
Instead of obtaining logger through: Logger.getLogger(AbstractIOSelector.class)
we make use of the logger field that is part of the AbstractIOSelector.


jerrinot:
:+1:

mdogan:
:+1:

PR #5696, opened 2015-07-13
[DO-NOT-MERGE] New Replicated Map Implementation

eminn:
It was used with intention of automatically solving conflicts but it is proven that approach is not working since vector clocks are not designed for that. Vector clocks were leading to the infinite message loops when they are trying to solve conflicts. That is a performance killer, in a big clusters they are biggest bottleneck to scale.


pveentjer:
Hi Emin,

what approach did you use so that all members converge to the same map content?

pveentjer:
@eminn ping.

eminn:
run-parallel

eminn:
run-parallel

eminn:
run-parallel

eminn:
run-parallel

PR #5697, opened 2015-07-13
FactoryIdHelper cleanup

pveentjer:
Cleanup of FactoryIdHelper
    
    1: all factory id's are now stored in the FactoryIdHelper. So one central place where
    you can see all the values.
    
    2: removed unused constants
    
    3: resolves issue with the count down latch serializer; it was pointing to the portable version
      and not the regular version
    
    4: resolved issue with the ringbuffer (wrong value being used).


jerrinot:
:+1:

PR #5700, opened 2015-07-14
arquillian test breaks mockito when it is run before them.

mustafaiman:
Arquillian does not shutdown properly until all the tests in the package is run. There is an open issue about that https://issues.jboss.org/browse/ARQ-1663 .

This commit splits the tests in JCA module in to two executions. First one runs non container tests and second one runs container tests.
https://developer.jboss.org/thread/230602?tstart=0

Mockito is not thread safe in stubbing. It is safer to run mockito tests in serial mode than parallel. https://groups.google.com/forum/#!topic/mockito/ctFkNxrwcWc

### DIFF
--- a/hazelcast-ra/hazelcast-jca/pom.xml
+++ b/hazelcast-ra/hazelcast-jca/pom.xml
@@ -109,52 +109,64 @@
             <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>2.12</version>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <phase>test</phase>
+                        <configuration>
+                        <excludes>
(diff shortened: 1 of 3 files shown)


kobalski:
:+1: @bilalyasar Can you please have a look at changes too? There is specific order for execution of tests in surefire plugin. Does it effect our maven build profiles?

mesutcelik:
@kobalski ,
where is this order defined?

kobalski:
@mesutcelik Sorry there is no specific order is defined, but inside pom.xml there are `default-test` and `arquillian` executions, which groups jca tests into two groups (unit and integration tests).  Is this configuration cause problems with our build profiles? For example, `test-coverage-nightly` profile has its own execution config. I am just asking whether changes in pom.xml breaks anything or not.

mesutcelik:
I don't think there will be an issue. All profiles are using the same config from JCA module. However, you can check this by running tests with `test-coverage-nightly` profile.

PR #5701, opened 2015-07-14
[3.1.8] Fixes #4888 - LockStore memory leak on map/multimap destroy

### LINKED ISSUE #4888: LockStore not cleared when map is destroyed causing OutOfMemory

mdogan:
**Backport to 3.1.8**

LockStore should be removed from registry when map/multimap is destroyed. Otherwise it will be a memory leak.

Backport of #5674 
Fixes #4888

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStoreContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStoreContainer.java
@@ -28,7 +28,7 @@ import java.util.concurrent.ConcurrentMap;
 /**
  * @author mdogan 2/12/13
  */
-final class LockStoreContainer {
+public final class LockStoreContainer {
 
     private final LockServiceImpl lockService;
     private final int partitionId;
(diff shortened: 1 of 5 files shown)


mdogan:
As far as I see, PR builder is not running tests for 3.1.x branch.

eminn:
:+1:

ahmetmircik:
:+1:

bilalyasar:
@mdogan: problem is that in the surefire plugin there is a line `<skipTests>true</skipTests>`
https://github.com/hazelcast/hazelcast/blob/3.1.8/pom.xml#L163

bilalyasar:
@mdogan in this PR you need to remove `<skipTests>true</skipTests>` line.

mdogan:
@bilalyasar: I know. We were using `parallel-test` profile before. That's why tests were not run on normal builds. I'm now running the tests on multiple machines manually.

PR #5704, opened 2015-07-14
[BACKPORT] Cache entry event improvements

serkan-ozal:
Before this improvement;

- Even though there is no cache entry listener registered and only invalidation listener is registered, cache entry events are created and sent to event system because of registered invalidation listener. By this improvement, cache entry events are sent to event system if there is any cache entry listener registered. 

- Same logic is available also for cache invalidation listener

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
@@ -26,6 +26,7 @@ import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.nio.IOUtil;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.partition.MigrationEndpoint;
+import com.hazelcast.spi.EventFilter;
 import com.hazelcast.spi.EventRegistration;
 import com.hazelcast.spi.EventService;
 import com.hazelcast.spi.NodeEngine;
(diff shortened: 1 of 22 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
```diff
@@ -52,11 +51,19 @@
         implements ICacheService, PostJoinAwareService {
 
     protected final ConcurrentMap<String, CacheConfig> configs = new ConcurrentHashMap<String, CacheConfig>();
+    protected final ConcurrentMap<String, CacheContext> cacheContexes = new ConcurrentHashMap<String, CacheContext>();
```
gurbuzali: `cacheContexts` ?

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheService.java
```diff
@@ -239,4 +240,86 @@ public void run() {
...
+            // We don't support this at the moment, because it is not used at the moment
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public void clear() {
+            // We don't support this at the moment, because
+            //   - It is not used at the moment
+            //   - It may or may not use "poll" method internally and this depends on the implementation
+            //     so it may change between different version of Java
```
gurbuzali: missing `throw new UnsupportedOperationException();` ?

serkan-ozal: yes, my bad. good point :+1:

gurbuzali:
some minor comments, overall good :+1:

gurbuzali:
I should add that some tests related with the newly introduced `NotifiableEventListener` would be good

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
serkan-ozal: right, my typo wrong

PR #5705, opened 2015-07-14
Fixes for FindBugs-raised issues

mtopolnik:
FindBugs discovered several issues with the Agrona backport classes. Some of them were legitimate and code was changed accordingly. Others just needed suppression.

Additionally this PR removes the usage of the outdated and deprecated `SuppressWarnings` annotation. The replacement is `SuppressFBWarnings` which avoids the name clash with the JDK standard annotation. The dependency to the FindBugs-annotations artifact was severely outdated (1.0.2 vs. current 3.0.0), which is the reason why the old annotation was not raising deprecation warnings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ target/
 .settings/
 .idea/
 .patch
+.surefire-*
 *.iml
 *.ipr
 *.iws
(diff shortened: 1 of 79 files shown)


pveentjer:
Looks good to me.

:+1:

PR #5707, opened 2015-07-15
change instance creation in client tests with TestHazelcastFactory to…

sancar:
… mock the connection and speed up the tests


hazelcast-client-new/src/test/java/com/hazelcast/client/ClientConnectionTest.java
```diff
@@ -42,36 +41,37 @@
 
 import static org.junit.Assert.assertEquals;
 
-@RunWith(HazelcastSerialClassRunner.class)
+@RunWith(HazelcastParallelClassRunner.class)
 @Category(QuickTest.class)
 public class ClientConnectionTest extends HazelcastTestSupport {
 
+    private final TestHazelcastFactory hazelcastFactory = new TestHazelcastFactory();
```
pveentjer: Can't we use the same approach as for the HazelcastTestSupport. E.g. HazelcastClientTestSupport which extends HazelcastTestSupport but then adds the ability to deal with clients. No factories exposed.

mdogan: :+1:

sancar: I am not really fond of that approach because I don`t find it fit to client. 
We have two kind of methods in HazelcastTestSupport

1) HazelcastInstance createHazelcastInstance(Config config);

This one create an instance and store it in HazelcastTestSupport. And we don`t use it ever again ?!?
For client I need to create at least two HazelcastInstance(one is server, other is client), 

2) TestHazelcastInstanceFactory createHazelcastInstanceFactory()
And for me this is not different than current one, this is just replacement for saying new TestHazelcastFactory()

I don`t really see any added value, and find current approach much more explicit. I might be missing something, if that is the case, I am listening.

pveentjer: Perhaps the HazelcastTestSupport should be cleaned up.

The main idea is that for a test you want to impose the least amount of environmental setup. Having an additional factory field violates this idea. Otherwise people will be copy pasting this line for every test they make and isn't very appealing. 

Apart from that I would like to see the setup of non-clients tests to be the same as client tests.

sancar: @pveentjer @mdogan These sound great but I want to this improvements if I have spare time. Lets not make these improvements a must for this pr .

pveentjer: Can you make a ticket for it? Then I'm happy to merge it. I'm even willing to help out changing the client tests so they can make use of the server side tests. This way you get a lot of duplicate test code removed

sancar: @pveentjer https://github.com/hazelcast/hazelcast/issues/5740 
Feel free to edit the explanation of ticket, if you think it needs more clarification.

gurbuzali:
:+1:

PR #5708, opened 2015-07-15
change instance creation in client tests with TestHazelcastFactory to…

sancar:
… mock the connection and speed up the tests


sancar:
I take look at the test-failures. Tests are failed in following scenerio
```
instance = hazelcastFactory.newHazelcastInstance();
hazelcastFactory.newHazelcastClient();
instance.shutdown();
hazelcastFactory.newHazelcastInstance();
```

It is failing because client does not know the address of second instance before it is created. One way to fix it is to revert back to create HazelcastInstanceFactory with total number server that is intended to created in the test.

I am trying a new approach with AddressProviders that we use for Aws.
https://github.com/sancar/hazelcast/commit/e30875f7e4c644390c9b8ea720c6d90d6badb1bc

@mdogan Does it look more complicated than it should be ? I am OK with reverting back to old way if we have to since it is much simpler. What do you think ?

sancar:
Merging this since master branch is merged.  We can discuss details later.

PR #5709, opened 2015-07-15
Clean up regexes in suppressions.xml

mtopolnik:
Our suppressions.xml contained a lot of `<suppress files="..." />` attributes which variably assumed they were java class names, platform-dependent pathnames, and pathname regexes. This PR turns them into the one thing they actually are, which is Java pathname regexes.


pveentjer:
:+1:

Donnerbart:
:+1:

We should adapt this for all other projects as well.

PR #5712, opened 2015-07-15
Cache entry event improvements

serkan-ozal:
Before this improvement;

- Even though there is no cache entry listener registered and only invalidation listener is registered, cache entry events are created and sent to event system because of registered invalidation listener. By this improvement, cache entry events are sent to event system if there is any cache entry listener registered. 

- Same logic is available also for cache invalidation listener


metanet:
:+1:

PR #5716, opened 2015-07-16
Transaction impl cleanup

pveentjer:
Moved operatons into operations package
Introduced TxBaseOperation to reduce duplication
Made TxOperations IdentifiedDataSerializable to reduce overhead for serialization


gurbuzali:
:+1: 
but needs rebase

PR #5717, opened 2015-07-16
TransactionLog renaming

pveentjer:
The TransactionLog is not a log for transactions, it is a single record in the transaction. So a single map.put is a single record within the transaction. That is why it has been renamed to TransactionRecord.

I also renamed the TransactionSupport to InternalTransaction and have it extend the Transaction interface.


ajermakovics:
:+1:

jerrinot:
:+1:

PR #5718, opened 2015-07-16
Independent sizing of buffers for connections opened by clients

jerrinot:
The PR introduces 2 new group properties to override buffers sizes for connections opened by clients.

It allows Hazelcast members to have large buffers for member-to-member connections and smaller buffers for connections opened by clients - as there might be 10,000's of clients, but typical cluster size is way smaller.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
@@ -23,6 +23,7 @@ import com.hazelcast.util.Clock;
 import com.hazelcast.util.EmptyStatement;
 
 import java.io.IOException;
+import java.net.SocketException;
 import java.nio.ByteBuffer;
 import java.nio.channels.SelectionKey;
 import java.util.Queue;
@@ -44,7 +45,7 @@ public final class WriteHandler extends AbstractSelectionHandler implements Runn
     private final Queue<SocketWritable> writeQueue = new ConcurrentLinkedQueue<SocketWritable>();
     private final Queue<SocketWritable> urgentWriteQueue = new ConcurrentLinkedQueue<SocketWritable>();
     private final AtomicBoolean scheduled = new AtomicBoolean(false);
-    private final ByteBuffer outputBuffer;
+    private ByteBuffer outputBuffer;
     private SocketWritable currentPacket;
     private SocketWriter socketWriter;
     private volatile long lastHandle;
@@ -58,7 +59,6 @@ public final class WriteHandler extends AbstractSelectionHandler implements Runn
 
     WriteHandler(TcpIpConnection connection, IOSelector ioSelector) {
         super(connection, ioSelector, SelectionKey.OP_WRITE);
-        this.outputBuffer = ByteBuffer.allocate(connectionManager.socketSendBufferSize);
     }
 
     long getLastHandle() {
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
```diff
@@ -93,15 +94,32 @@ private void createWriter(String protocol) {
...
     }
 
+    private void adjustBufferSizesForClientConnection()  {
+        if (connectionManager.socketClientSendBufferSize == connectionManager.socketSendBufferSize) {
+            return;
+        }
+
+        outputBuffer = ByteBuffer.allocate(connectionManager.socketClientSendBufferSize);
+        try {
+            connection.getSocketChannelWrapper().socket().setSendBufferSize(connectionManager.socketClientSendBufferSize);
```
serkan-ozal: I prefer giving responsibility of setting buffer sizes to `TcpIpConnection` class by introducing new two methods instead of this long call.
- `setReceiveBufferSize(int size)`
- `setSendBufferSize(int size)`

So inside itself, it can call `socketChannel.socket().setReceiveBufferSize(size)` and `socketChannel.socket().setSendBufferSize(size)`

jerrinot: that's a good one!

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
```diff
@@ -93,15 +94,32 @@ private void createWriter(String protocol) {
                 outputBuffer.put(stringToBytes(Protocols.CLUSTER));
                 registerOp(SelectionKey.OP_WRITE);
             } else if (Protocols.CLIENT_BINARY.equals(protocol)) {
+                adjustBufferSizesForClientConnection();
```
serkan-ozal: intstead of adjusting at here (after `outputBuffer` is allocated at constructor), what about initialize it here?
I know, generally, these properties will not be configured so `outputBuffer` will be used as it created at constructor.
But for cases like too many client like 1000 clients, it will create `1000 * 32k` temporary output buffers and then they will be replaced with the adjusted ones.

serkan-ozal:
Minor comments, and ready to go one PR builder is green.
@jerrinot Will you send this PR also to master?

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
jerrinot: the thing is you don't know the type of the connection when you are creating an instance of read/write buffer.

jerrinot:
@serkan-ozal: I will!

hazelcast/src/main/java/com/hazelcast/nio/tcp/WriteHandler.java
serkan-ozal: @jerrinot yes, I am saying creating `outputBuffer` is only at here by removing its initialization at constructor

jerrinot: I see, you meant to not creating the buffers in constructors at all. I could do it in a master, however I'm not sure about the maintenance branch. I don't want to risk NPE due funky access/usage pattern.

serkan-ozal:
```
    WriteHandler(TcpIpConnection connection, IOSelector ioSelector) {
        super(connection, ioSelector, SelectionKey.OP_WRITE);
        this.outputBuffer = ByteBuffer.allocate(connectionManager.socketSendBufferSize);
    }
```

Talking about removing this line `this.outputBuffer = ByteBuffer.allocate(connectionManager.socketSendBufferSize);`

serkan-ozal: ok enough

jerrinot: altought, I'm checking the code-flow and it's safe. I'll do it even in a maintenance branch.

pveentjer:
:+1:

PR #5719, opened 2015-07-16
Cleanup for TransactionRecord

pveentjer:
The TransactionRecord now exposes methods for creating the Commit/Prepare/Rollback operations.

The invocation logic has been placed in the AbstractTransactionRecord since it is the same for
all implementations. This cleanup is preparation for transaction optimizations for the 3.6
release.

The idea is that eventually all the invoke methods are going to be removed from the the TransactionRecord and the transaction will take care of it. Here we can play with optimizations like:
- batching operations for a single partition (no need for an invocation per item per partition)
- last cohort optimization: no need to first send a prepare operation and then a commit operation, this can be squashed into a single preparecommit operation. This will also speed up transaction that modify a single item since number of remote calls is reduced.


jerrinot:
:+1: once it's green

PR #5720, opened 2015-07-16
Minor renaming of TxLog to TxRecord

pveentjer:


hazelcast/src/main/java/com/hazelcast/transaction/impl/xa/XATransactionImpl.java
```diff
@@ -78,8 +78,8 @@
     private final SerializableXID xid;
     private final String txOwnerUuid;
 
-    private final List<TransactionRecord> txLogs = new LinkedList<TransactionRecord>();
-    private final Map<Object, TransactionRecord> txLogMap = new HashMap<Object, TransactionRecord>();
+    private final List<TransactionRecord> records = new LinkedList<TransactionRecord>();
```
jerrinot: what's the problem with the `txLogs`? isn't it more descriptive than just `records`?

pveentjer: the name logs is wrong since it stores records. The transaction is all about transacrtion records and if you prefix it all the time with transaction records, the code starts to become very verbose for no real reason.

jerrinot: ok

Donnerbart: Isn't this a common term? They also use the term "log record" here: https://en.wikipedia.org/wiki/Transaction_log

pveentjer: It is a common term and I'm changing the code so it reflects this naming style. Originally an entry in this log (e.g. a map.put) was called a TransactionLog.. but it isn't a lot.. it is a record in the log.

pveentjer:
@jerrinot can you have another look. 

I introduced TransactionLog which now wraps the list/map that was found in the transaction.

jerrinot:
:+1:

Donnerbart:
Nice de-duplication :)

:+1:

PR #5723, opened 2015-07-17
Add legal wording to Agrona backport

mtopolnik:
In addition to the copyright notice in each file and attribution in `NOTICE`, Apache License 2.0 also has this rule (§4.b):

> You must cause any modified files to carry prominent notices stating that You changed the files

To more explicitly comply with this rule I added four words to each affected file's copyright notice.

Related: [Martin Thompson's comment on an Agrona issue](https://github.com/real-logic/Agrona/issues/16#issuecomment-121973804)


pveentjer:
Can't we just shade the Agrona library instead of this manual copying?

mtopolnik:
We would of course prefer to do it like that, but Agrona is a Java 8 project.

pveentjer:
Ok.

PR #5724, opened 2015-07-17
TransactionLogRecord cleanup

pveentjer:
The LogRecords are responsible for creating operations, but not for the actual invocation.

This prevents code duplication, but it also makes it possible to do optimizations like last
cohort optimization, merging operations for a single partition etc.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/collection/impl/txncollection/CollectionTransactionLogRecord.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/impl/txncollection/CollectionTransactionLogRecord.java
@@ -19,20 +19,16 @@ package com.hazelcast.collection.impl.txncollection;
 import com.hazelcast.collection.impl.txncollection.operations.CollectionPrepareOperation;
 import com.hazelcast.collection.impl.txncollection.operations.CollectionRollbackOperation;
 import com.hazelcast.collection.impl.txncollection.operations.CollectionTxnRemoveOperation;
-import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
-import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.Operation;
-import com.hazelcast.spi.impl.operationservice.InternalOperationService;
 import com.hazelcast.transaction.impl.KeyAwareTransactionLogRecord;
-import com.hazelcast.util.ExceptionUtil;
+
 import java.io.IOException;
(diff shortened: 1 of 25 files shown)


hazelcast/src/main/java/com/hazelcast/collection/impl/txncollection/CollectionTransactionLogRecord.java
```diff
@@ -58,52 +50,26 @@ public CollectionTransactionLogRecord(long itemId,
...
-    @Override
-    public Future rollback(NodeEngine nodeEngine) {
-        boolean removeOperation = op instanceof CollectionTxnRemoveOperation;
-        CollectionRollbackOperation operation = new CollectionRollbackOperation(name, itemId, removeOperation);
-        try {
-            return nodeEngine.getOperationService().invokeOnPartition(serviceName, operation, partitionId);
-        } catch (Throwable t) {
-            throw ExceptionUtil.rethrow(t);
-        }
+    public Operation createCommitOperation() {
```
jerrinot: just a cosmetic thing: 
for sake of consistency would be better to use `getCommitOperation` - as in e.g. `BackupAwareOperation::getBackupOperation`

pveentjer: I don't think the name 'get' is appropriate here.

jerrinot: yeah, the `create` prefix makes more sense to me too. 
it's purely for consistency. anyway, it's not hugely important. I leave it up to you.

jerrinot:
test failures in tx tests:
```
com.hazelcast.collection.impl.queue.QueueTestsFrom2X.testQueuePollRollbackSize
com.hazelcast.collection.impl.queue.QueueTestsFrom2X.testQueueOrderAfterPollRollback
com.hazelcast.map.MapTransactionTest.testTxnGetForUpdateTxnFails
```

pveentjer:
I'm working on them.

hazelcast/src/main/java/com/hazelcast/collection/impl/txncollection/CollectionTransactionLogRecord.java
pveentjer: I'll go for correct names. You apply consistency for good things, so that more good things are caused bu other good things. But you don't apply consistency for inappropriate naming choices.

jerrinot:
:+1:

ajermakovics:
:+1: on green

PR #5726, opened 2015-07-17
Introduced map recordstore package

pveentjer:
To reduce the number of classes/interfaces in the map.impl package,
the recordstore package has been introduced containing all recordstore
related functionality.


gurbuzali:
:+1:

PR #5727, opened 2015-07-17
Introduced map nearcache package

pveentjer:
Near cache package was introduced to prevent the map.impl package
becoming too large to deal with.


gurbuzali:
:+1:

PR #5732, opened 2015-07-18
More logic into the TransactionLog

pveentjer:
The rollback, commit and prepare logic is now moved into the TransactionLog.

This is a preparation of making the TransactionLog smarter for future optimizations
like last cohort and merging operations for single partition.


gurbuzali:
:+1:

PR #5735, opened 2015-07-19
Removed InternalTransaction

pveentjer:
Class could not carry its weight since the Transaction also is
an internal instead of an external interface. So there is no point
in having 2 internal ones.


gurbuzali:
:+1:

PR #5736, opened 2015-07-19
XA Transaction cleanup

pveentjer:
* Moved operations for xa transactions to operations package
* Renamed XATransactionImpl to XATransaction
* Renamed XATransactionHolder to XATransactionDTO since it doesn't hold a pointer, but
is a dto version of the XATransaction


gurbuzali:
:+1:

PR #5737, opened 2015-07-19
Clean up map transactions

pveentjer:
* All operations/objects in the map.impl.tx package are now IdentifiedDataSerializable
* Operations are moved to map.impl.tx.operations
* Cleanup of the MapDataSerializableHook. Got rid of the messy array constructor and replaced
it by a simpler switch case version.


Donnerbart:
Should we move this to 3.7?

PR #5741, opened 2015-07-20
Deleted ClientCompatibleTest annotation

pveentjer:
It isn't used. And clients, with the fixes of sancar for client side connection mocking,
can probably make use of the serverside tests.


pveentjer:
Since it is only information and part of testing; I'll merge this one myself.

PR #5744, opened 2015-07-20
Listeners should continue even though client is disconnected from clust…

sancar:
…er and reconnect later, as long as client is not shutdown. Feature was broken with new client protocol. This pr contains the fix. Also added plenty of tests to make sure we dont broke the feature again.


sancar:
verify

gurbuzali:
:+1:

PR #5749, opened 2015-07-20
Removed the ordering from the TransactionLog

pveentjer:
The TransactionLogRecords were stored in a map and in a list. The list
was used to determine ordering. The ordering was only used in prepare/commit/rollback,
but these operations are executed in parallel anyway. So the current implementation
doesn't preservere ordering.

From a logical perspective I'm not convinced that these operations rely on any particular
ordering of records.


gurbuzali:
:+1:

pveentjer:
Ordering is needed. E.g. when there is a list and a bunch of adds, these adds need to be applied in the correct order.

The problem however is that the current mechanism (the one before this PR) being used to apply these changes is not guaranteed to provide this ordering. It can happen, when partitions migrate, that invocations are executed out of invocation-order. 

So we need to have a closer at how ordering needs to be applied correctly.

PR #5752, opened 2015-07-21
Removed unnecessary SoftLock reference in hibernate EntryProcesors

kobalski:
Hibernate EntryProcessor were using SoftLock interface unnecessarily.
Because of this, when 2nd Level Cache is used with client only mode,
remote cluster should have hibernate-core jars in the class path. With
this PR, such requirement is not needed any more.

### DIFF
--- a/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/distributed/IMapRegionCache.java
+++ b/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/distributed/IMapRegionCache.java
@@ -26,6 +26,8 @@ import com.hazelcast.hibernate.RegionCache;
 
 
 import com.hazelcast.hibernate.serialization.Expirable;
+import com.hazelcast.hibernate.serialization.MarkerWrapper;
+import com.hazelcast.hibernate.serialization.ExpiryMarker;
 import com.hazelcast.hibernate.serialization.Value;
 import org.hibernate.cache.CacheDataDescription;
 import org.hibernate.cache.access.SoftLock;
@@ -105,8 +107,13 @@ public class IMapRegionCache implements RegionCache {
     }
 
     public boolean update(final Object key, final Object newValue, final Object newVersion, final SoftLock lock) {
-        return (Boolean) map.executeOnKey(key, new UpdateEntryProcessor(lock, newValue, newVersion,
-                nextMarkerId(), nextTimestamp(hazelcastInstance)));
+        if (lock instanceof MarkerWrapper) {
+            final ExpiryMarker unwrappedMarker = ((MarkerWrapper) lock).getMarker();
+            return (Boolean) map.executeOnKey(key, new UpdateEntryProcessor(unwrappedMarker, newValue, newVersion,
+                    nextMarkerId(), nextTimestamp(hazelcastInstance)));
+        } else {
+            return false;
+        }
     }
 
     public boolean remove(final Object key) {
@@ -115,12 +122,17 @@ public class IMapRegionCache implements RegionCache {
 
     public SoftLock tryLock(final Object key, final Object version) {
         long timeout = nextTimestamp(hazelcastInstance) + lockTimeout;
-
-        return (SoftLock) map.executeOnKey(key, new LockEntryProcessor(nextMarkerId(), timeout, version));
+        final ExpiryMarker marker = (ExpiryMarker) map.executeOnKey(key,
+                new LockEntryProcessor(nextMarkerId(), timeout, version));
+        return new MarkerWrapper(marker);
     }
 
     public void unlock(final Object key, SoftLock lock) {
-        map.executeOnKey(key, new UnlockEntryProcessor(lock, nextMarkerId(), nextTimestamp(hazelcastInstance)));
+        if (lock instanceof MarkerWrapper) {
+            final ExpiryMarker unwrappedMarker = ((MarkerWrapper) lock).getMarker();
+            map.executeOnKey(key, new UnlockEntryProcessor(unwrappedMarker, nextMarkerId(),
+                    nextTimestamp(hazelcastInstance)));
+        }
     }
 
     public boolean contains(final Object key) {
--- a/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/distributed/UpdateEntryProcessor.java
+++ b/hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/distributed/UpdateEntryProcessor.java
@@ -22,7 +22,6 @@ import com.hazelcast.hibernate.serialization.HibernateDataSerializerHook;
 import com.hazelcast.hibernate.serialization.Value;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
-import org.hibernate.cache.access.SoftLock;
 
 
 import java.io.IOException;
@@ -34,7 +33,7 @@ import java.util.Map;
  */
 public class UpdateEntryProcessor extends AbstractRegionCacheEntryProcessor {
 
-    private SoftLock lock;
+    private ExpiryMarker lock;
     private Object newValue;
     private Object newVersion;
     private String nextMarkerId;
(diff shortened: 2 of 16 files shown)


hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/serialization/ExpiryMarker.java
```diff
@@ -33,7 +33,7 @@
  *     <li>It will always return a null value, resulting in a cache miss</li>
  *     <li>It is only replaceable when it is completely expired</li>
  *     <li>It can be marked by multiple transactions at the same time and will not expire until all transactions complete</li>
- *     <li>It should not be expired unless {@link #matches(SoftLock)} is true</li>
+ *     <li>It should not be expired unless  is true</li>
  * </ul>
  */
 public class ExpiryMarker extends Expirable implements SoftLock {
```
mesutcelik: you still have `SoftLock` dependency

kobalski: @mesutcelik  Yes, did not notice that. `SoftLock` is only used by Hibernate as a marker interface. `ExpiryMarker ` does not implement `SoftLock` now.  I have added new `MarkerWrapper` which only implements `SoftLock` and acts as a DTO between hibernate and IMap.  

`SoftLock` interface is used by `RegionAccessStrategy` interface of hibernate. No class used inside EntryProcessor need to implement it.
 
https://docs.jboss.org/hibernate/orm/4.2/javadocs/org/hibernate/cache/spi/access/RegionAccessStrategy.html

bilalyasar:
:+1:

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/serialization/MarkerWrapper.java
```diff
@@ -0,0 +1,37 @@
...
+
+
+import org.hibernate.cache.access.SoftLock;
+
+/**
+ * A wrapper class for ExpiryMarker for returning copy of marker object with SoftLock
+ * marker interface.
+ */
+
+public class MarkerWrapper extends ExpiryMarker implements SoftLock {
```
jhinch: Does this need to extend `ExpiryMarker`? Can it not use composition to include the marker?

```
public class MarkerWrapper implements SoftLock {
    private final ExpiryMarker marker;
    public ExpiryMarker getMarker() {
        return marker;
    }
} 
```

This avoids the need for all the setters on `ExpiryMarker` which I believe complicates reasoning on the state of `ExpiryMarker`. The original idea was that `ExpiryMarker` was effectively immutable, it only had immutable fields for serialization/deserialization.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/serialization/ExpiryMarker.java
```diff
@@ -33,10 +30,10 @@
  *     <li>It will always return a null value, resulting in a cache miss</li>
  *     <li>It is only replaceable when it is completely expired</li>
  *     <li>It can be marked by multiple transactions at the same time and will not expire until all transactions complete</li>
- *     <li>It should not be expired unless {@link #matches(SoftLock)} is true</li>
+ *     <li>It should not be expired unless  is true</li>
```
cszmajda: Minor point: this comment doesn't make sense any more...

hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/serialization/Expirable.java
```diff
@@ -71,13 +70,13 @@ public Object getVersion() {
...
      */
-    public abstract boolean matches(SoftLock lock);
+    public abstract boolean matches(Object lock);
 
     /**
      * Mark the entry for expiration with the given timeout and marker id.
      * <p/>
      * For every invocation a corresponding call to {@link ExpiryMarker#expire(long)} should be made, provided that
-     * the returned marker {@link #matches(SoftLock) matches}
+     * the returned marker
```
cszmajda: Minor point: this comment doesn't make sense any more...

kobalski:
@jhinch @cszmajda Thanks for your feed back. I made the changes as you have suggested.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/distributed/IMapRegionCache.java
```diff
@@ -115,8 +117,9 @@ public boolean remove(final Object key) {
...
     public SoftLock tryLock(final Object key, final Object version) {
         long timeout = nextTimestamp(hazelcastInstance) + lockTimeout;
-
-        return (SoftLock) map.executeOnKey(key, new LockEntryProcessor(nextMarkerId(), timeout, version));
+        final ExpiryMarker marker = (ExpiryMarker) map.executeOnKey(key,
+                new LockEntryProcessor(nextMarkerId(), timeout, version));
+        return new MarkerWrapper(marker);
     }
 
     public void unlock(final Object key, SoftLock lock) {
```
jhinch: You need to get the marker from the softlock here. `((MarkerWrapper) lock).getMarker()`.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/distributed/UpdateEntryProcessor.java
```diff
@@ -34,7 +33,7 @@
  */
 public class UpdateEntryProcessor extends AbstractRegionCacheEntryProcessor {
 
-    private SoftLock lock;
+    private Object lock;
```
jhinch: I would recommend using `ExpiryMarker` here.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/serialization/ExpiryMarker.java
```diff
@@ -95,7 +92,7 @@ public Object getValue(long txTimestamp) {
     }
 
     @Override
-    public boolean matches(SoftLock lock) {
+    public boolean matches(Object lock) {
```
jhinch: I would recommend using `ExpiryMarker` here as well.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/distributed/IMapRegionCache.java
jhinch: You'll probably want to do an `instanceof MarkerWrapper` check here as well.

kobalski: Yes, you are right I forgot to unwrap the value also added `instaceof`.

hazelcast-hibernate/hazelcast-hibernate3/src/main/java/com/hazelcast/hibernate/serialization/ExpiryMarker.java
```diff
@@ -33,10 +31,10 @@
  *     <li>It will always return a null value, resulting in a cache miss</li>
  *     <li>It is only replaceable when it is completely expired</li>
  *     <li>It can be marked by multiple transactions at the same time and will not expire until all transactions complete</li>
- *     <li>It should not be expired unless {@link #matches(SoftLock)} is true</li>
+ *     <li>It should not be expired unless {@link #matches(ExpiryMarker) matches} is true</li>
  * </ul>
  */
-public class ExpiryMarker extends Expirable implements SoftLock {
+public class ExpiryMarker extends Expirable implements Serializable {
```
kobalski: `ExpiryMarker ` is implementing `Serializable` interface, because it is being used inside `UpdateEntryProcessor` and `UnlockEntryProcessor`. Since `EntryProcessor` interface extends `Serializable`, all fields inside EntryProcessors must implement `Serializable`. Otherwise our findbugs build fails with `non-transient non-serializable instance field in Class` error.

PR #5753, opened 2015-07-21
Fixes the IsStillRunning storm

pveentjer:
It can happen that checking if an operation is running using the IsStillRunning operation,
leads itself to a checking of the IsStillRunning operation. So you can get an explosion
of such invocations which can crash a cluster.

Eventually we should discard the current invocation based approach completely. Either it needs to be taken care of by the replication system. Or till that is in place, we can periodically send a ping to each member in the cluster containing all the callid's of operations running. If a call has not received such a ping for a certain time, we can assume it isn't running. But this approach is fragile since you rely on a time based consensus.


metanet:
:+1:

jerrinot:
:+1:

mdogan:
:+1:

PR #5763, opened 2015-07-22
NodeEngine: removed hard dependency on QuorumServiceImpl

pveentjer:
The NodeEngine returned QuorumServiceImpl and not a QuorumService. We don't
want to expose concrete implementations of services in our API.

This change breaks compatibility. So if a user has made use of this API;
he will suffer from it.


pveentjer:
verify

jerrinot:
:+1:

PR #5764, opened 2015-07-22
NodeEngine: removed dependency on MemberImpl

pveentjer:
MemberImpl is an implementation detail and should not be exposed as API.
This PR fixes that by returning a Member interface, not MemberImpl.

This PR breaks compatibility!

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/Member.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/Member.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.core;
 
+import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.DataSerializable;
 
 import java.net.InetSocketAddress;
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/core/Member.java
```diff
@@ -40,6 +41,14 @@
     boolean localMember();
 
     /**
+     * Returns the Address of this Member.
+     *
+     * @return the address.
+     * @since 3.6
```
jerrinot: good one. we should use `@since` more

jerrinot:
:+1:

ajermakovics:
:+1:

hazelcast/src/main/java/com/hazelcast/core/Member.java
mdogan: BTW, this change makes `Address` a public API by exposing it here. Before it was an internal thing.

PR #5766, opened 2015-07-22
ClientDelegatingFuture was using same codecs for different responses.…

sancar:
… To make use of correct decoder ClientMessageDecoder is implemented. Via this interface decoder methods can be passed to ClientDelegatingFuture.


mdogan:
:+1:

Issue #5694, opened 2015-07-13
Near cache on client size for replicated map does not gets invalidated after replicated map changes


HaveNuIdea:
This issue was initially described in https://groups.google.com/forum/#!topic/hazelcast/QkSlH2UWhsI

I'am trying to add NearCache to ReplicatedMap and configure it to invalidate values on change.
```java
ClientConfig clientConfig = new ClientConfig();
NearCacheConfig nnc = new NearCacheConfig();
nnc.setInvalidateOnChange(true);
nnc.setInMemoryFormat(InMemoryFormat.OBJECT);
Map<String, NearCacheConfig> cachedMaps = new HashMap();
cachedMaps.put("TEST", nnc);
clientConfig.setNearCacheConfigMap(cachedMaps);
HazelcastInstance client = HazelcastClient.newHazelcastClient(clientConfig);
```
I have a Hazelcast 3.5 cluster with 2 nodes.
I create 2 clients using the above code.
1) First client populates ReplicatedMap "TEST" with some values
2) Then second client starts getting one of the values in a loop
3) First client changes the value in ReplicatedMap
4) Using jvisualvm i can see that values in the ReplicatedMap have changed
5) But second client only sees obsolete values

sancar:
I made straightforward fixes. I opened a new issue for some more improvements
https://github.com/hazelcast/hazelcast/issues/5769

jerrinot:
this is fixed by #5767

PR #5768, opened 2015-07-22
Added near cache listener to invalidate client replicated map when th…

sancar:
…ere is changes from other clients.


sancar:
fixes #5694

gurbuzali:
:+1:

PR #5776, opened 2015-07-24
Supress stacktrace when stringconstructor not found

pveentjer:
No reason to dump the stack trace if the string constructor can't be found.
Caused noise in the logging when debugging and doesn't add any value.


jerrinot:
:+1:

jerrinot:
:+1:

serkan-ozal:
:+1: 
Besides that I noticed that when `hazelcast.nio.faststring` is enabled, `UTFEncoderDecoder` tries to create `FastStringCreator` but if the reflection is not available due to security manager, `buildFastStringCreator` return null and then `StringCreator` becomes null

serkan-ozal:
Created a issue for this https://github.com/hazelcast/hazelcast/issues/5777

PR #5780, opened 2015-07-24
Moved MulticastServiceConstruction to MulticastService

pveentjer:
The construction of the MulticastService was part of node and is now moved to the right location.

No logic was changed, the method was just moved.


ajermakovics:
:+1:

PR #5781, opened 2015-07-24
Removed Node.getName

pveentjer:


jerrinot:
:+1: (assuming it's not considered to be a part of SPI)

pveentjer:
Node is an internal class. So it will not impact end users.

PR #5783, opened 2015-07-25
ReplicaSyncResponse removed code duplication

pveentjer:
The error response handler was a duplicate of what is provided in the
OperationResponseHandlerFactory


jerrinot:
:+1: (once it's green)

PR #5785, opened 2015-07-27
PacketHandler cleanup

pveentjer:
All services that can receive packets, implement the PacketHandler interface. 
There is no more hard dependency on services in the PacketTranseiver
Moved the response packet handling logic out of the operation executor
Made the InternalOperationService dumber by not needing to expose internal to handle packets.


hazelcast/src/main/java/com/hazelcast/wan/WanReplicationService.java
```diff
@@ -49,7 +50,7 @@
      *
      * @param packet the WAN replication packet to process
      */
-    void handleEvent(Packet packet);
```
jerrinot: I suspect this will have an impact on Hazelcast Enterprise?

hazelcast/src/main/java/com/hazelcast/spi/impl/PacketHandler.java
```diff
@@ -14,21 +14,22 @@
...
+public interface PacketHandler {
 
     /**
-     * Signals the ResponsePacketHandler that there is a response packet that should be handled.
+     * Signals the PacketHandler that there is a packet to be handled.
      *
      * @param packet the response packet to handle
-     * @throws Exception
      */
     void handle(Packet packet) throws Exception;
```
jerrinot: just a thought: what about renaming the method to something like `handlePacket()`? it seems like unnecessary, but what if a service implements multiple `XYHandler` interfaces?

hazelcast/src/main/java/com/hazelcast/wan/WanReplicationService.java
pveentjer: Maybe. But that should be a very easy fix.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/ResponsePacketHandlerImpl.java
```diff
@@ -19,31 +19,33 @@
...
-    private final NodeEngineImpl nodeEngine;
+    private final SerializationService serializationService;
+    private final InvocationRegistry invocationRegistry;
 
-    public ResponsePacketHandlerImpl(OperationServiceImpl operationService) {
-        this.operationService = operationService;
-        this.logger = operationService.logger;
-        this.nodeEngine = operationService.nodeEngine;
+    public ResponsePacketHandlerImpl(ILogger logger,
+                                     SerializationService serializationService,
```
jerrinot: good one!

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/AsyncResponsePacketHandler.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public int getQueueSize() {
+        return workQueue.size();
+    }
+
+    public void shutdown() {
+        responseThread.shutdown();
+    }
+
+    //@Override
```
jerrinot: typo?

hazelcast/src/main/java/com/hazelcast/spi/impl/PacketHandler.java
pveentjer: Then the type will define the method.

jerrinot: imagine you have: `ResponsePacketHandler` and let's say `RequestPacketHandler`..or any other type of `PacketHandler`.

hazelcast/src/main/java/com/hazelcast/wan/WanReplicationService.java
jerrinot: ok

jerrinot:
some minor comments, more like thoughts, not really issues. it's up to your judgements to address them or not. 

:+1: from me (once it's green)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/AsyncResponsePacketHandler.java
pveentjer: I'll merge that one directly in master when this PR is merged. Else we need to wait again for it to build. Unless of course there is another review comment that needs to get addressed.

jerrinot: ok

hazelcast/src/main/java/com/hazelcast/spi/impl/PacketHandler.java
pveentjer: Then the name of the method is not going to change anything since both will have the same 'conflicting' name. Then you need to demultiplex in the service; like I already do in the OperationService where I inspect the packet header and send to response handler or operation executor.

mdogan:
:+1:

PR #5787, opened 2015-07-27
Minor NodeEngineImpl dependencies cleanup

pveentjer:
Instead of refering to Node for logger/serialization, just get the dependencies.

Also some serialization duplication was removed from the NodeEngineImpl.


jerrinot:
:+1:

PR #5778, opened 2015-07-24
fixes aws different region bug - backport  of #5774

bilalyasar:
backport of #5774 

fixes #5653 #5446 #4731

PR #5789, opened 2015-07-27
Fix the v4 signer

igmar:
- The signature wasn't correct : It needs more then just the resulting
  hash
- x-amz-date is required


igmar:
I'll properly fix the pull, this is to start the discussion.

bilalyasar:
@igmar , in my local hz aws joiner works correctly. so is there a bug in the code?

igmar:
@bilalyasar You're sure that's a v4 signer ? I'm pretty sure it's broken in it's current state.

jarosite:
any updates for this?

mesutcelik:
fixed by https://github.com/hazelcast/hazelcast/pull/5778

PR #5792, opened 2015-07-27
Removed InternalOperationService asyncInvokeOnTarget

pveentjer:
Method was not used and wasn't tested. So no need to keep it in the system.


jerrinot:
:+1:

PR #5793, opened 2015-07-27
Cleanup: Implementation of Predicates extracted into top-level classes.

jerrinot:


hazelcast/src/main/java/com/hazelcast/query/Predicates.java
```diff
@@ -51,15 +42,6 @@ public static Predicate instanceOf(final Class klass) {
         return new InstanceOfPredicate(klass);
     }
 
-    private static Comparable readAttribute(Map.Entry entry, String attribute) {
```
jerrinot: this has been moved into a package protected class `com.hazelcast.query.impl.predicates.AttributeUtils`
I'm not exactly sure about it's purpose. It's used inside `LikePredicate` and `RegexPredicate`

There is another implementation of `readAttribute` in `AbstractPredicate`. It's quite confusing, but it might be caused by my lack of understanding.

jerrinot:
it's all just a pure refactoring -> move to top-level class. there is one exception - the `readAttribute` method I pointed out above as a review comment.

ajermakovics:
:+1:

pveentjer:
:+1:

jerrinot:
I'm afraid we'll have to revert this. As I just realized users might instantiate predicates directly without using the factory methods.

pveentjer:
Ah the joy of an accidentally exposed implementation detail.

Why not let the old classes be there; mark them as deprecated and focus on evolving the new classes.

PR #5794, opened 2015-07-27
Removed unnecessary while loop inside ClusteredSessionService which

kobalski:
causes high cpu usage.


hazelcast-wm/src/main/java/com/hazelcast/web/ClusteredSessionService.java
```diff
@@ -104,18 +103,14 @@ public ClusteredSessionService(FilterConfig filterConfig, Properties properties,
...
-                        }
-                        EmptyStatement.ignore(ignored);
-                    }
+                try {
+                    ensureInstance();
+                } catch (Exception ignored) {
+                    EmptyStatement.ignore(ignored);
                 }
             }
         }, 0, CLUSTER_CHECK_INTERVAL, TimeUnit.SECONDS);
```
bilalyasar: `CLUSTER_CHECK_INTERVAL = 1000L` , so every 1000 seconds is seems too much. And probably test is failed due to this interval... Maybe we can decrease this interval.

kobalski: @bilalyasar Yes, you are right. I made it 5 seconds for now. But maybe we should make it `CLUSTER_CHECK_INTERVAL ` configurable in the future.

bilalyasar:
:+1:  this PR seems ok to me, after merging we can observe nightly web filter tests.. because we are waiting https://github.com/hazelcast/hazelcast/blob/master/hazelcast-wm/src/test/java/com/hazelcast/wm/test/WebFilterSlowTests.java#L181 8 seconds for the cluster.

hazelcast-wm/src/main/java/com/hazelcast/web/ClusteredSessionService.java
jnehlmeier: Are you aware that after `es.scheduleWithFixedDelay()` is called the method `ensureInstance()` is called again? Because `es.scheduleWithFixedDelay()` is called with an initial delay of `0` the method `ensureInstance()` is effectively called twice when the service is created: once synchronously in the constructor and once asynchronously through the executor. 

Maybe it is better to change the ordering: First do it synchronously and then schedule the thread with an initial delay of `CLUSTER_CHECK_INTERVAL` instead of `0`

kobalski: @jnehlmeier Thanks, for your suggestion. You are right, making synchronous call first is better approach.  I accidentally closed this PR, you can see new PR here https://github.com/hazelcast/hazelcast/pull/5832.

PR #5799, opened 2015-07-28
Improved compatibility with JDK9

jerrinot:
JEP 254 introduces compacted String. As a result the value field is no longer char[].

See https://bugs.openjdk.java.net/browse/JDK-8054307 for additional details.


mdogan:
:+1:

pveentjer:
:+1: will merge once green

serkan-ozal:
Until `byte[]` based `UtfWriter` implemented, this is :+1: 
BTW, @jerrinot Instead of `Improved compatibility with JDK9`, I prefer `Fix compatibility issue with JDK9` so I think you should send backport of this one. Because without this fix, it is expected to get `ClassCastException` while trying to cast `byte[]` to `char[]`

PR #5802, opened 2015-07-28
Backport: Improved compatibility with JDK9

jerrinot:
Backport of #5799

Fixed a compatibility issue with JDK9.
JEP 254 introduces compacted String. As a result the value field is no longer char[].

See https://bugs.openjdk.java.net/browse/JDK-8054307 for additional details.
(cherry picked from commit e4b998b)


mdogan:
original PR #5799 is already merged.

PR #5803, opened 2015-07-28
PacketTransceiverImpl removed isAlive check

pveentjer:
The Connection.isAlive check is removed in the transmit method, since the same check is done by the connection when the connection.write is called. So no reason to repeat the logic.


jerrinot:
in other words: checking connection liveness is not a responsibility of the `PacketTransceiver`.
:+1:

mdogan:
:+1:

PR #5810, opened 2015-07-29
removed unneccessary plugin execution from pom

mustafaiman:
Embedded Glassfish starts and stops unnecessarily after test phase even when -DskipTests=true is set.
This pr fixes it.


bilalyasar:
so, where is the code snippet that stops,starts embedded glashfish?

mustafaiman:
JUnit starts glassfish for "@Runwith(Arquillian.class)" annotated classes

kobalski:
:+1:

PR #5814, opened 2015-07-29
Removed unnecessary put/set methods from RecordStore, combined same l…

ahmetmircik:
…ogic into one put method.


ahmetmircik:
Will be send in a different PR. Closing this one.

PR #5821, opened 2015-07-30
Explicit control on io thread count

pveentjer:
Currently you get equal number of io threads for input and output. With this pr you can control how  many threads you actually get. Default behavior remains the same; the existing io thread count mechanism works as default. But 2 additional properties have been added; if they are set, they will be preferred above the old property.

In tests we have seen that the number of threads doesn't need to be in balance.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -101,7 +101,24 @@ public class GroupProperties {
 
     public static final String PROP_VERSION_CHECK_ENABLED = "hazelcast.version.check.enabled";
     public static final String PROP_PREFER_IPv4_STACK = "hazelcast.prefer.ipv4.stack";
+
+    /**
+     * The number of threads doing socket input and the number of threads doing socket output.
+     *
+     * If e.g. 3 is configured, then you get 3 threads doing input and 3 doing output. For indidual control
+     * check PROP_IO_INPUT_THREAD_COUNT and PROP_IO_OUTPUT_THREAD_COUNT.
+     *
+     * Default value is 3 (so 6 threads)
+     */
     public static final String PROP_IO_THREAD_COUNT = "hazelcast.io.thread.count";
+    /**
+     * Controls the number of socket input threads. BY default it is the same as PROP_IO_THREAD_COUNT.
+     */
+    public static final String PROP_IO_INPUT_THREAD_COUNT = "hazelcast.io.input.thread.count";
+    /**
+     * Controls the number of socket output threads. BY default it is the same as PROP_IO_THREAD_COUNT.
+     */
+    public static final String PROP_IO_OUTPUT_THREAD_COUNT = "hazelcast.io.output.thread.count";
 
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -693,6 +712,9 @@ public GroupProperties(Config config) {
         VERSION_CHECK_ENABLED = new GroupProperty(config, PROP_VERSION_CHECK_ENABLED, "true");
         PREFER_IPv4_STACK = new GroupProperty(config, PROP_PREFER_IPv4_STACK, "true");
         IO_THREAD_COUNT = new GroupProperty(config, PROP_IO_THREAD_COUNT, "3");
+        IO_INPUT_THREAD_COUNT = new GroupProperty(config, PROP_IO_INPUT_THREAD_COUNT, "-1");
```
mdogan: Why not just?: 
`IO_INPUT_THREAD_COUNT = new GroupProperty(config, PROP_IO_INPUT_THREAD_COUNT, IO_THREAD_COUNT)`

This way its default value will be value of `IO_THREAD_COUNT`.

hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
```diff
@@ -231,7 +231,20 @@ public boolean getSocketNoDelay() {
     }
 
     @Override
-    public int getSelectorThreadCount() {
+    public int getInputSelectorThreadCount() {
+        int count = node.getGroupProperties().IO_INPUT_THREAD_COUNT.getInteger();
+        if (count != -1) {
```
jerrinot: what if I set it to 0? there should be a check and nice error message.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
pveentjer: Good one

hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
```diff
@@ -231,7 +231,20 @@ public boolean getSocketNoDelay() {
...
+        return node.groupProperties.IO_THREAD_COUNT.getInteger();
+    }
+
+    @Override
+    public int getOutputSelectorThreadCount() {
+        int count = node.getGroupProperties().IO_OUTPUT_THREAD_COUNT.getInteger();
+        if (count != -1) {
+            return count;
+        }
         return node.groupProperties.IO_THREAD_COUNT.getInteger();
```
jerrinot: same here. I know it has been like this since forever but the "boy-scout rule" should apply here.

hazelcast/src/main/java/com/hazelcast/nio/NodeIOService.java
pveentjer: It should not be done here. Group properties should be fixed so they have build in validation. This should not be added at every location you need to use the property.

mdogan:
:+1:

PR #5827, opened 2015-07-31
 Refactored client listeners so that responses can be decoded with correct decoders

sancar:
Removed unused two classes ResponseStream, ResponseHandler. ClientEventInvocation extracted from ClientInvocation and ClientEventFuture extracted from ClientInvocationFuture so that each class focused on specific responsibility.


sancar:
verify

PR #5829, opened 2015-07-31
When an invocation is not send, this will check if someone else dereg…

sancar:
…istred the invocation so that there will be no spurious retry.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -96,10 +96,17 @@ abstract class ClientInvocationServiceSupport implements ClientInvocationService
         ClientMessage clientMessage = invocation.getClientMessage();
         if (!isAllowedToSendRequest(connection, invocation) || !writeToConnection(connection, clientMessage)) {
             final int callId = clientMessage.getCorrelationId();
-            deRegisterCallId(callId);
+            ClientInvocation clientInvocation = deRegisterCallId(callId);
             deRegisterEventHandler(callId);
-            throw new IOException("Packet not send to " + connection.getRemoteEndpoint());
+            if (clientInvocation != null) {
+                throw new IOException("Packet not send to " + connection.getRemoteEndpoint());
+            } else {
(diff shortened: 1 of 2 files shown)


hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -96,10 +96,17 @@ protected void send(ClientInvocation invocation, ClientConnection connection) th
         ClientMessage clientMessage = invocation.getClientMessage();
         if (!isAllowedToSendRequest(connection, invocation) || !writeToConnection(connection, clientMessage)) {
             final int callId = clientMessage.getCorrelationId();
```
pveentjer: are call id's not a long?

sancar: No, they are not. I just checked operation call id it seems they are long, but as far as I remember client call ids were always int.

mdogan:
looks good... I wish you were able to write a test.. :+1:

sancar:
verify

bilalyasar:
run-parallel

bilalyasar:
run-parallel

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -96,10 +96,17 @@ protected void send(ClientInvocation invocation, ClientConnection connection) th
...
-            deRegisterCallId(callId);
+            ClientInvocation clientInvocation = deRegisterCallId(callId);
             deRegisterEventHandler(callId);
-            throw new IOException("Packet not send to " + connection.getRemoteEndpoint());
+            if (clientInvocation != null) {
+                throw new IOException("Packet not send to " + connection.getRemoteEndpoint());
+            } else {
+                if (logger.isFinestEnabled()) {
```
ihsandemir: Do we understand why such a case occurs? Who deregisters the call?

ihsandemir:
OK :+1: Let's change the correlationId to "long" in another PR (new protocol defines correlationId to be uint32).

Issue #5760, opened 2015-07-22
hazelcast-jca-rar/pom.xml references non-existent src/main/rar/ra.xml


juanavelez:
The hazelcast-jca-rar/pom.xml references non-existent src/main/rar/ra.xml. The correct ra.xml location is src/main/rar/META-INF/ra.xml but this is the default per the maven-rar-plugin.

The raXmlFile element in the configuration can be removed
```
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-rar-plugin</artifactId>
                <version>${maven.rar.plugin.version}</version>
                <configuration>
                    <raXmlFile>src/main/rar/ra.xml</raXmlFile>
                    <includeJar>false</includeJar>
                </configuration>
            </plugin>
```

PR #5833, opened 2015-07-31
removed invalid reference

mustafaiman:
Removes reference to non-existent "ra.xml". It is located under "src/main/rar/META-INF/" directory
maven-rar plugin copies everything under "src/main/rar" into ".rar" archive by default. So there is no need to specify the location.

This closes https://github.com/hazelcast/hazelcast/issues/5760


kobalski:
:+1:

pveentjer:
:+1:

PR #5835, opened 2015-07-31
Refactor MemberImpl usage

metanet:
* Address class is exposed by adding getAddress() method to Member interface in 06f7bdd83f28dfbd1cebb734fef8cc79de5ec6c5
* Use ClusterService.getMembers() instead of ClusterService.getMemberList() where only Member.getAddress() is used
* Remove unnecessary AbstractMember casts

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/ClusterService.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/ClusterService.java
@@ -49,13 +49,11 @@ public interface ClusterService extends CoreService {
     /**
      * Gets the collection of members.
      * <p/>
-     * TODO: The name of this method is confusing since it says that a list is returned, but a collection is returned.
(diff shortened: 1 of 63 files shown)


hazelcast/src/main/java/com/hazelcast/cluster/ClusterService.java
```diff
@@ -49,13 +49,11 @@
...
      * Gets the collection of members.
      * <p/>
-     * TODO: The name of this method is confusing since it says that a list is returned, but a collection is returned.
-     * TODO: This method also is a bit of a duplicate since there already is getMembers. So I think this method can be dropped
      * if we take care of the generics.
      *
      * @return the collection of member. Null will never be returned.
      */
-    Collection<MemberImpl> getMemberList();
+    Collection<MemberImpl> getMemberImpls();
```
pveentjer: Do we still need this? Always smelly to expose impl classes.

pveentjer:
:+1:

PR #5836, opened 2015-07-31
Rename of DefaultData to HeapData

pveentjer:
The name of DefaultData doesn't reveal the intent as good as HeapData.


metanet:
:+1:

serkan-ozal:
make sense :+1:

PR #5838, opened 2015-08-01
Remove Packet Data object litter

pveentjer:
To reduce the amount of object litter, the Packet nows extends HeapData instead of encapsulating it. Normally this isn't a preferred design approach; but in this case we can remove 4 object allocations per readonly invocation:

* caller side operation data
* receiver side operation data
* receiver side response data
* caller side response data

For a update operation (and a single sync backup) it will be 8 object allocations per invocation.

All these Data instances are now merged into the Packet instance. There is no difference in what goes over the wire; it is just an internal oo modeling thing.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -312,7 +312,7 @@
          of dependencies on these classes-->
     <suppress checks="ClassDataAbstractionCoupling" files="com/hazelcast/spi/impl/servicemanager/impl/ServiceManager"/>
 
-    <suppress checks="MethodCount"
+    <suppress checks="MethodCount|ExecutableStatementCount"
               files="com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl"/>
     <!-- the invocation just has many parameters because there are a lot of things to tune/-->
     <suppress checks="ParameterNumber"
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -101,8 +101,8 @@ abstract class ClientInvocationServiceSupport implements ClientInvocationService
         }
         registerInvocation(invocation);
         final SerializationService ss = client.getSerializationService();
-        final Data data = ss.toData(invocation.getRequest());
-        Packet packet = new Packet(data, invocation.getPartitionId());
+        final byte[] bytes = ss.toBytes(invocation.getRequest());
+        Packet packet = new Packet(bytes, invocation.getPartitionId());
         if (!isAllowedToSendRequest(connection, invocation.getRequest()) || !connection.write(packet)) {
             final int callId = invocation.getRequest().getCallId();
             deRegisterCallId(callId);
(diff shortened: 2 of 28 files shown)


gurbuzali:
looks good but I don't know how will this change affect the `NativeMemoryData`.
@mdogan @serkan-ozal can you guys take a look ?

pveentjer:
The main difference is that before this PR the bytearray from the Data was acquired when writing the packet to the bb. And now it is acquired when the Packet is created. Imho this is better because it reduces pressure on the IO thread. Slowing down the io thread hurts performance very quickly.

jerrinot:
I have the same concern regarding the `NativeMemoryData` as @ali. Now `Packet is HeapData`. Have you checked impact on Hazelcast Enterprise?
Other than that it's ok.

pveentjer:
@jerrinot of course i have checked enterprise; there is a PR for it.

jerrinot:
:+1: 

a note for future generations: There is a bytearray eventually created even for Off-Heap Data. The old-approach would create the array when data are being written into a socket. This new approach creates the array on packet creation. It makes sense as we should do as little work as possible on the IOThread.

serkan-ozal:
`NativeMemoryData` is not exposed outside of record store and converted to `HeapData`. I have not reviewed the whole PR but from the `NativeMemoryData` perspective, I have not seen a potential issue.

serkan-ozal:
@jerrinot 

Currently, there are a few `byte[]` and object creation for converting response to heap then creating the packet itself. 

I had introduced some improvements while working on my branches while Apple JCache tests but not merged them to Hazelcast repo because of they were risky. To make these improvements, I had added some new interfaces such as `DisposableData`, `SelftConvertableData`, etc ... With these interfaces, HD Cache operations and responses were not consumed heap and they were located on off-heap until they are written to socket in the IO queue. After the packet is destroyed (written to socket or connection closed), the data inside packet was disposed if the data is `DisposableData`. In our case, it was freeing allocated memory from off-heap.

jerrinot:
@serkan-ozal I knew there was a flow where the byte array wasn't created; I just didn't realize it was in your private branch! Thanks for clearing this up! 

I guess now the main question whether this PR isn't too restrictive towards optimizations like you did you the private branch.

pveentjer:
Since these optimizations aren't part of a public release, I'm not worried about them since nobody is using it and these optimization will be benefiting everyone (HZ and HZ Enterprise users).

serkan-ozal:
@jerrinot @pveentjer This PR is a staring point for these kind of optimizations and we can go on these optimizations step by step

jerrinot:
fair enough. I'm just thinking out-loud.

pveentjer:
Perhaps we can even let the 'packet' content be directly written in byte buffers which are recycled; instead of having 1 big byte buffer. I'm using this approach in Project-X. The only thing the io thread does for writing, is to toss the bytebuffers into the socketchannel.

serkan-ozal:
@pveentjer 

> Perhaps we can even let the 'packet' content be directly written in byte buffers which are recycled; 

It is very similar with my `SelfWriteableData`. I had used this approach for writing `NativeMemoryData without converting it to `byte[]` first.

pveentjer:
Cool.. I think that IO is waiting for quite an overhaul anyway. We really need to get started with framing to prevent big packets.

mdogan:
looks good.. :+1:

Issue #5823, opened 2015-07-30
NullPointerException in com.hazelcast.web.WebFilter$RequestWrapper.readSessionFromLocal


cszmajda:
We are seeing this error, seems to be new in Hazelcast 3.5.0:

```
30-Jul-2015 00:47:57.806 SEVERE [http-nio-7990-exec-4] org.apache.catalina.core.StandardWrapperValve.invoke Servlet.service() for servlet [default] in context with path [] threw exception
 java.lang.NullPointerException
        at com.hazelcast.web.WebFilter$RequestWrapper.readSessionFromLocal(WebFilter.java:486)
        at com.hazelcast.web.WebFilter$RequestWrapper.getSession(WebFilter.java:461)
        at com.hazelcast.web.WebFilter$RequestWrapper.getSession(WebFilter.java:392)
...
```

Was this introduced with https://github.com/hazelcast/hazelcast/pull/5471?

bilalyasar:
Do you use session replication with another framework? (like spring) 
also do you have a scenario?

cszmajda:
We are just using Hazelcast session replication, exactly as described in http://hazelcast.com/use-cases/web-session-clustering/generic-web-session-replication/.

It happens intermittently, to some users randomly.  Once a user triggers the problem, it seems to persist until they clear their `hazelcast.sessionId` cookie or the node is restarted. 

Looking at the code, should it not be more careful about a possible `null` return value from `sessions.get(hazelcastSessionId)`?

mesutcelik:
fixed by https://github.com/hazelcast/hazelcast/pull/5840

PR #5840, opened 2015-08-02
Added a minimal null check to WebFilter$RequestWrapper.readSessionFromLocal

cszmajda:
This is intended to fix observed NullPointerExceptions in Hazelcast 3.5.  See https://github.com/hazelcast/hazelcast/issues/5823.


cszmajda:
I'm afraid I'm not 100% sure precisely how to reproduce this issue, except that it definitely started happening when we upgraded from Hazelcast 3.4.2 to 3.5.0. 

Since the line of code that's crashing was introduced in https://github.com/hazelcast/hazelcast/pull/5471, the simplest fix to me seems to be just to insert a null check on this path. 

Suggestions / comments welcome.

PR #5842, opened 2015-08-03
Minor dependency cleanup

pveentjer:
Immediately making use of correct services instead of going through the IOService.

It simplifies the flow of a normal packet entering the system. Instead of going through the IOService, it is immediately being pushed into the PacketTransceiver.


gurbuzali:
:+1:

PR #5843, opened 2015-08-03
Client type is not required to send from client as one of the first bytes anymore

sancar:
It is a field of authentication from now on.


mdogan:
This change only applies to the new client protocol...  
:+1:

PR #5846, opened 2015-08-03
added replicated map to config xml generator

bilalyasar:
so that, management center can show replicatedMap config to users.


mesutcelik:
:+1:

PR #5847, opened 2015-08-03
NotEquals Predicate does not use index

jerrinot:
NotEquals predicate does not use index even when its attribute is indexed.

The `NotEqualsPredicate` has very low selectivity hence the full-map scan is better in vast majority of cases. Moreover the Index implementation is inefficient for `ComparisonType.NOT_EQUAL` as it's traversing the whole index anyway.

The same approach is used by most relation databases - a planner is usually not even trying to estimate cost of indexed access to evaluate NOT_EQUALS predicate and is simply doing a full table scan.


ajermakovics:
:+1:

pveentjer:
:+1:

PR #5851, opened 2015-08-04
Packet: findbugs issue on equals resolved

pveentjer:


gurbuzali:
:+1:

jerrinot:
do you know where are the `equals()/hashCode()` methods used? Is there any performance impact? as the methods may be rather expensive with larger payload.

pveentjer:
Only used for testing.

Issue #5850, opened 2015-08-04
[TEST-FAILURE] ObjectEqualsHashcodeTests.testConnectionFactoryImpl


hasancelik:
```
java.lang.AssertionError: java.lang.ClassCastException: java.lang.Object cannot be cast to javax.naming.RefAddr
For more information, go to: http://www.jqno.nl/equalsverifier/errormessages
	at nl.jqno.equalsverifier.EqualsVerifier.handleError(EqualsVerifier.java:406)
	at nl.jqno.equalsverifier.EqualsVerifier.verify(EqualsVerifier.java:392)
	at com.hazelcast.jca.ObjectEqualsHashcodeTests.testConnectionFactoryImpl(ObjectEqualsHashcodeTests.java:47)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:134)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
       ...
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8/com.hazelcast$hazelcast-jca/643/testReport/junit/com.hazelcast.jca/ObjectEqualsHashcodeTests/testConnectionFactoryImpl/

PR #5852, opened 2015-08-04
mock reference object which equals tester fails to create

mustafaiman:
Reference object this is used by application container to set JNDI reference cannot be mocked by the tester automatically.

This closes https://github.com/hazelcast/hazelcast/issues/5850


mesutcelik:
verify

PR #5853, opened 2015-08-04
Fixed lock lease backup issue - [Backported from #5603]

mdogan:
Fixed a bug which lease-time is not set correctly on backups and when backup is promoted, lease handling process doesn't kicks in. 

Backported from #5603.


gurbuzali:
verify

PR #5855, opened 2015-08-04
At most, poll from the cache invalidation queue as the current size of the queue before start to polling

serkan-ozal:
If there are so many cache invalidation events offered to invalidation queue and the `CacheService::CacheBatchInvalidationMessageSender` task is polling from the queue until it is empty, this may cause `CacheService::CacheBatchInvalidationMessageSender` task polls infinitely from the queue without publishing them. This may occur if the producing invalidation events are too high from the consuming invalidation events.


serkan-ozal:
verify

kobalski:
:+1:

PR #5860, opened 2015-08-04
applyIfAbsent method for Hazelcast CHM has been implemented

toktarev:
applyIfAbsent method for Hazelcast CHM has been implemented

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/ConcurrentReferenceHashMap.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/ConcurrentReferenceHashMap.java
@@ -22,6 +22,7 @@ package com.hazelcast.util;
  * http://creativecommons.org/licenses/publicdomain
  */
 
+import com.hazelcast.core.IFunction;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.io.IOException;
@@ -50,11 +51,11 @@ import java.util.concurrent.locks.ReentrantLock;
  * An advanced hash table supporting configurable garbage collection semantics
  * of keys and values, optional referential-equality, full concurrency of
  * retrievals, and adjustable expected concurrency for updates.
- * <p/>
+ * <p>
  * This table is designed around specific advanced use-cases. If there is any
  * doubt whether this table is for you, you most likely should be using
  * {@link java.util.concurrent.ConcurrentHashMap} instead.
- * <p/>
+ * <p>
  * This table supports strong, weak, and soft keys and values. By default keys
  * are weak, and values are strong. Such a configuration offers similar behavior
  * to {@link java.util.WeakHashMap}, entries of this table are periodically
@@ -67,13 +68,13 @@ import java.util.concurrent.locks.ReentrantLock;
  * <tt>isEmpty</tt> might return a value greater than the observed number of
  * entries. In order to support a high level of concurrency, stale entries are
  * only reclaimed during blocking (usually mutating) operations.
- * <p/>
+ * <p>
  * Enabling soft keys allows entries in this table to remain until their space
  * is absolutely needed by the garbage collector. This is unlike weak keys which
  * can be reclaimed as soon as they are no longer referenced by a normal strong
  * reference. The primary use case for soft keys is a cache, which ideally
  * occupies memory that is not in use for as long as possible.
- * <p/>
+ * <p>
  * By default, values are held using a normal strong reference. This provides
  * the commonly desired guarantee that a value will always have at least the
  * same life-span as it's key. For this reason, care should be taken to ensure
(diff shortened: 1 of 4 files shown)


hazelcast/src/test/java/com/hazelcast/util/SampleableConcurrentHashMapTest.java
```diff
@@ -44,4 +45,29 @@ public void samplesSuccessfullyRetrieved() {
...
+            public String apply(String input) {
+                return "value";
+            }
+        }), "value");
+
+        for (int i = 0; i < 10; i++)
+            new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    assertEquals(map.computeIfAbsent("key", new IFunction<String, String>() {
```
mdogan: Problem here is, test won't fail if this assertion fails (because it's off-loaded to a new thread).

toktarev: Fixed

hazelcast/src/main/java/com/hazelcast/util/ConcurrentReferenceHashMap.java
```diff
@@ -722,11 +722,46 @@ V put(K key, int hash, V value, boolean onlyIfAbsent) {
...
+                while (e != null && (e.hash != hash || !keyEq(key, e.key())))
+                    e = e.next;
+
+                V oldValue;
+                if (e != null) {
+                    oldValue = e.value();
+                } else {
+                    oldValue = null;
+                    ++modCount;
+                    V v = oldValue = function.apply(key);
```
jerrinot: what's the behaviour like if the function returns `null`?

toktarev: Good question
I suppose we should throw NPE ?

hazelcast/src/main/java/com/hazelcast/util/ConcurrentReferenceHashMap.java
```diff
@@ -722,11 +722,46 @@ V put(K key, int hash, V value, boolean onlyIfAbsent) {
                 if (e != null) {
                     oldValue = e.value();
                     if (!onlyIfAbsent)
-                        e.setValue(value, valueType, refQueue);
+                        e.setValue( value , valueType, refQueue);
                 } else {
                     oldValue = null;
                     ++modCount;
-                    tab[index] = newHashEntry(key, hash, first, value);
+                    V v =   value;
```
jerrinot: why you are adding this assignment?

hazelcast/src/main/java/com/hazelcast/util/ConcurrentReferenceHashMap.java
jerrinot: not sure. what's the Doug's implementation doing?

hazelcast/src/main/java/com/hazelcast/util/ConcurrentReferenceHashMap.java
toktarev: public class Test {
        ConcurrentMap<String,String> concurrentMap = new ConcurrentHashMap<>();

        concurrentMap.computeIfAbsent("key", new Function<String, String>() {
            @Override
            public String apply(String s) {
                return null;
            }
        });

        System.out.print("result: "+concurrentMap.get("key"));
    }
}

result: null

jerrinot:
There is one possible issue I can see - I'm not sure how serious is this in reality, but anyway:

Let's assume someone is using the map in a custom code compiled against JDK8. Then following code compiles right now:

```java
map.computeIfAbsent(foo, (i) -> {
  returnSomething(i);
}
```

I believe when you merge this PR then the code mentioned above will no longer compile as it'll become ambiguous and you will have to explicitly cast the lambda to the right interface(?). 

I haven't tested it and I have very little experience with Java8 so chances are I'm wrong. Moreover I'm not sure how much issue is this in reality. Perhaps someone else should have another look? @mdogan ?

toktarev:
We don't use  java.util.Function
We use hazelcast IFunction.

jerrinot:
but it's the same function interface when it comes to lambda? you don't (have to) specify the exact interface. see my code above.

toktarev:
You are right.
Second ..

toktarev:
I've renamed it to callIfAbsent

jerrinot:
what about `applyIfAbsent()` ?

toktarev:
Renamed to applyIfAbsent

jerrinot:
:+1: from me once it's green. 

I'd say at least 2 more reviewers (so 1 extra when compared to a normal PR) are needed, given the code is certainly not simple.

toktarev:
It was computeIfAbsent but there will be confilict with Doug's computeIfAbsent - if we will use it on JVM8 and lambdas.

That is why we renamed it to the applyIfAbsent.

jerrinot:
@mtopolnik: You have reviewed this, haven't you? What's the outcome?

mtopolnik:
:+1:

PR #5861, opened 2015-08-05
Ringbuffer for new client

pveentjer:


hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
```diff
@@ -0,0 +1,248 @@
...
+    @Override
+    protected void onInitialize() {
+        partitionId = getContext().getPartitionService().getPartitionId(getName());
+        final SerializationService serializationService = getContext().getSerializationService();
+
+        readManyAsyncResponseDecoder = new ClientMessageDecoder() {
+            @Override
+            public PortableReadResultSet decodeClientMessage(ClientMessage clientMessage) {
+                Data response = RingbufferReadManyCodec.decodeResponse(clientMessage).response;
+                PortableReadResultSet readResultSet = serializationService.toObject(response);
```
sancar: This usage does not seem to fit to client protocol. 
Only user specific data should come as Data in protocol. Anything primitive or structures encapsulating Data like List, Set , EntryView .. etc can be seen clearly in protocol.
You are gonna need to explain what is PortableResultSet in protocol. We used similar approaches in a couple of places. An example is EntryView 
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java#L76

https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java#L76

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
```diff
@@ -0,0 +1,248 @@
...
+    private static void checkSequence(long sequence) {
+        if (sequence < 0) {
+            throw new IllegalArgumentException("sequence can't be smaller than 0, but was: " + sequence);
+        }
+    }
+
+    @Override
+    protected <T> T invoke(ClientMessage req) {
+        req.setPartitionId(partitionId);
+        return super.invoke(req);
```
sancar: This will be invoked on a random node, and  I think intention as to send to request to owner of partition. 
This will call the 
```
 new ClientInvocation(getClient(), clientMessage).invoke();
```
We need the send via following
```
 new ClientInvocation(getClient(), clientMessage, partitionId).invoke();
```
And setting partition id on message will be done internally by client invocation system.

I think it make sense to add a method as following to ClientProxy for these kind of usages

```
  protected <T> T invoke(ClientMessage clientMessage, int partitionId) {
        try {
            final Future future = new ClientInvocation(getClient(), clientMessage, partitionId).invoke();
            return (T) future.get();
        } catch (Exception e) {
        }
    }
```

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
```diff
@@ -0,0 +1,248 @@
...
+        ClientMessage request = RingbufferReadManyCodec.encodeRequest(
+                getName(),
+                startSequence,
+                minCount,
+                maxCount,
+                toData(filter));
+        request.setPartitionId(partitionId);
+
+        try {
+            ClientInvocationFuture invocationFuture = new ClientInvocation(getClient(), request).invoke();
```
sancar: see https://github.com/hazelcast/hazelcast/pull/5861/files#r36734618

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
```diff
@@ -0,0 +1,248 @@
...
+        for (E e : collection) {
+            throwExceptionIfNull(e);
+            valueList.add(toData(e));
+        }
+
+        ClientMessage request = RingbufferAddAllCodec.encodeRequest(getName(), valueList, overflowPolicy.getId());
+        request.setPartitionId(partitionId);
+
+        try {
+            ClientInvocationFuture invocationFuture = new ClientInvocation(getClient(), request).invoke();
```
sancar: see https://github.com/hazelcast/hazelcast/pull/5861/files#r36734618

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
```diff
@@ -0,0 +1,248 @@
...
+    public ICompletableFuture<Long> addAsync(E item, OverflowPolicy overflowPolicy) {
+        checkNotNull(item, "item can't be null");
+        checkNotNull(overflowPolicy, "overflowPolicy can't be null");
+
+        Data element = toData(item);
+        ClientMessage request = RingbufferAddCodec.encodeRequest(getName(), overflowPolicy.getId(), element);
+        request.setPartitionId(partitionId);
+
+        try {
+            ClientInvocationFuture invocationFuture = new ClientInvocation(getClient(), request).invoke();
```
sancar: see https://github.com/hazelcast/hazelcast/pull/5861/files#r36734618

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
pveentjer: What should I see there?

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
pveentjer: Why do you think it is invoked on a random node? The partition id is set.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
sancar: I was trying to gave a link to my comment. I updated the link.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
sancar: As I explained, your code is calling wrong constructor of ClientInvocation, and if you follow the code, you will see that setting partition id on client message does not guarantee that your request will go to correct node.

hazelcast-client-new/src/test/java/com/hazelcast/client/ringbuffer/RingbufferTest.java
```diff
@@ -0,0 +1,223 @@
...
+    private Ringbuffer<String> clientRingbuffer;
+    private Ringbuffer<String> serverRingbuffer;
+
+    @BeforeClass
+    public static void init() {
+        setLoggingLog4j();
+        Config config = new Config();
+        config.addRingBufferConfig(new RingbufferConfig("rb*").setCapacity(CAPACITY));
+
+        server = Hazelcast.newHazelcastInstance(config);
```
sancar: Is there any reason for not to use mocked network here ?

pveentjer: Can you point me to an example test.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    protected RingbufferAddCodec.RequestParameters decodeClientMessage(ClientMessage clientMessage) {
+        return RingbufferAddCodec.decodeRequest(clientMessage);
+    }
+
+    @Override
+    public Object[] getParameters() {
+        return null;
```
sancar: `add` should return the parameter.value .
And `addAsync` should return parameter.value along with OverflowPolicy.getById(parameters.overflowPolicy))
We had these security checks and permissions for all distributed objects so far, and we tests of these also available in enterprise repo. Is it on your agenda to add these ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public Permission getRequiredPermission() {
+        return null;
+    }
+
+    @Override
+    public String getMethodName() {
+        return "add";
```
sancar: We made two distinct message types to be able to return correct method name for operations like map.put and map.putAsync. You might want to do the same here.

hazelcast-client-new/src/test/java/com/hazelcast/client/ringbuffer/RingbufferTest.java
sancar: Almost all client tests are using it right now. 
Example, https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
pveentjer: #1 add should return the parameter.value .
No idea what you mean. 

#And addAsync should return parameter.value along with 
No idea what you mean

#We had these security checks and permissions for all distributed objects so far, and we tests of these also available in enterprise repo. Is it on your agenda to add these ?  

No. It isn't on my agenda.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
pveentjer: I think that adds to code bloat. Don't know if that is the right way forward.

sancar: I am not really fond of that solution, but this way code is broken. 

Note: Old way was to send a boolean to server just to be able to return correct name, We prefered this solution over that one.

pveentjer: I fail to see what is broken.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
pveentjer: For #1; I guess you mean the content of the arguments. Why is this code not generated? Since it already is part of the parameters fields.

pveentjer: It is unclear from the documentation if getParameters can be null.

```
 @Override
        return null;
    }
```

So is the following valid?

hazelcast-client-new/src/test/java/com/hazelcast/client/ringbuffer/RingbufferTest.java
pveentjer: I have updated the test.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
sancar: Each proxy have as following
https://github.com/hazelcast/hazelcast-enterprise/blob/master/hazelcast-enterprise-client/src/test/java/com/hazelcast/client/security/MapSecurityInterceptorTest.java

If we implement one for RingBuffer, the test will fail, hence it is broken.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
sancar: #So is the following valid?  
It will not make the the supposedly written tests pass. 

Method should be as following for `ringbuffer.add ` :
```
 @Override
        return parameters.value;
    }
```

Method should be as following for `ringbuffer.addAsync ` :
```
 @Override
        return new Object[]{parameters.value, OverflowPolicy.getById(parameters.overflowPolicy) };
    }
```

We talked about we should generate those but we came across to some obstacles to do so. 
1) getParameters should be consistent with user api not with codec template, we should introduce an new annotation that we should put on top of public api. 
2) We need to generate a partial message task , some methods will be generated some methods will be hand-written. 

 2.1)One idea was to make generated code base class and extend from that, but it also did not fit into design. If we would have multiple inheritance in java, we can  implement this. 
.......
I am starting to getting into detail more than I should,  If you are interested we can discuss what we tried and your ideas to generate these parts also.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddMessageTask.java
pveentjer: I added the async versions of the task where needed and renamed where needed.

sancar:
:+1:

PR #5866, opened 2015-08-05
[DO NOT MERGE yet] A new Maven profile to run tests in multiple JVMs in parallel

jerrinot:
This PR introduces a new test category `ParallelTest`. 

When a new Maven profile `parallelTest` is selected then tests in this category will be executed in multiple JVMs in parallel. 

I also moved bunch of slow tests into the `Slow` category so they are not executed with each build. I moved only the tests with running time >30s and only if they cannot be executed in parallel. Moreover I slightly changed the Aggregations tests so they can be executed in parallel now. I also moved bunch of other tests to the `ParallelTest` category where I found it safe. 

Full build now takes less than 19 minutes on C4.4xLarge EC2 machine when the parallelTest profile is activated. 
Here are build times per category:
```
[INFO] Hazelcast Root .................................... SUCCESS [0.387s]
[INFO] hazelcast-code-generator .......................... SUCCESS [1.126s]
[INFO] hazelcast ......................................... SUCCESS [5:32.541s]
[INFO] hazelcast-cloud ................................... SUCCESS [1.751s]
[INFO] hazelcast-client .................................. SUCCESS [3:30.870s]
[INFO] hazelcast-client-new .............................. SUCCESS [3:16.694s]
[INFO] hazelcast-hibernate ............................... SUCCESS [0.220s]
[INFO] hazelcast-hibernate3 .............................. SUCCESS [1:55.815s]
[INFO] hazelcast-hibernate4 .............................. SUCCESS [2:04.291s]
[INFO] hazelcast-wm ...................................... SUCCESS [1:12.304s]
[INFO] hazelcast-spring .................................. SUCCESS [35.157s]
[INFO] hazelcast-ra ...................................... SUCCESS [0.011s]
[INFO] hazelcast-jca ..................................... SUCCESS [3.334s]
[INFO] hazelcast-jca-rar ................................. SUCCESS [0.017s]
[INFO] hazelcast-build-utils ............................. SUCCESS [0.671s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 18:15.475s

```

Known Issue:
We currently have 3 custom JUnit runners:
- `HazelcastParallelClassRunner`
- `HazelcastSerialClassRunner`
- `HazelcastTestRunner`

[...]

PR #5867, opened 2015-08-06
Aggregations tests re-factored so they can be executed in parallel

jerrinot:
I see no reason why Aggregations tests should be ran by SerialRunner. 

Cherry picked from #5866.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/mapreduce/aggregation/AbstractAggregationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/mapreduce/aggregation/AbstractAggregationTest.java
@@ -16,14 +16,10 @@
 
 package com.hazelcast.mapreduce.aggregation;
 
-import com.hazelcast.core.DistributedObject;
 import com.hazelcast.core.HazelcastInstance;
-import com.hazelcast.core.IMap;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.TestHazelcastInstanceFactory;
-import org.junit.After;
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
+import org.junit.Before;
 
 import java.io.Serializable;
 import java.lang.reflect.Array;
@@ -35,31 +31,17 @@ public class AbstractAggregationTest
     private static final int VALUES_COUNT = 10000;
     private static final Random RANDOM = new Random();
 
-    protected static HazelcastInstance HAZELCAST_INSTANCE;
-    protected static TestHazelcastInstanceFactory INSTANCE_FACTORY;
+    protected HazelcastInstance instance;
+    protected TestHazelcastInstanceFactory instanceFactory;
 
-    @BeforeClass
-    public static void startup() {
-        INSTANCE_FACTORY = new TestHazelcastInstanceFactory(2);
-        HAZELCAST_INSTANCE = INSTANCE_FACTORY.newHazelcastInstance();
-        HazelcastInstance hazelcastInstance = INSTANCE_FACTORY.newHazelcastInstance();
+    @Before
+    public void startup() {
+        instanceFactory = createHazelcastInstanceFactory(2);
+        instance = instanceFactory.newHazelcastInstance();
+        HazelcastInstance hazelcastInstance = instanceFactory.newHazelcastInstance();
 
-        assertClusterSizeEventually(2, HAZELCAST_INSTANCE);
+        assertClusterSizeEventually(2, instance);
         assertClusterSizeEventually(2, hazelcastInstance);
-    }
-
-    @AfterClass
-    public static void teardown() {
-        INSTANCE_FACTORY.shutdownAll();
-    }
-
-    @After
-    public void cleanup() {
-        for (DistributedObject object : HAZELCAST_INSTANCE.getDistributedObjects()) {
-            if (object instanceof IMap) {
-                ((IMap) object).destroy();
-            }
-        }
     }
 
     protected static int random(int min, int max) {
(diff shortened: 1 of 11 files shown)


toktarev:
It is ok for me

jerrinot:
@noctarius: Could you please have a look? Thanks!

Donnerbart:
I just had a very quick view at the code. The aggregations themselves seem to be stateless, so there shouldn't be a problem.

I see more a problem with your change from `@BeforeClass` to `@Before`, since you create a HZ instance now for each test. But most tests don't need this, since they are pure unit tests for the aggregation classes.

Also the tests ran fine with sharing the static map before, so there should be no need to change this in a parallalized version. We could ensure this by changing the abstract base class to not provide the `HAZELCAST_INSTANCE`, but a method which returns a random map. This would also de-duplicate this test code:
```
IMap<String, String> map = HAZELCAST_INSTANCE.getMap(mapName);
```

jerrinot:
@Donnerbart I want the tests to be independent that's why each test uses its own HZ instance. The difference is this changeset makes use of mock network - the cluster is formed much faster. 

Also the original tests were running serially - one-by-one - but this PR makes them parallel - the shared state might become a bigger problem as you cannot rely on cleanup, etc...

Donnerbart:
I just gave it a try (the changes are minimal, so easy to repeat).

You seem to be right, the tests fails if I share the HZ instances. Due to the random map names I expected another behavior. But fine, those HZ instances shouldn't make much of a difference.

Nevertheless there was an exception after a the first class completed:
```
Instances haven't been shut down: [HazelcastInstance{name='_hzInstance_10_dev', node=Address[127.0.0.1]:5010}, HazelcastInstance{name='_hzInstance_16_dev', node=Address[127.0.0.1]:5016}, HazelcastInstance{name='_hzInstance_6_dev', node=Address[127.0.0.1]:5007}, HazelcastInstance{name='_hzInstance_8_dev', node=Address[127.0.0.1]:5005}, HazelcastInstance{name='_hzInstance_17_dev', node=Address[127.0.0.1]:5017}, HazelcastInstance{name='_hzInstance_11_dev', node=Address[127.0.0.1]:5011}, HazelcastInstance{name='_hzInstance_12_dev', node=Address[127.0.0.1]:5012}, HazelcastInstance{name='_hzInstance_7_dev', node=Address[127.0.0.1]:5008}, HazelcastInstance{name='_hzInstance_14_dev', node=Address[127.0.0.1]:5014}, HazelcastInstance{name='_hzInstance_15_dev', node=Address[127.0.0.1]:5015}, HazelcastInstance{name='_hzInstance_4_dev', node=Address[127.0.0.1]:5004}, HazelcastInstance{name='_hzInstance_9_dev', node=Address[127.0.0.1]:5009}, HazelcastInstance{name='_hzInstance_1_dev', node=Address[127.0.0.1]:5003}, HazelcastInstance{name='_hzInstance_5_dev', node=Address[127.0.0.1]:5006}, HazelcastInstance{name='_hzInstance_13_dev', node=Address[127.0.0.1]:5013}, HazelcastInstance{name='_hzInstance_20_dev', node=Address[127.0.0.1]:5020}, HazelcastInstance{name='_hzInstance_19_dev', node=Address[127.0.0.1]:5019}, HazelcastInstance{name='_hzInstance_18_dev', node=Address[127.0.0.1]:5018}, HazelcastInstance{name='_hzInstance_2_dev', node=Address[127.0.0.1]:5002}, HazelcastInstance{name='_hzInstance_3_dev', node=Address[127.0.0.1]:5001}]

	at com.hazelcast.test.AbstractHazelcastClassRunner$1.evaluate(AbstractHazelcastClassRunner.java:189)
```

Adding this fixed the problem on my machine:
```
    @After
    public void teardown() {
        instanceFactory.shutdownAll();
    }
```

BTW: Runtime on my machine is about 4 minutes for the whole aggregration package.

Donnerbart:
Hmm, I just reverted the changes and ran the tests serialized: just about 2 minutes.

Maybe in this case the overhead of creating the HZ instances outweighs the effect of the parallel execution.

jerrinot:
@Donnerbart: Try it on top of #5896
Moreover the mock network allows the new fast PR builder to run this test in parallel with other tests.

Donnerbart:
Hmm, 4 to 5 minutes with more threads. I'm not sure these tests benefit that much from parallel execution. I repeated the serial execution and it's again 2 to 3 minutes.

I had a detailed look at the execution times. Most single tests take about 1 second on my machine in serial execution. There are just 2 or 3 tests which take up to 7 seconds. We have about 10 tests per class. So we have about 10 to 17 seconds of actual test time per class.

The start of the mocked instance takes about 1100 ms as well (1069 to 1726 ms). So we're adding about 11 seconds to each test class for the mocked instances. Due to the short test runs this doesn't seem to work out.

Also the mock network was already used before, wasn't it? So this should not prevent to execute even the serialized version in parallel with other tests.

I think we need a Runner for serialized test methods, but parallel test classes. This will basically work for every test which uses the mocked instances, but will suffer from the overhead of creating it for every method.

jerrinot:
@Donnerbart: Obviously we could use the SerialTestRunner, the mock network and still use JVM-level parallelism for fast merge builds. 

However I value when tests are independent and don't interfere with each other. Anyway, I don't have too strong opinion here so I can send an update if you think the value of independent test is outweighted by somewhat longer runs.

gurbuzali:
how about splitting the tests so that tests which do not need an `instance` don't create it ?

jerrinot:
@gurbuzali: I checked a few tests from the PR and it seems they all rely on `instance`. 

I'm closing this PR for now. The aggregation tests already use mock network so the new "Fast PR Builder" can still run it concurrently with other tests.

PR #5868, opened 2015-08-06
added isEmpty control to doDeleteSession

bilalyasar:


pveentjer:
:+1:

jerrinot:
@bilalyasar: Could you please include reasons for this change in a commit message? It's usually more helpful to describe WHY you are doing something than WHAT you are doing.

bilalyasar:
thanks, i updated the commit message and enabled the test

PR #5873, opened 2015-08-06
Fixes MapListener addition issue via EntryListenerConfig 

### LINKED ISSUE #5648: Not possible to add MapClearedListener to EntryListenerConfig as an implementation

ahmetmircik:
closes #5648 

Fixed two issues: first one is addition of a MapListener impl. via EntryListenerConfig object and second one is a left-over which prevents setting of instance to HazelcastInstanceAware listeners

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/EntryListenerConfigTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/EntryListenerConfigTest.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


ahmetmircik:
verify

gurbuzali:
:+1:

hazelcast/src/test/java/com/hazelcast/map/EntryListenerConfigTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.junit.runner.RunWith;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Keep this test serial. It relays on static shared variables.
```
mdogan: Why don't you convert statics to instance variables and use parallel testing?

ahmetmircik: Because i am creating a new listener from the class-name and want to reference to a field of that new instance in order to count fired events.

mdogan: I see. what about using some reflection magic to access internal listener instance? Is it worth it?

ahmetmircik: current approach seems simple to me.

mdogan:
:+1:

PR #5878, opened 2015-08-06
Slow tests moved into appropriate category

jerrinot:


pveentjer:
+1 once merged.

mesutcelik:
@jerrinot ,

What was the criteria to move those tests to `SlowTest`?
I think each test case has to be checked in this PR because we could have been missing some test cases which should be executed in PR Builder?

jerrinot:
@mesutcelik: 
I moved only tests running over 20s and only if they could not be paralleled well. 

There are two changes I'm not very sure about:
https://github.com/hazelcast/hazelcast/pull/5878/files#diff-f3de949df03ab3752d7cc6ea31e88846L53
https://github.com/hazelcast/hazelcast/pull/5878/files#diff-3731f3bd585ef495aaa6870d088fa46fL47

In theory these two tests should run in multiple JVMs just fine. However I was getting failures, perhaps it's a spurious failure/race triggered by a more powerful machine?

mesutcelik:
> I moved only tests running over 20s and only if they could not be paralleled well.

that is ok but what i mean is some test cases inside those test classes should still be running as QuickTest in PR Builder. For instance, `TcpIpJoinTest` and `MulticastJoinTest`. Noticing that they are failing the other day after the merge could be a bit late.

> There are two changes I'm not very sure about:
> https://github.com/hazelcast/hazelcast/pull/5878/files#diff-f3de949df03ab3752d7cc6ea31e88846L53
> https://github.com/hazelcast/hazelcast/pull/5878/files#diff-3731f3bd585ef495aaa6870d088fa46fL47

> In theory these two tests should run in multiple JVMs just fine. However I was getting failures, perhaps > it's a spurious failure/race triggered by a more powerful machine?

lets just keep them as QuickTest if they are not so slow or make them parallel and create a test failure issue.

jerrinot:
@mesutcelik: The idea here was that we rarely do changes as low to affect join mechanism specifically. The MulticastJoinTest takes 44s on my box. 

From the other hand I understand it's crucial part of functionality so I'm open to revert it.

mesutcelik:
We can divide such tests into `xxxTest` and `xxxSlowTest`(slow parts). Can we do this at least for joiner tests? For future generations, we should consider this approach more often...

jerrinot:
the problem is the Joiner integration tests are slower by theirs nature. 
I see some room anyway: There are tests checking whether 2 members with different configuration will refuse to form a cluster. IMHO the compatibility is not a responsibility of Joiner anyway and it could be tested with mock network.

jerrinot:
I'm closing it for now. As the tests affected truly rely on real networks and I don't see a quick way how to make them faster.

PR #5879, opened 2015-08-06
Packet bind request goes through transceiver

pveentjer:
The DefaultPacketReader had too many responsibilities. 

Before:
When a bind request comes in, it deserializes it and calls the right method on the TcpIpConnectionManager.

After
Now the reader passes all packets through the PacketTransceiver and here the packet
is send tot he connection manager where the connectionmanager deals with it. So
the bind request packet now follows the same flow as other packets.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
@@ -30,6 +30,7 @@ import com.hazelcast.internal.storage.Storage;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.LoggingServiceImpl;
 import com.hazelcast.nio.Address;
+import com.hazelcast.nio.Packet;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.nio.serialization.SerializationService;
 import com.hazelcast.partition.InternalPartitionService;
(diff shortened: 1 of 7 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -107,10 +109,23 @@ public NodeEngineImpl(Node node) {
...
                 executionService);
         quorumService = new QuorumServiceImpl(this);
     }
 
+    class ConnectionManagerPacketHandler implements PacketHandler {
+        // ConnectionManager is only available after the NodeEngineImpl is available. This is a
+        // way around this problem.
+        @Override
+        public void handle(Packet packet) throws Exception {
+            ConnectionManager connectionManager = node.getConnectionManager();
```
pveentjer: Once this PR is merged, I'll cast to TcpConnectionManager instead of checking for PacketHandler. I don't want to wait for another 1.5 hours build.

jerrinot: :+1:

jerrinot:
:+1:

mdogan:
Causes failure on enterprise:

>hazelcast-enterprise/hazelcast-enterprise/src/main/java/com/hazelcast/nio/tcp/SymmetricCipherPacketReader.java:[23,72] error: cannot find symbol

```java
public SymmetricCipherPacketReader(TcpIpConnection connection, IOService ioService) {
        super(connection, ioService.getSerializationService(), ioService.getPacketTransceiver());
...
```

pveentjer:
https://github.com/hazelcast/hazelcast-enterprise/pull/302

PR #5882, opened 2015-08-06
Cleanup of deprecated JUnit imports

Donnerbart:
`junit.framework` is deprecated, we should use `org.junit` instead.


pveentjer:
+1

PR #5883, opened 2015-08-07
Fix #5857: Add threadId to MapExecuteOnKeyRequest

### LINKED ISSUE #5857: MapExecuteOnKeyRequest should reference requester's threadId
ClientMapProxy and MapProxyImpl differ slightly in behaviour for the following methods:
* executeOnKey
* submitToKey (2 overloads)

In MapProxyImpl, these create an EntryOperation with the threadId set. This does not happen with ClientMapProxy. MapExecuteOnKeyRequest should hence take a threadId and set this on the generated EntryOperation.

This matters when a process explicitly locks a key in a distributed map, and then proceeds to invoke executeOnKey or submitToKey. This will work without issue for full members for the reasons described above, but may not work for clients.

LoneRifle:
* Add threadId as a long to MapExecuteOnKeyRequest, which should tag the
request as the id of the current thread, typically from ThreadUtils.getThreadId. Similarly, add threadId to the client protocol MapCodecTemplate methods relating to executeOnKey and submitToKey.

* Set threadId in the EntryOperation returned by MapExecuteOnKeyRequest.prepareOperation, MapSubmitToKeyRequest.prepareOperation and MapExecuteOnKeyMessageTask.prepareOperation.

* Supply threadId in ClientMapProxy when creating such requests.

* hazelcast-client-new/ClientMapProxy: Fix broken contract on
executeOnKey where the raw ClientMessage was being returned, by decoding
the message, retrieving the response, and deserializing the embedded
payload from Data to Object.

* provide unit-tests that prove all of the above.

Fixes #5857 and broken IMap contract, as described above.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -176,7 +176,7 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
             return (T) MapRemoveAsyncCodec.decodeResponse(clientMessage).response;
         }
     };
-
+    
     private static final ClientMessageDecoder submitToKeyResponseDecoder = new ClientMessageDecoder() {
         @Override
         public <T> T decodeClientMessage(ClientMessage clientMessage) {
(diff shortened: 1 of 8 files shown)


LoneRifle:
Hello Hazelcast,

As directed, I have cleaned up the previous pull request and provided unit-tests as well. While I was at it, I discovered one other defect where the new ClientMapProxy has broken its IMap.executeOnKey contract by returning the raw ClientMessage. If you want, I can create a new issue that describes this, and link it to this PR.

sancar:
Thanks for the fix and your efforts. You seem right,  IMap.executeOnKey for new client protocol implementation seem broken. Mentioning this with another issue would be great. 

And I found another issue while I was reviewing pr. The following also passes raw client message to callback. Would you like to fix this one also ? 
```
 public void submitToKey(K key, EntryProcessor entryProcessor, final ExecutionCallback callback) 
```

LoneRifle:
I'm currently busy for the next few days, so it will be a week or so before I can have a look. Will open a second issue to link to this PR along with the requested changes.

sancar:
:+1:

sancar:
I will fix the issue about submitToKey callback then. Thanks for the fix again.

PR #5888, opened 2015-08-07
Fixed Long2LongHashMap.size()

mdogan:
Size field should be reset after rehash() to avoid counting current entries multiple times.

Added size checking tests for `Long2LongHashMap`, `Int2ObjectHashMapTest` and `Long2ObjectHashMapTest`.

Fix to upstream: https://github.com/real-logic/Agrona/pull/19


jerrinot:
:+1:

pveentjer:
:+1:

PR #5894, opened 2015-08-08
Minor PacketTransceiver cleanup

pveentjer:
Added tests for the receive method.
Moved method down
Renamed packet handler fields.


jerrinot:
:+1:

PR #5859, opened 2015-08-04
Calculcation of MAX_THREADS for Parallel Runner has been customized

toktarev:
Calculcation of MAX_THREADS for Parallel Runner has been customized.

PR #5465, opened 2015-06-08
Amount of parallel threads has been reduced to prevent test timeout exceeds

toktarev:
MAX_THREADS in the HazelcastParallelClassRunner has been reduced to
Math.min(Runtime.getRuntime().availableProcessors()/2, 1)

PR #5896, opened 2015-08-08
Default `treadCount` for parallel test runner reverted to `max(coreCount, 8)`

jerrinot:
Default `treadCount` for parallel test runner reverted back to `max(coreCount, 8)`

It was like this originally, but it was changed by #5465 to prevent spurious MapReduce tests failures.
However since #5859 we are able to customize threadCount on per-test basis and there is no need to
slow-down all other tests just because of MapReduce.


jerrinot:
verify

pveentjer:
:+1:

toktarev:
:+1:

PR #5897, opened 2015-08-09
tcp: improved testing/javadoc

pveentjer:
Improvements
* explicit testing of TcpIpConnection
* explicit testing of TcpIpConnectionManager
* explicit testing of DefaultPacketReader
* explicit testing of DefaultSocketChannelWrapperFactory
* improved javadoc connection manager
* stress tests for sending packets to make sure we don't loose them
* removed MetricsRegistry from IO Service; pass it directly to TcpIpConnectionManager


pveentjer:
verify

jerrinot:
:+1: 

A minor side-note - I'm not sure about the terminology as a TDD purist (hello @ajermakovics ;-) ) could say your `MockIOService` is not really a mock, but I'm good with that.

pveentjer:
I'll rename it to stub :)

PR #5898, opened 2015-08-09
Got rid of some unnecessary suppressions.

Donnerbart:
Not much to say about this. I checked the suppressions for `PortableHook`, `SerializerHook` and `ConsoleApp` and removed those, which were not needed (anymore).

I also made the suppressions for `ConcurrentReferenceHashMap` more specific, since there are more complexity checks than `NPath` and `Cyclomatic`.

The PR builder will tell us if these modifications are okay.


pveentjer:
:+1: 

Will merge once green

PR #5899, opened 2015-08-09
Cleanup of ClusterServiceImpl

Donnerbart:
This PR will be a bit more fun after the very simple cleanups before. I mostly fixed simple CheckStyle issues like magic numbers, 130 line width, trailing comments etc. I also aligned the format of comments and log messages. After that I took care about NPath and Cyclomatic complexity and pulled out some methods or adjusted the code flow to get rid of those issues.

I have quite high confidence that I didn't break anything. A lot of changes were done via IDEA refactoring tools. I ran the cluster tests 3 times including code coverage. There where no failures with 84% line coverage. The biggest untested part is the `ping()` method, which I didn't change significantly. The remaining untested parts are mostly the `toString()` method and logging branches (see https://hazelcast-l337.sonar.cloudbees.com/resource/index/11063?display_title=true&metric=coverage).

I hope I didn't do any controversial things. There maybe things which we can discuss, but I hope we can agree that the class is in better shape than before with this PR.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -42,6 +42,7 @@ import com.hazelcast.core.MemberAttributeEvent;
 import com.hazelcast.core.MembershipEvent;
 import com.hazelcast.core.MembershipListener;
 import com.hazelcast.instance.BuildInfo;
+import com.hazelcast.instance.GroupProperties;
 import com.hazelcast.instance.HazelcastInstanceImpl;
 import com.hazelcast.instance.LifecycleServiceImpl;
 import com.hazelcast.instance.MemberImpl;
@@ -111,6 +112,7 @@ import static com.hazelcast.util.FutureUtil.logAllExceptions;
 import static com.hazelcast.util.FutureUtil.waitWithDeadline;
 import static com.hazelcast.util.Preconditions.checkNotNull;
 import static com.hazelcast.util.Preconditions.isNotNull;
+import static java.lang.String.format;
 import static java.util.Collections.unmodifiableMap;
 import static java.util.Collections.unmodifiableSet;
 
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -919,28 +952,6 @@ public void merge(final Address newTargetAddress) {
         }
     }
 
-    private <V> V waitOnFutureInterruptible(Future<V> future, long timeout, TimeUnit timeUnit)
```
Donnerbart: Moved this method down to the bottom into the `private class MergeTask` where it is used.

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -120,93 +122,112 @@
...
-        heartbeatIntervalMillis = heartbeatInterval > 0 ? heartbeatInterval * 1000L : 1000L;
-        pingIntervalMillis = heartbeatIntervalMillis * 10;
+
         node.connectionManager.addConnectionListener(this);
 
         registerMetrics();
     }
 
-    void registerMetrics() {
+    private static long getAsMilliSeconds(GroupProperties.GroupProperty groupProperty) {
```
Donnerbart: Methods like this maybe a candidate for a Util class, but I didn't pull out any method from this class during this refactoring step.

pveentjer:
I'll give you a tip with such classes. Don't do so many changes in 1 go. Cluster service is one of the most complex classes in the system and totally relies in implicit testing. So what I do, and it is very time consuming, is make a PR with just a few minor changes. But generate a whole sequence of them. This way you can transform the system one step at a time.

Donnerbart:
Yeah, I think none of those `checks=""` classes will be fun to touch and all have a chance to create huge PRs. For classes like this we maybe should do pair-programming with 3 people and merge it right away after the PR builder is green. This may save a lot of time, but of course is difficult in a distributed company like Hazelcast.

pveentjer:
I would really go for a continuous stream of small changes on such classes. Why do you think I have so many commits? :)

jerrinot:
I guess one PR could be ok, but a PR can contain multiple smaller commits - it'll make it easier to review. 
I'll check this one as it is as splitting it up would just increase a risk of a mistake.

jerrinot:
:+1:

pveentjer:
@Donnerbart what are we going to do with this pr? Perhaps @mdogan can have a look at it. My knowledge in the cluster code is not sufficient to review this PR.

Donnerbart:
Knowledge of the cluster code should not be necessary to review this PR. It's about resolving CheckStyle issues, not the logic of the cluster :)

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
gurbuzali: :+1:

gurbuzali:
verify

gurbuzali:
will merge once green

PR #5903, opened 2015-08-10
Minor SocketAcceptor cleanup

pveentjer:
Converted SocketAcceptor from Runnable to SocketAcceptorThread
Pulled SocketAcceptor shutdown logic into SocketAcceptorThread
Pulled out main selector loop on SocketAcceptThread into own method

Depends on this PR:
https://github.com/hazelcast/hazelcast/pull/5897

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
@@ -59,7 +59,7 @@ public class SocketAcceptorThread extends Thread {
             selector = Selector.open();
             serverSocketChannel.configureBlocking(false);
             serverSocketChannel.register(selector, SelectionKey.OP_ACCEPT);
-            mainLoop(selector);
+            acceptLoop(selector);
         } catch (OutOfMemoryError e) {
             OutOfMemoryErrorDispatcher.onOutOfMemory(e);
         } catch (IOException e) {
@@ -69,7 +69,7 @@ public class SocketAcceptorThread extends Thread {
         }
     }
 
-    private void mainLoop(Selector selector) throws IOException {
+    private void acceptLoop(Selector selector) throws IOException {
         while (connectionManager.isLive()) {
             // block until new connection or interruption.
             final int keyCount = selector.select();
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
```diff
@@ -51,26 +59,7 @@ public void run() {
...
-                while (it.hasNext()) {
-                    final SelectionKey sk = it.next();
-                    it.remove();
-                    // of course it is acceptable!
-                    if (sk.isValid() && sk.isAcceptable()) {
-                        acceptSocket();
-                    }
-                }
-            }
+            mainLoop(selector);
```
jerrinot: that just feels wrong. maybe because it's not a verb?

pveentjer: What do you suggest? 

Personally I think it makes sense since the main accept event loop is executed there.

jerrinot:
I don't quite grasp the shutdown sequence. When I call `TcpIpConnectionManager::shutdown`` then it'll first try to shutdown the acceptorThread.( `SocketAcceptorThread.shutdown()`). 

If I read the code correctly then it'll interrupt accepter thread and will try try to `join()` it. However the acceptor's  `mainLoop()` in active until the `manager.isAlive()` returns true. Perhaps I'm reading it wrong, but I reckon the `join()` will always timeout?

pveentjer:
I'll have a look. Perhaps this was already part of the system; I just moved the shutdown logic to the right place and now the issue is perhaps more obvious.

jerrinot:
that's very possible.

mdogan:
@jerrinot: see https://github.com/hazelcast/hazelcast/pull/5903/files#diff-ea6a789c729a89905029f0951bbdccbcR74

```java
while (connectionManager.isLive()) {
            // block until new connection or interruption.
            final int keyCount = selector.select();
            if (isInterrupted()) {
                break;
            }
....
```

jerrinot:
@mdogan: Thanks!

pveentjer:
The issue with the shutdown has been resolved. Apart from the loop name (which I'll directly push into master instead of waiting for another 2h build) is there anything else that prevents you from giving the green light.

jerrinot:
:+1:  (with the `acceptLoop()`). 

it's better to push the change here and merge without waiting to Green.

PR #5904, opened 2015-08-10
com.hazelcast.util.ConcurrentMap has been renamed to com.hazelcast.util.IConcurrentMap

toktarev:
com.hazelcast.util.ConcurrentMap has been renamed to com.hazelcast.util.IConcurrentMap


pveentjer:
Can't we give this interface a different name? Having a class with a name that is well known can be very confusing; especially if you rely on on the fly imports.

toktarev:
Which name do you think is good here ?

toktarev:
IConcurrentMap ?

pveentjer:
I think that is fine since it is a private class anyway. We can always rename it if we come up with a better name.

toktarev:
Done

pveentjer:
:+1: 

Can be merged once the build is green.

pveentjer:
verify

PR #5907, opened 2015-08-10
Renaming in non blocking package

pveentjer:
IOSelector --> deleted
AbstractIOSelector --> NonBlockingIOThread (so there is now a dependency on NonBlockingIOThread).
InputIOSelector -->NonBlockingInputThread
OutputIOSelector --> NonBlockingOutputThread

The naming was confusing because the IOSelector is not a selector at all. It is the thread that blocks on the selector and calls the handlers based on the results.

Probably NonBlockingInputThread/NonBlockingOutputThread will be deleted in the future. It can all be integrated into the NonBlockingIOThread.


jerrinot:
:+1: 

The input & output thread are divided on purpose. I believe @mdogan did in in the past as it yielded better results for him?

pveentjer:
I'm fine with dividing them on purpose, but it can be a configuration setting. But we only need a single generic NonBlockingIOThread that can take care of reads/writes.

pveentjer:
verify

pveentjer:
verify

pveentjer:
verify

mdogan:
:+1: 

PS: the biggest problem with this kind of continuous renaming/refactoring is, we are losing the history of the classes. I know it's possible to track with some `git` magic but still I even had hard times to remember which class is where...

pveentjer:
Thanks. In this case we don't loose the history of classes. 

But in splitting up the io threading model we did. especially read/write handlers.

PR #5910, opened 2015-08-10
[BACKPORT] At most, poll from the invalidation queue as the current size of the queue before start to polling

serkan-ozal:


kobalski:
:+1:

serkan-ozal:
verify

PR #5911, opened 2015-08-10
Moved iobalancer to nonblocking package

pveentjer:
The io balancer only has meaning for the non blocking threading model. So it is moved into that package.


jerrinot:
:+1:

pveentjer:
verify

PR #5912, opened 2015-08-10
Queue.addAll in new client should not use Set in the codec

cangencer:
Currently `Queue.addAll()` in new client uses `Set` in the codec, which causes items to be de duplicated and also lose the ordering when added. This has been changed to use `List`. We can further discuss if we want to change the other methods (`removeAll`, `retainAll` etc.) to use `List` as well.


sancar:
:+1:

enesakar:
verify

PR #5540, opened 2015-06-18
Unsafe disabled on platforms where unaligned access to memory is not allowed

jerrinot:
Fixes #5518

I appreciate any advice how to test it. Static initializers are tricky beasts. I could imagine some classloader magic, but it could cause more harm than good.

PR #5915, opened 2015-08-11
[Backport] Unsafe disabled on platforms where unaligned access to memory is not allowed.

jerrinot:
Backport of #5540
Unsafe disabled on platforms where unaligned access to memory is not allowed.
(cherry picked from commit 340017a)


serkan-ozal:
:+1:

gurbuzali:
verify

PR #5924, opened 2015-08-11
ClientCacheContextTest changed to use mock network

jerrinot:
1. ClientCacheContextTest(s) refactored to use mock network as it's faster and it allows to run test in parallel with other.

2. Both ClientCacheContextTest and CacheContextTest switched to use ParallelTestRunner as with mock network there is no need to run it sequentially.


serkan-ozal:
:+1:

pveentjer:
:+1:

PR #5925, opened 2015-08-11
ClientTxnMapProxy.values() should return List instead of Set

cangencer:
When returning the values `ClientTxnMapProxy.values()`should use List instead of Set


sancar:
@cangencer Could you add test to make sure we don`t break this again ?

pveentjer:
@cangencer can you have a look at this pr? Once the test is added, imho the code is good to go.

cangencer:
@pveentjer @sancar There was an issue on the server as well with the values being returned as a set when a predicate is supplied. I have fixed this as well and added a test for both.

sancar:
@cangencer ClientTxnMapProxy for old client also needs to be updated. Can you fix and add the same test in old client package ? 
And we are gonna need to have backport for maintenance-3.x

sancar:
verify

cangencer:
verify

pveentjer:
Can you squash the commits? After that it is a :+1:  from me

sancar:
:+1:

PR #5928, opened 2015-08-11
Regression test which does have to use real network moved into own class

jerrinot:
This is the only test from the ClientRegressionTest which does rely on real network.

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/ClientListenersTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/ClientListenersTest.java
@@ -48,7 +48,7 @@ public class ClientListenersTest extends HazelcastTestSupport {
             @Override
             public Portable create(int classId) {
                 if (classId == 6) {
-                    return new ClientRegressionTest.SamplePortable();
+                    return new ClientRegressionWithMockNetworkTest.SamplePortable();
                 }
                 return null;
             }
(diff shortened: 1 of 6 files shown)


Donnerbart:
Does it have to be a `SlowTest` (so moved out of PR-builder if I'm right) or is it enough to use the real network again?

jerrinot:
This very test is not particularly slow on its own. If you have better idea how to structure it then I'm open to suggestions!

Donnerbart:
My main criterias would be:
* importance of the test, as in:
  * importance of the component the test is testing
  * probability of breaking something which the tests is testing
* runtime of the test

Some tests may be important, but are nevertheless too slow/expensive for the PR-builder. Those would be candidates to move to `SlowTest`s to keep the PR-builder fast.

For this test I have no idea about the importance. I hope the client team can help out with their opinion and give advice where we should put this test in (PR-builder or daily run).

jerrinot:
@sancar: May I kindly ask you for a suggestion here? Thanks!

sancar:
It seems this test is written to prove an issue is fixed, hence it is important. Since it is fast, I would like to see this test in pr builder.

jerrinot:
@sancar: I pushed a new version where both test classes are in the QuickTest category. Feel free to suggest better names!

Donnerbart:
The names may not be beautiful, but they are very clear in their meaning.

:+1:

pveentjer:
verify

pveentjer:
:+1:

PR #5929, opened 2015-08-11
Old client Ringbuffer partition fix

pveentjer:
It now sends to correct member instead of random one (partition id was not set correctly).

Also upgraded ringbuffer test to make use of connecting mocking


sancar:
:+1:

PR #5930, opened 2015-08-12
PacketTransceiver simplified

pveentjer:
The PacketTransceiver has been simplified to only dispatching to the right service.
It doesn't have logic anymore trying to resend a packet. That has been moved to the
ConnectionManager.

The PacketTranseiver has been renamed to PacketDispatcher

Explicit tests for the transmit logic have been added to the TcpIpConnectionManager.

Also the retry functionality is now scheduling on its own scheduler in the TcpIpConnectionManager instead of relying on a central shared one. This provides better isolation.


jerrinot:
I'd like to have the re-transmission logic elsewhere as I think it's a nice responsibility. I talked about this with @pveentjer and we came to a conclusion so more cleanups are needed first. So :+1: from me

PR #5934, opened 2015-08-12
fixes #5212. testNearCacheFasterThanGoingToTheCluster is moved to nightly

### LINKED ISSUE #5212: [TEST-FAILURE] ClientMapNearCacheTest Failures
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-sonar/changes
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/3090/console
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/3091/testReport/
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.7/538/

sancar:
, because of its nature it needs to be on real network and not critical enough to be tested in each pr

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapNearCacheSpeedTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapNearCacheSpeedTest.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.map;
+
+import com.hazelcast.client.HazelcastClient;
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.config.NearCacheConfig;
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IMap;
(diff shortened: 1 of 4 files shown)


hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapNearCacheSpeedTest.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        long readFromClusterTime = System.currentTimeMillis() - begin;
+
+        begin = System.currentTimeMillis();
+        for (int i = 0; i < size; i++) {
+            map.get(i);
+        }
+        long readFromCacheTime = System.currentTimeMillis() - begin;
+
+        assertTrue("readFromCacheTime > readFromClusterTime", readFromCacheTime < readFromClusterTime);
```
jerrinot: this is very fragile by nature. perhaps it would help a bit to call `System.gc()` on lines 76 and 82?

sancar: I felt to same way, I want to remove this, but I saw that we had these tests since we first write client near cache.
We have tests making sure we are hitting near cache. This is just to make sure it is faster.
 If everyone OK with that I can happily remove this test. 

The test starting to fail after we replaced the network with mocked one because time difference is much closer.

jerrinot: what about monitoring number of executed operations on members?

sancar: We have the test to make sure near cache is hit or not. Your suggestion sound like another way of doing same thing.
see
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapNearCacheTest.java#L288

If your gonna measure how many call per second that will help, but when near cache is hit there are no operations running and there is nothing to compare.

sancar: I discussed this with @mdogan , it much more make sense to just remove this test. I will close the issue and remove the test with direct commit.

jerrinot: if there is another test to make sure the near cache is hit then I'd lean towards removing this one.

PR #5935, opened 2015-08-12
Renamed the readers/writers to be a bit more predictable

pveentjer:
Also added additional tests for a bunch of them.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
@@ -35,8 +35,8 @@ import com.hazelcast.nio.MemberSocketInterceptor;
 import com.hazelcast.nio.serialization.SerializationService;
 import com.hazelcast.nio.serialization.SerializationServiceBuilder;
 import com.hazelcast.nio.serialization.impl.DefaultSerializationServiceBuilder;
-import com.hazelcast.nio.tcp.DefaultPacketReader;
-import com.hazelcast.nio.tcp.DefaultPacketWriter;
+import com.hazelcast.nio.tcp.MemberPacketReader;
+import com.hazelcast.nio.tcp.MemberPacketWriter;
 import com.hazelcast.nio.tcp.DefaultSocketChannelWrapperFactory;
 import com.hazelcast.nio.tcp.PacketReader;
 import com.hazelcast.nio.tcp.PacketWriter;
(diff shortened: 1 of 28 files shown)


sancar:
I understand that you want to choose better names, but it is really hard to keep track of all the changes.  I think we should rename only if it causes a total misunderstanding. Whether it is good or bad, we are getting accustomed to names. In current state, we are changing names a lot and I had to look up a couple of places to get the code I am searching for, while before It was all in my memory.

pveentjer:
Hi Sancar, I understand your worried but the naming for me was very hard to follow. Every time i ended up in the wrong classes and I'm sure other engineers that need to deal with the tcp package are going to run into the same issues. So probably better to bite the bullet and have the naming fixed so that future engineers don't need to waste a lot of time. 

I think I'm done with most of the renaming.

sancar:
It feels like your renamings will not end soon. For example we intrdouced SocketClientMessageReader.java in 3.5, we get used to it and now you are changing it. Is there any guarantee that - for example in 3.7- you will not decide to change something I introduced in 3.6 ? 
I think, these are minor ones anyway but the core packages that is there since eternity changed a lot. 

I there was not much of a complain to change it anyway. I mean, if there were no side effect of changing it, I would also support changing to better names. But invalidating soo many engineers memory should prevent us from changing a lot.

pveentjer:
I clean op modules one at a time. I spend a lot of time on spi.impl and things are pretty stable and clean. I hope you agree that code there is easy to locate. 

I'm currently working on tcp since here is a probably a lot to gain performance wise and I want to have the code clean and well tested before I'm going to do more tweaking.

Your client reader/writer functionality is imho easy to understand:

screenshot_2015-08-12_14-59-17

It is close together. It follows predictable naming:
ClientXXXSomeInterface

pveentjer:
For the members you have now a similar symmetry. I like this; since it makes everything predictable. I don't need to read to code the understand what is happening. 

screenshot_2015-08-12_15-09-10

jerrinot:
I agree excessive renaming might be harmful (I like  @sancar's analogy with memory invalidation), but this particular one makes a lot of sense to me. It's rather small in scope and I don't see any controversial changes. 

:+1:  (assuming @pveentjer has checked impact on Enterprise)

metanet:
+1 for sancar's comment. 

:+1:

pveentjer:
I'll make the enterprise pr shortly. So don't hit the merge button yet. 

I'm currently setting up a whole bunch of test that validate tcpipconnection setup. So want to complete that first.

pveentjer:
The enterprise PR
https://github.com/hazelcast/hazelcast-enterprise/pull/313

pveentjer:
Hmmm.. why is this PR being rebuild? I didn't change any code.

PR #5938, opened 2015-08-12
testQueryCacheCleanup moved to Nightly

jerrinot:
The test is running over 1 minute. There was an attempt to categorize it as NighlyTest,
but the Category annotation does not work on method level.


mesutcelik:
what about `SlowTest`? That way coverage will be calculated...

jerrinot:
@mesutcelik: Could do. I wanted to keep the same category as was originalyl intended, but I can change it if you think it suits it better.

mesutcelik:
no need to change...
keep it as it is. We can move all similar NightlyTests to SlowTests in one seperate PR...

jerrinot:
ok

Issue #5862, opened 2015-08-05
InvocationNetworkSplitTest Failures


hasancelik:
new failure same as  #4779

```
java.lang.AssertionError: expected:<3> but was:<2>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationNetworkSplitTest.testWaitingInvocations_whenNodeSplitFromCluster(InvocationNetworkSplitTest.java:107)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationNetworkSplitTest.testWaitingInvocations_whenNodeSplitFromCluster(InvocationNetworkSplitTest.java:61)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.6/com.hazelcast$hazelcast/633/testReport/junit/com.hazelcast.spi.impl.operationservice.impl/InvocationNetworkSplitTest/testWaitingInvocations_whenNodeSplitFromCluster/

jerrinot:
I think I know what's going on here: The test creates [creates a cluster](https://github.com/hazelcast/hazelcast/blob/cb49de684405d9864deaef2e91e5bc62303bf129/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNetworkSplitTest.java#L80) of 3 members, introduces a [split-brain](https://github.com/hazelcast/hazelcast/blob/cb49de684405d9864deaef2e91e5bc62303bf129/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNetworkSplitTest.java#L230) and then attempts to [heal it.](https://github.com/hazelcast/hazelcast/blob/cb49de684405d9864deaef2e91e5bc62303bf129/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNetworkSplitTest.java#L105) 


[...]

PR #5939, opened 2015-08-12
Spurious test failure fixed

jerrinot:
Fixing #5862 as discussed with @mdogan 

The test creates creates a cluster of 3 members, introduces a split-brain and then attempts to heal it.

Now the healing process has two-steps: first it calls prepareMerge() which schedules the actual merge. However the test also calls the same merge() method on its own. So at the end of the day there are 2 threads attempting to call the merge() method. The merge() is protected by a AtomicBoolean to ensure only 1 thread is doing a merge. In most cases it's the test thread which win the CAS. But if the scheduled thread wins then the test-thread will lose the CAS on and will return before the merge is completed -> Test Failure.


jerrinot:
verify-parallel

gurbuzali:
:+1:

PR #5943, opened 2015-08-13
Issue was passing Response of submit to key  as raw…

sancar:
… clientMessage to user. Fixed the issue with using ClientDelegatingFuture. Related test is updated to verify the fix in both old and new client.
related to #5883.


sancar:
Merging this since https://github.com/hazelcast/hazelcast/pull/5944 is reviewed and merged.

PR #5948, opened 2015-08-13
Removed getIOHandler method

pveentjer:
getIoService and getIOHandler both did the same. So I removed the getIOHandler since it isn't
a logical name (the object returned is called ioservice).

HZ enterprise doesn't use this method. So no changes needed there.


jerrinot:
:+1:

PR #5950, opened 2015-08-13
Issue : ClientMapProxy executeOnKey returns raw client message to use…

sancar:
…r. Fix : Response is decoded and the resulting data deserialized to user object. Related tests to verify the fix are dadded. backport of fix done in #5883


asimarslan:
:+1:

ihsandemir:
:+1:

PR #5951, opened 2015-08-13
Fix Map Reduce codec mappings

metanet:
Test in the old client is not necessary but it is also sent to align the tests between clients


metanet:
verify

sancar:
:+1: I came across this bug just yesterday. It seems you beat me to it. :)

serkan-ozal:
:+1:

PR #5954, opened 2015-08-13
Ignore ReferenceId for transactional locks

### LINKED ISSUE #5923: Suspicious behavour of testRecovery_singleInstanceRemaining
https://github.com/hazelcast/hazelcast/blob/08857da91e497365437700243db026fda4c5c31a/hazelcast/src/test/java/com/hazelcast/xa/HazelcastXATest.java#L152  seems to have two operational modes:
- fast mode (<10s) 
- slow mode - (>2 minutes)

There is nothing in between. My guess is either our transaction recovery or the test itself is racy - sometimes a transaction is recovered successfully, but sometimes it's waiting for a timeout (2 minutes)

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK6/com.hazelcast$hazelcast/623/testReport/com.hazelcast.xa/HazelcastXATest/testRecovery_singleInstanceRemaining/history/

image

jerrinot:
Fixes #5923

For transactional locks the referenceId is not taken into consideration as lock/unlock operations might
be initiated by different members -> they might have the same referenceId by a chance and the unlock might be lost.

I also write some unit tests for `LockStoreImplTest` and extracted `InternalLockService` from `LockServiceImpl` See commit messages for details.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/InternalLockService.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/InternalLockService.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.concurrent.lock;
+
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.spi.ObjectNamespace;
+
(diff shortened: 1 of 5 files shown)


jerrinot:
run-parallel

bilalyasar:
run-parallel

mdogan:
@jerrinot: I think we should ignore `referenceId` check for transactional locks, because it seems they are not following general lock pattern, different endpoints and threads can call transactional `lock` and `unlock`. Or transactional locks should use cluster-wide unique `referenceId`s.

jerrinot:
@mdogan: cluster-wide unique referenceIDs sounds scary.

mdogan:
That's why I'm suggesting to ignore `referenceId` check for transactional locks:

```java
if (!transactional && this.referenceId == referenceId) {
     return true;
}
```

jerrinot:
run-parallel

jerrinot:
@mdogan: I pushed an update, can you please have a look? thanks!

mdogan:
@jerrinot: Did you decide to keep ` this.transactional = false;` line when a lock attempt fails?

jerrinot:
@mdogan: I'll send a separate PR with this. It's a somewhat risky change so I want to be able to revert it easily if needed.

mdogan:
@jerrinot: Ok, make sense

:+1:

gurbuzali:
:+1:

gurbuzali:
do we need a backport here ?

jerrinot:
The referenceId check was merged recently. It does not exist in the
maintainance branch.

PR #5956, opened 2015-08-13
Fixes spurious test failure for VersionCheckTest/

kobalski:
ClusterUpTime should not be negative but it can have 0 value due to System.currentTimeMilis() precision.


serkan-ozal:
:+1:

pveentjer:
:+1

PR #5960, opened 2015-08-14
Old client ringbuffer Filter problem

pveentjer:
Filter was not passed correctly

Tests have been added for various filter usages


jerrinot:
could you add the test into the `ParallelTest` category?

pveentjer:
Yes. But not as part of this PR.

hazelcast/src/main/java/com/hazelcast/client/ClientEngine.java
```diff
@@ -59,6 +60,13 @@
     SecurityContext getSecurityContext();
 
     /**
+     * Returns the SerializationService.
+     *
+     * @return the SerializationService
+     */
+    SerializationService getSerializationService();
```
jerrinot: is RingBuffer the only data structure which needs to expose `SerializationService`?

pveentjer:
Taking too much time. Merging.

PR #5961, opened 2015-08-14
NOT TO BE MERGED. REVIEW . Error codes for protocol.

sancar:


hazelcast-client-new/src/test/java/com/hazelcast/client/atomiclong/ClientAtomicLongTest.java
```diff
@@ -98,7 +99,7 @@ public void apply_whenException() {
         try {
             ref.apply(new FailingFunction());
             fail();
-        } catch (ExpectedRuntimeException expected) {
+        } catch (HazelcastException expected) {
```
sancar: We are losing this kind of support via protocol. 
One suggestion is to give a way to register error codes and exceptions to user.

hazelcast-client-new/src/test/java/com/hazelcast/client/executor/ClientExecutorServiceTest.java
```diff
@@ -206,7 +206,7 @@ public void onFailure(Throwable t) {
     }
 
 
-    @Test(expected = IllegalStateException.class)
+    @Test
     public void testSubmitFailingCallableReasonExceptionCause() throws Throwable {
```
sancar: This is another feature that we lost via my implementation. One solution is to add cause error code definition of exception in protocol. Question there is how many levels of cause we should put? For this test one more level is enough. But generally every exception may have a cause.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
```diff
@@ -0,0 +1,237 @@
...
+                return new IllegalStateException(message);
+            }
+        });
+        register(HAZELCAST, HazelcastException.class, new ExceptionFactory() {
+            @Override
+            public Throwable createException(String message) {
+                return new HazelcastException(message);
+            }
+        });
+        //TODO following list will be registered !?!?!
```
sancar: We will register all 64 exceptions that we use now. When we start to use a new exception that is not in the list, it will be send as HazelcastException. Are we OK with that ?

ihsandemir: I would suggest only add the exception if it has a meaning and use for the client implementor and wrap all user unrelated but server related errors in a generic error such as internal error. We should go through each of them and make this analysis. Furthermore, the user will most probably ask us which one is possible error in which operation and what action the user needs to take when a specific error is received (e.g. retry the operation?). What do you think about it?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/parameters/ExceptionResultParameters.java
```diff
@@ -33,24 +34,23 @@
      * ClientMessageType of this message
      */
     public static final int TYPE = ResponseMessageConst.EXCEPTION;
-    public String className;
-    public String causeClassName;
+    public int errorCode;
```
ihsandemir: Do we need a field such as human readable error name? (Would be nice for documentation)

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/parameters/ExceptionResultParameters.java
```diff
@@ -33,24 +34,23 @@
...
     public String message;
-    public String stacktrace;
+    public StackTraceElement[] stackTrace;
```
ihsandemir: Do we really need to communicate the stack trace for well known errors, e.g. AuthenticationFailure ? The stack trace exposes server internal operation to a client developer, should he/she be knowledgable to understand the stack trace? There may be a lot of questions from the client developers regarding this field .

what is a typical size for this field?

PR #5964, opened 2015-08-14
return type cacheRemove in protocol converted to boolean.

sancar:


jerrinot:
isn't this an incompatible protocol change?

sancar:
It does broke the compatibility. We will be finalizing the protocol with 3.6 release, so this period is the last time we can do the fixes. And, since release on maintenance branch is just beta, we have more flexibility to change.

sancar:
Team decided to not to back port protocol changes, closing this one.

PR #5965, opened 2015-08-14
Do not downgrade a transactional lock to a non-transactional lock after unsuccessful lock attempt

jerrinot:
Do not downgrade a transactional lock to a non-transactional lock after unsuccessful lock attempt

It seems like a left-over. We should not change state of a lock after unsuccessful lock-attempt.
It can result in premature lock release and jeopardize transaction recovery.

This PR is based on #5954 it should be merged only after #5954 is merged.


jerrinot:
I merged #5954 -> this one is ready for review&merge now.

eminn:
great finding :+1:

PR #5969, opened 2015-08-15
Removed exception count probes in read/write-handlers

pveentjer:
The NonBlockingReadHandler and writehandler have exception count probes, but since
every exception is terminal to the handler or connection there is no point in
counting them.


pveentjer:
run-parallel

jerrinot:
:+1:

pveentjer:
I already talked this through with Mehmet and he agreed. And since I already have a thumbs up, thanks Jaromir, I'll merge this one.

PR #5970, opened 2015-08-15
Minor readhandler fixes

pveentjer:
1: merged 2 try/catch statements into a single one. The whole handle method is now wrapped into a single try/catch (any exception is deadly anway)
2: simplified readbytes handling. It will use the readBytes from the socketChannel.read
to determine if something was read or not; instead of asking the bb.
3: removed one if branch on the happy path

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingReadHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingReadHandler.java
@@ -20,12 +20,12 @@ import com.hazelcast.internal.metrics.MetricsRegistry;
 import com.hazelcast.internal.metrics.Probe;
 import com.hazelcast.nio.Protocols;
 import com.hazelcast.nio.ascii.SocketTextReader;
-import com.hazelcast.nio.tcp.ClientPacketSocketReader;
 import com.hazelcast.nio.tcp.ClientMessageSocketReader;
+import com.hazelcast.nio.tcp.ClientPacketSocketReader;
 import com.hazelcast.nio.tcp.PacketSocketReader;
+import com.hazelcast.nio.tcp.ReadHandler;
 import com.hazelcast.nio.tcp.SocketReader;
 import com.hazelcast.nio.tcp.TcpIpConnection;
-import com.hazelcast.nio.tcp.ReadHandler;
 import com.hazelcast.util.Clock;
 import com.hazelcast.util.counters.Counter;
 import com.hazelcast.util.counters.SwCounter;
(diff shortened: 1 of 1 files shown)


jerrinot:
WIP?

pveentjer:
WIP = work in progress, Means that the work has not yet completed.

jerrinot:
why the PR for WIP? it should be at least marked as [DO NOT REVIEW] 
it's annoying when you start reviewing and it's changing under your hands.

pveentjer:
run-parallel

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingReadHandler.java
```diff
@@ -145,14 +145,17 @@ public void requestMigration(NonBlockingIOThread newOwner) {
...
     public void handle() {
-        eventCount.inc();
-        lastReadTime = Clock.currentTimeMillis();
-        if (!connection.isAlive()) {
-            String message = "We are being asked to read, but connection is not live so we won't";
-            logger.finest(message);
-            return;
-        }
         try {
+            eventCount.inc();
```
jerrinot: why is this now inside a `try` block?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingReadHandler.java
```diff
@@ -145,14 +145,17 @@ public void requestMigration(NonBlockingIOThread newOwner) {
...
+            // we are going to set the timestamp even if the socketChannel is going to fail reading. In that case
+            // the connection is going to be closed anyway.
+            lastReadTime = Clock.currentTimeMillis();
```
jerrinot: same here, I don't understand why it has to be inside the try block

pveentjer: To make it simple. There was tricky try/catch stuff done with multiple blocks and makes it hard to understand. So now the try/catch is around the whole method.

It is in preparation of another pr... getting rid of the try/catch block by placing it from the outside.

jerrinot: the tricky try/catch stuff is gone thanks to this PR. 
I'd still prefer the version things like checking/logging was placed outside the try/catch block... anyway, it's not terribly important.

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingReadHandler.java
pveentjer: I want to get rid of a lot of stuff in the loop :) Try catch already is available and io thread, so I'll let it take care completely of the exception handling and allow for a callback in case of an exception (close connection). If connection alive can also be removed since socket channel is closed; so you will run into that anyway.

PR #5971, opened 2015-08-15
Minor cleanup in handle selection keys

pveentjer:
got rid of local variable for the set but immediately jump to the iterator.
also removed some finals since they are not being used in this class.


jerrinot:
:+1:

PR #5972, opened 2015-08-15
TcpIpConnection.close was not thread-safe

pveentjer:
It is now made thread-safe by adding a cas on the close. So only
one thread will complete the close.


pveentjer:
run-parallel

pveentjer:
verify

jerrinot:
:+1:

PR #5973, opened 2015-08-15
Writehandler handle method cleanup

pveentjer:
the try/catch is now around the whole handle method.
write outputbuffer doesn't catch any exceptions
pulled out startMigration method
event counter is not volatile long but a swcounter; just as with read-handler
removed findbugs supression since no volatile long increment anymore


pveentjer:
run-parallel

pveentjer:
verify

pveentjer:
verify

pveentjer:
verify

jerrinot:
:+1:

PR #5974, opened 2015-08-15
Don't use new Config() as default test HZ config

mtopolnik:
This small PR fixes the behavior of the nullary `newHazelcastInstance()` in the test factory, such that it doesn't insert a `new Config()`; instead it passes `null` to the production factory, letting it construct the default in the usual way (which complies with the XML Schema defaults).

At the same time the behavior is fixed regarding the creation of named HZ instances: where previously `null` was passed as the name, triggering automatic name generation, now the setting in the supplied configuration is honored (if present).

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/IssuesTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/IssuesTest.java
@@ -57,7 +57,7 @@ public class IssuesTest extends HazelcastTestSupport {
         int n = 1;
         TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(n);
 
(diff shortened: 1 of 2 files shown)


pveentjer:
Looks good. But could you add a test and a few lines of comment so that someone doesn't undo what you did?

mtopolnik:
I could add a comment, but I'm not sure if it would be OK to write a test for test code?

mtopolnik:
Enes, I already have a commit ready with documentation :)

mtopolnik:
Can we revert, then merge after that commit?

pveentjer:
I would just push the comment directly to master.

PR #5981, opened 2015-08-17
TransactionalMap.values() should return List instead of Set. Back port of pull request #5925

cangencer:
Back port of pull request #5925


pveentjer:
:+1:

sancar:
:+1:

PR #5982, opened 2015-08-17
[DONT MERGE] Metrics Memory Optimization

pveentjer:


pveentjer:
There is already a different PR with an improved approach merged. So this one can be closed

PR #5986, opened 2015-08-17
Fixes problem in client side logic ringbuffer

pveentjer:
The problem was that the incorrect field was being used to determine the size of the result result set.
It should use the method size(), but it relied on an internal array which can have a different size (like
array in arraylist).

Also the ClientReliableTopicTest is now made a normal test instead of a nightly.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
@@ -12,6 +12,8 @@ import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.annotation.NightlyTest;
+import com.hazelcast.test.annotation.ParallelTest;
+import com.hazelcast.test.annotation.QuickTest;
 import com.hazelcast.topic.impl.reliable.ReliableMessageListenerMock;
 import com.hazelcast.util.Clock;
 import org.junit.After;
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
```diff
@@ -34,7 +35,7 @@
 import static org.junit.Assert.assertTrue;
 
 @RunWith(HazelcastParallelClassRunner.class)
-@Category(NightlyTest.class)
+@Category(QuickTest.class)
```
jerrinot: can you add the `ParallelTest` category too?

pveentjer: done

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/client/ReadManyRequest.java
```diff
@@ -64,11 +64,13 @@ public int getClassId() {
     @Override
     protected Object filter(Object response) {
         ReadResultSetImpl readResultSet = (ReadResultSetImpl) response;
+
         int readCount = readResultSet.readCount();
         List<Data> items = new ArrayList<Data>(readCount);
-        for (Data item : readResultSet.getDataItems()) {
-            items.add(item);
+        for (int k = 0; k < readResultSet.size(); k++) {
+            items.add(readResultSet.getDataItems()[k]);
```
jerrinot: perhaps C2 will optimize this away, but I'd still remove the `getDataItems()` outside of the loop.

jerrinot:
just a minor comment, I have no objection if you merge it as it is.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/client/ReadManyRequest.java
```diff
@@ -64,11 +64,13 @@ public int getClassId() {
     @Override
     protected Object filter(Object response) {
         ReadResultSetImpl readResultSet = (ReadResultSetImpl) response;
+
         int readCount = readResultSet.readCount();
         List<Data> items = new ArrayList<Data>(readCount);
```
jerrinot: not related to this PR, but doesn't this create unnecessary large arraylist? shouldn't the line be
```java
List<Data> items = new ArrayList<Data>(readResultSet.size());
```
?

pveentjer:
Review comments have been addressed.Thanks Jaromir

jerrinot:
my pleasure! :)
:+1:

PR #5990, opened 2015-08-17
Fixed default value of `backup-count` and `async-backup-count` properties for cache definition at XSD files

serkan-ozal:
Currently, in the cache config bean (programmatic configuration), `backup-count`'s default value is `1` and `async-backup-count`s default is `0`. But in the XSD definitions they are not the same.
 
- In the `hazelcast-config.xsd`, `backup-count`'s default value is `0` and `async-backup-count`s default is `0`.
- In the `hazelcast-spring.xsd`, there is no default value for `backup-count` and `async-backup-count` properties. In this case, they are not set to cache config bean so the default values are used of cache config bean as in the programmatic configuration.

With this PR, in XSD definitions, `backup-count`'s default value and `async-backup-count`s default value are removed. So default values are used from cache config bean. Also XSD documentations are updated as related changes.


pveentjer:
:+1:

PR #5991, opened 2015-08-17
introducing error codes to define exceptions in protocol.

sancar:


sancar:
verify

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -375,24 +375,11 @@ private void handleClientMessage(ClientMessage clientMessage) throws ClassNotFou
...
-                    Constructor<?> constructor = clazz.getDeclaredConstructor(new Class[]{String.class, Throwable.class});
-                    constructor.setAccessible(true);
-                    exception = (Throwable) constructor.newInstance(exceptionResultParameters.message, cause);
-                } else {
-                    Class<?> clazz = Class.forName(exceptionResultParameters.className);
-                    Constructor<?> constructor = clazz.getDeclaredConstructor(new Class[]{String.class});
-                    constructor.setAccessible(true);
-                    exception = (Throwable) constructor.newInstance(exceptionResultParameters.message);
-                }
+                ExceptionResultParameters exParameters = ExceptionResultParameters.decode(clientMessage);
```
ihsandemir: We changed all the names from *parameters to *Codec, hence may be we can change ExceptionResultParameters to ErrorCodec to be more consistent.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientExceptionFactory.java
```diff
@@ -0,0 +1,521 @@
+/*
```
ihsandemir: Can we just list the error codes in a more generic way in a template file and have this class generated? We may need similar error handlers in every language. If we provide a way to generate, it may help the developers.

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: updated.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/parameters/ExceptionResultParameters.java
```diff
@@ -33,77 +34,88 @@
      * ClientMessageType of this message
      */
     public static final int TYPE = ResponseMessageConst.EXCEPTION;
+    public int errorCode;
     public String className;
-    public String causeClassName;
     public String message;
-    public String stacktrace;
+    public StackTraceElement[] stackTrace;
+    public int causeErrorCode;
```
ihsandemir: Having two error codes in an error is confusing for the developers. All a protocol implementer would be interested is the actual cause for the error. Why provide error code if there is a causeErrorCode. May be we can just put the causeErrorCode value in errorCode field and eliminate causeErrorCode field.

sancar: This is necessary for our tests. We have tests that is testing if cause type correct. I added this to not lose that behaviour.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientExceptionFactory.java
sancar: May be in a different pr, we can work on this.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
```diff
@@ -78,9 +78,17 @@
...
+     * @param stacktrace     array of stack trace
+     * @param causeErrorCode error code of cause of this exception, if there is no cause -1
+     * @param causeClassName java class name of the cause of this exception
+     */
     @Response(ResponseMessageConst.EXCEPTION)
-    void Exception(int error, String className, String causeClassName, String message, String stacktrace);
+    void Exception(int errorCode, String className, @Nullable String message, StackTraceElement[] stacktrace
+            , int causeErrorCode, @Nullable String causeClassName);
 
     //FIXME review this one
```
ihsandemir: why is FIXME still there?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
```diff
@@ -0,0 +1,95 @@
...
+
+package com.hazelcast.client.impl.protocol;
+
+/**
+ * Each exception that are defined in client protocol have unique identifier which are error code.
+ * All error codes defined in protocol are listed in this class.
+ */
+public final class ClientProtocolErrorCodes {
+
+    public static final int UNDEFINED = 0;
```
ihsandemir: What is "UNDEFINED"?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
```diff
@@ -0,0 +1,95 @@
...
+    public static final int OPERATION_TIMEOUT = 38;
+    public static final int PARTITION_MIGRATING = 39;
+    public static final int QUERY = 40;
+    public static final int QUERY_RESULT_SIZE_EXCEEDED = 41;
+    public static final int QUORUM = 42;
+    public static final int REACHED_MAX_SIZE = 43;
+    public static final int REJECTED_EXECUTION = 44;
+    public static final int REMOTE_MAP_REDUCE = 45;
+    public static final int RESPONSE_ALREADY_SENT = 46;
+    public static final int RETRYABLE_HAZELCAST = 47;
```
ihsandemir: Does this relate to the "@retyryable" tag in the message templates?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
```diff
@@ -0,0 +1,95 @@
...
+    public static final int RETRYABLE_IO = 48;
+    public static final int RUNTIME = 49;
```
ihsandemir: To the user every error is a runtime error, i think it is too much Java specific. May be we can put these into another error code such as unknown error code.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
sancar: No, there is no direct relation. There are a group of exceptions that we might want to retry.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
sancar: There are others that we can argue, client developers does not has to use run time exception as is. It is just more info for other language developers. There are no harm to define in protocol for  other clients and if remove our java client will not be able to create RuntimeException as is.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
sancar: It is error code that we use when exception does not match any of the listed exceptions. I will add the javadoc later.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/template/ResponseTemplate.java
sancar: It is leftover from Peters pr. I will remove it later.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientExceptionFactory.java
ihsandemir: :+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/parameters/ExceptionResultParameters.java
ihsandemir: :+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
ihsandemir: :+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientProtocolErrorCodes.java
ihsandemir: :+1:

ihsandemir:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientExceptionFactory.java
mdogan: :100:

mdogan:
:+1:

asimarslan:
:+1:

PR #5993, opened 2015-08-18
[BACKPORT] Removed default value of `backup-count` and `async-backup-count` properties for cache definition at XSD files

serkan-ozal:
Currently, in the cache config bean (programmatic configuration), `backup-count`'s default value is `1` and `async-backup-count`s default is `0`. But in the XSD definitions they are not the same.
 
- In the `hazelcast-config.xsd`, `backup-count`'s default value is `0` and `async-backup-count`s default is `0`.
- In the `hazelcast-spring.xsd`, there is no default value for `backup-count` and `async-backup-count` properties. In this case, they are not set to cache config bean so the default values are used of cache config bean as in the programmatic configuration.

With this PR, in XSD definitions, `backup-count`'s default value and `async-backup-count`s default value are removed. So default values are used from cache config bean. Also XSD documentations are updated as related changes. 

Backport of https://github.com/hazelcast/hazelcast/pull/5990

### DIFF
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.5.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.5.xsd
@@ -542,18 +542,22 @@
                                 <xs:attribute name="backup-count" use="optional" type="xs:unsignedByte">
                                     <xs:annotation>
                                         <xs:documentation>
-                                            Number of sync backups. If 1 is set as the backup-count for example, then
-                                            all entries of the map will be copied to another JVM for fail-safety. Valid
-                                            numbers are 0 (no backup), 1, 2 ... 6.
+                                            Number of total (synchronous + asynchronous) backups.
+                                            For example, if `1` is set as the `backup-count`,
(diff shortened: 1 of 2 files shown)


jerrinot:
Do we want to backport changes of defaults? I understand the current situation is not great, but imagine you are running a cluster, you update Hazelcast to the latest patch release a boom: It doubles your memory consumption.

serkan-ozal:
@jerrinot Please look at the changes. Most of the changes are about XSD documentation update. The only XSD changes are removing default values from `hazelcast-config.xsd` and in fact at the moment, our `XmlConfigBulder` doesn't take care of default values (but default values are used for spring module). So it doesn't have any effect on runtime.

jerrinot:
"in fact at the moment, our XmlConfigBulder doesn't take care of default values" <-- This is the important part I didn't know about, thanks!

:+1:

serkan-ozal:
I had learned yesterday. You can ask @bilalyasar about details.

serkan-ozal:
Closing this PR, explanation is wrong

PR #5913, opened 2015-08-10
Spinning IOThreadingModel

pveentjer:
Since the TcpIpConnection threadmodel has been pulled into its own abstraction, it is easy to swap it out by a different implementation. Hazelcast by default using non blocking (selector based) io. This PR contains a spinning IO implementation.  It spins by looping on the read/write of the SocketChannel.

It is an experimental feature to learn more about throughput and latency. And perhaps in the future can be used in production.

PR #5995, opened 2015-08-18
SelectNow option for non-blocking tcp threading-model

pveentjer:
The NonBlocking threading model now supports spinning based on calling selectNow instead of select. This is one of the way to spin; another way to spin is using spinning on socketchannel. Which will be part of this pr:

https://github.com/hazelcast/hazelcast/pull/5913

Spinning using selectNow is disabled by default and is enabled through a system property. It is experimental, so we don't publish it yet in the group properties.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
@@ -30,56 +30,70 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 
 public abstract class NonBlockingIOThread extends Thread implements OperationHostileThread {
 
+    // WARNING: This value has significant effect on idle CPU usage!
     private static final int SELECT_WAIT_TIME_MILLIS = 5000;
     private static final int SELECT_FAILURE_PAUSE_MILLIS = 1000;
 
-    @Probe(name = "selectorQueueSize")
-    protected final Queue<Runnable> selectorQueue = new ConcurrentLinkedQueue<Runnable>();
+    @Probe(name = "taskQueueSize")
+    private final Queue<Runnable> taskQueue = new ConcurrentLinkedQueue<Runnable>();
 
     private final ILogger logger;
-
-    private final int waitTime;
 
     private final Selector selector;
 
     private final NonBlockingIOThreadOutOfMemoryHandler oomeHandler;
 
-    // field doesn't need to be volatile, is only accessed by this thread.
-    private boolean running = true;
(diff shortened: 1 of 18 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -87,22 +104,85 @@ public final void addTask(Runnable task) {
...
+            for (; ; ) {
+                try {
+                    if (selectNow) {
+                        runSelectNowLoop();
+                    } else {
+                        runSelectLoop();
+                    }
+                    // break the for loop; we are done
+                    break;
+                } catch (IOException noneFatalException) {
```
jerrinot: type. it should be `nonFatalException` ?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -87,22 +104,85 @@ public final void addTask(Runnable task) {
...
+        // * it calls runSelectNowLoop or runSelectLoop based on selectNow enabled or not.
+        // * handles backoff and retrying in case if io exception is thrown
+        // * it takes care of other exception handling.
+        //
+        // The idea about this approach is that the runSelectNowLoop and runSelectLoop are as clean as possible and don't contain
+        // any logic that isn't happening on the happy-path.
+        try {
+            for (; ; ) {
+                try {
+                    if (selectNow) {
```
jerrinot: this is ugly, but as already discussed: I don't know any better so I can live with it.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
pveentjer: Sure.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
pveentjer: Me neither.. well.. marco made suggestion to do somethng like
```
do{
  try{
  }catch(NonFatal e){
      continue;
  }
}while(false)
```

jerrinot: meh

gurbuzali:
:+1:

jerrinot:
+1

PR #5996, opened 2015-08-18
Fix naming of some internal cache API methods

mtopolnik:
Although this looks like a large commit, the changes are strictly method renaming/reordering and some improvement in Javadoc.

The names of some methods in the internal cache API seem to be a leftover from an earlier design. They tend to confuse new people studying the codebase (like Serkan and myself). This PR aligns the method names with the current state of things.


pveentjer:
:+1:

Issue #5905, opened 2015-08-10
Faster Map Statistics


wildnez:
Management Center is unstable in showing cluster view. I started a 3 members cluster (each member on different machine), connected with Management Center running on another machine and started loading cluster with data (only imap.set calls). 

Aim was to load 30 million entries, amounting to ~22GB in total.

15 million into loading process and ManCenter shakes - 2 out of 3 nodes disappear for a few seconds, a red warning sign flashes at the top of the screen that says - " Instance Connection Warning: xxxxxxxxx". Please see attached screenshot MC12.

Clicking on Map and the entire table of statistics is blank, I stayed on this screen for half a minute but nothing changed. See attached screenshot MC13.

After few more minutes, all three nodes came back on Home page, see MC14.
But clicking on Map (CreditCardCache) shows only 2 nodes, see MC15.

Few seconds later, another member disappears from the Map screen and red warning starts flashing again, see MC16. After a few more seconds, everything disappears, see MC17.

And Management Centre stops showing anything from this point onwards, even on Home, see MC18.

Do note that all this while, all the members were very much connected with each other, no disconnections, highly stable network (everything normal in logs). The machine that runs Management Centre was also in the same network as the member machines. Important thing is that Management Centre could never recover from this situation again.
Members were running with 10GB heap and only 1 Major GC seen on node 1 during the complete load of 30 million.

I have also attached some of the Health Monitor logs from all the members when the members were keep going out and coming into ManCentre.

Reproduction is very easy - just start a 3 nodes cluster and load 30 million entries of 1KB.

wildnez:
Two more issues with Management Centre:

Only started 6 server nodes with partition count set to 4099. 

1.Connect Management Centre and "Cluster Health" shows "Backup Synchronization in Progress". Please note that until this point, servers were idle, no client was connected but still this message appears, with a WARNING symbol.

2. Almost all of 6 server nodes are running with more than 18% heap occupancy - not sure if this is a Management Centre issue or issue related to servers.

emrahkocaman:
Hi @wildnez, 

**For the first part**, it was a known issue (reported before https://github.com/hazelcast/hazelcast/issues/4895) but never tested with this much entry count.

This is not about management center, this is about how we collect map statistics. Currently, some map stats (like cost, last access time) is calculated by traversing map entries. So calculation time exceeds the time interval reserved for management center state sending thread when entry count is too high.

[...]

PR #6004, opened 2015-08-18
Async stat gathering for management center

emrahkocaman:
Collect statistics asynchronously to be able to send stats to management center within update time interval.
related #5905

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
@@ -71,6 +71,7 @@ import java.util.Map;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


emrahkocaman:
verify

bilalyasar:
:+1: , i also made some tests with this fix. Management center was stable with 66 million entry. But 66 million is not upper limit.

kobalski:
:+1:

jerrinot:
Does it mean the state is now calculated even when no management centre is connected?

mesutcelik:
> Does it mean the state is now calculated even when no management centre is connected? that sounds very wasteful to me.

I see no change in that behavior by this PR. It is just that calculating `TimedMemberState` object is done by another thread. That is because if state calculation is costly , we just send the previous state to keep up with Management Center.

jerrinot:
sorry, I missed the thread is started in the (guarded) `start()` method. Please ignore my comment.

mesutcelik:
:+1:  looks good to merge

emrahkocaman:
Backport will follow.

PR #6008, opened 2015-08-19
Introduced Operation.onFailure() to deal with failures on operation execution

mdogan:
Introduced Operation.onFailure() to deal with failures on operation execution. Normally, this isn't needed for current heap based operations. But operations using/allocating native/un-managed resources require a better failure handling to be able release/clean up those resources.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/impl/MigrationOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/impl/MigrationOperation.java
@@ -55,6 +55,8 @@ public final class MigrationOperation extends BaseMigrationOperation {
     private long[] replicaVersions;
     private Collection<Operation> tasks;
 
+    private Throwable failureReason;
+
     public MigrationOperation() {
     }
 
@@ -77,6 +79,11 @@ public final class MigrationOperation extends BaseMigrationOperation {
             doRun();
         } catch (Throwable t) {
             logMigrationFailure(t);
+            failureReason = t;
+        } finally {
+            if (!success) {
+                onExecutionFailure(failureReason);
+            }
         }
     }
 
@@ -86,11 +93,13 @@ public final class MigrationOperation extends BaseMigrationOperation {
                 migrate();
             } catch (Throwable e) {
                 success = false;
+                failureReason = e;
                 getLogger().severe("Error while processing " + migrationInfo, e);
             } finally {
                 afterMigrate();
             }
         } else {
+            success = false;
             logMigrationCancelled();
         }
     }
@@ -141,10 +150,13 @@ public final class MigrationOperation extends BaseMigrationOperation {
         addActiveMigration();
 
         for (Operation op : tasks) {
+            prepareOperation(op);
             try {
                 runMigrationTask(op);
             } catch (Throwable e) {
                 getLogger().severe("An exception occurred while executing migration operation " + op, e);
+                success = false;
+                failureReason = e;
                 return;
             }
         }
@@ -157,11 +169,6 @@ public final class MigrationOperation extends BaseMigrationOperation {
     }
 
     private void runMigrationTask(Operation op) throws Exception {
-        op.setNodeEngine(getNodeEngine())
-                .setPartitionId(getPartitionId())
-                .setReplicaIndex(getReplicaIndex());
-        op.setResponseHandler(ERROR_RESPONSE_HANDLER);
-        OperationAccessor.setCallerAddress(op, migrationInfo.getSource());
         MigrationAwareService service = op.getService();
         PartitionMigrationEvent event =
                 new PartitionMigrationEvent(MigrationEndpoint.DESTINATION, migrationInfo.getPartitionId());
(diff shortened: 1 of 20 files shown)


hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -309,11 +309,29 @@ public final void setWaitTimeout(long timeout) {
...
+     * Called when an <tt>Exception</tt>/<tt>Error</tt> is thrown
+     * during an invocation. Invocation process will continue, retry
+     * or fail according to returned <tt>ExceptionAction</tt> result.
+     * <p/>
+     * This method is called on caller side of the invocation.
+     *
+     * @param throwable <tt>Exception</tt>/<tt>Error</tt> thrown during invocation
+     * @return <tt>ExceptionAction</tt>
+     */
+    public ExceptionAction onInvocationException(Throwable throwable) {
```
gurbuzali: how about naming this method `onInvocationFailure` to be inline with `onFailure` ?
or `onFailure` -> `onExecutionExcetion` ?

gurbuzali:
just a minor comment related with naming
:+1:

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
mdogan: I didn't rename it to `onInvocationFailure`, because it doesn't have to be an invocation failure. Invocation can retry or fail. It's up to operation. And it's also inlined with returned value class: `ExceptionAction`. 

On the other hand, `onFailure` is actually an operation failure.

hazelcast/src/main/java/com/hazelcast/partition/impl/MigrationOperation.java
```diff
@@ -171,6 +178,37 @@ private void runMigrationTask(Operation op) throws Exception {
...
+        op.setResponseHandler(ERROR_RESPONSE_HANDLER);
+        OperationAccessor.setCallerAddress(op, migrationInfo.getSource());
+    }
+
+    @Override
+    public void onFailure(Throwable e) {
+        if (e != null) {
+            getLogger().severe(e);
+        } else {
+            getLogger().severe("!!! FAILURE BUT NO REASON !!!");
```
mdogan: oops, this is a left over from testing!

serkan-ozal:
:+1:

mdogan:
PLEASE DON'T MERGE YET! I'LL SEND A TEST SOON!

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/OperationFailureTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                assertNotNull(backupOperationFailure.get());
+            }
+        });
+
+        Throwable failure = backupOperationFailure.getAndSet(null);
+        assertTrue(failure instanceof ExpectedRuntimeException);
```
pveentjer: We have assertInstanceOf :)

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
pveentjer: I guess I'm missing something, but what kind of functionality is added that wasn't added by onException? It looks like a method rename (which is fine with me) but I don't see added functionality.

pveentjer: Ahh.. ok.. this is a rename.. the method below is the interesting one.

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -345,6 +363,28 @@ short getFlags() {
...
+     * Called when an <tt>Exception</tt>/<tt>Error</tt> is thrown during operation execution.
+     * <p/>
+     * By default this method does nothing.
+     * Operation implementations can override this behaviour due to their needs.
+     * <p/>
+     * This method is called on node & thread that's executing the operation.
+     *
+     * @param e Exception/Error thrown during operation execution
+     */
+    public void onFailure(Throwable e) {
```
pveentjer: What about calling it onExecutionFailure (executing side) to make it clear that it is close to the onInvocationFailure (calling side)

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
mdogan: I just deprecated old `onException` and added `onExecutionExcetion` here. To see added functionality, you need to look down through the code :arrow_down:

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
pveentjer: The problem is that we add more capabilities to the operation. Can't we do this by an interface? So having an interface FailureCallbackOperation with a single method 'onFailure'. This way you keep the operation nice and clean.

mdogan: Then I need to check `instanceOf` on every place `onFailure` is called. This also applies to most of the methods in `Operation` class. If we extract an interface for each method, I guess code will be chaos. I don't see any benefit here.

PR #6012, opened 2015-08-19
Load jcache exceptions only when jcache is avaliable

### LINKED ISSUE #6010: Hazelcast lastest snapshot run fail due to ClientExceptionFactory brought in for creation from ClientEngineImpl.
The lastest hazelcast brought in ClientExceptionFactory due to @sancar did for exception processing. but it failed due to the initialization of hazelcast instance due to the creation of ClientExceptionFactory 
as it should addd jcache api for initialization. so it's better to build the jcache jar into hazelcast (to be the fat one).
@sancar, if you have time, please spare some time and take a look at.

sancar:
fixes #6010
bug caused by #5991


mdogan:
:+1:

PR #6014, opened 2015-08-20
Seperates the public API from the private/implementation.

pveentjer:
This doesnt' change anything for the end user because he will only be linking to the public API and
binary class compatibility is not violated.


pveentjer:
verify-parallel

pveentjer:
verify-parallel

pveentjer:
verify

sancar:
This pr hard to review because of its nature. We had to make sure 
1)No public api moved to internal and
1) All privates are moved to internal.

From my review, I saw only one problem. Data interface is private as far as I know. 
That has been said, we used it heavily in protocol but now I started to question it.  May be we should change, Data usages to byte[] and leave Data as private.

pveentjer:
Data is unfortunately already part of the public API due to the NodeEngine of SPI.

asimarslan:
I think we can merge this as is. If we miss a class we can move it later before code freeze
:+1:

sancar:
It is private to API and public to SPI. NodeEngine also part of SPI. I think, for purpose of this change, it should be moved to internal.

pveentjer:
I don't think it can be moved. If someone has written a customer data-structure on top of the SPI and we move the data class to a different package, we have a source and class level binary compatibility problem. 

SPI = API

pveentjer:
The corresponding enterprise fix:
https://github.com/hazelcast/hazelcast-enterprise/pull/334

PR #6015, opened 2015-08-20
Adding exceptions to protocol that is used in enterprise

sancar:
related to #5991


mdogan:
run-parallel

mdogan:
verify

sancar:
verify

mdogan:
:+1:

PR #6021, opened 2015-08-20
Cleanup of GroupProperties

Donnerbart:
This PR is an attempt to cleanup the `GroupProperties`, which are not in a good maintainable state. One of the symptoms I see in the code are rogue properties (which are just in the code, not in the `GroupProperties`) or even defined properties which are not used, e.g.:
```java
System.setProperty("hazelcast.local.localAddress", "127.0.0.1"); // not defined
System.setProperty("hazelcast.version.check.enabled", "false"); // defined, but not used
System.setProperty("hazelcast.socket.bind.any", "false"); // defined, but not used
```

The solution I came up with should make the definition way easier and the usage not much different than the old approach. It also adds methods to convert time-based properties to the wanted time unit. The new class `GroupProperty` contains all definitions and default values as well as the JavaDoc. The old class `GroupProperties` is just a container for the configured properties.

## Property definition

At the moment you have to add at least 3 lines of code to add a new property:
```java
/**
 * Maybe some JavaDoc here.
 */
public static final String PROP_PARTITION_COUNT = "hazelcast.partition.count";

public final GroupProperty PARTITION_COUNT;

public GroupProperties(Config config) {
    PARTITION_COUNT = new GroupProperty(config, PROP_PARTITION_COUNT, "271");
}
```
And this is one of the best readable examples, because of the short names and single lines.

A new definition in the class `GroupProperty` looks like this:
```java
/**
 * Total number of partitions in the Hazelcast cluster.
 */
PARTITION_COUNT("hazelcast.partition.count", 271),
```
For a time-based property you should add the time unit to the definition:
```java
EVENT_QUEUE_TIMEOUT_MILLIS("hazelcast.event.queue.timeout.millis", 250, MILLISECONDS),
```

## Usage with `Config`

```java
Config config = new Config();

// old definition
config.setProperty(GroupProperties.PROP_PARTITION_COUNT, "1");
String value = config.getProperty(GroupProperties.PROP_PARTITION_COUNT);

// new definition
config.setProperty(GroupProperty.PARTITION_COUNT, "1");
value = config.getProperty(GroupProperty.PARTITION_COUNT);
```
As you can see there is almost no difference.

## Usage with `System`

```java
// old definition
System.setProperty(GroupProperties.PROP_WAIT_SECONDS_BEFORE_JOIN, "0");
String value = System.getProperty(GroupProperties.PROP_WAIT_SECONDS_BEFORE_JOIN);

// new definition
System.setProperty(GroupProperty.WAIT_SECONDS_BEFORE_JOIN.getName(), "0");
value = System.getProperty(GroupProperty.WAIT_SECONDS_BEFORE_JOIN.getName());

GroupProperty.WAIT_SECONDS_BEFORE_JOIN.setSystemProperty("0");
value = GroupProperty.WAIT_SECONDS_BEFORE_JOIN.getSystemProperty();
```
You can either use the `getName()` or the built-in methods `setSystemProperty()` and `getSystemProperty()`.

### Normal properties

```java
// old definition
int threadCount = groupProperties.CLIENT_ENGINE_THREAD_COUNT.getInteger();

// new definition
int threadCount = groupProperties.getInteger(GroupProperty.CLIENT_ENGINE_THREAD_COUNT);
```
I'm not super happy with this "reverse" conversion, but I couldn't find a better solution.

### Time-based properties

As promised I could at least add some syntactic sugar to get rid of manual time conversions:
```java
// old definition
Thread.sleep(groupProperties.WAIT_SECONDS_BEFORE_JOIN.getInteger() * 1000L * 2);		

// new definition
Thread.sleep(2 * groupProperties.getMillis(GroupProperty.WAIT_SECONDS_BEFORE_JOIN));
```
This is again one of the better readable examples, there were much worse parts in the code regarding time conversions. This will also reduce magic numbers and `TimeUnit.SECONDS.toMillis()` boilerplate code.

## Todo

If you like this PR the next steps would be
* Enterprise PR to fix usage in that project (already prepared)
* convert the ClientProperties to the same structure
* add missing rogue properties to the `GroupProperty` class

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientConfig.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientConfig.java
@@ -27,6 +27,7 @@ import com.hazelcast.config.SerializationConfig;
 import com.hazelcast.config.SocketInterceptorConfig;
 import com.hazelcast.config.matcher.MatchingPointConfigPatternMatcher;
 import com.hazelcast.core.ManagedContext;
+import com.hazelcast.instance.HazelcastProperty;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 import com.hazelcast.security.Credentials;
@@ -117,7 +118,7 @@ public class ClientConfig {
     }
 
     /**
-     * Gets a property already set or from system properties if not exists.
+     * Gets a named property already set or from system properties if not exists.
      *
      * @param name property name
      * @return value of the property
@@ -128,7 +129,7 @@ public class ClientConfig {
     }
 
     /**
-     * Sets the value of a named property
+     * Sets the value of a named property.
      *
      * @param name  property name
      * @param value value of the property
@@ -138,6 +139,28 @@ public class ClientConfig {
     public ClientConfig setProperty(String name, String value) {
         properties.put(name, value);
         return this;
+    }
+
+    /**
+     * Gets a {@link HazelcastProperty} already set or from system properties if not exists.
+     *
+     * @param property {@link HazelcastProperty} to get
+     * @return value of the property
+     */
+    public String getProperty(HazelcastProperty property) {
+        return getProperty(property.getName());
+    }
+
+    /**
+     * Sets the value of a {@link HazelcastProperty}.
+     *
+     * @param property {@link HazelcastProperty} to set
+     * @param value    value of the property
+     * @return configured {@link com.hazelcast.client.config.ClientConfig} for chaining
+     * @see {@link com.hazelcast.client.config.ClientProperties} for properties that is used to configure client
+     */
+    public ClientConfig setProperty(HazelcastProperty property, String value) {
+        return setProperty(property.getName(), value);
     }
 
     /**
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
@@ -25,7 +25,7 @@ import com.hazelcast.client.spi.ClientProxy;
 import com.hazelcast.config.SerializationConfig;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.PartitioningStrategy;
-import com.hazelcast.instance.GroupProperties;
+import com.hazelcast.instance.GroupProperty;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 import com.hazelcast.map.impl.MapService;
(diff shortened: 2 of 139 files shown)


pveentjer:
We should get validation on them.

```
GroupProperty<Integer> partitionCount = new GroupPropertyBuilder()
   .withName("hazelcast.partitionCount")
   .withValidation(new PositiveNumber())
   ... build()
```

Currently we add validation on the place where we use the property (or don't validate). But it is part of the property.

Donnerbart:
Yes, I thought of that as well, but your API looks better than what I had in mind. I think this could be a good follow-up PR (this one is already big enough).

Donnerbart:
I don't see how we can get rid of them, since (at least the property names) are part of the public API. And they are a fixed set, so we need to define them (incl. their JavaDoc, a default value and a validation).

We already have those scattered and non-documented properties in the code. Even if we convert this to nicer code and DI, where should we define the names and values? Often there is no single place for a property, e.g. the partition count.

Donnerbart:
I think I cannot make my approach type based, since I use an `enum` for the properties.

BTW, here is a short usage statistic:
```
 3 getString()
25 getBoolean()
39 getInteger()
 3 getLong()
 1 getFloat()
 3 getNanos()
30 getMillis()
15 getSeconds()

 3 String
25 Boolean
43 Number
48 TimeUnit
```
So a major part of the properties are time-based values, where we really profit from the new getter methods. I not sure how we should implement this with a typed based approach, since they are still normal numbers, but have an additional `TimeUnit` to interpret them correctly.

pveentjer:
Can you rebase?

Donnerbart:
Rebase done.

bilalyasar:
@Donnerbart , Did you remove any unused property. Maybe we need to also remove from reference-manual. Also i think, this PR may requires doc update.

Serdaro:
I am checking all the files, but if a summary can be given I would be happy to update the manual.

Donnerbart:
I didn't remove any property, since all are referenced in the code somewhere. They are just moved from `GroupProperties` (which is just a container for the values now) to `GroupProperty` (which holds the constants we use to reference a property).

Also the `GroupProperties.PROP_xxx` constants are still there but marked as deprecated. I didn't remove them since we are not sure if anyone uses them. They are in a shadow zone of public API (never mentioned, but maybe used).

We don't mention the classes `GroupProperties` or `GroupProperty` in the System properties section of the reference manual. I think this is good, since they are our internal API for the properties. Our public API are the property names (e.g. `hazelcast.partition.count`). Those are documented and should be used.

Serdaro:
Ah ok, I thought this effort included the removal of non-used properties. Thank you for the explanation :)

Donnerbart:
I found one mention of those files in the reference manual, see https://github.com/hazelcast/hazelcast-reference-manual/issues/31

sancar:
:+1:

jerrinot:
@Donnerbart: this needs rebase

Donnerbart:
Done.

gurbuzali:
:+1:

PR #6029, opened 2015-08-20
Be sure that completion event is sent in all cases for `CacheRemoveAllOperation`. 

### LINKED ISSUE #5865: `CacheRemoveAllOperation` doesn't send completed event in some cases
`CacheRemoveAllOperation` doesn't send completed event
- If `CacheRecordStore` for that partition is not created yet
- If the filtered keys are empty (that partition is not owner of the any key to be removed)

The solution is sending completed event in these cases.

This failure can be reproduced by getting cache from the second node instead of first node in `BasicCacheTest::testCompletionTest`. I have tried the fix idea above on my local and it worked.

serkan-ozal:
Send completion events in these cases also:
- If `CacheRecordStore` for that partition is not created yet
- If the filtered keys are empty (that partition is not owner of the any key to be removed)

Fixes #5865


metanet:
:+1:

PR #6031, opened 2015-08-21
SelectionHandler exception handling

pveentjer:
Instead of the handle method doing the exception handling on top of its primary concern, handling the event, the exception handling logic has been pulled out by adding an additional onFailure method that gets called when the handle method throws an exception.

This makes the handle logic a lot cleaner since all the exception handling logic is moved out.

In the clients an unhandled exception leads to closing the connection; just like on server. In the past only the exception was logged but the continue allowed to continue. But this doesn't make sense since an exception is fatal.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/ClientReadHandler.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/ClientReadHandler.java
@@ -22,7 +22,6 @@ import com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThread;
 import com.hazelcast.util.Clock;
 
 import java.io.EOFException;
-import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.SelectionKey;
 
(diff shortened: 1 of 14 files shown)


jerrinot:
:+1:  once green

hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/ClientReadHandler.java
```diff
@@ -52,41 +51,32 @@ public void run() {
...
+        if (readBytes <= 0) {
             if (readBytes == -1) {
                 throw new EOFException("Remote socket closed!");
             }
-        } catch (IOException e) {
-            handleSocketException(e);
             return;
         }
-        try {
-            if (buffer.position() == 0) {
```
sancar: I see that these 3 lines are removed. Did you think that these are not necessary ? Or Is it changed by mistake when doing cleanup ?
``` 
if (buffer.position() == 0) {
      return;
} 

```

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientWriteHandler.java
```diff
@@ -60,17 +61,13 @@ public void handle() {
             ready = true;
             return;
         }
-        try {
-            writeBuffer();
-        } catch (Throwable t) {
-            logger.severe("Fatal Error at WriteHandler for endPoint: " + connection.getEndPoint(), t);
-        } finally {
```
sancar: There is also a behaviour change. The following lines will not be run when an writeBuffer throws Exception.
```
ready = false
registerWrite()
```
Aside from that this means that when `packet.writeTo` or `clientMessage.writeTo` throws a runtime exception connection will be closed. Before it was just logged and consequent packets may be written safely. Since I dont really see, how `writeTo` method can throw RunTimeException, I want to say it is OK, but I am not %100 sure. Leaving comment here to get attention from more people. ;)

hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/ClientReadHandler.java
pveentjer: Hi Sancar, they are already taken care of by using the readCount. So no need to check the readBuffer. The same approach is being used on the serverside.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientWriteHandler.java
pveentjer: The behaviorial change is intentional and that is why I need your input. On the serverside, any exception thrown in the read/write-handler automatically leads to a closing of the connection because you can't guarantee anything of the health of the connection.

pveentjer: ----> and consequent packets may be written safely.

THis is not going to work because you might have corrupted the stream. Some bytes may have been written to the bb and some have not.

sancar:
:+1:

PR #6033, opened 2015-08-21
Thread flattening in nio

pveentjer:
There is no need to have subclasses of the NonBlockingIOThread on the serverside. So the existing input/output thread have been collapsed in the NonBlockingIOThread.


jerrinot:
Looks Good to Me. :+1:

PR #6034, opened 2015-08-21
NonBlockingIOThread last select time incorrect

pveentjer:
The last select time was not set correctly in the NonBlockingIOThread.
This time is being used as a probe so isn't super exiting, but it was
not set on the selectNow and it was alwasy set on select, even it was
just the regular timeout.


jerrinot:
:+1:

pveentjer:
Since this is only relevant for a probe; I'll merge this.

PR #6038, opened 2015-08-21
Initial cleanup of CheckStyle suppressions for DefaultAddressPicker

Donnerbart:
I left the harder to resolve suppressions for the first iteration to keep the PR small.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/DefaultAddressPicker.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/DefaultAddressPicker.java
@@ -43,12 +43,17 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
 class DefaultAddressPicker implements AddressPicker {
 
-    private final ILogger logger;
+    private static final int SOCKET_BACKLOG_LENGTH = 100;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/instance/DefaultAddressPicker.java
```diff
@@ -364,18 +365,16 @@ private void log(Level level, String message) {
     }
 
     private class InterfaceDefinition {
+
         String host;
         String address;
 
-        private InterfaceDefinition() {
```
Donnerbart: I removed the private constructor since CheckStyle insists on making a class final if it has just private constructors. But AddressDefinition extends InterfaceDefinition, so this won't work. Since the inner class is private, there is no need for private constructors anyway, so safe to delete.

hazelcast/src/main/java/com/hazelcast/instance/DefaultAddressPicker.java
```diff
@@ -412,23 +411,21 @@ public int hashCode() {
     }
 
     private class AddressDefinition extends InterfaceDefinition {
+
         InetAddress inetAddress;
         int port;
 
-        private AddressDefinition() {
```
Donnerbart: Same for AddressDefinition. I also removed the private keyword from the other constructors for the same reason.

pveentjer:
:+1:  good work David. Thanks for cleaning up!

jerrinot:
:+1:

PR #6042, opened 2015-08-24
Fixed tcp last-read/write-time issue

pveentjer:
The bug was in the unit test. Instead of creating a list of packets in the
setup method, it was created in the class constructor.

Also the lastReadTime and lastWriteTime have been renamed to lastReadTimeMillis
and lastWriteTimeMillis to remove any confusion about their unit of time.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
@@ -128,12 +128,12 @@ public class ClientConnection implements Connection, Closeable {
     }
 
     @Override
-    public long lastReadTime() {
+    public long lastReadTimeMillis() {
         return readHandler.getLastHandle();
     }
 
     @Override
-    public long lastWriteTime() {
+    public long lastWriteTimeMillis() {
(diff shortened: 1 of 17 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -34,7 +34,7 @@
 public abstract class NonBlockingIOThread extends Thread implements OperationHostileThread {
 
     // WARNING: This value has significant effect on idle CPU usage!
-    private static final int SELECT_WAIT_TIME_MILLIS = 5000;
+    private static final int SELECT_WAIT_TIME_MILLIS = 15000;
```
jerrinot: is this a left-over?

pveentjer: Oooooopsss

jerrinot:
:+1: (on green)

Donnerbart:
Looks good to me (if PR builder likes it as well).

pveentjer:
verify

PR #6043, opened 2015-08-24
Query Performance: InflatableSet Implementation

jerrinot:
This PR avoid potentially slow HashSet creation. 

Our tests have shown adding data into HashSet might be a bottleneck - specially when equals/hashdode methods are somewhat slow. This PR introduces `InflatableSet` - we are playing a bet most Sets obtained via querying a map are simple iterated over and they are rarely modified. I have a test (obtained from a real customer) which shows 3-5x improvement in throughput with this PR.

See commit messages for details. This is the test: https://github.com/hazelcast/hazelcast-simulator/pull/728

This modifies the old Java client only. I have yet to check how to ports these changes into the new client.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -702,13 +702,12 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapKeySetRequest request = new MapKeySetRequest(name);
         MapKeySet mapKeySet = invoke(request);
         Set<Data> keySetData = mapKeySet.getKeySet();
-        InflatableSet<K> keySet = new InflatableSet<K>(keySetData.size());
+        InflatableSet.Builder<K> setBuilder = InflatableSet.newBuilder(keySetData.size());
         for (Data data : keySetData) {
             final K key = toObject(data);
-            keySet.add(key);
+            setBuilder.add(key);
         }
-        keySet.close();
-        return keySet;
+        return setBuilder.build();
     }
 
     @Override
@@ -768,16 +767,15 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapEntrySet result = invoke(request);
 
         Set<Entry<Data, Data>> entries = result.getEntrySet();
-        InflatableSet<Entry<K, V>> entrySet = new InflatableSet<Entry<K, V>>(entries.size());
+        InflatableSet.Builder<Entry<K, V>> setBuilder = InflatableSet.newBuilder(entries.size());
         for (Entry<Data, Data> dataEntry : entries) {
             Data keyData = dataEntry.getKey();
             Data valueData = dataEntry.getValue();
             K key = toObject(keyData);
             V value = toObject(valueData);
-            entrySet.add(new AbstractMap.SimpleEntry<K, V>(key, value));
+            setBuilder.add(new AbstractMap.SimpleEntry<K, V>(key, value));
         }
-        entrySet.close();
-        return entrySet;
+        return setBuilder.build();
     }
 
     @Override
@@ -791,13 +789,12 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapQueryRequest request = new MapQueryRequest(name, predicate, IterationType.KEY);
         QueryResultSet result = invoke(request);
         if (pagingPredicate == null) {
-            final InflatableSet<K> keySet = new InflatableSet<K>(result.size());
+            InflatableSet.Builder<K> setBuilder = InflatableSet.newBuilder(result.size());
             for (Object o : result) {
                 final K key = toObject(o);
-                keySet.add(key);
+                setBuilder.add(key);
             }
-            keySet.close();
-            return keySet;
+            return setBuilder.build();
         }
 
         Iterator<Entry> iterator = result.rawIterator();
@@ -823,14 +820,13 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         QueryResultSet result = invoke(request);
         if (pagingPredicate == null) {
             SerializationService serializationService = getContext().getSerializationService();
-            InflatableSet<Entry<K, V>> entrySet = new InflatableSet<Entry<K, V>>(result.size());
+            InflatableSet.Builder<Entry<K, V>> setBuilder = InflatableSet.newBuilder(result.size());
             for (Object data : result) {
                 AbstractMap.SimpleImmutableEntry<Data, Data> dataEntry = (AbstractMap.SimpleImmutableEntry<Data, Data>) data;
                 LazyMapEntry lazyEntry = new LazyMapEntry(dataEntry.getKey(), dataEntry.getValue(), serializationService);
-                entrySet.add(lazyEntry);
+                setBuilder.add(lazyEntry);
             }
-            entrySet.close();
-            return entrySet;
+            return setBuilder.build();
         }
         ArrayList<Map.Entry> resultList = new ArrayList<Map.Entry>();
         for (Object data : result) {
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -36,10 +36,12 @@ import com.hazelcast.core.IMap;
 import com.hazelcast.core.IMapEvent;
 import com.hazelcast.core.MapEvent;
 import com.hazelcast.core.Member;
+import com.hazelcast.internal.serialization.SerializationService;
 import com.hazelcast.logging.Logger;
 import com.hazelcast.map.EntryProcessor;
 import com.hazelcast.map.MapInterceptor;
 import com.hazelcast.map.MapPartitionLostEvent;
+import com.hazelcast.map.impl.LazyMapEntry;
 import com.hazelcast.map.impl.ListenerAdapter;
 import com.hazelcast.map.impl.MapEntrySet;
 import com.hazelcast.map.impl.MapKeySet;
@@ -114,6 +116,7 @@ import com.hazelcast.util.IterationType;
 import com.hazelcast.util.Preconditions;
 import com.hazelcast.util.QueryResultSet;
 import com.hazelcast.util.ThreadUtil;
+import com.hazelcast.util.collection.InflatableSet;
 import com.hazelcast.util.executor.CompletedFuture;
 import com.hazelcast.util.executor.DelegatingFuture;
 
@@ -699,11 +702,12 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapKeySetRequest request = new MapKeySetRequest(name);
         MapKeySet mapKeySet = invoke(request);
         Set<Data> keySetData = mapKeySet.getKeySet();
-        Set<K> keySet = new HashSet<K>(keySetData.size());
+        InflatableSet<K> keySet = new InflatableSet<K>(keySetData.size());
         for (Data data : keySetData) {
             final K key = toObject(data);
             keySet.add(key);
         }
+        keySet.close();
         return keySet;
     }
 
@@ -763,8 +767,8 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapEntrySetRequest request = new MapEntrySetRequest(name);
         MapEntrySet result = invoke(request);
 
-        Set<Entry<K, V>> entrySet = new HashSet<Entry<K, V>>();
         Set<Entry<Data, Data>> entries = result.getEntrySet();
+        InflatableSet<Entry<K, V>> entrySet = new InflatableSet<Entry<K, V>>(entries.size());
         for (Entry<Data, Data> dataEntry : entries) {
             Data keyData = dataEntry.getKey();
             Data valueData = dataEntry.getValue();
@@ -772,6 +776,7 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
             V value = toObject(valueData);
             entrySet.add(new AbstractMap.SimpleEntry<K, V>(key, value));
         }
+        entrySet.close();
         return entrySet;
     }
 
@@ -786,11 +791,12 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapQueryRequest request = new MapQueryRequest(name, predicate, IterationType.KEY);
         QueryResultSet result = invoke(request);
         if (pagingPredicate == null) {
-            final HashSet<K> keySet = new HashSet<K>();
+            final InflatableSet<K> keySet = new InflatableSet<K>(result.size());
             for (Object o : result) {
                 final K key = toObject(o);
                 keySet.add(key);
             }
+            keySet.close();
             return keySet;
         }
 
@@ -816,13 +822,14 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapQueryRequest request = new MapQueryRequest(name, predicate, IterationType.ENTRY);
         QueryResultSet result = invoke(request);
         if (pagingPredicate == null) {
-            Set<Entry<K, V>> entrySet = new HashSet<Entry<K, V>>(result.size());
+            SerializationService serializationService = getContext().getSerializationService();
+            InflatableSet<Entry<K, V>> entrySet = new InflatableSet<Entry<K, V>>(result.size());
             for (Object data : result) {
                 AbstractMap.SimpleImmutableEntry<Data, Data> dataEntry = (AbstractMap.SimpleImmutableEntry<Data, Data>) data;
-                K key = toObject(dataEntry.getKey());
-                V value = toObject(dataEntry.getValue());
-                entrySet.add(new AbstractMap.SimpleEntry<K, V>(key, value));
+                LazyMapEntry lazyEntry = new LazyMapEntry(dataEntry.getKey(), dataEntry.getValue(), serializationService);
+                entrySet.add(lazyEntry);
             }
+            entrySet.close();
             return entrySet;
         }
         ArrayList<Map.Entry> resultList = new ArrayList<Map.Entry>();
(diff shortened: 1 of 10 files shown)


jerrinot:
@ajermakovics, @ahmetmircik: Can you please have a look?

jerrinot:
verify

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -698,11 +699,12 @@ public void loadAll(Set<K> keys, boolean replaceExistingValues) {
         MapKeySetRequest request = new MapKeySetRequest(name);
         MapKeySet mapKeySet = invoke(request);
         Set<Data> keySetData = mapKeySet.getKeySet();
-        Set<K> keySet = new HashSet<K>(keySetData.size());
+        InflatableSet<K> keySet = new InflatableSet<K>(keySetData.size());
         for (Data data : keySetData) {
             final K key = toObject(data);
```
ahmetmircik: Can we also introduce lazy de-serialization for `InflatableSet`?

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
```diff
@@ -0,0 +1,223 @@
...
+ * and {@link #containsAll(Collection)} calls to enable fast look-ups.
+ *
+ * It's intended to be use in cases where a contract mandates us to return Set,
+ * but we know our data contains not duplicates. It performs the best in cases
+ * biased towards sequential iterations.
+ *
+ *
+ * @param <T>
+ */
+public class InflatableSet<T> extends AbstractSet<T> implements Set<T> {
```
ahmetmircik: Maybe you can add some note about thread-safety.

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
```diff
@@ -0,0 +1,223 @@
...
+    private final List<T> compactList;
+    private Set<T> inflatedSet;
+    private State state = State.INITIAL_LOAD;
+
+    public InflatableSet(int initialCapacity) {
+        if (initialCapacity < 0) {
+            throw new IllegalArgumentException("Initial Capacity must be a positive number. Current capacity: "
+                    + initialCapacity);
+        }
+        compactList = new ArrayList<T>(initialCapacity);
```
ahmetmircik: What about using an `ArrayDeque` instead ?

hazelcast/src/test/java/com/hazelcast/util/BitSetUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.runner.RunWith;
+
+import java.util.ArrayList;
+import java.util.BitSet;
+import java.util.List;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(HazelcastSerialClassRunner.class)
```
ahmetmircik: Can we use `HazelcastParallelClassRunner` here?

hazelcast/src/test/java/com/hazelcast/util/collection/InflatableSetTest.java
```diff
@@ -0,0 +1,226 @@
...
+import org.junit.runner.RunWith;
+
+import java.util.ConcurrentModificationException;
+import java.util.Iterator;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThat;
+import static org.hamcrest.Matchers.*;
+
+@RunWith(HazelcastSerialClassRunner.class)
```
ahmetmircik: Can we use `HazelcastParallelClassRunner` here?

hazelcast/src/test/java/com/hazelcast/util/BitSetUtilsTest.java
jerrinot: good one!

hazelcast/src/test/java/com/hazelcast/util/collection/InflatableSetTest.java
jerrinot: thanks!

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
jerrinot: then it'd make it dependent on SerializationService. that's not what I want.

gurbuzali: what's wrong to depend on SerializationService? why do you want to avoid it ?

jerrinot: as this is meant to be a Set with no external dependencies. I'll think about how to make it possible.

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
jerrinot: good idea!

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
jerrinot: why would be the advantage?

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
jerrinot: a better better expressed objections against introducing dependency on SS: Mixing concerns. Responsibility of this class is to inflate on modification/lookup. I don't want to deal with de-serialization in the same class - it has a nice responsibility right now. However I have an idea, stay tuned :)

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
ahmetmircik: I can not see a disadvantage but if you have a scenario which requires `add` and `remove` combinations in INITIAL_LOAD state, it will perform better.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
jerrinot: @gurbuzali, @ahmetmircik:

see this: https://github.com/jerrinot/hazelcast/blob/experiments/inflatableSet-lazyEntry/master/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java#L828-828

No reason to poison the Set with deserilization logic. How does it look? I think we can apply this composition for other methods too.

ahmetmircik: Sounds better but with that approach, isn't there a possibility that we are creating too many lazy-entries even we don't need to? Imagining a scenario that only needs to check the size of the returned key-set, in that case there is no need a set pre-populated with lazy-entries, i am expecting an approach which creates lazy-entries on demand, maybe i misunderstood your approach...

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
ahmetmircik: Although I am fine with this `Set` impl, i just only want to note that a minor point, previously returned `HashSet` was `Cloneable and Serializable` but now we are removing that functionality.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
jerrinot: Sure. However that's yet another responsibility. There are no serialized entries to begin with. We would need some sort of lazy-generator.

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
jerrinot: that's a very good point!

jerrinot:
verify

pveentjer:
@jerrinot is there any update? Perhaps collection more input from the reviewers?

jerrinot:
@pveentjer: I'll push an update today.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
jerrinot: I added the LazyEntry. I don't want to introduce another trickery at this point

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
jerrinot: I don't understand how ArrayDequee would be faster. (unless I expose methods like `removeLast()`, etc..

jerrinot:
verify

jerrinot:
So I consider this one to be done with the exception of the FindBugs issue - I need @Donnerbart's help here.

Testing scenario https://github.com/hazelcast/hazelcast-simulator/blob/master/tests/src/main/java/com/hazelcast/simulator/tests/map/SerializationStrategyTest.java
```
class=com.hazelcast.simulator.tests.map.SerializationStrategyTest
threadCount=1
itemCount=10000000
```
2 members, 1 client. 

Before
```
INFO  17:58:28 Running 00d 00h 00m 30s   10.00% complete              8 ops            0.39 ops/s
INFO  17:58:58 Running 00d 00h 01m 00s   20.00% complete             17 ops            0.33 ops/s
INFO  17:59:28 Running 00d 00h 01m 30s   30.00% complete             25 ops            0.33 ops/s
INFO  17:59:58 Running 00d 00h 02m 00s   40.00% complete             34 ops            0.33 ops/s
INFO  18:00:28 Running 00d 00h 02m 30s   50.00% complete             42 ops            0.17 ops/s
INFO  18:00:58 Running 00d 00h 03m 00s   60.00% complete             51 ops            0.33 ops/s
```

After
```
INFO  18:11:04 Running 00d 00h 00m 30s   10.00% complete            348 ops           15.24 ops/s
INFO  18:11:34 Running 00d 00h 01m 00s   20.00% complete            774 ops           14.54 ops/s
INFO  18:12:04 Running 00d 00h 01m 30s   30.00% complete          1,231 ops           14.94 ops/s
```

jerrinot:
@ahmetmircik @gurbuzali: Guys, can you please  have another look? Thanks!

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
```diff
@@ -0,0 +1,245 @@
...
+ * Provides fast {@link Set} implementation for cases where items are known to not
+ * contain duplicates.
+ *
+ * It doesn't not call equals/hash methods on initial data insertion hence it avoids
+ * performance penalty in the case these methods are expensive. It also means it does
+ * not detect duplicates - it's a responsibility of a caller to make sure no duplicated
+ * entries are inserted.
+ *
+ * Once the initial load is done then caller should indicate it
+ * by calling {@link #close()}. It switches operational mode and if any item is
```
ahmetmircik: Can you please point that what the implications are if one forgets to call "close" method ? I assume that may be done easily.

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
ahmetmircik: I see this as a micro-enhancement and thinking similar with you that `ArrayList` is also fine in most cases. But for my personal favour refactored your implementation and compared array-list-backed-inflatable-set with array-deque-backed one, sharing the results here (tested by adding and removing same number of elements) :

>
JMH 1.10.5 
VM version: JDK 1.8.0_60, VM 25.60-b23
Benchmark                                       Mode  Cnt      Score       Error  Units
inflatableSet.ArrayListVsArrayDeque.dequeuTest  avgt    3  14471.093 ± 10099.430  ns/op
inflatableSet.ArrayListVsArrayDeque.listTest    avgt    3  61844.015 ±  6738.956  ns/op

In the end, if you think it is worth to use `ArrayDeque` you can change it, but i am fine with `ArrayList` also.

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
```diff
@@ -142,5 +150,16 @@ public int hashCode() {
...
         return getKey() + "=" + getValue();
     }
+
+    private void readObject(java.io.ObjectInputStream s) throws IOException, ClassNotFoundException {
+        keyObject = s.readObject();
+        valueObject = s.readObject();
+
+    }
+
+    private void writeObject(java.io.ObjectOutputStream s) throws IOException {
```
ahmetmircik: What about calling `defaultWriteObject` method here? According to Effective Java it is a recommended practice. http://allenlsy.com/NOTES-of-Effective-Java-10/

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
```diff
@@ -142,5 +150,16 @@ public int hashCode() {
     public String toString() {
         return getKey() + "=" + getValue();
     }
+
+    private void readObject(java.io.ObjectInputStream s) throws IOException, ClassNotFoundException {
```
ahmetmircik: What about calling `defaultReadObject` method here? According to Effective Java it is a recommended practice. http://allenlsy.com/NOTES-of-Effective-Java-10/

jerrinot: done, thanks for your feedback!

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
jerrinot: done

pveentjer:
*impressed by discussions going on here*

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
jerrinot: thanks! I'll have a closer look and possible I'll send another PR with the change.

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
```diff
@@ -0,0 +1,251 @@
...
+ * never pass unclosed InfltableSet to user code.
+ *
+ * It's intended to be use in cases where a contract mandates us to return Set,
+ * but we know our data contains not duplicates. It performs the best in cases
+ * biased towards sequential iterations.
+ *
+ *
+ * @param <T>
+ */
+public final class InflatableSet<T> extends AbstractSet<T> implements Set<T>, Serializable, Cloneable {
```
ahmetmircik: Do you think we need to test serializability & de-serializability of `InflatableSet`?

jerrinot: good one, I'll add it.

jerrinot: done

ahmetmircik:
:+1:

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
```diff
@@ -0,0 +1,251 @@
...
+                break;
+            case INFLATED:
+                inflatedSet.clear();
+                break;
+            default:
+                compactList.clear();
+        }
+    }
+
+    public void close() {
```
ajermakovics: Closing a collection might be a bit misleading. If I close something I wouldn't expect to add stuff to it but here I can. Perhaps we could use the Builder pattern to construct an InflatableSet and pass it a List as a constructor argument.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -815,13 +821,14 @@ public void loadAll(Set<K> keys, boolean replaceExistingValues) {
         MapQueryRequest request = new MapQueryRequest(name, predicate, IterationType.ENTRY);
         QueryResultSet result = invoke(request);
```
ajermakovics: QueryResultSet has entries in a Set which are then added to a List here. It might be another optimisation to store QueryResultSet.entries as a List when it is deserialized.

jerrinot: but what if someone add a new item into the `QueryResultSet` after deserialization? We could  have `QueryResultSet` backed by `InflatableSet` but I don't think it's related to this PR.

hazelcast/src/main/java/com/hazelcast/util/collection/InflatableSet.java
jerrinot: @ajermakovics: as discussed: 
I don't think a builder pattern it justifiable here. I agree the `close()` is not the best name here; I appreciate your suggestion how to name it better.

jerrinot:
verify

jerrinot:
@ahmetmircik: I have slightly changed it. It now uses the builder pattern as recommended (and force:)) by @ajermakovics; It prevents accidental passing of unclosed InflatableSet to user code. Could you please have another look? Thanks!

ajermakovics:
the builder also makes InflatableSet's internal implementation bit simpler :+1:

ajermakovics:
Findbugs is complaining about the Cloneable interface

jerrinot:
oh, I'll have a look

jerrinot:
fixed thanks to @Donnerbart!

sancar:
@jerrinot How about new client ? 

And we have an interesting enterprise test failures related to this.
java.lang.NoClassDefFoundError: com.hazelcast.util.collection.InflatableSet

 https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-ee-IbmJDK1.6/com.hazelcast$hazelcast-enterprise-client/460/testReport/junit/com.hazelcast.client.security/MapSecurityInterceptorTest/test1_entrySet/

jerrinot:
@sancar: I'll have a look at the new client.

The failures is highly suspicious. Do you have any explanation what's going on?

sancar:
@jerrinot It is suspicious. It says class def not found but it knows name of the class.

PR #6044, opened 2015-08-24
Fixes a bug when a target thread is waiting on Selector.select()

jerrinot:
When a task does not belong to the current thread then the current thread should move the
 task using `addTaskAndWakeup()` as otherwise it's possible the target thread will not process
 it until the Selector.select() times out.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
@@ -138,7 +138,7 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
      * @param task the task to add.
      * @throws NullPointerException if task is null
      */
-    public final void addTaskAndWakeup(Runnable task) {
+    public void addTaskAndWakeup(Runnable task) {
         taskQueue.add(task);
         if (!selectNow) {
             selector.wakeup();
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -109,7 +109,7 @@ public final void addTask(Runnable task) {
      * @param task the task to add.
      * @throws NullPointerException if task is null
      */
-    public final void addTaskAndWakeup(Runnable task) {
+    public void addTaskAndWakeup(Runnable task) {
```
jerrinot: I'm not quite sure how to test it without making this method non-final. (as final methods can't be mocked)

pveentjer: Difficult. I think it is fine without adding a test here. Since it the same behavior as before + adds notification on top of that.

jerrinot:
I'l create a backport once this is reviewed and merged.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
jerrinot: so do you prefer to keep this final and not to have a test for this?

pveentjer:
verify

jerrinot:
run-parallel

jerrinot:
run-parallel

jerrinot:
verify

pveentjer:
:+1:

mdogan:
:+1:

PR #6047, opened 2015-08-25
Minor io thread exception handling tweak

pveentjer:
When a selection key is invalid, the handler is being called with a cancelled key exception. So we now feed this exception back into the handler so it can deal with it (e.g. close the connection). Before it was just ignored.

Also added a lot of tests for faulty and select behavior.

Also added a probe that counts the number of io exceptions the selector.select/selectNow methods have run into. These exceptions are not terminal.. so we want to know how many of such exceptions we got.

Also pulled out a 'coolDown' method from the run method to simplify the code a bit.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
@@ -23,6 +23,7 @@ import com.hazelcast.spi.impl.operationexecutor.OperationHostileThread;
 import com.hazelcast.util.counters.SwCounter;
 
 import java.io.IOException;
+import java.nio.channels.CancelledKeyException;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.util.Iterator;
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -66,12 +69,25 @@ public NonBlockingIOThread(ThreadGroup threadGroup,
                                ILogger logger,
                                NonBlockingIOThreadOutOfMemoryHandler oomeHandler,
                                boolean selectNow) {
+        this(threadGroup, threadName, logger, oomeHandler, selectNow, newSelector());
+    }
+
+    public NonBlockingIOThread(ThreadGroup threadGroup,
```
jerrinot: Do we really need 3 constructors?

pveentjer: Well. you can always reduce to a single constructor :)

hazelcast/src/test/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThreadTest.java
```diff
@@ -0,0 +1,288 @@
...
+        startThread();
+        assertEquals(thread.getName(), thread.toString());
+    }
+
+    @Test
+    public void addTaskAndWakeup() {
+        startThread();
+
+        // give the thread some time to start; it will probably be blocking on the selector by now
+        sleepSeconds(1);
```
jerrinot: ಠ_ಠ

jerrinot:
:+1:

gurbuzali:
what happened to our at least 2 person other than the sender should review ?

pveentjer:
We parked that some time ago since otherwise no pr's would get through.

gurbuzali:
who parked that ? who made the decission ?

PR #6049, opened 2015-08-25
Changes after reviewing protocol AtomicLong AtomicReference and JCache

sancar:
1) AtomicLong.alter response converted to void from Long.
2) JCache.managementConfig response converted to void from Data.
3) New response type CACHE_KEY_ITEREATOR_RESULT is introduced.
4) JCache.iterate response converted to CACHE_KEY_ITEREATOR_RESULT from Data.
5) Basic iterator test is implemented.
6) A couple null checks removed in cache proxy where it is already null checked or impossible
to be null.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
@@ -229,7 +229,7 @@ abstract class AbstractClientCacheProxy<K, V>
         if (keys.isEmpty()) {
             return Collections.EMPTY_MAP;
         }
-        final Set<Data> keySet = new HashSet(keys.size());
+        final Set<Data> keySet = new HashSet<Data>(keys.size());
         for (K key : keys) {
             final Data k = toData(key);
             keySet.add(k);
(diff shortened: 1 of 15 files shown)


sancar:
verify

sancar:
verify

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
```diff
@@ -235,12 +235,9 @@ public void clear() {
         }
         final Data keyData = toData(key);
         Data epData = toData(entryProcessor);
-        List<Data> argumentsData = null;
-        if (arguments != null) {
-            argumentsData = new ArrayList<Data>(arguments.length);
-            for (int i = 0; i < arguments.length; i++) {
-                argumentsData.add(toData(arguments[i]));
-            }
+        List<Data> argumentsData = new ArrayList<Data>(arguments.length);
```
ihsandemir: Is the arguments variable guaranteed to be non-null?

sancar: Yes

sancar: It seems if user passes null we get NULL arguments. I was expecting to get a list with 1 element whose first element is null. Fixing it now.

ihsandemir:
:+1:

asimarslan:
:+1:

PR #6050, opened 2015-08-25
Partition movement detection uses partitionStateVersion instead just comparing owned partitions

jerrinot:
It's faster and more importantly it fixes the situation when there is a partition
movement while Query is running, but the member ends-up owning the same partitions
as in the beginning.


gurbuzali:
:+1:

jerrinot:
verify

ahmetmircik:
:+1:

jerrinot:
verify

PR #6051, opened 2015-08-25
Unneccessary deserialization removed

jerrinot:
When source map is in OBJECT in memory format then there we don't have to serialize and deserialize
entries.

I don't expect this was intentional.


gurbuzali:
:+1:

pveentjer:
1: Always good to remove things not needed
2: Do you know already the impact on performance
3: Is there a test that verified this code?

jerrinot:
Throughput of this tests is almost doubled with this PR: https://github.com/jerrinot/hazelcast-stabilizer/commit/841e0d1aaaab8e097388bc1d889d5a3eacb716e0 (when OBJECT format is used)

There are existing test for this in Hazelcast code. I'm not changing behaviour of `MapKeyValueSource`, just its internal implementation.

pveentjer:
:+1:

PR #6053, opened 2015-08-25
[DONT MERGE] Api/3.x/serializationservice

pveentjer:


sancar:
It looks like you changed all SerializationService usages to InternalSerializationService. I think it is better to leave SerializationService where it is sufficient. Especially distributed object proxy and service classes.

pveentjer:
I already converted many back.. but sure.. I'll try to 'dumb' down where e possible.

PR #6057, opened 2015-08-26
Changes after protocol review

sancar:
Rest of the work started with #6049.

1) Manually written MemberAttributeChangeCodec is removed and its definition is moved to codec template.
2) MultiMapKeySetMessageTask return type changed to set of data from list of data.
3) EnterpriseMap.publisherCreate return type changed to set of data since it is returning keys.
4) DisrnibutedObject name made plural in response template because it returns list of DistributedObject info.
5) Fixed javadoc mistakes encountered.


asimarslan:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/multimap/MultiMapKeySetMessageTask.java
```diff
@@ -54,7 +54,7 @@ protected OperationFactory createOperationFactory() {
 
     @Override
     protected Object reduce(Map<Integer, Object> map) {
-        List<Data> keys = new ArrayList<Data>();
+        Set<Data> keys = new HashSet<Data>();
```
ihsandemir: How do we know no duplicates will exist? If there are duplicates, this change may cause loss of data.

gurbuzali: there can't be duplicates, it is after all `keySet` not `values`

ihsandemir:
:+1:

PR #6058, opened 2015-08-26
Improved Packet version mismatch exception message

pveentjer:


hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -188,8 +188,11 @@ private boolean readVersion(ByteBuffer source) {
             byte version = source.get();
             setPersistStatus(PERSIST_VERSION);
             if (VERSION != version) {
-                throw new IllegalArgumentException("Packet versions are not matching! Expected -> "
-                        + VERSION + ", Incoming -> " + version);
+                throw new IllegalArgumentException(
+                        "Packet version mismatch; the incoming packet version doesn't match expected packet version,"
+                                + " incoming packet version: " + version
+                                + " expected packet version: " + VERSION + '.'
+                                + " If the incoming version is negative or very large, it could indicate a corrupted stream.");
```
jerrinot: just an idea: what about including this only if the incoming is negative or large ?

jerrinot:
Good One! :+1:

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
pveentjer: +1

I'll play with that a bit.

jerrinot: perhaps you could give a hint: There is a different version of Hazelcast running.

jerrinot: @pveentjer: ping

PR #6062, opened 2015-08-26
Fixes #6020. 

### LINKED ISSUE #6020: [TEST-FAILURE] executor.CompletableFutureTest.postregisterCallback
```
java.lang.AssertionError: CountDownLatch failed to complete within 120 seconds , count left: 1
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:718)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:711)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:703)
	at com.hazelcast.executor.CompletableFutureTest.postregisterCallback(CompletableFutureTest.java:104)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3630/testReport/junit/com.hazelcast.executor/CompletableFutureTest/postregisterCallback/

tombujok:
Fix for #6020 

* As suggested by Peter the AbstractCompletableFuture class has been extended so that it may be a base class for all CompletableFuture classes - in this way we omit duplication in the execution of callbacks, which was a root cause of #6020. Some classes executed the callbacks correctly - some, like the CompletableFutureTask did not - now it's unified.
* Currently, AbstractCompletableFuture class handles the cancellation and get(timeout). It enables us to remove the `extends FutureTask` in the ComputableFutureTask. As a result we create less litter and do not track the state of the task in two classes at once. Earlier it was done by the  ComputableFutureTask and the base FutureTask; what is more it was not fully in sync either.
* This change was supposed to be minimalistic so no other changes has been done in the logic of the AbstractCompletableFuture class - apart from the things mentioned above. It may be a topic for future improvements. 
* There's some more classes that encompass the logic of callbacks' execution in the codebase (listed below). They could be refactored now, so that they extend the AbstractCompletableFuture - in this way we would omit even more duplication.
   * InvocationFuture
   * ClientInvocationFuture
   * CompletedFuture

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
@@ -22,6 +22,7 @@ import com.hazelcast.logging.ILogger;
 import com.hazelcast.spi.ExecutionService;
 import com.hazelcast.spi.NodeEngine;
 
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
@@ -32,9 +33,17 @@ import static com.hazelcast.util.ExceptionUtil.sneakyThrow;
 import static com.hazelcast.util.Preconditions.isNotNull;
 import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
+/**
+ * A base {@link ICompletableFuture} implementation that may be explicitly completed by setting its
+ * value through setResult.
+ * Implements the logic of cancellation and callbacks execution.
+ *
+ * @param <V> The result type returned by this Future's {@code get} method
+ */
 public abstract class AbstractCompletableFuture<V> implements ICompletableFuture<V> {
 
-    static final Object INITIAL_STATE = new ExecutionCallbackNode(null, null, null);
+    private static final Object INITIAL_STATE = new ExecutionCallbackNode(null, null, null);
+    private static final Object CANCELLED_STATE = new Object();
 
     private static final AtomicReferenceFieldUpdater<AbstractCompletableFuture, Object> STATE
             = newUpdater(AbstractCompletableFuture.class, Object.class, "state");
@@ -45,7 +54,7 @@ public abstract class AbstractCompletableFuture<V> implements ICompletableFuture
     // The reason for this abuse of the type system is to deal with the head node and the result
     // in a single cas. Using a single cas prevent a thread that calls and then do this concurrently
     // with a thread that calls setResult.
-    volatile Object state = INITIAL_STATE;
+    private volatile Object state = INITIAL_STATE;
 
     private final ILogger logger;
     private final Executor defaultExecutor;
@@ -72,7 +81,11 @@ public abstract class AbstractCompletableFuture<V> implements ICompletableFuture
         for (; ; ) {
             Object currentState = this.state;
 
-            if (isDone(currentState)) {
+            if (isCancelledState(currentState)) {
+                return;
+            }
+
+            if (isDoneState(currentState)) {
                 runAsynchronous(callback, executor, currentState);
                 return;
             }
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
```diff
@@ -110,6 +167,32 @@ public V get() throws InterruptedException, ExecutionException {
...
+            if (Thread.interrupted()) {
+                throw new InterruptedException();
+            }
+
+            nanosToWait = deadlineNanoTime - System.nanoTime();
+            if (nanosToWait < 0) {
+                throw new TimeoutException();
+            }
+
+            synchronized (this) {
```
jerrinot: scope of this synchronized block is suspiciously small. the smells like a race-condition to me. usually you want to check your condition while owning the mutex.

jerrinot: Imagine 2 threads: 

- Thread1 - Waiter
- Thread2 - Producer

Waiter calls `get()`. Result is still not set so the Waiter will fall through up to line 190. But before Waiter enters the synchronized block the Producer thread will set the result. The Producer will call `notifyThreadsWaitingOnGet()`, but there is no thread waiting! (as the Waiter is right *before* the synchronized block)

Eventually Waiter reaches the synchronized block and calls `wait()`, but there is no one to call notify! (as the Producer is already gone). Did I miss something?

mtopolnik:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
tombujok: Thanks @jerrinot  I fixed it!

jerrinot: resolved

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
```diff
@@ -110,25 +165,71 @@ public V get() throws InterruptedException, ExecutionException {
         }
     }
 
+    @Override
+    public V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+        final long deadlineNanoTime = System.nanoTime() + unit.toNanos(timeout);
```
jerrinot: just a small one: It's easier to work in ms rather then ns. `Object.wait()` will convert it to ms anyway -> you can save potentially expensive modulo operation, but more importantly - it will simplify the code.

jerrinot:
verify

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
tombujok: Thx. Fixed as suggested.

jerrinot:
:+1: from me. 

Normally having two :+1: would be enough to merge it. However this is a very core part of the system and I believe one more pair of eyes would be great. I'm nominating @metanet as he has a good knowledge in this area.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
```diff
@@ -110,25 +163,71 @@ public V get() throws InterruptedException, ExecutionException {
...
+                throw new InterruptedException();
+            }
+
+            millisToWait = deadlineTimeMillis - System.currentTimeMillis();
+            if (millisToWait <= 0) {
+                throw new TimeoutException();
+            }
+
+            synchronized (this) {
+                if (!isDoneState(this.state)) {
```
gurbuzali: I think we may have still race here for cancellation

jerrinot: I believe it's fine. As a cancelled future is "done". @tombujok: Perhaps you should rename the method to `isFinishedOrCancelled`? or something like this. I made the same mistake while reviewing it.

gurbuzali: I see

mtopolnik: @jerrinot My suggestion was `isFinalState` to imply that no more state changes can occur, but avoid the implications of having completed successfully. But I must say I also argued for the removal of comments explaining what "done" means for a Future because it is already documented in the official Oracle docs for `Future`. OTOH I do suggest writing a short Javadoc comment on top of `isDoneState` which reinforces the meaning.

jerrinot: +1 to `isFinalState()`

tombujok: The idea was to have an analogous naming convention to the Future's `isDone()` and `isCancelled()` one, thus we named it `isDoneState()` and `isCancelledState()`.

Introducing another word like `final` doesn't help much here IMHO, because somebody still has to know if `finalState` encompasses `cancelled` or not.
I would call the method `isDoneOrCancelled()` - then it's obvious what it means.

WDYT?

mtopolnik: I just hate admitting in a method name that we don't know the semantics of a "done" future :) Keeping the current name and adding a Javadoc which reminds of the meaning of "done" would sit nicer with me.

tombujok: OK, I've added a javadoc to the isDoneState()

hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
```diff
@@ -23,158 +23,350 @@
...
+        f.andThen(storeTaskResponseToReference(reference2));
+
+        releaseAwaitingTask();
+
+        assertCallbacksExecutedEventually();
+        assertTestExceptionThrown(reference1, reference2);
     }
 
     @Test
+    @Repeat(10)
```
metanet: I suppose there is a @Repeat left-over here.

metanet:
:+1:  with a minor comment

hazelcast/src/test/java/com/hazelcast/executor/CompletableFutureTest.java
tombujok: Thx for spotting - I left it intentionally there just increase to chance of a failing test should there be any hidden bug. Would you prefer to have all the `@Repeat`s deleted? These tests are rather fast, so it's not a huge bottleneck

gurbuzali:
I would like `Repeat`s removed, after all our CI is repeating the tests daily.
I'll merge this one, @tombujok can you please remove the `Repeat`s with a direct commit, no need a PR for that.

tombujok:
Thx for the merge. `@Repeat`s removed with a direct commit:
https://github.com/hazelcast/hazelcast/commit/11ababded2d01fccf0b5f0ab4f26ac122429c291

PR #6067, opened 2015-08-26
Cleanup of CheckStyle suppressions for DebugUtil

Donnerbart:
Not much to say, was close to directly commit this.


pveentjer:
Just remove the class. I can get back when I need to.

Donnerbart:
I added another commit to delete the file. But I kept the previous one, so we have a CheckStyle compliant version in the history for restore.

pveentjer:
:+1:

jerrinot:
:+1:

PR #6070, opened 2015-08-27
add ordering to code generation.

sancar:
Distributed objects and their operations will always be generated in same order.

### DIFF
--- a/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodecModel.java
+++ b/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodecModel.java
@@ -36,6 +36,7 @@ public class CodecModel implements Model{
     static final Map<String, TypeElement> CUSTOM_CODEC_MAP = new HashMap<String, TypeElement>();
 
     private final Lang lang;
+    private short requestId;
     private String id;
     private String name;
     private String className;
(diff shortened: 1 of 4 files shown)


hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodecModel.java
```diff
@@ -153,9 +154,9 @@ public CodecModel(TypeElement parent, ExecutableElement methodElement, Executabl
         this.lang = lang;
 
         name = methodElement.getSimpleName().toString();
-        short requestId = methodElement.getAnnotation(Request.class).id();
+        requestId = methodElement.getAnnotation(Request.class).id();
         short masterId = parent.getAnnotation(GenerateCodec.class).id();
-        id = CodeGenerationUtils.mergeIds(masterId, requestId);
+        id = CodeGenerationUtils.addHexPrefix(CodeGenerationUtils.mergeIds(masterId, requestId));
```
ihsandemir: Lets decrease this to a single call.

sancar: I subtracted the method from mergeId un purpose because I needed the int without hex prefix somewhere else. I dont see any value creating a new method like `mergeIdsWithHexPrefix`. 
This way two concerns are in separate methods.

ihsandemir:
:+1:

Issue #5992, opened 2015-08-18
[TEST-FAILURE] ClientMapWriteQuorumTest failures


hasancelik:
```
java.lang.AssertionError: Expected exception: com.hazelcast.quorum.QuorumException
	at org.junit.internal.runners.statements.ExpectException.evaluate(ExpectException.java:32)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:134)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at com.hazelcast.test.HazelcastSerialClassRunner.runChild(HazelcastSerialClassRunner.java:44)
	at com.hazelcast.test.HazelcastSerialClassRunner.runChild(HazelcastSerialClassRunner.java:26)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/com.hazelcast$hazelcast-client-new/155/#showFailuresLink

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/3618/

Issue #5813, opened 2015-07-29
[TEST-FAILURE] TransactionalMapReadQuorumTest failures


hasancelik:
com.hazelcast.quorum.TransactionalMapReadQuorumTest.testTxGetThrowsExceptionWhenQuorumSizeNotMet[Executing: TransactionOptions{timeoutMillis=120000, durability=1, txType=TWO_PHASE}]

com.hazelcast.quorum.TransactionalMapReadQuorumTest.testTxContainsKeyThrowsExceptionWhenQuorumSizeNotMet[Executing: TransactionOptions{timeoutMillis=120000, durability=1, txType=TWO_PHASE}]

com.hazelcast.quorum.TransactionalMapReadQuorumTest.testTxGetThrowsExceptionWhenQuorumSizeNotMet[Executing: TransactionOptions{timeoutMillis=120000, durability=1, txType=LOCAL}]

com.hazelcast.quorum.TransactionalMapReadQuorumTest.testTxContainsKeyThrowsExceptionWhenQuorumSizeNotMet[Executing: TransactionOptions{timeoutMillis=120000, durability=1, txType=LOCAL}]

```
java.lang.AssertionError: Expected exception: com.hazelcast.transaction.TransactionException
	at org.junit.internal.runners.statements.ExpectException.evaluate(ExpectException.java:32)
	at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:134)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at com.hazelcast.test.HazelcastParallelClassRunner.access$201(HazelcastParallelClassRunner.java:29)
	at com.hazelcast.test.HazelcastParallelClassRunner$TestRunner.run(HazelcastParallelClassRunner.java:89)
	at java.lang.Thread.run(Thread.java:738)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.6/626/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/151/

PR #6071, opened 2015-08-27
fix race condition in quorum tests startup

mustafaiman:
Quorum tests starts after a successful split of a five member cluster. There is a short interval between successful split of the cluster and quorum service getting notified about it. This situation causes some tests to fail if the test starts before quorum service getting notified. This pr makes the tests wait for quorum service getting successfully notified.

Closes https://github.com/hazelcast/hazelcast/issues/5992 https://github.com/hazelcast/hazelcast/issues/5813


mesutcelik:
:+1:

mesutcelik:
Can you please create backport for this PR? I'm sure there are test-failures happening in our maintenance builds as well.

Issue #6072, opened 2015-08-27
hazelcast-wm >= 3.3 OSGI Manifest spring package imports should not be required


fwassmer:
Starting with hazelcast-wm 3.3.0 spring packages are imported and required in the Manifest.mf,
so hazlecast-wm can only be started when spring is also installed:

>org.springframework.context;version="[3.1,4)",
>org.springframework.security.core.session;version="[3.1,4)",
>org.springframework.security.web.session;version="[3.1,4)",
>org.springframework.web.context;version="[3.1,4)",
>org.springframework.web.context.support;versio n="[3.1,4)"

Those dependencies should be set to be optional with resolution:=optional
in the pom (  `<Import-Package>...` ):
e.g.: `org.springframework.context;version="[3.1,4)";resolution:=optional`

bilalyasar:
@fwassmer yes your findings seems logical. If you want, you can create PR about this issue..

fwassmer:
PR created: https://github.com/hazelcast/hazelcast/pull/6073

serkan-ozal:
closed via https://github.com/hazelcast/hazelcast/pull/6073

PR #6073, opened 2015-08-27
hazelcast-wm : OSGI package imports set to resolution:=optional

fwassmer:
PR for issue: https://github.com/hazelcast/hazelcast/issues/6072

### DIFF
--- a/hazelcast-wm/pom.xml
+++ b/hazelcast-wm/pom.xml
@@ -105,6 +105,11 @@
                                     org.apache.logging.log4j;resolution:=optional,
                                     org.apache.logging.log4j.*;resolution:=optional,
                                     org.slf4j;resolution:=optional,
+                                    org.springframework.context;resolution:=optional,
+                                    org.springframework.security.core.session;version="[3.2,4)";resolution:=optional,
+                                    org.springframework.security.web.session;version="[3.2,4)";resolution:=optional,
+                                    org.springframework.web.context;resolution:=optional,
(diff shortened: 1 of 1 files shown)


mesutcelik:
Many thanks for the contribution :+1: 

Can you please sign our contributor agreement so that we can merge the PR?

hazelcast-wm/pom.xml
```diff
@@ -105,6 +105,11 @@
                                     org.apache.logging.log4j;resolution:=optional,
                                     org.apache.logging.log4j.*;resolution:=optional,
                                     org.slf4j;resolution:=optional,
+                                    org.springframework.context;resolution:=optional,
+                                    org.springframework.security.core.session;version="[3.2,4)";resolution:=optional,
```
bilalyasar: @fwassmer , Why do you provide version here? Is it intentional? What is the logic behind this?

fwassmer: I've added the imports as they were generated by bnd (maven-bundle-plugin), so that there is no change compared to the current manifest (if you look at the current manifest, there is also a version added to those two packages).
bnd resolves the version by looking into the manifest of the dependency and when there is a version declared on the package export, it's also applied on the import package.
Adding this version range makes sure, that the the osgi container wires at least the version, which the bundle was built against.

bilalyasar: :+1:

PR #6075, opened 2015-08-28
added putAll count to MapStats

bilalyasar:


hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllOperation.java
```diff
@@ -72,6 +74,7 @@ public void run() {
         backupEntrySet = new ArrayList<Map.Entry<Data, Data>>();
         int partitionId = getPartitionId();
         final MapServiceContext mapServiceContext = mapService.getMapServiceContext();
+        LocalMapStatsImpl localMapStats = mapServiceContext.getLocalMapStatsProvider().getLocalMapStatsImpl(name);
```
ahmetmircik: This part seems running on remote instead of local. So i guess we can not count it as a local map stat. To add `putAll` count in local map stats, it should be calculated and should be added on caller side or am i missing something?

hazelcast/src/test/java/com/hazelcast/monitor/TimedMemberStateTest.java
```diff
@@ -80,4 +85,10 @@ public void testReplicatedMapGetStats() {
         ReplicatedMapService replicatedMapService = nodeEngine.getService(ReplicatedMapService.SERVICE_NAME);
         assertNotNull(replicatedMapService.getStats().get("replicatedMap"));
     }
+
+    @Test
+    public void testMapPutAllStats() {
+        LocalMapStatsImpl localMapStats = (LocalMapStatsImpl) timedMemberState.getMemberState().getLocalMapStats("name");
```
ahmetmircik: I believe, to run some `TimedMemberStateTest ` tests on multiple nodes and to assert results according to each node, if possible, will make tests more reliable.

bilalyasar:
i will send another PR. closing this one.

PR #6077, opened 2015-08-28
configurationURL inconsistencies fixed

### LINKED ISSUE #6061: XmlBuilder does not set URL in the config it parses
I am integrating Hazelcast in Spring Boot and I found what I think is an inconsistency. If I create the `HazelcastInstance` with the _default_ `newHazelcastInstance` static method then I get an `HazelcastInstance` whose `Config` has a valid `configurationUrl` property.

However, the following code gives me a null `configurationUrl`:

```
Config cfg = new XmlConfigBuilder(url).build();
```

Looking at `XmlConfigBuilder` it doesn't do anything with the parameter (except initializing the input stream). How about specifying the `configurationUrl` as well?

mesutcelik:
closes #6061

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
@@ -116,6 +116,7 @@ public class XmlConfigBuilder extends AbstractConfigBuilder implements ConfigBui
      */
     public XmlConfigBuilder(String xmlFileName) throws FileNotFoundException {
         this(new FileInputStream(xmlFileName));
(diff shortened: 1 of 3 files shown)


bilalyasar:
https://github.com/hazelcast/hazelcast/pull/6077/files#diff-3dffd84a99122128d983e9c2491c4668R118 
Do we need to set `configurationFile` here?

mesutcelik:
@bilalyasar ,
updated the PR.

bilalyasar:
:+1:  looks good to merge.

pveentjer:
:+1:

pveentjer:
Do we need a backport?

PR #6078, opened 2015-08-28
Missing security interceptor related methods are implemented.

sancar:


eminn:
:+1:

gurbuzali:
we need to add tests for these to the enterprise repository

sancar:
It is ready on my local. I will send the pr soon.

Issue #6080, opened 2015-08-28
Inconsistent and potentially buggy design in BasicCompletableFuture


tombujok:
While working on a follow up of the #6062 together with Marko we've noticed some inconsistent and potentially buggy design in BasicCompletableFuture.
A pull request with an Unit Test will document what is meant here.

We've also came up with a couple of improvements that may be done in BasicCompletableFuture and AbstractCompletableFuture as soon as #6062 is merged.

jerrinot:
@tombujok: What you are planning to do with this issue?

tombujok:
@jerrinot Should be fixed through #6106

PR #6081, opened 2015-08-28
Tests BasicCompletableFuture and documents its inconsistencies.

tombujok:
Related to #6080 

* Tests BasicCompletableFuture's behaviour to find out what its purpose actually is. 
* Documents its inconsistencies.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/executionservice/impl/BasicCompletableFutureTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/executionservice/impl/BasicCompletableFutureTest.java
@@ -1,0 +1,329 @@
+package com.hazelcast.spi.impl.executionservice.impl;
+
+import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.spi.ExecutionService;
+import com.hazelcast.spi.NodeEngine;
+import com.hazelcast.test.annotation.QuickTest;
+import com.hazelcast.util.executor.ManagedExecutorService;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/spi/impl/executionservice/impl/BasicCompletableFutureTest.java
```diff
@@ -0,0 +1,275 @@
...
+        delegateThrowException = true;
+
+        delegateFuture.run();
+
+        expected.expectMessage("Exception in execution");
+        outerFuture.get();
+
+        // callbacks will be never reached
+        outerFuture.andThen(callback);
+
```
mtopolnik: Doesn't `outerFuture.get` legitimately fail with an exception? Then you never even register the callback. I think the registration should happen before `delegateFuture.run()`.

tombujok: Yes, that's true - that's the reason I've put this comment `// callbacks will be never reached`
But I agree it may be confusing - I'll remove everything below `outerFuture.get()` and just leave a one line of comment there.

mtopolnik:
:+1:

hazelcast/src/test/java/com/hazelcast/spi/impl/executionservice/impl/BasicCompletableFutureTest.java
```diff
@@ -0,0 +1,296 @@
...
+    public void completeDelegate_bothDone_outerAskedFirst() {
+        delegateFuture.run();
+
+        assertTrue(outerFuture.isDone());
+        assertTrue(delegateFuture.isDone());
+    }
+
+    @Test
+    // PROBLEM NR.1
+    // If you complete the BasicCompletableFuture through setResult it will return isDone() == true,
```
jerrinot: what's a remedy here? perhaps `BasicCompletableFuture` should override `setResult()` with a final modifier?

mtopolnik: This is the current idea: AbstractCompletableFuture should not expose setResult, it should be a protected method. Note that ICompletableFuture actually defines the API of a _listenable_,  _noncompletable_ future. This will now cause major confusion as we also plan a subclass of ACF which exposes setResult -- what should we name it? ThisTimeForRealCompletableFuture?

pveentjer: LessAbstractCompletableFuture :)

tombujok: As Marko said the fix for Nr.1 is marking the setResult() method `protected` - it definitely shouldn't be `public` by default. It will be the implementer's responsibility to either expose the `protected setrResult()` method or not. 

For now, I think that we will skip the `subclass of ACF which exposes setResult`. It may be too confusing. 

@jerrinot Could you be the second reviewer? I think you had a thorough look at this test anyway :)

jerrinot:
:+1:

Issue #6083, opened 2015-08-29
MapProxySupport.putAllInternal should use setInternal, ClientMapProxy.putAll/getAll should invoke on key owners and not random target


LoneRifle:
The following improvements to IMap.putAll/getAll implementations are proposed:

* Given that IMap.putAll has a void return type, MapProxySupport.putAllInternal should use setInternal (and not putInternal) if the number of entries to be put is fewer than thrice the partition count. setInternal returns void, while putInternal returns the previous value mapped to the key, so using setInternal will avoid sending unwanted data over the cluster.

* ClientMapProxy.putAll/getAll should behave more like their MapProxyImpl counterparts. It currently implements putAll and getAll by invoking the corresponding request on a random target, even if the target may not be the owner of all the keys to be put/get. This incurs the following unnecessary overhead:
  * there is one extra hop in network activity, since the client is not directly going to key owners, and;
  * the random target has to devote heap memory to deal with objects it does not own.

This can be avoided by making ClientMapProxy.putAll/getAll work out which keys belong to which key owners and invoking on them accordingly.

PR #6084, opened 2015-08-29
MapProxySupport.putAllInternal: use setInternal instead of putInternal

LoneRifle:
setInternal has a void return type, which fits the return type of
putAllInternal, and avoids retrieving the previous value mapped to the
key that putInternal does

Implements part of #6083


jerrinot:
verify

jerrinot:
This looks like a good improvement. :+1: from me.

jerrinot:
verify

jerrinot:
verify

LoneRifle:
Let me close this PR and file a new one, doesn't seem like the CI job is kicking off correctly.

PR #6085, opened 2015-08-29
[DONT-MERGE] ClientMapProxy should go direct to key owners

LoneRifle:
Port code from MapProxySupport so that MapPutAllRequests are sent
directly to key owners rather than mediated by a randomly chosen member
of the cluster

Implements part of #6083 

DO NOT MERGE - This PR currently breaks behaviour in that events relating to putAll operations will no longer fire since set instead of put is being called. This will be replaced by something else pending the acceptance of #6093

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -10,6 +10,7 @@
     <suppress checks="StrictDuplicateCode" files="\.java" lines="1-15"/>
 
     <!-- Suppress checking of copyright notice -->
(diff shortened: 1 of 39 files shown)


jerrinot:
Hi @LoneRifle: This looks like a very nice contribution, many thanks for your effort!
Let's wait for feedback from our client experts! @sancar, @asimarslan

LoneRifle:
Messed up the commit history, starting over.

PR #6086, opened 2015-08-29
Minor nio cleanup

pveentjer:
* Renamed bytebuffer method variables to src/dst where appropriate. So that it is in line with jdk and other parts of the system
* Improved javadoc
* Changed a few method from position in the file.


jerrinot:
:+1:

Donnerbart:
:+1:

PR #6088, opened 2015-08-29
MapProxySupport.putAllInternal: use setInternal instead of putInternal

LoneRifle:
setInternal has a void return type, which fits the return type of
putAllInternal, and avoids retrieving the previous value mapped to the key that putInternal does

Partially implements #6083

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -862,7 +862,7 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
                     checkNotNull(entry.getKey(), NULL_KEY_IS_NOT_ALLOWED);
                     checkNotNull(entry.getValue(), NULL_VALUE_IS_NOT_ALLOWED);
 
-                    putInternal(mapService.getMapServiceContext().toData(entry.getKey(), partitionStrategy),
+                    setInternal(mapService.getMapServiceContext().toData(entry.getKey(), partitionStrategy),
                             mapService.getMapServiceContext().toData(entry.getValue()),
                             -1,
                             TimeUnit.MILLISECONDS);


Donnerbart:
I checked both methods and the operations and cannot see any magic that `putInternal()` is and `setInternal()` is not doing. So this change should be safe.

:+1:

jerrinot:
@LoneRifle: Thank you for your contribution, it's much appreciated!

ahmetmircik:
Seems this PR should not have been merged, it is breaking the backward compatibility from the perspective of fired events. Please see the warning [here] (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/core/IMap.java#L564). Since this merge, `old-value` is no more available in fired events.

jerrinot:
@ahmetmircik: That's a good point, thanks for spotting this!
I'll revert it and write a test which is obviously missing.

Donnerbart:
Could you also add this warning to the `setInternal()` method, please? It was not obvious for me, even though I had a look down to the operations of put and set :(

jerrinot:
@Donnerbart: Done

LoneRifle:
@ahmetmircik - thanks for pointing this out. I'm still interested in avoiding the unwanted retval in putAll however, so would it be okay that I replace the putInternal call with a PutAllOperation containing a single entry, ie like:
```java
for (Entry entry : entries.entrySet()) {
    checkNotNull(entry.getKey(), NULL_KEY_IS_NOT_ALLOWED);
    checkNotNull(entry.getValue(), NULL_VALUE_IS_NOT_ALLOWED);
    
    int partitionId = partitionService.getPartitionId(entry.getKey());
    Map.Entry<Data,Data> dataEntry = 
        new AbstractMap.SimpleImmutableEntry<Data, Data>(
            mapService.getMapServiceContext().toData(entry.getKey(), partitionStrategy),
            mapService.getMapServiceContext().toData(entry.getValue())
        );

    MapEntrySet mapEntrySet = new MapEntrySet(Collections.singleton(dataEntry));
    final PutAllOperation op = new PutAllOperation(name, mapEntrySet);
    op.setPartitionId(partitionId);
    operationService.invokeOnPartition(SERVICE_NAME, op, partitionId).get();
}
```

or similar.

ahmetmircik:
Hi @LoneRifle, there will be no difference, but maybe we can think of adding a `IMap#setAll` method instead.

LoneRifle:
I would disagree, given that PutAllOperation.getResponse returns a boolean, which is inline with SetOperation.getResponse, and so avoiding the old value returned by PutOperation.getResponse, but this conversation may be moot anyway given #6094 . 

Fwiw, if the only difference between setAll and putAll is the firing of event listeners, I would think users would be happy just having putAll.

ahmetmircik:
I see your point and https://github.com/hazelcast/hazelcast/pull/6094 is an obvious plus. But i wanted to point the case that if there is a `map-store`, `putAll` will still be calling the `load` method of `map-store` in order to return `old-value`. `setAll` will be handy in that case also.

PR #6089, opened 2015-08-29
Remove PacketReader/Writer

pveentjer:
The following classes have been removed:
* PacketReader
* PacketWriter 
* PacketSocketReader
* PacketSocketWriter

The abstraction for reading and writing a packet can't carry its weight. It can be done already through the SocketReader/Writer abstraction. 

This Enterprise PR needs to be checked once this one is merged:
https://github.com/hazelcast/hazelcast-enterprise/pull/345


pveentjer:
verify

jerrinot:
The changes is relatively small LoC-wise. However I'm not if I am able to to asses its full impact. 

What are these abstractions for? What are we losing by removing them? @mdogan might know the answers here.

pveentjer:
It is an abstraction for reading/writing the packet. But it can easily be realized by the SocketReader; no need to have 4 classes for it. It an architectural artifact that can't carry its weight.

jerrinot:
:+1:  
The changes look good to me, but I don't have a great knowledge about this. 
I'd recommend someone with deeper knowledge of this area to review it too.

mtopolnik:
I can only say the current design gives me dizzy spells, it seems to run around in circles doing nothing. So :+1:

pveentjer:
It also caused my dizzy spells, but the actual design is quite nice. The problem is that there is a bit too much indirection, and the naming of the classes is not self explanatory. I have another PR to clean this up. 

Actually I'm very happy with the abstraction for socket communication for the connection; because it has proven to be very flexible (e.g. now easy to plug in different threading models).

mtopolnik:
This is what I'm referring to: 

- `PacketSocketReader implements SocketReader`, wraps `PacketReader`
- `SocketReader` has method `read(ByteBuffer src)`
- `PacketReader` has method `read(ByteBuffer src)`

pveentjer:
Yes. These issues are resolved in this PR:
ReadHandler -> SocketReader
WriteHandler -> SocketWriter
SocketReader -> ReadHandler
SocketWriter -> WriteHandler

So a SocketReader reads from the socket and forwards to the ReadHandler to process the incoming data (if you know better names.. please provide them). But I found this a lot more intuitive. 

Anyhow.. lets continue this discussion on that PR :)

mdogan:
Looks good.

@jerrinot: As far as I remember, `PacketWriter/Reader` were internal abstractions of `SocketWriter/Reader` in Hazelcast 2. Their purpose was to support plain write/read and symmetric & asymmetric encryptions via delegation. 

See: 
https://github.com/hazelcast/hazelcast/blob/maintenance-2.x/hazelcast/src/main/java/com/hazelcast/nio/SocketPacketWriter.java#L60

https://github.com/hazelcast/hazelcast/blob/maintenance-2.x/hazelcast/src/main/java/com/hazelcast/nio/SocketPacketReader.java#L78

pveentjer:
Their abstraction makes sense, but it is a feature which can already be provided by the SocketReader/SocketWriter.

jerrinot:
@mdogan: Thanks for clarification!

mdogan:
@pveentjer: We did have `SocketReader/SocketWriter`. But there were two impls of them; one for `Packet` and one for text/ascii. Both were used for member-2-member and client-2-member communications. 
In 2.x, `SocketPacketReader/SocketPacketWriter` were handling different types via delegation. Now we have concrete impls for each type; plain, encrypted, client, member etc.

PR #6093, opened 2015-08-31
MapProxySupport.putAllInternal: use single-entry operation instead of putInternal

LoneRifle:
* Replace the call to putInternal with the creation of single-entry
PutAllOperation that gets invoked asynchronously.

* Move the LinkedList of Futures to outside the if-block, since we now
create futures even if !tooManyEntries. Move the for loop that blocks on
the futures outside too.

Implements part of #6083

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -827,8 +827,9 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
         int partitionCount = partitionService.getPartitionCount();
         boolean tooManyEntries = entries.size() > (partitionCount * factor);
         try {
+            List<Future<?>> futures = new LinkedList<Future<?>>();
+
             if (tooManyEntries) {
-                List<Future> futures = new LinkedList<Future>();
                 Map<Integer, MapEntrySet> entryMap
                         = new HashMap<Integer, MapEntrySet>(nodeEngine.getPartitionService().getPartitionCount());
                 for (Entry entry : entries.entrySet()) {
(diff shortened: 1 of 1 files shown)


pveentjer:
Hi LoneRifle, thanks for your contribution. I'll find someone reviewers for this PR; potentially myself.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -827,8 +827,9 @@ protected void putAllInternal(final Map<? extends Object, ? extends Object> entr
         int partitionCount = partitionService.getPartitionCount();
         boolean tooManyEntries = entries.size() > (partitionCount * factor);
         try {
+            List<Future<?>> futures = new LinkedList<Future<?>>();
```
pveentjer: I would make use of an ArrayList and initialize it with entries.size. This way you create a constant amount of a litter instead of litter for every item in the entries set.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -827,8 +827,9 @@ protected void putAllInternal(final Map<? extends Object, ? extends Object> entr
...
+
```
pveentjer: I'm not following the flow of the MapProxySupport.putAllInternal method. 

If there are many operations, the system will merge the operations per partition. But if there are not enough operations, the system will not do so. 

AFAIK it is good to do always because making remote calls is 'extremely' expensive compared to the amount of work to be done. Imho we should get rid of the complete (!tooManyEntries) branch and always execute initial branch.

So perhaps your code isn't needed at all.

pveentjer: I created a new PR dealing with the strange behavior of this method:

https://github.com/hazelcast/hazelcast/pull/6094

Could you have a look at it?

LoneRifle: Gladly. I share your opinions, though with this PR I was looking to preserve code flow. Given that you've now created the new PR that embodies your suggestions, I'll close this one.

PR #6094, opened 2015-08-31
Map.putAll always merges operation per partition

pveentjer:
There was some strange and undocumented logic that only above a certain
threshold the put operations were being merged per partition. This PR fixes
that by always doing so.

Even if there are 2 entries, but they both fall in the same partition, it is already very beneficial to merge these operations. There is really no reason for a threshold. 

In the future we can speed it up even more by making use of the PartitionIteratingOperation so that multiple partitions for the same member, can be merged into a single remote call. But this needs to be investigated.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -121,7 +121,6 @@ import java.util.EventListener;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -157,7 +156,7 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
         this.partitionService = getNodeEngine().getPartitionService();
 
         lockSupport = new LockProxySupport(new DefaultObjectNamespace(MapService.SERVICE_NAME, name),
-                    LockServiceImpl.getMaxLeaseTimeInMillis(nodeEngine.getGroupProperties()));
+                LockServiceImpl.getMaxLeaseTimeInMillis(nodeEngine.getGroupProperties()));
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -818,57 +817,57 @@ public boolean isEmpty() {
...
-                    }
-                    entryMap.get(partitionId).add(
-                            new AbstractMap.SimpleImmutableEntry<Data, Data>(mapService.getMapServiceContext().toData(
-                                    entry.getKey(),
-                                    partitionStrategy),
-                                    mapService.getMapServiceContext()
-                                            .toData(entry.getValue())
-                            ));
+            List<Future> futures = new ArrayList<Future>(partitionCount);
+            MapEntrySet[] entrySetPerPartition = new MapEntrySet[partitionCount];
```
LoneRifle: I foresee this being somewhat expensive where entries.size() << partitionCount (with hazelcast defaults, that's 271 elements), but as entries.size() approaches partitionCount, the HashMap used by the original code would be worse. Ideally we would be able to easily switch between the two depending on how many entries we're putting in, but that's beyond the scope of this PR.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -818,57 +817,57 @@ public boolean isEmpty() {
...
+                entrySet.add(new AbstractMap.SimpleImmutableEntry<Data, Data>(keyData, valueData));
+            }
 
-                for (Future future : futures) {
-                    future.get();
+            // then we invoke the operations
+            for (int partitionId = 0; partitionId < entrySetPerPartition.length; partitionId++) {
+                MapEntrySet entrySet = entrySetPerPartition[partitionId];
+                if (entrySet != null) {
+                    // If there is a single entry, we could make use of a PutOperation since that is a bit cheaper
```
LoneRifle: We probably want to avoid a PutOperation given that we want to avoid getting back the old value that maps to the key. See #6083.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
pveentjer: Why should this be expensive. You can a constant amount of litter independent of the number of items. With a hashmap approach you create multiple objects per partition so you get a lot more litter.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
pveentjer: Its an old comment. But yet.. set would be better.

LoneRifle: Yup, except @ahmetmircik points out set doesn't trigger the EventListeners for put, which we need for putAll. I suppose this can be revisited.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
LoneRifle: Fine, though it would be nice if we can reduce that constant amount.

LoneRifle:
:+1: for me. I'll make corresponding changes to ClientMapProxy.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
pveentjer: Yes. 

We should not work around a bug.. but we should fix the bug :)

jerrinot:
It makes sense to me. but I already burned my fingers with IMap recently. See: https://github.com/hazelcast/hazelcast/pull/6088#issuecomment-136025968

Hence I'd recommend @ahmetmircik to look at this too.

ahmetmircik:
@pveentjer It would be better if you can also provide a test. Other that that this PR seems  mergeable to me.
:+1:

pveentjer:
Hi Ahmet, 

there are already putAll tests for map. So I don't know what can be added.

ahmetmircik:
Sure that is why i did not say  "this requires a test".  I am fine in any case.

PR #6100, opened 2015-08-31
ClientMapProxy.putAll should go direct to key owners

LoneRifle:
Port code from MapProxySupport so that MapPutAllRequests are sent directly to key owners rather than mediated by a randomly chosen member of the cluster

Implements part of #6083

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -1244,26 +1244,23 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
             final Data keyData = toData(entry.getKey());
             invalidateNearCache(keyData);
             
-            int partitionId = partitionService.getPartitionId(entry.getKey());
-            if (!entryMap.containsKey(partitionId)) {
-                entryMap.put(partitionId, new HashMap<Data, Data>());
+            int partitionId = partitionService.getPartitionId(keyData);
+            Map<Data, Data> partition = entryMap.get(partitionId); 
+            if (partition == null) {
+                partition = new HashMap<Data, Data>();
+                entryMap.put(partitionId, partition);
             }
-            
-            entryMap.get(partitionId).put(keyData, toData(entry.getValue()));
+
+            partition.put(keyData, toData(entry.getValue()));
         }
         
         List<Future<?>> futures = new ArrayList<Future<?>>(entryMap.size());
         for (final Map.Entry<Integer, Map<Data, Data>> entry : entryMap.entrySet()) {
             final Integer partitionId = entry.getKey();
-            //If there is only one entry, consider how we can use MapPutRequest without
-            //having to get back the return value.
+            //If there is only one entry, consider how we can use MapPutRequest
+            //without having to get back the return value.
             ClientMessage request = MapPutAllCodec.encodeRequest(name, entry.getValue());
-            
-            //invoke by partition owner address rather than id, as
-            //the latter will embed the partitionId into the Packet payload,
-            //routing the invocation to the operation service, which does 
-            //not handle MapPutAllRequest and other similar requests
-            futures.add(new ClientInvocation(getClient(), request, partitionService.getPartitionOwner(partitionId)).invoke());
+            futures.add(new ClientInvocation(getClient(), request, partitionId).invoke());
         }
 
         try {
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -79,6 +79,7 @@ import com.hazelcast.client.impl.protocol.codec.MapValuesWithPredicateCodec;
 import com.hazelcast.client.nearcache.ClientHeapNearCache;
 import com.hazelcast.client.nearcache.ClientNearCache;
 import com.hazelcast.client.spi.ClientListenerService;
+import com.hazelcast.client.spi.ClientPartitionService;
 import com.hazelcast.client.spi.ClientProxy;
 import com.hazelcast.client.spi.EventHandler;
 import com.hazelcast.client.spi.impl.ClientInvocation;
@@ -1233,15 +1234,46 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
 
     @Override
     public void putAll(Map<? extends K, ? extends V> m) {
-        Map<Data, Data> map = new HashMap<Data, Data>();
+        ClientPartitionService partitionService = getContext().getPartitionService();
+        Map<Integer, Map<Data, Data>> entryMap = new HashMap<Integer, Map<Data, Data>>(partitionService.getPartitionCount());
+        
         for (Entry<? extends K, ? extends V> entry : m.entrySet()) {
+            checkNotNull(entry.getKey(), NULL_KEY_IS_NOT_ALLOWED);
+            checkNotNull(entry.getValue(), NULL_VALUE_IS_NOT_ALLOWED);
+            
             final Data keyData = toData(entry.getKey());
             invalidateNearCache(keyData);
-            map.put(keyData, toData(entry.getValue()));
+            
+            int partitionId = partitionService.getPartitionId(entry.getKey());
+            if (!entryMap.containsKey(partitionId)) {
+                entryMap.put(partitionId, new HashMap<Data, Data>());
+            }
+            
+            entryMap.get(partitionId).put(keyData, toData(entry.getValue()));
+        }
+        
+        List<Future<?>> futures = new ArrayList<Future<?>>(entryMap.size());
+        for (final Map.Entry<Integer, Map<Data, Data>> entry : entryMap.entrySet()) {
+            final Integer partitionId = entry.getKey();
+            //If there is only one entry, consider how we can use MapPutRequest without
+            //having to get back the return value.
+            ClientMessage request = MapPutAllCodec.encodeRequest(name, entry.getValue());
+            
+            //invoke by partition owner address rather than id, as
+            //the latter will embed the partitionId into the Packet payload,
+            //routing the invocation to the operation service, which does 
+            //not handle MapPutAllRequest and other similar requests
+            futures.add(new ClientInvocation(getClient(), request, partitionService.getPartitionOwner(partitionId)).invoke());
         }
 
-        ClientMessage request = MapPutAllCodec.encodeRequest(name, map);
-        invoke(request);
+        try {
+            for (Future<?> future : futures) {
+                future.get();
+            }
+        } catch (Exception e) {
+            ExceptionUtil.rethrow(e);
+        }
+        
     }
 
     @Override
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
@@ -226,6 +226,20 @@ public class ClientMapTest {
     }
 
     @Test
+    public void testPutAllWithTooManyEntries() {
+        final IMap map = createMap();
+        Map mm = new HashMap();
+        for (int i = 0; i < 1000; i++) {
+            mm.put(i, i);
+        }
+        map.putAll(mm);
+        assertEquals(map.size(), 1000);
+        for (int i = 0; i < 1000; i++) {
+            assertEquals(map.get(i), i);
+        }
+    }
+    
+    @Test
     public void testAsyncGet() throws Exception {
         final IMap map = createMap();
         fillMap(map);
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -1049,7 +1049,7 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
             final Data keyData = toData(entry.getKey());
             invalidateNearCache(keyData);
             
-            int partitionId = partitionService.getPartitionId(entry.getKey());
+            int partitionId = partitionService.getPartitionId(keyData);
             MapEntrySet entrySet = entrySetPerPartition[partitionId];
             if (entrySet == null) {
                 entrySet = new MapEntrySet();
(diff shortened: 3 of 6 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1036,14 +1037,50 @@ public boolean isEmpty() {
...
+        MapEntrySet[] entrySetPerPartition = new MapEntrySet[partitionCount];
+        
         for (Entry<? extends K, ? extends V> entry : m.entrySet()) {
+            checkNotNull(entry.getKey(), NULL_KEY_IS_NOT_ALLOWED);
+            checkNotNull(entry.getValue(), NULL_VALUE_IS_NOT_ALLOWED);
+            
             final Data keyData = toData(entry.getKey());
             invalidateNearCache(keyData);
+            
+            int partitionId = partitionService.getPartitionId(entry.getKey());
```
sancar: This will deserialize the key again, use following instead

```
int partitionId = partitionService.getPartitionId(keyData);
```

LoneRifle: Thanks for the catch, will amend when I get to my workstation.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1036,14 +1037,50 @@ public boolean isEmpty() {
...
+            if (entrySet != null) {
+                //If there is only one entry, consider how we can use MapPutRequest without
+                //having to get back the return value.
+                MapPutAllRequest request = new MapPutAllRequest(name, entrySet);
+                
+                //invoke by partition owner address rather than id, as
+                //the latter will embed the partitionId into the Packet payload,
+                //routing the invocation to the operation service, which does 
+                //not handle MapPutAllRequest and other similar requests
+                futures.add(new ClientInvocation(getClient(), request, partitionService.getPartitionOwner(partitionId)).invoke());
```
sancar: This does not look right. We already gather the keys that should go to one partition. We should not make use of a multi partition request here. 

The problem with this approach there will be no check on server side if this requests is running in correct node. Meaning of correct here is the node that owns of the partition. And partition table can change anytime.

We can modify MapPutAllRequest so that it will invoke an operation that will run on one node. It will be PartitionClientRequest instead of AllPartitionRequest. Also we need to the similar for MessageTasks(new client)

LoneRifle: So if we modify MapPutAllRequest and its new client counterpart as you proposed, should I be able to call the ClientInvocation by partitionId rather than address? I assume that invocation-by-id will ensure that the invocation will be processed by the correct node.

sancar: Yes, purpose if the proposed change is to be able to send via partition id to find the correct place to run safely.

LoneRifle: Done, along with re-use of serialized key in old client.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1233,15 +1234,41 @@ public boolean isEmpty() {
...
+        
         for (Entry<? extends K, ? extends V> entry : m.entrySet()) {
+            checkNotNull(entry.getKey(), NULL_KEY_IS_NOT_ALLOWED);
+            checkNotNull(entry.getValue(), NULL_VALUE_IS_NOT_ALLOWED);
+            
             final Data keyData = toData(entry.getKey());
             invalidateNearCache(keyData);
-            map.put(keyData, toData(entry.getValue()));
+            
+            int partitionId = partitionService.getPartitionId(entry.getKey());
```
sancar: same here, use keyData.

LoneRifle: Can't believe I missed that. Doing it now.

sancar:
:+1:  Looks good

sancar:
verify

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1233,15 +1234,41 @@ public boolean isEmpty() {
...
         for (Entry<? extends K, ? extends V> entry : m.entrySet()) {
+            checkNotNull(entry.getKey(), NULL_KEY_IS_NOT_ALLOWED);
+            checkNotNull(entry.getValue(), NULL_VALUE_IS_NOT_ALLOWED);
+            
             final Data keyData = toData(entry.getKey());
             invalidateNearCache(keyData);
-            map.put(keyData, toData(entry.getValue()));
+            
+            int partitionId = partitionService.getPartitionId(keyData);
+            if (!entryMap.containsKey(partitionId)) {
```
ihsandemir: No need for this extra containsKey call. Just use the below entryMap.get(partitionId) call and check if null.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1233,15 +1234,41 @@ public boolean isEmpty() {
 
     @Override
     public void putAll(Map<? extends K, ? extends V> m) {
-        Map<Data, Data> map = new HashMap<Data, Data>();
+        ClientPartitionService partitionService = getContext().getPartitionService();
+        Map<Integer, Map<Data, Data>> entryMap = new HashMap<Integer, Map<Data, Data>>(partitionService.getPartitionCount());
+        
         for (Entry<? extends K, ? extends V> entry : m.entrySet()) {
+            checkNotNull(entry.getKey(), NULL_KEY_IS_NOT_ALLOWED);
```
sancar: I dont think it make sense to iterate same thing twice just for that. It will slowdown the best case scenario.
And as a defence ConcurentHashMap.putAll in java does the same. Puts some and then throw null pointer exception if came across null.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1233,15 +1234,41 @@ public boolean isEmpty() {
...
+            //without having to get back the return value.
+            ClientMessage request = MapPutAllCodec.encodeRequest(name, entry.getValue());
+            futures.add(new ClientInvocation(getClient(), request, partitionId).invoke());
         }
 
-        ClientMessage request = MapPutAllCodec.encodeRequest(name, map);
-        invoke(request);
+        try {
+            for (Future<?> future : futures) {
+                future.get();
```
ihsandemir: Will it block indefinitely if one of the members does not respond? Why not  use "get(long timeout, TimeUnit unit)" here?

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1233,15 +1234,41 @@ public boolean isEmpty() {
...
-            map.put(keyData, toData(entry.getValue()));
+            
+            int partitionId = partitionService.getPartitionId(keyData);
+            if (!entryMap.containsKey(partitionId)) {
+                entryMap.put(partitionId, new HashMap<Data, Data>());
+            }
+            
+            entryMap.get(partitionId).put(keyData, toData(entry.getValue()));
+        }
+
```
ihsandemir: The code above this line is trying to classify all the entries matching to the same partition ID but do we really need this information? We are not doing any batch send to the member, we are still sending each request one by one. Hence, we do we need the entryMap at all? Just iterate through m, find the partition id for the key and send it right away.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1036,14 +1037,45 @@ public boolean isEmpty() {
 
     @Override
     public void putAll(Map<? extends K, ? extends V> m) {
```
ihsandemir: Comments for hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java putAll apply to this method as well.

hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
```diff
@@ -226,6 +226,20 @@ public void testGetAllPutAll() {
     }
 
     @Test
+    public void testPutAllWithTooManyEntries() {
+        final IMap map = createMap();
+        Map mm = new HashMap();
+        for (int i = 0; i < 1000; i++) {
```
ihsandemir: Is 1000 too many? If so, Why?

hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
```diff
@@ -226,6 +226,20 @@ public void testGetAllPutAll() {
     }
 
     @Test
+    public void testPutAllWithTooManyEntries() {
```
ihsandemir: What is the test case description here? Does this test really verify that the request is not randomly sent but to the specific partition Id? When sent to a random partition id, will the call "map.get(i)" fail?

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
sancar: All our other invocations are doing the same thing. Timeout for client is limited to heartbeat timeout for now.

hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
LoneRifle: This was originally intended to exercise the case where the number of entries exceeded partitionCount * 3, which if so, would execute the group-by-partition logic you see in this PR; otherwise, it would iterate through the entries sending each out as a PutOperation. 1000 was > 3 * 271, the default partition count in Hazelcast, which I wasn't able to immediately work out how to extract.

hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
LoneRifle: See earlier comments. The point of the test was largely to exercise the block that used to form one branch, and wasn't to check if the operations were being sent to the correct partitions.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
LoneRifle: Point taken, though I note this is more inline with the implementation found in MapProxySupport.putAllInternal. I also note that the requests are being batched at the partition level since as @sancar notes we do not know for sure which member would own the partition at the time we send, which renders member-level sending moot.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
LoneRifle: That's the intent - putAll does not specify a timeout, unlike the overrides for put.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
LoneRifle: Done.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
sancar: We are doing batch send for each partition. It is different and efficient from sending one by one. 
Batching all partitions that is owned by same node can be one step further improvement but it will be more complicated than this one. 

@pveentjer If I don`t understand wrong, we could have utilize the following for map.putAll on MapProxyImpl, which will batch the partitions going to same member.
```
Map<Integer, Object> invokeOnPartitions(String serviceName, OperationFactory operationFactory,
                                            Collection<Integer> partitions) throws Exception;
```

@LoneRifle 
But for the client this means you are gonna need to implement a similar thing yourself. And if you intend to do that, keep in mind that same proxy is for both smart and non-smart(connected to only one node) client.

LoneRifle: @sancar - noted, if we must change this. I believe similar comments were made in #6094 about using PartitionIteratingOperation, but to be done at a later date.

hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
sancar: It does not verify exactly but this change can not be tested outside. It is only a performance improvement. I think we are OK even if there is no extra test introduced.

LoneRifle:
By-the-by, is there a public real-time chat channel that external contributors like me can use to discuss items with the hazelcast devs?

jerrinot:
@LoneRifle: Some people are hanging out at https://gitter.im/hazelcast/hazelcast

sancar:
verify

sancar:
Unrelated test failed:
@eminn 
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast/15252/testReport/junit/com.hazelcast.quorum.cache/CacheWriteQuorumTest/com_hazelcast_quorum_cache_CacheWriteQuorumTest/
```
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at com.hazelcast.quorum.PartitionedCluster$3.run(PartitionedCluster.java:118)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:751)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:765)
	at com.hazelcast.quorum.PartitionedCluster.splitFiveMembersThreeAndTwo(PartitionedCluster.java:113)
	at com.hazelcast.quorum.PartitionedCluster.partitionFiveMembersThreeAndTwo(PartitionedCluster.java:62)
	at com.hazelcast.quorum.cache.CacheWriteQuorumTest.initialize(CacheWriteQuorumTest.java:85)
```

sancar:
verify

hazelcast-client-new/src/test/java/com/hazelcast/client/map/ClientMapTest.java
ihsandemir: :+1:

PR #6101, opened 2015-08-31
Added new CheckStyle rule to prevent UUID.randomUUID() calls

Donnerbart:
Added a new CheckStyle rule to prevent `UUID.randomUUID()` calls (static import and direct method call). We should use `UuidUtil.buildRandomUUID()` instead. Fixed all classes which used `UUID` so far.

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -281,6 +281,13 @@
         <module name="ArrayTypeStyle"/>
         <module name="UpperEll"/>
 
+        <!-- Check for usage of UuidUtil instead of UUID -->
+        <module name="RegexpSinglelineJava">
+            <property name="format" value="(import static java\.util\.UUID\.randomUUID|UUID\.randomUUID\(\))"/>
+            <property name="ignoreComments" value="true"/>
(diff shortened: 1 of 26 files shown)


Donnerbart:
verify

Donnerbart:
After a short discussion about `Random` and `SecureRandom` to create a `UUID` I changed this PR to add both versions into our `UuidUtil` class.

I renamed the method `buildRandomUuidString()` to `newUnsecureUUID()` (which is based on `Random`). I also added `newSecureUUID()` (which is based on `SecureRandom`). Finally I improved the JavaDoc of the whole class.

**I replaced all existing calls to `UUID.randomUUID()`, so we have to check if we need the secure or unsecure method call here.**

I used the secure method for `ClusteredSessionService.jvmId` and `Webfilter.generateSessionId()`.

All other methods were related to add listeners. Since I found already listener methods which used the old, unsecure `UuidUtil` method, I used it as well to replace the following `UUID.randomUUID()` calls:
* `ClientReliableTopicProxy.addMessageListener()`
* `ClientClusterServiceImpl.addMembershipListenerWithoutInit()`
* `EventServiceImpl.registerListenerInternal()`
* `ReliableTopicProxy.addMessageListener()`

sancar:
:+1:

jerrinot:
:+1:

PR #6103, opened 2015-09-01
Added target-aware transaction support

mdogan:
Before this change, a transaction record can be only mapped to a partition. Now it's possible to map a transaction record to a specific target member directly.

_This feature is required to provide transactional cluster state change for hot-restart module._

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/transaction/impl/TargetAwareTransactionLogRecord.java
+++ b/hazelcast/src/main/java/com/hazelcast/transaction/impl/TargetAwareTransactionLogRecord.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


pveentjer:
:+1:

jerrinot:
Just 2 minor notes: 
- Is `TargetAwareTransactionLogRecord` a good name? Isn't in better to name it `AddressAware`? Target is more abstract concept to me, a partition could be a target too. 

- The test names aren't very descriptive. `test_prepare()` inside `TransactionLogTest` class does not indicate the test is specifici for `TargetAware` records. 

None of these notes is serious enough to block this. Consider this as :+1: & feel free to address or not address my notes.

mdogan:
@jerrinot:

> The test names aren't very descriptive. test_prepare() inside TransactionLogTest class does not indicate the test is specifici for TargetAware records.

It is NOT specific for target-aware records. They are testing  `TransactionLog.prepere()/commit()/rollback()` methods for both normal/partition-specific and target-aware transaction records. 
Do you prefer to have separate tests for each two type?

jerrinot:
yes. I'd prefer to have separated test methods. as it's a different scenario.

mdogan:
done.

jerrinot:
:+1:

jerrinot:
Thanks!

PR #6105, opened 2015-09-01
Added a specific exception to denote response to an invocation cannot sent

mdogan:
Otherwise, failure of sending a response causes another failure.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/exception/ResponseNotSentException.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/exception/ResponseNotSentException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


pveentjer:
Can you add to the javadoc a very concrete example of this behavior?

Can you also add some tests that verify that this behavior gets triggered and interpreted correctly.

mdogan:
@pveentjer:
This is not easy to test/verify because many things needed to be mocked/spied and when problem occurs, only difference you see is two failure logs instead of one. So, there's no caller visible side effect.

ahmetmircik:
@mdogan How should we deal with this defect on maintenance branch? WDYT?

mdogan:
@ahmetmircik: this is not a severe issue. I just faced this working on something completely different. As I said, it's not caller visible issue.

ahmetmircik:
@mdogan i see thanks

pveentjer:
Can you add some documentation; especially a very specific example? So under which circumstances can this happen?

hazelcast/src/main/java/com/hazelcast/spi/exception/ResponseNotSentException.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.hazelcast.core.HazelcastException;
+
+/**
+ * A HazelcastException indicating that there is some kind of system error causing a response NOT to be sent
+ * for some operation.
+ */
+public class ResponseNotSentException extends HazelcastException {
+
+    public ResponseNotSentException() {
```
pveentjer: We don't need this constructor. It will only invite bad behavior :)

pveentjer:
@mdogan what should be done with this PR?

mdogan:
Rebased and updated.

pveentjer:
Looks good. Can you add some tests?

mdogan:
See above :arrow_up: 

>This is not easy to test/verify because many things needed to be mocked/spied and when problem occurs, only difference you see is two failure logs instead of one. So, there's no caller visible side effect.

pveentjer:
Ok.

PR #6106, opened 2015-09-01
Fix for #6080

### LINKED ISSUE #6080: Inconsistent and potentially buggy design in BasicCompletableFuture

tombujok:
Follow up on #6020 #6062 and #6081; Fixes #6080

Key points:
* All the Futures extending AbstractCompletableFuture have been refactored in order to take advantage of the features added to AbstractCompletableFuture. Thus, all the latches and synchronisation mechanisms could have been removed making the code of a couple of Future's much simpler. 
* BasicCompletableFuture has been fixed, documented, unit-tested and made more "digestible".
* After all classes that extend AbstractCompletableFuture had been refactored the class itself was also modified to make some methods final / protected and basically to keeps its state "more" secure. There have also been some minor fixes in the get() or getResult() methods

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapReduceProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapReduceProxy.java
@@ -29,6 +29,7 @@ import com.hazelcast.client.spi.impl.ClientInvocation;
 import com.hazelcast.client.spi.impl.ClientInvocationFuture;
 import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.core.ICompletableFuture;
+import com.hazelcast.internal.serialization.SerializationService;
 import com.hazelcast.logging.Logger;
 import com.hazelcast.mapreduce.Collator;
 import com.hazelcast.mapreduce.CombinerFactory;
@@ -51,7 +52,6 @@ import com.hazelcast.mapreduce.impl.SetKeyValueSource;
 import com.hazelcast.mapreduce.impl.task.TransferableJobProcessInformation;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.Data;
-import com.hazelcast.internal.serialization.SerializationService;
 import com.hazelcast.spi.impl.AbstractCompletableFuture;
 import com.hazelcast.util.EmptyStatement;
 import com.hazelcast.util.UuidUtil;
@@ -65,12 +65,7 @@ import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-
-import static com.hazelcast.util.Preconditions.isNotNull;
 
 public class ClientMapReduceProxy
         extends ClientProxy
@@ -241,14 +236,10 @@ public class ClientMapReduceProxy
             implements JobCompletableFuture<V> {
 
         private final String jobId;
-        private final CountDownLatch latch;
-
-        private volatile boolean cancelled;
 
         protected ClientCompletableFuture(String jobId) {
             super(getContext().getExecutionService().getAsyncExecutor(), Logger.getLogger(ClientCompletableFuture.class));
             this.jobId = jobId;
-            this.latch = new CountDownLatch(1);
         }
 
         @Override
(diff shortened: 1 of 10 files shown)


hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapReduceProxy.java
```diff
@@ -269,25 +261,10 @@ public boolean cancel(boolean mayInterruptIfRunning) {
...
         @Override
-        public boolean isCancelled() {
-            return cancelled;
-        }
-
-        @Override
-        public void setResult(Object result) {
+        protected void setResult(Object result) {
             super.setResult(result);
-            latch.countDown();
```
mtopolnik: The entire `setResult` shoudl be deleted now.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
```diff
@@ -130,6 +131,13 @@ public boolean cancel(boolean mayInterruptIfRunning) {
                 return false;
             }
 
+            if (shouldCancel == null) {
+                shouldCancel = shouldCancel(mayInterruptIfRunning);
+                if (!shouldCancel) {
+                    return false;
+                }
+            }
+
```
mtopolnik: For me this would be clearer if the local variable was `alreadyAskedShouldCancel` or similar because you only check the actual `shouldCancel` result once. Currently the semantics of the tri-state variable are confusing.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
```diff
@@ -161,17 +181,23 @@ public boolean isCancelled() {
...
-            return get(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
-        } catch (TimeoutException e) {
-            logger.severe("Unexpected timeout while processing " + this, e);
-            return null;
+    public final V get() throws InterruptedException, ExecutionException {
+        for (; ; ) {
+            try {
+                return get(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+            } catch (TimeoutException e) {
+            }
```
mtopolnik: To forestall future raised eyebrows and suspicion, maybe add a comment here, such as: 
```
 // A timeout here can only be a spurious artifact.
 // It should never happen and even if it does, we must retry.
```

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapReduceProxy.java
tombujok: Unfortunately it's used in line 120 and 133, so I can't get rid of it completely

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
tombujok: Done

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractCompletableFuture.java
tombujok: Argh, this wasn't actually meant to be like this. Thx for spotting.  I fixed it - it's called once only but the result should be saved for next iterations should the first one fail to set the result.
```
if (shouldCancel == null) {
                shouldCancel = shouldCancel(mayInterruptIfRunning);
            }
            if (!shouldCancel) {
                return false;
            }
```

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapReduceProxy.java
mtopolnik: I don't get it... if you delete `@Override protected setResult(x) { super.setResult(x); }`, you should be able to continue using it just the same, unless there's another class in the same package that wants to call `setResult` from the outside.

tombujok: It's kind of an SCJP-style problem :) 
These are two independent inner classes (ClientJob & ClientCompletableFuture) that are based in one outer class (ClientMapReduceProxy).
ClientJob can't access a protected method from ClientCompletableFuture's super class (AbstractCompletableFuture) since it's neither in AbstractCompletableFuture's package nor in it's inheritance tree.
However, ClientJob could access ClientCompletableFuture's protected method since they are both located in the same outer class, thus we need this setResult() method.

tombujok:
verify

tombujok:
verify

metanet:
:+1:

PR #6107, opened 2015-09-01
Store IdentifiedDataSerializable factories in a specialized HashMap

jerrinot:
It avoids integer boxing and pointer indirection.


mdogan:
:+1:

pveentjer:
I don't think this is going to have a big impact. The values for serialization are within the -128..127 range, meaning that there is not a lot of overhead due to autoboxing.

If you really want to have something fast, I would also not go for a hashmap. Just go for an array.

jerrinot:
I agree it won't have a *big* impact. It removes a layer of indirection. Difference between this and array is this method: https://github.com/jerrinot/hazelcast/blob/e61f7e40991728a3ddbb6590df0f7565ea7fa64d/hazelcast/src/main/java/com/hazelcast/util/collection/Int2ObjectHashMap.java#L410-410

(beside possible conflicts)

pveentjer:
:+1:

PR #6109, opened 2015-09-01
SelectionKey exception handling

pveentjer:
In the old approach the exceptions around the selection key were caught and the connection was closed; but the flow was not aborted (so the methods keep running).
    
That has been resolved in this pr; so a selection key exception is propagated and caught just as any other  exception thrown from a handler (handler allows for checked exceptions to be thrown)


jerrinot:
:+1: 
there is still quite some room for further improvements as discussed. eg. the `SocketWritable` interface...

jerrinot:
this needs rebase

Issue #2630, opened 2014-06-05
AWS plugin does not support IAM roles


aberaud:
Currently the Hazelcast AWS plugin requires EC2 credentials hard-coded in the xml configuration file (or provided directly in Java), while Amazon provides the option to access APIs from running instances using IAM roles. This is suboptimal in term of security (by encouraging storing API keys in repositories and AMIs) and ease of use.

If AWS is enabled and creds are not provided in the configuration, the default behavior should be to check for IAM role membership, retrieve a token, and use it to access the API.

ddossot:
If anyone is interested, here is an alternative implementation of `TcpIpJoinerOverAWS` that uses the AWS EC2 client, which we have created at Unbounce in order to use IAM roles: https://gist.github.com/ddossot/814fadc6e2cf577f87f7

noctarius:
If somebody is interested in a general SPI for all kinds of discovery mechanisms, please read https://hazelcast.atlassian.net/wiki/display/COM/HEP+2+-+Hazelcast+Discovery+SPI and join the discussion at https://gitter.im/hazelcast-incubator/hazelcast-discovery-spi

unixunion:
+1 for fixing IAM roles.

w32-blaster:
+1 for IAM roles

danfink:
+1

jmiddleton:
+1 for IAM roles. I just got this issue so I used memcached instead.

ddossot:
@jmiddleton You could have used the code I posted above :wink: We use it in production.

jmiddleton:
Yes, that's true. I just saw it but I will try it soon. Is there any plan to incorporate that code to hazelcast?

ddossot:
@jmiddleton I don't think so, I think the official feature will come as part of a bigger revamp of the discovery mechanism (see: https://hazelcast.atlassian.net/wiki/display/COM/HEP+2+-+Hazelcast+Discovery+SPI )

PR #6110, opened 2015-09-01
added IAM Role support to hz-cloud

bilalyasar:
closes https://github.com/hazelcast/hazelcast/issues/2630

### DIFF
--- a/hazelcast-cloud/src/main/java/com/hazelcast/aws/impl/DescribeInstances.java
+++ b/hazelcast-cloud/src/main/java/com/hazelcast/aws/impl/DescribeInstances.java
@@ -19,8 +19,12 @@ package com.hazelcast.aws.impl;
 import com.hazelcast.aws.security.EC2RequestSigner;
 import com.hazelcast.aws.utility.CloudyUtility;
 import com.hazelcast.config.AwsConfig;
+import com.hazelcast.config.InvalidConfigurationException;
 
+import java.io.BufferedReader;
+import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.text.SimpleDateFormat;
@@ -28,11 +32,14 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static com.hazelcast.aws.impl.Constants.DOC_VERSION;
 import static com.hazelcast.aws.impl.Constants.SIGNATURE_METHOD_V4;
 
 public class DescribeInstances {
+    private static final String IAM_ROLE_ENDPOINT = "169.254.169.254";
     String timeStamp = getFormattedTimestamp();
     private EC2RequestSigner rs;
     private AwsConfig awsConfig;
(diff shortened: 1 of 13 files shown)


hazelcast-cloud/src/main/java/com/hazelcast/aws/impl/DescribeInstances.java
```diff
@@ -59,6 +66,29 @@ public DescribeInstances(AwsConfig awsConfig, String endpoint) {
...
+                    awsConfig.setAccessKey(matcher.group(1));
+                }
+                if (line.contains("SecretAccessKey") && matcher.find()) {
+                    awsConfig.setSecretKey(matcher.group(1));
+                }
+                if (line.contains("Token") && matcher.find()) {
+                    attributes.put("X-Amz-Security-Token", matcher.group(1));
+                }
+            }
+        } catch (IOException io) {
```
mesutcelik: empty catch.. we should fail-fast here.

hazelcast-cloud/src/main/java/com/hazelcast/aws/impl/DescribeInstances.java
```diff
@@ -59,6 +66,29 @@ public DescribeInstances(AwsConfig awsConfig, String endpoint) {
         attributes.put("X-Amz-Expires", "30");
     }
 
+    private void getKeysFromIamRole() {
+        try {
+            String query = "latest/meta-data/iam/security-credentials/" + awsConfig.getIamRole();
+            URL url = new URL("http", "169.254.169.254", -1, query);
```
mesutcelik: you can use following constructor here + pull out IP address into a static final.

```    
public URL(String protocol, String host, String file)
            throws MalformedURLException {
        this(protocol, host, -1, file);
    }
```

mesutcelik:
can you please test with clients both inside and outside the cloud? Also you should add fail fast in case an outside client which tries to connect to a cluster with a role should be disallowed...

hazelcast-cloud/src/main/java/com/hazelcast/aws/impl/DescribeInstances.java
```diff
@@ -59,6 +66,29 @@ public DescribeInstances(AwsConfig awsConfig, String endpoint) {
...
     }
 
+    private void getKeysFromIamRole() {
+        try {
+            String query = "latest/meta-data/iam/security-credentials/" + awsConfig.getIamRole();
+            URL url = new URL("http", "169.254.169.254", -1, query);
+            InputStreamReader is = new InputStreamReader(url.openStream(), "UTF-8");
+            BufferedReader reader = new BufferedReader(is);
+            Pattern pattern = Pattern.compile(": \"(.*?)\",");
+            for (String line; (line = reader.readLine()) != null; ) {
```
mesutcelik: I am sure you can create a static util method which parses response string and returns a map which contains required parameters. That makes testing this part easier...

hazelcast-cloud/src/main/java/com/hazelcast/aws/impl/DescribeInstances.java
bilalyasar: thanks, updated

hazelcast-cloud/src/main/java/com/hazelcast/aws/impl/DescribeInstances.java
bilalyasar: i have separated them.

bilalyasar:
added new tests, also rebased.

PR #6116, opened 2015-09-03
Introduced OutboundFrame

pveentjer:
Frame is the merging of the SocketReadable/SocketWritable.

A Frame has only a single method; isUrgent. The read/write methods
can all be handled in the Frame-specific ReadHandler/WriterHandler
classes. E.g. the MemberRead/Write-Handlers know how to read or
write a packet. So these methods don't belong in the interface since
they will never be invoked in a frame.read; but always in a packet.read
or a message.read.

Also the name 'Socket' is removed since frames can be exchanged
in other ways than just sockets. E.g. one could create a shared
memory IPC version.


hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientWriteHandler.java
```diff
@@ -88,8 +89,8 @@ private void writeBuffer() throws IOException {
         }
     }
 
-    public void enqueueSocketWritable(SocketWritable socketWritable) {
-        writeQueue.offer(socketWritable);
+    public void enqueue(Frame frame) {
```
tombujok: Shouldn't it be `public void enqueue(Packet packet)`? We do an explicit cast anyway and the method is not a part of an implemented interface. WDYT?

hazelcast/src/main/java/com/hazelcast/nio/tcp/MemberWriteHandler.java
```diff
@@ -30,7 +30,7 @@
 public class MemberWriteHandler implements WriteHandler<Packet> {
 
     @Override
-    public boolean onWrite(Packet packet, ByteBuffer dst) {
-        return packet.writeTo(dst);
+    public boolean onWrite(Packet frame, ByteBuffer dst) {
+        return frame.writeTo(dst);
     }
```
tombujok: Analogously here: `Packet packet`?

hazelcast/src/main/java/com/hazelcast/nio/tcp/OldClientWriteHandler.java
```diff
@@ -30,7 +30,7 @@
 public class OldClientWriteHandler implements WriteHandler<Packet> {
 
     @Override
-    public boolean onWrite(Packet packet, ByteBuffer dst) throws Exception {
-        return packet.writeTo(dst);
+    public boolean onWrite(Packet frame, ByteBuffer dst) throws Exception {
+        return frame.writeTo(dst);
     }
```
tombujok: Analogously here: `Packet packet`?

tombujok:
great job!
:+1: (with minor comments)

mdogan:
@tombujok: this PR is already closed. you should review: https://github.com/hazelcast/hazelcast/pull/6117

PR #6119, opened 2015-09-03
OSGI Improvements

serkan-ozal:
This PR contains:
- Hazelcast bundles provide OSGI services so users can manage (create, access, shutdown) Hazelcast instances through this service on OSGI environment.
- Created Hazelcast instances can be served as OSGI service to be accessed by other bundles.
- Each Hazelcast bundle provide a different OSGI service and their instances can be grouped (clustered) together for preventing possible compatibility issues between different Hazelcast versions/bundles.
- Hazelcast OSGI service's lifecycle (and so also the owned/created instances's lifecycles) are same with the owner Hazelcast bundles. When the bundle is stopped (deactivated), owned service and Hazelcast instances are also deactivated/shutdown and deregistered automatically. Then when the bundle is activated again, its service is registered again.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
+++ b/hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
@@ -14,10 +14,8 @@
  * limitations under the License.
  */
 
-package com.hazelcast.internal.osgi;
+package com.hazelcast.osgi.impl;
 
-import com.hazelcast.core.Hazelcast;
-import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 import org.osgi.framework.BundleActivator;
@@ -26,18 +24,20 @@ import org.osgi.framework.BundleContext;
 import java.lang.reflect.Method;
 
 /**
- * Hazelcast OSGi bundle activator
- * <p/>
- * initializes script Engines
+ * Hazelcast OSGi bundle activator.
+ *
+ * @see org.osgi.framework.BundleActivator
  */
 public class Activator
         implements BundleActivator {
 
-    private static final String HAZELCAST_OSGI_START = "hazelcast.osgi.start";
-
     private static final ILogger LOGGER = Logger.getLogger(Activator.class);
 
-    private HazelcastInstance hazelcastInstance;
+    // Defined as `volatile` since Javadoc of `BundleActivator` says that
+    // `The Framework must not concurrently call a BundleActivator object`
+    // and also it is marked as `@NotThreadSafe`.
+    // So it can be called from different threads but cannot be called concurrently.
+    private volatile HazelcastInternalOSGiService hazelcastOSGiService;
 
     @Override
     public void start(BundleContext context)
@@ -45,17 +45,22 @@ public class Activator
         // Try to start javax.scripting - JSR 223
         activateJavaxScripting(context);
 
-        if (Boolean.getBoolean(HAZELCAST_OSGI_START)) {
-            hazelcastInstance = Hazelcast.newHazelcastInstance();
-        }
+        assert hazelcastOSGiService == null : "Hazelcast OSGI service should be null while starting!";
+
+        // Create a new Hazelcast OSGI service with given bundle and its context
+        hazelcastOSGiService = new HazelcastOSGiServiceImpl(context.getBundle());
+
+        // Activate new created Hazelcast OSGI service
+        hazelcastOSGiService.activate();
     }
 
     @Override
     public void stop(BundleContext context)
             throws Exception {
-        if (hazelcastInstance != null) {
-            hazelcastInstance.shutdown();
-        }
+        assert hazelcastOSGiService != null : "Hazelcast OSGI service should not be null while stopping!";
+
+        hazelcastOSGiService.deactivate();
+        hazelcastOSGiService = null;
     }
 
     private void activateJavaxScripting(BundleContext context)
@@ -65,8 +70,9 @@ public class Activator
             return;
         }
 
-        Class<?> clazz = context.getBundle().loadClass("com.hazelcast.internal.osgi.ScriptEngineActivator");
+        Class<?> clazz = context.getBundle().loadClass("com.hazelcast.osgi.impl.ScriptEngineActivator");
         Method register = clazz.getDeclaredMethod("registerOsgiScriptEngineManager", BundleContext.class);
+        register.setAccessible(true);
         register.invoke(clazz, context);
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
@@ -1,0 +1,377 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.osgi.impl;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.config.GroupConfig;
+import com.hazelcast.config.XmlConfigBuilder;
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.instance.BuildInfoProvider;
+import com.hazelcast.osgi.HazelcastOSGiInstance;
+import com.hazelcast.osgi.HazelcastOSGiService;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.Logger;
+import com.hazelcast.util.ExceptionUtil;
+import com.hazelcast.util.StringUtil;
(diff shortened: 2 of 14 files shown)


noctarius:
Does that support on the fly registration/deregistration for serializers and model bundles?

serkan-ozal:
@noctarius Can you a little bit explain more?

noctarius:
Well one of the main aspects of OSGi is to split your application into smaller bundles that are installed separately. Those can be installed, removed, updated at runtime. Versions of the bundles may change (due to updates) and classes may disappear or appear again. Same goes for the Hazelcast bundle itself, it might be updated at runtime and the created HazelcastOSGIInstance should be updated with a HazelcastInstance delegate of the new version.

Not sure if all those things are not part of the requirements (since I'm not sure if this is based on my PRD ;-)).

serkan-ozal:
First of all these improvements are not based on your PRD, we have different PRD for a customer. 
We mentioned about our functionalities and design decisions and they are OK. You can see the related PRD and its design document from the confluence.

> Same goes for the Hazelcast bundle itself, it might be updated at runtime and the created HazelcastOSGIInstance should be updated with a HazelcastInstance delegate of the new version.

Every `HazelcastOSGiInstance` and `HazelcastOSGiService` instances are bind to a specific Hazelcast OSGI bundle such as 3.6-OSS, 3.6-EE, 3.6.1-OSS, 3.6.1-EE, .... When you deploy a new bundle, it will have its own `HazelcastOSGiService` and the other bundles will continue living on the container unless the bundle is stopped. When a bundle is stopped, its owned `HazelcastOSGiService` and `HazelcastOSGiInstance`'s are terminated. 

As a result, when you install a new bundle with newer version, there is no connection between itself and the instances of previous/other bundles. Every bundle has its own ecosystem.

serkan-ozal:
verify

hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
```diff
@@ -45,17 +45,28 @@ public void start(BundleContext context)
         // Try to start javax.scripting - JSR 223
         activateJavaxScripting(context);
 
-        if (Boolean.getBoolean(HAZELCAST_OSGI_START)) {
-            hazelcastInstance = Hazelcast.newHazelcastInstance();
+        // TODO Is it possible that OSGI service is not null (deactivated) on start?
```
mesutcelik: is this `TODO` leftover? You create github issues instead of using `TODO` marker

hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
```diff
@@ -45,17 +45,28 @@ public void start(BundleContext context)
...
+        if (hazelcastOSGiService != null) {
+            // If there is an active Hazelcast OSGI service leftover, deactivate it.
+            hazelcastOSGiService.deactivate();
+            hazelcastOSGiService = null;
         }
```
mesutcelik: you can just call `stop` method as far as i see instead of if block...

hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
```diff
@@ -45,17 +45,28 @@ public void start(BundleContext context)
...
     @Override
     public void stop(BundleContext context)
             throws Exception {
-        if (hazelcastInstance != null) {
-            hazelcastInstance.shutdown();
+        if (hazelcastOSGiService != null) {
+            hazelcastOSGiService.deactivate();
+            hazelcastOSGiService = null;
         }
+        // TODO Is it possible that OSGI service is null (not activated) on stop?
```
mesutcelik: another `TODO`

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
```diff
@@ -0,0 +1,387 @@
...
+                || StringUtil.isNullOrEmpty(groupConfig.getName())
+                || GroupConfig.DEFAULT_GROUP_NAME.equals(groupConfig.getName())) {
+            if (!Boolean.getBoolean(HAZELCAST_OSGI_GROUPING_DISABLED)) {
+                String groupName = id;
+                if (groupConfig == null) {
+                    config.setGroupConfig(new GroupConfig(groupName));
+                } else {
+                    groupConfig.setName(groupName);
+                }
+            }
```
mesutcelik: That block feels too complex already. :)

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
```diff
@@ -0,0 +1,387 @@
...
+                    } catch (Throwable t) {
+                        LOGGER.finest("Error occurred while deregistering " + this, t);
+                    }
+                    serviceRegistration = null;
+                    active = false;
+                    LOGGER.info(this + " has been deregistered as OSGI service and deactivated now");
+                } catch (Throwable t) {
+                    // Forcefully deactive the service itself.
+                    serviceRegistration = null;
+                    active = false;
```
mesutcelik: finally block could be alternative for the part
```
serviceRegistration = null;
active = false;
```

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
```diff
@@ -0,0 +1,387 @@
...
+            hazelcastOSGiInstance = new HazelcastOSGiInstanceImpl(instance, this);
+        }
+        ServiceRegistration serviceRegistration;
+        if (!Boolean.getBoolean(HAZELCAST_OSGI_REGISTER_DISABLED)) {
+            serviceRegistration =
+                    ownerBundleContext.registerService(HazelcastInstance.class.getName(), hazelcastOSGiInstance, null);
+        } else {
+            serviceRegistration = EMPTY_SERVICE_REGISTRATION;
+        }
+        instanceServiceRegistrationMap.put(hazelcastOSGiInstance, serviceRegistration);
```
kobalski: Do we need to add `try catch`  block here, maybe if service registration fails, we should not put entries to `instanceServiceRegistrationMap`,`instanceMap `.

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    private void shutdownAllInternal() {
+        try {
+            for (Map.Entry<HazelcastOSGiInstance, ServiceRegistration> entry : instanceServiceRegistrationMap.entrySet()) {
+                HazelcastOSGiInstance instance = entry.getKey();
+                ServiceRegistration serviceRegistration = entry.getValue();
+                if (serviceRegistration != EMPTY_SERVICE_REGISTRATION) {
+                    try {
+                        ownerBundleContext.ungetService(serviceRegistration.getReference());
```
cangencer: Could call deregisterInstance here first before shutting down (so it's consistent with shutdownInstance)

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
```diff
@@ -0,0 +1,387 @@
...
+    @Override
+    public boolean isActive() {
+        return active;
+    }
+
+    @Override
+    public void activate() {
+        // No need to complex lock-free approaches since this is not called frequently. Simple is good.
+        synchronized (serviceMutex) {
+            if (!active) {
```
cangencer: Why do we need an extra active flag here? AFAIK the container should keep track of if a bundle is active or not - we shouldn't need an additional flag.

hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
serkan-ozal: Even though this case seems not possible by OSGI container, instead of `TODO`, I will throw `IllegalStateException`

hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
serkan-ozal: I will remove this block as I mentioned in my previous comment and will throw `IllegalStateException`

hazelcast/src/main/java/com/hazelcast/osgi/impl/Activator.java
serkan-ozal: Even though this case seems not possible by OSGI container, instead of TODO, I will throw `IllegalStateException`

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
serkan-ozal: right, will make it more simple

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
serkan-ozal: If registration fails, the exception returns to caller directly without adding any put to both of `instanceServiceRegistrationMap`, `instanceMap`. So I think, no need to `try/catch` block.

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
serkan-ozal: make sense

kobalski:
:+1:

hazelcast/src/main/java/com/hazelcast/osgi/impl/HazelcastOSGiServiceImpl.java
kobalski: Maybe we can use `Bundle.getState` here. https://osgi.org/javadoc/r4v43/core/org/osgi/framework/Bundle.html#getState()

PR #6122, opened 2015-09-03
Little cleanup of Quorum Cache tests

Donnerbart:
Motivated by warnings in the class files.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
@@ -16,27 +16,22 @@
 
 package com.hazelcast.quorum.cache;
 
-import com.hazelcast.cache.HazelcastCachingProvider;
-import com.hazelcast.cache.ICache;
 import com.hazelcast.cache.impl.HazelcastServerCachingProvider;
-import com.hazelcast.config.CacheConfig;
 import com.hazelcast.config.CacheSimpleConfig;
 import com.hazelcast.config.Config;
-import com.hazelcast.config.MapConfig;
 import com.hazelcast.config.QuorumConfig;
 import com.hazelcast.config.QuorumListenerConfig;
 import com.hazelcast.core.HazelcastInstance;
-import com.hazelcast.core.IMap;
 import com.hazelcast.core.Member;
-import com.hazelcast.quorum.Quorum;
 import com.hazelcast.quorum.QuorumEvent;
+import com.hazelcast.quorum.QuorumException;
 import com.hazelcast.quorum.QuorumFunction;
 import com.hazelcast.quorum.QuorumListener;
-import com.hazelcast.quorum.QuorumService;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.TestHazelcastInstanceFactory;
 import com.hazelcast.test.annotation.QuickTest;
+import com.hazelcast.util.EmptyStatement;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
@@ -50,8 +45,7 @@ import static org.junit.Assert.fail;
 
 @RunWith(HazelcastSerialClassRunner.class)
 @Category(QuickTest.class)
-public class CacheQuorumListenerTest
-        extends HazelcastTestSupport {
+public class CacheQuorumListenerTest extends HazelcastTestSupport {
 
     @Test
     public void testQuorumFailureEventFiredWhenNodeCountBelowThreshold() throws Exception {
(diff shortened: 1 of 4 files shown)


Donnerbart:
run-parallel

hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
```diff
@@ -166,14 +160,15 @@ public void onChange(QuorumEvent quorumEvent) {
             threeNode.put(generateKeyOwnedBy(h1), "bar");
             fail();
         } catch (Exception e) {
+            EmptyStatement.ignore(e);
         }
         try {
             fourNode.put(generateKeyOwnedBy(h1), "bar");
             fail();
         } catch (Exception e) {
+            EmptyStatement.ignore(e);
```
pveentjer: rename the exception to expected and pleace remove the EmptyStatement.ignore. We do this for production code; but we should not do this for test code.

hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
```diff
@@ -240,12 +234,13 @@ public void onChange(QuorumEvent quorumEvent) {
...
         TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(3);
         HazelcastInstance instance1 = factory.newHazelcastInstance(config);
-        HazelcastInstance instance2 = factory.newHazelcastInstance(config);
+        factory.newHazelcastInstance(config);
         HazelcastServerCachingProvider cachingProvider = HazelcastServerCachingProvider.createCachingProvider(instance1);
         Cache<Object, Object> cache = cachingProvider.getCacheManager().getCache(cacheName);
         try {
             cache.put(generateKeyOwnedBy(instance1), 1);
         } catch (Exception e) {
+            e.printStackTrace();
```
pveentjer: Don't print stack traces. Either it is a failure and the test should be aborted; or it is something which should happen and then you also don't want to see the stacktrace.

pveentjer:
Minor review comments regarding exception handling

hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
Donnerbart: IDEA still creates a warning for the empty catch. This always grabs my attention (yellow alert). Renamed the exception though.

hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
Donnerbart: I wasn't sure about this one, but you're right. I'll remove the try-catch and we'll see if it crashes.

pveentjer:
The tests could be better; one of the problem is that I see with multipl try/catch clauses. This is an indication that the test is 

action
assert (exception)
action
assert (exception)

This is a smell and normally indicates that there are multiple tests hidden in a single test method. Once that is done.. you can put an @Test(expected=blaexception.class) around it and you get rid of the try catch completely.

Donnerbart:
I just improved the try-catch in `CacheQuorumListenerTest`. At least it's visible now that we expected a `QuorumException` all over the class. But this is not what the test is about, it is about the listener event to be fired (in case of a `QuorumException`). So it should be okay to catch the exceptions and wait for the event to be fired.

mustafaiman:
:+1: 
these tests and map's quorum tests are very similar. it would be good to apply the same refactoring to them to be consistent.

Donnerbart:
@pveentjer: Can you have another look? I think the tests should be okay now. The method calls which throw the exception are needed to trigger the listeners, which are the real test subject.

So it should be perfectly fine to call one or two methods before (and check that the needed exception is really thrown, since this is the happy path to the final assert).

hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
```diff
@@ -203,18 +204,21 @@ public boolean apply(Collection<Member> members) {
         });
         config.getCacheConfig(cacheName).setQuorumName(quorumName);
         config.addQuorumConfig(quorumConfig);
+
         HazelcastInstance instance = createHazelcastInstance(config);
         QuorumService quorumService = instance.getQuorumService();
-        Quorum quorum = quorumService.getQuorum(quorumName);
+        quorumService.getQuorum(quorumName);
```
mesutcelik: why are we calling this?

mesutcelik:
:+1: with a minor comment.

hazelcast/src/test/java/com/hazelcast/quorum/cache/CacheQuorumListenerTest.java
mustafaiman: i think that's a leftover. It has no use in the test.

Donnerbart: Good catch, thx!

PR #6127, opened 2015-09-04
Clean up of "Removed deprecated Hz.getLock(Object) method"

asimarslan:
minor clean-up left from a previous clean up


pveentjer:
:+1:

PR #6132, opened 2015-09-04
Skip transactionlog replication if singleton transaction

pveentjer:
If the transactionlog contains a single transactionlog record, there is no added value
to copying the transactionlog. The goal of copying the transactionlog is that the system
can't be left in an inconsistent state if the transaction initiator fails after the
prepare and has written some log records, but fails to write the rest.

If there is a single item in the transactionlog, it can't happen that some log records are written
and others are not since there is only a single item.

So this optimization check the number of items in the transactionlog after the prepare
and if there is one or less, it skips the replication of the transactionlog.


pveentjer:
Somehow my git or idea is doing very strange things.

PR #6133, opened 2015-09-04
Speeds up single item transactions

pveentjer:
Single item transaction don't need to be prepared. The txlog doesn't
need to be replicated since a single item transaction log can't be
partially written.

And it doesn't need to be prepared, since the commit will fail if
it tries to write something it doesn't own.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionContextImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionContextImpl.java
@@ -42,7 +42,6 @@ import javax.transaction.xa.XAResource;
 import java.util.HashMap;
 import java.util.Map;
 
-import static com.hazelcast.transaction.TransactionOptions.TransactionType.TWO_PHASE;
 import static com.hazelcast.transaction.impl.Transaction.State.ACTIVE;
 
 final class TransactionContextImpl implements TransactionContext {
@@ -70,7 +69,7 @@ final class TransactionContextImpl implements TransactionContext {
 
     @Override
     public void commitTransaction() throws TransactionException {
-        if (transaction.getTransactionType().equals(TWO_PHASE)) {
+        if (transaction.requiresPrepare()) {
             transaction.prepare();
         }
         transaction.commit();
@@ -89,14 +88,14 @@ final class TransactionContextImpl implements TransactionContext {
 
     @SuppressWarnings("unchecked")
     @Override
-    public <E> TransactionalQueue<E> getQueue(String name) {
-        return (TransactionalQueue<E>) getTransactionalObject(QueueService.SERVICE_NAME, name);
+    public <K, V> TransactionalMultiMap<K, V> getMultiMap(String name) {
+        return (TransactionalMultiMap<K, V>) getTransactionalObject(MultiMapService.SERVICE_NAME, name);
     }
 
     @SuppressWarnings("unchecked")
     @Override
-    public <K, V> TransactionalMultiMap<K, V> getMultiMap(String name) {
-        return (TransactionalMultiMap<K, V>) getTransactionalObject(MultiMapService.SERVICE_NAME, name);
+    public <E> TransactionalQueue<E> getQueue(String name) {
+        return (TransactionalQueue<E>) getTransactionalObject(QueueService.SERVICE_NAME, name);
     }
 
     @SuppressWarnings("unchecked")
@@ -116,6 +115,10 @@ final class TransactionContextImpl implements TransactionContext {
     public TransactionalObject getTransactionalObject(String serviceName, String name) {
         checkActive(serviceName, name);
 
+        if (requiresBackupLogs(serviceName)) {
+            transaction.ensureBackupLogsExist();
+        }
+
         TransactionalObjectKey key = new TransactionalObjectKey(serviceName, name);
         TransactionalObject obj = txnObjectMap.get(key);
         if (obj != null) {
@@ -127,6 +130,18 @@ final class TransactionContextImpl implements TransactionContext {
         obj = transactionalService.createTransactionalObject(name, transaction);
         txnObjectMap.put(key, obj);
         return obj;
+    }
+
+    private boolean requiresBackupLogs(String serviceName) {
+        if (serviceName.equals(MapService.SERVICE_NAME)) {
+            return false;
+        }
+
+        if (serviceName.equals(MultiMapService.SERVICE_NAME)) {
+            return false;
+        }
+
+        return true;
     }
 
     private TransactionalService getTransactionalService(String serviceName) {
--- a/hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionImpl.java
@@ -16,31 +16,29 @@
 
 package com.hazelcast.transaction.impl;
 
+import com.hazelcast.cluster.ClusterService;
 import com.hazelcast.core.MemberLeftException;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Address;
 import com.hazelcast.spi.NodeEngine;
+import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.OperationService;
 import com.hazelcast.spi.exception.TargetNotMemberException;
 import com.hazelcast.transaction.TransactionException;
 import com.hazelcast.transaction.TransactionNotActiveException;
 import com.hazelcast.transaction.TransactionOptions;
-import com.hazelcast.transaction.impl.operations.BeginTxBackupOperation;
-import com.hazelcast.transaction.impl.operations.PurgeTxBackupOperation;
-import com.hazelcast.transaction.impl.operations.ReplicateTxOperation;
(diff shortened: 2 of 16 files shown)


gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionContextImpl.java
```diff
@@ -70,7 +70,7 @@ public void beginTransaction() {
 
     @Override
     public void commitTransaction() throws TransactionException {
-        if (transaction.getTransactionType().equals(TWO_PHASE)) {
+        if (transaction.getTransactionType().equals(TWO_PHASE) || transaction.getTransactionLog().size() > 1) {
```
Donnerbart: Am I stupid or should this be an `&&` instead of `||`? Prepare the commit if it's `TWO_PHASE` AND has more than one entry.

pveentjer: Is it possible to give the answer 'yes' twice? :)

jerrinot:
is a potential exception propagated to user code?

gurbuzali:
`TransactionException`

jerrinot:
ok, as I know that in the past we had an issue where prepare was successful, but commit failed. in this case the exception was simple eaten. so I just wanted to make sure it's not a problem here.

pveentjer:
In both cases.. prepare and failed commit ---> TransactionException

jerrinot:
ok, good.
:+1: once green

pveentjer:
Some additional changes have been made; so it needs to be re-reviewed. 

One of the problems is that not only the txlog is copied; but when the tx begins, the tx-id is copied so that a remote txlog backup is created.The problem is that in case of a single item transaction, this is going to be an additional performance penalty.

This has been changed so that before a prepare the txlog is copied. I believe we should get rid of this particular copy completely. After the tx prepares, let the replication of the txlog take care of creating the remote tx log; there is no need to do it in 2 separate actions.

pveentjer:
These are the performance results of the MapTransactionTest on the test lab:
```
Local
	688,452.84 ops/s

Master: 2phase + 1 durability
	265,584.93 ops/s

Master: 2phase + 0 durability
	413,763.40 ops/s

Peter branch: 2phase + 1 durability
	 675,336.67 ops/s 

Peter branch: 2phase + 0 durability
	 676,016.40 ops/s
```
As can be seen, the performance of a single item 2phase transaction has the same performance as a local transaction and the durability doesn't have any impact.

hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionContextImpl.java
```diff
@@ -129,6 +132,18 @@ public TransactionalObject getTransactionalObject(String serviceName, String nam
         return obj;
     }
 
+    private boolean requiresBackupLogs(String serviceName) {
```
gurbuzali: I believe it would be safer to check if `serviceName` is equal to one of the Collection services.
It is fine currently but what if we introduce another `TransactionalService`, say `ReplicatedMap` for example.
Developer can easily miss to update this method.

gurbuzali:
a lot of good tests

hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionContextImpl.java
pveentjer: If the developer would miss it, then it would not lead to a bad thing; the backup logs are created. That is why I went for the 'on by default unless explicitly disabled'. 

But I can change it to 'off by default'

gurbuzali:
can we get rid of the WIP commit

hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionContextImpl.java
gurbuzali: correct, `on by default` is better/safer.

pveentjer:
Np.. I'm going to add more tests that verify the correct triggering of the backup logs. Once done I'll clean up the commits.

pveentjer:
There is also a change it the commit behavior. Instead of just logging the exception and not propagating and marking the commit as a success, the exception is now always thrown and the transaction marked as COMMIT_FAILED. This is needed for the single item transaction that skips the prepare phase.

hazelcast/src/main/java/com/hazelcast/transaction/impl/TransactionImpl.java
```diff
@@ -222,70 +205,76 @@ public void prepare() throws TransactionException {
...
                 state = COMMITTING;
-
                 List<Future> futures = transactionLog.commit(nodeEngine);
-
-                // We should rethrow exception if transaction is not TWO_PHASE
-                ExceptionHandler exceptionHandler = transactionType.equals(TWO_PHASE)
-                        ? commitExceptionHandler : FutureUtil.RETHROW_TRANSACTION_EXCEPTION;
-                waitWithDeadline(futures, COMMIT_TIMEOUT_MINUTES, MINUTES, exceptionHandler);
-
+                waitWithDeadline(futures, COMMIT_TIMEOUT_MINUTES, MINUTES, RETHROW_TRANSACTION_EXCEPTION);
```
gurbuzali: a question:
we were not throwing any exception while committing but we were logging.
with this we are changing that, how will it affect a user ?
```
try {
    do something
    commit
} catch(Exception) {
    rollback
}
```
What will happen for the above case ? user will get the exception and will try to rollback it, will it throw exception too ?

gurbuzali:
throwing exception while committing in a `TWO_PHASE` transaction seems like a behavioral change to me. We should be careful about this. For example, we call store methods of `MapStore` at commit phase, in this case the exception will propagate to the user which was not the case before.

pveentjer:
Hi Ali, it is a behavior chance and imho a good once. The old approach hides the problem and throwing the exception to the end user makes it clear there is a problem. It is one of our architectural principles 'fail fast'.

gurbuzali:
yes I know but we should be careful even while doing good changes, I just wanted the point that out.
And what about the above situation, what will happen when user calls `rollback` on a commit-failed transaction? How should we behave ? Do we have tests for that situation ?

pveentjer:
I added a test for this particular case:

```
 @Test
    public void rollback_whenRollingBackCommitFailedTransaction() {
        TransactionOptions options = new TransactionOptions().setTransactionType(TWO_PHASE).setDurability(0);
        TransactionImpl tx = new TransactionImpl(txManagerService, nodeEngine, options, "dummy-uuid");
        tx.begin();
        tx.add(new MockTransactionLogRecord().failCommit());
        try {
            tx.commit();
            fail();
        }catch (TransactionException expected){
        }

        tx.rollback();
        assertEquals(ROLLED_BACK, tx.getState());
    }
```

gurbuzali:
BitTest failed, I don't think it is relevant
```
10:39:01 Tests run: 25, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.004 sec - in com.hazelcast.nio.BitsTest
10:39:02 
10:39:02 Results :
10:39:02 
10:39:02 Failed tests: 
10:39:02   AbstractCompletableFutureTest.get_exceptionResultSet_exceptionThrown 
10:39:02 Expected: (an instance of java.lang.RuntimeException and exception with message a string containing "You screwed buddy!")
10:39:02      but: exception with message a string containing "You screwed buddy!" message was null
10:39:02 Stacktrace was: java.lang.NullPointerException
10:39:02 	at java.lang.management.ThreadInfo.<init>(ThreadInfo.java:168)
10:39:02 	at sun.management.ThreadImpl.dumpThreads0(Native Method)
10:39:02 	at sun.management.ThreadImpl.dumpAllThreads(ThreadImpl.java:446)
10:39:02 	at com.hazelcast.test.AbstractHazelcastClassRunner.generateThreadDump(AbstractHazelcastClassRunner.java:228)
10:39:02 	at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:139)
10:39:02 	at org.junit.rules.ExpectedException$ExpectedExceptionStatement.evaluate(ExpectedException.java:239)
10:39:02 	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
10:39:02 	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
10:39:02 	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
10:39:02 	at com.hazelcast.test.HazelcastParallelClassRunner.access$201(HazelcastParallelClassRunner.java:35)
10:39:02 	at com.hazelcast.test.HazelcastParallelClassRunner$TestRunner.run(HazelcastParallelClassRunner.java:119)
10:39:02 	at java.lang.Thread.run(Thread.java:745)
10:39:02 
```

gurbuzali:
verify

pveentjer:
'''You screwed buddy!'''

pveentjer:
verify-parallel

pveentjer:
verify

gurbuzali:
good to go

jerrinot:
:+1:

PR #6135, opened 2015-09-05
Rewrite ClientMapProxy.getAll to invoke-by-partition

LoneRifle:
* Convert MapGetAllRequest/MessageTask to ClientPartitionRequest to
allow invocation-by-partition

* Re-implement ClientMapProxy.getAll by porting code from putAll, so
that keys are grouped by corresponding partitions. For each key set,
send a request to the partition owner. This approach has two advantages:
  * Allows the client to gather data directly from owners rather than
depend on a randomly-chosen node
  * Because the requests are made asynchronously, the client can work on
caching and populating the return value with some of the data while
waiting for the other data to be fully received.

* Delete MapPutAllOperationFactory since it is no longer used

Implements the final part of #6083


sancar:
:+1: Looks good. 
@cangencer @ihsandemir Can you guys also review this one ?

cangencer:
:+1: looks good to me

ahmetmircik:
@LoneRifle, Is it possible to share some metrics also (i guess you already had some) which show that after merging this PR how much improvement, especially for different cluster and requested keySet sizes, will be there?

LoneRifle:
I don't have this to hand, but I'll work on it.

LoneRifle:
Gist https://gist.github.com/LoneRifle/ff9c6c658327ffe39e79 created.
I obtained old vs new times for 1, 2, 4 nodes for integer keys mapping to randomly-generated UUIDs for increasing powers of two up till 2^20.

Charts plotted are both in normal and log axes:
ClientMapProxy.getAll - old vs new

sancar:
@LoneRifle Scales of first two graphs are different. Can you share the graphs again and change the scale of the graph at the right top corner ? It would be a lot easier to compare if it would start from 20.000 rather than 100.000.

LoneRifle:
Apologies, I neglected to correct the scale before publishing the charts. Corrected above.

LoneRifle:
Apologies yet again, realised that the charts weren't actually displaying correctly. I've updated them and checked for correctness. Note that getAll timings are now independent of cluster size.

LoneRifle:
Hello @sancar, @ihsandemir - are you happy to have this merged? I'm planning to propose IMap.setAsync and would prefer to ensure that this is resolved before proceeding with further changes.

sancar:
Hi @LoneRifle 
The numbers for 1 node is confusing to me. While this operation can be done with one remote call to server, when we do number of partition times we get speed-up. I was expecting worst performance when cluster size is small. We may want to do these tests in an environment more close to real time scenario

These changes also require us to change c++ and c# client code, since team a little bit busy we might postpone the merge of this pr a little bit.

Other than that code looks good and if we verify the performance numbers it will probably in 3.6 release.
Note: My comments also apply to putAll.

LoneRifle:
@sancar, noted. Fwiw, the tests were done with the entire cluster on one machine, so there may have been overhead experienced by the target node when using the old code that may not be so significant if the cluster nodes are hosted by > 1 machine.

pveentjer:
@sancar can you have another look?

pveentjer:
@sancar can you have another look?

pveentjer:
@LoneRifle can you rebase?

LoneRifle:
I'm currently out of the country but will do so at earliest convenience

LoneRifle:
I have decided to submit a fresh PR after rebasing, since the changes will warrant a fresh review. Please see #6555.

PR #6139, opened 2015-09-07
`EvictionPolicy` of `EvictionConfig` can be `NONE` but JCache doesn't allow `NONE` for itself. 

### LINKED ISSUE #5251: JCache declarative eviction config is counter intuitive when EvictionPolicy is set to NONE
If i set it to `NONE` via `<eviction eviction-policy="NONE"/>`, it gives
`java.lang.IllegalArgumentException: Eviction policy cannot be "NONE" `. 

It seems counter intuitive, expecting no exception when i set it to `NONE`.

serkan-ozal:
Without this PR, `EvictionPolicy` of `EvictionConfig` cannot be `NONE` due to check in `EvictionConfig`. But in fact, non-allowed `NONE` value is specific to JCache and this check should not be handled in common place like `EvictionConfig` since this can be used by other data structures like `IMap`.

Fixes #5251


ahmetmircik:
:+1:  (once green)

serkan-ozal:
The following task should be moving eviction package to a common place to be used by other parts rathet than JCache since it is not JCache specific already.

gurbuzali:
:+1:

PR #6140, opened 2015-09-07
Moved eviction package from `com.hazelcast.cache.impl.eviction` to `com.hazelcast.eviction`

serkan-ozal:
Since eviction package is not dependent to JCache and it is designed for common usage, no need to locate it under cache package. So it is moved to under a common package `com.hazelcast.eviction`.


serkan-ozal:
Closing this one and will move to `com.hazelcast.internal.eviction` instead of `com.hazelcast.eviction`

PR #6143, opened 2015-09-07
Moved eviction package from `com.hazelcast.cache.impl.eviction` to `com.hazelcast.internal.eviction` 

serkan-ozal:
Since eviction package is not dependent to JCache and it is designed for common usage, no need to locate it under cache package. So it is moved to under a common package `com.hazelcast.internal.eviction`.


ahmetmircik:
Makes sense to me and :+1:  once green

pveentjer:
Are any classes being used by end users? Or is it a 100 private party?

serkan-ozal:
Yes, none of the are exposed to end user but some of them will be exposed with `Custom Eviction Policy` feature later. So I have moved them to our internal package after I had talked with @ahmetmircik

PR #6148, opened 2015-09-07
MapProxySupport sanitation

pveentjer:
Mostly removing excess of get calls and setting them in fields of the MapProxySupport


ahmetmircik:
Good cleanup :+1:

PR #6155, opened 2015-09-08
[backport] added aws iam support

bilalyasar:


mesutcelik:
closes https://github.com/hazelcast/hazelcast/issues/2630

mesutcelik:
@bilalyasar ,
can you please update master & maintenance documentation?

bilalyasar:
@mesutcelik , yes sure.

anirudh-wa:
Can you please provide the documentation link for this ?

danbeaulieu:
@bilalyasar I know this changeset has already been merged but I believe it misses the mark a bit. IAM roles are typically not part of application config, IME it is more of an instance provisioning config, so my applications may not know exactly what it's IAM role is, it just knows that it is using them. (Or rather assumes. We have a policy of no AWS keys in code or config and use the zero constructor arg for AWS service clients and let the client find the keys itself)

Ideally this changeset would loosen the requirement to need to provide keys/IAM roles at all but it looks like you aren't using the AWS EC2 java client and are hand rolling your own requests, and I am guessing that is for a reason?  

At the least, do you see any issues with modifying the logic to assume the instance is using IAM roles and querying the metadata to find the keys based on the IAM role which can be discovered via the metadata service? An instance can only have a single IAM role which can generally be discovered at http://169.254.169.254/latest/meta-data/iam/security-credentials/ and this way users won't need to know which IAM role they are using ahead of time but all of the error checking can remain.

PR #6157, opened 2015-09-08
Adding missing client cache tests

sancar:
Beside adding tests a couple if minor fixes added where new added
tests failed
1)deregisterCacheEntryListener removes from local map if only remote
deregistration is succesfull. Old way can lead to memory leaks when
there is an exception on remote call.
2)CacheEventDataCodec is updated so that it can handle null keys.
This issue has seen in testCompletionEvent
3)Response deserialization is fixed for async methods on old client
cache.


sancar:
verify

asimarslan:
:+1:

cangencer:
:+1:

PR #6159, opened 2015-09-09
Introduced map.impl.event package

pveentjer:
To reduce the size of the map.impl package, a subpackage is created where the event
logic is copied into.

See for Enteprise PR:
https://github.com/hazelcast/hazelcast-enterprise/pull/369


jerrinot:
it looks good to me. :+1: However I don't have a great big picture here.

ahmetmircik:
:+1:

Issue #6120, opened 2015-09-03
[TEST-FAILURE] CacheReadWriteQuorumTest.testPutGetWhenQuorumSizeMet


Donnerbart:
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at com.hazelcast.quorum.PartitionedCluster$3.run(PartitionedCluster.java:118)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:751)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:765)
	at com.hazelcast.quorum.PartitionedCluster.splitFiveMembersThreeAndTwo(PartitionedCluster.java:113)
	at com.hazelcast.quorum.PartitionedCluster.partitionFiveMembersThreeAndTwo(PartitionedCluster.java:62)
	at com.hazelcast.quorum.cache.CacheReadWriteQuorumTest.initialize(CacheReadWriteQuorumTest.java:93)
```
https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-OpenJDK8/674/com.hazelcast$hazelcast/testReport/com.hazelcast.quorum.cache/CacheReadWriteQuorumTest/com_hazelcast_quorum_cache_CacheReadWriteQuorumTest/

mustafaiman:
closed via https://github.com/hazelcast/hazelcast/pull/6164

PR #6164, opened 2015-09-09
fix ordering of membership events execution

mustafaiman:
ClusterServiceImpl does not guarantee membership events' execution order. We come across this behaviour when trying to solve recent quorum failures with @eminn .

Behaviour is as follows:
When 3rd node joins in a 2-node cluster, 3rd node gets two MembershipEvents with cluster size of 2 and 3. 
Node 3 sometimes finishes execution of the second event before the first event. If required quorum size of the cluster is 3, Node 3 begins with quorum state 'false', changes to 'true' and 'false' respectively. In the end, Node 3 thinks quorum is not present although cluster has 3 active nodes.

ClusterServiceImpl should guarantee membership events' execution order.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -217,6 +217,7 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
 
         node.connectionManager.addConnectionListener(this);
 
+        //MEMBERSHIP_EVENT_EXECUTOR is a single threaded executor to ensure that events are executed in correct order.
         nodeEngine.getExecutionService().register(MEMBERSHIP_EVENT_EXECUTOR_NAME, 1, Integer.MAX_VALUE, ExecutorType.CACHED);
 
         registerMetrics();
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
@@ -95,6 +95,8 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -184,6 +186,8 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
     private long firstJoinRequest;
 
     private volatile boolean joinInProgress;
+
+    private final ExecutorService membershipEventExecutor = Executors.newSingleThreadExecutor();
 
     @Probe(name = "lastHeartBeat")
     private volatile long lastHeartBeat;
@@ -1285,7 +1289,7 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
             final MembershipServiceEvent event = new MembershipServiceEvent(membershipEvent);
             for (final MembershipAwareService service : membershipAwareServices) {
                 // service events should not block each other
-                nodeEngine.getExecutionService().execute(ExecutionService.SYSTEM_EXECUTOR, new Runnable() {
+                membershipEventExecutor.execute(new Runnable() {
                     public void run() {
                         if (added) {
                             service.memberAdded(event);
@@ -1384,6 +1388,7 @@ public final class ClusterServiceImpl implements ClusterService, ConnectionListe
 
     @Override
     public void shutdown(boolean terminate) {
+        membershipEventExecutor.shutdown();
         reset();
     }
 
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -185,6 +187,8 @@
 
     private volatile boolean joinInProgress;
 
+    private final ExecutorService membershipEventExecutor = Executors.newSingleThreadExecutor();
```
gurbuzali: rather than creating a standalone `ExecutorService`, we can register one via out `ExecutionService`

```
private static final String MEMBERSHIP_EVENT_EXECUTOR_NAME = "hz:cluster:event";
```
and in `init` method:
```
executionService.register(MEMBERSHIP_EVENT_EXECUTOR_NAME, 1, Integer.MAX_VALUE, ExecutorType.CACHED);
```
and the usage:
```
nodeEngine.getExecutionService().execute(MEMBERSHIP_EVENT_EXECUTOR_NAME, new Runnable() {
                    public void run() {
                        if (added) {
                            service.memberAdded(event);
                        } else {
                            service.memberRemoved(event);
                        }
                    }
                });
```

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1384,6 +1388,7 @@ private void setMembersRef(Map<Address, MemberImpl> memberMap) {
 
     @Override
     public void shutdown(boolean terminate) {
+        membershipEventExecutor.shutdown();
```
gurbuzali: with above mentioned solution, you don't need to handle the life-cycle of the executor

mustafaiman:
I updated the pr as you advised @gurbuzali

hasancelik:
verify

mesutcelik:
closes #6120

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1285,7 +1293,7 @@ private void sendMembershipEventNotifications(MemberImpl member, Set<Member> mem
             final MembershipServiceEvent event = new MembershipServiceEvent(membershipEvent);
             for (final MembershipAwareService service : membershipAwareServices) {
                 // service events should not block each other
```
ahmetmircik: If all events are consumed by 1 threaded executor, don't we making this comment useless?
With this PR, as far as i understand, a service should wait event processing of a previous service in executors queue.

gurbuzali: I've thought the same thing while commenting.
I wonder what was the reasoning behind this comment? Will we break something by blocking?
If yes than we may consider a striped executor here

mustafaiman: If this breaks something, I think striped executor is ok too for quorum issues

hasancelik:
closes #6170

gurbuzali:
here is a better solution from @mdogan 
```
nodeEngine.getEventService().executeEventCallback(new StripedRunnable() {
                    @Override
                    public int getKey() {
                        return service.hashCode();
                    }

                    @Override
                    public void run() {
                        if (added) {
                            service.memberAdded(event);
                        } else {
                            service.memberRemoved(event);
                        }
                    }
                });
```
you don't need to register anything and services will not block each other

eminn:
:+1:

hazelcast/src/main/java/com/hazelcast/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1285,7 +1293,7 @@ private void sendMembershipEventNotifications(MemberImpl member, Set<Member> mem
...
-                nodeEngine.getExecutionService().execute(ExecutionService.SYSTEM_EXECUTOR, new Runnable() {
+                nodeEngine.getExecutionService().execute(MEMBERSHIP_EVENT_EXECUTOR_NAME, new Runnable() {
```
jerrinot: is there any reason why it did use `SYSTEM_EXECUTOR` originally? The Event executor can be quite busy (specially when you use listeners, etc..) and it's dropping events when a queue is full - that's consider a business as usual. Are we ok if we drop a membership event? Can't this make the whole cluster unstable?

I just checked source code and apparently SYSTEM_EXECUTOR has capacity defined as `Integer.MAX_VALUE`.

gurbuzali: correct, we shouldn't let events drop. by advising event-executor I was aiming to use `StripedRunnable` but in this case it is not safe to use.

jerrinot: should we StripedExecutor for SYSTEM_EXECUTOR? (I have no idea about side-effects)

mustafaiman: How about using a single threaded executor and adding a comment to memberAdded and memberRemoved methods of MembershipAwareService as "Implementing classes should not block in this method" ?

mustafaiman:
I added MemberShipAwareService a comment that says memberRemoved and memberAdded methods of implementing classes should be non-blocking as suggested by @gurbuzali . MembershipEvent notifications are executed in single thread now. TransactionManagerService was blocking. I made its memberRemoved method non-blocking.
Please review again.

gurbuzali:
:+1:  for me

serkan-ozal:
:+1:

eminn:
I think we will need a backport to the maintenance for this fix

PR #6165, opened 2015-09-10
Simplified Query execution logic

pveentjer:
All the logic has now been moved inside of the MapQueryEngine.
The QueryOperation is now dumb.


hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -76,17 +97,161 @@ public MapQueryEngineImpl(MapServiceContext mapServiceContext) {
...
         this.queryResultSizeLimiter = new QueryResultSizeLimiter(mapServiceContext, logger);
+        this.operationService = nodeEngine.getOperationService();
+        this.clusterService = nodeEngine.getClusterService();
+        this.localMapStatsProvider = mapServiceContext.getLocalMapStatsProvider();
+        this.parallelEvaluation = nodeEngine.getGroupProperties().getBoolean(QUERY_PREDICATE_PARALLEL_EVALUATION);
+        this.executor = nodeEngine.getExecutionService().getExecutor(QUERY_EXECUTOR);
+    }
+
+    @Override
+    public QueryResult query(String mapName, Predicate predicate) throws ExecutionException, InterruptedException {
```
jerrinot: the name is misleading as it does not indicate it queries local partitions only

hasancelik:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
pveentjer: What is your suggestion? Query local?

jerrinot: queryLocalPartitions? + describe its contract well as now you are making it to be (internal) API.

jerrinot: also check with @ahmetmircik how it affects his work. it sort of effects mine too, but I can solve a few merge conflicts :)

pveentjer: Hmmm.. I see some calls do make remote calls.. ugly...

jerrinot:
:+1: looks like a very good move to me.
@ahmetmircik should have a look too as it could clash with his work.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -80,17 +100,167 @@ public MapQueryEngineImpl(MapServiceContext mapServiceContext, QueryOptimizer op
...
         this.queryResultSizeLimiter = new QueryResultSizeLimiter(mapServiceContext, logger);
         this.queryOptimizer = optimizer;
+        this.operationService = nodeEngine.getOperationService();
+        this.clusterService = nodeEngine.getClusterService();
+        this.localMapStatsProvider = mapServiceContext.getLocalMapStatsProvider();
+        this.parallelEvaluation = nodeEngine.getGroupProperties().getBoolean(QUERY_PREDICATE_PARALLEL_EVALUATION);
+        this.executor = nodeEngine.getExecutionService().getExecutor(QUERY_EXECUTOR);
+    }
+
+    public QueryResultSizeLimiter getQueryResultSizeLimiter() {
```
Donnerbart: Could be package private and with comment `// just for testing`

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -80,17 +100,167 @@ public MapQueryEngineImpl(MapServiceContext mapServiceContext, QueryOptimizer op
...
+    public QueryResult queryLocalPartitions(String mapName, Predicate predicate) throws ExecutionException, InterruptedException {
+        int initialPartitionStateVersion = partitionService.getPartitionStateVersion();
+        Collection<Integer> initialPartitions = mapServiceContext.getOwnedPartitions();
+        MapContainer mapContainer = mapServiceContext.getMapContainer(mapName);
+
+        // first we optimize the query
+        predicate = queryOptimizer.optimize(predicate, mapContainer.getIndexes());
+
+        // then we try to run using an index, but if that doesn't work, we'll try a full table scan
+        // This would be the point where a query-plan should be added. Is should determine if a full table scan
```
Donnerbart: Should we mark this with **TODO**? Also minor typo:
> Is should determine...
> **It** should determine...

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -80,17 +100,167 @@ public MapQueryEngineImpl(MapServiceContext mapServiceContext, QueryOptimizer op
...
+        List<QueryableEntry> sortedSubList = getSortedSubList(toMerge, predicate, nearestAnchorEntry);
+        result.addAll(sortedSubList);
+        return result;
+    }
+
+    private static Collection<Collection<QueryableEntry>> getResult(List<Future<Collection<QueryableEntry>>> lsFutures) {
+        return returnWithDeadline(lsFutures, QUERY_EXECUTION_TIMEOUT_MINUTES, MINUTES, RETHROW_EVERYTHING);
+    }
+
+    private boolean checkPartitionState(int expectedPartitionStateVersion, Predicate predicate) {
```
Donnerbart: I always struggled with the meaning of this method (due to my missing knowledge what a partition state is and the not very helpful method name). What about `isPartitionStateUnchanged` or `hasPartitionStateChanged` as method name? That would at least make clear what `true` and `false` means.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResult.java
```diff
@@ -47,6 +47,11 @@ public QueryResult(long resultLimit) {
         this.resultLimit = resultLimit;
     }
 
+    // just for testing
+    public long getResultLimit() {
```
Donnerbart: Could be package private.

Donnerbart:
:+1: (with minor comments)

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
pveentjer: It isn't really a todo. It is more a reminder that this is the place to be when this stuff gets added.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
pveentjer: Having badly named methods is indeed a problem that keeps popping up. What do you think about this:

```
  private boolean hasPartitionMigrated(int expectedPartitionStateVersion, Predicate predicate) {
        if (expectedPartitionStateVersion != partitionService.getPartitionStateVersion()) {
            logger.info("Partition assignments changed while executing query: " + predicate);
            return true;
        }
        return false;
    }
```

jerrinot: `isPartitionTableAtVersion()` ?

but then you should probably move the logging to the caller.

pveentjer:
I'm going to merge this. This PR has been pending for 2 weeks so @ahmetmircik had enough time to review it. And mostly it is copying logic to the correct place; there is not a functional change. I'll address David his comments directly.

Issue #6156, opened 2015-09-08
TxQueue ordering on rollback can be violated


pveentjer:
Imagine you have the following queue

[1,2,3,4,5]

and you do tx{
  q.poll // 1
  q.poll // 2
  tx.abort
} 

Then the queue content could become [2,1,3,4,5]

The reason is that the QueueContainer doesn't rollback the changes in the oposite order of being added. It is a random order due to the fact that the changes are stored in a hashmap.

PR #6172, opened 2015-09-10
in case of rollback we are putting back the item by checking with itemId, so it will be in order

gurbuzali:
fixes https://github.com/hazelcast/hazelcast/issues/6156

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListContainer.java
@@ -18,6 +18,7 @@ package com.hazelcast.collection.impl.list;
 
 import com.hazelcast.collection.impl.collection.CollectionContainer;
 import com.hazelcast.collection.impl.collection.CollectionItem;
+import com.hazelcast.collection.impl.collection.TxCollectionItem;
 import com.hazelcast.config.ListConfig;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.spi.NodeEngine;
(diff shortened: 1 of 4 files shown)


jerrinot:
@gurbuzali: checkstyle issue: `QueueContainer.java:603:12: Declaring variables, return values or parameters of type 'LinkedList' is not allowed.`

gurbuzali:
fixed it

gurbuzali:
verify

hazelcast/src/main/java/com/hazelcast/collection/impl/list/ListContainer.java
```diff
@@ -50,6 +51,29 @@ public ListConfig getConfig() {
         return config;
     }
 
+    public void rollbackRemove(long itemId) {
```
jerrinot: `@Override` is missing

gurbuzali:
@pveentjer can you also have a look ?

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueContainer.java
```diff
@@ -172,17 +168,29 @@ public Data txnCommitPollBackup(long itemId) {
...
         if (!backup) {
-            getItemQueue().offerFirst(item);
+            addTxItemOrdered(item);
         }
         cancelEvictionIfExists();
         return true;
     }
 
+    private void addTxItemOrdered(TxQueueItem txQueueItem) {
+        ListIterator<QueueItem> iterator = ((List) getItemQueue()).listIterator();
```
mdogan: What about using `itemQueue` field directly instead of cast:
`ListIterator<QueueItem> iterator = itemQueue.listIterator();`

mdogan:
minor comment.

:+1:

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueContainer.java
gurbuzali: we have to use `getItemQueue()` method because there is no queue created for backup but will be created in case of promote via that method. And check-style is bitching about returning `LinkedList`

mdogan: ok. np.

PR #6173, opened 2015-09-10
[BACK-PORT] in case of rollback we are putting back the item by checking with itemId, so it will be in order

gurbuzali:
back-port of https://github.com/hazelcast/hazelcast/pull/6172


gurbuzali:
verify

mdogan:
original PR is already merged.

PR #6176, opened 2015-09-10
Re-implementation of Replicated map

eminn:
Re-implementation of ReplicatedMap with anti-entropy system. ReplicatedMap will no longer use events to replicate events, it’ll use operations instead. It is a weak consistent implementation, although it will behave like eventually consistent system if there are no machine failures. It will converge to the same value if one or more nodes are missing updates.


eminn:
verify

eminn:
verify

hazelcast/src/main/java/com/hazelcast/config/ReplicatedMapConfig.java
```diff
@@ -97,13 +107,14 @@ public ReplicatedMapConfig setName(String name) {
...
      * to other nodes. During this time, multiple puts can be operated and cached up to be sent
-     * out all at once after the delay. This highers the latency for eventually consistency but lowers IO operations.
+     * out all at once after the delay.
      * Default value is 100ms before a replication is operated. If 0, no delay is used and
      * all values are replicated one by one.
      *
      * @return the number of milliseconds after a put is executed before the value is replicated
      * to other nodes.
      */
+    @Deprecated
```
jerrinot: as I user when I see something is deprecated then I'd like see:
1. why?
2. what to used instead where applicable. if it's not applicable then it should be explained in the `why?` part.

jerrinot:
a test failure: `ReplicatedMapAntiEntropyTest.testMapConvergesToLatestValueWhenPrimaryCrashesAndReplicasMissingUpdate:111->HazelcastTestSupport.assertTrueEventually:823 expected:<2d5488b5-a464-4907-a527-b85db8f638ee> but was:<null>`

eminn:
That is the test which we are falling short in terms of eventual consistency. It should be ignored or removed from the system because that will never pass :)

eminn:
Particular `ReplicatedMapAntiEntropyTest.testMapConvergesToLatestValueWhenPrimaryCrashesAndReplicasMissingUpdate` test removed from the codebase.

eminn:
verify

hazelcast/src/main/java/com/hazelcast/monitor/LocalReplicatedMapStats.java
```diff
@@ -17,7 +17,4 @@
 package com.hazelcast.monitor;
 
 public interface LocalReplicatedMapStats extends LocalMapStats {
-
-    long getReplicationEventCount();
```
metanet: We are making API change, right?

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapProxy.java
```diff
@@ -16,45 +16,119 @@
...
+    public void initialize() {
+        service.initializeListeners(name);
+        if (nodeEngine.getClusterService().getSize() == 1) {
+            return;
+        }
+        fireMapDataLoadingTasks();
+        if (!config.isAsyncFillup()) {
+            Collection<ReplicatedRecordStore> stores = service.getAllReplicatedRecordStores(name);
+            for (ReplicatedRecordStore store : stores) {
+                while (!store.isLoaded()) {
```
metanet: should we add a timeout mechanism here?

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/record/AbstractReplicatedRecordStore.java
```diff
@@ -229,105 +210,226 @@ public boolean containsValue(Object value) {
...
+            VersionResponsePair responsePair = new VersionResponsePair(mergingEntry.getValue(), getVersion());
+            sendReplicationOperation(false, getName(), dataKey, dataValue, record.getTtlMillis(), responsePair);
+
+        }
+        return true;
+    }
+
+    protected void sendReplicationOperation(final boolean isRemove, String name, Data key, Data value, long ttl,
+                                            VersionResponsePair response) {
+        Collection<Member> members = nodeEngine.getClusterService().getMembers();
```
metanet: I think we should skip lite members here. We can use clusterService::getMembers(MemberSelector) method. What about this node? Should we also skip it?

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/AbstractReplicatedMapOperation.java
```diff
@@ -16,40 +16,82 @@
...
+            invoke(isRemove, operationService, address, name, key, value, ttl, response);
+        }
     }
 
-    @Override
-    public void afterRun()
-            throws Exception {
+    protected Collection<Address> getMemberAddresses() {
+        Address thisAddress = getNodeEngine().getThisAddress();
+        Collection<Member> members = getNodeEngine().getClusterService().getMembers();
```
metanet: We should skip lite members here.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/PartitionContainer.java
```diff
@@ -0,0 +1,105 @@
...
+                        throw new IllegalStateException("Native memory not yet supported for replicated map");
+                    default:
+                        throw new IllegalStateException("Unhandled in memory format:" + inMemoryFormat);
+                }
+                return replicatedRecordStorage;
+            }
+        };
+    }
+
+    public ConcurrentHashMap<String, ReplicatedRecordStore> getStores() {
```
metanet: minor comment: we can return ConcurrentMap here.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapService.java
```diff
@@ -16,86 +16,151 @@
...
+        for (int i = 0; i < nodeEngine.getPartitionService().getPartitionCount(); i++) {
+            partitionContainers[i] = new PartitionContainer(this, i);
+        }
+        nodeEngine.getExecutionService().getDefaultScheduledExecutor().scheduleWithFixedDelay(new Runnable() {
+            @Override
+            public void run() {
+                if (clusterService.getSize() == 1) {
+                    return;
+                }
+                Collection<Address> addresses = new ArrayList<Address>(clusterService.getMemberAddresses());
```
metanet: We should skip lite members here.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapService.java
```diff
@@ -16,86 +16,151 @@
...
+                addresses.remove(nodeEngine.getThisAddress());
+                for (int i = 0; i < partitionContainers.length; i++) {
+                    Address thisAddress = nodeEngine.getThisAddress();
+                    Address ownerAddress = partitionService.getPartitionOwner(i);
+                    if (!thisAddress.equals(ownerAddress)) {
+                        continue;
+                    }
+                    PartitionContainer partitionContainer = partitionContainers[i];
+                    for (Address address : addresses) {
+                        CheckReplicaVersion checkReplicaVersion = new CheckReplicaVersion(partitionContainer);
```
metanet: Can't we populate versions map before for loop and share it between operations?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationServiceImpl.java
```diff
@@ -491,7 +491,7 @@ private SerializerAdapter createSerializerAdapter(Serializer serializer) {
 
     protected final SerializerAdapter serializerFor(final Class type) {
         SerializerAdapter serializer;
-        if (overrideCustomSerialization) {
+        if (overrideCustomSerialization && constantTypesMap.get(type) == null) {
```
gurbuzali: why is this lookup necessary ?

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapEventPublishingService.java
```diff
@@ -0,0 +1,134 @@
...
+                break;
+            case REMOVED:
+                entryListener.entryRemoved(entryEvent);
+                break;
+            // TODO handle evictAll and clearAll event
+            default:
+                throw new IllegalArgumentException("event type " + entryEvent.getEventType() + " not supported");
+        }
+        String mapName = ((EntryEventData) event).getMapName();
+        if (config.findReplicatedMapConfig(mapName).isStatisticsEnabled()) {
```
gurbuzali: if a wildcard used in the repl-map's name, this lookup becomes very costly (pattern matching). do we have any other choice here ?

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapProxy.java
gurbuzali: +1

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/EntrySetOperation.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public EntrySetOperation(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public void run() throws Exception {
+        ReplicatedMapService service = getService();
+        Collection<ReplicatedRecordStore> stores = service.getAllReplicatedRecordStores(name);
+        Set<Map.Entry<Object, ReplicatedRecord>> entrySet = new HashSet<Map.Entry<Object, ReplicatedRecord>>();
```
metanet: We know that these entries are distinct by their key. Then I think we can put them into an array list etc. (some collection does not require hashing) and return it that way to the user. I suppose we did something similar with the previous impl. Something like lazy collection etc. Additionally, I think we can remove one of these ``entrySet`` and ``entries`` and use only one.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/KeySetOperation.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public KeySetOperation(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public void run() throws Exception {
+        ReplicatedMapService service = getService();
+        Collection<ReplicatedRecordStore> stores = service.getAllReplicatedRecordStores(name);
+        Set keySet = new HashSet();
```
metanet: My comment about EntrySetOperation also applies here.

eminn:
verify

eminn:
verify

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationServiceImpl.java
eminn: This is there for to prevent constant types are not picked up by the custom serializer when serialization override is enabled. This logic will be changed by the @asimarslan work on serialization.

hazelcast/src/main/java/com/hazelcast/monitor/LocalReplicatedMapStats.java
eminn: I think depreciation would be more appropriate.

eminn:
run-parallel

eminn:
run-parallel

eminn:
run-parallel

eminn:
run-parallel

eminn:
run-parallel

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapService.java
```diff
@@ -16,86 +16,152 @@
...
+            @Override
+            public void run() {
+                if (clusterService.getSize() == 1) {
+                    return;
+                }
+                Collection<Address> addresses = new ArrayList<Address>(getMemberAddresses(DATA_MEMBER_SELECTOR));
+                addresses.remove(nodeEngine.getThisAddress());
+                for (int i = 0; i < partitionContainers.length; i++) {
+                    Address thisAddress = nodeEngine.getThisAddress();
+                    Address ownerAddress = partitionService.getPartitionOwner(i);
```
mdogan: This line triggers the partition assignment even if there's no data. You can change it to:

```java
InternalPartition partition = partitionService.getPartition(i, false);
Address ownerAddress = partition.getOwnerOrNull();
```

mdogan: that's why some cluster-state tests fail.

eminn: Thanks for the pointer, good catch! :+1:

metanet:
:+1:

gurbuzali:
:+1:

eminn:
run-parallel

PR #6177, opened 2015-09-10
Deserialization of client message moved to partition thread

sancar:


pveentjer:
I guess the deserialization of the ClientMessage was done on the io thread; right? That would be a very bad thing.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -114,6 +113,7 @@ private void initializeAndProcessMessage() {
         if (!node.joined()) {
             throw new HazelcastInstanceNotActiveException("Hazelcast instance is not ready yet!");
         }
+        parameters = decodeClientMessage(clientMessage);
```
pveentjer: Why just not just call it 'decode' since the client only deals with ClientMessages.

pveentjer:
@sancar any update?

sancar:
Deserialization of client message does not include `serializationService.toObject(data)` , but it can still problem. I will update this pr, asap.

sancar:
verify

sancar:
verify

pveentjer:
verify

sancar:
run-lub-run

sancar:
run-lab-run

hasancelik:
run-lab-run

hasancelik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
sancar: @pveentjer 
This method is used by all MessageTask classes. It will mean that touching all of them. If you still think that we should rename it, I would like to do that as separate pr, right after merging this one. What do you think ?

cangencer:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
sancar: @pveentjer 
Or since we are just in bug fix period. We should postpone the renaming even further ?

ihsandemir:
:+1:

PR #6180, opened 2015-09-10
Applied corrections per Serdar's review on confluence page. Also corrected the md ftl file

ihsandemir:
Applied corrections per Serdar's review on confluence page. Also corrected the md ftl file


ihsandemir:
This is only comment changes and the md ftl change. Hence, merging it without further review.

PR #6181, opened 2015-09-10
Moved internal cluster functionality to internal.cluster

pveentjer:


mdogan:
Is this really needed now? I think this kind of big refactors/moves should be done only at the beginning of development phase, not in the middle. I'm working on atomic cluster shutdown for hot-restart and already touched many places in cluster package. Now if this gets merged, I need to solve lots of conflicts and that's just a pain for me.

pveentjer:
Ok, np.

Issue #6151, opened 2015-09-08
NullPointerException when adding index


wimdeblauwe:
The following unit test shows the NullPointerException:

```
package com.traficon.tmsng.server;

import com.hazelcast.config.Config;
import com.hazelcast.config.MapConfig;
import com.hazelcast.config.MapIndexConfig;
import com.hazelcast.core.Hazelcast;
import com.hazelcast.core.HazelcastInstance;
import com.hazelcast.core.IMap;

import java.io.Serializable;

/**
 * Created by wdb on 07/09/15.
 */
public class HazelcastIndexTest
{
	public static void main( String[] args )
	{
		Config config = new Config();
		MapConfig mapConfig = new MapConfig();
		mapConfig.setName( "test" );
		mapConfig.addMapIndexConfig( new MapIndexConfig( "name", false ) );
		mapConfig.addMapIndexConfig( new MapIndexConfig( "testInterface.foo", false ) );
		config.addMapConfig( mapConfig );
		HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance( config );

		IMap<Object, Object> testMap = hazelcastInstance.getMap( "test" );

		testMap.put( "key", new Person() );
	}

	private static class Person implements Serializable
	{
		private String m_name;
		private TestInterface m_testInterface;

		public String getName()
		{
			return m_name;
		}

		public void setName( String name )
		{
			m_name = name;
		}

		public TestInterface getTestInterface()
		{
			return m_testInterface;
		}

		public void setTestInterface( TestInterface testInterface )
		{
			m_testInterface = testInterface;
		}
	}

	private static interface TestInterface
	{

	}

	private static class TestInterfaceImpl implements TestInterface
	{
		private String m_foo;

		public String getFoo()
		{
			return m_foo;
		}

		public void setFoo( String foo )
		{
			m_foo = foo;
		}
	}
}
```

And this is the resulting exception:

```
Exception in thread "main" com.hazelcast.query.QueryException: java.lang.NullPointerException
	at com.hazelcast.query.impl.getters.ReflectionHelper.createGetter(ReflectionHelper.java:204)
	at com.hazelcast.query.impl.getters.ReflectionHelper.getAttributeType(ReflectionHelper.java:128)
	at com.hazelcast.query.impl.QueryEntry.getAttributeType(QueryEntry.java:174)
	at com.hazelcast.query.impl.IndexImpl.saveEntryIndex(IndexImpl.java:83)
[...]

PR #6182, opened 2015-09-10
Do not throw NPE when intermediate field is an interface

jerrinot:
and the interface does not contain a requested attribute

The issue was originally reported by @wimdeblauwe
 at #6151 Thanks to @serkan-ozal for analysis!

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/getters/ReflectionHelper.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/getters/ReflectionHelper.java
@@ -175,7 +175,7 @@ public final class ReflectionHelper {
                     }
                     if (localGetter == null) {
                         Class c = clazz;
-                        while (!Object.class.equals(c)) {
+                        while (!c.isInterface() && !Object.class.equals(c)) {
                             try {
                                 final Field field = c.getDeclaredField(name);
                                 field.setAccessible(true);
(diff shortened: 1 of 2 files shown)


jerrinot:
I'll prepare a backport once this is reviewed.

jerrinot:
@serkan-ozal: can you please have a look at this one? thanks!

serkan-ozal:
:+1:

serkan-ozal:
This PR fixes NPE but doesn't cover all issue.

@jerrinot will you work on the other problem that is about finding field on field's dynamic type rather than static field type?

jerrinot:
@serkan-ozal: I know it's not fixing the whole issue. The requested use-case is not supported at the moment. This is just fixing the NPE which is obviously a clear bug.

I have hacked an experiment to add somewhat more dynamic behaviour - the naive implementation is rather simple: See this: https://github.com/jerrinot/hazelcast/commit/7b2811f688e9565ca2243907c8f29ec673f68615

but it's not without side-effects: The getters are cached - so it means it uses the first passed object as q blueprint to get actual (=not just declared) types. However what if another object instance returns a different implementation? It's all solvable, but it won't be easy (without performance regressions)

serkan-ozal:
@jerrinot I am not sure also performance overhead of `method.invoke(...)` like you.
Even though this doesn't solve all issues, I think we may use dynamic type checking only if field type is interface or abstract class. If not (I know, still it may be a non-abstract base class), we can simple skip dynamic type checking and can go as it is.

ajermakovics:
:+1:

PR #6183, opened 2015-09-10
Get expiryDelayMillis once and store in a final field

jerrinot:
There is no good reason to fetch the delay from GroupProperties and convert it into milliseconds for each call.


jerrinot:
@ahmetmircik: Can you please have a look at this one? Thanks!

Donnerbart:
The values in a `GroupProperties` instance are final, also the `GroupProperties` instance in `Node` is final. So a value cannot change during runtime. The change should be fine.

:+1:

ahmetmircik:
:+1:

PR #6184, opened 2015-09-11
Minor fixes in GroupProperties

### LINKED ISSUE #6174: GroupProperty defaulting not working properly when programmatic configuration is used
Example: 
Properties `IO_INPUT_THREAD_COUNT` and `IO_OUTPUT_THREAD_COUNT` are configurable. The both defaults to `IO_THREAD_COUNT`

```java
    IO_THREAD_COUNT("hazelcast.io.thread.count", 3),
    IO_INPUT_THREAD_COUNT("hazelcast.io.input.thread.count", IO_THREAD_COUNT),
    IO_OUTPUT_THREAD_COUNT("hazelcast.io.output.thread.count", IO_THREAD_COUNT),
```

This seems to work fine. However when I change the `IO_THREAD_COUNT` property then I'd intuitively expect the new value will be reflected in dependent properties.

In other words: I'd expect following test to pass:
```java
@Test
    public void testInheritance_whenDefaultIsChangedProgramatically_thenDependentPropertyIsChangedToo() {
        Config config = new Config();
        config.setProperty(GroupProperty.IO_THREAD_COUNT, "1");
        HazelcastInstance instance = createHazelcastInstance(config);

        Node node = getNode(instance);
        int inputIOThreadCount = node.getGroupProperties().getInteger(GroupProperty.IO_INPUT_THREAD_COUNT);

        assertEquals(1, inputIOThreadCount);
    }
```

However the test is failing due the way how properties are resolved. I'm not sure if the issue is solvable with the current GrouProperty design. If not that the defaulting mechanism should be removed from (GroupProperty) as it's misleading.

Donnerbart:
* Fixed inheritance of values for `GroupProperties`.
* Fixed unchecked warning in `GroupProperties.getEnum()`.
* Increased code coverage of `GroupProperty` and `GroupProperties`.

Fixes #6174.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -300,6 +300,11 @@ public class GroupProperties {
                 properties[groupProperty.ordinal()] = propertyValue;
                 continue;
             }
+            GroupProperty parent = groupProperty.getParent();
+            if (parent != null) {
+                properties[groupProperty.ordinal()] = properties[parent.ordinal()];
+                continue;
+            }
             properties[groupProperty.ordinal()] = groupProperty.getDefaultValue();
         }
     }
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -300,6 +300,11 @@ public GroupProperties(Config config) {
                 properties[groupProperty.ordinal()] = propertyValue;
                 continue;
             }
+            GroupProperty parent = groupProperty.getParent();
+            if (parent != null) {
+                properties[groupProperty.ordinal()] = properties[parent.ordinal()];
```
jerrinot: what if the parent has not been resolved yet?

Donnerbart: You cannot use an `enum` constant as parent before it's declared. So in `GroupProperty` a parent entry is always declared before its children. Furthermore the iteration is in the order of the declaration (at least the JavaDoc of enum.values() states that):
> Returns an array containing the constants of this enum type, in the order they're declared.

jerrinot: good one, I didn't know that!

jerrinot:
:+1:

Donnerbart:
Do you think we need another reviewer for this?

hazelcast/src/test/java/com/hazelcast/instance/GroupPropertiesTest.java
```diff
@@ -3,38 +3,160 @@
...
+import java.util.concurrent.TimeUnit;
+
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
 
 @RunWith(HazelcastParallelClassRunner.class)
 @Category(QuickTest.class)
-public class GroupPropertiesTest {
+public class GroupPropertiesTest extends HazelcastTestSupport {
```
serkan-ozal: Am I missing something, why extending from `HazelcastTestSupport` is needed?

serkan-ozal:
Minor comment :+1:

PR #5023, opened 2015-04-08
NotEqualPredicate and Exception on unordered index with null value fixed

muratayan:
forward port of #5026

Fixes #4525 
Fixes #4373

PR #6186, opened 2015-09-11
NotEquals predicate is again just a negated Equals predicate

### LINKED ISSUE #6124: Inconsistencies with null handling by predicates
I added it in a comment to https://github.com/hazelcast/hazelcast/issues/4525 but it was ignored so I am opening a new issue.

Since Hazelcast 3.4.3 (or .4) you changed how nulls are treated by equal/notEqual (and probably other) predicates.
Previously if I had three entries with property say "field" being null, "value" and "other" respectively, then searching for Predicates.notEqual("field", "value") would return me the "other" and null one. Starting with 3.4.3 it will only return "other".

You say it is to make it more like SQL. However not(equal()) will still return me null value both and it is NOT what SQL would do.

```
mysql> select * from test1;
+------+
| id   |
+------+
|    2 |
| NULL |
|    1 |
+------+
3 rows in set (0.00 sec)

mysql> select * from test1 where not id=2;
+------+
| id   |
+------+
|    1 |
+------+
1 row in set (0.00 sec)

mysql> 

```

PS: I am not sure what is the correct way here. To me personally, you should not change the original behaviour - after all we are in Java world and null is NOT equal to "value" so this is what people expect...

jerrinot:
In other words: not(equals(p)) <-> notEquals(p)
Fixes #6124

It's partially reverting a changeset #5023.
The original change was introduced to treat NULL-values more like relational databases do.
SQL standard mandates special treating of NULL values, but it order to be compliant we'd
need to introduce three-values logic as specified in SQL. Having just a single predicate
behaving like SQL mandates creates inconsistencies and logical paradoxes.

The #5023 was also fixing a NPE when NotEquals used an index, however this revert will not
re-introduce the NPE  indexes are no longer used for NotEquals predicate. See #5847

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/NotEqualPredicate.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/NotEqualPredicate.java
@@ -16,8 +16,6 @@
 
 package com.hazelcast.query.impl.predicates;
 
-import com.hazelcast.query.impl.ComparisonType;
-import com.hazelcast.query.impl.Index;
 import com.hazelcast.query.impl.QueryContext;
 import com.hazelcast.query.impl.QueryableEntry;
 
@@ -37,11 +35,6 @@ public class NotEqualPredicate extends EqualPredicate {
 
     @Override
     public boolean apply(Map.Entry entry) {
-        Comparable entryValue = readAttribute(entry);
-        if (entryValue == null) {
-            return false;
-        }
-
         return !super.apply(entry);
     }
 
@@ -52,12 +45,7 @@ public class NotEqualPredicate extends EqualPredicate {
 
     @Override
     public Set<QueryableEntry> filter(QueryContext queryContext) {
-        Index index = getIndex(queryContext);
-        if (index != null) {
-            return index.getSubRecords(ComparisonType.NOT_EQUAL, value);
-        } else {
-            return null;
-        }
+        return null;
     }
 
     @Override
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/query/QueryNullIndexingTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/query/QueryNullIndexingTest.java
@@ -83,9 +83,9 @@ public class QueryNullIndexingTest extends HazelcastTestSupport {
         final List<Long> dates =
                 queryIndexedDateFieldAsNullValue(false,
                                                  Predicates.notEqual("date", 2000000L));
-        assertEquals(4, dates.size());
+        assertEquals(9, dates.size());
         assertTrue(dates.containsAll(
-                Arrays.asList(4000000L, 6000000L, 8000000L, 10000000L)));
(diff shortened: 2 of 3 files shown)


jerrinot:
I'm not too sure about a backport here.

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/NotEqualPredicate.java
```diff
@@ -37,11 +35,6 @@ public NotEqualPredicate(String attribute, Comparable value) {
 
     @Override
     public boolean apply(Map.Entry entry) {
-        Comparable entryValue = readAttribute(entry);
```
jerrinot: these 4 lines are the fix.

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/NotEqualPredicate.java
```diff
@@ -52,12 +45,7 @@ public boolean isIndexed(QueryContext queryContext) {
 
     @Override
     public Set<QueryableEntry> filter(QueryContext queryContext) {
-        Index index = getIndex(queryContext);
```
jerrinot: I removed this code as the `NotEqualsPredicates::isIndexed` always returns false. It's not strictly related to this changeset, I followed the boy-scout rule.

hazelcast/src/test/java/com/hazelcast/map/impl/query/QueryNullIndexingTest.java
```diff
@@ -83,9 +83,9 @@ public void testIndexedNullValueOnUnorderedIndexStoreWithNotEqualPredicate() {
         final List<Long> dates =
                 queryIndexedDateFieldAsNullValue(false,
                                                  Predicates.notEqual("date", 2000000L));
-        assertEquals(4, dates.size());
+        assertEquals(9, dates.size());
         assertTrue(dates.containsAll(
-                Arrays.asList(4000000L, 6000000L, 8000000L, 10000000L)));
+                Arrays.asList(4000000L, 6000000L, 8000000L, 10000000L, null)));
```
jerrinot: this is a side-effect of this change. "notEquals(attribute, not-NULL)" not returns values where the `attribute` is set to null.

ajermakovics:
:+1:

muratayan:
:+1:

jerrinot:
@ajermakovics, @muratayan: thanks! What do you think about backport?

muratayan:
@jerrinot I've introduced this via #5026 which was for 3.x. I think we need a backport?

jerrinot:
@muratayan: that's the big question. the behaviour is inconsistent, but someone might rely on it. @ajermakovics WDYT?

ajermakovics:
Neither is ideal but I favor a backport more. Afaik it was an undocumented feature / bug so if someone is relying on it then... tough. And it will break anyway once they upgrade to 3.6 so might as well break early.
With a backport at least the branches won't diverge as much and I don't think we want to support two different behaviors.

jerrinot:
@ajermakovics: alright, I'll send a backport.

PR #6191, opened 2015-09-12
Implement lite members feature

metanet:


metanet:
verify

metanet:
verify

hazelcast/src/main/java/com/hazelcast/cluster/impl/MemberSelectingCollection.java
```diff
@@ -0,0 +1,193 @@
...
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends M> c) {
+        return false;
```
mdogan: `throw new UnsupportedOperationException()`?

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientReplicatedMapProxy.java
```diff
@@ -102,7 +102,7 @@ public V put(K key, V value, long ttl, TimeUnit timeUnit) {
         Data valueData = toData(value);
         Data keyData = toData(key);
         ClientMessage request = ReplicatedMapPutCodec.encodeRequest(name, keyData, valueData, timeUnit.toMillis(ttl));
-        ClientMessage response = invoke(request);
+        ClientMessage response = invokeOnPartitionOfProxyName(request);
```
sancar: As far as I understand we can not send to random node anymore,  because that random node might be lite member and does not have any data in it. So what will happen in case of non-smart client. Are users can not access to replicated map anymore if they are using non-smart client ? Or will they be routed to a non-lite member by lite-member ?

metanet:
verify

metanet:
verify

mdogan:
Some tests has serial runner but also categorized as `ParallelTest`. Is this intentional or typo?

```java
@Category({QuickTest.class, ParallelTest.class})
```

metanet:
It is by mistake. I will fix those.

metanet:
It is by mistake. I will fix those.

mdogan:
added some minor comments. 

definitely good work, every aspect seems taken care of  :+1:

mdogan:
a small suggestion:

In cases where `isClusterFormedByOnlyLiteMembers()` is called, we only need to know there is at least a single data member, I mean there's no need to count all data members:

```java
private boolean isClusterFormedByOnlyLiteMembers() {
       final ClusterServiceImpl clusterService = node.getClusterService();
       return clusterService.getSize(DATA_MEMBER_SELECTOR) == 0;
}
```

In Hazelcast 2, I was using following method, which returns immediately when the first data member is found:

```java
private boolean hasStorageMember() {
    for (MemberImpl member : concurrentMapManager.lsMembers) {
        if (!member.isLiteMember()) {
            return true;
        }
    }
    return false;
}
```

metanet:
Good point! I will do that

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientReplicatedMapProxy.java
```diff
@@ -79,8 +82,14 @@
...
     protected static final String NULL_VALUE_IS_NOT_ALLOWED = "Null value is not allowed!";
 
+    private static final AtomicIntegerFieldUpdater<ClientReplicatedMapProxy> TARGET_PARTITION_ID_UPDATER =
+            AtomicIntegerFieldUpdater.newUpdater(ClientReplicatedMapProxy.class, "targetPartitionId");
+
     private volatile ClientHeapNearCache<Object> nearCache;
     private final AtomicBoolean nearCacheInitialized = new AtomicBoolean();
+    // all requests are forwarded to the same node via a random partition id
+    // do not use this field directly. use getOrInitTargetPartitionId() instead.
+    private volatile int targetPartitionId = Operation.GENERIC_PARTITION_ID;
```
sancar: How about we use PartitionSpecificClientProxy ? 
It was used by List , Set and Queue Proxies of client. It will simplify the code.

metanet: Doesn't it pick the partition of the name? I wanted to pick a random partition without triggering a partition table assignment on proxy creation so that there is no behavioural change while creating the proxy. If it is not a problem, I can get a random partition on proxy init.

sancar: Sorry. I was thinking about suggesting onInitialize method rather than lazy initialization. I forget the fact that name is used in PartitionSpecificClientProxy.

sancar:
I could not find the changes on ReplicatedMapMessageTasks. I think they should extend AbstractPartitionMessageTask rather than AbstractCallableMessageTask.

sancar:
:+1:  Looks good.

PR #6193, opened 2015-09-13
Introduction to local and thread-local system properties for serial and parallel test runners

serkan-ozal:
With this PR:
- `HazelcastSerialClassRunner` uses local system properties for each serial test so even a test sets a system property, this doesn't effect other tests.
- `HazelcastParallelClassRunner` uses thread-local system properties (`InheritableThreadLocal` is used so child threads share same thread-local system properties with their parents) for each parallel test so parallel tests don't effect other test's system properties even they run at the same time.

Inspired by this test failure (https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3736/testReport/junit/com.hazelcast.instance/GroupPropertiesTest/setProperty_ensureUsageOfDefaultValue/) due to shared system properties between parallel test runs.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/HazelcastParallelClassRunner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/HazelcastParallelClassRunner.java
@@ -22,19 +22,34 @@ import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.Statement;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.io.PrintWriter;
+import java.io.Reader;
+import java.io.Writer;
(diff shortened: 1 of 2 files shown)


jerrinot:
very clever! 

I'm not sure if I can reason about all possible side-effects, but let's give it a try. 
:+1: from me!

serkan-ozal:
A few notes on this PR:

- In fact, this PR doesn't protect system properties from corruption in all cases but just makes system property changes invisible to others (parallel tests methods and next test methods). For example, if a test class sets a system property inside `@BeforeClass` (or static initializer block) and forget to clear it on `@AfterClass`, this property will be visible for next tests also.

- If a class (test or application) is loaded first time inside the test method (which is very possible) and if it sets a global (used in general) system property inside its static initializer, since this changes will be handled as local change, it will not exist in the real system properties. So this property will be disappeared in the system.

PR #6194, opened 2015-09-13
Query Optimizations, part 1

jerrinot:
Introduces Query Optimizer based on static rewriting rules. It treats
Predicates as immutable and returns a modifed copy when optimized is
found. This is the first part where only a handful of predicates are
optimized. Other PRs will follow.

Optimized can be disabled by setting property
hazelcast.query.optimizer.type to "none"

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
@@ -22,6 +22,7 @@ import com.hazelcast.internal.monitors.HealthMonitorLevel;
 import com.hazelcast.map.QueryResultSizeExceededException;
 import com.hazelcast.map.impl.query.QueryResultSizeLimiter;
 import com.hazelcast.query.TruePredicate;
+import com.hazelcast.query.impl.predicates.QueryOptimizerFactory;
 
 import java.util.concurrent.TimeUnit;
 
@@ -533,6 +534,19 @@ public enum GroupProperty implements HazelcastProperty {
      * @see #QUERY_RESULT_SIZE_LIMIT
      */
     QUERY_MAX_LOCAL_PARTITION_LIMIT_FOR_PRE_CHECK("hazelcast.query.max.local.partition.limit.for.precheck", 3),
+
+    /**
+     * Type of Query Optimizer.
+     * Valid Values:
+     * <ul>
+     *     <li>RULES - for optimizations based on static rules</li>
+     *     <li>NONE - optimization are disabled</li>
+     * </ul>
+     *
+     * Values are case sensitive
+     *
+     */
+    QUERY_OPTIMIZER_TYPE("hazelcast.query.optimizer.type", QueryOptimizerFactory.Type.RULES.toString()),
 
     /**
      * Forces the JCache provider, which can have values client or server, to force the provider type.
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
@@ -18,6 +18,7 @@ package com.hazelcast.map.impl.query;
 
 import com.hazelcast.query.PagingPredicate;
 import com.hazelcast.query.Predicate;
+import com.hazelcast.query.impl.Indexes;
 import com.hazelcast.query.impl.QueryableEntry;
 import com.hazelcast.util.IterationType;
 
@@ -93,4 +94,15 @@ public interface MapQueryEngine {
      * @return {@link QueryResult}
      */
     QueryResult newQueryResult(int numberOfPartitions);
+
+
+    /**
+     * Return optimized version of the query. The input predicate itself must not be modified in anyway.
+     * If no optimization is done then it may return the original instanceg.
+     *
+     * @param predicate
+     * @param indexes
+     * @return optimized version of input predicate or the predicate itself if no optimization was performed
+     */
+    Predicate optimize(Predicate predicate, Indexes indexes);
 }
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
@@ -30,9 +30,11 @@ import com.hazelcast.query.PagingPredicate;
 import com.hazelcast.query.PagingPredicateAccessor;
 import com.hazelcast.query.Predicate;
 import com.hazelcast.query.TruePredicate;
+import com.hazelcast.query.impl.Indexes;
 import com.hazelcast.query.impl.QueryEntry;
 import com.hazelcast.query.impl.QueryResultEntry;
 import com.hazelcast.query.impl.QueryableEntry;
+import com.hazelcast.query.impl.predicates.QueryOptimizer;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.OperationService;
 import com.hazelcast.util.Clock;
@@ -68,14 +70,16 @@ public class MapQueryEngineImpl implements MapQueryEngine {
     private final QueryResultSizeLimiter queryResultSizeLimiter;
     private final SerializationService serializationService;
     private final InternalPartitionService partitionService;
+    private final QueryOptimizer queryOptimizer;
 
-    public MapQueryEngineImpl(MapServiceContext mapServiceContext) {
+    public MapQueryEngineImpl(MapServiceContext mapServiceContext, QueryOptimizer optimizer) {
         this.mapServiceContext = mapServiceContext;
         this.nodeEngine = mapServiceContext.getNodeEngine();
         this.serializationService = nodeEngine.getSerializationService();
         this.partitionService = nodeEngine.getPartitionService();
         this.logger = nodeEngine.getLogger(getClass());
         this.queryResultSizeLimiter = new QueryResultSizeLimiter(mapServiceContext, logger);
+        this.queryOptimizer = optimizer;
     }
 
     @Override
@@ -273,6 +277,11 @@ public class MapQueryEngineImpl implements MapQueryEngine {
     @Override
     public QueryResult newQueryResult(int numberOfPartitions) {
         return new QueryResult(queryResultSizeLimiter.getNodeResultLimit(numberOfPartitions));
+    }
+
+    @Override
+    public Predicate optimize(Predicate predicate, Indexes indexes) {
+        return queryOptimizer.optimize(predicate, indexes);
     }
 
     private void checkIfNotPagingPredicate(Predicate predicate) {
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/Indexes.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/Indexes.java
@@ -84,7 +84,13 @@ public class Indexes {
         }
     }
 
-    Index getIndex(String attribute) {
+    /**
+     * Get index for a given attribute. If the index does not exist then returns null.
+     *
+     * @param attribute
+     * @return Index for attribute or null if the index does not exist.
+     */
+    public Index getIndex(String attribute) {
         return mapIndexes.get(attribute);
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/OrToInVisitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/OrToInVisitor.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.query.impl.predicates;
+
+import com.hazelcast.query.Predicate;
+import com.hazelcast.query.impl.Indexes;
+import com.hazelcast.util.collection.InternalMultiMap;
+
+import java.util.List;
+import java.util.Map;
+
+import static com.hazelcast.util.collection.ArrayUtils.createCopy;
+
+/**
+ * Transforms predicate (attribute = 1 or attribute = 2 or attribute = 3) into
+ * (attribute in (1, 2, 3)
+ *
+ * InPredicate is easier to evaluate in both indexed and non-indexed scenarios.
+ *
+ * When index is not used then {@link Predicate} require just a single reflective call
+ * the extract an attribute for {@link com.hazelcast.query.impl.QueryableEntry} while
+ * disjunction of equalPredicate(s) requires one reflective call for each equalPredicate.
+ *
+ * The performance is even more significant when tha {@link InPredicate#attribute} is indexed.
+ * As then the InPrecicate can be evaluated by just a single hit into index.
+ *
+ *
+ * When index is used then
+ *
+ *
+ */
+public class OrToInVisitor extends AbstractVisitor {
+
+    private static final int MINIMUM_NUMBER_OF_OR_TO_REPLACE = 5;
+
+    @Override
+    public Predicate visit(OrPredicate orPredicate, Indexes indexes) {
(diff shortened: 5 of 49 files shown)


jerrinot:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
```diff
@@ -93,4 +94,15 @@ Set query(String mapName, Predicate predicate,
      * @return {@link QueryResult}
      */
     QueryResult newQueryResult(int numberOfPartitions);
+
+
+    /**
+     * Return optimized version of the query. The input predicate itself must not be modified in anyway.
+     * If no optimization is performed then it can return the original instance of predicate.
```
pveentjer: If no optimization is performed then it can return the original instance

hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
```diff
@@ -535,6 +535,17 @@
...
     /**
+     * Type of Query Optimizer.
+     * Valid Values:
+     * <ul>
+     *     <li>rules - for optimizations based on static rules</li>
+     *     <li>none - optimization are disabled</li>
+     * </ul>
+     *
+     */
+    QUERY_OPTIMIZER_TYPE("hazelcast.query.optimizer.type", "rules"),
```
pveentjer: Is there no enum?

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -76,6 +80,7 @@ public MapQueryEngineImpl(MapServiceContext mapServiceContext) {
         this.partitionService = nodeEngine.getPartitionService();
         this.logger = nodeEngine.getLogger(getClass());
         this.queryResultSizeLimiter = new QueryResultSizeLimiter(mapServiceContext, logger);
+        this.queryOptimizer = QueryOptimizerFactory.newOptimizer(nodeEngine.getGroupProperties());
```
pveentjer: Personally I'm not crazy about such factories. Perhaps it is a flavor thing; but can't you inject the right query optimizer as constructor arg and deal with the construction when the QueryEngineImpl is made?

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -275,6 +280,11 @@ public QueryResult newQueryResult(int numberOfPartitions) {
         return new QueryResult(queryResultSizeLimiter.getNodeResultLimit(numberOfPartitions));
     }
 
+    @Override
+    public Predicate optimize(Predicate predicate, Indexes indexes) {
```
pveentjer: Should this not be an implementation detail? Now you expose the fact that queries can be optimized. It is not relevant to the outside that this is in place.

pveentjer: I guess it is caused by the call the 'indexes' ? The indexes should be behind the QueryEngine.

hazelcast/src/main/java/com/hazelcast/query/Visitable.java
```diff
@@ -0,0 +1,39 @@
...
+package com.hazelcast.query;
+
+import com.hazelcast.query.impl.Indexes;
+import com.hazelcast.query.impl.predicates.Visitor;
+
+/**
+ * Predicates which can be visited by optimizer.
+ *
+ */
+public interface Visitable {
```
pveentjer: Personally I would go for a VisiblePredicate extends Predicate.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
jerrinot: in theory - yes. in practice - no.
you want to minimize number of optimization attempts as it can be quite costly.

hazelcast/src/main/java/com/hazelcast/query/Visitable.java
```diff
@@ -0,0 +1,39 @@
...
+     * Accept visitor. Predicate can either return it's own instance if no modification
+     * was done as a result of the visit. In the case there is a change needed then
+     * the predicate has to return changed copy of itself. Predicates has to be treated
+     * as immutable for optimization purposes.
+     *
+     * @param visitor visitor to accept
+     * @param indexes indexes
+     * @return itself or its changed copy
+     */
+    Predicate accept(Visitor visitor, Indexes indexes);
```
pveentjer: This looks a bit fishy. If you need context e.g. Indexes, you can set it as a field in your Visitor implementation. There is no need to pass it as an argument next to the visitor.

hazelcast/src/main/java/com/hazelcast/query/impl/Indexes.java
```diff
@@ -84,7 +84,7 @@ public void saveEntryIndex(QueryableEntry queryableEntry) throws QueryException
         }
     }
 
-    Index getIndex(String attribute) {
+    public Index getIndex(String attribute) {
```
pveentjer: Can you add a line of javadoc (if that doesnt' exist) about the behavior? Especially what happens when the index for the attribute doesn't exist.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
jerrinot: good one, I'll fix it

hazelcast/src/main/java/com/hazelcast/query/Visitable.java
jerrinot: I can do that.

hazelcast/src/main/java/com/hazelcast/query/Visitable.java
pveentjer: Also the return value isn't needed. If in your visitor you need to deal with 'return' values, you can just store them internally in a field. 

A visitor normally looks like this:

```
void accept(Visitor visitor)
```

jerrinot: I do agree. but then I'd need to create a visitor per map. As each map has a different instance of `Indexes`. or some look-up mechanism.

hazelcast/src/main/java/com/hazelcast/query/impl/Indexes.java
jerrinot: sure

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
jerrinot: I didn't want to include Optimizer creation logic inside MapQueryEngineImpl. 
I can pass it as an argument, I'm fine with that.

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/Negatable.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.hazelcast.query.impl.predicates;
+
+import com.hazelcast.query.Predicate;
+
+/**
+ * Predicates can implement this interface if they can convert itself into logical negation
+ *
+ */
+public interface Negatable {
```
pveentjer: I would call this NegatablePredicate since it is only meant for predicates.

jerrinot: ok

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/RuleBasedQueryOptimizer.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+public final class RuleBasedQueryOptimizer implements QueryOptimizer {
+    private final Visitor betweenVisitor = new BetweenVisitor();
+    private final Visitor flatteningVisitor = new FlatteningVisitor();
+    private final Visitor orToInVisitor = new OrToInVisitor();
+
+    public <K, V> Predicate<K, V> optimize(Predicate<K, V> predicate, Indexes indexes) {
+        Predicate optimized = predicate;
+        if (optimized instanceof Visitable) {
```
pveentjer: You are now traversing the tree 3 times. In the future we could have a look at traversing the three only once but chaining the visitors.

jerrinot: Indeed. that's one of the things I want to look at. (it's not as easy as the first impression could be)

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
pveentjer: Sure. But eventually it is best to inject a QueryOptimizer into the QueryEngine as argument. Anyhow currently not a big deal. Once we move towards DI we can fix this easily

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
pveentjer: I don't understand why exposing it is relevant to the number of calls to optimize. The underlying implementation is smelly because the queryengine doesnt deal with queries using index. 

Imho this should really be an implementation detail.

pveentjer:
It looks ok to me. We need to clean up the QueryEngine and QueryOperation since logic is smeared over these 2 classes. QueryOperation should be dumb. 

Apart from some design issue, I'm ok with these changes. But we need to spend some time on query cleanup.

ahmetmircik:
Does re-writing happen in every execution of the same query? Is there a cache mechanism for previously rewritten queries? Or is it worth to implement something like that? 
WDYT @jerrinot ?

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/OrToInVisitor.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Map;
+
+import static com.hazelcast.util.collection.ArrayUtils.createCopy;
+
+/**
+ * Transforms predicate (attribute = 1 or attribute = 2 or attribute = 3) into
+ * (attribute in (1, 2, 3)
+ *
+ */
+public class OrToInVisitor extends AbstractVisitor {
```
ahmetmircik: A simple note which explains when/why this transformation is needed would be nice.

jerrinot:
@ahmetmircik: as of now it rewrites it for every execution. query cache definitely makes sense, but it will require further changes. I'm planning to introduce a concept of "prepared queries" and it will include the cache.

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/OrToInVisitor.java
jerrinot: good idea!

jerrinot: done

jerrinot:
verify

ahmetmircik:
What about a design like this: all optimizers are under `com.hazelcast.query.impl.optimizer` package, 
predicates do not know any optimizer specific thing like `negate` or `visitible`..etc. Optimizer receives predicate and does its job. Just trying to isolate things.

I am also ok with current structure, but i expect more isolation. That's my only remaining concern for this PR, otherwise :+1: once green.

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/FlatteningVisitorTest.java
```diff
@@ -0,0 +1,114 @@
...
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static com.hazelcast.query.impl.TypeConverters.INTEGER_CONVERTER;
+import static org.mockito.Mockito.withSettings;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class FlatteningVisitorTest {
+
```
tombujok: In general this test doesn't check nothing apart from the length of the outer predicate which is not sufficient IMHO.
The optimizer may return the same predicate three times and the test would pass...
It applies to all Visitor tests.

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/FlatteningVisitorTest.java
```diff
@@ -0,0 +1,114 @@
...
+        Predicate a3 = equal("a3", 3);
+
+        OrPredicate innerOr = (OrPredicate) or(a2, a3);
+        OrPredicate outerOr = (OrPredicate) or(a1, innerOr);
+
+        OrPredicate result = (OrPredicate) visitor.visit(outerOr, mockIndexes);
+        Predicate[] inners = result.predicates;
+        assertEquals(3, inners.length);
+    }
+
```
tombujok: I've just added a test with two levels of inner-predicates and it fails. Is it intended so that we do not support nesting with more then one level? 
```java
@Test
    public void visitAndPredicate_whenHasMoreThanOneInnerAndPredicate_thenFlattenIt() {
        //(a1 = 1 and (a2 = 2 and (a3 = 3 and a4 = 4))  -->  (a1 = 1 and a2 = 2 and a3 = 3 and a4 = 4)

        Predicate a1 = equal("a1", 1);
        Predicate a2 = equal("a2", 2);
        Predicate a3 = equal("a3", 3);
        Predicate a4 = equal("a4", 4);

        AndPredicate innerInnerAnd = (AndPredicate) and(a3, a4);
        AndPredicate innerAnd = (AndPredicate) and(a2, innerInnerAnd);
        AndPredicate outerAnd = (AndPredicate) and(a1, innerAnd);

        AndPredicate result = (AndPredicate) visitor.visit(outerAnd, mockIndexes);
        Predicate[] inners = result.predicates;
        assertEquals(4, inners.length);
    }
```

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/BetweenVisitorTest.java
```diff
@@ -0,0 +1,208 @@
...
+import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.hamcrest.Matchers.*;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class BetweenVisitorTest {
+
```
tombujok: Would be great to see how the visitor behaves if there's nesting in the expression.

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/FlatteningVisitorTest.java
jerrinot: good point

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/FlatteningVisitorTest.java
jerrinot: traversing is not a responsibility of visitors. I'll add an integration test to make sure nested cases are covered as well.

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/BetweenVisitorTest.java
jerrinot: can you give me an example?

tombujok: I mean expressions that couldn't be flattened.
I know that this visitor expects flattened input, but there are certainly cases where the input to this visitor won't be flat, e.g. 
`(age >= 18 or ((height >= 170 and weight<55) || (height >= 180 and weight<60)) or age <= 28)`

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/BetweenVisitorTest.java
```diff
@@ -0,0 +1,208 @@
...
+        //(age >= 5 and age <= 5)  -->  (age = 5)
+        Predicate left = greaterEqual("attribute", 5);
+        Predicate right = lessEqual("attribute", 5);
+        Predicate and = and(left, right);
+
+        EqualPredicate p = (EqualPredicate) visitor.visit((AndPredicate) and, mockIndexes);
+        assertEquals("attribute", p.attribute);
+        assertEquals(5, p.value);
+    }
+
```
tombujok: Would be great to have a test that covers optimization of the `between` operation when the operator is the same in both cases, like
`age > 5 and age > 8`

tombujok:
:+1: great job (just minor comments related to tests)

PR #6198, opened 2015-09-14
[DONT MERGE] Still running

pveentjer:
Hazelcast has a mechanism to detect if an operation/response got lost and the invocation won't wait indefinitely. 

However there are a few issues with the current approach:
* The logic for waiting for the future is very complex; the waitForResponse perhaps is one of the most complicated methods in our code base.
* It relies on operations. If the system is under a lot of pressure, adding additional operations on top of the load is not going to help out. Also if there are no threads available to execute the operations, the invocation will falsely assume the operation it is not running.
* Async calls don't timeout since there is no thread that call the remote side to ask if the operation is still running.

So instead of asking if an operation is still running, each member will periodically scan all executing operations and send back the call id's (an array of longs wrapped into a packet) to the calling member.

When this packet is received, is put into the OperationDeadMansSwitch; which has its own thread to deal with it. So there is no reliance on operations. The packet is a priority packet to make sure it is preferred above regular packets.

When the OperationDeadMansSwitch receives the callids packets, it notifies the invocation registry for each of these callid's. Each invocation is then 'tagged' with a new timestamp. The invocationregistry will expire any invocation that has not been tagged for a too long period.


StanAccy:
Peter

Doesnt this just flip the responsibility around to the other side of the operation - what if the processing node goes down? Are there going to be mechanisms in place to detect failures on either side of this async invoke / wait-for-response mechanism?

pveentjer:
Hi StanAccy,

if the processing node goes down, then memberlistener will kick in and all invocations for that member get automatically aborted. 

What do you mean with the other mechanisms?

PR #6200, opened 2015-09-14
Fixes IOBalancer Memory Leak 

### LINKED ISSUE #6199: IOBalancer has a memory leak
IOBalancer keeps references of all the socket reader/writers but when destroying the connection they release the references for only the ones which has endpoints.

It could be seen from the image below that socket writers consume %77 of the total memory after all the connections to hazelcast are destroyed.

screen shot 2015-09-14 at 10 55 32 am

eminn:
If the connection is destroyed and removed from active connections, it should be removed from iobalancer regardless of connection endpoint. 

After this fix memory consumption pattern is like below when hundreds of connections are opened/closed to the hazelcast instance. 

screen shot 2015-09-14 at 10 54 47 am

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
@@ -393,34 +393,39 @@ public class TcpIpConnectionManager implements ConnectionManager, PacketHandler 
             logger.finest("Destroying " + connection);
         }
 
-        activeConnections.remove(connection);
-        final Address endPoint = connection.getEndPoint();
-        if (endPoint != null) {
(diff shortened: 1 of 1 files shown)


pveentjer:
Good work. 

a few questions:
- can you create a unit test which verifies the behavior?
- when does this particular exception happen; so how did you stumble on this issue?

eminn:
I won't write a stress test by checking the memory size for detecting a leak since it is fragile but I will try to write a unit test with mocks if possible.

This issue happens when REST API is used, after some time oome happens even though cluster is not doing any data storage etc.

pveentjer:
Hi Emin, a regular unit test is perfectly fine.

We should add rest simulator tests :)

bwzhang2011:
@eminn, could it be back port ?

eminn:
@bwzhang2011 see #6201

PR #6201, opened 2015-09-14
[Backport] Fixes IOBalancer Memory Leak

eminn:
If the connection is destroyed and removed from active connections, it should be removed from iobalancer regardless of connection endpoint. 

After this fix memory consumption pattern is like below when hundreds of connections are opened/closed to the hazelcast instance. 

screen shot 2015-09-14 at 10 54 47 am


eminn:
Pull request for tests : #6206 .

Can we prioritize this pull request and merge this ?

pveentjer:
:+1:

PR #6204, opened 2015-09-14
Simplified MultiMapContainer concurrency

pveentjer:
The container was using field updaters; but since only a single thread
will write to the container (operation thread) there is no need for
a mechanism that deals with concurrent writes. A simple volatile write
is sufficient.

The idgenerator has been made thread-unsafe. So a simple non volatile
long is sufficient.


hasancelik:
:+1:

ahmetmircik:
:+1:

PR #6210, opened 2015-09-14
Fixed for [TEST-FAILURE] map.BasicMapTest.testPutWithTtl

### LINKED ISSUE #6196: [TEST-FAILURE] BasicMapTest.testPutWithTtl
```
java.lang.AssertionError: expected:<value> but was:<null>
	at com.hazelcast.map.BasicMapTest.testPutWithTtl(BasicMapTest.java:1107)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3739/testReport/junit/com.hazelcast.map/BasicMapTest/testPutWithTtl/

tombujok:


gurbuzali:
how about this 
```
    @Test
    public void testPutWithTtl() throws InterruptedException {
        final IMap<String, String> map = getInstance().getMap("testPutWithTtl");

        map.put("key", "value", 2, TimeUnit.SECONDS);
            @Override
                assertNull(map.get("key"));
            }
        });
    }
```

tombujok:
@gurbuzali Looks much simpler, but I thought this test is also supposed to verify whether the `entryEvicted` listener has been fired. If we can skip it, I prefer your option too.

gurbuzali:
listener is used for detecting that the ttl expired for this test. we may want to test `entryEvicted` listener too in a separate test, maybe we already have tests for that, I don't know.

pveentjer:
Can you add short description in title? This makes an overview of issues resolves easier to understand.

tombujok:
Fixed as suggested

gurbuzali:
do we need a back-port here ?

PR #6192, opened 2015-09-12
IdentifyConverter converts NULL-values into NULL-objects

jerrinot:
Fixed #5807

It changes TypeConverter to be an abstract class instead of interface. It's not part of user API and it's the simplest way to make sure all converters deal with NULL input values correctly. 

Also a few small consistency related clean-ups. See commits for more informations.

PR #6214, opened 2015-09-14
[Backport] IdentifyConverter converts NULL-values into NULL-objects

jerrinot:
Backport of #6192


jerrinot:
@ajermakovics, @serkan-ozal: can you guys have a look? thanks!

hazelcast/src/main/java/com/hazelcast/query/impl/UUIDConverter.java
```diff
@@ -33,6 +33,9 @@ private UUIDConverter() {
 
     @Override
     public Comparable convert(Comparable value) {
+        if (value == null) {
```
jerrinot: this was another converter where null-value handling was not implemented properly

serkan-ozal:
:+1:

ajermakovics:
:+1:

PR #6215, opened 2015-09-15
[DONT MERGE] Queries keyset not download values + other improvements

pveentjer:
IMap.keySet is implemented inefficiently since it also downloads the values. Also plenty of other not needed complexity that is removed; e.g. why force using a set to store results of a list can be used (e.g. map.values).


pveentjer:
This PR has been replaced by this cleaned up version:

https://github.com/hazelcast/hazelcast/pull/6288

So I'm going to close this one.

PR #6216, opened 2015-09-15
Renamed IndexService to Indexes

pveentjer:


jerrinot:
is there actually any dependency on map?

pveentjer:
Conceptually there is because every map has a MapIndexes instance.

jerrinot:
sure, but you could use it outside the map you wanted, can't you?

pveentjer:
it is linked to querable entry. 

But I could rename it.. to Indexes?

jerrinot:
"Indexes" sounds better. to me anyway. 

I understand the issue with "IndexService" is that it implies there is just one. But in reality there is one per each map. Am I correct?

pveentjer:
Yes. One for each map. Hence 'service' is confusing.

pveentjer:
Renamed to Indexes

ahmetmircik:
`Indexes` makes sense to me. Also this PR needs an enterprise counterpart.
:+1: once green.

jerrinot:
:+1:

pveentjer:
Enterprise PR:
https://github.com/hazelcast/hazelcast-enterprise/pull/376

PR #6224, opened 2015-09-15
[DON'T MERGE] JCache Split-Brain Implemantation

serkan-ozal:


hazelcast/src/main/java/com/hazelcast/cache/BuiltInCacheMergePolicies.java
```diff
@@ -0,0 +1,124 @@
...
+ * <p>
+ * <b>Note:</b>
+ *      When a new built-in {@link com.hazelcast.cache.CacheMergePolicy} is implemented,
+ *      its definition should be added here also.
+ * </p>
+ */
+public enum BuiltInCacheMergePolicies {
+
+    /**
+     * Cache merge policy that merges cache entry to from source to destination directly.
```
mesutcelik: unnecessary `to` in the javadoc.

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
```diff
@@ -0,0 +1,75 @@
...
+            // Register `CacheMergePolicy` by its constant
+            mergePolicyMap.put(mergePolicy.name(), cacheMergePolicy);
+            // Register `CacheMergePolicy` by its name
+            mergePolicyMap.put(mergePolicy.getName(), cacheMergePolicy);
+        }
+    }
+
+    public CacheMergePolicy getMergePolicy(String className) {
+        if (className == null) {
+            throw new NullPointerException("Class name is mandatory!");
```
mesutcelik: It might be better to use `InvalidConfigurationException`. @bilalyasar ?

bilalyasar: Yes, and I expect fail fast here or null checking should be done somewhere else.

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
```diff
@@ -0,0 +1,75 @@
...
+            } catch (Exception e) {
+                nodeEngine.getLogger(getClass()).severe(e);
+                throw ExceptionUtil.rethrow(e);
+            }
+        }
+    };
+
+    public CacheMergePolicyProvider(NodeEngine nodeEngine) {
+        this.nodeEngine = nodeEngine;
+        mergePolicyMap = new ConcurrentHashMap<String, CacheMergePolicy>();
```
mesutcelik: you can use `this.mergePolicyMap` to make it consistent...

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private final ConstructorFunction<String, CacheMergePolicy> policyConstructorFunction
+            = new ConstructorFunction<String, CacheMergePolicy>() {
+        @Override
+        public CacheMergePolicy createNew(String className) {
+            try {
+                return newInstance(nodeEngine.getConfigClassLoader(), className);
+            } catch (Exception e) {
+                nodeEngine.getLogger(getClass()).severe(e);
+                throw ExceptionUtil.rethrow(e);
```
mesutcelik: this place could be a candidate for `InvalidConfigurationException`.

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
```diff
@@ -0,0 +1,75 @@
...
+        addOutOfBoxPolicies();
+    }
+
+    private void addOutOfBoxPolicies() {
+        for (BuiltInCacheMergePolicies mergePolicy : BuiltInCacheMergePolicies.values()) {
+            final CacheMergePolicy cacheMergePolicy = mergePolicy.newInstance();
+            // Register `CacheMergePolicy` by its constant
+            mergePolicyMap.put(mergePolicy.name(), cacheMergePolicy);
+            // Register `CacheMergePolicy` by its name
+            mergePolicyMap.put(mergePolicy.getName(), cacheMergePolicy);
```
mesutcelik: `getImpClassName` might be more readable? When I checked this piece, `name` and `getName` didn't make difference. I had to go and check `BuiltInCacheMergePolicies` implementation.

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/PutIfAbsentCacheMergePolicy.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.hazelcast.cache.CacheEntryView;
+import com.hazelcast.cache.CacheMergePolicy;
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+
+import java.io.IOException;
+
+/**
+ * `PassThroughCacheMergePolicy` policy merges cache entry to from source to destination
```
mesutcelik: javadoc typo....

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheRemovePartitionLostListenerMessageTask.java
```diff
@@ -38,8 +37,8 @@ public CacheRemovePartitionLostListenerMessageTask(ClientMessage clientMessage,
     @Override
     protected Object call() {
         ICacheService service = getService(CacheService.SERVICE_NAME);
```
emrahkocaman: I think accessing service name using different classes is confusing. Can't we use `ICacheService.SERVICE_NAME` also here?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheAddPartitionLostListenerMessageTask.java
```diff
@@ -61,9 +61,8 @@ public void partitionLost(CachePartitionLostEvent event) {
                 new InternalCachePartitionLostListenerAdapter(listener);
         final EventFilter filter = new CachePartitionLostEventFilter();
         final CacheService service = getService(CacheService.SERVICE_NAME);
-        final EventRegistration registration = service.getNodeEngine().
-                getEventService().registerListener(AbstractCacheService.SERVICE_NAME,
-                parameters.name, filter, listenerAdapter);
+        final EventRegistration registration = service.getNodeEngine().getEventService()
+                    .registerListener(ICacheService.SERVICE_NAME, parameters.name, filter, listenerAdapter);
         final String registrationId = registration.getId();
         endpoint.setListenerRegistration(CacheService.SERVICE_NAME, parameters.name, registrationId);
```
emrahkocaman: Should be `ICacheService.SERVICE_NAME`.

mesutcelik:
can you please squash commits?

bilalyasar:
can you update `ConfigXmlGenerator` class according to your xsd-xml config changes?

hazelcast/src/main/java/com/hazelcast/config/CacheConfig.java
```diff
@@ -71,6 +71,7 @@
     private WanReplicationRef wanReplicationRef;
     private List<CachePartitionLostListenerConfig> partitionLostListenerConfigs;
     private String quorumName;
+    private String mergePolicy = CacheSimpleConfig.DEFAULT_CACHE_MERGE_POLICY;
```
bilalyasar: do we need to update `toString` method.

hazelcast/src/main/java/com/hazelcast/cache/impl/client/CacheAddPartitionLostListenerRequest.java
```diff
@@ -68,9 +67,8 @@ public void partitionLost(CachePartitionLostEvent event) {
                 new InternalCachePartitionLostListenerAdapter(listener);
         final EventFilter filter = new CachePartitionLostEventFilter();
         final ICacheService service = getService();
-        final EventRegistration registration = service.getNodeEngine().
-                getEventService().registerListener(AbstractCacheService.SERVICE_NAME,
-                name, filter, listenerAdapter);
+        final EventRegistration registration = service.getNodeEngine().getEventService()
+                .registerListener(ICacheService.SERVICE_NAME, name, filter, listenerAdapter);
         final String registrationId = registration.getId();
         endpoint.setListenerRegistration(CacheService.SERVICE_NAME, name, registrationId);
```
bilalyasar: should be `ICacheService.SERVICE_NAME`

hazelcast/src/main/java/com/hazelcast/cache/BuiltInCacheMergePolicies.java
serkan-ozal: right

hazelcast/src/main/java/com/hazelcast/cache/impl/client/CacheAddPartitionLostListenerRequest.java
serkan-ozal: Good catch. I have updated most of them as `ICacheService.SERVICE_NAME` but I had missed a few of them :)

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
serkan-ozal: ok

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
serkan-ozal: make sense

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
serkan-ozal: ok

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/CacheMergePolicyProvider.java
serkan-ozal: ok

hazelcast/src/main/java/com/hazelcast/cache/impl/merge/policy/PutIfAbsentCacheMergePolicy.java
serkan-ozal: right

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheRemovePartitionLostListenerMessageTask.java
serkan-ozal: Good catch. I have updated most of them as `ICacheService.SERVICE_NAME` but I had missed a few of them :)

hazelcast/src/main/java/com/hazelcast/config/CacheConfig.java
serkan-ozal: I have removed `quorumName` from the `toString` method since I think it is not necessary giving information at there.

hazelcast/src/main/java/com/hazelcast/cache/OriginalTypeAwareCacheMergePolicy.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * <p>
+ * At worst case value is returned from the merge method as selected and this means that at all cases
+ * value is accessed. So even the the convertion is done as lazy, it will be processed at this point.
+ * But by default, their (key and value) storage types are used unless this {@link com.hazelcast.cache.CacheMergePolicy} is used.
+ * </p>
+ *
+ * @see com.hazelcast.cache.CacheMergePolicy
+ */
+public interface OriginalTypeAwareCacheMergePolicy extends CacheMergePolicy {
```
gurbuzali: The motivation behind this interface is reasonable, most of the users will not require `value` while merging the entries.

This way if a user does not implement this interface `getValue` may return a `Data` depending on the configured in-memory-format. Can't we do it other way around, to be safe ? So if a user knows that s/he will not need the value then s/he will implement this interface to get rid of `value` deserialization. 

What do you think @ahmetmircik ? Should we go for optimal solution or for safe one ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheAddPartitionLostListenerMessageTask.java
serkan-ozal: Good catch. I have updated most of them as ICacheService.SERVICE_NAME but I had missed a few of them :)

hazelcast/src/main/java/com/hazelcast/cache/OriginalTypeAwareCacheMergePolicy.java
ahmetmircik: @gurbuzali +1 for safe one.

serkan-ozal: @gurbuzali @ahmetmircik Even though I prefer optimal solution, I will switch to safe one. So I will make the default one converts to object (still built-in merge policies still won't convert to object but user defined ones). If user wants the optimized one, s(he) will need to implement `RawTypeAwareCacheMergePolicy` (which extends from `CacheMergePolicy`) and this won't convert stored values to their original types.  

But one more thing, if the put is done through client and if the client classpath is different from server classpath, converting the value to object by default will cause `ClassNotFoundException` since its original type (class) is only exist at client side. Have you considered this case?

ahmetmircik: @serkan-ozal you are right about client case it may be problematic but here, in my opinion, for most use cases, it will be expected to see an object value and if one needs raw types, your raw-type-aware solution can be used.

serkan-ozal: @gurbuzali @ahmetmircik thanks for your comments. will update PR as raw type aware

serkan-ozal:
closed in favor of https://github.com/hazelcast/hazelcast/pull/6270

PR #6229, opened 2015-09-16
An exception during wan replication should never drop an event

henri-tremblay:


emrahkocaman:
Hi @henri-tremblay , This is a nice finding thanks for your contribution.
Build is failed due to some `checkstyle` errors, could you please take a look at them?
You can run `checkstyle` locally by using `mvn clean install -DskipTests -Pcheckstyle` quickly.
I think we can merge this than.

hazelcast/src/test/java/com/hazelcast/wan/impl/WanNoDelayReplicationTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertSame;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.spy;
+
+/**
+ * @author Henri Tremblay
+ */
```
emrahkocaman: Could you please replace this javadoc with a more explanatory one?

emrahkocaman: Or better, there already is a test class related to wan replication -> [WanReplicationTest](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/wan/WanReplicationTest.java)
Could you please move your test cases to that class?

henri-tremblay:
I have fixed the checkstyle issues. I've removed the comment on the test class since your tests usually don't have such

Donnerbart:
verify

hazelcast/src/test/java/com/hazelcast/wan/impl/WanNoDelayReplicationTest.java
henri-tremblay: In fact I did that on purpose. Because the other test is a NightlyTest.
Mine is a real unit test.

Donnerbart: FYI: Nightly tests don't count for the code coverage measurement and are not executed in the PR builder. So if these tests are important for one of those two things, we should keep them as separate quick (PR builder + coverage) or slow test (coverage).

emrahkocaman: Agreed

emrahkocaman:
verify

PR #6230, opened 2015-09-17
Improved null checking ICountdownLatch.await

pveentjer:
The unit was not checked for null, but just ignored in case of null.
This is a violation of the fail fast principle.

Also the javadoc was updated + test was added.


jerrinot:
:+1:

Issue #4199, opened 2014-12-01
ClientExecutorServiceTest.testCancelFutureAfterCancellationAwareTaskTimeOut


bilalyasar:
```
java.lang.AssertionError: null
    at org.junit.Assert.fail(Assert.java:86)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at org.junit.Assert.assertTrue(Assert.java:52)
    at com.hazelcast.client.executor.ClientExecutorServiceTest.testCancelFutureAfterCancellationAwareTaskTimeOut(ClientExecutorServiceTest.java:157)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-OpenJDK8/com.hazelcast$hazelcast-client/72/testReport/junit/com.hazelcast.client.executor/ClientExecutorServiceTest/testCancelFutureAfterCancellationAwareTaskTimeOut/

hasancelik:
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.client.executor.ClientExecutorServiceTest.testCancelFutureAfterCancellationAwareTaskTimeOut(ClientExecutorServiceTest.java:164)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK6/com.hazelcast$hazelcast-client-new/635/testReport/junit/com.hazelcast.client.executor/ClientExecutorServiceTest/testCancelFutureAfterCancellationAwareTaskTimeOut/

sancar:
Related to https://github.com/hazelcast/hazelcast/issues/4677

Holmistr:
I'm closing this issue, since the test is ignored. The root cause is https://github.com/hazelcast/hazelcast/issues/4677 , so the test should be enabled and fixed, once the root cause is fixed. I created a PR just to link the issue also to the test https://github.com/hazelcast/hazelcast/pull/12522

PR #6232, opened 2015-09-17
Fixed the ExecutorServiceCancelTest

pveentjer:
Fixed the test by adding a latch to guarantee that a cancel doesn't overtake a task submit
and you run into reordering problems.


Donnerbart:
:+1:

sancar:
@pveentjer We also have similar problem here ?
https://github.com/hazelcast/hazelcast/issues/4199

Issue #6234, opened 2015-09-17
[TEST-FAILURE]  com.hazelcast.client.cache.CacheBasicClientTest.testIteratorDuringInsertion


serkan-ozal:
```
Tests run: 29, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 14.03 sec <<< FAILURE! - in com.hazelcast.client.cache.CacheBasicClientTest
testIteratorDuringInsertion(com.hazelcast.client.cache.CacheBasicClientTest)  Time elapsed: 0.188 sec  <<< ERROR!
java.lang.NullPointerException: null
    at com.hazelcast.cache.CacheBasicAbstractTest.testIteratorDuringInsertion(CacheBasicAbstractTest.java:338)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
    at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
    at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:134)
    at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
    at com.hazelcast.test.HazelcastParallelClassRunner.access$301(HazelcastParallelClassRunner.java:47)
    at com.hazelcast.test.HazelcastParallelClassRunner$TestRunner.run(HazelcastParallelClassRunner.java:144)
    at java.lang.Thread.run(Thread.java:744)
```

Reported by @dbrimley

Donnerbart:
In the test there is
```
[...]

PR #6236, opened 2015-09-17
Cleanup of basic cache tests and temporary fix of test failure

Donnerbart:
Most of this PR is cleanup of the tests to get rid of the "yellow alert" due to missing generics. Also we got rid of the sleep calls to make the tests more predictable.

This PR also temporary fixes a test failure (#6234) due to a race between cache eviction and cache iterator. The fix is to ensure no eviction happens (by limiting a random put to size of the pre-filled cache). The underlying issue will be resolved later and will need new tests which enforce the eviction to happen.

I know that this PR is quite noisy, due to the cleanups. I recommend to review it commit by commit.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
@@ -7,13 +7,8 @@ import com.hazelcast.internal.serialization.SerializationService;
 import com.hazelcast.internal.serialization.impl.DefaultSerializationServiceBuilder;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.test.AssertTask;
-import com.hazelcast.test.HazelcastParallelClassRunner;
-import com.hazelcast.test.annotation.ParallelTest;
-import com.hazelcast.test.annotation.QuickTest;
 import com.hazelcast.util.CacheConcurrentHashMap;
 import org.junit.Test;
-import org.junit.experimental.categories.Category;
-import org.junit.runner.RunWith;
 
 import javax.cache.Cache;
 import javax.cache.configuration.FactoryBuilder;
@@ -50,13 +45,11 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
-@RunWith(HazelcastParallelClassRunner.class)
-@Category({QuickTest.class, ParallelTest.class})
 public abstract class CacheBasicAbstractTest extends CacheTestSupport {
 
     @Test
     public void testPutGetRemoveReplace() throws InterruptedException, ExecutionException {
-        ICache cache = createCache();
+        ICache<String, String> cache = createCache();
 
         cache.put("key1", "value1");
         assertEquals("value1", cache.get("key1"));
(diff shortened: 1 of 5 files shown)


hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
```diff
@@ -534,7 +500,7 @@ public void testJSRCreateDestroyCreate()
         Cache<Integer, String> cache = cacheManager.createCache(cacheName, config);
         assertNotNull(cache);
 
-        Thread.sleep(1000);
+        sleepSeconds(1);
```
Donnerbart: I'm not sure why we have this sleep here. Is it safe to remove or are we waiting for some internal state of the created cache?

hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
```diff
@@ -274,175 +272,144 @@ public void testIterator() {
...
-                    try {
-                        cache.put(i, i);
-                        LockSupport.parkNanos(1);
-                    } catch (Throwable ignored) {
-                    }
-                }
+        AbstractCacheWorker worker = new AbstractCacheWorker() {
+            @Override
+            void doRun(Random random) {
+                int i = random.nextInt(size);
```
Donnerbart: This is the actual fix of the spurious test failure, since an unlimited key range will lead to eviction (which breaks the concurrent iterator).

hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
serkan-ozal: @Donnerbart safe to remove. It is left over. Cache config creation on other nodes is changed to sync by not blocking partition threads long time ago.

hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
serkan-ozal: Default max-size policy is `ENTRY_COUNT` and its default size is `10.000`. Based on our estimation based max-size evaluation algorithm, partition's max size, which is used internally, can be calculated via 
`EntryCountCacheMaxSizeChecker::calculateMaxPartitionSize(int maxEntryCount, int partitionCount)`. Eviction is done for a partition if its size is reached to max partition size.

So, in theory, it is guaranteed that if the element count is less than max partition size, there is no eviction even though data is distributed as unbalanced. But if the element count is bigger than max partition size, there may be eviction due to unbalanced data (data can be stored only one partition).

Also, in theory, it is guaranteed that if the element count is bigger than `max_partition_size * partition_count`,
there is eviction even though data is distributed as perfectly balanced.

As a result, if we want to be sure that there is no eviction, we should set the `size` to max partition size, so it is guaranteed that there is no eviction. To find the max partition size, the following code snippet can be used:
``` java
final int partitionCount = ...;
final int maxPartitionSize = 
	EntryCountCacheMaxSizeChecker.calculateMaxPartitionSize(
		EvictionConfig.DEFAULT_MAX_ENTRY_COUNT, partitionCount);
		
```

For example, max partition size and max global size with default size (which is `10.000`) and default partition count (which is `271`):
```
Max partition size : 55
Max global size    : 14905
```

I know, in this test, we are giving keys sequencially from `0` to `size` and there is no unbalanced state, it would be nice if we don't trust this fact in the test and set the size to max partition size. 

Also, for the `testIteratorDuringInsertion_withoutEviction`, I prefer putting half (or less than size) of the elements before worker thread start and so worker thread can make an insertion (not update).

Donnerbart: That sounds very reasonable. I'll prepare an update of the PR on Monday. Thanks again for those insights about the eviction, that should make the changes very easy.

serkan-ozal:
once green :+1:

PR #6237, opened 2015-09-17
Added NodeStates [ACTIVE, SHUTTING_DOWN, SHUTDOWN]

mdogan:
Before there was only an `active` flag in Node. It was set to `false` after partition information check during shutdown. Problem is, during shutdown, before Node itself becomes inactive, all operations are allowed to execute (including operations mutating the data). This can lead to modification of partition data after partition is synced with its backups.

Now, we have one state to separate begin (`SHUTTING_DOWN`) and end (`SHUTDOWN`) of shutdown process. In `SHUTTING_DOWN` state, only certain operations are allowed to run. These operations are (should be) marked with `AllowedDuringShutdown` interface.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/NodeState.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/NodeState.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.instance;
+
+/**
+ * Possible states of a {@link Node} during its lifecycle.
+ * Some actions/operations may be allowed or denied
+ * according to current state of the {@link Node}.
+ *
+ * @see Node#start()
+ * @see Node#shutdown(boolean)
+ * @see com.hazelcast.spi.impl.AllowedDuringShutdown
+ *
+ * @since 3.6
+ */
+public enum NodeState {
+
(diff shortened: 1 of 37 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -234,7 +235,7 @@ public Object toObject(Object object) {
 
     @Override
     public boolean isActive() {
-        return node.isActive();
+        return node.getState() == NodeState.ACTIVE;
```
metanet: How about moving this check into node.isActve() method? Then all other node.isActive() checks remain as they are

mdogan: My intention is to make caller aware of the all node states by removing `node.isActive()`, it's not `active/passive` anymore. 

`NodeEngine.isActive()` is called by over SPI level, mostly at that level only important thing is if instance is `ACTIVE` or not. But `Node.getState()` is used on lower levels where difference between `SHUTTING_DOWN` and `SHUTDOWN` is important.

metanet: ok, thanks for the clarification!

metanet:
:+1:

hazelcast/src/main/java/com/hazelcast/instance/NodeState.java
```diff
@@ -0,0 +1,56 @@
...
+ * @see Node#start()
+ * @see Node#shutdown(boolean)
+ * @see com.hazelcast.spi.impl.AllowedDuringShutdown
+ *
+ * @since 3.6
+ */
+public enum NodeState {
+
+    /**
+     * Initial state of the Node. An {@code ACTIVE} node is allowed
```
jerrinot: is this another change? as currently the node is not active until it's explicitly activated, is it correct?

mdogan: no change here. Currently node activated during start process. It's same in this branch too.

mdogan: Before: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/instance/Node.java#L290

After: https://github.com/mdogan/hazelcast/blob/node-state/hazelcast/src/main/java/com/hazelcast/instance/Node.java#L292

jerrinot:
:+1:

PR #6239, opened 2015-09-17
Cleanup of the test package.

Donnerbart:
The normal, boring cleanup stuff, driven by IntelliJ and our current CheckStyle rules (which are not automatically applied here).

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
@@ -19,6 +19,13 @@ package com.hazelcast.test;
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.test.annotation.Repeat;
+import org.junit.After;
+import org.junit.internal.runners.statements.RunAfters;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.InitializationError;
+import org.junit.runners.model.MultipleFailureException;
+import org.junit.runners.model.Statement;
+
 import java.lang.management.ManagementFactory;
 import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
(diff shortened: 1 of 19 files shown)


Donnerbart:
verify

hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
```diff
@@ -76,17 +78,16 @@ public static String getTestMethodName() {
     }
 
     /**
-     * Creates a BlockJUnit4ClassRunner to run {@code klass}
+     * Creates a BlockJUnit4ClassRunner to run {@code classType}
      *
      * @throws org.junit.runners.model.InitializationError if the test class is malformed.
      */
-    public AbstractHazelcastClassRunner(Class<?> klass) throws InitializationError {
-        super(klass);
```
pveentjer: classType is not an improvements over klass. With klass it is at at least obvious that a 'class' is expected, but that the name is in use. classType is going to be the type of a class? So either use klass or clazz..

hazelcast/src/test/java/com/hazelcast/test/TestHazelcastInstanceFactory.java
```diff
@@ -149,10 +151,10 @@ public void terminateAll() {
     private static List<Address> createAddresses(AtomicInteger ports, String... addressArray) {
         checkElementsNotNull(addressArray);
 
-        final int count = addressArray.length;
+        int count = addressArray.length;
         List<Address> addresses = new ArrayList<Address>(count);
-        for (int i = 0; i < count; i++) {
-            addresses.add(createAddress(addressArray[i], ports.incrementAndGet()));
+        for (String anAddressArray : addressArray) {
```
pveentjer: anAddressArray to indicate a field of type string? I think it should be 'address'.

hazelcast/src/test/java/com/hazelcast/test/TestPartitionUtils.java
```diff
@@ -170,17 +161,15 @@ public int getPartitionId() {
 
         @Override
         public void run() {
-            final InternalPartitionService partitionService = node.nodeEngine.getPartitionService();
-            final long[] replicaVersions = partitionService.getPartitionReplicaVersions(partitionId);
-            this.replicaVersions = Arrays.copyOf(replicaVersions, replicaVersions.length);
+            InternalPartitionService partitionService = node.nodeEngine.getPartitionService();
+            long[] replicaVersionsArray = partitionService.getPartitionReplicaVersions(partitionId);
```
pveentjer: I prefer the replicaVersions. Array in this case only adds noise; doesn't provide any added value.

pveentjer:
Minor but important comments. Once resolved I'm happy to put my thumbs up sign on it.

hazelcast/src/test/java/com/hazelcast/test/TestPartitionUtils.java
Donnerbart: `replicaVersions` is already the name of a field. So in the old code `final long[] replicaVersions` was hiding this field, so a `this.replicaVersions` was needed a line below. Hiding fields seems not to be a good practice for me (it's also a SonarQube issue).

hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
Donnerbart: I just wanted to give the names a unified and more English name. I think I saw `classType` in the JDK itself, but I may be wrong. I picked `clazz` for now.

pveentjer: With type of class you start to enter complex type systems. Where a type of a type (class) is also a 'kind'. So it puts me on the wrong foot.

pveentjer:
:+1:

PR #6240, opened 2015-09-17
Fixes #5662 - InvocationNestedRemoteTest failure

### LINKED ISSUE #5662: [TEST-FAILURE] InvocationNestedRemoteTest.whenPartition_callsIncorrectPartition
```
	at org.junit.Assert.fail(Assert.java:95)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationNestedRemoteTest.whenPartition_callsIncorrectPartition(InvocationNestedRemoteTest.java:105)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.7/com.hazelcast$hazelcast/166/testReport/junit/com.hazelcast.spi.impl.operationservice.impl/InvocationNestedRemoteTest/whenPartition_callsIncorrectPartition/

tombujok:
Fix for #5662 - InvocationNestedRemoteTest failure

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
@@ -186,13 +186,13 @@ public final class ClassicOperationExecutor implements OperationExecutor {
         return threads;
     }
 
-    @SuppressFBWarnings({"EI_EXPOSE_REP" })
+    @SuppressFBWarnings({ "EI_EXPOSE_REP" })
     @Override
     public OperationRunner[] getPartitionOperationRunners() {
         return partitionOperationRunners;
     }
 
-    @SuppressFBWarnings({"EI_EXPOSE_REP" })
+    @SuppressFBWarnings({ "EI_EXPOSE_REP" })
     @Override
     public OperationRunner[] getGenericOperationRunners() {
         return genericOperationRunners;
@@ -248,18 +248,24 @@ public final class ClassicOperationExecutor implements OperationExecutor {
             return true;
         }
 
+        // allowed to invoke non partition specific task
         if (op.getPartitionId() < 0) {
             return true;
         }
 
-        // we are allowed to invoke from non PartitionOperationThreads (including GenericOperationThread)
+        // allowed to invoke from non PartitionOperationThreads (including GenericOperationThread)
         if (!(currentThread instanceof PartitionOperationThread)) {
             return true;
         }
 
-        // we are only allowed to invoke from a PartitionOperationThread if the operation belongs to that
-        // PartitionOperationThread.
         PartitionOperationThread partitionThread = (PartitionOperationThread) currentThread;
+        OperationRunner runner = partitionThread.getCurrentOperationRunner();
+        if (runner != null) {
+            // non null runner means it's a nested call
+            // in this case partitionId of both inner and outer operations have to match
+            return runner.getPartitionId() == op.getPartitionId();
+        }
+
         return toPartitionThreadIndex(op.getPartitionId()) == partitionThread.threadId;
     }
 
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
@@ -36,14 +36,11 @@ public class InvocationFutureTest extends HazelcastTestSupport {
         operationService = getOperationService(local);
     }
 
-    // ================= isDone ==========================================
-
     @Test
     public void isDone_whenNullResponse() throws ExecutionException, InterruptedException {
         DummyOperation op = new DummyOperation(null);
 
         InternalCompletableFuture future = operationService.invokeOnTarget(null, op, getAddress(local));
-        // first we wait for the future to complete.
         future.get();
 
         assertTrue(future.isDone());
@@ -51,76 +48,40 @@ public class InvocationFutureTest extends HazelcastTestSupport {
 
     @Test
     public void isDone_whenWaitResponse() {
-        DummyOperation op = new DummyOperation() {
-            {
-                // we need to set the call-id to prevent running the operation on the calling-thread.
-                setPartitionId(1);
-            }
-
-            @Override
-            public void run() throws Exception {
-                Thread.sleep(5000);
-            }
-        };
+        DummyOperation op = new GetLostPartitionOperation();
 
         InvocationFuture future = (InvocationFuture) operationService.invokeOnTarget(null, op, getAddress(local));
         future.set(InternalResponse.WAIT_RESPONSE);
+
         assertFalse(future.isDone());
     }
 
     @Test
     public void isDone_whenInterruptedResponse() {
-        DummyOperation op = new DummyOperation() {
-            {
-                // we need to set the call-id to prevent running the operation on the calling-thread.
-                setPartitionId(1);
-            }
-
-            @Override
-            public void run() throws Exception {
-                Thread.sleep(5000);
-            }
-        };
+        DummyOperation op = new GetLostPartitionOperation();
 
         InvocationFuture future = (InvocationFuture) operationService.invokeOnTarget(null, op, getAddress(local));
         future.set(InternalResponse.INTERRUPTED_RESPONSE);
+
         assertTrue(future.isDone());
     }
 
     @Test
     public void isDone_whenTimeoutResponse() {
-        DummyOperation op = new DummyOperation() {
-            {
-                // we need to set the call-id to prevent running the operation on the calling-thread.
-                setPartitionId(1);
-            }
-
-            @Override
-            public void run() throws Exception {
-                Thread.sleep(5000);
-            }
-        };
+        DummyOperation op = new GetLostPartitionOperation();
 
         InvocationFuture future = (InvocationFuture) operationService.invokeOnTarget(null, op, getAddress(local));
         future.set(InternalResponse.TIMEOUT_RESPONSE);
+
         assertTrue(future.isDone());
     }
 
     @Test
     public void isDone_whenNoResponse() {
-        DummyOperation op = new DummyOperation() {
-            {
-                // we need to set the call-id to prevent running the operation on the calling-thread.
-                setPartitionId(1);
-            }
-
-            @Override
-            public void run() throws Exception {
-                Thread.sleep(5000);
-            }
-        };
+        DummyOperation op = new GetLostPartitionOperation();
 
         InternalCompletableFuture future = operationService.invokeOnTarget(null, op, getAddress(local));
+
         assertFalse(future.isDone());
     }
 
@@ -129,22 +90,19 @@ public class InvocationFutureTest extends HazelcastTestSupport {
         DummyOperation op = new DummyOperation("foobar");
 
         InternalCompletableFuture future = operationService.invokeOnTarget(null, op, getAddress(local));
+
         assertTrue(future.isDone());
     }
-
-    // ========================= andThen ==================================
 
     // There is a bug: https://github.com/hazelcast/hazelcast/issues/5001
     @Test
     public void andThen_whenNullResponse_thenCallbackExecuted() throws ExecutionException, InterruptedException {
         DummyOperation op = new DummyOperation(null);
-
+        final ExecutionCallback callback = mock(ExecutionCallback.class);
         InternalCompletableFuture future = operationService.invokeOnTarget(null, op, getAddress(local));
-        // first we wait for the future to complete.
         future.get();
 
-        // if we now register a callback, it could complete immediately since a response already has been set (NULL_RESPONSE)
-        final ExecutionCallback callback = mock(ExecutionCallback.class);
+        // Callback can be completed immediately since a response (NULL_RESPONSE) has been already set.
         future.andThen(callback);
 
         assertTrueEventually(new AssertTask() {
@@ -154,4 +112,18 @@ public class InvocationFutureTest extends HazelcastTestSupport {
             }
         });
     }
+
+    // Needed to have an invocation and this is the easiest way how to get one and do not bother with its result.
+    private static class GetLostPartitionOperation extends DummyOperation {
+        {
+            // we need to set the call-id to prevent running the operation on the calling-thread.
+            setPartitionId(1);
+        }
+
+        @Override
+        public void run() throws Exception {
+            Thread.sleep(5000);
+        }
+    }
+
 }
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
@@ -8,115 +8,120 @@ import com.hazelcast.spi.OperationService;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.annotation.ParallelTest;
 import com.hazelcast.test.annotation.QuickTest;
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
 
 @RunWith(HazelcastParallelClassRunner.class)
 @Category({QuickTest.class, ParallelTest.class})
 public class InvocationNestedLocalTest extends InvocationNestedTest {
 
-    private final String response = "someresponse";
+    private static final String RESPONSE = "someresponse";
+
+    @Rule
+    public ExpectedException expected = ExpectedException.none();
 
     @Test
-    public void whenPartition_callsGeneric() {
-        HazelcastInstance hz = createHazelcastInstance();
-        OperationService operationService = getOperationService(hz);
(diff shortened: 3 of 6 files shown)


hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
```diff
@@ -8,115 +8,120 @@
...
-        HazelcastInstance hz = createHazelcastInstance();
-        OperationService operationService = getOperationService(hz);
+    public void invokeOnPartition_outerGeneric_innerGeneric_forbidden() {
+        HazelcastInstance local = createHazelcastInstance();
+        OperationService operationService = getOperationService(local);
+
+        InnerOperation innerOperation = new InnerOperation(RESPONSE, GENERIC_OPERATION);
+        OuterOperation outerOperation = new OuterOperation(innerOperation, GENERIC_OPERATION);
+
+        expected.expect(Exception.class);
```
bilalyasar: just friendly advice :)
in general, in hazelcast we use (expected = SomeExceptionClassName) pattern.
You can see an example here: https://github.com/hazelcast/hazelcast/blob/08857da91e497365437700243db026fda4c5c31a/hazelcast/src/test/java/com/hazelcast/config/QueryCacheConfigTest.java#L15

tombujok: Thx. I know this one - but it's not specific enough IMHO in many cases. 
You don't really know which line threw the exception. 
It's also impossible to match what the message was which is useful in some cases, where the same type of exception may be thrown twice from the same method, like it's matched here:
https://github.com/hazelcast/hazelcast/pull/6240/files#diff-efd38b559d7c6a3cc584ef18e630e672R79

There are cases where notation proposed by you is OK, but I don't see a reason why we should be dogmatic about it ;)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
```diff
@@ -248,18 +248,25 @@ public boolean isInvocationAllowedFromCurrentThread(Operation op, boolean isAsyn
...
+        // if already running an operation for a partition (id >= 0) it means that it's a nested invocation
+        // in this case partitionId of both nested operations has to match
         PartitionOperationThread partitionThread = (PartitionOperationThread) currentThread;
+        OperationRunner runner = partitionThread.getCurrentOperationRunner();
+        if (runner != null) {
+            return runner.getPartitionId() == op.getPartitionId();
+        }
+
+        // TODO: this is racy... what if the partition is being migrated at the same time?
         return toPartitionThreadIndex(op.getPartitionId()) == partitionThread.threadId;
```
pveentjer: being local or not is currently not the question.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
```diff
@@ -248,18 +248,25 @@ public boolean isInvocationAllowedFromCurrentThread(Operation op, boolean isAsyn
...
         }
 
-        // we are only allowed to invoke from a PartitionOperationThread if the operation belongs to that
-        // PartitionOperationThread.
+        // if already running an operation for a partition (id >= 0) it means that it's a nested invocation
+        // in this case partitionId of both nested operations has to match
         PartitionOperationThread partitionThread = (PartitionOperationThread) currentThread;
+        OperationRunner runner = partitionThread.getCurrentOperationRunner();
+        if (runner != null) {
+            return runner.getPartitionId() == op.getPartitionId();
```
pveentjer: Can you add a line of comment explaining that a nested call is happening here.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
pveentjer: Can you add a line of comment that this is the first call (so not nested).

tombujok: Do you suggest removing the TODO?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
tombujok: There's one just above the line you quoted
```
// if already running an operation for a partition (id >= 0) it means that it's a nested invocation
// in this case partitionId of both nested operations has to match
```
Would like to have another comment here?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
```diff
@@ -36,91 +36,52 @@ public void setup() {
...
-    // ================= isDone ==========================================
-
     @Test
     public void isDone_whenNullResponse() throws ExecutionException, InterruptedException {
         DummyOperation op = new DummyOperation(null);
 
         InternalCompletableFuture future = operationService.invokeOnTarget(null, op, getAddress(local));
-        // first we wait for the future to complete.
-        future.get();
+        future.get(); // first we wait for the future to complete.
```
pveentjer: Checkstyle doesn't like trailing comments; so we don't do them in our regular code. Personally I have no problem with it; perhaps we should make checkstyle lenient towards trailing comments.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
```diff
@@ -36,91 +36,52 @@ public void setup() {
...
-                // we need to set the call-id to prevent running the operation on the calling-thread.
-                setPartitionId(1);
-            }
-
-            @Override
-            public void run() throws Exception {
-                Thread.sleep(5000);
-            }
-        };
+        DummyOperation op = new DummySleepingNullReturningOperation();
```
pveentjer: Now it becomes a bit more difficult to understand the operation. From the name it isn't clear that it is a partition specific operation.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
tombujok: It's a test - so no checkstyle.
In a test I personally prefer this notation to make the test smaller in terms of LOC.
I have no problems in moving it up again.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
```diff
@@ -129,22 +90,19 @@ public void isDone_whenObjectResponse() {
...
     }
 
-    // ========================= andThen ==================================
-
     // There is a bug: https://github.com/hazelcast/hazelcast/issues/5001
     @Test
     public void andThen_whenNullResponse_thenCallbackExecuted() throws ExecutionException, InterruptedException {
         DummyOperation op = new DummyOperation(null);
-
+        final ExecutionCallback callback = mock(ExecutionCallback.class);
```
pveentjer: Does this field need to be final?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
bilalyasar: yes you are right, it is not obligation as you said :+1:

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
tombujok: For me the test is much more unreadable with the inner class. I have to analyse its code reading every single unit-test.
What name do you suggest?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
tombujok: Unfortunately yes since it's used in the inner class. 
If we used some proxying mechanism that e.g. Awaitility provides we could remove most of the inner classes from our tests...

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
```diff
@@ -8,115 +8,120 @@
...
+        OuterOperation outerOperation = new OuterOperation(innerOperation, outerPartitionId);
         InternalCompletableFuture f = operationService.invokeOnPartition(null, outerOperation, outerOperation.getPartitionId());
 
-        try {
-            f.getSafely();
-            fail();
-        } catch (IllegalThreadStateException e) {
-        }
+        expected.expect(IllegalThreadStateException.class);
+        expected.expectMessage("cannot make remote call");
```
pveentjer: I would not test for the message.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
```diff
@@ -8,115 +8,120 @@
...
+        f.getSafely();
     }
 
     @Test
-    public void whenPartition_callsDifferentPartition_butMappedToSameThread() throws ExecutionException, InterruptedException {
+    public void invokeOnPartition_outerLocal_innerSameInstance_callsDifferentPartition_mappedToSameThread() throws ExecutionException, InterruptedException {
```
pveentjer: Maybe it is an idea to split up this test into 2 files: one testing partition specific outer call.. and the other testing non partition specific outer call. Otherwise you get such long names.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
pveentjer: I think we should. We should also update the documentation to clearly explain what this method is doing.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
pveentjer: Only when you have received the runner, you know you are in a nested call.  

So this conclusion:
```
 // if already running an operation for a partition (id >= 0) it means that it's a nested invocation
        // in this case partitionId of both nested operations has to match
```

I'm not following.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
tombujok: I know there are various opinions about it - I wouldn't like to have another dogmatic discussion about it and would prefer to stay pragmatic.
We don't have any error codes, thus it's difficult to assert precisely what you want to assert. 
Checking exception type only is in many cases very imprecise, you can also get:
`cannot make invocation on multiple partitions`
`cannot be run in current thread`

I prefer to be sure what I'm asserting and where the exception is thrown - otherwise the test may not be testing what you are willing to test. I've seen many of them already...

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
pveentjer: Perhaps we should update checkstyle. I have little problems with a trailing comment. Can you suggest it on the dev channel?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
tombujok: There's already `InvocationNestedLocalTest` and `InvocationNestedRemoteTest`
Wouldn't like to split a test to two files because of a name that's descriptive and tells me everything about it. The name can be shorter actually since `callsDifferentPartition` is redundant.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
pveentjer: For me it isn't clear because this DummySleepingNullReturningOperation doesn't explain the intent to me.

pveentjer: Let me have a look at the whole source. Judging changes based on fragements sucks.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
tombujok: Actually I'll remove the comment since it's obvious that the future has completed when get returns.
Feel free to discuss on dev channel whether we should have checkstyle in tests.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
tombujok: Done

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
tombujok: Done. Thx

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
pveentjer: The main purpose of the DummySleeping one is to not be bother with the actual result of the operation. I need to have an invocation and this is the way how to get one. 

So perhaps name it to GetLostOperation :) Something that clearly explains the intent that we don't want to be bother with the actual operation.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
pveentjer: Sure.. Again.. judging code based on fragments sucks

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationNestedLocalTest.java
tombujok: Thx!

pveentjer: I think this is a relatively new unit feature. I recently discovered it while reading 

https://pragprog.com/book/utj2/pragmatic-unit-testing-in-java-8-with-junit

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
tombujok: I love the GetLostOperation :)
How about `DummyPartitionOperation`? or `TestPartitionOperation`?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
tombujok: Oki doki

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
pveentjer: The problem with DummyPartitionOperation/TestPartitionOperation names is that they all give the impression that they matter. In our case, the GetLostOperation is meant to create an invocation and to shoot it so much into the future that it won't bother the actual test. So I don't care about its results.. I just want to get access to an Invocation instance I can test.

tombujok: OK, so I'll change it to `GetLostPartitionOperation` since you asked once if it's partition operation or not.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
```diff
@@ -154,4 +112,17 @@ public void run() throws Exception {
             }
         });
     }
+
+    private static class GetLostPartitionOperation extends DummyOperation {
```
pveentjer: Can you add a few lines of comment explaining the intend of this operation?

metanet:
:+1:

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFutureTest.java
tombujok: Done

tombujok:
We've got two thumbs-up already. 
@pveentjer Are you also ok with it? I'm asking since you deliver significant input here.

PR #6245, opened 2015-09-18
Support recursive class structures for portable objects

henri-tremblay:


sancar:
Hi @henri-tremblay . 
I have take a look at your pr. Can you share a use case scenario ? 
I wrote a test with my understanding of what you are trying to support and test did not pass. Either I got the feature wrong or code is not achieving what it intends to do. 
Here is my test: https://gist.github.com/sancar/262b4dbe6fa5011d08b0

A test from user perspective can help us understand this pr.

henri-tremblay:
My unit test shows the problem. It will fail with the original code. I think you gist is working because it doesn't register a class definition. It is the recursive class definition that causes the problem

sancar:
To prevent misunderstanding, I will retry, my gist did not pass with the changes in this pr. I wanted to repeat because in your last comment you stated `I think you gist is working because .... ` , although probably a typo.

In hazelcast, you do not need to register class definition manually in most of the cases. Do you think in that case user has to create a class definition and register via SerializationConfig.addClassDefinition ?
Is it possible to give same feature without need of registering class definition manually ? 

And in the end , does user have to create a fake class definition to be able to use recursive structures ? 
1) If not , can you share a cleaner use case ?
2) If yes, I don't think we will want to give such a complicated API. We need to find a better solution.

henri-tremblay:
No. It wasn't a typo. But indeed I misread you. So I might have solve only part of the problem. Let me see.

In my case I was indeed requiring to register a class definition. Because I have Map and List attributes if I remember correctly.

The fake is required. I didn't want to modify that part myself to make the API cleaner. The quickest and easiest would be to have a `addPortableXXXField(String fieldName, int factoryId, int classId)`.

henri-tremblay:
I had a quick look and indeed, the ClassDefinitionWriter doesn't support recursive structures either. That's why your test doesn't work.

ClassDefinitionWriter should notice that a ClassDefinition was already registered. However, I won't have the time to find out how to fix this right now. So my fix makes it possible when ClassDefinitions are explicitly registered. So at least, it is possible to have a recursive structure.

I've pushed an improved version that doesn't require a local Set. It's doing a lookup instead

henri-tremblay:
In fact I do remember with I need a ClassDefinition... Because I have optional fields.

pveentjer:
@sancar can you have a look at this PR and decide how to proceed.

sancar:
Hi @henri-tremblay 
I have created an issue for this. We will consider your second suggestion to solve the problem. 

```
The quickest and easiest would be to have a addPortableXXXField(String fieldName, int factoryId, int classId)
```

Closing this one. Thanks for the suggested fix and pull request.

sancar:
Created issue https://github.com/hazelcast/hazelcast/issues/6810

PR #6249, opened 2015-09-18
ReturnCount checkstyle setting adjusted to 8

tombujok:


Donnerbart:
:+1:

I also talked with Peter about this, so we have a thumbs up for the rule change from him.

Thanks a lot for taking care about this!

PR #6251, opened 2015-09-18
Fixed Types in Count Aggregations

jerrinot:
We should not force Supplier to provide raw Object(s).


jerrinot:
@noctarius: can you please have a look?

noctarius:
It actually doesn't make a difference for a count aggregation but yeah as long as everything still compiles I'm fine with that :)

jerrinot:
Type do not make difference for aggregations, but it effectively forced Supplier to supply Objects -> You would have to use different Suppliers for different aggregations on the same map.

noctarius:
Oh yeah ok I see your point! Good catch.

jerrinot:
so is it :+1: from you? :)

noctarius:
sure :D

toktarev:
:+1:

PR #6261, opened 2015-09-19
Added tiny improvement for PartitionWideEntryWithPredicateOperation

ahmetmircik:
Returned true or false immediately if we know the supplied predicate is equal to one of TruePredicate or FalsePredicate.

* `PartitionWideEntryWithPredicateOperation` extends `PartitionWideEntryOperation` and overrides `getPredicate` method


jerrinot:
:+1:

PR #6263, opened 2015-09-21
Fixed logging when Level = OFF

jerrinot:
The OFF level was interpreted as INFO in some cases.
The led to unnecessary logging noise.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/logging/Log4jFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/logging/Log4jFactory.java
@@ -36,16 +36,19 @@ public class Log4jFactory extends LoggerFactorySupport implements LoggerFactory 
 
         public Log4jLogger(Logger logger) {
             this.logger = logger;
-            if (logger.getLevel() == org.apache.log4j.Level.DEBUG) {
-                level = Level.FINEST;
-            } else if (logger.getLevel() == org.apache.log4j.Level.INFO) {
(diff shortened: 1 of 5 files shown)


mdogan:
looks good :+1:

hazelcast/src/main/java/com/hazelcast/logging/Log4jFactory.java
```diff
@@ -36,16 +36,19 @@ protected ILogger createLogger(String name) {
...
+            } else if (log4jLevel == org.apache.log4j.Level.WARN) {
+                this.level = Level.WARNING;
+            } else if (log4jLevel == org.apache.log4j.Level.FATAL) {
+                this.level = Level.SEVERE;
+            } else if (log4jLevel == org.apache.log4j.Level.OFF) {
+                this.level = Level.OFF;
             } else {
-                level = Level.INFO;
+                this.level = Level.INFO;
             }
```
tombujok: Here the default level is INFO, where as in the inverse getLevel() method it's OFF.
I know that there's no OFF level there, but shouldn't it be analogous?

jerrinot: good catch! I'll send an update

tombujok:
:+1: This change looks good 
As discussed with Jaromir - it would need some proper cleanup in the future (e.g. how we handle logger.isTraceEnabled())

PR #6265, opened 2015-09-21
Improved JavaDoc of SharedService interface.

Donnerbart:


hazelcast/src/main/java/com/hazelcast/spi/SharedService.java
```diff
@@ -17,8 +17,11 @@
 package com.hazelcast.spi;
 
 /**
- * A shared service (not clear what the difference is between shared and unshared though).
+ * Marks a service which exposes data (e.g. states) to other services.
```
jerrinot: What if the service exposes behaviour and not data?

Donnerbart: That's a very good point (and the reason I made a PR and not a direct commit) :)

jerrinot:
:+1:

Issue #6264, opened 2015-09-21
[TEST-FAILURE] ClientRegressionWithMockNetworkTest.testDeadlock_WhenDoingOperationFromLifecycleListener_and_NearCache


sancar:
Error Message
```
Can not connected to Address[127.0.0.1]:5904: instance does not exist
```
Stacktrace
```
com.hazelcast.client.HazelcastClientNotActiveException: Can not connected to Address[127.0.0.1]:5904: instance does not exist
	at com.hazelcast.client.spi.impl.ClientInvocation.notifyException(ClientInvocation.java:175)
	at com.hazelcast.client.spi.impl.ClientInvocation.invoke(ClientInvocation.java:133)
	at com.hazelcast.client.spi.impl.ClientListenerServiceImpl.stopListening(ClientListenerServiceImpl.java:96)
	at com.hazelcast.client.proxy.ClientMapProxy.removeNearCacheInvalidationListener(ClientMapProxy.java:1368)
	at com.hazelcast.client.proxy.ClientMapProxy.destroyNearCache(ClientMapProxy.java:1288)
	at com.hazelcast.client.proxy.ClientMapProxy.onShutdown(ClientMapProxy.java:1295)
	at com.hazelcast.client.spi.ProxyManager.destroy(ProxyManager.java:237)
	at com.hazelcast.client.impl.HazelcastClientInstanceImpl.doShutdown(HazelcastClientInstanceImpl.java:533)
	at com.hazelcast.client.impl.LifecycleServiceImpl.shutdown(LifecycleServiceImpl.java:108)
	at com.hazelcast.client.impl.LifecycleServiceImpl.terminate(LifecycleServiceImpl.java:114)
	at com.hazelcast.client.test.TestHazelcastFactory.terminateAll(TestHazelcastFactory.java:126)
	at com.hazelcast.client.ClientRegressionWithMockNetworkTest.cleanup(ClientRegressionWithMockNetworkTest.java:86)
```
Standard Output
```
15:46:04,101  INFO [LifecycleService] hz.client_769_dev.cached-1 - HazelcastClient[hz.client_769_dev][3.6-SNAPSHOT] is CLIENT_CONNECTED
[...]

PR #6268, opened 2015-09-21
Makes all listener remove requests retryable

sancar:
Issue seen in https://github.com/hazelcast/hazelcast/issues/6264
An internal listener remove fails with exception because the node that it tries to run the request on is no longer available. I decided to make client listener remove requests retryable to solve the problem. In hazelcast client, only idempotent operations are retryable. Listener remove request are also idempotent since they run twice they do not get any exception and second call just returns silently, except when they are not. Only thing that is changes when remove request runs twice is its response. If first call is successfull and then we lose the connection, the second retry will return false. This will lead to wrong response to user when request is made retryable. To overcome that, a second change is made so that remove listener is always returns true to user without looking at what is returned from remote. Any response other than exception means listener is removed successfully.

A second small fix in ClientMembershipListener. The call applying 'member remove' to the member list is moved before related connections are destroyed, because when a connection is destroyed it will be retried(if retryable), on a member in memberlist And we don't want to retry on a member that will already know that is dead.

fixes https://github.com/hazelcast/hazelcast/issues/6264

backport of https://github.com/hazelcast/hazelcast/pull/6267

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientListenerServiceImpl.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientListenerServiceImpl.java
@@ -93,8 +93,9 @@ public final class ClientListenerServiceImpl implements ClientListenerService {
                 return false;
             }
             ClientMessage removeRequest = listenerRemoveCodec.encodeRequest(realRegistrationId);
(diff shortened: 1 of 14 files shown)


metanet:
:+1:

mdogan:
:+1:

PR #6271, opened 2015-09-21
ClientInvocation should not sleep when waiting for the next retry attempt (Fixes #6253)

### LINKED ISSUE #6253: [TEST-FAILURE] ClientExecutionPoolSizeLowTest.testOwnerNodeTerminateWithAsyncOperations & testOwnerNodeTerminateWithAsyncOperations
```
java.lang.AssertionError: the size of the map is not correct expected:<1000> but was:<999>
	at com.hazelcast.test.HazelcastTestSupport$7.run(HazelcastTestSupport.java:623)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:737)
	at com.hazelcast.test.HazelcastTestSupport.assertSizeEventually(HazelcastTestSupport.java:620)
	at com.hazelcast.test.HazelcastTestSupport.assertSizeEventually(HazelcastTestSupport.java:616)
	at com.hazelcast.client.io.ClientExecutionPoolSizeLowTest.testOwnerNodeTerminateWithAsyncOperations(ClientExecutionPoolSizeLowTest.java:93)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-nightly/com.hazelcast$hazelcast-client/608/testReport/junit/com.hazelcast.client.io/ClientExecutionPoolSizeLowTest/testOwnerNodeTerminateWithAsyncOperations/

cangencer:
…empt.

ClientExecutionService should be consistent with the use of the internal executor when running scheduled tasks.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
@@ -24,14 +24,15 @@ import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.spi.ClientExecutionService;
 import com.hazelcast.client.spi.ClientInvocationService;
+import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
+import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.LifecycleService;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
 import com.hazelcast.spi.exception.RetryableHazelcastException;
-import com.hazelcast.util.EmptyStatement;
 
 import java.io.IOException;
 import java.util.concurrent.RejectedExecutionException;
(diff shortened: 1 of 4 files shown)


sancar:
We had a problem with this approach. We need to find a solution that fits to both problems. 
https://github.com/hazelcast/hazelcast/pull/4593/files

cangencer:
Updated to use ICompletableFuture instead, so the exception can be set in the invocation.

hazelcast-client-new/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -203,35 +204,46 @@ private boolean handleRetry() {
...
-
         try {
-            sleep();
-            ((ClientExecutionServiceImpl) executionService).executeInternal(this);
+            rescheduleInvocation();
         } catch (RejectedExecutionException e) {
             if (LOGGER.isFinestEnabled()) {
                 LOGGER.finest("Retry could not be scheduled ", e);
             }
-            clientInvocationFuture.setResponse(e);
```
sancar: Did you remove this line by mistake? I think we need to leave this line as is.

cangencer: yes copy paste mistake. Updated now, thanks!

sancar:
:+1:  Looks good. Do we need to create an issue for c# client ? 
No need for c++/ There is no executionService in c++ client.
And also waiting for backport.

cangencer:
@sancar it should not affect the c# client at the moment as there is only one thread pool for all tasks and it is quite large, and not configurable for now.

cangencer:
verify

pveentjer:
Looks go to me, but I'm not an expert on the client side.

PR #6272, opened 2015-09-21
Backport of #6269 - Timeout original operation when no response is received for IsStillExecutingOperation

### LINKED ISSUE #6250: InvocationFuture async calls don't detect lost operations
A sync call (future.get) will call is still running and if returns false, then timeout.

But future using an andThen won't trigger this behavior. So if a future is only used this way and the operation/response got lost, it won't be detected.

metanet:
The problem is as follows: IsStillExecutingOperation is used to timeout a remote invocation by InvocationRegistry or either making the call sync by calling .get() on the future reference. If no response is received for the IsStillExecutingOperation invocation, then the original invocation does not times out.

Fixes #6250


metanet:
Merging this one myself since it is already merged in master.

PR #6275, opened 2015-09-22
Minor test cleanup in opservice.impl

pveentjer:
Mostly test class renaming to make grouping them a lot easier.

Also split of one test into its own testfile.


jerrinot:
:+1: 
good changeset!

pveentjer:
You should see my invocationfuture cleanup branch where this comes from :)

PR #6277, opened 2015-09-23
Rename of FieldUpdater fields

pveentjer:
Instead of having FOO_FIELD_UPDATER, it is now just FOO.

Also switched to static imports for the newUpdater method.


Donnerbart:
:+1: (once green)

PR #6278, opened 2015-09-23
No need to explicitly call recordValues.remove()

jerrinot:
Minor Improvement - no need to explicitly call `recordValues.remove()` as the old value will be replaced anyway

+ value sanitization extracted into its own method.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/IndexImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/IndexImpl.java
@@ -94,14 +94,9 @@ public class IndexImpl implements Index {
         }
 
         Data key = e.getIndexKey();
-        Comparable oldValue = recordValues.remove(key);
         Comparable newValue = e.getAttribute(attribute);
-        if (newValue == null) {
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/query/impl/IndexImpl.java
```diff
@@ -94,14 +94,9 @@ public void saveEntryIndex(QueryableEntry e) throws QueryException {
...
-        Comparable oldValue = recordValues.remove(key);
         Comparable newValue = e.getAttribute(attribute);
-        if (newValue == null) {
-            newValue = NULL;
-        } else if (newValue.getClass().isEnum()) {
-            newValue = TypeConverters.ENUM_CONVERTER.convert(newValue);
-        }
-        recordValues.put(key, newValue);
+        newValue = sanitizeValue(newValue);
+        Comparable oldValue = recordValues.put(key, newValue);
```
jerrinot: we can get the old value here, no need to call `remove()`

pveentjer:
:+1:

Issue #6144, opened 2015-09-07
Extra '\r\n' in body of REST api responses


sreis:
I'm running a single instance of the latest Hazelcast docker build and trying out the REST API. 

So it's `Hazelcast 3.5.2` with `java 1.7.0_79 OpenJDK Runtime Environment (IcedTea 2.5.6) (7u79-2.5.6-1~deb8u1)`

My simple golang client was complaining of `Unsolicited response received on idle HTTP channel starting` so I checked with `curl` and got this `* Excess found in a non pipelined read: excess = 2, size = 3, maxdownload = 3, bytecount = 0`

This is a screenshot of Wireshark showing an extra 2 bytes `\r\n` at the end of the message.
![hazelcast get debug](https://cloud.githubusercontent.com/assets/495294/9717918/b610449a-556e-11e5-8086-68d6237dab70.png)

Steps with `curl`:
`curl -v -X POST http://192.168.99.100:5701/hazelcast/rest/maps/example/foo -d bar`
`curl -v http://192.168.99.100:5701/hazelcast/rest/maps/example/foo`

This does the right thing:
`curl -v http://192.168.99.100:5701/hazelcast/rest/maps/example/doesnotexist`

Donnerbart:
The REST client code sample is broken due to this issue as well. You can use it to easily check if the defect has been fixed:
```
Value for key1 is value1

Exception in thread "main" java.io.StreamCorruptedException: invalid stream header: FFFFFFE9
	at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:806)
	at java.io.ObjectInputStream.<init>(ObjectInputStream.java:299)
	at RestClient.main(RestClient.java:41)
```
The problematic line is:
```
objectInputStream = new ObjectInputStream(byteArrayInputStream);
```
(see https://github.com/hazelcast/hazelcast-code-samples/blob/master/clients/client-rest/src/main/java/RestClient.java#L40)

jerrinot:
it's fixed in both maintenance and master branches.

PR #6285, opened 2015-09-24
Remove extra new line after message body in HTTP response for the REST API

cangencer:
Fixes issue #6144


pveentjer:
Can you add a test for this?

cangencer:
No new tests but I've updated the tests so that they don't use readLine() to read the response. This way they fail with the extra line new line at the end of the reponse body.

Donnerbart:
:+1: Looks good for new, but would appreciate another pair of eyes :)

If we need an explicit test we could use the REST client code from the code samples as base (which is not working due to #6144 at the moment).

cangencer:
I did test it against the sample code, and it works from there as well. Some more useful info here:

http://stackoverflow.com/questions/13821263/should-newline-be-included-in-http-response-content-length/30677511#30677511

pveentjer:
:+1:

Issue #5130, opened 2015-04-21
ClientMapProxy.entrySet() seems to have O(n²) complexity (all clients)


Donnerbart:
While writing tests for my unbounded return values feature I had massive problems with executing a simple `IMap.entrySet()` on the client side with about 150k entries in the map.

So I changed my failing test until I came up with the following code to measure the duration of several `IMap` methods on client side with a different number of server nodes as well as local and remote setup: https://gist.github.com/Donnerbart/ad25717a52b4521da221

These are the results:
![test1a](https://cloud.githubusercontent.com/assets/4196298/7253598/3020bc68-e83e-11e4-86e9-48118b7120b7.png)
![test1b](https://cloud.githubusercontent.com/assets/4196298/7253601/335fd350-e83e-11e4-83da-f2ddd6764081.png)
![test2a](https://cloud.githubusercontent.com/assets/4196298/7253605/364d5e7a-e83e-11e4-87ff-355ca969599c.png)
![test2b](https://cloud.githubusercontent.com/assets/4196298/7253611/3a361cd4-e83e-11e4-85d1-48c5955fa7c7.png)
![test3a](https://cloud.githubusercontent.com/assets/4196298/7253612/3a4e194c-e83e-11e4-940a-bf6f53e40200.png)
![test3b](https://cloud.githubusercontent.com/assets/4196298/7253621/41d48e30-e83e-11e4-9bdb-ffe71e8fb401.png)
![test4a](https://cloud.githubusercontent.com/assets/4196298/7253620/41d42ce2-e83e-11e4-99ed-d519f3a34193.png)
![test4b](https://cloud.githubusercontent.com/assets/4196298/7253623/41d58ab0-e83e-11e4-981a-4b58eadf85f7.png)
![test5a](https://cloud.githubusercontent.com/assets/4196298/7253624/41d573c2-e83e-11e4-9e96-896624dbb65e.png)
![test5b](https://cloud.githubusercontent.com/assets/4196298/7253622/41d52a98-e83e-11e4-982b-28210d0080a9.png)
![test6a](https://cloud.githubusercontent.com/assets/4196298/7253625/41d78720-e83e-11e4-9376-87b1fb22378e.png)
![test6b](https://cloud.githubusercontent.com/assets/4196298/7253626/41ef5a26-e83e-11e4-9bcf-eb1a9743477a.png)

The local setup were client and server nodes on the same machine. The remote setup were client on one and all server nodes on another machine.

[...]

PR #6288, opened 2015-09-25
Performance: keyset/values download too much

pveentjer:
For keyset the values are downloaded so this is a big performance problem. 

This PR is a clean up of the querying mechanism and resolves a bunch of performance problems.

It also resolves a problem with old cliend Map.entryset:
https://github.com/hazelcast/hazelcast/issues/5130

Other problems resolved for the old client is that keySet(predicate),values(predicate) always download both key and value.


hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -875,7 +836,7 @@ public void loadAll(Set<K> keys, boolean replaceExistingValues) {
 
     @Override
     public Set<K> localKeySet() {
-        throw new UnsupportedOperationException("Locality is ambiguous for client!!!");
+        return localKeySet(TruePredicate.INSTANCE);
```
ahmetmircik: Is removing this `UnsupportedOperationException` intentional?

pveentjer: Yes. So that there is a single method you jump to.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResultRow.java
```diff
@@ -25,60 +25,50 @@
 import java.io.IOException;
 
 /**
- *  Multiple result set for Predicates
+ * This interfaces provides methods related to entry of the query result.
```
ahmetmircik: Seems a copy/paste javaDoc error.

ahmetmircik:
:+1:  except one minor comment.

sancar:
This one broke C++ and C# client builds since it changes return type of a request. I am opening related issues to both projects. And adding CLIENT label to this one.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -697,18 +693,6 @@ public void loadAll(Set<K> keys, boolean replaceExistingValues) {
         return dataKeys;
     }
 
-    @Override
-    public Set<K> keySet() {
-        MapKeySetRequest request = new MapKeySetRequest(name);
```
sancar: @pveentjer Why did we changed keySet() , values() and entrySet() that does not use predicates ?

It seems they are replaced by QueryRequest using TruePredicate. I cant do the changes in c++ client since Predicates are not supported there.(because they are not Portable or IdentifiedDataSerializable) . I am not sure what is the situation with c# client.(There was a solution with full java class names, I wonder it works for this one)

hazelcast/src/main/java/com/hazelcast/map/impl/client/AbstractMapQueryRequest.java
```diff
@@ -65,7 +65,7 @@ public AbstractMapQueryRequest(String name, IterationType iterationType) {
 
     @Override
     protected final void invoke() {
-        QueryResultSet result = new QueryResultSet(null, iterationType, true);
+        List<QueryResultRow> result = new ArrayList<QueryResultRow>();
```
sancar: Collection of a IdentifiedDataSerializable/Portable is not supported in other clients. We can't send this as response directly.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
pveentjer: The old requests are still in place.

hazelcast/src/main/java/com/hazelcast/map/impl/client/AbstractMapQueryRequest.java
pveentjer: Ok. This should be not that difficult to resolve.

andrey-yemelyanov:
Hi,

I am seeing this problem in Hazelcast 4.1.1. When calling map.keySet(), values will be deserialized. Is it a feature now? Please advise how I can retrieve just the keys in a certain map.

PR #6293, opened 2015-09-26
Performance: MapEntrySet is not hashmap based anymore

pveentjer:
MapEntrySet is slow because it stores the content in a hashmap and
this makes insertion quite complex due to all the hashing. When the data-structure is send
over the wire, on the other side you pay the price again. Apart from that, it also generates quite a lot of litter.

By switching to a list based approach, since it is unlikely that there are any duplicates, this problem disappears.

For the enterprise PR see:
https://github.com/hazelcast/hazelcast-enterprise/pull/387


hazelcast/src/main/java/com/hazelcast/map/impl/MapEntries.java
```diff
@@ -85,4 +77,28 @@ public int getFactoryId() {
     public int getId() {
         return MapDataSerializerHook.ENTRY_SET;
```
jerrinot: this should be renamed as well

jerrinot:
why can't you use just a regular `ArrayList` instead of the `Entries` classes?

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
```diff
@@ -662,27 +662,27 @@ public Data readBackupData(Data key) {
     }
 
     @Override
-    public MapEntrySet getAll(Set<Data> keys) {
+    public MapEntries getAll(Set<Data> keys) {
         checkIfLoaded();
         final long now = getNow();
 
-        final MapEntrySet mapEntrySet = new MapEntrySet();
```
jerrinot: just a detail: the naming is a bit confusing. you have "entries" and "loadedEntries" in the scope of this method.

hazelcast/src/main/java/com/hazelcast/map/impl/MapEntries.java
```diff
@@ -85,4 +77,28 @@ public int getFactoryId() {
...
+                out.writeData(o.getKey());
+                out.writeData(o.getValue());
+            }
+        }
+    }
+
+    @Override
+    public void readData(ObjectDataInput in) throws IOException {
+        int size = in.readInt();
+        for (int i = 0; i < size; i++) {
```
jerrinot: a small optimization: when `size > X` then you could re-create the array with right capacity - to avoid copying when growing the array-list.

jerrinot:
:+1: with just a small comments. feel free to ignore them as they are not that important. 
I was trying to find a place where this could yield duplicate results, but I couldn't not find any. 

I believe someone else should have a look too; Preferably a person with a solid knowledge of both clients.

pveentjer:
Thanks jaromir. The new client is not effected by this since they don't use the MapEntrysSet in the putAll. I have already contacted them to find a solution

sancar:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
pveentjer: What is your suggestion?

hazelcast/src/main/java/com/hazelcast/map/impl/MapEntries.java
pveentjer: I know. The problem is that the no arg constructor; needed for serialization, is also used for regular calls. And for regular calls, you need to have an initialized collections; hence it is made in the constructor. I'm not very happy with it, but don't see a clear way out which some ugly code. What is your suggestion.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
jerrinot: actually, the scope of `loadedEntries` is extremely small. It's fine as it is now.

hazelcast/src/main/java/com/hazelcast/map/impl/MapEntries.java
jerrinot: it is a pain. I'd do something like this:
```java
private static final int ARRAY_RECREATION_THRESHOLD = 200;
[...]
public void readData() {
  [...]
  if (size > ARRAY_RECREATION_THRESHOLD) {
    assert entries.isEmpty()
    entries = new ArrayList<>(size);
  }
  [...]
}
```

It's ugly, the threshold might seems artificial, but I *believe* it will work better. The ugly code is isolated. Imagine you have extremely large result site. with 10,000's of entries.

pveentjer: I have a better approach. The list is created lazily.

PR #6298, opened 2015-09-28
Minor JCache Improvements

serkan-ozal:
- Ignores checking completion latch if its ignored in cache proxies
- Gets existing cache record for backup record on put without extra record map access


ahmetmircik:
:+1:

metanet:
:+1:

PR #6299, opened 2015-09-28
native client test for hibernate

mustafaiman:


hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/NativeClientTest.java
```diff
@@ -0,0 +1,81 @@
...
+@Category({QuickTest.class, ParallelTest.class})
+public class NativeClientTest
+        extends HibernateSlowTestSupport {
+
+    protected SessionFactory clientSf;
+
+    @Override
+    protected Properties getCacheProperties() {
+        Properties props = new Properties();
+        props.setProperty(Environment.CACHE_REGION_FACTORY, HazelcastCacheRegionFactory.class.getName());
```
mesutcelik: do we need to clear this inside `@After`?

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/NativeClientTest.java
```diff
@@ -0,0 +1,81 @@
...
+        session = clientSf.openSession();
+        DummyEntity retrieved = (DummyEntity) session.get(DummyEntity.class, (long)1);
+        assertEquals("dummy:1", retrieved.getName());
+    }
+
+    @After
+    public void tearDown() {
+        if(clientSf !=null) {
+            clientSf.close();
+            clientSf = null;
```
mesutcelik: I don't see any reason to set it to `null`.

mesutcelik:
two minor comments :+1:

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/NativeClientTest.java
mustafaiman: this is not an environment variable. it is passed to hibernate on creation. it is gone once hibernate's session factory is closed

kobalski:
:+1:  Merging, but will change these tests as slow test with direct commit, because hibernate tests which does not use mock network must be slow test.

PR #6300, opened 2015-09-28
Renamed EventPacket to EventEnvelope

pveentjer:
EventPacket is not a correct name since it isn't a type of a Packet.
Envelope is more clear since it is placed around an arbitrary event
object and some meta-data is added.


mtopolnik:
:+1:

Donnerbart:
:+1: (if we have consensus on the name change, the rest of the cleanup looks good)

toktarev:
:+1:

pveentjer:
verify

PR #6301, opened 2015-09-28
Don't convert backup operation to `Data` before serializing `Backup` if there is only one backup for preventing unnecessary temporary memory allocation and copy

serkan-ozal:
By default, Hazelcast uses only one backup.
If there is only one backup, backup operation is sent to only one backup node.
If backup operation is converted to `Data`, there will be these operations as below:
* a temporary memory allocation (`byte[]`) for `Data`
* serialize backup operation to allocated memory
* copy the temporary allocated memory (backup operation data) to output while serializing `Backup`

In this flow, there are two redundant operations (allocating temporary memory and copying it to output). So in this case (there is only one backup), we don't convert backup operation to `Data` as temporary before `Backup` is serialized but backup operation is already serialized directly into output without any unnecessary memory allocation and copy when it is used as object inside `Backup`.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationBackupHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationBackupHandler.java
@@ -149,14 +149,57 @@ final class OperationBackupHandler {
 
     private int makeBackups(BackupAwareOperation backupAwareOp, int partitionId, long[] replicaVersions,
                             int syncBackups, int asyncBackups) {
-        int sendSyncBackups = 0;
+        int sendSyncBackups;
+        int totalBackups = syncBackups + asyncBackups;
 
         InternalPartitionService partitionService = node.getPartitionService();
         InternalPartition partition = partitionService.getPartition(partitionId);
 
-        Data backupOpData = getBackupOperationData(backupAwareOp);
+        if (totalBackups == 1) {
+            sendSyncBackups = sendSingleBackup(backupAwareOp, partition, replicaVersions, syncBackups);
+        } else {
+            sendSyncBackups = sendMultipleBackups(backupAwareOp, partition, replicaVersions, syncBackups, totalBackups);
+        }
+        return sendSyncBackups;
+    }
 
-        for (int replicaIndex = 1; replicaIndex <= syncBackups + asyncBackups; replicaIndex++) {
+    private int sendSingleBackup(BackupAwareOperation backupAwareOp, InternalPartition partition,
+                                 long[] replicaVersions, int syncBackups) {
+        // Since there is only one replica, replica index is `1`
+        Address target = partition.getReplicaAddress(1);
+        if (target != null) {
+            // Since there is only one backup, backup operation is sent to only one node.
+            // If backup operation is converted to `Data`, there will be these operations as below:
+            //      - a temporary memory allocation (byte[]) for `Data`
+            //      - serialize backup operation to allocated memory
+            //      - copy the temporary allocated memory (backup operation data) to output while serializing `Backup`
+            // In this flow, there are two redundant operations (allocating temporary memory and copying it to output).
+            // So in this case (there is only one backup), we don't convert backup operation to `Data` as temporary
+            // before `Backup` is serialized but backup operation is already serialized directly into output
+            // without any unnecessary memory allocation and copy when it is used as object inside `Backup`.
+            Operation backupOp = getBackupOperation(backupAwareOp);
+
+            assertNoBackupOnPrimaryMember(partition, target);
+
+            boolean isSyncBackup = syncBackups == 1;
+
+            Backup backup = newBackup(backupAwareOp, backupOp, replicaVersions, 1, isSyncBackup);
+            operationService.send(backup, target);
+
+            if (isSyncBackup) {
+                return 1;
+            }
+        }
+        return 0;
+    }
+
+    private int sendMultipleBackups(BackupAwareOperation backupAwareOp, InternalPartition partition,
+                                    long[] replicaVersions, int syncBackups, int totalBackups) {
+        int sendSyncBackups = 0;
+        Operation backupOp = getBackupOperation(backupAwareOp);
+        Data backupOpData = nodeEngine.getSerializationService().toData(backupOp);
+
+        for (int replicaIndex = 1; replicaIndex <= totalBackups; replicaIndex++) {
             Address target = partition.getReplicaAddress(replicaIndex);
 
             if (target == null) {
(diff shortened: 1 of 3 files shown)


serkan-ozal:
verify

serkan-ozal:
verify

serkan-ozal:
verify

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationBackupHandler.java
```diff
@@ -150,35 +150,65 @@ private int requestedTotalBackups(BackupAwareOperation op) {
     private int makeBackups(BackupAwareOperation backupAwareOp, int partitionId, long[] replicaVersions,
                             int syncBackups, int asyncBackups) {
         int sendSyncBackups = 0;
+        int totalBackups = syncBackups + asyncBackups;
 
         InternalPartitionService partitionService = node.getPartitionService();
         InternalPartition partition = partitionService.getPartition(partitionId);
 
-        Data backupOpData = getBackupOperationData(backupAwareOp);
+        if (totalBackups == 1) {
```
jerrinot: I'd split this if/else branch into 2 methods.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationBackupHandler.java
```diff
@@ -188,14 +218,21 @@ private Data getBackupOperationData(BackupAwareOperation backupAwareOp) {
...
 
-    private Backup newBackup(BackupAwareOperation backupAwareOp, Data backupOpData, long[] replicaVersions,
+    private Backup newBackup(BackupAwareOperation backupAwareOp, Object backupOp, long[] replicaVersions,
             int replicaIndex, boolean respondBack) {
-
         Operation op = (Operation) backupAwareOp;
-        Backup backup = new Backup(backupOpData, op.getCallerAddress(), replicaVersions, respondBack);
+        Backup backup;
+        if (backupOp instanceof Operation) {
+            backup = new Backup((Operation) backupOp, op.getCallerAddress(), replicaVersions, respondBack);
```
jerrinot: just a thought: is this if/else needed here? Could it be better to move it to `Backup` class?

serkan-ozal: You mean, constructor of `Backup` takes `Object` type in a single constructor instead of two different constructor? Then it decides which one will be used, `Data` or `Operation`

jerrinot: yeah, this was the idea. probably a bad idea, forget it. 
I have to say I don't like how the code is cluttered with if/else branches. so I was thinking how to improve it. but this wouldn't solve anything.

serkan-ozal: In fact, I don't like having two fields for same purpose inside `Backup` and only one of them is used. 
Another solution I had though was using two `Backup` implementation as `Data` based and `Operation` based for backup operation but it seemed to me more complex than this one.

jerrinot:
could you write a test for both branches? I think an integration test is fine.

serkan-ozal:
ok, I will think about how can I test it

jerrinot:
an integration test should be easy, right? you just need to start 3 members and set backup count to `2` or `1`.

it's probably more valuable than just a unit test in this case.

serkan-ozal:
I mean, verifying no `Data` convertion for backup operation is done when backup count is 1? Are you saying no need to verify this?

jerrinot:
well, you should verify the backups have been made?

jerrinot:
in a nutshell: you have created a special case. hence the test should verify:
1. the backup mechanism works in this new special case
2. there is no regression in the non-special case.

serkan-ozal:
Yes, with this test, we verify this PR doesn't break any part of backup mechanism. But in fact, I was thinking of verifying it really prevents redundant convertion.

sancar:
Looks good :+1: 
Personal comment: 
We are adding special cases to code base for performance. Normally we all hate to see this kind of code but it seems we just need to get used to it.

serkan-ozal:
As I said, another solution I had thought was using two `Backup` implementation as `Data` based and `Operation` based for backup operation but it seemed to me more complex than this one.

metanet:
:+1:

hazelcast/src/test/java/com/hazelcast/cache/CacheBackupTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.junit.Test;
+
+import javax.cache.Cache;
+import javax.cache.CacheManager;
+import javax.cache.spi.CachingProvider;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class CacheBackupTest extends HazelcastTestSupport {
```
ahmetmircik: test-runner? category?

serkan-ozal: ah, ok. good catch

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationBackupHandler.java
```diff
@@ -188,14 +231,21 @@ private Data getBackupOperationData(BackupAwareOperation backupAwareOp) {
...
     }
 
-    private Backup newBackup(BackupAwareOperation backupAwareOp, Data backupOpData, long[] replicaVersions,
+    private Backup newBackup(BackupAwareOperation backupAwareOp, Object backupOp, long[] replicaVersions,
             int replicaIndex, boolean respondBack) {
-
         Operation op = (Operation) backupAwareOp;
-        Backup backup = new Backup(backupOpData, op.getCallerAddress(), replicaVersions, respondBack);
+        Backup backup;
+        if (backupOp instanceof Operation) {
```
ahmetmircik: hard to read these namings `backup` and `backupOp` what about `backupWrapper` instead of `backup` or some other name more clear?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
```diff
@@ -43,25 +43,36 @@
 
 public final class Backup extends Operation implements BackupOperation, IdentifiedDataSerializable {
 
-    private Data backupOpData;
     private Address originalCaller;
     private long[] replicaVersions;
     private boolean sync;
 
     private Operation backupOp;
+    private Data backupOpData;
```
ahmetmircik: do we need to prepend names with `backup`, like `backupOpData` or `backupOp`, they can be understood from context.

ahmetmircik:
with minor comments and seems mergeable once green :+1:

jerrinot:
:+1: (on green and the test category added)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
serkan-ozal: The old name was `backupOpData` so I added new field as `backupOp` to be consistent with the old name. I didn't change the existing names.

serkan-ozal:
run-parallel

PR #6302, opened 2015-09-28
Removed RecordStore#put(Map.Entry<Data, Object> entry) method.

ahmetmircik:
Instead of it RecordStore#put(key,value,ttl) should be used, they are doing same job.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
@@ -21,19 +21,19 @@ import com.hazelcast.config.MapConfig;
 import com.hazelcast.core.EntryEventType;
 import com.hazelcast.core.EntryView;
 import com.hazelcast.core.ManagedContext;
+import com.hazelcast.internal.serialization.SerializationService;
 import com.hazelcast.map.EntryBackupProcessor;
 import com.hazelcast.map.EntryProcessor;
 import com.hazelcast.map.impl.LazyMapEntry;
 import com.hazelcast.map.impl.LocalMapStatsProvider;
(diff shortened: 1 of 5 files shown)


ahmetmircik:
verify-parallel

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
```diff
@@ -292,6 +292,8 @@
      **/
     void loadAll(boolean replaceExistingValues);
 
-    /** Performs initial loading from a MapLoader if it has not been done before  **/
+    /**
+     * Performs initial loading from a MapLoader if it has not been done before
+     **/
     void maybeDoInitialLoad();
```
pveentjer: Although not part of this PR; perhaps rename this method to 'ensureLoaded' ?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
```diff
@@ -211,8 +211,8 @@ private EntryEventType pickEventTypeOrNull(Map.Entry entry) {
         return null;
     }
 
-    private void put(Object value) {
-        recordStore.put(new AbstractMap.SimpleImmutableEntry<Data, Object>(dataKey, value));
+    private void put(Data dataKey, Object value) {
```
pveentjer: Why not call it key? From the context we know it is Data.

pveentjer:
Minor comments. So :+1:

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
ahmetmircik: done.

PR #6304, opened 2015-09-28
Reduce index memory usage

ajermakovics:
A few optimisations to reduce index memory usage:
- Avoiding storing value in serialized and unserialized form inside QueryEntry. Now only storing value in original form.
- Not storing old value in the index anymore (IndexImpl.recordValues). Instead fetching it from Record itself.


jerrinot:
@ajermakovics: We talked about a switch to re-enable deserialized value caching in `QueryableEntry`. I can't see it in this PR - is it because you simply didn't have enough time or is there any fundamental issue with that?

jerrinot:
I'm closing this in favour of #6374

PR #6308, opened 2015-09-29
Added EntryExpiredListener in order to listen expired map entries.

ahmetmircik:
closes https://github.com/hazelcast/hazelcast/issues/6311

- Created an `EntryExpiredListener` which extends `MapListener`. 
- Kept current `EntryEvictedListener` notification mechanism as is; means it will be receiving both expiration and eviction events as in now. If one implements both `EntryExpiredListener` and `EntryEvictedListener` interfaces, there is a probability that more than one event per entry at the same time may be seen. So you may receive both EVICTED and EXPIRED events for the same entry. This is because, eviction may also sweep expired entries. This may be appear when both size-based-eviction and expiration are configured for the same `IMap`.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapListenerAdaptors.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapListenerAdaptors.java
@@ -22,6 +22,7 @@ import com.hazelcast.core.IMapEvent;
 import com.hazelcast.core.MapEvent;
 import com.hazelcast.map.listener.EntryAddedListener;
 import com.hazelcast.map.listener.EntryEvictedListener;
+import com.hazelcast.map.listener.EntryExpiredListener;
 import com.hazelcast.map.listener.EntryMergedListener;
 import com.hazelcast.map.listener.EntryRemovedListener;
 import com.hazelcast.map.listener.EntryUpdatedListener;
@@ -192,6 +193,26 @@ public final class MapListenerAdaptors {
 
 
     /**
+     * Converts an {@link EntryExpiredListener} to a {@link com.hazelcast.map.impl.ListenerAdapter}.
+     */
+    private static final ConstructorFunction<MapListener, ListenerAdapter> ENTRY_EXPIRED_LISTENER_ADAPTER_CONSTRUCTOR =
+            new ConstructorFunction<MapListener, ListenerAdapter>() {
+                @Override
+                public ListenerAdapter createNew(MapListener mapListener) {
+                    if (!(mapListener instanceof EntryExpiredListener)) {
+                        return null;
+                    }
+                    final EntryExpiredListener listener = (EntryExpiredListener) mapListener;
+                    return new ListenerAdapter() {
+                        @Override
+                        public void onEvent(IMapEvent event) {
+                            listener.entryExpired((EntryEvent) event);
+                        }
+                    };
+                }
+            };
+
+    /**
      * Register all {@link com.hazelcast.map.impl.ListenerAdapter} constructors
      * according to {@link com.hazelcast.core.EntryEventType}s.
      */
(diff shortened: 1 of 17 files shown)


eminn:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictionOperator.java
```diff
@@ -113,6 +109,30 @@ public void removeEvictableRecords(RecordStore recordStore, int evictableSize, M
...
+            return false;
+        }
+
+        if (!backup) {
+            mapServiceContext.interceptAfterRemove(mapName, value);
+
+            boolean expired = recordStore.isExpired(record, now, false);
+            if (expired) {
+                // Fire EXPIRED event also because there is a possibility that
+                // evicted entry may be also an expired one.
```
jerrinot: is this important? what if the entry is not expired now, but it's gonna be expired in 1s. then we are not firing the expire event anyway.

jerrinot: resolved. from a discussion:
```
given that we have to maintain compatibility with the old eviction listener then we need to fire eviction event for expired entries. and because the entry is expired then we have to fire the expire event too.

later on we can add another listener which will be notified for (size-based) evictions only.
```

hazelcast/src/test/java/com/hazelcast/map/ExpirationListenerTest.java
```diff
@@ -0,0 +1,146 @@
...
+        CountDownLatch expirationEventArrivalCount = new CountDownLatch(numberOfPutOperations);
+
+        map.addEntryListener(new ExpirationListener(expirationEventArrivalCount), true);
+
+        for (int i = 0; i < numberOfPutOperations; i++) {
+            map.put(i, i, 100, TimeUnit.MILLISECONDS);
+        }
+
+        // wait expiration of entries.
+        sleepSeconds(1);
```
jerrinot: do we have to wait 1s? isn't eg. 200ms enough?

hazelcast/src/main/java/com/hazelcast/map/impl/MapListenerAdaptors.java
```diff
@@ -200,9 +221,10 @@ public void onEvent(IMapEvent event) {
         CONSTRUCTORS.put(EntryEventType.REMOVED, ENTRY_REMOVED_LISTENER_ADAPTER_CONSTRUCTOR);
         CONSTRUCTORS.put(EntryEventType.EVICTED, ENTRY_EVICTED_LISTENER_ADAPTER_CONSTRUCTOR);
         CONSTRUCTORS.put(EntryEventType.UPDATED, ENTRY_UPDATED_LISTENER_ADAPTER_CONSTRUCTOR);
+        CONSTRUCTORS.put(EntryEventType.MERGED, ENTRY_MERGED_LISTENER_ADAPTER_CONSTRUCTOR);
```
jerrinot: is this something forgotten? Do we have a test for this?

hazelcast/src/test/java/com/hazelcast/map/ExpirationListenerTest.java
Donnerbart: Could we get rid of the sleep at all? Maybe with assertTrueEventually? With the upcoming parallel running tests I worry about more and more racy tests otherwise.

jerrinot: @Donnerbart: This is tricky, as the logic the test is exercising is actually time-based.

Donnerbart: Yeah, I just saw the whole test context and actually `assertOpenEventually` is already used later on. Maybe a usage of `sleepAtLeastMillis()` with a smaller delay would be a good solution then.

ahmetmircik: done.

hazelcast/src/main/java/com/hazelcast/map/impl/MapListenerAdaptors.java
ahmetmircik: asking about `EntryEventType.MERGED`?

jerrinot: indeed

ahmetmircik: it is tested in `WanReplicationTest` class.

Serdaro:
This should have a documentation impact right?

hazelcast/src/main/java/com/hazelcast/map/impl/MapListenerAdaptors.java
jerrinot: oh, I missed you just moved it. I thought you added a new constructor

jerrinot:
:+1:

ahmetmircik:
@Serdaro correct, we need to document this.

PR #6310, opened 2015-09-29
Added EntryExpiredListener in order to listen expired map entries.

ahmetmircik:
closes https://github.com/hazelcast/hazelcast/issues/6311

backport of https://github.com/hazelcast/hazelcast/pull/6308

- Created an `EntryExpiredListener` which extends `MapListener`. 
- Kept current `EntryEvictedListener` notification mechanism as is; means it will be receiving both expiration and eviction events as in now. If one implements both `EntryExpiredListener` and `EntryEvictedListener` interfaces, there is a probability that more than one event per entry at the same time may be seen. So you may receive both EVICTED and EXPIRED events for the same entry. This is because, eviction may also sweep expired entries. This may be appear when both size-based-eviction and expiration are configured for the same `IMap`.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -1379,6 +1379,7 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
                 case REMOVED:
                 case UPDATED:
                 case EVICTED:
+                case EXPIRED:
                 case MERGED:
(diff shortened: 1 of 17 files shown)


ahmetmircik:
This PR seems not applicable to maintenance. If we merge this PR, in a cluster which has both 3.5.3 and 3.5.2 nodes, 3.5.2 nodes will throw lots of NPEs due to the received EXPIRED events. The code part that throws NPE is [here](https://github.com/hazelcast/hazelcast/blob/v3.5.2/hazelcast/src/main/java/com/hazelcast/map/impl/InternalMapListenerAdapter.java#L46).

Closing this PR.

PR #6312, opened 2015-09-29
Combined put, set and tryPut methods of RecordStore impl.

ahmetmircik:


hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
```diff
@@ -57,6 +57,9 @@
 
     boolean tryPut(Data dataKey, Object value, long ttl);
 
+    /**
+     * Returns {@code true} if key doesn't exist previously, otherwise returns {@code false}.
```
pveentjer: Can you add more documentation than just the return value? especially what the purpose is of this method,

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
```diff
@@ -749,7 +753,7 @@ public Object put(Data key, Object value, long ttl) {
         Record record = getRecordOrNull(key, now, false);
         Object oldValue;
         if (record == null) {
-            oldValue = mapDataStore.load(key);
+            oldValue = loadFromStore ? mapDataStore.load(key) : null;
```
pveentjer: Minor change at the end of the method. Can you move the duplication of ' saveIndex(record);' till after the if/else branch instead of repeating it in each branch?

pveentjer:
Minor comments. Simplifications always are a good thing to have.  So :+1:

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
ahmetmircik: done.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
ahmetmircik: done.

PR #6313, opened 2015-09-29
Necessary changes for Enterprise Wan Replication Screens

muratayan:


bilalyasar:
can you add some tests?

hazelcast/src/main/java/com/hazelcast/monitor/LocalPublisherStats.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package com.hazelcast.monitor;
+
+import com.hazelcast.internal.management.JsonSerializable;
+
+public interface LocalPublisherStats extends JsonSerializable {
```
pveentjer: Can you rename it to LocalWanPublisherStats or something? Publisher is such a generic term.

Also this interface needs to be documented since it is public API; the same goes for the other changes you make in the com.hazelcast.monitor package.

hazelcast/src/main/java/com/hazelcast/wan/impl/WanReplicationServiceImpl.java
```diff
@@ -163,4 +164,9 @@ public void shutdown() {
     private ConcurrentHashMap<String, WanReplicationPublisherDelegate> initializeWanReplicationPublisherMapping() {
         return new ConcurrentHashMap<String, WanReplicationPublisherDelegate>(2);
     }
+
+    @Override
+    public Map<String, LocalWanStats> getStats() {
```
pveentjer: How can it be that an interface needs to be implemented, but the actual implementation returns null?

Is additional work required? If so, please add a comment.

And returning null if there is nothing to do is often a PITA since it forces the caller to do null checks. Often it is best to return an empty map (see EMPTY_MAP from Collections). This way you get less branching on the calling side.

muratayan:
Updated added tests

PR #6319, opened 2015-09-30
added locale default

bilalyasar:


hazelcast-client-new/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -132,6 +133,7 @@ protected Document parse(InputStream inputStream) throws Exception {
         DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
         dbf.setNamespaceAware(true);
         final DocumentBuilder builder = dbf.newDocumentBuilder();
+        Locale.setDefault(Locale.getDefault());
```
pveentjer: Changing the default Locale for the whole JVM doesn't seem like a very friendly solution for anyone using Hazelcast.

pveentjer:
I think we should not merge this PR since it change behavior for the whole JVM.

hazelcast-client-new/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
mdogan: I'm not sure I understand what's done here. but it looks like doing `this.x = this.x`. It's setting default locale to its current value again.

mesutcelik: I think JVM is already using default locale if locale not provided.. check [toUpperCase](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#toUpperCase())

Instead of this PR, we should add a new complete feature like overriding default locale if there is such need...

serkan-ozal:
`AbstractXmlConfigBuilder::cleanNodeName` should take care of locale while converting to lowercase instead of this PR.

bilalyasar:
i will send another pr according to feedbacks. thanks all. closing this one.

PR #6320, opened 2015-09-30
refactored document builder factory

bilalyasar:
XML parser should be configured so that it does not allow external entities as part of an incoming XML document. so i disabled the doctype declaration which is enough according to my researches.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
@@ -131,6 +131,7 @@ public class XmlClientConfigBuilder extends AbstractConfigBuilder {
     protected Document parse(InputStream inputStream) throws Exception {
         DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
         dbf.setNamespaceAware(true);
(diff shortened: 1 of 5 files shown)


mesutcelik:
@bilalyasar , can you please add description for the PR?

bilalyasar:
added description

jerrinot:
@bilalyasar: Is it possible to write a test for this?

bilalyasar:
@jerrinot , added test.. thanks.

hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
```diff
@@ -131,6 +131,7 @@ public XmlClientConfigBuilder() {
     protected Document parse(InputStream inputStream) throws Exception {
         DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
         dbf.setNamespaceAware(true);
+        dbf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
```
jerrinot: I believe this is a Xerces 2 -specific feature. Other parsers will probably ignore it. 
Anyway, we accept configuration input from trusted sources only. So I think this is ok.

jerrinot:
:+1: once green

bilalyasar:
it is green now.

mesutcelik:
:+1:

PR #6321, opened 2015-09-30
Additional tests for map values/keyset/entryset methods

pveentjer:


hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResultCollection.java
```diff
@@ -44,6 +44,11 @@ public QueryResultCollection(SerializationService serializationService, Iteratio
         }
     }
 
+    // just for testing
+    public Collection<QueryResultRow> getRows() {
```
jerrinot: does it have to be public?

hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
```diff
@@ -631,6 +631,13 @@ public static void assertIterableEquals(Iterable iterable, Object... values) {
         assertEquals(expected, actual);
     }
 
+    public static void assertSetEquals(Set actual, Object... values) {
```
jerrinot: this is a bit unusual. Not really intuitive, to me anyway. Check `TestCollectionUtils`. You can use this:
```java
assertEquals(expected, setOf("1", "2", "3"));
```

This seems cleaner to me, it express its intend better.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResultCollection.java
pveentjer: Its internal anyway.

jerrinot: yeah, I appreciate it's not a biggie; but still no reason to expose more than is necessary.

hazelcast/src/test/java/com/hazelcast/map/impl/query/MapKeySetTest.java
```diff
@@ -0,0 +1,121 @@
...
+        public boolean apply(Map.Entry<String, String> mapEntry) {
+            return mapEntry.getValue().startsWith("good");
+        }
+    }
+
+//    /**
+//     * Test for issue 711
+//     */
+//    @Test(timeout = 1000 * 60)
+//    public void testPredicateWithEntryKeyObject() {
```
jerrinot: I assume this is not your code. Do you know why it's commented out?

hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
pveentjer: Good point. 

We need to clean up the testing because the Asserts are in multiple classes.

hazelcast/src/test/java/com/hazelcast/map/impl/query/MapKeySetTest.java
pveentjer: I have removed it since this test has remained in the BasicQueryTest. I didn't have the stamina to clean that thing up.

jerrinot:
:+1: on green.
Green Addition!

PR #6324, opened 2015-09-30
converted to lowerCaseInternal

bilalyasar:


serkan-ozal:
:+1:

jerrinot:
:+1: looks good to me.

A note future generations: This is needed because some locales have odd;-) rules when it comes to conversion to lower case. Like the turkish "I" with and without dot. 

@bilalyasar: Can you confirm it? Are there any other places where we should fix this?

bilalyasar:
@jerrinot , i couldn't find any other places.

PR #6325, opened 2015-09-30
Fixed DelegatingFuture Issue

stefanLeo:
Delegated Future .get() not able to beeing called more than once.

See additional JUnit Test -which fails with current master-branch.

In our application we need to abort a remote execution, but before we do so we grant the task another second or so to finish. Unfortunately one cannot call task.get(1, TimeUnit.Second) if another Thread has already called task.get(). The 2nd call needs to wait until the 1st call has finished not throwing a TimeoutException as expected. 
This works however for ordinary Futures like Guavas ListenableFutureTask.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/executor/DelegatingFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/executor/DelegatingFuture.java
@@ -61,23 +61,15 @@ public class DelegatingFuture<V> implements ICompletableFuture<V> {
         }
     }
 
-
     @SuppressFBWarnings("IS2_INCONSISTENT_SYNC")
     @Override
-    public final V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-        if (!done) {
-            synchronized (mutex) {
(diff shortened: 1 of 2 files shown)


pveentjer:
I see the bug in the code. The thread that called get should not modify the state of the future.

pveentjer:
Hmmm.. forget my previous comment.

stefanLeo:
Hi 

@pveentjer : do you need any further information?

The code I provided was just to prove my thoughts on the issue and provide an according unit test. 
So if you are seeking for another solution feel free to just skip my pull request! 
If you would like me to create a Bug report for it pls. let me know.

Thank you for your support.

noctarius:
Hey @stefanLeo as you can see @pveentjer is working on another form to solve this issue. Are you ok we close your PR for the sake of removing DelegatingFuture completely? :)

stefanLeo:
@noctarius :+1:  Fine with me

pveentjer:
Closing this PR. For the fix see:
https://github.com/hazelcast/hazelcast/pull/8925

PR #6332, opened 2015-09-30
MapQueryEngine cleanup

pveentjer:
Instead of injecting a collection to store the result, the QueryEngine returns a QueryResult and it is up to the caller to copy it in the right collection e.g. set. This is needed to deal with enterprise + is cleaner.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResultCollection.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResultCollection.java
@@ -44,6 +44,14 @@ public class QueryResultCollection<E> extends AbstractSet<E> {
         }
     }
 
+    public QueryResultCollection(SerializationService serializationService, IterationType iterationType, boolean binary,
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResultCollection.java
```diff
@@ -44,6 +44,14 @@ public QueryResultCollection(SerializationService serializationService, Iteratio
         }
     }
 
+    public QueryResultCollection(SerializationService serializationService, IterationType iterationType, boolean binary,
+                                 boolean unique, QueryResult queryResult) {
+
+        this(serializationService, iterationType, binary, unique);
+        allAllRows(queryResult.getRows());
```
ahmetmircik: `allAllRows` ?

ahmetmircik:
with minor comment :+1:  (once green)

pveentjer:
I'll change the name in a direct push. Thanks.

PR #6333, opened 2015-09-30
Added LocalKeySet test

pveentjer:


hazelcast/src/test/java/com/hazelcast/map/impl/query/LocalKeySetTest.java
```diff
@@ -0,0 +1,121 @@
...
+        map.put(remoteKey2, "bad");
+        map.put(remoteKey3, "good4");
+
+        Set<String> result = map.localKeySet(new GoodPredicate());
+
+        assertEquals(setOf(localKey1, localKey3), result);
+    }
+
+    @Test
+    public void testResultType() {
```
jerrinot: what is this test good for? it the `QueryResultCollection` part of the contract? or do we make any assumption about it?

pveentjer: We  need to do certain verification that it contains exactly the right type of info. e.g. if you are asking for keys and it also contains values; I want to know about that.

tombujok:
:+1:

hazelcast/src/test/java/com/hazelcast/map/impl/query/LocalKeySetTest.java
jerrinot: I checked the content of `QueryResultCollection` and it makes sense now.

jerrinot:
:+1:

hazelcast/src/test/java/com/hazelcast/map/impl/query/LocalKeySetTest.java
pveentjer: Since there is no more need to send more information than needed, perhaps we can use the same approach as client. On the client the whole collection is fully deserialized and returned. So you get a regular collection with real objects. Instead of this intercepted thing.

pveentjer: @jerrinot 

One of the problems with the QueryResultCollection.. when it is being used as a set.. that methods like contains.. rely on a scan over the items. So complexity will be O(N) instead of O(1). This was already an issue with the QueryResultSet btw.

PR #6334, opened 2015-09-30
QueryResult implements IdentifiedDataSerializable

pveentjer:
QueryResult implements IdentifiedDataSerializable instead of DataSerializable


jerrinot:
:+1:

PR #6337, opened 2015-10-01
Send MigrationOperation even when there's no migration task

mdogan:
MigrationOperation should be sent even when none of the services
provides a migration task to be executed on destination. This is
required to commit or rollback migration on destination when migration
completes. Otherwise, MigrationAwareService.commitMigration() is not
called on destination when no data is migrated.


jerrinot:
:+1: 
@Donnerbart, @mdogan: Great finding!

Donnerbart:
:+1:

Looks good to me. I'll also give this a try on my reproduction scenario later :)

Donnerbart:
run-parallel

Donnerbart:
Fix seems to work fine, couldn't reproduce the wrong state of `ownedPartitions` a single time.

Donnerbart:
verify

PR #6338, opened 2015-10-01
Send MigrationOperation even when there's no migration task

mdogan:
MigrationOperation should be sent even when none of the services
provides a migration task to be executed on destination. This is
required to commit or rollback migration on destination when migration
completes. Otherwise, MigrationAwareService.commitMigration() is not
called on destination when no data is migrated.

(cherry picked from commit 2f450db8684cc8897064164064b39d7f43d94cf6)

backport of #6337


jerrinot:
:+1:

Donnerbart:
:+1:

PR #6342, opened 2015-10-01
Fixes security issue on old client

pveentjer:
The security got the true predicate for method without explicit predicate like values().
This has been solved by using null as special true predicate. This is a local issue only
in the proxy and request.


pveentjer:
verify

sancar:
:+1:

pveentjer:
@sancar this pr contains your suggested null checking:

https://github.com/hazelcast/hazelcast/pull/6345

PR #6331, opened 2015-09-30
Cache stats improvement

serkan-ozal:
Added 
* `long getCreationTime()`
* `long getLastAccessTime()`
* `long getLastUpdateTime()` 
* `long getOwnedEntryCount()` 
methods to cache stats API

Also fixes https://github.com/hazelcast/hazelcast/issues/6314

PR #6343, opened 2015-10-01
[BACKPORT] Cache stats improvement 

serkan-ozal:
Backport of https://github.com/hazelcast/hazelcast/pull/6331

I have added new fields to `readData` and `writeData` methods of `CacheStatisticsImpl` but in fact this class is not serialized/deserialized between nodes since it is local. So there is no backward compatibility problem.


serkan-ozal:
Forward port is already merged. Merging this one myself.

PR #6347, opened 2015-10-01
MapPartitionDestroyOperation is now BackupAware to destroy containers on replicas

### LINKED ISSUE #6317: Hazelcast 3.5.x Memory Leak using IMap
We are experimenting memory leaks when using IMaps.

Based on our observations, the leak seems related to the fact that we're regularly destroying maps.
Based on Heap analysis, the culprit seems to be around com.hazelcast.map.impl.MapServiceContextImpl.

Setup: 
- Hazelcast 3.5.2, JDK 1.7/1.8, reproduced on Linux/Windows 7 and Mac OSX 10.9.5
- 2 nodes with embedded Hazelcast (e.g. not using separate client JVMs)

Scenario:
- One of our node is creating a new IMap roughly once per minute and put stuff in it at regular interval (few hundreds to few thousands per minute).
- Few seconds/minutes later, the other node is removing data from the map, and if the map is empty, destroys the map. 
- We do notice a steady growth of the heap usage, which seems related to Hazelcast IMaps internal management.

We did run a couple of variants with the following observations:
- Hazelcast versions 3.5.1 and 3.5.2 exhibits the same memory leak behavior
- We tested both with JMX/ManCenter enabled and disabled, without changes on the results.
- We disable Map "rotation" (e.g. a single map is used instead) and we didn't noticed the memory leak.
- We also tested with version 3.4.6, problem is NOT present - seems something introduced in 3.5.x

We wanted to update from 3.4.x since as our production system is affected by another memory leak documented here: https://github.com/hazelcast/hazelcast/issues/4888

Please find a small project on GitHub which demonstrates the problem: https://github.com/jocelynlepage/hz-map-leak
This project includes a couple of heap dump files (produced with jmap) provided for analysis.

Thanks,
Jocelyn

jerrinot:
Fixes #6317

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MapPartitionDestroyOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MapPartitionDestroyOperation.java
@@ -18,15 +18,20 @@ package com.hazelcast.map.impl.operation;
 
 
 import java.io.IOException;
+
+import com.hazelcast.map.impl.MapContainer;
+import com.hazelcast.map.impl.MapService;
 import com.hazelcast.map.impl.PartitionContainer;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.spi.BackupAwareOperation;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.PartitionAwareOperation;
 
-public class MapPartitionDestroyOperation extends Operation implements PartitionAwareOperation {
+public class MapPartitionDestroyOperation extends Operation implements PartitionAwareOperation, BackupAwareOperation {
     private final PartitionContainer partitionContainer;
     private final String mapName;
+    private int totalBackupCount;
 
     public MapPartitionDestroyOperation(PartitionContainer partitionContainer, String mapName) {
         this.partitionContainer = partitionContainer;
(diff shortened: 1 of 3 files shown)


jerrinot:
A backport will follow

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MapPartitionDestroyOperation.java
```diff
@@ -72,4 +79,24 @@ protected void readInternal(ObjectDataInput in) throws IOException {
...
+    }
+
+    @Override
+    public int getSyncBackupCount() {
+        return 0;
+    }
+
+    @Override
+    public int getAsyncBackupCount() {
+        return totalBackupCount;
```
eminn: Is there any particular reason to make backups async ?

jerrinot: I could find good reasons to wait for backup responses. This makes the `destroy()` method somewhat faster.

jerrinot: actually, you are right - it should be sync. as otherwise there might be operation reordering if there is a migration in progress.

jerrinot: @eminn: resolved

eminn:
Looks good to merge :+1:

ahmetmircik:
seems mergeable to me. :+1:

jerrinot:
thanks guys. I'll merge it on Green. 
I'll send a backport PR in a short while.

PR #6349, opened 2015-10-02
Prevents sending of not requested events to a MapListener

ahmetmircik:
Filters events according the implemented MapListener sub-interfaces.

Specifically, for example, if a listener is registered via an implementation like this :

```
 public class MyMapListener implements EntryAddedListener, EntryRemovedListener {
  ...
  }
``` 
That listener will only be notified for `EntryEventType#ADDED` and `EntryEventType#REMOVED` events.
  Other events, like `EntryEventType#EVICTED` or `EntryEventType#EXPIRED`, will not be sent over wire.
 This may help to reduce load on eventing system and network.


mdogan:
looks good to me

eminn:
:+1:

PR #6352, opened 2015-10-02
hibernate access types cleanup

mustafaiman:
cleanup hibernate initialisation of region factory with different access types

adds natural id tests for each access type


kobalski:
:+1: can be merged once green.

PR #6353, opened 2015-10-02
added missing tests to hz-cloud, removed unused methods

bilalyasar:


bilalyasar:
i will send another pr.

PR #6354, opened 2015-10-02
[BACKPORT] Execute the MapPartitionDestroyOperation on all local partitions even they are just backups replicas

### LINKED ISSUE #6317: Hazelcast 3.5.x Memory Leak using IMap

jerrinot:
Fixes #6317
It's not a direct backport of #6347 as we cannot add a new operation into a maintenance branch.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
@@ -25,8 +25,8 @@ import com.hazelcast.map.impl.eviction.ExpirationManager;
 import com.hazelcast.map.impl.operation.MapPartitionDestroyOperation;
 import com.hazelcast.map.listener.MapPartitionLostListener;
 import com.hazelcast.map.merge.MergePolicyProvider;
+import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.Data;
-import com.hazelcast.partition.InternalPartition;
 import com.hazelcast.partition.InternalPartitionService;
 import com.hazelcast.spi.EventFilter;
 import com.hazelcast.spi.EventRegistration;
@@ -181,19 +181,21 @@ class MapServiceContextImpl implements MapServiceContext {
     public void destroyMap(String mapName) {
         final PartitionContainer[] containers = partitionContainers;
         final List<Future> futures = new ArrayList<Future>(containers.length);
+        Address thisAddress = getNodeEngine().getThisAddress();
+        OperationService operationService = nodeEngine.getOperationService();
         for (PartitionContainer container : containers) {
-            if (container != null) {
-                int partitionId = container.getPartitionId();
-                InternalPartition partition = nodeEngine.getPartitionService().getPartition(partitionId);
-
-                if (partition.isLocal()) {
-                    OperationService operationService = nodeEngine.getOperationService();
-                    Operation operation = new MapPartitionDestroyOperation(container, mapName);
-
-                    Future f = operationService.invokeOnPartition(SERVICE_NAME, operation, partitionId);
-                    futures.add(f);
-                }
+            if (container == null) {
+                continue;
             }
+            int partitionId = container.getPartitionId();
+            Operation operation = new MapPartitionDestroyOperation(container, mapName);
+            operation.setPartitionId(partitionId);
+            // Execute the MapPartitionDestroyOperation on all partitions.
+            // We don't care if they are primary replicas or just backups
+            operation.setValidateTarget(false);
+
+            Future f = operationService.invokeOnTarget(SERVICE_NAME, operation, thisAddress);
+            futures.add(f);
         }
 
         try {
--- a/hazelcast/src/test/java/com/hazelcast/map/MapDestroyTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/MapDestroyTest.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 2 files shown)


eminn:
There are some checkstyle issue, apart from that, looks good to merge :+1:

jerrinot:
checkstyle issues resolved

hazelcast/src/test/java/com/hazelcast/map/MapDestroyTest.java
```diff
@@ -0,0 +1,99 @@
...
+@Category(QuickTest.class)
+public class MapDestroyTest extends HazelcastTestSupport {
+
+    private HazelcastInstance instance1;
+    private HazelcastInstance instance2;
+
+    @Before
+    public void setUp() {
+        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(2);
+        instance1 = factory.newHazelcastInstance();
```
pveentjer: I would rename them to local and remote. The reason is that you are also a map and the map you define on the local.

hazelcast/src/test/java/com/hazelcast/map/MapDestroyTest.java
```diff
@@ -0,0 +1,99 @@
...
+    private void assertAllPartitionContainersAreEmpty(HazelcastInstance instance) {
+        MapServiceContext context = getMapServiceContext(instance);
+        int partitionCount = getPartitionCount(instance);
+
+        for (int i = 0; i < partitionCount; i++) {
+            PartitionContainer container = context.getPartitionContainer(i);
+            ConcurrentMap<String, RecordStore> maps = container.getMaps();
+            assertTrue(maps.isEmpty());
+        }
+
```
pveentjer: why the empty line?

hazelcast/src/test/java/com/hazelcast/map/MapDestroyTest.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private MapServiceContext getMapServiceContext(HazelcastInstance instance) {
+        NodeEngineImpl nodeEngine1 = getNodeEngineImpl(instance);
```
pveentjer: why nodeEngine1?

hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
```diff
@@ -181,18 +181,19 @@ public void flushMaps() {
     public void destroyMap(String mapName) {
         final PartitionContainer[] containers = partitionContainers;
         final List<Future> futures = new ArrayList<Future>(containers.length);
+        Address localAddress = getNodeEngine().getLocalMember().getAddress();
         for (PartitionContainer container : containers) {
             if (container != null) {
```
pveentjer: To prevent excessive nesting, I normally do a if(container == null) continue;

Jump out of something as soon as you have no reason to be there.

hazelcast/src/test/java/com/hazelcast/map/MapDestroyTest.java
jerrinot: an artefact from refactoring.

hazelcast/src/test/java/com/hazelcast/map/MapDestroyTest.java
jerrinot: good one

hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
```diff
@@ -181,18 +181,19 @@ public void flushMaps() {
...
-                InternalPartition partition = nodeEngine.getPartitionService().getPartition(partitionId);
-
-                if (partition.isLocal()) {
-                    OperationService operationService = nodeEngine.getOperationService();
-                    Operation operation = new MapPartitionDestroyOperation(container, mapName);
-
-                    Future f = operationService.invokeOnPartition(SERVICE_NAME, operation, partitionId);
-                    futures.add(f);
-                }
+                OperationService operationService = nodeEngine.getOperationService();
```
pveentjer: Nagging old lady here; I would move the operationService retrievel to the last moment you need it. So first create operation, then retrieve the operationservice and do invoke.

hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
```diff
@@ -181,18 +181,19 @@ public void flushMaps() {
     public void destroyMap(String mapName) {
         final PartitionContainer[] containers = partitionContainers;
         final List<Future> futures = new ArrayList<Future>(containers.length);
+        Address localAddress = getNodeEngine().getLocalMember().getAddress();
```
pveentjer: Rename it to 'thisAddress'. And you can get it directly from NodeEngine.

pveentjer:
Looks good.

hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
jerrinot: thanks for pointing this out. I actually moved it out of the loop - so it's retrieved just once.

jerrinot:
@pveentjer: thanks for review, I pushed a new version.

pveentjer:
:+1:

PR #6357, opened 2015-10-02
Treat <> as an alias for !=

### LINKED ISSUE #6195: SQL<> should be an alias for !=
One more thing... In SQL there is not !=, the syntax is <>. Could <> be aliased?

jerrinot:
ANSI SQL defines <> only
Fixes #6195


Donnerbart:
:+1:

innerverse:
I did the same enhancement, but found a few more places to add tests:
https://github.com/hazelcast/hazelcast/compare/hazelcast:master...innerverse:enhancement/3.x/6195?expand=1

Specifically - adding some tests in ParserTest and QueryIndexTest might be useful.

jerrinot:
@innerverse: Hi, that makes sense. Would you like to send a PR with your changes instead of this one?

innerverse:
@jerrinot I submitted my version as a separate PR in case we want to use it.  My signed Contributor Agreement was emailed this morning.

jerrinot:
I'm closing this one in favour of #6362

PR #6359, opened 2015-10-02
Cleanup of ClientMapBasicTest

Donnerbart:
Cleanup of ClientMapBasicTest (removed final, added generics, no warnings anymore).

Nothing interesting here, just normal cleanup to match our actual coding style and to get rid of the yellow alert in those files.


sancar:
:+1:

eminn:
:+1:

PR #6360, opened 2015-10-02
Cleanup of ClientProperties

Donnerbart:
* Cleanup of `ClientProperties` according to `GroupProperties`.
* Pulled out `HazelcastProperties` to reduce code duplication.
* Added `@PrivateApi` annotation to `GroupProperty`, `ClientProperty` and `HazelcastProperty` classes.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientProperty.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientProperty.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.config;
+
+import com.hazelcast.instance.GroupProperty;
+import com.hazelcast.instance.HazelcastProperty;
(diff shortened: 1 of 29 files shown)


Donnerbart:
verify

hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientProperty.java
```diff
@@ -0,0 +1,143 @@
...
+
+    ClientProperty(String name, Integer defaultValue) {
+        this(name, String.valueOf(defaultValue));
+    }
+
+    ClientProperty(String name, String defaultValue) {
+        this(name, defaultValue, null);
+    }
+
+    ClientProperty(String name, Integer defaultValue, TimeUnit timeUnit) {
```
pveentjer: Something in the design goes wrong here. A timeunit should not be imposed on a property.

Donnerbart: It's the same as we have it in the `GroupProperties`. Also about 50% of the properties are `TimeUnit` based values, so it makes the usage more robust, since you now can retrieve the value in the correct time unit.

Donnerbart: I know that this is all far from a perfect solution. I just wanted to start moving this whole thing into a better direction by
* deprecating a grey-zoned public API (`PROP_` constants) so we can freely implement a better solution in 3.7/3.8
* ease the biggest pain with the usage of properties (which are the time unit based ones)

Donnerbart:
verify

hazelcast-client-new/src/main/java/com/hazelcast/client/config/ClientProperty.java
Donnerbart: @pveentjer Do you have any objections to merge this as an intermediate step? This is already done in this way on the member side, so right now it's inconsistent. Also we have to deprecate the old constants anyway, before you can replace it with a proper DI system.

pveentjer:
:+1:

Donnerbart:
I agree that this is just an intermediate solution. It should nevertheless ease the pain of adding new properties, until we have a better design. And more important it makes clear that this is not a public API.

I'm merging this one, since it's just an adaption of the `GroupProperties` refactoring which we already reviewed with >2 engineers.

PR #6361, opened 2015-10-02
Cleanup of GroupPropery usage

Donnerbart:
Made usage of `GroupyProperty` constants where string literals where used in code.


pveentjer:
:+1:

Issue #6363, opened 2015-10-03
`ClientInvocationFuture` may hang when deserialized response is null


serkan-ozal:
When the `responseDeserialized` field set in `ClientInvocationFuture`, after the received response is deserialized, response may be null (in cases such as missed get operation). So in this case, even though response is received, `get` can wait indefinetely since it sees `response` as `null`.

I have faced with this issue when I add `ExecutionCallback` to an invocation which returns `null` (missed cache get operation). `resolveResponse` method is called by async execution of added `ExecutionCallback` and after received response is deserialized in this call, actual `response` is set to `null` but the `get` method of `ClientInvocationFuture` waits for not `null` response to return. And in this case, caller hangs.

**P.S**: I have faced with this issue on old-client and I will send a PR for fix.

PR #6364, opened 2015-10-04
Handle null response after deserialized in `ClientInvocationFuture::resolveResponse`

serkan-ozal:
With this PR, if response is deserialized in `resolveResponse`, it checked if it is null or not. If it is null, `response` is set to a pre-defined constant object `NULL_RESPONSE` that represents null value for `response`. Also `resolveResponse` returns null to caller if `response` is `NULL_RESPONSE`.

Fixes https://github.com/hazelcast/hazelcast/issues/6363


sancar:
:+1:

sancar:
@serkan-ozal backport ?

serkan-ozal:
@sancar there is no problem with the maintenance branch. Because, as far as I see, in the maintenance branch deserialized response is not saved into `response` field so there is no problem like in the master branch. I mean, each time response is deserialized (I know it is not efficient way, but I didn't want to touch there)

PR #6366, opened 2015-10-04
[BACKPORT] Client side cache statistics implementation

### LINKED ISSUE #6262: Support cache stats at client side also
Currently, cache stats are only supported at server side and exposed by `CacheProxy`. But at client side cache stats are not calculated and supported so `UnsupportedOperationException` is thrown at `ClientCacheProxy`.

serkan-ozal:
This PR handles client side cache statistics.

Closes #6262

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
@@ -54,6 +54,7 @@ import com.hazelcast.client.util.ClientDelegatingFuture;
 import com.hazelcast.config.CacheConfig;
 import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.core.Client;
+import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
 import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.Member;
@@ -154,6 +155,9 @@ abstract class AbstractClientInternalCacheProxy<K, V>
     protected String nearCacheMembershipRegistrationId;
     protected final ConcurrentMap<Member, String> nearCacheInvalidationListeners = new ConcurrentHashMap<Member, String>();
 
+    protected final ClientCacheStatisticsImpl statistics;
+    protected final boolean statisticsEnabled;
+
     private boolean cacheOnUpdate;
     private final ConcurrentMap<CacheEntryListenerConfiguration, String> asyncListenerRegistrations;
     private final ConcurrentMap<CacheEntryListenerConfiguration, String> syncListenerRegistrations;
@@ -164,12 +168,20 @@ abstract class AbstractClientInternalCacheProxy<K, V>
                                                HazelcastClientCacheManager cacheManager) {
         super(cacheConfig, clientContext);
         this.cacheManager = cacheManager;
-        nearCacheManager = clientContext.getNearCacheManager();
-        asyncListenerRegistrations = new ConcurrentHashMap<CacheEntryListenerConfiguration, String>();
-        syncListenerRegistrations = new ConcurrentHashMap<CacheEntryListenerConfiguration, String>();
-        syncLocks = new ConcurrentHashMap<Integer, CountDownLatch>();
+        this.nearCacheManager = clientContext.getNearCacheManager();
+        this.asyncListenerRegistrations = new ConcurrentHashMap<CacheEntryListenerConfiguration, String>();
+        this.syncListenerRegistrations = new ConcurrentHashMap<CacheEntryListenerConfiguration, String>();
+        this.syncLocks = new ConcurrentHashMap<Integer, CountDownLatch>();
 
         initNearCache();
+
+        if (nearCache != null) {
+            this.statistics = new ClientCacheStatisticsImpl(System.currentTimeMillis(),
+                                                            nearCache.getNearCacheStats());
+        } else {
+            this.statistics = new ClientCacheStatisticsImpl(System.currentTimeMillis());
+        }
+        this.statisticsEnabled = cacheConfig.isStatisticsEnabled();
     }
 
     private void initNearCache() {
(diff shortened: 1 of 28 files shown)


serkan-ozal:
verify

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/stats/ClientCacheStatsTest.java
```diff
@@ -0,0 +1,156 @@
...
+        return HazelcastClientCachingProvider.createCachingProvider(client);
+    }
+
+    protected ClientConfig createClientConfig() {
+        return new ClientConfig();
+    }
+
+    @Test
+    public void testStatisticsDisabled() {
+        _testStatisticsDisabled();
```
pveentjer: what is the purpose of this strange naming convention?

hazelcast-client-new/src/test/java/com/hazelcast/client/cache/stats/ClientCacheStatsTest.java
```diff
@@ -0,0 +1,156 @@
...
+    public void testNearCacheStatsWhenNearCacheEnabled() {
+        String cacheName = randomName();
+        CacheConfig cacheConfig = createCacheConfig();
+        cacheConfig.setName(cacheName);
+        ClientConfig clientConfig = ((HazelcastClientProxy) client).getClientConfig();
+        clientConfig.addNearCacheConfig(new NearCacheConfig().setName(cacheName));
+        ICache<Integer, String> cache = createCache(cacheName, cacheConfig);
+        CacheStatistics stats = cache.getLocalCacheStatistics();
+
+        assertNotNull(stats);
```
pveentjer: I see in this test and in the tests above that you are testing if statistics is not null. Please create a dedicated test for that and remove this assert from all the other tests. A test that does:

do
assert
do 
assert

Is probably a broken test since it contains multiple situations.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheStatisticsImpl.java
```diff
@@ -0,0 +1,74 @@
...
+
+
+    @Override
+    public long getOwnedEntryCount() {
+        throw new UnsupportedOperationException("This statistic is not supported for client.");
+    }
+
+    @Override
+    public NearCacheStats getNearCacheStatistics() {
+        if (nearCacheStats != null) {
```
pveentjer: I would flip the 2 statements. Normally you first check for the unhappy situation and then you continue to the real logic. For example argument validation.

hazelcast/src/test/java/com/hazelcast/cache/stats/CacheStatsTestSupport.java
```diff
@@ -33,19 +29,63 @@ protected HazelcastInstance getHazelcastInstance() {
...
-    public void testCreationTime() {
+    protected void _testStatisticsDisabled() {
+        long now = System.currentTimeMillis();
+
+        CacheConfig cacheConfig = createCacheConfig();
+        cacheConfig.setStatisticsEnabled(false);
+        ICache<Integer, String> cache = createCache(cacheConfig);
+        CacheStatistics stats = cache.getLocalCacheStatistics();
+
+        assertNotNull(stats);
```
pveentjer: Again an example of do assert do assert.

Split it up in 2 different tests.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -409,11 +556,35 @@ protected ClientInvocationFuture invoke(ClientMessage req, Data keyData, int com
         } catch (Exception e) {
             throw ExceptionUtil.rethrow(e);
         }
-        return new ClientDelegatingFuture<Boolean>(future, clientContext.getSerializationService(),
-                putIfAbsentResponseDecoder);
+        ClientDelegatingFuture delegatingFuture =
+                new ClientDelegatingFuture<Boolean>(future, clientContext.getSerializationService(),
+                                                    putIfAbsentResponseDecoder);
+        if (async && statisticsEnabled) {
+            delegatingFuture.andThen(new ExecutionCallback<Object>() {
```
sancar: Can't we add this callback after `putIfAbsentAsyncInternal ` returned ?  They are also returning completable future. 
Async methods with isAsync boolean looks odd. I am basically trying to get rid of that boolean .
And my comment applies to all other async methods with async boolean.

sancar:
I had one comment. Other than that, pr looks good.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
serkan-ozal: @sancar at first version if this PR, I had implemented as you mentioned. But then I thought that it is not good idea using thread pools (executor) for sync operations for just statistics since its overhead to overall system may be more that making it as sync.

serkan-ozal:
Please do not merge it yet. I will override `getCacheEvictions` method at client cache statistics as unsupported operation.

serkan-ozal:
done!

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
sancar: @serkan-ozal  I did not understand. As far as I know, they are running on same executor, aren't they ? I mean, where is it running on when we choose to make the call inside the `putIfAbsentAsyncInternal ` method?

serkan-ozal: I don't mean there are different executors, I say that this statistics are just simple calculation and no need to handle them by creating an extra object (ExecutionCallback) and processing it on another executor thread.

serkan-ozal:
verify

serkan-ozal:
verify

enesakar:
verify

ihsandemir:
:+1:

serkan-ozal:
verify

PR #6381, opened 2015-10-06
compile fix for java 1.6

sancar:


serkan-ozal:
:+1:

jerrinot:
:+1:

serkan-ozal:
Is this problem only for new-client? what about `ClientCacheProxy` in old-client?

sancar:
It did not gave any error. Let me check.

sancar:
Compile error caused by usage of specifically usage of toObject method and applies only to new client. And as far as I understand, toObject is not needed anyway. I will remove toObject methods and update the pr.

sancar:
@jerrinot @serkan-ozal . I have updated the pr.

jerrinot:
:+1: once green

serkan-ozal:
:+1:

sancar:
verify

sancar:
Merging since tests are passed here https://github.com/hazelcast/hazelcast/pull/6387

PR #6382, opened 2015-10-06
compile fix for java 1.6

sancar:

backport of https://github.com/hazelcast/hazelcast/pull/6381


sancar:
verify

hasancelik:
verify

PR #6384, opened 2015-10-06
Performance: keyset/values download too much applied to new client

sancar:
JMH test results .

```
New client

Benchmark                 Mode  Cnt    Score    Error  Units
PTest.entrySet           thrpt    5  481.196 ± 30.155  ops/s
PTest.entrySetPredicate  thrpt    5  481.303 ± 36.378  ops/s
PTest.keyset             thrpt    5  582.044 ± 26.960  ops/s
PTest.keysetPredicate    thrpt    5  593.332 ± 25.451  ops/s
PTest.values             thrpt    5  578.870 ± 31.411  ops/s
PTest.valuesPredicate    thrpt    5  589.302 ± 11.811  ops/s

Old Client

Benchmark                 Mode  Cnt    Score    Error  Units
PTest.entrySet           thrpt    5  471.154 ± 16.446  ops/s
PTest.entrySetPredicate  thrpt    5  488.504 ± 11.056  ops/s
PTest.keyset             thrpt    5  536.753 ± 29.754  ops/s
PTest.keysetPredicate    thrpt    5  524.883 ± 35.712  ops/s
PTest.values             thrpt    5  549.128 ± 19.177  ops/s
PTest.valuesPredicate    thrpt    5  529.379 ± 21.721  ops/s
```


Donnerbart:
I don't see the migration of `ClientMapProxy.values()` to `ClientMapProxy.values(TruePredicate.INSTANCE)` in this PR. Do you plan to do this in a next step?

I'm asking because I took care about adpating the JavaDoc of `IMap` and the reference manual about the scope of the `QueryResultSizeExceededException`.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapEntriesWithPredicateMessageTask.java
```diff
@@ -38,11 +39,11 @@ public MapEntriesWithPredicateMessageTask(ClientMessage clientMessage, Node node
 
     @Override
     protected Object reduce(Collection<QueryResultRow> result) {
-        HashMap<Data, Data> map = new HashMap<Data, Data>();
-        for (QueryResultRow resultEntry : result) {
-            map.put(resultEntry.getKey(), resultEntry.getValue());
+        List<Map.Entry<Data, Data>> entries = new ArrayList<Map.Entry<Data, Data>>(result.size());
```
pveentjer: Smart using an ArrayList here.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapEntrySetMessageTask.java
```diff
@@ -18,60 +18,57 @@
...
         super(clientMessage, node, connection);
     }
 
     @Override
-    protected OperationFactory createOperationFactory() {
-        return new EntrySetOperationFactory(parameters.name);
+    protected Object reduce(Collection<QueryResultRow> result) {
+        List<Map.Entry<Data, Data>> entries = new ArrayList<Map.Entry<Data, Data>>(result.size());
+        for (QueryResultRow row : result) {
+            entries.add(new AbstractMap.SimpleEntry<Data, Data>(row.getKey(), row.getValue()));
```
pveentjer: Taste issue; but I try to prevent long lines. I would use a static import for SimpleEntry:

entries.add(new SimpleEntry<Data, Data>(row.getKey(), row.getValue()));

And I would probably pull out the entry:

SimpleEntry entry = new SimpleEntry<Data, Data>(row.getKey(), row.getValue());
entries.add(entry)

This way you don't need to analyze the line.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapEntrySetMessageTask.java
```diff
@@ -18,60 +18,57 @@
...
-        }
-        return dataMap;
+    protected Predicate getPredicate() {
+        return TruePredicate.INSTANCE;
     }
 
     @Override
-    protected MapEntrySetCodec.RequestParameters decodeClientMessage(ClientMessage clientMessage) {
-        return MapEntrySetCodec.decodeRequest(clientMessage);
+    protected IterationType getIterationType() {
```
pveentjer: What about the map.entryset(predicate) ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapKeySetMessageTask.java
```diff
@@ -18,44 +18,45 @@
...
+    protected Object reduce(Collection<QueryResultRow> result) {
+        List<Data> keys = new ArrayList<Data>(result.size());
+        for (QueryResultRow resultEntry : result) {
+            keys.add(resultEntry.getKey());
+        }
+        return keys;
     }
 
+    @Override
+    protected Predicate getPredicate() {
```
pveentjer: What about imap.keyset(predidcate) ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapValuesMessageTask.java
```diff
@@ -18,42 +18,45 @@
...
 
     @Override
-    protected Object reduce(Map<Integer, Object> results) {
-        List<Data> values = new ArrayList<Data>();
-        MapService mapService = getService(MapService.SERVICE_NAME);
-        for (Object result : results.values()) {
-            values.addAll(((MapValueCollection) mapService.getMapServiceContext().toObject(result)).getValues());
-        }
-        return values;
+    protected Predicate getPredicate() {
```
pveentjer: What about imap.values(predicate) ?

sancar: see MapValuesWithPredicateMessageTask
https://github.com/hazelcast/hazelcast/pull/6384/files#diff-d678d5e7b5bb8a10931b5aff20b0c8b1R50

sancar:
@Donnerbart I solved it diffrently since this one more sutiable to client. I do not chage the message send over the wire. It is more like ValuesRequest extends QueryRequest in the old client perspective. Same code runs on the server side.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapEntrySetMessageTask.java
sancar: see MapEntriesWithPredicateMessageTask

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapKeySetMessageTask.java
sancar: see MapKeySetWithPredicateMessageTask

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapEntrySetMessageTask.java
sancar: Upps. I forget to change this. QueryResultRow is already Map.Entry<Data, Data>. 
I mean I made it so that I don't have to create object here. I will update it.

Donnerbart:
So it will use the queries internally and may throw the exception if too much data is returned? Actually we'll see, since there are nightly tests checking those methods, and they will fail if you changed it under the hood :D

sancar:
run-parallel

sancar:
@pveentjer @Donnerbart Can we merge this pr now ? I have added the performance measurements with jmh at the top.

Donnerbart:
I was just worried about the queries not being used (since this would be a difference to the old client and the possible exception). Since this seems to be resolved I have no objections :)

After this PR being merged I expect the `ClientMapUnboundReturnValues_BasicTest` nightly test to fail (PR builder doesn't execute it). I will be happy to adapt the test, since I know what to do there.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
```diff
@@ -61,66 +63,86 @@ public Permission getRequiredPermission() {
...
     protected final Object call() throws Exception {
         Collection<QueryResultRow> result = new LinkedList<QueryResultRow>();
+        try {
+            Predicate predicate = getPredicate();
+            int partitionCount = clientEngine.getPartitionService().getPartitionCount();
+
+            BitSet finishedPartitions = invokeOnMembers(result, predicate, partitionCount);
+            invokeOnMissingPartitions(result, predicate, finishedPartitions, partitionCount);
+        } catch (Throwable t) {
+            throw ExceptionUtil.rethrow(t);
```
pveentjer: Is a non exception thrown here?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
```diff
@@ -61,66 +63,86 @@ public Permission getRequiredPermission() {
...
 
-    private boolean hasMissingPartitions(Set<Integer> finishedPartitions, int partitionCount) {
-        return finishedPartitions.size() != partitionCount;
+    private boolean hasMissingPartitions(BitSet finishedPartitions, int partitionCount) {
+        return finishedPartitions.nextClearBit(0) == partitionCount;
     }
 
-    private List<Integer> findMissingPartitions(Set<Integer> finishedPartitions, int partitionCount) {
+    private List<Integer> findMissingPartitions(BitSet finishedPartitions, int partitionCount) {
         List<Integer> missingList = new ArrayList<Integer>();
```
pveentjer: Not terribly exciting. but every place where we go through partitions and need to store them, I would go for an array. Arrays are cheap since no wrapper objects are needed + extremely fast is used correctly. And in case of partitions, you always know how big the array needs to be; so there is no need for growing.

pveentjer:
Looks good. Do we have enough tests for clients?

E.g. map[(1,peter), (2.peter)]

when you ask values.. how many peters do you get.

Also the paging predicate needs enough coverage.

In 3.7 we should clean up the client tests so they can reuse the server tests for logical testing. No need to repeat.

sancar:
@pveentjer I ported the changes from MapQueryRequest. I did not think on that that much. And there are tests in ClientMapBasicTest with and without predicates and checking the sizes.
And I see tests about PagingPredicate in ClientMapIssueTest.

PR #6385, opened 2015-10-06
compile fix for java 1.6

sancar:

backport of #6381


sancar:
verify

pveentjer:
It looks good to me. However Jenkins is failing for obscure reasons.

sancar:
Merging since tests are passed here #6387

PR #6386, opened 2015-10-06
[BACKPORT] adds partition lost listener for map config to spring

mustafaiman:


mustafaiman:
merging this one myself since it is already merged into master

PR #6389, opened 2015-10-06
rest: Added listing map entries function.

onderaycicek:
Added listing map entries function on rest api module.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
@@ -27,6 +27,7 @@ import static com.hazelcast.util.StringUtil.stringToBytes;
 public class HttpGetCommandProcessor extends HttpCommandProcessor<HttpGetCommand> {
 
     public static final String QUEUE_SIZE_COMMAND = "size";
+    public static final String MAP_ENTRIES_COMMAND = "entries";
 
     public HttpGetCommandProcessor(TextCommandService textCommandService) {
         super(textCommandService);
@@ -78,7 +79,12 @@ public class HttpGetCommandProcessor extends HttpCommandProcessor<HttpGetCommand
         int indexEnd = uri.indexOf('/', URI_MAPS.length());
         String mapName = uri.substring(URI_MAPS.length(), indexEnd);
         String key = uri.substring(indexEnd + 1);
-        Object value = textCommandService.get(mapName, key);
+        Object value;
+        if(MAP_ENTRIES_COMMAND.equalsIgnoreCase(key)){
+            value = textCommandService.entries(mapName);
+        }else{
+            value = textCommandService.get(mapName, key);
+        }
         prepareResponse(command, value);
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/TextCommandService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/TextCommandService.java
@@ -85,4 +85,6 @@ public interface TextCommandService {
     Node getNode();
 
     byte[] toByteArray(Object value);
+    
+    Object entries(String mapName);
 }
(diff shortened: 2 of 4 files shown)


jerrinot:
Hello @onderaycicek,

thank you for your contribution, it's much appreciated. There seems to be checkstyle issues - we use checkstyle to enforce a consistent coding style. You can run checkstyle locally if you activate `checkstyle` Maven profile: `mvn clean install -Pcheckstyle`

It would be great if you could provide a test for your changes - to make sure we'll not break your changes in the future. Cheers!

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
```diff
@@ -78,7 +79,12 @@ private void handleMap(HttpGetCommand command, String uri) {
         int indexEnd = uri.indexOf('/', URI_MAPS.length());
         String mapName = uri.substring(URI_MAPS.length(), indexEnd);
         String key = uri.substring(indexEnd + 1);
-        Object value = textCommandService.get(mapName, key);
+        Object value;
+        if(MAP_ENTRIES_COMMAND.equalsIgnoreCase(key)){
```
Donnerbart: If I read the surrounding code correctly you won't be able to fetch the value for the key `entries` of a map anymore. This maybe a corner case, but is at least very unexpected.

onderaycicek: Yes ,You are right to make a problem on setting keys like "entries"
i think it is better to define new URI for console commands. For example "
/hazelcast/rest/console/maps/mapName/command(entries,size,keys vs.) "
What do you think about it?

pveentjer: @Donnerbart any update?

Donnerbart: An additional REST command should not break any compatibility.
@asimarslan @sancar WDYT?

sancar: Hi @onderaycicek 
New addition should not be a problem. 
We had a quick discussion internally. console does not sound straightforward to us, although it  is ok for your use case.

Can you make it as follows ?
```
/hazelcast/rest/maps-entries/mapName/entries
```
I think, It can be implemented easily without breaking old behaviour. I downside is this change will not be sync with queue.

In queue we can get the size as
```
/hazelcast/rest/maps/queueName/size
```
Note: To `poll` from queue, one needs to give integer value as timeout value . this is why `size` was not a problem in `queue`, but a problem in `map`.

I think we should add the following also for queue, just for the sake symmetry.
```
/hazelcast/rest/queues-size/queueName/command/size
```

Later any command can be added following same pattern.

sancar:
Also can you change the commit messages to English ?

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
onderaycicek: Hi @sancar 
i think it is better if  we change to 
` /hazelcast/rest/maps-command/mapName/entries|size|keys|clear`

it is more useful for common usage with this solution.

For queues ;
`/hazelcast/rest/queues-command/queueName/size|iterator|clear `

What do you think about it?

if you think the same way , i will open new PR .

sancar: Sorry, there was a redundant part in the url in my previous comment I have updated the urls. Here is result of internal discussion with team:
```
/hazelcast/rest/maps-entries/mapName/
/hazelcast/rest/queues-size/queueName/
```  
Thanks for the contribution.

hasancelik:
run-lab-run

sancar:
@onderaycicek I am closing this one. If you are planning to update the pr, please feel free to reopen .

PR #6393, opened 2015-10-06
Serialization compatibility

asimarslan:


asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

asimarslan:
verify

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java
```diff
@@ -0,0 +1,466 @@
...
+    private SerializerAdapter registerFromSuperType(final Class type, final Class superType) {
+        final SerializerAdapter serializer = typeMap.get(superType);
+        if (serializer != null) {
+            safeRegister(type, serializer);
+        }
+        return serializer;
+    }
+
+    protected final SerializerAdapter serializerFor(final int typeId) {
+        if (typeId == 0) {
```
sancar: Is this 0 means CONSTANT_TYPE_NULL ?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java
```diff
@@ -0,0 +1,466 @@
...
+            }
+            writeDataInternal(out, data);
+        } catch (Throwable e) {
+            throw handleException(e);
+        }
+    }
+
+    @Override
+    public Data readData(ObjectDataInput input) {
+        return SerializationServiceHelper.readData(input);
```
sancar: `readData` and `writeData` looks odd. One is in SerializationServiceHelper and the other is in AbstractSerializationService ?!?!

sancar:
:+1:  A couple of comments, other than that it looks good.
Eline sağlık abi.
https://translate.google.com.tr/#tr/en/Eline%20sa%C4%9Fl%C4%B1k%20abi.

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
```diff
@@ -80,6 +81,15 @@ public void authenticate(ClientConnection connection) throws AuthenticationExcep
...
+        AuthenticationStatus authenticationStatus = AuthenticationStatus.findByValue(result.status);
+        switch (authenticationStatus) {
+            case AUTHENTICATED:
+                connection.setRemoteEndpoint(result.address);
+                return;
+            case CREDENTIALS_FAILED:
+                throw new AuthenticationException("Invalid credentials!");
+            default:
+                throw new AuthenticationException("Authentication status code not supported.");
+        }
```
ihsandemir: Are we missing SERIALIZATION_VERSION_MISMATCH check here? (similar in ClusterListenerSupport)

hazelcast/src/main/java/com/hazelcast/internal/serialization/ObjectDataOutputStream.java
```diff
@@ -205,11 +214,27 @@ public void writeShortArray(short[] shorts) throws IOException {
...
+
+    @Override
     public void writeUTF(String str) throws IOException {
-        if (utfBuffer == null) {
-            utfBuffer = new byte[UTF_BUFFER_SIZE];
+        int len = (str == null) ? 0 : str.length();
+        writeInt(len);
+        if (len > 0) {
+            for (int i = 0; i < len; i++) {
+                byte[] buffer = new byte[3];
```
ihsandemir: Optimization possibility: Could this "byte[] buffer = new byte[3];" be put out of the for loop and this buffer is reused?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataOutput.java
```diff
@@ -249,15 +246,19 @@ public void writeShort(int position, int v, ByteOrder byteOrder) throws IOExcept
 
     @Override
     public void writeUTF(final String str) throws IOException {
-        if (utfBuffer == null) {
-            utfBuffer = new byte[UTF_BUFFER_SIZE];
+        int len = (str != null) ? str.length() : NULL_ARRAY;
+        writeInt(len);
+        if (len > 0) {
+            ensureAvailable(len * 3);
```
ihsandemir: Is utf-8 maximum 4 bytes or 3 bytes? My understanding from https://en.wikipedia.org/wiki/UTF-8, it may be 4 bytes, can we double check on this?

Donnerbart: Should be up to 4 bytes: http://tools.ietf.org/html/rfc3629#section-3

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
```diff
@@ -63,14 +64,19 @@ public int totalSize() {
...
         return hashCode();
     }
 
     @Override
     public boolean hasPartitionHash() {
-        return totalSize() != 0 && payload[PARTITION_HASH_BIT_OFFSET] != 0;
+        int i = 1;
+        boolean payloadCheck = payload != null && payload.length > INT_SIZE_IN_BYTES;
+        return payloadCheck && payload[payload.length - i++] != 0
+                && payload[payload.length - i++] != 0
```
ihsandemir: Should this be || (OR) instead of &&?

hazelcast/src/main/java/com/hazelcast/internal/serialization/ObjectDataOutputStream.java
asimarslan: You're right but moreover these ObjectDataOutputStream and ObjectDataInputStream looks like should be in the test scope so I will move them there.

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
asimarslan: ClusterAuthenticator does not need that because it is only used to login to an already authenticated clusters.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
asimarslan: Checkstype do not like magic numbers. I personally prefer that but checkstyle do not allow so I wrote as above :)
It is AND. payload is not null and we need at least 4 bytes and all of them should be zero.

asimarslan:
@sancar @ihsandemir : this is the latest single commit. You may realize that I removed read/write Data form serialization service and move to streams.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
cangencer: What if the first byte of the partition hash is 0? It will short circuit and return false.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataOutput.java
```diff
@@ -249,15 +246,19 @@ public void writeShort(int position, int v, ByteOrder byteOrder) throws IOExcept
...
+            for (int i = 0; i < len; i++) {
+                pos += Bits.writeUtf8Char(buffer, pos, str.charAt(i));
```
cangencer: I have a suspicion this might not be right. If the char is part of a surrogate pair then you need both the low surrogate and the surrogate to convert it into a utf-8 character.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataOutput.java
asimarslan: It is upto 3 byte. See the converter method in Bits util

asimarslan:
verify

hasancelik:
verify

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataOutput.java
asimarslan: We don't support surrogates. We encode them but it result in non utf-8. Maybe we should fix that

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
```diff
@@ -63,14 +67,15 @@ public int totalSize() {
...
-            return Bits.readIntB(payload, payload.length - Bits.INT_SIZE_IN_BYTES);
+            return Bits.readIntB(payload, payload.length - INT_SIZE_IN_BYTES);
         }
         return hashCode();
     }
 
     @Override
     public boolean hasPartitionHash() {
-        return totalSize() != 0 && payload[PARTITION_HASH_BIT_OFFSET] != 0;
+        return payload != null && payload.length > INT_SIZE_IN_BYTES
```
ihsandemir: Should we have payload.length >= HEAP_DATA_OVERHEAD instead of payload.length > INT_SIZE_IN_BYTES?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DefaultPortableWriter.java
```diff
@@ -195,9 +202,16 @@ public void writeShortArray(String fieldName, short[] values) throws IOException
...
+            throws IOException {
+        setPosition(fieldName, FieldType.UTF_ARRAY);
+        out.writeUTFArray(values);
+    }
+
+    @Override
     public void writePortableArray(String fieldName, Portable[] portables) throws IOException {
         FieldDefinition fd = setPosition(fieldName, FieldType.PORTABLE_ARRAY);
-        final int len = portables == null ? 0 : portables.length;
+        final int len = portables == null ? -1 : portables.length;
```
cangencer: can use NULL_ARRAY here instead of -1

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DefaultPortableReader.java
```diff
@@ -277,6 +301,9 @@ public Portable readPortable(String fieldName) throws IOException {
             int len = in.readInt();
             int factoryId = in.readInt();
             int classId = in.readInt();
+            if (len == Bits.NULL_ARRAY) {
```
cangencer: minor but we could move this check to just after reading the length.

asimarslan: we have to read factoryId and classId , we cannot ignore them if it is null

asimarslan:
verify

asimarslan:
if (cloudBees.testsHang) {
   comment("verify");
}
//does this work :smile:

asimarslan:
verify

pveentjer:
Before we are going to get this merged, I would like to run a set of benchmarks to make sure we don't get a performance compression. We can use existing simulator tests for it, so that should not be an issue.

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
```diff
@@ -80,6 +81,16 @@ public void authenticate(ClientConnection connection) throws AuthenticationExcep
...
+        switch (authenticationStatus) {
+            case AUTHENTICATED:
+                connection.setRemoteEndpoint(result.address);
+                return;
+            case CREDENTIALS_FAILED:
+                throw new AuthenticationException("Invalid credentials!");
+            default:
+                //we do not need serialization version here as we already connected to master and agreed on the version
+                throw new AuthenticationException("Authentication status code not supported.");
+        }
```
pveentjer: Can you include 'authenticationStatus' in the exception?

asimarslan:
@pveentjer How long will that take. Today is EA day.

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
pveentjer: Can you include the authenticationStatus in the message?

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
```diff
@@ -69,6 +69,9 @@ public SerializationService createSerializationService() {
             SerializationServiceBuilder builder = new DefaultSerializationServiceBuilder();
             SerializationConfig serializationConfig = config.getSerializationConfig() != null ? config
                     .getSerializationConfig() : new SerializationConfig();
+            if (version > 0) {
```
pveentjer: What if version <= 0 ? Unfortunately bytes in java are signed, so -1 is a possibility.

pveentjer: I see below that -1 is accepted; this can be fine.. but can you add it to the javadoc?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/AuthenticationStatus.java
```diff
@@ -0,0 +1,46 @@
...
+    CREDENTIALS_FAILED(1),
+    SERIALIZATION_VERSION_MISMATCH(2);
+
+    private final byte status;
+
+    AuthenticationStatus(int status) {
+        this.status = (byte) status;
+    }
+
+    public byte getStatus() {
```
pveentjer: I would call this 'id' This is done at other parts in the system else you get status.getStatus() which looks funny.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/AuthenticationStatus.java
```diff
@@ -0,0 +1,46 @@
...
+        return status;
+    }
+
+    public static AuthenticationStatus findByValue(int status) {
```
pveentjer: At other parts in the system this is called 'getById'

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -76,65 +76,80 @@ private void handleEndpointNotCreatedConnectionNotAlive() {
...
+        switch (authenticationStatus) {
+            case AUTHENTICATED:
+                return handleAuthenticated();
+            case CREDENTIALS_FAILED:
+                return handleUnauthenticated();
+            case SERIALIZATION_VERSION_MISMATCH:
+                return handleSerializationVersionMismatch();
+            default:
         }
+        throw new IllegalStateException("Execution cannot reach here.");
```
pveentjer: Place this exception in the default: clause. There is no really a point in providing an empty default class since that is the 'default' behavior anyway.

pveentjer:
I would rename 'serialization version' to 'serial version' to allign it with Serializable:
Serial Version Uid
+ it also makes it a bit snappier to read. Else you get such long lines.

pveentjer:
Does this change wire compatibility? Is this version send over the wire and is therefor the wire content changed? So can a 3.5 client work with a 3.6 cluster e.g. ?

pveentjer:
@asimarslan I don't know. It depends on the number of tests being run. Keep in mind the the QuSP team is busy with finishing up their own stuff, so you need to run tests yourself. I'm currently trying to understand exactly what is being changed.. afaik it isn't that risky from a performance point of view from what I have seen so far. But I have not yet completed the PR.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
asimarslan: As the method must return something, I moved that from default to out of switch.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ArrayListStreamSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.StreamSerializer;
+
+import java.io.IOException;
+import java.util.ArrayList;
+
+/**
+ * The {@link java.util.ArrayList} serializer
+ */
+public class ArrayListStreamSerializer
```
pveentjer: Does this resolve the ArrayList serialization issue?

https://github.com/hazelcast/hazelcast/issues/6455

If so, tests are included as part of this PR? (I have not yet seen the whole PR)

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ArrayListStreamSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+        implements StreamSerializer<ArrayList> {
+
+    @Override
+    public void write(ObjectDataOutput out, ArrayList object)
```
pveentjer: Can't we rename object to list or something?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ArrayListStreamSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+            for (int i = 0; i < size; i++) {
+                out.writeObject(object.get(i));
+            }
+        }
+    }
+
+    @Override
+    public ArrayList read(ObjectDataInput in)
+            throws IOException {
+        if (in.readBoolean()) {
```
pveentjer: It is a flavor thing but I would always load this in a variable 'notNull' or something.

boolean notNull = in.readBoolean();
if(notNull){...}

Now the code becomes self documenting.

Apart from that I would flip the flag. So write true in case of null, since afaik that is the most used approach in our code I believe.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataInput.java
```diff
@@ -394,8 +399,27 @@ public short readShort(int position, ByteOrder byteOrder) throws IOException {
...
+            return null;
+        }
+        if (len > 0) {
+            boolean[] values = new boolean[len];
+            for (int i = 0; i < len; i++) {
+                values[i] = readBoolean();
+            }
+            return values;
+        }
+        return new boolean[0];
```
pveentjer: Probably we can cache the boolean[0]. No point in creating a new one.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataInput.java
```diff
@@ -385,6 +387,9 @@ public short readShort(int position, ByteOrder byteOrder) throws IOException {
     @Override
     public byte[] readByteArray() throws IOException {
         int len = readInt();
+        if (len == NULL_ARRAY) {
```
pveentjer: I would rename NULL_ARRAY to NULL_ARRAY_LENGTH or NULL_LENGTH. 

Because NULL_ARRAY makes you believe that the field is of type byte[], but actually it is a length indicator. So it puts the reader on the wrong foot.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataInput.java
```diff
@@ -522,10 +577,23 @@ public int readUnsignedShort() throws IOException {
...
+        for (int i = 0; i < charCount; i++) {
+            b = readByte();
+            if (b < 0) {
+                charBuffer[i] = Bits.readUtf8Char(this, b);
+            } else {
+                charBuffer[i] = (char) b;
+            }
         }
-        return UTFEncoderDecoder.readUTF(this, utfBuffer);
+        return new String(charBuffer, 0, charCount);
```
pveentjer: @jerrinot @serkan-ozal @noctarius can you have a look at this? You know quite a lot about the UTFEncoderDecoder.

What is the reason the UTFEncoderDecoder is dropped?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataInput.java
```diff
@@ -535,7 +603,9 @@ public final Object readObject() throws IOException {
 
     @Override
     public final Data readData() throws IOException {
-        return service.readData(this);
+        byte[] bytes = readByteArray();
+        Data data = bytes != null ? new HeapData(bytes) : null;
```
pveentjer: Flavor thing; I always try to get rid of !. 

Data data = bytes == null ? null : new HeapData(bytes);

Makes it easier to read because I don't need to flip the expression.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/LinkedListStreamSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+        implements StreamSerializer<LinkedList> {
+
+    @Override
+    public void write(ObjectDataOutput out, LinkedList object)
+            throws IOException {
+        out.writeBoolean(object != null);
+        if (object != null) {
+            int size = object.size();
+            out.writeInt(size);
+            Iterator iterator = object.iterator();
```
pveentjer: Probably best to make a check when the size == 0, in this case no iterator needs to be created.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationServiceHelper.java
```diff
@@ -0,0 +1,99 @@
...
+import com.hazelcast.nio.serialization.HazelcastSerializationException;
+import com.hazelcast.nio.serialization.Serializer;
+import com.hazelcast.nio.serialization.StreamSerializer;
+
+import java.util.Collections;
+import java.util.Set;
+
+import static java.lang.Boolean.parseBoolean;
+
+public final class SerializationServiceHelper {
```
pveentjer: Since this is a utility class,can you call it SerializationServiceUtils? Or just SerializationUtils since you don't care about the Service?

pveentjer: Personally I'm always a bit careful introducing helpers/utils. Should this not be part of the SerializationService?

pveentjer:
A large number of files have been changed but not a lot of test have been added. Can you check the test coverage of the classes you have created (Should be 85+%) or changed.

pveentjer:
@asimarslan I have a question about the design. I see there is a SerializationServiceV1; so I guess that once serialization changes are made we get a V2, V3 etc? My experience is that using class hierarchies are often problematic in combination with versioning because subtile differences in behavior can lead to certain methods not fitting into the super class. 

A more technical problem I have this this approach is that as soon as there are multiple different classes for the same 'interface' are in memory of the JVM that the JIT will have a much harder time. Normally it can easily inline something but with multiple implementations, this will be very difficult. 

Imho we should reconsider this approach for the reasons I described above.

@jerrinot WDYT.

pveentjer:
I have been thinking to make sure that we don't disturb the happy flow and we don't cause a lot of extra work.

The key is that the SerializationService should only have a single implementation. 

So what we can do is to create a super interface:
* CompatibilitySerializationService (please suggest better name). All methods from the SerializationService are moved to this interface
All compatibility serializers implement this interface; not SerializationInterface.
And only the latest and greatest serializer implements the SerializationService. 

In 99% of the code the SerializationService is used, and since they are only a single class implementing this interface, the JIT can inline. Only at places where there is is a potential need for compatibility, the CompatibilitySerializationService is used.

hazelcast-client-new/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
asimarslan:
```
    /**
     * Creates a {@link SerializationService} instance to be used by this client.
     *
     * @param version serialization version to be created. Values less than 1 will be ignored and max supported version
     * will be used
     * @return the created {@link SerializationService} instance
     */
    SerializationService createSerializationService(byte version);
```
Already defined in ClientExtension interface

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ArrayListStreamSerializer.java
asimarslan: This does not fix #6455, I will just refactor and fix that issue later. It's actually a part of bigger problem about resolving serializers from super classes.

PR #6395, opened 2015-10-07
More fluent properties setter as in server style

mmsolutions:


pveentjer:
Question is if this is going to cause a binary compatibility. The issue is is that the return type of a method is part of the signature of the call; so if the return type changes from void to an object, the call isn't valid anymore (the bytecode has become incompatible).

jerrinot:
@mmsolutions: Thank you for your contribution, we appreciate your effort! Unfortunately we can't merge your change as it would break binary compatibility. It's easy to demonstrate it: Compile some code against Hazelcast master and try to run it with your version of Hazelcast on a classpath - it will fail during runtime. 

@pveentjer: There is a Maven plugin to generate a synthetic method with the original signature, but I don't think it's worth it.

mmsolutions:
Thank you for your responses.

PR #6402, opened 2015-10-08
Remove dependency on (Map) Record in the indexing module

jerrinot:
The dependency was introduced by mistake at #6391
Indexing is used by the QueryCache as well -> it can't
depends on map-specific classes.


serkan-ozal:
:+1:

ahmetmircik:
:+1:

PR #6408, opened 2015-10-09
disabled memcache and rest as default

bilalyasar:


mesutcelik:
verify

bilalyasar:
verify

mesutcelik:
:+1:

hazelcast/src/test/java/com/hazelcast/internal/ascii/MemcacheTest.java
```diff
@@ -57,11 +58,17 @@
     final static Config config = new XmlConfigBuilder().build();
 
     @BeforeClass
+    public static void setup() throws IOException {
+        System.setProperty(GroupProperty.MEMCACHE_ENABLED.getName(),"true");
```
sancar: It will probably not affect the other tests but it is better to set the property on config

config.setProperty(GroupProperty.MEMCACHE_ENABLED.getName(),"true")

jerrinot:
This should be in big fat letters in 3.6 release notes. 
CC @Serdaro

sancar:
One minor comment.
:+1:  Looks good .

hazelcast/src/test/java/com/hazelcast/internal/ascii/MemcacheTest.java
mesutcelik: :+1:  as @sancar suggested.
this will also remove cleanup section in `@AfterClass`

bilalyasar:
@sancar @mesutcelik @jerrinot thanks for comments, updated.

bilalyasar:
it is green now.

PR #6409, opened 2015-10-09
The removeIndex() is not called when current record is null.

jerrinot:
Fixes findbugs issues. One of them turns out to be a real bug.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
@@ -629,7 +629,6 @@ public class DefaultRecordStore extends AbstractEvictableRecordStore implements 
 
         final Record record = getRecordOrNull(key, now, false);
         if (record == null) {
-            removeIndex(record);
             mapDataStore.remove(key, now);
         } else {
             return removeRecord(key, record, now) != null;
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
```diff
@@ -629,7 +629,6 @@ public boolean delete(Data key) {
 
         final Record record = getRecordOrNull(key, now, false);
         if (record == null) {
-            removeIndex(record);
```
jerrinot: this must have been some left-over not covered by test. 
The `removeIndex()` really does not make sense when the record is not stored locally. This change makes it aligned with the `remove()` method.

ahmetmircik:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
serkan-ozal: @jerrinot you seems right but maybe there was another intention. @ahmetmircik can you have a look?

serkan-ozal:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
jerrinot: @serkan-ozal: I have already discussed this with @ahmetmircik over slack.

serkan-ozal: ok, lets merge it

PR #6414, opened 2015-10-09
Minor query cleanups

pveentjer:


hazelcast/src/main/java/com/hazelcast/query/impl/CachedQueryEntry.java
```diff
@@ -110,34 +107,44 @@ public Comparable getAttribute(String attributeName) throws QueryException {
...
-                    targetObject = getKey();
-                }
+            return getTargetKey();
+        } else {
+            return getTargetValue();
+        }
+    }
+
+    private Object getTargetValue() {
+        Object targetObject;
```
Donnerbart: You could safe some lines of code and indentation with direct `return` statements. Also the definition order of `getTargetValue()` and `getTargetKey()` methods should be the same as the invocation order.

hazelcast/src/main/java/com/hazelcast/query/impl/IndexImpl.java
```diff
@@ -50,10 +50,10 @@
 
     private final SerializationService ss;
 
-    public IndexImpl(String attribute, boolean ordered, SerializationService ss) {
+    public IndexImpl(String attribute, boolean ordered, SerializationService serializationService) {
         this.attribute = attribute;
         this.ordered = ordered;
-        this.ss = ss;
+        this.ss = serializationService;
```
Donnerbart: Forgot to rename the field as well?

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryEntryFactory.java
```diff
@@ -24,13 +24,14 @@
 
 public final class QueryEntryFactory {
     private final boolean useCache;
+    private final SerializationService serializationService;
```
Donnerbart: Could you please align the field and constructor parameter order if you touch this PR anyway?

Pleaaaze

Donnerbart:
Minor comments :+1:

jerrinot:
:+1:  thanks!

PR #6423, opened 2015-10-10
Do not deserialize key when not needed

jerrinot:
Old implementation eagerly deserialized key. This is not needed as in most
cases desired attribute is extracted from value and not from key.

Moreover eager de-serialization means ClassNotFoundError when domain class is not
available on a classpath.


pveentjer:
:+1:

PR #6424, opened 2015-10-10
added warning log for host name configuration

bilalyasar:


mesutcelik:
verify

pveentjer:
Using a hostname in HZ is not recommended at all. It is fragile because at some points the ip address is used and other hostname which can lead to strange behavior in a cluster. At least that is what I remember.

bilalyasar:
run-parallel

bilalyasar:
run-parallel

bilalyasar:
verify

bilalyasar:
verify

PR #6429, opened 2015-10-12
Introduced InvocationMonitor

pveentjer:
The InvocationMonitor logic was pulled out from the InvocationRegistry. The InvocationRegistry is now a dumb container for invocations. And the monitoring (checking for timeouts, member leaving the cluster) and responding to it is a responsibility of the InvocationMonitor.

So it is a seperation of concerns.

One of the reasons, apart from having cleaner code, to have this seperation, is that in 3.7 I want to
experiment with different types of invocation-storage. Currently we use a CHM for it but it generates
quite a lot of waste.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
@@ -18,31 +18,23 @@ package com.hazelcast.spi.impl.operationservice.impl;
 
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
 import com.hazelcast.core.MemberLeftException;
-import com.hazelcast.instance.GroupProperties;
-import com.hazelcast.instance.GroupProperty;
-import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.internal.metrics.Probe;
 import com.hazelcast.logging.ILogger;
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -65,21 +58,16 @@
...
     private static final int INITIAL_CAPACITY = 1000;
     private static final float LOAD_FACTOR = 0.75f;
-    private static final int DELAY_MILLIS = 1000;
     private static final double HUNDRED_PERCENT = 100d;
 
-    private final long backupTimeoutMillis;
-
     @Probe(name = "invocations.pending", level = MANDATORY)
     private final ConcurrentMap<Long, Invocation> invocations;
+    //todo: this dependency on nodeEngineImpl sucks.
```
jerrinot: Do we need this? You could add this to (almost) every single field pointing to `NodeEngineImpl`.

pveentjer: Think so :) We need to get rid of it.

jerrinot:
:+1: on green

Donnerbart:
:+1:

gurbuzali:
the only comment should be addressed is a removal of comment, it can be done via direct commit
merging

PR #6431, opened 2015-10-12
Call `Backup::onExecutionFailure` instead of returning silently if backup is not valid (current node is not backup of this replica anymore)

### LINKED ISSUE #6420: `onExecutionFailure` should be called before returning from `run` if backup is not valid
After this PR https://github.com/hazelcast/hazelcast/pull/6301, Backup operation can be read directly as object without reading it first as `Data` temporary. 

However after the backup operation is read (deserialized) as object, if it is decided (at `Backup::beforeRun` so `valid` field is `false`) that current node is not backup of this replica anymore, `Backup::run` returns silently. But before it returns, it should call `onExecutionFailure` of the backup operation otherwise it causes native memory leaks due to undisposed native memory.

serkan-ozal:
Currently after the backup operation is read (deserialized) as object, if it is decided (at `Backup::beforeRun` so valid field is false) that current node is not backup of this replica anymore, `Backup::run` returns silently. But before it returns, it should call `onExecutionFailure` of the backup operation otherwise it causes native memory leaks due to undisposed native memory. 

With this PR, f backup is not valid (current node is not backup of this replica anymore), backup operation's `onExecutionFailure` is called so it can handle its own logic such as disposing allocated native memory for HD Cache/IMap operations.

Closes #6420 
Fixes https://github.com/hazelcast/hazelcast-enterprise/issues/340

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
@@ -95,9 +95,23 @@ public final class Backup extends Operation implements BackupOperation, Identifi
         }
     }
 
+    private void ensureBackupOperationInitialized() {
+        if (backupOp.getNodeEngine() == null) {
+            backupOp.setNodeEngine(getNodeEngine());
+            backupOp.setPartitionId(getPartitionId());
+            backupOp.setReplicaIndex(getReplicaIndex());
+            backupOp.setCallerUuid(getCallerUuid());
+            OperationAccessor.setCallerAddress(backupOp, getCallerAddress());
+            OperationAccessor.setInvocationTime(backupOp, Clock.currentTimeMillis());
+            backupOp.setOperationResponseHandler(createEmptyResponseHandler());
+        }
+    }
+
     @Override
     public void run() throws Exception {
         if (!valid) {
+            onExecutionFailure(
+                    new IllegalStateException("Wrong target! " + toString() + " cannot be processed!"));
             return;
         }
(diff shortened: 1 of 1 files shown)


gurbuzali:
looks good, @mdogan can you have a look too ?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
```diff
@@ -161,12 +170,10 @@ public boolean validatesTarget() {
     public void onExecutionFailure(Throwable e) {
         if (backupOp != null) {
```
ahmetmircik: What worries me is, if we have more than 1 backup, means we are operating by using `backupOpData`, the exception will not be print as in 1 backup-ed case. Here this null check can only be passed if we have 1 backup. 1 and more backup-ed cases are not parallel any more from this exception printing perspective.

mdogan:
There's another return point in `beforeRun()`:

```java
@Override
public void beforeRun() throws Exception {
    NodeEngine nodeEngine = getNodeEngine();
    InternalPartition partition = nodeEngine.getPartitionService().getPartition(partitionId);
    Address owner = partition.getReplicaAddress(getReplicaIndex());
    if (nodeEngine.getThisAddress().equals(owner)) {
        return;  // < ----------------- HERE
    }
...
```

serkan-ozal:
@mdogan I didn't get your point. `return` means there is no problem and then `run` is called.

serkan-ozal:
run-parallel

serkan-ozal:
verify

PR #6433, opened 2015-10-12
Changed TransactionalMapProxy to use predicates for keySet() and values()

Donnerbart:
* Changed `TransactionalMapProxy` to use predicates for `keySet()` and `values()` as well.
* Replaced deprecated `QueryResult` with usage of `QueryEngine`.
* Changed `MapUnboundedReturnValues_TxnTest` to expect `QueryResultSizeExceededException` on all txn methods.
* Cleanup of `TransactionalMapProxySupport`.
* Removed unused operations and factories.
* Adapted `ClientMapUnboundReturnValues_TxnTest` to check for `QueryResultSizeExceededException` on all txn methods.

We should do this in 3.6 since this changes the on-the-wire compatibility (otherwise we'll have to wait for 3.7 with aligning `IMap` and `TransactionalMap`). I couldn't resolve the `TODO`s with the duplicated data structures yet, but that can be done later since it won't change the protocol anymore.


gurbuzali:
:+1:

Donnerbart:
I just noticed that the old client seems to need no change, since the `AbstractTxnMapRequest` just delegates to the member methods. I changed the nightly `ClientMapUnboundReturnValues_TxnTest` accordingly.

Donnerbart:
The new client is delegating the `values()` and `keySet()` calls as well to the member methods, so there is no need to change anything beside the JavaDoc (already done) and the nightly tests.

It's nevertheless the question if we want to reduce the number of requests/codecs, since we could handle them with parameters (e.g. passing a null predicate to receive all keys/values). @sancar wdyt?

Donnerbart:
verify

sancar:
:+1:  Looks good. 
We intentionally make separate codecs for near future plans about client txn implemetations.

Donnerbart:
Well then, thanks for the review and the feedback. I'll merge this and keep in mind that we have to adapt the documentation regarding the scope of the query result size limiter (this will now be in effect of the whole IMap and TransactionalMap) \o/

PR #6435, opened 2015-10-12
Adds jclouds integration with hazelcast via using Discovery SPI API

kobalski:


hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsProperties.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public static final PropertyDefinition REGIONS = property("regions", STRING);
+    /**
+     * Property used to define tag keys for node filtering
+     */
+    public static final PropertyDefinition TAG_KEYS = property("tag-key", STRING);
+    /**
+     * Property used to define tag values for node filtering
+     */
+    public static final PropertyDefinition TAG_VALUES = property("tag-value", STRING);
```
bilalyasar: tag_keys, tag_values can be comma separated. I think this could be mentioned in java doc.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsProperties.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Property used to define group keys for node filtering
+     */
+    public static final PropertyDefinition GROUP = property("group", INTEGER);
```
bilalyasar: does this property correspond to `security groups` in ec2?

bilalyasar:
Will we have jclouds config as we have awsConfig?

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategy.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    public void destroy() {
+        computeServiceBuilder.destroy();
+    }
+
+    private DiscoveredNode buildDiscoveredNode(NodeMetadata metadata) {
+        Address privateAddressInstance = null;
+        if (!metadata.getPrivateAddresses().isEmpty()) {
+            InetAddress privateAddress = mapAddress(metadata.getPrivateAddresses().iterator().next());
+            privateAddressInstance =  new Address(privateAddress, NetworkConfig.DEFAULT_PORT);
```
mesutcelik: so we only support `5701` as member port?

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategy.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private InetAddress mapAddress(String address) {
+        if (address == null) {
+            return null;
+        }
+        try {
+            return InetAddress.getByName(address);
+        } catch (UnknownHostException e) {
+            LOGGER.warning("Address '" + address + "' could not be resolved");
```
mesutcelik: can we fail fast here instead of setting the discovered address to `null`?

pveentjer:
Exciting! You have no idea how much extra value this is going to provide! :)

I'll do a review shortly.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsProperties.java
kobalski: Done. Also changed property key tag-key to tag-keys

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsProperties.java
kobalski: Yes, it should work.

kobalski:
@bilalyasar maybe we should have a separate PR for configuration of this feature like aws-config. Right now we are using only discovery spi config apis.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategy.java
kobalski: No, port is configurable default port is 5701.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategy.java
kobalski: Done.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -0,0 +1,282 @@
...
+    private ComputeService computeService;
+
+    /**
+     * Instantiates a new Compute service builder.
+     *
+     * @param properties the properties
+     */
+    ComputeServiceBuilder(Map<String, Comparable> properties) {
+        if (properties == null) {
+            throw new NullPointerException("Properties can't be null");
```
mesutcelik: isn't `java.lang.IllegalArgumentException` better here?

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -0,0 +1,282 @@
...
+            ((HashSet<NodeMetadata>) filteredResult).add(metadata);
+        }
+        return filteredResult;
+    }
+
+    public int getServicePort() {
+        return getOrDefault(JCloudsProperties.HZ_PORT, NetworkConfig.DEFAULT_PORT);
+    }
+
+    private boolean checkRegionZoneConfig(NodeMetadata metadata) {
```
mesutcelik: i think this method can be implemented in simpler manner without that many branching and local variables.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private InetAddress mapAddress(String address) {
+        if (address == null) {
+            return null;
+        }
+        try {
+            return InetAddress.getByName(address);
+        } catch (UnknownHostException e) {
+            throw new UnsupportedOperationException("Address '" + address + "' could not be resolved");
```
mesutcelik: `UnsupportedOperationException` is not proper exception here.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -0,0 +1,282 @@
...
+                    break;
+                case REGION:
+                    region = location.getId();
+                    break;
+                default:
+                    break;
+            }
+            location = location.getParent();
+        }
+        if (regionsSet != null && region != null && regionsSet.contains(region)) {
```
mesutcelik: regionsSet can't be null i guess.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
kobalski: Done.

kobalski:
Tomorrow is the last day of code-freeze and getting a green light from PR builder is too hard these days, if there will be other comments to this pr, please make them before this noon.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+public class ComputeServiceBuilder {
+
+    private static final String JCLOUD_CONNECTION_TIMEOUT = "10000";
+    private static final ILogger LOGGER = Logger.getLogger(ComputeServiceBuilder.class);
+
+    private final Map<String, Comparable> properties;
+    private Set<String> regionsSet = new HashSet<String>();
+    private Set<String> zonesSet = new HashSet<String>();
+    private Set<String> tagsKeysSet = new HashSet<String>();
+    private Set<String> tagsValuesSet = new HashSet<String>();
```
bilalyasar: as far as i know, hash set doesn't preserve insertion order. But in that case, we need to preserve insertion order. Because every tag corresponds 1 value.. So maybe we can use `linkedHashSet`..Because it preserves insertion order.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+    private ComputeService computeService;
+
+    /**
+     * Instantiates a new Compute service builder.
+     *
+     * @param properties the properties
+     */
+    ComputeServiceBuilder(Map<String, Comparable> properties) {
+        if (properties == null) {
+            throw new IllegalArgumentException("Properties can't be null");
```
bilalyasar: you can use `preconditions` util class here. i.e.: `checkNotNull` method..

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -0,0 +1,279 @@
...
+        nodesFilter = new Predicate<ComputeMetadata>() {
+             @Override
+             public boolean apply(ComputeMetadata nodeMetadata) {
+                if (nodeMetadata == null) {
+                    return false;
+                }
+                for (String key: tagsKeysSet) {
+                    for (String value: tagsValuesSet) {
+                        if (!nodeMetadata.getUserMetadata().get(key).equals(value)) {
+                            return false;
```
bilalyasar: in ec2, every node can have a tag. ( key,value pair)
But here you look all combinations of (key-value) pairs..

kobalski: @bilalyasar Thanks for pointing out that. I did not test for multiple key value pairs, but this implementation would cause problems. I create a pair for each (tag-key,tag-value) tuple with `AbstractMap.SimpleImmutableEntry ` and make checks that way.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
kobalski: Done.

mesutcelik:
:+1:

bilalyasar:
Lgtm

PR #6445, opened 2015-10-13
Fixed Operation.toString

pveentjer:
The Operation has a nice toString method which is quite informative, unfortunately it was
overridden on quite a few places by one that doesn't include all this information.

This pr fixes that by replacing all Operation.toString methods by a better one that
can be glued into the one of the  operation

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheGetAllOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheGetAllOperation.java
@@ -71,8 +71,11 @@ public class CacheGetAllOperation
     }
 
     @Override
-    public String toString() {
-        return "CacheGetAllOperation{" + "keys:" + keys.toString() + "expiryPolicy:" + expiryPolicy + '}';
+    protected void toString(StringBuilder sb) {
+        super.toString(sb);
+
+        sb.append(", keys=").append(keys.toString());
+        sb.append(", expiryPolicy=").append(expiryPolicy);
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
@@ -571,15 +571,30 @@ public abstract class Operation implements DataSerializable {
 
     protected abstract void readInternal(ObjectDataInput in) throws IOException;
 
+    /**
+     * A template method allows for additional information to be passed into the {@link #toString()} method. So an Operation
+     * subclass can override this method and add additional debugging content. The default implementation does nothing so
+     * one is not forced to provide an empty implementation.
+     *
+     * It is a good practice always to call the super.toString(stringBuffer) when implementing this method to make sure
+     * that the super class can inject content if needed.
+     *
+     * @param sb the StringBuilder to add the debug info to.
+     */
+    protected void toString(StringBuilder sb) {
+    }
+
     @Override
     public String toString() {
-        final StringBuilder sb = new StringBuilder(getClass().getName()).append('{');
+        StringBuilder sb = new StringBuilder(getClass().getName()).append('{');
         sb.append("serviceName='").append(getServiceName()).append('\'');
         sb.append(", partitionId=").append(partitionId);
+        sb.append(", replicaIndex=").append(replicaIndex);
         sb.append(", callId=").append(callId);
         sb.append(", invocationTime=").append(invocationTime);
         sb.append(", waitTimeout=").append(waitTimeout);
         sb.append(", callTimeout=").append(callTimeout);
+        toString(sb);
         sb.append('}');
         return sb.toString();
     }
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/operations/FinalizeJoinOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/operations/FinalizeJoinOperation.java
@@ -128,11 +128,10 @@ public class FinalizeJoinOperation extends MemberInfoUpdateOperation implements 
     }
 
     @Override
-    public String toString() {
-        return "FinalizeJoinOperation{"
-                + "postJoinOp=" + postJoinOp
-                + "} "
-                + super.toString();
+    protected void toString(StringBuilder sb) {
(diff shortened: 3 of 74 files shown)


hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -571,15 +571,20 @@ public final void readData(ObjectDataInput in) throws IOException {
...
-        final StringBuilder sb = new StringBuilder(getClass().getName()).append('{');
+        StringBuilder sb = new StringBuilder(getClass().getName()).append('{');
         sb.append("serviceName='").append(getServiceName()).append('\'');
         sb.append(", partitionId=").append(partitionId);
+        sb.append(", replicaIndex=").append(replicaIndex);
         sb.append(", callId=").append(callId);
         sb.append(", invocationTime=").append(invocationTime);
         sb.append(", waitTimeout=").append(waitTimeout);
         sb.append(", callTimeout=").append(callTimeout);
+        sb.append(sb);
```
Donnerbart: Should this be `sb.append(toString(sb));`?

pveentjer: good one.

Donnerbart:
All specific information of the operations are still in the new methods, sometimes even more information was added. Beside one issue this PR seems very fine to me :+1:

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheGetAllOperation.java
```diff
@@ -71,8 +71,9 @@ public int getId() {
     }
 
     @Override
-    public String toString() {
-        return "CacheGetAllOperation{" + "keys:" + keys.toString() + "expiryPolicy:" + expiryPolicy + '}';
+    protected void toString(StringBuilder sb) {
+        sb.append(", keys=").append(keys.toString());
+        sb.append(", expiryPolicy=").append(expiryPolicy);
     }
```
tombujok: The hierarchy looks like this here:
```
CacheGetAllOperation extends PartitionWideCacheOperation
PartitionWideCacheOperation extends AbstractNamedOperation
```

As I reckon, the `name` should be added to the toString's output, but will be not.
Shouldn't the `toString(StringBuilder sb)` call `super.toString(sb)` at the beginning to make sure that all classes up the hierachy tree could add their input too (like in serialization)?

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheGetAllOperation.java
pveentjer: Good one

pveentjer:
@Donnerbart @tombujok can you have another look

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -571,15 +571,20 @@ public final void readData(ObjectDataInput in) throws IOException {
 
     protected abstract void readInternal(ObjectDataInput in) throws IOException;
 
+    protected void toString(StringBuilder sb) {
+    }
+
```
tombujok: There's still a slight problem here. Let's look at the hierarchy again:
```
CacheGetAllOperation extends PartitionWideCacheOperation
PartitionWideCacheOperation extends AbstractNamedOperation
```

I don't see PartitionWideCacheOperation in this PR, which means it probably doesn't implement the `toString(StringBuilder)` method. It means that it uses the default one from `Operation` - which is empty. 
Thus, it will break the chain of invocations of the toString() hierarchy and `AbstractNamedOperation.toString(StringBuilder)` will never be called.

Possible solution: making this method abstract? Then it's sure that very op implements this method.
But we still need a convention to call super.toString(sb) inside...

Donnerbart: I like the abstract method, since you cannot forget about the implementation. You anyway did it for most of the classes.

And what convention should be needed? `Operation` is abstract so it's fine to have `public abstract String toString(StringBuilder sb);` there. Then we make `public final String toString()` so it cannot be overridden by accident and use the abstract method in there. What else should be needed?

tombujok: Yep, that will work fine. The convention is just to call super.toString(sb) in the implementation of this abstract method in every subclass. If it's not done, everything up the inheritance hierarchy will not be included in the toString(sb) output. 
But I guess it's fine - we should just put it in the javadoc.

Donnerbart: Ah, got it. Yes, good point.

pveentjer: Making the toString(StringBuffer) abstract is not good. It forces every Operation implementation to come up with an implementation, even if it is empty. In that case it is better to provide a default implementation (empty) which is good enough for most. It is very important that you always take care of the use of use of an API. If something becomes painful, people will come up with a solution that might be a lot more ugly than you expect.

tombujok: Yes, true, but do you see that the empty implementation is broken in this case?

pveentjer: Can you point me to the problem? And you have looked at the latest changes?

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
tombujok: Sure, plz look at my comment beginning with "There's still a slight problem here. " (it's folded now)

pveentjer:
verify

tombujok:
After a discussion :+1: 
There's no easy way to solve this inheritance problem - let the implementers decide what they want to have in the toString() output. If they want the whole hierarchies - they may implement the toString(sb) methods in the class hierarchies they want to debug.

pveentjer:
When a caller forget to implement or forgets to call super, the most important information will always be available; the information from the Operation itself.

Donnerbart:
:+1: (once green)

PR #6446, opened 2015-10-13
Fixed JavaDoc of SERVICE_NAME constants.

Donnerbart:
Just a very minor change, but due to the number of files worthy for a PR.

Motivation for this PR was a broken JavaDoc reference to this outdated method name: `Service name of map service used to register {@link com.hazelcast.spi.impl.ServiceManager#registerService}`


pveentjer:
I would remove the part where you say that the name of service can be used to retrieve it from the service. It is copy pasted documentation.

Donnerbart:
So no JavaDoc for those constants at all? I could live with that.
```
/**
* Service name
*/
```
or
```
/**
* Service name for this service.
*/
```
or
```
/**
* Service name for {@link TheSameClass}.
*/
```
is not providing any useful information, since the constant name is already `SERVICE_NAME` in a `XXXService` class.

pveentjer:
So just get rid of it then :)

PR #6448, opened 2015-10-13
Update for hazelcast scripts

kobalski:
Updates server.sh script and writes hazelcast process id to a file then adds a stop script which reads current PID and shuts down hazelcast instance.

### DIFF
--- a/hazelcast/src/main/resources/server.sh
+++ b/hazelcast/src/main/resources/server.sh
@@ -3,13 +3,14 @@
 PRG="$0"
 PRGDIR=`dirname "$PRG"`
 HAZELCAST_HOME=`cd "$PRGDIR/.." >/dev/null; pwd`
+PID_FILE=/tmp/hazelcast_instance.pid
 
 if [ $JAVA_HOME ]
 then
-    echo "JAVA_HOME found at $JAVA_HOME"
-    RUN_JAVA=$JAVA_HOME/bin/java
+	echo "JAVA_HOME found at $JAVA_HOME"
+	RUN_JAVA=$JAVA_HOME/bin/java
 else
-    echo "JAVA_HOME environment variable not available."
+	echo "JAVA_HOME environment variable not available."
     RUN_JAVA=`which java 2>/dev/null`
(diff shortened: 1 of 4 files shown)


hasancelik:
verify

hazelcast/src/main/resources/server.sh
```diff
@@ -3,13 +3,14 @@
 PRG="$0"
 PRGDIR=`dirname "$PRG"`
 HAZELCAST_HOME=`cd "$PRGDIR/.." >/dev/null; pwd`
+PID_FILE=/tmp/hazelcast_instance.pid
```
mesutcelik: would be good to create it in local folder?

hazelcast/src/main/resources/stop.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+PRG="$0"
+PRGDIR=`dirname "$PRG"`
+HAZELCAST_HOME=`cd "$PRGDIR/.." >/dev/null; pwd`
+PID_FILE=/tmp/hazelcast_instance.pid
```
mesutcelik: same here.

hazelcast/src/main/resources/stop.sh
```diff
@@ -0,0 +1,13 @@
...
+PID=$(cat ${PID_FILE});
+
+if [[ -z "${PID}" ]]; then
+    echo "${PID_FILE}.pid is not running (missing PID)."
+else
+   kill -15 $1 ${PID}
```
mesutcelik: no need to give `-15` as it is default signal for `kill`

kobalski:
verify

mesutcelik:
:+1:

mesutcelik:
@kobalski , I think you can squash and send direct commit..

kobalski:
@mesutcelik Changes squashed and sent with direct commit  https://github.com/hazelcast/hazelcast/commit/16cf5e275902449dc09ece9ec0e02ce93af7ef44

Issue #6036, opened 2015-08-21
There should be a way to create JCache Cache manager declaratively when using Spring Configuration


jerrinot:
There is not a nice & supported way to create an instance of JCache CacheManager when using declarative Spring configuration. 

JCache CacheManager is needed to glue JCacheCacheManager (provided by Spring) and Hazelcast. I could come-up with something like this (not tested):

```xml
<beans>
  <hz:hazelcast id="hazelcastInstance" >
<!--		[...] -->
  </hz:hazelcast>

  <bean id="cachingProvider" 
    class="com.hazelcast.client.cache.impl.HazelcastServerCachingProvider"  <!-- this is a problem as it relies on a impl class -->
    factory-method="createCachingProvider">
   <constructor-args ref="hazelcastInstance" />
  </bean>

  <bean id="jcacheCacheManager" factory-bean="cachingProvider" factory-method="getCacheManager" />

  <bean id="cacheManager" class="org.springframework.cache.jcache.JCacheCacheManager">
    <constructor-args ref="jcacheCacheManager" />
  </bean>
</beans>
```

It's not exactly concise and it relies on our implementation details. 

Something like this would be nice:
```xml
<hz:hazelcast id="hazelcastInstance" >
  <hz:config id="hazelCastConfig">
<!--  [...]  -->
  </hz:config>
           
  <hz:cache-manager id="jcacheCacheManager"> <!-- This would create a JCache CacheManager from this Hazelcast instance -->
    <some configuration, URI, etc.. />
  </hz:cache-manager>
</hz:hazelcast>

  
<bean id="springCacheManager" class="org.springframework.cache.jcache.JCacheCacheManager">
  <constructor-args ref="jcacheCacheManager" />
</bean>
```

mustafaiman:
Hi @jerrinot 
I am trying to implement what you suggested. However I see a problem with your configuration.
Entry point for JCache API is `Caching.getCachingProvider()` method. This creates a new Hazelcast instance(server or client) and returns `HazelcastCachingProvider` based on this instance. Therefore we cannot use JCache API's entry point when parsing Spring XML configuration.

[...]

PR #6450, opened 2015-10-13
jcache support for spring cache manager

mustafaiman:
Spring applications can use Hazelcast's JCache support for caching. With this PR one can create a `javax.cache.CacheManager` bean backed by Hazelcast in their Spring xml configuration and link it with `SpringJCacheCacheManager`. This way Spring internally uses Hazelcast's JCache implementations for caching.

Related issue: https://github.com/hazelcast/hazelcast/issues/6036

Sample application-context.xml:
```xml
    <cache:annotation-driven cache-manager="cacheManager" />

    <hz:hazelcast id="instance">
        <hz:config>
            <hz:group name="dev" password="dev-pass"/>
            <hz:network port="5701" port-auto-increment="false">
                <hz:join>
                    <hz:multicast enabled="true"/>
                    <hz:tcp-ip enabled="false">
                        <hz:interface>127.0.0.1:5701</hz:interface>
                    </hz:tcp-ip>
                </hz:join>
                <hz:interfaces enabled="true">
                    <hz:interface>127.0.0.1</hz:interface>
                </hz:interfaces>
            </hz:network>
            <hz:cache name="name" />
            <hz:cache name="city" />
        </hz:config>
    </hz:hazelcast>

    <hz:cache-manager id="hazelcastJCacheCacheManager" instance-ref="instance" name="hazelcastJCacheCacheManager"/>

    <bean id="cacheManager" class="org.springframework.cache.jcache.JCacheCacheManager">
        <constructor-arg ref="hazelcastJCacheCacheManager" />
    </bean>

    <bean id="dummy" class="com.hazelcast.spring.cache.TestJCacheCacheManager.DummyBean" />
</beans>
```

### DIFF
--- a/hazelcast-spring/pom.xml
+++ b/hazelcast-spring/pom.xml
@@ -30,7 +30,7 @@
     </parent>
 
     <properties>
-        <org.springframework.version>3.1.0.RELEASE</org.springframework.version>
+        <org.springframework.version>4.0.2.RELEASE</org.springframework.version>
         <jsr250.api.version>1.0</jsr250.api.version>
         <!-- needed for checkstyle/findbugs -->
         <main.basedir>${project.parent.basedir}</main.basedir>
@@ -190,6 +190,12 @@
                     <version>4.3.8.Final</version>
                     <scope>test</scope>
                 </dependency>
+                <dependency>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-context-support</artifactId>
+                    <version>${org.springframework.version}</version>
+                    <scope>test</scope>
+                </dependency>
             </dependencies>
         </profile>
         <profile>
@@ -249,6 +255,18 @@
                     <groupId>org.hibernate</groupId>
                     <artifactId>hibernate-core</artifactId>
                     <version>3.6.10.Final</version>
+                    <scope>test</scope>
+                </dependency>
+                <dependency>
+                    <groupId>javax.cache</groupId>
+                    <artifactId>cache-api</artifactId>
+                    <version>1.0.0</version>
+                    <scope>test</scope>
+                </dependency>
+                <dependency>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-context-support</artifactId>
+                    <version>${org.springframework.version}</version>
                     <scope>test</scope>
                 </dependency>
             </dependencies>
--- a/hazelcast-spring/src/main/java/com/hazelcast/spring/CacheManagerBeanDefinitionParser.java
+++ b/hazelcast-spring/src/main/java/com/hazelcast/spring/CacheManagerBeanDefinitionParser.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.spring;
+
+import com.hazelcast.spring.cache.SpringHazelcastCachingProvider;
+import org.springframework.beans.factory.support.AbstractBeanDefinition;
+import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.xml.ParserContext;
+import org.w3c.dom.Element;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
+
+/**
+ * Bean definition parser for JCache {@link javax.cache.CacheManager}
+ * <p/>
+ * <b>Sample bean</b>
+ * <pre>
+ * &lt;hz:cache-manager id="cacheManager" instance-ref="instance" name="cacheManager" /&gt;
+ * </pre>
+ */
+public class CacheManagerBeanDefinitionParser
+        extends AbstractHazelcastBeanDefinitionParser {
+
+    @Override
+    protected AbstractBeanDefinition parseInternal(Element element, ParserContext parserContext) {
+        final SpringXmlBuilder springXmlBuilder = new SpringXmlBuilder(SpringHazelcastCachingProvider.class, parserContext);
+        springXmlBuilder.handle(element);
+        final BeanDefinitionBuilder builder = springXmlBuilder.getBuilder();
+        return builder.getBeanDefinition();
+    }
+
+    private class SpringXmlBuilder extends AbstractHazelcastBeanDefinitionParser.SpringXmlBuilderHelper {
+
+        private final ParserContext parserContext;
+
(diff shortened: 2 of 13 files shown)


kobalski:
:+1:

hazelcast-spring/pom.xml
```diff
@@ -30,7 +30,7 @@
     </parent>
 
     <properties>
-        <org.springframework.version>3.1.0.RELEASE</org.springframework.version>
+        <org.springframework.version>4.0.2.RELEASE</org.springframework.version>
```
jerrinot: why is this needed? does it mean we no longer support Spring Framework 3.x ?

hazelcast-spring/src/main/java/com/hazelcast/spring/CacheManagerBeanDefinitionParser.java
```diff
@@ -0,0 +1,80 @@
...
+
+        public BeanDefinitionBuilder getBuilder() {
+            return this.builder;
+        }
+
+        public void handle(Element element) {
+            handleCommonBeanAttributes(element, builder, parserContext);
+            final NamedNodeMap attrs = element.getAttributes();
+            if (attrs != null) {
+                Node instanceRefNode = attrs.getNamedItem("instance-ref");
```
jerrinot: is this the only way to pass the instance? Is there any other way to pass the instance? I do not know spring that well, I'm just asking.

hazelcast-spring/pom.xml
mustafaiman: JCacheCacheManager is present in 3.2.0 at least. However Spring 3.2.0 depends on JCache 0.5.
Spring 4.0.2 both have JCacheCacheManager and depends on JCache 1.0.0. This extension works under Spring 4 and JCache 1.0. It may work with Spring 3.x and JCache 0.5 but I could not test it yet.
Rest of the Spring module should work fine with Spring 3.x

hazelcast-spring/src/main/java/com/hazelcast/spring/CacheManagerBeanDefinitionParser.java
mustafaiman: This is how all the other custom beans of Hazelcast (`<hz:map> , <hz:list>, etc`) passes the instance reference. I do not know any other way to do that.

hazelcast-spring/pom.xml
mustafaiman: I have just tested it. It works with Spring 3.2.0 and JCache 0.5. We should mention in documentation that those who want to use this feature should depend on JCache 0.5.

jerrinot: can we make a runtime check and show a hint to a user?

hazelcast-spring/src/main/java/com/hazelcast/spring/CacheManagerBeanDefinitionParser.java
jerrinot: ok, as long as its consistent with the rest I'm fine with it.

hazelcast-spring/src/test/java/com/hazelcast/spring/cache/TestJCacheCacheManager.java
```diff
@@ -0,0 +1,128 @@
...
+import javax.annotation.Resource;
+import javax.cache.CacheManager;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+@RunWith(CustomSpringJUnit4ClassRunner.class)
+@ContextConfiguration(locations = {"jCacheCacheManager-applicationContext-hazelcast.xml"})
+@Category(QuickTest.class)
+public class TestJCacheCacheManager {
```
jerrinot: that's an unusual naming. why not `JCacheCacheManagerTest` ?

hazelcast-spring/src/test/java/com/hazelcast/spring/cache/TestJCacheCacheManager.java
```diff
@@ -0,0 +1,128 @@
...
+    @Resource(name = "cacheManager")
+    private JCacheCacheManager springCacheManager;
+
+    @Resource(name = "cacheManager2")
+    private CacheManager cacheManager2;
+
+    @BeforeClass
+    @AfterClass
+    public static void start() {
+        Hazelcast.shutdownAll();
```
jerrinot: shouldn't  the lifecycle be driven by Spring container? I know it's not related to this changeset, but I would expect container to manage lifecycle of Hazelcast instance here.

hazelcast-spring/src/test/java/com/hazelcast/spring/cache/TestJCacheCacheManager.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Before
+    public void reset() {
+        Iterable<String> cacheNames = springCacheManager.getCacheManager().getCacheNames();
+        for (String cacheName :  cacheNames) {
+            springCacheManager.getCacheManager().getCache(cacheName).removeAll();
```
jerrinot: why is this needed?

mustafaiman: caches are created once before tests start. therefore caches may have leftover entries from other tests. each test should run with empty caches

hazelcast-spring/pom.xml
mustafaiman: Spring 3 already crashes with JCache 1.0.0. I do not think we have to give users a hint about that.

hazelcast-spring/src/test/java/com/hazelcast/spring/cache/JCacheCacheManagerTest.java
```diff
@@ -0,0 +1,135 @@
...
+    @Resource(name = "cacheManager")
+    private JCacheCacheManager springCacheManager;
+
+    @Resource(name = "cacheManager2")
+    private CacheManager cacheManager2;
+
+    @BeforeClass
+    @AfterClass
+    public static void start() {
+        Hazelcast.shutdownAll();
```
jerrinot: I still do not understand this one. Why is it needed? Hazelcast instance is started by the container, shouldn't the container stop it on its destruction?

mustafaiman: Sorry, I should have missed it when you first asked.
Sure Spring container stops the instance on its destruction. That statement is to strictly ensure that instances are closed in any failure or success scenario. If `@AfterClass` annotation is removed, behavior will not change. You too can verify that. However I prefer keeping the statement as it is present in all other Spring tests.
I can remove the annotation if you still think I should. What do you think?

mesutcelik: It seems `@AfterClass` could be removed.

mesutcelik:
:+1: with minor comment.

hazelcast-spring/src/test/java/com/hazelcast/spring/cache/JCacheCacheManagerTest.java
jerrinot: If an instance is not closed -> it's a bug -> we want to know it -> test failure is OK to me. 
Did I miss anything?

jerrinot:
:+1: 

What about a documentation? CC @Serdaro ?

mustafaiman:
Once it is merged, documentation will follow!

jerrinot:
awesome!

PR #6460, opened 2015-10-14
Refactored Record related stuff

ahmetmircik:


gurbuzali:
:+1:

eminn:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
```diff
@@ -23,11 +23,17 @@
  */
 public interface SizeEstimator<T> {
```
pveentjer: In this case it would be a good thing to add documentation to the interface.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
```diff
@@ -23,11 +23,17 @@
...
 
+    /**
+     * Assume compressed object reference is used.
```
pveentjer: This is a bad assumption. References can be 32 or 64 bit depending on JVM being 32 or 64 and OOP compression.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimators.java
```diff
@@ -0,0 +1,83 @@
...
+     * Returns zero for all estimations.
+     */
+    private static final SizeEstimator ZERO_SIZE_ESTIMATOR = new ZeroSizeEstimator();
+
+    private SizeEstimators() {
+    }
+
+    public static SizeEstimator createMapSizeEstimator(InMemoryFormat inMemoryFormat) {
+        if (BINARY.equals(inMemoryFormat)) {
+            return new BinaryMapSizeEstimator();
```
pveentjer: Afaik there is no reason to create an instance of the BinaryMapSizeEstimator. Since it doesn't contain state.

hazelcast/src/main/java/com/hazelcast/map/impl/record/AbstractRecord.java
```diff
@@ -18,85 +18,140 @@
...
+        return creationTime;
+    }
+
+    @Override
+    public void setCreationTime(long creationTime) {
+        this.creationTime = creationTime;
     }
 
     @Override
     public long getCost() {
```
pveentjer: You know that this calculation is broken, right? Because you do not include the size of the CHM.Node for example.

hazelcast/src/main/java/com/hazelcast/map/impl/record/AbstractRecordWithStats.java
```diff
@@ -26,11 +24,8 @@
     protected RecordStatistics recordStatistics;
 
     protected AbstractRecordWithStats() {
-    }
-
-    protected AbstractRecordWithStats(Data key) {
-        super(key);
-        this.recordStatistics = new RecordStatistics();
+        super();
```
pveentjer: There is no need to call super. If you do not provide one, one is emitted by the Java Compiler. So this is only noise and can be removed.

hazelcast/src/main/java/com/hazelcast/map/impl/record/DataRecordWithStats.java
```diff
@@ -22,12 +22,13 @@
 
     protected Data value;
 
-    DataRecordWithStats(Data keyData, Data value) {
-        super(keyData);
-        this.value = value;
+    DataRecordWithStats() {
+        super();
```
pveentjer: No need for super.

hazelcast/src/main/java/com/hazelcast/map/impl/record/EmptyRecordStatistics.java
```diff
@@ -0,0 +1,86 @@
...
+
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+
+import java.io.IOException;
+
+/**
+ * Empty {@link RecordStatistics} implementation which returns zero for all stats.
+ */
+public class EmptyRecordStatistics implements RecordStatistics {
```
pveentjer: This is a good example of where you want to define a singleton. This class is stateless.

hazelcast/src/main/java/com/hazelcast/map/impl/record/ObjectRecord.java
```diff
@@ -16,17 +16,16 @@
...
 package com.hazelcast.map.impl.record;
 
-import com.hazelcast.nio.serialization.Data;
-
 class ObjectRecord extends AbstractRecord<Object> implements Record<Object> {
 
     private Object value;
 
     ObjectRecord() {
+        super();
```
pveentjer: No super needed.

hazelcast/src/main/java/com/hazelcast/map/impl/record/ObjectRecordWithStats.java
```diff
@@ -16,17 +16,16 @@
...
 package com.hazelcast.map.impl.record;
 
-import com.hazelcast.nio.serialization.Data;
-
 class ObjectRecordWithStats extends AbstractRecordWithStats<Object> {
 
     private Object value;
 
     public ObjectRecordWithStats() {
+        super();
```
pveentjer: No super.

hazelcast/src/main/java/com/hazelcast/map/impl/record/ObjectRecordWithStats.java
```diff
@@ -16,17 +16,16 @@
...
     }
 
-    public ObjectRecordWithStats(Data key, Object value) {
-        super(key);
+    public ObjectRecordWithStats(Object value) {
+        super();
```
pveentjer: No super

hazelcast/src/main/java/com/hazelcast/map/impl/record/Record.java
```diff
@@ -19,7 +19,7 @@
 import com.hazelcast.nio.serialization.Data;
 
 /**
- * @param <V>
+ * @param <V> the type of value which is in the Record
```
pveentjer: Afaik we don't need this type parameter at all. I have not seen a single usage of of Record with a concrete type other then Data or object.

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordFactory.java
```diff
@@ -16,21 +16,16 @@
 
 package com.hazelcast.map.impl.record;
 
-import com.hazelcast.config.InMemoryFormat;
-import com.hazelcast.nio.serialization.Data;
-
 /**
- * TODO need a better name than RecordFactory!
+ * Factory for creating records. Created for every partition.
```
pveentjer: Is this done per map instance?

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordFactory.java
```diff
@@ -16,21 +16,16 @@
...
  *
  * @param <T> the type of object which is going to be created.
  */
 public interface RecordFactory<T> {
 
-    Record<T> newRecord(Data key, Object value);
+    Record<T> newRecord(Object value);
 
     void setValue(Record<T> record, Object value);
```
pveentjer: These are methods that do not belong on a factory. A factory has a create method. Things like setValue and isEquals are not part of factory behavior. So what you have created here is something else than a factory.

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordStatistics.java
```diff
@@ -16,78 +16,34 @@
...
+public interface RecordStatistics extends DataSerializable {
 
-    // TODO is volatile needed? if yes then hits should be atomicnumber
-    protected int hits;
-    protected long lastStoredTime;
-    protected long expirationTime;
+    /**
+     * Used instead of null.
+     */
+    RecordStatistics EMPTY_STATS = new EmptyRecordStatistics();
```
pveentjer: I would not define this singleton here, I would define it on the EmptyRecordStatistics. And call it INSTANCE.

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordStatisticsImpl.java
```diff
@@ -0,0 +1,101 @@
...
+import com.hazelcast.util.Clock;
+
+import java.io.IOException;
+
+/**
+ * Statistics of a {@link com.hazelcast.map.impl.record.Record Record}
+ */
+public class RecordStatisticsImpl implements RecordStatistics {
+
+    // TODO is volatile needed? if yes then hits should be atomicnumber
```
pveentjer: Perhaps resolve this todo?

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordStatisticsImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public void setLastStoredTime(long lastStoredTime) {
+        this.lastStoredTime = lastStoredTime;
+    }
+
+    @Override
+    public long getMemoryCost() {
+        //size of the instance.
+        final int numberOfLongVariables = 3;
```
pveentjer: Define this as a constant for the class.  Also there are only 2 longs and 1 int. So 3 is not correct.

Apart from that the calculation is wrong. You are not including the size of the header of the object which is around 12 bytes normally. 

Also the Bits class where Long size etc is already defined. No need to first get the bits and then convert to bytes.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractEvictableRecordStore.java
```diff
@@ -287,6 +280,8 @@ protected void markRecordStoreExpirable(long ttl) {
 
     abstract Object evictInternal(Data key, boolean backup);
 
+    abstract Object evictInternal(Data key, Record removedRecord, boolean backup);
```
pveentjer: What does this mean? Please add javadoc. Since there is a complex architecture for the Map, documentation is required so that other engineers can work easily with map as well.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractEvictableRecordStore.java
```diff
@@ -426,6 +421,13 @@ protected void mergeRecordExpiration(Record record, EntryView mergingEntry) {
     }
 
     /**
+     * Returns value of removed record.
```
pveentjer: This is not proper javadoc. You only explain the return value, but not the actual behavior.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
```diff
@@ -314,5 +310,18 @@
      **/
     void maybeDoInitialLoad();
 
+    Storage createStorage(RecordFactory<R> recordFactory, InMemoryFormat memoryFormat);
```
pveentjer: There is no reason this method should be in the interface. The AbstractRecordStore wants to create the Storage, and you want to define this method in the sub class. 

1: Just make this method a protected abstract method which only needs to be seen by the AbstractRecordStore subclasses. There is no need to expose this in the API since nobody else apart from the AbstractRecordStore subclass wants to call this method.

2: Such methods can be very dangerous because due to the callback in the constructor, it can happen that the implementing methods sees a partially constructed object. I often prefer to inject the right Storage through the constructor; thereby following DI.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
```diff
@@ -314,5 +310,18 @@
...
+
+    Record createRecord(Object value, long ttlMillis, long now);
+
+    /**
+     * This can be used to release unused resources.
+     */
+    void dispose();
+
+    void destroy();
```
pveentjer: What is the difference between dispose and destroy? please add javadoc.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
```diff
@@ -314,5 +310,18 @@
...
+
+    Storage getStorage();
```
pveentjer: Now you are exposing the underlying abstraction. I would not do this. If someone needs to do something on the Storage, just add a method on the RecordStore. If you get a lot of such pass through calls, then probably the abstraction is broken.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
```diff
@@ -0,0 +1,60 @@
...
+
+package com.hazelcast.map.impl.recordstore;
+
+import com.hazelcast.map.impl.SizeEstimator;
+
+import java.util.Collection;
+import java.util.Set;
+
+/**
+ * Represents actual storage layer behind a {@link RecordStore}.
```
pveentjer: This doesn't add much value to me. The RecordStore is already the abstraction for record storage. You initially called it InternalRecordStore, but now you renamed to to another obscure name. What is the purpose of having another layer of indirection?

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
```diff
@@ -0,0 +1,60 @@
...
+ * Includes basic storage operations.
+ *
+ * @param <K> the key type to be put in this storage.
+ * @param <V> the value type to be put in this storage.
+ */
+public interface Storage<K, V> {
```
pveentjer: Zero javadoc. Please fix that since the imap internals is its own architecture.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
```diff
@@ -0,0 +1,60 @@
...
+
+    boolean isEmpty();
+
+    void clear();
+
+    void destroy();
+
+    SizeEstimator getSizeEstimator();
+
+    void setSizeEstimator(SizeEstimator sizeEstimator);
```
pveentjer: Why do we want to set it? I checked the code and it is only set in the EvictionMaxSizePolicyTest. Since it is just used for testing, don't expose it through your API. BUt just do a hard cast and create a setter on the implementation if needed.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
pveentjer: The name of this class is not self documenting for the following reasons:

* it doesn't track size, it tracks memory usage. Size in map and record store is being used to define the number of elements; not the memory usage. So the name 'size' is misleading.
* Size estimator gives impression of being a function like objects, so not tracking any state. However in this case you are tracking size state to prevent doing a full scan (which is good). However the name of the class doesn't reveal this. 

I would rename it to StorageMemoryUsageTracker.

Using this name it automatically resolves the issues above. Also is makes it immediately clear that it tracks it not from a map, but from a Storage. Afaik Storage is only used by single thread at any given moment, so therefor doesn't need to be threadsafe.

pveentjer:
I don't see much tests being added. Classes like the size estimators, storage etc.. they are perfect candidates for direct unit tests. There is no need to make use of higher level integration tests to test these classes. So please add additional tests.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
ahmetmircik: Existing code. It may be addressed later.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
ahmetmircik: Existing code. Subject of different discussion. Only made it more visible by moving it here.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimators.java
ahmetmircik: Can you please state your alternative name? This is the size estimator of binary in-memory-formatted map.

hazelcast/src/main/java/com/hazelcast/map/impl/record/AbstractRecord.java
ahmetmircik: It may not be the right calculation but this is not what i am trying to address with this PR and it should be the same calculation what we had before this PR, this PR does not touch it.

hazelcast/src/main/java/com/hazelcast/map/impl/record/AbstractRecordWithStats.java
ahmetmircik: right no need, it seems a leftover to me

hazelcast/src/main/java/com/hazelcast/map/impl/record/DataRecordWithStats.java
ahmetmircik: right no need, it seems a leftover to me

hazelcast/src/main/java/com/hazelcast/map/impl/record/ObjectRecord.java
ahmetmircik: right no need, it seems a leftover to me

hazelcast/src/main/java/com/hazelcast/map/impl/record/ObjectRecordWithStats.java
ahmetmircik: right no need, it seems a leftover to me

hazelcast/src/main/java/com/hazelcast/map/impl/record/ObjectRecordWithStats.java
ahmetmircik: right no need, it seems a leftover to me

hazelcast/src/main/java/com/hazelcast/map/impl/record/Record.java
ahmetmircik: seems you are right.

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordFactory.java
ahmetmircik: I think so. Mostly existing code. Subject of different PR.

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordStatisticsImpl.java
ahmetmircik: I think so. It was existing code. Not trying to address that.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractEvictableRecordStore.java
ahmetmircik: will do this.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractEvictableRecordStore.java
ahmetmircik: will do.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
ahmetmircik: wil do.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
ahmetmircik: Thanks for pointing this, will address it later.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
ahmetmircik: Need a layer which implements simple get/put/remove. This is the job of Storage. It is different from RecordStores methods because there is a complex logic behind every method of RecordStore like loading a record from MapLoader, intercepting operations, evict/expire of records and all these are using simple methods of Storage.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
ahmetmircik: will do.

ahmetmircik:
will add test for Storage especially.

PR #6461, opened 2015-10-14
Introduce shutdown method on Cluster interface for shutting down all nodes

metanet:
* Also wait until all replica syncs are completed since replica sync operations are allowed on PASSIVE state


metanet:
verify

mdogan:
:+1:

eminn:
:+1:

PR #6462, opened 2015-10-15
NullPointerException error

onderaycicek:
NullPointerException error occurs at line 375 and than REST service stoped.


hasancelik:
verify

pveentjer:
I doubt if catching the NPE is the right way to solve the problem. Can you send the NPE stacktrace so we can fix the cause instead of the symptom?

pveentjer:
@onderaycicek do you have the original stacktrace?

mesutcelik:
why was this PR sent to `3.4-gem` branch?

sancar:
Hi @onderaycicek , 
I am closing the issue. If you have more insights about to issue, please reopen the pr. To take action on pr we are gonna need more info from you.

PR #6468, opened 2015-10-15
Refactored NearCache related stuff + some tiny changes

ahmetmircik:
In this PR:
- As of now, with this PR, all near-caches implement the same `NearCache` interface. Moved all near-cache related logic to `NearCachedMapProxyImpl` and `ClientNearCachedMapProxyImpl`.
- Refactored MapCodecTemplate#getAll to use List internally instead of Set to hold binary keys.
- Added native-to-heap-copy for internal map-store implementations.
- Some additional tiny refactoring for extensibility


ahmetmircik:
run-parallel

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
```diff
@@ -0,0 +1,411 @@
...
+
+    @Override
+    public LocalMapStats getLocalMapStats() {
+        LocalMapStats localMapStats = super.getLocalMapStats();
+        NearCacheStats nearCacheStats = nearCache.getNearCacheStats();
+        ((LocalMapStatsImpl) localMapStats).setNearCacheStats(((NearCacheStatsImpl) nearCacheStats));
+        return localMapStats;
+    }
+
+    // TODO near-cache invalidation.
```
gurbuzali: todo?
should we invalidate returning map's keyset ?

hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
```diff
@@ -0,0 +1,395 @@
...
+
+    @Override
+    public LocalMapStats getLocalMapStats() {
+        LocalMapStats localMapStats = super.getLocalMapStats();
+        NearCacheStats nearCacheStats = nearCache.getNearCacheStats();
+        ((LocalMapStatsImpl) localMapStats).setNearCacheStats(((NearCacheStatsImpl) nearCacheStats));
+        return localMapStats;
+    }
+
+    // TODO near-cache invalidation.
```
gurbuzali: same here

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheDataRecordStore.java
```diff
@@ -79,6 +81,10 @@ protected NearCacheDataRecord valueToRecord(V value) {
 
     @Override
     protected V recordToValue(NearCacheDataRecord record) {
+        if (record.getValue() == null) {
+            nearCacheStats.incrementMisses();
+            return (V) NULL_OBJECT;
```
serkan-ozal: What about handling `null` <=> `NULL_OBJECT` in `NearCache` level not `NearCacheRecordStore` level?

I mean, `NearCacheRecordStore` should not allow any `null` but `NearCache` handles convertion between  `null` and `NULL_OBJECT`. 

* If the value on the put is `null`, `NearCache` converts value to `NULL_OBJECT` and passes it to underlying `NearCacheRecordStore`. 
* If the value is `NULL_OBJECT` on get returning from underlying `NearCacheRecordStore`, `NearCache` returns `null` to caller.

serkan-ozal: Also, maybe `NearCache` can take a parameter to its constuctor like `nullValueAllowed` and its default value is `false`. So `nullValueAllowed` decides `null` value is allowed and converts it to `NULL_OBJECT`, otherwise, throws `InvalidParameterException`.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheProvider.java
```diff
@@ -40,28 +41,28 @@
  */
 public class NearCacheProvider {
```
serkan-ozal: Not necessary at the moment but it would be nice if we could integrate `NearCacheProvider` with `NearCacheManager`.

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/NearCacheDataRecordStore.java
ahmetmircik: It would be better but will do it later improvements.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheProvider.java
ahmetmircik: I also noticed that but left it to future improvements.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
ahmetmircik: done.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
ahmetmircik: done.

PR #6476, opened 2015-10-16
Add IAM Role support for jclouds integration.

kobalski:
This pr makes call AWS API defined here http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials and parses and builds `Credentials` for JClouds API. Parser code is copied from `hazelcast-clouds` module.

### DIFF
--- a/hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
+++ b/hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
@@ -57,6 +57,8 @@ import static com.hazelcast.util.Preconditions.checkNotNull;
  */
 public class ComputeServiceBuilder {
 
+    private static final String GOOGLE_COMPUTE_ENGINE = "google-compute-engine";
+    private static final String AWS_EC2 = "aws-ec2";
     private static final String JCLOUD_CONNECTION_TIMEOUT = "10000";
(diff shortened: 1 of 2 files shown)


hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -144,7 +146,6 @@ ComputeService build() {
         final String identity = getOrNull(JCloudsProperties.IDENTITY);
         String credential = getOrNull(JCloudsProperties.CREDENTIAL);
         final String credentialPath = getOrNull(JCloudsProperties.CREDENTIAL_PATH);
-
         checkNotNull(cloudProvider, "Cloud provider is not set");
         checkNotNull(identity, "Cloud provider identity is not set");
```
bilalyasar: what if user only set `iam-role` ... in that case , user doesn't have to set `identity`. right?

kobalski: Yes, right moved checks here : https://github.com/hazelcast/hazelcast/pull/6476/files#diff-627080c6781ee2f36b9958aafc799128R256 If iam-role is set no identity or credential should be set. If aim-role is not set identity or credentials must be present.

kobalski:
verify

bilalyasar:
:+1: , once 💚

kobalski:
verify

PR #6478, opened 2015-10-16
Fix cluster state transaction's partition & migration handling

mdogan:
- Initial partition assignment must be rejected & fail when cluster state
is not ACTIVE. Partition specific/related invocations must fail immediately
if cluster state is not ACTIVE and initial partition assignments are empty.

- Fixed race issue between cluster state transaction and initial partition
assignment (and also partition rebalancing).


mdogan:
@metanet FYI

metanet:
very good finding and great work! 
:+1:

PR #6480, opened 2015-10-16
Client side cache batch putAll implementation

serkan-ozal:
This PR implements batching support for cache's putAll operation at client side.

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
@@ -24,8 +24,10 @@ import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.impl.protocol.codec.CacheGetAllCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheGetCodec;
+import com.hazelcast.client.impl.protocol.codec.CachePutAllCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheSizeCodec;
 import com.hazelcast.client.spi.ClientContext;
+import com.hazelcast.client.spi.ClientPartitionService;
 import com.hazelcast.client.spi.impl.ClientInvocation;
 import com.hazelcast.client.spi.impl.ClientInvocationFuture;
 import com.hazelcast.client.util.ClientDelegatingFuture;
@@ -38,12 +40,16 @@ import com.hazelcast.util.ExceptionUtil;
 
 import javax.cache.CacheException;
 import javax.cache.expiry.ExpiryPolicy;
+import java.util.AbstractMap;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 import static com.hazelcast.cache.impl.CacheProxyUtil.validateNotNull;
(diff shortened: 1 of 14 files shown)


hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
```diff
@@ -321,14 +327,142 @@ public V getAndPut(K key, V value, ExpiryPolicy expiryPolicy) {
...
+        for (int partitionId = 0; partitionId < entriesPerPartition.length; partitionId++) {
+            List<Map.Entry<Data, Data>> entries = entriesPerPartition[partitionId];
+            if (entries != null) {
+                int completionId = nextCompletionId();
+                // TODO If there is a single entry, we could make use of a put operation since that is a bit cheaper
+                ClientMessage request =
+                        CachePutAllCodec.encodeRequest(nameWithPrefix, entries, expiryPolicyData, partitionId);
+                Future f = invoke(request, partitionId, completionId);
+                futureEntryCountTuples.add(new FutureEntryCountTuple(f, entries.size()));
+                if (nearCache != null) {
```
metanet: what about updating near cache after future.get returns successfully, instead of here?

serkan-ozal: At first, I had thought like that but then I didn't want to use creating `ExecutionCallback` for each partition and process them on executor. 

Personally I don't prefer using  `ExecutionCallback` if it is not really needed and also I think until `Future` is done (between `future.invoke()` and `future.get()`), it is better to handling invalidating/updating near-cache at this time as sync.

serkan-ozal: BTW, I think you have not said that I should use `ExecutionCallback` so ignore that sentence but consider this one:

> I think until Future is done (between future.invoke() and future.get()), it is better to handling > invalidating/updating near-cache at this time as sync since putAll is sync here.

serkan-ozal: But from the consistency perspective, it seems better invalidating/updating near-cache after actual putAll is done successfully (after `future.get()`).

metanet: Yeah, that was my main point. We can update near cache after future.get() returns. I don't know if this may be necessary but we can also remove the corresponding entries from near cache if future.get() throws an exception. Then, there won't be a inconsistency case in near cache.

hazelcast-client-new/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
```diff
@@ -321,12 +327,150 @@ public V getAndPut(K key, V value, ExpiryPolicy expiryPolicy) {
...
+        Data expiryPolicyData = toData(expiryPolicy);
+        List<FutureEntriesTuple> futureEntriesTuples =
+                new ArrayList<FutureEntriesTuple>(entriesPerPartition.length);
+        for (int partitionId = 0; partitionId < entriesPerPartition.length; partitionId++) {
+            List<Map.Entry<Data, Data>> entries = entriesPerPartition[partitionId];
+            if (entries != null) {
+                int completionId = nextCompletionId();
+                // TODO If there is a single entry, we could make use of a put operation since that is a bit cheaper
+                ClientMessage request =
+                        CachePutAllCodec.encodeRequest(nameWithPrefix, entries, expiryPolicyData, partitionId);
```
sancar: Should last parameter be completionId ?

serkan-ozal: good catch!

sancar:
:+1:

metanet:
:+1:

PR #6482, opened 2015-10-16
DiscoverySPI 2nd/final batch

noctarius:
1. Added tests to higher the code coverage
2. Added Member parameter to DiscoveryStrategy::start and DiscoveryStrategy::destroy to offer the possibility for strategies to register themselves as services (or for multicast to offer themselves ;-))
3. Slightly changed DiscoveryServiceFactory interface to meet the new requirements
4. Added @Beta annotations and @since Javadoc tags
5. Added AbstractDiscoveryStrategy to offer convenience and common methods to implement strategies in even less code

**PS: To test the Discovery SPI we cannot use the test factories since those use a hardcoded version of the Joiner and ignore all configuration!**

**PPS: I expect the Member to, eventually, have public and private addresses, isn't that the plan?**

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/spi/impl/discovery/ClientDiscoverySpiTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/spi/impl/discovery/ClientDiscoverySpiTest.java
@@ -16,30 +16,37 @@
 
 package com.hazelcast.client.spi.impl.discovery;
 
+import com.hazelcast.client.HazelcastClient;
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.client.config.ClientNetworkConfig;
 import com.hazelcast.client.config.XmlClientConfigBuilder;
-import com.hazelcast.client.test.TestHazelcastFactory;
+import com.hazelcast.client.connection.AddressTranslator;
 import com.hazelcast.config.AwsConfig;
 import com.hazelcast.config.Config;
-import com.hazelcast.config.DiscoveryStrategiesConfig;
+import com.hazelcast.config.DiscoveryConfig;
 import com.hazelcast.config.DiscoveryStrategyConfig;
 import com.hazelcast.config.InterfacesConfig;
+import com.hazelcast.config.JoinConfig;
 import com.hazelcast.config.properties.PropertyDefinition;
 import com.hazelcast.config.properties.PropertyTypeConverter;
 import com.hazelcast.config.properties.SimplePropertyDefinition;
+import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.Logger;
 import com.hazelcast.nio.Address;
-import com.hazelcast.spi.discovery.DiscoveredNode;
-import com.hazelcast.spi.discovery.DiscoveryMode;
-import com.hazelcast.spi.discovery.integration.DiscoveryService;
-import com.hazelcast.spi.discovery.integration.DiscoveryServiceProvider;
+import com.hazelcast.spi.discovery.AbstractDiscoveryStrategy;
+import com.hazelcast.spi.discovery.DiscoveryNode;
 import com.hazelcast.spi.discovery.DiscoveryStrategy;
 import com.hazelcast.spi.discovery.DiscoveryStrategyFactory;
 import com.hazelcast.spi.discovery.NodeFilter;
-import com.hazelcast.spi.discovery.SimpleDiscoveredNode;
+import com.hazelcast.spi.discovery.SimpleDiscoveryNode;
 import com.hazelcast.spi.discovery.impl.DefaultDiscoveryService;
 import com.hazelcast.spi.discovery.impl.DefaultDiscoveryServiceProvider;
+import com.hazelcast.spi.discovery.integration.DiscoveryMode;
+import com.hazelcast.spi.discovery.integration.DiscoveryService;
+import com.hazelcast.spi.discovery.integration.DiscoveryServiceProvider;
+import com.hazelcast.spi.discovery.integration.DiscoveryServiceSettings;
 import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
@@ -47,7 +54,6 @@ import com.hazelcast.test.annotation.QuickTest;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
-import org.xml.sax.SAXException;
 
 import javax.xml.XMLConstants;
 import javax.xml.transform.Source;
@@ -63,12 +69,18 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 @RunWith(HazelcastSerialClassRunner.class)
 @Category(QuickTest.class)
 public class ClientDiscoverySpiTest extends HazelcastTestSupport {
+
+    private static final ILogger LOGGER = Logger.getLogger(ClientDiscoverySpiTest.class);
 
     @Test
     public void testSchema() throws Exception {
@@ -95,12 +107,12 @@ public class ClientDiscoverySpiTest extends HazelcastTestSupport {
         AwsConfig awsConfig = networkConfig.getAwsConfig();
         assertNull(awsConfig);
 
-        DiscoveryStrategiesConfig discoveryStrategiesConfig = networkConfig.getDiscoveryStrategiesConfig();
-        assertTrue(discoveryStrategiesConfig.isEnabled());
+        DiscoveryConfig discoveryConfig = networkConfig.getDiscoveryConfig();
+        assertTrue(discoveryConfig.isEnabled());
 
-        assertEquals(1, discoveryStrategiesConfig.getDiscoveryStrategyConfigs().size());
+        assertEquals(1, discoveryConfig.getDiscoveryStrategyConfigs().size());
 
-        DiscoveryStrategyConfig providerConfig = discoveryStrategiesConfig.getDiscoveryStrategyConfigs().iterator().next();
+        DiscoveryStrategyConfig providerConfig = discoveryConfig.getDiscoveryStrategyConfigs().iterator().next();
 
         assertEquals(3, providerConfig.getProperties().size());
         assertEquals("foo", providerConfig.getProperties().get("key-string"));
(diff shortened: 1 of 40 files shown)


noctarius:
verify

hazelcast-client-new/src/test/java/com/hazelcast/client/spi/impl/discovery/ClientDiscoverySpiTest.java
```diff
@@ -179,10 +208,36 @@ public void testNodeFilter_from_xml() throws Exception {
...
+        String xmlFileName = "hazelcast-client-discovery-spi-test.xml";
+        InputStream xmlResource = ClientDiscoverySpiTest.class.getClassLoader().getResourceAsStream(xmlFileName);
+        ClientConfig clientConfig = new XmlClientConfigBuilder(xmlResource).build();
+
+        ClientNetworkConfig networkConfig = clientConfig.getNetworkConfig();
+
+        DiscoveryStrategiesConfig discoveryStrategiesConfig = networkConfig.getDiscoveryStrategiesConfig();
+        assertNotNull(discoveryStrategiesConfig);
+        assertNotNull(discoveryStrategiesConfig.getNodeFilterClass());
+        ILogger logger = Logger.getLogger(ClientDiscoverySpiTest.class);
```
pveentjer: I would throw this in a static field. Only adds noise to the test method. (so the logger)

pveentjer: Also this is an example of a test doing too much. You are testing the discoveryStrategiesConfig and discoveryStrategiesConfig.getNodeFilterClass(). But that should not be part of this particular test; this test tests translation.. not constructions. So move that to another test.

noctarius: I would like to do this but wasn't that the problem with configuration since the loggers are not configured from the xml (like loggertype set to log4j?)

hazelcast-client-new/src/test/java/com/hazelcast/client/spi/impl/discovery/ClientDiscoverySpiTest.java
```diff
@@ -179,10 +208,36 @@ public void testNodeFilter_from_xml() throws Exception {
...
+
+        DiscoveryServiceProvider provider = new DefaultDiscoveryServiceProvider();
+        DiscoveryService discoveryService = provider.newDiscoveryService(null, logger, DiscoveryMode.Client,
+                discoveryStrategiesConfig, ClientDiscoverySpiTest.class.getClassLoader());
+
+        AddressTranslator translator = new DiscoveryAddressTranslator(discoveryService);
+
+        Address address = new Address("127.0.0.1", 50001);
+
+        translator.translate(null);
```
pveentjer: I'm missing something here; but what is being tested? Should the return value of the translate method not be verified?

hazelcast/src/main/java/com/hazelcast/config/properties/PropertyTypeConverter.java
```diff
@@ -29,7 +29,10 @@
     STRING {
         @Override
         public Comparable convert(Comparable value) {
-            return value.toString();
+            if (value != null) {
```
pveentjer: It is a flavor thing, but the initial code of my method is the unhappy stuff so that is out of the way when the real work begins. I normally do something like:

if(value == null){.....}

And in case of Hazelcast, we use the Preconditions. This is cleaner and less lines. I would also include the reason why you can't convert to string btw. Now a user might be a bit baffled.

hazelcast/src/main/java/com/hazelcast/spi/discovery/integration/DiscoveryServiceProvider.java
```diff
@@ -25,18 +27,23 @@
...
+     * @param discoveryMode             the current discovery mode
      * @param discoveryStrategiesConfig the configuration parsed from the XML or provided using the programmatic API
-     * @param configClassLoader        the classloader set in the configuration
+     * @param configClassLoader         the classloader set in the configuration
      * @return a new instance of the discovery service
      */
-    DiscoveryService newDiscoveryService(DiscoveryMode discoveryMode, DiscoveryStrategiesConfig discoveryStrategiesConfig,
-                                         ClassLoader configClassLoader);
+    DiscoveryService newDiscoveryService(Member member, ILogger logger, DiscoveryMode discoveryMode,
+                                         DiscoveryStrategiesConfig discoveryStrategiesConfig, ClassLoader configClassLoader);
```
pveentjer: perhaps rename to config? From the context it is clear that it is a DiscoveryStrategiesConfig.

hazelcast/src/test/java/com/hazelcast/config/properties/PropertyTypeConverterTest.java
```diff
@@ -0,0 +1,132 @@
...
+        PropertyTypeConverter.STRING.convert(null);
+    }
+
+    @Test
+    public void test_string_converter_success() throws Exception {
+        assertEquals("test", PropertyTypeConverter.STRING.convert("test"));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void test_short_converter_illegal_argument() throws Exception {
```
pveentjer: In the testing we use mostly the following naming practice:

test_shortConverted_whenIllegalArgument
shortConverted_whenIllegalArgument
or 
test_shortConverted_whenIllegalFormat_thenNumberFormatException
shortConverted_whenIllegalFormat_thenNumberFormatException.

hazelcast/src/main/java/com/hazelcast/config/DiscoveryStrategiesConfig.java
```diff
@@ -96,7 +96,7 @@ public boolean isEnabled() {
      * @return all enabled {@link DiscoveryStrategy} configurations
      */
     public Collection<DiscoveryStrategyConfig> getDiscoveryStrategyConfigs() {
-        return Collections.unmodifiableCollection(discoveryStrategyConfigs);
+        return discoveryStrategyConfigs;
```
kobalski: Can we also rename of this class, `DiscoveryStrategiesConfig` to `DiscoveryStrategyConfigs`. We have such a naming convention in our config collections: replicatedMapConfigs, wanReplicationConfigs, multiMapConfigs

hazelcast/src/test/java/com/hazelcast/config/properties/PropertyTypeConverterTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test(expected = NumberFormatException.class)
+    public void test_float_converter_number_format_exception() throws Exception {
+        PropertyTypeConverter.FLOAT.convert("test");
+    }
+
+    @Test
+    public void test_float_converter_success() throws Exception {
+        assertEquals(Float.MAX_VALUE, PropertyTypeConverter.FLOAT.convert(String.valueOf(Float.MAX_VALUE)));
```
pveentjer: I flavor thing but I prefer:

```
 float actual = PropertyTypeConverter.FLOAT.convert(String.valueOf(Float.MAX_VALUE))

 assertEquals(Float.MAX_VALUE, actual );
```

pveentjer:
One question: I see that member is being used in Discovery api's. Does this make sense? I would expect that first I do discovery and then a can connect to members. So why do we need members at this level?

hazelcast/src/main/java/com/hazelcast/spi/discovery/integration/DiscoveryServiceProvider.java
noctarius: What do you mean it is not clear? You mean it is not clear what this config is?

hazelcast/src/main/java/com/hazelcast/config/DiscoveryStrategiesConfig.java
noctarius: But this is another class whereas the other ones are lists of configs. This is a general difference but we probably just want to rename it to "DiscoveryConfig"?

noctarius:
The member is used to register yourself at a service discovery or in case of multicast to announce yourself to others. This is the integral part that was missing in the first shot and you would never be able to implement those things. Jaromir actually figured that out.

hazelcast/src/main/java/com/hazelcast/config/DiscoveryStrategiesConfig.java
kobalski: Yes, `DiscoveryConfig` might be better to indicate that this class is top-level config for the  Discovery SPI than the `DiscoveryStrategiesConfig`

hazelcast/src/main/java/com/hazelcast/spi/discovery/integration/DiscoveryServiceProvider.java
Donnerbart: I think he meant you could rename the variable `discoveryStrategiesConfig` to `config` to make it shorter, since it's clear what kind of config it is due to the context.

noctarius:
@kobalski please see the above PS (in the main description), I found that while looking through the code coverage, your 2nd jclouds test is useless for that matter.

kobalski:
@noctarius Yes it seems to be useless configs for jclouds is not touched , Can you please remove the second test in this PR? or it can stay as it is I will send another PR for increasing test coverage module, I can remove it in that PR.

noctarius:
Removed and pushed :)

kobalski:
@noctarius Thanks.

pveentjer:
I'm not that crazy about having a dependency on member because this increases the amount of cyclic dependencies. 

What do you actually need from the member?

noctarius:
As discussed with Peter I changed Member to DiscoveryNode to separate the Hazelcast API and Discovery SPI completely.

noctarius:
verify

hazelcast/src/main/java/com/hazelcast/spi/discovery/DiscoveryStrategy.java
```diff
@@ -29,17 +32,18 @@
...
+@Beta
 public interface DiscoveryStrategy {
 
     /**
      * The <tt>start</tt> method is used to initialize internal state and perform any kind of
      * startup procedure like multicast socket creation. The behavior of this method might
-     * change based on the passed {@link DiscoveredNode}.
-     *
-     * @param discoveryMode the provided discovery mode
+     * change based on the {@link Member} instance passed to the {@link DiscoveryStrategyFactory}.
```
pveentjer: Do we want to have any reference to Member in this package?

pveentjer: I see that DefaultDiscoveryService has a reference to Member. 

You are only using it for this method:
```
   this.discoveryNode = buildDiscoveryNode(member);
```

where the following logic is executed:

```
   private DiscoveryNode buildDiscoveryNode(Member member) {
        return member == null ? null : new SimpleDiscoveryNode(member.getAddress(), member.getAttributes());
    }
```

I would just pass the argument into the constructor and remove any reference to member from this whole package.

hazelcast/src/main/java/com/hazelcast/spi/discovery/DiscoveryNode.java
```diff
@@ -40,8 +41,11 @@
  * {@link com.hazelcast.spi.discovery.NodeFilter} is configured, these
  * properties might be used for further refinement of the discovered nodes
  * based on whatever the filter decides.
+ *
+ * @since 3.6
  */
-public interface DiscoveredNode {
+@Beta
```
pveentjer: For future maintainers it would be helpful to explain why an abstract class instead of an interface is being used. A lot of people see abstract classes as dirty and everything should go through interfaces.

hazelcast/src/main/java/com/hazelcast/spi/discovery/impl/DefaultDiscoveryService.java
```diff
@@ -174,13 +188,28 @@ private DiscoveryStrategy buildDiscoveryProvider(DiscoveryStrategyFactory factor
         Class<? extends DiscoveryStrategy> discoveryProviderType = factory.getDiscoveryStrategyType();
         String className = discoveryProviderType.getName();
 
-        for (DiscoveryStrategyConfig config : discoveryStrategyConfigs) {
-            if (config.getClassName().equals(className)) {
+        Iterator<DiscoveryStrategyConfig> iterator = discoveryStrategyConfigs.iterator();
```
pveentjer: Afaik you can use a for each here.

hazelcast/src/main/java/com/hazelcast/spi/discovery/integration/DiscoveryService.java
```diff
@@ -53,13 +57,13 @@
...
-    Iterable<DiscoveredNode> discoverNodes();
+    Iterable<DiscoveryNode> discoverNodes();
 
     /**
      * The <tt>start</tt> method is called on system startup to implement simple
      * lifecycle management. This method is expected to call
-     * {@link DiscoveryStrategy#destroy()} on all discovered and activated strategies
-     * before the service itself will shutdown.
+     * {@link DiscoveryStrategy#destroy()} on all discovered
+     * and activated strategies before the service itself will shutdown.
```
pveentjer: I would align the name of the method with the documentation. So either talk about destroy in the documentation, or rename the method to shutdown.

hazelcast/src/main/java/com/hazelcast/spi/discovery/integration/DiscoveryServiceProvider.java
```diff
@@ -25,18 +27,23 @@
...
+     * @param member            the current <tt>Member</tt> instance if created on member nodes, otherwise <tt>null</tt>
+     * @param logger            the <tt>ILogger</tt> instance to be used for logging output
+     * @param discoveryMode     the current discovery mode
+     * @param discoveryConfig   the configuration parsed from the XML or provided using the programmatic API
+     * @param configClassLoader the classloader set in the configuration
      * @return a new instance of the discovery service
      */
-    DiscoveryService newDiscoveryService(DiscoveryMode discoveryMode, DiscoveryStrategiesConfig discoveryStrategiesConfig,
-                                         ClassLoader configClassLoader);
+    DiscoveryService newDiscoveryService(Member member, ILogger logger, DiscoveryMode discoveryMode,
```
pveentjer: One design trick. Passing arguments sucks because if you add an argument, you break the interface. If you use an argument object, you can easily add new arguments without causing any problems. 

Something like this:
```
public interface DiscoveryServiceProvider {
    DiscoveryService newDiscoveryService(DiscoveryServiceProviderSetting settings);
}

public class DiscoveryServiceProviderSetting{
    public Member member;
    public ILogger logger;
    public DiscoveryMode discoveryMode;
    public DiscoveryConfig discoveryConfig;
    public ClassLoader configClassLoader;
}
```

This gives you a lot more flexibility + I find methods with many arguments awful to read due to lack of named arguments.

noctarius:
Applied all comments

pveentjer:
:+1

kobalski:
:+1:

noctarius:
Thanks guys, I like the changes we did :) Group work is still the best!

noctarius:
verify

noctarius:
verify

noctarius:
verify

PR #6485, opened 2015-10-18
[DONT MERGE] Invocation Registry Improvements

pveentjer:
This PR Contains various improvements:
- back pressure is always enabled for low priority operations. Having uncontrolled number of invocations is danger to the stability of the system since one can run into oome
- there is a lot less litter. Normal for a readonly low priority operations 3 Longs (callid's) and 1 CHM.Node is created. No litter is created.


pveentjer:
run-lab-run

pveentjer:
I'm going to close this one. Will work on it 3.8 again. The litter reduction + build in back pressure is something I like. But there are certain things that need to get addressed like retrying and back pressure on client requests.

PR #6492, opened 2015-10-18
Serialization tests

asimarslan:


hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/ByteArrayObjectDataInputTest.java
```diff
@@ -0,0 +1,641 @@
...
+        in.init(INIT_DATA, 2);
+        assertArrayEquals(INIT_DATA, in.data);
+        assertEquals(INIT_DATA.length, in.size);
+        assertEquals(2, in.pos);
+    }
+
+    @Test
+    public void testClear() throws Exception {
+        in.clear();
+        assertNull(in.data);
```
pveentjer: Minor comment: I always place a line between the action and the asserts. This way it is immediately clear where the verification begins based on the visual structure. This is a recommended practice in a lot of testing literature and mentioned on our wiki.

pveentjer:
:+1:

PR #6498, opened 2015-10-19
Add async management center request handling

mdogan:
Added AsyncConsoleRequest marker interface to denote management
center requests which should be offloaded to a separate thread,
because they are long running or blocking.

PS: This is to change cluster state and shutdown cluster from 
management center UI. Both are blocking operations.


mdogan:
@muratayan & @emrahkocaman: FYI

muratayan:
@emrahkocaman we can also use this for pausing Wan Replication I guess

emrahkocaman:
Thanks @mdogan , :+1: when it's green.

muratayan:
:+1:

PR #6500, opened 2015-10-19
Use specified instance if instance name is given regardless from URI is specified or not

### LINKED ISSUE #6454: HazelcastCachingProvider does not use specified instance when instance-name not specified
Both `HazelcastServerCachingProvider` and `HazelcastClientCachingProvider` does not use the instance specified by in object when `uri` is not null and `instance-name` not specified in Properties. They try to start a new instance instead.

serkan-ozal:
Fixes #6454


serkan-ozal:
verify

serkan-ozal:
verify

serkan-ozal:
verify

serkan-ozal:
verify

sancar:
:+1:

mustafaiman:
:+1:

PR #6501, opened 2015-10-19
When optimizerQuery is set then cache deserialized objects

jerrinot:
This part should be rewritten, this commit is just fixing buggy behaviour, not the problematic design.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
@@ -55,12 +55,20 @@ public final class Records {
             value = record.getValue();
         } else if (value == null) {
             value = record.getValue();
-            if (record instanceof CachedDataRecord && value instanceof Data && !((Data) value).isPortable()) {
+            if (shouldCache(record, value)) {
                 value = serializationService.toObject(value);
                 record.setCachedValue(value);
             }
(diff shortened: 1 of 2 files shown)


eminn:
I am not that familiar with the whole picture of the query mechanism, but this change looks good to merge :+1:

ahmetmircik:
:+1:

jerrinot:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
```diff
@@ -63,4 +63,9 @@ public static Object getValueOrCachedValue(Record record, SerializationService s
         return value;
     }
 
+    private static boolean shouldCache(Record record, Object value) {
+        return (record instanceof CachedDataRecord || record instanceof CachedDataRecordWithStats)
```
pveentjer: I would simplify this a bit. e.g.

```
if(!((record instanceof CachedDataRecord || record instanceof CachedDataRecordWithStats))
   return false;

if(....
```

The current implementation is painful to read.

pveentjer:
Minor comment. :+1:

jerrinot:
I refactored the `shouldCache()`  to make it more readable.

jerrinot:
verify

PR #6503, opened 2015-10-19
MorphingPortableReader tests

asimarslan:


sancar:
:+1:

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/MorphingPortableReader.java
```diff
@@ -198,10 +198,22 @@ public short readShort(String fieldName) throws IOException {
     }
 
     @Override
+    public boolean[] readBooleanArray(String fieldName) throws IOException {
+        FieldDefinition fd = cd.getField(fieldName);
+        if (fd == null) {
+            return null;
+        }
+        if (fd.getType() != FieldType.BOOLEAN_ARRAY) {
+            throw new IncompatibleClassChangeError();
```
pveentjer: For a future PR it would be a good thing to tell exactly what was incompatible. Currently you only signal that there is an issue, but not which field and the type conflict. Putting in a simple message can really help out the end user because now he has the complex task to figure out what exactly went wrong.

pveentjer: I have been checking the documentation of the PortableReader and it is unclear that a null or empty array is returned in certain cases. Can you add this to the documentation? Can be done by writing it once and copy pasting it to each method.

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/ClassAndFieldDefinitionTest.java
```diff
@@ -55,8 +56,7 @@ public void testClassDef_getField_properIndex() throws Exception {
     @Test
     public void testClassDef_hasField() throws Exception {
         for (int i = 0; i < classDefinition.getFieldCount(); i++) {
-            FieldDefinition field = classDefinition.getField(i);
-            assertNotNull(field);
+            assertTrue(classDefinition.hasField(fieldNames[i]));
```
pveentjer: I would be careful with such 'complex' expression in an assertTrue. Because it isn't immediately clear what is being asserted here (it doesn't read as a good book).

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/ClassAndFieldDefinitionTest.java
```diff
@@ -68,6 +68,19 @@ public void testClassDef_getFieldType() throws Exception {
         }
     }
 
+    @Test
+    public void testClassDef_getFieldClassId() throws Exception {
+        for (String fieldName : fieldNames) {
+            int classId = classDefinition.getFieldClassId(fieldName);
```
pveentjer: This is super clean.

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/MorphingPortableReaderTest.java
```diff
@@ -0,0 +1,218 @@
...
+        assertEquals(3, aShort);
+        assertEquals(2, aChar);
+        assertEquals(4, aInt);
+        assertEquals(0, reader.readInt("NO SUCH FIELD"));
+
+    }
+
+    @Test
+    public void testReadLong() throws Exception {
+        long aByte = reader.readLong("byte");
```
pveentjer: should the readLong("long") not be tested?

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/MorphingPortableReaderTest.java
```diff
@@ -0,0 +1,218 @@
...
+        service1.destroy();
+        service2.destroy();
+    }
+
+    @Test
+    public void testReadInt() throws Exception {
+        int aByte = reader.readInt("byte");
+        int aShort = reader.readInt("short");
+        int aChar = reader.readInt("char");
+        int aInt = reader.readInt("int");
```
pveentjer: What about the readInt("long") 

What happens in such a case?

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/MorphingPortableReaderTest.java
```diff
@@ -0,0 +1,218 @@
...
+        int aByte = reader.readShort("byte");
+        int aShort = reader.readShort("short");
+
+        assertEquals(1, aByte);
+        assertEquals(3, aShort);
+        assertEquals(0, reader.readShort("NO SUCH FIELD"));
+
+    }
+
+    @Test(expected = IncompatibleClassChangeError.class)
```
pveentjer: I'm confused with the below tests. They all expect an exception to be thrown, which is perfectly fine. But they also all have an assert as the last line. Afaik this assert will never be executed.

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/MorphingPortableReaderTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    }
+
+    @Test
+    public void testReadUTF() throws Exception {
+        assertNull(reader.readUTF("NO SUCH FIELD"));
+    }
+
+    @Test
+    public void testReadBoolean() throws Exception {
```
pveentjer: Should the readBoolean not be tested on a boolean field? Because afaik it is only tested what happens when the field is not found, but we have no idea what happens when the boolean field is found.

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/MorphingPortableReaderTest.java
```diff
@@ -0,0 +1,218 @@
...
+    public void testReadByte() throws Exception {
+        assertEquals(0, reader.readByte("NO SUCH FIELD"));
+    }
+
+    @Test
+    public void testReadChar() throws Exception {
+        assertEquals(0, reader.readChar("NO SUCH FIELD"));
+    }
+
+    @Test
```
pveentjer: What happens when the field is long?

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/MorphingPortableReaderTest.java
pveentjer: And what about float/double fields? Should an error be thrown?

pveentjer:
I have some comments. Some of them are flavor based. But there are also some missing cases and strange behavior in the tests which need to be addressed. Once fixed I'm happy to put my thumbs up..

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/MorphingPortableReader.java
asimarslan: I don't think we put each method doc that it may return null. Our serialization has a contract of "what you put , that you get" . That I will put into tutorial.

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/MorphingPortableReaderTest.java
asimarslan: That's incompatible type case, I will add tests for that

pveentjer: Thanks.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/MorphingPortableReader.java
```diff
@@ -47,7 +46,7 @@ public int readInt(String fieldName) throws IOException {
             case SHORT:
                 return super.readShort(fieldName);
             default:
-                throw new IncompatibleClassChangeError();
+                throw new IncompatibleClassChangeError("This type is incompatible to read from :" + fd.getType());
```
pveentjer: I would pull out the IncompatibleClassChangeError into a single method instead of copy pasting it.

Also the message isn't that informative. Since the message doesn't include:
*  the class name
* the expected type. 

I would include the ClassDefinition which includes version? So the user has all information at hand to deal with the IncompatibleClassChangeError.

asimarslan: We don't have class name there. But we can include classId of the portable and other detail you have mentioned

pveentjer:
:+1:

asimarslan:
verify

asimarslan:
verify

PR #6505, opened 2015-10-19
Retry cluster shutdown until it succeeds in ClusterShutdownTest

### LINKED ISSUE #6504: [TEST-FAILURE] ClusterShutdownTest.cluster_mustBeShutDown_by_singleMember_when_clusterState_ACTIVE
```
java.lang.AssertionError: expected:<SHUT_DOWN> but was:<ACTIVE>
	at com.hazelcast.cluster.ClusterShutdownTest$3.run(ClusterShutdownTest.java:97)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:828)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:842)
	at com.hazelcast.cluster.ClusterShutdownTest.assertNodesShutDownEventually(ClusterShutdownTest.java:93)
	at com.hazelcast.cluster.ClusterShutdownTest.testClusterShutdownWithSingleMember(ClusterShutdownTest.java:65)
	at com.hazelcast.cluster.ClusterShutdownTest.cluster_mustBeShutDown_by_singleMember_when_clusterState_ACTIVE(ClusterShutdownTest.java:26)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast/16376/testReport/junit/com.hazelcast.cluster/ClusterShutdownTest/cluster_mustBeShutDown_by_singleMember_when_clusterState_ACTIVE/

metanet:
* It may fail because of the ongoing replications or migrations and it will succeed once they are completed

Fixes #6504

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
@@ -14,12 +14,12 @@ import org.junit.runner.RunWith;
 
 import java.util.concurrent.CountDownLatch;
 
-import static com.hazelcast.cluster.AdvancedClusterStateTest.changeClusterStateEventually;
 import static org.junit.Assert.assertEquals;
 
 @RunWith(HazelcastParallelClassRunner.class)
 @Category({QuickTest.class, ParallelTest.class})
-public class ClusterShutdownTest extends HazelcastTestSupport {
+public class ClusterShutdownTest
+        extends HazelcastTestSupport {
 
     @Test
     public void cluster_mustBeShutDown_by_singleMember_when_clusterState_ACTIVE() {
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
```diff
@@ -56,11 +57,19 @@ private void testClusterShutdownWithSingleMember(final ClusterState clusterState
         final Thread thread = new Thread(new Runnable() {
             @Override
             public void run() {
-                instances[0].getCluster().shutdown();
+                while (true) {
```
pveentjer: Hmmm.. This looks strange to me. So it can happen that a shutdown call throws an exception and is ignored??

hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
```diff
@@ -56,11 +57,19 @@ private void testClusterShutdownWithSingleMember(final ClusterState clusterState
...
+                    try {
+                        instances[0].getCluster().shutdown();
+                        break;
+                    } catch (Exception ignored) {
```
pveentjer: I would definitely log the exception.

mdogan:
@metanet: what about removing ` warmUpPartitions(instances)` line from the shutdown tests? Is it required?

hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
mdogan: log why? this is a test and it's known that `shutdown()` can fail. Logging them will be just noise in the logs.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
mdogan: Why strange?

metanet:
@mdogan actually it is intentional. I wanted to test the cluster shutdown while partitions are assigned. Actually, after the fix we made yesterday on node shutdown, maybe we can put some data to increment partition versions in this tests too.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
pveentjer: I would expect the shutdown not to fail. Also in most applications, people don't take that much care of protecting the shutdown and a whole bunch of things are shutdown sequentially. If an exception is thrown, then the following services (which could be db connection pool etc) are not closed.

So I would expect a shutdown not to fail. If it does; it must be a programming error and this should be resolved, not suppressed.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
pveentjer: See previous comments. In my book a failing shutdown is a system error and should be resolved.

metanet: It is a cluster shutdown which terminates all nodes one by one. It will not fail while actually shutting down the nodes, it will fail during moving the cluster to the shut-down-able state. Since we know it will fail because of replica sync operations, I ignored the exception. How can a cluster shutdown operation fail is well-documented in javadocs.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
mdogan: This is a distributed shutdown and shutdowns all cluster members. A distributed operation is expected to fail. And also it's clearly defined in its doc. 

```
* @throws IllegalStateException    if member-list changes during the transaction
 * @throws TransactionException     if there's already an ongoing transaction
 *                                  or this transaction fails
 *                                  or this transaction timeouts
```

mdogan:
@metanet: We are already doing cluster-state changes while partitions are assigned. I think we can remove warmup here and just test shutdown behaviour.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterShutdownTest.java
pveentjer: Ok. 

It still feels funny for the above reasons. In my book shutdowns should not fails because they can cause problems for other components in a system that also need to shut down. It can leave a system in an undefined half shut down state

metanet: there is an old saying, you know :)

"timeo hominem unius libri"

Joking aside, I fully understand your concern. As I stated, if cluster shutdown fails, it means that:
- Either cluster state failed to move into a shut-down-able (PASSIVE) state which creates no problem since everything will work fine. 
- Or cluster state is moved to (PASSIVE) state, some of the nodes has actually shut down but some others hasn't shutdown because of any failure (message loss etc.). In this case, still-running nodes must be shut-down by triggering the shutdown process on these nodes. Otherwise, they will not accept any operations since they are in PASSIVE state. This is still a safe state.

metanet:
unrelated test failure

metanet:
verify

mdogan:
:+1:

PR #6452, opened 2015-10-14
Refactorings for IMap eviction

ahmetmircik:
- Renamed old EvictionOperator as Evictor and made some distillation
- Removed boolean parameter backup from evict methods.
- Some tiny changes to provide extensibility.

PR #6507, opened 2015-10-20
[DO NOT MERGE] Evictor interface refactored according to review comments

ahmetmircik:
Addressed some of review comments of merged PR https://github.com/hazelcast/hazelcast/pull/6452


ahmetmircik:
@serkan-ozal can you please have a look?

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/Evictor.java
```diff
@@ -32,20 +32,21 @@
      * @param recordStore the recordStore
      * @return removal size.
      */
-    int findRemovalSize(RecordStore recordStore);
+    int findSizeToEvict(RecordStore recordStore);
```
pveentjer: The javadoc doesn't explain at all what this method is doing and the method name either. So what is this method actually doing?

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/Evictor.java
```diff
@@ -32,20 +32,21 @@
...
 
     /**
      * Remove supplied number of elements from record-store.
      *
-     * @param removalSize supplied size to remove.
+     * @param sizeToEvict supplied size to remove.
      * @param recordStore the recordStore
      */
-    void removeSize(int removalSize, RecordStore recordStore);
+    void evictBySize(int sizeToEvict, RecordStore recordStore);
```
pveentjer: What does this mean? evict by size? And what does size mean? Is that item count or is that memory usage?

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/Evictor.java
```diff
@@ -32,20 +32,21 @@
...
     /**
-     * Get eviction checker for this {@link Evictor}
+     * Check whether the supplied record-store needs eviction.
      *
-     * @return the {@link EvictionChecker}
+     * @param recordStore the recordStore
+     * @return {@code true} if eviction is required, {@code false} otherwise.
      */
-    EvictionChecker getEvictionChecker();
+    boolean checkEvictionPossible(RecordStore recordStore);
```
pveentjer: I would rename this to is eviction required or something.

serkan-ozal:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictorImpl.java
```diff
@@ -38,22 +47,26 @@
...
         this.mapServiceContext = mapServiceContext;
+        this.memoryInfoAccessor = new RuntimeMemoryInfoAccessor();
     }
 
     @Override
-    public EvictionChecker getEvictionChecker() {
-        return evictionChecker;
-    }
+    public void evictBySize(int sizeToEvict, RecordStore recordStore) {
+        if (recordStore.size() < 1) {
```
pveentjer: if(recordStore.size()==0) ?

pveentjer:
A lot of code has been changed, but I don't see a lot of changes in test. The Evictor is a perfect class to have a lot of explicit tests on.  It is super important that eviction works correctly.

pveentjer:
Can you enhance the description of the PR to include the actual changes being made? This way it is a lot easier for future maintainers of the code to see what the purpose is of these changes.

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictorImpl.java
```diff
@@ -151,10 +164,14 @@ protected int getEvictionStartIndex(long[] criterias, int evictableSize) {
     }
 
     @Override
-    public int findRemovalSize(RecordStore recordStore) {
+    public int findSizeToEvict(RecordStore recordStore) {
+        int currentPartitionSize = recordStore.size();
```
pveentjer: Since a recordStore is always per partiton, what is the purpose if naming this variable currentPartitionSize? It is less informative than currenRecordStoreSize.

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictorImpl.java
```diff
@@ -204,4 +221,208 @@ protected long getEvictionCriteriaValue(Record record, EvictionPolicy evictionPo
...
+
+        String mapName = recordStore.getName();
+        int partitionId = recordStore.getPartitionId();
+
+        MapContainer mapContainer = recordStore.getMapContainer();
+        MaxSizeConfig maxSizeConfig = mapContainer.getMapConfig().getMaxSizeConfig();
+        MaxSizeConfig.MaxSizePolicy maxSizePolicy = maxSizeConfig.getMaxSizePolicy();
+
+        boolean result;
+        switch (maxSizePolicy) {
```
pveentjer: I would not introduce a local variable with the maxSizePolicy. Just say

'''
switch(maxSizeConfig.getMaxSizePolicy()){
 ....
}
...

It reduces the amount of noise in the method. 

Another flavor things is to not store the result in the 'result' local variable, but immediately return from each case. Saves you another 2 lines on the method.

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictorImpl.java
```diff
@@ -204,4 +221,208 @@ protected long getEvictionCriteriaValue(Record record, EvictionPolicy evictionPo
...
+                result = checkHeapSizeEviction(mapName, maxSizeConfig);
+                break;
+            case FREE_HEAP_PERCENTAGE:
+                result = checkFreeHeapPercentageEviction(maxSizeConfig);
+                break;
+            case FREE_HEAP_SIZE:
+                result = checkFreeHeapSizeEviction(maxSizeConfig);
+                break;
+            default:
+                throw new IllegalArgumentException("Not an appropriate max size policy [" + maxSizePolicy + ']');
```
pveentjer: Probably this should be IllegalStateException since it isn't an argument to a method.

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictorImpl.java
```diff
@@ -38,22 +47,26 @@
 public class EvictorImpl implements Evictor {
```
pveentjer: Would it not be an idea to EvictorImpl per RecordStore instead of this stateless object? Then you can move all eviction related logic to this class instead of having part of it in the evictable record store and another part in the evictor.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractEvictableRecordStore.java
```diff
@@ -213,42 +212,23 @@ private void cleanUp(long now) {
         if (size() == 0) {
             return;
         }
+
         if (shouldEvict(now)) {
-            evict();
-            lastEvictionTime = now;
-            readCountBeforeCleanUp = 0;
+            int sizeToEvict = evictor.findSizeToEvict(this);
```
pveentjer: This is an unclear API like I suggested above. You expose less of your implementation detail if you can say to the evictor.evict().

The evictor will keep track of stuff like lastEvictionTime, readCountBeforeCleanup etc etc. 

This way the Evictor becomes a very 'simple' abstraction with potentially complex behavior. Currently it is a bit smeared over multiple classes.

ahmetmircik:
```
com.hazelcast.client.cache.nearcache.ClientNearCacheTest.putToCacheAndThenGetFromClientNearCacheWithObjectInMemoryFormat
```

ahmetmircik:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/Evictor.java
```diff
@@ -20,32 +20,28 @@
...
-
-    /**
-     * Get eviction checker for this {@link Evictor}
+     * @return {@code true} if eviction is required, {@code false} otherwise.
      *
-     * @return the {@link EvictionChecker}
+     * @see com.hazelcast.map.impl.recordstore.DefaultRecordStore#putFromLoad
      */
-    EvictionChecker getEvictionChecker();
+    boolean isReachedMaxSize(RecordStore recordStore);
```
pveentjer: I would go for:
isMaxSizeReached or isFull.

ahmetmircik:
Closing this one, it needs some more work. I will try to send it for 3.7.

Issue #5031, opened 2015-04-09
[TEST-FAILURE]  MapTransactionTest.testTxnBackupDies


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.6/198/com.hazelcast$hazelcast/testReport/com.hazelcast.map/MapTransactionTest/testTxnBackupDies/

eminn:
Closing the issue since the test is not failing recently.

hasancelik:
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.map.MapTransactionTest.testTxnBackupDies(MapTransactionTest.java:214)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3669/testReport/junit/com.hazelcast.map/MapTransactionTest/testTxnBackupDies/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK6/com.hazelcast$hazelcast/693/testReport/junit/com.hazelcast.map/MapTransactionTest/testTxnBackupDies/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3894/testReport/junit/com.hazelcast.map/MapTransactionTest/testTxnBackupDies/

mdogan:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast/16424/testReport/junit/com.hazelcast.map/MapTransactionTest/testTxnBackupDies/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.7/com.hazelcast$hazelcast/278/testReport/junit/com.hazelcast.map/MapTransactionTest/testTxnBackupDies/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3968/testReport/junit/com.hazelcast.map/MapTransactionTest/testTxnBackupDies/

[...]

PR #6510, opened 2015-10-20
Add method to execute a Runnable on all PartitionOperationThreads

mdogan:
_Required to run a task on each partition thread while restoring hot-restart data._

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutorTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutorTest.java
@@ -9,6 +9,9 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.TimeUnit;
+
 import static org.junit.Assert.assertEquals;
 
 @RunWith(HazelcastSerialClassRunner.class)
@@ -82,5 +85,39 @@ public class ClassicOperationExecutorTest extends AbstractClassicOperationExecut
                 assertEquals(expectedCount, executor.getOperationExecutorQueueSize());
             }
         });
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void test_runOnAllPartitionThreads_whenTaskNull() {
+        initExecutor();
+        executor.runOnAllPartitionThreads(null);
+    }
+
+    @Test
+    public void test_runOnAllPartitionThreads() throws Exception {
+        initExecutor();
+
+        int threadCount = executor.getPartitionOperationThreadCount();
+        final CyclicBarrier barrier = new CyclicBarrier(threadCount + 1);
+
+        executor.runOnAllPartitionThreads(new Runnable() {
+            @Override
+            public void run() {
+                // current thread must be a PartitionOperationThread
+                if (Thread.currentThread() instanceof PartitionOperationThread) {
+                    try {
+                        awaitBarrier(barrier);
+                    } catch (Exception e) {
+                        e.printStackTrace();
+                    }
+                }
+            }
+        });
+
+        awaitBarrier(barrier);
+    }
+
+    private static void awaitBarrier(CyclicBarrier barrier) throws Exception {
+        barrier.await(ASSERT_TRUE_EVENTUALLY_TIMEOUT, TimeUnit.SECONDS);
     }
 }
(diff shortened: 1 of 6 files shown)


pveentjer:
:+1:

metanet:
:+1:

mdogan:
Added tests.

pveentjer:
:+1:

mdogan:
verify

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutorTest.java
```diff
@@ -83,4 +85,28 @@ public void run() throws Exception {
...
+        final CountDownLatch latch = new CountDownLatch(threadCount);
+
+        executor.runOnAllPartitionThreads(new Runnable() {
+            @Override
+            public void run() {
+                latch.countDown();
+            }
+        });
+
+        assertOpenEventually(latch);
```
jerrinot: perhaps it's a nitpicking, but this test only makes sure the task is executed `threadCount`-times. Not that it's executed by all threads.

mdogan:
```java
@Test
public void test_runOnAllPartitionThreads() {
    initExecutor();

    int threadCount = executor.getPartitionOperationThreadCount();
    final CountDownLatch latch = new CountDownLatch(threadCount);

    executor.runOnAllPartitionThreads(new PartitionedRunnable(latch));

    assertOpenEventually(latch);
}

private static class PartitionedRunnable implements Runnable {

    private final ThreadLocal threadLocal = new ThreadLocal() {
        @Override
        protected Object initialValue() {
            latch.countDown();
            return null;
        }
    };

    private final CountDownLatch latch;

    public PartitionedRunnable(CountDownLatch latch) {this.latch = latch;}

    @Override
    public void run() {
        if (Thread.currentThread() instanceof PartitionOperationThread) {
            threadLocal.get();
        }
    }
}
```
WDYT?

mdogan:
MapTransactionTest failed. https://github.com/hazelcast/hazelcast/issues/5031

```
com.hazelcast.map.MapTransactionTest
17:07:38 testTxnBackupDies(com.hazelcast.map.MapTransactionTest)  Time elapsed: 67.378 sec  <<< FAILURE!
17:07:38 java.lang.AssertionError: null
17:07:38 	at org.junit.Assert.fail(Assert.java:86)
17:07:38 	at org.junit.Assert.assertTrue(Assert.java:41)
17:07:38 	at org.junit.Assert.assertTrue(Assert.java:52)
17:07:38 	at com.hazelcast.map.MapTransactionTest.testTxnBackupDies(MapTransactionTest.java:252)
```

hazelcast/src/test/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutorTest.java
jerrinot: That's indeed quite clever. Now please pass me some eye bleach please! :)

I would say a `CyclicBarrier` would make your intent clearer. 
Let's create a cyclic barrier with `threadCount + 1`. and use `await` instead of `assertOpenEventually`. ?

mdogan: hmm.. makes sense.. 

```java
@Test
public void test_runOnAllPartitionThreads() throws Exception {
    initExecutor();

    int threadCount = executor.getPartitionOperationThreadCount();
    final CyclicBarrier barrier = new CyclicBarrier(threadCount + 1);

    executor.runOnAllPartitionThreads(new Runnable() {
        @Override
        public void run() {
            // current thread must be a PartitionOperationThread
            if (Thread.currentThread() instanceof PartitionOperationThread) {
                try {
                    barrier.await(ASSERT_TRUE_EVENTUALLY_TIMEOUT, TimeUnit.SECONDS);
                } catch (Exception e) {
                    e.printStackTrace();
                }
            }
        }
    });

    barrier.await(ASSERT_TRUE_EVENTUALLY_TIMEOUT, TimeUnit.SECONDS);
}
```

jerrinot: I have to admit your solution was very creative! :+1:

jerrinot: actually the inner class should use a longer timeout then the main test thread? I am not too sure, but it can't hurt. 

from CyclicBarrier contract: "`await()` [...] throws TimeoutException if the specified timeout elapses. *In this case the barrier will be broken.*"

pveentjer:
Can the commits be squashed?

PR #6511, opened 2015-10-20
Hazelcast Client Rename.

asimarslan:
Maven module rename; hazelcast-client-new to hazelcast-client and hazelcast-client to hazelcast-client-legacy

IMPORTANT NOTE: NO SOURCE CODE FILE IS UPDATED IN THIS COMMIT.

ONLY FOLDER NAMES ARE UPDATED.

The updated configuration files are pom.xml's, readme.html and assembly.xml.


pveentjer:
I don't know if this should be merged. Because due to this rename, the new client will be selected by default.

pveentjer:
Ok.. we can commit this. I missed that we are going to select the new client as default.

pveentjer:
I guess git(hub) has a hard time to due flipping of files. Because it mentions many changes. Another problem is that we could loose history or get a messed up history.

Perhaps an idea that we do the rename in 2 phase? First rename existing client to legacy... And then we rename new-client to 'client; but do this in a different pr.

asimarslan:
What about splitting this into two commits in this PR, isn't it same?

asimarslan:
The real history is; at some point of history, call it NEW_START, we copied everything to client_new from client, so maybe I should create a branch there and cherry pick every commit from client_new into that branch. Then send  PR to merge into client ??

asimarslan:
We are working on a solution not to lose history, so I will close this

PR #6518, opened 2015-10-20
refactor client to client-legacy

asimarslan:
This is the first step of renaming clients.


pveentjer:
images

pveentjer:
images

pveentjer:
@asimarslan can you rebase? We should get this PR merged ASAP so it doesn't suffer from conflicts.

asimarslan:
Two commits in single PR, will do the same trick.

pveentjer:
Now we are back to the original problem. The commit history is messed up. Can you rollback the 'client-new to client rename' and do a rebase of the'client to client legacy' rename?

asimarslan:
@pveentjer we checked it locally, git history just work fine.

mdogan:
@asimarslan: will you do the same renaming on EE?

PR #6522, opened 2015-10-21
Improved HazelcastSerializationException error message

pveentjer:
It now gives a clue what the cause is of the 'there is no suitable ....'


pveentjer:
verify

jerrinot:
:+1: 
we should have more error messages like this.

Donnerbart:
:+1:

serkan-ozal:
:+1:

PR #6449, opened 2015-10-13
Queries on collections / arrays

jerrinot:
This is still work in progress. 

I'm creating the PR to have tests after every push.

PR #6525, opened 2015-10-21
Predicate on multi-value attribute and custom attribute values extractors

jerrinot:
See the commit message for details.

This is just a squashed version of #6449 which was already reviewed.


gurbuzali:
green on the first try

PR #6526, opened 2015-10-21
fix unneccessary instance-name set in properties

mustafaiman:
Setting instance name in properties is unneccessary now as https://github.com/hazelcast/hazelcast/issues/6454 solved.


serkan-ozal:
:+1:

PR #6527, opened 2015-10-21
Cleanup xml config

mtopolnik:
This PR fixes several issues with XML config builder and related tests.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/config/InvalidConfigurationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/config/InvalidConfigurationTest.java
@@ -20,22 +20,29 @@ import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.annotation.ParallelTest;
 import com.hazelcast.test.annotation.QuickTest;
 
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
+import org.xml.sax.SAXParseException;
 
 import java.io.ByteArrayInputStream;
 import java.util.Properties;
 import java.util.Random;
 
+import static com.hazelcast.config.XMLConfigBuilderTest.HAZELCAST_START_TAG;
 import static org.junit.Assert.assertNotNull;
 
 @RunWith(HazelcastParallelClassRunner.class)
 @Category({QuickTest.class, ParallelTest.class})
 public class InvalidConfigurationTest {
+    @Rule
+    public ExpectedException rule = ExpectedException.none();
 
-    @Test(expected = InvalidConfigurationException.class)
+    @Test
     public void testWhenTwoJoinMethodEnabled() {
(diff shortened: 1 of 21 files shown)


sancar:
:+1:

hazelcast/src/test/java/com/hazelcast/config/InvalidConfigurationTest.java
```diff
@@ -109,13 +123,15 @@ public void testWhenValid_MapEvictionPolicy() {
         buildConfig("map-eviction-policy", "NONE");
     }
 
-    @Test(expected = InvalidConfigurationException.class)
+    @Test
     public void testWhenInvalid_MapEvictionPercentage() {
+        expectInvalid(" Value '101' is not facet-valid with respect to maxInclusive '100'");
```
pveentjer: Personally i'm never crazy about checking exception messages. But in this case it is probably the best thing to do.

hazelcast/src/test/java/com/hazelcast/config/XMLConfigBuilderTest.java
```diff
@@ -147,79 +124,33 @@ public void testSecurityInterceptorConfig() {
...
-                        "</hazelcast>";
+                HAZELCAST_START_TAG +
+                "   <group>\n" +
+                "        <name>dev</name>\n" +
+                "        <password>dev-pass</password>\n" +
+                "    </group>\n" +
+                "    <network>\n" +
+                "        <port auto-increment=\"true\">5701</port>\n" +
+                "        <join>\n" +
+                "            <multicast enabled=\"false\">\n" +
```
pveentjer: We should really get this test class split up so that every data-structure like IQueue, IMap gets its own test file.

pveentjer:
Looks good to me. Once you have fixed the fail fast violations with integer etc parsing, I would be happy to put my thumbs up.

bilalyasar:
for the refactoring things i am OK once green.

but for the asserting exception messages, are you sure about those messages are same for all jdk versions and compabilities.?
Because validation factory may be different in some jdks and compatibilities as default..

hazelcast/src/test/java/com/hazelcast/config/XMLConfigWithSystemPropertyTest.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.URL;
+
+import static com.hazelcast.config.XMLConfigBuilderTest.HAZELCAST_START_TAG;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+// These tests manipulate system properties, therefore they must be run in serial mode.
+@RunWith(HazelcastSerialClassRunner.class)
```
bilalyasar: missing  category

mtopolnik:
@bilalyasar I'm not sure, but if that doesn't work out then we're in trouble. The best I could then do is assert schema validation error and once again let unrelated errors creep in.

mtopolnik:
verify

enesakar:
Danny will run tests before simulator runs, so merging this

PR #6528, opened 2015-10-21
Dependency on JDK7 class removed

jerrinot:


jerrinot:
It's a trivial change in test code only. Not need to wait for PR builder, I run the affected tests manually on my local.

metanet:
why dont you send a direct commit, then?

jerrinot:
visibility.

PR #6532, opened 2015-10-21
Added size-operation-factory to MapOperationProvider

ahmetmircik:


gurbuzali:
since this is a very trivial change, only moving the creation of size-operation to `OperationProvider`, I'll merge it

PR #6535, opened 2015-10-21
UpdateEntryProcessor test on real network

mustafaiman:
Tests UpdateEntryProcessor in Hibernate. Also increases test coverage by testing toSrring methods.

### DIFF
--- a/hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/RegionFactoryDefaultSlowTest.java
+++ b/hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/RegionFactoryDefaultSlowTest.java
@@ -17,15 +17,19 @@
 package com.hazelcast.hibernate;
 
 import com.hazelcast.config.MapConfig;
+import com.hazelcast.hibernate.entity.DummyEntity;
 import com.hazelcast.hibernate.region.HazelcastQueryResultsRegion;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.annotation.SlowTest;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
 import org.hibernate.cfg.Environment;
 import org.hibernate.impl.SessionFactoryImpl;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
+import java.util.ArrayList;
 import java.util.Properties;
 
 import static org.junit.Assert.assertEquals;
@@ -61,4 +65,19 @@ public class RegionFactoryDefaultSlowTest extends HibernateSlowTestSupport {
         assertEquals(numberOfEntities - evictedItemCount, queryRegion.getCache().size());
     }
 
+    @Test
+    public void testUpdate() {
+        insertDummyEntities(1);
+        Session session = sf.openSession();
+        Transaction tx = session.beginTransaction();
+        DummyEntity ent = (DummyEntity) session.get(DummyEntity.class, 0L);
+        ent.setName("updatedName");
+        session.update(ent);
+        tx.commit();
+        session.close();
+
+        session = sf2.openSession();
+        DummyEntity entity = (DummyEntity) session.get(DummyEntity.class, 0L);
+        assertEquals("updatedName", entity.getName());
+    }
 }
--- a/hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/local/LocalRegionCacheTest.java
+++ b/hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/local/LocalRegionCacheTest.java
@@ -15,6 +15,7 @@ import org.junit.runner.RunWith;
 
 import java.util.Comparator;
 
+import static org.junit.Assert.assertNotNull;
 import static org.mockito.Mockito.*;
 
 @RunWith(HazelcastSerialClassRunner.class)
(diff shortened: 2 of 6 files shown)


pveentjer:
Can you a description? The purpose of this PR isn't clear to me based on the title.

hazelcast-hibernate/hazelcast-hibernate4/src/test/java/com/hazelcast/hibernate/local/LocalRegionCacheTest.java
```diff
@@ -94,6 +95,16 @@ public void testThreeArgConstructorRegistersTopicListener() {
         verify(topic).addMessageListener(isNotNull(MessageListener.class));
     }
 
+    @Test
+    public void testTimestampToString() {
+        assertNotNull(new Timestamp().toString());
```
pveentjer: Just testing on not null doesn't make a lot of sense to me. It leads to a false sense of security in the sense that the code seems tested, but actually there is not a lot of validation logic. Looks nice for coverage; but not a lot of actual value.

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/local/LocalRegionCacheTest.java
```diff
@@ -94,6 +95,16 @@ public void testThreeArgConstructorRegistersTopicListener() {
...
+    public void testTimestampToString() {
+        assertNotNull(new Timestamp().toString());
+    }
+
+    @Test
+    public void testInvalidationToString() {
+        assertNotNull(new Invalidation().toString());
+    }
+
     public static void runCleanup(LocalRegionCache cache) {
```
pveentjer: what is the purpose of this method since it just forwards to cache.cleanup.

pveentjer: Can you check the other toString methods as well. So either test them well by checking the content, or don't test them at all.

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/RegionFactoryDefaultSlowTest.java
```diff
@@ -61,4 +65,19 @@ public void testQueryCacheCleanup() {
         assertEquals(numberOfEntities - evictedItemCount, queryRegion.getCache().size());
     }
 
+    @Test
+    public void testUpdate() {
+        insertDummyEntities(1);
+        Session session = sf.openSession();
+        Transaction tx = session.beginTransaction();
+        DummyEntity ent = (DummyEntity) session.get(DummyEntity.class, 0L);
```
pveentjer: The test is a bit obscure because they key '0' appears out of thin air. I guess you know there is an entry created with key 0 when insertDummyEntities(1) is being called.

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/local/LocalRegionCacheTest.java
mustafaiman: I canot see any other toString method in Hibernate. Which other toString methods are you referring to?

mustafaiman: As far as I understand, `cleanup` method of `LocalRegionCache` is package-protected and it is needed by a `HazelcastQueryResultsRegionTest` in a different package. It simply delegates the task for a test which cannot access the method.

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/RegionFactoryDefaultSlowTest.java
mustafaiman: Yes it is like you said. I added JavaDoc comments to two `insertDummyEntities` methods with different signatures explaining their behavior.

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/local/LocalRegionCacheTest.java
pveentjer: I was refering to the other toString test methods. They don't make much sense since they just check for not null.

hazelcast-hibernate/hazelcast-hibernate3/src/test/java/com/hazelcast/hibernate/RegionFactoryDefaultSlowTest.java
pveentjer: I would introduce a consistant for it. e.g start id. 

DummyEntity ent = (DummyEntity) session.get(DummyEntity.class, START_ID);

pveentjer: Or even better.. let the insert method return a list of the id's inserted. This way all this information is directly available instead through some global state.

int[] ids = insertDummyEnties(1)

Or.. simpler:

int id = insertDummyEntity();

mustafaiman:
Thanks for the review. I will send another PR with changes in other tests that use `insertDummyEntities` method.

PR #6543, opened 2015-10-22
Fix for `CachePartitionSegment` clear and close

serkan-ozal:
In `CachePartitionSegment` , `ICacheRecordStore` should not be removed from internal `recordStores` map if it is not destroyed otherwise it causes to memory leak since we lost its reference but its storage memory is not disposed.

Fixes https://github.com/hazelcast/hazelcast-enterprise/issues/340

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
@@ -70,14 +70,17 @@ public class CachePartitionSegment implements ConstructorFunction<String, ICache
     }
 
     public void deleteRecordStore(String name, boolean destroy) {
-        ICacheRecordStore store = recordStores.remove(name);
-        if (store == null) {
-            return;
-        }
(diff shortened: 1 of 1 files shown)


ahmetmircik:
Is it testable?

serkan-ozal:
tested at enterprise and fixes hazelcast/hazelcast-enterprise#340

ahmetmircik:
ok

ahmetmircik:
:+1:  good catch. Seems mergeable.

ahmetmircik:
What about porting it to EA? @enesakar  @serkan-ozal WDYT?

serkan-ozal:
I think this should be there

enesakar:
yes please port it to 3.6-EA

ahmetmircik:
i will port

ahmetmircik:
done https://github.com/hazelcast/hazelcast/commit/0bfc9a624c23117a941f53f9c0ddad3bc01df158

serkan-ozal:
Thanks @ahmetmircik

PR #6552, opened 2015-10-23
Hot Restart Requests for mancenter

muratayan:
This PR makes mancenter enable to manage Hot Restart events such as `GetState` `ChangeState` and `Shutdown`

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.management.request;
+
+import com.eclipsesource.json.JsonObject;
+import com.hazelcast.cluster.ClusterState;
+import com.hazelcast.core.Cluster;
+import com.hazelcast.internal.management.ManagementCenterService;
+import com.hazelcast.logging.ILogger;
+
+import java.io.IOException;
+
+import static com.hazelcast.util.JsonUtil.getString;
+
+/**
+ * Request coming from Management Center for changing the {@link ClusterState}
+ */
+public class ChangeClusterStateRequest implements AsyncConsoleRequest {
+
+    private String state;
+
+    public ChangeClusterStateRequest() {
+    }
+
+    public ChangeClusterStateRequest(String state) {
+        this.state = state;
+    }
+
+    @Override
+    public int getType() {
+        return ConsoleRequestConstants.REQUEST_TYPE_CHANGE_CLUSTER_STATE;
+    }
+
+    @Override
+    public Object readResponse(JsonObject in) throws IOException {
+        return getString(in, "result", "FAIL");
+    }
+
+    @Override
+    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
+        String resultString = "SUCCESS";
+        try {
+            Cluster cluster = mcs.getHazelcastInstance().getCluster();
+            cluster.changeClusterState(getClusterState(state));
+        } catch (Exception e) {
+            ILogger logger = mcs.getHazelcastInstance().node.nodeEngine.getLogger(getClass());
+            logger.warning("Cluster state can not be changed: ", e);
+            resultString = e.getMessage();
+        }
+        JsonObject result = new JsonObject().add("result", resultString);
+        out.add("result", result);
+    }
+
+    private static ClusterState getClusterState(String state) {
+        return ClusterState.valueOf(state);
+    }
+
+    @Override
+    public JsonObject toJson() {
+        JsonObject root = new JsonObject();
+        root.add("state", state);
+        return root;
+    }
+
+    @Override
+    public void fromJson(JsonObject json) {
+        state = getString(json, "state");
+    }
+}
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/request/GetClusterStateRequest.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/request/GetClusterStateRequest.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.management.request;
+
+import com.eclipsesource.json.JsonObject;
+import com.hazelcast.cluster.ClusterState;
+import com.hazelcast.internal.management.ManagementCenterService;
+
+import java.io.IOException;
+
+import static com.hazelcast.util.JsonUtil.getString;
+
+/**
+ * Request coming from Management Center for getting the {@link ClusterState}
+ */
+public class GetClusterStateRequest implements ConsoleRequest {
+
+    @Override
+    public int getType() {
+        return ConsoleRequestConstants.REQUEST_TYPE_GET_CLUSTER_STATE;
+    }
+
+    @Override
+    public Object readResponse(JsonObject in) throws IOException {
+        return getString(in, "result", "UNKNOWN");
+    }
+
+    @Override
+    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
+        ClusterState clusterState = mcs.getHazelcastInstance().getCluster().getClusterState();
+        JsonObject result = new JsonObject();
+        result.add("result", clusterState.toString());
+        out.add("result", result);
+    }
+
+    @Override
+    public JsonObject toJson() {
+        return new JsonObject();
+    }
+
+    @Override
+    public void fromJson(JsonObject json) {
+    }
+}
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/request/ShutdownClusterRequest.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/request/ShutdownClusterRequest.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.management.request;
+
+import com.eclipsesource.json.JsonObject;
+import com.hazelcast.internal.management.ManagementCenterService;
+import com.hazelcast.logging.ILogger;
+
+import java.io.IOException;
+
+import static com.hazelcast.util.JsonUtil.getString;
+
+/**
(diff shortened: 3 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
```diff
@@ -0,0 +1,64 @@
...
+        return getString(in, "result", "FAIL");
+    }
+
+    @Override
+    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
+        JsonObject result = new JsonObject();
+        String resultString = "SUCCESS";
+        try {
+            if (state.equalsIgnoreCase("ACTIVE")) {
+                mcs.getHazelcastInstance().getCluster().changeClusterState(ClusterState.ACTIVE);
```
mdogan: You can pull out `mcs.getHazelcastInstance().getCluster()` to a local var, that will make code cleaner to eyes.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
```diff
@@ -0,0 +1,64 @@
...
+
+            } else if (state.equalsIgnoreCase("FROZEN")) {
+                mcs.getHazelcastInstance().getCluster().changeClusterState(ClusterState.FROZEN);
+
+            } else if (state.equalsIgnoreCase("PASSIVE")) {
+                mcs.getHazelcastInstance().getCluster().changeClusterState(ClusterState.PASSIVE);
+            }
```
mdogan: Will you see failure on management center if state doesn't match any of the above?

muratayan: Yes. However, that case is not possible under normal circumstances since it comes directly from the UI.

Only way is to someone spoof the call to Management Center server somehow, which ends in a failure message.

muratayan:
Updated

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
```diff
@@ -0,0 +1,84 @@
...
+        return ConsoleRequestConstants.REQUEST_TYPE_CHANGE_CLUSTER_STATE;
+    }
+
+    @Override
+    public Object readResponse(JsonObject in) throws IOException {
+        return getString(in, "result", "FAIL");
+    }
+
+    @Override
+    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
```
pveentjer: This method name or content is very confusing. Write response doesn't only write the response, it actually does the execution of the logic. Having this method called run/execute/process etc is more revealing.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
```diff
@@ -0,0 +1,84 @@
...
+        JsonObject result = new JsonObject();
+        String resultString = "SUCCESS";
+        Cluster cluster = mcs.getHazelcastInstance().getCluster();
+        try {
+            if (state.equalsIgnoreCase("ACTIVE")) {
+                cluster.changeClusterState(ClusterState.ACTIVE);
+            } else if (state.equalsIgnoreCase("FROZEN")) {
+                cluster.changeClusterState(ClusterState.FROZEN);
+
```
pveentjer: Why this empty line?

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
muratayan: I understand, however this needs a rather big refactoring since it is used this way in all other request classes. We can open an issue for it.

muratayan: https://github.com/hazelcast/hazelcast/issues/6559

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
```diff
@@ -0,0 +1,84 @@
...
+            } else if (state.equalsIgnoreCase("PASSIVE")) {
+                cluster.changeClusterState(ClusterState.PASSIVE);
+            }
+        } catch (Exception e) {
+            resultString = e.toString();
+        }
+        result.add("result", resultString);
+        out.add("result", result);
+    }
+
```
pveentjer: I would pull out a method that converts the cluster state like this:

```
  @Override
    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
        JsonObject result = new JsonObject();
        String resultString = "SUCCESS";
        try {
    	   Cluster cluster = mcs.getHazelcastInstance().getCluster();
           cluster.setClusterState(getClusterState(state));		
        } catch (Exception e) {
            resultString = e.toString();
        }
        result.add("result", resultString);
        out.add("result", result);
    }

	
   private static ClusterState getClusterState(String state){
	if ("ACTIVE".equalsIgnoreCase()) {
           return ClusterState.ACTIVE;
        } else if ("FROZEN".equalsIgnoreCase()) {
              return ClusterState.FROZEN;
        } else if("PASSIVE".equalsIgnoreCase(state)){
               return ClusterState.PASSIVE;
        }else{
           throw new IllegalArgumentExcepton(...);
        }
   }
```
If the state is not found; don't ignore the call, but throw an exception. Because this is a problematic situation.

Also when an exception is thrown, you want to log it. Currently it is gobbled up. This makes debugging a lot more complicated.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ShutdownClusterRequest.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
+        JsonObject result = new JsonObject();
+        String resultString = "SUCCESS";
+        try {
+            mcs.getHazelcastInstance().getCluster().shutdown();
+        } catch (Exception e) {
+            resultString = e.toString();
```
pveentjer: Exception gobbling.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ShutdownClusterRequest.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public JsonObject toJson() {
+        return new JsonObject();
+    }
+
+    @Override
+    public void fromJson(JsonObject json) {
+
```
pveentjer: Why this empty line?

pveentjer:
There is no test for this behavior. This logic should be tested in the core module.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
pveentjer: Afaik no big change; just rename the method writeResponse to process or execute or something.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
pveentjer: Another change I would make is that currently yuo have copy pasted the names of the ClusterState fields; just use the .name() method from the ClusterState to get the strings. Copy paste = bad.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
muratayan: I still think this should be a separate PR. I will do it this afternoon.

pveentjer: Anyhow.. not important now. I just got confused understanding the logic.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
pveentjer: And even better it to make use of the ClusterState.valueOf. No need to do the if/else.

hazelcast/src/main/java/com/hazelcast/internal/management/request/ChangeClusterStateRequest.java
```diff
@@ -0,0 +1,93 @@
...
+        } catch (Exception e) {
+            logger.warning("Cluster state can not be changed: " + e);
+            resultString = e.toString();
+        }
+        result.add("result", resultString);
+        out.add("result", result);
+    }
+
+
+    private static ClusterState getClusterState(String state) {
```
pveentjer: You could have a look at ClusterState.valueOf(String name). This way you can remove the if else range. Also adding new cluster states gets immediately resolved since you don't need to add any code here.

hazelcast/src/main/java/com/hazelcast/internal/management/request/GetClusterStateRequest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public JsonObject toJson() {
+        return new JsonObject();
+    }
+
+    @Override
+    public void fromJson(JsonObject json) {
+
```
pveentjer: Why this empty line?

hazelcast/src/main/java/com/hazelcast/internal/management/request/ShutdownClusterRequest.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
+        ILogger logger = mcs.getHazelcastInstance().node.nodeEngine.getLogger(getClass());
+        JsonObject result = new JsonObject();
+        String resultString = "SUCCESS";
+        try {
+            mcs.getHazelcastInstance().getCluster().shutdown();
+        } catch (Exception e) {
+            resultString = e.toString();
+            logger.warning("Cluster can not be shutdown: " + e);
```
pveentjer: You need to pass the exception using

```
logger.warning("Cluster failed to shutdown", e);
```
Otherwise you will loose the exception stacktrace.

pveentjer: I would clean it up like this:

```
 @Override
    public void writeResponse(ManagementCenterService mcs, JsonObject out) throws Exception {
        String resultString = "SUCCESS";
        try {
            mcs.getHazelcastInstance().getCluster().shutdown();
        } catch (Exception e) {
            ILogger logger = mcs.getHazelcastInstance().node.nodeEngine.getLogger(getClass());
            logger.warning("Cluster can not be shutdown: " + e);
            resultString = e.getMessage();
         }

        JsonObject result = new JsonObject();
        result.add("result", resultString);
        out.add("result", result);
    }
```

You can see that the result also has been moved to the place where it is needed.

pveentjer:
Looks good.. but I'm missing tests.

muratayan:
Updated & Added tests

hazelcast/src/test/java/com/hazelcast/internal/management/ShutdownClusterRequestTest.java
```diff
@@ -0,0 +1,70 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class ShutdownClusterRequestTest extends HazelcastTestSupport {
+
+    final String SUCCESS = "SUCCESS";
+
+    HazelcastInstance hz;
```
pveentjer: Can you make these fields private and the field above as well.

hazelcast/src/test/java/com/hazelcast/internal/management/GetClusterStateRequestTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class GetClusterStateRequestTest extends HazelcastTestSupport{
+
+    HazelcastInstance hz;
```
pveentjer: Can you make these fields private?

hazelcast/src/test/java/com/hazelcast/internal/management/ShutdownClusterRequestTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                assertFalse(lifecycleService.isRunning());
+            }
+        });
+    }
+
+    @After
```
pveentjer: Can you move this below the setup.

hazelcast/src/test/java/com/hazelcast/internal/management/GetClusterStateRequestTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public void testGetClusterState() throws Exception {
+        ClusterState clusterState = cluster.getClusterState();
+        JsonObject jsonObject = new JsonObject();
+        getClusterStateRequest.writeResponse(managementCenterService, jsonObject);
+
+        JsonObject result = (JsonObject) jsonObject.get("result");
+        assertEquals(clusterState.name(), getClusterStateRequest.readResponse(result));
+    }
+
+    @After
```
pveentjer: I don't think this  is needed. HZ instances are terminated anyway because it is part of HazelcastTestSupport.

pveentjer:
:+1: once green

pveentjer:
verify

pveentjer:
verify

PR #6555, opened 2015-10-24
ClientMapProxy.getAll to invoke-by-partition

### LINKED ISSUE #6083: MapProxySupport.putAllInternal should use setInternal, ClientMapProxy.putAll/getAll should invoke on key owners and not random target

LoneRifle:
* Convert MapGetAllRequest/MessageTask to ClientPartitionRequest to allow invocation-by-partition

* Re-implement ClientMapProxy.getAll by porting code from putAll, so that keys are grouped by corresponding partitions. For each key set, send a request to the partition owner. This approach has two advantages:
  * Allows the client to gather data directly from owners rather than depend on a randomly-chosen node
  * Because the requests are made asynchronously, the client can work on caching and populating the return value with some of the data while waiting for the other data to be fully received.

This fixes #6083, and is a rebase of #6135

Previously reviewed by @sancar and @cangencer.

Create a gist found at https://gist.github.com/LoneRifle/ff9c6c658327ffe39e79 documenting ad-hoc benchmark.
I obtained old vs new times for 1, 2, 4 nodes to getAll integer keys mapping to randomly-generated UUIDs for key counts in increasing powers of two up till 2^20.

I expect that similar timings will be obtained for getAlls on fewer keys but more expensive values.

ClientMapProxy.getAll - old vs new

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
@@ -47,6 +47,8 @@ import static com.hazelcast.cache.impl.nearcache.NearCache.NULL_OBJECT;
 import static com.hazelcast.util.MapUtil.createHashMap;
 import static java.util.Collections.emptyMap;
 
+import java.util.ArrayList;
+
 /**
  * A Client-side {@code IMap} implementation which is fronted by a near-cache.
  *
@@ -55,7 +57,7 @@ import static java.util.Collections.emptyMap;
  */
 public class NearCachedClientMapProxy<K, V> extends ClientMapProxy<K, V> {
 
-    protected NearCache nearCache;
+    protected NearCache<Data, Object> nearCache;
     protected volatile String invalidationListenerId;
 
     public NearCachedClientMapProxy(String serviceName, String name) {
@@ -235,24 +237,36 @@ public class NearCachedClientMapProxy<K, V> extends ClientMapProxy<K, V> {
     }
 
     @Override
-    protected MapEntries getAllInternal(List<Data> keys, Map<K, V> result) {
-        Iterator<Data> iterator = keys.iterator();
-        while (iterator.hasNext()) {
-            Data key = iterator.next();
-            Object cached = nearCache.get(key);
-            if (cached != null && NULL_OBJECT != cached) {
-                result.put((K) toObject(key), (V) cached);
-                iterator.remove();
+    protected List<MapEntries> getAllInternal(Map<Integer, List<Data>> partitionToKeyData, Map<K, V> result) {
+        List<Integer> partitionsWithCachedEntries = new ArrayList<Integer>(partitionToKeyData.size());
+        for (Entry<Integer, List<Data>> partitionKeyEntry : partitionToKeyData.entrySet()) {
+            List<Data> keyList = partitionKeyEntry.getValue();
+            Iterator<Data> iterator = keyList.iterator();
+            while (iterator.hasNext()) {
+                Data key = iterator.next();
+                Object cached = nearCache.get(key);
+                if (cached != null && NULL_OBJECT != cached) {
+                    result.put((K) toObject(key), (V) cached);
+                    iterator.remove();
+                }
+            }
+            if (keyList.isEmpty()) {
+                partitionsWithCachedEntries.add(partitionKeyEntry.getKey());
             }
         }
 
-        MapEntries entries = super.getAllInternal(keys, result);
+        for (Integer partitionId : partitionsWithCachedEntries) {
+            partitionToKeyData.remove(partitionId);
+        }
 
-        for (Entry<Data, Data> entry : entries.entries()) {
-            nearCache.put(entry.getKey(), entry.getValue());
+        List<MapEntries> responses = super.getAllInternal(partitionToKeyData, result);
+        for (MapEntries entries : responses) {
+            for (Entry<Data, Data> entry : entries.entries()) {
+                nearCache.put(entry.getKey(), entry.getValue());
+            }
         }
         //TODO: This returned value is not be used.
-        return entries;
+        return responses;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


pveentjer:
@sancar can you have a look at this one

sancar:
:+1: Looks good. 
@ihsandemir, can you review and merge this one asap?  And can you also check if protocol documentation of putAll and getAll needs change. Since we are changing the behaviour of a request, we need to describe it clearly on javadoc. 
We are going to need to made the related changes in C# and C++ @ihsandemir @cangencer .

hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
```diff
@@ -235,24 +237,36 @@ protected void loadAllInternal(boolean replaceExistingValues, List<Data> dataKey
...
+        List<MapEntries> responses = super.getAllInternal(partitionToKeyData, result);
+        for (MapEntries entries : responses) {
+            for (Entry<Data, Data> entry : entries.entries()) {
+                nearCache.put(entry.getKey(), entry.getValue());
+            }
         }
         //TODO: This returned value is not be used.
-        return entries;
+        return responses;
     }
```
cangencer: we are potentially retrieving items here which are already cached if they belong to a partition in which some items are cached and some not. What is the reason for changing this method from it's default behaviour (only retrieve non-cached items)

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -961,25 +961,53 @@ protected void loadAllInternal(boolean replaceExistingValues, Set<Data> dataKeys
...
-        ClientMessage response = invoke(request);
-        return MapGetAllCodec.decodeResponse(response);
+    
+    protected List<MapGetAllCodec.ResponseParameters> getAllInternal(Map<Integer, List<Data>> partitionToKeyData, Map<K, V> result) {
+        List<Future<ClientMessage>> futures = new ArrayList<Future<ClientMessage>>(partitionToKeyData.size());
+        List<MapGetAllCodec.ResponseParameters> responses = new ArrayList<MapGetAllCodec.ResponseParameters>(partitionToKeyData.size());
+        
+        for (final Map.Entry<Integer, List<Data>> entry : partitionToKeyData.entrySet()) {
+            int partitionId = entry.getKey();
+            List<Data> keyList = entry.getValue();
```
cangencer: should we check if the list is empty here?

LoneRifle: getAll guarantees that the list is non-empty in lines 968 to 975.

cangencer:
:+1: looks good to me

noctarius:
verify

hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
```diff
@@ -235,24 +237,36 @@ protected void loadAllInternal(boolean replaceExistingValues, List<Data> dataKey
...
+            Iterator<Data> iterator = keyList.iterator();
+            while (iterator.hasNext()) {
+                Data key = iterator.next();
+                Object cached = nearCache.get(key);
+                if (cached != null && NULL_OBJECT != cached) {
+                    result.put((K) toObject(key), (V) cached);
+                    iterator.remove();
+                }
+            }
+            if (keyList.isEmpty()) {
```
ihsandemir: Do we really need to check for isEmpty condition? Lets say for partition 5 we have 10 entries and 9 of them are in cache, then all we want is to refresh the remaining 1 item, but in you implementation, it seems like we are trying to retrieve all 10 entries from the map for refreshing the cache. This is a change in behaviour. We can change it to refresh only the 1 item instead of the 9 items to keep the behaviour the same.

cangencer: I think this method could be simplified by leaving it as before and just adding a list.isEmpty() check on the getAllInternal call in the map. So you iterate over the map once, remove the cached entries, and then call the method on the base class. The super.getAllInternal will ensure that it doesn't make any requests for partitions with no keys.

ihsandemir:
@sancar Added a section "Performance considerations" to the implementation guide regarding this implementation.

ihsandemir:
@sancar Added the javadoc https://github.com/hazelcast/hazelcast/pull/6708. Also added a performance consideration section in the implementation guide about this.

ihsandemir:
:+1:

hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
LoneRifle: So you're suggesting that the isEmpty check be moved to the base method? I agree with that, since we avoid having to manipulate partitionToKeyData later on in 258-260. Give me a day to rework the commit.

ihsandemir: @LoneRifle I saw your comment after the merge. Can you send this change in a separate PR?

LoneRifle: Aye, sir.

Issue #3341, opened 2014-08-21
[TEST-FAILURE] ClientMultiMapListenerStressTest.listenerAddStressTest


bilalyasar:
```
java.lang.AssertionError: Timeout waiting for thread Thread-52 to terminate
	at org.junit.Assert.fail(Assert.java:88)
	at com.hazelcast.test.HazelcastTestSupport.assertJoinable(HazelcastTestSupport.java:149)
	at com.hazelcast.client.multimap.ClientMultiMapListenerStressTest.listenerAddStressTest(ClientMultiMapListenerStressTest.java:70)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast-client/280/testReport/junit/com.hazelcast.client.multimap/ClientMultiMapListenerStressTest/listenerAddStressTest/

hasancelik:
```
java.lang.AssertionError: expected:<3200> but was:<6400>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.multimap.ClientMultiMapListenerStressTest$PutItemsThread$1.run(ClientMultiMapListenerStressTest.java:108)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:827)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:841)
	at com.hazelcast.client.multimap.ClientMultiMapListenerStressTest$PutItemsThread.assertResult(ClientMultiMapListenerStressTest.java:105)
	at com.hazelcast.client.multimap.ClientMultiMapListenerStressTest.assertReceivedEventsSize(ClientMultiMapListenerStressTest.java:79)
	at com.hazelcast.client.multimap.ClientMultiMapListenerStressTest.listenerAddStressTest(ClientMultiMapListenerStressTest.java:73)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast-client/748/testReport/junit/com.hazelcast.client.multimap/ClientMultiMapListenerStressTest/listenerAddStressTest/

PR #6560, opened 2015-10-26
Added hashCode and equals methods to ClientEventRegistration 

### LINKED ISSUE #6545: [TEST-FAILURE] ClientReplicatedMapListenerTest.testEntryUpdated
```
java.lang.AssertionError: expected:<1> but was:<2>
	at com.hazelcast.client.replicatedmap.ClientReplicatedMapListenerTest$2.run(ClientReplicatedMapListenerTest.java:79)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:827)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:841)
	at com.hazelcast.client.replicatedmap.ClientReplicatedMapListenerTest.testEntryUpdated(ClientReplicatedMapListenerTest.java:76)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK7/com.hazelcast$hazelcast-client/696/testReport/junit/com.hazelcast.client.replicatedmap/ClientReplicatedMapListenerTest/testEntryUpdated/

eminn:
since it is used in a set to decide whether registration is already made or not. Tries to fix #6545 and #3341

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/ClientEventRegistration.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/ClientEventRegistration.java
@@ -18,6 +18,8 @@ package com.hazelcast.client.spi.impl.listener;
 
 import com.hazelcast.nio.Address;
 
+import static com.hazelcast.util.Preconditions.isNotNull;
+
 /**
  * Keeps the information related to to an event registration made by clients
  */
@@ -29,6 +31,7 @@ public class ClientEventRegistration {
 
     public ClientEventRegistration(String serverRegistrationId,
                                    int callId, Address subscriber) {
+        isNotNull(serverRegistrationId, "serverRegistrationId");
         this.serverRegistrationId = serverRegistrationId;
         this.callId = callId;
         this.subscriber = subscriber;
@@ -65,6 +68,26 @@ public class ClientEventRegistration {
      */
     public int getCallId() {
         return callId;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        ClientEventRegistration that = (ClientEventRegistration) o;
+
+        return serverRegistrationId.equals(that.serverRegistrationId);
+
+    }
+
+    @Override
+    public int hashCode() {
+        return serverRegistrationId.hashCode();
     }
 
 }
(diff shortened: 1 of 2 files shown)


hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/ClientEventRegistration.java
```diff
@@ -67,5 +67,27 @@ public int getCallId() {
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        ClientEventRegistration that = (ClientEventRegistration) o;
+
+        return !(serverRegistrationId != null
```
pveentjer: Can serverRegistrationId ever be not null? If it can't be null, add the null check here and add a null check using Preconditions in the constructor.

pveentjer:
Since you are trying to solve a problem; in most cases it is better to make a test that can consistently reproduce it. In some cases it is very difficult, so I just add some delays in the system to force certain behavior. But at least a test can be made so you have some idea if the fix really fixes the problem.

eminn:
This PR fixes a problem regardless of test failure.

Consequences of missing equals/hashcode method  is that same registration id for listener could be added to the set twice, and this could lead to the duplicate events which can be seen from the test failure.

pveentjer:
Sure. But if there is a presumed bug, probably it is best to expose the bug in a test before fixing it.

eminn:
I am not against writing a test case, the reason this PR is missing a test is the case is so hard to reproduce.

pveentjer:
Ok. can you have a look at the initial comments regarding registration id afaik not possible to be null? So no need to add complexity to equals/hash.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/ClientEventRegistration.java
```diff
@@ -29,6 +31,7 @@
 
     public ClientEventRegistration(String serverRegistrationId,
                                    int callId, Address subscriber) {
+        isNotNull(serverRegistrationId, "serverRegistrationId");
```
pveentjer: It can be simplified:

this.serverRegistrationId = checkNotNull(serverRegistrationId, "serverRegistrationId");

But doesn't need to be changed now.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/ClientEventRegistration.java
```diff
@@ -67,5 +70,25 @@ public int getCallId() {
...
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        ClientEventRegistration that = (ClientEventRegistration) o;
+
+        return serverRegistrationId.equals(that.serverRegistrationId);
+
```
pveentjer: Why this empty line?

pveentjer:
Looks good. Minor comment. 

:+1:

PR #6561, opened 2015-10-26
Introduce the 'fine' logging level

mtopolnik:
The Hot Restart module needs a distinction between `debug` and `trace` logging levels. Currently we have `finest` which maps to `debug`. This PR changes `finest` to `trace` and introduces `fine` which now maps to `debug`.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/logging/NoLogFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/logging/NoLogFactory.java
@@ -27,70 +27,60 @@ public class NoLogFactory implements LoggerFactory {
     }
 
     static class NoLogger implements ILogger {
-        @Override
-        public void finest(String message) {
+        @Override public void finest(String message) {
         }
 
-        @Override
-        public void finest(String message, Throwable thrown) {
+        @Override public void finest(String message, Throwable thrown) {
         }
 
-        @Override
-        public void finest(Throwable thrown) {
+        @Override public void finest(Throwable thrown) {
         }
 
-        @Override
-        public boolean isFinestEnabled() {
+        @Override public boolean isFinestEnabled() {
             return false;
         }
 
-        @Override
-        public void info(String message) {
+        @Override public void fine(String message) {
         }
 
-        @Override
-        public void severe(String message) {
+        @Override public boolean isFineEnabled() {
+            return false;
         }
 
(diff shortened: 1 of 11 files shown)


hazelcast/src/main/java/com/hazelcast/logging/NoLogFactory.java
```diff
@@ -27,70 +27,26 @@ public ILogger getLogger(String name) {
...
-        public void finest(String message, Throwable thrown) {
-        }
-
-        @Override
-        public void finest(Throwable thrown) {
-        }
-
-        @Override
-        public boolean isFinestEnabled() {
+        @Override public void finest(String message) { }
```
pveentjer: I don't think Checkstyle is going to appreciate this code.

pveentjer:
Apart from style related issue, :+1: from me.

mtopolnik:
I checked it before submitting. Do you object to the style, should we change the rules? Personally I prefer boilerplate empty methods to be condensed because I don't have to scroll through three screenfuls to ensure that there's no interesting  code somewhere in there.

pveentjer:
IntelliJ can collapse empty methods and through the system we don't use this single line approach. That is one of the reasons we introduced checkstyle so that code is consistent.

Donnerbart:
Since the PR builder is already running tests CheckStyle doesn't seem to care. But I don't like these single-line empty methods either, especially the annotation in the same line :( If you want I can have a look at the CheckStyle config regarding this later.

Beside that the change seems to be fine (although I didn't to a very detailed review, just was triggered by the CheckStyle discussion).

pveentjer:
I just tried my local auto format in IntelliJ and this:

```
  @Override public void log(Level level, String message) {}
``

is immediately converted to:
```
  @Override
  public void log(Level level, String message) {
  }
```

**Comment by mtopolnik** [comment]
@pveentjer That's _your_ autoformatter :) We have to enforce this through CheckStyle, otherwise it's all about personal preferences.

**Comment by Donnerbart** [comment]
I also think this is the style we should go for:
```
  @Override
  public void log(Level level, String message) {
  }
```
I will try if we can configure CheckStyle accordingly.

mtopolnik:
All checks have passed. At this point I can force-push a style change, but will merge without waiting again.

PR #6568, opened 2015-10-26
Moved Maven SonarQube plugin from profiles to build section

Donnerbart:
Since there is no further configuration we can safely move the plugin to the most general section. This should fix the Sonar build, since the plugin version should be takes for all profiles and builds.


hasancelik:
:+1:

Donnerbart:
This solution seems to work for EE, so I'll merge it here as well.

PR #6569, opened 2015-10-26
Fixes for Long/IntHashSet

mtopolnik:
1. The set entered an infinite loop when full; now works correctly when full and `add()`
   throws `IllegalStateException` if capacity is exceeded.
1. The set did not ensure a good load factor (could be almost 1 in the worst case).
2. `toArray()` copied "missing value" sentinels into the array as if they were set members.
3. Also fixed a legal issue: a copyright line in QuickMath.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/QuickMath.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/QuickMath.java
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ * Portions Copyright 2014 Real Logic Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


pveentjer:
:+1:

hazelcast/src/main/java/com/hazelcast/util/QuickMath.java
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ * Portions Copyright 2014 Real Logic Ltd.
```
mdogan: For which part?

mtopolnik: Our friend `nextPowerOfTwo` :)

mdogan:
👍

PR #6574, opened 2015-10-27
Initial partition state improvements

mdogan:
- When initial partition assignment is triggered, partition listeners
should be called, as expected.
- When initial partition table is restored explicitly, no event should
be fired. Otherwise, it causes redundant backup replica syncs.
- Added tests to verify these behaviours.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/partition/impl/InternalPartitionServiceImplTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/partition/impl/InternalPartitionServiceImplTest.java
@@ -16,8 +16,10 @@
 
 package com.hazelcast.partition.impl;
 
+import com.hazelcast.cluster.ClusterState;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
+import com.hazelcast.nio.Address;
 import com.hazelcast.partition.InternalPartitionService;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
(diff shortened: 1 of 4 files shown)


metanet:
:+1:

hazelcast/src/test/java/com/hazelcast/partition/impl/InternalPartitionServiceImplTest.java
```diff
@@ -27,16 +29,106 @@
...
+        partitionService.firstArrangement();
         partitionService.getPartitionOwnerOrWait(0);
     }
-}
\ No newline at end of file
+
+    @Test
+    public void test_setInitialState() {
+        Address thisAddress = getNode(instance).getThisAddress();
+        int partitionCount = partitionService.getPartitionCount();
```
pveentjer: I would pull out these 2 fields into the setup method so that your tests remain a bit cleaner (shorter).

hazelcast/src/test/java/com/hazelcast/partition/impl/InternalPartitionServiceImplTest.java
```diff
@@ -27,16 +29,106 @@
...
+    @Test(expected = IllegalStateException.class)
+    public void test_setInitialState_multipleTimes() {
+        Address thisAddress = getNode(instance).getThisAddress();
+        int partitionCount = partitionService.getPartitionCount();
+        Address[][] addresses = new Address[partitionCount][MAX_REPLICA_COUNT];
+        for (int i = 0; i < partitionCount; i++) {
+            addresses[i][0] = thisAddress;
+        }
+
+        InternalPartitionServiceImpl partitionServiceImpl = (InternalPartitionServiceImpl) partitionService;
```
pveentjer: I would do this in the setup as well. Just make a field partitionService of type InternalPartitionServiceImpl and do the casting in the setup.

gurbuzali:
merging this one, mehmet can address the comments with direct commits, no need to trigger another build

mdogan:
done https://github.com/hazelcast/hazelcast/commit/9f900099ac2f36e7b410758e6470b6d914862fbd

PR #6577, opened 2015-10-27
Serialization binary format change to match 8 byte memory alignment

asimarslan:
Design comment from @mdogan :
>To support SPARC like cpus we will need to avoid unaligned memory access. This is handled by JVM >for heap memory but in off-heap region, this is something we have to handle ourselves. 
>If you use the same layout for offheap too then I suggest you to move Partition hash above byte[]. That >way Partition hash access will be guaranteed to be aligned. Also we will have 8 bytes before byte[] , >which will allow 8 bytes alignment in byte[]. 
>Even we can force byte[] to be 8 bytes aligned by adding padding at the end if needed.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java
@@ -121,11 +121,12 @@ public abstract class AbstractSerializationService
         BufferObjectDataOutput out = pool.takeOutputBuffer();
         try {
             SerializerAdapter serializer = serializerFor(obj);
-            out.writeInt(serializer.getTypeId(), ByteOrder.BIG_ENDIAN);
-            serializer.write(out, obj);
-
             int partitionHash = calculatePartitionHash(obj, strategy);
             out.writeInt(partitionHash, ByteOrder.BIG_ENDIAN);
+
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java
```diff
@@ -121,11 +121,12 @@ public final Data toData(Object obj, PartitioningStrategy strategy) {
         BufferObjectDataOutput out = pool.takeOutputBuffer();
         try {
             SerializerAdapter serializer = serializerFor(obj);
-            out.writeInt(serializer.getTypeId(), ByteOrder.BIG_ENDIAN);
-            serializer.write(out, obj);
-
             int partitionHash = calculatePartitionHash(obj, strategy);
             out.writeInt(partitionHash, ByteOrder.BIG_ENDIAN);
+
```
pveentjer: I\m fine with changing the order. But I'm curious about the reasoning. Personally I like it in the sense that it becomes very easy to gather all the meta data like partition id since the payload is at the end.

asimarslan:
@mdogan : is this change meet the 8 byte alignment requirements ? Can you review this?

pveentjer:
:+1:  from me. Probably good to have Mehmet have a look as well.

mdogan:
:+1:

PR #6581, opened 2015-10-27
Introduce hot-restart config

mdogan:
- Added initial support for Hot Restart configuration
- Added hot-restart to hazelcast and spring xml schema
- Added cache hot-restart-enabled flag to hazelcast and spring xml schema
- Added to ConfigXmlGenerator

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -98,6 +98,7 @@
     <suppress checks="CyclomaticComplexity" files="com/hazelcast/config/AbstractXmlConfigHelper"/>
     <suppress checks="ClassDataAbstractionCoupling" files="com/hazelcast/config/AbstractXmlConfigHelper"/>
     <suppress checks="FanOutComplexity" files="com/hazelcast/config/ConfigXmlGenerator"/>
+    <suppress checks="ClassFanOutComplexity" files="com/hazelcast/config/Config"/>
 
     <!-- Couldn't change structure because of API -->
     <suppress checks="CyclomaticComplexity" files="com/hazelcast/config/MapConfig"/>
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
@@ -642,6 +642,14 @@
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
+                                <xs:attribute name="hot-restart-enabled" use="optional" type="xs:string">
+                                    <xs:annotation>
+                                        <xs:documentation>
+                                            This boolean parameter enables hot-restart feature when set as true.
+                                            Only available on Hazelcast Enterprise.
+                                        </xs:documentation>
+                                    </xs:annotation>
+                                </xs:attribute>
                             </xs:complexType>
                         </xs:element>
                         <xs:element name="multimap" minOccurs="0" maxOccurs="unbounded">
@@ -972,7 +980,18 @@
                         </xs:element>
                         <xs:element name="services" type="services" minOccurs="0" maxOccurs="1"/>
                         <xs:element name="quorum" type="quorum" minOccurs="0" maxOccurs="1"/>
-
+                        <xs:element name="lite-member" minOccurs="0" maxOccurs="1">
+                            <xs:complexType>
+                                <xs:attribute name="enabled" type="xs:string">
+                                    <xs:annotation>
+                                        <xs:documentation>
+                                            True to set the node as a lite member, false otherwise.
+                                        </xs:documentation>
+                                    </xs:annotation>
+                                </xs:attribute>
+                            </xs:complexType>
+                        </xs:element>
+                        <xs:element name="hot-restart" type="hot-restart" minOccurs="0" maxOccurs="1" />
                     </xs:sequence>
                 </xs:extension>
             </xs:complexContent>
@@ -2415,4 +2434,17 @@
     <xs:simpleType name="wan-ack-type">
         <xs:union memberTypes="wan-ack-type-format-enum non-space-string"/>
     </xs:simpleType>
+
+    <xs:complexType name="hot-restart">
+        <xs:sequence>
+            <xs:element name="home-dir" type="xs:string" minOccurs="0" maxOccurs="1" default="hot-restart"/>
+        </xs:sequence>
+        <xs:attribute name="enabled" type="xs:string">
+            <xs:annotation>
+                <xs:documentation>
+                    True to enable hot-restart, false otherwise.
+                </xs:documentation>
+            </xs:annotation>
+        </xs:attribute>
+    </xs:complexType>
 </xs:schema>
--- a/hazelcast/src/main/java/com/hazelcast/config/CacheSimpleConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/CacheSimpleConfig.java
@@ -99,6 +99,9 @@ public class CacheSimpleConfig {
 
     private String mergePolicy = BuiltInCacheMergePolicies.getDefault().getImplementationClassName();
 
+    private boolean hotRestartEnabled;
+
+    @SuppressWarnings("checkstyle:executablestatementcount")
     public CacheSimpleConfig(CacheSimpleConfig cacheSimpleConfig) {
         this.name = cacheSimpleConfig.name;
         this.keyType = cacheSimpleConfig.keyType;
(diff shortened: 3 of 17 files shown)


mtopolnik:
:+1:

checkstyle/suppressions.xml
```diff
@@ -98,6 +98,7 @@
     <suppress checks="CyclomaticComplexity" files="com/hazelcast/config/AbstractXmlConfigHelper"/>
     <suppress checks="ClassDataAbstractionCoupling" files="com/hazelcast/config/AbstractXmlConfigHelper"/>
     <suppress checks="FanOutComplexity" files="com/hazelcast/config/ConfigXmlGenerator"/>
+    <suppress checks="ClassFanOutComplexityCheck" files="com/hazelcast/config/Config"/>
```
mtopolnik: It should be `ClassFanOutComplexity`, without `Check` in the end.

hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
```diff
@@ -642,6 +642,15 @@
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
+                                <xs:attribute name="hot-restart-enabled" use="optional" type="xs:string">
+                                    <xs:annotation>
+                                        <xs:documentation>
+                                            This boolean parameter enables hot-restart feature when set as
+                                            `true`.
```
mtopolnik: No point in using backticks here, is there? Or is this scraped by some markdown-honoring tool?

checkstyle/suppressions.xml
mdogan: I just copied from another suppression :)

hazelcast/src/main/java/com/hazelcast/config/CacheSimpleConfig.java
```diff
@@ -99,6 +99,9 @@
 
     private String mergePolicy = BuiltInCacheMergePolicies.getDefault().getImplementationClassName();
 
+    private boolean hotRestartEnabled;
+
+    //CHECKSTYLE:OFF
     public CacheSimpleConfig(CacheSimpleConfig cacheSimpleConfig) {
```
mtopolnik: Instead of the `CHECKSTYLE:OFF` sledgehammer I would prefer a focused `@SuppressWarnings("checkstyle:whatever")`.

checkstyle/suppressions.xml
mdogan: It seems both versions work.

Donnerbart: Hmm, than the other suppression might not work. The shorter name should be correct:
http://checkstyle.sourceforge.net/config_metrics.html#ClassFanOutComplexity

The longer name is the name of the class which defines the check.

mtopolnik: They do, but the one without the suffix should be preferred.

mtopolnik: This is actually a regex matching any part of the name. For example, just `Complexity` would also word and exclude all complexity checks.

hazelcast/src/main/java/com/hazelcast/config/CacheSimpleConfig.java
```diff
@@ -551,6 +555,19 @@ public CacheSimpleConfig setQuorumName(String quorumName) {
...
     /**
+     * Sets whether hot restart is enabled on this cache.
+     * @return this
+     */
+    public CacheSimpleConfig setHotRestartEnabled(boolean hotRestart) {
+        this.hotRestartEnabled = hotRestart;
+        return this;
+    }
+
+    public boolean isHotRestartEnabled() {
```
pveentjer: Javadoc is missing here.

hazelcast/src/main/java/com/hazelcast/config/Config.java
```diff
@@ -993,6 +995,14 @@ public Config setPartitionGroupConfig(PartitionGroupConfig partitionGroupConfig)
         return this;
     }
 
+    public HotRestartConfig getHotRestartConfig() {
+        return hotRestartConfig;
+    }
+
+    public void setHotRestartConfig(HotRestartConfig hrConfig) {
+        this.hotRestartConfig = hrConfig;
```
pveentjer: Should we have a null check here? And there is no javadoc.

hazelcast/src/main/java/com/hazelcast/config/HotRestartConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Default hot-restart home directory
+     */
+    public static final String HOT_RESTART_HOME_DEFAULT = "hot-restart";
+
+    private boolean enabled;
+    private File homeDir = new File(HOT_RESTART_HOME_DEFAULT);
+
+    public boolean isEnabled() {
```
pveentjer: Javadoc is missing here.

hazelcast/src/main/java/com/hazelcast/config/HotRestartConfig.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package com.hazelcast.config;
+
+import java.io.File;
+
+/**
+ * Configures the Hot Restart stores.
+ */
+public class HotRestartConfig {
```
pveentjer: We definitely need to add javadoc to the class so a user has a clue what the purpose is.

hazelcast/src/main/java/com/hazelcast/config/HotRestartConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Home directory for all Hot Restart stores.
+     */
+    public File getHomeDir() {
+        return homeDir;
+    }
+
+    public HotRestartConfig setHomeDir(File homeDir) {
+        this.homeDir = homeDir;
```
pveentjer: null check?

pveentjer:
Looks good. Minor comments regarding javadoc.

hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
```diff
@@ -2415,4 +2434,17 @@
     <xs:simpleType name="wan-ack-type">
         <xs:union memberTypes="wan-ack-type-format-enum non-space-string"/>
     </xs:simpleType>
+
+    <xs:complexType name="hot-restart">
+        <xs:sequence>
+            <xs:element name="home-dir" type="xs:string" minOccurs="0" maxOccurs="1" default="hot-restart"/>
+        </xs:sequence>
+        <xs:attribute name="enabled" type="xs:string">
```
bilalyasar: why this is `xs:string`? in hazelcast-config.xsd this is `boolean`.

mtopolnik: This reminds me that our `getBooleanValue` is unneccessarily complex, allowing the synonyms "on" and "yes".

mdogan: @bilalyasar: because in Spring you can replace config values with variable expressions, like `${hotrestart.enabled}`. But you can't write that expression to a `xs:boolean` type. All elements & attributes in spring xml should be `xs:string`, if you ask me.

mtopolnik: @mdogan Does the property value then propagate directly to our Java code? So the `getBooleanValue` method does make sense as-is?

serkan-ozal: what about `parameterized-boolean` type in our xsd? It can be `parameterizedValueType` or `xs:boolean`

bilalyasar: @mdogan you are definitely right. actually for those cases we have `parameterised-boolean`
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd#L268

mdogan: @mtopolnik: Spring handles basic type conversions, string -> primitives, file; array -> collections etc.

mdogan: @bilalyasar, @serkan-ozal: I see. `parameterized-boolean` may be a better choice. But I don't want to start a build just to change an XSD definition. 

Also, spring schema seems very inconsistent in this matter. Some elements are using `parameterized-X` types, some are just `xs:string`.

mdogan:
@bilalyasar, @serkan-ozal: see https://github.com/hazelcast/hazelcast/commit/d336e23baaed9e499e8a2f9cf18561429bf0b652

PR #6587, opened 2015-10-28
Additional testing for NotPredicate

pveentjer:


pveentjer:
Merging since this test is not the cause of the build failure and the test is very deterministic.

PR #6589, opened 2015-10-28
lazy deserialization of event to map,multimap,queue,topic,list,set

### LINKED ISSUE #6582: Lazy Deserialization required on events processing
If 2 clients are listening on entryEvicted of EntryEvictedListener only, a map.put on one client results in a callback ClientMapProxy$ClientMapEventHandler.creteEntryEvent() on other clients, which further leads to deserialisation of the entry on those clients. This causes GC and excess CPU issues on the clients which are not even actively making map calls. 

It requires lazy deserialization on clients when the event arrives. This will prevent excess CPU and memory usage spikes.

sancar:
fixes #6582

### DIFF
--- a/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMultiMapProxy.java
+++ b/hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMultiMapProxy.java
@@ -52,6 +52,7 @@ import com.hazelcast.core.IMapEvent;
 import com.hazelcast.core.MapEvent;
 import com.hazelcast.core.Member;
 import com.hazelcast.core.MultiMap;
+import com.hazelcast.map.impl.DataAwareEntryEvent;
 import com.hazelcast.map.impl.ListenerAdapter;
 import com.hazelcast.mapreduce.Collator;
 import com.hazelcast.mapreduce.CombinerFactory;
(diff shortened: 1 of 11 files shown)


hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientMultiMapProxy.java
```diff
@@ -478,16 +479,9 @@ private MapEvent createMapEvent(int eventType, int numberOfAffectedEntries, Memb
...
-            V oldValue = null;
-            V mergingValue = null;
-            if (includeValue) {
-                value = toObject(valueData);
-                oldValue = toObject(oldValueData);
-                mergingValue = toObject(mergingValueData);
-            }
-            K key = toObject(keyData);
-            return new EntryEvent<K, V>(name, member, eventType, key, oldValue, value, mergingValue);
+            return new DataAwareEntryEvent(member,
```
pveentjer: Please rename this class to LazyDeserializingEntryEvent.

pveentjer:
Some tests need to be added to test this behavior.

hazelcast-client-new/src/main/java/com/hazelcast/client/proxy/ClientTopicProxy.java
```diff
@@ -112,12 +113,11 @@ private TopicItemHandler(MessageListener<E> listener) {
...
         public void handle(Data item, long publishTime, String uuid) {
-            final SerializationService serializationService = getContext().getSerializationService();
-            final ClientClusterService clusterService = getContext().getClusterService();
+            SerializationService serializationService = getContext().getSerializationService();
+            ClientClusterService clusterService = getContext().getClusterService();
 
-            E messageObject = serializationService.toObject(item);
             Member member = clusterService.getMember(uuid);
-            Message<E> message = new Message<E>(name, messageObject, publishTime, member);
+            Message<E> message = new DataAwareMessage(name, item, publishTime, member, serializationService);
```
pveentjer: I would rename this to LazyDeserializingMessage.

sancar:
@pveentjer Do you think we should add a test to check if event has data and not serialized yet, then we called the getValue then it is deserialized? If you dont think it is necessary, other listener tests are already available.

sancar:
Merging this since it is reviewed and merged here. https://github.com/hazelcast/hazelcast/pull/6590
I take my note about test, I will think about how can I implement and send another pr later.

PR #6590, opened 2015-10-28
lazy deserialization of event to map,multimap,queue,topic,list,set,re…

### LINKED ISSUE #6582: Lazy Deserialization required on events processing

sancar:
…plicated map

fixes #6582

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/AbstractClientCollectionProxy.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/AbstractClientCollectionProxy.java
@@ -19,6 +19,7 @@ package com.hazelcast.client.proxy;
 import com.hazelcast.client.impl.client.ClientRequest;
 import com.hazelcast.client.spi.ClientProxy;
 import com.hazelcast.client.spi.EventHandler;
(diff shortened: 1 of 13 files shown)


pveentjer:
Can ItemEvent.getItem be called concurrently? getItem isn't threadsafe; and I'm not pointing at a potential double deserialization.

pveentjer:
And can we rename DataAwareItemEvent to e.g. LazyDeserializingItemEvent. The name DataAware is very confusing to me.

sancar:
@pveentjer 
We are passing new DataAwareItemEvent to every listener. They are not shared, so concurrency should not be a problem.

Lets do renaming later. We need these asap. And we dont want to wait for tests to pass again.

pveentjer:
:+1:

pveentjer:
One problem. There are not tests. Is it possible to write tests for this behavior?

PR #6591, opened 2015-10-28
lazy deserialization of event to map,multimap,queue,topic,list,set

sancar:
related to #6582


sancar:
Merging this since it is reviewed and merged here. https://github.com/hazelcast/hazelcast/pull/6590

PR #6592, opened 2015-10-28
Added more unittests for higher code coverage

noctarius:


hazelcast/src/main/java/com/hazelcast/config/properties/ValidationException.java
```diff
@@ -35,6 +36,6 @@ public ValidationException(String message, Throwable cause) {
     }
 
     public ValidationException(Throwable cause) {
-        super(cause);
+        this(cause != null ? cause.getMessage() : null, cause);
```
pveentjer: I would not expect somebody every to call this with a null cause. So I would remove this logic. If they have a null cause, just call the no arg constructor.

pveentjer:
:+1:

Issue #6600, opened 2015-10-29
Could IExecutorService threads have their thread name altered


StanAccy:
Currently, all threads related to IExecutorService instances are allocated from the "cached" thread pool and are named accordingly: cached1..cachedN.

Would it be possible to append/insert the IExecutorService name to the created thread so its easier to see/diagnose in the debugger/profiler.

Eg.  IExecutorService.setName("foo");  would result in:
cached_foo_1
cached_foo_2
..
cached_foo_N

pveentjer:
I never understood why a cached threadpool is being used here. That is the cause of the names of the threads being cached instead of concrete. I'll make a pr for it; I'm curious about the reasons behind this choice.

StanAccy:
Great, thanks.  Perhaps the thread names could be execsrv-{IES_NAME}-1 (or something along those lines)

pveentjer:
When I name the threadpool 'foo' this is how the thread is named

![screenshot_2015-10-29_23-25-56](https://cloud.githubusercontent.com/assets/105243/10832496/71452bf6-7e94-11e5-8d0d-96e9fa81dc1d.png)

StanAccy:
The threadpool, yes.  But we can only name the IExecutorService - that's the name I think would be helpful.

pveentjer:
But that is the name you use to retrieve the threapool from hz.

IExecutorService executor = hz.getExecutorService("thisismycoolexecutor")

StanAccy:
When I name the IExecutorService "foo", I just get a series of threads named cached1..cachedN where N is the IExecutorService.poolSize

pveentjer:
yes. But with my fix that should be addressed. If you name your IExecutorService foo, your threads will be called something with foo.

StanAccy:
That sounds perfect.

pveentjer:
https://github.com/hazelcast/hazelcast/pull/6601

jerrinot:
You probably do not want to create a new thread pool for each IExecutorService - that's a big change when compared to the current implementation with negative side-effects. 

I understand descriptive thread names are useful during troubleshooting and for operation monitoring in general. I guess the way to move forward is to (temporary) rename the thread when task is about to be executed. This is a nice place: https://github.com/jerrinot/hazelcast/blob/d0fcf0bb8da8265c41458b3c6f4be023cabe5397/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java#L205-L205

[...]

PR #6601, opened 2015-10-29
[DONT MERGE] IExecutorService gets private NamedThreadPoolExecutor

pveentjer:
Having a shared one causes isolation and contention problems. It also makes debugging problematic since each thread is called 'cached'.  So no more cached, but concrete. 

See https://github.com/hazelcast/hazelcast/issues/6600


StanAccy:
Peter - have I understood correctly that this change wont go in to 3.6.0.  Will it be able to go into 3.6.1 or only in 3.7?

pveentjer:
Probably 3.7.

hasancelik:
run-lab-run

hasancelik:
run-lab-run

PR #6604, opened 2015-10-30
Processed packet count should be decremented for events packet also.

sancar:


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientListenerServiceImpl.java
```diff
@@ -164,7 +164,11 @@ public void run() {
             final ClientResponse clientResponse = serializationService.toObject(packet.getData());
             final int callId = clientResponse.getCallId();
             final Data response = clientResponse.getResponse();
-            handleEvent(response, callId, conn);
+            try {
+                handleEvent(response, callId, conn);
+            } finally {
+                conn.decrementPacketCount();
```
pveentjer: What does packet count mean? Number of pending packets? If so, why isn't it called 'decrementPendingPacketCount'? This way code becomes self documenting.

sancar:
@pveentjer I updated the pr .

mdogan:
:+1:

PR #6608, opened 2015-10-30
Decrement Pending packets for events also

sancar:
Pending packet count is kept per connection incremented each time
a new packet from that connection is arrived and should be
decremented back when processing is over. For event packages
decrementing was missing. It is added in this pr.


sancar:
Merging this one since it is reviewed here https://github.com/hazelcast/hazelcast/pull/6612

PR #6609, opened 2015-10-30
jclouds discoverystrategy tests

mustafaiman:
adds tests for jclouds dicovery strategy builder finding running nodes

### DIFF
--- a/hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
+++ b/hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
@@ -1,0 +1,100 @@
+package com.hazelcast.jclouds;
+
+import com.hazelcast.core.HazelcastException;
+import com.hazelcast.nio.Address;
+import com.hazelcast.spi.discovery.DiscoveryNode;
+import com.hazelcast.test.HazelcastParallelClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
+import com.hazelcast.test.annotation.QuickTest;
+import org.jclouds.compute.domain.NodeMetadata;
+import org.jclouds.compute.domain.internal.NodeMetadataImpl;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.net.UnknownHostException;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category(QuickTest.class)
+public class JCloudsDiscoveryStrategyTest extends HazelcastTestSupport {
+    private static final int STARTING_PORT = 5701;
+    private static final int NUMBER_OF_NODES = 10;
+    private static final int NUMBER_OF_RUNNING_NODES = 7;
+
+    private Set<NodeMetadata> nodes = new HashSet<NodeMetadata>();
+    private Set<Address> addressesOfRunningInstances = new HashSet<Address>();
+
+    @Before
+    public void setup()
+            throws UnknownHostException {
+        for (int i = 0; i < NUMBER_OF_NODES; i++) {
+            HashSet<String> privateAdressses = new HashSet<String>();
+            privateAdressses.add("127.0.0." + (i+1));
+            NodeMetadata.Status status = NodeMetadata.Status.PENDING;
+            if (i < NUMBER_OF_RUNNING_NODES) {
+                status = NodeMetadata.Status.RUNNING;
+                addressesOfRunningInstances.add(new Address(privateAdressses.iterator().next(),0));
+            }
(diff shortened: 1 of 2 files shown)


hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category(QuickTest.class)
+public class JCloudsDiscoveryStrategyTest extends HazelcastTestSupport {
+    private static final Set<NodeMetadata> nodes = new HashSet<NodeMetadata>();
```
pveentjer: Can't this cause problems? Because these fields are static, so shared between threads. And the tests of this class can be run in parallel, so potentially multiple threads on the same HashSet(s).

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final Set<Address> addressesOfRunningInstances = new HashSet<Address>();
+
+    private static final int STARTING_PORT = 5701;
+    private static final int NUMBER_OF_NODES = 10;
+    private static final int NUMBER_OF_RUNNING_NODES = 7;
+
+    @BeforeClass
+    public static void start()
```
pveentjer: I would rename this to 'beforeClass'.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+    public void testDestroyCalled() {
+        ComputeServiceBuilder mockComputeServiceBuilder = mock(ComputeServiceBuilder.class);
+        JCloudsDiscoveryStrategy jCloudsDiscoveryStrategy = new JCloudsDiscoveryStrategy(mockComputeServiceBuilder);
+        jCloudsDiscoveryStrategy.destroy();
+
+        verify(mockComputeServiceBuilder).destroy();
+    }
+
+    @Test(expected = HazelcastException.class)
+    public void testInvalidAddressShouldThrow() {
```
pveentjer: I would suggest using the following naming

whenInvalidAddress_thenHazelcastException

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @BeforeClass
+    public static void start()
+            throws UnknownHostException {
+        for (int i = 0; i < NUMBER_OF_NODES; i++) {
+            HashSet<String> privateAdressses = new HashSet<String>();
+            privateAdressses.add("127.0.0." + (i+1));
+            NodeMetadata.Status status = NodeMetadata.Status.PENDING;
+            if (i < NUMBER_OF_RUNNING_NODES) {
+                status = NodeMetadata.Status.RUNNING;
+                addressesOfRunningInstances.add(new Address(privateAdressses.iterator().next(),0));
```
pveentjer: what does this mean? I'm confused with the iterator.next thing because you are grabbing a arbitrary address from the privateAddresses.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
mustafaiman: These sets are not changed in tests. So there should not be any problems. But I will make them non-static to not confuse future developers.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
mustafaiman: There is only one address in `privateAddresses`. I used plural name for variable because it is a set.

mustafaiman:
I have updated the pr @pveentjer

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/JCloudsDiscoveryStrategyTest.java
pveentjer: But if there is only a single item, why make it a set?

mustafaiman: This is how JClouds API works. `NodeMetadata` takes a list of IP addresses for both private and public IPs. It supports multiple IP addresses but this test uses only 1 public and private address.

pveentjer:
Can you squash the commits? Apart from that

:+1:

pveentjer:
Do you plan to add integration tests so that we can actually see if the integration works? It is good to have some mock tests to get basic behavior tested, but we also should verify it truly works. I know this is a huge pain to test.

Donnerbart:
If you plan to add an integration test, please be aware that you cannot use the test factory with the mocked network. It will punch in its own discovery, so your test may run successfully but will not test your own discovery strategy. Chris had this issue with the discovery SPI before and we only saw this in the low code coverage during the QA verification.

PR #6610, opened 2015-10-30
Fixed errors on listeners with predicates on replicated map

eminn:
predicates applied on query entries not the raw objects

### DIFF
--- a/hazelcast-client-legacy/src/test/java/com/hazelcast/client/replicatedmap/ClientReplicatedMapListenerTest.java
+++ b/hazelcast-client-legacy/src/test/java/com/hazelcast/client/replicatedmap/ClientReplicatedMapListenerTest.java
@@ -22,16 +22,20 @@ import com.hazelcast.core.EntryListener;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.MapEvent;
 import com.hazelcast.core.ReplicatedMap;
+import com.hazelcast.query.impl.FalsePredicate;
 import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.annotation.QuickTest;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicLong;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertEquals;
 
 @Category(QuickTest.class)
 @RunWith(HazelcastParallelClassRunner.class)
@@ -46,10 +50,7 @@ public class ClientReplicatedMapListenerTest extends HazelcastTestSupport {
 
     @Test
     public void testEntryAdded() throws Exception {
-        HazelcastInstance instance = factory.newHazelcastInstance();
-        HazelcastInstance client = factory.newHazelcastClient();
-        String mapName = randomMapName();
-        ReplicatedMap<Object, Object> replicatedMap = client.getReplicatedMap(mapName);
+        ReplicatedMap<Object, Object> replicatedMap = createClusterAndGetRandomReplicatedMap();
         final EventCountingListener listener = new EventCountingListener();
         replicatedMap.addEntryListener(listener);
         replicatedMap.put(1, 1);
@@ -63,10 +64,7 @@ public class ClientReplicatedMapListenerTest extends HazelcastTestSupport {
 
     @Test
     public void testEntryUpdated() throws Exception {
-        HazelcastInstance instance = factory.newHazelcastInstance();
-        HazelcastInstance client = factory.newHazelcastClient();
-        String mapName = randomMapName();
-        ReplicatedMap<Object, Object> replicatedMap = client.getReplicatedMap(mapName);
+        ReplicatedMap<Object, Object> replicatedMap = createClusterAndGetRandomReplicatedMap();
         final EventCountingListener listener = new EventCountingListener();
         replicatedMap.addEntryListener(listener);
         replicatedMap.put(1, 1);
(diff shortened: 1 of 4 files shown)


hazelcast-client-legacy/src/test/java/com/hazelcast/client/replicatedmap/ClientReplicatedMapListenerTest.java
```diff
@@ -115,6 +117,50 @@ public void run() throws Exception {
         });
     }
 
+
+    @Test
+    public void testListenToKeyForEntryAdded() throws Exception {
+        HazelcastInstance instance = factory.newHazelcastInstance();
+        HazelcastInstance client = factory.newHazelcastClient();
+        String mapName = randomMapName();
+        ReplicatedMap<Object, Object> replicatedMap = client.getReplicatedMap(mapName);
```
pveentjer: Can't this be pulled out into the setup of the test since each test has exactly the same beginning.  So everything above this comment.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/replicatedmap/ClientReplicatedMapListenerTest.java
```diff
@@ -115,6 +117,50 @@ public void run() throws Exception {
...
+        final EventCountingListener listener = new EventCountingListener();
+        replicatedMap.addEntryListener(listener, 1);
+        replicatedMap.put(1, 1);
+        replicatedMap.put(2, 2);
+        assertTrueEventually(new AssertTask() {
```
pveentjer: Testing flavor; I add line of white to quickly find the 'do' and the verification phase.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/replicatedmap/ClientReplicatedMapListenerTest.java
```diff
@@ -115,6 +117,50 @@ public void run() throws Exception {
...
+    @Test
+    public void testListenWithPredicate() throws Exception {
+        HazelcastInstance instance = factory.newHazelcastInstance();
+        HazelcastInstance client = factory.newHazelcastClient();
+        String mapName = randomMapName();
+        ReplicatedMap<Object, Object> replicatedMap = client.getReplicatedMap(mapName);
+        final EventCountingListener listener = new EventCountingListener();
+        replicatedMap.addEntryListener(listener, new FalsePredicate());
+        replicatedMap.put(2, 2);
+        assertTrueEventually(new AssertTask() {
```
pveentjer: using assertTrueEventually is probably not the right one. Because in theory it can happen that the assertTrueEventually is run immediately and sees there are not listeners. However the event got pending somewhere and eventually caused the listener to be called due to potential bug. You won't see this because the test thread has already completed the test.

Probably this one is better: assertTrueFiveSeconds

hazelcast-client-legacy/src/test/java/com/hazelcast/client/replicatedmap/ClientReplicatedMapListenerTest.java
```diff
@@ -115,6 +117,50 @@ public void run() throws Exception {
...
+            @Override
+            public void run() throws Exception {
+                Assert.assertEquals(0, listener.addCount.get());
+            }
+        });
+    }
+
+    private static class FalsePredicate implements Predicate<Object, Object> {
```
pveentjer: We have a false predicate: FalsePredicate.INSTANCE.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/replicatedmap/ClientReplicatedMapListenerTest.java
```diff
@@ -115,6 +117,50 @@ public void run() throws Exception {
...
+        String mapName = randomMapName();
+        ReplicatedMap<Object, Object> replicatedMap = client.getReplicatedMap(mapName);
+        final EventCountingListener listener = new EventCountingListener();
+        replicatedMap.addEntryListener(listener, 1);
+        replicatedMap.put(1, 1);
+        replicatedMap.put(2, 2);
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                Assert.assertEquals(1, listener.addCount.get());
```
pveentjer: If the event would not contain the right content, your count based test is not going to detect this; perhaps 2 was received instead of 1. If you would store the events in a threadsafe list in the entry listener, you could do assertEquals(asList(1), listener.items())

There is another testing problem that can be hidden. Imagine that the predicate would not be triggered correctly (ignored) and therefor the listener received 1 first and then it received 2. 

But you end the test immediately after you have found the first event.. you are not going to check for some time if the second event is not going to be received.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapEventPublishingService.java
```diff
@@ -187,13 +189,26 @@ public void fireEntryListenerEvent(Data key, Data oldValue, Data value, EntryEve
...
+            if (!shouldPublish(key, oldValue, value, eventType, registration.getFilter())) {
+                continue;
             }
+            eventService.publishEvent(SERVICE_NAME, registration, eventData, key.hashCode());
         }
     }
 
+    private boolean shouldPublish(Data key, Data oldValue, Data value, EntryEventType eventType, EventFilter filter) {
+        QueryEntry queryEntry = null;
+        if (filter instanceof ReplicatedQueryEventFilter) {
```
pveentjer: Can it be that a different type of filter than a ReplicatedQueryEventFilter is going to be passed? If so, is this not erronous behavior? It could very well be that I'm wrong; I'm trying to figure out if we should fail fast or not.. or if this is a normal flow.

pveentjer: Ah ok.. I get it.  Forget my previous comment

pveentjer:
:+1:

PR #6612, opened 2015-10-30
Decrement Pending packets for events also

sancar:
Pending packet count is kept per connection, incremented each
time a new packet from that connection is arrived and should be
decremented back when processing is over. For event packages
decrementing was missing. It is added in this pr.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientListenerServiceImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientListenerServiceImpl.java
@@ -160,11 +160,15 @@ public final class ClientListenerServiceImpl implements ClientListenerService {
 
         @Override
         public void run() {
-            final ClientConnection conn = (ClientConnection) packet.getConn();
-            final ClientResponse clientResponse = serializationService.toObject(packet.getData());
-            final int callId = clientResponse.getCallId();
-            final Data response = clientResponse.getResponse();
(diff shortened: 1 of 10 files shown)


pveentjer:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientListenerServiceImpl.java
```diff
@@ -160,11 +160,15 @@ private ClientEventProcessor(Packet packet) {
 
         @Override
         public void run() {
-            final ClientConnection conn = (ClientConnection) packet.getConn();
-            final ClientResponse clientResponse = serializationService.toObject(packet.getData());
-            final int callId = clientResponse.getCallId();
-            final Data response = clientResponse.getResponse();
-            handleEvent(response, callId, conn);
+            ClientConnection conn = (ClientConnection) packet.getConn();
+            ClientResponse clientResponse = serializationService.toObject(packet.getData());
```
Donnerbart: Can this line (or one of the others) throw an exception? Should we maybe put those 4 lines into the try-finally block like in the legacy client?

sancar: You are right, serialization could throw exception. I will move these into try block.

Donnerbart:
:+1: (once green)

Issue #5880, opened 2015-08-06
VersionCheck uses hardcoded domain name and doesn't follow redirects


CleverMoose:
We have noticed a new error message in our logs saying:
ERROR [STDERR] [Fatal Error] :6:3: The element type "hr" must be terminated by the matching end-tag "&lt;/hr&gt;".

It seems the source of the error is VersionCheck which uses hardcoded domain name to fetch XML version document at https://hazelcast.com/version.jsp?version=2.6.1.
This url redirects to a new domain. However, VersionCheck doesn't follow HTTP redirect and attempts to parse returned HTML document.

kobalski:
@CleverMoose Thanks for the feedback. We have disabled HTTP to HTTPS redirection on our website. You should not see the exception in your logs now. I am closing the issue, please feel free to reopen, if you see any other problems.

CleverMoose:
I've checked latest version of VersionCheck and it seems that only purpose of this class is to report HZ cluster specific information to Hazelcast:
```
        parameterCreator.addParam("version", version);
        parameterCreator.addParam("m", hazelcastNode.getLocalMember().getUuid());
        parameterCreator.addParam("e", Boolean.toString(isEnterprise));
        parameterCreator.addParam("l", MD5Util.toMD5String(hazelcastNode.getConfig().getLicenseKey()));
        parameterCreator.addParam("p", downloadId);
        parameterCreator.addParam("c", clusterService.getClusterId());
        parameterCreator.addParam("crsz", convertToLetter(clusterService.getMembers().size()));
        parameterCreator.addParam("cssz", convertToLetter(hazelcastNode.clientEngine.getClientEndpointCount()));
        parameterCreator.addParam("hdgb", nativeMemoryParameter);
        parameterCreator.addParam("ccpp", Integer.toString(clusterClientStats.get(ClientType.CPP)));
        parameterCreator.addParam("cdn", Integer.toString(clusterClientStats.get(ClientType.CSHARP)));
        parameterCreator.addParam("cjv", Integer.toString(clusterClientStats.get(ClientType.JAVA)));
        parameterCreator.addParam("cuptm", Long.toString(clusterUpTime));
        parameterCreator.addParam("nuptm", Long.toString(runtimeMxBean.getUptime()));
        String urlStr = BASE_VERSION_CHECK_URL + parameterCreator.build();
```
To me this seems to be very different from declared (and legacy) functionality:
```
[...]

PR #6626, opened 2015-11-02
Version Check code is refactored with phone home code. 

kobalski:
After hazelcast 3.x version check functionality is deprecated but class
names,properties and urls were still containing `version check` related
stuff. This pr refactors `VersionCheck` to `PhoneHome` Fixes https://github.com/hazelcast/hazelcast/issues/5880

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
@@ -181,7 +181,10 @@ public enum GroupProperty implements HazelcastProperty {
 
     PREFER_IPv4_STACK("hazelcast.prefer.ipv4.stack", true),
 
+    @Deprecated
     VERSION_CHECK_ENABLED("hazelcast.version.check.enabled", true),
+
+    PHONE_HOME_ENABLED("hazelcast.phone.home.enabled", true),
 
     CONNECT_ALL_WAIT_SECONDS("hazelcast.connect.all.wait.seconds", 120, SECONDS),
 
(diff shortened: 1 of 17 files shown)


hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
```diff
@@ -181,7 +181,7 @@
 
     PREFER_IPv4_STACK("hazelcast.prefer.ipv4.stack", true),
 
-    VERSION_CHECK_ENABLED("hazelcast.version.check.enabled", true),
+    PHONE_HOME_ENABLED("hazelcast.phone.home.enabled", true),
```
bilalyasar: so this change requires doc update.

kobalski: Yes, I will send another pr to reference-manual repo.

Serdaro: I've already been making the related changes. I will share the updates with you @kobalski.

kobalski:
@mesutcelik @bilalyasar I have updated PR, because it was effecting more places than I initially thought. Can you take a look again?

mesutcelik:
I see no logic change. Only correcting the naming so :+1:

pveentjer:
Can't the rename not lead to backwards compatibility problems? So an XML file now becoming invalid.

pveentjer:
Perhaps we should add a warning that if somebody uses the old system property to disable phone home, he needs to switch to the new? Because in 3.6 the system will now secretly phone home again.

kobalski:
@pveentjer Old property is still present under `GroupProperty` but annotated with `Deprecated` and if default value of the old property is changed, we give a proper warning log in here https://github.com/hazelcast/hazelcast/pull/6626/files#diff-032f44a5b4672f69049bd4c3622e037cR65. I updated PR.

pveentjer:
Just to make sure I understand correctly.

What if I have configured phone home with the old property name and have have said disable and I'm using e.g. a system property.

When I upgrade to 3.6, will it still be disabled?

It would be annoying if Hazelcast unexpectedly starts to phone home again and way say; you should have read the change log or the logging.

kobalski:
@pveentjer yes, it can be annoying if user did not notice the warning log. I changed the code so that both
old property and the new one works the same and old one still logs the warning. Also we should clearly state this change What's New section in our 3.6 documentation.

emrahkocaman:
:+1:

Issue #6534, opened 2015-10-21
"Invalid regular expression" in hazelcast-spring.xsd


stoerr:
Altova XML Spy complains about an invalid regular expression in https://hazelcast.com/schema/spring/hazelcast-spring-3.5.xsd (actually in versions 3.3 and 3.4 too). This is about the following line:

    <xs:simpleType name="parameterizedValueType">
        <xs:restriction base="xs:string">
            <xs:pattern value="(\$|\#)\{([^=^:]+)\}"/>
        </xs:restriction>
    </xs:simpleType>

The reason is that $ is not a meta-character in XML regular expressions, and thus quoting it with backslash is dubious. I suggest changing that regex to 
`[$#]\{([^=^:]+)\}`
which means the same thing.

By the way: `[^=^:]` looks a little strange, too. I'm not quite sure what that the intuition is behind this - should that regex really mean "everything except equals, colon and caret"?

PR #6627, opened 2015-11-02
edited spring xsd regex

bilalyasar:
closes https://github.com/hazelcast/hazelcast/issues/6534

### DIFF
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
@@ -2007,7 +2007,7 @@
 
     <xs:simpleType name="parameterizedValueType">
         <xs:restriction base="xs:string">
-            <xs:pattern value="(\$|\#)\{([^=^:]+)\}"/>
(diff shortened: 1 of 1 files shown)


mesutcelik:
verify

mesutcelik:
Do we have a test for this particular change? Can you please send a backport as well?

hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
```diff
@@ -2007,7 +2007,7 @@
 
     <xs:simpleType name="parameterizedValueType">
         <xs:restriction base="xs:string">
-            <xs:pattern value="(\$|\#)\{([^=^:]+)\}"/>
+            <xs:pattern value="[$#]\{([^=^:]+)\}"/>
```
mesutcelik: do we really need to exclude those chars `^=^:`? If yes, do we need to have some other?

bilalyasar: i didn't remember real cause, but we can leave as it is.

PR #6629, opened 2015-11-02
Adds more tests to jclouds module.

kobalski:


hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/ComputeServiceBuilderTest.java
```diff
@@ -0,0 +1,82 @@
...
+            nodes.add(new NodeMetadataBuilder().id(UuidUtil.newSecureUuidString()).group("group" + i)
+                    .status(NodeMetadata.Status.RUNNING).build());
+        }
+        nodes.add(new NodeMetadataBuilder().id(UuidUtil.newSecureUuidString()).group(groupUnderTest)
+                .status(NodeMetadata.Status.RUNNING).build());
+        doReturn(nodes).when(mockComputeService).listNodesDetailsMatching(null);
+
+        Set<NodeMetadata> result = (Set) builder.getFilteredNodes();
+
+        assertEquals(2, result.size());
```
mesutcelik: you add 5 but you assert 2... I couldn't get the logic?

mesutcelik:
:+1:

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/ComputeServiceBuilderTest.java
kobalski: I add 5 with increasing group number, `group1`,`group2` etc and add one extra `group3` outside of the for loop and assert `builder.getFilteredNodes ` filters nodes correctly for `group3` which should be 2 nodes.

serkan-ozal:
:+1:

PR #6631, opened 2015-11-02
post processed value included to entry event

### LINKED ISSUE #6588: EntryListener shows unprocessed value in combination with PostProcessingMapStore
From stackoverflow:
http://stackoverflow.com/questions/33375564/hazelcast-map-listener-not-getting-the-updated-object-after-updation-in-map-stor

I am using Map store for a map in Hazelcast.I am modifying an object inside the Map store.My map store also implements PostProcessingMapStore interface.

I create a client which does a Map.put to the cache.This put triggers the store on MapStore.I have a listener to the Map on client side also.When I try to do a map.get, I get the modified object back(because of PostProcessingMapStore) , but the entry added method inside the listener shows the unmodified object.Am I missing some configuration?

eminn:
Fixes #6588

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllOperation.java
@@ -100,6 +100,7 @@ public class PutAllOperation extends MapOperation implements PartitionAwareOpera
         mapServiceContext.interceptAfterPut(name, dataValue);
         EntryEventType eventType = dataOldValue == null ? EntryEventType.ADDED : EntryEventType.UPDATED;
         final MapEventPublisher mapEventPublisher = mapServiceContext.getMapEventPublisher();
+        dataValue = getPostProcessedValueIfAvailable(mapServiceContext, recordStore, dataKey, dataValue);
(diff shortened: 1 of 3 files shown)


ahmetmircik:
This PR is mergeable but the fix should also be applied to `Entry Processor operations` and `putAll`.
And it seems we need a backport for 3.5.x.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllOperation.java
```diff
@@ -119,6 +120,19 @@ private void put(int partitionId, MapServiceContext mapServiceContext, RecordSto
         evict();
     }
 
+    private Data getPostProcessedValueIfAvailable(MapServiceContext mapServiceContext, RecordStore recordStore,
```
ahmetmircik: Maybe `getValueOrPostProcessedValue`?

ahmetmircik:
just one minor comment :+1:

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllOperation.java
eminn: I will do a direct commit for that when it's merged

PR #6632, opened 2015-11-03
Fix for MergeOperation's shouldBackup logic.

emrahkocaman:
When a merge policy returns null, it endsup with the deletion of related entry. (see https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java#L747)

In that case backup of this entry should also be deleted, this is supposed to be done by `RemoveBackupOperation` generated in the `getBackupOperation` of `MergeOperation`. (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MergeOperation.java#L93)

But this method is never called in that case since `shouldBackup` method returns `false` since it checks the existence of the record. 

Checking only `merged` property should be enough.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MergeOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MergeOperation.java
@@ -70,8 +70,7 @@ public class MergeOperation extends BasePutOperation {
 
     @Override
     public boolean shouldBackup() {
-        final Record record = recordStore.getRecord(dataKey);
-        return merged && record != null;
+        return merged;
     }
 
     @Override
@@ -93,7 +92,7 @@ public class MergeOperation extends BasePutOperation {
             return new RemoveBackupOperation(name, dataKey);
         } else {
             final Record record = recordStore.getRecord(dataKey);
-            final RecordInfo replicationInfo = record != null ? Records.buildRecordInfo(record) : null;
+            final RecordInfo replicationInfo = Records.buildRecordInfo(record);
             return new PutBackupOperation(name, dataKey, dataValue, replicationInfo);
         }
     }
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/operation/MergeOperation.java
```diff
@@ -70,8 +70,7 @@ public Object getResponse() {
 
     @Override
     public boolean shouldBackup() {
-        final Record record = recordStore.getRecord(dataKey);
-        return merged && record != null;
+        return merged;
```
emrahkocaman: Record existence should not be checked since it can be removed if a merge policy returns `null`

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MergeOperation.java
```diff
@@ -93,7 +92,7 @@ public Operation getBackupOperation() {
             return new RemoveBackupOperation(name, dataKey);
         } else {
             final Record record = recordStore.getRecord(dataKey);
-            final RecordInfo replicationInfo = record != null ? Records.buildRecordInfo(record) : null;
+            final RecordInfo replicationInfo = Records.buildRecordInfo(record);
```
emrahkocaman: Removed `null` check since `record` can not be null here.

emrahkocaman:
@ahmetmircik , Can you please have a look?

ahmetmircik:
:+1:  for this one. 
Could you please also create a backporter PR?

emrahkocaman:
sure

emrahkocaman:
Backport: https://github.com/hazelcast/hazelcast/pull/6652

Issue #6580, opened 2015-10-27
[TEST-FAILURE] ClientServiceTest Timeout


hasancelik:
```
00:33:05 Running com.hazelcast.client.ClientServiceTest
02:28:31 Build timed out (after 180 minutes). Marking the build as aborted.
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/3956/console

hasancelik:
```
05:45:07 Running com.hazelcast.client.ClientServiceTest
07:14:49 Build timed out (after 180 minutes). Marking the build as aborted.
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/606/console

hasancelik:
After the #6635 , these test failed at Jenkins @cangencer 

com.hazelcast.client.ClientServiceTest.testClientListenerDisconnected
com.hazelcast.client.ClientServiceTest.testClientListenerDisconnected
com.hazelcast.client.ClientServiceTest.com.hazelcast.client.ClientServiceTest

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-OracleJDK1.7/293/

hasancelik:
```
java.lang.AssertionError: CountDownLatch failed to complete within 30 seconds , count left: 2
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:746)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:739)
	at com.hazelcast.client.ClientServiceTest.testClientListenerDisconnected(ClientServiceTest.java:335)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast-client/719/testReport/junit/com.hazelcast.client/ClientServiceTest/testClientListenerDisconnected/

hasancelik:
#6694 
#6695
These two tests related this failure.

sancar:
fixed by #6728

PR #6635, opened 2015-11-03
Add overall timeout to ClientServiceTests to detect stuck test

cangencer:
Also, added assertions to removeClientListener tests

See issue #6580


cangencer:
verify

sancar:
:+1:

hasancelik:
:+1:

pveentjer:
:+1: 

But we need to change this mechanism. Any tests that is not configured to have longer timeout, should by default have a system imposed timeout. So it should be an opt out feature instead of an opt in feature.

PR #6642, opened 2015-11-04
Check NodeState before updating partition table

mdogan:
_Fixes some issues found during hot-restart testing_

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/impl/PartitionStateOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/impl/PartitionStateOperation.java
@@ -17,12 +17,14 @@
 package com.hazelcast.partition.impl;
 
 import com.hazelcast.cluster.impl.operations.JoinOperation;
+import com.hazelcast.instance.NodeState;
 import com.hazelcast.nio.ObjectDataInput;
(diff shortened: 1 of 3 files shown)


metanet:
:+1:

hazelcast/src/main/java/com/hazelcast/partition/impl/PartitionStateOperation.java
```diff
@@ -46,6 +48,11 @@ public PartitionStateOperation(PartitionRuntimeState partitionState, boolean syn
 
     @Override
     public void run() {
+        NodeEngineImpl nodeEngine = (NodeEngineImpl) getNodeEngine();
+        if (nodeEngine.getNode().getState() == NodeState.PASSIVE) {
+            getLogger().warning("Partition table received before startup is not completed. Caller: " + getCallerAddress());
```
pveentjer: should 'not' be added in the message? Should it be 'before startup is complete..'

pveentjer:
I guess it will be tricky to add tests for this since it is difficult to force the cluster in a certain state.

mdogan:
@pveentjer: we have tests on enterprise repo, since we can't test hot-restart on OSS. See https://github.com/hazelcast/hazelcast-enterprise/pull/453

Issue #6615, opened 2015-10-30
Custom credintials class is not de-serialized on the server side


dsukhoroslov:
HZ 3.6-EA. 
I use custom cridentials class which inherits from UsernamePasswordCredential class. I need to provide some additional auth data from client to server. Till version 3.5.3 it works fine: the custom credentials class was deserialized and used on the server side. In 3.6-EA the standard UsernamePasswordCredential class deserialized instead. Auth message created on the server side is AuthenticationMessageTask, but it must be AuthenticationCustomCredentialsMessageTask, I believe.

Please have a look.
Thanks, Denis.

enesakar:
fixed by https://github.com/hazelcast/hazelcast/pull/6644

PR #6644, opened 2015-11-04
Stricter type check when deciding if ClientAuthenticationCustomCodec should be used during authentication

cangencer:
Fixes the issue when the custom credentials given extends UsernamePasswordCredentials. Added tests for both new and legacy client, however the issue only effects the new client. Fixes issue #6615


cangencer:
@gurbuzali is it possible for you to have a quick look at this? Is this a valid use case? see: https://github.com/dsukhoroslov/bagri/blob/master/bagri-server/bagri-server-hazelcast/src/main/java/com/bagri/xdm/cache/hazelcast/serialize/SecureCredentials.java

sancar:
:+1:
Do we have backport of this pr ?

gurbuzali:
:+1:

cangencer:
agreed that no backport needed as it only affects new client on 3.5

PR #6645, opened 2015-11-04
Enhancing Extractors API & fixing ignored tests

tombujok:
Extraction tests refactored. New tests added. Documentation added.
Adding custom attributes to hazelcast-fullconfig.xml
Adding documentation to the test structure
Changing package names
Some more corner cases found and fixed
First version of the enhanced extraction API
Tests adjusted according to null-handling policy
Fixing bugs and inconsistencies in collection & array extraction


hazelcast/src/main/java/com/hazelcast/query/extractor/ValueCollector.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Collects a value extracted by a ValueExtractor.
+     * <p/>
+     * More than one value may be collected in a single extraction, for more details see
+     * {@link com.hazelcast.query.extractor.ValueExtractor}
+     *
+     * @param value value to be collected
+     */
+    public abstract void collect(Object value);
```
jerrinot: what about `collectObject()`? as later on we might want to introduce methods like `collectInt()`, `collectLong()`, etc.

tombujok: Good point. Adding Object suffix would be fine.
I'm also thinking of renaming the base method to `add*()` so that there's no redundancy like `collector.collect`. The name `collector` already says it will be collecting, so it looks weird...
`collector.addObject(object)` or `collector.addInt(4)` look better to me. WDYT?

jerrinot: @pveentjer: What would you suggest?

tombujok:
This PR will be replaced by a couple of smaller ones.

PR #6651, opened 2015-11-05
Fixes major bug which makes jclouds discovery unusable.

kobalski:
Due to a bug in zone and regions filtering current jclouds implementation discovery is
broken in 3.6-EA. Also added more logs and unit tests, configuration for live
testing is coming after this pr is merged.

### DIFF
--- a/hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
+++ b/hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
@@ -33,6 +33,7 @@ import org.jclouds.compute.domain.ComputeMetadata;
 import org.jclouds.compute.domain.NodeMetadata;
 import org.jclouds.domain.Credentials;
 import org.jclouds.domain.Location;
+import org.jclouds.domain.LocationScope;
 import org.jclouds.googlecloud.GoogleCredentialsFromJson;
 import org.jclouds.location.reference.LocationConstants;
 
@@ -80,6 +81,21 @@ public class ComputeServiceBuilder {
         this.properties = properties;
     }
 
+    public Map<String, Comparable> getProperties() {
+        return properties;
+    }
+
+    public Set<String> getRegionsSet() {
+        return regionsSet;
+    }
+
+    public Set<String> getZonesSet() {
+        return zonesSet;
+    }
+
+    public List<AbstractMap.SimpleImmutableEntry> getTagPairs() {
+        return tagPairs;
+    }
 
     /**
      * Injects already built ComputeService
@@ -102,7 +118,7 @@ public class ComputeServiceBuilder {
             if (group != null && !group.equals(metadata.getGroup())) {
                 continue;
             }
-            if (checkRegionZoneConfig(metadata)) {
+            if (!isNodeInsideZones(metadata) || !isNodeInsideRegions(metadata)) {
                 continue;
             }
             ((HashSet<NodeMetadata>) filteredResult).add(metadata);
@@ -114,27 +130,31 @@ public class ComputeServiceBuilder {
         return getOrDefault(JCloudsProperties.HZ_PORT, NetworkConfig.DEFAULT_PORT);
     }
 
-    private boolean checkRegionZoneConfig(NodeMetadata metadata) {
+    public boolean isNodeInsideZones(NodeMetadata metadata) {
         Location location = metadata.getLocation();
         while (location != null) {
             String id = location.getId();
-            switch (location.getScope()) {
-                case ZONE:
-                    if (id != null && regionsSet.contains(id)) {
-                        return true;
-                    }
-                    break;
-                case REGION:
-                    if (id != null && !zonesSet.contains(id)) {
-                        return true;
-                    }
-                    break;
-                default:
-                    break;
+            if (location.getScope().equals(LocationScope.ZONE)) {
+                if (id != null && !zonesSet.isEmpty() && !zonesSet.contains(id)) {
+                    return false;
+                }
             }
             location = location.getParent();
         }
-        return false;
+        return true;
+    }
+    public boolean isNodeInsideRegions(NodeMetadata metadata) {
+        Location location = metadata.getLocation();
+        while (location != null) {
+            String id = location.getId();
+            if (location.getScope().equals(LocationScope.REGION)) {
+                if (id != null && !regionsSet.isEmpty() && !regionsSet.contains(id)) {
+                    return false;
+                }
+            }
+            location = location.getParent();
+        }
+        return true;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -114,27 +133,32 @@ public int getServicePort() {
...
+        while (location != null) {
+            String id = location.getId();
+            if (location.getScope().equals(LocationScope.REGION)) {
+                if (id != null && !regionsSet.isEmpty() && !regionsSet.contains(id)) {
+                    return false;
+                }
+            }
+            location = location.getParent();
+        }
+        return true;
```
mesutcelik: if location itself or any of its parents in the graph is not in `REGION` scope, it returns true. That means to me `node is not inside that region` but method name is `isNodeInsideRegions`.
It feels strange...

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -218,17 +252,18 @@ public boolean apply(ComputeMetadata nodeMetadata) {
                     return false;
                 }
                 for (AbstractMap.SimpleImmutableEntry entry: tagPairs) {
-                    if (!nodeMetadata.getUserMetadata().get(entry.getKey()).
-                            equals(entry.getValue())) {
+                    String value = nodeMetadata.getUserMetadata().get(entry.getKey());
+                    if (value != null && equals(entry.getValue())) {
```
mesutcelik: isn't this `value.equals(entry.getValue())`?

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/ComputeServiceBuilderTest.java
```diff
@@ -79,4 +90,217 @@ public void test_destroy_calls_computeService_destroy() throws Exception {
...
+
+    @Test
+    public void test_buildZonesConfig_parse_single_zone() throws Exception {
+        Map<String, Comparable> properties = new HashMap<String, Comparable>();
+        properties.put("zones", "zone1");
+        ComputeServiceBuilder builder = new ComputeServiceBuilder(properties);
+        builder.buildRegionZonesConfig();
+        assertEquals(1, builder.getZonesSet().size());
+        assertTrue(builder.getZonesSet().contains("zone1"));
+        assertFalse(builder.getZonesSet().contains("zone2"));
```
mesutcelik: i think `assertFalse` is redundant. We are already checking the size.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/ComputeServiceBuilderTest.java
```diff
@@ -79,4 +90,217 @@ public void test_destroy_calls_computeService_destroy() throws Exception {
...
+        assertFalse(builder.getRegionsSet().contains("region2"));
+    }
+
+    @Test
+    public void test_isNodeInsideZones_when_zone_set_is_empty() throws Exception {
+        ComputeServiceBuilder builder = new ComputeServiceBuilder(new HashMap<String, Comparable>());
+        builder.buildRegionZonesConfig();
+
+        assertTrue(builder.isNodeInsideZones(
+                getRunningNodeMetaDataAtLocation(LocationScope.ZONE, "testZone")));
```
mesutcelik: what does `testZone` mean in the context of the test case?

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
kobalski: Yes, it might be feel strange, but a `Node` in jclouds  always has `REGION` scope, because you cannot define a `ComputeResource` without binding it to a region or zone. So your case will never happen.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
kobalski: Good catch also method should return true if tag key/value pair matches. Corrected.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/ComputeServiceBuilderTest.java
kobalski: So true also removed other redundant false statements in other tests.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/ComputeServiceBuilderTest.java
kobalski: `testZone ` is used to differ verified zone from other zone names. I named zones which are not verified as `zone1`,`zone2` etc.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -182,20 +205,30 @@ ComputeService build() {
...
+
+    public Properties buildRegionZonesConfig() {
         final String regions = getOrNull(JCloudsProperties.REGIONS);
         final String zones = getOrNull(JCloudsProperties.ZONES);
+
         Properties jcloudsProperties = newOverrideProperties();
         if (regions != null) {
+            List<String> regionList = Arrays.asList(regions.split(","));
+            for (String region : regionList) {
+                regionsSet.add(region);
```
bilalyasar: so now we are supporting multi regions and zones. (with comma separated) right?

kobalski: Correct.

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -102,7 +118,10 @@ public void setComputeService(ComputeService computeService) {
             if (group != null && !group.equals(metadata.getGroup())) {
                 continue;
             }
-            if (checkRegionZoneConfig(metadata)) {
+            if (!isNodeInsideZones(metadata)) {
+                continue;
```
bilalyasar: just for readability, i prefer
```
if (!isNodeInsideZones(metadata) || !isNodeInsideRegions(metadata) ) {
continue;
}
```

hazelcast-jclouds/src/main/java/com/hazelcast/jclouds/ComputeServiceBuilder.java
```diff
@@ -218,17 +251,18 @@ public boolean apply(ComputeMetadata nodeMetadata) {
                     return false;
                 }
                 for (AbstractMap.SimpleImmutableEntry entry: tagPairs) {
-                    if (!nodeMetadata.getUserMetadata().get(entry.getKey()).
-                            equals(entry.getValue())) {
-                        return false;
+                    String value = nodeMetadata.getUserMetadata().get(entry.getKey());
+                    if (value != null && value.equals(entry.getValue())) {
```
bilalyasar: now we are make `OR` operation between tags. It was `AND` operation before...
which one is more usable? @mesutcelik

mesutcelik: The picture above says we need AND operation here. Otherwise, following config will return both machines instead of `production` machine.

```
<property name="tag-keys">owner,Stack</property>
 <property name="tag-values">DbAdmin,Production</property>
```

kobalski:
Verify

kobalski:
Thanks for your comments. Will merge if no more comments are present today.

kobalski:
verify

mesutcelik:
verify

kobalski:
verify

mesutcelik:
verify

PR #6653, opened 2015-11-05
Fixed unnecessarily longer lock holding of keys in transactional map

### LINKED ISSUE #6419: TransactionalMapProxySupport::replaceIfSameInternal method keeps key locked even if the replace did not succeed
It seems that TransactionalMapProxySupport::replaceIfSameInternal keeps the lock in place for the key even if the replace was unsuccessful.

```
    public boolean replaceIfSameInternal(Data key, Object oldValue, Data newValue) {
        VersionedValue versionedValue = lockAndGet(key, tx.getTimeoutMillis());
        if (!getService().getMapServiceContext().compare(name, oldValue, versionedValue.value)) {
            addUnlockTransactionRecord(key, versionedValue.version);
            return false;
        }
        final TxnSetOperation op = new TxnSetOperation(name, key, newValue, versionedValue.version);
        tx.add(new MapTransactionLogRecord(name, key, getPartitionId(key), op, versionedValue.version, tx.getOwnerUuid()));
        return true;
```
We think that 
```
        if (!getService().getMapServiceContext().compare(name, oldValue, versionedValue.value)) {
```
should unlock the key immediately and not wait for a commit/rollback (or lock timeout) to do it, since it did not replace the value.

Other methods within the same class exhibiting the same behavior are: putIfAbsentInternal, replaceInternal, removeIfSameInternal.

eminn:
In case of conditional operation failure, locks are immediately unlocked. Fixes #6419

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
@@ -38,7 +38,6 @@ import com.hazelcast.transaction.TransactionalObject;
 import com.hazelcast.transaction.impl.Transaction;
 import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.ThreadUtil;
-
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -161,8 +160,13 @@ public abstract class TransactionalMapProxySupport extends AbstractDistributedOb
     }
 
     public Data putIfAbsentInternal(Data key, Data value) {
+        boolean unlockImmediately = !valueMap.containsKey(key);
         VersionedValue versionedValue = lockAndGet(key, tx.getTimeoutMillis());
         if (versionedValue.value != null) {
+            if (unlockImmediately) {
+                unlock(key, versionedValue);
+                return versionedValue.value;
+            }
             addUnlockTransactionRecord(key, versionedValue.version);
             return versionedValue.value;
         }
@@ -173,8 +177,13 @@ public abstract class TransactionalMapProxySupport extends AbstractDistributedOb
     }
 
     public Data replaceInternal(Data key, Data value) {
+        boolean unlockImmediately = !valueMap.containsKey(key);
         VersionedValue versionedValue = lockAndGet(key, tx.getTimeoutMillis());
         if (versionedValue.value == null) {
+            if (unlockImmediately) {
+                unlock(key, versionedValue);
+                return null;
+            }
             addUnlockTransactionRecord(key, versionedValue.version);
             return null;
         }
@@ -184,8 +193,13 @@ public abstract class TransactionalMapProxySupport extends AbstractDistributedOb
     }
 
     public boolean replaceIfSameInternal(Data key, Object oldValue, Data newValue) {
+        boolean unlockImmediately = !valueMap.containsKey(key);
         VersionedValue versionedValue = lockAndGet(key, tx.getTimeoutMillis());
         if (!isEquals(oldValue, versionedValue.value)) {
+            if (unlockImmediately) {
+                unlock(key, versionedValue);
+                return false;
+            }
             addUnlockTransactionRecord(key, versionedValue.version);
             return false;
         }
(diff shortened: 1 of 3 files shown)


eminn:
verify

hazelcast/src/test/java/com/hazelcast/map/MapTransactionLockingTest.java
```diff
@@ -0,0 +1,450 @@
...
+        final String value = "v";
+        final TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(1);
+        final HazelcastInstance hazelcastInstance = factory.newHazelcastInstance(getConfig());
+        IMap<String, Object> map = hazelcastInstance.getMap(mapName);
+
+        final CountDownLatch startAssertion = new CountDownLatch(1);
+        final CountDownLatch finishAssertion = new CountDownLatch(1);
+        final CountDownLatch finishTransaction = new CountDownLatch(1);
+
+        new Thread(new Runnable() {
```
gurbuzali: no need to introduce a thread here, you can use `map.isLocked(key)` after `map.replace`
this will dramatically decrease the test complexity

hazelcast/src/test/java/com/hazelcast/map/MapTransactionLockingTest.java
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    @Test
+    public void testTxnReplace_whenReplaceIfSameFails_keyShouldRemainUnlockedDuringTransaction() throws InterruptedException {
+        final String mapName = randomMapName();
+        final String key = "k";
+        final String value = "v";
+        final TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(1);
+        final HazelcastInstance hazelcastInstance = factory.newHazelcastInstance(getConfig());
+        IMap<String, Object> map = hazelcastInstance.getMap(mapName);
```
gurbuzali: you can pull above code to `setup`

gurbuzali:
minor comments regarding test structure
:+1:

pveentjer:
Can you do a minor title cleanup so that we get a useful short text when we list the bugs being resolved?

hazelcast/src/test/java/com/hazelcast/map/MapTransactionLockingTest.java
```diff
@@ -0,0 +1,450 @@
...
+        final CountDownLatch startAssertion = new CountDownLatch(1);
+        final CountDownLatch finishAssertion = new CountDownLatch(1);
+        final CountDownLatch finishTransaction = new CountDownLatch(1);
+
+        map.put(key, value);
+
+        new Thread(new Runnable() {
```
pveentjer: we have a HazelcastTestSupport.spawn for this btw.

pveentjer:
I agree with Ali his suggestions. Minor test cleanup is required. Apart from that it looks good.

eminn:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
```diff
@@ -215,6 +233,23 @@ public boolean removeIfSameInternal(Data key, Object value) {
...
+        try {
+            NodeEngine nodeEngine = getNodeEngine();
+            TxnUnlockOperation unlockOperation = new TxnUnlockOperation(name, key, versionedValue.version);
+            unlockOperation.setThreadId(ThreadUtil.getThreadId());
+            unlockOperation.setOwnerUuid(tx.getOwnerUuid());
+            int partitionId = nodeEngine.getPartitionService().getPartitionId(key);
+            Future<VersionedValue> future = nodeEngine.getOperationService()
+                    .invokeOnPartition(MapService.SERVICE_NAME, unlockOperation, partitionId);
+            future.get();
+            valueMap.remove(key);
```
juanavelez: There is no need for this. All instances where the unlock is being called are made because the key does not exist in the map.

eminn: The check (`valueMap.containsKey()`) is made to learn that whether lock is acquired by someone or I will be the one who will lock it first. If I'm the first one, I acquire the lock with `lockAndGet()` call which leads to a put call to the `valueMap`, so it is needed to remove the key when the key is unlocked.

juanavelez: My apologies, you are correct.

PR #6658, opened 2015-11-05
introduced init for RecordStore

gurbuzali:
The reason to introduce an `init` method is `createStorage` is overridden on `EnterpriseRecordStore` and we need to use some fields defined in `EnterpriseRecordStore` while creating storage.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
@@ -95,7 +95,9 @@ public class PartitionContainer {
         keyLoader.setMaxSize(getMaxSizePerNode(mapConfig.getMaxSizeConfig()));
         keyLoader.setHasBackup(mapConfig.getTotalBackupCount() > 0);
         keyLoader.setMapOperationProvider(serviceContext.getMapOperationProvider(name));
-        return serviceContext.createRecordStore(mapContainer, partitionId, keyLoader);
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/StorageImpl.java
```diff
@@ -36,9 +36,10 @@
  */
 class StorageImpl<R extends Record> implements Storage<Data, R> {
 
-    private final RecordFactory<R> recordFactory;
     // Concurrency level is 1 since at most one thread can write at a time.
-    private final ConcurrentMap<Data, R> records = new ConcurrentHashMap<Data, R>(1000, 0.75f, 1);
+    protected final ConcurrentMap<Data, R> records = new ConcurrentHashMap<Data, R>(1000, 0.75f, 1);
```
gurbuzali: changed to `protected` to use it on extended enterprise storages.

gurbuzali: ahmet pointed out that this change is not necessary

ahmetmircik:
:+1:  once green

eminn:
:+1:

mdogan:
See https://github.com/hazelcast/hazelcast-enterprise/commit/85a3a65032906f1afd5090100f48aab3805ab5e6

gurbuzali:
Thank you for the fix

PR #6659, opened 2015-11-05
Adding null check to AbstractPredicate and using it overall

tombujok:
Saves a lot of pain...


hazelcast/src/test/java/com/hazelcast/query/impl/predicates/PredicatesTest.java
```diff
@@ -76,6 +76,8 @@
 @Category(QuickTest.class)
 public class PredicatesTest extends HazelcastTestSupport {
 
+    private static final String DUMMY = "DUMMY_ATTRIBUTE_IGNORED";
```
jerrinot: a nitpicking: `DUMMY_ATTRIBUTE_NAME` would be better. or just an `ATTRIBUTE_NAME`

jerrinot:
a small comment only :+1: othewise

eminn:
:+1:

PR #6661, opened 2015-11-05
Fixed race-condition in IO during shutdown

jerrinot:
When a shutdown task was executed by IOThread then we do not want to continue with further processing as the channel is closed & selection key is cancelled anyway.

I'm not quite sure how to test this, perhaps @pveentjer might have an idea.


jerrinot:
Related discussion: https://groups.google.com/forum/#!topic/hazelcast/GFNtzlHwnKU

hasancelik:
verify

jerrinot:
actually I just reproduce the issue even with this fix. I'm closing it for now and will continue investigating.

pveentjer:
I'll have a look at the discussion. I know the shutdown behavior is not water tight in the sense that you can get some exception noise. Regarding the fix; i try to prevent exceptional cases from the fast path; especially since IO is very performance sensitive.

PR #6662, opened 2015-11-05
Enhanced extraction API

tombujok:
Enhanced extraction API - ValueCollector added to make the collection of extracted results easier.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/DefaultValueCollector.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/DefaultValueCollector.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.query.impl;
+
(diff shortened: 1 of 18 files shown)


hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
```diff
@@ -74,6 +74,10 @@ public MultiResult() {
         this.results = new ArrayList<T>();
     }
 
+    public MultiResult(List<T> results) {
+        this.results = new ArrayList<T>(results);
```
jerrinot: is it necessary to create a defensive copy? it's just internal API now.

hazelcast/src/main/java/com/hazelcast/query/impl/DefaultValueCollector.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.hazelcast.query.extractor.ValueCollector;
+import com.hazelcast.query.impl.getters.MultiResult;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DefaultValueCollector extends ValueCollector {
+
+    private List<Object> values = new ArrayList<Object>();
```
jerrinot: this seems to be quite wasteful. I can imagine a lot of extractor will produce just a single value.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
tombujok: It's was defensively copied with the assumption that the ValueCollector that calls this constructor will be recycled & reused so we didn't want to get this list reset when the Collector was recycled.
But for now, the collector is not reused so I can just take the reference as-is.

hazelcast/src/main/java/com/hazelcast/query/impl/DefaultValueCollector.java
tombujok: Good hint. Fixed.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
tombujok: Fixed

jerrinot:
:+1:

metanet:
:+1:

PR #6664, opened 2015-11-05
Fixed unnecessary deserializtion in ICache  near cache

sancar:


hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
```diff
@@ -0,0 +1,266 @@
...
+
+        @Override
+        public void readPortable(PortableReader reader) throws IOException {
+            deserializeCount.incrementAndGet();
+        }
+
+    }
+
+    @After
+    public void tearDown() {
```
serkan-ozal: `serializeCount` and `deserializeCount` should be reset also here since in case of error, they might left unresetted and this may lead to fail of other tests.

serkan-ozal: or better approach can be each test has its own `serializeCount` and `deserializeCount` as local variable and pass them to `assertAndReset` method. So we can run these tests with parallel test runner also.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
```diff
@@ -0,0 +1,266 @@
...
+    @Test
+    public void testDeserializationCountWith_ObjectNearCache_cacheLocalUpdatePolicy() {
+        NearCacheConfig nearCacheConfig = createNearCacheConfig(InMemoryFormat.OBJECT
+                , NearCacheConfig.LocalUpdatePolicy.CACHE);
+        prepareCache(nearCacheConfig);
+
+        final String key = randomString();
+        SerializationCountingData value = new SerializationCountingData();
+        cache.put(key, value);
+        assertAndReset(1, 0);
```
serkan-ozal: since putting to near-cache is done as async, it would be better making this assertion after `assertEquals(1, nearCache.size());` so we can be sure that deserialization count is `0` after the record is put to near-cache

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
serkan-ozal: forget second (local variable based) approach :)

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
sancar: Ok. Updating it.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
sancar: Done the resetting part.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
```diff
@@ -0,0 +1,266 @@
...
+
+    @Test
+    public void testDeserializationCountWith_BinaryNearCache_cacheLocalUpdatePolicy() {
+        NearCacheConfig nearCacheConfig = createNearCacheConfig(InMemoryFormat.BINARY
+                , NearCacheConfig.LocalUpdatePolicy.CACHE);
+        prepareCache(nearCacheConfig);
+
+        String key = randomString();
+        SerializationCountingData value = new SerializationCountingData();
+        cache.put(key, value);
```
serkan-ozal: I think, we need `assertEquals(1, nearCache.size());` check here also

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    @Test
+    public void testDeserializationCountWithoutNearCache() {
+        prepareCache(null);
+
+        SerializationCountingData value = new SerializationCountingData();
+
+        String key = randomString();
+        cache.put(key, value);
```
serkan-ozal: I think, we need assertEquals(1, nearCache.size()); check here also because we put to near-cache on put to cache

serkan-ozal:
my comments for `client-legacy` is applicable for `client` other than :+1:

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheSerializationCountTest.java
sancar: You want me to check the near-cache size in a test without near cache !??

serkan-ozal: ah sorry, i though that we are using near-cache with default configs. had not look at the test name. my bad. no need to any update on this test

serkan-ozal:
merge once green

serkan-ozal:
verify

PR #6637, opened 2015-11-03
Add namespace declaration for XmlClientConfig tests which test invalid configuration

cangencer:
Additionally, add a test for missing namespace declaration. Fixes issue #6633

PR #6675, opened 2015-11-06
Remove XML namespace injection during schema validation

cangencer:
This injection doesn't work on JDKs other than the IBM one. The user
should be required to specify the namespace in all cases.

### DIFF
--- a/hazelcast-client-legacy/src/test/java/com/hazelcast/client/config/XmlClientConfigBuilderTest.java
+++ b/hazelcast-client-legacy/src/test/java/com/hazelcast/client/config/XmlClientConfigBuilderTest.java
@@ -321,6 +321,18 @@ public class XmlClientConfigBuilderTest {
         buildConfig(xml);
     }
 
+    @Test(expected = InvalidConfigurationException.class)
+    public void testInvalidNamespace() {
+        String xml = "<hazelcast-client xmlns=\"http://foo.bar\"/>";
+        buildConfig(xml);
+    }
+
+    @Test
+    public void testValidNamespace() {
+        String xml = HAZELCAST_CLIENT_START_TAG + "</hazelcast-client>";
+        buildConfig(xml);
+    }
+
     private void testXSDConfigXML(String xmlFileName) throws SAXException, IOException {
         SchemaFactory factory = SchemaFactory.newInstance(XMLConstants.W3C_XML_SCHEMA_NS_URI);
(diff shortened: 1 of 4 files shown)


cangencer:
@mtopolnik @bilalyasar @pveentjer One of the tests on my previous PR #6637 failed on IBM JDK and further investigation revealed that we had code to make sure XMLs without namespace would be validated, however it was not working as intended.

mtopolnik:
But why is the XML namespace inserted into the DOM tree? If the declaration exists in the XML file, then after parsing all the nodes wil have it. I don't think we should fix the problem for the user if they left out the NS declaration.

cangencer:
@mtopolnik There was already some code inject the NS declaration if missing, so it seems that was the intention from the beginning. However it was not working as intended.

mtopolnik:
OK, but now that it's brought up, it seems like an awkward policy. It can cause weird and counterintuitive errors. For example, some XML element declares a slightly wrong namespace---then it fails, but it doesn't fail with no namespace at all. Users are not accustomed to XML files acquiring a namespace out of thin air.

mtopolnik:
However, if the current Hazelcast is happy without XML declaration, then insisting on it would be breaking behavior.

cangencer:
@mtopolnik It is not happy currently except if you are using IBM JDK, which seems to treat setting the NS attribute directly differently. We should either support the config without NS fully (which was this PR) or require the user to have it always.

mtopolnik:
When I say "currently" I mean "what the users currently have" (the released version).

mtopolnik:
More importantly: does the currently released version tolerate the absence of namespace declaration? If it doesn't, let us not introduce it now. Just delete all the patchwork code and let it fail, just like the unit tests were asserting.

cangencer:
@mtopolnik it will fail in the released version without namespace as well. In that case I will just remove all the code relating to namespace injection.

bilalyasar:
@cangencer, can you write minimum xml file that works ( if this PR is merged) ?

mtopolnik:
Let me add one more point: specifying the schema location should not be required (I see that was also injected into the DOM tree). The unit tests were passing without it.

So the minimum XML should be this:

```xml
<hazelcast xmlns="http://www.hazelcast.com/schema/config" />
```

cangencer:
@mtopolnik yes that was redundant. I've now added some more tests.

hasancelik:
verify

bilalyasar:
:+1:

PR #6677, opened 2015-11-06
Reverting a precondition in AbstractPredicate introduced by #6659

jerrinot:
Reverting a precondition in AbstractPredicate introduced by #6659
PredicateBuilder may init the abstract predicate with attributeName set to null.

It's a quick revert to unblock PR builder.


mtopolnik:
:+1: 

Does it pass CheckStyle? :)

jerrinot:
@mtopolnik: It does. Now. Thanks!

ahmetmircik:
@jerrinot What about the tests in #6659?

jerrinot:
@ahmetmircik: The tests are passing fine. 
They should not use null is an attribute in the 1st place. I think @tombujok is already investigating it.

ahmetmircik:
@jerrinot i see thanks

tombujok:
:+1:

PR #6678, opened 2015-11-06
Removed unnecessary `CacheLoader` dependency on client and server side at cache proxies

### LINKED ISSUE #6676: No need to using `CacheLoader` inside client/server side cache proxies
At the moment both of client and server side cache proxies create configured `CacheLoader` through their factories however they are not used but just created, validated and closed. So seems that no need to using them at cache proxy level since they are handled at record store level at server side.

See http://stackoverflow.com/questions/33556896/why-hazelcast-cacheloader-class-needs-to-be-visible-by-all-clients

serkan-ozal:
* Removed unnecessary `CacheLoader` creation and closing codes at client and server side cache proxies
* If factory of `CacheLoader` and `CacheWriter` is `HazelcastInstanceAware`, `HazelcastInstance` is injected
* Refactor on `ClientReadWriteThroughJCacheTests` so tests are using mock instances (client and server) and can run with parallel runner. 

Fixes #6676


asimarslan:
If TCK pass then :+1:

serkan-ozal:
@asimarslan yes, both of client and server instance based TCK tests pass

Issue #5983, opened 2015-08-17
hazelcast map - slow load on startup - 2 seconds for each map


sarvesh279:
Class:
class HazelcastUserAssignmentStore implements MapStore<String, UserAssignmentCache>, MapLoader<String, UserAssignmentCache>, MapLoaderLifecycleSupport

the above class has all necessary overridden methods.

On server startup, there is 2 sec time lag after data load to each map.
The sequence of methods called,
1) init
2) loadAllKeys
3) loadAll

All the three methods are executed less than 5ms. Then the system waits for 2 sec to load the next map. How to avoid this delay?

jerrinot:
Hi @sarvesh279, 
thank you for reporting the issue. It seems like a duplicate of #5349. The fix is already pending, we are aiming to include it in the 3.5.2 release.

jerrinot:
I'm closing this one in favour of #5349. Feel free to re-open it if you feel it's not a duplicate.

sarvesh279:
release date for 3.5.2?

jerrinot:
Hi @sarvesh279: It's out, but be careful, it's still hot! ;-)
http://hazelcast.org/download/
https://github.com/hazelcast/hazelcast/tree/v3.5.2

(It might take a while until it's available Maven central)

sarvesh279:
I get the same issue in 3.5.2 version.

On server startup, there is 2 sec time lag after data load to each map.
The sequence of methods called,
1) init
2) loadAllKeys
3) loadAll

All the three methods are executed less than 5ms. Then the system waits for 2 sec to load the next map.

sarvesh279:
@jerrinot is the issue fixed in 3.5.2?

jerrinot:
hm, it's fixed in 3.5.2. 
Maybe there is another source of lag. @ajermakovics: Can you please look at this?

sarvesh279:
@ajermakovics Any updates?

ajermakovics:
This test case passes and doesn't have a delay: https://github.com/hazelcast/hazelcast/issues/5632 
@sarvesh279 Perhaps you have a test case or can create one?

PR #6689, opened 2015-11-09
Avoiding sleep when loading empty maps in eager mode

eminn:
Rebased version of @ajermakovics 's PR (#5677).

Issue #5349 , #5983

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionCheckIfLoadedOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionCheckIfLoadedOperation.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.map.impl.operation;
 
+import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.map.impl.MapServiceContext;
 import com.hazelcast.map.impl.recordstore.RecordStore;
 import com.hazelcast.nio.ObjectDataInput;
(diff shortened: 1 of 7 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionCheckIfLoadedOperation.java
```diff
@@ -42,14 +44,26 @@ public PartitionCheckIfLoadedOperation(String name, boolean doLoad) {
         this.doLoad = doLoad;
     }
 
+    public PartitionCheckIfLoadedOperation(String name, boolean doLoad, boolean waitForKeyLoad) {
```
jerrinot: just a small one: the other 2 constructors should call this one. 
```java
    public PartitionCheckIfLoadedOperation(String name) {
        this(name, false);
    }

    public PartitionCheckIfLoadedOperation(String name, boolean doLoad) {
        this(name, doLoad, false);
    }

    public PartitionCheckIfLoadedOperation(String name, boolean doLoad, boolean wait) {
        super(name);
        this.doLoad = doLoad;
        this.waitForKeyLoad = waitForKeyLoad;
    }
```

I think it makes code cleaner.

jerrinot:
it seems it's just a simple forward-port of the PR from the maintenance branch. 
:+1:

gurbuzali:
:+1:

PR #6691, opened 2015-11-09
Hibernate UpdateEntryProcessor test on real network

mustafaiman:
Tests update operation works properly in a cluster.


pveentjer:
:+1:  And since it is just a test update, I'll merge this one.

PR #6693, opened 2015-11-09
Revived: Map/reduce type variance improvements

mtopolnik:
Revives #4872, which needed rebasing.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/mapreduce/impl/AbstractJob.java
+++ b/hazelcast/src/main/java/com/hazelcast/mapreduce/impl/AbstractJob.java
@@ -75,7 +75,8 @@ public abstract class AbstractJob<KeyIn, ValueIn>
     }
 
     @Override
-    public <KeyOut, ValueOut> MappingJob<KeyIn, KeyOut, ValueOut> mapper(Mapper<KeyIn, ValueIn, KeyOut, ValueOut> mapper) {
+    public <KeyOut, ValueOut> MappingJob<KeyIn, KeyOut, ValueOut> mapper(
+            Mapper<KeyIn, ValueIn, KeyOut, ValueOut> mapper) {
         isNotNull(mapper, "mapper");
         if (this.mapper != null) {
             throw new IllegalStateException("mapper already set");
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/mapreduce/impl/AbstractJob.java
```diff
@@ -85,7 +86,7 @@ public AbstractJob(String name, JobTracker jobTracker, KeyValueSource<KeyIn, Val
     }
 
     @Override
-    public Job<KeyIn, ValueIn> onKeys(Iterable<KeyIn> keys) {
+    public Job<KeyIn, ValueIn> onKeys(Iterable<? extends KeyIn> keys) {
```
Donnerbart: This method seems to lack code coverage: https://hazelcast-l337.sonar.cloudbees.com/dashboard/index/12063?display_title=true&metric=coverage

hazelcast/src/main/java/com/hazelcast/mapreduce/impl/AbstractJob.java
```diff
@@ -138,7 +139,7 @@ protected void prepareKeyPredicate() {
         }
     }
 
-    private void addKeys(Iterable<KeyIn> keys) {
+    private void addKeys(Iterable<? extends KeyIn> keys) {
```
Donnerbart: No code coverage as well: https://hazelcast-l337.sonar.cloudbees.com/dashboard/index/12063?display_title=true&metric=coverage

Donnerbart:
It's not visible in SonarQube if the interfaces are used (so I had to check that in IntelliJ). `Job` is implemented by `AbstractJob` and `MappingJob` by the inner class `AbstractJob.MappingJobImpl`.

There is a lot of missing code coverage in the `AbstractJob`, especially the inner class `MappingJobImpl`. This might be an JaCoCo issue or real missing code coverage.

Beside that the compiler should tell us if the changes are okay. And since all changed classes are somewhere used in tests we should see a failure in the PR builder.

So :+1: (once green).

mtopolnik:
Even though I am the formal submitter, this is not my code so I should be allowed to vote.

:+1:

PR #6698, opened 2015-11-10
Async connection creation for client invocations

sancar:
When an invocation needs to be done and connection to destination
is not established yet, connection establishment used to be a blocking
operation. With this pr, if connection is not available yet, we only
trigger a new connect task async and related user request will be
handled by retry mechanism.

Secondly, invocation timeout handling is changed. # of times an
invocation made is used to be count with 1 second sleeps. Since
invocation itself can take time, user configured timeout was
used to be violated. It is now checked by comparing
System.currentTimeMillis and starting time.

backport https://github.com/hazelcast/hazelcast/pull/6702


sancar:
run-lab-run

PR #4768, opened 2015-03-07
to add getAll/putAll methods in TransactionalMap interface; issue #4698

dsukhoroslov:
to explain the changes I did in PR:
- moved method getAll(Set<K> keys) from IMap to it's parent BaseMap interface;
- moved method keySet() from IMap to BaseMap interface;
- moved method keySet(Predicate p) from IMap to BaseMap interface;
- added putAll(Map<K, V> entries) method to BaseMap interface;
- removed method keySet() from TransactionalMap interface (now it is in it's parent BaseMap interface);
- removed method keySet(Predicate p) from TransactionalMap interface;
- implemented getAll/putAll methods in TransactionalMapProxy class;
- implemented getAll/putAll methods in ClientTxnMapProxy class;
- extended MapPortableHook class with two new requests to process getAll/putAll calls;
- implemented getAll/putAll requests in TxnMapRequestWithKeySet/TxnMapRequestWithDataMap classes;
- extended AbstractTxnMapRequest.innerCallInternal method with getAll/putAll request processing;
- added testTxnMapPutGetAll test to ClientTxnMapTest class;

Issue #4698, opened 2015-03-02
getAll/utAll methods in TransactionalMap interface


dsukhoroslov:
It would be beneficial to have most common Map methods in a base interface (BaseMap) which are inherited by both IMap and TransactionalMap. Forum discussion is here: https://groups.google.com/forum/#!topic/hazelcast/6Dvz14D6nU8. 
I'm ready to provide a PR for this. I've generalized methods keySet(), keySet(Predicate p) into the BaseMap interface too.
We can discuss implementation details here. As a first attempt I just extend TransactionalMapProxy class with getAll/putAll methods which use getInternal/putInternal sequentially. Is it ok or it'll be better to implement getAllInternal/putAllInternal and corresponding operations for them?

Thanks, Denis.

noctarius:
Sounds like a good thing to me. What do others think?

dsukhoroslov:
Honestly, I'd extend BaseMap interface from the regular java.util.Map. It is probably the major cache structure anyway, doesn't matter it the cache/map transactional or not. But this is up to you of course. So, what about the implementation? If the sequential implementation is ok then I'll submit the PR, if not then I'll extend it further with getAll/putAll operations.

dsukhoroslov:
I submit a PR for this: #4768

mmedenjak:
Related : https://github.com/hazelcast/hazelcast/issues/9080

PR #6705, opened 2015-11-10
add getAll/putAll methods in TransactionalMap interface

mtopolnik:
Revives #4768 which needed rebasing.

Related to issue #4698.

Note that this PR contains client code from the days before the New Client. Therefore the implementation of the added getAll and putAll methods needs to be filled in.

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/txn/proxy/ClientTxnMapProxy.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/txn/proxy/ClientTxnMapProxy.java
@@ -22,13 +22,17 @@ import com.hazelcast.map.impl.MapKeySet;
 import com.hazelcast.map.impl.MapService;
 import com.hazelcast.map.impl.MapValueCollection;
 import com.hazelcast.map.impl.client.TxnMapRequest;
+import com.hazelcast.map.impl.client.TxnMapRequestWithDataMap;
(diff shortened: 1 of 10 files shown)


bwzhang2011:
@mtopolnik, thanks a lot for batch operation provided. how about batch support for TransactionQueue
for addAll support

enesakar:
Old code was doing puts and gets one by one in for loop. Do you plan to improve/fix this inefficiency?

mtopolnik:
I am just the messenger. I have rebased the code after the original author went unresponsive. If we are not interested in this work anymore, let us decide to close.

pveentjer:
@mtopolnik sorry for the long delay. What are we going to do with this PR? Probably the module owners should get involved here.

mtopolnik:
I think the proposed feature is interesting, but not the naïve implementation with no batching. In my view this could be turned into an _issue_ with the enhancement request and this PR could be closed as probably no code it contributes will be used in the actual solution.

tombujok:
I agree with Marko, without batching is just a dummy wrapper that may confuse people who could expect that it's implemented in a performant way.
I would close it and add to backlog should we want to add it in the future.

LoneRifle:
I would be happy to take this on and rework the impl, having had a hand in recent changes to getAll/putAll on both legacy and current ClientMapProxy. Having said that, I would like to take this over only after we're sure #7129 needs no further changes and can be merged.

PR #6685, opened 2015-11-08
Fix migration state problems on node shutdown and join

metanet:
* Do not trigger RepartitioningTask if cluster state is not active. It causes infinite retries and prevents nodes to shutdown since they think that there is an ongoing migration
* Introduce NodeExtension.isStartCompleted() method and use it in PartitionStateOperation instead of checking NodeState. It is a no-op method for OSS and has an implementation on enterprise.
* Send initial partition table with finalize-join operation 

Fixes #6663
Fixes #6655
Fixes #6656

PR #6706, opened 2015-11-11
Reject migration operations when node is not started yet

mdogan:
- Newly joining node should reject migration operations when node is not started yet,
for example it's still loading hot-restart data.
- Pause migration process for a small amount of time, if a migration attempt is failed.
 Otherwise, migration failures can do a busy spin until migration problem is resolved.
 Migration can fail either a node's just joined and not completed start yet or it's just left the cluster.

Follow up #6685

_There are & will be tests to verify this fix in enterprise repo hot-restart test._


metanet:
:+1:  very good finding :)

gurbuzali:
:+1:

PR #6707, opened 2015-11-11
Map.set update event missing old value

pveentjer:
The SetOperation didn't store the old value unlike the PutOperation. So when
the event is send, the old value isn't set correct in the event.

This pr fixes:
https://github.com/hazelcast/hazelcast-enterprise/issues/461

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/SetOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/SetOperation.java
@@ -21,7 +21,7 @@ import com.hazelcast.nio.serialization.Data;
 
 public class SetOperation extends BasePutOperation {
 
-    private boolean newRecord;
+    private boolean response;
 
     public SetOperation() {
     }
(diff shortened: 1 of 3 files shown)


pveentjer:
@ahmetmircik can you have a look at this one.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/SetOperation.java
```diff
@@ -32,17 +32,19 @@ public SetOperation(String name, Data dataKey, Data value, long ttl) {
...
     public void afterRun() {
-        eventType = newRecord ? EntryEventType.ADDED : EntryEventType.UPDATED;
+        eventType = response ? EntryEventType.ADDED : EntryEventType.UPDATED;
         super.afterRun();
     }
 
     @Override
     public void run() {
-        newRecord = recordStore.set(dataKey, dataValue, ttl);
+        Object oldValue = recordStore.put(dataKey, dataValue, ttl);
```
ahmetmircik: `recordStore.set` should be used here because `recordStore.put` loads old-value from map-store.

gurbuzali:
I think https://github.com/hazelcast/hazelcast-enterprise/pull/484 already addressed the issue, I'm closing this one

PR #6708, opened 2015-11-11
Added javadoc for reflecting this change on the Map getAll and putAll requests meaning

ihsandemir:
Map  getAll and putAll javadoc update.


sancar:
:+1:

PR #6709, opened 2015-11-11
Fix distributed object listener unnecessary proxy creation 

### LINKED ISSUE #5554: IMap.destroy() - map is destroyed and then immediately recreated
Creating this issue as suggested by @bilalyasar (see [discussion on the Hazelcast Google Groups forum] (https://groups.google.com/forum/#!topic/hazelcast/G1P43tORJPA))

Hazelcast version that you use: ```3.4.2```, ```3.5```
Cluster size: ```1```
Number of the clients: ```1```
Version of Java: ```OracleJDK 1.7.0_80, x86_64```, ```OracleJDK 1.8.0_45, x86_64```
Operating system: ```OS X Yosemite 10.10.3 (14D136)```
Logs and stack traces: See the console output from executing the TestNG test case below
Detailed description of the steps to reproduce your issue: See the TestNG test case below
Unit test with the hazelcast.xml file: 

```java
import static org.testng.Assert.assertEquals;
import static org.testng.Assert.assertNotNull;

import org.testng.annotations.AfterClass;
import org.testng.annotations.BeforeClass;
import org.testng.annotations.Test;

import com.hazelcast.core.DistributedObjectEvent;
import com.hazelcast.core.DistributedObjectListener;
import com.hazelcast.core.IMap;

public class IMapDestroyTest {

	private HazelcastInstance server;
	private HazelcastInstance client;

	@BeforeClass
	public void setUpBeforeClass() {
		server = Hazelcast.newHazelcastInstance();
		sleep(2000);
		client = HazelcastClient.newHazelcastClient();
		client.addDistributedObjectListener(new DistributedObjectListener() {
			@Override public void distributedObjectDestroyed(final DistributedObjectEvent event) {
				System.out.println("\n\tdistributedObjectDestroyed(): " + event);
			}
			@Override public void distributedObjectCreated(final DistributedObjectEvent event) {
				System.out.println("\n\tdistributedObjectCreated(): " + event);
			}
		});
	}

	@AfterClass
	public void tearDownAfterClass() {
		client.shutdown();
		server.shutdown();
	}

	@Test
	public void testIMapDestroy() {
		System.out.print("Creating 'foo' map... ");
		IMap<String, String> fooMap = client.getMap("foo");
		assertNotNull(fooMap);
		System.out.println("done.");
		System.out.println("All known distributed objects: " + client.getDistributedObjects());
		System.out.print("Inserting an entry... ");
		fooMap.put("fu", "bar");
		System.out.print("done.\nRetrieving the entry... ");
		String value = fooMap.get("fu");
		assertEquals(value, "bar");
		System.out.print("done.\nDestroying the map... ");
		fooMap.destroy();
		System.out.println("done.");
		System.out.println("All known distributed objects: " + client.getDistributedObjects());
	}

	private void sleep(final int millis) {
		try {
			Thread.sleep(millis);
			Thread.currentThread().interrupt();
		}
	}
}
```

Console output:
```
Creating 'foo' map...
        distributedObjectCreated(): DistributedObjectEvent{eventType=CREATED, serviceName='hz:impl:mapService', distributedObject=IMap{name='foo'}}
done.
All known distributed objects: [IMap{name='foo'}]
Inserting an entry... done.
Retrieving the entry... done.
Destroying the map...
        distributedObjectDestroyed(): DistributedObjectEvent{eventType=DESTROYED, serviceName='hz:impl:mapService', distributedObject=IMap{name='foo'}}
        distributedObjectCreated(): DistributedObjectEvent{eventType=CREATED, serviceName='hz:impl:mapService', distributedObject=IMap{name='foo'}}
done.
All known distributed objects: [IMap{name='foo'}]
```

eminn:
Distributed object listener no longer re-creates the proxy on DESTROYED event and proxy creation made lazily. 

Fixes #5554

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/LazyDistributedObjectEvent.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/LazyDistributedObjectEvent.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi.impl.listener;
+
+import com.hazelcast.client.spi.ProxyManager;
+import com.hazelcast.core.DistributedObject;
+import com.hazelcast.core.DistributedObjectEvent;
+
+public final class LazyDistributedObjectEvent extends DistributedObjectEvent {
+
+    private ProxyManager proxyManager;
+
+    /**
+     * Constructs a DistributedObject Event.
+     *
+     * @param eventType         The event type as an enum {@link EventType} integer.
+     * @param serviceName       The service name of the DistributedObject.
+     * @param objectName        The name of the DistributedObject.
+     * @param distributedObject The DistributedObject for the event.
+     * @param proxyManager      The ProxyManager for lazily creating the proxy if not available on the client.
+     */
+    public LazyDistributedObjectEvent(EventType eventType, String serviceName, String objectName,
+                                      DistributedObject distributedObject,
+                                      ProxyManager proxyManager) {
+        super(eventType, serviceName, objectName, distributedObject);
+        this.proxyManager = proxyManager;
+    }
+
+    @Override
+    public DistributedObject getDistributedObject() {
+        distributedObject = super.getDistributedObject();
+        if (distributedObject == null) {
+            distributedObject = proxyManager.getOrCreateProxy(getServiceName(), (String) getObjectName());
+        }
+        return distributedObject;
+    }
+
+}
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/LazyDistributedObjectEvent.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/LazyDistributedObjectEvent.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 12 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/proxy/DistributedObjectListenerTest.java
```diff
@@ -65,6 +69,13 @@ public void distributedObjectDestroyed(DistributedObjectEvent event) {
         assertOpenEventually(createdLatch, 10);
         topic.destroy();
         assertOpenEventually(destroyedLatch, 10);
+        final Collection<DistributedObject> distributedObjects = client.getDistributedObjects();
```
sancar: Should not be this inside assertTrueAllTheTime ?

sancar:
How about backport to maintenance branch ?

hazelcast-client/src/test/java/com/hazelcast/client/proxy/DistributedObjectListenerTest.java
eminn: Indeed it should.

eminn:
On the way

hazelcast/src/main/java/com/hazelcast/core/DistributedObjectEvent.java
```diff
@@ -25,11 +25,11 @@
  */
 public class DistributedObjectEvent {
 
-    private EventType eventType;
+    protected EventType eventType;
```
jerrinot: this is dangerous is it effectively makes the fields to be part of API. I'd avoid it if possible.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/LazyDistributedObjectEvent.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi.impl.listener;
+
+import com.hazelcast.client.spi.ProxyManager;
+import com.hazelcast.core.DistributedObject;
+import com.hazelcast.core.DistributedObjectEvent;
+
+public class LazyDistributedObjectEvent extends DistributedObjectEvent {
```
jerrinot: it should be final?

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/listener/LazyDistributedObjectEvent.java
```diff
@@ -0,0 +1,59 @@
...
+                                      ProxyManager proxyManager) {
+        super(eventType, serviceName, distributedObject);
+        this.objectName = objectName;
+        this.proxyManager = proxyManager;
+    }
+
+    @Override
+    public DistributedObject getDistributedObject() {
+        if (EventType.DESTROYED.equals(eventType)) {
+            return null;
```
jerrinot: have you considered throwing an exception here? in every case the behaviour should be explicitly documented in contract.

jerrinot:
:+1: on green

jerrinot:
backport of #6709

sancar:
:+1:

PR #6711, opened 2015-11-11
Remove the distinction between async and sync protocol calls

cangencer:
Whether a client API is async or sync is a client side concern and should
not be reflected within the protocol. Cases where async and sync were
separate calls in the protocol have been merged.


hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/client/AddRequest.java
```diff
@@ -68,12 +73,12 @@ public Permission getRequiredPermission() {
...
     @Override
     public Object[] getParameters() {
-        return new Object[]{item};
+        return new Object[]{item, overflowPolicy};
     }
 
     @Override
     public String getMethodName() {
-        return "add";
+        return "addAsync";
```
sancar: Is this intended ? Don't we want to return "add" here ?

cangencer:
@sancar: yes that was a mistake! updated the PR now.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
```diff
@@ -103,7 +102,7 @@ public long remainingCapacity() {
     public long add(E item) {
         checkNotNull(item, "item can't be null");
 
-        AddRequest request = new AddRequest(getName(), toData(item));
+        AddRequest request = new AddRequest(getName(), toData(item), OverflowPolicy.OVERWRITE);
```
sancar: This looks like change in the behaviour. Before in prepareOperation method OverflowPolicy.FAIL was used , now it looks like we are using  OverflowPolicy.OVERWRITE .

cangencer: I just made it the same as the Add for the new client:

```@Override
    public long add(E item) {
        checkNotNull(item, "item can't be null");

        Data element = toData(item);
        ClientMessage request = RingbufferAddCodec.encodeRequest(name, OverflowPolicy.OVERWRITE.getId(), element);
        ClientMessage response = invoke(request, partitionId);
        RingbufferAddCodec.ResponseParameters resultParameters = RingbufferAddCodec.decodeResponse(response);
        return resultParameters.response;
    }```

Comment (sancar) [inline hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java]: Interesting, then we new client and old client is not matching . 
@pveentjer can you take a look ? Which OverflowPolicy should we use in sync add method ?

Comment (cangencer) [inline hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java]: I think looks like a bug in the old client, as the Javadoc says

```  /**
     * Adds an item to the tail of the Ringbuffer. If there is no space in the Ringbuffer, the add will overwrite the oldest
     * item in the ringbuffer no matter what the ttl is. For more control on this behavior, check the
     * {@link #addAsync(Object, OverflowPolicy)} and the {@link OverflowPolicy}.
```

sancar:
:+1:

hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
pveentjer: I'll have a look.

pveentjer:
@cangencer For my understanding; apart from the potential issue with the policy in the ringbuffer, does this PR solve a bug or is this a code cleanup?

cangencer:
@pveentjer It's more of a protocol clean up. The *Async calls at the protocol level are redundant and making this change now before 3.6 release will ensure we will be able to stay backwards compatible.

pveentjer:
Tricky one. I love cleanups and nothing makes me happier than removing redundant code. And I understan the 'do it now instead of needing to provide backwards compatibility'. But on the other side we are in bug fixing period and code cleanups should be postponed till 3.6 has been released.  So I'm a bit in a split here.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/ringbuffer/RingbufferAddAllMessageTask.java
```diff
@@ -76,7 +76,7 @@ public Permission getRequiredPermission() {
 
     @Override
     public String getMethodName() {
-        return "addAllAsync";
```
pveentjer: Can this lead to security problems? Because one can't control the async methods independently. 

To me it doesn't make sense to provide granularity on such a level; but I'm wondering.

hazelcast/src/main/java/com/hazelcast/map/impl/client/MapGetRequest.java
```diff
@@ -126,9 +126,6 @@ public String getDistributedObjectName() {
 
     @Override
     public String getMethodName() {
```
pveentjer: Can this lead to security of backwards incompatibility problems? This method name is afaik exposed in the security mechanism, so by changing it, the security mechanism is influenced and this could lead to sneaky incompatibility issues.

cangencer: I think the only issue is if you are doing security filtering by the async method, but not the sync one, which I can't see why someone would want to do. The security filtering should be done based on the actual operation being done, not what method name is being called in the client. This is even more true when we have multiple clients and not just Java.

gurbuzali: yes this could lead to backwards incompatibility problems but we don't have a customer using this feature afaik (need to check). we've introduced this interceptor for a particular prospect who didn't purchase us at the end. 
I'm in favor of removing all those async method interceptors, if no one is using it.
the sooner we get rid of them the better

pveentjer:
The technology guy in me also wants to get rid of duplication as much as possible. 

My main concerns are the following:
- it isn't a bug fix, so it is a violation of the bug fixing period
- it isn't done consistently? So do we replace all async ones by sync ones?
- even though we don't know of any customer using it; perhaps there are some and to sneak it on could lead to a big security violation. Imagine they added security interceptor for async call and they only do async calls; but now the method is changed and their security mechanism should either be too strict or too relaxed.

ihsandemir:
looks good except for Sancar's comment on OverflowPolicy.FAIL. :+1:

cangencer:
* It could be considered a protocol bug, if there is such a thing
* Yes it is done consistently
* Security is an enterprise only feature and we are not aware of any customer using it..

enesakar:
because it fixes the protocol, we can not postpone this.

PR #6713, opened 2015-11-11
Fix for persisting Nonexistent ExecutorServiceStats

### LINKED ISSUE #6474: LocalExecutorStats are still persisted after IExecutorService.destroy()
IExecutorService.destroy() does not work when invoked on the same client instance which initialised the service - the service is still seen in cluster. When invoked from a new client instance, destroy removes the service from the cluster.

Do the following:
        HazelcastInstance hazelcastInstance = HazelcastClient.newHazelcastClient();
        hazelcastInstance.getExecutorService(NAME).execute(new TestMiscRunnable());
/****** At this point, ManagementCentre shows the service under "Executors", which is normal ******/

        Thread.sleep(15000);
        hazelcastInstance.getExecutorService(NAME).destroy();

/****** The service was expected to be removed from the cluster but ManCentre shows the service still present ******/

Now, create another HazelcastInstace and try to destroy the service and it works normal. Do the following:

        HazelcastInstance hazelcastInstance = HazelcastClient.newHazelcastClient();
        hazelcastInstance.getExecutorService(NAME).destroy();

/******** And at this point, the service is removed from the cluster and can not be seen in Management Centre ********/

This is consistently reproducible in 3.4.1.

In 3.5.3, IExecutorService.destroy() does not work at all - either from same client instance or a new one, the service instance always remains in the cluster. In other words, once a service is created in 3.5.3, currently there is no means to destroy it.

muratayan:
Fixes #6474

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
+++ b/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
@@ -136,6 +136,7 @@ public class DistributedExecutorService implements ManagedService, RemoteService
     public void destroyDistributedObject(String name) {
         shutdownExecutors.remove(name);
         executionService.shutdownExecutor(name);
+        statsMap.remove(name);
     }
 
     LocalExecutorStatsImpl getLocalExecutorStats(String name) {


eminn:
Will you backport this to maintenance branch ?

muratayan:
I think it'd be useful, so yes.

PR #6714, opened 2015-11-11
Preparation for hot restart cache integration

mdogan:
To support hot-restart, a cache record can be a tombstone. For hot-restart,
when a record is requested to be removed (either explicitly or by eviction or expiration),
only it's value is cleared and it becomes a tombstone.
A tombstone is only removed when requested by hot-restart store.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
@@ -861,10 +861,10 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
         long now = Clock.currentTimeMillis();
         Object value = null;
         R record = records.get(key);
-        final boolean isExpired = processExpiredEntry(key, record, now);
+        processExpiredEntry(key, record, now);
 
         try {
-            if (record == null || isExpired) {
+            if (recordNotExistOrExpiredOrTombstone(record, now)) {
                 if (isStatisticsEnabled()) {
                     statistics.increaseCacheMisses(1);
                 }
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
```diff
@@ -1367,7 +1369,7 @@ public Object invoke(Data key, EntryProcessor entryProcessor, Object[] arguments
             record = null;
         }
         if (isStatisticsEnabled()) {
-            if (record == null || isExpired) {
+            if (record == null || record.isTombstone() || isExpired) {
```
mtopolnik: I would suggest extracting this repeated check to a method

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheReplicationOperation.java
```diff
@@ -128,10 +128,9 @@ protected void writeInternal(ObjectDataOutput out)
             for (Map.Entry<Data, CacheRecord> e : cacheMap.entrySet()) {
                 final Data key = e.getKey();
                 final CacheRecord record = e.getValue();
-                final long expirationTime = record.getExpirationTime();
 
-                // If entry is already expired we skip it
-                if (expirationTime > now) {
+                // If entry is already expired or it's tombstone, we just skip it
+                if (!record.isExpiredAt(now) && !record.isTombstone()) {
```
mtopolnik: Consider rewriting to `if (record.isExpired(now) || record.isTombstone()) continue;` Then it nicely matches the comment, which can be deleted as redundant.

mtopolnik:
+1

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheReplicationOperation.java
mdogan: Make sense.

gurbuzali:
:+1:

PR #6715, opened 2015-11-11
Simplify getBooleanValue

mtopolnik:
`getBoolean()` was needlessly permissive and complex because XML Schema already prevents anything else but "true" and "false"

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/AbstractXmlConfigHelper.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/AbstractXmlConfigHelper.java
@@ -58,6 +58,7 @@ import java.util.Properties;
 
 import static com.hazelcast.nio.IOUtil.closeResource;
 import static com.hazelcast.util.StringUtil.upperCaseInternal;
+import static java.lang.Boolean.parseBoolean;
 
 /**
  * Contains Hazelcast Xml Configuration helper methods and variables.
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/config/AbstractXmlConfigHelper.java
```diff
@@ -288,10 +289,7 @@ public static String cleanNodeName(final Node node) {
     }
 
     protected static boolean getBooleanValue(final String value) {
-        String lowerCaseValue = StringUtil.lowerCaseInternal(value);
-        return "true".equals(lowerCaseValue)
-                || "yes".equals(lowerCaseValue)
-                || "on".equals(lowerCaseValue);
+        return parseBoolean(value);
```
jerrinot: the old implementation always uses `Locale.ENGLISH` for a conversion to lower-case. 

The new one will use whatever is platform default. Can't this lead to issues with some locales? I know Turkish locale has somewhat unusual rules for conversion between i / I.

mtopolnik: Actually `equalsIgnoringCase` is black art all on its own, it doesn't rely just on `toLowerCase`. But I changed it back just to remove this worry.

jerrinot:
:+1:

pveentjer:
:+1:

PR #6716, opened 2015-11-11
Avoid starting threads from constructor

mtopolnik:
The threads access the object being constructed, therefore starting them from the constructor results in data races.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
@@ -183,6 +183,10 @@ public final class OperationServiceImpl implements InternalOperationService, Pac
         metricsRegistry.scanAndRegister(this, "operation");
     }
 
+    public void start() {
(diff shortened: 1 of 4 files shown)


hasancelik:
run-lab-run

mtopolnik:
verify

pveentjer:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
```diff
@@ -98,7 +98,7 @@
     private static final long TERMINATION_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(10);
 
     final InvocationRegistry invocationsRegistry;
-    final OperationExecutor operationExecutor;
```
jerrinot: wouldn't it be better to add lifecycle method to the interface?

jerrinot:
:+1: with a small comment. I don't think it's a big deal here.

mtopolnik:
verify

mtopolnik:
verify

PR #6718, opened 2015-11-12
Fixes OperationTimeoutException on long running executor tasks

### LINKED ISSUE #5564: OperationTimeoutException when using MultiExecutionCallback for long running tasks
Hi,

as issue #4398 is already closed and already fixes most of the issues mentioned there, I open a new issue here.

The problem described in issue #4398 is fixed in all cases, except when using MultiExecutionCallback to wait for completion of long running tasks.

Here is the test that shows the problem:

```java
package com.nm.test.hazelcast.task;

import com.hazelcast.core.*;
import java.io.Serializable;
import java.util.ArrayList;
import java.util.List;
import java.util.Map;
import java.util.concurrent.CountDownLatch;

/**
 * Test if a task that runs longer than the call timeout is executed correctly when using
 * callbacks.
 */
public class TestTask4 extends TestCase {

	private static final Logger logger = Logger.getLogger(TestTask4.class);

	private static final long timeoutMs = 3000;

	@Override
	protected void setUp() throws Exception {

		// configure logging
		}
	}

	public void testTask4() throws Exception {

		// create hazelcast config

		// set call timeout to 3 seconds to make the problem appear quicker
		config.setProperty("hazelcast.operation.call.timeout.millis", String.valueOf(timeoutMs));

		// start two hazelcast instances
		HazelcastInstance hcInstance2 = Hazelcast.newHazelcastInstance(config);
		logger.info("Hazelcast instances started.");

		// var to collect execution results
		List<String> results = new ArrayList<String>();

		// try-finally to stop hazelcast
		try {

			// create callback
			CountDownLatch latch = new CountDownLatch(1);
			Callback callback = new Callback(latch);

			// submit the long-running task
			logger.info("Executing task...");
			IExecutorService executorService = hcInstance1.getExecutorService("default");
			executorService.submitToAllMembers(new SleepingTask(), callback);
			logger.info("Task submitted.");

			// wait for all results
			logger.info("Waiting for results...");
			callback.await();
			for (Object value : callback.values.values()) {
				results.add(value.toString());
			}
			logger.info("Results received.");

		} catch (Exception e) {

			// log and rethrow exception
			logger.error("Exception: " + e.getMessage(), e);
			throw e;

		} finally {

			// shutdown
			hcInstance1.getLifecycleService().shutdown();
			hcInstance2.getLifecycleService().shutdown();
		}

		// ensure success
		for (String result : results) {
			if (!result.equals("Success")) {
				fail("Non-success result: " + result);
			}
		}

		logger.info("Test done.");
	}

	private static class SleepingTask implements Callable<String>, Serializable {

		private static final long serialVersionUID = 1L;

		@Override
		public String call() throws Exception {
			Thread.sleep(15000);
			return "Success";
		}

	}

	private static class Callback implements MultiExecutionCallback {

		private final CountDownLatch latch;

		private Map<Member, Object> values;

		public Callback(CountDownLatch latch) {
			this.latch = latch;
		}

		@Override
		public void onResponse(Member member, Object value) {
			logger.info("Member " + member + " done: " + value);
		}

		@Override
		public void onComplete(Map<Member, Object> values) {
			this.values = values;
			logger.info("All members done: " + values.values());
			latch.countDown();
		}

		public void await() throws InterruptedException {
			latch.await();
		}

	}

}
```

Best,
Lukas

eminn:
Added missing uuid to MemberCallableTaskOperation on ExecutorServiceProxy

Fixes #5564 

Thanks @lukasblu for the bug report and a reproducer !

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningServiceTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningServiceTest.java
@@ -2,6 +2,9 @@ package com.hazelcast.spi.impl.operationservice.impl;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IExecutorService;
+import com.hazelcast.core.Member;
+import com.hazelcast.core.MultiExecutionCallback;
 import com.hazelcast.instance.GroupProperty;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.ObjectDataInput;
@@ -18,17 +21,20 @@ import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.TestHazelcastInstanceFactory;
 import com.hazelcast.test.annotation.ParallelTest;
 import com.hazelcast.test.annotation.QuickTest;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
-import java.io.IOException;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
-
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 @RunWith(HazelcastParallelClassRunner.class)
 @Category({QuickTest.class, ParallelTest.class})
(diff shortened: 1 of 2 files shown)


eminn:
run-lab-run

gurbuzali:
:+1: 
back-port?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningServiceTest.java
```diff
@@ -231,6 +237,63 @@ public void run()
...
+        HazelcastInstance instance2 = factory.newHazelcastInstance(config);
+
+        CountDownLatch latch = new CountDownLatch(1);
+        Callback callback = new Callback(latch);
+
+        IExecutorService executorService = instance1.getExecutorService(randomName());
+        executorService.submitToAllMembers(new SleepingTask(), callback);
+        assertOpenEventually(latch);
+        for (Object result : callback.values.values()) {
+            if (!result.equals("Success")) {
```
Donnerbart: Maybe this is easier to read:
```
assertNotEquals("Expected success result, but was: " + result, "Success", result);
```

eminn:
verify

eminn:
Backport : #6719

hasancelik:
run-lab-run

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningServiceTest.java
```diff
@@ -231,6 +237,63 @@ public void run()
         }, 2);
     }
 
+
+    @Test
+    public void testExecutionShouldNotTimeoutIfTraceable_WithMultiCallback() throws Exception {
```
ahmetmircik: is this a `quick` or  a `slow` test?

Donnerbart: `@Category({QuickTest.class, ParallelTest.class})`

ahmetmircik: Thats right @Donnerbart,  but this particular test seems not `quick` at first glance. There is a very long sleep `Thread.sleep(15000)`.

Donnerbart: Ah, now I get want you meant, sorry for the confusion. The test indeed seems to sleep for 15 seconds to trigger the operation timeout of 3 seconds (the callable result is checked, so the full time will elapse).

That one might indeed be a good candidate for a slow test.

eminn: I will merge it as it is and add `SlowTest` annotation later.

PR #6732, opened 2015-11-13
Transactional map getForUpdate() method now loads key from maploader

### LINKED ISSUE #6104: MapLoader does not get invoked for TransactionalMap.getForUpdate()
See http://stackoverflow.com/questions/32311284/hazelcast-maploader-does-not-get-invoked-for-transactionalmap-getforupdate for details

Credits for reporting the bug goes to http://stackoverflow.com/users/4185938/andres

eminn:
Fixes #6104

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
@@ -171,8 +171,8 @@ public class DefaultMapOperationProvider implements MapOperationProvider {
 
     @Override
     public MapOperation createTxnLockAndGetOperation(String name, Data dataKey, long timeout, long ttl, String
-            ownerUuid) {
-        return new TxnLockAndGetOperation(name, dataKey, timeout, ttl, ownerUuid);
+            ownerUuid, boolean shouldLoad) {
+        return new TxnLockAndGetOperation(name, dataKey, timeout, ttl, ownerUuid, shouldLoad);
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
@@ -121,7 +121,7 @@ public abstract class TransactionalMapProxySupport extends AbstractDistributedOb
     }
 
     public Object getForUpdateInternal(Data key) {
-        VersionedValue versionedValue = lockAndGet(key, tx.getTimeoutMillis());
+        VersionedValue versionedValue = lockAndGet(key, tx.getTimeoutMillis(), true);
         addUnlockTransactionRecord(key, versionedValue.version);
         return versionedValue.value;
     }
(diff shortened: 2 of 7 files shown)


gurbuzali:
:+1: 
I guess there is no need  to change anything for client-side

eminn:
verify

eminn:
run-lab-run

eminn:
run-lab-run

eminn:
run-lab-run

hasancelik:
run-lab-run

eminn:
verify

eminn:
verify

eminn:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
```diff
@@ -171,8 +171,8 @@ public MapOperation createTxnDeleteOperation(String name, Data dataKey, long ver
 
     @Override
     public MapOperation createTxnLockAndGetOperation(String name, Data dataKey, long timeout, long ttl, String
-            ownerUuid) {
-        return new TxnLockAndGetOperation(name, dataKey, timeout, ttl, ownerUuid);
+            ownerUuid, boolean loadKey) {
```
pveentjer: Afaik the whole entry is loaded; key & value. Perhaps it makes more sense to call it loadFromStorage?

pveentjer: A better name would be 'read-through'.

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java
```diff
@@ -277,6 +278,11 @@ private VersionedValue lockAndGet(Data key, long timeout) {
         } catch (Throwable t) {
             throw ExceptionUtil.rethrow(t);
         }
+
+    }
+
+    private VersionedValue lockAndGet(Data key, long timeout) {
```
pveentjer: Add a line of javadoc that this method doesn't do a read through.

hazelcast/src/test/java/com/hazelcast/map/MapTransactionTest.java
```diff
@@ -1329,4 +1327,61 @@ public Object execute(TransactionalTaskContext context) throws TransactionExcept
             }
         });
     }
+
+
+    @Test
+    public void testGetForUpdateLoadsKeyFromMapLoader() {
```
pveentjer: Personally I would go over the different permutations.

1: get for updated when locked
2: get for update when item doesn't exist in db.
3: get for update when item does exist in db.

Perhaps there are more combination.

Also use the following naming:

getForUpdate_whenLockedByOtherThread

or with a test prefix.

Going over all permutations helps to discover current + future bugs.

hazelcast/src/test/java/com/hazelcast/map/MapTransactionTest.java
```diff
@@ -1329,4 +1327,61 @@ public Object execute(TransactionalTaskContext context) throws TransactionExcept
...
+        final String mapName = randomMapName();
+        final CountDownLatch latch = new CountDownLatch(1);
+        Config config = new Config();
+        MapStoreConfig storeConfig = new MapStoreConfig();
+        storeConfig.setEnabled(true).setImplementation(new MockMapLoader(latch));
```
pveentjer: You don't need to create a MockMapLoader; just use Mockito.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
pveentjer: Forget about the read-through. It would suggest that even if the data would be in memory, the data is still loaded from the db. But we do need to find a more self explaining name than 'loadKey'.

hazelcast/src/test/java/com/hazelcast/map/MapTransactionTest.java
```diff
@@ -1329,4 +1332,97 @@ public Object execute(TransactionalTaskContext context) throws TransactionExcept
             }
         });
     }
+
+
+    @Test
+    public void testGetForUpdate_LoadsKeyFromMapLoader_whenKeyDoesNotExistsInDb() {
+        final String mapName = randomMapName();
+        final MapStoreAdapter mock = mock(MapStoreAdapter.class);
+        when(mock.load(anyObject())).thenReturn(null);
```
pveentjer: Afaik this is default behavior.

pveentjer:
Minor comment. 

:+1:

Issue #6486, opened 2015-10-18
[TEST-FAILURE] ClientMapNearCacheTest.testServerMapExpiration_doesNotInvalidateClientNearCache


hasancelik:
```
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.map.ClientMapNearCacheTest$6.run(ClientMapNearCacheTest.java:541)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:828)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:842)
	at com.hazelcast.client.map.ClientMapNearCacheTest.testServerMapExpiration_doesNotInvalidateClientNearCache(ClientMapNearCacheTest.java:536)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x/com.hazelcast$hazelcast-client-new/3903/testReport/junit/com.hazelcast.client.map/ClientMapNearCacheTest/testServerMapExpiration_doesNotInvalidateClientNearCache/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK7/com.hazelcast$hazelcast-client/692/testReport/junit/com.hazelcast.client.map/ClientMapNearCacheTest/testServerMapExpiration_doesNotInvalidateClientNearCache/

ahmetmircik:
Fix attempt via https://github.com/hazelcast/hazelcast/commit/11be05660a096922893c51683a25e11fba923193

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast-client-legacy/704/testReport/junit/com.hazelcast.client.map/ClientMapNearCacheTest/testServerMapExpiration_doesNotInvalidateClientNearCache/

hasancelik:
[...]

PR #6733, opened 2015-11-13
Race-free test of client near cache via listener

Donnerbart:
* Added method to register a near cache invalidation listener to the `NearCachedClientMapProxy`.
* Used the listener in client near cache tests to get rid of race conditions.

I don't know if it's acceptable to have a public method in a specialized proxy class. If so this might be a way to write a race-free test for the client near cache.

@ahmetmircik: Could you please also have a look if this is okay?

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
@@ -343,9 +343,13 @@ public class NearCachedClientMapProxy<K, V> extends ClientMapProxy<K, V> {
     }
 
     protected void addNearCacheInvalidateListener() {
+        EventHandler handler = new InvalidationListener(this.nearCache);
+        addNearCacheInvalidateListener(handler);
+    }
+
+    public void addNearCacheInvalidateListener(EventHandler handler) {
         try {
             MapAddNearCacheEntryListenerRequest addRequest = new MapAddNearCacheEntryListenerRequest(name, false);
-            EventHandler handler = new InvalidationListener(this.nearCache);
             BaseClientRemoveListenerRequest removeRequest = new MapRemoveEntryListenerRequest(name);
 
             invalidationListenerId = registerListener(addRequest, removeRequest, handler);
@@ -354,7 +358,6 @@ public class NearCachedClientMapProxy<K, V> extends ClientMapProxy<K, V> {
                     "-----------------\n Near Cache is not initialized!!! \n-----------------", e);
         }
     }
-
 
     protected static final class InvalidationListener implements EventHandler<PortableEntryEvent> {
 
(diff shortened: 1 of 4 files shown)


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

pveentjer:
@Donnerbart is this a bug fix?

About public method ---> no problem. Nobody will be using the actual proxy implementation.

sancar:
@Donnerbart 
I think this will fix fixes https://github.com/hazelcast/hazelcast/issues/6486 , right ?

sancar:
@Donnerbart 
From what I understand from code and comments problem that is trying to be solved this pr is to do the test after we get invalidation event from first put. 
But it looks like this problem is already trying to solved by creating client after put. This did not solve the problem, because we can still get the invalidation event, even in new client. New client opens gets map, add invalidation listener, and then server send the updated event. Even Though it is very unlikely, since there is no ordering between client authentication and puts, it is possible.

As a solution, you explicitly wait for first event to come to continue the testing.
If my understanding is correct so far I give :+1:

sancar:
Not concern of this pull request, but I did not understand why we don't want client nearcache to not get invalidated when key is expired. In config, we set invalidateOnChange as true. Does expiration not counts as change ? Why ? 

From the javadoc, change means "updated or removed"
True to evict the cached entries if the entries are changed (updated or removed).

Having the value in client near cache, even though it is not available on server, sounds wrong to me. 

Then I found this commit explaining the answer. :)
d931c2b767bba27ab2ea547d09c8a768b2907bdf
Near-cache has its own eviction/expiration mechanism, eviction/expiration on IMap should not force any near-cache eviction/expiration

Leaving the my question and found answer for future reference.

Donnerbart:
> From what I understand from code and comments, the problem that is trying to be solved this pr is to do the test after we get invalidation event from first put.
Yes, the put creates an invalidation event. After that we get the value again on the client side, so it should be available in the near cache. Then the entry should expire via TTL, which should not affect the client near cache.

> But it looks like this problem is already trying to solved by creating client after put. This did not solve the problem, because we can still get the invalidation event, even in new client. New client opens gets map, add invalidation listener, and then server send the updated event. Even Though it is very unlikely, since there is no ordering between client authentication and puts, it is possible.
Almost. The idea of creating a new client is from @ahmetmircik. And I have to admit, I didn't understand it in the first way, because I didn't know of the invalidation event. But that is solved with creating a new client, there seems to be no race there.

The actual race is between the TTL lifetime and the creation of the new client. We had 3 and 6 seconds, but sometimes this is not enough time to create a new client instance on CloudBees. We can use a very high value here, but this doesn't solve the race condition, just hides it.

> As a solution, you explicitly wait for first event to come to continue the testing.
Yes. My idea was to create a register method for an invalidation listener, so we can use the already existing client instance and wait exactly until the event has arrived and then continue with the test.

So if you're fine with the additional code in the proxy class and the cast in the test code, I think this should be a good solution to make this test race free.

Donnerbart:
Since Peter and Sancar are fine with the proxy change I'll merge it. If we have other issues with that test I'll revert it to the solution of Ahmet, which seemed to be the most stable one we had.

Issue #6566, opened 2015-10-26
Query inconsistency due to race condition


pveentjer:
The behavior of retrieving data, deserializing it and setting it on the record is racy since an update could have happened in the same time. This can lead to permanently inconsistent query results.

jerrinot:
fixed by #6737

PR #6737, opened 2015-11-13
Fix the race when optimizeQuery = true

jerrinot:
A version without a lock on a hot path.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecord.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecord.java
@@ -18,10 +18,15 @@ package com.hazelcast.map.impl.record;
 
 import com.hazelcast.nio.serialization.Data;
 
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
 /**
  * CachedDataRecord.
  */
 class CachedDataRecord extends DataRecord {
+    private static final AtomicReferenceFieldUpdater<CachedDataRecord, Object> CACHED_VALUE =
+            AtomicReferenceFieldUpdater.newUpdater(CachedDataRecord.class, Object.class, "cachedValue");
+
 
     private transient volatile Object cachedValue;
 
@@ -34,18 +39,18 @@ class CachedDataRecord extends DataRecord {
 
     @Override
     public void setValue(Data o) {
-        cachedValue = null;
         super.setValue(o);
+        cachedValue = null;
     }
 
     @Override
-    public Object getCachedValue() {
+    public Object getCachedValueUnsafe() {
         return cachedValue;
     }
 
     @Override
-    public void setCachedValue(Object cachedValue) {
-        this.cachedValue = cachedValue;
+    public boolean casCachedValue(Object expectedValue, Object newValue) {
+        return CACHED_VALUE.compareAndSet(this, expectedValue, newValue);
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecordWithStats.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecordWithStats.java
@@ -18,10 +18,14 @@ package com.hazelcast.map.impl.record;
 
 import com.hazelcast.nio.serialization.Data;
 
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
 /**
  * CachedDataRecordWithStats.
  */
 class CachedDataRecordWithStats extends DataRecordWithStats {
+    private static final AtomicReferenceFieldUpdater<CachedDataRecordWithStats, Object> CACHED_VALUE =
+            AtomicReferenceFieldUpdater.newUpdater(CachedDataRecordWithStats.class, Object.class, "cachedValue");
 
     private transient volatile Object cachedValue;
 
@@ -35,18 +39,18 @@ class CachedDataRecordWithStats extends DataRecordWithStats {
 
     @Override
     public void setValue(Data o) {
-        cachedValue = null;
         super.setValue(o);
+        cachedValue = null;
     }
 
     @Override
-    public Object getCachedValue() {
+    public Object getCachedValueUnsafe() {
         return cachedValue;
     }
 
     @Override
-    public void setCachedValue(Object cachedValue) {
-        this.cachedValue = cachedValue;
+    public boolean casCachedValue(Object expectedValue, Object newValue) {
+        return CACHED_VALUE.compareAndSet(this, expectedValue, newValue);
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/record/Record.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/record/Record.java
@@ -56,9 +56,26 @@ public interface Record<V> {
 
     long getEvictionCriteriaNumber();
 
-    Object getCachedValue();
+    /**
+     * Get current cache value or null.
+     *
+     * Warning: Do not use this method directly as it might expose arbitrary objects acting as a lock.
+     * Use {@link Records#getCachedValue(Record)} instead.
+     *
+     * @return current cached value or null or cached record mutex.
+     */
+    Object getCachedValueUnsafe();
 
-    void setCachedValue(Object cachedValue);
+    /**
+     * Atomically sets the cached value to the given new value
+     * if the current cached value {@code ==} the expected value.
+     *
+     * @param expectedValue the expected cached value
+     * @param newValue the new cached value
+     * @return {@code true} if successful. False return indicates that
+     * the actual cached value was not equal to the expected cached value.
+     */
+    boolean casCachedValue(Object expectedValue, Object newValue);
 
     long getTtl();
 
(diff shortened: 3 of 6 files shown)


jerrinot:
run-lab-run

hasancelik:
run-lab-run

hasancelik:
run-lab-run

hasancelik:
run-lab-run

jerrinot:
verify

mtopolnik:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
```diff
@@ -50,25 +50,97 @@ public static RecordInfo buildRecordInfo(Record record) {
...
-        return value;
     }
 
-    private static boolean shouldCache(Record record, Object value) {
-        boolean isCachableRecordType = record instanceof CachedDataRecordWithStats || record instanceof CachedDataRecord;
-        if (!isCachableRecordType) {
-            return false;
+    private static Object tryToLoadIntoCache(Record record, SerializationService serializationService) {
+        Object valueBeforeCas = record.getValue();
+        if (!shouldCache(valueBeforeCas)) {
```
mtopolnik: There is a data race on the reference loaded into `valueBeforeCas`.

jerrinot: do you think it's alright to merge this and fix the race in a separate issue? 
as the data race is not specific to caching data records.

mtopolnik:
If that is the case, yes. :+1:

jerrinot:
see https://github.com/hazelcast/hazelcast/pull/6774

jerrinot:
Fixes #6566

jerrinot:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecord.java
```diff
@@ -18,10 +18,15 @@
 
 import com.hazelcast.nio.serialization.Data;
 
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
 /**
  * CachedDataRecord.
  */
 class CachedDataRecord extends DataRecord {
+    private static final AtomicReferenceFieldUpdater<CachedDataRecord, Object> CACHED_VALUE_UPDATER =
```
pveentjer: Just a flavor thing; I always removed the _UPDATER section. It is already clear from the CACHED_VALUE you are talking to a constant. And I find this a bit easier to read:

CACHED_VALUE.compareAndSet(this, ....)

But this is flavor.

hazelcast/src/main/java/com/hazelcast/map/impl/record/Record.java
```diff
@@ -58,7 +58,7 @@
 
     Object getCachedValue();
 
-    void setCachedValue(Object cachedValue);
+    boolean casCachedValue(Object expectedValue, Object newValue);
```
pveentjer: Perhaps a line of comment. True if success, false if failure.

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
```diff
@@ -50,25 +50,97 @@ public static RecordInfo buildRecordInfo(Record record) {
...
+                //the cachedValue is either locked by another thread or it contains a wrapped thread
+                cachedValue = ThreadWrapper.unwrapOrNull(cachedValue);
+                if (cachedValue != null) {
+                    //exceptional case: the cachedValue is not locked, it just contains an instance of Thread.
+                    //this can happen when user put an instance of Thread into a map
+                    //(=it should never happen, but never say never...)
+                    return cachedValue;
+                }
+                //it looks like some other thread actually locked the cachedValue
+                Thread.yield();
```
pveentjer: Why is yielding done only in case of Thread? I don't understand that in this case you want to be nice; but in other case you just spin.

hazelcast/src/test/java/com/hazelcast/map/impl/record/RecordsTest.java
```diff
@@ -49,32 +46,36 @@ public void setUp() {
...
+        Object firstDeserilizedValue = Records.getValueOrCachedValue(record, serializationService);
+        assertEquals(objectPayload, firstDeserilizedValue);
 
-        Records.getValueOrCachedValue(record, serializationService);
-        verify(record, times(1)).setCachedValue(objectPayload);
+        //we don't need serialization service for the 2nd call
+        Object secondDeserilizedValue = Records.getValueOrCachedValue(record, null);
+        assertSame(firstDeserilizedValue, secondDeserilizedValue);
     }
+
```
pveentjer: Do we have test for the thread as value?

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
```diff
@@ -50,25 +50,97 @@ public static RecordInfo buildRecordInfo(Record record) {
...
 
+
+    /**
+     * currentThread inside cachedValue acts as "deserialization in-progress" marker
+     * if the actual deserialized value is instance of Thread then we need to wrap it
+     * otherwise it might be mistaken for the "deserialization in-progress" marker.
+     *
+     */
+    private static final class ThreadWrapper extends Thread {
+        private final Thread delegate;
```
pveentjer: why is this called delegate? Nothing is being delegated to this thread.

hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecordWithStats.java
```diff
@@ -18,10 +18,14 @@
...
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
 /**
  * CachedDataRecordWithStats.
  */
 class CachedDataRecordWithStats extends DataRecordWithStats {
+    private static final AtomicReferenceFieldUpdater<CachedDataRecordWithStats, Object> CACHED_VALUE_UPDATER =
+            AtomicReferenceFieldUpdater.newUpdater(CachedDataRecordWithStats.class, Object.class, "cachedValue");
 
     private transient volatile Object cachedValue;
```
pveentjer: It would be a good thing to add documentation here explaining that the cachedValue not only contains the value or null, but could also have a thread/thread-wrapper to indicate it is currently being locked.

If somebody would just read out the cachedValue, then bogus values could appear out of thin air.

hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecord.java
mtopolnik: I bet you actually mean "taste" :) In any case, :100: to your taste :)

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
mtopolnik: While we're at it, it would be even better to use the complete `BackoffIdleStrategy` to have incremental backpressure. This is currently trapped on the EE side, but it would be easy to transfer.

pveentjer: Ahh  I remember again. Thread == locked.

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
jerrinot: fixed by #6774

hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecordWithStats.java
jerrinot: that's a good one, thanks!

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
jerrinot: good point. I'm moving the yielding to the end of the loop.

hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecord.java
jerrinot: done.

hazelcast/src/main/java/com/hazelcast/map/impl/record/Record.java
jerrinot: done

hazelcast/src/test/java/com/hazelcast/map/impl/record/RecordsTest.java
jerrinot: added. that was actually a bug, so thanks for spotting this!

hazelcast/src/main/java/com/hazelcast/map/impl/record/Records.java
jerrinot: changed to `storedThread`. It's still not a great name, but it has very small scope.

hasancelik:
run-lab-run

jerrinot:
verify

pveentjer:
:+1: good work from.

PR #6738, opened 2015-11-13
Fix for accessing expired cache entry

serkan-ozal:
If `processExpiredEntry` expires and removes record, `isExpiredAt` should not be called on that record since it may be native memory based record and it is disposed.

Without this PR, `CacheMemoryLeakStressTest::testNativeMemoryLeakWithExpiryPolicy` test causes JVM crash.


serkan-ozal:
verify

serkan-ozal:
verify

serkan-ozal:
verify

mdogan:
:+1: 

fortunately we have `CacheMemoryLeakStressTest::testNativeMemoryLeakWithExpiryPolicy` test...

mdogan:
I'm merging this one because it fixes a critical JVM crash issue.

PR #6740, opened 2015-11-14
Member list order fixed for clients

sancar:
Member list order was broken because we used HashSet in protocol
related code generation part. ArrayList will be used in all
collections in code generation.
A test is added to verify memberlist of server and client is in
same order.

### DIFF
--- a/hazelcast-client-legacy/src/test/java/com/hazelcast/client/cluster/ClientClusterServiceMemberListTest.java
+++ b/hazelcast-client-legacy/src/test/java/com/hazelcast/client/cluster/ClientClusterServiceMemberListTest.java
@@ -16,6 +16,7 @@ import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
 import java.util.Collection;
+import java.util.Set;
 
 import static com.hazelcast.client.impl.ClientTestUtil.getClientClusterService;
 import static com.hazelcast.cluster.memberselector.MemberSelectors.DATA_MEMBER_SELECTOR;
@@ -25,6 +26,7 @@ import static com.hazelcast.test.HazelcastTestSupport.getNode;
 import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
 import static junit.framework.Assert.assertEquals;
+import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertTrue;
 
 @RunWith(HazelcastParallelClassRunner.class)
@@ -85,6 +87,13 @@ public class ClientClusterServiceMemberListTest {
                 assertEquals(2, clusterService.getSize(DATA_MEMBER_SELECTOR));
             }
         });
+    }
+
+    @Test
+    public void testMemberListOrderConsistentWithServer() {
+        Set<Member> membersFromClient = client.getCluster().getMembers();
+        Set<Member> membersFromServer = dataInstance.getCluster().getMembers();
+        assertArrayEquals(membersFromClient.toArray(), membersFromServer.toArray());
     }
 
     private void verifyMembers(Collection<Member> membersToCheck, Collection<HazelcastInstance> membersToExpect) {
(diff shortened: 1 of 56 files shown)


sancar:
run-lab-run

sancar:
run-lab-run

sancar:
verify

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cluster/ClientClusterServiceMemberListTest.java
```diff
@@ -87,6 +89,13 @@ public void run()
         });
     }
 
+    @Test
+    public void testMemberListOrderConsistentWithServer() {
+        Set<Member> membersFromClient = client.getCluster().getMembers();
+        Set<Member> membersFromServer = dataInstance.getCluster().getMembers();
```
pveentjer: why do you cal the server instance a dataInstance? This is not very intuitive.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
```diff
@@ -260,8 +260,8 @@ public V get(K key, ExpiryPolicy expiryPolicy) {
         Data expiryPolicyData = toData(expiryPolicy);
         ClientMessage request = CacheGetAllCodec.encodeRequest(nameWithPrefix, keySet, expiryPolicyData);
         ClientMessage responseMessage = invoke(request);
-        Set<Map.Entry<Data, Data>> entrySet = CacheGetAllCodec.decodeResponse(responseMessage).entrySet;
-        for (Map.Entry<Data, Data> dataEntry : entrySet) {
+        List<Map.Entry<Data, Data>> entries = CacheGetAllCodec.decodeResponse(responseMessage).entries;
+        for (Map.Entry<Data, Data> dataEntry : entries) {
```
pveentjer: why do you call the collection entries, but a single item 'dataEntry'. So either uses entries/entry or dataEntries/dataEntry.

pveentjer:
This PR contains a lot more than just fixing the ordering of the member listing. Also a rewrite was done for a lot of api's where sets were used and now replaced by lists and collections. This goes beyond what should happen during bug fixing period. 

I would suggest splitting the PR in 2 parts:
1: the bug fix for member ordering
2: the code rewriting; which can be merged in 3.7.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cluster/ClientClusterServiceMemberListTest.java
sancar: I did not call it dataInstance. It was there before I wrote the test. I just used it.

sancar:
Code rewriting part happened since we decided to remove the Set from Code Generation in protocol. And this decision came with the realizing it from member list response. Since we used HashSet wherever a Set is used in protocol, we lost the order that we have on server side. I don't know, if changing the original order causing a bug in other parts that Set is used.(sometimes performance is an issue when using HashSet as you know). So since it is a straightforward change, I decided to do all at once.

asimarslan:
@pveentjer This PR contains client protocol related changes. Any required fix for the protocol cannot be delayed as such change directly broke the protocol compatibility. I know we do not make changes other than bug fixes. And I think this is a bug in protocol itself. Moreover postponing some part of this PR is not an option here due to compatibility concerns.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/cluster/ClientClusterServiceMemberListTest.java
pveentjer: Ok.

pveentjer:
The arguments make sense. I'm 50% yes and no on this issue.

asimarslan:
:+1:

pveentjer:
I would like to move forward with this issue; so either get it merged or moved to 3.7. Can you make a ping at the architecture group and ask them for opinion.

enesakar:
Peter, I also think arguments make sense. While both of us think in this direction no need to take architecture group's time. I approve the fixes.

PR #6741, opened 2015-11-14
Cleanup usage of "this." before local method calls

Donnerbart:
Due to the late state of the version we maybe should wait for 3.7 to merge this.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
@@ -298,7 +298,7 @@ public class ClientCacheProxy<K, V>
         for (K key : keys) {
             CacheEntryProcessorResult<T> ceResult;
             try {
-                final T result = this.invoke(key, entryProcessor, arguments);
+                final T result = invoke(key, entryProcessor, arguments);
                 ceResult = result != null ? new CacheEntryProcessorResult<T>(result) : null;
             } catch (Exception e) {
                 ceResult = new CacheEntryProcessorResult<T>(e);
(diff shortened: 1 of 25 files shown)


hazelcast-ra/hazelcast-jca/src/main/java/com/hazelcast/jca/ManagedConnectionImpl.java
```diff
@@ -55,16 +55,16 @@
     private HazelcastTransactionImpl localTransaction;
 
     public ManagedConnectionImpl(ConnectionRequestInfo cxRequestInfo, ManagedConnectionFactoryImpl factory) {
-        this.setLogWriter(factory.getLogWriter());
-        log(Level.FINEST, "ManagedConnectionImpl");
+        this.id = ID_GEN.incrementAndGet();
```
Donnerbart: This file contains the biggest changes, since I re-ordered the field assignments according to the field order in the class and moved the logging part to the end of the constructor.

Donnerbart:
run-lab-run

pveentjer:
Personally I'm in favor of this.

However; do we have consensus on it? And do we either have a checkstyle or wiki code style manual?

Donnerbart:
* If you have a look how many local method calls we have in Hazelcast vs. the changes in this PR I assume we have a general consensus of not using `this.` for local methods. A lot of occurrences are duplicates due to the legacy/new client. So the absolute usage of this style is very, very low.
* CheckStyle doesn't support this out of the box (just the opposite via `RequireThis`).
* I don't think we have an actual code page for this topic at the moment. But this special case will most probably fall under the general "consistency", since of the majority of code not using `this.` already. I really don't understand the big discussion about this specific topic, since it's so clear that the actual code base doesn't support the `this.` usage at all.
* Most occurrences seem to be in constructors, where it's not unusual to use `this.` to assign fields. I think this is more of an (accidental) usage to have a conform looking vertical code. I resolved this by adding a newline between the different sections in a constructor:
```
public ClassName() {
  super();

  this.field1  = field1;
  this.field2 = field2;

  localMethodCall();
}
```

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

hasancelik:
run-lab-run

Donnerbart:
run-lab-run

pveentjer:
:+1:

jerrinot:
:+1:

PR #6744, opened 2015-11-16
removed not needed recursive call to the destroy on replicatedmap

### LINKED ISSUE #6754: replicatedmap.destroy() goes into an infinite loop
`destroyDistributedObject()` method on `ReplicatedMapService` iterates over partition containers and record stores and destroys them. While destroying, record store calls `destroyDistributedObject()` which leads to an infinite loop.

eminn:
Fixes #6754

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/replicatedmap/ReplicatedMapTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/replicatedmap/ReplicatedMapTest.java
@@ -18,6 +18,7 @@ package com.hazelcast.replicatedmap;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.config.InMemoryFormat;
+import com.hazelcast.core.DistributedObject;
 import com.hazelcast.core.EntryAdapter;
 import com.hazelcast.core.EntryEvent;
 import com.hazelcast.core.EntryEventType;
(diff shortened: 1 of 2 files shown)


eminn:
verify

eminn:
run-lab-run

hazelcast/src/test/java/com/hazelcast/replicatedmap/ReplicatedMapTest.java
```diff
@@ -1030,6 +1031,18 @@ public void testSizeAfterRemove() throws Exception {
...
     }
 
+
+    @Test
+    public void testDestroy() throws Exception {
+        HazelcastInstance instance = createHazelcastInstance();
+        ReplicatedMap<Object, Object> replicatedMap = instance.getReplicatedMap(randomName());
+        replicatedMap.put(1, 1);
+        replicatedMap.destroy();
+        Collection<DistributedObject> objects = instance.getDistributedObjects();
```
pveentjer: Add an empty line between 'doing' and 'asserting'. This style is recommended on the wiki to make it easier to see what a test is doing.

pveentjer:
I don't see any reference to a bug. If this is a cleanup, it needs to be parked till 3.7.

hazelcast/src/test/java/com/hazelcast/replicatedmap/ReplicatedMapTest.java
eminn: Will make a direct commit after merge

pveentjer:
After a discussion with Emin.

If there are P partitions, then the total number of destroy calls is going to be P+P*P. 

This can become a performance problem if P is large enough. So we can consider this a performance bug and it is fine to get it merged.

PR #6747, opened 2015-11-16
Adds Live API Tests for jclouds

kobalski:
- Introduced LiveTest category to distinguish jclouds integration test from out normal test.
- Test marked with LiveTest annotation runs in our `code-coverage` profile
- To run tests run this command from maven ` mvn clean install -Pjclouds-test -Dtest.aws-ec2.identity="<aws-key>" -Dtest.aws-ec2.credential="<aws-credential>"` valid credentials must be provided as System property.

What we do is basically create five cloud instances with different zones,tag configurations via jclouds apis and tests are run over `JCloudsDiscoveryStrategy`. We are only verifying our jclouds integration over live apis.

Also, could not run google-compute-engine live tests, because I could not create more than one cloud instances with java api of jclouds I get a time-out exception.(https://www.mail-archive.com/user@jclouds.apache.org/msg01300.html). Right now, we only have live tests for AWS.

### DIFF
--- a/hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/AWSEC2LiveTest.java
+++ b/hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/AWSEC2LiveTest.java
@@ -1,0 +1,55 @@
+package com.hazelcast.jclouds.integration;
+
+import com.hazelcast.test.HazelcastSerialClassRunner;
+import com.hazelcast.test.annotation.JCloudsTest;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Category(JCloudsTest.class)
+@RunWith(HazelcastSerialClassRunner.class)
+public class AWSEC2LiveTest extends AbstractLiveTest {
+
+    public static final String AWS_PROVIDER   = "aws-ec2";
+    public static final String AWS_IDENTITY   =  System.getProperty("test.aws-ec2.identity");
+    public static final String AWS_CREDENTIAL =  System.getProperty("test.aws-ec2.credential");
+
+    public static LiveComputeServiceUtil builder =
+            new LiveComputeServiceUtil(AWS_PROVIDER, AWS_IDENTITY,
+                    AWS_CREDENTIAL);
+
+
+    @BeforeClass
+    public static void provisionNodes() throws Exception {
+        builder.provisionNodes();
+    }
+
+    @AfterClass
+    public static void destroyNodes() throws Exception {
+        builder.destroyNodes();
+    }
+
+
+    @Override
+    protected Map<String, Comparable> getProperties() {
+        Map<String, Comparable> properties = new HashMap<String, Comparable>();
+        properties.put("provider", AWS_PROVIDER);
+        properties.put("identity", AWS_IDENTITY);
+        properties.put("credential", AWS_CREDENTIAL);
+        return properties;
+    }
+
+    @Override
+    protected String getRegion1() {
+        return LiveComputeServiceUtil.AWS_REGION1;
+    }
+
+    @Override
+    protected String getRegion2() {
+        return LiveComputeServiceUtil.AWS_REGION2;
+    }
+}
--- a/hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/AbstractLiveTest.java
+++ b/hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/AbstractLiveTest.java
@@ -1,0 +1,83 @@
+package com.hazelcast.jclouds.integration;
+
+
+import com.google.common.collect.Iterators;
+import com.hazelcast.jclouds.JCloudsDiscoveryStrategy;
+import org.junit.Test;
+
+import java.util.Map;
+
+import static com.hazelcast.jclouds.integration.LiveComputeServiceUtil.GROUP_NAME1;
+import static com.hazelcast.jclouds.integration.LiveComputeServiceUtil.GROUP_NAME2;
+import static com.hazelcast.jclouds.integration.LiveComputeServiceUtil.tag1;
+import static com.hazelcast.jclouds.integration.LiveComputeServiceUtil.tag2;
+import static com.hazelcast.jclouds.integration.LiveComputeServiceUtil.tag3;
+import static org.junit.Assert.assertEquals;
+
+public abstract class AbstractLiveTest {
+
+    protected abstract Map<String,Comparable> getProperties();
+    protected abstract String getRegion1();
+    protected abstract String getRegion2();
+
+    @Test
+    public void test_DiscoveryStrategyFiltersNodesByGroup() throws Exception {
+        Map<String, Comparable> properties1 = getProperties();
+        properties1.put("group", GROUP_NAME1);
+        JCloudsDiscoveryStrategy strategy1 = new JCloudsDiscoveryStrategy(properties1);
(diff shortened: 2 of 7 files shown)


bilalyasar:
:+1: very cool testing feature,
there some `system out println` codes, other than this looks good.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/AbstractLiveTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static com.hazelcast.jclouds.integration.LiveComputeServiceBuilder.tag1;
+import static com.hazelcast.jclouds.integration.LiveComputeServiceBuilder.tag2;
+import static com.hazelcast.jclouds.integration.LiveComputeServiceBuilder.tag3;
+import static org.junit.Assert.assertEquals;
+
+public abstract class AbstractLiveTest {
+
+    public static final String AWS_PROVIDER   = "aws-ec2";
+    public static final String AWS_IDENTITY   =  System.getProperty("test.aws-ec2.identity");
+    public static final String AWS_CREDENTIAL =  System.getProperty("test.aws-ec2.credential");
```
mesutcelik: I think those static fields should go into `AWSEC2LiveTest`

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/LiveComputeServiceBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+            new AbstractMap.SimpleImmutableEntry<String, String>("Type", "Micro");
+
+    public LiveComputeServiceBuilder(String provider) {
+        this.provider = provider;
+        init();
+    }
+
+    public void init() {
+        if (computeService == null) {
+            computeService = initComputeService(AWS_PROVIDER, AWS_IDENTITY, AWS_CREDENTIAL);
```
mesutcelik: `LiveComputeServiceBuilder` feels more Test Util class so using Util in the name might be more readable. Additionally, accepting `initComputeService` parameters from outside will make this class independent of CloudProvider...

hazelcast/src/test/java/com/hazelcast/test/annotation/LiveTest.java
```diff
@@ -0,0 +1,10 @@
+package com.hazelcast.test.annotation;
+
+
+/**
+ * Annotates tests which requires integration tests over a live api.
+ * <p/>
+ * Will be executed in nightly builds and for jclouds live tests
+ */
+public final class LiveTest {
```
mesutcelik: I don't know if LiveTest is a good name and including it in sonar is a good option. That makes Sonra Build fragile in every Cloud API down cases. On the other hand, not having this in Sonar Build makes code-coverage report unhappy. There is a trade-off here.

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/AWSEC2LiveTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+
+    @BeforeClass
+    public static void provisionNodes() throws Exception {
+        builder.provisionNodes();
+    }
+
+    @AfterClass
+    public static void destroyNodes() throws Exception {
+        builder.destroyNodes();
```
mesutcelik: What happens if we can't destroy Cloud Nodes? Are they gonna run indefinitely if we don't stop them manually? 
We might need an alert mechanism but i don't know if such thing is available...

Donnerbart: The first thing we start on a node in Simulator is a HarakiriMonitor process, which will shutdown the node after a specific time.

hazelcast/src/test/java/com/hazelcast/test/annotation/LiveTest.java
Donnerbart: With Cloud API you mean AWS, right? We had this kind of issue quite often with Simulator nightly builds with our Provisioner tool which uses jclouds. Danny switched the test setup to use AWS command line tools which is more reliable. So I cannot really recommend to use this in an automated test.

For Simulator I went down the hard road and mocked the `ComputeService` to test the Provisioner tool without any need for AWS to be working. This even works completely offline. The trade-off here is heavy usage of mocking. So whenever there is a significant change in the jclouds API we have to adjust the tests as well.

kobalski: @Donnerbart Yes, you are right using jclouds own `ComputeService` has problems when provisioning multiple machines, I could not even start more than two machines on google-compute engine, the tests just hangs and gives timeout exception. But on AWS, I did not encounter reliability issues , but I am pretty sure it will create problems during running automated tests. Actually what I want is provide more live tests for different cloud vendors easily so that we can see problems due to API changes ahead of time with the help of live tests, so depending on AWS CLI tool might be problem, what I want is a generic provisioning tool for different cloud vendors which is jclouds. But it seems, this makes our code coverage built dependent of other APIs. Maybe we might use your approach, we also used heavy mocking for unit tests without any dependency to AWS and maybe we can put all live tests in separate repo and run them on different built once a week. We need to have same kind of verification so that in each release we are sure that our code is working over Live API.

Donnerbart: Sure, Simulator also uses jclouds in the Provisioner tool to support all kind of cloud services. But as I said, during the nightly tests we had issues with it (and sometimes even with the AWS CLI utils, I suspect peaks at AWS causing issues sometimes).

I think we can run those tests regularly, but maybe not nightly (the API will surely not change that often). For a nightly test the mocking approach maybe more suitable.

This is my abstract test class where I have an init method for the `ComputeService` mock:
https://github.com/hazelcast/hazelcast-simulator/blob/master/simulator/src/test/java/com/hazelcast/simulator/provisioner/AbstractComputeServiceTest.java

This is an example usage where I use the `initComputeServiceMock()` in the `@Before` method and some `verify()` checks in the `@Test` methods:
https://github.com/hazelcast/hazelcast-simulator/blob/master/simulator/src/test/java/com/hazelcast/simulator/provisioner/TemplateBuilderTest.java

hazelcast-jclouds/src/test/java/com/hazelcast/jclouds/integration/AWSEC2LiveTest.java
kobalski: When destroy operation nodes fails, we might have leftover nodes. Right now, I could not integrate Harakiri monitor because it seem to be a separate jvm process started by this script, https://github.com/hazelcast/hazelcast-simulator/blob/master/dist/src/main/dist/bin/harakiri-monitor
but I will find a way to start  same kind of monitoring mechanism for our tests too. but not in this PR.

hazelcast/src/test/java/com/hazelcast/test/annotation/LiveTest.java
kobalski: I made a separate profile for integration tests. Our coverage profile will no longer be dependent to jclouds. We might run integration test weekly for example we mostly cover our jclouds classes with unit tests and maybe run jclouds test weekly with  `mvn clean install -Pjclouds-test -Dtest.aws-ec2.identity="<aws-key>" -Dtest.aws-ec2.credential="<aws-credential>"` on seperate build on cloudbeees.

kobalski:
@bilalyasar Switched `printf` statements inside tests with logger statements.

kobalski:
run-lab-run

mesutcelik:
:+1:

kobalski:
verify

PR #6749, opened 2015-11-16
For JCache, put to near-cache as sync if put is sync, put as async if put is async

serkan-ozal:
If put to cache is sync no need to put it to near-cache as async but sync.
Async puts are still same by putting to near-cache as async

This PR also contains some updates on eviction policies:
* Compare creation times of records if access times are same for LRU based eviction
* Compare creation times of records if access hits are same for LFU based eviction


serkan-ozal:
verify

serkan-ozal:
verify

serkan-ozal:
run-lab-run

asimarslan:
:+1:

PR #6752, opened 2015-11-16
Fixed memory leak caused by insertion of null-valued records into ima…

ahmetmircik:
…p via MapLoader

closes https://github.com/hazelcast/hazelcast/issues/6830

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
@@ -824,6 +824,10 @@ public class DefaultRecordStore extends AbstractEvictableRecordStore {
 
     @Override
     public Object putFromLoad(Data key, Object value, long ttl) {
+        if (!isKeyAndValueLoadable(key, value)) {
+            return null;
+        }
+
         final long now = getNow();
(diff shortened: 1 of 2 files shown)


mdogan:
@ahmetmircik: I think it's not a common to get a null key from map-loader, I prefer to print a warning when at least `key` is null.

pveentjer:
@ahmetmircik can you clean up the title so we get a nice title in the listing of 3.6 issues resolved?

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
```diff
@@ -828,6 +828,10 @@ public Object putFromLoad(Data key, Object value) {
 
     @Override
     public Object putFromLoad(Data key, Object value, long ttl) {
+        if (key == null || value == null) {
```
pveentjer: In case of doubt it is best to fail fast; one of our architectural principles. So I would also go for an exception (that isn't gobbled).

ahmetmircik:
run-lab-run

ahmetmircik:
@mdogan done.

jerrinot:
:+1:

mdogan:
:+1:

PR #6753, opened 2015-11-16
Fixed memory leak caused by insertion of null-valued records into ima…

ahmetmircik:
…p via MapLoader


jerrinot:
it's a backport of #6752

jerrinot:
:+1: 
we do not want any behaviour changes in patch releases -> this patch is good to me.

mdogan:
:+1:

PR #6756, opened 2015-11-17
Check node's start completion before applying received partition table

mdogan:
In PartitionStateOperation checks node start before processing partition
table but FinalizeJoinOperation directly applies without doing any check.
This is causing failure when a node restarts while restoring hot-restart data.

_This problem is introduced by commit: a1851f7a8598745035f9b64eec43d93d4de5bd32_

Fixes hazelcast/hazelcast-enterprise#490

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
@@ -718,6 +718,11 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
         lock.lock();
         try {
             final Address sender = partitionState.getEndpoint();
+            if (!node.getNodeExtension().isStartCompleted()) {
+                logger.warning("Ignoring received partition table, startup is not completed yet. Sender: " + sender);
+                return;
+            }
+
             final Address master = node.getMasterAddress();
             if (node.isMaster()) {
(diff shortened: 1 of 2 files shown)


metanet:
:+1:  once green

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -718,6 +718,11 @@ public void processPartitionRuntimeState(PartitionRuntimeState partitionState) {
         lock.lock();
         try {
             final Address sender = partitionState.getEndpoint();
+            if (!node.getNodeExtension().isStartCompleted()) {
```
pveentjer: Just for my understanding; will this method be called multiple times? Because now we are existing because the start has not completed; but is a new attempt going to be made that eventually can succeed?

mdogan: It'll be called when master publishes partition table and that happens after every migration and also periodically (15 sec by default).

That means, it will eventually succeed. 

Normally, this doesn't happen without hot-restart. Because when a node joins to the cluster, it will be ready to receive the partition table.

mdogan:
@pveentjer: are you ok with the change?

gurbuzali:
:+1:

PR #6758, opened 2015-11-17
Extractor parameters

tombujok:


jerrinot:
```
15:22:53 Tests in error: 
15:22:53   IssuesTest.testIssue1067GlobalSerializer:207 » ClassCast com.hazelcast.map.Iss...
```

tombujok:
It passes locally. I guess it's an randomly failing test...

tombujok:
verify

hazelcast/src/main/java/com/hazelcast/query/extractor/Arguments.java
```diff
@@ -0,0 +1,49 @@
...
+ * that uses a ValueExtractor, like: customAttribute[argumentString]
+ * <p/>
+ * Let's have a look at the following query: 'currency[incoming] == EUR'
+ * Let's assume that currency is an custom attribute that uses com.test.CurrencyExtractor for extraction.
+ * The string "incoming" is an argument that will be passed to the {@link ArgumentsParser} during the extraction.
+ * The parser will parse the string according to the parser's custom logic and it will return an Arguments object.
+ * The Arguments object may return a single object, array, collection, etc. It's up to the ValueExtractor implementor's
+ * to understand the semantics of the arguments.
+ * <p/>
+ * By default, if there's no custom {@link ArgumentsParser} registered, the default parser follows the pass-through
```
eminn: Since we are lacking the ability of registering custom argument parsers, it is better to rephrase that. Later in the javadoc mentions the inability but there is no need to have such confusion.

hazelcast/src/main/java/com/hazelcast/query/impl/QueryableEntry.java
```diff
@@ -209,19 +188,19 @@ private AttributeType extractAttributeTypeIfAttributeQueryConstant(String attrib
...
     }
 
-    private AttributeType extractAttributeTypeFromSingleResult(Object extractedObject) {
-        if (extractedObject == null) {
+    private AttributeType extractAttributeTypeFromSingleResult(Object extractedSingleResult) {
+        if (extractedSingleResult == null) {
             return null;
         } else {
-            return ReflectionHelper.getAttributeType(extractedObject.getClass());
+            return ReflectionHelper.getAttributeType(extractedSingleResult.getClass());
```
eminn: I think we can remove the else keyword and curly braces.

hazelcast/src/main/java/com/hazelcast/query/extractor/Arguments.java
tombujok: Done. Thx!

hazelcast/src/main/java/com/hazelcast/query/impl/QueryableEntry.java
tombujok: Done.

eminn:
:+1:

hazelcast/src/main/java/com/hazelcast/query/extractor/Arguments.java
```diff
@@ -0,0 +1,49 @@
...
+ * For now, it's not possible to register a custom {@link ArgumentsParser}, thus a default parser is used.
+ * It follows a "pass-through" semantics, which means that the string located in the square-brackets is passed
+ * "as-is" to the Arguments class.
+ * <p/>
+ * <b>It is not allowed to use square brackets within the argument string.</b>
+ * <p/>
+ *
+ * @param <T> type of the argument object encompassed by this class
+ */
+public abstract class Arguments<T> {
```
jerrinot: I am still not sure about this class. What's the motivation to have it?

tombujok: It's the same motivation as the one for having ValueCollector.
In the future we can add `int getInt()`, `double getDouble()` to avoid allocations and boxing/unboxing for primitive types, etc.

jerrinot: then do we want users to have ability to provide a custom implementation?

tombujok: As discussed, the Arguments object has been removed since the same effect may be achieved without it.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/Extractors.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+        return null;
+    }
+
+    Getter getGetter(Object targetObject, String attributeName) {
+        Getter getter = getterCache.getGetter(targetObject.getClass(), attributeName);
+        if (getter == null) {
+            getter = instantiateGetter(targetObject, attributeName);
+        }
+        if (getter.isCacheable()) {
```
jerrinot: should this be in the `(getter == null)` branch only?

hazelcast/src/main/java/com/hazelcast/query/impl/getters/Extractors.java
```diff
@@ -0,0 +1,93 @@
...
+        this.argumentsParser = new DefaultArgumentParser();
+    }
+
+    public Object extract(Object targetObject, String attributeName) {
+        if (targetObject != null) {
+            Getter getter = getGetter(targetObject, attributeName);
+            try {
+                return getter.getValue(targetObject);
+            } catch (Exception ex) {
+                ExceptionUtil.sneakyThrow(ex);
```
jerrinot: ಠ_ಠ

hazelcast/src/main/java/com/hazelcast/query/impl/getters/Extractors.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public Extractors(List<MapAttributeConfig> mapAttributeConfigs) {
+        this.extractors = instantiateExtractors(mapAttributeConfigs);
+        this.getterCache = new EvictableGetterCache(MAX_CLASSES_IN_CACHE, MAX_GETTERS_PER_CLASS_IN_CACHE,
+                EVICTION_PERCENTAGE);
+        this.argumentsParser = new DefaultArgumentParser();
+    }
+
+    public Object extract(Object targetObject, String attributeName) {
+        if (targetObject != null) {
```
jerrinot: I believe this could be simplified.
```java
if (targetObject == null) {
  return null;
}
```

It eliminates the second branch. it's a taste thing anyway.

jerrinot:
this is the only real issue I can see: https://github.com/hazelcast/hazelcast/pull/6758#discussion_r45196151

rest looks fine, the other comments I've made are mostly taste-related.
:+1: once the 1st issue is fixed and PR builder is green.


A general note: I am not sure if JavaDoc should use terms like 'Swap' to explain concepts. It should use will-known terms.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/Extractors.java
tombujok: It's basically caused by the smell of Getter;s `abstract Object getValue(Object obj) throws Exception;`
I mean I can't do anything with this exception so IMHO it should be a runtime exception.

The other workaround I saw was in the ReflectionHelper: 
`throw ExceptionUtil.rethrow(e);`

Top level QueryableEntry looks like this `public Object getAttributeValue(String attributeName) throws QueryException` but I don't want to drag this Checked exception through all signatures on the way.

tombujok: Actually QueryException extends RuntimeException, so I changed it to:
`throw new QueryException(ex.getMessage(), ex);`

hazelcast/src/main/java/com/hazelcast/query/impl/getters/Extractors.java
tombujok: Thx! Fixed.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/Extractors.java
tombujok: Fixed.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/Extractors.java
tombujok: Thx. I skipped it for now.

tombujok:
Thx for the reviews. It will be merged once green.

tombujok:
verify

PR #6765, opened 2015-11-17
Correct request should be used when removing item listeners

cangencer:


sancar:
run-lab-run

sancar:
Can you also include the ClientQueueProxy in this fix. It looks like I made the same mistake also there.

cangencer:
done, there was the same problem in TopicProxy as well.

pveentjer:
Can you create a test which demonstrates that the original problem existed and verifies that the problem is now resolved?

cangencer:
@pveentjer There are tests in enterprise repo, where this problem originated in the first place. See hazelcast/hazelcast-enterprise#493

sancar:
:+1:

pveentjer:
If this functionality is used on HZ OS, we should have test for HZ OS.

cangencer:
As all remove listener requests are very similar (they just pass a topic name + registration id), the only way to test them is using SecurityInterceptors which can intercept the actual method name. Those are available only on Enterprise.

hasancelik:
run-lab-run

PR #6767, opened 2015-11-18
Increment stats on async map operations

### LINKED ISSUE #6731: putAsync does not affect LocalMapStats.getPutOperationCount()
I am using Hazelcast 3.5.3. It seems that putAsync does not affect LocalMapStats.getPutOperationCount(). Is it possible to add all async operations to the Put or Get or Delete OperationsCounts in LocalMapStats?

metanet:
Fixes #6731

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -388,6 +388,7 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
             final InvocationBuilder invocationBuilder
                     = operationService.createInvocationBuilder(SERVICE_NAME,
                     operation, partitionId).setResultDeserialized(false);
+            final long startTime = System.currentTimeMillis();
             final InternalCompletableFuture<Data> future = invocationBuilder.invoke();
             future.andThen(new ExecutionCallback<Data>() {
                 @Override
@@ -406,6 +407,9 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
                 public void onFailure(Throwable t) {
                 }
             });
+            if (getMapConfig().isStatisticsEnabled()) {
+                future.andThen(new IncrementStatsExecutionCallback(operation, startTime));
+            }
             return future;
         } catch (Throwable t) {
             throw ExceptionUtil.rethrow(t);
@@ -454,14 +458,7 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
                         .setResultDeserialized(false)
                         .invoke();
                 o = f.get();
-                if (operation instanceof BasePutOperation) {
-                    localMapStats.incrementPuts(System.currentTimeMillis() - time);
-                } else if (operation instanceof BaseRemoveOperation) {
-                    localMapStats.incrementRemoves(System.currentTimeMillis() - time);
-                } else if (operation instanceof GetOperation) {
-                    localMapStats.incrementGets(System.currentTimeMillis() - time);
-                }
-
+                incrementStats(operation, System.currentTimeMillis() - time);
             } else {
                 f = operationService.createInvocationBuilder(SERVICE_NAME, operation, partitionId)
                         .setResultDeserialized(false).invoke();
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -406,6 +407,9 @@ public void onResponse(Data response) {
                 public void onFailure(Throwable t) {
                 }
             });
+            if (getMapConfig().isStatisticsEnabled()) {
```
pveentjer: I don't know if this is a good thing. Now you are going to introduce a new object for every map update and this causes jitter.

hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
```diff
@@ -54,6 +54,23 @@ public void testPutAndHitsGenerated() throws Exception {
...
+        HazelcastInstance h1 = createHazelcastInstance();
+        IMap<Integer, Integer> map = h1.getMap(randomMapName());
+        for (int i = 0; i < 100; i++) {
+            map.putAsync(i, i);
+        }
+        final LocalMapStats localMapStats = map.getLocalMapStats();
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run()
+                    throws Exception {
```
pveentjer: Why are you doing exceptions like this? I see it more frequent in the code but we have not listed this as an official style and we should apply it consistently throughout the codebase.

metanet: It is the hazelcast code style I imported to my intellij idea. If it is not preferred, we should remove it from the code style file that exist in our confluence page.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
metanet: yeah, I know that I am creating an extra object for every update but I don't have a better solution for this bug. Just like the qusp team, core team goes to great lengths to fix bugs and solve issues.

pveentjer: Sure. But I always get a bit frustrated when hard work gets undone. 

Actually the problem is bigger than 1 object; actually it is 2. Also the node object in the invocation future is created. I have this on my list for 3.7; so optimization for invocation future with just a single callback.

hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
pveentjer: Is code automatically formatted so that the exception is parked underneath the method declaration?

metanet: Normally, I do not hit "format code" in intellij and the "auto-format" feature is also disabled but from time to time intellij makes re-formatting somehow and I miss it.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
metanet: maybe we can share a single callback for multiple futures.

pveentjer: The problem in your case is that you have to have the start time; you are now using the callback as a container for the start time so you can calculate the difference on completion. 

The problem is that map was not designed for speed and statistics are enabled by default, but quite expensive; also in memory usage since additional object + fields are tracked. Making the memory overhead by default larger than needs to be.

hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
pveentjer: I hit the auto format button often :) In most cases there are no changes in the code.

gurbuzali:
:+1:

gurbuzali:
run-lab

metanet:
run-lab-run

mdogan:
:+1: 
As far as I think there's no other way of supporting stats for async ops other than registering a callback.

Issue #6620, opened 2015-11-02
[TEST-FAILURE] ClientPartitionLostListenerTest.test_partitionLostListener_removed


hasancelik:
```
java.lang.AssertionError: expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.partitionservice.ClientPartitionLostListenerTest$1.run(ClientPartitionLostListenerTest.java:131)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:825)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:839)
	at com.hazelcast.client.partitionservice.ClientPartitionLostListenerTest.assertRegistrationsSizeEventually(ClientPartitionLostListenerTest.java:123)
	at com.hazelcast.client.partitionservice.ClientPartitionLostListenerTest.test_partitionLostListener_removed(ClientPartitionLostListenerTest.java:75)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-ZuluJDK6/com.hazelcast$hazelcast-client-legacy/7/testReport/junit/com.hazelcast.client.partitionservice/ClientPartitionLostListenerTest/test_partitionLostListener_removed/

sancar:
fixed by #6728

hasancelik:
```
java.lang.AssertionError: expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.partitionservice.ClientPartitionLostListenerTest$1.run(ClientPartitionLostListenerTest.java:131)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:825)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:839)
[...]

PR #6768, opened 2015-11-18
Listener service should listen for events after clusterService has been started to prevent receiving duplicate events.

cangencer:
Fix for #6620


hasancelik:
run-lab-run

sancar:
run-lab-run

PR #6769, opened 2015-11-18
Discovery SPI minor changes.

kobalski:
Adds fail fast to JoinConfig for Discovery SPI when enabled with TCP/IP discovery.
Adds a property to enable use of public IP addresses in DiscoveryJoiner.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/JoinConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/JoinConfig.java
@@ -118,16 +118,27 @@ public class JoinConfig {
         if (getMulticastConfig().isEnabled() && getAwsConfig().isEnabled()) {
             throw new InvalidConfigurationException("Multicast and AWS join can't be enabled at the same time");
         }
+        verifyDiscoveryProviderConfig();
+    }
 
+    /**
+     * Verifies this JoinConfig is valid. When Discovery SPI enabled other discovery
+     * methods should be disabled
+     *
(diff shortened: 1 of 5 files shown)


noctarius:
It is perfectly legal to have predefined IPs + discovery, this is for mixed private+public clouds. You're right about AWS though.

kobalski:
Ok, removed restriction on having predefined IPs + discovery together. Thanks for the warning.

kobalski:
verify

hazelcast/src/main/java/com/hazelcast/config/JoinConfig.java
```diff
@@ -118,16 +118,27 @@ public void verify() {
...
             throw new InvalidConfigurationException("Multicast and AWS join can't be enabled at the same time");
         }
+        verifyDiscoveryProviderConfig();
+    }
 
+    /**
+     * Verifies this JoinConfig is valid. When Discovery SPI enabled other discovery
+     * methods should be disabled
+     *
+     * @throws IllegalStateException when the join config is not valid.
```
mesutcelik: `@throws IllegalStateException when the join config is not valid.` <-- that is not correct.

mesutcelik:
:+1:

kobalski:
verify

kobalski:
verify

bilalyasar:
LGTM :+1: , also needs some docs change.

kobalski:
verify

PR #6774, opened 2015-11-18
Fix Data Race in Record classes

jerrinot:
Multiple threads access the value field.
It has to be volatile to avoid data races.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/record/DataRecord.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/record/DataRecord.java
@@ -20,7 +20,7 @@ import com.hazelcast.nio.serialization.Data;
 
 class DataRecord extends AbstractRecord<Data> {
 
-    protected Data value;
+    protected volatile Data value;
 
     DataRecord(Data value) {
         this.value = value;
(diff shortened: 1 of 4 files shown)


jerrinot:
run-lab-run

jerrinot:
run-lab-run

gurbuzali:
`ObjectRecordWithStats` and `DataRecordWithStats` should be changed accordingly too

jerrinot:
thank you Ali, this is a very good catch!

jerrinot:
Done

hazelcast/src/main/java/com/hazelcast/map/impl/record/DataRecord.java
```diff
@@ -20,7 +20,7 @@
 
 class DataRecord extends AbstractRecord<Data> {
 
-    protected Data value;
+    protected volatile Data value;
```
pveentjer: Could you add a line of comments on the volatiles? You can do this with a hard push; but then other people are not going to get confused and remove it.

pveentjer:
:+1:

pveentjer:
@gurbuzali can you have a look

PR #6788, opened 2015-11-19
Fixes Ringbuffer readManyAsync when minValue is 0.

### LINKED ISSUE #6787: Problem with ringbuffer.readManyAsync and minSize is 0
Hello,

I'm a bit stumped by the non-blocking behaviour of readManyAsync method (Hazelcast 3.5).

If I want a non-blocking call (minCount = 0) :

ReadResultSet<String> rs = rb.readManyAsync(sequence, 0, 10, null).get();

the ReadResultSet *NEVER* returns anything.

If I make the readManyAsync call blocking with a minCount > 0, then it works fine.

ReadResultSet<String> rs = rb.readManyAsync(sequence, 1, 10, null).get();

(If I want an interruptible thread, I can also add a timeout on the get, it works fine too) :

ReadResultSet<String> rs = rb.readManyAsync(seq.get(), 1, 10, null).get(500, TimeUnit.MILLISECONDS);

But if I specify a minCount of 0, it never reads anything.

I attach a minimal test case, but really wondering what I'm missing here, any help appreciated :-)

RingbufferTest .java

```
import static org.junit.Assert.fail;

import java.util.concurrent.atomic.AtomicBoolean;
import java.util.concurrent.atomic.AtomicLong;

import org.junit.Test;

import com.hazelcast.ringbuffer.ReadResultSet;
import com.hazelcast.ringbuffer.Ringbuffer;


public class RingbufferTest {

	@Test
	public void test() throws InterruptedException, ExecutionException {
		Ringbuffer<String> rb = hz.getRingbuffer("testrb");
		
		
		final StringBuffer result = new StringBuffer();
		final AtomicLong seq = new AtomicLong(rb.headSequence());
		final AtomicBoolean running = new AtomicBoolean(true);
		
		Thread t = new Thread(new Runnable(){

			@Override
			public void run() {
				while(running.get()){
					System.out.print("readManyAsync...");
					ReadResultSet<String> rs;
					try {
                                                // 
						rs = rb.readManyAsync(seq.get(), 0, 10, null).get();
						rs.forEach((str) ->{
							result.append(str);
							// stop the thread
							running.set(false);
						});
						System.out.println(rs.readCount());
						seq.addAndGet(rs.readCount());
						Thread.sleep(500);
					} catch (Exception e) {
						e.printStackTrace();
						fail(e.getMessage());
					} 
				}
				
			}
			
		});
		rb.add("string1");
		assertEquals(1,rb.size());

		t.start();
		
		Thread.sleep(5000);
		System.out.print("Stopping...");
		// stop the Thread
		running.set(false);
		Thread.sleep(1000);
		assertTrue(result.length() > 0);
		hz.shutdown();
		
	}

}
```

pveentjer:
There was a logical error in the ReadManyOperation that in case of minSize = 0,
that the ringbuffer would not be read at all. So nothing gets returned.

Fixes #6787

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/ReadManyOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/operations/ReadManyOperation.java
@@ -73,13 +73,22 @@ public class ReadManyOperation extends AbstractRingBufferOperation implements Wa
             sequence = startSequence;
         }
 
-        if (resultSet.isMinSizeReached()) {
-            // if the minimum number of items are found, we don't need to wait anymore.
+        RingbufferContainer ringbuffer = getRingBufferContainer();
+        if (minSize == 0) {
+            if (!ringbuffer.shouldWait(sequence)) {
+                sequence = ringbuffer.readMany(sequence, resultSet);
+            }
+
             return false;
         }
 
-        RingbufferContainer ringbuffer = getRingBufferContainer();
+        if (resultSet.isMinSizeReached()) {
+            // enough items have been read, we are done.
+            return false;
+        }
+
         if (ringbuffer.shouldWait(sequence)) {
+            // the sequence is not readable
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


cangencer:
ringbuffer.readManyAsync(0, 0, 0, null);

will give an assertion error. Probably there should be a check in the proxy not to allow maxSize = 0

pveentjer:
Thanks. I'll fix that in a different PR since it isn't related to the actual problem.

pveentjer:
verify

cangencer:
:+1:

jerrinot:
:+1:

PR #6791, opened 2015-11-19
Ignore lite members on cluster merge

metanet:
If a sub-cluster has less number of data members with some additional lite members, it should join to the other sub-cluster with more data members.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/JoinMessage.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/JoinMessage.java
@@ -35,17 +35,18 @@ public class JoinMessage implements DataSerializable {
     protected boolean liteMember;
     protected ConfigCheck configCheck;
     protected Collection<Address> memberAddresses;
+    protected int dataMemberCount;
 
     public JoinMessage() {
     }
 
     public JoinMessage(byte packetVersion, int buildNumber, Address address,
                        String uuid, boolean liteMember, ConfigCheck configCheck) {
-        this(packetVersion, buildNumber, address, uuid, liteMember, configCheck, Collections.<Address>emptySet());
+        this(packetVersion, buildNumber, address, uuid, liteMember, configCheck, Collections.<Address>emptySet(), 0);
     }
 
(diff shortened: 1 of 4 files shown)


metanet:
unrelated test failure

metanet:
verify

hazelcast/src/main/java/com/hazelcast/cluster/impl/JoinMessage.java
```diff
@@ -34,25 +36,25 @@
     protected String uuid;
     protected boolean liteMember;
     protected ConfigCheck configCheck;
-    protected Collection<Address> memberAddresses;
+    protected Collection<Member> members;
```
mdogan: What's the reason of sending members instead of addresses?

metanet: I used them to count data members on the merging side but I think we can only send number of data members

mdogan: I see. I missed the place you filtered the members. Or you can filter data-members on sending side and only send addresses of data-members.

metanet: Yes but we wait until 2 sub-clusters become completely disjoint before merging them back. Because of this I still send lite members.

mdogan: Right! withdrawing my comment :)

metanet: Should I convert it to data member count or keep it as it is now?

mdogan: I think I prefer adding data-member count and sending only addresses.

metanet: ok

metanet:
@mdogan pr is updated accordingly

mdogan:
:+1:

PR #6792, opened 2015-11-19
Added eviction if value loaded from maploader

### LINKED ISSUE #6517: Possible memory leak with IMap.containsKey
I am making some tests with IMap in Hazelcast 3.5.3.  These tests are being run with -Xms100M -Xmx100M.

My Map has <Integer, String> where I had 2MB Strings, running with no eviction  will exhaust memory around 30 items.

First I added an eviction policy at 20 items. This made the test run perfectly with items being evicted when needed.

I then added a MapStore. As these are just local tests the store simply saves and loads to a file in disk, in a real scenario persistence would be done to a distributed persistence. With the MapStore i can recover items that were present in the past and update them.

Problem happens when I use containsKey for a value that exists in the MapStore. I see the Item being loaded from the Store but it does not get added to the Map (no event being triggered in this case), eventually the system crashes with no memory.

The code is a bit long so I added it to a gist: https://gist.github.com/gfelisberto/1075ea5428920c9a2746 I can move it here if needed.

gurbuzali:
Added eviction if value loaded from maploader for readonly operations like get and containsKey, removed postReadCleanup since it is now redundant.
fixes #6517

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractEvictableRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractEvictableRecordStore.java
@@ -48,37 +48,22 @@ import static com.hazelcast.map.impl.MapService.SERVICE_NAME;
  */
 abstract class AbstractEvictableRecordStore extends AbstractRecordStore {
 
-    /**
-     * Number of reads before clean up.
-     * A nice number such as 2^n - 1.
-     */
-    private static final int POST_READ_CHECK_POINT = 63;
-
+    private final long expiryDelayMillis;
+    private final Evictor evictor;
     /**
      * Iterates over a pre-set entry count/percentage in one round.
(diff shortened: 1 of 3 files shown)


gurbuzali:
@ahmetmircik should we do a back-port, WDYT?

gurbuzali:
unrelated test failure
```
	Test Result (1 failure / ±0)

    com.hazelcast.map.IssuesTest.testIssue1067GlobalSerializer
```

gurbuzali:
verify

ahmetmircik:
@gurbuzali this seems good. Can you please also remove unused constants/fields like `POST_READ_CHECK_POINT` ? There should also be some other unneeded ones after removal of `postReadCleanUp` method. Also we need a backport.

:+1:

gurbuzali:
back-port https://github.com/hazelcast/hazelcast/pull/6805

hasancelik:
run-lab-run

gurbuzali:
verify

jerrinot:
:+1:

PR #6797, opened 2015-11-19
Clear near-caches of caches while node is shutdown (in `SHUTTING_DOWN` state)

### LINKED ISSUE #6778: Remaining cache invalidation messages should be flushed on `ICacheService` shutdownwhile node is shutdown (in `SHUTTING_DOWN` state)

serkan-ozal:
At the moment, while `ICacheService` is shutdown, there might be remaining and non-flushed invalidation events for cache when batch invalidation mode is enabled (which is default).

With this PR, while node is shutting down (in `SHUTTING_DOWN` state), `ICacheService` registers and listens state changes through `LifecycleService` and send clear near-cache events for all caches when state is `SHUTTING_DOWN`. We send clear near-cache event instead of flushing remaning invalidation events since sending all remaining invalidation events are more expensive than just sending a single clear event.  

However, even though events are published, they may not be received by clients due to these reasons:
* Events may not be processed in the event queue due to shutdown event queue executor. This may be happend after events are offered to event queue, before  event queue executor polls and dispatches, the executor itself might be shutdown while `EventService` shutdown. 
* The events may be dispatched to local invalidation event listeners and write to client endpoint but they may not be processed and sent to client endpoint due to `ClientWriteHandler` shutdown.
* ...

With this PR, as mentioned above there is still no guarantee to deliver invalidation events to clients in case of shutdown but at least it tries :)

**P.S:** Maybe we can skip event system layer (just for this case) and directly send to all registered client end points but as mentioned above, still there is no guarantee.

Fixes #6778

### DIFF
--- a/hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTest.java
+++ b/hazelcast-client-legacy/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTest.java
@@ -68,6 +68,16 @@ public class ClientNearCacheTest extends ClientNearCacheTestSupport {
     }
 
     @Test
+    public void putToCacheAndGetInvalidationEventWhenNodeShutdownWithBinaryInMemoryFormat() {
+        putToCacheAndGetInvalidationEventWhenNodeShutdown(InMemoryFormat.BINARY);
+    }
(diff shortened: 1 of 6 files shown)


ahmetmircik:
:+1:

PR #6805, opened 2015-11-20
Added eviction if value loaded from maploader for readonly operations

gurbuzali:
added eviction if value loaded from maploader for readonly operations like get and containsKey, removed postReadCleanup since it is now redundant


hasancelik:
run-lab-run

PR #6812, opened 2015-11-20
[Backport] Reconnecting to instance on same address should preserve member list correctly

cangencer:
When the client reconnects to the cluster and receives the initial member list, it will rebuild a list of events according to the changes. The ordering of these events should be removals first and then the new members.


asimarslan:
:+1:

PR #6820, opened 2015-11-23
Removed one extra toData conversion when finding partition-id of key

ahmetmircik:
thanks @serkan-ozal for pointing this


gurbuzali:
merging this one since it is very trivial

Issue #6565, opened 2015-10-26
Query Performance regression


pveentjer:
By default the IMap will no longer cache object instance when a field is accessed multiple times in queries. This will lead to extremely bad performance.

jerrinot:
fixed by #6831

PR #6831, opened 2015-11-23
Introduces CacheDeserializedValues into MapConfig

jerrinot:
Allows to configure caching with finer granularity
than (now deprecated) optimizeQueries boolean flag.

### DIFF
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
@@ -275,13 +275,30 @@
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
-                                <xs:attribute name="optimize-queries" use="optional" type="parameterized-boolean"
-                                              default="false">
+                                <xs:attribute name="optimize-queries" use="optional" type="parameterized-boolean">
                                     <xs:annotation>
                                         <xs:documentation>
+                                            This parameter is deprecated as of Hazelcast 3.6
+                                            Use cache-deserialized-values attribute instead.
+
+                                            When both optimize-query and cache-deserialized-values are used at the same time
+                                            Hazelcast will do its best to detect possible conflicts. Conflict detection
+                                            is done on best-effort basis and you should not rely on it. 
+
                                             This parameter is used to increase the speed of query processes in the map.
                                             It only works when `in-memory-format` is set as `BINARY` and performs
                                             a pre-caching on the entries queried.
+                                        </xs:documentation>
+                                    </xs:annotation>
+                                </xs:attribute>
+                                <xs:attribute name="cache-deserialized-values" use="optional" type="cache-deserialized">
+                                    <xs:annotation>
+                                        <xs:documentation>
+                                            Control caching of de-serialized values. Caching makes query evaluation faster, but it cost memory.
+                                            Possible Values:
+                                            NEVER: Never cache de-serialized object
+                                            INDEX-ONLY: Cache values only when they are inserted into an index.
+                                            ALWAYS: Always cache de-serialized values.
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
@@ -1487,6 +1504,14 @@
 
     <xs:simpleType name="interface">
         <xs:restriction base="xs:string"/>
+    </xs:simpleType>
+
+    <xs:simpleType name="cache-deserialized">
+        <xs:restriction base="non-space-string">
+            <xs:enumeration value="NEVER"/>
+            <xs:enumeration value="ALWAYS"/>
+            <xs:enumeration value="INDEX-ONLY"/>
+        </xs:restriction>
     </xs:simpleType>
 
     <xs:simpleType name="member">
--- a/hazelcast/src/main/java/com/hazelcast/config/CacheDeserializedValues.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/CacheDeserializedValues.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.config;
+
+import com.hazelcast.util.StringUtil;
+
+import java.util.Arrays;
+
+/**
+ *
+ * Control caching of de-serialized values. Caching makes query evaluation faster but it cost memory.
+ *
+ * To maintain reusability cached values are used by read-only operations only and they are never
+ * passed to user-code where they could be accidentally mutated. Users will always get a fresh object,
+ * which they are free to mutate.
+ *
+ * It does not have any effect when {@link com.hazelcast.nio.serialization.Portable} serialization or
+ * {@link InMemoryFormat#OBJECT} format is used.
+ *
+ * @since 3.6
+ */
+public enum CacheDeserializedValues {
+
+    /**
+     * Never cache de-serialized value
+     *
+     */
+    NEVER,
+
+    /**
+     * Cache values only when using search indexes.
+     *
+     */
+    INDEX_ONLY,
+
+    /**
+     * Always cache de-serialized values
+     *
+     */
+    ALWAYS;
+
+    /**
+     * Create instance from String
+     *
+     * @param string
+     * @return instance of {@link CacheDeserializedValues}
+     * @throws IllegalArgumentException when unknown value is passed
+     */
+    public static CacheDeserializedValues parseString(String string) {
+        String upperCase = StringUtil.upperCaseInternal(string);
+        if ("NEVER".equals(upperCase)) {
+            return NEVER;
+        } else if ("INDEX_ONLY".equals(upperCase) || "INDEX-ONLY".equals(upperCase)) {
+            return INDEX_ONLY;
+        } else if ("ALWAYS".equals(upperCase)) {
+            return ALWAYS;
+        } else {
+            throw new IllegalArgumentException("Unknown CacheDeserializedValues option '" + string + ". "
+                    + "Possible options: " + Arrays.toString(CacheDeserializedValues.values()) + " .");
+        }
+    }
+}
--- a/hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
@@ -87,6 +87,11 @@ public class MapConfig {
      */
     public static final InMemoryFormat DEFAULT_IN_MEMORY_FORMAT = InMemoryFormat.BINARY;
 
+    /**
+     * We want to cache values only when an index is defined.
+     */
+    public static final CacheDeserializedValues DEFAULT_CACHED_DESERIALIZED_VALUES = CacheDeserializedValues.INDEX_ONLY;
+
     private String name;
 
     private int backupCount = DEFAULT_BACKUP_COUNT;
@@ -111,7 +116,7 @@ public class MapConfig {
 
     private boolean readBackupData;
 
-    private boolean optimizeQueries;
+    private CacheDeserializedValues cacheDeserializedValues = DEFAULT_CACHED_DESERIALIZED_VALUES;
 
     private String mergePolicy = DEFAULT_MAP_MERGE_POLICY;
 
@@ -139,6 +144,11 @@ public class MapConfig {
 
     private MapConfigReadOnly readOnly;
 
+    // we use these 2 flags to detect a conflict between (deprecated) #setOptimizeQueries()
+    // and #setCacheDeserializedValues()
+    private boolean optimizeQueryExplicitlyInvoked;
+    private boolean setCacheDeserializedValuesExplicitlyInvoked;
+
     public MapConfig(String name) {
         this.name = name;
     }
@@ -161,7 +171,7 @@ public class MapConfig {
         this.mapStoreConfig = config.mapStoreConfig != null ? new MapStoreConfig(config.mapStoreConfig) : null;
         this.nearCacheConfig = config.nearCacheConfig != null ? new NearCacheConfig(config.nearCacheConfig) : null;
         this.readBackupData = config.readBackupData;
-        this.optimizeQueries = config.optimizeQueries;
+        this.cacheDeserializedValues = config.cacheDeserializedValues;
         this.statisticsEnabled = config.statisticsEnabled;
         this.mergePolicy = config.mergePolicy;
         this.wanReplicationRef = config.wanReplicationRef != null ? new WanReplicationRef(config.wanReplicationRef) : null;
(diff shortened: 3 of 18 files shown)


hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -674,15 +679,66 @@ public boolean isNearCacheEnabled() {
...
+     *
+     * Enable de-serialized value caching when evaluating predicates.
+     *
+     * @param optimizeQueries
+     * @return this {@code MapConfig} instance.
+     * @deprecated use {@link #setCacheDeserializedValues(CacheDeserializedValues)} instead
+     */
     public MapConfig setOptimizeQueries(boolean optimizeQueries) {
-        this.optimizeQueries = optimizeQueries;
+        if (optimizeQueries) {
```
jerrinot: I am not sure whether (and how aggressively) we should try to detect conflict between (now deprecated) `optimizeQueries` and `CacheDeserializedValues` options

jerrinot:
Fixes #6565

jerrinot:
run-lab-run

jerrinot:
run-lab-run

jerrinot:
run-lab-run

jerrinot:
run-lab-run

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -674,15 +679,66 @@ public boolean isNearCacheEnabled() {
...
+     * @deprecated user {@link #getQueryCacheConfigs()} instead.
+     */
     public boolean isOptimizeQueries() {
-        return optimizeQueries;
+        return cacheDeserializedValues == CacheDeserializedValues.ALWAYS;
     }
 
+    /**
+     *
+     * Enable de-serialized value caching when evaluating predicates.
```
ahmetmircik: Ca you also add to javaDoc here that what will be the behaviour when InMemoryFormat is OBJECT? Or is this only valid for InMemoryFormat#BINARY ?

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -674,15 +679,66 @@ public boolean isNearCacheEnabled() {
...
+    /**
+     * Configure value de-serialized value caching.
+     * Default: {@link CacheDeserializedValues#INDEX_ONLY}
+     *
+     * @param cacheDeserializedValues
+     * @return this {@code MapConfig} instance.
+     * @since 3.6
+     * @see {@link CacheDeserializedValues}
+     */
+    public MapConfig setCacheDeserializedValues(CacheDeserializedValues cacheDeserializedValues) {
```
ahmetmircik: Ca you also add to javaDoc here that what will be the behaviour when InMemoryFormat is OBJECT? Or is this only valid for InMemoryFormat#BINARY ?

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
jerrinot: if InMemoryFormat is OBJECT then this has no effect of query evaluation. I'll amend the JavaDoc.

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
jerrinot: it's described in details in the linked `CacheDeserializedValues`. See the line 725.

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -674,15 +679,66 @@ public boolean isNearCacheEnabled() {
...
+                throw new ConfigurationException("Deprecated option 'optimize-queries' is set to false, "
+                        + "but 'cacheDeserializedValues' is set to ALWAYS. These are conflicting options. Please "
+                        + "remove the `optimize-queries'");
+            }
+        }
         return this;
     }
 
+    /**
+     * Configure value de-serialized value caching.
```
ahmetmircik: `Configure de-serialized value caching` ?

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
ahmetmircik: ops just saw it

ahmetmircik:
:+1:  with minor comments

jerrinot:
updated.

tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/config/CacheDeserializedValues.java
```diff
@@ -0,0 +1,76 @@
...
+        } else if ("INDEX_ONLY".equals(upperCase) || "INDEX-ONLY".equals(upperCase)) {
+            return INDEX_ONLY;
+        } else if ("ALWAYS".equals(upperCase)) {
+            return ALWAYS;
+        } else {
+            throw new IllegalArgumentException("Unknown CacheDeserializedValues option '" + string + ". "
+                    + "Possible options: " + Arrays.toString(CacheDeserializedValues.values()) + " .");
+        }
+    }
+}
```
tombujok: You could do `CacheDeserializedValues.valueOf(string)` to use the internal static method that does the same thing.

hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
```diff
@@ -275,16 +275,29 @@
...
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
+                                <xs:attribute name="cache-deserialized-values" use="optional" type="cache-deserialized">
+                                    <xs:annotation>
+                                        <xs:documentation>
+                                            Control caching of de-serialized values. Caching makes query evaluation faster, but it cost memory.
+                                            Possible Values:
+                                            NEVER: Never cache de-serialized object
+                                            INDEX-ONLY: Caches values only when they are inserted into an index.
```
tombujok: Small inconsistenty: NEVER and ALWAYS descriptions are written in imperative thus it would be better to use it in all descrpitions like: `Cache values...` without `s`

jerrinot: good one, thanks!

hazelcast/src/main/java/com/hazelcast/config/CacheDeserializedValues.java
jerrinot: well, almost. except it would not work for `INDEX-ONLY` (with dash, not with underscore) and underscores look alien in XML configuration.

hazelcast/src/main/resources/hazelcast-config-3.6.xsd
```diff
@@ -103,6 +103,17 @@
...
             </xs:element>
+            <xs:element name="cache-deserialized-values" type="cache-deserialized-values" minOccurs="0" maxOccurs="1" default="INDEX-ONLY">
+                <xs:annotation>
+                    <xs:documentation>
+                        Control caching of de-serialized values. Caching makes query evaluation faster, but it cost memory.
+                        Possible Values:
+                        NEVER: Never cache de-serialized object
+                        INDEX-ONLY: Caches values only when they are inserted into an index.
+                        ALWAYS: Always cache de-serialized values.
+                    </xs:documentation>
```
tombujok: The same here.

hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
```diff
@@ -275,16 +275,29 @@
...
+                                <xs:attribute name="optimize-queries" use="optional" type="parameterized-boolean">
                                     <xs:annotation>
                                         <xs:documentation>
+                                            This parameter is deprecated as of Hazelcast 3.6
+                                            Use cache-deserialized-values attribute instead.
+
                                             This parameter is used to increase the speed of query processes in the map.
                                             It only works when `in-memory-format` is set as `BINARY` and performs
                                             a pre-caching on the entries queried.
                                         </xs:documentation>
```
tombujok: Maybe we should add a comment somewhere what happens if both options are used at once?

hazelcast/src/main/java/com/hazelcast/config/CacheDeserializedValues.java
tombujok: ah, ok. In enum, wouldn't `INDEX-ONLY` work due to checkstyle?

jerrinot: I am not sure. however that would look alien in Java anyway :)

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -87,6 +87,11 @@
      */
     public static final InMemoryFormat DEFAULT_IN_MEMORY_FORMAT = InMemoryFormat.BINARY;
 
+    /**
+     * We want to cache values only when an index is defined.
+     */
+    public static final CacheDeserializedValues DEFAULT_CACHED_DESERIALIZED_VALUES = CacheDeserializedValues.INDEX_ONLY;
+
```
tombujok: This would be a great piece of information for the reference-manual (I haven't looked at the PRs there, so don't know if it's there)

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -674,15 +679,68 @@ public boolean isNearCacheEnabled() {
...
+                        + "remove the `optimize-queries'");
+            }
+        } else {
+            if (cacheDeserializedValues == CacheDeserializedValues.ALWAYS) {
+                throw new ConfigurationException("Deprecated option 'optimize-queries' is set to false, "
+                        + "but 'cacheDeserializedValues' is set to ALWAYS. These are conflicting options. Please "
+                        + "remove the `optimize-queries'");
+            }
+        }
         return this;
```
tombujok: Hmm, this relies on the invocation order of `setOptimizeQueries` and `setCacheDeserializedValues`.
I think we should invoke method `validate` in each method so that if somebody changes the order it still works...

hazelcast/src/main/java/com/hazelcast/config/CacheDeserializedValues.java
tombujok: ok, let's keep it then ;)

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -796,6 +855,7 @@ public String toString() {
                 + ", mapAttributeConfigs=" + mapAttributeConfigs
                 + ", quorumName=" + quorumName
                 + ", queryCacheConfigs=" + queryCacheConfigs
+                + ", cacheDeserializedValues=" + cacheDeserializedValues
                 + '}';
```
tombujok: Is the `cacheDeserializedValues` skipped in the hashCode for speed?

hazelcast/src/test/java/com/hazelcast/map/impl/record/DataRecordFactoryTest.java
```diff
@@ -0,0 +1,89 @@
...
+        object = new Object();
+        data = new HeapData();
+        when(mockSerializationService.toData(object, mockPartitioningStrategy)).thenReturn(data);
+
+    }
+
+    @Test
+    public void givenStatisticsEnabledAndCacheDeserializedValuesIsNEVER_thenCreateDataRecordWithStats() {
+        MapConfig mapConfig = new MapConfig().setStatisticsEnabled(true).setCacheDeserializedValues(CacheDeserializedValues.NEVER);
+        DataRecordFactory dataRecordFactory = new DataRecordFactory(mapConfig, mockSerializationService, mockPartitioningStrategy);
```
tombujok: if no `// GIVEN`, etc comments it would be great to have blank lines to separate the test stages, otherwise it's difficult to asses what the test is doing at the first look at this class

enesakar:
run-lab-run

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
jerrinot: will do

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
jerrinot: that's a good question. the `optimizeQuery` is not included in `hashCode` / `equals` calculation. cacheDeserializedValues is just an extension of the same idea so I didn't want to change hashCode / equals behaviour. @ahmetmircik: WDYT?

hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
jerrinot: done. 
```
When both optimize-query and cache-deserialized-values are used at the same time
Hazelcast will do its best to detect possible conflicts. Conflict detection
is done on best-effort basis and you should not rely on it. 
```
WDYT?

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
jerrinot: this is a bit tricky since I have removed the `optimizeQuery` flag. The `setOptimizeQueries()` actually set the `cacheDeserializedValues` under the hood. The beauty of encapsulation.


Any I might re-introduce the `optimizeQuery` flag just to help with conflict detection. WDYT?

hazelcast/src/test/java/com/hazelcast/map/impl/record/DataRecordFactoryTest.java
jerrinot: done, thanks

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
tombujok: Thx

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
tombujok: Ah, I see. 
Hmm, yes, IMHO anything that would help to detect this situation would be fine. It's a great piece of work, but these flags may lead to some confusion if not used properly.

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
tombujok: I meant that the `cacheDeserializedValues` it's used in equals and not used in the hashCode method. In theory it's fine but a bit unusual, thus my question. The only reason it's like this I could figure out was the performance of hashCode. On the other hand it's just a boolean.

hazelcast/src/test/java/com/hazelcast/map/impl/record/DataRecordFactoryTest.java
tombujok: Thx!

hazelcast-spring/src/main/resources/hazelcast-spring-3.6.xsd
tombujok: Sounds good. Thx!

tombujok:
:+1:  great job
(my minor comments are the hashCode method and conflict detection - but I don't have a very strict opinion on those so plz decide for yourself)

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
ahmetmircik: @jerrinot i do not have a reason to not to add `cacheDeserializedValues ` into hashCode and equals methods. Seems `optimizeQuery` is a missed one also like `hotRestartEnabled`. 
They can be added.

jerrinot: thanks!

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -139,6 +144,11 @@
 
     private MapConfigReadOnly readOnly;
 
+    // we use these 2 flags to detect a conflict between (deprecated) #setOptimizeQueries()
+    // and #setCacheDeserializedValues()
+    private boolean optimizeQueryExplicitlyInvoked;
+    private boolean setCacheDeserializedValuesExplicitlyInvoked;
+
```
tombujok: nice!

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -674,15 +684,96 @@ public boolean isNearCacheEnabled() {
...
+                        + "but 'cacheDeserializedValues' is set to NEVER. "
+                        + "These are conflicting options. Please remove the `optimize-queries'");
+            } else if (!optimizeQueries && cacheDeserializedValues == CacheDeserializedValues.ALWAYS) {
+                throw new ConfigurationException("Deprecated option 'optimize-queries' is set to false, "
+                        + "but 'cacheDeserializedValues' is set to ALWAYS. "
+                        + "These are conflicting options. Please remove the `optimize-queries'");
+            }
+        }
+    }
+
```
tombujok: Looks good. `INDEX_ONLY` is kind of neutral which is fine.

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -674,15 +684,96 @@ public boolean isNearCacheEnabled() {
...
+            if (cacheDeserializedValues != validatedCacheDeserializedValues) {
+                boolean optimizeQueriesFlagState = cacheDeserializedValues == CacheDeserializedValues.ALWAYS;
+                throw new ConfigurationException("Deprecated option 'optimize-queries' is set to "
+                        + optimizeQueriesFlagState + " but 'cacheDeserializedValues' is set to "
+                        + validatedCacheDeserializedValues + ". These are conflicting options. "
+                        + "Please remove the `optimize-queries'");
+            }
+        }
+    }
+
```
tombujok: Looks good.

hazelcast/src/test/java/com/hazelcast/config/MapConfigTest.java
```diff
@@ -289,4 +290,120 @@ public void testMapPartitionLostListener_equalsWithImplementation() {
...
+    @Test(expected = ConfigurationException.class)
+    public void givenCacheDeserializedValuesSetToNEVER_whenSetOptimizeQueriesToTrue_thenThrowConfigurationException() {
+        //given
+        MapConfig mapConfig = new MapConfig();
+        mapConfig.setCacheDeserializedValues(CacheDeserializedValues.NEVER);
+
+        //when
+        mapConfig.setOptimizeQueries(true);
+    }
+
```
tombujok: cool that the conflict detection is covered

PR #6839, opened 2015-11-23
On server shutdown, the client should shutdown without temporary dead…

### LINKED ISSUE #6829: Client shutdown takes more than 30 seconds
When all the nodes client connected is closed client should try to connect to the nodes predefined number of times and shutdown itself if no available connection is found. In that scenario, client is unable to shutdown itself and waits for 30 seconds.

Problematic behavior can be found in `ClusterListenerSupport.connectionRemoved`, in case of an exception, client is shutdown. Shutdown process tries to shutdown `clusterExecutor` but it cannot since the shutdown process is running over there.  The process is deadlocked for 30 seconds.

```
HazelcastInstance server = Hazelcast.newHazelcastInstance();
server.shutdown();
```

cangencer:
…lock

After server shutdown, the client will shutdown after connection attempts
have failed. However there was a deadlock preventing the ClusterService
from being shutdown. Fixes #6829


cangencer:
verify

hasancelik:
run-lab-run

sancar:
:+1:

cangencer:
run-lab-run

cangencer:
run-lab-run

sancar:
:+1:

PR #6851, opened 2015-11-24
Code generator changes for CPP code generation are added.

ihsandemir:
Code generation related changes for CPP code generation are implemented. I had to change some conflicting parameter names in templates such as "register" which is a keyword in C++ and the names of response parameters to response where possible. Added the cpp codec templates.

### DIFF
--- a/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
+++ b/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
@@ -29,7 +29,6 @@ public final class CodeGenerationUtils {
     public static final String CODEC_PACKAGE = "com.hazelcast.client.impl.protocol.codec.";
     public static final String DATA_FULL_NAME = "com.hazelcast.nio.serialization.Data";
 
-
     private static final Map<String, String> JAVA_TO_CSHARP_TYPES = new HashMap<String, String>() {{
         put(DATA_FULL_NAME, "IData");
         put("java.lang.String", "string");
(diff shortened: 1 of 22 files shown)


hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
```diff
@@ -47,6 +46,30 @@
...
+        put("char", "int8_t");
+        put("byte", "uint8_t");
+        put("long", "int64_t");
+        put(DATA_FULL_NAME, "serialization::pimpl::Data");
+        put("java.lang.String", "std::string");
+        put("byte[]", "std::vector<byte>");
+        put("java.util.List", "std::vector");
+        put("java.util.Set", "std::vector");
+        put("java.util.Collection", "std::vector");
+        put("java.util.Map", "std::vector<std::pair");
```
mdogan: Is closing bracket `>` missing? Shouldn't this be `std::vector<std::pair>`?

ihsandemir: No, it was intentional, the closing bracket is inserted at the end after the inner types in getLanguageType method. It looks weird but it was the easiest way for me.

hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
```diff
@@ -251,18 +280,34 @@ public static String getCSharpType(String type) {
...
+                builder.append(getLanguageType(iterator.next(), languageMapping));
                 if (iterator.hasNext()) {
-                    builder.append(',');
+                    builder.append(", ");
                 }
             }
-            builder.append(">");
-            return builder.toString();
+
+            if (type.startsWith("java.util.Map<")) {
```
cangencer: is this c++ specific? as it's not required for c#

hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
cangencer: We should not have Map type in any of the codecs, but rather List<Entry<>>. This will mean you don't need this hack either.

ihsandemir: In MapCodec putAll and ResponseTemplate still uses map. Hence, I shall keep Map mapping in place for now.

hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGenerationUtils.java
ihsandemir: Yes, it is for c++.

cangencer: Ok but then It will break the code generated for C# with an extra >

ihsandemir: @cangencer OK, I got you point now. Updated the code so that the change is only for CPP.

cangencer:
:+1:

PR #6857, opened 2015-11-24
Tiny addition: Fetched near-cache config dynamically

ahmetmircik:
Closes https://github.com/hazelcast/hazelcast/issues/5485. Btw dynamic configuration of near-cache is not an officially supported feature. This PR only makes the config dynamically reachable. Current near-cache tests should be enough to prove this. 

PS: There is also a PR from the requester https://github.com/hazelcast/hazelcast/pull/5547

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
@@ -17,11 +17,12 @@
 package com.hazelcast.map.impl.nearcache;
 
 import com.hazelcast.cache.impl.nearcache.NearCache;
-import com.hazelcast.config.Config;
 import com.hazelcast.config.EvictionPolicy;
 import com.hazelcast.config.InMemoryFormat;
+import com.hazelcast.config.MapConfig;
 import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.internal.serialization.SerializationService;
+import com.hazelcast.map.impl.MapContainer;
 import com.hazelcast.map.impl.SizeEstimator;
 import com.hazelcast.monitor.impl.NearCacheStatsImpl;
 import com.hazelcast.nio.serialization.Data;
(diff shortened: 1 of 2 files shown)


pveentjer:
Tests are missing. 

You have changed behavior, so tests are needed for that. 

Also it apparently solves a bug, and I don't see a test for that either.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
```diff
@@ -88,17 +88,35 @@ public NearCacheImpl(String mapName, NodeEngine nodeEngine) {
...
+        NearCacheConfig nearCacheConfig = mapConfig.getNearCacheConfig();
+        return TimeUnit.SECONDS.toMillis(nearCacheConfig.getTimeToLiveSeconds());
     }
 
     // TODO this operation returns the given value in near-cache memory format (data or object)?
     @Override
     public void put(Data key, Object value) {
         fireTtlCleanup();
-        if (evictionPolicy == NONE && cache.size() >= maxSize) {
+        if (evictionPolicy == NONE && cache.size() >= getMaxSize()) {
```
pveentjer: Why are you retrieving this value every time? It obfuscates the regular flow since the CPU needs to chew through a lot of additional code.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
```diff
@@ -185,7 +203,7 @@ public void run() {
                         try {
                             lastCleanup = Clock.currentTimeMillis();
                             for (Map.Entry<Data, NearCacheRecord> entry : cache.entrySet()) {
-                                if (entry.getValue().isExpired(maxIdleMillis, timeToLiveMillis)) {
+                                if (entry.getValue().isExpired(getMaxIdleMillis(), getTimeToLiveMillis())) {
```
pveentjer: The same goes for this code; additional fat getters the CPU needs to chew through.

pveentjer:
This PR looks more like a feature than a bug fix + there are no tests + it makes code more complex than it needs to be.

jerrinot:
I don't understand how the existing tests are able to validate the newly added behaviour.

ahmetmircik:
closing this one, dynamic configuration of near-cache is not an officially supported feature.

PR #6870, opened 2015-11-25
Increased code coverage of UnsafeHelper

Donnerbart:
File is missing code coverage in at least one feature.

Do you think it's safe to assume that `sun.misc.Unsafe` is available on our build machines? Otherwise two of these tests may fail.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/UnsafeHelper.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/UnsafeHelper.java
@@ -72,7 +72,6 @@ public final class UnsafeHelper {
             + "seem to support unaligned access to memory. Unsafe usage has been enforced via System Property "
             + UNSAFE_MODE_PROPERTY_NAME + " This is not a supported configuration and it can crash JVM or corrupt your data!";
 
-
     static {
         Unsafe unsafe;
         try {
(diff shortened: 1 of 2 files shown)


Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
verify

Donnerbart:
@serkan-ozal @noctarius Could you have a look at this one, please? I don't know if it's a good idea to introduce tests which actually test if Unsafe could be retrieved. I actually like the idea, because we would see failures immediately in JDKs which don't support this. But this could also mean broken builds in Jenkins in the long term. WDYT?

(the cause for this test is the missing code coverage for a feature, so I looked for some low hanging fruits to increase the coverage of that class)

noctarius:
:+1:

serkan-ozal:
@Donnerbart good job :+1: 

BTW, in fact you should be able to use parallel runner even though you set system properties in the tests. 
I had added local and thread-local system properties for serial and parallel test runners.
See my previous PR #6193

But this is not necessary, since tests are already fast no need to make them parallel. I have just wanted to inform you about this behaviour.

Donnerbart:
Oh, that very good to know. I was unsure about that, so I preferred to run this test in isolation. As you mentioned, it's very fast, so no need to change this at the moment.

Thanks for the feedback :)

Issue #6858, opened 2015-11-25
PublicAddressTest was ignored due to running time.


hasancelik:
This situation affected pr-builder's completion time.
https://github.com/hazelcast/hazelcast/commit/e734fbc9a3472e599dae112d282db4c7a3008211

gurbuzali:
fixed via https://github.com/hazelcast/hazelcast/pull/6884

PR #6884, opened 2015-11-26
MulticastJoiner try count should be bounded

### LINKED ISSUE #5598: Multicast joiner slow
I noticed that MulticastJoiner can be much slower than configuration parameters imply.

See the code below (MulticastJoiner.calculateTryCount()).
tryCount is adjusted by a value which can increase significantly the time a node is waiting to decide if is a master or not. 

For example, even if I set multicast timeout seconds to 1, it still takes 8 seconds for a node to decide that is a master node. In my case tryCount is initially 10 and then adjusted by 65 to a total value of 75. So even if the MulsticastJoiner is configured to make a decision in one second, it still takes 8 seconds to do so.

This is especially painful in tests where 1 or 2 seconds would be just fine and running tests faster is important.

```java
private int calculateTryCount() {
        final NetworkConfig networkConfig = config.getNetworkConfig();
        int timeoutSeconds = networkConfig.getJoin().getMulticastConfig().getMulticastTimeoutSeconds();
        int tryCountCoefficient = 1000 / PUBLISH_INTERVAL;
        int tryCount = timeoutSeconds * tryCountCoefficient;
        String host = node.getThisAddress().getHost();
        int lastDigits;
        try {
            lastDigits = Integer.parseInt(host.substring(host.lastIndexOf('.') + 1));
        } catch (NumberFormatException e) {
            lastDigits = RandomPicker.getInt(512);
        }
        lastDigits = lastDigits % 100;
        int portDiff = node.getThisAddress().getPort() - networkConfig.getPort();
        tryCount += lastDigits + portDiff * timeoutSeconds * 3;
        return tryCount;
    }
```

cangencer:
Currently MulticastJoiner has no bounds, especially when portDiff is large
it can result in a very large try count (>10000)

Removing ignore from tests which were ignored because of slowness with this issue.

Fixes #5598 and #6858

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cluster/impl/MulticastJoiner.java
+++ b/hazelcast/src/main/java/com/hazelcast/cluster/impl/MulticastJoiner.java
@@ -31,7 +31,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 public class MulticastJoiner extends AbstractJoiner {
 
-    private static final int PUBLISH_INTERVAL = 100;
+    private static final int PUBLISH_INTERVAL_MIN = 50;
+    private static final int PUBLISH_INTERVAL_MAX = 200;
     private static final long JOIN_RETRY_INTERVAL = 1000L;
 
     private final AtomicInteger currentTryCount = new AtomicInteger(0);
@@ -164,7 +165,7 @@ public class MulticastJoiner extends AbstractJoiner {
                 node.multicastService.send(joinRequest);
                 if (node.getMasterAddress() == null) {
                     //noinspection BusyWait
-                    Thread.sleep(PUBLISH_INTERVAL);
+                    Thread.sleep(getPublishInterval());
                 } else {
                     return node.getMasterAddress();
                 }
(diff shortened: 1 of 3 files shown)


cangencer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/cluster/impl/MulticastJoiner.java
```diff
@@ -191,9 +191,8 @@ private int calculateTryCount() {
         } catch (NumberFormatException e) {
             lastDigits = RandomPicker.getInt(512);
         }
-        lastDigits = lastDigits % 100;
         int portDiff = node.getThisAddress().getPort() - networkConfig.getPort();
-        tryCount += lastDigits + portDiff * timeoutSeconds * 3;
+        tryCount += (lastDigits + portDiff) % 10;
```
mdogan: Additionally, we can change `PUBLISH_INTERVAL` to a variable interval, like between 50 ~ 200 for example, to avoid all nodes publishing at the same time, instead of a constant 100.

cangencer: I've randomized the interval now to be between 50-200

mdogan:
looks good to me.

PR #6890, opened 2015-11-26
cleanup PR related to ignored CORE module tests

hasancelik:
Please check `Ignored Tests Report` page at Confluence.

* This PR decreases ignore count 48 to 27 for the core module.
* Deleted forgotten and useless tests


hasancelik:
run-lab-run

hasancelik:
run-lab-run

hazelcast/src/test/java/com/hazelcast/nio/ConnectionTest.java
```diff
@@ -65,7 +65,7 @@ public void cleanup() throws Exception {
     }
 
     @Test
-    @Ignore
+//    @Ignore
```
Donnerbart: The whole class is ignored, so I don't know if this will change anything. The class is also missing `@RunWith` and `@Category`:
```
@Ignore("See testBlockedClientSockets and testBlockedClientSockets2 tests. " +
        "Currently we couldn't find a way to make them pass...")
public class ConnectionTest {
```

Donnerbart:
:+1: for the `@Ignore("Not a JUnit test")` removals (those seem to be outdated tests from the pre-CI area). I cannot say if all other ignored tests are useless or just disabled because they had failures.

hasancelik:
run-lab-run

hazelcast/src/test/java/com/hazelcast/quorum/map/MapReadWriteQuorumTest.java
```diff
@@ -455,35 +455,4 @@ public void testSubmmitToKeyOperationThrowsExceptionWhenQuorumSizeNotMet() throw
         foo.get();
     }
 
-    @Ignore
```
gurbuzali: can you ask @eminn about these tests, why are they ignored ?

gurbuzali:
for `*ClientRequestTest` classes they were meant to test requests without creating a full-blown client.
But we've never updated the tests other than ignoring when they've failed
With the new client they are totally redundant

hazelcast/src/test/java/com/hazelcast/quorum/map/MapReadWriteQuorumTest.java
eminn: Quorum for lock operations has never been implemented. I've left the tests if that's implemented in near future but it didn't happened. you can delete them.

hasancelik:
run-lab-run

PR #6894, opened 2015-11-26
EWR related fixes to multiple entry ops

emrahkocaman:
EWR related cleanup and fixes.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MultipleEntryOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MultipleEntryOperation.java
@@ -104,7 +104,12 @@ public class MultipleEntryOperation extends AbstractMultipleEntryOperation imple
     @Override
     public Operation getBackupOperation() {
         EntryBackupProcessor backupProcessor = entryProcessor.getBackupProcessor();
-        return backupProcessor != null ? new MultipleEntryBackupOperation(name, keys, backupProcessor) : null;
+        MultipleEntryBackupOperation backupOperation = null;
+        if (backupProcessor != null) {
+            backupOperation = new MultipleEntryBackupOperation(name, keys, backupProcessor);
+            backupOperation.setWanEventList(wanEventList);
+        }
+        return backupOperation;
     }
 
     @Override
--- a/hazelcast/src/test/java/com/hazelcast/wan/WanReplicationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/wan/WanReplicationTest.java
@@ -414,7 +414,7 @@ public class WanReplicationTest extends HazelcastTestSupport {
         assertGivenDataAppliedEventually(clusterB, "map", 0, 1000, "TEST");
     }
 
-    static class MyEntryProcessor implements EntryProcessor {
+    static class MyEntryProcessor implements EntryProcessor, EntryBackupProcessor {
 
         @Override
         public Object process(Entry entry) {
@@ -423,7 +423,12 @@ public class WanReplicationTest extends HazelcastTestSupport {
 
         @Override
         public EntryBackupProcessor getBackupProcessor() {
-            return null;
+            return MyEntryProcessor.this;
+        }
+
+        @Override
+        public void processBackup(Entry entry) {
+            entry.setValue("TEST");
         }
     }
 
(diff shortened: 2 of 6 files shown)


emrahkocaman:
run-lan-run

mesutcelik:
run-lan-run

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MultipleEntryOperation.java
```diff
@@ -104,7 +104,12 @@ public int getAsyncBackupCount() {
     @Override
     public Operation getBackupOperation() {
         EntryBackupProcessor backupProcessor = entryProcessor.getBackupProcessor();
-        return backupProcessor != null ? new MultipleEntryBackupOperation(name, keys, backupProcessor) : null;
+        MultipleEntryBackupOperation backupOperation = null;
+        if (backupProcessor != null) {
+            backupOperation = new MultipleEntryBackupOperation(name, keys, backupProcessor);
+            backupOperation.setWanEventList(wanEventList);
+        }
+        return backupOperation;
```
mesutcelik: is it ok to return `null` backup operation?

hazelcast/src/test/java/com/hazelcast/wan/WanReplicationTest.java
```diff
@@ -423,7 +423,12 @@ public Object process(Entry entry) {
...
         @Override
         public EntryBackupProcessor getBackupProcessor() {
-            return null;
+            return MyEntryProcessor.this;
+        }
+
+        @Override
+        public void processBackup(Entry entry) {
+            entry.setValue("TEST");
         }
```
mesutcelik: is this a fix for a failing test? I don't see an addition into testing part of the code...

mesutcelik:
:+1: with two minor comments

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MultipleEntryOperation.java
emrahkocaman: yes, this is the original behaviour, nothing has changed there.

hazelcast/src/test/java/com/hazelcast/wan/WanReplicationTest.java
emrahkocaman: This is to increase test coverage. Otherwise backup operations are not called.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MultipleEntryOperation.java
```diff
@@ -104,7 +104,12 @@ public int getAsyncBackupCount() {
     @Override
     public Operation getBackupOperation() {
         EntryBackupProcessor backupProcessor = entryProcessor.getBackupProcessor();
-        return backupProcessor != null ? new MultipleEntryBackupOperation(name, keys, backupProcessor) : null;
+        MultipleEntryBackupOperation backupOperation = null;
+        if (backupProcessor != null) {
+            backupOperation = new MultipleEntryBackupOperation(name, keys, backupProcessor);
+            backupOperation.setWanEventList(wanEventList);
```
Donnerbart: Did I get it right that this `setWanEventList()` call copies the existing events from the `MultipleEntryOperation` instead of creating new ones in the `AbstractMultipleEntryBackupOperation` via `wanEventList.add(new WanEventWrapper(key, null, EntryEventType.REMOVED|UPDATED));`?

Donnerbart:
The code changes look reasonable to me. I also couldn't find any usage of the deleted constructor. And if I understood the event list copy (see inline comment) I think the functionality didn't change.

:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MultipleEntryOperation.java
emrahkocaman: Yes, this is needed to keep events ordered.

Donnerbart: I first didn't see where the `wanEventList` entries were created (since the `new WanEventWrapper` is gone), but then I realized they were just copied. Makes sense and looks good to me.

PR #6900, opened 2015-11-27
`CacheEntryCountResolver` and its usage by cache stats are simplified

serkan-ozal:


hazelcast/src/main/java/com/hazelcast/cache/impl/CacheStatisticsImpl.java
```diff
@@ -98,7 +102,11 @@ public long getLastUpdateTime() {
 
     @Override
     public long getOwnedEntryCount() {
-        return cacheEntryCountResolver.getEntryCount();
+        if (cacheEntryCountResolver != null) {
```
pveentjer: Isn't it easier to introduce a NullCacheEntryCountResolver which returns 0 so you don't need to do any if else checking? Not terribly important.

pveentjer:
:+1:

Issue #6141, opened 2015-09-07
LeaseLock simulator test hanging, 3-6.SNAPSHOT


Danny-Hazelcast:
I see all 10 threads of all 4 boxes hanging for ever 

```
grep -r --include *.out LeaseLockTest .
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest$Worker.run(LeaseLockTest.java:64)
./54.209.174.3/54.209.174.3-serverJstack.out:	at com.hazelcast.simulator.tests.concurrent.lock.LeaseLockTest.run(LeaseLockTest.java:48)
[...]

PR #6904, opened 2015-11-27
Fix SecondsBasedEntryTaskScheduler issues

mdogan:
- SecondsBasedEntryTaskScheduler was using System.nanoTime() as unique
identifier and order key for scheduled entries. But it's known that
System.nanoTime() can return the same value on consecutive calls.

- While preparing entries to process, scheduled entries was being stored
in a set and then copied to a list to sort.
There are two problems with this approach:
	* Extra copy, which is obvious
	* When key/value pair is scheduled more than once (FOR_EACH)
	then only one instance of entry is used for processing.
	But every entry must be processed regardless of their values,
	once they are scheduled (when schedule() method returns true).

_Lock lease uses this scheduler, probably fixes https://github.com/hazelcast/hazelcast/issues/6141_

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/scheduler/CompositeKey.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/scheduler/CompositeKey.java
@@ -19,16 +19,16 @@ package com.hazelcast.util.scheduler;
 /**
  * This object is first needed for having different key objects that are created for the same key updated in
  * different times so should be persisted consequently.
- * So using ScheduleType.FOR_EACH, all the updates will be scheduled seperately.
+ * So using ScheduleType.FOR_EACH, all the updates will be scheduled separately.
  */
-final class TimeKey {
+final class CompositeKey {
 
-    private Object key;
-    private long time;
+    private final Object key;
+    private final long id;
 
-    public TimeKey(Object key, long time) {
+    public CompositeKey(Object key, long id) {
         this.key = key;
-        this.time = time;
+        this.id = id;
     }
 
     public Object getKey() {
@@ -44,9 +44,9 @@ final class TimeKey {
             return false;
         }
 
-        TimeKey that = (TimeKey) o;
+        CompositeKey that = (CompositeKey) o;
 
-        if (time != that.time) {
+        if (id != that.id) {
             return false;
         }
         if (key != null ? !key.equals(that.key) : that.key != null) {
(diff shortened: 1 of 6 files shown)


hazelcast/src/test/java/com/hazelcast/util/scheduler/SecondsBasedEntryTaskSchedulerTest.java
```diff
@@ -120,23 +106,25 @@ public void test_rescheduleEntry_postpone() {
     }
 
     @Test
+    // setting system property, not a ParallelTest !
+    @Category(QuickTest.class)
```
jerrinot: is per-method category actually working?

hazelcast/src/main/java/com/hazelcast/util/scheduler/CompositeKey.java
```diff
@@ -59,12 +59,12 @@ public boolean equals(Object o) {
...
         int result = key != null ? key.hashCode() : 0;
-        result = 31 * result + (int) (time ^ (time >>> 32));
+        result = 31 * result + (int) (id ^ (id >>> 32));
         return result;
     }
 
     @Override
     public String toString() {
-        return "TimeKey{" + "key=" + key + ",time=" + time + '}';
+        return "Key{" + "key=" + key + ",time=" + id + '}';
```
ahmetmircik: seems it should be `id` instead of `time` in message string.

hazelcast/src/test/java/com/hazelcast/util/scheduler/SecondsBasedEntryTaskSchedulerTest.java
mdogan: yes, it was working when I last tried.

hazelcast/src/main/java/com/hazelcast/util/scheduler/CompositeKey.java
mdogan: will fix

hazelcast/src/test/java/com/hazelcast/util/scheduler/SecondsBasedEntryTaskSchedulerTest.java
serkan-ozal: @mdogan in fact, this test should work even with parallel runner without effecting other tests.
I had added local and thread-local system properties for serial and parallel test runners.
See my previous PR https://github.com/hazelcast/hazelcast/pull/6193

gurbuzali:
:+1:

hazelcast/src/test/java/com/hazelcast/util/scheduler/SecondsBasedEntryTaskSchedulerTest.java
mdogan: Thanks @serkan-ozal. I had seen your before, but somehow I forgot it. You're right, I'll remove the comment.

jerrinot:
:+1:

PR #6905, opened 2015-11-27
[BACKPORT] Fix SecondsBasedEntryTaskScheduler issues

mdogan:
- SecondsBasedEntryTaskScheduler was using System.nanoTime() as unique
identifier and order key for scheduled entries. But it's known that
System.nanoTime() can return the same value on consecutive calls.

- While preparing entries to process, scheduled entries was being stored
in a set and then copied to a list to sort.
There are two problems with this approach:
	* Extra copy, which is obvious
	* When key/value pair is scheduled more than once (FOR_EACH)
	then only one instance of entry is used for processing.
	But every entry must be processed regardless of their values,
	once they are scheduled (when schedule() method returns true).

Backport of https://github.com/hazelcast/hazelcast/pull/6904


jerrinot:
:+1:

mdogan:
Original PR is merged.

Issue #5128, opened 2015-04-21
ExecutorServiceTest.testInvokeAllTimeoutSuccess


bilalyasar:
```
java.lang.AssertionError: expected:<231> but was:<1000>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at com.hazelcast.executor.ExecutorServiceTest.testInvokeAllTimeoutSuccess(ExecutorServiceTest.java:740)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/com.hazelcast$hazelcast/215/testReport/junit/com.hazelcast.executor/ExecutorServiceTest/testInvokeAllTimeoutSuccess/

bilalyasar:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.7/com.hazelcast$hazelcast/241/testReport/junit/com.hazelcast.executor/ExecutorServiceTest/testInvokeAllTimeoutSuccess/

Donnerbart:
```
Regression

com.hazelcast.executor.ExecutorServiceTest.testInvokeAllTimeoutSuccess

Failing for the past 1 build (Since Unstable#568 )
Took 0.16 sec.
Error Message

expected:<null> but was:<Task completed>
Stacktrace

java.lang.AssertionError: expected:<null> but was:<Task completed>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at com.hazelcast.executor.ExecutorServiceTest.testInvokeAllTimeoutSuccess(ExecutorServiceTest.java:836)
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/com.hazelcast$hazelcast/568/testReport/junit/com.hazelcast.executor/ExecutorServiceTest/testInvokeAllTimeoutSuccess/

hasancelik:
```
java.util.concurrent.CancellationException: null
[...]

PR #6908, opened 2015-11-29
If value is null call inner future's get even if the future is marked as done

gurbuzali:
The reasoning behind this change is there is a race between `cancel` and `get` in`CancellableDelegatingFuture` which occurs while executing `IExecutorService.invokeAll(tasks, time, TimeUnit)`. race is in `CancellableDelegatingFuture.cancel` method:

```
@Override
    public boolean cancel(boolean mayInterruptIfRunning) {
        if (isDone() || cancelled) {
            return false;
        }

        Future f = invokeCancelOperation(mayInterruptIfRunning);
        try {
            Boolean b = (Boolean) f.get();
            if (b != null && b) {
                setError(new CancellationException());
                cancelled = true;
                return true;
            }
            return false;
        } catch (Exception e) {
            throw rethrow(e);
        } finally {
            setDone();
        }
    }
```
While executing cancel, it may happen that future was not done so we continue on sending cancel operation. But before cancel operation executed the inner future is completed which results returning false and setting the future as done (in finally block). Since the future is set as done `get` method in `DelegatingFuture` does not try to get the result from inner future but returns the `value` which is null.

With this change `get` method will try to get the inner future's result if the value is null, ignoring that the future is marked as done. 

fixes https://github.com/hazelcast/hazelcast/issues/5128

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -504,6 +504,8 @@
     <suppress checks="JavadocType" files="com/hazelcast/util/executor/"/>
     <suppress checks="JavadocMethod" files="com/hazelcast/util/executor/"/>
     <suppress checks="JavadocVariable" files="com/hazelcast/util/executor/"/>
+    <suppress checks="NPathComplexity" files="com.hazelcast.util.executor.DelegatingFuture"/>
+    <suppress checks="NPathComplexity|CyclomaticComplexity" files="com.hazelcast.client.util.ClientDelegatingFuture"/>
 
     <!-- Written by Doug Lea  -->
     <suppress checks="MagicNumber|FileLength|DeclarationOrder|RedundantModifier|InnerAssignment|NPath|Cyclomatic"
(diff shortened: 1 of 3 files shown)


gurbuzali:
```
java.lang.AssertionError: expected:<100> but was:<0>
	at com.hazelcast.collection.impl.list.ListMigrationTest.test(ListMigrationTest.java:48)
```
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/217/

gurbuzali:
run-lab-run

eminn:
:+1:

sancar:
What if request return with exception instead of a value ? As far as I understand it should not be a problem. But I want to initiate a bit of brainstorming if this is not considered yet. 
Pr seems to be fine, if I am not missing anything. 

And we probably have same problem with ClientDelegatingFuture.

gurbuzali:
Yes, throwing the exception should not cause any problem. I will check the client side

gurbuzali:
unrelated test-failure
```
Test Result (1 failure / +1)

    com.hazelcast.mapreduce.aggregation.MapMinAggregationTest.com.hazelcast.mapreduce.aggregation.MapMinAggregationTest
```
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/263/

gurbuzali:
verify

gurbuzali:
unrelated test failure
```
	Test Result (1 failure / +1)

    com.hazelcast.spi.impl.operationservice.impl.Invocation_NetworkSplitTest.testWaitingInvocations_whenNodeSplitFromCluster
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/17253/

gurbuzali:
verify

gurbuzali:
verify

Issue #6911, opened 2015-11-30
Badly named method in Discovery SPI configuration


jerrinot:
https://github.com/hazelcast/hazelcast/blob/9733142144f2b882f69bec5b389ca37c8b05ca43/hazelcast/src/main/java/com/hazelcast/config/DiscoveryConfig.java#L110-L110 

@noctarius has confirmed it's a left-over.  it should be `addDiscoveryStrategyConfig()`. 

Given it has never been released as a final I it's ok just to rename it.

jerrinot:
fixed by #6912

PR #6912, opened 2015-11-30
Misleading method name in Discovery SPI config object fixed

jerrinot:
It's a left-over from old Discovery SPI implementation. I am not deprecating the old method as it has never been released as a Final.
Fixing #6911


noctarius:
:+1: lovely :D

Donnerbart:
:+1:

mesutcelik:
:+1:  Thanks @jerrinot !

Issue #6153, opened 2015-09-08
[TEST-FAILURE] client.ClientRegressionWithMockNetworkTest timeout failures


hasancelik:
**ClientRegressionWithMockNetworkTest.testOperationRedo**
**ClientRegressionWithMockNetworkTest.testGithubIssue3557**

```
org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
	at sun.misc.URLClassPath$JarLoader.validIndex(URLClassPath.java:1032)
	at sun.misc.URLClassPath$JarLoader.getResource(URLClassPath.java:1166)
	at sun.misc.URLClassPath$JarLoader.getResource(URLClassPath.java:1077)
	at sun.misc.URLClassPath$JarLoader.findResource(URLClassPath.java:1047)
	at sun.misc.URLClassPath.findResource(URLClassPath.java:298)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:901)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:899)
	at java.security.AccessController.doPrivileged(AccessController.java:252)
```

**ClientRegressionWithMockNetworkTest.com.hazelcast.client.ClientRegressionWithMockNetworkTest**

```
java.lang.IllegalStateException: Instances haven't been shut down: [HazelcastInstance{name='_hzInstance_14_dev', node=Address[127.0.0.1]:5022}, HazelcastInstance{name='_hzInstance_27_dev', node=Address[127.0.0.1]:5027}]
	at com.hazelcast.test.AbstractHazelcastClassRunner$1.evaluate(AbstractHazelcastClassRunner.java:189)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-IbmJDK1.7/692/

sancar:
I did not do a detailed research but it looks like highly parallel tests are somehow causing deadlock in classLoader related parts.
@serkan-ozal Do you have an idea ?

hasancelik:
**I updated ibm version for this build but still we are getting this failure.**

[...]

PR #6917, opened 2015-11-30
Unignore ClientRegressionWithMockNetwork tests

cangencer:
These tests are being unignored, because it's not clear if the specific
failure in issue #6153 is related to these tests. The tests can be ignored
again if they're still failing.


sancar:
:+1:

PR #6923, opened 2015-11-30
Small fixes for issues found during hot-restart test

mdogan:
-  `InternalPartitionService.setInitialState()` should not set initialized flag if all replicas are empty
- `ClusterServiceImpl.addMembersRemovedInNotActiveState()` should skip local address


metanet:
:+1:

PR #6926, opened 2015-11-30
Server should respond to authentication message when node is not active

cangencer:
This seems to be a regression after the message decoding was moved to partition thread. When the node is not active, the error message is not sent back to the client


sancar:
:+1:

PR #6929, opened 2015-11-30
Adds rest api for cluster shutdown. Updated stop.sh script

kobalski:


hazelcast/src/main/resources/stop.sh
```diff
@@ -1,13 +1,104 @@
...
+
+if [[ -z "$SCOPE" ]]; then
+ 	echo "No scope is defined, running script with default scope : 'member'"
+ 	SCOPE="member"
+fi
+	
+if [ $SCOPE = "cluster" ]; then
+	
+	if [[ -z "$PORT" ]]; then
+		echo "No port is defined running script with default port : '5701'"
```
emrahkocaman: "," or "." needed after `No port is defined`. (Same thing for below messages)

hazelcast/src/test/java/com/hazelcast/internal/ascii/RestTest.java
```diff
@@ -182,4 +184,28 @@ public void testDisabledRest() throws IOException {
...
+                assertFalse(instance2.getLifecycleService().isRunning());
+            }
+        });
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run()
+                    throws Exception {
+                assertFalse(instance2.getLifecycleService().isRunning());
+            }
+        });
```
emrahkocaman: Looks like both `assertTrueEventually` checks the same condition.

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
```diff
@@ -44,6 +48,9 @@ public void handle(HttpPostCommand command) {
                 handleManagementCenterUrlChange(command);
             } else if (uri.startsWith(URI_QUEUES)) {
                 handleQueue(command, uri);
+            } else if (uri.startsWith(URI_SHUTDOWN_CLUSTER_URL)) {
+                handleClusterShutdown(command);
+                return;
```
emrahkocaman: Is `return` necessary here? Can't we just set the response of `command` and let the response be sent in line 60.

emrahkocaman: ah ok. we need to send response before shutdown.

noctarius:
Is it really a good idea to use groupname and grouppassword as a security feature, in general those two properties are not considered to be used for security? In general is it really a good idea to provide shutdown on the same port as typical rest api calls?

mesutcelik:
I believe user/pass is used for pre-caution rather than security feature. We can even add one more warning like `You are shutting down the cluster. please provide user/pass?`. @kobalski ?

Rest API is already used by management center to change the management center URL so this is another addition.

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
kobalski: Yes, exactly.

hazelcast/src/test/java/com/hazelcast/internal/ascii/RestTest.java
kobalski: Done. Also did not need two separate `assertTrueEventually `.

hazelcast/src/main/resources/stop.sh
kobalski: Added ','.

kobalski:
@noctarius @mesutcelik groupname and pass is only used for precaution. This PR is intended to provide cluster shutdown with our `stop.sh` script for Hot Restart feature.  In 3.7, we should have more secure and more well designed Cluster Management API and discuss all these concerns during the design of that API.

kobalski:
@bilalyasar I renamed `server.sh` to `start.sh`, Do we need to update our release script after this change.

mesutcelik:
renaming might not be a good option. It was always `server.sh` and people are used to this naming...

kobalski:
@mesutcelik I reverted renaming of `server.sh`

mesutcelik:
:+1:

emrahkocaman:
:+1:

hazelcast/src/main/resources/stop.sh
```diff
@@ -1,13 +1,105 @@
...
+
+	if [[ -z "$PASSWORD" ]]; then
+		echo "No password is defined, running script with default password : 'dev-pass'"
+		PASSWORD="dev-pass"
+	fi
+	echo "You are shutting down the cluster. Please make sure you provide valid user/pass."
+	echo "Shutting down cluster from member on this machine on port ${PORT}"
+
+	request="http://127.0.0.1:${PORT}/hazelcast/rest/management/cluster/shutdown"
+ 	response=$(curl --data "${GROUPNAME}&${PASSWORD}" --silent "${request}");
```
mdogan: Shouldn't you check if `curl` is available and print a warning if not?

kobalski: Good idea, I added check here : https://github.com/hazelcast/hazelcast/pull/6929/files#diff-d8a0ca814fb592a4a2d706bc7d79e76dR48

mdogan:
:+1:

PR #6935, opened 2015-12-01
Atomicity supported between cache record store and `CacheWriter` while adding (not for update) new entry

serkan-ozal:
Currently, writing to `CacheWriter` is done before record is added to cache record store. But in case of failure while adding to record store (such as native OOME at EE),  record is written to `CacheWriter` more than once while trying to adding to record store.

With this PR, writing to `CacheWriter` is done after record is added to cache record store successfully. Also to be consistent if writing to `CacheWriter` fails, new added records are reverted (removed from record store).

**P.S:** Note that this PR is only for add (new put) not for update. Update case will be handled in another PR.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheReadWriteThroughTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheReadWriteThroughTest.java
@@ -18,9 +18,9 @@ package com.hazelcast.cache;
 
 import com.hazelcast.cache.impl.HazelcastServerCachingProvider;
 import com.hazelcast.config.CacheConfig;
+import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
-import com.hazelcast.test.AssertTask;
-import com.hazelcast.test.HazelcastSerialClassRunner;
+import com.hazelcast.test.HazelcastParallelClassRunner;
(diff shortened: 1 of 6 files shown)


ahmetmircik:
:+1:

hazelcast/src/test/java/com/hazelcast/cache/CacheReadWriteThroughTest.java
```diff
@@ -60,47 +64,48 @@
...
     private HazelcastServerCachingProvider cachingProvider2;
 
+    protected Config createConfig() {
+        return new Config();
+    }
+
     @Before
-    public void init() {
+    public void setup() {
         factory = new TestHazelcastInstanceFactory(2);
```
emrahkocaman: Can't we use `HazelcastTestSupport#createHazelcastInstanceFacrory(int nodeCount)` here?

serkan-ozal: yes, we could. In fact, i had not looked at that part. Just renamed the method and added `createConfig()` to use same test at EE

emrahkocaman:
:+1:  (I've a minor comment, can be ignored as it's not directly related to this pr)

serkan-ozal:
@ahmetmircik @emrahkocaman thanks for review

PR #6940, opened 2015-12-01
Client proxy initialization should take care of any failure

mdogan:
Currently proxy creation/initialization only handles `Exception`
but when it fails with an `Error`, `ClientProxyFuture` is remains unset
and client shutdown and a subsequent getProxy call will hang forever.


pveentjer:
verify

pveentjer:
:+1:  once green

sancar:
:+1:

PR #6941, opened 2015-12-01
[BACKPORT] Client proxy initialization should take care of any failure

mdogan:
Currently proxy creation/initialization only handles `Exception`
but when it fails with an `Error`, `ClientProxyFuture` is remains unset
and client shutdown and a subsequent getProxy call will hang forever.

Backport of https://github.com/hazelcast/hazelcast/pull/6940


pveentjer:
:+1:

PR #6939, opened 2015-12-01
Wait for only cluster members in the same JVM during shutdown

metanet:
* NodeShutdownLatch waits until other hazelcast instances in the same JVM to notice the shutting down member. It retrieves all of the hazelcast instances in the same JVM even if they are in a different cluster. It also waits for the instances that have not completed the start process yet. Because of this, it slows down the shutdown since when a node tries to shutdown in a test method, it can wait for other Hazelcast instances starting in other test methods.

Fixes #6372

PR #6943, opened 2015-12-01
[BACKPORT] Wait for only cluster members in the same JVM during shutdown

metanet:
* NodeShutdownLatch waits until other hazelcast instances in the same JVM to notice the shutting down member. It retrieves all of the hazelcast instances in the same JVM even if they are in a different cluster. It also waits for the instances that have not completed the start process yet. Because of this, it slows down the shutdown since when a node tries to shutdown in a test method, it can wait for other Hazelcast instances starting in other test methods.

Backport of #6939


gurbuzali:
forward-port already merged

PR #6950, opened 2015-12-02
Query collections & arrays feature - code coverage supplemented

tombujok:
* code coverage supplemented
* bug fixed in `FastMultiResultSet.next()` method

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/FastMultiResultSet.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/FastMultiResultSet.java
@@ -27,7 +27,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 
 /**
- *  Multiple result set for Predicates.
+ * Multiple result set for Predicates.
  */
 public class FastMultiResultSet extends AbstractSet<QueryableEntry> implements MultiResultSet {
 
@@ -103,7 +103,16 @@ public class FastMultiResultSet extends AbstractSet<QueryableEntry> implements M
             if (resultSets.size() == 0) {
                 return null;
             }
-            return currentIterator.next();
+            if (currentIterator != null && currentIterator.hasNext()) {
+                return currentIterator.next();
+            }
+            while (currentIndex < resultSets.size()) {
+                currentIterator = resultSets.get(currentIndex++).values().iterator();
+                if (currentIterator.hasNext()) {
+                    return currentIterator.next();
+                }
+            }
+            return null;
         }
 
         @Override
--- a/hazelcast/src/test/java/com/hazelcast/query/impl/CachedQueryEntryTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/query/impl/CachedQueryEntryTest.java
@@ -17,13 +17,11 @@
 package com.hazelcast.query.impl;
 
 import com.hazelcast.internal.serialization.SerializationService;
-import com.hazelcast.internal.serialization.impl.DefaultSerializationServiceBuilder;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.query.impl.getters.Extractors;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.annotation.ParallelTest;
 import com.hazelcast.test.annotation.QuickTest;
-import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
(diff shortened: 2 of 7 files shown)


hazelcast/src/main/java/com/hazelcast/query/impl/FastMultiResultSet.java
```diff
@@ -103,7 +103,16 @@ public QueryableEntry next() {
             if (resultSets.size() == 0) {
                 return null;
             }
-            return currentIterator.next();
```
tombujok: This was a bug

hazelcast/src/test/java/com/hazelcast/query/impl/CachedQueryEntryTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class CachedQueryEntryTest extends QueryEntryTest {
+
+    @Test
+    public void getKey_caching() {
+        QueryableEntry entry = entry("key", "value");
+
+        assertThat(entry.getKey(), sameInstance(entry.getKey()));
```
pveentjer: afaik we are not using this approach in tests. We should have consistency in our tests.

Personally I don't see the added value compared to:
```
assertSame(entry.getKey(), entry.getKey())
```

hazelcast/src/test/java/com/hazelcast/query/impl/DuplicateDetectingMultiResultTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class DuplicateDetectingMultiResultTest {
+
+    private DuplicateDetectingMultiResult result = new DuplicateDetectingMultiResult();
+
+    @Test
+    public void testAddResultSet_empty() throws Exception {
+        assertThat(result.size(), is(0));
```
pveentjer: assertEquals(0, result.size())

hazelcast/src/test/java/com/hazelcast/query/impl/QueryEntryTest.java
```diff
@@ -48,7 +53,7 @@ public void getAttribute_whenValueIsPortableObject_thenConvertedToData() {
         Data key = serializationService.toData("indexedKey");
 
         Portable value = new SampleObjects.PortableEmployee(30, "peter");
-        QueryEntry queryEntry = new QueryEntry(serializationService, key, value, Extractors.empty());
+        QueryableEntry queryEntry = entry(serializationService, key, value, Extractors.empty());
```
pveentjer: why does serialization service need to be passed as argument since it a field of the test? This makes the line nosier than needs to be.

hazelcast/src/test/java/com/hazelcast/query/impl/getters/GetterFactoryTest.java
```diff
@@ -58,6 +77,15 @@ public void newFieldGetter_whenExtractingFromSimpleField_thenInferTypeFromFieldT
     }
 
     @Test
+    public void newMethodGetter_whenExtractingFromSimpleField_thenInferTypeFromFieldType() throws Exception {
```
pveentjer: what I normally do when running into very long test method-names, it to create a subclass of the test that only deals with e.g. method getter. So this part doesn't need to be repeated in the name.

pveentjer:
Looks good apart from the assertThat. I understand it can be read as a sentence, even though there is a lot of java syntactical noise. My main problem is that this isn't a style being used in HZ. I'm in favour of a single style; either yours or the existing one. But we should not have multiple styles in the code base.

tombujok:
Thanks for expressing your personal opinion. I acknowledge it.

AFAIK I see 275 usages of assertThat in tests so I disagree with your statement that it's not a style used in HZ.

I also see added value in `assertThat` in contrary to archaic junit `assert` statements that are hardly readable.

Consistency or single-style doesn't mean that everybody codes perfectly in the same way. 
One developer will do `trim().toLowerCase()`, other one will do `toLowerCase().trim()`.
One will use a couple of `if` statements, the other one will use a `switch`.
One will do one `if` braces and a return statement afterwards, another will do an `if-else` statement.
Would you like to ban using `switch` statements and describe it on the wiki? Probably not a good idea.
The same applies to `assertEquals(0, result.size())` vs. `assertThat(result.size(), is(0));`.
It's just a flavour.

I would rather be discussing things like this:
- is the test readable? Yes
- is the test testing what it's supposed to test? Yes
- is the test clean? Yes

Personally, I would rather work on bringing other tests from the codebase to a state where we could answer Yes to all these 3 questions instead of loosing energy on fighting about minor things.

Code is not a thing that may be standardised with ISO 9001. 
Developers are not robots either and thus I would be careful with too strict standardisation of minor things. 

I understand that people have and will have personal opinions about things (like you right now), but it would be great not to try to impose such opinions on others - that's my personal opinion.

hazelcast/src/test/java/com/hazelcast/query/impl/CachedQueryEntryTest.java
tombujok: Thanks for expressing your personal opinion - see the comment below.

hazelcast/src/test/java/com/hazelcast/query/impl/DuplicateDetectingMultiResultTest.java
tombujok: Thanks for expressing your personal opinion - see the comment below.

hazelcast/src/test/java/com/hazelcast/query/impl/QueryEntryTest.java
tombujok: Good catch - will fix it!

hazelcast/src/test/java/com/hazelcast/query/impl/getters/GetterFactoryTest.java
tombujok: Thx for your comment. IMHO long names are often the only way to express the intent of this test.
I think we can live with this ones too.

pveentjer:
We have a company standard for testing:

https://hazelcast.atlassian.net/wiki/display/EN/Testing+Practices

And I'm not imposing an opinion. The only thing we should prevent is that we get many styles of testing in the system. So as long as we stick with one style; whatever that style may be; I'm fine with it.

pveentjer:
Apart from the assertThat, there should be reasonably clear guidelines how to name a method. I have zero problems with your naming strategy; it isn't better or worse than other approaches we use in the system. But there should be consistency.

So if we select your naming approach, I would suggest doing one sweep over the code base to get it all fixed. And I'm sure that as part of the sweep, we get a lot of badly named methods resolved as well.

hazelcast/src/test/java/com/hazelcast/query/impl/getters/GetterFactoryTest.java
pveentjer: In this case I would create a class 

NewMethdoGetterFactoryTest

so that a test with name

newMethodGetter_whenExtractingFromSimpleField_thenInferTypeFromFieldType

can be simplified to

whenExtractingFromSimpleField_thenInferTypeFromFieldType

tombujok:
`company standard` - well I would abstain for using this name for assertions. 
Let's have a company standard for `if` and `switch` statements too in this case.

These are not different styles of testing. Different styles would be using spock instead of junit.
We are discussing a flavor of asserting.

` clear guidelines ` - yes that's true with a degree of freedom I suppose. On the other hand you can't align everything (and it's also not good to align everything), you can't make people work the way you would like everything to work either.

Donnerbart:
Can we please get this merged ASAP? We need to get done with the code coverage verification. If we decide later to not support `assertThat()` we can easily change it together with the other 275 occurrences.

@tombujok Can you please rebase?

tombujok:
@Donnerbart Yep, I've been working on rebasing for 1.5h now, there was a bit of overlap. But it should be fine now.

@jerrinot @pveentjer Will you give a thumbs up here?

tombujok:
BTW, It would be great if somebody reviewed the change in `FastMultiResultSet`. There was a bug there in the next method that I found just because we extended the test coverage.

jerrinot:
:+1:  - great addition.

pveentjer:
:+1:

PR #6960, opened 2015-12-02
`LoadAllTask` for client and server side cache proxies also handles `Throwable`, not just `Exception`

### LINKED ISSUE #6944: `LoadAllTask` for client and server side cache proxies should also handle `Throwable`, not just `Exception`
Currently, `LoadAllTask`, used by loadAll operation on client/server side caches, doesn't call `CompletionListener::onException` if received error is a `Throwable` but not an `Exception`.

`javax.cache.integration.CompletionListener::onException(Exception e)` takes `Exception` not `Throwable` so if received error is not an `Exception` but `Throwable` (such as `Error`), it should be wrapped inside an `Exception` as `Throwable` is cause and can be passed to `CompletionListener`.

serkan-ozal:
Fixes #6944


hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxyBase.java
```diff
@@ -201,6 +201,10 @@ public void run() {
                 if (completionListener != null) {
                     completionListener.onException(e);
                 }
+            } catch (Throwable t) {
```
sancar: What happens if t is OutOfMemoryError. How should we handle that ?

serkan-ozal: @sancar This error is not directly dispatched to user thread but passed to `CompletionListener`. So what about throwing it if it is `OutOfMemoryError` after `CompletionListener` is notified. WDYT?

sancar: It looks like it will work. I am thinking about if NativeOutOfMemoryError should have been Exception not Error in the first place. Other than that, pr looks good.

serkan-ozal:
run-lab-run

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxyBase.java
serkan-ozal: @sancar 

Even though there is eviction, force eviction and cross partition (but owned by same partition thread) based force eviction, still there might be `NativeOutOfMemoryError`.

The simplest case is that for example all native memory occupied by *parition_thread_1*, .., *parition_thread_n-1* and *partition_thread_n* has not used any native memory yet because there was no operation targeted to partitions owned by *partition_thread_n*. And then an operation comes to partition owned by *partition_thread_n* but all native memory is already in use. Also since there is no used native memory by neither that partition nor that partition thread, all evictions will have no effect so in this case, there will be `NativeOutOfMemoryError`. 

In this case, `CompletionListener` should be notified since `NativeOutOfMemoryError` might occur because there is no %100 guarantee

serkan-ozal:
run-lab-run

ahmetmircik:
Seems mergeable to me :+1:

sancar:
:+1:

PR #6962, opened 2015-12-02
Return true from task cancellation only if cancellation is sent to original invocation

### LINKED ISSUE #6283: Is this a bug in DistributedExecutionService.CallableProcessor.run()?
I apologize in advance if this is not really a bug but from what I can understand from the code in DistributedExecutionService.CallableProcessor.run() when checking for isCancelled() if true, a null **might** be sent as response which will override the sending of an exception for example in DistributorExecutionService.cancel().

Should the sendResponse in the run()'s finally block be part of the block of code called only it is not cancelled?

```
                sendResponse(result);
                if (!isCancelled()) {
                    finishExecution(name, Clock.currentTimeMillis() - start);
                }
```
Please feel free to change the summary/description for this issue if it is correctly identified as a bug.

Thanks!

metanet:
Fixes #6283


gurbuzali:
:+1:  I guess we need a backport here

metanet:
Yes, I am already preparing the backport

mdogan:
:+1:

PR #6965, opened 2015-12-02
Additional unit tests for query-related code added

jerrinot:
- Additional unit tests for query-related code added
- Bug in equals()/hashCode in QueryEventFilter fixed - it didn't call its superclass

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryEventFilter.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryEventFilter.java
@@ -70,6 +70,9 @@ public class QueryEventFilter extends EntryEventFilter {
             return false;
         }
         QueryEventFilter that = (QueryEventFilter) o;
+        if (!super.equals(o)) {
+            return false;
+        }
         if (!predicate.equals(that.predicate)) {
             return false;
         }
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryEventFilter.java
```diff
@@ -70,6 +70,9 @@ public boolean equals(Object o) {
             return false;
         }
         QueryEventFilter that = (QueryEventFilter) o;
+        if (!super.equals(o)) {
```
jerrinot: this was a bug as it didn't call the superclass.

hazelcast/src/main/java/com/hazelcast/query/impl/CachedQueryEntry.java
```diff
@@ -92,18 +89,11 @@ public Data getValueData() {
     protected Object getTargetObject(boolean key) {
         Object targetObject;
         if (key) {
-            if (keyObject == null) {
-                if (keyData.isPortable()) {
-                    targetObject = keyData;
-                } else {
-                    targetObject = getKey();
-                }
+            //keyData is never null
```
jerrinot: this is a simplification based on fact keyData is never null.

Donnerbart:
Dead code and bug are reasonable. Looks good to me.

:+1:

jerrinot:
verify

jerrinot:
Unrelated test failure issue created: #6966

pveentjer:
:+1:

PR #6971, opened 2015-12-03
Fix leaking local map stats

### LINKED ISSUE #6972: IMap Memory Leak Problems Upon Destroy
In the map data structure, there are multiple places which leaks memory by either re-creating objects redundantly or object that are not cleanup ( which should be ).

These places are, map containers, local map statistics and near cache invalidation queues.

eminn:
Below the results from MapMemoryUsageStressTest

Fixes #6972

Heap usage on upstream
screen shot 2015-12-03 at 10 38 48 am

Heap usage with the fixes in this PR
screen shot 2015-12-03 at 10 38 39 am

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
@@ -6,6 +6,10 @@ import com.hazelcast.config.PartitionGroupConfig;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
 import com.hazelcast.core.MultiMap;
+import com.hazelcast.instance.Node;
+import com.hazelcast.map.impl.LocalMapStatsProvider;
+import com.hazelcast.map.impl.MapService;
+import com.hazelcast.map.impl.MapServiceContext;
 import com.hazelcast.monitor.LocalMapStats;
 import com.hazelcast.nio.Address;
 import com.hazelcast.partition.InternalPartitionService;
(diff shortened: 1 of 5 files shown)


eminn:
run-lab-run

eminn:
run-lab-run

eminn:
run-lab-run

jerrinot:
@eminn: could you write a test? I think you could create a map, do some operations, then destroy&re-create the map again -> the stats should be empty?

eminn:
run-lab-run

ahmetmircik:
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/326/
```
Test Result (1 failure / ±0)
com.hazelcast.client.spi.impl.ClientInvocationTest.executionCallback_FailOnShutdown
```

ahmetmircik:
run-lab-run

hazelcast/src/test/java/com/hazelcast/map/LocalMapStatsTest.java
```diff
@@ -29,6 +34,25 @@
 public class LocalMapStatsTest extends HazelcastTestSupport {
 
     @Test
+    public void testMapStatsAreNotLeaking() throws Exception {
```
jerrinot: it's something like this sufficient? it makes less assumption about internals. 
```java
    @Test
    public void givenMapContainsStatistics_whenDestroyedAndCreatedAgain_thenStatisticsAreReset()  {
        HazelcastInstance instance = createHazelcastInstance();
        IMap<Integer, Integer> map = instance.getMap(mapName);
        map.put(1, 1);

        map.destroy();
        map = instance.getMap(mapName);

        long putOperationCount = map.getLocalMapStats().getPutOperationCount();
        assertEquals(0, putOperationCount);
    }
```

eminn: It doesn't imply that there no leakage of statistics objects. It could happen that statistics are reset when map is destroyed and the test will pass even they are leaking.

To me both ways is ugly/hacky.

eminn: Afaik, when getLocalMapStats() is called, statistics object is re-created so it won't test the actual scenario.

jerrinot: this test as it is is failing on the current master -> it shows the stats. data are retained even when the map is destroyed -> leak. 

my approach is not bullet-proof and I understand your concert about a hidden leak. yours is more fragile as it makes more assumptions about implementation. It's hard to tell what's better -> :+1: from me and the decision what approach is better on you.

jerrinot:
:+1:

Issue #6966, opened 2015-12-02
SingleNodeTest.executorServiceStats


jerrinot:
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/297/

Jenkins is down right now so I can't paste more, however I can reproduce it locally with `Repeat(100)`

hasancelik:
```
10:47:48 executorServiceStats(com.hazelcast.executor.SingleNodeTest)  Time elapsed: 120.241 sec  <<< FAILURE!
10:47:48 java.lang.AssertionError: expected:<1> but was:<0>
10:47:48 	at org.junit.Assert.fail(Assert.java:88)
10:47:48 	at org.junit.Assert.failNotEquals(Assert.java:834)
10:47:48 	at org.junit.Assert.assertEquals(Assert.java:645)
10:47:48 	at org.junit.Assert.assertEquals(Assert.java:631)
10:47:48 	at com.hazelcast.executor.SingleNodeTest$5.run(SingleNodeTest.java:359)
10:47:48 	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:835)
10:47:48 	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:849)
10:47:48 	at com.hazelcast.executor.SingleNodeTest.executorServiceStats(SingleNodeTest.java:351)
```

gurbuzali:
should be fixed via https://github.com/hazelcast/hazelcast/pull/6975

PR #6975, opened 2015-12-03
added isCanceled check before sending response

### LINKED ISSUE #6969: ClientExecutorServiceTest.testGetFutureAfterCancel
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast-client/17278/console

```
Tests run: 15, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 23.423 sec <<< FAILURE! - in com.hazelcast.client.executor.ClientExecutorServiceTest
testGetFutureAfterCancel(com.hazelcast.client.executor.ClientExecutorServiceTest)  Time elapsed: 21.842 sec  <<< FAILURE!
java.lang.AssertionError: Expected exception: java.util.concurrent.CancellationException
	at org.junit.internal.runners.statements.ExpectException.evaluate(ExpectException.java:32)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:132)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at com.hazelcast.test.HazelcastParallelClassRunner.access$301(HazelcastParallelClassRunner.java:46)
	at com.hazelcast.test.HazelcastParallelClassRunner$TestRunner.run(HazelcastParallelClassRunner.java:139)
	at java.lang.Thread.run(Thread.java:662)
```

gurbuzali:
Reasoning behind this change:
When a task is canceled, it tries to send the response (which maybe null) without checking `isCanceled`
If the task is cancelled (`task.cancel()` returns true) the canceller part should be responsible sending the response (`CancellationException`)

fixes #6969 and #6966

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
+++ b/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
@@ -211,8 +211,8 @@ public class DistributedExecutorService implements ManagedService, RemoteService
                 if (uuid != null) {
                     submittedTasks.remove(uuid);
                 }
-                sendResponse(result);
                 if (!isCancelled()) {
+                    sendResponse(result);
                     finishExecution(name, Clock.currentTimeMillis() - start);
                 }
             }


metanet:
Isn't sendResponse sending the response only once by doing a CAS? A task is cancelled only if its result has been sent to the original invocation. I didn't apply this change intentionally since I though that it is not necessary.

sancar:
:+1: 
@metanet  When task is cancelled, there is a race. When we don't do the check, there is a possibility that, the thread that successfully canceled the operation returning `false`(Not able to cancel). And the cancelled operation returning as if it is not cancelled.

gurbuzali:
I think if this line is returning true, it should send the `CancellationException`
`processor != null && processor.cancel(interrupt)`

metanet:
I dont think that this fix solves the case @sancar mentioned, completely. A task can be cancelled just before run method tries to send the actual response within the "if not cancelled" block. I am not sure if this is a race. I think this is a concurrent operation that touches multiple variables.

gurbuzali:
@metanet a task cannot be cancelled just before sending the response because `get` is called after `super.run()` which moves the state of the task from `NEW`. A task can only be cancelled only if it is `NEW`

metanet:
@gurbuzali ok, now I got it. I didn't know that. Thanks 

:+1:

juanavelez:
This issue had been raised a while ago #6283

PR #6981, opened 2015-12-03
Adds cluster state list and change REST APIS.

kobalski:
Also wrote a newclusterState.sh script which uses this API. Scripts must run on a machine at least one hazelcast member is running.

Sample usage : 

For getting cluster state : 
`./clusterState.sh -- operation get -port 5701 --groupname dev --password devpass`

For changing cluster state : 
`./clusterState.sh --operation change --state passive --port 5701 --groupname dev --password devpass`

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpCommandProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpCommandProcessor.java
@@ -23,7 +23,10 @@ public abstract class HttpCommandProcessor<T> extends AbstractTextCommandProcess
     public static final String URI_MAPS = "/hazelcast/rest/maps/";
     public static final String URI_QUEUES = "/hazelcast/rest/queues/";
     public static final String URI_CLUSTER = "/hazelcast/rest/cluster";
-    public static final String URI_SHUTDOWN_CLUSTER_URL = "/hazelcast/rest/management/cluster/shutdown";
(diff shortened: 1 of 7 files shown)


mesutcelik:
I would suggest to rename `clusterState.sh` to `cluster.sh` and add shutdown functionality into this script instead of stop.sh. That helps us only maintain one briefly complex script instead of 2

mdogan:
@mesutcelik: :+1:

kobalski:
Good Idea. `stop.sh` only responsible for shutdown of member which is started with `server.sh`.
`cluster.sh` will be responsible for cluster operations. Valid operations are `get`, `change`, `shutdown`.

Update sample usage : 

For getting cluster state : 

`./cluster.sh  --operation get -port 5701 --groupname dev --password dev-pass`
OR 
`./cluster.sh  -o get -p 5701 -g dev -P dev-pass`

For changing cluster state : 

`./cluster.sh  --operation change --state passive -port 5701 --groupname dev --password dev-pass`
OR 
`./cluster.sh  -o get -s passive -p 5701 -g dev -P dev-pass`

For cluster shutdown : 

`./cluster.sh  --operation shutdown -port 5701 --groupname dev --password dev-pass`
OR 
`./cluster.sh  -o shutdown -p 5701 -g dev -P dev-pass`


Defaults for parameters will be : 
--operation : `get`. Script will show cluster state by default
--port : `5701`. Script will try to call api on port `5701` by default
--groupname : `dev`. Script will use default hazelcast group-name by default
--password : `dev-pass`. Script will use default hazelcast group password by default.

mesutcelik:
:+1:

emrahkocaman:
:+1:

PR #6982, opened 2015-12-03
New option on map that allows to configure if wants a defensive copy of object contained

angel9484:
This enhacement is really cool with paths where the speed is critical over data integrity.
This modification is only for map.
It add's a new option on mapConfig: 
The option let choose if a map can retrieve the object without serialize and deserialize (a defensive copy).
The RecordStore has been enhaced and the readBackupData has been refactored: now exists a new method readBackup that contais all the code in the old readBackupData except the "toData method call". The readBackupData now calls to readBackup and then call to the toData method on mapServiceContext.
Then the MapProxySupport does this:
if true, hazelcast will work as ever.
If false, and in memory type is not Object, hazelcast will work as ever.
If false and in memory type is Object and isReadBackupData is active: it will use the readBackup and retrieve the object as is. With 35KB serialized value, and a bit of bussiness logic, the response time was around 200ms. Now without this defensive copy, the response time is around 30ms

### DIFF
--- a/hazelcast-client-new/src/test/java/com/hazelcast/client/lock/ClientConditionTest.java
+++ b/hazelcast-client-new/src/test/java/com/hazelcast/client/lock/ClientConditionTest.java
@@ -23,7 +23,7 @@ import static org.junit.Assert.assertTrue;
 
 @RunWith(HazelcastParallelClassRunner.class)
 @Category(QuickTest.class)
-public class ClientConditionTest extends HazelcastTestSupport{
+public class ClientConditionTest extends HazelcastTestSupport {
 
     private final TestHazelcastFactory hazelcastFactory = new TestHazelcastFactory();
     private HazelcastInstance client;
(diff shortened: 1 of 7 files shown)


pveentjer:
Hi @angel9484, thanks you for your PR.

We are currently at feature freeze and this feature needs to wait till 3.6 has been released.

@ahmetmircik @asimarslan can you have a look at this PR?

angel9484:
Aww! Ok! Sorry for the inconvenience! I will try to do this PR on the 3.6-snapshot branch!
Actually we can't work with non stable releases. Do you have an ETA for 3.6?.
Do you have the style code format in some place? if you have it to eclipse or intellij it would be great!

Thanks!

Donnerbart:
To get this eventually merged you have to base it on the master branch, which is the base for the upcoming releases. We don't merge new features into other branches than master (and of course not during feature freeze).

For the code style we have CheckStyle which you can run via Maven, e.g. by executing `mvn clean compile -Pcheckstyle`. That should cover most style issues. We don't have an up to date code formatter, but the out-of-the-box formatter of InetlliJ 15 works quite well for Hazelcast.

angel9484:
Ok @Donnerbart Thanks! now is the new PR on master: #7013 
In spain is holidays today and tomorrow, I will sign the contribution agreement on wednesday utc zone.
Regards!

PR #6984, opened 2015-12-03
Atomicity supported between cache record store and `CacheWriter` while updating existing entry

serkan-ozal:
Currently, writing to `CacheWriter` is done before record is updated in cache record store. But in case of failure while updating in record store (such as native OOME at EE while doing native memory allocation for new value), record is written to `CacheWriter` more than once while trying to putting to record store.

With this PR, writing to `CacheWriter` is done after new value is allocated but not just set inside existing record as new value. So if writing to `CacheWriter` fails, since we have not updated record value yet with its new value, no need to any action to revert it.

Fixes https://github.com/hazelcast/hazelcast-enterprise/issues/537
with #6935

**P.S 1:** No need to additional tests for Client and EE modules because client (`Client CacheReadWriteThroughTest `) and EE (`HiDensityCacheReadWriteThroughTest `) counterparts already extend from this test `CacheReadWriteThroughTest` so added tests to `CacheReadWriteThroughTest` are applied to them also by inheritance.

**P.S 2:** `HiDensityNativeMemoryCacheRecordStore::onCreateRecordWithExpiryError` should be updated as `HiDensityNativeMemoryCacheRecordStore::onCreateRecordError` after this PR merged because `HiDensityNativeMemoryCacheRecordStore` overrides this method from `AbstractCacheRecordStore` and this method name is updated to `onCreateRecordError` from `onCreateRecordWithExpiryError` with this PR.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
@@ -468,53 +468,63 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
     }
 
     protected R createRecord(Data keyData, Object value, long expirationTime, int completionId) {
-        return createRecord(keyData, value, expirationTime, completionId, null);
-    }
-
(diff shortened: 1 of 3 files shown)


ahmetmircik:
seems mergeable to me :+1:

PR #6989, opened 2015-12-04
Add more logging to TransactionManagerService

mdogan:
Added detailed logs while committing/rollingback replicated transactions when transaction manager dies.


gurbuzali:
merging this one since it is just adding finest logging

PR #6990, opened 2015-12-04
unit tests for predicate-related stuff added

jerrinot:


hazelcast/src/test/java/com/hazelcast/query/impl/predicates/AbstractVisitorTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import static org.junit.Assert.assertSame;
+import static org.mockito.Mockito.mock;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class AbstractVisitorTest {
+
+    @Test
+    public void testAllVisitMethodReturnTheOriginalPredicate() throws IllegalAccessException, InstantiationException, InvocationTargetException {
```
jerrinot: this can be quite fragile. I still think it's worth giving it a try.

Donnerbart:
:+1: (once green)

It's just new tests, if they pass I think we can merge them directly.

PR #6992, opened 2015-12-04
Cache putAllBackup should skip tombstone records

mdogan:
Tombstone records are created when an entry is removed
or during eviction/expiration when hot-restart is enabled for the cache.

Fixes hazelcast/hazelcast-enterprise#560


serkan-ozal:
:+1:

PR #6993, opened 2015-12-04
Update `primary` field of `AbstractCacheRecordStore` after migration through `ICacheRecordStore::init()`

### LINKED ISSUE #6983: `AbstractCacheRecordStore` should update `isOwner` field while it is being cleared after migration
After migration, `ICacheService` clears `ICacheRecordStore` instances related with migrated partitions.

`AbstractCacheRecordStore` has a boolean field named `isOwner` and this flag represents that partition, has this record store, is primary or backup. If it is backup, events and invalidation messages are ignored. 

So far so good, but while record store is being cleared after migration, this field should be updated by checking that current node is the owner of the partition or not anymore. 

A new `reset` method can be added `ICacheRecordStore` and this is called by `ICacheService` instead of `clear` so this reset method can be implemented inside `AbstractCacheRecordStore` by calling `clear` first and then updating `isOwner` field.

serkan-ozal:
Fixes #6983

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
@@ -33,7 +33,9 @@ import com.hazelcast.internal.eviction.EvictionPolicyEvaluatorProvider;
 import com.hazelcast.internal.eviction.EvictionStrategy;
 import com.hazelcast.internal.eviction.EvictionStrategyProvider;
 import com.hazelcast.map.impl.MapEntries;
+import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.partition.InternalPartition;
 import com.hazelcast.partition.InternalPartitionService;
 import com.hazelcast.spi.EventRegistration;
 import com.hazelcast.spi.NodeEngine;
@@ -80,7 +82,6 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
     protected final int partitionId;
     protected final int partitionCount;
     protected final NodeEngine nodeEngine;
-    protected final InternalPartitionService partitionService;
     protected final AbstractCacheService cacheService;
     protected final CacheConfig cacheConfig;
     protected CRM records;
@@ -98,17 +99,15 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
     protected final EvictionChecker evictionChecker;
     protected final EvictionStrategy<Data, R, CRM> evictionStrategy;
     protected final boolean wanReplicationEnabled;
-    protected final boolean isOwner;
+    protected boolean primary;
 
     //CHECKSTYLE:OFF
     public AbstractCacheRecordStore(final String name, final int partitionId, final NodeEngine nodeEngine,
                                     final AbstractCacheService cacheService) {
         this.name = name;
         this.partitionId = partitionId;
-        this.isOwner = nodeEngine.getPartitionService().isPartitionOwner(partitionId);
         this.nodeEngine = nodeEngine;
-        this.partitionService = nodeEngine.getPartitionService();
-        this.partitionCount = partitionService.getPartitionCount();
+        this.partitionCount = nodeEngine.getPartitionService().getPartitionCount();
         this.cacheService = cacheService;
         this.cacheConfig = cacheService.getCacheConfig(name);
         this.cacheContext = cacheService.getOrCreateCacheContext(name);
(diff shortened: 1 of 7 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
```diff
@@ -157,12 +160,19 @@ public void commitMigration(PartitionMigrationEvent event) {
     public void rollbackMigration(PartitionMigrationEvent event) {
         if (event.getMigrationEndpoint() == MigrationEndpoint.DESTINATION) {
             clearPartitionReplica(event.getPartitionId());
+        } else {
+            // MigrationEndpoint.DESTINATION
```
mdogan: MigrationEndpoint.DESTINATION? or MigrationEndpoint.SOURCE?

serkan-ozal: ah yes, comment is wrong. it will be `MigrationEndpoint.SOURCE`

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
```diff
@@ -157,12 +160,19 @@ public void commitMigration(PartitionMigrationEvent event) {
...
         }
     }
 
     @Override
     public void clearPartitionReplica(int partitionId) {
-        segments[partitionId].clear();
+        resetPartitionReplica(partitionId, true);
+    }
+
+    private void resetPartitionReplica(int partitionId, boolean clear) {
```
mdogan: Only thing that bothers me is, names of `clearPartitionReplica` and `resetPartitionReplica` methods are sounds similar. It may be confusing.

mdogan:
other than my minor comment, :+1:

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
```diff
@@ -1434,6 +1434,14 @@ protected void onClear() {
     }
 
     @Override
+    public void reset(boolean clear) {
+        if (clear) {
+            clear();
+        }
+        isOwner = nodeEngine.getPartitionService().isPartitionOwner(partitionId);
```
ahmetmircik: Do we need to trigger partition owner assignment here? 

```
 @Override
    public boolean isPartitionOwner(int partitionId) {
        InternalPartitionImpl partition = getPartition(partitionId);
        return node.getThisAddress().equals(partition.getOwnerOrNull());
    }
@Override
    public InternalPartitionImpl getPartition(int partitionId) {
        return getPartition(partitionId, true);
    }

    @Override
    public InternalPartitionImpl getPartition(int partitionId, boolean triggerOwnerAssignment) {
        InternalPartitionImpl p = getPartitionImpl(partitionId);
        if (triggerOwnerAssignment && p.getOwnerOrNull() == null) {
            // probably ownerships are not set yet.
            // force it.
            getPartitionOwner(partitionId);
        }
        return p;
    }
```

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
ahmetmircik: +1

hazelcast/src/test/java/com/hazelcast/cache/impl/InternalCacheRecordStoreTest.java
```diff
@@ -64,17 +88,114 @@ public void testBatchEventMapShouldBeCleanedAfterRemoveAll() {
         assertEquals(0, recordStore.batchEvent.size());
     }
 
-    @Override
-    protected void onSetup() {
+    /**
+     * Test for issue: https://github.com/hazelcast/hazelcast/issues/6983
+     */
+    @Test
+    public void ownerStateShouldBeUpdatedAfterMigration() throws ExecutionException, InterruptedException {
```
ahmetmircik: if it seems possible to you, a shorter test makes much sense, this seems a hard one to understand to me.

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
serkan-ozal: will do

hazelcast/src/test/java/com/hazelcast/cache/impl/InternalCacheRecordStoreTest.java
serkan-ozal: ok, i will try make it shorter and split it into sub-methods

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
serkan-ozal: I can remove `reset` (which contains clearing and refreshing cache record store) so cache service will call `clear` and `init` (or `refresh` or any other name. do you have any suggesstion for better name) sequentially after migration instead of a single `reset` call. 

@mdogan @ahmetmircik WDYT?

ahmetmircik: calling clear init makes sense to me

PR #7009, opened 2015-12-07
added metainf/services tests

bilalyasar:


hazelcast-client-legacy/src/test/java/com/hazelcast/client/ClientServiceTest.java
```diff
@@ -144,7 +148,23 @@ public void testNumberOfClients_afterUnAuthenticatedClient_withTwoNode() {
         assertEquals(0, instance2.getClientService().getConnectedClients().size());
     }
 
-
+    @Test
+    public void loadServiceHooksClassNames() throws IOException {
```
mesutcelik: i see this method all over test classes... why not adding to HazelcastTestSupport or into a util class?

pveentjer: +1

if you are doing copy paste, something is going wrong. Copy paste is violation or our architectural principles.

PR #7010, opened 2015-12-07
Removed unused QueryResultSet

Donnerbart:
This is a left-over class which has 0% code coverage and blocks the Reduce Index Memory Consumption feature from code coverage verification.


jerrinot:
:+1: 
I wanted to remove it myself, thanks!

Donnerbart:
I think it's safe to merge this, since all checks passed.

PR #7012, opened 2015-12-07
Server should respond to authentication message when node is not active

sancar:
This seems to be a regression after the message decoding was moved
 to partition thread. When the node is not active, the error message
is not sent back to the client

Backport of : https://github.com/hazelcast/hazelcast/pull/6926


sancar:
Merging this since this change reviewed and merge here
https://github.com/hazelcast/hazelcast/pull/6926

PR #7013, opened 2015-12-07
[3.7] New option on map that allows to configure if wants a defensive copy of object contained

angel9484:
from PR #6982 
This enhacement is really cool with paths where the speed is critical over data integrity.
This modification is only for map.
It add's a new option on mapConfig: 
The option let choose if a map can retrieve the object without serialize and deserialize (a defensive copy).
The RecordStore has been enhaced and the readBackupData has been refactored: now exists a new method readBackup that contais all the code in the old readBackupData except the "toData method call". The readBackupData now calls to readBackup and then call to the toData method on mapServiceContext.
Then the MapProxySupport does this:
if true, hazelcast will work as ever.
If false, and in memory type is not Object, hazelcast will work as ever.
If false and in memory type is Object and isReadBackupData is active: it will use the readBackup and retrieve the object as is. With 35KB serialized value, and a bit of bussiness logic, the response time was around 200ms. Now without this defensive copy, the response time is around 30ms


pveentjer:
Hi @angel9484 thank you for your contribution. We are currently in feature freeze since 3.6 is going to be released any day. So we'll need to park this for 3.7.

angel9484:
Hi @pveentjer! Ok! I saw your calendar and I think we can update later :-D
Thanks for the 1194 issue reference, I will study it, and help you if possible.

Regards!

angel9484:
Hello! Is there any possibility to review / Integrate this PR?
Thanks!

jerrinot:
@angel9484: Hi, thanks for you contribution. Can you please elaborate the way your benchmarked it? you wrote "With 35KB serialized value, and a bit of bussiness logic, the response time was around 200ms. Now without this defensive copy, the response time is around 30ms".  

What exactly is a response here time? I suppose it's your end-to-end response of you business application. Is this correct? Can you describe your testing scenario a bit more? Was is with just a single node? The difference in performance seems to be massive. 

I'm undecided here as this has an impact only when reading from a local member. In all other cases it has no effect. I'm not sure if this is worth the (configuration) complexity and decreased safety guarantees.

jerrinot:
I'm closing this PR now. Please feel free to re-open it when you have additional data about your benchmarks.

gAmUssA:
@angel9484 could you provide more data about your benchmark?

Issue #6970, opened 2015-12-03
ClientMapLoadAllTest testLoadAll_givenKeys


jerrinot:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast-client/17278/consoleText

```
Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 120.593 sec <<< FAILURE! - in com.hazelcast.client.map.ClientMapLoadAllTest
testLoadAll_givenKeys(com.hazelcast.client.map.ClientMapLoadAllTest)  Time elapsed: 120.593 sec  <<< ERROR!
com.hazelcast.core.OperationTimeoutException: null
	at com.hazelcast.client.impl.protocol.ClientExceptionFactory$40.createException(ClientExceptionFactory.java:332)
	at com.hazelcast.client.impl.protocol.ClientExceptionFactory.createException(ClientExceptionFactory.java:553)
	at com.hazelcast.client.impl.protocol.ClientExceptionFactory.createException(ClientExceptionFactory.java:539)
	at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport$ResponseThread.handleClientMessage(ClientInvocationServiceSupport.java:333)
	at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport$ResponseThread.process(ClientInvocationServiceSupport.java:313)
	at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport$ResponseThread.doRun(ClientInvocationServiceSupport.java:306)
	at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport$ResponseThread.run(ClientInvocationServiceSupport.java:283)
	at ------ End remote and begin local stack-trace ------.(Unknown Source)
	at com.hazelcast.client.spi.impl.ClientInvocationFuture.resolveResponse(ClientInvocationFuture.java:134)
	at com.hazelcast.client.spi.impl.ClientInvocationFuture.get(ClientInvocationFuture.java:101)
	at com.hazelcast.client.spi.impl.ClientInvocationFuture.get(ClientInvocationFuture.java:76)
	at com.hazelcast.client.spi.impl.ClientInvocationFuture.get(ClientInvocationFuture.java:39)
	at com.hazelcast.client.spi.ClientProxy.invoke(ClientProxy.java:144)
[...]

PR #7015, opened 2015-12-07
Send correct exceptions to client

sancar:
Before exceptions send to client, they will get through a similar
method like ExceptionUtil.rethrow so that clients can get real
exception.

related to https://github.com/hazelcast/hazelcast/issues/6970

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
@@ -48,23 +48,39 @@ public final class ExceptionUtil {
         return sw.toString();
     }
 
+    public static RuntimeException peel(final Throwable t) {
+        return (RuntimeException) peel(t, null);
+    }
+
+    public static <T extends Throwable> Throwable peel(final Throwable t, Class<T> allowedType) {
+        if (t instanceof RuntimeException) {
+            return t;
+        }
+
+        if (t instanceof ExecutionException) {
+            final Throwable cause = t.getCause();
+            if (cause != null) {
+                return peel(cause, allowedType);
+            } else {
+                return new HazelcastException(t);
+            }
+        }
+
+        if (allowedType != null && allowedType.isAssignableFrom(t.getClass())) {
+            return t;
+        }
+
+        return new HazelcastException(t);
+    }
+
     public static RuntimeException rethrow(final Throwable t) {
         if (t instanceof Error) {
             if (t instanceof OutOfMemoryError) {
                 OutOfMemoryErrorDispatcher.onOutOfMemory((OutOfMemoryError) t);
             }
             throw (Error) t;
-        } else if (t instanceof RuntimeException) {
-            throw (RuntimeException) t;
-        } else if (t instanceof ExecutionException) {
-            final Throwable cause = t.getCause();
-            if (cause != null) {
-                throw rethrow(cause);
-            } else {
-                throw new HazelcastException(t);
-            }
         } else {
-            throw new HazelcastException(t);
+            throw peel(t);
         }
     }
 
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
```diff
@@ -48,23 +48,35 @@ public static String toString(Throwable cause) {
...
     }
 
+    public static RuntimeException peel(final Throwable t) {
+        return (RuntimeException) peel(t, null);
+    }
+
+    public static <T extends Throwable> Throwable peel(final Throwable t, Class<T> allowedType) {
+        if (t instanceof RuntimeException) {
+            return (RuntimeException) t;
+        } else if (t instanceof ExecutionException) {
```
gurbuzali: `else if`s are redundant

gurbuzali:
minor comment :+1:

hasancelik:
run-lab-run

hasancelik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
```diff
@@ -48,23 +48,39 @@ public static String toString(Throwable cause) {
         return sw.toString();
     }
 
+    public static RuntimeException peel(final Throwable t) {
+        return (RuntimeException) peel(t, null);
+    }
+
+    public static <T extends Throwable> Throwable peel(final Throwable t, Class<T> allowedType) {
```
pveentjer: why are some parameters final and others not?

pveentjer:
There are not tests + javadoc for the peel functionality. Please add them.

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
```diff
@@ -48,23 +48,39 @@ public static String toString(Throwable cause) {
...
+        if (t instanceof RuntimeException) {
+            return t;
+        }
+
+        if (t instanceof ExecutionException) {
+            final Throwable cause = t.getCause();
+            if (cause != null) {
+                return peel(cause, allowedType);
+            } else {
+                return new HazelcastException(t);
```
pveentjer: This could be problematic since the InterruptedException might be wrapped inside a HazelcastException. 

@mtopolnik WDYT?

mtopolnik: Yes, one of many such examples in our codebase.

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
```diff
@@ -48,23 +48,39 @@ public static String toString(Throwable cause) {
...
+            }
+        }
+
+        if (allowedType != null && allowedType.isAssignableFrom(t.getClass())) {
+            return t;
+        }
+
+        return new HazelcastException(t);
```
pveentjer: Same here.

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
pveentjer: @sancar please fix these bugs and add tests. Also add tests for dealing with interrupt status.

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
mtopolnik: However, perhaps this is not the place where to fix this: the method is already trapped into returning an unchecked exception. We'll have to make `InterruptedException` a special case earlier on and we'll probably need to separately cover the cases where an `InterruptedException` may have occurred and where not.

pveentjer: SHould we leave it as it is?

mtopolnik: I think the whole design needs a careful review. It is OK to have a method which unchecks an exception; what is not OK is to use this method indiscriminately. We'll probably need one more utility method which will `throws InterruptedException` and uncheck everything else.

PR #7026, opened 2015-12-08
Fixed inconsistency in MapListeners null behaviours

### LINKED ISSUE #7001: Inconsistent behavior of IMap.addEntryListener
Calling `IMap.addEntryListener` with null values results in behavior inconsistent with API as well as different behavior between members and clients and between Hazelcast versions 3.5.4 and 3.6-EA2.

Below is the code I used to test this. The full testing was done on Windows 8.1 with Java 8, but similar behavior was observed on Windows 7 and CentOS 6.6, both running Java 8
```java
HazelcastInstance instance = [new member or client instance]
IMap<String, String> map = instance.getMap("test_map");
MapListener listener = null;
Predicate<String, String> predicate = null;
String key = null;
// one of the following
map.addEntryListner(listener, false);
map.addEntryListner(listener, key, false);
map.addEntryListner(listener, predicate, false);
map.addEntryListner(listener, predicate, key, false);
```
Below are inconsistencies that I've found between the documentation and behavior on members on either version 3.5.4 or 3.6-EA2.
* The API states that `IMap.addEntryListener(MapListener, K, boolean)` will throw a `NullPointerException` if the key is null, but no exception is thrown when `key` is null (and other arguments are not null).
* No other `addEntryListener` methods document `NullPointerException`s that may be thrown. However, if `listener` or `predicate` are null, then a `NullPointerException` will be thrown.

Below are behaviors of clients on version 3.5.4 which differ from members.
* Using a null `listener` will result in an `IllegalArgumentException`.
* Using either a null `listener` or `key` results in no exception being thrown.

Below are behaviors of clients on version 3.6-EA2 which differ from members.
* Using a null `listener` will result in an `IllegalArgumentException`.
* Using either a null `listener` or `key` results in a `HazelcastException` being thrown.

muratayan:
fixes #7001 

We need to decide to backport and/or client-legacy change as well.

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -138,6 +138,8 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
 
     protected static final String NULL_KEY_IS_NOT_ALLOWED = "Null key is not allowed!";
     protected static final String NULL_VALUE_IS_NOT_ALLOWED = "Null value is not allowed!";
+    protected static final String NULL_LISTENER_IS_NOT_ALLOWED = "Null listener is not allowed!";
+    protected static final String NULL_PREDICATE_IS_NOT_ALLOWED = "Predicate should not be null!";
 
     protected final String name;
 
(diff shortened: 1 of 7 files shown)


gurbuzali:
shouldn't we fix the issue for client-legacy too ?
can you add tests so that we don't break this in the future

muratayan:
@gurbuzali I've fixed legacy as well. I will go over the other methods in `map proxies` too. But I don't know how can we add tests for javadoc consistency actually.

gurbuzali:
@muratayan you can create a test which just calls those methods with null values and expect the specified exception mentioned in the javadoc

muratayan:
Updated

ttaomae:
Is there a reason why the `add[local]EntryListener(listener, predicate, key, includeValue)` methods do not check for a null key even though the `addEntryListener(listener, key, includeValue)` methods do check?

gurbuzali:
@ttaomae I think we should

muratayan:
@ttaomae @gurbuzali This is for not breaking the compatibility. There are some code out there using `add[local]EntryListener(listener, predicate, key, includeValue)` with `null` key instead of `add[local]EntryListener(listener, predicate, includeValue)`. So if a `null` key is passed in the first it's same functionality as the second one. We can create an issue for next major release.

ahmetmircik:
@muratayan's reasoning makes sense to me. 

LGTM :+1:

gurbuzali:
:+1:

muratayan:
run-lab-run

PR #7029, opened 2015-12-08
Additional unit tests for predicates-related code

tombujok:


Donnerbart:
:+1: (just new tests, I think it can be merged right away once green)

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/PredicatesTest.java
```diff
@@ -252,6 +253,12 @@ public void testILike() {
     }
 
     @Test
+    public void testILike_Id() {
+        ILikePredicate predicate = (ILikePredicate) ilike(ATTRIBUTE, "Java_World");
+        assertEquals(PredicateDataSerializerHook.ILIKE_PREDICATE, predicate.getId());
```
jerrinot: wouldn't it make more sense for test to use the actual ID instead of linking the constant? so if you change the constant by accident then the test fails.

tombujok: hmm, I know what you mean, but referencing 6 would be a magic number that nobody would understand. This test is not very clever anyway...

tombujok: OK, I added a double assertion, so it should be super safe now ;)

jerrinot:
:+1:

PR #7030, opened 2015-12-08
Code coverage and fix for client specific MemberImpl

Donnerbart:
* Increased code coverage of client specific `MemberImpl`
* Fixed bug in `getLongAttribute()`

The client class was accidentically covered by the QA whitelisting. Actually the `MemberImpl.getLongAttribute()` was using a wrong method to retrieve the Long value, which was not detected due to missing code coverage.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/MemberImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/MemberImpl.java
@@ -27,12 +27,10 @@ import java.util.Map;
 /**
  * Client side specific Member implementation.
  *
- * <p>Caution: This class is required by protocol encoder/decoder which are on hazelcast module
- * so this impl also stays in the same module, although it is totally client side specific</p>
+ * <p>Caution: This class is required by protocol encoder/decoder which are on the Hazelcast module.
+ * So this implementation also stays in the same module, although it is totally client side specific.</p>
  */
-public final class MemberImpl
-        extends AbstractMember
-        implements Member {
+public final class MemberImpl extends AbstractMember implements Member {
 
     public MemberImpl() {
     }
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/MemberImpl.java
```diff
@@ -70,94 +68,93 @@ public String getStringAttribute(String key) {
...
     @Override
     public void setIntAttribute(String key, int value) {
-        notSupportedOnClient();
+        throw notSupportedOnClient();
     }
 
     @Override
     public Long getLongAttribute(String key) {
-        final Object attribute = getAttribute(key);
-        return attribute != null ? Long.getLong(attribute.toString()) : null;
```
Donnerbart: `Long.getLong()` tries to read from System properties, not from the given parameter.

serkan-ozal: :+1:

serkan-ozal:
:+1:

hazelcast/src/test/java/com/hazelcast/client/impl/MemberImplTest.java
```diff
@@ -0,0 +1,196 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class MemberImplTest extends HazelcastTestSupport {
+
+    private static Address address;
+
+    private MemberImpl member;
```
pveentjer: I'm always very careful putting in such a field in my test. For example the method 'assertMemberImpl' makes no sense until you realize that it is about the field member.

hazelcast/src/test/java/com/hazelcast/client/impl/MemberImplTest.java
```diff
@@ -0,0 +1,196 @@
...
+        member.removeAttribute("remoteMemberRemove");
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void testSetAttribute_onRemoteMember() {
+        member = new MemberImpl(address);
+        member.setStringAttribute("remoteMemberSet", "wontWork");
+    }
+
+    private void assertMemberImpl() {
```
pveentjer: This method is a bit strangely named. Initially I thought it checks if it is a memberImpl, but you are checking all kinds of properties.  So its name is not self describing.

Donnerbart: What's your suggestion?

(your comment got outdated, since I just removed two useless test methods which were copied from the other `MemberImplTest` class).

hazelcast/src/test/java/com/hazelcast/client/impl/MemberImplTest.java
mtopolnik: instance fields make sense when they are initialized by `@Setup`, `@Parameters` or similar. But this one seems to be just used as a local variable. So the only point is saving one keyword per test method?

Donnerbart:
Can we get this merged today, please?

mtopolnik:
:+1:

PR #7031, opened 2015-12-08
If cookie-secure is false, mark the Hazelcast session cookie as "Secure" if the request that initiated the corresponding session used HTTPS. 

cszmajda:
This conforms to the Java EE Servlet 3.0 API http://docs.oracle.com/javaee/7/api/javax/servlet/SessionCookieConfig.html#isSecure--.


emrahkocaman:
Thanks @cszmajda , looks good to me :+1:

jerrinot:
:+1: 

@cszmajda: Thanks for the patch!

PR #7032, opened 2015-12-09
Fix hibernate natural id caching in local mode

cszmajda:
Previously, `HazelcastNaturalIdRegion` was hard coded to create an `IMapRegionCache` even when being created from `HazelcastLocalCacheRegionFactory`.  This change brings natural id regions in line with entity, collection, and timestamp regions.

This change is required only in the `hibernate4` tree, as `hibernate3` does not have natural id caching.

### DIFF
--- a/hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
+++ b/hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
@@ -20,13 +20,10 @@ import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.hibernate.distributed.IMapRegionCache;
 import com.hazelcast.hibernate.region.HazelcastCollectionRegion;
 import com.hazelcast.hibernate.region.HazelcastEntityRegion;
+import com.hazelcast.hibernate.region.HazelcastNaturalIdRegion;
 import com.hazelcast.hibernate.region.HazelcastTimestampsRegion;
 import org.hibernate.cache.CacheException;
-import org.hibernate.cache.spi.RegionFactory;
-import org.hibernate.cache.spi.CacheDataDescription;
-import org.hibernate.cache.spi.CollectionRegion;
-import org.hibernate.cache.spi.EntityRegion;
-import org.hibernate.cache.spi.TimestampsRegion;
+import org.hibernate.cache.spi.*;
 
 import java.util.Properties;
 
@@ -56,6 +53,13 @@ public class HazelcastCacheRegionFactory extends AbstractHazelcastCacheRegionFac
                                           final CacheDataDescription metadata) throws CacheException {
         return new HazelcastEntityRegion<IMapRegionCache>(instance, regionName, properties, metadata,
                 new IMapRegionCache(regionName, instance, properties, metadata));
+    }
+
+    public NaturalIdRegion buildNaturalIdRegion(final String regionName, final Properties properties,
+                                                final CacheDataDescription metadata)
+            throws CacheException {
+        return new HazelcastNaturalIdRegion<IMapRegionCache>(instance, regionName, properties, metadata,
+                new IMapRegionCache(regionName, instance, properties, null));
     }
 
     public TimestampsRegion buildTimestampsRegion(final String regionName, final Properties properties)
--- a/hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
+++ b/hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
@@ -23,7 +23,12 @@ import com.hazelcast.hibernate.region.HazelcastEntityRegion;
 import com.hazelcast.hibernate.region.HazelcastNaturalIdRegion;
 import com.hazelcast.hibernate.region.HazelcastTimestampsRegion;
 import org.hibernate.cache.CacheException;
-import org.hibernate.cache.spi.*;
+import org.hibernate.cache.spi.RegionFactory;
+import org.hibernate.cache.spi.CacheDataDescription;
+import org.hibernate.cache.spi.CollectionRegion;
+import org.hibernate.cache.spi.EntityRegion;
+import org.hibernate.cache.spi.NaturalIdRegion;
+import org.hibernate.cache.spi.TimestampsRegion;
 
 import java.util.Properties;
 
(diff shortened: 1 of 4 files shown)


hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
```diff
@@ -20,13 +20,10 @@
...
 import com.hazelcast.hibernate.region.HazelcastEntityRegion;
+import com.hazelcast.hibernate.region.HazelcastNaturalIdRegion;
 import com.hazelcast.hibernate.region.HazelcastTimestampsRegion;
 import org.hibernate.cache.CacheException;
-import org.hibernate.cache.spi.RegionFactory;
-import org.hibernate.cache.spi.CacheDataDescription;
-import org.hibernate.cache.spi.CollectionRegion;
-import org.hibernate.cache.spi.EntityRegion;
-import org.hibernate.cache.spi.TimestampsRegion;
+import org.hibernate.cache.spi.*;
```
mustafaiman: Please do not use star imports.

hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
```diff
@@ -48,8 +45,9 @@ public HazelcastCacheRegionFactory(final Properties properties) {
 
     public CollectionRegion buildCollectionRegion(final String regionName, final Properties properties,
                                                   final CacheDataDescription metadata) throws CacheException {
+        /* Collection regions are never versioned, so pass in null for metadata */
         return new HazelcastCollectionRegion<IMapRegionCache>(instance, regionName, properties, metadata,
-                new IMapRegionCache(regionName, instance, properties, metadata));
+                new IMapRegionCache(regionName, instance, properties, null));
```
mustafaiman: Why do you override `metadata` with null? Can somebody not provide metadata for `CollectionRegion`?

hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
```diff
@@ -58,6 +56,14 @@ public EntityRegion buildEntityRegion(final String regionName, final Properties
                 new IMapRegionCache(regionName, instance, properties, metadata));
     }
 
+    public NaturalIdRegion buildNaturalIdRegion(final String regionName, final Properties properties,
+                                                final CacheDataDescription metadata)
+            throws CacheException {
+        /* Natural id regions are never versioned, so pass in null for metadata */
+        return new HazelcastNaturalIdRegion<IMapRegionCache>(instance, regionName, properties, metadata,
+                new IMapRegionCache(regionName, instance, properties, null));
```
mustafaiman: Why, again, overriding `metadata`?

mustafaiman:
Thanks for your effort. Moving `NaturalIdRegion` creation out of `AbstractHazelcastCacheRegionFactory` makes sense. Can you do the same thing for `QueryResultsRegion` too so regions are all consistent.

hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
cszmajda: The comment `/* Natural id regions are never versioned, so pass in null for metadata */` above explains why. 

The *only* place `IMapRegionCache` uses the `metadata` for is [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/distributed/IMapRegionCache.java#L68) where it gets the entity's `versionComparator`.  The only place this is used for is [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/distributed/IMapRegionCache.java#L102) where a version comparison is used to decide which is the latest version of an entity that should be cached. 

For the `HazelcastEntityRegion`, this version comparison makes sense as the latest version of an entity (assuming a version field exists) is always the freshest version of the object.  This is why `buildEntityRegion` passes in the `metadata` to the `IMapRegionCache()` constructor. 

But for `HazelcastCollectionRegion` (which stores the `@OneToMany` etc. relationships between entities and not the entities themselves) and all other region types, this version comparison makes no sense, and it would actually be incorrect to use an entity's version to decide which is the freshest copy of the data.  An entity's relationships to other objects can easily be modified without incrementing its version field, so using the `versionComparator` in non-entity regions could conceivably lead to stale data being cached.

hazelcast-hibernate/hazelcast-hibernate4/src/main/java/com/hazelcast/hibernate/HazelcastCacheRegionFactory.java
cszmajda: Please see my note "The comment `/* Natural id regions are never versioned, so pass in null for metadata */` above explains why." above

cszmajda:
Hey @mustafaiman, thanks for your feedback.  On this question:

> Moving `NaturalIdRegion` creation out of `AbstractHazelcastCacheRegionFactory` makes sense. Can you do the same thing for `QueryResultsRegion` too so regions are all consistent.

AFAICT the `QueryResultsRegion` creation in `AbstractHazelcastCacheRegionFactory` is actually *deliberately* inconsistent with all the other region types. While all the other regions are constructed as `IMapRegionCache`s in DISTRIBUTED mode and `LocalRegionCache`s in LOCAL mode, the `QueryResultsRegion` alone is always constructed as a `LocalRegionCache` in both modes. 

Why it was done this way isn't 100% clear to me, this code is a few years old and seems to have been originally contributed by someone outside Hazelcast.  But it's possible it could have been to avoid severe performance regressions in situations that only occur with the Hibernate query cache.  The query cache can receive a very large number of gets and puts (after all, it is accessed on every cached query) and often does not achieve a very high hit rate.  Distributing this cache region across the cluster with `IMap`s (no matter how they're configured) could place a very high load on remote operations, while the `LocalRegionCache` with `ITopic` notifications used for invalidations could be much more efficient and scalable.  The only downside of the `LocalRegionCache`, slightly reduced consistency semantics, doesn't seem to be a problem for the query cache. 

But this is just my speculation and I'm not really sure why it was done this way. But changing it could have a significant effect on the performance and stability of the Hibernate integration, especially for users who put their query caches under heavy load in DISTRIBUTED mode.  It would also be completely untested, whereas the changes in this PR for `NaturalIdRegion`s have been tested by us and our customers under production workloads for months without issue. 

So can I suggest if you would like to make changes to the `QueryResultsRegion` too that we split this out into a different PR where the potential effects of this can be discussed separately?

cszmajda:
I don't have access to see what the failing check is about.  Is this blocking this PR and can someone with access say what the failure is?

mustafaiman:
verify

mustafaiman:
run-lab-run

hasancelik:
run-lab-run

mustafaiman:
run-lab-run

mustafaiman:
run-lab-run

mustafaiman:
run-lab-run

kobalski:
@cszmajda  We have created a new repo for hibernate modules in order to have more flexible release cycles. Hopefully, separating integration module will enable us releases independent from main hazelcast module. Could you please send your PR to new hibernate repo(https://github.com/hazelcast/hazelcast-hibernate)? Thanks.

pveentjer:
I'm going to close this one since hibernate has been moved to a different repo

PR #7038, opened 2015-12-09
Code coverage and bugfix of IOUtil

Donnerbart:
* Increased code coverage of `IOUtil`
* Fixed bug with empty String in `IOUtil.compress()`

Class is modified in hot restart feature, so it's on my radar.

I deleted the following unused methods (are also not used in EE):
* `public static int copyToDirectBuffer(ByteBuffer src, ByteBuffer dst)`
* `public static void writeLongString(DataOutput dos, String str) throws IOException`
* `public static String readLongString(DataInput in) throws IOException`


hasancelik:
run-lab-run

Donnerbart:
run-lab-run

gurbuzali:
:+1:

Issue #7022, opened 2015-12-08
ReplicatedMapEntryListenerOnReconnectTest.testListenersSmartRouting


hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 120000 milliseconds
	at java.lang.Thread.sleep(Native Method)
	at java.lang.Thread.sleep(Thread.java:321)
	at java.util.concurrent.TimeUnit.sleep(TimeUnit.java:356)
	at com.hazelcast.test.HazelcastTestSupport.sleepMillis(HazelcastTestSupport.java:254)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:843)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:849)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersInternal(AbstractListenersOnReconnectTest.java:72)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersSmartRouting(AbstractListenersOnReconnectTest.java:108)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:622)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:334)
	at java.util.concurrent.FutureTask.run(FutureTask.java:166)
	at java.lang.Thread.run(Thread.java:701)
```

[...]

PR #7053, opened 2015-12-11
Add more debugging for Listener reconnect test

ihsandemir:
Added more debugging for Listener reconnect test. This is put due to non produceable issue https://github.com/hazelcast/hazelcast/issues/7022 . In the future failures, we can use this log for problem identification.

### DIFF
--- a/hazelcast-client-legacy/src/test/java/com/hazelcast/client/listeners/AbstractListenersOnReconnectTest.java
+++ b/hazelcast-client-legacy/src/test/java/com/hazelcast/client/listeners/AbstractListenersOnReconnectTest.java
@@ -71,8 +71,12 @@ public abstract class AbstractListenersOnReconnectTest extends HazelcastTestSupp
 
         assertTrueEventually(new AssertTask() {
             @Override
-            public void run() throws Exception {
-                assertEquals(EVENT_COUNT, eventCount.get());
+            public void run()
+                    throws Exception {
+                int current = eventCount.get();
+                String msg = "[testListenersInternal] Current event count is " + current + " and expected event count is "
+                        + EVENT_COUNT;
+                assertEquals(msg, EVENT_COUNT, current);
             }
         });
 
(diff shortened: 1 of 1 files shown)


cangencer:
my only comment would be that the test fails due to timeout because the overall timeout is the same as the duration given for assertTrueEventually. So if the test failed due to assertTrueEventually then you would be able to see the real fail reason in the logs.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/listeners/AbstractListenersOnReconnectTest.java
```diff
@@ -72,7 +73,9 @@ public void run() throws Exception {
         assertTrueEventually(new AssertTask() {
             @Override
             public void run() throws Exception {
-                assertEquals(EVENT_COUNT, eventCount.get());
+                int current = eventCount.get();
+                Logger.getLogger(this.getClass()).info("[testListenersInternal] Current event count is " + current);
+                assertEquals(EVENT_COUNT, current);
```
sancar: Pass the error message to logger as parameter. This way when test failed, you will have the error message printed.

ihsandemir: @sancar @cangencer Can pointed out a good point that the time for test and assertTrueEventually are both 120 secs, hence the JUnit timeout will always kick in before assertTrueEventually timeout. The issue log also verifies this. The exception in the issue is reported to be "org.junit.runners.model.TestTimedOutException: test timed out after 120000 milliseconds". 

Hence, if we make the logging as Sancar suggested, then we should play with the test timeout so that it is guaranteed that the assertTrueEventually timeout is caught. Otherwise, we can not log using the message in  assertTrueEventually.

ihsandemir:
@cangencer  @sancar Changed the test time from 120 to 150 secs to catch the exception correctly. Also put the log into the exception message as suggested.

ihsandemir:
run-lan-run

cangencer:
:+1:

PR #7061, opened 2015-12-12
Removing remote calls in internalExecutor of ClientExecutionService

sancar:
Since internalExecutor is used by critical operations like
CleanResourcesTask(responsible for setting free all invocations
waiting on dead connection) or HeartBeat task, there shouldn't
be any blocking operation on this executor.

Refactored LoadAllTask and PartitionService since they were doing
remote calls.

Fixed a bug on rescheduleInvocation. There were a rare change that
An invcation is scheduled but not able to send to internal executor.
In that case, the rescheduled invocation will not be aware that
it is not able to run. Added a callback to be able to notify it in
this situation.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
@@ -17,6 +17,7 @@
 package com.hazelcast.client.spi.impl;
 
 import com.hazelcast.client.spi.ClientExecutionService;
+import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.Logger;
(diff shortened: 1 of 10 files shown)


sancar:
@serkan-ozal Can you review the changes in AbstractClientCacheProxyBase ?

sancar:
And for the reviewers, there is one more remote call operation done in executors. It is the authentication. I wanted to change that but it got tricky. I will try to prepare a second pr for that.

sancar:
rub-lab-run

sancar:
verify

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
```diff
@@ -142,11 +168,11 @@ public void shutdown() {
         shutdownExecutor("internal", internalExecutor);
     }
 
-    private void executeInternalSafely(Runnable command) {
+    private void executeInternalSafely(Runnable command, ExecutionCallback executionCallback) {
         try {
             executeInternal(command);
```
cangencer: we are not trapping failures in executeInternal(command) to ExecutionCallback, I think this sould be using the submitInternal and then passing the callback with .andThen()

mdogan:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
sancar: Yes, usage seems not right by itself but my fix was for schedule methods. This method used by them. And if we get an exception when scheduled task trying to call executeInternalSafely, we may miss an exception.

PR #7063, opened 2015-12-13
ClientMessage size calculation problem causes NegativeArraySizeException exception

ihsandemir:
Some tests returns large message (e.g. a query returning 250000 entries which requires a message size of 1090944467 bytes). The QuickMath.nextPowerOfTwo(initialCapacity); overflows the integer max value.

Solution: if such a large value is required, just set the buffer size to be equal to the max int value.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.client.impl.protocol;
 
+import com.hazelcast.client.impl.protocol.exception.MaxMessageSizeExceeded;
 import com.hazelcast.client.impl.protocol.util.ClientProtocolBuffer;
 import com.hazelcast.client.impl.protocol.util.MessageFlyweight;
 import com.hazelcast.client.impl.protocol.util.SafeBuffer;
@@ -130,12 +131,23 @@ public class ClientMessage
     }
 
     public static ClientMessage createForEncode(int initialCapacity) {
-        initialCapacity = QuickMath.nextPowerOfTwo(initialCapacity);
+        initialCapacity = findSuitableMessageSize(initialCapacity);
         if (USE_UNSAFE) {
             return createForEncode(new UnsafeBuffer(new byte[initialCapacity]), 0);
         } else {
             return createForEncode(new SafeBuffer(new byte[initialCapacity]), 0);
         }
+    }
+
+    public static int findSuitableMessageSize(int desiredMessageSize) {
+        if (desiredMessageSize < 0) {
+            throw new MaxMessageSizeExceeded();
+        }
+        desiredMessageSize = QuickMath.nextPowerOfTwo(desiredMessageSize);
+        if (desiredMessageSize < 0) {
+            desiredMessageSize = Integer.MAX_VALUE;
+        }
+        return desiredMessageSize;
     }
 
     public static ClientMessage createForEncode(ClientProtocolBuffer buffer, int offset) {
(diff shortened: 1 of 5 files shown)


asimarslan:
:+1:

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
```diff
@@ -129,6 +129,9 @@ public static ClientMessage create() {
 
     public static ClientMessage createForEncode(int initialCapacity) {
         initialCapacity = QuickMath.nextPowerOfTwo(initialCapacity);
+        if (initialCapacity < 0) {
+            initialCapacity = Integer.MAX_VALUE;
+        }
         if (USE_UNSAFE) {
```
cangencer: would it not be better to have this fix in QuickMath.nextPowerOfTwo? I can see it's called from some other places also

jerrinot: I do not think so. the point of the `QuickMath` class is to do as little work as possible. Its contract explicitly says:
```
This method is not suitable for {@link Integer#MIN_VALUE} or numbers greater than 2^30.
```

ihsandemir: QuickMath.nextPowerOfTwo are being called from places such as Int2ObjectHashMap. We may need to analyse and understand what would be side affects of changing this method. I would prefer not touching this method now as it is a bug fix period but surely we can open an issue for examining this change for later releases.

ihsandemir:
run-lan-run

jerrinot:
this PR solves only a part of the problem - when the size requested is actually <2GB, but the next power of two is > MAX_LONG.

However there is still a problem when it overflows during the size calculation - here: https://github.com/jerrinot/hazelcast/blob/f0fedcbaf6f67e5fce81500e07fc531c6536a941/hazelcast-code-generator/src/main/resources/codec-template-java.ftl#L27-L27

perhaps we should add an overflow check into the template?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
cangencer: Those places will have the same overflow problem, such as trying to create an array with negative capacity

ihsandemir:
@jerrinot Yes, you are right. We need this check on all codecs both generated and manually written codecs. So, what should we do if we detect such a large size message, throw exception? We normally introduced the framing in the protocol but it is not fully implemented yet, which would actually solve such a problem.

jerrinot:
it should propagate the error to the caller (client)

ihsandemir:
@asimarslan We may need a new error code for this case. I checked at the existing codes at ClientExceptionFactory, there is QueryResultSizeExceededException but this error may occur for other protocol messages as well. What do you think? Should we add a new error code to the protocol such as ResponseMaxMessageSizeExceeded exception?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
ihsandemir: @jerrinot @cangencer  The method contract explicitly limits for possible values of this method parameter but i also don't want another duplicate method for this purpose. Currently, I did put checks outside the QuickMath.nextPowerOfTwo where it is being called in two places for the protocol implementation. I still make a call with a value not matching the method contract in this fix for the sake of not duplicating the same code. Let me know if it is good enough for this fix.

cangencer:
:+1: (when tests pass)

ihsandemir:
run-lan-run

ihsandemir:
run-lan-run

ihsandemir:
verify

ihsandemir:
run-lan-run

ihsandemir:
run-lan-run

ihsandemir:
run-lan-run

PR #7064, opened 2015-12-13
Added default timeout (5min) to tests

### LINKED ISSUE #6978: Unit tests should have default timeout
Unittests currently don't have a default timeout and this can lead to a single unit tests stalling the build till the build run into a timeout. This is a problem because apart from this problematic test, any other problematic test is hidden.

Therefor by default each unit test should have a default timeout of a few minutes unless one is defined explicitly.

gurbuzali:
refactored `XATestWithJCA` sot that `getConnection()` is called from the same thread
fixes #6978


gurbuzali:
verify

jerrinot:
:+1:

ahmetmircik:
:+1:

PR #7068, opened 2015-12-14
Fixes memory leak in TestNodeRegistry by removing node upon instance …

ahmetmircik:
…termination

This can be seen in continuous `HazelcastInstance create-terminate cycles`, as in some stress tests.


ahmetmircik:
run-lab-run

Donnerbart:
I don't know the whole context of the mock framework, but the change looks reasonable to me. Actually the added test explains quite well for me what it does.

:+1:

metanet:
:+1:

PR #7071, opened 2015-12-15
Introducing max allowed invocation for clients

sancar:
Its default value is int max. If total number of invocations that
did not return response yet passes the configured threshold, user
gets HazelcastOverloadException.

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
@@ -60,7 +60,7 @@ public class ClusterAuthenticator implements Authenticator {
         //contains remoteAddress and principal
         SerializableList collectionWrapper;
         final ClientInvocation clientInvocation = new ClientInvocation(client, auth, connection);
-        final Future<SerializableList> future = clientInvocation.invoke();
+        final Future<SerializableList> future = clientInvocation.invokeUrgent();
         try {
             collectionWrapper = ss.toObject(future.get());
         } catch (Exception e) {
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/CallIdSequence.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/CallIdSequence.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi.impl;
+
+
+import com.hazelcast.core.HazelcastOverloadException;
+
+import java.util.concurrent.atomic.AtomicLongArray;
+
+import static com.hazelcast.nio.Bits.CACHE_LINE_LENGTH;
+import static com.hazelcast.nio.Bits.LONG_SIZE_IN_BYTES;
+
+
+public abstract class CallIdSequence {
+
+    /**
+     * Creates the next call-id.
+     * <p/>
+     *
+     * @return the generated callId.
+     */
+    public abstract long next();
+
+    /**
+     * Creates the call id without applying any backpressure
+     *
+     * @return the generated call id.
+     */
+    public abstract long renew();
+
+    public abstract void complete();
+
+    /**
+     * A {@link com.hazelcast.spi.impl.operationservice.impl.CallIdSequence} that provided backpressure by taking
+     * the number in flight operations into account when a call-id needs to be determined.
+     * <p/>
+     * It is possible to temporary exceed the capacity:
+     * - due to system operations
+     * - due to racy nature of checking if space is available and getting the next sequence.
+     * <p/>
+     */
+    public static final class CallIdSequenceWithBackpressureViaException extends CallIdSequence {
+        private static final int INDEX_HEAD = 7;
+        private static final int INDEX_TAIL = INDEX_HEAD + CACHE_LINE_LENGTH / LONG_SIZE_IN_BYTES;
+
+        // instead of using 2 AtomicLongs, we use an array if width of 3 cache lines to prevent any false sharing.
+        private final AtomicLongArray longs = new AtomicLongArray(3 * CACHE_LINE_LENGTH / LONG_SIZE_IN_BYTES);
+
+        private final int maxConcurrentInvocations;
+
+        public CallIdSequenceWithBackpressureViaException(int maxConcurrentInvocations) {
+            this.maxConcurrentInvocations = maxConcurrentInvocations;
+        }
+
+        @Override
+        public long next() {
+            if (!hasSpace()) {
+                throw new HazelcastOverloadException("Failed to get a callId for invocation: ");
+            }
+
+            return longs.incrementAndGet(INDEX_HEAD);
+        }
+
+        @Override
+        public long renew() {
+            return longs.incrementAndGet(INDEX_HEAD);
+        }
+
+        private boolean hasSpace() {
+            return longs.get(INDEX_HEAD) - longs.get(INDEX_TAIL) < maxConcurrentInvocations;
+        }
+
+        @Override
+        public void complete() {
+            longs.incrementAndGet(INDEX_TAIL);
+        }
+    }
+}
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -50,8 +50,8 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.atomic.AtomicInteger;
 
+import static com.hazelcast.client.config.ClientProperty.MAX_CONCURRENT_INVOCATIONS;
 import static com.hazelcast.instance.OutOfMemoryErrorDispatcher.onOutOfMemory;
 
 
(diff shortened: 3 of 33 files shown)


sancar:
From the tests you will notice that, I am throwing the exception from future. I am still not convinced that it is thrown from async methods directly. I think, most of the users will not expect an an exception there. 

Secondly if I try to throw the exception directly from async methods I need to do lots of refactoring on the codebase. I need to separate where invocation is called for the first time from the retry of invocation(one will send expception, other needs to set to future since it is no longer in user thread). And whole internal things that is using invocation should expect HazelcastOverLoadExpcetion from invoke method directly.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -148,6 +154,17 @@ private void registerInvocation(ClientInvocation clientInvocation) {
...
+        }
+        long currentConcurrentInvocations = callIdIncrementer.get() - tailCallId.get();
+        if (maxAllowedConcurrentInvocations <= currentConcurrentInvocations) {
+            throw new HazelcastOverloadException("Max allowed concurrent invocations is reached : "
+                    + maxAllowedConcurrentInvocations);
+        }
+    }
+
     private ClientInvocation deRegisterCallId(int callId) {
         return callIdMap.remove(callId);
```
serkan-ozal: `tailCallId` must be incremented at here

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -153,7 +158,19 @@ private void registerInvocation(ClientInvocation clientInvocation) {
...
+        }
+        long currentConcurrentInvocations = callIdIncrementer.get() - tailCallId.get();
+        if (maxAllowedConcurrentInvocations <= currentConcurrentInvocations) {
+            throw new HazelcastOverloadException("Max allowed concurrent invocations is reached : "
+                    + maxAllowedConcurrentInvocations);
+        }
+    }
+
     private ClientInvocation deRegisterCallId(long callId) {
+        tailCallId.incrementAndGet();
```
serkan-ozal: what about incrementing `tailCallId` if `callIdMap.remove(callId);` returns not-null?

hazelcast-client/src/test/java/com/hazelcast/client/ClientMaxAllowedInvocationTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        Future future = map.getAsync(1);
+        try {
+            future.get();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        } catch (ExecutionException e) {
+            assertTrue(e.getCause() instanceof HazelcastOverloadException);
+        }
+
```
serkan-ozal: test should fail if there is no exception (`HazelcastOverloadException`) here

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -148,6 +154,17 @@ private void registerInvocation(ClientInvocation clientInvocation) {
         }
     }
 
+    private void checkMaxAllowedReached() {
+        if (maxAllowedConcurrentInvocations == Integer.MAX_VALUE) {
+            return;
+        }
+        long currentConcurrentInvocations = callIdIncrementer.get() - tailCallId.get();
+        if (maxAllowedConcurrentInvocations <= currentConcurrentInvocations) {
+            throw new HazelcastOverloadException("Max allowed concurrent invocations is reached : "
```
serkan-ozal: I think, we should increase `tailCallId` even when there is exception. Otherwise, there will be inconsistency between `callIdIncrementer` and `tailCallId`

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: Mistake from copy paste for the old client. Good catch

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: Yes, it is already like that. I am incrementing the tailCallId as soon as I get incoming message. And this is not correct  place if you are searching for it. It is incremented in deregisterCallId method.

serkan-ozal: no, I am mentioning about `deregisterCallId` is not called when `HazelcastOverloadException` is thrown here. Is it called?

serkan-ozal: For example, max allowed concurrent invocation count is `10`, and you submit one more. In this case, `callIdIncrementer` is `11`, you throw `HazelcastOverloadException` but `tailCallId` is `0`. Then the first `10` invocation is finished and you incremented `tailCallId` and it is `10` but  `callIdIncrementer` is `11`. This is the inconsistency I have mentioned.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: Yes, you might be right. I think I need to decrease only if there is invocation on the map. But I want to think about this a little bit more.

hazelcast-client/src/test/java/com/hazelcast/client/ClientMaxAllowedInvocationTest.java
sancar: Yes, you are right. I will fix this one.

pveentjer:
I think that also for an async call the exception should be immediately put in the face of the user.

See for example the RejectedExecutionException when e..g an ExecutorService.submit is called. Even though a future normally is returned; if the executor doesn't want to execute your task, no future is returned but the caller immediately gets a exception in his face.

Imho the HazelcastOverloadException is similar.

pveentjer:
@serkan-ozal perfect consistency is not very important here. Nobody is going to care of e.g. 10 invocations are done more than specified. In my CallIdSequence I explicitly mention this; otherwise you end up in a lot more complex concurrency logic and will probably be a lot slower.

sancar:
@pveentjer We can't always send the OverLoadExpception to user. It  might be case that invocation is retried and we are no longer in user thread but in our executors. In this case, we have no chance to set the exception to future to let the user know. And for the sake of consistency , I think setting it to future makes more sense.

serkan-ozal:
@pveentjer are you sure, you have read my comments carefully?  it is not just inconsistency, it is a bug if I don't get wrong the code

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: Yes, there seems to be an issue. I will look into it.

pveentjer:
I guess I didn't read it carefully.

sancar:
I integrated CalIdSequence from https://github.com/hazelcast/hazelcast/pull/7075
Only one version of CallIdSequence implemented. Others can be implemented later.
@serkan-ozal  I addressed the consistency issue that you mentioned.
Introduced invokeUrgent to bypass backpressure for important messages.
@pveentjer Exception is send to user directly as we discussed for the first invocation, if it is retried it will be set to future instead.

sancar:
verify

hazelcast-client-legacy/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
```diff
@@ -60,7 +60,7 @@ public void authenticate(ClientConnection connection) throws AuthenticationExcep
         //contains remoteAddress and principal
         SerializableList collectionWrapper;
         final ClientInvocation clientInvocation = new ClientInvocation(client, auth, connection);
-        final Future<SerializableList> future = clientInvocation.invoke();
+        final Future<SerializableList> future = clientInvocation.invokeUrgent();
```
serkan-ozal: It might be better introducing a new interface like `UrgentClientRequest` and so `ClientInvocation` takes a request as urgent automatically if it implements `UrgentClientRequest`. 

I think, this way, when there is new client request that is needed to be sent urgent, it will be easier to integrate instead of changing the invocation call code to `invokeUrgent` from `invoke` in all places it is called.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/CallIdSequence.java
```diff
@@ -0,0 +1,92 @@
...
+     * the number in flight operations into account when a call-id needs to be determined.
+     * <p/>
+     * It is possible to temporary exceed the capacity:
+     * - due to system operations
+     * - due to racy nature of checking if space is available and getting the next sequence.
+     * <p/>
+     */
+    public static final class CallIdSequenceWithBackpressureViaException extends CallIdSequence {
+        private static final int INDEX_HEAD = 7;
+        private static final int INDEX_TAIL = 15;
```
serkan-ozal: I know, this part is taken from PR #7075
It would be better that calculating `INDEX_TAIL` as `INDEX_HEAD + CACHE_LINE_LENGTH / LONG_SIZE_IN_BYTES` so if `com.hazelcast.nio.Bits#CACHE_LINE_LENGTH` is changed, `INDEX_TAIL` will be updated accordingly.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/impl/ClusterAuthenticator.java
sancar: We dont have requests anymore in new client. And it does not fit to new design.

serkan-ozal: I know, this is why I have made this comment under `hazelcast-client-legacy` not `hazelcast-client`

sancar: Ok than. I don't want to spend time on different solution on new and old client. Old client will be deleted soon anyway.

sancar:
@pveentjer @serkan-ozal Can I get a couple of :+1:  so that we can merge this ?

serkan-ozal:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -130,12 +132,20 @@ public ClientInvocationFuture invoke() {
         try {
             invokeOnSelection();
         } catch (Exception e) {
+            if (e instanceof HazelcastOverloadException) {
+                throw (HazelcastOverloadException) e;
+            }
```
cangencer: could use multiple catch block here and just rethrow it instead

cangencer:
Would it be more consistent to call everything CorrelationId , rather than callId in some places and correlationId in the message?

enesakar:
@pveentjer Can you make your final review please? We need to merge this today to include into 3.6-EA3.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
mdogan: agree. :+1:

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/CallIdSequence.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return the generated callId.
+     */
+    public abstract long next();
+
+    /**
+     * Creates the call id without applying any backpressure
+     *
+     * @return the generated call id.
+     */
+    public abstract long renew();
```
mdogan: not sure about naming...

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/CallIdSequence.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * A {@link com.hazelcast.spi.impl.operationservice.impl.CallIdSequence} that provided backpressure by taking
+     * the number in flight operations into account when a call-id needs to be determined.
+     * <p/>
+     * It is possible to temporary exceed the capacity:
+     * - due to system operations
+     * - due to racy nature of checking if space is available and getting the next sequence.
+     * <p/>
+     */
+    public static final class CallIdSequenceWithBackpressureViaException extends CallIdSequence {
```
mdogan: You can create a _without-backpressure_ version if max is equal to long.max. not a must... but easy to switch..

mdogan:
I added a few minor comments. other than those, it's good. :+1:

enesakar:
merging. @sancar can you handle commented issues in a separate PR?

sancar:
@enesakar OK

PR #7074, opened 2015-12-15
Add fsync flag to map & cache hot restart configurations

mdogan:
- Renamed HotRestartConfig to HotRestartPersistenceConfig
- Created HotRestartConfig for map & cache which included fsync flag
which defaults to false.

Also includes a non-related trivial change where more diagnostic data is added to `CallerNotMemberException`.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
@@ -140,7 +140,7 @@ public class MapConfig {
 
     private String quorumName;
 
-    private boolean hotRestartEnabled;
+    private HotRestartConfig hotRestartConfig = new HotRestartConfig();
 
     private MapConfigReadOnly readOnly;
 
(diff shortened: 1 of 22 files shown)


hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
```diff
@@ -195,15 +195,6 @@ public MapConfigReadOnly getAsReadOnly() {
         return readOnly;
     }
 
-    public boolean isHotRestartEnabled() {
```
gurbuzali: how about not removing these methods so that enterprise does not break ?

mdogan:
verify

mtopolnik:
:+1:

hazelcast/src/main/java/com/hazelcast/config/MapConfig.java
mdogan: you mean to make `isHotRestartEnabled()` to delegate to hot-restart-config?

mtopolnik: But I suppose delete them as soon as EE PR is merged.

mdogan: @gurbuzali: Even if we keep these ones, still EE won't compile. Because I renamed HotRestartConfig to HotRestartPersistenceConfig and HotRestartConfig is now something else.

gurbuzali:
:+1:

PR #7075, opened 2015-12-15
[DONT MERGE] client side back pressure

pveentjer:


jerrinot:
The customer requires throwing an exception instead of blocking. It appears to me it should be easy to implement it within this PR; it just a different back-off strategy after all. 

Once the "`THROW_EXCEPTION`" strategy is implemented then it's the same as #7071. This PR uses the same abstraction (`CallIdSequence`) to implemented the backpressure as the member-to-member communication so it makes sense to use this. @sancar @asimarslan: WDYT?

serkan-ozal:
@jerrinot you are mentioning about a configurable strategy for throwing exception or waiting, right?

jerrinot:
indeed

serkan-ozal:
@jerrinot i am also definitely agree on it

pveentjer:
Just set change this:
```
    long maxWaitingForCallIdMs = 0;
```

PR #7077, opened 2015-12-15
Changes required for integration with Hazelcast Jet

toktarev:
Changes
Hazelcast Server

1) Introduced interface CustomServiceConfig.
   It let us to add custom configs for each external service.
2) HazelcastInstanceFactory has been customized. It let  us to create custom instances of HazelcastInstanceFactory adding them to META-INF/services
By default DefaultHazelcastInstanceFactory will be used.

3) HazelcastInstanceImpl and have been extended.
Now it is possible to inherit from these classes.

Hazelcast Client

1) Introduced interface HazelcastClientFactory to create custom instances of HazelcastClientInstanceImpl

2) CodeGeneratorMessageTaskFactory has been customized
Added opportunity to generate Tasks from Jet component: JetMessageTaskFactoryImpl.

3) TemplateConstants - added JET_TEMPLATE_ID = 26.

4) Opportunity to work with many MessageTaskFactory instances
Introduced method createMessageTaskFactories in NodeExtension.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientFactory.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientFactory.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client;
+
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.client.impl.HazelcastClientProxy;
+import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
+import com.hazelcast.client.impl.ClientConnectionManagerFactory;
+
+/***
+ * This is interface which provides capability for Hazelcast client factories customization;
+ * It's implementation can be changed and passed to the HazelcastClientManager's constructors;
+ */
+public interface HazelcastClientFactory<T extends HazelcastClientInstanceImpl,
+        V extends HazelcastClientProxy,
+        C extends ClientConfig> {
+    T createHazelcastInstanceClient(C config,
+                                    ClientConnectionManagerFactory hazelcastClientFactory);
+
+    V createProxy(T client);
+}
--- a/hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client;
+
+import java.util.HashSet;
+import java.util.Collection;
+import java.util.Collections;
+
+import com.hazelcast.util.EmptyStatement;
+
+import java.util.concurrent.ConcurrentMap;
+
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.OutOfMemoryHandler;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.client.impl.HazelcastClientProxy;
+import com.hazelcast.instance.OutOfMemoryErrorDispatcher;
+import com.hazelcast.core.DuplicateInstanceNameException;
+import com.hazelcast.client.config.XmlClientConfigBuilder;
+import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
+import com.hazelcast.client.impl.ClientConnectionManagerFactory;
+import com.hazelcast.client.impl.DefaultClientConnectionManagerFactory;
+
+/***
+ * This is central manager for all Hazelcast clients in JVM;
+ * All creation functionality will be stored here and particular;
+ * instance of Client will delegate here;
+ */
+public final class HazelcastClientManager {
+    /***
+     * Instance for clientManagers
+     */
+    public static final HazelcastClientManager INSTANCE = new HazelcastClientManager();
+
+    static {
+        OutOfMemoryErrorDispatcher.setClientHandler(new ClientOutOfMemoryHandler());
+    }
+
+    private final ConcurrentMap<String, HazelcastClientProxy> clients
+            = new ConcurrentHashMap<String, HazelcastClientProxy>(5);
+
+    private HazelcastClientManager() {
+    }
+
+    public static HazelcastInstance newHazelcastClient(HazelcastClientFactory hazelcastClientFactory) {
+        return newHazelcastClient(new XmlClientConfigBuilder().build(), hazelcastClientFactory);
+    }
+
+    @SuppressWarnings("unchecked")
+    public static HazelcastInstance newHazelcastClient(ClientConfig config, HazelcastClientFactory hazelcastClientFactory) {
+        if (config == null) {
+            config = new XmlClientConfigBuilder().build();
+        }
+
+        final ClassLoader tccl = Thread.currentThread().getContextClassLoader();
+        HazelcastClientProxy proxy;
+        try {
+            Thread.currentThread().setContextClassLoader(HazelcastClient.class.getClassLoader());
+            ClientConnectionManagerFactory clientConnectionManagerFactory = new DefaultClientConnectionManagerFactory();
+            final HazelcastClientInstanceImpl client = hazelcastClientFactory.createHazelcastInstanceClient(
+                    config,
+                    clientConnectionManagerFactory
+            );
+            client.start();
+            OutOfMemoryErrorDispatcher.registerClient(client);
+            proxy = hazelcastClientFactory.createProxy(client);
+
+            if (INSTANCE.clients.putIfAbsent(client.getName(), proxy) != null) {
+                throw new DuplicateInstanceNameException("HazelcastClientInstance with name '" + client.getName()
+                        + "' already exists!");
+            }
+        } finally {
+            Thread.currentThread().setContextClassLoader(tccl);
+        }
+        return proxy;
+    }
+
+    public static HazelcastInstance getHazelcastClientByName(String instanceName) {
+        return INSTANCE.clients.get(instanceName);
+    }
+
+    public static Collection<HazelcastInstance> getAllHazelcastClients() {
+        Collection<HazelcastClientProxy> values = INSTANCE.clients.values();
+        return Collections.unmodifiableCollection(new HashSet<HazelcastInstance>(values));
+    }
+
+    public static void shutdownAll() {
+        for (HazelcastClientProxy proxy : INSTANCE.clients.values()) {
+            HazelcastClientInstanceImpl client = proxy.client;
+            if (client == null) {
+                continue;
+            }
+            proxy.client = null;
+            try {
+                client.shutdown();
+            } catch (Throwable ignored) {
+                EmptyStatement.ignore(ignored);
+            }
+        }
+        OutOfMemoryErrorDispatcher.clearClients();
+        INSTANCE.clients.clear();
+    }
+
+
+    public static void shutdown(HazelcastInstance instance) {
+        if (instance instanceof HazelcastClientProxy) {
+            final HazelcastClientProxy proxy = (HazelcastClientProxy) instance;
+            HazelcastClientInstanceImpl client = proxy.client;
+            if (client == null) {
+                return;
+            }
+            proxy.client = null;
+            INSTANCE.clients.remove(client.getName());
+
+            try {
+                client.shutdown();
+            } catch (Throwable ignored) {
+                EmptyStatement.ignore(ignored);
+            } finally {
+                OutOfMemoryErrorDispatcher.deregisterClient(client);
+            }
+        }
+    }
+
+    public static void shutdown(String instanceName) {
+        HazelcastClientProxy proxy = INSTANCE.clients.remove(instanceName);
+        if (proxy == null) {
+            return;
+        }
+        HazelcastClientInstanceImpl client = proxy.client;
+        if (client == null) {
+            return;
+        }
+        proxy.client = null;
+        try {
+            client.shutdown();
+        } catch (Throwable ignored) {
+            EmptyStatement.ignore(ignored);
+        } finally {
+            OutOfMemoryErrorDispatcher.deregisterClient(client);
+        }
+    }
+
+    public static void setOutOfMemoryHandler(OutOfMemoryHandler outOfMemoryHandler) {
+        OutOfMemoryErrorDispatcher.setClientHandler(outOfMemoryHandler);
+    }
+}
--- a/hazelcast/src/main/java/com/hazelcast/core/PartitioningStrategy.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/PartitioningStrategy.java
@@ -16,13 +16,15 @@
 
 package com.hazelcast.core;
 
+import java.io.Serializable;
+
 /**
  * PartitioningStrategy allows keys to be located on the same member.
  * This causes related data to be stored in the same location. (See data-affinity.)
  *
  * @param <K> key type
  */
-public interface PartitioningStrategy<K> {
+public interface PartitioningStrategy<K> extends Serializable {
 
     /**
      * Returns the key object that will be used by Hazelcast to specify the partition.
--- a/hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.instance;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.config.XmlConfigBuilder;
+
+/***
+ * Default implementation for Hazelcast instance factoriy;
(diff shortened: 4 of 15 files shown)


hazelcast/src/main/java/com/hazelcast/config/Config.java
```diff
@@ -853,6 +855,22 @@ public JobTrackerConfig findJobTrackerConfig(String name) {
         return getJobTrackerConfig(name);
     }
 
+    public <T extends CustomServiceConfig> T getServiceConfig(String name) {
```
eminn: If you are introducing a new type of configuration, I think it should also configurable from XML and Spring.

toktarev: Ok, I will add it to XML and Spring

hazelcast/src/main/java/com/hazelcast/spi/impl/servicemanager/impl/ServiceManagerImpl.java
```diff
@@ -274,7 +274,7 @@ private void shutdownService(final ManagedService service, final boolean termina
         }
     }
 
-    private synchronized void registerService(String serviceName, Object service) {
+    protected synchronized void registerService(String serviceName, Object service) {
```
eminn: If we provide a way to register services with a programmatic API, we might not need this change since this class is opened for extension just for service registration purposes (See https://github.com/hazelcast-incubator/cascading-prototype/blob/transport/hazelcast-jet/src/main/java/com/hazelcast/jet/impl/hazelcast/JetServiceManagerImpl.java).

toktarev: I don't understand what do you mean here

eminn: Hazelcast currently has a way to register user defined services via XML (http://docs.hazelcast.org/docs/3.5/manual/html-single/index.html#user-defined-services). If Hazelcast provides, similar way with programmatic API, you don't need to make this change by opening the class for extension.

toktarev: I didn't find a programatic way how to register custom service.
Could you please help here ?

eminn: I think we can use ` Map<String, Object> createExtensionServices();` method on the `NodeExtension` to register extension services.

toktarev: Fixed

hazelcast/src/main/java/com/hazelcast/config/Config.java
toktarev: Emin, I suppose you have better knowledge in Hazelcast config subsystem.
Could you please assist a little bit here ?

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -74,22 +74,22 @@
  * we don't leak {@link com.hazelcast.spi.impl.operationservice.impl.OperationServiceImpl} to the outside.
  */
 public class NodeEngineImpl implements NodeEngine {
+    protected final ILogger logger;
+    protected final EventServiceImpl eventService;
+    protected final OperationServiceImpl operationService;
+    protected final ServiceManagerImpl serviceManager;
+    protected final WanReplicationService wanReplicationService;
```
eminn: Do we really need to make these fields protected ?

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -99,21 +99,33 @@ public NodeEngineImpl(final Node node) {
...
         );
-        quorumService = new QuorumServiceImpl(this);
+    }
+
+    protected PacketHandler connectionManagerPacketHandler() {
+        return new ConnectionManagerPacketHandler();
+    }
+
+    protected ServiceManagerImpl createServiceManager() {
+        return new ServiceManagerImpl(this);
```
eminn: I think we already solved problem with service registration, no need to extend service managers.

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
toktarev: No needs. You are right, will be fixed

hazelcast/src/main/java/com/hazelcast/spi/impl/packetdispatcher/impl/PacketDispatcherImpl.java
```diff
@@ -32,7 +32,7 @@
  */
 public class PacketDispatcherImpl implements PacketDispatcher {
 
-    private final ILogger logger;
+    protected final ILogger logger;
```
eminn: Is this also needed ?

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
toktarev: yep

hazelcast/src/main/java/com/hazelcast/spi/impl/packetdispatcher/impl/PacketDispatcherImpl.java
toktarev: yep, will fix

hazelcast/src/main/java/com/hazelcast/instance/IHazelcastInstanceFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.Member;
+
+public interface IHazelcastInstanceFactory {
```
pveentjer: We need to have documentation on this class.

hazelcast/src/main/java/com/hazelcast/instance/IHazelcastInstanceFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+    Set<HazelcastInstanceImpl> getInstanceImpls(Collection<Member> members);
+
+    void remove(HazelcastInstanceImpl instance);
+
+    void shutdownAll();
+
+    void terminateAll();
+
+    Map<MemberImpl, HazelcastInstanceImpl> getInstanceImplMap();
```
pveentjer: I would move this method underneath the 'getInstanceImp' so that is closely placed to another related method 'getInstanceImpls'

hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
```diff
@@ -130,7 +131,7 @@
     /**
      * Creates factory method that creates server side client message handlers
      */
-    MessageTaskFactory createMessageTaskFactory();
+    List<? extends MessageTaskFactory> createMessageTaskFactories();
```
pveentjer: The Javadoc doesn't match the method name AFAIK. And why is a list needed? Is ordering relevant?

hazelcast/src/main/java/com/hazelcast/util/LinkedMapBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+    public static <K, V> Map<K, V> of(K key1, V value1, K key2, V value2, K key3, V value3) {
+        return LinkedMapBuilder.<K, V>builder().put(key1, value1).put(key2, value2).put(key3, value3).build();
+    }
+
+    public static <K, V> Map<K, V> of(K key1, V value1, K key2, V value2, K key3, V value3, K key4, V value4) {
+        return LinkedMapBuilder.<K, V>builder().put(key1, value1).put(key2, value2).put(key3, value3).put(key4, value4).build();
+    }
+
+    //CHECKSTYLE:OFF
+    public static <K, V> Map<K, V> of(K key1, V value1, K key2, V value2, K key3, V value3, K key4, V value4, K key5, V value5) {
```
pveentjer: Why do you need a builder at all here? It just creates another short lived object without real value.

hazelcast/src/main/java/com/hazelcast/util/LinkedMapBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Map;
+import java.util.LinkedHashMap;
+
+/***
+ * This class provides builder pattern to create LinkedHashMap
+ *
+ * @param <K> - type of map of the key
+ * @param <V> - type of map of the value
+ */
+public final class LinkedMapBuilder<K, V> {
```
pveentjer: Is this class used in Hazelcast or only in JET? In this PR I don't see any usage of LinkedMapBuilder.

hazelcast/src/main/java/com/hazelcast/instance/IHazelcastInstanceFactory.java
toktarev: Currently it is used by Jet. But can be used by any external component via META-INF/services.
Ok, I will add documentation.

hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
toktarev: List is needed because we can have many factories for component.
Some factories are being provided by Hazelcast and by Jet

hazelcast/src/main/java/com/hazelcast/instance/IHazelcastInstanceFactory.java
toktarev: I've added documentation.

hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
toktarev: We need list because different components can provide their own factories.
Yes ordering is relevant.

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
```diff
@@ -0,0 +1,156 @@
...
+import com.hazelcast.client.config.XmlClientConfigBuilder;
+import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
+import com.hazelcast.client.impl.ClientConnectionManagerFactory;
+import com.hazelcast.client.impl.DefaultClientConnectionManagerFactory;
+
+public class HazelcastClientManager {
+    /***
+     * Instance for clientManagers
+     */
+    public static final HazelcastClientManager HAZELCAST_CLIENT_MANAGER = new HazelcastClientManager();
```
pveentjer: Not easier to call this INSTANCE?

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
```diff
@@ -0,0 +1,156 @@
...
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.client.impl.HazelcastClientProxy;
+import com.hazelcast.instance.OutOfMemoryErrorDispatcher;
+import com.hazelcast.core.DuplicateInstanceNameException;
+import com.hazelcast.client.config.XmlClientConfigBuilder;
+import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
+import com.hazelcast.client.impl.ClientConnectionManagerFactory;
+import com.hazelcast.client.impl.DefaultClientConnectionManagerFactory;
+
+public class HazelcastClientManager {
```
pveentjer: Can you add a few lines of Javadoc what the purpose of this class is?

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
toktarev: Hmm, easier ? for me it is deal of taste.

You insist on INSTANCE ?

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
toktarev: ok

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
```diff
@@ -0,0 +1,156 @@
...
+    /***
+     * Instance for clientManagers
+     */
+    public static final HazelcastClientManager HAZELCAST_CLIENT_MANAGER = new HazelcastClientManager();
+
+    static {
+        OutOfMemoryErrorDispatcher.setClientHandler(new ClientOutOfMemoryHandler());
+    }
+
```
pveentjer: Do we want to make the constructor private? Since this class also acts like a Singleton; we probably want to prevent having multiple instances.

toktarev: yes, you are right

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
```diff
@@ -0,0 +1,156 @@
...
+    private static final ConcurrentMap<String, HazelcastClientProxy> CLIENTS
```
pveentjer: Why is this field static since HazelcastClientManager already is a singleton. Using the singleton I can access the clients map anway;

```
HazelcastClientManager.INSTANCE.clients
```

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientManager.java
```diff
@@ -0,0 +1,156 @@
...
+            ClientConnectionManagerFactory clientConnectionManagerFactory = new DefaultClientConnectionManagerFactory();
+            final HazelcastClientInstanceImpl client = hazelcastClientFactory.createHazelcastInstanceClient(
+                    config,
+                    clientConnectionManagerFactory
+            );
+            client.start();
+            OutOfMemoryErrorDispatcher.registerClient(client);
+            proxy = hazelcastClientFactory.createProxy(client);
+
+            if (CLIENTS.containsKey(client.getName())) {
```
pveentjer: Could it be that we get some racy behavior since the check/modify is not atomic?

toktarev: Hm, We have the same code before and currently have this code in main HZ :)
I didn't introduce here my own logic)
The race is possible of two thread will try to create two clients with the same name.
Do you think we need to fix it ?

toktarev: Fixed it

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -174,6 +179,21 @@ public void handleClientMessage(ClientMessage clientMessage, Connection connecti
         }
     }
 
+    private MessageTask resolveClientMessage(ClientMessage clientMessage, Connection connection) {
```
pveentjer: I'm not that crazy about this part. It means that processing a client request/message needs to pass through additional logic and we have no idea how many task factories there are.

hazelcast/src/main/java/com/hazelcast/core/PartitioningStrategy.java
```diff
@@ -16,13 +16,15 @@
...
+import java.io.Serializable;
+
 /**
  * PartitioningStrategy allows keys to be located on the same member.
  * This causes related data to be stored in the same location. (See data-affinity.)
  *
  * @param <K> key type
  */
-public interface PartitioningStrategy<K> {
+public interface PartitioningStrategy<K> extends Serializable {
```
pveentjer: Do you know how often this is send over the wire?

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
toktarev: Well generally factories count is the same as components count.
Currently - 2.

hazelcast/src/main/java/com/hazelcast/core/PartitioningStrategy.java
toktarev: I need to send it in Jet

hazelcast/src/main/java/com/hazelcast/instance/IHazelcastInstanceFactory.java
```diff
@@ -0,0 +1,54 @@
...
+
+
+import java.util.Map;
+import java.util.Set;
+import java.util.Collection;
+
+import com.hazelcast.core.Member;
+import com.hazelcast.config.Config;
+import com.hazelcast.core.HazelcastInstance;
+
```
pveentjer: Why don't we clean up the HazelcastInstanceFactory? Because this naming schema is very confusing; 2 classes with simular names, but different purposes makes it confusing.

toktarev: HazelcastInstanceFactory - this is class
IHazelcastInstanceFactory - interface

hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
pveentjer: Can you add it to the documentation for future generations? :)

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
pveentjer: I'm not crazy about it; especially when factory count increases. But also the litter; since you get an additional iterator instance.

hazelcast/src/main/java/com/hazelcast/core/PartitioningStrategy.java
pveentjer: But this just happens a few times; right?

pveentjer: After talking with @toktarev it is fine to be serializable.  Values won't be send very often.

hazelcast-client/src/main/java/com/hazelcast/client/HazelcastClientFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.hazelcast.client;
+
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.client.impl.HazelcastClientProxy;
+import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
+import com.hazelcast.client.impl.ClientConnectionManagerFactory;
+
+public interface HazelcastClientFactory<T extends HazelcastClientInstanceImpl,
```
pveentjer: Could you add some javadoc to this interface for future generations?

hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.instance;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.config.XmlConfigBuilder;
+
+public class DefaultHazelcastInstanceFactory implements HazelcastInstanceFactory {
```
pveentjer: Could you add some javadoc to the class?

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceManager.java
```diff
@@ -47,13 +50,26 @@
 @SuppressWarnings("SynchronizationOnStaticField")
 @PrivateApi
 public final class HazelcastInstanceManager {
```
pveentjer: Could you allign this with the HazelcastClientManager so that either all methods are static, or a singleton is used and the methods are not static? Now we have 1 class with static methods, and another with instance methods.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -140,8 +140,17 @@
 
     private final HazelcastThreadGroup hazelcastThreadGroup;
 
+
     private final boolean liteMember;
 
+    protected NodeEngineImpl createNodeEngine() {
```
pveentjer: Can you add some comments that this method needs to be protected for outside projects? Else someone could make it private

pveentjer:
Minor review comments. Apart from that, it looks good to go!

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceManager.java
toktarev: Done

pveentjer:
:+1:

sancar:
:+1:

PR #7089, opened 2015-12-16
[DONT MERGE] Put data immediately to near cache when LocalInvalidation is CACHE

sancar:
Behaviour was different between sync and async API.
This pr makes sures that data is put the nearcache before remote
operation to cluster when LocalInvalidation is CACHE for both sync
and async put/putIfAbsent operations.

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
@@ -386,15 +386,14 @@ abstract class AbstractClientInternalCacheProxy<K, V>
         } catch (Exception e) {
             throw ExceptionUtil.rethrow(e);
         }
+        if (cacheOnUpdate) {
+            storeInNearCache(keyData, valueData, value);
+        }
         if (!async) {
             try {
                 Object response = future.get();
-                if (nearCache != null) {
-                    if (cacheOnUpdate) {
-                        storeInNearCache(keyData, valueData, value);
-                    } else {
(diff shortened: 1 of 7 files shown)


sancar:
PutIfAbsent is failed ?

com.hazelcast.client.cache.nearcache.ClientNearCacheTest.putIfAbsentAsyncToCacheAndThenGetFromClientNearCacheWithObjectInMemoryFormat
```
Failing for the past 1 build (Since Failed#608 )
Took 1.4 sec.
Error Message

expected:<Value-16> but was:<null>
Stacktrace

java.lang.AssertionError: expected:<Value-16> but was:<null>
	at com.hazelcast.client.cache.nearcache.ClientNearCacheTestSupport.putToCacheAndThenGetFromClientNearCacheInternal(ClientNearCacheTestSupport.java:201)
	at com.hazelcast.client.cache.nearcache.ClientNearCacheTestSupport.putIfAbsentToCacheAndThenGetFromClientNearCache(ClientNearCacheTestSupport.java:210)
	at com.hazelcast.client.cache.nearcache.ClientNearCacheTest.putIfAbsentAsyncToCacheAndThenGetFromClientNearCacheWithObjectInMemoryFormat(ClientNearCacheTest.java:77)
```

serkan-ozal:
@sancar the first version of this behaviour was as your PR. But the problem is that putting to near-cache immediately might lead to inconsistencies because async operation might fail for some reason and in this case we have entry in our near-cache but not in cache. This is why I had moved putting to near-cache logic after response is returned.

cangencer:
could we improve the JavaDoc for LocalUpdatePolicy ? It's not quite clear how it's supposed to act when it's set to CACHE vs INVALIDATE

PR #7097, opened 2015-12-17
NearCache heap cost calculation is wrong when the cache gets concurrent misses

### LINKED ISSUE #7057: NearCacheSizeEstimator raises and raises and compromises Eviction
The attached test program reproduces that the attribute heapCost of the localMapStats of an IMap does not go down to 0 after stuff is added, read and removed, but raises and raises without limit. Debugging into Hazelcast shows that its NearCacheSizeEstimator.size is not 0.

Please notice that this causes serious problems when USED_HEAP_SIZE is used for eviction, since hazelcast also uses this estimator to calculate the maps overall size. Thus, the map will be limited to be smaller and smaller, since the estimator thinks its size is much bigger than it actually is. In our application, this repeatably causes a crash: the hibernate second level cache becomes too small after a while, causing more and more unneccesary SQL fetches for data missing in the cache, leading ultimately to the applications death.

The test program has to be started two times in parallel. It waits until a cluster of two nodes is formed, and then repeatedly fills up the IMap, reads all its contents back from a number of threads in parallel, and removes all items from the IMap with remove. Normally, the heapCost should be 0 after that, but it isn't - in my last tries it was about 600000.

I tried this with JDK 1.7.0_75 with Windows and -Xmx1024m, Windows 8, and both hazelcast 3.4.6 and 3.5.4.

[BrokenMemoryLimitHazelcast.zip](https://github.com/hazelcast/hazelcast/files/59444/BrokenMemoryLimitHazelcast.zip)

cangencer:
Fixes #7057

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
@@ -46,6 +46,8 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
```diff
@@ -196,6 +198,49 @@ public void testNearCacheStats() throws Exception {
...
+            }
+        };
+
+        for (int i = 0; i < threadCount; i++) {
+            executorService.execute(task);
+        }
+        assertOpenEventually(countDownLatch);
+
+        for (int i = 0; i < itemCount; i++) {
+            map.remove(i);
```
ahmetmircik: According to javaDocs [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/config/NearCacheConfig.java#L286), this `map#remove` should not remove entry from near-cache. This seems we have another bug here related with entry removal when `invalidateOnChange` is false.

cangencer: Looks like any put/remove on member will invalidate the cache directly. I think otherwise it would be absurd (you could remove an item and get would still return it). InvalidateOnChange is only for changes from other members, maybe the JavaDoc should be updated accordingly?

ahmetmircik: Updating javaDoc makes sense to me.

cangencer: Ok I will send that in another PR, as it's not related to this change.

ahmetmircik:
:+1:

Issue #7114, opened 2015-12-18
 NormalResponse already responseReceived for callback -  MapLoader regression


tombujok:
MapLoaderMultiNodeTest.testLoadsAll_whenMapCreatedInEager():
```
14:16:31,712 ERROR || - [MapKeyLoader$LoadFinishedFuture] cached17 - Failed asynchronous execution of execution callback: com.hazelcast.map.impl.operation.PartitionCheckIfLoadedOperation$CallbackResponseSender@1656a7d4for call class com.hazelcast.map.impl.MapKeyLoader$LoadFinishedFuture
com.hazelcast.spi.exception.ResponseAlreadySentException: NormalResponse already responseReceived for callback: Invocation{serviceName='hz:impl:mapService', op=com.hazelcast.map.impl.operation.PartitionCheckIfLoadedOperation{serviceName='hz:impl:mapService', partitionId=19, replicaIndex=0, callId=0, invocationTime=1450444591574 (Fri Dec 18 14:16:31 CET 2015), waitTimeout=-1, callTimeout=60000, name=MapLoaderMultiNodeTest}, partitionId=19, replicaIndex=0, tryCount=250, tryPauseMillis=500, invokeCount=1, callTimeout=60000, target=Address[127.0.0.1]:5001, backupsExpected=0, backupsCompleted=0, connection=null}, current-response: : true
	at com.hazelcast.spi.impl.operationservice.impl.Invocation.sendResponse(Invocation.java:360)
	at com.hazelcast.spi.impl.operationservice.impl.PartitionInvocation.sendResponse(PartitionInvocation.java:29)
	at com.hazelcast.spi.Operation.sendResponse(Operation.java:277)
	at com.hazelcast.map.impl.operation.PartitionCheckIfLoadedOperation$CallbackResponseSender.onResponse(PartitionCheckIfLoadedOperation.java:97)
	at com.hazelcast.map.impl.operation.PartitionCheckIfLoadedOperation$CallbackResponseSender.onResponse(PartitionCheckIfLoadedOperation.java:93)
	at com.hazelcast.spi.impl.AbstractCompletableFuture$ExecutionCallbackRunnable.run(AbstractCompletableFuture.java:339)
	at com.hazelcast.util.executor.CachedExecutorServiceDelegate$Worker.run(CachedExecutorServiceDelegate.java:212)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #7056, opened 2015-12-11
Timeout for key-load increased in the MapProxySupport

tombujok:
Fixes #6328

PR #7115, opened 2015-12-18
Operation retry fixed in PartitionCheckIfLoadedOperation

tombujok:
Fix for #7114, regression from #7056
The decoupling of operation result from operation retry mechanism will be followed on in 3.7 - there's a PRD for this. For now it's a fix for MapLoader that doesn't change the default behavior.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionCheckIfLoadedOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionCheckIfLoadedOperation.java
@@ -72,8 +72,17 @@ public class PartitionCheckIfLoadedOperation extends MapOperation implements Par
 
     @Override
     public boolean returnsResponse() {
-        // has to return true in all cases in order to enable the default retry invocation mechanism
-        return true;
+        return !waitForKeyLoad;
+    }
+
+    @Override
+    public void onExecutionFailure(Throwable e) {
+        if (!returnsResponse()) {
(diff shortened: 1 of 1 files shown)


tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
```diff
@@ -341,7 +341,7 @@ private void handleOperationError(Operation operation, Throwable e) {
 
         operation.logError(e);
 
-        if (e instanceof ResponseNotSentException || !operation.returnsResponse()) {
+        if (e instanceof ResponseNotSentException || !operation.isRetryable()) {
```
gurbuzali: Imagine an operation which is not retryable but returnsResponse=true, it will not be able to return response.
I see that currently `isRetryable`=`returnsResponse` except `PartitionCheckIfLoadedOperation` but I think we should address that case too.

tombujok: Thx Ali. Yes, I agree. 
I'm trying to find a minimal subset of changes to fix the PartitionCheckIfLoadedOperation. This change that I propose here will be reverted.

gurbuzali:
:+1:

ahmetmircik:
:+1:

eminn:
Do we need to backport this  ?

Issue #7044, opened 2015-12-10
ClientServiceTest.testClientListener


hasancelik:
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.client.ClientServiceTest.testClientListener(ClientServiceTest.java:219)
```

https://hazelcast-l337.ci.cloudbees.com/view/Quality-Outreach/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/com.hazelcast$hazelcast-client/266/testReport/junit/com.hazelcast.client/ClientServiceTest/testClientListener/

hasancelik:
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.client.ClientServiceTest.testClientListener(ClientServiceTest.java:220)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast/802/testReport/com.hazelcast.topic.impl.reliable/ReliableTopicStressTest/test/

cangencer:
@hasancelik I don't think this failure is related to the test in this issue?

ihsandemir:
@hasancelik last test report is as @cangencer mentioned belongs to ReliableTopicStressTest.

hasancelik:
@cangencer @ihsandemir sorry guys. I updated the link.

PR #7120, opened 2015-12-18
ClientServiceTest sometimes may fail at 

ihsandemir:
Sometimes on ZuluJDK the testClientListener failed on latchAdd.await failing to get all the new client add events. With a small possibility the client shutdown may be faster which may cause the server client service registration to be cleaned up before the client add event is actually delivered. Since the logs are mixed due to parallel test execution it is not easy to verify this but changing the latch wait before client shutdown shall eliminate such a small possibility. I could not generate the failure in local machine.

Fix for https://github.com/hazelcast/hazelcast/issues/7044 .


ihsandemir:
run-lan-run

cangencer:
I don't think it should matter how quickly the clients connect/disconnect, the event should still be published by the server correctly and in correct order.

ihsandemir:
https://github.com/hazelcast/hazelcast/pull/7146 solves the issue.

Issue #6726, opened 2015-11-12
IMap should expose setAsync


LoneRifle:
Users of IMap.set do not have a corresponding async method, unlike put and putAsync. putAsync is not entirely suitable as an async set, since put returns the previous value mapped to the key, and triggers EntryListeners, which may not be desirable. 

It is hence proposed that IMap exposes a dedicated setAsync, to fulfill the contract for set and have the means to do so asynchronously.

PR #7129, opened 2015-12-20
Implement #6726 - Add support for setAsync

LoneRifle:
Declare and implement IMap.setAsync


LoneRifle:
I've tried very hard to run checkstyle on my local machine, but the command line tool refuses to use the hazelcast config/suppressions, and the Eclipse plugin seems to read those but blatantly ignores them (giving me Javadoc violations in com.hazelcast.util.scheduler.ScheduleType despite being on the suppressions list, for instance). This makes it very hard for me to spot any checkstyle problems with my PR, so I apologise in advance.

cangencer:
Can you try doing `mvn verify -Pcheckstyle -DskipTests` ?

LoneRifle:
same outcome - still getting spurious violations.

cangencer:
I ran checkstyle on your branch, and get this error:

`/Users/can/src/hazelcast/hazelcast/src/main/java/com/hazelcast/core/IMap.java:1: File length is 1,525 lines (max allowed is 1,500).`

LoneRifle:
Figured out that Maven does not take well to Windows paths, so will have a look into those. As regards the IMap error - is that the only checkstyle problem? I can look into cutting down the amount of javadoc that I have made.

cangencer:
Yes that was the only checkstyle error. You can also put an exception for IMap file length in `suppressions.xml` as it'll break if we add another method anyway.  As we are in feature freeze for 3.6 release, this can be a candidate for 3.6.1 or 3.7.

LoneRifle:
Okay, I'll add an exception for IMap line count. I'm alright with this being tagged for 3.6.1.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
```diff
@@ -314,6 +314,72 @@ public void testPutAsyncTTL_afterExpireWhenKeyExists() throws Exception {
...
+    @Test
+    public void testSetAsyncTTL_afterExpireWhenKeyExists() throws Exception {
+        IMap<String, String> map = client.getMap(randomString());
+        String key = "Key";
+        String oldValue = "oldValue";
+        String newValue = "Val";
+
+        map.set(key, oldValue);
+        Future<Void> result = map.setAsync(key, newValue, 1, TimeUnit.SECONDS);
+        sleepSeconds(2);
```
pveentjer: Doing sleeps in tests makes a test either fragile or slow. Why do you need to sleep at all? Why not immediately call get?

hazelcast-client-legacy/src/test/java/com/hazelcast/client/map/ClientMapTest.java
```diff
@@ -275,6 +276,33 @@ public void entryEvicted(EntryEvent<String, String> event) {
...
+            public void entryEvicted(EntryEvent<String, String> event) {
+                latch.countDown();
+            }
+        }, true);
+
+        Future<Void> future = map.setAsync("key", "value1", 3, TimeUnit.SECONDS);
+        future.get();
+        assertEquals("value1", map.get("key"));
+
+        assertTrue(latch.await(10, TimeUnit.SECONDS));
```
pveentjer: We have a better method for this; HazelcastTestSupport.assertOpenEventually

LoneRifle: Okay. Let me make the changes that you recommend - these tests were largely
taken from their putAsync equivalents.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/map/ClientMapBasicTest.java
LoneRifle: The sleep is necessary to ensure that the desired number of seconds has elapsed after which the value would be evicted from the Map. I can rework the test such that the test blocks on a latch that is triggered by an eviction listener

LoneRifle:
I've had to revert my test fixes because GitHub insists they create conflicts, but git refuses to acknowledge any conflicts between my branch and hazelcast/master.

LoneRifle:
My rebase has gone horribly wrong. Nuking and starting over.

PR #7131, opened 2015-12-21
Refactored test config creations

ahmetmircik:
This will be used in enterprise


gurbuzali:
:+1:  only test refactorings to be able to extend them on enterprise and some small changes like package relocating.

PR #7135, opened 2015-12-21
[3.6-EA3] At load-all operation, exception should not be wrapped and underlying exception should be propagated

serkan-ozal:
…  Also on cache close, on-going load-all calls should be waited to finish.


serkan-ozal:
merging this one since this is already merged into master

PR #7142, opened 2015-12-21
Use thread name for each time limited test thread to clarify thread dumps

cangencer:
This should help to clarify what thread is running what test in thread dumps
for parallel tests


gurbuzali:
:+1:

PR #7145, opened 2015-12-22
Adds force-start operation to cluster.sh script.

kobalski:
Sample usage : 
`sh cluster.sh --operation force-start`

### DIFF
--- a/hazelcast/src/main/resources/cluster.sh
+++ b/hazelcast/src/main/resources/cluster.sh
@@ -2,8 +2,8 @@
 
 if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
    	echo "parameters : "
-   	echo "	-o, --operation	    : Defines state operation of the script. Operation can be 'get' or 'change'."
-    echo "	-s, --state 	    : Defines new state of the cluster. State can be 'active', 'frozen', 'passive'"
+   	echo "	-o, --operation	    : Executes cluster-wide operation. Operation can be 'get-state','change-state','shutdown','force-start'."
+    echo "	-s, --state 	    : Updates state of the cluster to new state. New state can be 'active', 'frozen', 'passive'"
    	echo "	-p, --port  	    : Defines which port hazelcast is running. Default value is '5701'."
    	echo "	-g, --groupname     : Defines groupname of the cluster. Default value is 'dev'."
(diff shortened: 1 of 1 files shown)


kobalski:
run-lab-run

mdogan:
:+1: 

no need to run or wait for full build, this is only a script change.

hazelcast/src/main/resources/cluster.sh
```diff
@@ -2,7 +2,7 @@
 
 if [ "$1" = "--help" ] || [ "$1" = "-h" ]; then
    	echo "parameters : "
-   	echo "	-o, --operation	    : Defines state operation of the script. Operation can be 'get' or 'change'."
+   	echo "	-o, --operation	    : Defines state operation of the script. Operation can be 'get','change','shutdown','force-start'."
```
mesutcelik: `get-state`,`change-state` would be better names for those operations.

Another naming suggestion :
`Defines state operation of the script` --> `executes cluster-wide operation`

hazelcast/src/main/resources/cluster.sh
```diff
@@ -2,7 +2,7 @@
...
     echo "	-s, --state 	    : Defines new state of the cluster. State can be 'active', 'frozen', 'passive'"
```
mesutcelik: `Defines new state of the cluster` --> `updates cluster state` or something similar.

hazelcast/src/main/resources/cluster.sh
mdogan: :+1:

hazelcast/src/main/resources/cluster.sh
```diff
@@ -131,6 +131,33 @@ if [ "$OPERATION" = "change" ]; then
     exit 0
 fi
 
+if [ "$OPERATION" = "force-start" ]; then
+    echo "You are force starting the cluster. Please make sure you provide valid user/pass."
```
mesutcelik: `You are force starting the cluster`. Any better explanation for this @mdogan @metanet ?

mesutcelik: `force-start makes cluster operational if cluster start is blocked by problematic members.` could be an option

kobalski:
Merging this, PR will update force update info message with direct commit.

PR #7147, opened 2015-12-22
Log and suppress phone-home scheduling rejection

mdogan:
PhoneHome scheduling is rejected if node fails to start. This rejection
exception suppresses actual exception behind the failure of node start.


kobalski:
:+1:

emrahkocaman:
:+1:

PR #7152, opened 2015-12-22
Add support for read-through

snicoll:
Update HazelcastCache to implement a new method introduced in Spring Framework 4.3.

See https://jira.spring.io/browse/SPR-9254 for more details.

You'll need our snapshot repository for the time being if you want to build this PR. Adding the following to your main pom and running `mvn package -PspringSnapshot` should do the trick in the meantime.

```xml
<profile>
	<id>springSnapshot</id>
	<repositories>
		<repository>
			<id>spring-snapshots</id>
			<name>Spring Snapshots</name>
			<url>http://repo.spring.io/snapshot</url>
			<snapshots>
				<enabled>true</enabled>
			</snapshots>
		</repository>
	</repositories>
</profile>

### DIFF
--- a/hazelcast-spring/pom.xml
+++ b/hazelcast-spring/pom.xml
@@ -33,7 +33,7 @@
         <!-- needed for checkstyle/findbugs -->
         <main.basedir>${project.parent.basedir}</main.basedir>
 
-        <org.springframework.version>4.0.2.RELEASE</org.springframework.version>
+        <org.springframework.version>4.3.0.RELEASE</org.springframework.version>
         <jsr250.api.version>1.0</jsr250.api.version>
         <hazelcast.latest.version>3.6</hazelcast.latest.version>
     </properties>
--- a/hazelcast-spring/src/main/java/com/hazelcast/spring/cache/HazelcastCache.java
+++ b/hazelcast-spring/src/main/java/com/hazelcast/spring/cache/HazelcastCache.java
@@ -24,6 +24,7 @@ import org.springframework.cache.Cache;
 import org.springframework.cache.support.SimpleValueWrapper;
 
 import java.io.IOException;
+import java.util.concurrent.Callable;
 
 /**
  * @author mdogan 4/3/12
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by jhoeller** [comment]
Note that this will remain compatible with earlier Spring versions at runtime even when compiled against 4.3 that way. The new ´get(key, valueLoader)´ method simply won't be usable there but the rest still works as before. The new method signature doesn't declare any 4.3-specific types either, so it also won't fail if some code reflectively introspects the ´HazelcastCache´ class on older Spring versions.

**Comment by emrahkocaman** [comment]
verify

**Comment by snicoll** [comment]
I didn't add the repository definition in the PR (see my comment).

**Comment by emrahkocaman** [comment]
Uh, got it. I think we need to park this PR till the release of Spring Framework 4.3. 
Other than that, I'm fine with it. Thanks for your effort @snicoll :+1:

**Comment by noctarius** [comment]
Hey @snicoll any release date for Spring 4.3? Seen there's RC2 out :)

**Comment by snicoll** [comment]
Current date is [June 1st](https://jira.spring.io/browse/SPR/fixforversion/15453/?selectedTab=com.atlassian.jira.plugins.jira-development-integration-plugin:release-report-tabpanel). We're on track.

**Comment by snicoll** [comment]
@noctarius 4.3 GA is out!

**Comment by noctarius** [comment]
@emrahkocaman you heard it ;)

**Comment by emrahkocaman** [comment]
verify

File: hazelcast-spring/pom.xml [hunk 1]
```diff
@@ -33,7 +33,7 @@
         <!-- needed for checkstyle/findbugs -->
         <main.basedir>${project.parent.basedir}</main.basedir>
 
-        <org.springframework.version>4.0.2.RELEASE</org.springframework.version>
+        <org.springframework.version>4.3.0.BUILD-SNAPSHOT</org.springframework.version>
```
emrahkocaman: This needs to be updated to '4.3.0.RELEASE'

snicoll:
I rebased and pushed an update with `4.3.0`.

emrahkocaman:
verify

emrahkocaman:
Thanks @snicoll, there are some checkstyle errors, could you please  check the build results?
(You can check locally by running `mvn compile -Pcheckstyle`.)

snicoll:
Fixed

emrahkocaman:
verify

snicoll:
Yup that test fail for me as well and I have a hard time to see the link with my change.

snicoll:
@noctarius is it going to make it in 3.7? When is 3.7 due by the way? Would be nice to be able to upgrade in Boot 1.4 that is due early July.

emrahkocaman:
@snicoll , Looks like this failure is due to a behaviour change in the Spring 4.3.0.
In the `IDummyBean`, `getNull` method is annotated with `@Cacheable("temp")` (https://github.com/hazelcast/hazelcast/blob/master/hazelcast-spring/src/test/java/com/hazelcast/spring/cache/IDummyBean.java#L32)

But it's implementation is annoted with `@Cacheable("null-map")`. (https://github.com/hazelcast/hazelcast/blob/master/hazelcast-spring/src/test/java/com/hazelcast/spring/cache/TestCacheManager.java#L150)

In previous version (4.0.3), `getNull` is cached with the name "null-map", but with version 4.3.0 the name in the interface is used so cache name becomes "temp".

Cause of failure is the difference of the configs of these two maps, "temp" has a map-store configured so values are loaded from that mapstore and method doesn't return null.

I'll fix this and give another try to your PR.

snicoll:
We refined annotation lookup, yes. If you expose the interface, we used that which might explain the difference. That's a bit of an unusual setup (especially adding such annotations on an interface) but I guess you need that for testing.

jhoeller:
Hmm, arguably, the class-level annotation should override the interface-specified one. Let's see what we can do for 4.3.1 here.

emrahkocaman:
verify

emrahkocaman:
LGTBM 👍

mesutcelik:
👍

mesutcelik:
Hi @snicoll @jhoeller ,

We have a new [issue](https://github.com/hazelcast/hazelcast/issues/9023) raised which is conflicting [the comment](https://github.com/hazelcast/hazelcast/pull/7152#issuecomment-166631805) above. `ValueRetrievalException` which is introduced in this PR is available since 4.3 and that is blocking hazelcast version upgrades to 3.7 for people who use the older versions of spring framework.

Can we use something other than `ValueRetrievalException` ?

emrahkocaman:
@snicoll Any comments?

PR #7156, opened 2015-12-23
Create scheduler for locks per partition instead of per node

gurbuzali:
We've converted our scheduler to synchronized due to race problems and this change introduced performance regression on locks (with lease). To overcome this contention we are now creating a scheduler for each partition. 

PS: ReplicatedMap and IQueue already use this approach(per-partition) while using scheduler.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/InternalLockService.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/InternalLockService.java
@@ -26,17 +26,19 @@ public interface InternalLockService extends LockService {
      * at the time of eviction.
      *
      * @param namespace namespace of the lock
+     * @param partitionId partitionId the key falls into
      * @param key locked key
      * @param version version of a lock to evict.
      * @param delayMillis delay in ms
      */
-    void scheduleEviction(ObjectNamespace namespace, Data key, int version, long delayMillis);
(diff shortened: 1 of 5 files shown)


gurbuzali:
weird test failure. `MapSumAggregationTest` failed with `Node failed to start!`
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/705/

gurbuzali:
run-lab-run

mdogan:
:+1:

metanet:
:+1:

pveentjer:
There are other related problems using a scheduler btw.

You get contention on the queue; especially the head if you have multiple threads posting items.

But even worse is that these queues can grow and can lead to resource problems (OOME). Otherwise a queue needs to be capped and either blocking or items need to be discarded; both situations are problematic.

Imho we should reconsider using schedulers for releasing resources.

gurbuzali:
- what do you mean with `There are other related problems using a scheduler btw.` ? can you elaborate?
- you say `You get contention on the queue`. which queue you are talking about ?
- `Apart from that, using schedulers can lead to litter (either queue and/or task objects).` I haven’t introduced scheduler here, it was there since the beginning. And this PR is just changing the scope of the scheduler.
- `Imho we should reconsider using schedulers for releasing resources.` I think this goes to the next release. can you create a PRD for that

PR #7157, opened 2015-12-23
Removed listeners from JMX Beans

muratayan:
Clustered statistics shouldn't be provided from here.

PR #7163, opened 2015-12-24
[DON'T MERGE] [BACKPORT] Removed listeners from JMX Beans

muratayan:
Backport of #7157


muratayan:
See https://github.com/hazelcast/hazelcast/pull/7188 instead

PR #7165, opened 2015-12-24
Client and server member Mock connection port number collision

ihsandemir:
Too many repeated tests or tests using too many connections may cause the number of ports used for the client and the server members to collide. Currently the separation between client and the server port is 1000 (6000-1000) which shall be consumed if too many instances are started and closed. The port number is always incremented. This change shall increase this interval to 40000-5000=35000.


sancar:
:+1:

Donnerbart:
:+1:

PR #7167, opened 2015-12-24
Async puts in cache reflects to near-cache immediately when its `future.get()` returns

### LINKED ISSUE #7155: Async puts should reflect to near-cache immediately when its `future.get()` returns
Currently at client side, when `LocalUpdatePolicy` is `CACHE`, in async puts, entries are put to near-cache in async way from another task when response is received. But with this way, when `future.get()` returns, entry might not be inside near-cache (but it will be there eventually).

serkan-ozal:
… or when its scheduled task executed eventually as it is currently

Fixes #7155

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.cache.impl;
+
+import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.util.Clock;
+
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+
+import static com.hazelcast.util.ExceptionUtil.sneakyThrow;
+
+/**
+ * <p>
+ * A specific {@link ExecutionCallback} base implementation which has this behaviour:
+ * <ul>
+ *      <li>If it has not been called yet, it will be called and it will be waited to finish.</li>
+ *      <li>If it has been called but not finished yet, it will be waited to finish.</li>
+ *      <li>If it has been called and finished already, it will return immediately.</li>
+ * </ul>
+ * </p>
+ *
+ * @param <V> type of the response
+ */
+abstract class OneShotExecutionCallback<V> implements ExecutionCallback<V> {
+
+    private static final AtomicIntegerFieldUpdater CALL_STATE_UPDATER
+            = AtomicIntegerFieldUpdater.newUpdater(OneShotExecutionCallback.class, "callState");
+
+    private static final int NOT_CALLED = 0;
+    private static final int CALL_IN_PROGRESS = 1;
+    private static final int CALL_FINISHED = 2;
+
+    private volatile int callState = NOT_CALLED;
+
+    @Override
+    public final void onResponse(V response) {
+        onInternal(response, false, Long.MAX_VALUE);
+    }
+
+    @Override
+    public final void onFailure(Throwable t) {
+        onInternal(t, true, Long.MAX_VALUE);
+    }
+
+    void onResponse(V response, long finishTime) {
+        onInternal(response, false, finishTime);
+    }
+
+    void onFailure(Throwable t, long finishTime) {
+        onInternal(t, true, finishTime);
+    }
+
+    protected abstract void onResponseInternal(V response);
+    protected abstract void onFailureInternal(Throwable t);
+
+    private void onInternal(Object obj, boolean failure, long finishTime) {
+        if (callState != CALL_FINISHED) {
+            // Call not finished yet so be sure that it finished by calling it or waiting it to finish
+            if (CALL_STATE_UPDATER.compareAndSet(this, NOT_CALLED, CALL_IN_PROGRESS)) {
+                // Will be called only by this thread
+                try {
+                    if (failure) {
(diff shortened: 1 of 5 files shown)


serkan-ozal:
@sancar could you please have a look?

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -401,9 +401,9 @@ protected Object putInternal(final K key, final V value, final ExpiryPolicy expi
             }
         } else {
             if (nearCache != null || statisticsEnabled) {
-                future.andThen(new ExecutionCallback<Object>() {
```
ahmetmircik: Could it be a solution to use same thread for executor service? Something like this:

```
// executes tasks in current thread.
public class SameThreadExecutor implements Executor {
   public void execute(Runnable task) 
}

future.andThen(new ExecutionCallback(), sameThreadExecutor);

Comment (serkan-ozal) [inline hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java]: Sorry i couldn't understand. Can you explain more detailed ?

Comment (sancar) [inline hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java]: We don't want to use same thread. That thread is ResponseThread in this case. 
We want to run on user thread. And only way seems to be by overwriting future that we gave to user.

Comment (sancar) [inline hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java]: @serkan-ozal @ahmetmircik 
That has been said, this solution although is doing all we want, seems complicated. 
Question I am trying to answer is that is trying to offload the work to user thread worth putting a complex code to system ?

Comment (serkan-ozal) [inline hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java]: Yes, adds conplexity but just for cache async put. And you know this request comes from customer. If they are ok without this, I am OK

**Comment by sancar** [comment]
:+1:

**Comment by sancar** [comment]
@serkan-ozal We probably need to update legacy-client also.

File: hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.cache.impl;
+
+import com.hazelcast.core.ExecutionCallback;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+abstract class OneShotExecutionCallback<V> implements ExecutionCallback<V> {
```
ahmetmircik: Can you  add some javaDoc which describes responsibility of this class?

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.cache.impl;
+
+import com.hazelcast.core.ExecutionCallback;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+abstract class OneShotExecutionCallback<V> implements ExecutionCallback<V> {
```
ahmetmircik: javaDoc?

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
```diff
@@ -0,0 +1,72 @@
...
+                        onResponseInternal((V) obj);
+                    }
+                } finally {
+                    // Call finished
+                    callState.set(CALL_FINISHED);
+                }
+            } else {
+                // Call has been started by another thread but not finished yet.
+                // So wait until it finished by busy-spin.
+                for (;;) {
```
ahmetmircik: What about adding a time-out here in order to prevent possible indefinite waiting?

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
serkan-ozal: sure

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
serkan-ozal: sure

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+abstract class OneShotExecutionCallback<V> implements ExecutionCallback<V> {
+
+    private static final int NOT_CALLED = 0;
+    private static final int CALL_IN_PROGRESS = 1;
+    private static final int CALL_FINISHED = 2;
+
+    private final AtomicInteger callState = new AtomicInteger(0);
```
pveentjer: Why not an AtomicLongFieldUpdater? Now you create more litter than needs to be.

pveentjer:
A lot more tests need to be added. We had many problems with future's in the past and there is hardly any coverage. I'm going to suggest to get this fixed in 3.7 and to push this through an official PRD. It is too easy to cause problems here.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
serkan-ozal: yes, `AtomicLongFieldUpdater` makes sense because we may create many `OneShotExecutionCallback`

serkan-ozal:
I am thinking of adding a few more tests but as I said this request comes from a customer this is why I have sent this PR before 3.6 release and why I have make this PR effect only cache async put.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
serkan-ozal: done

hazelcast-client-legacy/src/main/java/com/hazelcast/client/cache/impl/OneShotExecutionCallback.java
serkan-ozal: done

sancar:
:+1:

PR #7186, opened 2015-12-25
Adds fail earlier behaviour and prevents creation of proxy in case of…

ahmetmircik:
… NATIVE in-memory-format usage

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapRemoteService.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapRemoteService.java
@@ -22,8 +22,10 @@ import com.hazelcast.map.impl.proxy.MapProxyImpl;
 import com.hazelcast.map.impl.proxy.NearCachedMapProxyImpl;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.RemoteService;
+
 import java.util.Map;
 
+import static com.hazelcast.map.impl.MapConfigValidator.checkInMemoryFormat;
 import static com.hazelcast.map.impl.MapService.SERVICE_NAME;
 
 /**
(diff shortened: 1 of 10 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/MapRemoteService.java
```diff
@@ -44,8 +47,15 @@
     @Override
     public DistributedObject createDistributedObject(String name) {
         MapConfig mapConfig = nodeEngine.getConfig().findMapConfig(name);
+        checkInMemoryFormat(mapConfig.getInMemoryFormat());
 
         if (mapConfig.isNearCacheEnabled()) {
+
+            NearCacheConfig nearCacheConfig = mapConfig.getNearCacheConfig();
+            if (nearCacheConfig != null) {
```
jerrinot: can this ever be false? It would mean NearCacheIsEnabled, but config is null? Is this possible?

hazelcast/src/main/java/com/hazelcast/map/impl/MapRemoteService.java
```diff
@@ -44,8 +47,15 @@
     @Override
     public DistributedObject createDistributedObject(String name) {
         MapConfig mapConfig = nodeEngine.getConfig().findMapConfig(name);
+        checkInMemoryFormat(mapConfig.getInMemoryFormat());
```
jerrinot: how is this overridden in Enterprise context?

hazelcast/src/main/java/com/hazelcast/map/impl/MapRemoteService.java
ahmetmircik: fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/MapRemoteService.java
ahmetmircik: Enterprise has its own proxies, means no call to oss `createDistributedObject` method, so there should not be any problem.

jerrinot: ok

jerrinot:
:+1:

eminn:
:+1:

PR #7188, opened 2015-12-25
Made JMX listeners optional

muratayan:
We are removing these listeners in 3.6 which will most probably break the compatibility. However, since it's a major performance overhead, we need to get rid of them on 3.5.5 as well. 

This way, unless user specifically gives the `GroupProperty` `ENABLE_JMX_DETAILED("hazelcast.jmx.detailed", true)`, they are not created, and total stats return `0` even though API stays the same.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/jmx/ListMBean.java
+++ b/hazelcast/src/main/java/com/hazelcast/jmx/ListMBean.java
@@ -19,6 +19,7 @@ package com.hazelcast.jmx;
 import com.hazelcast.core.IList;
 import com.hazelcast.core.ItemEvent;
 import com.hazelcast.core.ItemListener;
+
 import java.util.concurrent.atomic.AtomicLong;
 
 import static com.hazelcast.util.EmptyStatement.ignore;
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/jmx/ListMBean.java
```diff
@@ -29,27 +30,34 @@
 @ManagedDescription("IList")
 public class ListMBean extends HazelcastMBean<IList<?>> {
 
+    private boolean totalEnabled;
```
emrahkocaman: It should be `final`

hazelcast/src/main/java/com/hazelcast/jmx/ListMBean.java
```diff
@@ -29,27 +30,34 @@
...
 
+    private final boolean totalEnabled;
+
     private AtomicLong totalAddedItemCount = new AtomicLong();
     private AtomicLong totalRemovedItemCount = new AtomicLong();
     private final String registrationId;
 
     protected ListMBean(IList<?> managedObject, ManagementService service) {
         super(managedObject, service);
+        totalEnabled = service.instance.node.groupProperties.ENABLE_JMX_DETAILED.getBoolean();
```
emrahkocaman: I think the field `totalEnabled` and this line should be moved to `HazelcastMBean` class. This way we can get rid of code duplication in all other `..MBean` impls.

emrahkocaman:
:+1:

PR #7195, opened 2015-12-28
Distinguish map/cache service reset, shutdown and single map/cache destroy

mdogan:
Added a flag to map/cache clear/close operations to denote service shutdown.
It's possible to avoid some operations during service shutdown,
which are required when map/cache is closed on normal runtime.

_This is required to fix issue: https://github.com/hazelcast/hazelcast-enterprise/issues/613_


ahmetmircik:
:+1:

gurbuzali:
:+1:

Issue #7194, opened 2015-12-28
ClientMapNearCacheTest.testAfterSubmitToKeyKeyIsInvalidatedFromNearCache[batchInvalidationEnabled:true]


hasancelik:
```
java.lang.AssertionError: expected:<999> but was:<900>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheTest.assertThatOwnedEntryCountEquals(ClientMapNearCacheTest.java:986)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-OracleJDK8/com.hazelcast$hazelcast-client/664/testReport/junit/com.hazelcast.client.map.impl.nearcache/ClientMapNearCacheTest/testAfterSubmitToKeyKeyIsInvalidatedFromNearCache_batchInvalidationEnabled_true_/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/view/Quality-Outreach/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/com.hazelcast$hazelcast-client-legacy/282/testReport/junit/com.hazelcast.client.map.impl.nearcache/ClientMapNearCacheTest/testAfterExecuteOnKeyKeyIsInvalidatedFromNearCache_batchInvalidationEnabled_true_/

PR #7201, opened 2015-12-28
changed test structure of ClientMapNearCacheTest

hasancelik:
possible fix for #7194


cangencer:
what is the reason for this change? What difference will it make to make puts on the server rather than the client?

PR #7203, opened 2015-12-28
Fixes repeatedly checks made from InvocationMonitor for timeout

### LINKED ISSUE #7170: InvocationMonitor checks for the same invocations continuously
We observed in two different runs that InvocationMonitor is checking for the same invocations continuously. You can see the logs of one in the attachment. 

There are continuous `Handled 3 invocation timeouts and 0 backupTimeouts` logs while all of the three invocations checks are returning `is-executing': false`.

It seems the invocations in the `invocationRegistry` left deregistered for some reason.
[hazelcast_log.txt](https://github.com/hazelcast/hazelcast/files/71751/hazelcast_log.txt)

eminn:
Problematic scenario :
If an invocation with callId 1 is retried twice for any reason, two new innovations created and registered to invocation registry with callId’s 2 and 3 respectively. Both new invocations are sharing the same operation.

When one of the new invocations, say the one with callId 2 finishes, it de-registers itself from the invocation registry. When doing the de-registration it sets the shared operation’s callId to 0. After that when the invocation with the callId 3 completes, it tries to de-register itself from invocation registry but fails to do so since the invocation callId and the callId on the operation are not matching anymore. 

When InvocationMonitor thread kicks in, it sees that there is an invocation in the registry, and asks whether invocation is finished or not. Even if the remote node replies with invocation is timed out, It can’t be de-registered from the registry because of aforementioned non-matching callId scenario.

Workaround:
When InvocationMonitor kicks in, it will do a check for invocations that are completed but their callId's are not matching with their operations. If any invocation found for that type, it is removed from the invocation registry.

Actual solution to this problem could be re-designing the invocations by making only one retry can be done at the same time.

Fixes #7170

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -504,13 +504,14 @@ abstract class Invocation implements OperationResponseHandler, Runnable {
         long maxCallTimeout = invocationFuture.getMaxCallTimeout();
         long expirationTime = op.getInvocationTime() + maxCallTimeout;
 
+        boolean done = invocationFuture.isDone();
         boolean hasResponse = pendingResponse != null;
         boolean hasWaitingThreads = invocationFuture.getWaitingThreadsCount() > 0;
         boolean notExpired = maxCallTimeout == Long.MAX_VALUE
                 || expirationTime < 0
                 || expirationTime >= Clock.currentTimeMillis();
 
-        if (hasResponse || hasWaitingThreads || notExpired) {
+        if (hasResponse || hasWaitingThreads || notExpired || done) {
             return false;
         }
 
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
@@ -30,7 +30,9 @@ import com.hazelcast.spi.impl.operationexecutor.OperationHostileThread;
 import com.hazelcast.util.Clock;
 import com.hazelcast.util.EmptyStatement;
 import com.hazelcast.util.counters.SwCounter;
-
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import static com.hazelcast.instance.OutOfMemoryErrorDispatcher.inspectOutputMemoryError;
@@ -40,7 +42,7 @@ import static com.hazelcast.util.counters.SwCounter.newSwCounter;
 /**
  * The InvocationMonitor monitors all pending invocations and determines if there are any problems like timeouts. It uses the
  * {@link InvocationRegistry} to access the pending invocations.
- *
+ * <p/>
  * An experimental feature to support debugging is the slow invocation detector. So it can log any invocation that takes
  * more than x seconds. See {@link GroupProperty#SLOW_INVOCATION_DETECTOR_THRESHOLD_MILLIS} for more information.
  */
@@ -99,10 +101,10 @@ public class InvocationMonitor {
 
     /**
      * The MonitorThread iterates over all pending invocations and sees what needs to be done
-     *
+     * <p/>
      * But it should also check if a 'is still running' check needs to be done. This removed complexity from
      * the invocation.waitForResponse which is too complicated too understand.
-     *
+     * <p/>
      * This class needs to implement the {@link OperationHostileThread} interface to make sure that the OperationExecutor
      * is not going to schedule any operations on this task due to retry.
      */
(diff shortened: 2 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -504,13 +504,14 @@ boolean checkInvocationTimeout() {
...
 
+        boolean done = invocationFuture.isDone();
         boolean hasResponse = pendingResponse != null;
         boolean hasWaitingThreads = invocationFuture.getWaitingThreadsCount() > 0;
         boolean notExpired = maxCallTimeout == Long.MAX_VALUE
                 || expirationTime < 0
                 || expirationTime >= Clock.currentTimeMillis();
 
-        if (hasResponse || hasWaitingThreads || notExpired) {
+        if (hasResponse || hasWaitingThreads || notExpired || done) {
```
pveentjer: So apparently we have invocations that have received a response (otherwise future.set would not have been called), but the invocation is not deregistered from the registry?

pveentjer: Apart from that, I don't think the 'isDone' check should be repeated in the checkInvocationTimeout method. You already do that in the isInvocationDone method.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -172,6 +181,11 @@ private void scan() {
             log(backupTimeouts, invocationTimeouts);
         }
 
+        private boolean isInvocationDone(Invocation invocation) {
+            long callId = invocation.op.getCallId();
+            return callId == 0 && invocation.invocationFuture.isDone();
```
pveentjer: Why is the 0 check needed?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_RetryTest.java
```diff
@@ -69,6 +74,50 @@ public void whenTargetMemberDiesThenOperationAbortedWithMembersLeftException() t
...
+        InvocationFuture future = (InvocationFuture) operationService.invokeOnTarget(null
+                , op, remoteNodeEngine.getThisAddress());
+        Field invocationField = InvocationFuture.class.getDeclaredField("invocation");
+        invocationField.setAccessible(true);
+        Invocation invocation = (Invocation) invocationField.get(future);
+
+        invocation.notifyError(new RetryableHazelcastException());
+        invocation.notifyError(new RetryableHazelcastException());
+
+        sleepSeconds(10);
```
pveentjer: I would use an assertTrueEventually. sleepSeconds only slows down the test for no good reason.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -172,6 +183,14 @@ private void scan() {
             log(backupTimeouts, invocationTimeouts);
         }
 
+        private boolean isCallIdsMatch(Long callId, Invocation invocation) {
```
pveentjer: Not very important but I would call this method callIdMatches. And why don't you use a primitive long?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -157,6 +158,10 @@ public int size() {
         return invocations.values();
     }
 
+    public Iterator<Map.Entry<Long, Invocation>> registryIterator() {
```
pveentjer: why don't you return an entry set and let the method be called entrySet?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -172,6 +183,14 @@ private void scan() {
...
+            return callId.equals(invocation.op.getCallId());
+        }
+
+        private boolean isInvocationDone(Invocation invocation) {
```
pveentjer: I would rename it to 'isDone' since the argument already explains the type.

pveentjer:
Please add one or more explicit tests for the behavior on the InvocationRegistry/Invocation. The current test is very 'big' and this stuff can easily be tested here.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
eminn: It is checked in the context of `stuck` case in combination of callId matching. This is just for if the invocation is done, do not make any attempt to ask if it's still running.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -172,6 +185,14 @@ private void scan() {
             log(backupTimeouts, invocationTimeouts);
         }
 
+        private boolean callIdMatches(Long callId, Invocation invocation) {
+            return callId.equals(invocation.op.getCallId());
```
mdogan: Can you replace `equals` with primitive equality check?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -151,10 +153,21 @@ private void scan() {
...
+            Iterator<Map.Entry<Long, Invocation>> iterator = invocations.iterator();
+            while (iterator.hasNext()) {
                 if (shutdown) {
                     return;
                 }
+                Map.Entry<Long, Invocation> entry = iterator.next();
+                Long callId = entry.getKey();
+                Invocation invocation = entry.getValue();
+
+                if (!callIdMatches(callId, invocation) && isDone(invocation)) {
```
mdogan: Can you explain in a comment what is done here, what's the reason behind this check and why invocation is removed?

pveentjer: The 'phantom' invocation is removed. So if you get 2 concurrent retries the same invocation is registered under 2 different calld ids' (first retry callid1 and second retry calldi2) So you can potentially end up with phantom invocations in your registry.

mdogan:
:+1:

gurbuzali:
:+1:

pveentjer:
I think there is a more solid approach. However it is still mostly an idea and I'm not that sure of the consequences.

Add new volatile transient boolean field is added to the operation called e.g. retrying. This field has a get/set/cas available.

When a response requiring a retry comes in, this field is cas'ed. 
* If failure, somebody else his problem since the system is already retrying.
* If success, then we can schedule a retry. When the invocation is actually retried, we update the target, call id etc etc and finally we clear the retry field.

If the member left event comes in and scans all invocations (we need to make sure invocation.target is also volatile) it will try to retry the invocations with the desired target. It will do so by doing a cas on this retrying field. 
* if it fails. somebody else his problem
* if it succeeds, it first needs to recheck if the target is still desired. If no, skip it. If the target is valid, than we can initiate the retry process and eventually the retry field is unset.

If the member left event comes in first instead of last, the system should first check if the call-id of the operation matches that of the response. If it doesn't match, the operation has been retried and there is no need to process the response. However this works fine for a response that didn't change anything, but could be problematic for a regular response.

Using this approach there will mutual exclusive behavior when retrying and you can't get duplicate retries. 
Using the current approach the 'phantom' registration is removed. However since it only tried to heal the invocation registry, it doesn't prevent the operation to be executed twice.

PR #7211, opened 2015-12-30
Improve resource cleanup logic in MapService.shutdown

mdogan:
Moved shutdown resource cleanup handling logic to RecordStore.


mdogan:
@ahmetmircik: FYI

PR #7215, opened 2015-12-30
Client Partition Service Update Problem

ihsandemir:
No need to submit the partition refresh response processing for the client into a separate thread using the executor. If it is offloaded to such an executor then ClientRegressionWithMockNetworkTest.testDeadlock_WhenDoingOperationFromLifecycleListener test fails.

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java
@@ -18,10 +18,8 @@ package com.hazelcast.client.spi.impl;
 
 import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.client.spi.ClientClusterService;
-import com.hazelcast.client.spi.ClientExecutionService;
 import com.hazelcast.client.spi.ClientPartitionService;
 import com.hazelcast.core.ExecutionCallback;
-import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.Member;
 import com.hazelcast.core.Partition;
 import com.hazelcast.logging.ILogger;
@@ -35,6 +33,7 @@ import com.hazelcast.partition.client.PartitionsResponse;
 import com.hazelcast.util.EmptyStatement;
 
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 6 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java
```diff
@@ -148,6 +136,7 @@ private ClientInvocationFuture getPartitionsFrom(Connection connection) {
     }
 
     private boolean processPartitionResponse(ClientGetPartitionsCodec.ResponseParameters response) {
+        LOGGER.finest("Processing partition response.");
```
cangencer: is this extra logging leftover from debugging or was meant to be added?

hazelcast-client-legacy/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java
```diff
@@ -71,15 +68,10 @@ public void start() {
...
 
     public void refreshPartitions() {
-        if (!updating.compareAndSet(false, true)) {
-            return;
-        }
         ClientExecutionService executionService = client.getClientExecutionService();
         try {
-            ICompletableFuture future = executionService.submit(new RefreshTask());
-            future.andThen(refreshTaskCallback);
+            executionService.submit(new RefreshTask());
```
cangencer: the RefreshTask is using the User Executor rather than the internal executor. Should that be the case?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java
ihsandemir: We added this log intentionally. There was no log for partition updates.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java
```diff
@@ -69,21 +68,19 @@ public ClientPartitionServiceImpl(HazelcastClientInstanceImpl client) {
...
-        }
-        ClientExecutionService executionService = client.getClientExecutionService();
+        ClientExecutionServiceImpl executionService = (ClientExecutionServiceImpl) client.getClientExecutionService();
         try {
-            ICompletableFuture future = executionService.submit(new RefreshTask());
-            future.andThen(refreshTaskCallback);
-
+            // Use internal execution service for all partition refresh process (Do not use the user executor thread)
+            ExecutorService internalExecutor = executionService.getInternalExecutor();
+            executionService.submitInternal(new RefreshTask(internalExecutor));
```
cangencer: shouldn't there be a future.andThen() here?

ihsandemir: No, we off loaded it to RefreshTask.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientPartitionServiceImpl.java
```diff
@@ -130,10 +127,7 @@ private boolean getPartitions() {
             Future<ClientMessage> future = getPartitionsFrom(connection);
             ClientMessage responseMessage = future.get();
             ClientGetPartitionsCodec.ResponseParameters response = ClientGetPartitionsCodec.decodeResponse(responseMessage);
-            if (response == null) {
-                return false;
-            }
-            return processPartitionResponse(response);
+            return response != null && processPartitionResponse(response);
```
cangencer: I think the code was easier to read before (if not more verbose)

ihsandemir: idea code review tool suggested this change :) i also prefer the older version.

sancar: Idea suggests all crazy things. Don't listen it . I also prefer the older version.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/ClientRegressionWithMockNetworkTest.java
```diff
@@ -616,17 +615,23 @@ public void entryAdded(EntryEvent<Object, Object> event) {
     public void testDeadlock_WhenDoingOperationFromLifecycleListener() {
```
sancar: Can we add the modified test as new test ? The old one with getting the partition table as on-demand also a valuable test to have. I think we need to have both versions at the same time.

ihsandemir: Ok, added a it as a new test with name testDeadlock_WhenDoingOperationFromLifecycleListenerWithInitialPartitionTable.

sancar:
:+1:

PR #7218, opened 2015-12-31
Extract migration queue from InternalPartitionServiceImpl

### LINKED ISSUE #7191: ClusterRollingRestartTest.test_rollingRestart[clusterState:FROZEN,partitionAssignmentType:NEVER]
```
java.lang.AssertionError: Cluster size is not correct. Expected: 3 Actual: 1
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSize(HazelcastTestSupport.java:694)
	at com.hazelcast.test.HazelcastTestSupport$10.run(HazelcastTestSupport.java:713)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:835)
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSizeEventually(HazelcastTestSupport.java:709)
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSizeEventually(HazelcastTestSupport.java:704)
	at com.hazelcast.cluster.impl.ClusterRollingRestartTest.test_rollingRestart(ClusterRollingRestartTest.java:86)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.8/78/com.hazelcast$hazelcast/testReport/com.hazelcast.cluster.impl/ClusterRollingRestartTest/test_rollingRestart_clusterState_FROZEN_partitionAssignmentType_NEVER_/

metanet:
Extract management of migrations tasks into a separate class. It makes it easier to manage migration state of the partition service. It also solves the race problem occurs during starting a migration task polled from the migration queue. 

Fixes #7191

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
@@ -86,11 +86,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Future;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
@@ -148,8 +146,8 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
 
     @Probe
     private final AtomicInteger stateVersion = new AtomicInteger();
-    @Probe(name = "migrationQueueSize")
-    private final BlockingQueue<Runnable> migrationQueue = new LinkedBlockingQueue<Runnable>();
+
+    private final MigrationQueue migrationQueue = new MigrationQueue();
     private final AtomicBoolean migrationAllowed = new AtomicBoolean(true);
     @Probe(name = "lastRepartitionTime")
     private final AtomicLong lastRepartitionTime = new AtomicLong();
@@ -1254,8 +1252,8 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
 
     @Override
     public boolean hasOnGoingMigrationLocal() {
-        return !activeMigrations.isEmpty() || !migrationQueue.isEmpty()
-                || migrationThread.isMigrating();
+        return !activeMigrations.isEmpty() || migrationQueue.isNonEmpty()
+                || migrationQueue.hasMigrationTasks();
     }
 
     private boolean isReplicaInSyncState() {
@@ -1650,6 +1648,7 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
     }
 
     @Override
+    @Probe(name = "migrationQueueSize")
     public long getMigrationQueueSize() {
         return migrationQueue.size();
     }
@@ -1809,7 +1808,7 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
         @Override
         public void run() {
             if (node.isMaster() && node.getState() == NodeState.ACTIVE) {
-                if (!migrationQueue.isEmpty() && isMigrationAllowed()) {
+                if (migrationQueue.isNonEmpty() && isMigrationAllowed()) {
                     logger.info("Remaining migration tasks in queue => " + migrationQueue.size());
                 }
                 publishPartitionRuntimeState();
@@ -1924,10 +1923,7 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
         private void migratePartitionToNewOwner(int partitionId, Address[] replicas, Address currentOwner, Address newOwner) {
             MigrationInfo info = new MigrationInfo(partitionId, currentOwner, newOwner);
             MigrateTask migrateTask = new MigrateTask(info, replicas);
-            boolean offered = migrationQueue.offer(migrateTask);
-            if (!offered) {
-                logger.severe("Failed to offer: " + migrateTask);
-            }
+            migrationQueue.add(migrateTask);
         }
 
         private void assignNewPartitionOwner(int partitionId, Address[] replicas, InternalPartitionImpl currentPartition,
@@ -1947,7 +1943,7 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
         }
     }
 
-    private class MigrateTask implements Runnable {
+    class MigrateTask implements Runnable {
         final MigrationInfo migrationInfo;
         final Address[] addresses;
 
@@ -2077,7 +2073,6 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
 
     private class MigrationThread extends Thread implements Runnable {
         private final long sleepTime = max(250L, partitionMigrationInterval);
-        private volatile boolean migrating;
 
         MigrationThread(Node node) {
             super(node.getHazelcastThreadGroup().getInternalThreadGroup(),
@@ -2102,7 +2097,7 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
         }
 
         private void doRun() throws InterruptedException {
-            for (; ; ) {
+            for (; ;) {
                 if (!isMigrationAllowed()) {
                     break;
                 }
@@ -2118,8 +2113,7 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
             }
             boolean hasNoTasks = migrationQueue.isEmpty();
             if (hasNoTasks) {
-                if (migrating) {
-                    migrating = false;
+                if (!migrationQueue.hasMigrationTasks()) {
                     logger.info("All migration tasks have been completed, queues are empty.");
                 }
                 evictCompletedMigrations();
@@ -2130,15 +2124,18 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
         }
 
         boolean processTask(Runnable r) {
-            if (r == null || isInterrupted()) {
-                return false;
-            }
-            migrating = (r instanceof MigrateTask);
             try {
+                if (r == null || isInterrupted()) {
+                    return false;
+                }
+
                 r.run();
             } catch (Throwable t) {
                 logger.warning(t);
+            } finally {
+                migrationQueue.afterTaskCompletion(r);
             }
+
             return true;
         }
 
@@ -2147,9 +2144,6 @@ public class InternalPartitionServiceImpl implements InternalPartitionService, M
             interrupt();
         }
 
-        boolean isMigrating() {
-            return migrating;
-        }
     }
 
     private static final class InternalPartitionListener implements PartitionListener {
--- a/hazelcast/src/main/java/com/hazelcast/partition/impl/MigrationQueue.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/impl/MigrationQueue.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.partition.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Manages migration tasks and migration status flag for {@link InternalPartitionServiceImpl} safely. Once a migration task
+ * is added to the queue, queue has to be notified via {@link MigrationQueue#afterTaskCompletion(Runnable)} after its execution.
+ */
+class MigrationQueue {
+
+    private final AtomicInteger migrateTaskCount = new AtomicInteger();
+
+    private final BlockingQueue<Runnable> queue = new LinkedBlockingQueue<Runnable>();
+
+    public void add(Runnable task) {
+        if (task instanceof InternalPartitionServiceImpl.MigrateTask) {
+            migrateTaskCount.incrementAndGet();
+        }
+
+        queue.offer(task);
+    }
+
+    public Runnable poll(int timeout, TimeUnit unit)
+            throws InterruptedException {
+        return queue.poll(timeout, unit);
+    }
+
+    public void clear() {
+        List<Runnable> sink = new ArrayList<Runnable>();
+        queue.drainTo(sink);
+
+        for (Runnable task : sink) {
+            afterTaskCompletion(task);
+        }
+    }
+
+    public void afterTaskCompletion(Runnable task) {
+        if (task instanceof InternalPartitionServiceImpl.MigrateTask) {
+            if (migrateTaskCount.decrementAndGet() < 0) {
+                throw new IllegalStateException();
+            }
+        }
+    }
+
+    public boolean hasMigrationTasks() {
+        return migrateTaskCount.get() > 0;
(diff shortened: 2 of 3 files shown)


metanet:
run-lab-run

mdogan:
:+1:

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -2102,15 +2102,22 @@ public void run() {
...
-            for (; ; ) {
+            for (; ;) {
                 if (!isMigrationAllowed()) {
                     break;
                 }
-                Runnable r = migrationQueue.poll(1, TimeUnit.SECONDS);
+
+                Runnable r = migrationQueue.peek();
                 if (r == null) {
+                    Thread.sleep(100);
```
pveentjer: We are not checking 10 more frequent than before? Because initially the frequency was 1/second and now 10/second.

pveentjer:
Could you add a short explanation of the race?

metanet:
MigrationThread polls a migration task from migrationQueue and sets the migrating flag but this is not an atomic operation. hasOnGoingMigrationLocal() returns false when a migration task is polled from the queue but the flag is not set yet. This fix first sets the flag before polling the task from the queue.

metanet:
@pveentjer pr description is updated

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
pveentjer: @metanet ping.

metanet: Before this change, ``poll()`` is used with 1 second of timeout. So if there is no migration task, it will be checked every 1 second. Once a migration task is put into the queue, it will be polled immediately. I used 100 ms sleep duration to get the migration task without too much delay.

pveentjer:
Sorry for my stupidity; so apparently another thread than the MigrationThread is suffering from this non atomic behavior. And this thread makes the wrong conclusion; it thinks there are no migrations since the flag isn't set, but actually there are migrations. Is my understanding correct?

Which thread is this other thread?
What are the consequences of this race?

metanet:
Any code that checks about the safety of the partitions may think that there is no migration while there is a migration task to be runned. For example, instance shut down.

pveentjer:
But if the calling code would have called e.g. just before the MigrationThread would have done its work, it would also see a 'stale' value since the flag hasn't been set. So why is it important that the other threads needs to see 'the most actual' value?

metanet:
There are 2 cases:
1) There is migration task in the queue
2) Migration task is polled from the queue and being run

For both of these cases, ``hasOnGoingMigrationLocal()``, which can be called by any other thread,must return true. Between step 1 and step 2, there is a race which is fixed by this test.

pveentjer:
Sorry for my never ending stupidity. Would it not make sense to have an AtomicLong that is incremented with every migration task and decremented when a tasks is completed? To check if there are ongoing migration, check if the value > 0. 

My lack of understanding the threading for this particular section is causing problems.

metanet:
Np. That was our other solution but it creates its own problems because for some parts of the codes, migrationQueue is updated within a lock; for some other parts, it is updated without a lock. 

We will refactor these parts while working on partition service for 3.7

pveentjer:
So something like this:

'''
 private class MigrationThread extends Thread implements Runnable {
        private final long sleepTime = max(250L, partitionMigrationInterval);
        private final AtomicLong migrationTaskCount = new AtomicLong();

        MigrationThread(Node node) {
            super(node.getHazelcastThreadGroup().getInternalThreadGroup(),
                    node.getHazelcastThreadGroup().getThreadNamePrefix("migration"));
        }

        @Override
        public void run() {
            try {
                while (!isInterrupted()) {
                    doRun();
                }
                    logger.finest("MigrationThread is interrupted: " + e.getMessage());
                }
            } catch (OutOfMemoryError e) {
                OutOfMemoryErrorDispatcher.onOutOfMemory(e);
            } finally {
                migrationQueue.clear();
            }
        }
        
        private void doRun() throws InterruptedException {
            for (; ; ) {
                if (!isMigrationAllowed()) {
                    break;
                }
                Runnable r = migrationQueue.poll(1, TimeUnit.SECONDS);
                if (r == null) {
                    break;
                }

                processTask(r);
                if (partitionMigrationInterval > 0) {
                    Thread.sleep(partitionMigrationInterval);
                }
            }
            boolean hasNoTasks = migrationQueue.isEmpty();
            if (hasNoTasks) {
                if (migrating) {
                    migrating = false;
                    logger.info("All migration tasks have been completed, queues are empty.");
                }
                evictCompletedMigrations();
                Thread.sleep(sleepTime);
            } else if (!isMigrationAllowed()) {
                Thread.sleep(sleepTime);
            }
        }
        private void postTask(Runnable task){
            if(task instanceof MigrateTask){
                migrationTaskCount.incrementAndGet();
            }
            migrationQueue.add(task);
        }
        boolean processTask(Runnable r) {
            if (r == null || isInterrupted()) {
                return false;
            }

            try {
                r.run();
            } catch (Throwable t) {
                logger.warning(t);
            }

            if(r instanceof MigrateTask){
                migrationTaskCount.decrementAndGet();
            }

            return true;
        }

        void stopNow() {
            migrationQueue.clear();
            interrupt();
        }

        boolean isMigrating() {
            return migrationTaskCount.get()>0;
        }
    }
'''

pveentjer:
After a short discussion with Basri, the solution with the counter makes sense to some point. However there is a lot of complex queue interaction like clear which could leave the counter in a permanently inconsistent state. I'll do some more thinking.

pveentjer:
@metanet I currently have a bunch of other tasks on my plate. Please have a look at the following code that should simplify the logic.

```
 static class MigrationQueue {
        private final AtomicInteger migrationTaskCount = new AtomicInteger();
        private final BlockingQueue<Runnable> taskQueue = new LinkedBlockingQueue<Runnable>();

        public Runnable poll(int timeout, TimeUnit unit) throws InterruptedException {
            return taskQueue.poll(timeout, unit);
        }

        public void add(Runnable task){
            if(task instanceof MigrateTask){
                migrationTaskCount.incrementAndGet();
                taskQueue.add(new MigrationTaskWrapper((MigrateTask)task));
            }else{
                taskQueue.add(task);
            }
        }

        public void clear() {
            List<Runnable> sink = new LinkedList<Runnable>();
            taskQueue.drainTo(sink);

            for(Runnable task: sink){
                if(task instanceof MigrateTask){
                    migrationTaskCount.decrementAndGet();
                }
            }
        }

        public boolean hasMigrations(){
            return migrationTaskCount.get()>0;
        }

        public boolean isEmpty() {
            return taskQueue.isEmpty();
        }

        public int size() {
            return taskQueue.size();
        }

         class MigrationTaskWrapper implements Runnable{
            private final MigrateTask task;

            public MigrationTaskWrapper(MigrateTask task) {
                this.task = task;
            }

            @Override
            public void run() {
                try {
                    task.run();
                }finally {
                    migrationTaskCount.decrementAndGet();
                }
            }
        }
    }
```

metanet:
@pveentjer pr is updated

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -1254,8 +1252,8 @@ private boolean hasOnGoingMigrationMaster(Level level) {
 
     @Override
     public boolean hasOnGoingMigrationLocal() {
-        return !activeMigrations.isEmpty() || !migrationQueue.isEmpty()
-                || migrationThread.isMigrating();
+        return !activeMigrations.isEmpty() || migrationQueue.isNonEmpty()
+                || migrationQueue.hasMigrations();
```
pveentjer: why the complex query? The MigrationQueue knows if there is a MigrationTask pending or executing. So we does activeMigrations and migrationQueue no empty be used?

So why not just:

```
 public boolean hasOnGoingMigrationLocal() {return migrationQueue.hasMigrations();}
```

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -2130,15 +2124,18 @@ private void doRun() throws InterruptedException {
...
             try {
+                if (r == null || isInterrupted()) {
+                    return false;
+                }
+
                 r.run();
             } catch (Throwable t) {
                 logger.warning(t);
+            } finally {
+                migrationQueue.afterTaskCompletion(r);
```
pveentjer: Why did you remove the wrapper task? With the wrapper task all this stuff remains an implementation detail of the queue. Now you have added the complexity to the thread.

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+
+@Category(QuickTest.class)
+public class MigrationQueueTest {
+
+    private final MigrationQueue migrationQueue = new MigrationQueue();
+
+    @Test
+    public void test_migrationTaskCount_incremented_onMigrateTask() {
```
pveentjer: Why does every test have 'test_migrationTaskCount' as prefix? It doesn't seem to add a lot of additional info,

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
```diff
@@ -0,0 +1,77 @@
...
+        migrationQueue.add(mock(InternalPartitionServiceImpl.MigrateTask.class));
+        migrationQueue.poll(1, TimeUnit.SECONDS);
+
+        assertTrue(migrationQueue.hasMigrations());
+    }
+
+    @Test
+    public void test_migrateTaskCount_decremented_afterMigrateTaskCompleted()
+            throws InterruptedException {
+        final InternalPartitionServiceImpl.MigrateTask migrateTask = mock(InternalPartitionServiceImpl.MigrateTask.class);
```
pveentjer: Can you make this a static import?

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
mdogan: `migrationQueue` itself is also a detail, it's directly bound to the partition service. Moving complexity from one side to another doesn't make any difference.

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
mdogan: Only master node knows `migrationQueue`, on other nodes it's always empty. `activeMigrations ` check is required for non-master nodes.

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public void test_migrateTaskCount_notDecremented_afterNonMigrateTaskCompleted()
+            throws InterruptedException {
+        migrationQueue.add(mock(InternalPartitionServiceImpl.MigrateTask.class));
+        migrationQueue.afterTaskCompletion(mock(Runnable.class));
+
+        assertTrue(migrationQueue.hasMigrations());
+    }
+
+    @Test
+    public void test_migrateTaskCount_decremented_onClear() {
```
pveentjer: most test in HZ have the following naming convention:

nameMethod_whenFoo_thenBla.

This makes it very easy to figure out where the test of the method is found and which cases are being tested. 

In this case I would expect the method to be called clear_whenMigrationTaskQueued_thenMigrationCountDecreased

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
```diff
@@ -0,0 +1,77 @@
...
+        migrationQueue.add(mock(InternalPartitionServiceImpl.MigrateTask.class));
+        migrationQueue.clear();
```
pveentjer: Also add a line of white space between your setup and your action. Even though your test methods are very small, it still helps to see the general structure faster.

hazelcast/src/main/java/com/hazelcast/partition/impl/MigrationQueue.java
```diff
@@ -0,0 +1,78 @@
...
+package com.hazelcast.partition.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+class MigrationQueue {
```
pveentjer: Can you add a few lines of javadoc for other engineers what the purpose is of this class?

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
pveentjer: Probably it would make sense to add some comments to explain to for others.

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
metanet: No, i won't. Test is already too short

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
metanet: This is my preference. Unless there is a strict rule against my preference, i wont change it.

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
pveentjer: It does. Now you have a mo ecomplex method than needs to be. With the wrapper task approach I have given, this is fully an implementation detail of the MigrationQueue. The logic for this thread is simplified since it doesn't need to do any callbacks to the queue. In my book this is cleaner code since incrementing and decrementing the counter is a implementation detail of the MigrationQueue.

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
metanet: It is a non static inner class

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
metanet: It is just my preference

metanet:
I am done with this pr. I wont make any change related to its style.

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
mdogan: If we start to talk things in our books then I think there's no need to discuss about this PR anymore. Discussion here started to become about esthetics and choices, again and again.

hazelcast/src/test/java/com/hazelcast/partition/impl/MigrationQueueTest.java
pveentjer: Do you think it is wise if we allow everybody to have their own preference when writing a test? Imho it makes code harder to understand. 

We have a test practices page on the wiki where we have a naming approach. If you don't like it and want to improve it, I'm all ear. But imho we should do this consistently using the whole system. I

pveentjer:
There are some other comments that need to be addressed like a few lines  of Javadoc. Just to give the reader a quick understanding of the purpose (providing info of a migration is pending/executing) of the class (don't care about the methods).

hazelcast/src/main/java/com/hazelcast/partition/impl/InternalPartitionServiceImpl.java
metanet: Partition service already manages the whole migration process. Migration queue is only responsible for managing migration tasks. Therefore, having an extra callback method doesn't seem problematic to me. The other approach, wrapping the given runnable, is also ok but I don't think it is necessary and I want to keep our current approach.

metanet:
I added Javadoc for ``MigrationQueue`` class. I also updated commit message and pr description to make purpose of the pr more understandable. Lastly, I answered your concern about the callback in my reply above.

pveentjer:
Good.

pveentjer:
:+1:

PR #7226, opened 2015-12-31
LockStoreImpl has to serialize its partitionId during serialization

jerrinot:
#7156 introduced per-partition locks, but it was missing partitionId field in serialization methods
as a result the partition was set to 0 on during migration and this result in contention issues.


ahmetmircik:
:+1:  Good catch.

gurbuzali:
:+1:

PR #7227, opened 2015-12-31
Moved scheduler into LockStore

gurbuzali:
removed partitionId from LockStore since it will be redundant
with this change, the fix for the issue https://github.com/hazelcast/hazelcast/issues/6141 will be backportable.


jerrinot:
looks good.
:+1:

eminn:
:+1:

PR #7237, opened 2016-01-04
Fix typo in Python client protocol ID

cangencer:


jerrinot:
:+1: 
why are the client ids important? is it for debugging? what's a difference from let's say python client and cpp client? they both use the same protocol, don't they?

noctarius:
The more important question, what happens to clients that are thrid-party made?

cangencer:
I believe they get marked as "Other", if they send some other 3 byte string. I don't think it serves any purpose other than some metric information. @sancar or @asimarslan , can you confirm?

asimarslan:
This id is used for statistics in man. center etc.

asimarslan:
:+1:

PR #7247, opened 2016-01-05
Operation toString includes System.identityHash

pveentjer:
This PR contains 2 improvements:
* Included the System.identityHash in the Operation.toString. The consequence is when this toString is called, biased locking doesn't work anymore for the object. Which probably is fine. The reason for this is that we currently have no good way to identify an operation since callId can be updated on retry. This makes dealing with invocation issues a lot more problematic.
*  The IsStillRunnningService prints itself instead of the invocation it is dealing with. So this prevent doing proper analysis of what is happening.


jerrinot:
:+1: on green

PR #7248, opened 2016-01-05
Operation.toString includes identity hash

pveentjer:
Operation.toString includes identityHash for improved log analysis.

Also fixed issue with toString in isstillrunningservice where the service.toString is called instead of the invocation.

Backport of #7247


pveentjer:
Since this is a backport and the other one is reviewed, I'll merge this one.

PR #7250, opened 2016-01-05
Don't update expiry time on update if update duration of used expiry policy is not available

### LINKED ISSUE #7236: Cache expiry is lost after an update
Hi all,

We are using Hazelcast as a JCache provider. As JSR 107 Javadoc states, *an update should not reset expiry time* for CreatedExpiryPolicy. However, if we update a cache entry, it does not expire.

Here is a test case: 

 ```
@Test
public void testExpiryWithUpdate() throws InterruptedException {
    final int expiry = 2000;
    Duration duration = new Duration(TimeUnit.MILLISECONDS, expiry);
    Configuration<Integer, String> config = new MutableConfiguration<Integer, String>()
            .setExpiryPolicyFactory(CreatedExpiryPolicy.factoryOf(duration));

    try (Cache<Integer, String> cache = Caching.getCachingProvider().getCacheManager()
            .createCache("testCache", config)) {
        cache.put(1, "value");
        //cache.put(1, "value");
        Assert.assertNotNull(cache.get(1));
        Thread.sleep(expiry);
        Assert.assertNull(cache.get(1));
    }
}
```

The entries expire as expected after 2 seconds, if we did not update them (via put or replace methods), so above test succeeds. 

But if we update the entry (ie: uncomment out second put call) then the entry doesn't expire and test fails. Moreover, an update does not update expire time (ie: expiry time is not extended), but cache behaves as if expiry policy is eternal. That is, the following code block prints out "...still not expired" forever (using same config as above one, with CreatedExpiry policy with 2 second expiry time).

```
try (Cache<Integer, String> cache = Caching.getCachingProvider().getCacheManager()
                .createCache("testCache", config)) {
    cache.put(1, "value");
    cache.put(1, "value");
    Assert.assertNotNull(cache.get(1));
    String value;
    while((value = cache.get(1)) != null) {
        System.out.println("...still not expired");
        Thread.sleep(1000);
    }
    System.out.println("Entry expired!");
}
``` 

I tried this with hazelcast versions 3.4.6, 3.5.4, and 3.6-EA3. 

I can confirm that only Apache Ignite runs as expected (an update does not change/reset expiry time for CreatedExpiryPolicy). Hazelcast and Infinispan fails as I summarized above.

### PR DESCRIPTION
**Pull request by serkan-ozal**
Fixes #7236

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java
@@ -14,6 +14,7 @@ import javax.cache.Cache;
 import javax.cache.configuration.FactoryBuilder;
 import javax.cache.configuration.MutableCacheEntryListenerConfiguration;
 import javax.cache.configuration.MutableConfiguration;
+import javax.cache.expiry.CreatedExpiryPolicy;
 import javax.cache.expiry.Duration;
 import javax.cache.expiry.ExpiryPolicy;
 import javax.cache.expiry.ModifiedExpiryPolicy;
@@ -306,8 +307,8 @@ public abstract class CacheBasicAbstractTest extends CacheTestSupport {
         final CacheConfig<Integer, Integer> config = getCacheConfigWithMaxSize(MAX_SIZE);
         final ICache<Integer, Integer> cache = createCache(config);
         final int maxSize = withoutEviction
-                                ? getMaxCacheSizeWithoutEviction(config)
-                                : getMaxCacheSizeWithEviction(config);
+                ? getMaxCacheSizeWithoutEviction(config)
+                : getMaxCacheSizeWithEviction(config);
 
         // we prefill the cache with half of the max size, so there will be new inserts from the AbstractCacheWorker
         int prefillSize = maxSize / 2;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by emrahkocaman** [comment]
:+1:

File: hazelcast/src/test/java/com/hazelcast/cache/CacheBasicAbstractTest.java [hunk 1]
```diff
@@ -816,4 +817,22 @@ public void run() {
 
         abstract void doRun(Random random);
     }
+
+    // https://github.com/hazelcast/hazelcast/issues/7236
+    @Test
+    public void expiryTimeShouldNotBeChangedOnUpdateWhenCreatedExpiryPolicyIsUsed() {
+        final int CRETAED_EXPIRY_TIME_IN_MSEC = 1000;
```
jerrinot: typo in `*CRETAED*_EXPIRY_TIME_IN_MSEC`

serkan-ozal: ah, ok thanks

jerrinot: does it have to be 1000ms? could you make it 100ms or even less? 1s for single test is quite a lot.

serkan-ozal: not necessary. I will decrease it

jerrinot:
:+1:

serkan-ozal:
@jerrinot I will handle your comments after PR merged because I think no need to wait one more PR builder (1 PR builder if I am lucky, in general it is 1+ :) )

serkan-ozal:
@jerrinot your comments are handled via https://github.com/hazelcast/hazelcast/commit/dc050d4adc218769b52b57d7bb1263200c945ccf

PR #7262, opened 2016-01-06
Fix migration completion log condition

metanet:
* Migration completion log should be printed when migration is started and completed.


mdogan:
:+1:

mdogan:
run-lan-run

PR #7264, opened 2016-01-07
Improves IMap test coverage

ahmetmircik:
Also removed some never-entered branches and redundant methods for coverage. See `EntryOperation` and `EntryBackupOperation`

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryBackupOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryBackupOperation.java
@@ -66,7 +66,7 @@ public class EntryBackupOperation extends KeyBasedMapOperation implements Backup
 
         Map.Entry entry = createMapEntry(dataKey, oldValue);
 
-        processBackup(entry);
+        entryProcessor.processBackup(entry);
 
         if (noOpBackup(entry)) {
             return;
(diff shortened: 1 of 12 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryBackupOperation.java
```diff
@@ -123,14 +114,10 @@ private boolean entryRemovedBackup(Map.Entry entry) {
         return false;
     }
 
-    private boolean entryAddedOrUpdatedBackup(Map.Entry entry) {
-        final Object value = entry.getValue();
-        if (value != null) {
```
jerrinot: do you know why there was the null check?

gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryBackupOperation.java
gurbuzali: `entryRemovedBackup` method is handling `null` value, redundant to check it here I guess

ahmetmircik: I wrote it before. 

But it is unnecessary because value can not be null there, previous `if` checks in `run method` handles null-ness case. But here is also a fragility because order of `if` checks should not be changed, i added a comment for the importance of `if` check orders for `EntryOperation`, also one similar comment can be added here.

jerrinot:
:+1:

PR #7265, opened 2016-01-07
Adding client reliableTopic tests

sancar:
Added missing overloadPolicy tests to client.
Javadoc of TopicOverloadPolicy is fixed.
Distributed version of OverloadTest is implemented and added

### DIFF
--- a/hazelcast-client-legacy/src/main/java/com/hazelcast/client/config/ClientConfig.java
+++ b/hazelcast-client-legacy/src/main/java/com/hazelcast/client/config/ClientConfig.java
@@ -244,12 +244,12 @@ public class ClientConfig {
      * @return the found config. If none is found, a default configured one is returned.
      */
     public ClientReliableTopicConfig getReliableTopicConfig(String name) {
-        ClientReliableTopicConfig nearCacheConfig = lookupByPattern(reliableTopicConfigMap, name);
-        if (nearCacheConfig == null) {
-            nearCacheConfig = new ClientReliableTopicConfig(name);
-            addReliableTopicConfig(nearCacheConfig);
+        ClientReliableTopicConfig reliableTopicConfig = lookupByPattern(reliableTopicConfigMap, name);
+        if (reliableTopicConfig == null) {
+            reliableTopicConfig = new ClientReliableTopicConfig(name);
+            addReliableTopicConfig(reliableTopicConfig);
         }
-        return nearCacheConfig;
+        return reliableTopicConfig;
     }
 
     /**
--- a/hazelcast-client-legacy/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
+++ b/hazelcast-client-legacy/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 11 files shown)


sancar:
@pveentjer Can you review ?

pveentjer:
First thing on my list tomorrow morning.

hazelcast-client-legacy/src/main/java/com/hazelcast/client/config/ClientConfig.java
```diff
@@ -244,12 +244,12 @@ public ClientConfig addReliableTopicConfig(ClientReliableTopicConfig reliableTop
      * @return the found config. If none is found, a default configured one is returned.
      */
     public ClientReliableTopicConfig getReliableTopicConfig(String name) {
-        ClientReliableTopicConfig nearCacheConfig = lookupByPattern(reliableTopicConfigMap, name);
```
pveentjer: Did I make this? I should drink more or less coffee...

hazelcast-client/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
```diff
@@ -33,20 +50,17 @@
 import static org.junit.Assert.assertTrue;
 
 @RunWith(HazelcastParallelClassRunner.class)
-@Category(NightlyTest.class)
```
pveentjer: Pfff..  what did I think....

pveentjer:
Pfff..  too only a few hours. The cause is a one line addition to your setup:

```
warmUpPartitions(hazelcastInstances);
```

The problem is that the partitions are not assigned and in case of the backup this leads to a rejection of the backup execution. Eventually causing the invocation future to wait till a the backup timeout expires.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
```diff
@@ -67,13 +79,13 @@ public void testCreation() {
 
     @Test(expected = NullPointerException.class)
     public void addMessageListener_whenNull() {
-        ITopic topic = client.getReliableTopic("foo");
+        ITopic topic = client.getReliableTopic(randomString());
```
pveentjer: What is the added value of randomization of the topic name? 

Marko suggested removing randomization from tests:
https://hazelcast.atlassian.net/wiki/display/PM/Remove+Randomness+from+Unit+Tests

hazelcast-client/src/test/java/com/hazelcast/client/topic/ClientReliableTopicOverloadTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+        serializationService = ((HazelcastClientProxy) client).getSerializationService();
+
+        String topicName = getTestMethodName();
+        topic = (ClientReliableTopicProxy<String>) client.<String>getReliableTopic(topicName);
+
+        ringbuffer = topic.getRingbuffer();
+    }
+
+    @Override
```
pveentjer: Isn't it easier if the TopicOverloadAbstract test has fields like ringbuffer and the subclass is responsible for initializing them? Adding these 'getters' is only code noise.

hazelcast-client-legacy/src/test/java/com/hazelcast/client/topic/ClientReliableTopicTest.java
sancar: I tend to use random names for the data put to hazelcast or in hazelcast data structures. The reason with this for me to differentiate which value is needed to be exact or which value does not matter what it is. When I see random name, I directly deduce that it does not matter which partition it lands to etc . Think of it as `_` in Haskell.
But random number if iterations or random sleep, they can lead to when trying to reproduce the issue back. This is close to idea of Fuzzy testing, but we are missing `the reporting the random used values` part to use them again when reproducing the test.(There is also finding the minimal failing test in Fuzzy test, but this is another story)

hazelcast-client/src/test/java/com/hazelcast/client/topic/ClientReliableTopicOverloadTest.java
sancar: Yes, you are right. I will simplify these parts.

sancar:
@pveentjer thanks for the finding the problem in the test. I have updated the pr according to your comments. 
Is it OK to merge now ?

pveentjer:
:+1:

PR #7278, opened 2016-01-08
Shutdown `MetricRegistery` instances at the end of test

serkan-ozal:
… because they have their internal threads so non-terminated instances causes zombie threads after each test


jerrinot:
:+1:

muratayan:
Nice catch :+1:

PR #7274, opened 2016-01-08
Return cache config as response even though found and created cache config could not put into cache configs inside cache service

serkan-ozal:
… in case of parallel access

Fixes #7208

PR #7279, opened 2016-01-08
[BACKPORT] Return cache config as response even though found and created cache config could not put into cache configs inside cache service

serkan-ozal:
… in case of parallel access

Backport of #7274


serkan-ozal:
Merging this one since its forward port is already merged.

PR #7282, opened 2016-01-09
Operation.toString for identityhash prints an unwanted quote

pveentjer:


pveentjer:
Since this is a removal of an unwanted character in a toString, I'll merge this one myself.

PR #7283, opened 2016-01-09
PartitionIteratingOperation.toString include factory name

pveentjer:
The operationFactory is included in the toString output so we can at least see, if a bug reports come in that involves a PartitionIteratingOperation, what kind of operation this PartitionIteratingOperation actually is executing. 

For more information see the log message from the following github ticket:
https://github.com/hazelcast/hazelcast/issues/7280


jerrinot:
:+1:

Donnerbart:
:+1:

PR #7284, opened 2016-01-11
IsStillRunning logging under debug

pveentjer:
No more warning logging, but everything under debug.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
@@ -63,6 +63,7 @@ public class IsStillRunningService {
 
     /**
      * Checks if an operation is still running.
+     *
      * @param invocation The invocation for this operation.
      * @return true if the operation is running, false otherwise.
      */
@@ -86,7 +87,9 @@ public class IsStillRunningService {
         } catch (Exception e) {
             invocation.logger.warning("While asking 'is-executing': " + invocation, e);
         }
-        invocation.logger.warning("'is-executing': " + executing + " -> " + invocation);
+        if (invocation.logger.isFinestEnabled()) {
+            invocation.logger.finest("'is-executing': " + executing + " -> " + invocation);
+        }
         return executing;
     }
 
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
```diff
@@ -86,7 +87,9 @@ public boolean isOperationExecuting(Invocation invocation) {
         } catch (Exception e) {
             invocation.logger.warning("While asking 'is-executing': " + invocation, e);
         }
-        invocation.logger.warning("'is-executing': " + executing + " -> " + invocation);
+        if (invocation.logger.isFineEnabled()) {
+            invocation.logger.finest("'is-executing': " + executing + " -> " + invocation);
```
Donnerbart: You check for `isFineEnabled()` but use `finest()`. There is a `isFinestEnabled()` method which should most probably be used here.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
```diff
@@ -224,7 +229,9 @@ public IsOperationStillRunningCallback(Invocation invocation) {
         @Override
         public void onResponse(Object response) {
             boolean executing = Boolean.TRUE.equals(response);
-            invocation.logger.warning("'is-executing': " + executing + " -> " + invocation);
+            if (invocation.logger.isFineEnabled()) {
```
Donnerbart: Same here as above.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/IsStillRunningService.java
```diff
@@ -264,7 +271,9 @@ public void run() {
                     invocation.nodeEngine, invocation.serviceName, isStillRunningOperation,
                     invocation.getTarget(), 0, 0, IS_EXECUTING_CALL_TIMEOUT, callback, true);
 
-            invocation.logger.warning("Asking if operation execution has been started: " + invocation);
+            if (invocation.logger.isFineEnabled()) {
```
Donnerbart: And a third time.

jerrinot:
moving to 3.6.1 as discussed

jerrinot:
:+1:

pveentjer:
A fix has been made that prevent running into this problem in the first place. So we can close this PR

PR #7286, opened 2016-01-11
Invocation count logging improvements

pveentjer:
This pr contains the following 3 improvements:
* Includes the number of scanned invocations in the logging of the InvocationMonitor
* if debug is enabled, the InvocationMonitor will always display what it does while scanning
* The number of invocations is now a trigger for the HealthMonitor to trigger health logging

This will help us dealing with large number of invocation problems.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
@@ -30,14 +30,18 @@ import com.hazelcast.spi.impl.operationexecutor.OperationHostileThread;
 import com.hazelcast.util.Clock;
 import com.hazelcast.util.EmptyStatement;
 import com.hazelcast.util.counters.SwCounter;
+
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
 
 import static com.hazelcast.instance.OutOfMemoryErrorDispatcher.inspectOutputMemoryError;
 import static com.hazelcast.internal.metrics.ProbeLevel.MANDATORY;
 import static com.hazelcast.util.counters.SwCounter.newSwCounter;
+import static java.util.logging.Level.FINE;
+import static java.util.logging.Level.INFO;
 
 /**
  * The InvocationMonitor monitors all pending invocations and determines if there are any problems like timeouts. It uses the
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -153,10 +157,13 @@ private void scan() {
             long now = Clock.currentTimeMillis();
             int backupTimeouts = 0;
             int invocationTimeouts = 0;
+            int invocationCount = 0;
```
Donnerbart: Is an `int` sufficient for this value?

Donnerbart:
:+1: beside a minor comment

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
pveentjer: Yes :) Maximum number of concurrent invocations is bound to the size of the CHM.... and therefor an int.

Donnerbart: Ah, forgive me, it looked like a field of `InvocationMonitor`, not like a local variable. Makes total sense now.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -247,9 +254,18 @@ private boolean checkBackupTimeout(Invocation invocation) {
             }
         }
 
-        private void log(int backupTimeouts, int invocationTimeouts) {
+        private void log(int invocationCount, int backupTimeouts, int invocationTimeouts) {
+            Level logLevel = null;
             if (backupTimeouts > 0 || invocationTimeouts > 0) {
-                logger.info("Handled " + invocationTimeouts + " invocation timeouts and " + backupTimeouts + " backupTimeouts");
+                logLevel = INFO;
+            } else if (logger.isFineEnabled()) {
```
jerrinot: I'd guard it by `logLevel == null && logger.isFineEnabled`

as otherwise you are downgrading INFO to FINE -> confusing.

eminn:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
pveentjer: I don't see the downgrade. Did you see the if else branch? If the else would be missing, then there is a downgrade.

jerrinot: you are right, I missed it was in `else` branch.

jerrinot:
:+1:

PR #7293, opened 2016-01-11
SENDER_BACKUP not allowed to update SENDER's load status.

tombujok:
Fix for one of the MapLoader's bugs.
It doesn't fully solve the problem when there's more than one replica - another PR will follow (#7295)


jerrinot:
:+1:

gurbuzali:
:+1:

PR #7295, opened 2016-01-11
Enforces only one SENDER_BACKUP loader

tombujok:
**WAS:**
With a couple of nodes and couple of backups configured for a map hazelcast would create more than one `SENDER_BACKUP` loader. Given that the `SENDER_BACKUP` nodes are notified by a call with `setReplicaIndex(1)` only one `SENDER_BACKUP`` is notified about the load statu, and the other ones are just stuck with the loading status.

**IS:**
There will be just one SENDER_BACKUP - the one with the `replicaIndex == 1`


tombujok:
verify

jerrinot:
:+1:

gurbuzali:
:+1:

PR #7301, opened 2016-01-12
Fix cache record creation with expiry time

mdogan:
Cache record should be created using already available `now` time,
instead of `System.currentMillis()`. Because expiry time is calculated
and relative to `now`.
Otherwise, when System.currentMillis() is used and TTL is very small,
difference between  expiry time and creation time can be less than 0,
which leads to no expiry at all.

Fixes https://github.com/hazelcast/hazelcast-enterprise/issues/679


serkan-ozal:
This is trivial fix. I am merging this one myself

PR #7305, opened 2016-01-13
Trigger quorum listener when memberlist changes

### LINKED ISSUE #7300: QuorumListener not invoked
Hi,
I'd like to have 3 node cluster and for reacting on quorum lost events I'd like to use QuorumListener registered in QuorumConfig. But it seems to me that it doesn't work as I expected.

Scenario (3 nodes => quorum = 2):
- Node1 starts
- Node2 starts
- Node2 crashes => Node 1 lost quorum but my QuorumListener was not invoked on Node1.

Using Hazelcast 3.5.4.

Is it bug or am I missing something? Is there another mechanism for quorum events which I can use?

eminn:
when a change happens in the member list, it is also reflected to quorum listeners besides quorum state. Fixes #7300


gurbuzali:
:+1:

mustafaiman:
:+1:

PR #7307, opened 2016-01-13
UNDEFINED error code should not be in the protocol

ihsandemir:
Removing the error code UNDEFINED. There is no such a communicated message.


sancar:
:+1:

PR #7314, opened 2016-01-13
[DONT MERGE] IO thread starvation problem

pveentjer:
The problem which can lead to OOME is there can be an infinite loop of SocketWriters getting processed in the NonBlockingIOThread.processTaskQueue method. It can happen that the same SocketWriter is picked up again and again in the same loop. This has been confirmed with a  demo program in combination with some detection. If the NonBlockingIOThread get 'stuck' in this loop, it will not be processing any connections that rely on a write event to get their data written to socket, and for these connections a buildup of packet in the connection queue is the consequence. This buildup eventually can lead to an OOME;  especially with a large number of connections (e,g. many clients).

This PR contains 2 fixes:
- there is a hard cap on the number of items to be processed.
- there is detection of a duplicate. As soon as the same item is found twice, the loop ends and the selection keys get processed first.

We are currently experimenting to see the impact on performance. For the time being it seems that setting the cap to 1, even leads to a performance increase of 3..5% which of course is a nice bonus.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
@@ -40,6 +40,8 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
     private static final int SELECT_WAIT_TIME_MILLIS = 5000;
     private static final int SELECT_FAILURE_PAUSE_MILLIS = 1000;
 
+    public static final int MAXIMUM_ITEMS_TAKEN_FROM_TASK_QUEUE_RENAME_ME_I_AM_SILLY = Integer.getInteger("ioselector.batchsize", 10);
+
     @Probe(name = "taskQueueSize")
     private final Queue<Runnable> taskQueue = new ConcurrentLinkedQueue<Runnable>();
     @Probe
@@ -198,9 +200,9 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
 
     private void runSelectLoop() throws IOException {
         while (!isInterrupted()) {
-            processTaskQueue();
+            boolean queueFullyProcessed = processTaskQueue();
 
-            int selectedKeys = selector.select(SELECT_WAIT_TIME_MILLIS);
+            int selectedKeys = queueFullyProcessed ? selector.select(SELECT_WAIT_TIME_MILLIS) : selector.selectNow();
             if (selectedKeys > 0) {
                 lastSelectTimeMs = currentTimeMillis();
                 handleSelectionKeys();
@@ -220,14 +222,18 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
         }
     }
 
-    private void processTaskQueue() {
-        while (!isInterrupted()) {
+    /**
+     * @return true is all tasks have been processed
+     */
+    private boolean processTaskQueue() {
+        for (int i = 0; i < MAXIMUM_ITEMS_TAKEN_FROM_TASK_QUEUE_RENAME_ME_I_AM_SILLY && !isInterrupted(); i++) {
             Runnable task = taskQueue.poll();
             if (task == null) {
-                return;
+                return true;
             }
             executeTask(task);
         }
+        return false;
     }
 
     private void executeTask(Runnable task) {
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
@@ -222,15 +222,30 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
         }
     }
 
+    private Runnable pending;
+
     /**
      * @return true is all tasks have been processed
      */
     private boolean processTaskQueue() {
+        Runnable first = pending;
+
+        if (first != null) {
+            executeTask(first);
+            pending = null;
+        }
+
         for (int i = 0; i < MAXIMUM_ITEMS_TAKEN_FROM_TASK_QUEUE_RENAME_ME_I_AM_SILLY && !isInterrupted(); i++) {
             Runnable task = taskQueue.poll();
             if (task == null) {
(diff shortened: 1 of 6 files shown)


bwzhang2011:
@pveentjer, when I merged your modification of NonBlockingIOThread (and also with other classes) into 3.6 branch and test with IQueue, I found some error like this:

2016-01-25T10:37:40.310+0800 WARN com.hazelcast.nio.tcp.nonblocking.NonBlockingSocketWriter. hz._hzInstance_1_ReliableMsgCenterGroup.IO.thread-out-5 MDC{} [192.168.115.10]:9800 [ReliableMsgCenterGroup] [3.6] hz._hzInstance_1_ReliableMsgCenterGroup.IO.thread-out-5 Closing socket to endpoint Address[10.100.11.31]:59789, Cause:java.nio.channels.CancelledKeyException
java.nio.channels.CancelledKeyException
	at sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73) ~[?:1.8.0_51]
	at sun.nio.ch.SelectionKeyImpl.interestOps(SelectionKeyImpl.java:77) ~[?:1.8.0_51]
	at com.hazelcast.nio.tcp.nonblocking.AbstractHandler.unregisterOp(AbstractHandler.java:73) ~[hazelcast-all-3.6.jar:3.6]
	at com.hazelcast.nio.tcp.nonblocking.NonBlockingSocketWriter.unschedule(NonBlockingSocketWriter.java:295) ~[hazelcast-all-3.6.jar:3.6]
	at com.hazelcast.nio.tcp.nonblocking.NonBlockingSocketWriter.handle(NonBlockingSocketWriter.java:344) ~[hazelcast-all-3.6.jar:3.6]
	at com.hazelcast.nio.tcp.nonblocking.NonBlockingSocketWriter.run(NonBlockingSocketWriter.java:423) ~[hazelcast-all-3.6.jar:3.6]
	at com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThread.executeTask(NonBlockingIOThread.java:257) ~[hazelcast-all-3.6.jar:3.6]
	at com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThread.processTaskQueue(NonBlockingIOThread.java:249) ~[hazelcast-all-3.6.jar:3.6]
	at com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThread.runSelectLoop(NonBlockingIOThread.java:203) ~[hazelcast-all-3.6.jar:3.6]
	at com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThread.run(NonBlockingIOThread.java:165) ~[hazelcast-all-3.6.jar:3.6]
2016-01-25T10:37:40.317+0800 INFO com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThreadingModel. cached4 MDC{} [192.168.115.10]:9800 [ReliableMsgCenterGroup] [3.6] TcpIpConnectionManager configured with Non Blocking IO-threading model: 6 input threads and 6 output threads
2016-01-25T10:37:40.317+0800 DEBUG com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThreadingModel. cached4 MDC{} [192.168.115.10]:9800 [ReliableMsgCenterGroup] [3.6] InputThreads selectNow enabled=false
2016-01-25T10:37:40.317+0800 DEBUG com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThreadingModel. cached4 MDC{} [192.168.115.10]:9800 [ReliableMsgCenterGroup] [3.6] OutputThreads selectNow enabled=false

by the way: I setup some cluster with 2 members (192.168.115.9 and 192.168.115.10) and when I tested IQueue and message sent to 192.168.115.10 and then error thrown like above. I found that
such member dropped out of the cluster for both i the client and the server of 192.168.115.9 lost its connection with 192.168.115.10 so I had to restart the cluster again to setup the cluster both for client and server.

please leave me some message if I were wrong to merge such for just testing.

pveentjer:
Hi @bwzhang2011 this code has not been merged for a reason :) So testing it is always at your own risk.

The exception doesn't need to mean something bad; it is at closing the connection you can get this exception. That it is being logged perhaps is confusing. The main question is: why was the connection closed? Can you create a test which reproduces the behavior?

bwzhang2011:
@pveentjer, I just tested it with IQueue component with hazelcast client connecting to the cluster(like above described) it's hard to make it clearly that how your could reproduce it that's because it was a simple test case with IQueue offer with loop cases and reach a certain amount before it break e.g set the maxcount to be 1M.  
1. I test it under my machine connecting to the remote machine with the cluster, the ip enviroment is different but could be accessed through the switch and ping responsed arranged from 2 to 5ms

2. when the connection closed, testing continued and the client choose to send data to another cluster member, but to my surprise the member closed shown in the client also dropped out the connection between the cluster members that's to say it was not due to the network reason for it also diconnecting the relationship between the cluster and I didn't know why so that the partion operated was not invoked as expected 

3. my question was just to ask whether such modification would impact the connection or the cluster or something else. if so I would revert those changes. 

4. during those days testing around hazelcast3.5.4, I fould it was easy to induce future timeout(120s)  if continuing to send data wth IQueue(I didn't know about other distribute structures) which I though it was a long way to improve the nio implementation for I compared with IQueue with other broker like kafka or rocketmq especially for the latter based on netty component it was stable that no matter count of the data I would send for testing, it would not induce the timeout exception so I thought it was not a pure network reason for such.

pveentjer:
Did you get logging regarding OOME? Can you check with a profiler/jconsole what is happening? If you are running into an OOME, it can explain the cause of connection closing.

bwzhang2011:
@pveentjer, thanks for your advice and I will have a try next day with your suggestion and try to list more properties to get more debug and metrics info. by the way I just test it with only one client and with main class for testing (for I did consider it to be some simple loop testing)

pveentjer:
We saw this problem when testing with 600 hundreds clients and tests that generate a lot of network traffic.

bwzhang2011:
@pveentjer, I think unstable network env also induce such problem though without many clients connecting. and I was curious about how to solve the network traffic if many clients setup the connection. during IQueue or other testing, the data will always be sent to the same member of the cluster and will not be balanced each other and I want to know whether hazelcast could think about kafka way to process messaging between different brokers.

PR #7323, opened 2016-01-14
[DO NOT MERGE] Avoids missed states in key loader

tombujok:


hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
```diff
@@ -70,7 +70,8 @@
     protected final RecordStoreLoader recordStoreLoader;
     protected final MapKeyLoader keyLoader;
     // loadingFutures are modified by partition threads and could be accessed by query threads
-    protected final Collection<Future> loadingFutures = new ConcurrentLinkedQueue<Future>();
+    protected final Collection<Future> keyLoadingFutures = new ConcurrentLinkedQueue<Future>();
+    protected final Collection<Future> valueLoadingFutures = new ConcurrentLinkedQueue<Future>();
```
tombujok: I could either have one collection and check the type of each future being added in order not to add two key-loader futures, or have two collections for key-loader and value-loader futures.

PR #7324, opened 2016-01-14
Defaulting properties when negative value is generalized

sancar:
`get` methods of Hazelcast properties made return the default
value in case a negative value is configured.

Advantages of this change:
1) I see some places when a a property is negative, a different
hardcoded default value is used instead of one in the Properties
file. This way, same mistake won't be happening again.

2) There are lots of places that there are no check to if a value
is negative. This way all will be checked, a log will be printed
and it will fallback to default value,

3) There are places that negative value checked and correct value
is used, but this is code repeation. This logic to one place to
prevent code repetation.

4) Before this change, in some places we convert to defualt value
silently, in some places we throw exception when a negative value
configured. This change will make the behavuour consistent across
all propreties so that potential confusing behaviour for our users
is eliminated.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
@@ -17,6 +17,7 @@
 package com.hazelcast.instance;
 
 import com.hazelcast.config.Config;
+import com.hazelcast.logging.ILogger;
 
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
@@ -37,6 +38,7 @@ import static java.lang.String.format;
 public abstract class HazelcastProperties {
 
     private final String[] properties = createProperties();
+    private ILogger logger;
 
     /**
      * Created the properties array.
@@ -75,6 +77,10 @@ public abstract class HazelcastProperties {
         }
     }
 
+    public void setLogger(ILogger logger) {
+        this.logger = logger;
+    }
+
     /**
      * Returns the configured value of a {@link HazelcastProperty} as String.
      *
@@ -103,7 +109,15 @@ public abstract class HazelcastProperties {
      * @throws NumberFormatException if the value cannot be parsed
      */
     public int getInteger(HazelcastProperty groupProperty) {
-        return Integer.parseInt(properties[groupProperty.getIndex()]);
+        int configuredValue = Integer.parseInt(properties[groupProperty.getIndex()]);
+        if (configuredValue == -1 && groupProperty.isDisablable()) {
+            return -1;
+        }
+        if (configuredValue < 0) {
+            logIllegalValueMessage(groupProperty, configuredValue);
+            return Integer.parseInt(groupProperty.getDefaultValue());
+        }
+        return configuredValue;
     }
 
     /**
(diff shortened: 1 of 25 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -83,9 +83,7 @@ public InvocationMonitor(InvocationRegistry invocationRegistry, ILogger logger,
 
     private long initSlowInvocationThresholdMs(GroupProperties props) {
         long thresholdMs = props.getMillis(GroupProperty.SLOW_INVOCATION_DETECTOR_THRESHOLD_MILLIS);
-        if (thresholdMs > -1) {
-            logger.info("Slow invocation detector enabled, using threshold: " + thresholdMs + " ms");
-        }
+        logger.info("Slow invocation detector enabled, using threshold: " + thresholdMs + " ms");
```
Donnerbart: I'm not sure if this check should be removed, since `-1` means disabled for `SLOW_INVOCATION_DETECTOR_THRESHOLD_MILLIS`, not `0`, see the definition:
> The default is -1 indicating there is no detection.

So actually a negative value should be fine here.

hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
```diff
@@ -103,7 +109,12 @@ public boolean getBoolean(HazelcastProperty groupProperty) {
      * @throws NumberFormatException if the value cannot be parsed
      */
     public int getInteger(HazelcastProperty groupProperty) {
-        return Integer.parseInt(properties[groupProperty.getIndex()]);
+        int configuredValue = Integer.parseInt(properties[groupProperty.getIndex()]);
+        if (configuredValue < 0) {
```
Donnerbart: Since we have at least one property where there is a negative default value, we should maybe check this here. We could have another field in GroupProperty to allow negative values or check if the default is negative.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
sancar: Yes, you are right. I am adding this check back.

hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
sancar: The ones with default value `-1` has different meanings. 
For example we have
"hazelcast.operation.thread.count", "hazelcast.operation.generic.thread.count"
in which `-1` use a default value that hazelcast see fits(mostly dynamically calculated)

We have `-1` where it means feature is disabled.

I could add a check like following
```
if(configuredValue == -1 && defaultValue == -1){
     return -1;
}
```
So that when user specifically writes `-1`  that which have `-1`  as default value, we will not print anything to user.
I have found an exception that does not fit to my suggestion
"hazelcast.query.max.local.partition.limit.for.precheck" default value is 3 and can be disabled by -1

Then, adding a field to GroupProperty to differentiate `-1` means `disable` the feature.
@Donnerbart Do you think is that OK ?

Donnerbart: Seems to be a good idea, yes. We should not print a warning if a user explicitly defines a negative default value.

pveentjer:
What should be done is when a property is defined, it should have some kind of validator.

Pseudo code:
```
GroupProperty<Integer> partitionCount = new GroupProperty<Integer>("foo", 271, new Extractor<Integer(){
     Integer parse(String s){
         ...  if s not integer or smaller than 1 then exception ...
     }
}
```

This way you have everything local:
- name
- default
- validation

@Donnerbart WDYT

Donnerbart:
At the moment GroupProperty is an `enum`, so you can't have generics. Without an `enum` there were some other trade-offs I don't remember by heart at the moment. I already tried to add some validators in a branch, but gave up for some reason.

PR #7327, opened 2016-01-14
JMX collects calculation heavy stats periodically

muratayan:
@ahmetmircik @mdogan Can you take a look?

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
@@ -182,6 +182,7 @@ public class GroupProperties {
     public static final String PROP_LOGGING_TYPE = "hazelcast.logging.type";
     public static final String PROP_ENABLE_JMX = "hazelcast.jmx";
     public static final String PROP_ENABLE_JMX_DETAILED = "hazelcast.jmx.detailed";
+    public static final String PROP_JMX_UPDATE_INTERVAL_SECONDS = "hazelcast.jmx.update.interval.seconds";
     public static final String PROP_MC_MAX_VISIBLE_INSTANCE_COUNT = "hazelcast.mc.max.visible.instance.count";
     public static final String PROP_MC_MAX_VISIBLE_SLOW_OPERATION_COUNT = "hazelcast.mc.max.visible.slow.operations.count";
     public static final String PROP_MC_URL_CHANGE_ENABLED = "hazelcast.mc.url.change.enabled";
@@ -549,6 +550,8 @@ public class GroupProperties {
 
     public final GroupProperty ENABLE_JMX_DETAILED;
 
+    public final GroupProperty JMX_UPDATE_INTERVAL_SECONDS;
+
     public final GroupProperty MC_MAX_INSTANCE_COUNT;
 
     public final GroupProperty MC_MAX_SLOW_OPERATION_COUNT;
@@ -700,6 +703,7 @@ public class GroupProperties {
         LOGGING_TYPE = new GroupProperty(config, PROP_LOGGING_TYPE, "jdk");
         ENABLE_JMX = new GroupProperty(config, PROP_ENABLE_JMX, "false");
         ENABLE_JMX_DETAILED = new GroupProperty(config, PROP_ENABLE_JMX_DETAILED, "false");
+        JMX_UPDATE_INTERVAL_SECONDS = new GroupProperty(config, PROP_JMX_UPDATE_INTERVAL_SECONDS, "5");
         MC_MAX_INSTANCE_COUNT = new GroupProperty(config, PROP_MC_MAX_VISIBLE_INSTANCE_COUNT, "100");
         MC_MAX_SLOW_OPERATION_COUNT = new GroupProperty(config, PROP_MC_MAX_VISIBLE_SLOW_OPERATION_COUNT, "10");
         MC_URL_CHANGE_ENABLED = new GroupProperty(config, PROP_MC_URL_CHANGE_ENABLED, "true");
--- a/hazelcast/src/main/java/com/hazelcast/jmx/HazelcastMBean.java
+++ b/hazelcast/src/main/java/com/hazelcast/jmx/HazelcastMBean.java
@@ -16,11 +16,6 @@
 
 package com.hazelcast.jmx;
 
-import java.lang.management.ManagementFactory;
-import java.lang.reflect.Method;
-import java.util.HashMap;
-import java.util.Hashtable;
-import java.util.Map;
 import javax.management.Attribute;
 import javax.management.AttributeList;
 import javax.management.AttributeNotFoundException;
@@ -36,6 +31,11 @@ import javax.management.MBeanServer;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 import javax.management.ReflectionException;
+import java.lang.management.ManagementFactory;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Hashtable;
+import java.util.Map;
 
 /**
  * Base class for beans registered to JMX by Hazelcast.
@@ -47,14 +47,19 @@ public abstract class HazelcastMBean<T> implements DynamicMBean, MBeanRegistrati
     protected HashMap<String, BeanInfo> attributeMap = new HashMap<String, BeanInfo>();
     protected HashMap<String, BeanInfo> operationMap = new HashMap<String, BeanInfo>();
 
+    protected final long updateIntervalSec;
+
     final T managedObject;
     final ManagementService service;
+
     String description;
     ObjectName objectName;
+
 
     protected HazelcastMBean(T managedObject, ManagementService service) {
         this.managedObject = managedObject;
         this.service = service;
+        updateIntervalSec = service.instance.node.groupProperties.JMX_UPDATE_INTERVAL_SECONDS.getLong();
     }
 
     public void register(HazelcastMBean mbean) {
--- a/hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
+++ b/hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.jmx;
+
+import com.hazelcast.jmx.suppliers.StatsSupplier;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Provides LocalStats for {@link MapMBean}, {@link MultiMapMBean}, {@link QueueMBean}
+ * So that statistics are only created in given internal
+ * @param <T> LocalStats type
+ */
+public class LocalStatsDelegate<T> {
+
+    private volatile T localStats;
+    private final StatsSupplier<T> supplier;
+    private final long intervalMs;
+    private final AtomicLong lastUpdated = new AtomicLong(0);
+    private final AtomicBoolean inProgress = new AtomicBoolean(false);
+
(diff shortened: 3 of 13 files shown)


muratayan:
run-lab-run

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -182,6 +182,7 @@
     public static final String PROP_LOGGING_TYPE = "hazelcast.logging.type";
     public static final String PROP_ENABLE_JMX = "hazelcast.jmx";
     public static final String PROP_ENABLE_JMX_DETAILED = "hazelcast.jmx.detailed";
+    public static final String PROP_JMX_INTERVAL = "hazelcast.jmx.interval";
```
alparslanavci: I think ''hazelcast.jmx.update.interval" will be better for the sake of comprehensibility.

Donnerbart: Maybe we should also add the time unit to the name, so it's clear if we have to set seconds or milliseconds here.

muratayan: Sure thing. Quick suggestion would be awesome as I am about the fix findbugs and push it.

Donnerbart: Don't know, depending on what we have to set here maybe:
```
hazelcast.jmx.update.interval.seconds
hazelcast.jmx.update.interval.millis
```

muratayan: Danke! `hazelcast.jmx.update.interval.seconds` it is.

alparslanavci: `hazelcast.jmx.update.interval.seconds` maybe?

muratayan:
run-lab-run

hazelcast/src/main/java/com/hazelcast/jmx/HazelcastMBean.java
```diff
@@ -44,17 +46,26 @@
  */
 public abstract class HazelcastMBean<T> implements DynamicMBean, MBeanRegistration {
 
+    protected static final long NO_DELAY = 0;
```
ahmetmircik: is this used anywhere?

hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.hazelcast.jmx;
+
+import com.hazelcast.jmx.suppliers.StatSupplier;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class LocalStatsDelegate<T> {
```
ahmetmircik: Can you please simply document this class?

hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
```diff
@@ -0,0 +1,56 @@
...
+    private final long intervalMs;
+    private final AtomicLong lastUpdated = new AtomicLong(0);
+    private final AtomicBoolean inProgress = new AtomicBoolean(false);
+
+    public LocalStatsDelegate(StatSupplier<T> supplier, long intervalSec) {
+        this.supplier = supplier;
+        this.intervalMs = intervalSec * 1000;
+    }
+
+    public T getLocalStats() {
```
ahmetmircik: Can you please document what this method will return according to possible conditions?

hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
```diff
@@ -0,0 +1,56 @@
...
+        long delta = System.currentTimeMillis() - lastUpdated.get();
+        if (delta > intervalMs) {
+            if (inProgress.compareAndSet(false, true)) {
+                localStats = supplier.get();
```
ahmetmircik: what if we get an exception here?
In an exceptional case, the below part will not be executed.
```
lastUpdated.set(System.currentTimeMillis());
inProgress.set(false);
```

hazelcast/src/main/java/com/hazelcast/jmx/MapMBean.java
```diff
@@ -41,7 +45,10 @@
     private final AtomicLong totalEvictedEntryCount = new AtomicLong();
     private final String listenerId;
 
-    protected MapMBean(IMap managedObject, ManagementService service) {
+    private StatSupplier<LocalMapStats> localMapStatsSupplier;
```
ahmetmircik: seems we don't need this as a global variable.

hazelcast/src/main/java/com/hazelcast/jmx/MultiMapMBean.java
```diff
@@ -24,147 +30,152 @@
 @ManagedDescription("MultiMap")
 public class MultiMapMBean extends HazelcastMBean<MultiMap> {
 
-    protected MultiMapMBean(MultiMap managedObject, ManagementService service) {
+    private StatSupplier<LocalMultiMapStats> localMultiMapStatsSupplier;
```
ahmetmircik: seems we don't need this as a global variable.

hazelcast/src/main/java/com/hazelcast/jmx/QueueMBean.java
```diff
@@ -35,7 +39,10 @@
     private final AtomicLong totalRemovedItemCount = new AtomicLong();
     private final String registrationId;
 
-    protected QueueMBean(IQueue managedObject, ManagementService service) {
+    private StatSupplier<LocalQueueStats> localQueueStatsSupplier;
```
ahmetmircik: seems we don't need this as a global variable.

hazelcast/src/main/java/com/hazelcast/jmx/MapMBean.java
```diff
@@ -41,7 +45,10 @@
...
+    private LocalStatsDelegate<LocalMapStats> localMapStatsDelegate;
```
ahmetmircik: seems this can be a `final`

ahmetmircik:
:+1:

emrahkocaman:
:+1:

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -182,6 +182,7 @@
     public static final String PROP_LOGGING_TYPE = "hazelcast.logging.type";
     public static final String PROP_ENABLE_JMX = "hazelcast.jmx";
     public static final String PROP_ENABLE_JMX_DETAILED = "hazelcast.jmx.detailed";
+    public static final String PROP_JMX_UPDATE_INTERVAL_SECONDS = "hazelcast.update.jmx.interval.seconds";
```
mdogan: `hazelcast.jmx.update.interval.seconds`?

hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
```diff
@@ -0,0 +1,71 @@
...
+                try {
+                    localStats = supplier.get();
+                    lastUpdated.set(System.currentTimeMillis());
+                } finally {
+                    inProgress.set(false);
+                }
+            }
+        }
+
+        while (localStats == null && inProgress.get()) {
```
mdogan: What's the purpose of `localStats == null` check?

muratayan: If two threads enter at the same time at the beginning, and the thread that prepares the stats for the first time stops etc.

mdogan: But `localStats` is null only at the first time, right? Do you want to make caller wait only when `localStats == null`? 
If yes, then what `&& inProgress.get()` is for?

muratayan: I think that should be `|| inProgress.get()`. `inProgress.get()` is for, if calculation is already on the fly, we should wait for it.

hazelcast/src/main/java/com/hazelcast/jmx/MapMBean.java
```diff
@@ -74,6 +80,8 @@ public void mapCleared(MapEvent event) {
             }
         };
         listenerId = managedObject.addEntryListener(entryListener, false);
+        StatSupplier<LocalMapStats> localMapStatsSupplier = new LocalMapStatsSupplier(managedObject);
```
mdogan: Isn't making `MapMBean` to implement `StatSupplier<LocalMapStats>` simpler? instead of creating another object?

hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
serkan-ozal: Doesn't `localStats == null || inProgress.get()` cause infinite loop when `getLocalStats` is called first time and `delta > intervalMs`? Because in this case `localStats` is still null.

serkan-ozal: but anyway, it might be called eventually by another thread so it finishes

muratayan: You are right, in that case I was assuming another thread will make it not null because `inProgress.get()` will be setted to `false` no matter what.

mdogan: I think there's no need to block caller unless `localStats == null`. At worst it will return an old stats which will be updated very soon. 
And if you can make `localStats` non-null at the beginning, in constructor, then you will get rid of that waiting block completely.

hazelcast/src/main/java/com/hazelcast/jmx/MapMBean.java
muratayan: It might be make it more complicated. Will it hold a reference to itself that way?

`this.localMapStatsDelegate = new LocalStatsDelegate<LocalMapStats>(**this**, updateInterval);`

hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
mdogan: To me, blocking just to get latest stats is overkill. just return what you have.

muratayan: @ahmetmircik Also said it might be ok to return one-step-older stats. So I am changing it this way then.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
Donnerbart: BTW, this should not be added here anymore. All those String constants are deprecated. Just add an entry to `GroupProperty`.

hazelcast/src/main/java/com/hazelcast/jmx/LocalStatsDelegate.java
muratayan: @mdogan What if we get an exception in the constructor?

mdogan: What I prefer would be creating an empty stats object in c'tor. I don't know if that's possible.

muratayan: This would do it, yeah?

https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/monitor/impl/LocalMapStatsImpl.java#L94

mdogan: then you need to find a way to get that empty stats from `StatSupplier`.

hazelcast/src/main/java/com/hazelcast/jmx/suppliers/StatSupplier.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.jmx.suppliers;
+
+/**
+ * Interface used in MBeans for retrieving the Local Stats
+ * @param <T> LocalStats type
+ */
+public interface StatSupplier<T> {
```
mdogan: Isn't `StatsSupplier` more suitable name? if you think what it returns is `LocalStats`...

hazelcast/src/main/java/com/hazelcast/jmx/suppliers/LocalMapStatsSupplier.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final IMap map;
+
+    public LocalMapStatsSupplier(IMap map) {
+        this.map = map;
+    }
+
+    @Override
+    public LocalMapStats getEmpty() {
+        return new LocalMultiMapStatsImpl();
```
mdogan: `LocalMapStatsImpl`?

mdogan:
:+1:

PR #7331, opened 2016-01-14
[DONT MERGE] Zero copy on client write handler

sancar:
With this change, the byterray that is created at the proxy level is  written directly to socketChannel without any copy. BUT , SimpleMapTest with two nodes shows %16 performance degradation
with this improvement!

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientWriteHandler.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientWriteHandler.java
@@ -34,17 +34,14 @@ public class ClientWriteHandler extends AbstractClientSelectionHandler implement
 
     private final AtomicBoolean informSelector = new AtomicBoolean(true);
 
-    private final ByteBuffer buffer;
-
     private boolean ready;
 
-    private ClientMessage lastMessage;
+    private ByteBuffer lastMessage;
 
     private volatile long lastHandle;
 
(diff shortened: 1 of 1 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientWriteHandler.java
```diff
@@ -69,28 +66,15 @@ public void handle() throws Exception {
     }
 
     private void writeBuffer() throws IOException {
-        while (buffer.hasRemaining() && lastMessage != null) {
-            boolean complete = lastMessage.writeTo(buffer);
-            if (complete) {
+        while (lastMessage != null) {
+            socketChannel.write(lastMessage);
```
ihsandemir: I think that the performance drop may be caused due to change in number of socket write calls. In the new version, you make a call for each message. Let's assume that the number of bytes in the message is 1024 on average, then if buffer is 1024 K bytes, you increase the number of IO system call from 1024 K / 024 = 1024 instead of 1. So, we may need to reconsider this. May be this is a better choice to use an extra buffer :) We may also try this for C++ client as well.

jerrinot: @ihsandemir: so it disables "smart-batching", is it what you are saying? 
it makes sense as copying a few bytes and there is cheaper than a system call.

PR #7333, opened 2016-01-15
Don't send cache invalidation events if the node is backup for that partition

serkan-ozal:
Fixes https://github.com/hazelcast/hazelcast-enterprise/issues/701


ahmetmircik:
:+1:

PR #7334, opened 2016-01-15
Invalid partition id should not crash partition threads

mdogan:
When an operation, a partition-specific-runnable or a packet with an
invalid partition id (greater than max partition id) is submitted to
operation executor, it causes partition operation thread to crash with
ArrayIndexOutOfBoundsException. After that point, no task can be executed
on any that thread and also thread's task queue can lead to oome.

Even this can be considered a security failure, one can send a corrupted
client message to a server and server becomes stuck.

see zendesk tickets: 1361 & 1370


eminn:
:+1:

metanet:
:+1:

sancar:
:+1:

ahmetmircik:
:100:

Donnerbart:
Nice catch!

:+1:

PR #7340, opened 2016-01-15
delete legacy client

asimarslan:
This PR removes client-legacy module and client-legacy related code from hazelcast module.

More refactoring can be applied after this PR:

* Portable Hooks can be cleaned
* some portables can be made IdentifiedDataserializable 
* dead code cleanup


Donnerbart:
Maybe we should do a diff between the both clients to check that the still shared files and tests have been fixed equally.

sancar:
:+1:

PR #7351, opened 2016-01-18
Fixes #7347

### LINKED ISSUE #7347: DiscoveryService start and destroy methods should be called during Client start and shutdown when Discovery SPI is enabled.

kobalski:


hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
```diff
@@ -193,6 +194,10 @@ public HazelcastClientInstanceImpl(ClientConfig config,
             addressProviders.add(new DiscoveryAddressProvider(discoveryService));
         }
 
+        if (Boolean.valueOf(getClientConfig().getProperty(GroupProperty.DISCOVERY_SPI_ENABLED))) {
```
pveentjer: @Donnerbart can't we clean this up?

Donnerbart: Yes, this should work in this context:
`if (clientProperties.getBoolean(GroupProperty.DISCOVERY_SPI_ENABLED)) {`

kobalski: @Donnerbart `DISCOVERY_SPI_ENABLED` was not defined in `ClientProperty`. I added the definition. Can you look at the PR again?  Is this the correct way?
``` 
if (clientProperties.getBoolean(ClientProperty.DISCOVERY_SPI_ENABLED)) {
       discoveryService.start();
}
```

bilalyasar:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
Donnerbart: I think it should be fine, yes.

Donnerbart:
In the test I just see the `hazelcast.discovery.enabled` property being set on the server config. Is the client side code `discoveryService.start();` executed anyway?

kobalski:
@Donnerbart Good catch. Since both client and server use the same `DiscoveryStrategy`, `start` and `destroy` was called by the server not client. I corrected the test. Increased the latch count to 2 and since `hazelcast.discovery.enabled` is set in both server and client, I expect `start` and `destroy` methods to be called twice. I updated test name accordingly.

Donnerbart:
Okay, then it's a :+1: from me.

I would always run the tests with code coverage in IDEA, so you can easily see if your (new) code is executed. It should not be more work than having the `Coverage` plugin installed and push another button to run the test:

run-with-coverage

PR #7355, opened 2016-01-18
Fix backport typo in NearCache heap cost calculation

cangencer:
Fixes bug in backport #7110


mdogan:
:+1:

pveentjer:
Can you also copy the test to the backport? Once that is done, :+1: from me

PR #7356, opened 2016-01-19
Remove logging of client requests failures

sancar:
These exceptions are already send to client and handled on client
side.


pveentjer:
:+1:

sancar:
@pveentjer Sorry but I have changed the pr on the fly. I found a couple of places where try catch is used but actually not needed. And instead of deleting the logging. I made all logged in FINEST level. You may need to re-review.

pveentjer:
:+1:

mdogan:
:+1:

PR #7359, opened 2016-01-19
Remove logging of client requests failures

sancar:
These exceptions are already send to client and handled on client
side.
 I found a couple of places where try catch is used but actually not needed. And instead of deleting the logging. I made all logged in FINEST level. 

Backport of https://github.com/hazelcast/hazelcast/pull/7356


sancar:
verify

sancar:
verify

pveentjer:
:+1:

mdogan:
:+1:

PR #7365, opened 2016-01-19
[3.7] ILock expiry operation should be able run both owner and backup

mdogan:
UnlockIfLeaseExpiredOperation runs on both owner and backup nodes. When expiry
isn't triggered on a backup and owner dies just around expiry time, before sending
unlocking backup operation, lock remains as locked even though it's supposed to be expired.
That's why `Operation.validateTarget` is set to false. Otherwise it doesn't
run on backup node and logs a `WrongTargetException`.


gurbuzali:
backport is already merged

PR #7370, opened 2016-01-19
Removes Map representation from protocol

sancar:
This change does not affect wire level protocol.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -1013,7 +1013,7 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
             try {
                 ClientMessage response = future.get();
                 MapGetAllCodec.ResponseParameters resultParameters = MapGetAllCodec.decodeResponse(response);
-        
+
                 for (Entry<Data, Data> entry : resultParameters.response) {
                     final V value = toObject(entry.getValue());
                     final K key = toObject(entry.getKey());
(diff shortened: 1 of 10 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1361,21 +1362,21 @@ public void putAll(Map<? extends K, ? extends V> m) {
             Data keyData = toData(entry.getKey());
 
             int partitionId = partitionService.getPartitionId(keyData);
-            Map<Data, Data> partition = entryMap.get(partitionId);
+            List<Map.Entry<Data, Data>> partition = entryMap.get(partitionId);
             if (partition == null) {
-                partition = new HashMap<Data, Data>();
+                partition = new LinkedList<Entry<Data, Data>>();
```
LoneRifle: Given http://stackoverflow.com/questions/322715/when-to-use-linkedlist-over-arraylist , would it make better sense to use ArrayList here? I get that if the backing array has to be resized we incur a time penalty, but that may be mitigated by estimating the size needed on construction.

cangencer:
Does the code templates for other languages need updating also to support "MAPENTRY"?

sancar:
Yes, they need to be updated. Since I removed "MAP", code generation will be broken for other languages. But I could not do it because I don't know what to put instead of new AbstractMap.Entry class.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1361,21 +1361,21 @@ public void putAll(Map<? extends K, ? extends V> m) {
             Data keyData = toData(entry.getKey());
 
             int partitionId = partitionService.getPartitionId(keyData);
-            Map<Data, Data> partition = entryMap.get(partitionId);
+            List<Map.Entry<Data, Data>> partition = entryMap.get(partitionId);
             if (partition == null) {
-                partition = new HashMap<Data, Data>();
+                partition = new ArrayList<Entry<Data, Data>>();
```
ihsandemir: is this supposed to be Map.Entry instead of just Entry?

sancar:
run-lab-run

cangencer:
:+1:

sancar:
run-lab-run

PR #7373, opened 2016-01-20
Improved BufferPoolThreadLocal

pveentjer:
The previous implementation was very inefficient due to ConcurrentReferenceHashMap. The new implementation relies on a ThreadLocal followed by a WeakReference on the happy path. 

On my machine the following test:
```
 @Test
    public void testLong() {
        Long value = new Long(234533455);

        long startMs = System.currentTimeMillis();
        for (int k = 0; k < iterations; k++) {
            Data data = serializationService.toData(value);
            serializationService.toObject(data);

            if(k%10000000==0){
                System.out.println("at:"+k);
            }
        }
        long durationMs = System.currentTimeMillis() - startMs;
        double performance = (1000.0 * iterations) / durationMs;
        System.out.println("Performance: " + performance + " serializations/second");
    }
```

 gives the following performance improvements:

before:
Performance: 3.674.107 serializations/second

after:
Performance 5.081.817 serializations/second

That is a 38% performance improvement

This week I'll prepare another PR that pushes it to 6M serialization/seconds by improving serializer lookup.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/bufferpool/BufferPoolThreadLocal.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/bufferpool/BufferPoolThreadLocal.java
@@ -16,47 +16,97 @@
 
 package com.hazelcast.internal.serialization.impl.bufferpool;
 
+import com.hazelcast.core.HazelcastInstanceNotActiveException;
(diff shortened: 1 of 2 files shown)


pveentjer:
@jerrinot can you have a look at this.

jerrinot:
:+1:

tombujok:
:+1:

PR #7374, opened 2016-01-21
Moved evicting entries to the end of the loadRecordOrNull method

gurbuzali:
Purpose of this change is that if the loaded entry is picked for eviction, we add the index regardless of the eviction. With this PR, eviction happens at the end so that it can clear indexes of the evicted entry


ahmetmircik:
:+1:

ahmetmircik:
run-lab-run

jerrinot:
:+1:

PR #7376, opened 2016-01-21
Introduce a way to test hearbeat features of client

sancar:
This is currently only avaliable for mocked network tests.
One bug fixed and related tests are added.

Fixed bug is related to async invocations made from client.
ClientHeartbat check was used to be made from only `future.get` .
When user does not call `future.get` method, heartbeat never kicks
in to make the waiting party free. Related logic is moved to
lientInvocationServiceSupport via ConnectionHeartbeatListener.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.heartbeat;
+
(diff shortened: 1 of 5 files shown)


sancar:
run-lab-run

hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
```diff
@@ -0,0 +1,147 @@
...
+        clientConfig.setProperty(ClientProperty.HEARTBEAT_TIMEOUT, "3000");
+        clientConfig.setProperty(ClientProperty.HEARTBEAT_INTERVAL, "500");
+        HazelcastInstance client = hazelcastFactory.newHazelcastClient(clientConfig);
+        HazelcastInstance instance2 = hazelcastFactory.newHazelcastInstance();
+
+        IMap map = client.getMap(randomString());
+        String keyOwnedByInstance2 = generateKeyOwnedBy(instance2);
+        map.put(keyOwnedByInstance2, randomString());
+        blockMessagesFromInstance(instance2, client);
+        ClientDelegatingFuture future = (ClientDelegatingFuture) map.putAsync(keyOwnedByInstance2, randomString());
```
ihsandemir: Don't you need to do async call before blocking the connection? Don't you need to simulate that the request arrived at the server but the response did not arrive due to connection.

sancar: I am blocking when response turning back.  

How can I make sure `getAsync` not finished when I call to `blockMessagesFromInstance` method after `getAsync` ? Before executing next line, it could be the case that response already got back.

ihsandemir: OK, i thought that blockMessagesFromInstance also blocked the outgoing messages from client to server but this is not correct, right? It only blocks the incoming messages. Then it looks fine.

ihsandemir:
:+1:

mdogan:
:+1:

PR #7377, opened 2016-01-21
Introduce a way to test hearbeat features of client

sancar:
This is currently only avaliable for mocked network tests.
One bug fixed and related tests are added.

Fixed bug is related to async invocations made from client.
ClientHeartbat check was used to be made from only `future.get` .
When user does not call `future.get` method, heartbeat never kicks
in to make the waiting party free. Related logic is moved to
lientInvocationServiceSupport via ConnectionHeartbeatListener.


sancar:
verify

sancar:
run-lab-run

Issue #7210, opened 2015-12-30
Select_TcpIpConnection_TransferStressTest.testLargePackets


hasancelik:
```
java.lang.AssertionError: Thread WriteThread-1 is still alive
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at com.hazelcast.test.TestThread$1.run(TestThread.java:52)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:835)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:849)
	at com.hazelcast.test.TestThread.assertTerminates(TestThread.java:49)
	at com.hazelcast.test.TestThread.assertSucceedsEventually(TestThread.java:61)
	at com.hazelcast.nio.tcp.TcpIpConnection_TransferStressBaseTest.test(TcpIpConnection_TransferStressBaseTest.java:84)
	at com.hazelcast.nio.tcp.TcpIpConnection_TransferStressBaseTest.testLargePackets(TcpIpConnection_TransferStressBaseTest.java:61)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:88)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:82)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.lang.Thread.run(Thread.java:745)
```

[...]

PR #7378, opened 2016-01-21
Increase timeout for `TcpIpConnection_TransferStressBaseTest::testLargePackets`

serkan-ozal:
I have added logs (total/empty select count, passed time on selector.select, ...) inside IO threads and have seen that sometimes there were higher passed time than usual so the last verification phase (waits for pending packets to be sent and received) couldn't finish in time (2 mins). I have seen that there were more CPU/user times in IO threads and wait time on selector until getting selected keys. I have also seen that there were some high CPU usage issues related with JDK on UNIX for selector epoll wait but as I read, they had been fixed.

Anyway, even in these slow cases, the verification phase eventually finishes always.
So I have increased the timeout for testing large packets and made two 5 hours (10 hours in total) run without any fail.

Fixes https://github.com/hazelcast/hazelcast/issues/7210

Also, thread CPU/user times and waited/blocked times are printed if they are available inside thread dump on test exception.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/nio/tcp/TcpIpConnection_TransferStressBaseTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/nio/tcp/TcpIpConnection_TransferStressBaseTest.java
@@ -21,14 +21,14 @@ import static org.junit.Assert.assertEquals;
  */
 public abstract class TcpIpConnection_TransferStressBaseTest extends TcpIpConnection_AbstractTest {
 
-    // total running time
-    private static final long DURATION_SECONDS = TimeUnit.MINUTES.toSeconds(2);
+    // total running time for writer threads
+    private static final long WRITER_THREAD_RUNNING_TIME_IN_SECONDS = TimeUnit.MINUTES.toSeconds(2);
     // maximum number of pending packets
     private static final int maxPendingPacketCount = 10000;
     // we create the payloads up front and select randomly from them. This is the number of payloads we are creating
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/nio/tcp/TcpIpConnection_TransferStressBaseTest.java
```diff
@@ -91,14 +95,25 @@ public void test() throws Exception {
...
-        final SocketReader XReadHandler = ((TcpIpConnection) connManagerB.getConnection(addressA)).getSocketReader();
+        final SocketWriter writer = c.getSocketWriter();
+        final SocketReader reader = ((TcpIpConnection) connManagerB.getConnection(addressA)).getSocketReader();
+        long start = System.currentTimeMillis();
         assertTrueEventually(new AssertTask() {
             @Override
             public void run() throws Exception {
-                assertEquals(expectedNormalPackets, XReadHandler.getNormalFramesReadCounter().get());
-                assertEquals(expectedUrgentPackets, XReadHandler.getPriorityFramesReadCounter().get());
+                logger.info("writer total frames pending   : " + writer.totalFramesPending());
```
mesutcelik: are those loggings for debugging?

serkan-ozal: yes, to see the progress when there is failure

pveentjer:
:+1:

PR #7382, opened 2016-01-21
Client txn deserialization on server-side fix

gurbuzali:
Client transaction mechanism relies on server-side transaction mechanism, client transaction proxies basically delegates calls to server-side transaction proxies. Because of this design values are deserialized on server-side. 
For example a call to `TransactionalMap.put` from client side, sends a message to the server which calls the server-side transactional-map-proxy's `put` method. And server-side transactional-map-proxy deserializes the value before returning.

The aim of this PR is to mark transactional-proxies on server-side which are created by clients, so that deserialization of the values can be bypassed for those proxies. In order to achieve that first I've centralized transactional-proxies by extending them from `TransactionalDistributedObject`. In this class we check if the proxy originated from client and bypass the deserialization.

PS: This PR may affect the performance positively since we are removing deserialization/serialization cycle.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/txn/ClientTxnMapTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/txn/ClientTxnMapTest.java
@@ -20,6 +20,9 @@ import com.hazelcast.client.test.TestHazelcastFactory;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.IMap;
 import com.hazelcast.core.TransactionalMap;
+import com.hazelcast.nio.ObjectDataInput;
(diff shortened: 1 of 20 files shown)


sancar:
:+1:

gurbuzali:
build passed
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/new-lab-fast-pr/1035/

jerrinot:
:+1:

codingtony:
Is it possible that this issue occurs also when using ReplicatedMap ? 
(I'm currently running 3.7.1) and recently changed from IMap to Replicated map and starting to get class not found exception during serialization.

Issue #3338, opened 2014-08-20
Parallel execution of MapStore#store method for the same key triggered by IMap#flush


atschofen:
Hi, 

this issue exists in Hazelcast 3.3-EA and also a build of the 3.3 development branch from August 7, 2014. It is related to issue #2128.

If you have a map with write-behind and a map store configured (eviction is not needed), and you call the flush method in the IMap, the map store's store method can be called concurrently for the same key, namely for those keys which are in the write-behind queue and then forcibly stored by the flush. This is because the flush operation storing all entries in the write-behind queue seems to be executed in the operation thread, while the periodic processing of the write-behind queue is done by an executor service defined in the WriteBehindQueueManager.

The following piece of code is a unit test to reproduce the issue:

```java
public class TestMapStore16 extends TestCase {

	private static final String mapName = "testMap" + TestMapStore15.class.getSimpleName();

	private static final int writeDelaySeconds = 1;

	@Override
	protected void setUp() throws Exception {

		// configure logging
		if (!TestHazelcast.loggingInitialized) {
			TestHazelcast.loggingInitialized = true;
			BasicConfigurator.configure();
		}
	}

	public void testNoStoreConcurrency() throws Exception {

		// create shared hazelcast instance config
		final Config config = new XmlConfigBuilder().build();
		config.setProperty("hazelcast.logging.type", "log4j");

		// create shared map store implementation
		SlowConcurrencyCheckingMapStore store = new SlowConcurrencyCheckingMapStore();

		// configure map store
		MapStoreConfig mapStoreConfig = new MapStoreConfig();
		mapStoreConfig.setEnabled(true);
		mapStoreConfig.setWriteDelaySeconds(writeDelaySeconds);
		mapStoreConfig.setClassName(null);
		mapStoreConfig.setImplementation(store);
		MapConfig mapConfig = config.getMapConfig(mapName);
		mapConfig.setMapStoreConfig(mapStoreConfig);

		// start hazelcast instance
		HazelcastInstance hcInstance = Hazelcast.newHazelcastInstance(config);

		IMap<String, String> testMap = hcInstance.getMap(mapName);

		// This will trigger a write-behind store in roughly writeDelaySeconds, the store itself is artificially delayed to take 10 seconds
		testMap.put("key", "value");
		// Wait until the store operation has started
		Thread.sleep((writeDelaySeconds + 2) * 1000);
		// Flush the map, causing the not yet stored entries to be stored 
		testMap.flush();
		
[...]

PR #7383, opened 2016-01-22
Fixes concurrent store operation occurence probability upon IMap#flush

ahmetmircik:
- With this PR partition threads will not be blocked by IMap#flush operations any more (previously IMap#flush operations were executed in partition threads.) What we do now is, marking recordstores write-behind-queues as flushable by calling IMap#flush and periodically checking if the write-behind-queues are all flushed. This periodic check will be done from user thread by calling `CheckIfPartitionFlushedOperation`.

- Also this PR includes some changes to reduce code complexity.

closes https://github.com/hazelcast/hazelcast/issues/3338


gurbuzali:
:+1:

eminn:
:+1:

jerrinot:
can't this lead to significantly higher latencies on `flush`?

PR #7387, opened 2016-01-22
Renamed HazelcastInstanceFactory to HazelcastInstanceManager

pveentjer:
A factory creates objects and doesn't track them. The current implementation
is more of a manager since it manages a lot of access and lifecycle methods.

This renaming if pre-work for JET integration; to prevent making this PR
too large.


Donnerbart:
:+1:

Simple renaming and one import cleanup in a test file. Should be mergeable with a single review once green.

jerrinot:
this requires EE-counterpart.

pveentjer:
EE change added.

https://github.com/hazelcast/hazelcast-enterprise/pull/713

PR #7388, opened 2016-01-22
MultiMap entry listener should only listen for events for specific key

cangencer:
When a listener is added to a MultiMap from client to listen for events
on a specific key, events on all keys were being emitted instead.


sancar:
:+1:

asimarslan:
:1

cangencer:
run-lab-run

PR #7392, opened 2016-01-24
Reuse computePerturbationValue()

mtopolnik:
Code moved from other classes in Enterprise so it can be reused.


jerrinot:
:+1: (after labels are added)

PS: Is this really passing checkstyle?

mtopolnik:
As a trivial PR (pure movement of existing code and satisfaction of CheckStyle), I'm merging now.

PR #7399, opened 2016-01-25
Improved PerformanceMonitor

pveentjer:
It includes a lot more flexible monitortask plugin system and it contains a bunch of interesting tasks like packet sampler and invocation occurrence counter that count the number of pending operations aggregated based on occurrence.

A lot of work was done to make sure the PerformanceMonitor/PerformanceLog when enabled, doesn't generate a lot of litter. Most of the litter created is from the MetricsRegistry due to reflection method calls. Unfortunately all returned primitives are wrapped in an object.


hazelcast/src/main/java/com/hazelcast/internal/monitors/PendingInvocationsMonitorTask.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    private void render(BufferedWriter writer) throws IOException {
+        writer.write("PendingInvocations[");
+        writeKeyValue(writer, "count", invocationRegistry.size());
+        writer.write(getIndent() + "invocations[");
+        incIndentLevel();
+
+        for (Class op : occurrenceMap.keySet()) {
```
jerrinot: it would be nice to sort categories by number of occurencies

jerrinot:
It would be nice to have an ability to monitor reasons why invocations are pending. 

I observed issues with huge numbers (800+) of smart clients. What I see is that most of pending invocations have a response already set (because the target member is local), but they are waiting on backup response.

hazelcast/src/main/java/com/hazelcast/internal/metrics/renderers/CommaSeparatedKeyValueProbeRenderer.java
```diff
@@ -32,7 +32,7 @@
     @Override
     public void start() {
         sb.setLength(0);
-        sb.append("{").append("time").append("=").append(System.currentTimeMillis());
+        sb.append("").append("time").append("=").append(System.currentTimeMillis());
```
Donnerbart: This line could need some optimizing :)
`sb.append("time=").append(System.currentTimeMillis());`

hazelcast/src/main/java/com/hazelcast/internal/monitors/BuildInfoPlugin.java
```diff
@@ -0,0 +1,60 @@
...
+    private final BuildInfo buildInfo = BuildInfoProvider.getBuildInfo();
+    private final ILogger logger;
+
+    public BuildInfoPlugin(NodeEngineImpl nodeEngine) {
+        this.logger = nodeEngine.getLogger(BuildInfoPlugin.class);
+    }
+
+    @Override
+    public long getPeriodMillis() {
+        return -1;
```
Donnerbart: This looks like a magic number. I think I'll see this further in the code, but if this means "execute once" it would be nice to have a constant in `PerformanceMonitorPlugin` for this special value.

hazelcast/src/main/java/com/hazelcast/internal/monitors/BuildInfoPlugin.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public void onStart() {
+        logger.info("Plugin:active, period-seconds:" + getPeriodMillis());
```
Donnerbart: `period-seconds` in the String vs. `getPeriodMillis()` as method name

hazelcast/src/main/java/com/hazelcast/internal/monitors/MetricsPlugin.java
```diff
@@ -0,0 +1,107 @@
...
+    private final MetricsRegistry metricsRegistry;
+    private final int periodSeconds;
+    private final ILogger logger;
+    private final NodeEngineImpl nodeEngine;
+    private ProbeRenderer probeRenderer;
+
+    public MetricsPlugin(NodeEngineImpl nodeEngine) {
+        this.nodeEngine = nodeEngine;
+        this.metricsRegistry = nodeEngine.getMetricsRegistry();
+        this.periodSeconds = nodeEngine.getGroupProperties().getInteger(PERFORMANCE_MONITOR_METRICS_PERIOD_SECONDS);
```
Donnerbart: You could use `getMillis()` instead of `getInteger()` here, so you don't have to convert the value later.

hazelcast/src/main/java/com/hazelcast/internal/monitors/OverloadedConnectionsPlugin.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.Set;
+
+import static com.hazelcast.instance.GroupProperty.PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_PERIOD_SECONDS;
+import static com.hazelcast.instance.GroupProperty.PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_SAMPLES;
+import static com.hazelcast.instance.GroupProperty.PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_THRESHOLD;
+import static java.lang.Math.min;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * The PacketSampler is a debug aid that periodically scans the Connection.writeQueues and samples the content.
```
Donnerbart: `PacketSampler` in JavaDoc vs. `OverloadedConnectionsPlugin` as class name

hazelcast/src/main/java/com/hazelcast/internal/monitors/OverloadedConnectionsPlugin.java
```diff
@@ -0,0 +1,229 @@
...
+    private Set<TcpIpConnection> getTcpIpConnections() {
+        ConnectionManager connectionManager = nodeEngine.getNode().getConnectionManager();
+        if (connectionManager instanceof TcpIpConnectionManager) {
+            return ((TcpIpConnectionManager) connectionManager).getActiveConnections();
+        } else {
+            return Collections.emptySet();
+        }
+    }
+
+    private void scan(BufferedWriter w, TcpIpConnection connection, boolean priority) throws IOException {
```
Donnerbart: I would prefer `writer` instead of `w`, since that name is used in all other methods of this class.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -296,4 +296,5 @@ private boolean hasMemberLeft(Invocation invocation) {
             }
         }
     }
+
```
Donnerbart: Useless new empty line? :D

hazelcast/src/test/java/com/hazelcast/Main.java
```diff
@@ -0,0 +1,72 @@
...
+import com.hazelcast.map.EntryBackupProcessor;
+import com.hazelcast.map.EntryProcessor;
+import com.hazelcast.test.HazelcastTestSupport;
+
+import java.util.Map;
+
+/**
+ * Created by alarmnummer on 1/23/16.
+ */
+public class Main extends HazelcastTestSupport {
```
Donnerbart: Should this class stay in the codebase? Looks like a WIP for a new test.

hazelcast/src/test/java/com/hazelcast/internal/monitors/OverloadedConnectionsPluginTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static com.hazelcast.instance.GroupProperty.PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_PERIOD_SECONDS;
+import static com.hazelcast.instance.GroupProperty.PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_SAMPLES;
+import static com.hazelcast.instance.GroupProperty.PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_THRESHOLD;
+
+/**
+ * This test can't run with test-hazelcast instances since we rely on a real TcpIpConnectionManager.
+ */
+@RunWith(HazelcastSerialClassRunner.class)
+@Category(QuickTest.class)
+@Ignore
```
Donnerbart: Is the test intentionally ignored (can't tell by just looking at it, if it should work already)?

hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
```diff
@@ -148,6 +141,21 @@
      */
     PERFORMANCE_MONITOR_HUMAN_FRIENDLY_FORMAT("hazelcast.performance.monitor.human.friendly.format", true),
 
+    PERFORMANCE_MONITOR_SLOW_OPERATIONS_PERIOD_SECONDS("hazelcast.performance.monitor.slowoperations.period.seconds", 60),
```
Donnerbart: You should define these properties with a `TimeUnit`, so you can benefit from the `getMillis()`, `getSeconds()` methods in `HazelcastProperties`, e.g. `PERFORMANCE_MONITOR_SLOW_OPERATIONS_PERIOD_SECONDS("hazelcast.performance.monitor.slowoperations.period.seconds", 60, SECONDS),`

hazelcast/src/test/java/com/hazelcast/Main.java
pveentjer: Will be deleted. I'm using it to check the content of the generated output.

hazelcast/src/main/java/com/hazelcast/internal/monitors/SlowOperationPlugin.java
```diff
@@ -0,0 +1,125 @@
...
+    public SlowOperationPlugin(NodeEngineImpl nodeEngine) {
+        this.logger = nodeEngine.getLogger(SlowOperationPlugin.class);
+        this.operationService = nodeEngine.getOperationService();
+        this.periodSeconds = getPeriodSeconds(nodeEngine);
+    }
+
+    private long getPeriodSeconds(NodeEngineImpl nodeEngine) {
+        GroupProperties groupProperties = nodeEngine.getGroupProperties();
+        if (!groupProperties.getBoolean(SLOW_OPERATION_DETECTOR_ENABLED)) {
+            return -1;
```
Donnerbart: I think the disabled magic number was `0`, not `-1`. Maybe another good candidate for a constant.

mdogan:
As @jerrinot said, having more data about invocations would be nice, like; 
- how many local and remote
- how many waiting for actual response
- how many have done more than X retries
- how many waiting for backup response
- how many pending per target (?)

mdogan:
Can we also monitor heartbeat timestamps? When each member did send last heartbeat...

pveentjer:
@mdogan I believe that a lot of your nice to haves can be implemented (if they are not already) as metrics. There is a metrics performancemonitor plugin that dumps everything in the metrics registry.

I'll check them and if missing, I'll add missing probes.

The backup stuff I'll add; good to have. 
Pending per target ----> good to have.

hazelcast/src/main/java/com/hazelcast/internal/metrics/renderers/CommaSeparatedKeyValueProbeRenderer.java
```diff
@@ -32,7 +32,7 @@
     @Override
     public void start() {
         sb.setLength(0);
-        sb.append("{").append("time").append("=").append(System.currentTimeMillis());
+        sb.append("").append("time=").append(System.currentTimeMillis());
```
Donnerbart: Line still contains an empty `.append("")`.

checkstyle/suppressions.xml
```diff
@@ -272,6 +272,9 @@
     <suppress checks="MethodCount|NPathComplexity|MagicNumber" files="com/hazelcast/spi/Operation"/>
     <suppress checks="MethodCount" files="com/hazelcast/spi/impl/NodeEngineImpl"/>
     <suppress checks="ClassFanOutComplexity" files="com/hazelcast/spi/impl/NodeEngineImpl"/>
+    <!-- node engine constructs objects, so it knows many of them -->
+    <suppress checks="ClassDataAbstractionCoupling" files="com/hazelcast/spi/impl/NodeEngineImpl"/>
```
mtopolnik: I strongly suggest migrating away from suppressions.xml towards `@SuppressWarnings("checkstyle:classdataabstractioncoupling")`. It makes it much more visible that something was suppressed, plus it discourages wildcard suppressions.

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceMonitor.java
```diff
@@ -16,111 +16,135 @@
...
+        checkNotNull(performanceMonitorPlugin, "monitorTask can't be null");
+
         if (!enabled) {
-            return null;
+            return;
+        }
+
+        long periodMillis = performanceMonitorPlugin.getPeriodMillis();
+
+        logger.finest(performanceMonitorPlugin.getClass().toString() + " is " + (periodMillis == 0 ? "disabled" : "enabled"));
```
Donnerbart: You could use the `DISABLED` in here as well.

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceMonitor.java
```diff
@@ -16,111 +16,135 @@
...
+
+        if (periodMillis == 0) {
```
Donnerbart: Use case for `DISABLED` as well.

hazelcast/src/main/java/com/hazelcast/internal/monitors/PerformanceMonitor.java
```diff
@@ -16,111 +16,135 @@
...
-        HazelcastThreadGroup threadGroup = node.getHazelcastThreadGroup();
-        int delaySeconds = node.getGroupProperties().getSeconds(GroupProperty.PERFORMANCE_MONITOR_DELAY_SECONDS);
+        performanceMonitorPlugin.setSingleLine(singleLine);
+        performanceMonitorPlugin.onStart();
 
-        return new MonitorThread(threadGroup, delaySeconds);
+        if (periodMillis > 0) {
+            // it is a periodic task
+            scheduler.scheduleAtFixedRate(new MonitorTaskRunnable(performanceMonitorPlugin), 0, periodMillis, MILLISECONDS);
+        } else {
```
Donnerbart: This looks nice and clean, but in fact every negative `periodMillis` will execute the plugin once, although it's documented to be `-1` (aka `RUN_ONCE`). Maybe we should check with `else if (periodMillis == RUN_ONCE)` here and throw an `IllegalArgumentException` for all other negative values? Don't know what our policy for wrong configurations is (if there is any).

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -138,6 +148,18 @@ public void start() {
         proxyService.init();
         operationService.start();
         quorumService.start();
+        performanceMonitor.start();
+
+        // static loggers at beginning of file
+        performanceMonitor.register(new BuildInfoPlugin(this));
```
Donnerbart: Maybe we can do an automatic registration via reflection and annotations? With this solution you have to know about this part of code (and NodeEngineImpl has to know about all plugins). We could also get rid of the `ClassDataAbstractionCoupling` suppression. Could be an optimization for another PR.

checkstyle/suppressions.xml
Donnerbart: I'm in favor of the annotation style as well. Another advantage: If you refactor a class you directly see the suppressions and can give it a try to remove them. A suppression in the XML is easily forgotten, even if the class is improved.

But I won't make this a blocker if we can't get a consensus on this. Would be an easy task to convert most of the XML config to annoations in a separate PR.

I also made a suggestion to get rid of this specific suppression for `NodeEngineImpl` (no suppression at all should be the best solution anyway).

Donnerbart:
I like the general design of the new PerformanceMonitor. The plugins are a very good idea to break down this complex task.

There are some minor remarks here and there, but I see no big blocker for an initial version. Missing metrics can be added anytime later, as well as some optimiztations in some of the plugins.

checkstyle/suppressions.xml
mtopolnik: Another case in point: I already had a PR prepared which removes the blanket XML suppression from the complete `c.h.elastic` hierarchy. The PR was delayed until after the release; but in the meantime another PR was accepted which turned out to add new code to the same package hierarchy, even though in a completely separate module. After rebasing I was welcomed with 300 new CheckStyle errors to take care of.

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
pveentjer: THe NodeEngineImpl is acting like a application context; so the application gets constructed here. That is why it is perfectly fine that the NodeEngineImpl is violating this class data abstraction coupling.

hazelcast/src/main/java/com/hazelcast/internal/monitors/MetricsPlugin.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        @Override
+        public void renderDouble(String name, double value) {
+            writer.writeKeyValueEntry(name, value);
+        }
+
+        @Override
+        public void renderException(String name, Exception e) {
+            writer.writeKeyValueEntry(name, e.getClass().getName() + ":" + e.getMessage());
```
mtopolnik: `":"` can be `':'`

mtopolnik:
:+1:

hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
```diff
@@ -114,14 +114,7 @@
      */
     PERFORMANCE_METRICS_LEVEL("hazelcast.performance.metric.level", ProbeLevel.MANDATORY.name()),
 
-    /**
-     * The delay in seconds between monitor of the performance.
-     * <p/>
-     * The default is 30 seconds.
-     */
-    PERFORMANCE_MONITOR_DELAY_SECONDS("hazelcast.performance.monitor.delay.seconds", 30, SECONDS),
```
jerrinot: should we deprecate it / fail-fast? now the property will be simple ignored -> not nice.

hazelcast/src/main/java/com/hazelcast/internal/monitors/MultiLinePerformanceLogWriter.java
```diff
@@ -0,0 +1,191 @@
...
+ * A {@link PerformanceLogWriter} that writes over multiple lines. Useful for human reading.
+ */
+public class MultiLinePerformanceLogWriter extends PerformanceLogWriter {
+
+    private static final String STR_LONG_MIN_VALUE = String.format("%,d", Long.MIN_VALUE);
+
+    private static final char[] DIGITS = new char[]{'0', '1', '2', '3', '4', '5', '6', '7', '8', '9'};
+
+    private static final String[] INDENTS = new String[]{
+            "\n                          ",
```
jerrinot: platform-specific stuff

hazelcast/src/main/java/com/hazelcast/internal/monitors/OverloadedConnectionsPlugin.java
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * Currently the sampling has a lot of overhead since the value needs to be deserialized. That is why this
+ * plugin is disabled by default.
+ */
+public class OverloadedConnectionsPlugin extends PerformanceMonitorPlugin {
+    private static final Queue<OutboundFrame> EMPTY_QUEUE = new LinkedList<OutboundFrame>();
+
+    private final SerializationService serializationService;
+    private final ILogger logger;
+    private final Map<String, MutableInteger> occurrenceMap = new HashMap<String, MutableInteger>();
```
jerrinot: a small suggestion: `ItemCounter` might be a more suitable abstraction for this. It miss iterations, but it would be easier to add.

hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
Donnerbart: Hmm, maybe we can create a section where we can put outdated properties and some kind of help text (reference to new property, explanation etc.). With this we could keep the existing properties clean and have a central place for the historical stuff.

jerrinot:
:+1:  once the platform specific end of lines are removed.

hazelcast/src/main/java/com/hazelcast/internal/monitors/MultiLinePerformanceLogWriter.java
Donnerbart: In Simulator I put this in our `FormatUtils`, so I can use it in the rest of the code:
`public static final String NEW_LINE = System.getProperty("line.separator");`

In Hazelcast it seems to capsulated in this method: `StringUtil.getLineSeperator()`

pveentjer: I don't think it is platform specific.

1: any decent editor on windows or osx will deal with /n correctly
2: this stuff is typically run on linux; but viewed on a range of platforms like windows/linux/osx. So you already got a difference between your running platform and your viewing platform.

jerrinot: fair enough. my opinion is not too strong here.

Donnerbart:
:+1:

hazelcast/src/main/java/com/hazelcast/internal/monitors/OverloadedConnectionsPlugin.java
pveentjer: Can you elaborate?

hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
pveentjer: These should not be officially supported properties. They are there for debugging only

hazelcast/src/main/java/com/hazelcast/internal/monitors/MultiLinePerformanceLogWriter.java
mtopolnik: A sidenote: writing `String.format("%n                              ")` takes care of the line separator, if you ever feel the need for that.

PR #7407, opened 2016-01-26
Don't count cache records as entry count if they are backup

### LINKED ISSUE #7403: Incorrect cache owned entry count in `CacheStatistics`
At the moment, also backup records increase owned cache entry count but they should be ignored.
So `CacheStatistics` returns incorrect entry count via `getOwnedEntryCount`.

serkan-ozal:
Fixes #7403 

I will send backport of this PR and there will be enterprise side also.

EE counter-part: https://github.com/hazelcast/hazelcast-enterprise/pull/723

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/record/CacheRecordHashMap.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/record/CacheRecordHashMap.java
@@ -22,11 +22,9 @@ import com.hazelcast.internal.eviction.Evictable;
 import com.hazelcast.internal.eviction.EvictionCandidate;
 import com.hazelcast.internal.eviction.EvictionListener;
 import com.hazelcast.nio.serialization.Data;
-import com.hazelcast.util.ConcurrentReferenceHashMap;
 import com.hazelcast.util.SampleableConcurrentHashMap;
 
 import java.util.ArrayList;
-import java.util.EnumSet;
 import java.util.List;
 
 public class CacheRecordHashMap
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/record/CacheRecordHashMap.java
```diff
@@ -36,24 +34,36 @@
...
+            if (!entryCountingEnable) {
+                // It was disable before but now it will be enable.
+                // Therefore, we increase the entry count as size of records.
+                cacheContext.increaseEntryCount(size());
+            }
+        } else {
+            if (entryCountingEnable) {
+                // It was enable before but now it will be disable.
+                // Therefore, we decrease the entry count as size of records.
+                cacheContext.decreaseEntryCount(size());
```
ahmetmircik: is there any possibility that we can see `entryCount` negative after this decrease?

serkan-ozal: The decrease part is called on source node after migration is committed. (on rollback migration since there is no primary state change, this code block should not be executed because it expects `enable != entryCountingEnable`) 

I mean the node was the owner of the partition but now it is not owner anymore. But in fact this part is called after `clearPartitionReplica` so when this decrease part is called, there is not record in cache record map because it is already cleared and its counts are already decreased from entry count. 

So in fact this part should not have any effect but I had added it to be consistent with increase part.

ahmetmircik:
:+1:

hazelcast/src/main/java/com/hazelcast/cache/impl/record/CacheRecordHashMap.java
ahmetmircik: maybe an `assert` can be added in any case.

PR #7410, opened 2016-01-27
Hazelcast-Jet integration

toktarev:
1) MessageTaskFactories has been replaced on MessageTaskProvider
2) RemoteServiceProvider has been implemented for custom service loading
3) ClientProxyProvider has been implemented for custom client proxy loading

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi;
+
+/**
+ * Descriptor of Hazelcast client proxies
+ * <p/>
+ * Provided by corresponding provider (see com.hazelcast.client.spi.ClientProxyDescriptorProvider).
+ * Contains information about client proxy service.
+ * <p/>
+ * <p/>
+ * Examples:
+ * <pre>
+ *           JetClientProxyDescriptor
+ *           TreeClientProxyDescriptor
+ *                    ....
+ * </pre>
+ */
+public interface ClientProxyDescriptor {
+    /**
+     * @return - name of the corresponding client service
+     */
+    String getServiceName();
+
+    /**
+     * @return - class of the corresponding proxy class
+     */
+    Class<? extends ClientProxy> getClientProxyClass();
+}
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptorProvider.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptorProvider.java
@@ -14,9 +14,24 @@
  * limitations under the License.
  */
 
-/**
- * <p>Client protocol tasks implementations for set</p>
- */
-@GenerateMessageTaskFactoryProvider package com.hazelcast.client.impl.protocol.task.set;
+package com.hazelcast.client.spi;
 
-import com.hazelcast.annotation.GenerateMessageTaskFactoryProvider;
+
+/**
+ * It will be discovered from some external source (like META-INF/services)
+ * and registered in the system.
+ * <p/>
+ * Examples:
+ * <pre>
+ *           JetClientProxyDescriptorProvider
+ *           TreeClientProxyDescriptorProvider
+ * </pre>
+ * <p/>
+ * Descriptor of Hazelcast proxy descriptors
+ */
+public interface ClientProxyDescriptorProvider {
+    /**
+     * Return registered clientProxy descriptors
+     */
+    ClientProxyDescriptor[] createClientProxyDescriptors();
+}
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java
@@ -87,6 +87,7 @@ import com.hazelcast.util.ExceptionUtil;
 import java.io.IOException;
 import java.lang.reflect.Constructor;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
@@ -100,9 +101,11 @@ import static com.hazelcast.client.config.ClientProperty.INVOCATION_TIMEOUT_SECO
  * The ProxyManager handles client proxy instantiation and retrieval at start- and runtime by registering
  * corresponding service manager names and their {@link com.hazelcast.client.spi.ClientProxyFactory}s.
  */
+@SuppressWarnings("checkstyle:classfanoutcomplexity")
 public final class ProxyManager {
-
+    private static final String PROVIDER_ID = "com.hazelcast.client.spi.ClientProxyDescriptorProvider";
     private static final Class[] CONSTRUCTOR_ARGUMENT_TYPES = new Class[]{String.class, String.class};
+
 
     private final HazelcastClientInstanceImpl client;
     private final ConcurrentMap<String, ClientProxyFactory> proxyFactories = new ConcurrentHashMap<String, ClientProxyFactory>();
@@ -186,6 +189,29 @@ public final class ProxyManager {
             } catch (Exception e) {
                 throw ExceptionUtil.rethrow(e);
             }
+        }
+
+        readProxyDescriptors();
+    }
+
+    private void readProxyDescriptors() {
+        try {
+            ClassLoader classLoader = client.getClientConfig().getClassLoader();
+            Iterator<Class<ClientProxyDescriptorProvider>> iter =
+                    com.hazelcast.util.ServiceLoader.classIterator(PROVIDER_ID, classLoader);
+
+            while (iter.hasNext()) {
+                Class<ClientProxyDescriptorProvider> clazz = iter.next();
+                Constructor<ClientProxyDescriptorProvider> constructor = clazz.getDeclaredConstructor();
+                ClientProxyDescriptorProvider provider = constructor.newInstance();
+                ClientProxyDescriptor[] services = provider.createClientProxyDescriptors();
+
+                for (ClientProxyDescriptor serviceDescriptor : services) {
+                    register(serviceDescriptor.getServiceName(), serviceDescriptor.getClientProxyClass());
+                }
+            }
+        } catch (Exception e) {
+            throw ExceptionUtil.rethrow(e);
         }
     }
 
--- a/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGeneratorMessageTaskFactoryProvider.java
+++ b/hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGeneratorMessageTaskFactoryProvider.java
@@ -16,7 +16,7 @@
 
 package com.hazelcast.client.protocol.generator;
 
-import com.hazelcast.annotation.GenerateMessageTaskFactory;
+import com.hazelcast.annotation.GenerateMessageTaskFactoryProvider;
 import freemarker.cache.ClassTemplateLoader;
 import freemarker.log.Logger;
 import freemarker.template.Configuration;
@@ -47,14 +47,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-@SupportedAnnotationTypes("com.hazelcast.annotation.GenerateMessageTaskFactory")
+@SupportedAnnotationTypes("com.hazelcast.annotation.GenerateMessageTaskFactoryProvider")
 @SupportedSourceVersion(SourceVersion.RELEASE_6)
-public class CodeGeneratorMessageTaskFactory
+public class CodeGeneratorMessageTaskFactoryProvider
         extends AbstractProcessor {
 
     private Filer filer;
     private Messager messager;
-    private Template messageFactoryTemplate;
+    private Template messageFactoryProviderTemplate;
     private Map<String, Map<String, String>> map = new HashMap<String, Map<String, String>>();
 
     private Elements elementUtils;
@@ -73,7 +73,7 @@ public class CodeGeneratorMessageTaskFactory
         Configuration cfg = new Configuration(Configuration.VERSION_2_3_22);
         cfg.setTemplateLoader(new ClassTemplateLoader(getClass(), "/"));
         try {
-            messageFactoryTemplate = cfg.getTemplate("messagefactory-template-java.ftl");
+            messageFactoryProviderTemplate = cfg.getTemplate("messagefactoryprovider-template-java.ftl");
         } catch (IOException e) {
             messager.printMessage(Diagnostic.Kind.ERROR, e.getMessage());
             throw new RuntimeException(e);
@@ -82,7 +82,7 @@ public class CodeGeneratorMessageTaskFactory
 
     @Override
     public boolean process(Set<? extends TypeElement> elements, RoundEnvironment env) {
-        final Set<? extends Element> elementsAnnotatedWith = env.getElementsAnnotatedWith(GenerateMessageTaskFactory.class);
+        final Set<? extends Element> elementsAnnotatedWith = env.getElementsAnnotatedWith(GenerateMessageTaskFactoryProvider.class);
         if (elementsAnnotatedWith == null || elementsAnnotatedWith.size() == 0) {
             return false;
         }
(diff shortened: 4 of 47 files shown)


eminn:
The changeset is quite straightforward and looks good the be merged, but I have a missing pieces in the contextual information, so I guess @pveentjer should also have a look at it.

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
```diff
@@ -29,4 +29,6 @@ HazelcastInstance newHazelcastInstance(Config config, String instanceName,
                                            NodeContext nodeContext) throws Exception;
 
     HazelcastInstance newHazelcastInstance(Config config) throws Exception;
+
+    HazelcastInstanceProxy newHazelcastInstanceProxy(HazelcastInstanceImpl hazelcastInstance);
```
serkan-ozal: Is there another way rather than using `HazelcastInstanceImpl` because AFAIK it is in private API?

toktarev: protected HazelcastInstanceProxy(HazelcastInstanceImpl original)
Proxy constructor asks me to pass here impl.
protected HazelcastInstanceProxy(HazelcastInstanceImpl original)

I have the same for JET

protected JetHazelcastInstanceProxy(HazelcastInstanceImpl original)

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceProxy.java
```diff
@@ -68,7 +68,7 @@
  * </li>
  * </ol>
  */
-public final class HazelcastInstanceProxy implements HazelcastInstance, SerializationServiceSupport {
+public class HazelcastInstanceProxy implements HazelcastInstance, SerializationServiceSupport {
     protected volatile HazelcastInstanceImpl original;
```
tombujok: Are you planning on extending HazelcastInstanceProxy in Jet?
Removing final here may be a bit tricky & hazardous.

toktarev: Yes 
public class JetHazelcastInstanceProxy extends HazelcastInstanceProxy

toktarev: For the back compatibility I need to return something which would extend HazelcastInstanceProxy but with JET functionality

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
serkan-ozal: yes, I know. My point is that adding impl class to interface seems not good way to me. 

Not sure but maybe you can use `HazelcastInstance` in the interface and make a type check and cast in the instance factory.

toktarev: Ok I will fix it

serkan-ozal:
:+1:

jerrinot:
Guys, is there any way to get a coherent picture how the integration between Hazelcast - JET should look like? 

What are all integration points? I'm very concerned about this as we are creating another API into Hazelcast. Stuff we considered to be purely internal is now an integration point. This is something we should plan extremely carefully otherwise it'll bite us. 

I'd love to see a (simple) diagram showing what the integration points are before this is merged.

toktarev:
It is quite simple

Hazelcast                             JetHazelcast
HazelcastInstance               JetHazelcastInstance
Node                                    JetNode
DefaultNodeExtension         JetNodeExtension
EnterpriseNodeExtension    JetEnterpriseNodeExtension
Config                                  JetConfig
HazelcastClient                   JetHazelcastClient
HazelcastClientProxy          JetHazelcastClientProxy
HazelcastInstanceProxy      JetHazelcastInstanceProxy

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceProxy.java
jerrinot: at very least this class should be moved into `internal` package.

pveentjer: Moving to internal could lead to problems with e.g. simulator since we rely on this class for reflection.

I would add a @PrivateAPI annotation on top of the class. That should be sufficient. This instance package is a mixture of public and private classes.

jerrinot: yes. that's why I'm quite reluctant to open it for extensions.
we are in a bad position due having too open internal classes. and now we are opening them even more. 

I'm willing to accept it in the cases where it's necessary. but then I want to see a good reasoning behind it.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi;
+
+/***
+ * Descriptor of Hazelcast client proxies
+ */
+public interface ClientProxyDescriptor {
+    String getServiceName();
```
pveentjer: Can you add javadoc to the methods since this is AFAIK public API.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
```diff
@@ -0,0 +1,26 @@
...
+
+    Class<? extends ClientProxy> getClientProxy();
```
pveentjer: Would it not make more sense to call this method getClientProxyClass since a class is returned, not a proxy instance.

Perhaps the name can even be shortened, since client is already implied by this interface. So getProxyClass.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptorProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi;
+
+
+/***
+ * Descriptor of Hazelcast proxy descriptors
```
pveentjer: Can you add some lines of javadoc how this class fits into the big picture?

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptorProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public interface ClientProxyDescriptorProvider {
+    ClientProxyDescriptor[] createClientProxyDescriptors();
```
pveentjer: And some javadoc here.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptorProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi;
+
+
+/***
```
pveentjer: Do we need 3 of those '*' signs?

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.spi;
+
+/***
```
pveentjer: Do we need 3 of those '*' signs?

hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java
```diff
@@ -100,10 +101,12 @@
  * The ProxyManager handles client proxy instantiation and retrieval at start- and runtime by registering
  * corresponding service manager names and their {@link com.hazelcast.client.spi.ClientProxyFactory}s.
  */
+@SuppressWarnings("checkstyle:classfanoutcomplexity")
 public final class ProxyManager {
-
+    private static final String PROVIDER_ID = "com.hazelcast.client.spi.ClientProxyDescriptorProvider";
```
pveentjer: Why not use the ClientProxyDescriptorProvider.getName?

hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java
```diff
@@ -187,6 +190,29 @@ public ClientProxy create(String id) {
                 throw ExceptionUtil.rethrow(e);
             }
         }
+
+        readProxyDescriptors();
+    }
+
+    private void readProxyDescriptors() {
```
pveentjer: Should this method not be called readServiceDescriptor?

hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java
```diff
@@ -187,6 +190,29 @@ public ClientProxy create(String id) {
...
+        try {
+            ClassLoader classLoader = client.getClientConfig().getClassLoader();
+            Iterator<Class<ClientProxyDescriptorProvider>> iter =
+                    com.hazelcast.util.ServiceLoader.classIterator(PROVIDER_ID, classLoader);
+
+            while (iter.hasNext()) {
+                Class<ClientProxyDescriptorProvider> clazz = iter.next();
+                Constructor<ClientProxyDescriptorProvider> constructor = clazz.getDeclaredConstructor();
```
pveentjer: It would be best to do this step by step and help out the user if something goes wrong. For example if the user fails to provide a no arg-constructor, there should be a clear message.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/MessageTaskFactoryProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.impl.protocol;
+
+/***
+ * Provider for client's Message task factories;
```
pveentjer: Can you add some comment how this interface fits into the picture?

Should this class not be in the com.hazelcast.client package since AFAIK it is public API. If outside developer wants to add support for a new data-structure, he needs to make a implementation of the MessageTaskFactoryProvider interface.

hazelcast/src/main/java/com/hazelcast/spi/impl/servicemanager/RemoteServiceDescriptor.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.spi.impl.servicemanager;
+
+import com.hazelcast.spi.NodeEngine;
+import com.hazelcast.spi.RemoteService;
+
+/***
+ * Descriptor of Hazelcast remote service
```
pveentjer: Can you add some documentation how this class fits into the big picture? 

Also this interface AFAIK is public API, because if a user wants to integrate his service, like TreeService, he needs to come up with an implementation.

And do we need an interface at all here? I think that RemoteServiceDescriptor can better be modified as a final class since it has just 2 fields. The only reason you need this class is because you need to have a pair <ServiceName,RemoteService>. So it is just a 'dumb' record.

hazelcast/src/main/java/com/hazelcast/spi/impl/servicemanager/RemoteServiceDescriptorProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.spi.impl.servicemanager;
+
+/***
+ * Provider for the remote service descriptors;
+ * Will be read from META-INF/services
+ */
+public interface RemoteServiceDescriptorProvider {
```
pveentjer: This class is public API AFAIK and should be moved somewhere outside of an impl package.

Can you add some documentation how this class fits into the big picture? If I'm a third party developer, I want to create my own e.g. Tree Service.. then I need to implement this class and do something with it (put it in special location).

pveentjer:
Can you also add tests.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java
toktarev: Because we read proxies here, not services

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
toktarev: It is some sign for javaDoc, isn't it ?

hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java
toktarev: The same style was like like in NodeExtensionFactory:
    private static final String FACTORY_ID = "com.hazelcast.instance.NodeExtension";

hazelcast-code-generator/src/main/java/com/hazelcast/client/protocol/generator/CodeGeneratorMessageTaskFactoryProvider.java
```diff
@@ -82,7 +82,7 @@ public void init(ProcessingEnvironment env) {
 
     @Override
     public boolean process(Set<? extends TypeElement> elements, RoundEnvironment env) {
```
pveentjer: I'm not that familair with this part of the code generator. But if I would be a third party developer and designing e.g. a Tree and want to expose it through a client. How can this be done? Because I see hard coded support for Enterprise/Jet/OS... but how can I get my tree in?

toktarev: Now it is impossible for 3-rd party developer - integrated any client task factory.

To points:
        if (isEnterprise) {
            className = "EnterpriseMessageTaskFactoryProvider";
        } else if (isJet) {
            className = "JetMessageTaskFactoryProvider";
        } else {
            className = "DefaultMessageTaskFactoryProvider";
        }

2) public static final int JET_TEMPLATE_ID = 26;
Register of templateID is static

We need to discuss it with client team

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
toktarev: done

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptor.java
toktarev: done

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptorProvider.java
toktarev: Done

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxyDescriptorProvider.java
toktarev: Done

hazelcast/src/main/java/com/hazelcast/spi/impl/servicemanager/RemoteServiceDescriptorProvider.java
toktarev: Done

pveentjer:
Proper client integration requires some additional work.

So I'm fine getting this merged so that the JET team can continue with their work. And let the integration be a worry of the client team.

:+1:

pveentjer:
verify

jerrinot:
:+1: 
this affects client. @asimarslan or @sancar should have a look too before merging it. 
Also avoid merging it before required changes in EE are ready and reviewed.

pveentjer:
For EE PR see: https://github.com/hazelcast/hazelcast-enterprise/pull/737

toktarev:
I will finish with some remarks + write tests.

Thanks guys !!!

pveentjer:
Thanks for the huge amount of cleanup and especially the introduces flexibility by moving away from the NodeExtension. Even though the code changes might not be that big, it makes Hazelcast more flexible and less complicated.

sancar:
@toktarev 
I could not have a chance to review this pr. I have a question about following Client proxy registration
>> ClientProxyProvider has been implemented for custom client proxy loading
ClientProxyDescriptor
ClientProxyDescriptorProvider

Are these classes to register custom proxy and services for clients ? If yes, we already have this feature.
See ProxyFactoryTest
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/test/java/com/hazelcast/client/spi/ProxyFactoryTest.java

sancar:
Adding missing client label to this pull request!!!!!

toktarev:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/test/java/com/hazelcast/client/spi/ProxyFactoryTest.java

This is registration via config.
What if I need use service with default config ?

PR #7420, opened 2016-01-28
JMX Enhancements

muratayan:
Gets rid of the JMX stat collecting overhead. Stats are created for given distributed object type for once in given interval. Before it was every time there was a method call.

Logic reviewed and approved here: https://github.com/hazelcast/hazelcast/pull/7327

Please comment on compatibility.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
@@ -291,6 +291,7 @@ public enum GroupProperty implements HazelcastProperty {
 
     ENABLE_JMX("hazelcast.jmx", false),
     ENABLE_JMX_DETAILED("hazelcast.jmx.detailed", false),
+    JMX_UPDATE_INTERVAL_SECONDS("hazelcast.jmx.update.interval.seconds", 5, SECONDS),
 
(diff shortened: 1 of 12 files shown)


hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -168,6 +168,8 @@
     @Deprecated
     public static final String PROP_ENABLE_JMX_DETAILED = GroupProperty.ENABLE_JMX_DETAILED.getName();
     @Deprecated
+    public static final String PROP_JMX_UPDATE_INTERVAL_SECONDS = GroupProperty.JMX_UPDATE_INTERVAL_SECONDS.getName();
+    @Deprecated
```
emrahkocaman: Adding an already deprecated field seems odd to me. Isn't adding it to only `GroupProperty` enough?

emrahkocaman:
:+1: with a minor comment

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
Donnerbart: Same as in the other PR. `GroupProperty` should be enough, since `GroupProperties` was never intended to be a public API (but we don't know for sure, so I deprecated everything in 3.6).

PR #7421, opened 2016-01-28
[Forward-port] JMX Enhancements

muratayan:
Forward port of https://github.com/hazelcast/hazelcast/pull/7420

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
@@ -353,6 +353,7 @@ public enum GroupProperty implements HazelcastProperty {
 
     ENABLE_JMX("hazelcast.jmx", false),
     ENABLE_JMX_DETAILED("hazelcast.jmx.detailed", false),
+    JMX_UPDATE_INTERVAL_SECONDS("hazelcast.jmx.update.interval.seconds", 5, SECONDS),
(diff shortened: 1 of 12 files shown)


hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
```diff
@@ -168,6 +168,8 @@
     @Deprecated
     public static final String PROP_ENABLE_JMX_DETAILED = GroupProperty.ENABLE_JMX_DETAILED.getName();
     @Deprecated
+    public static final String PROP_JMX_UPDATE_INTERVAL_SECONDS = GroupProperty.JMX_UPDATE_INTERVAL_SECONDS.getName();
```
emrahkocaman: I think there is no need to add an already deprecated field.

emrahkocaman:
:+1: with only a minor comment

hazelcast/src/main/java/com/hazelcast/instance/GroupProperties.java
Donnerbart: Yes, please don't add anything to `GroupProperties` anymore. This is the old way and a grey-zone between public and private API. Just put stuff in `GroupProperty`, which is new since 3.6 and clearly marked as `@PrivateAPI`, so we can change things as we need to in the future.

PR #7426, opened 2016-01-29
Litter reduction in MethodProbe

pveentjer:
After doing some allocation analysis in JProfiler, it became clear that litter was being created
when calling Method.invoke(source) without an arguments array. The litter created is an empty var-args array.

After adding an explicit empty varargs array, no more litter was visible.

Very strange that the JIT doesn't optimize for this. Another option is that either JProfiler is lying or the JVM behaves differently when a profiler is connected.

Anyhow.. it should not do any harm.


mtopolnik:
:+1:

jerrinot:
:+1:

PR #7432, opened 2016-01-30
Removal of SerializerAdapter

pveentjer:
The SerializerAdapter has been removed and replaced by the usage of StreamSerializer.
The ByteArraySerializer is wrapped inside a StreamSerializer using the adapter design pattern.

The reason for this is simplification; the SerializerAdapters adds a layer of indirection
and causes complexity without added value.

Also renamed SingletonSerializer to AbstractStreamSerializer since the name is confusing.


pveentjer:
@sancar @asimarslan can you have a look at this PR. I have some other PR's that depend on this one.

sancar:
What are those ?
hazelcast/.attach_pid25507
hazelcast/.attach_pid25768

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java
```diff
@@ -323,44 +325,44 @@ protected final int calculatePartitionHash(Object obj, PartitioningStrategy stra
...
 
-    protected final boolean safeRegister(final Class type, final SerializerAdapter serializer) {
+    protected final boolean safeRegister(final Class type, final StreamSerializer serializer) {
         if (constantTypesMap.containsKey(type)) {
             throw new IllegalArgumentException("[" + type + "] serializer cannot be overridden!");
         }
-        SerializerAdapter current = typeMap.putIfAbsent(type, serializer);
-        if (current != null && current.getImpl().getClass() != serializer.getImpl().getClass()) {
+        StreamSerializer current = typeMap.putIfAbsent(type, serializer);
+        if (current != null && current.getClass() != serializer.getClass()) {
```
sancar: There seems to be a behaviour change. If user configures two different ByteArraySerializer for same type , it will not get an exception. Because both classes will be `ByteArraySerializerStreamSerializer` . I reproduced this in `AbstractSerializationServiceTest` as unittest. But when I try to do a integration test, I could not because there seems to be another check for this in `SerializerHookLoader.register`

```
   @Test(expected = IllegalStateException.class)
    public void testGlobalRegister_doubleRegistration_byteArraySerializer() throws Exception {
        abstractSerializationService.safeRegister(SampleObjects.Employee.class, new FooByteArraySerializer());
        abstractSerializationService.safeRegister(SampleObjects.Employee.class, new BarByteArraySerializer());
    }
```

pveentjer:
The pid files I'll remove. They are from Java Flight Recorder. I'll have a look at your other comment.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java
pveentjer: So if I understand you correctly; this test used to fail but withy my change it doesn't fail anymore. Right?

sancar: Yes. By the way, there is no such test in system. I just wrote a new test and shared with you to give an idea. You may want to add this test in your pr.

pveentjer: Yes. I have added the test; I got sidetracked with some JET stuff. But i'm back to this.

pveentjer:
I have changed the logic so it doesn't work on .class comparison, but it relies on the equals of the serializer. The ByteArraySerializerStreamSerializerAdapter has equals (hash) implemented so that it will only return equals when the wrapped ByteArraySerializer are the same instance.

pveentjer:
For the time being I'm going to close this one. I have a different approach in mind that can get rid of a bunch of instanceof checks and virtual method calls.

PR #7434, opened 2016-02-01
Papercut: Introduce static final field NEW_LINE and use it instead of getLineSeperator()

jerrinot:
The field is slightly less awkward to use, it's a standard practice in other projects
and it could have performance benefits as well.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
@@ -30,7 +30,6 @@ import java.util.logging.Level;
 
 import static com.hazelcast.internal.monitors.HealthMonitorLevel.OFF;
 import static com.hazelcast.internal.monitors.HealthMonitorLevel.valueOf;
-import static com.hazelcast.util.StringUtil.getLineSeperator;
 import static java.lang.String.format;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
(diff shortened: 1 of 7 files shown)


hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
```diff
@@ -154,7 +154,7 @@ private void logPerformanceMonitorHint() {
 
             logger.info(String.format("The HealthMonitor has detected a high load on the system. For more detailed information,%s"
                             + "enable the PerformanceMonitor by adding the property -D%s=true",
-                    getLineSeperator(), GroupProperty.PERFORMANCE_MONITOR_ENABLED));
+                    NEW_LINE, GroupProperty.PERFORMANCE_MONITOR_ENABLED));
```
Donnerbart: In this case you can just use `%n` in the text, if I'm not wrong.

jerrinot: correct

Donnerbart:
You could use the constant in the deprecated method itself as well to save a bit of performance if it's still used somewhere. But that's just nitpicking :)

:+1: (just minor comments)

jerrinot:
@Donnerbart: thanks, the PR is updated.

jerrinot:
run-lab-run

jerrinot:
run-lab-run

pveentjer:
Please check StringUtils. LINE_SEPARATOR

jerrinot:
I'm closing this one as @pveentjer introduced the `LINE_SEPARATOR` in a different PR.

PR #7440, opened 2016-02-01
Moved jmx to com.hazelcast.internal

pveentjer:
Since these classes are not used by end users, they can be moved to the com.hazelcast.internal package so they are hidden from the end user.


jerrinot:
:+1:

Donnerbart:
Just a small comment for future clean-ups like this: Right now a lot of imports are out of order. To prevent this you can use the `Reformat Code` on the top level folder and limit the scope to `Changed Files` (so no other files will be touched), just before you commit.

reformat-code-with-scope

This should not hurt, since the code formatter itself of IDEA 15/16 is good with its auto-detection of indentation etc. So it should not change too much else.

The rest of the commit looks good :+1:

pveentjer:
I have explicitely disabled this feature because you sometimes get changes for no apparent reason.

Donnerbart:
I'm not talking about an `auto-format-before-commit`. I'm talking about running the `Reformat Code` manually after these kind of changes (which need an optimize import to be nice and clean). The scope to `Changed Files` is the trick to not mess up the whole project.

gurbuzali:
:+1:

PR #7449, opened 2016-02-02
Moves com.hazelcast.cluster to internal

pveentjer:
Everything from the com.hazelcast.cluster package which isn't public API,
has been moved to the com.hazelcast.internal.cluster.

There are no other changes than import/checkstyle and javadoc.


pveentjer:
verify

tombujok:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -16,6 +16,7 @@
 
 package com.hazelcast.spi;
 
+import com.hazelcast.internal.cluster.ClusterClock;
```
Donnerbart: Just a note (and little reminder to myself): This will lead to an unused import issue in SonarQube, since `ClusterClock` is not used in the class itself, just JavaDoc. This is why we have normally have the FQN in JavaDoc. I don't know by heart from which tool this complain comes and why we have it just in SonarQube, so either we play by that rule or change it.

Donnerbart:
:+1:

jerrinot:
suspicious errors:
```
com.hazelcast.nio.serialization.HazelcastSerializationException: java.lang.ClassNotFoundException: com.hazelcast.examples.SerializerFactory
```

@hasancelik: is it a known issue in the fast pr builder?

mdogan:
@jerrinot: it's because of a new addition (most probably a leftover): https://github.com/hazelcast/hazelcast/blob/eef987b9920f3f7b4c506e433b0d067cab9ba93c/hazelcast/src/main/resources/hazelcast-default.xml#L253

jerrinot:
@mdogan: well spotted!

jerrinot:
:+1:

PR #7458, opened 2016-02-04
Moved connection to com.hazelcast.internal.connection

pveentjer:
The connection functionality is moved to com.hazelcast.internal.connection since it is
private API. If it is placed in com.hazelcast.nio, the private API is being placed in
the same directory as public API and this is very confusing.


pveentjer:
It causes an incompatibility on Operation since it has the method getConnection.

mdogan:
It's not possible to see actual changes. 
Can you paste the contents of `com.hazelcast.internal.connection` and `com.hazelcast.nio` packages here?

PR #7459, opened 2016-02-04
memberselectors is moved back to com.hazelcast.cluster

pveentjer:
It is moved back due to being a public API.


jerrinot:
:+1:

Donnerbart:
:+1: (once green)

PR #7471, opened 2016-02-05
 Improves size estimator heap cost calculation logic by introducing co…

ahmetmircik:
…mpressed-oops detection

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -462,6 +462,7 @@
     <suppress checks="" files="com.hazelcast.util.Base64"/>
 
     <suppress checks="" files="com.hazelcast.util.DebugUtil"/>
+    <suppress checks="" files="com.hazelcast.util.JVMUtil"/>
 
     <suppress checks="IllegalImport" files="com.hazelcast.util.counters.SwCounter"/>
     <suppress checks="InnerAssignment" files="com.hazelcast.util.counters.SwCounter"/>
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/NearCacheSizeEstimator.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/NearCacheSizeEstimator.java
@@ -18,11 +18,12 @@ package com.hazelcast.map.impl;
 
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
+import static com.hazelcast.util.JVMUtil.OBJECT_HEADER_IN_BYTES;
+
 /**
  * Size estimator for near cache.
  */
-class NearCacheSizeEstimator
-        implements SizeEstimator<NearCacheRecord> {
+class NearCacheSizeEstimator implements SizeEstimator<NearCacheRecord> {
 
     private static final AtomicLongFieldUpdater<NearCacheSizeEstimator> SIZE_UPDATER = AtomicLongFieldUpdater
             .newUpdater(NearCacheSizeEstimator.class, "size");
@@ -45,12 +46,9 @@ class NearCacheSizeEstimator
         if (cost == 0) {
             return 0;
         }
-        final int numberOfIntegers = 4;
-        long size = 0;
         // entry size in CHM
-        size += numberOfIntegers * ((Integer.SIZE / Byte.SIZE));
-        size += cost;
-        return size;
+        final int numberOfIntegers = 4;
+        return numberOfIntegers * (Integer.SIZE / Byte.SIZE) + OBJECT_HEADER_IN_BYTES + cost;
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
@@ -21,6 +21,9 @@ import com.hazelcast.util.HashUtil;
 
 import java.util.Arrays;
 
+import static com.hazelcast.util.JVMUtil.REFERENCE_COST_IN_BYTES;
+
+
 @edu.umd.cs.findbugs.annotations.SuppressWarnings("EI_EXPOSE_REP")
 public final class DefaultData implements Data {
 
@@ -84,9 +87,7 @@ public final class DefaultData implements Data {
 
     @Override
     public int getHeapCost() {
-        // reference (assuming compressed oops)
-        int objectRef = Bits.INT_SIZE_IN_BYTES;
-        return objectRef + (data != null ? ARRAY_HEADER_SIZE_IN_BYTES + data.length : 0);
+        return REFERENCE_COST_IN_BYTES + (data != null ? ARRAY_HEADER_SIZE_IN_BYTES + data.length : 0);
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/util/JVMUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/JVMUtil.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.util;
+
+import javax.management.MBeanServer;
+import javax.management.ObjectName;
+import javax.management.openmbean.CompositeDataSupport;
+import java.lang.management.ManagementFactory;
+
+import static com.hazelcast.logging.Logger.getLogger;
+import static com.hazelcast.nio.UnsafeHelper.UNSAFE;
+import static com.hazelcast.nio.UnsafeHelper.UNSAFE_AVAILABLE;
+import static java.lang.System.getProperty;
+
+/**
+ * Helper class for retrieving JVM specific information.
+ */
+public final class JVMUtil {
+
+    private static final boolean DEFAULT_USE_COMPRESSED_OOPS = true;
+    private static final boolean USE_COMPRESSED_OOPS = useCompressedOops();
+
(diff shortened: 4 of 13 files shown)


hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
```diff
@@ -21,6 +21,8 @@
 
 import java.util.Arrays;
 
+import static com.hazelcast.map.impl.SizeEstimator.REFERENCE_COST_IN_BYTES;
```
mdogan: `.map.impl.` import in serialization?

hazelcast/src/test/java/com/hazelcast/map/SizeEstimatorTest.java
```diff
@@ -47,7 +47,7 @@ public void smoke() throws InterruptedException {
 
     @Test
     public void testSinglePut() throws InterruptedException {
-        final long expectedPerEntryHeapCost = 166L;
+        final long expectedPerEntryHeapCost = 218L;
```
mdogan: Does this number assume compressed oops?

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
```diff
@@ -16,18 +16,101 @@
...
  *
  * @param <T> the type of object which's size going to be estimated.
  */
-public interface SizeEstimator<T> {
+public abstract class SizeEstimator<T> {
+
+    public static final boolean HAS_COMPRESSED_OOPS = hasCompressedOops();
+
+    @SuppressWarnings("CheckStyle")
+    public static final int OBJECT_HEADER_IN_BYTES = HAS_COMPRESSED_OOPS ? 12 : 16;
```
serkan-ozal: if we handle 32 bit systems also, this is `8` on these systems. `4 byte mark oops` + `4 byte class pointer`

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
```diff
@@ -16,18 +16,101 @@
...
  * Size estimator general contract.
  *
  * @param <T> the type of object which's size going to be estimated.
  */
-public interface SizeEstimator<T> {
+public abstract class SizeEstimator<T> {
+
+    public static final boolean HAS_COMPRESSED_OOPS = hasCompressedOops();
+
+    @SuppressWarnings("CheckStyle")
```
Donnerbart: You can replace the entry in the `suppressions.xml` with `@SuppressWarnings("checkstyle:magicnumber")` here. That should work (must be all lowercase).

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
```diff
@@ -16,18 +16,101 @@
...
+    public static final int OBJECT_HEADER_IN_BYTES = HAS_COMPRESSED_OOPS ? 12 : 16;
+
+    @SuppressWarnings("CheckStyle")
```
Donnerbart: Same here: `@SuppressWarnings("checkstyle:magicnumber")`

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
```diff
@@ -16,18 +16,101 @@
...
+import javax.management.openmbean.CompositeDataSupport;
+import java.lang.management.ManagementFactory;
+
 /**
  * Size estimator general contract.
  *
  * @param <T> the type of object which's size going to be estimated.
  */
-public interface SizeEstimator<T> {
+public abstract class SizeEstimator<T> {
```
Donnerbart: You can try this here, but I don't know if this is the correct level to put this annotation: `@SuppressWarnings("checkstyle:illegalimport")`

checkstyle/suppressions.xml
```diff
@@ -574,6 +574,7 @@
     <suppress checks="MethodCount|ClassDataAbstractionCoupling" files="com.hazelcast.map.impl.MapServiceContextImpl"/>
     <suppress checks="ClassFanOutComplexityCheck" files="com.hazelcast.map.impl.MapServiceContextImpl"/>
     <suppress checks="MethodCount" files="com.hazelcast.map.impl.MapServiceContext"/>
+    <suppress checks="IllegalImport|MagicNumber" files="com.hazelcast.map.impl.SizeEstimator"/>
```
Donnerbart: Can be removed if annotation based suppressions work.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
mdogan: @serkan-ozal; 
Just out of curiosity, are these Hotspot specific details or from jvm spec?

serkan-ozal: not from spec. in fact more reliable way might be getting field offset of a sample field on a sample class. Such as

``` java
class Foo {
    int i;
}
```

in here, field offset of `i` should be the header size of object (I assume jvm impl doesn't have stupid layout mechanism).

Similar approach is generally used for getting shallow size of objects by getting max field offset and aligning it to heap word size (which is `8` by default)

serkan-ozal: In fact, there is no public way to detect compressed-oops even though via `Unsafe`.
The best way is through Hotspot SA agent but it is hard to use in embedded mode.
Another hacky way is Andrei Pangin's `Helfy`: https://github.com/apangin/helfy/
As far as I have experienced, it can be used to access JVM internal symbols so you can get some JVM details without Hotspot SA agent

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
```diff
@@ -16,18 +16,101 @@
...
+     *
+     * Borrowed from http://openjdk.java.net/projects/code-tools/jol/
+     */
+    private static boolean isCompressedOopsEnabled() {
+        // When running with CompressedOops on 64-bit platform, the address size
+        // reported by Unsafe is still 8, while the real reference fields are 4 bytes long.
+        // Try to guess the reference field size with this naive trick.
+        int oopSize;
+        try {
+            long off1 = UNSAFE.objectFieldOffset(TestCompressedOops.class.getField("obj1"));
```
mdogan: You need to check if `Unsafe` is available and NOT fail when not available.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
serkan-ozal: This is the Jemstone way to access these internals (compressed-oops) via HotSpot SA agent: https://github.com/serkan-ozal/jemstone/blob/master/src/main/java/tr/com/serkanozal/jemstone/sa/impl/compressedrefs/HotSpotSACompressedReferencesWorker.java

This is the Helfy way: https://github.com/apangin/helfy/blob/master/test/one/helfy/ObjectAddress.java

I have not tried Helfy on another JVMs rather than Hotspot, but Jemstone is only for HotSpot

mdogan: @ahmetmircik: is it really required to calculate that much precise?

hazelcast/src/test/java/com/hazelcast/map/SizeEstimatorTest.java
ahmetmircik: Yes

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
ahmetmircik: put all to xml.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
ahmetmircik: done.

hazelcast/src/main/java/com/hazelcast/map/impl/SizeEstimator.java
ahmetmircik: trying to make it to be as close as possible

hazelcast/src/main/java/com/hazelcast/nio/serialization/DefaultData.java
ahmetmircik: fixed

hazelcast/src/main/java/com/hazelcast/util/JVMUtil.java
```diff
@@ -0,0 +1,122 @@
...
+        // Fallback check for other JVM implementations.
+        enabled = isCompressedOopsOrNull();
+        if (enabled != null) {
+            return Boolean.valueOf(enabled);
+        }
+
+        return DEFAULT_COMPRESSED_OOPS;
+    }
+
+    private static String isHotSpotCompressedOopsEnabledOrNull() {
```
serkan-ozal: why are we dealing with `String`? this method (`isHotSpotCompressedOopsEnabledOrNull`) and the other `isCompressedOopsOrNull` could be return boolean.

serkan-ozal:
:+1:

hazelcast/src/main/java/com/hazelcast/util/JVMUtil.java
ahmetmircik: makes sense.

hasancelik:
verify

hasancelik:
verify

hasancelik:
verify

hasancelik:
This build has a green pr-builder after last commit.
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-pr-builder/17965/

hazelcast/src/main/java/com/hazelcast/map/impl/NearCacheSizeEstimator.java
```diff
@@ -45,12 +46,9 @@ public long getCost(NearCacheRecord record) {
...
             return 0;
         }
-        final int numberOfIntegers = 4;
-        long size = 0;
         // entry size in CHM
-        size += numberOfIntegers * ((Integer.SIZE / Byte.SIZE));
-        size += cost;
-        return size;
+        final int numberOfIntegers = 4;
+        return numberOfIntegers * (Integer.SIZE / Byte.SIZE) + OBJECT_HEADER_IN_BYTES + cost;
```
mdogan: minor comment: `Bits.INT_SIZE_IN_BYTES` is easier to read, instead of `(Integer.SIZE / Byte.SIZE)`.

mdogan:
:+1:

Issue #7443, opened 2016-02-01
All Topic messages are processed by a single thread


dsukhoroslov:
I have a single HZ 3.5.4 server. There is a Topic. Both publisher and subscriber are within the server and they're singletons. The rate of messages published to the topic is quite high, it is > 1000 per sec. But all of them are consumed by only one HZ event thread, as I see in my logs. My EventService pool size is default = 5. So it is not clear why only one thread handles all the load. Look like you assign thread per subscriber? Is it possible to improve the situation somehow?

Thanks, Denis.

jerrinot:
Hi @dsukhoroslov,

the same topic use the same thread. I agree this could be relaxed when ordering is not important. The workaround would be to either use multiple topics or offload message processing to another thread.

dsukhoroslov:
Hi @jerrinot, 

I like this new label :). Ok, I'll think about it. Besides ordering do you have any other concens using the full EventService thread pool for message processing?

jerrinot:
I am glad you like the label! ;-)

as far as I know ordering is the only reason for this. 

[This could be a good starting point](https://github.com/jerrinot/hazelcast/blob/46d5aa590f0598b7fa226f448b99bc8c2ca4f330/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java#L148-L148)  See the last argument - `name.hashCode())` 
Also please try to read the [documentation on this](http://docs.hazelcast.org/docs/3.6/manual/html-single/index.html#topic first). It's quite interesting. I didn't know about the `globalOrderEnabled` flag.

dsukhoroslov:
@jerrinot, I looked at the code around TopicService and EventService. The improvement looks pretty clear:

<code>
add Registration.isOrdered() property, set it at construntion from nodeEngine.getHzInstance().getConfig().getTopicConfig(topic).isGlobalOrderingEnabled();

make two new classes which are not StrippedRunable, move most of LocalEventDispatcher/EventPacketPubliser fields to them (except orderKey);
inherit LocalEventDispatcher/EventPacketPubliser from these two classes, leave StrippedRunable implementation only;

in eventService.executeLocal construct LocalEventDisparcher or its non-StrinnedRunable parent depending on the reg.isOrdered() flag;

do the same for remote events with EventPacketPublisher/it's parent class;
</code>

[...]

PR #7472, opened 2016-02-05
Multithreaded topics, issue #7443

dsukhoroslov:
Hi,

here is my imlementation for the issue mentioned. All tests were passed, new tests for the multithreaded functionality added. Some event handling logic was straithen, so no need in EventProcessor/RemoteEventProcessor classes any more. Schemas and configuration processing changed for Hazelcast and Spring configs.
In the Spring config schema I changed xs:sequence compartment type to xs:all compartment type. This is how it is specified in the Hazelcast schema and it is much better for future changes from compatibility point of view.

Please have a look and merge :)


dsukhoroslov:
Hi,

Can someone check/verify it please?

Thanks, Denis.

jerrinot:
Hi @dsukhoroslov,

many thanks for the contribution! I'll have a look today or tomorrow latest.

hazelcast/src/main/java/com/hazelcast/config/TopicConfig.java
```diff
@@ -114,10 +115,44 @@ public boolean isGlobalOrderingEnabled() {
...
+     * thread pool can be used for incoming message processing. Otherwise only one
+     * dedicated thread will be used to handle topic messages.    
+     * Note: it can be enabled only in case when global ordering is disabled.
+     *
+     * @param multiThreadingEnabled set to true to enable multi-threaded message processing, false to disable
+     * @return The updated TopicConfig
+     */
+    public TopicConfig setMultiThreadingEnabled(boolean multiThreadingEnabled) {
+    	if (this.globalOrderingEnabled) {
+    		// log warn/error ?
```
jerrinot: you can throw `ConfigurationException`. 
prefer to fail-fast rather than silently "fix" conflicting configuration options.

dsukhoroslov: Ok, will fix now.

dsukhoroslov: Done, please have a look

hazelcast/src/main/java/com/hazelcast/config/TopicConfig.java
```diff
@@ -188,12 +221,11 @@ public boolean equals(Object obj) {
             return false;
         }
         TopicConfig other = (TopicConfig) obj;
-        return
-                (this.name != null ? this.name.equals(other.name) : other.name == null)
-                        && this.globalOrderingEnabled == other.globalOrderingEnabled;
+        return (this.name != null ? this.name.equals(other.name) : other.name == null);
```
jerrinot: this is changing semantic of equality. is it intentional?

jerrinot:
Hi Dennis, I'm still reviewing it to make sure I understand its implications. 

Just a few style-related notes: 
- Please remove the code you commented out 
- You can squash the changes into a single commit. To have a nice history. 

@eminn will have a look tomorrow - he has pretty deep understanding of ITopic. If everything is good then we are good to merge.

hazelcast/src/main/java/com/hazelcast/config/TopicConfig.java
dsukhoroslov: Yes, this was agreed in the initial issue discussion #7443

jerrinot: I see. I must have missed this. thanks for clarification!

hazelcast/src/main/java/com/hazelcast/spi/impl/eventservice/impl/operations/SendEventOperation.java
```diff
@@ -41,7 +41,10 @@ public SendEventOperation(EventEnvelope eventEnvelope, int orderKey) {
     @Override
     public void run() throws Exception {
         EventServiceImpl eventService = (EventServiceImpl) getNodeEngine().getEventService();
-        eventService.executeEventCallback(new EventProcessor(eventService, eventEnvelope, orderKey));
+        Registration registration = eventService.getRegistration(eventEnvelope);
```
jerrinot: `eventService.getRegistration()` assumes registrations are always local. I don't fully understand the flow, but it's somewhat suspicious.

dsukhoroslov: It happens when remote event (initiated on other node) has arrived on destination node. Here it founds local EventService impl and reuse its existing (local) functionality.

jerrinot: I see, makes sense.

dsukhoroslov:
@jerrinot , I removed the obsolete comments. No need in them when diff can be used, sure.
Regarding the squash.. Not sure how to do it properly, acually. I tried it couple of time with no luck :(

mtopolnik:
I advise the following, assuming that `upstream` is the name of the "blessed" HZ remote repository:

```
$ git fetch upstream
$ git rebase upstream/master
... resolve any merge conflicts ...
$ git reset upstream/master
$ git commit -a -m "Improve topic processing"
```

hazelcast/src/main/java/com/hazelcast/config/TopicConfig.java
```diff
@@ -113,11 +115,43 @@ public boolean isGlobalOrderingEnabled() {
...
+        return multiThreadingEnabled;
+    }
+    
 
     /**
+     * Enable multi-threaded message handling. When enabled any thread from events 
+     * thread pool can be used for incoming message processing. Otherwise only one
+     * dedicated thread will be used to handle topic messages.    
+     * Note: it can be enabled only in case when global ordering is disabled.
+     *
```
eminn: You are only mentioning that the global order will be broken when the multi-threading enabled. We also have an implicit order which is same publisher's messages are consumed in order, which multi-threading will also break this one.

eminn:
Also these changes should be reflected and documented on the client side. You can have a look at `TopicAddMessageListenerMessageTask.java` line 108, where the topic event's sent to the client and executed over there on a striped executor.

hazelcast/src/main/java/com/hazelcast/config/TopicConfig.java
dsukhoroslov: Agree, explained this in javadoc.

dsukhoroslov:
Client message processing is quite different. I do not see any possibility to configure topic behaviour on the client side at all: there is no corresponding config element. So, I wouldn't change message handling model on the client side right away, it'll make the system non-backward-compatible and not usable for some projects which relies on the message ordering.

dsukhoroslov:
@jerrinot, I did squash, please have a look

eminn:
Then, with the multi-threading enabled, client consumers and member consumers will behave differently which is something we wouldn't want.

dsukhoroslov:
@eminn, what would you suggest then? 
Yes, client and member consumers can behave differently (when multithreading is used). They can behave differently even if topic configurtation will be added on the client side... I'd separate server and client message processing into different PRs.

sancar:
This can be solved in two ways as far as I see. 
1) It can be configured from node only and effect the client by refactoring the sending part. 
2) And we can refactor the client code and put the config on client side. 

First option seems easier to implement. The latter will be a little bit more complicated to implement.

We will discuss this internally with Client Team  and see which one works better. I will try to get back to you asap.

dsukhoroslov:
@sancar, thank you for the quick response. I'd like to add my view on this:
Client and server nodes usually have completelly different purpose and resource power. On the server side multithreading is usefull when there is a need to process a high volume of incoming messages regardless of their ordering. 
On the client side the scenario as above is not so often. Usually client subscribes to some kind of notifications (state changes) which should be procesed in the right order. 
So, I wouldn't couple these two kind of topic message processing at all. Also, there are cases when all processing is done on the server or client side only.

sancar:
General approach of hazelcast so far, we have the configuration of all distributed objects only from server. This makes things a lot easier to understand. If multithreaded is enabled on server side, as @eminn suggested here https://github.com/hazelcast/hazelcast/pull/7472#issuecomment-182759417
, we can send refactor the server to send to random partition id. 

It should be a couple of lines of work. 
We will not introduce any config on client side.

sancar:
>> I did squash, please have a look

And there seems to be problem with squash. I still see 8 commits on pr. And the last commit is saying merge with master and it has no changes ?!?

dsukhoroslov:
Yep, that is how eclipse plugin for git does work. It creates unnecesary commits sometime resolving diffs. All previous commits were aggregated into a new one.
Regarding the client message processing: could you advise these two lines? From the code pointed by @eminn I do not see what has to be done.

eminn:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/topic/TopicAddMessageListenerMessageTask.java#L108

This line should be replaced with something like following,

```
if multiThreadingEnabled:
	sendClientMessage(randomPartitionKey)
else:
	sendClientMessage(partitionKey)
```

dsukhoroslov:
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java#L209

is null key used as random?

hazelcast/src/main/java/com/hazelcast/spi/impl/eventservice/impl/EventServiceImpl.java
```diff
@@ -441,16 +492,10 @@ public void executeEventCallback(Runnable callback) {
...
-            eventExecutor.execute(new RemoteEventProcessor(this, packet));
-        } catch (RejectedExecutionException e) {
-            rejectedCount.inc();
-
-            if (eventExecutor.isLive()) {
-                Connection conn = packet.getConn();
-                String endpoint = conn.getEndPoint() != null ? conn.getEndPoint().toString() : conn.toString();
-                logFailure("EventQueue overloaded! Failed to process event packet sent from: %s", endpoint);
-            }
+        EventEnvelope eventEnvelope = (EventEnvelope) nodeEngine.toObject(packet);
```
ahmetmircik: Can you please point what is your reasoning behind this change (i mean new de-serialization step)? It seems to me a cause for performance degradation at first glance

dsukhoroslov: but the code in the RemoteEventProcessor was exactly the same:

`            EventEnvelope eventEnvelope = (EventEnvelope) eventService.nodeEngine.toObject(packet);
            process(eventEnvelope);
 `

which new deserialization step do you mean?

eminn:
@dsukhoroslov 

1- The flag will only be configured from server side, so no need to carry the flag to the client.
By setting random partition id, you are already defining the multi-threaded behavior for client side on the server side.

2- No, `null` key doesn't mean random. You can generate a random number between 0 and partition count.

hazelcast/src/main/java/com/hazelcast/spi/impl/eventservice/impl/EventServiceImpl.java
ahmetmircik: I was meaning, `RemoteEventProcessor` is used to offload that de-serialization work to another thread than IO thread, now, with your change, it will be an extra work for IO thread (beware that method `handle(Packet packet)` is called by IO thread). And given that IO thread is a shared resource and it mostly only dispatches packets to handlers, seems better to revert your change for that.

dsukhoroslov: Ah, I see what you mean now. Ok, I'll revert this change back.

dsukhoroslov:
@eminn 

So `TopicAddMessageListenerMessageTask` invoked on the servere side? Ok, now it make sense. Then we can set isMultithreaded (local) flag right in the `call()` method, right?

eminn:
Yes it is invoked on the server side. You can check-then-act based on the multithreaded flag retrieved from node configuration on the `onMessage` method.

hazelcast/src/main/java/com/hazelcast/spi/impl/eventservice/impl/EventServiceImpl.java
dsukhoroslov: @ahmetmircik,

there is a problem with this approach. I can understand - is message should be processed on dedicated thread or on any event service thread - after message de-serialization only:

`

        EventEnvelope eventEnvelope = (EventEnvelope) nodeEngine.toObject(packet);
        Registration registration = getRegistration(eventEnvelope);
        if (registration.isMultithreaded()) {
            // process it via regular Runable task
        } else {
            // process it via StripedRunable task
        }
`

what would you suggest?

ahmetmircik: Why don't we just give a random number [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java#L148) instead of `name.hashCode` when multi-threaded mode is on? 

By doing that can't we provide the desired fix for your issue?

dsukhoroslov: Hi @ahmetmircik,

Yes, it can be solved this way. Then most of the changes I did in this PR are not needed. How can I get random key? I wouldn't duplicate StripedExecutor functionality. What about hashCode from current timestamp?

ahmetmircik: Hi, 
`System.currentTimeMillis()` can give same value in frequent calls under high load, maybe you can use `TopicEvent.hashCode` or `System.nanoTime` instead.

dsukhoroslov: Well, under high load `TopicEvent.hashCode` will also be the same, as it is produced via System.currentTimeMillis, afaik. And `System.nanoTime` produce collisions too. What about a simple AtomicLong counter?

ahmetmircik: I am also fine with atomiclong

jerrinot: not that it's terribly important, but no HotSpot version I know of uses `System.currentTimeMillis()` to generate identity hashcode. 

See [this](http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/981cb5ddafc8/src/share/vm/runtime/synchronizer.cpp#l557) for available strategies. The last strategy is used by default (in recent versions of HotSpot). The strategy no. 2 could be a lot of fun. `-XX:hashCode=2` :)
`</smartass-mode>`

dsukhoroslov: yes, just did a quick test too. No collisions on TopicEvent, I agree. But I seen the confusing output many times when objects created at the same time (with default toString() impl) has the same output.
The output from my test is:

`
time taken for TopicEvent: 405; collisions: 0
time taken for AtomicLong: 65; collisions: 0
time taken for currentTimeMillis: 52; collisions: 999948
time taken for nanoTime: 126; collisions: 752389
`

the number if iterations is 1 million. So, I'd use AtomicLong as it is a bit faster. Do you agree?

ahmetmircik: i am fine with `AtomicLong`.

dsukhoroslov:
The most changes I did in EventServiceIMpl and around it are not relevant any more. Then I'll close this PR and will create a new one with less changes.

PR #7476, opened 2016-02-07
Renamed to InternalClusterService

pveentjer:
The com.hazelcast.internal.cluster.ClusterService has been renamed to InternalClusterService.

The reason behind this rename is that we already have a public com.hazelcast.core.ClusterService
and having a second class with the same name is confusing (especially with code completion).

Also internally it is confusing to have 2 interfaces with the same name, but different purposes.

Apart from that there is a naming schema that for internal API's, especially in case of name
conflicts, we use the Internal prefix. For example:
* InternalOperationService
* InternalPartitionService
* InternalEventService

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/NodeEngine.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/NodeEngine.java
@@ -16,7 +16,7 @@
 
 package com.hazelcast.spi;
 
-import com.hazelcast.internal.cluster.ClusterService;
+import com.hazelcast.internal.cluster.InternalClusterService;
 import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.Member;
@@ -57,7 +57,7 @@ public interface NodeEngine {
      *
      * @return the ClusterService.
      */
-    ClusterService getClusterService();
+    InternalClusterService getClusterService();
 
     /**
      * Gets the InternalPartitionService.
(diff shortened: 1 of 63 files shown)


hazelcast/src/main/java/com/hazelcast/spi/NodeEngine.java
```diff
@@ -57,7 +57,7 @@
      *
      * @return the ClusterService.
      */
-    ClusterService getClusterService();
+    InternalClusterService getClusterService();
```
Donnerbart: This is a _public_ SPI method, isn't it? Can this break existing code?

mdogan:
>The reason behind this rename is that we already have a public com.hazelcast.core.ClusterService
and having a second class with the same name is confusing (especially with code completion).
Also internally it is confusing to have 2 interfaces with the same name, but different purposes.


There's no class named `com.hazelcast.core.ClusterService`. Means we don't have 2 interfaces with the same name.  It seems main reasoning is broken. 

Also to me, it's redundant to have *internal* both in package and class names.

pveentjer:
I got confused with Cluster vs ClusterService.

Normally it is redundant; only in case of conflicting names like PartitionService/InternalPartitionService, OperationService/InternalOperationService.

hazelcast/src/main/java/com/hazelcast/spi/NodeEngine.java
pveentjer: Indeed. Another example of broken API where internal data-structures like into the outside like InternalPartitionService.

PR #7483, opened 2016-02-08
Prevent blocking for Transactions (Map/MultiMap)

gurbuzali:
Previously read operations were blocked if an entry is locked transactionaly. 
I've introduced a `blockReads` field on Lock, it is set to true on `prepare` phase of the transaction and cleared on `commit/rollback` phase of the transaction (while unlocking).
Now read operations are checking `blockReads` field to block/wait instead of `transactional` field.

One thing to remember, since `ONE_PHASE` transactions do not have a `prepare` phase, the old behavior is not changed.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockResourceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockResourceImpl.java
@@ -42,6 +42,7 @@ final class LockResourceImpl implements DataSerializable, LockResource {
     private long expirationTime = -1;
     private long acquireTime = -1L;
     private boolean transactional;
+    private boolean blockReads;
     private Map<String, ConditionInfo> conditions;
     private List<ConditionKey> signalKeys;
     private List<AwaitOperation> expiredAwaitOps;
@@ -74,7 +75,7 @@ final class LockResourceImpl implements DataSerializable, LockResource {
         return (this.threadId == threadId && owner != null && owner.equals(this.owner));
     }
 
-    boolean lock(String owner, long threadId, long referenceId, long leaseTime, boolean transactional) {
+    boolean lock(String owner, long threadId, long referenceId, long leaseTime, boolean transactional, boolean blockReads) {
         if (lockCount == 0) {
             this.owner = owner;
             this.threadId = threadId;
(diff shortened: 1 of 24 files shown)


hazelcast/src/test/java/com/hazelcast/concurrent/lock/LockStoreImplTest.java
```diff
@@ -398,21 +398,21 @@ public void testTxnLock_whenLockedBySameCallerAndDifferentThreadId_thenReturnFal
 
     @Test
     public void testIsTransactionallyLocked_whenLockDoesNotExist_thenReturnFalse() {
-        boolean locked = lockStore.isTransactionallyLocked(key);
+        boolean locked = lockStore.shouldBlockReads(key);
```
eminn: I think we should also rename the names of the test methods

hazelcast/src/test/java/com/hazelcast/map/MapTransactionTest.java
```diff
@@ -746,59 +806,43 @@ public Boolean execute(TransactionalTaskContext context) throws TransactionExcep
...
-        }
-
-        @Override
-        public MapOperation createContainsKeyOperation(String name, Data dataKey) {
-            MapOperation containsKeyOperation = operationProvider.createContainsKeyOperation(name, dataKey);
-            containsKeyOperation.setWaitTimeout(TimeUnit.SECONDS.toMillis(3));
-            return containsKeyOperation;
-        }
-    }
+//    @Test(expected = OperationTimeoutException.class)
```
eminn: Either we should remove this test or make the test uncommented.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockResourceImpl.java
```diff
@@ -101,7 +104,7 @@ boolean extendLeaseTime(String caller, long threadId, long leaseTime) {
         if (!isLockedBy(caller, threadId)) {
             return false;
         }
-
+        this.blockReads = true;
```
eminn: I think we can make this parameterized, because we are assuming that this method is used only by two phase transactions.

eminn:
Besides minor comments, this looks good to be merged :+1:

gurbuzali:
I've addressed your review comments

metanet:
:+1:

gurbuzali:
unrelated test failure
`com.hazelcast.internal.cluster.impl.ClusterStateManagerTest.test_lockClusterState_expiry`
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/2146/

gurbuzali:
run-lab-run

PR #7486, opened 2016-02-08
Moved collection.common to impl

pveentjer:
Moved the collection.common to collection.impl.common since it isn't public API.


Donnerbart:
:+1:

mtopolnik:
:+1:

PR #7487, opened 2016-02-08
Moved com.hazelcast.instance to internal

pveentjer:
Since all these classes are private and not public, it is best to move them to
the com.hazelcast.internal package.


Donnerbart:
Hmm, reviewing this doesn't make sense to a certain point. I think we can trust IDEA that it did its job. So let's have a look at the classes which have been moved and check them for possible public usage.

I'm not sure about `BuildInfo`. At least in Simulator we use it to check for the Hazelcast version, e.g. to decide if we can use the async map streamer. I tried to find other projects via the GitHub search which may use the class, but after a quick check I found just internal usage. At least I can tell for sure, that we have to adapt EE and EE sessions repo, please prepare a PR before this gets merged!

`GroupProperties` etc. should be fine. The constants which may be used in public have been deprecated in 3.6, so it should be fine to move the class in 3.7. I will do a cleanup PR to remove the old stuff after this has been merged.

I think the hardest candidate to judge will be `Node`. Are we sure that we didn't use it in production to get internal information from Hazelcast? Maybe our SA can tell us more.

At least Simulator will break in several places with this change. We have a `HazelcastTestUtils` which uses `BuildInfo`, `BuildInfoProvider`, `HazelcastInstanceImpl`, `HazelcastInstanceProxy` and `Node`. Also your `MockIOService` and the `NetworkTest` will break due to usage of `HazelcastThreadGroup` and `Node`. If we adapt those classes to the new packages, we can't use Hazelcast 3.6 during runtime anymore. Tricky...

noctarius:
I agree with @Donnerbart, BuildInfo is used by external tools / applications to figure out the version of Hazelcast running (incl Enterprise yes / no). Those classes should not be moved, at least not without legacy path (new classes in old place, extending the moved classes and deprecating them).

pveentjer:
@noctarius we'll change HZ Enterprise to point to the new classes. 

You bring up a good point. The primary reason to do these migrations is that it is often unclear of something is public API or not. And it seems we have confusion here; so it is important to get resolved.

@Donnerbart I know this is going to break the simulator, but that is a problem we need to resolve in the simulator. Probably we'll need to apply some reflection magic to make it running in a version independent way. It is not a problem that can't be resolved.

noctarius:
@pveentjer I'm not talking about Hazelcast Enterprise but about external / third party tools using this class to determine running on HZ / HZE. We have partners that embed HZ into their own systems and activate / deactivate features depending on what their users deploy.

pveentjer:
Ok. I moved the BuildInfo/BuildInfoProvider back to the com.hazelcast.instance package.

Donnerbart:
Just see Simulator as an example for a "3rd party" tool. If we need to use those classes, there will be other projects as well. Maybe we should also check the other classes I mentioned above for possible external usage.

For example we have this method, which looks like it could be a common use case:
```
public static boolean isMemberNode(HazelcastInstance instance) {
    return (instance instanceof HazelcastInstanceProxy);
}
```

And then we have the famous `getNode()`, which is actually just used in your `NetworkTest` at the moment (was also used in the `SyntheticTest`):
```
public static Node getNode(HazelcastInstance hz) {
    HazelcastInstanceImpl impl = getHazelcastInstanceImpl(hz);
    return impl != null ? impl.node : null;
}

private static HazelcastInstanceImpl getHazelcastInstanceImpl(HazelcastInstance hz) {
    HazelcastInstanceImpl impl = null;
    if (hz instanceof HazelcastInstanceProxy) {
        return getFieldValue(hz, "original");
    } else if (hz instanceof HazelcastInstanceImpl) {
        impl = (HazelcastInstanceImpl) hz;
    }
    return impl;
}
```
And of course `getNode()` used internally again to retrieve the `OperationService`, which is used in the synthetic test package, to retrieve the operation count (used in 16 tests) and in two other tests...
```
public static OperationService getOperationService(HazelcastInstance hz) {
    Node node = getNode(hz);
    if (node == null) {
        throw new NullPointerException("node is null in Hazelcast instance " + hz);
    }
    NodeEngineImpl nodeEngine = node.getNodeEngine();
    try {
        return nodeEngine.getOperationService();
    } catch (NoSuchMethodError e) {
        // fallback for a binary incompatible change (see commit http://git.io/vtfKU)
        return getOperationServiceViaReflection(nodeEngine);
    }
}

private static OperationService getOperationServiceViaReflection(NodeEngineImpl nodeEngine) {
    try {
        Method method = NodeEngineImpl.class.getMethod("getOperationService");
        return (OperationService) method.invoke(nodeEngine);
    } catch (NoSuchMethodException e) {
        throw new IllegalStateException(e);
    } catch (InvocationTargetException e) {
        throw new IllegalStateException(e);
    } catch (IllegalAccessException e) {
        throw new IllegalStateException(e);
    }
}
```
So this PR will break Simulator in quite a lot of places. But the important question is if other projects use the same classes as we do. I guess the idea of Chris with the deprecated wrapper classes is a good one. Just a simple move may cause too much collateral damage.

pveentjer:
The problem here is that these API's were not set up with a separation of public and private api in mind. Node etc are internal classes. It is impossible to provide backwards guarantees on it and even highly undesirable.

PR #7489, opened 2016-02-08
Fixes MapLoader#load call to one per key in EntryProcessor related op…

ahmetmircik:
…erations


gurbuzali:
:+1:

ahmetmircik:
simple one merging it.

PR #7490, opened 2016-02-09
Using StringUtil.LINE_SEPARATOR

pveentjer:
Using a constant instead of static method which does a System property lookup.


jerrinot:
:+1:

Donnerbart:
:+1: (once green)

PR #7491, opened 2016-02-09
Moved util.counters to internal.util.counters

pveentjer:
The com.hazelcast.util package is internal API so should be moved to com.hazelcast.internal.util. This is one of many smaller pr's.


Donnerbart:
Those classes look safe to be moved. Also the `util` package is documented to be private (although there is always a risk of people ignoring that). Nevertheless, in this case I don't expect much pain.

:+1:

jerrinot:
:+1:

PR #7493, opened 2016-02-09
Moved Preconditions to com.hazelcast.internal.util

pveentjer:


Donnerbart:
I could not find any 3rd party usage of these classes. But EE and Cascading (Incubator project) will be affected. If there are no objections from our solution architects, this change should be fine.

:+1:

BTW: PR needs a rebase.

PR #7496, opened 2016-02-09
Fixed test failure on MetricsPluginTest

pveentjer:
the problem is that rending the probes is eventually consistent due to the internals
of MetricsRegistry being eventually consistent. So eventually the right probes are
seen; but doesn't need to be immediately.

This cases the problematic probe not to be seen.


jerrinot:
:+1:

Donnerbart:
:+1:

PR #7500, opened 2016-02-09
Cleanup of CacheProxyUtil

Donnerbart:
Little cleanup of `CacheProxyUtil`, which used unnecessary `map.keySet().contains()` and `map.values().contains()` calls.


serkan-ozal:
:+1:

Issue #7484, opened 2016-02-08
TransactionalMapProxy should be made generic


LoneRifle:
Given that ClientTxnMapProxy has generic types, it makes sense for TransactionalMapProxy to do so to. This will allow it to better support new method calls where the return type is a collection of keys and values, such as the getAll method proposed in #4698

Chelsea31:
I would like to work on this.

Chelsea31:
I have created a [PR](https://github.com/hazelcast/hazelcast/pull/19750) for this, please review.

PR #7502, opened 2016-02-11
Make TranscationalMapProxy generic

LoneRifle:
* Introduce generic parameters to TxnValueWrapper and CachedQueryEntry.
* Introduce a generic cast for toObjectIfNeeded. Given most callers of
this method will be casting to desired type anyway, it makes sense to do
so.
* Convert all Object types in TransactionalMapProxy to K/V generics
where appropriate

implements #7484

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
@@ -52,9 +52,9 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 /**
  * Proxy implementation of {@link com.hazelcast.core.TransactionalMap} interface.
  */
-public class TransactionalMapProxy extends TransactionalMapProxySupport implements TransactionalMap {
+public class TransactionalMapProxy<K, V> extends TransactionalMapProxySupport implements TransactionalMap<K, V> {
 
-    private final Map<Data, TxnValueWrapper> txMap = new HashMap<Data, TxnValueWrapper>();
+    private final Map<Data, TxnValueWrapper<V>> txMap = new HashMap<Data, TxnValueWrapper<V>>();
 
     public TransactionalMapProxy(String name, MapService mapService, NodeEngine nodeEngine, Transaction transaction) {
         super(name, mapService, nodeEngine, transaction);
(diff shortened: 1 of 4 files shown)


LoneRifle:
Rebased to account for introduction of InternalSerializationService

gurbuzali:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java
```diff
@@ -126,29 +126,29 @@ public Object getForUpdate(Object key) {
...
 
     @Override
-    public Object put(Object key, Object value, long ttl, TimeUnit timeUnit) {
+    public V put(K key, V value, long ttl, TimeUnit timeUnit) {
         checkTransactionState();
         MapService service = getService();
         MapServiceContext mapServiceContext = service.getMapServiceContext();
         Data keyData = mapServiceContext.toData(key, partitionStrategy);
-        Object valueBeforeTxn = toObjectIfNeeded(putInternal(keyData, mapServiceContext.toData(value), ttl, timeUnit));
+        V valueBeforeTxn = toObjectIfNeeded(putInternal(keyData, mapServiceContext.toData(value), ttl, timeUnit));
```
jerrinot: this is not correct. 
`valueBeforeTxn` is not always `V`. It can be kept as `Data`  when a transaction was initiated by a client. see the `toObjectIfNeeded()`

LoneRifle: Given that line 147 can return valueBeforeTxn, that would imply that put could possibly return Data instead of V. Was this really intended? If so, I assume the rationale was so that we save the cost of serialization?

jerrinot: Hello @LoneRifle. yes, this is a correct observation. 
when the proxy is used on behalf of clients then it might return `Data` instead of deserialized value. It has 2 effects:
1. better performance as it avoids de-serialization and subsequent serialization back to Data (to be sent to a client)
2. the domain class does not need to be available on server classpath.(this is very important)

I think a better solution here is to use a different (internal?) API when executing on behalf of a client. but there might be other solutions as well. We are going to discuss it with @gurbuzali on Monday. @LoneRifle: are you available as well? we could include you in the discussion if you are interested. ;-)

LoneRifle: I'll follow up with you both on gitter.

gurbuzali:
`14:45:08 /appdisk/jenkins/jenkins/workspace/new-lab-fast-pr/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxy.java:320: Line has trailing spaces.`

LoneRifle:
Closing after discussions about ongoing changes to internal/external APIs. A PR will be sent later once TranscationalMapProxy is cleanly separated from its duties to clients.

PR #7504, opened 2016-02-11
Moved util.scheduler to internal

pveentjer:


Donnerbart:
I found this external project which uses some of the moved classes: https://github.com/taskurotta/taskurotta. But that should be no big issue for us and them.

As long as none of our Solution Architects or other Developers know about any important external usage, the change looks fine to me: :+1:

PR #7509, opened 2016-02-12
PerformanceMonitor should not rely on NodeEngineImpl

pveentjer:
This is needed refactoring to let the client make use of the PerformanceMonitor.


jerrinot:
:+1: (even I'm not too excited about `PerformanceLog` directly accessing fields from `PerformanceMonitor`)

pveentjer:
If it was an inner class, it would be fine. These classes belong to each other.

tombujok:
:+1: looks good

PR #7512, opened 2016-02-12
Reduced reliance performance plugins on NodeEngineImpl

pveentjer:
This cleanup is needed so that the same plugins can be used on the client side.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
@@ -36,7 +36,8 @@ import static java.lang.String.format;
  */
 public abstract class HazelcastProperties {
 
-    private final String[] properties = createProperties();
+    private final String[] values = createProperties();
+    private Properties properties;
 
     /**
      * Created the properties array.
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
```diff
@@ -55,34 +56,46 @@
      * @param hazelcastProperties array of {@link HazelcastProperty} to configure
      */
     protected void initProperties(Properties properties, HazelcastProperty[] hazelcastProperties) {
+        this.properties = properties;
         for (HazelcastProperty property : hazelcastProperties) {
```
tombujok: Hmm, I know you haven't changed this, but Properties is mutable. Don't we want' to guard it?

pveentjer: How do you want to guard it? There is no immutable wrapper available afaik.

Donnerbart: I think a user cannot change properties at runtime anyway, can't he?

pveentjer: I'm working on a proper cleanup PR for the HazelcastProperties. There is some complexity here in the constructor and methods that need to be implemented that can be removed by simplification.

Donnerbart:
:+1: (once green)

hazelcast/src/main/java/com/hazelcast/instance/HazelcastProperties.java
pveentjer: A user should not be.

This is more a question about responsible programming. The question is if we need to protect the exposing collections and perhaps I should protect it. 

But @tombujok if you are fine with it; I can do this cleanup while dealing with the HazelcastProperties. I can get rid of the Properties completley and expose a keySet method which returns a immutable set and a getValue method. This way nobody can by accident modify a Properties-instance while accessing the HazelcastProperties

tombujok:
:+1: let's rework the properties in other PR

PR #7514, opened 2016-02-12
Adds Metrics and PerformanceLog to the client.

pveentjer:
It is disabled by default.

The Hazelcast Java client can now benefit from the same metrics (and additional performance plugins)
as the Hazelcast Member. This will help to speed up debugging since we can see better what is happening
inside of the Java client.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -148,11 +148,14 @@ public class ClientConnectionManagerImpl implements ClientConnectionManager {
                 client.getName() + ".ClientInSelector",
                 loggingService.getLogger(NonBlockingIOThread.class),
                 OUT_OF_MEMORY_HANDLER);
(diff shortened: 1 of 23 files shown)


pveentjer:
This PR depends on the following PR:
https://github.com/hazelcast/hazelcast/pull/7516/

sancar:
:+1:

pveentjer:
There is a problem in the performance log when accessing the client instance. Need to fix that before this gets merged.

pveentjer:
run-lab-run

pveentjer:
verify

sancar:
I am seeing 3 scanAndRegister calls in pr. for input and output threads and for pending invocations. Is that all metrics we have related to a running client right now ? 

Are these metrics will call toString methods of given objects ? If that is the case, will input and output threads just print their name ?
See https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java#L310

pveentjer:
There are more probes; also the standard metrics sets like gc, os etc are included.

I'll create a different PR where I'll expose all kinds of internals of the connection manager, connections and more details of the invocation service. But the PR is already big enough. 

The probes for the io thread are based on the name of the thread. The advantage is that this name can also be connected to other external tooling like JFR that also uses the name of the thread.

PR #7516, opened 2016-02-12
Increased flexibility of HazelcastProperty

pveentjer:
Properties are currently inflexible + lot of duplication. With this change properties don't
need to be defined on the either Client or GroupProperty; but they can just
be used. This allows for a more flexible approach.

One of the reason to clean this up is that the metrics were using group property,
and this caused big problems when using them on a client.

So we can do stuff like:

```
public class OverloadedConnectionsPlugin extends PerformanceMonitorPlugin {

    public static final HazelcastProperty PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_PERIOD_SECONDS
            = new HazelcastProperty("hazelcast.performance.monitor.overloaded.connections.period.seconds", 0, SECONDS);

    public static final HazelcastProperty PERFORMANCE_MONITOR_OVERLOADED_CONNECTIONS_THRESHOLD
            = new HazelcastProperty("hazelcast.performance.monitor.overloaded.connections.threshold", 10000);
```

Where these properties are not defined anymore in GroupProperty or ClientProperty. And this plugin can be used on both client and server because the properties are not bound to either client or server.


Donnerbart:
I'm not super happy with the runtime resolution of the property value in `HazelcastProperties.getString()`. This was done once in the constructor before. But as long as we don't use properties somewhere in a performance critical path this should be fine for now.

Also the syntax was a bit nicer with an `enum`, since everything fit into a single line and was less code.

But I see the need for the flexibility between member and client properties. So we have to change it and maybe optimize later.

So :+1: with a

HappySad

jerrinot:
:+1:  (after a side-channel conversation with @pveentjer)
I'd recommend someone from client to have a look as well (CC @asimarslan)

pveentjer:
The key driver is that a HazelcastProperty is independent of Client/Server. And therefor one can define a property on e.g. a components (in my case the performance plugins) and reuse this property on both client and server.

pveentjer:
No Enterprise changes are needed.

sancar:
As far as I understand, this means properties will not be in one place anymore. And I was very fond of that, it was easy to find a configuration when you are not sure the exact name of it. There was only one file to check.

Can't we solve same problem by introducing a CommonProperties class like Client/GroupProperties? These two can extend CommonProperties. And things like PerformanceMonitor can go into CommonProperties. 

How does that sound ?

pveentjer:
Hi Sancar, 

all regular properties will still be in the same place. However there are a few 'hidden' properties that should only be used by people that know what they are doing. 

We can introduce a commons properties; the problem however is that the current style of the ClientProperties/MemberProperties doesn't allow for that since its an enum.  Also I don't want to include these private properties into this common file since they are private

sancar:
I see that you already removed enums. With new code, we can introduce CommonProperties easily.
My main concern is here to not to lose ability to find all configs in a few classes. And it will be more  convenient not just for external users, also for us.

pveentjer:
Sure. But this is matter of agreeing that all public properties should be in ClientConfig or MemberConfig. 

And one can still find all configuration; just search for all HazelcastProperty instances being created.

There is also some hidden configuration triggered by directly reading out system properties which can now be cleaned up as well.

sancar:
This will solve one problem. We can find a better solution for my concern later. 
I think it ok to merge this for now. :+1:

PR #7521, opened 2016-02-15
Client logging structure is improved.

sancar:
Static Logger.getLogger usage is removed.
ClientLoggingService.getLogger is used instead.
This logger has following format
ClientInstanceName [Cluster Name] [Version]

When client started it prints buildNumber and github commit id
as server.A sample line is as follows
hz.client_0 [dev] [3.7-SNAPSHOT] HazelcastClient 3.7-SNAPSHOT (20160201 - bc4f0b5) is CLIENT_CONNECTED

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientLoggingService.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientLoggingService.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.impl;
+
+import com.hazelcast.instance.BuildInfo;
+import com.hazelcast.logging.AbstractLogger;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.LogEvent;
+import com.hazelcast.logging.LogListener;
+import com.hazelcast.logging.Logger;
+import com.hazelcast.logging.LoggerFactory;
+import com.hazelcast.logging.LoggingService;
+import com.hazelcast.util.ConstructorFunction;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.logging.Level;
+
+import static com.hazelcast.util.ConcurrencyUtil.getOrPutIfAbsent;
+
+public class ClientLoggingService implements LoggingService {
+
+    private final String groupName;
+
+    private final ConcurrentMap<String, ILogger> mapLoggers = new ConcurrentHashMap<String, ILogger>(100);
+
+    private final ConstructorFunction<String, ILogger> loggerConstructor
+            = new ConstructorFunction<String, ILogger>() {
+
+        @Override
+        public ILogger createNew(String key) {
+            return new DefaultLogger(key);
+        }
+    };
+
+    private final LoggerFactory loggerFactory;
+    private final BuildInfo buildInfo;
+    private final String clientName;
+
+    public ClientLoggingService(String groupName, String loggingType, BuildInfo buildInfo,
+                                String clientName) {
+        this.groupName = groupName;
+        this.clientName = clientName;
+        this.loggerFactory = Logger.newLoggerFactory(loggingType);
+        this.buildInfo = buildInfo;
+    }
+
+    @Override
+    public void addLogListener(Level level, LogListener logListener) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void removeLogListener(LogListener logListener) {
+        throw new UnsupportedOperationException();
+    }
+
+    public ILogger getLogger(String name) {
+        return getOrPutIfAbsent(mapLoggers, name, loggerConstructor);
+    }
+
+    public ILogger getLogger(Class clazz) {
+        return getOrPutIfAbsent(mapLoggers, clazz.getName(), loggerConstructor);
+    }
+
+    private class DefaultLogger extends AbstractLogger {
+        final String name;
+        final ILogger logger;
+
+        DefaultLogger(String name) {
+            this.name = name;
+            this.logger = loggerFactory.getLogger(name);
+        }
+
+        @Override
+        public void log(Level level, String message) {
+            log(level, message, null);
+        }
+
+        @Override
+        public void log(Level level, String message, Throwable thrown) {
+            boolean loggable = logger.isLoggable(level);
+            if (loggable) {
+                String logMessage = clientName
+                        + " [" + groupName + "] [" + buildInfo.getVersion() + "] " + message;
+                logger.log(level, logMessage, thrown);
+            }
+        }
+
+        public void log(LogEvent logEvent) {
+            //unsued
+        }
+
+        @Override
+        public Level getLevel() {
+            return logger.getLevel();
+        }
+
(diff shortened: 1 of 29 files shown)


sancar:
run-lab-run

ihsandemir:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientLoggingService.java
```diff
@@ -0,0 +1,106 @@
...
+import com.hazelcast.logging.LoggerFactory;
+import com.hazelcast.util.ConstructorFunction;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.logging.Level;
+
+import static com.hazelcast.util.ConcurrencyUtil.getOrPutIfAbsent;
+
+public class ClientLoggingService {
```
pveentjer: Q1: Why not make use of com.hazelcast.logging.LoggingService and expose that in your system instead of this ClientLoggingService (and let this class implement that interface)

Q2: The LoggingServiceImpl can't be reused for the client? Perhaps an additional constructor can be added that solves your problems?

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
```diff
@@ -555,6 +561,10 @@ public NearCacheManager getNearCacheManager() {
         return nearCacheManager;
     }
 
+    public ClientLoggingService getClientLoggingService() {
```
pveentjer: I would return the interface 'LoggingService' and rename this method to getLoggingService.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientContext.java
```diff
@@ -84,6 +87,10 @@ public NearCacheManager getNearCacheManager() {
         return nearCacheManager;
     }
 
+    public ClientLoggingService getClientLoggingService() {
```
pveentjer: I would rename this to 'getLoggingService' and return a LoggingService instance.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -89,7 +88,7 @@ protected ClientInvocation(HazelcastClientInstanceImpl client,
         retryTimeoutPointInMillis = System.currentTimeMillis() + waitTimeResolved;
 
         clientInvocationFuture = new ClientInvocationFuture(this, client, clientMessage);
-
+        logger = client.getClientLoggingService().getLogger(ClientInvocation.class);
```
pveentjer: You don't want to do this lookup for every invocation. I would move this method to the invocation-service and if a logger is needed here do:

```
logger = invocationService.invocationLogger
```

Perhaps less pretty, but no unneeded work is added at the fast path.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
```diff
@@ -51,6 +50,7 @@ public ClientInvocationFuture(ClientInvocation invocation, HazelcastClientInstan
         this.executionService = (ClientExecutionServiceImpl) client.getClientExecutionService();
         this.clientMessage = clientMessage;
         this.invocation = invocation;
+        this.logger = client.getClientLoggingService().getLogger(ClientInvocationFuture.class);
```
pveentjer: The same here; you don't want to lookup the logger for every invocation.

hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientLoggingService.java
sancar: Two methods on the LoggingService are the reason I decided to no reuse it.

void addLogListener(Level level, LogListener logListener);
void removeLogListener(LogListener logListener);

This LogListener fired with a LogEvent class in which there is `Member` passed. For client this has no meaning. I can not refactor these parts, because it seems like it is public API.  I may pass `null ` for this Member. And there is the question of returning LoggingService from HazelcastInstance.getLoggingService or not for clients.

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
sancar: see https://github.com/hazelcast/hazelcast/pull/7521#discussion_r52975964

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientContext.java
sancar: see https://github.com/hazelcast/hazelcast/pull/7521#discussion_r52975964

hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientLoggingService.java
pveentjer: I think we should deprecate those methods. And in case of the client, I would throw an UnsupportedOperationException.

The problem with providing a different interface for a similar structure is apart from additional code complexity, you also get tight coupling. E.g. if a component expects LoggingService; then the client can't use it.

jerrinot: I see little reasons why a logging interface should be aware who is using it (client, server,...) 
Certainly the interface should be agnostic to its users. 

The current implementation depends on `MemberImpl`, but this dependency can be easily removed - it depends on it just to create a prefix. If you supply the prefix externally then you can fully re-use the implementation. Do I miss anything?

sancar:
@pveentjer I updated pr according to your comments.

pveentjer:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/impl/LifecycleServiceImpl.java
```diff
@@ -80,8 +79,10 @@ public boolean removeLifecycleListener(String registrationId) {
 
     public void fireLifecycleEvent(LifecycleEvent.LifecycleState lifecycleState) {
         final LifecycleEvent lifecycleEvent = new LifecycleEvent(lifecycleState);
-        getLogger().info("HazelcastClient[" + client.getName() + "]" + "["
-                + buildInfo.getVersion() + "] is " + lifecycleEvent.getState());
+        String revision = buildInfo.getRevision();
+        getLogger().info("HazelcastClient " + buildInfo.getVersion() + " ("
+                + buildInfo.getBuild() + (revision != null ? " - " + revision : "") + ") is "
```
cangencer: seems like the buildInfo here is now redundant

hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientLoggingService.java
```diff
@@ -0,0 +1,118 @@
...
+        public void log(Level level, String message) {
+            log(level, message, null);
+        }
+
+        @Override
+        public void log(Level level, String message, Throwable thrown) {
+            boolean loggable = logger.isLoggable(level);
+            if (loggable) {
+                String logMessage = clientName
+                        + " [" + groupName + "] [" + buildInfo.getVersion() + "] " + message;
```
cangencer: would a StringBuilder be quicker here?

hazelcast-client/src/main/java/com/hazelcast/client/impl/LifecycleServiceImpl.java
sancar: Output of this line is as follows:
hz.client_0 [dev] [3.7-SNAPSHOT] HazelcastClient 3.7-SNAPSHOT (20160201 - bc4f0b5) 
Where following is from DefaultLogger
hz.client_0 [dev] [3.7-SNAPSHOT]  
And more detailed info that includes buildInfo is from implementation in LifeCycleService 
HazelcastClient 3.7-SNAPSHOT (20160201 - bc4f0b5) 

Only version is duplicate here which was requirement and also in line with server 
Similar line of server is as follows:
INFO: [127.0.0.1]:5701 [dev] [3.7-SNAPSHOT] Hazelcast 3.7-SNAPSHOT (20160216 - bc4f0b5) starting at Address[127.0.0.1]:5701

ihsandemir:
:+1:

PR #7526, opened 2016-02-15
Packet read/write simplification

pveentjer:
Instead of dealing with each parameter from the head individually, they are done in 1 go.

So we check if there is space to read them all, or write them all. If not, we don't read any,
and if there is, we read them all. This reduces the amount and complexity of the code considerably.

Doesn't have a negative impact on performance:

```
Benchmark            Mode  Cnt   Score   Error  Units
PacketBenchmark.run  avgt   20  50.052 ± 1.840  ns/op

new:

Benchmark            Mode  Cnt   Score   Error  Units
PacketBenchmark.run  avgt   20  48.389 ± 0.578  ns/op
```

Using the following benchmark:

```
package hazelcast;

import com.hazelcast.nio.Packet;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;

import java.nio.ByteBuffer;

@BenchmarkMode(Mode.AverageTime)
@Warmup(iterations = 1)
@Measurement(iterations = 2)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Thread)
public class PacketBenchmark {

    private Packet sourceBuffer;
    private Packet dstBuffer;
    private ByteBuffer buffer;

    @Setup
    public void init() {
        sourceBuffer = new Packet(new byte[10]);
        dstBuffer = new Packet();
        buffer = ByteBuffer.allocate(100);
    }

    @Benchmark
    public void run() {
        buffer.clear();

        sourceBuffer.writeTo(buffer);
        sourceBuffer.reset();

        buffer.flip();

        dstBuffer.readFrom(buffer);
        dstBuffer.reset();
    }

    public static void main(String[] args) throws RunnerException {
        Options opt = new OptionsBuilder()
                .include( ".*" + PacketBenchmark.class.getSimpleName() + ".*" )
                .build();
        new Runner( opt ).run();
    }
}
```

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/Packet.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/Packet.java
@@ -20,11 +20,12 @@ import com.hazelcast.internal.serialization.impl.HeapData;
 
 import java.nio.ByteBuffer;
 
+import static com.hazelcast.nio.Bits.BYTE_SIZE_IN_BYTES;
 import static com.hazelcast.nio.Bits.INT_SIZE_IN_BYTES;
+import static com.hazelcast.nio.Bits.SHORT_SIZE_IN_BYTES;
 
 /**
- * A Packet is a piece of data send over the line. The Packet is used for member to member communication and old-client to
- * member communication.
+ * A Packet is a piece of data send over the line. The Packet is used for member to member communication.
  *
  * The Packet extends HeapData instead of wrapping it. From a design point of view this is often not the preferred solution (
  * prefer composition over inheritance), but in this case that would mean more object litter.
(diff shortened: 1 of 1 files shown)


pveentjer:
When I switch to a direct buffer, the performance difference is bigger (in favor of the cleanup)

before
Benchmark            Mode  Cnt   Score   Error  Units
PacketBenchmark.run  avgt   20  41.211 ± 1.987  ns/op

after
Benchmark            Mode  Cnt   Score   Error  Units
PacketBenchmark.run  avgt   20  27.879 ± 0.358  ns/op

jerrinot:
:+1:

mdogan:
:+1:

PR #7529, opened 2016-02-16
Restructure layout of partition package

mdogan:
Layout of partition package is adapted to the new _internal_ package
structure. All internal/private interfaces/classes are moved under
`com.hazelcast.internal.partition` package. Public (SPI) ones are left
in `com.hazelcast.partition` package.

`com.hazelcast.partition.InternalPartitionService` is split into two interfaces;
- `com.hazelcast.partition.IPartitionService`: as spi service interface
- `com.hazelcast.internal.partition.InternalPartitionService` which extends
`IPartitionService`: as internal interface

`com.hazelcast.partition.InternalPartition` is split into two interfaces;
- `com.hazelcast.partition.IPartition`: as spi partition interface
- `com.hazelcast.internal.partition.InternalPartition` which extends
`IPartition`: as internal interface

PS: *I* prefix used for public/SPI interfaces is subject to change before release,
it's not a final name yet.

EE counter-part: https://github.com/hazelcast/hazelcast-enterprise/pull/762


mdogan:
Test failures: https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/1388/console

```
12:21:31 java.lang.AssertionError: replicatedRecord.getUpdateTime() should return a greater update time
12:21:31 	at org.junit.Assert.fail(Assert.java:88)
12:21:31 	at org.junit.Assert.assertTrue(Assert.java:41)
12:21:31 	at com.hazelcast.replicatedmap.impl.record.ReplicatedRecordTest.testGetUpdateTime(ReplicatedRecordTest.java:111)
12:21:31 
12:21:31 getLastAccessTime(com.hazelcast.replicatedmap.impl.record.ReplicatedRecordTest)  Time elapsed: 0.114 sec  <<< FAILURE!
12:21:31 java.lang.AssertionError: replicatedRecord.getLastAccessTime() should return a greater access time
12:21:31 	at org.junit.Assert.fail(Assert.java:88)
12:21:31 	at org.junit.Assert.assertTrue(Assert.java:41)
12:21:31 	at com.hazelcast.replicatedmap.impl.record.ReplicatedRecordTest.getLastAccessTime(ReplicatedRecordTest.java:133)
12:21:31 
```

mdogan:
run-lab-run

metanet:
:+1:

pveentjer:
Grabbing one of these and starting review

635806393717942917-2045720344_coffee-cup-and-beans-wood-background

pveentjer:
Looks good to me. I have 2 issues.

1: Can we come up with a cooler name than IPartition/IPartitionService? Since you provide enhanced/extended abilities compared to a regular Partition/PartitionService, perhaps we can name it e.g. ExtendedPartition, ExtendedPartitionService? I'm completely open for suggestions here, as long as we don't use I as prefix.

2: We need to wait till everyone, especially the big boss, agrees that we need to continue with these package migrations. So we can't merge yet unless he gives the thumbs up. All other migrations have been parked as well.

mdogan:
@pveentjer: I agree about naming and already mentioned that. For second part, I think we've already done many changes in this area, one less or more won't change the situation any further. 

If we won't change structure of partition package according to the new layout then that means it will be postponed until 3.8. @metanet and I are planning to do structural changes and cleanups in this package for our tasks (data loss, graceful shutdown). I don't want to struggle with resolve lots of conflicts after a rebase in the middle of the sprint.

@enesakar, @jerrinot, @gurbuzali: your thoughts?

gurbuzali:
normally I was against these changes and I've stated this to both @jerrinot and @enesakar.
but we've agreed on that they are necessary and should be done at the beginning of the release cycle.
we've already merged a lot of moving/renaming PR, we should merge this one too

enesakar:
@mdogan: agree with you as I know we are doing such changes in internal classes that we do not expect to be used by users. 

@gregrluck has some worry about breaking any user code. He is doing an investigation with help of  @bilalyasar to be sure that we are not breaking any user code.

metanet:
This PR is not a refactoring for just code structure and aesthetics. We already have an ongoing plan for making improvements in migration system and this PR will help us a lot to clean up migration code and apply our changes more easily. 

I think we should merge this PR when we ensure that it doesn't break any public API.

gregrluck:
"I think we should merge it when we ensure that it doesn't break any public API."

I agree with this. How do we determine this? I have Bilal working on this topic on Thursday/Friday.

How do you believe we gain assurance of not breaking code.

If we believe we are going to break code in this change and the others it makes me lean towards calling this 4.0. That is one thing.

But I worry about disruption to many users and frameworks out there. 

So we need to gain assurance this and other changes will not break code.

mdogan:
If you count SPI as part of public API in this context, this PR definitely breaks it like some of the previous PRs merged before. 

Note that, in the near future, we may be forced into breaking SPI to be able to fix challenging issues in the top list.

metanet:
If we are talking about user level API, this PR should not break anything. If we are talking about SPI level, we have been already doing changes in SPI level in new releases. Even in this release, there are a few merged PRs making class movements etc (=breaking SPI).

gregrluck:
Do we know of any external users of the SPI?

If the nature of the SPI is that it is constantly changing low-level interface, then perhaps we should describe it as such and put a “subject to change without notice” in the package info for it. 

Regards

Greg Luck

web: http://gregluck.com <http://gregluck.com/>
skype: gregrluck
mobile US: +1 650 924 6244
mobile Australia: +61 408 061 622

pveentjer:
@gregrluck the SPI contains many parts. For example the discovery API com.hazelcast.spi.discovery; imho this is a stable API and we should be very careful to prevent causing incompatibilities. 

Afaik there are users of the Discovery SPI; see the blogpost/twitter messages of people writing their own discovery integration

The SPI to write your own distributed data-structures, which can mostly be found in com.hazelcast.spi root is a direct reflection of our internal engine for operation execution. This is a lot more in flux since we are improving the engine all the time. If this is fixated; progress in the engine will be halted.

Afaik we have no users of the this part of the SPI. Since we have a continuous stream of incompatibilities here (there are some leaks so that internal code is exposed) an incompatibility and get these problematic parts removed from the SPI. And that is exactly what Basri/Mehmet have been doing. They have fixed a design flaw of the SPI.

PR #7530, opened 2016-02-16
Fixed Hazelcast properties API

Donnerbart:
Fixed properties API by restoring public classes from Hazelcast 3.6 and deprecating them properly. Moved changed implementations from Hazelcast 3.7 to internal packages.

This should work, since we just exposed the interface `HazelcastProperty` as a parameter in two methods in `Config` and `ClientConfig` in Hazelcast 3.6. So the implementations `GroupProperty` and `ClientProperty` were never returned to the user. Those implementations were also marked as `@PrivateAPI`.

I restored the interface `HazelcastProperty` with the original content. I also restored the abstract class `HazelcastProperties`, just in case someone implemented it. Also the classes `GroupProperties` and `ClientProperties` are still there, which were always a grey-zone regarding public or private API.

All classes are deprecated, also the methods in `Config` and `ClientConfig` which used `HazelcastProperty`. I refactored all usages of the deprecated methods.

The new implementation from Peter is now in new internal packages, so it's 100% clear that they are private API. As far as I can see they are not exposed to the user.


Donnerbart:
I know, the PR is big and the topic is complex. Here is more information what this PR will do.

I removed some files:
```
com.hazelcast.instance.GroupProperty (@PrivateAPI, never exposed to public API)
com.hazelcast.client.config.ClientProperty (@PrivateAPI, never exposed to public API)
```

I reverted some files to their 3.6 version:
```
com.hazelcast.instance.GroupProperties (was always there and people might have used it)
com.hazelcast.instance.HazelcastProperties (abstract, there could be implementations)
com.hazelcast.instance.HazelcastProperty (was used in Config and ClientConfig)
com.hazelcast.client.config.ClientProperties (was always there and people might have used it)
```
All files have been completely deprecated. I also had to change the implementation of `HazelcastProperties.getEnum()` to throw `UnsupportedOperationException`, because it had a internal `GroupProperty` dependency.

I created new files with Peter's 3.7 changes and used them in our codebase:
```
com.hazelcast.internal.properties.GroupProperties (used in Node)
com.hazelcast.internal.properties.GroupProperty (defines the member properties)
com.hazelcast.internal.properties.HazelcastProperties (abstract class)
com.hazelcast.internal.properties.HazelcastProperty (base class for a property)
com.hazelcast.client.internal.properties.ClientProperties (used in HazelcastClientInstanceImpl)
com.hazelcast.client.internal.properties.ClientProperty (defines the client properties)
```

I also deprecated the two methods in `Config` and `ClientConfig` which use the deprecated `com.hazelcast.instance.HazelcastProperty`. I changed the code to not use those methods anymore.

So the public API is restored and we have a new, completely private API which is used by us.

**PRO:** The 3.6 public API is restored. The FindBugs issue is resolved.

**CONTRA:** Same class names in public and private API, but that can be improved with another PR. For example we can get rid of `ClientProperties` and `GroupProperties` and just use a common `HazelcastProperties` with a constructor, which takes `Properties` instead of `Config` and `ClientConfig`. Then we just have a collision on `HazelcastProperties` and `HazelcastProperty`, which maybe resolved with a simple rename of the new **internal** classes.

Donnerbart:
run-lab-run

jerrinot:
:+1: from me. it can't be merged before the Enterprise counter-part is ready (if needed)
@pveentjer: can you please have a look?

Donnerbart:
If we have consensus on a merge, I'll prepare the Enterprise PR. This PR already needs a rebase every day, so I don't want to maintain two PRs until this one is reviewed :)

bilalyasar:
If user has (in 3.6)
```
clientConfig.setProperty(ClientProperty.INVOCATION_TIMEOUT_SECONDS, String.valueOf(Integer.MAX_VALUE));
```
now it should
```
clientConfig.setProperty(ClientProperty.INVOCATION_TIMEOUT_SECONDS.getName(), String.valueOf(Integer.MAX_VALUE));
```
right?
So user should change the code?

Donnerbart:
Internally we should use them with `.getName()`, yes. An external user should not use `ClientProperty` or `GroupProperty` at all. That was always marked as private API and should never have been exposed via the config classes (sorry for that, slipped through in 3.6). They should use the string identifiers, which are documented in the reference manual.

What is missing is to retrieve the resulting value from the group properties (which could be set via configuration, system property or the default value). But that should be another PR, since it's a new public API, which didn't exist before.

bilalyasar:
theoretically you are right. An external user shouldn't use those api since it is private api.
But in [code samples](https://github.com/hazelcast/hazelcast-code-samples/search?utf8=✓&q=GroupProperty) and [documentation](https://github.com/hazelcast/hazelcast-reference-manual/search?utf8=✓&q=groupproperty) we already have examples that supports using them.

anyway, i am ok with your PR. just want to underline some points :)

Donnerbart:
Good findings! We should fix those leakages. I checked the documentation in 3.6 and there was no reference to the old `GroupProperties` at that time. And `GroupProperty` should really not be used in public, since it's even marked as `@PrivateAPI`.

Donnerbart:
Code samples should be fixed, PR for reference manual is pending: https://github.com/hazelcast/hazelcast-reference-manual/pull/133

Donnerbart:
What do we do with this PR? I had to resolve merge conflict in 30 files today. It's not really fun to keep this PR mergeable :(

@bilalyasar @jerrinot

jerrinot:
@Donnerbart: what's preventing your from pressing the merge button? 
@bilalyasar: I read your comment as :+1: , is this correct?

Donnerbart:
As discussed with Jaromir and Peter I will merge this PR once it's green again. I think the result of the PR will be quite clear if you see the new package in IDEA on your machine :) If there are any unforeseen consequences I'll be happy to fix them in a new PR.

PR #7532, opened 2016-02-16
Added option for direct bytebuffers

pveentjer:
for the tcp functionality one can now configure direct bytebuffers.
By default it is configured as false, and will behave the same as
the system behaved before this change.


jerrinot:
:+1:

mtopolnik:
:+1:

PR #7536, opened 2016-02-17
Fix #7291 concurrent request in webfilter

### LINKED ISSUE #7291: WebFilter(3.6-RC1): session attributes being lost/overwritten with old values using deferred-write by reading them
Assume two parallel http requests A and B, sticky-session set to false, deferred-write to true:

__Time tick 1:__
A: reads attributes X=1 and Y=1 from session

__Time tick 2:__
A: sets attribute X=2

__Time tick 3:__
B: reads attributes X=1 and Y=1 from session

__Time tick 4:__
B: sets attribute Y= 2

__Time tick 5:__
A: request finished. WebFilter "commits" session to distributed map

Note:
Distributed map contains X=2 and Y=1

__Time tick 6:__
B: request finished. WebFilter "commits" session to distributed map

The distributed map now contains Y=2 but again the old X=1 that was read by B but was never set. X=2 is overwritten and the data is lost.

The expected behavior is, that attributes which were only read and not modified, will not be written to the distributed map.

In general:
Reading attributes with deferred-write enabled, triggers a write operation for that value in the distributed map (maybe only, when another attribute was changed - not tested). All other changes to that value after the value was read, are overwritten.

Question regarding performance:
If you check many (lets say a few 10k values) session attributes via getAttribute (deferred-write is enabled), will all these values be written to the distributed map? It might be, that we are hitting this performance bug. Only 2-3 attributes are changed, but thousands are written to the map.

bilalyasar:
you can find full explanation at #7291

### DIFF
--- a/hazelcast-wm/src/main/java/com/hazelcast/web/HazelcastHttpSession.java
+++ b/hazelcast-wm/src/main/java/com/hazelcast/web/HazelcastHttpSession.java
@@ -338,7 +338,9 @@ public class HazelcastHttpSession implements HttpSession {
 
     public void updateReloadFlag() {
         for (Map.Entry<String, LocalCacheEntry> entry : localCache.entrySet()) {
-            entry.getValue().setReload(true);
+            if (!entry.getValue().isDirty()) {
+                entry.getValue().setReload(true);
+            }
         }
 
     }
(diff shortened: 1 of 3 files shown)


mesutcelik:
can you please send forward port to new repo?

kobalski:
:+1:

emrahkocaman:
:+1:

PR #7539, opened 2016-02-17
Rename MemoryAccessor.DEFAULT to MEM

mtopolnik:
... making it convenient for import static.

Also improve some Javadoc.

There is an enterprise PR which depends on this one and cannot be merged until this is merged.


serkan-ozal:
:+1: merge once green

PR #7540, opened 2016-02-17
Add test rule: AssertEnabledRule

mtopolnik:
Allows a test method to be marked as "requires Java assertions to be enabled"

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/AssertEnabledFilterRule.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/AssertEnabledFilterRule.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.test;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/test/AssertEnabledFilterRule.java
```diff
@@ -0,0 +1,45 @@
...
+ * Typically such a test will expect an {@code AssertionError} to be thrown.
+ */
+public class AssertEnabledFilterRule implements TestRule {
+    @Override public Statement apply(final Statement base, final Description description) {
+        if (description.getAnnotation(RequireAssertEnabled.class) == null) {
+            return base;
+        }
+        return new Statement() {
+            @Override public void evaluate() throws Throwable {
+                boolean assertEnabled = false;
```
serkan-ozal: **This is enough.**

On the other hand, in fact assertions can be specified classloader based, package based or even class based.
So in theory this means that, assertion might be enabled in here but disabled at the tested class itself (`Long2LongElasticMapHsa` in our case at another PR and also I know this is very edge case that we do not need to consider here).

So the another way (but complex) might be like this:

``` java
@Retention(RUNTIME)
@interface RequireAssertEnabled {
    
    Class<?> targetClass() default Object.class;
    
}

class AssertEnabledFilterRule implements TestRule {
    
    @Override 
    public Statement apply(final Statement base, final Description description) {
        final RequireAssertEnabled assertEnabled = 
                description.getAnnotation(RequireAssertEnabled.class);
        if (assertEnabled == null) {
            return base;
        }
        return new Statement() {
            @Override 
            public void evaluate() throws Throwable {
                boolean assertionEnabled = false;
                Class<?> targetClass = assertEnabled.targetClass();
                if (Object.class.equals(targetClass)) {
                    // We are considering test class's assertion status
                    assertionEnabled = description.getTestClass().desiredAssertionStatus();
                } else {
                    // We are considering specified class's assertion status
                    assertionEnabled = targetClass.desiredAssertionStatus();
                }
                if (assertionEnabled) {
                    base.evaluate();
                } else {
                    System.err.println("WARNING! Test cannot run when Java assertions are not enabled (java -ea ...): "
                            + description.getDisplayName());
                }
            }
        };
    }

}
```

serkan-ozal:
minor comment :+1:

mtopolnik:
But my `assert assertEnabled = true` statement runs just prior to running the actual test method. There is no more direct way to test it than that.

serkan-ozal:
As I said, in theory it doesn't matter the statement runs prior to running the actual test method or not. This means that, assertion might be enabled in here but disabled at the tested class itself. The best way is checking the assertion status if there is single class with asserts to be tested.

Issue #1181, opened 2013-11-21
Adding an InitialMembershipListener to ClientConfig using ClientConfig.addListenerConfig throws NullPointerException.


alexturc:
Version used is 3.1.2

java.lang.NullPointerException
    at com.hazelcast.client.proxy.ClientClusterProxy.getMembers(ClientClusterProxy.java:53)
    at com.hazelcast.client.spi.impl.ClientClusterServiceImpl.addMembershipListener(ClientClusterServiceImpl.java:319)
    at com.hazelcast.client.spi.impl.ClientClusterServiceImpl.<init>(ClientClusterServiceImpl.java:91)
    at com.hazelcast.client.HazelcastClient.<init>(HazelcastClient.java:119)
    at com.hazelcast.client.HazelcastClient.newHazelcastClient(HazelcastClient.java:149)

PR #7543, opened 2016-02-18
Fix fired events when initialMembersipListener is used on client

### LINKED ISSUE #7430: Callback order and content of InitialMembershipEvent
For clients, `InitialMembershipListener.init` is called _after_ all `MembershipListener.memberAdded`, which contradicts the docs: 
> The InitializingMembershipListener is a `MembershipListener` that first receives a `InitialMembershipEvent` when it is registered so it immediately knows which members are available. After that event has been received, it will receive the normal MembershipEvents

For members, `InitialMembershipListener.init` is called immediately, but `getMembers` only contains itself, which also seems to contradict the intent described in the docs.

sancar:
When initial membership event added before client is started,
it was firing individual ADDED events for current memberlist,
along with InitialMembershipEvent. Only latter is preserved,
other events are removed.

Related tests are added.
Membershiplistener related tests in ClientRegressionWithMockNetworkTest
are moved to MembershipListenerTest.

fixes #7430

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/MembershipListenerTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/MembershipListenerTest.java
@@ -1,11 +1,11 @@
 package com.hazelcast.client;
 
-/**
- * User: danny Date: 11/28/13
- */
-
+import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.client.test.TestHazelcastFactory;
+import com.hazelcast.config.ListenerConfig;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.InitialMembershipEvent;
+import com.hazelcast.core.InitialMembershipListener;
 import com.hazelcast.core.Member;
 import com.hazelcast.core.MemberAttributeEvent;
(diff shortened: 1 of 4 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/MembershipListenerTest.java
```diff
@@ -124,20 +163,83 @@ public void addNullListener_thenException() throws Exception {
         return result;
     }
 
+    /**
+     * related to issue #1181
+     */
+    @Test
+    public void testAddInitialMembership_whenListenerAddedViaClientConfig() throws InterruptedException {
```
ihsandemir: why is there no assert? What is being tested?

sancar: Before the fix, It was throwing NullPointer exception according to issue mentioned in #1181 . It is testing that it does not throw any exception.

sancar:
clientRelated

sancar:
clientRelated

ihsandemir:
:+1:

asimarslan:
:+1:

PR #7544, opened 2016-02-18
Fix fired events when initialMembersipListener is used on client

### LINKED ISSUE #7430: Callback order and content of InitialMembershipEvent

sancar:
When initial membership event added before client is started,
it was firing individual ADDED events for current memberlist,
along with InitialMembershipEvent. Only latter is preserved,
other events are removed.

Related tests are added.
Membershiplistener related tests in ClientRegressionWithMockNetworkTest
are moved to MembershipListenerTest.

fixes #7430
backport of #7543


ihsandemir:
:+1:

PR #7546, opened 2016-02-18
added multi-threading processing for topic messages; see #7443

dsukhoroslov:
The second try, instead of #7472. Most changes are around TopicConfig now. Please have a look.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java
+++ b/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java
@@ -40,6 +40,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -63,6 +64,7 @@ public class TopicService implements ManagedService, RemoteService, EventPublish
             };
     private EventService eventService;
     private ILogger logger;
+    private final AtomicInteger counter = new AtomicInteger(0);
 
     @Override
     public void init(NodeEngine nodeEngine, Properties properties) {
@@ -143,9 +145,10 @@ public class TopicService implements ManagedService, RemoteService, EventPublish
         getLocalTopicStats(topicName).incrementReceives();
     }
 
-    public void publishEvent(String name, TopicEvent event) {
+    public void publishEvent(String name, TopicEvent event, boolean multithreaded) {
         Collection<EventRegistration> registrations = eventService.getRegistrations(TopicService.SERVICE_NAME, name);
-        eventService.publishEvent(TopicService.SERVICE_NAME, registrations, event, name.hashCode());
+        int partitionId = multithreaded ? counter.incrementAndGet() : name.hashCode();
+        eventService.publishEvent(TopicService.SERVICE_NAME, registrations, event, partitionId);
     }
 
     public String addMessageListener(String name, MessageListener listener, boolean localOnly) {
--- a/hazelcast/src/test/java/com/hazelcast/topic/TopicStressTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/topic/TopicStressTest.java
@@ -14,6 +14,7 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -51,11 +52,16 @@ public class TopicStressTest extends HazelcastTestSupport {
     private CountDownLatch startLatch;
     private PublishThread[] publishThreads;
     private Map<String, List<MessageListenerImpl>> listenerMap;
+    
+    @Parameterized.Parameter
+    public boolean multiThreadingEnabled;
+    
 
     @Before
     public void setUp() {
         TopicConfig topicConfig = new TopicConfig();
         topicConfig.setName("topic*");
+        topicConfig.setMultiThreadingEnabled(multiThreadingEnabled);
 
         Config config = new Config();
         config.addTopicConfig(topicConfig);
--- a/hazelcast/src/test/java/com/hazelcast/topic/TopicTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/topic/TopicTest.java
@@ -32,6 +32,7 @@ import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.TestHazelcastInstanceFactory;
+import com.hazelcast.test.annotation.NightlyTest;
 import com.hazelcast.test.annotation.ParallelTest;
 import com.hazelcast.test.annotation.QuickTest;
 import com.hazelcast.topic.impl.TopicService;
@@ -39,7 +40,9 @@ import com.hazelcast.util.UuidUtil;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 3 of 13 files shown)


hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java
```diff
@@ -145,7 +147,14 @@ public void incrementReceivedMessages(String topicName) {
 
     public void publishEvent(String name, TopicEvent event) {
```
ahmetmircik: What about this one? (by making `counter` an `AtomicInteger`)
- no need to create Long
- no need to find TopicConfig in every run

```
 public void publishEvent(String name, TopicEvent event, boolean multiThreadingEnabled) {
        Collection<EventRegistration> registrations = eventService.getRegistrations(TopicService.SERVICE_NAME, name);
        int orderKey = multiThreadingEnabled ? counter.incrementAndGet() : name.hashCode();
        eventService.publishEvent(TopicService.SERVICE_NAME, registrations, event, orderKey);

eventService.publishEvent(TopicService.SERVICE_NAME, registrations, event, multiThreadingEnabled ? orderKey)
    }
```

dsukhoroslov: Yes, looks reasonable. Will do.

ahmetmircik: You can put [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicProxySupport.java#L40) a field like `private final TopicConfig config` and can pass `config.multiThreadingEnabled ` to [this method](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicProxySupport.java#L94)

dsukhoroslov:
@ahmetmircik, is it ok now? what else shoud be done in scope of this PR?

hazelcast/src/test/java/com/hazelcast/topic/TopicTest.java
```diff
@@ -627,4 +629,76 @@ public void onMessage(Message msg) {
         assertEquals(1000, stats.getPublishOperationCount());
         assertEquals(2000, stats.getReceiveOperationCount());
     }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testTopicMultiThreading() throws Exception {
```
ahmetmircik: Seems a long running test (longer than 10 seconds) please annotate it with `NightlyTest ` as in [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/map/impl/InterceptorRegistryTest.java#L67)

hazelcast/src/test/java/com/hazelcast/topic/TopicStressTest.java
```diff
@@ -56,6 +56,7 @@
     public void setUp() {
         TopicConfig topicConfig = new TopicConfig();
         topicConfig.setName("topic*");
+        topicConfig.setMultiThreadingEnabled(true);
```
ahmetmircik: You can do this test parameterized on `multiThreadingEnabled` by setting it to `true/false` in each run as in [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java#L82)

hazelcast/src/test/java/com/hazelcast/topic/TopicTest.java
Donnerbart: Just a note: If that method if important for code coverage, please use `SlowTest`. If it's just for correctness or a kind of integration/stress test use `NightlyTest`. All categories will be executed, but just `SlowTest` is considered for our code coverage reports.

ahmetmircik:
@dsukhoroslov i added my last comments when those are fixed i am ok with this PR, thanks for the contribution.

hazelcast/src/test/java/com/hazelcast/topic/TopicTest.java
dsukhoroslov: this depends on the machine where tests are runing. on my machine it states
`Finished Running Test: testTopicMultiThreading in 8,576 seconds.`
the test code is the same as in `testTopicLocalOrder` and `testTopicGlobalOrder` and takes similair time. But these tests are not anotated.

dsukhoroslov:
done, please have a look.

ahmetmircik:
seems mergeable to me, guys @eminn @jerrinot can you also have a look please?

eminn:
looks good to merge :+1: 

@dsukhoroslov thanks for the contribution.

ahmetmircik:
run-lab-run

ahmetmircik:
@dsukhoroslov can you please fix check-style errors in this PR? 
To see the errors, you can run the checks via : `mvn verify -Pcheckstyle -DskipTests`

dsukhoroslov:
@ahmetmircik I did, pleasee have a look

hasancelik:
run-lab-run

hasancelik:
@dsukhoroslov pr-builder failed again due to checkstyle errors.

dsukhoroslov:
yep, I see thouthands of checkstyle errors. But they do not relate to the changes that I did.

dsukhoroslov:
@hasancelik found some errors in TopicConfig, fixed

ahmetmircik:
run-lab-run

hazelcast/src/test/java/com/hazelcast/topic/TopicStressTest.java
```diff
@@ -51,11 +52,16 @@
     private CountDownLatch startLatch;
     private PublishThread[] publishThreads;
     private Map<String, List<MessageListenerImpl>> listenerMap;
+    
+    @Parameterized.Parameter
```
jerrinot: is parametrization working with the `HazelcastSerialClassRunner` ?

dsukhoroslov: I have no idea. This change was requested by @ahmetmircik, see above.

dsukhoroslov: Hi @jerrinot,

what the status of the PR, can it be merged?

jerrinot: apparently the `HazelcastSerialClassRunner` supports parametrization thanks to #7720.

jerrinot:
I think this PR is (almost) good to merge: The last thing is this: The property is now called `multi-threading`. It could be (mis-)interpreted as this promise: "When you set this property to `true` then Hazelcast will always use multiple threads to invoke the listener". 

I think something like `concurrent` might be more suitable. It's not a promise of what Hazelcast will do, it's more a capability of the listener: "This listener can be invoked concurrently". 

So for now there won't be any difference, but it reduces a change of misunderstanding. I can foresee a github issues saying: "I set the `multi-threading`, but Hazelcast is using only 1, (2, 3, 4...) threads to invoke my listener."

WDYT? Does it make sense? I think it's this is resolved then we can hit the merge button.

dsukhoroslov:
I assume people will read javadoc at least. The javadoc states:
    /**
     * Enable multi-threaded message handling. When enabled any thread from events
     * thread pool can be used for incoming message processing. Otherwise only one
     * dedicated thread will be used to handle topic messages.
     * Note: it can be enabled only in case when global ordering is disabled. Moreover,
     * the local message ordering is not supported in this mode also. This means the
     * messages produced by local publisher can be processed by several threads with
     * no ordering guarantee.
     *
     * @param multiThreadingEnabled set to true to enable multi-threaded message processing, false to disable
     * @return The updated TopicConfig
     */

Do you find it confusing?

jerrinot:
@dsukhoroslov: I see a subtle difference between:
1. Explicit configuration to use multiple threads. 
vs.
2. Configuration to indicate this listener has a *capability* to be invoked by concurrently

What do you think?

Anyway, even the current PR looks good. I just wanted to show it for completeness. If you think there is no real difference from a user-point of view then let me know. We can merge it as it is.

dsukhoroslov:
Well, feature naming is a bit subjective thing, you know. Someone can find it confusing when others are not.. To be honest, I don't see the difference between `multiThreading` and `concurrent`. Let's use the same naming conventions as for other Topic properties: `globalOrdering`, `statistics`? I'd say they're more about _explicit configuration_ then a _capability_..

jerrinot:
@dsukhoroslov: that's a good point with the other names. I'm going to merge this one. 

Many thanks for this contribution, it's very much appreciated!

jerrinot:
@dsukhoroslov: and a very last thing: could you squash your commit into 1? There is no reason to have isolated commits related to checkstyle only. 

GitHub now has an ability to squash commits before merging them, but I want to make sure your name will stay in git history to give you the credit you surely deserve!

dsukhoroslov:
Welcome, @jerrinot. Regarding the squash: it'll take some time. My present env is quite limited in connection, the `git fetch origin` throws error: `fatal: unable to access 'https://github.com/dsukhoroslov/hazelcast.git/': Failed to connect to github.com port 443: Connection refused`. Will try to do it from home in the evening.

dsukhoroslov:
I did squash, please have a look. The last commit done by eclipse egit plugin after merge. Don't know how to avoid it.

Donnerbart:
run-lab-run

Donnerbart:
verify

Donnerbart:
Git rebase is your friend. With that command you can pull the changes from another branch in yours and add your commit(s) on top of it again. But I don't think it's mandatory that you change this PR again.

hasancelik:
run-lab-run

Donnerbart:
run-lab-run

jerrinot:
I merged this PR as extra commit does not overweight usefulness of this PR.
@dsukhoroslov: thank you for this contribution, it's much appreciated!

dsukhoroslov:
Glad to hear this, thanks!

PR #7552, opened 2016-02-18
Get rid of `UnsafeHelper` and use `MemoryAccessor` instead

serkan-ozal:
`UnsafeHelper` is not removed because it is still used by other modules (for example hibernate modules) and so it is just deprecated.

This PR just replaces `UnsafeHelper` with `MemoryAccessor` and there is no renaming about **Unsafe** in properties, fields, classes, etc ... But I think they should be renamed later so there will be no **Unsafe** code, term, name, etc in our code base.


Donnerbart:
:+1: (once green)

pveentjer:
:+1:

PR #7559, opened 2016-02-19
Resolves findbugs problems GroupProperties

pveentjer:
; group properties need to be checked for i…ncomaptibilities


Donnerbart:
:+1:

PR #7562, opened 2016-02-19
Introducing `@AutoRegisteredTestRule` …

serkan-ozal:
… for connecting test rule requirement annotation with test rule implementation and registering test rule automatically. By this way, no need to define a `@TestRule` annotated field inside test to specific test rule, you just need to mark the whole test class or required methods with the required test rule annotation.

For example, just annotate methods which need to assertion enabled with `@RequireAssertEnabled` and since `@RequireAssertEnabled`  is connected to its implementation `AssertEnabledFilterRule` via `@AutoRegisteredTestRule` on its definition, the rule will be created and registered automatically.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
@@ -23,17 +23,21 @@ import org.apache.log4j.MDC;
 import org.junit.After;
 import org.junit.Test;
 import org.junit.internal.runners.statements.RunAfters;
+import org.junit.rules.TestRule;
 import org.junit.runners.model.FrameworkMethod;
 import org.junit.runners.model.InitializationError;
 import org.junit.runners.model.MultipleFailureException;
 import org.junit.runners.model.Statement;
+import org.junit.runners.model.TestClass;
 
+import java.lang.annotation.Annotation;
 import java.lang.management.ManagementFactory;
 import java.lang.management.ThreadInfo;
 import java.lang.management.ThreadMXBean;
 import java.lang.reflect.Method;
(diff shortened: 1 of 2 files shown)


serkan-ozal:
run-lab-run

hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
```diff
@@ -183,6 +187,53 @@ private Repeat getRepeatable(FrameworkMethod method) {
...
+                Class<? extends TestRule> testRuleClass = autoFilterRule.testRule();
+                testRuleClasses.add(testRuleClass);
+            }
+        }
+
+        // Find the required test rule classes from methods
+        List<FrameworkMethod> annotatedMethods = testClass.getAnnotatedMethods();
+        for (FrameworkMethod annotatedMethod : annotatedMethods) {
+            Annotation[] methodAnnotations = annotatedMethod.getAnnotations();
+            for (Annotation annotation : methodAnnotations) {
```
bilalyasar: i couldn't get the idea here, 
first question:  this `getTestRules` is called for each methods or for each classes?
second question: if it calls for each class, you collect all rules and apply all methods?
Can you clarify please?

serkan-ozal: > first question: this getTestRules is called for each methods or for each classes?

Called for each method.

> second question: if it calls for each class, you collect all rules and apply all methods?

There is no behavioral change. It is like defining rule as field. Rules are applied for all test methods and rule checks the method itself as done in `AssertEnabledFilterRule`

bilalyasar:
:+1:

PR #7564, opened 2016-02-20
Add @RequireAssertEnabled where applicable

mtopolnik:
Recently a JUnit test rule was added which can filter out tests that depend on Java `assert` being enabled, when `assert` is not enabled. Those tests would otherwise inappropriately fail when the problem is not in the code, but in the test setup.

This PR applies the above test rule where applicable.


serkan-ozal:
Rule definition by field should not be needed after my merged PR https://github.com/hazelcast/hazelcast/pull/7562
Just `@RequireAssertEnabled` should be enough

mtopolnik:
Confirmed that it indeed works without it. Removing and rebasing.

serkan-ozal:
:+1:

PR #7572, opened 2016-02-22
Prefer instance name property or instance name config over config url …

### LINKED ISSUE #7567: Instance name should not be overriden while creating cache manager from specified config file
Currently, while creating `CacheManager` backed by custom hazelcast instance from given config file, regardless from instance name is specified or not in the config file, instance name is set as config url but when there is specified instance name in the config file, that instance name should be used instead of config url.

Also, it would be better to specify instance name via `(HazelcastCachingProvider.HAZELCAST_INSTANCE_NAME` property when instance config is taken from specified config file via `HazelcastCachingProvider.HAZELCAST_CONFIG_LOCATION`.

serkan-ozal:
… as name of backed hazelcast instance while creating `HazelcastServerCacheManager`

Fixes #7567

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
@@ -74,7 +74,9 @@ public final class HazelcastServerCachingProvider
     private HazelcastInstance getOrCreateInstance(ClassLoader classLoader, Properties properties, boolean isDefaultURI)
             throws URISyntaxException, IOException {
         String location = properties.getProperty(HazelcastCachingProvider.HAZELCAST_CONFIG_LOCATION);
-        // If config location is specified, get instance through it.
+        String instanceName = properties.getProperty(HazelcastCachingProvider.HAZELCAST_INSTANCE_NAME);
+
+        // If config location is specified via properties, get instance through it.
         if (location != null) {
             URI uri = new URI(location);
             String scheme = uri.getScheme();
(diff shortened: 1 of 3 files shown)


hazelcast/src/test/resources/test-hazelcast-jcache-without-instanceName.xml
```diff
@@ -0,0 +1,25 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<hazelcast xsi:schemaLocation="http://www.hazelcast.com/schema/config hazelcast-config-3.6.xsd"
+           xmlns="http://www.hazelcast.com/schema/config"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+    <cache name="xmlConfiguredCacheWithInstanceName">
```
emrahkocaman: I think cache name `xmlConfiguredCacheWithoutInstanceName` will fit better.

hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
```diff
@@ -102,7 +102,6 @@ private HazelcastInstance getOrCreateInstance(ClassLoader classLoader, Propertie
         }
 
         // If config location is specified, get instance with its name.
```
emrahkocaman: Should be "If instanceName is specified"

hazelcast/src/test/resources/test-hazelcast-jcache-without-instanceName.xml
serkan-ozal: right. copy-paste issue :smile:

hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
serkan-ozal: :+1:

emrahkocaman:
:+1: (minor comments)

PR #7576, opened 2016-02-22
CheckStyle improvements

Donnerbart:
* Updated CheckStyle, Findbugs, Failsafe and Cobertura to latest versions.
* Removed duplicate InnerAssignment CheckStyle rule (already defined in line 231).
* Removed unnecessary whitespaces around brackets in suppress warnings annotations.

I've updated CheckStyle and Findbugs to the same versions we're using for a long time in Simulator (since May 2015, so they are battle proven). Failsafe and Cobertura are just used for the profile `test-coverage-Local`, so there is no implication on the CloudBees environment.

With this change we finally get rid of a false positive error regarding missing whitespaces in `@SuppressWarnings` and `@SuppressFBWarnings` annotations. This was very annoying when using a code formatter, which always removed those whitespaces and lead to a CheckStyle error in the PR builder.

I also checked FindBugs with a local run and it still compiled without errors (there are some new checks in 3.0.1, so that was important).

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -129,7 +129,9 @@
 
         <!-- Checks for imports                              -->
         <!-- See http://checkstyle.sf.net/config_import.html -->
-        <module name="AvoidStarImport"/>
+        <module name="AvoidStarImport">
+            <property name="excludes" value="java.awt,javax.swing"/>
+        </module>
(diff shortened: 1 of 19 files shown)


jerrinot:
:+1:

ahmetmircik:
:+1:

PR #7578, opened 2016-02-22
Support parameterized values for native memory size

emrahkocaman:


hazelcast-spring/src/main/resources/hazelcast-spring-3.7.xsd
```diff
@@ -2420,7 +2420,7 @@
         <xs:attribute name="enabled" default="true" type="parameterized-boolean"/>
     </xs:complexType>
     <xs:complexType name="memory-size">
-        <xs:attribute name="value" type="xs:int" default="128"/>
+        <xs:attribute name="value" type="intMin0" default="128"/>
```
serkan-ozal: I think, `intMin1` (positive integer) might be better instead of `intMin0` (non negative integer)

serkan-ozal:
just minor comment :+1:

hazelcast-spring/src/main/resources/hazelcast-spring-3.7.xsd
emrahkocaman: I see but `MemorySize` allows 0. Please see https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/memory/MemorySize.java#L39

PR #7581, opened 2016-02-23
Support HazelcastInstanceAware for QuorumFunction

### LINKED ISSUE #7499: Support HazelcastInstanceAware for QuorumFunction
If a `QuorumFunction` implements `HazelcastInstanceAware`, it should have instance injected.

jerrinot:
Fixes #7499 

This is a small fix for this very issue to be applied on the maintenance branch only. For master I'm working on a new abstraction to fix this on multiple places.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/quorum/impl/QuorumImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/quorum/impl/QuorumImpl.java
@@ -17,6 +17,8 @@
 package com.hazelcast.quorum.impl;
 
 import com.hazelcast.config.QuorumConfig;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.HazelcastInstanceAware;
 import com.hazelcast.core.Member;
 import com.hazelcast.nio.ClassLoaderUtil;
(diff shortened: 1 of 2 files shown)


eminn:
:+1:

pveentjer:
:+1:

nilskp:
@jerrinot, @eminn, @pveentjer 
I just came across this fix and it's incomplete. It only works for class name based quorum functions, not for provided instances.

jerrinot:
@nilskp: this is expected. When you create an instance then it's your responsibility to set all dependencies. When Hazelcast creates an instance then it will set dependencies. It's the same on other places.

gurbuzali:
@jerrinot @nilskp  `When you create an instance then it's your responsibility to set all dependencies` this is simply wrong, one should set to `QuorumConfig` object and quorumFunction before creating the instance therefore s/he cannot handle dependencies properly.
This logic applies to other places too, please take a look at listeners for example.

jerrinot:
@gurbuzali it makes sense. I was wrong. Thanks!

nilskp:
Exactly. `HazelcastInstance` is the _one_ dependency I cannot set at construction time.

PR #7585, opened 2016-02-23
Bits refactoring

toktarev:
1) Now all bits - classes redirects call to the DirectMemoryBits
2) ByteBufferMemoryAccessor has been implemented


hazelcast/src/main/java/com/hazelcast/internal/memory/impl/ByteBufferMemoryAccessor.java
```diff
@@ -0,0 +1,396 @@
...
+public class ByteBufferMemoryAccessor implements MemoryAccessor {
+    private static final boolean BIG_ENDIAN = ByteOrder.nativeOrder() == ByteOrder.BIG_ENDIAN;
+
+    private void validateObject(Object srcObj) {
+        Preconditions.checkNotNull(srcObj);
+        Preconditions.checkTrue(srcObj instanceof byte[], "object should be byte-array");
+    }
+
+    private void checkOffset(long offset) {
+        if (offset < Integer.MIN_VALUE || offset > Integer.MAX_VALUE) {
```
serkan-ozal: can `offset` be negative? shouldn't `offset < Integer.MIN_VALUE` be `offset < 0`?

toktarev: yes good remark

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
```diff
@@ -393,53 +553,79 @@ public static char readCharVolatile(long address, boolean bigEndian) {
     }
 
     public static char readCharVolatile(byte[] buffer, int pos, boolean bigEndian) {
-        return readCharVolatile(buffer, (long) (BYTE_ARRAY_BASE_OFFSET + pos), bigEndian);
+        return readCharVolatile(BYTE_BUFFER_MEMORY_ACCESSOR, buffer, pos, bigEndian);
```
serkan-ozal: Currently (without this PR), when `Unsafe` is enabled, we can read from `byte[]` as volatile by using `readXYZVolatile(byte[] buffer, int pos, boolean bigEndian)` signature by passing our `byte[]` object. 

But with this PR, since `BYTE_BUFFER_MEMORY_ACCESSOR` doesn't support volatile read, we must cast our `byte[]` instance to `Object` for calling `readXYZVolatile(Object base, long offset, boolean bigEndian)` explicitly to enforce using `UNSAFE_MEMORY_ACCESSOR`.

Same comment is also for volatile writes.

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
```diff
@@ -37,7 +41,7 @@ public static char readChar(long address, boolean bigEndian) {
     }
 
     public static char readChar(byte[] buffer, int pos, boolean bigEndian) {
-        return readChar(buffer, (long) (BYTE_ARRAY_BASE_OFFSET + pos), bigEndian);
+        return readChar(BYTE_BUFFER_MEMORY_ACCESSOR, buffer, pos, bigEndian);
```
serkan-ozal: I think, `Bits` itself should pass `BYTE_BUFFER_MEMORY_ACCESSOR` to `DirectMemoryBits` by calling `char readChar(MemoryAccessor memoryAccessor, Object base, long offset, boolean bigEndian)` and by default `DirectMemoryBits` should use `UNSAFE_MEMORY_ACCESSOR` even though passed object is `byte[]`

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
```diff
@@ -16,16 +16,20 @@
...
+import java.io.IOException;
+import java.io.UTFDataFormatException;
 
 /**
  * Utility class to read/write bits to given location (by base object/offset or native memory address)
  * by specified byte order (little/big endian).
  */
 public final class DirectMemoryBits {
+    private static final MemoryAccessor BYTE_BUFFER_MEMORY_ACCESSOR = MemoryAccessor.HEAP_BYTE_BUFFER_MEM;
+    private static final MemoryAccessor UNSAFE_MEMORY_ACCESSOR = MemoryAccessor.AMEM;
```
serkan-ozal: We are trying to clean all `Unsafe` terms (names, classes, etc ...) in our code base and it is still in progress. So please don't use `Unsafe`. You may use `ALIGNED_MEMORY_ACCESSOR` here instead of `UNSAFE_MEMORY_ACCESSOR`

hazelcast/src/main/java/com/hazelcast/nio/Bits.java
```diff
@@ -69,265 +70,124 @@
...
-     * */
+     */
     public static final Charset ISO_8859_1 = Charset.forName("ISO-8859-1");
 
     private Bits() {
     }
 
     public static char readChar(byte[] buffer, int pos, boolean bigEndian) {
-        return bigEndian ? readCharB(buffer, pos) : readCharL(buffer, pos);
+        return DirectMemoryBits.readChar(buffer, pos, bigEndian);
```
serkan-ozal: As I commented above, `Bits` should use `ByteBufferMemoryAccessor` directly here by calling `DirectMemoryBits .readChar(BYTE_BUFFER_MEMORY_ACCESSOR, buffer, pos, bigEndian)` and `DirectMemoryBits` should use `AlignedMemoryAccessor` by default.

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
toktarev: Why ?
    public static char readCharVolatileB(Object base, long offset) {
        int byte1 = UNSAFE.getByte(base, offset) & 0xFF;
        int byte0 = UNSAFE.getByte(base, offset + 1) & 0xFF;
        return (char) ((byte1 << 8) + byte0);
    }

    public static char readCharVolatileL(Object base, long offset) {
        int byte1 = UNSAFE.getByte(base, offset) & 0xFF;
        int byte0 = UNSAFE.getByte(base, offset + 1) & 0xFF;
        return (char) ((byte0 << 8) + byte1);
    }
I don't see here usage of some readXYZVolatile in current codeBase

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
toktarev: Why do we need to use some unsafe stuff if byte[] is passed explicitly

From my point of view it is ineffective.

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
toktarev: Ok

hazelcast/src/main/java/com/hazelcast/nio/Bits.java
toktarev: Commented previous remark

mdogan:
I did not understand the use case of `ByteBufferMemoryAccessor`.

toktarev:
It is used to read from passed byte[] directly

mdogan:
I see what it does. But what's the use-case? Why & where do you need a memory accessor to read from a plain `byte[]`?

toktarev:
If we have old Bits class it uses direct access
I used BBMemoryAccessor to generalize this approach and get rid of copy past

hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessor.java
```diff
@@ -61,6 +62,12 @@
      */
     MemoryAccessor AMEM = MemoryAccessorProvider.getMemoryAccessor(STANDARD);
 
+
+    /**
+     * The instance of {@link MemoryAccessor} which  handles data in specified heap-byteBuffer (byte[] buffer).
+     */
+    MemoryAccessor HEAP_BYTE_BUFFER_MEM = MemoryAccessorProvider.getMemoryAccessor(HEAP_BYTE_BUFFER);
```
mtopolnik: This file has a huge diff, but this seems to be the only semantic change. Could you make the diff more focused on the actual change? I don't see the value in any of the other changes.

toktarev: Idea changed spaces, it is bad i know

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
mtopolnik: `MemoryAccessor` calls into `DirectMemoryBits`, which calls back into the same `MemoryAccessor`. Am I getting this right?

toktarev: there is no cycles :)
but yes

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/ByteBufferMemoryAccessor.java
```diff
@@ -0,0 +1,400 @@
...
+        }
+
+        if (offset < Integer.MIN_VALUE || offset > Integer.MAX_VALUE) {
+            throw new IllegalArgumentException(offset + " cannot be cast to int without changing its value.");
+        }
+    }
+
+    @Override
+    public long objectFieldOffset(Field field) {
+        throw new IllegalStateException("Not supported");
```
mtopolnik: Why `IllegalState`? Should be `UnsupportedOperation`.

toktarev: ok

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/ByteBufferMemoryAccessor.java
```diff
@@ -0,0 +1,400 @@
...
+public class ByteBufferMemoryAccessor implements MemoryAccessor {
+    private static final boolean BIG_ENDIAN = ByteOrder.nativeOrder() == ByteOrder.BIG_ENDIAN;
+
+    private void validateObject(Object srcObj) {
+        Preconditions.checkNotNull(srcObj);
+        Preconditions.checkTrue(srcObj instanceof byte[], "object should be byte-array");
+    }
+
+    private void checkOffset(long offset) {
+        if (offset < 0) {
```
mtopolnik: We could consider `assert` instead of always-on checks.

toktarev: ok

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/DirectMemoryBits.java
```diff
@@ -263,77 +371,103 @@ public static long readLong(long address, boolean bigEndian) {
...
+        long byte4 = (long) (memoryAccessor.getByte(base, offset + 3) & 0xFF) << 32;
+        long byte3 = (long) (memoryAccessor.getByte(base, offset + 4) & 0xFF) << 24;
+        long byte2 = (long) (memoryAccessor.getByte(base, offset + 5) & 0xFF) << 16;
+        long byte1 = (long) (memoryAccessor.getByte(base, offset + 6) & 0xFF) << 8;
+        long byte0 = (long) (memoryAccessor.getByte(base, offset + 7) & 0xFF);
+        return byte7 + byte6 + byte5 + byte4 + byte3 + byte2 + byte1 + byte0;
+    }
+
+    public static long readLongL(MemoryAccessor memoryAccessor, Object base, long offset) {
+        long byte7 = (long) (memoryAccessor.getByte(base, offset) & 0xFF);
```
mtopolnik: As a consequence of the current design, this method will perform eight validation checks, one for each invocation of `getByte`. Maybe we should rethink this.

mtopolnik:
:+1:

serkan-ozal:
:+1:

PR #7598, opened 2016-02-24
Fixed issues found by new FindBugs version.

Donnerbart:
Fixed FindBugs issues found by a new FindBugs version. I thought I had the OS version properly tested before making my PR, something must have been wrong, since I got a successful build, sorry :(

Nevertheless, here is a PR to fix and/or suppress those discovered issues. Some may need a closer look to decide if they are a false positive or not.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
@@ -62,7 +62,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.Future;
 
 import static com.hazelcast.cache.impl.CacheProxyUtil.validateNotNull;
 
@@ -395,7 +394,6 @@ public class ClientCacheProxy<K, V>
                                                          boolean isRegister) {
         final Collection<Member> members = clientContext.getClusterService().getMemberList();
         final HazelcastClientInstanceImpl client = (HazelcastClientInstanceImpl) clientContext.getHazelcastInstance();
-        final Collection<Future> futures = new ArrayList<Future>();
         for (Member member : members) {
             try {
                 final Address address = member.getAddress();
@@ -403,8 +401,7 @@ public class ClientCacheProxy<K, V>
                 final ClientMessage request =
                         CacheListenerRegistrationCodec.encodeRequest(nameWithPrefix, configData, isRegister, address);
                 final ClientInvocation invocation = new ClientInvocation(client, request, address);
-                final Future future = invocation.invoke();
-                futures.add(future);
+                invocation.invoke();
             } catch (Exception e) {
                 ExceptionUtil.sneakyThrow(e);
             }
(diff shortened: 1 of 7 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
```diff
@@ -403,8 +401,7 @@ protected void updateCacheListenerConfigOnOtherNodes(CacheEntryListenerConfigura
                 final ClientMessage request =
                         CacheListenerRegistrationCodec.encodeRequest(nameWithPrefix, configData, isRegister, address);
                 final ClientInvocation invocation = new ClientInvocation(client, request, address);
-                final Future future = invocation.invoke();
-                futures.add(future);
+                invocation.invoke();
```
Donnerbart: This actually needs a closer look. The invocation futures were collected, but never used. Is this a bug?

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
```diff
@@ -792,6 +793,7 @@ protected void deleteCacheEntry(Data key) {
         }
     }
 
+    @SuppressFBWarnings("WMI_WRONG_MAP_ITERATOR")
```
Donnerbart: FindBugs complained about `keySet()` used ineffectively. But the usage is quite complex and this should be a false positive. At least I will not optimize this in this PR.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -665,6 +665,7 @@ public MemberImpl getMember(String uuid) {
         return null;
     }
 
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT")
```
Donnerbart: `values()`, `keySet()` and `values()` are used to fill a cache. Should be fine to ignore the return values.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
Donnerbart: Okay, there was code after this, which used the futures:
```
        //make sure all configs are created
        //TODO do we need this ???s
        //        try {
        //            FutureUtil.waitWithDeadline(futures, CacheProxyUtil.AWAIT_COMPLETION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
        //        } catch (TimeoutException e) {
        //            logger.warning(e);
        //        }
```
This commented out version was removed on 2015/10/05 by @serkan-ozal in commit 90a1a4f5d0d93b3c367e81f6f83e599e3983ce11
> Client side cache statistics implementation

So I assume it's safe to delete those unused futures, but could you please confirm this, Serkan?

jerrinot:
:+1:

hasancelik:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
serkan-ozal: @Donnerbart yes, it is not used at the moment and I think it is safe to delete because it is about updating cache config not the actual listener registration which is sync. So I think, no need to wait update cache configs as sync, async should be OK.

Donnerbart: Thanks a lot for the confirmation!

PR #7602, opened 2016-02-25
MemoryStats refactoring

toktarev:
1) MemoryStats has been decoupled from native memory


mdogan:
Why? `MemoryStats` shows stats for `heap` and `native` memory regions? If you remove `native` prefix, what will `getMaxMemory()` mean?

PR #7607, opened 2016-02-25
Remove defaulting to the CLUSTER protocol…

eminn:
since the protocols are completely different, defaulting to CLUSTER protocol won't solve anything. SpinningSocketWriter was suffering from this issue when it is used with client connections.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketWriter.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketWriter.java
@@ -18,6 +18,7 @@ package com.hazelcast.nio.tcp.nonblocking;
 
 import com.hazelcast.internal.metrics.MetricsRegistry;
 import com.hazelcast.internal.metrics.Probe;
+import com.hazelcast.internal.util.counters.SwCounter;
 import com.hazelcast.nio.IOUtil;
 import com.hazelcast.nio.OutboundFrame;
 import com.hazelcast.nio.Packet;
@@ -26,8 +27,6 @@ import com.hazelcast.nio.tcp.NewClientWriteHandler;
 import com.hazelcast.nio.tcp.SocketWriter;
 import com.hazelcast.nio.tcp.TcpIpConnection;
 import com.hazelcast.nio.tcp.WriteHandler;
-import com.hazelcast.internal.util.counters.SwCounter;
-
 import java.io.IOException;
 import java.net.SocketException;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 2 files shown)


mdogan:
:+1:

gurbuzali:
will `not defaulting to CLUSTER protocol` do any harm ?

eminn:
In case of write handler is not initialized(that means no protocol information received from the socket), you just cannot write to that socket.

Defaulting to any protocol and letting it to write that particular protocol is not a solution to the problem, but makes the things worse since it could happen that we'll just receive protocol information and continue to write data in correct protocol. But when defaulted to some protocol, we cannot change the protocol after it is set.

pveentjer:
The more exceptional logic (in all but one case) is removed from the fast path, the better.

However I believe this is needed when there a duplicate connection and the system doesn't receive the correct information from the other side since the other side has provided the correct info.

So I don't know if this code improvement, doesn't cause problems.

@mdogan @jerrinot wdyt 

We should allocate to take care of the handshaking of connections so the duplicate connection doesn't happen

eminn:
The aforementioned scenario can happen immediately with the Spinning I/O model regardless of duplicate connections. Even if it was there for duplicate connections, if it's a client connection, you are forcing to use `CLUSTER` protocol, which is a problem.

pveentjer:
Duplicate connections can't happen with clients since it is always the client that contacts the server. So in case of a client, the protocol would already have been set.

Don't get me wrong; I would love to get rid of this extra branch in the code.

pveentjer:
Let me try it again. Forget about the spinning one for the time being; Afaik this default is needed to deal when 2 nodes connect to each other at the same time (e.g. startup) and you end up with 2 connections between members. It can happen that one connection takes care of one side of the traffic and the other connection for the reverse side of the traffic. If this default wilt not be added, one of these connections could end up without the protocol being set because the info is send over the other connection.

eminn:
I see, the first job of the socket writer is to process the protocol information provided by the socket reader.

Do you have a clear picture of the duplicate connections and how this code could solve any problem on that scenario ?

pveentjer:
This code solves the problem that the other connection was used to publish it was a member, but this isn't used, because due to the duplicate connection, the other connection is being used for reading. So this duplicated channel, doesn't contain the protocol information since it was send (and not used) on the other connection.

mdogan:
As far as I remember, the very first thing connecting side is to write protocol bytes to the socket. So, when accepting side reads from socket, it will be reading the protocol as the first thing.  

There are joiner stress tests which will trigger this duplicate connection case. 
@eminn: you can run and see if this change causes any trouble.

pveentjer:
My suggestion would be to create a test that at the same moment tries to launch many jvm's in parallel. This will trigger the duplicate connection behavior. And put this creation logic in a loop because duplicate connections don't need to happen every time. But after a few hundred times, duplicate connections will be there.

pveentjer:
What are we going to do with this one? @eminn

PR #7608, opened 2016-02-25
[BACKPORT] Remove defaulting to the CLUSTER protocol …

eminn:
since the protocols are completely different, defualting to CLUSTER protocol won't solve anything. SpinningSocketWriter was suffering from this issue when it is used with client connections.


eminn:
verify

PR #7609, opened 2016-02-25
Skip unaligned memory access tests if platform doesn't support unaligned memory access

serkan-ozal:
With this PR, all our OSS tests (core, client, spring) pass at SPARC.

EE is in progress ...


jerrinot:
:+1:

mdogan:
:+1:

PR #7610, opened 2016-02-25
MemoryStats refactoring

toktarev:
https://hazelcast.atlassian.net/wiki/display/EN/Proposed+cleanup+of+the+MemoryManager+abstraction

OSS specific aspects of new MemoryManager design have been implemented:

1) Old MemoryStats interface has become JVMMemoryStats.
2) MemoryAccessor has been split into MemoryAccessor and MemoryAccessStrategy.
3) Corresponding refactoring in the related classes has been done


pveentjer:
2 questions:

1: you know that you can access this information using the metrics system? The same probes are available

2: we need to be careful here if something is public API or not. There is some public stuff in here like MemoryUnit. I don't know if MemoryStats is public.

toktarev:
2) I don't know it as well

Can we clarify it ?

checkstyle/suppressions.xml
```diff
@@ -30,7 +30,8 @@
               files="SerializerHook\.java$"/>
 
     <!-- ConsoleApp -->
-    <suppress checks="MethodCountCheck|FileLengthCheck|ClassFanOutComplexityCheck" files="com/hazelcast/console/ConsoleApp"/>
+    <suppress checks="MethodCountCheck|FileLengthCheck|ClassFanOutComplexityCheck"
```
mtopolnik: We should not be adding more suppressions to `suppressions.xml`. Could you please move these to the Java source code file, as a class annotation `@SuppressWarnings("checkstyle:methodcount")` etc.? Except for `FileLength`, `IllegalImport` and others which are beyond the class scope.

jerrinot:
I am confused. The description of this PR says: "Old MemoryStats interface has become JVMMemoryStats." but I can see quite big changes, not just the rename. @toktarev: Can you clarify your goal?

toktarev:
@jerrinot: I will add detailed PR explanations

toktarev:
Added detailed PR description

mtopolnik:
The fact that you'

hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessStrategy.java
```diff
@@ -0,0 +1,576 @@
...
+ * @param <R> - type of resource
+ */
+public interface MemoryAccessStrategy<R> extends ByteMemoryAccessStrategy<R> {
+    MemoryAccessStrategy<Object> STANDARD_MEM = MemoryAccessorStrategyProvider.getMemoryAccessStrategy(MemoryAccessStrategyType.STANDARD);
+
+    MemoryAccessStrategy<Object> STANDARD_ALIGNED_MEM = MemoryAccessorStrategyProvider.getMemoryAccessStrategy(MemoryAccessStrategyType.STANDARD_ALIGNMENT);
+
+    MemoryAccessStrategy<byte[]> HEAP_BYTE_ARRAY_MEM = MemoryAccessorStrategyProvider.getMemoryAccessStrategy(MemoryAccessStrategyType.HEAP_BYTE_ARRAY);
+
+    MemoryAccessStrategy<Object> MEM = MemoryAccessorStrategyProvider.getMemoryAccessStrategy(MemoryAccessStrategyType.PLATFORM_AWARE);
```
mtopolnik: We must improve these constant names. `MEM` is what `MEM` was before, i think. It should be an always-safe-to-use accessor of native CPU address space. `AMEM` is gone; it was meant to be like `MEM`, but only safe to use when the caller guarantees aligned memory access. I wonder whether we need at all the `STANDARD_ALIGNED_MEM` constant and the name is confusing (not clear whether the caller must ensure aligned access, or the accessor does it).

hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessStrategy.java
```diff
@@ -0,0 +1,576 @@
...
+
+    /**
+     * Index scale of double[]
+     */
+    int ARRAY_DOUBLE_INDEX_SCALE = STANDARD_MEM != null ? STANDARD_MEM.arrayIndexScale(double[].class) : -1;
+
+    /**
+     * Index scale of any type of Object[]
+     */
+    int ARRAY_OBJECT_INDEX_SCALE = STANDARD_MEM != null ? STANDARD_MEM.arrayIndexScale(Object[].class) : -1;
```
mtopolnik: All these constants and the `...offset` methods below don't belong to the general memory access strategy. These would have to be moved to an interface specialized for native CPU memory access. This can be done in a further PR.

hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessStrategyType.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package com.hazelcast.internal.memory;
+
+/**
+ * Provides different types to access data from external resources (objects , byte arrays)
+ */
+public enum MemoryAccessStrategyType {
+    /**
+     * Represents the UNSAFE-based {@link MemoryAccessor}.
```
mtopolnik: These now represent `MemoryAccessStrategy`s, not `MemoryAccessor`s.

hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessStrategy.java
mtopolnik: Retracting the above comment... I thought `MemoryAccessStrategy` would replace the current usages of `MEM` and `AMEM`, but these will remain `MemoryAccesor`s. I guess then the name `MEM` should not be reused.

hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessStrategyType.java
toktarev: MemoryAccessStrategyType represents MemoryAccessStrategy
MemoryAccessorType represents MemoryAccessor

PR #7612, opened 2016-02-25
Renamed Packet.header to Packet.flags

pveentjer:
The problem with header is that it isn't the full header of the Packet; there
are more things in there like size, version etc.

I renamed header to flags, similar as in TCP


pveentjer:
No Enterprise PR needed.

jerrinot:
:+1:

mtopolnik:
:+1:

PR #7613, opened 2016-02-25
Improve Javadoc of MemoryAccessor

mtopolnik:
Better explain the range of use cases of MemoryAccessor and the meaning of its associated address space.


pveentjer:
:+1: 

Since this is just javadoc update; I'll merge this one.

PR #7615, opened 2016-02-26
Partially revert #7377

sancar:
Heartbeat fix in #7377 removed a workaround for another bug.

This bug is a race condition when a connection is closed.
It is possible to not be able to set exception to invocations that are
send right before connection closed.

Although it is not the best way, to not to take a new risk,
we are reverting the change that works as workaround for  this bug.


gurbuzali:
:+1:

jerrinot:
:+1:

PR #7618, opened 2016-02-28
Moved a test class for jmx to internal

pveentjer:
Since the jmx package is moved to internal, this class should be there as well.


tombujok:
:+1:

pveentjer:
Thanks. Since this is just moving a test class, I'll merge this one

PR #7619, opened 2016-02-29
Add metaspace stats to HealthMonitor

mtopolnik:
The optimal configuration of native memory metadata percentage is difficult to determine without the support of runtime statistics. This PR therefore adds the key metrics to the output of Health Monitor:

1. used metadata memory;
2. free metadata memory;
3. current metadata percentage (used metadata memory / used "main" native memory).


pveentjer:
:+1:

PR #7623, opened 2016-02-29
Added probes to let handlers show their thread.id

pveentjer:
Each handler is run on an io thread and with these probes in place,
in the metrics one can see which iothread a handler is using. This
is very useful to figure out of there are any imbalances.


mtopolnik:
:+1:

jerrinot:
:+1:

pveentjer:
run-lab-run

PR #7625, opened 2016-02-29
Fixed broken replication logic in putAll operation

### LINKED ISSUE #7617: ReplicatedMap updates in 3.6 takes very long time
in version 3.6 there is a massive ,IMO, regression in the ReplicatedMap updates.
sometimes the update loses some of the entries to update.
here is a test i prepared that demonstrate the problem :  

hazelcast.xml : 
```
<hazelcast xsi:schemaLocation="http://www.hazelcast.com/schema/config https://hazelcast.com/schema/config/hazelcast-config-3.6.xsd"
           xmlns="http://www.hazelcast.com/schema/config"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <group>
        <name>mygroup</name>
        <password>admin</password>
    </group>
    <management-center enabled="false">http://localhost:8080/mancenter</management-center>
    <instance-name>BOTH_LAYERS</instance-name>
    <network>
        <port auto-increment="true">5711</port>
        <join>
            <multicast enabled="true">
                <multicast-group>224.2.2.3</multicast-group>
                <multicast-port>54327</multicast-port>
                <multicast-timeout-seconds>4</multicast-timeout-seconds>
            </multicast>
            <tcp-ip enabled="false"/>
            <aws enabled="false"/>
        </join>
    </network>
    <replicatedmap name="configurationSetCache">
        <statistics-enabled>true</statistics-enabled>
    </replicatedmap>
</hazelcast>

```

ReplicatedMapIssue.java :  
```
package test;

import java.io.BufferedReader;
import java.io.IOException;
import java.io.InputStreamReader;
import java.io.Serializable;
import java.util.HashMap;
import java.util.Map;

public class ReplicatedMapIssue implements Serializable{

    public static final int NEW_ENTRIES_COUNT = 7;

    private Map<Integer,DummyObject> replicatedMap;

    private BufferedReader stdin;


    public void test() {

        replicatedMap = hz.getReplicatedMap("configurationSetCache");


        try {
            boolean exit = false;
            stdin = new BufferedReader(new InputStreamReader(System.in));
            do {
                System.out.print(
                        "1.replicated map test\n" +
                        "0.exit\n");
                System.out.print("choose: ");
                String line = stdin.readLine();

                switch(Integer.parseInt(line)) {
                    case 1: {
                        runReplicatedMapTest();
                        break;
                    }
                    case 0: {
                        exit = true;
                        break;
                    }
                }
            } while (!exit) ;
            hz.shutdown();
        } catch (IOException e) {
            e.printStackTrace();
        }

    }

    private void runReplicatedMapTest() {
        Map<Integer,DummyObject> map = new HashMap<>();
        for(int i=0;i < NEW_ENTRIES_COUNT;i++){
            map.put((int) (Math.random() * 100), new DummyObject());
        }
        replicatedMap.clear();
        replicatedMap.putAll(map);
//        for(Map.Entry<Integer,DummyObject> e:map.entrySet()) replicatedMap.put(e.getKey(),e.getValue());

        long start = System.currentTimeMillis();
        do {
            try {
                Thread.sleep(1000);
            } catch (InterruptedException ignored) {}
        }while(replicatedMap.size() != NEW_ENTRIES_COUNT);
        long finish = System.currentTimeMillis();
        System.out.format("replicated map (size %d), took %d ms and contains the following values: %s \n", replicatedMap.size(), (finish - start), replicatedMap.values());
    }


        ReplicatedMapIssue issue = new ReplicatedMapIssue();
        issue.test();
        System.out.println("hazelcast stopped");
        System.exit(0);
    }
}

```

DummyObject.java :  

```
package test;

import java.io.Serializable;
import java.util.UUID;

public class DummyObject implements Serializable {
    private String a;

    public DummyObject() {
        a = UUID.randomUUID().toString();
    }

    @Override
    public String toString() {
        return a.substring(0,8);
    }
}

```
-------------------------------------------------------------------

using java 8 and hazelcast 3.6 

reproduce steps:
1) run the main
2) on the menu click 1 few times one by one
--> after some times either the update will take more than a minute or the program will simply freeze (its because the map contains only 6 entries and the seventh entry will never get updated.

in 3.5.5, the scenario is much better, but around 20% of the times, also the seventh entry is lost.

eminn:
Fixes #7617

The `PutAllOperation` is issued for `putAll` calls to the replicated map on the keys' partition owners. After value is put to the record store, update should be replicated to all members excluding the partition owner node itself. But the buggy implementation was also excluding the caller member which is the source of the bug that blocks the replication of the update. When the replication of update is blocked, it might take 30(default period of anti-entropy check) seconds for anti-entropy system to kick in and sync the values of replicas with partition owner.

In this PR the exclusion of caller members are removed.

Thanks @sabag for reporting the issue and providing a reproducer.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/replicatedmap/ReplicatedMapReadYourWritesTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/replicatedmap/ReplicatedMapReadYourWritesTest.java
@@ -18,10 +18,12 @@ package com.hazelcast.replicatedmap;
 
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.ReplicatedMap;
+import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.TestHazelcastInstanceFactory;
 import com.hazelcast.test.annotation.ParallelTest;
 import com.hazelcast.test.annotation.QuickTest;
+import java.util.HashMap;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
@@ -33,6 +35,31 @@ import static org.junit.Assert.assertTrue;
 @RunWith(HazelcastParallelClassRunner.class)
 @Category(value = {QuickTest.class, ParallelTest.class})
 public class ReplicatedMapReadYourWritesTest extends ReplicatedMapBaseTest {
+
+    @Test
+    public void testReadYourWritesBySize() throws Exception {
+        TestHazelcastInstanceFactory nodeFactory = createHazelcastInstanceFactory();
+        HazelcastInstance instance1 = nodeFactory.newHazelcastInstance();
+        HazelcastInstance instance2 = nodeFactory.newHazelcastInstance();
+        final ReplicatedMap<Integer, Integer> map1 = instance1.getReplicatedMap("default");
+        final ReplicatedMap<Integer, Integer> map2 = instance2.getReplicatedMap("default");
+
+
+        HashMap<Integer, Integer> map = new HashMap<Integer, Integer>();
+        final int count = 100;
+        for (int i = 0; i < count; i++) {
+            map.put(i, i);
+        }
+        map1.putAll(map);
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                assertEquals(count, map1.size());
+                assertEquals(count, map2.size());
+            }
+        }, 15);
+    }
 
     @Test
     public void testReadYourWritesByGet() throws Exception {
--- a/hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/PutAllOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/PutAllOperation.java
@@ -81,7 +81,7 @@ public class PutAllOperation extends AbstractOperation {
         Collection<Member> members = getNodeEngine().getClusterService().getMembers();
         for (Member member : members) {
             Address address = member.getAddress();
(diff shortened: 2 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/replicatedmap/ReplicatedMapReadYourWritesTest.java
```diff
@@ -35,6 +37,31 @@
...
+        }
+        map1.putAll(map);
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                assertEquals(count, map1.size());
+                assertEquals(count, map2.size());
+            }
+        }, 5);
```
jerrinot: this is a rather aggressive timeout. 

I understand you want to rule out false positives caused by anti-entropy system, but this smells like a lengthy GC could cause spurious failures. 15s or so seems better. it's an upper bound anyway, so a passing test will not slower because of this.

eminn: Yes we can relax it to 15 seconds.

jerrinot: another idea: I'd be nice to have a comment why the timeout shouldn't be too long. 
as otherwise it's too tempting to increase it to 30+ seconds to "fix" a failing test.

jerrinot:
:+1:

ahmetmircik:
:+1:

sabag:
Hello, 
Was this fix included in 3.6.2 ? because I don't find it in the release notes.

jerrinot:
@sabag: The issue was fixed as we backported the fix. See: https://github.com/hazelcast/hazelcast/pull/7626

Issue #7527, opened 2016-02-16
CacheClientListenerTest.testSyncListener_shouldNotHang_whenHazelcastInstanceShutdown


hasancelik:
```
java.lang.AssertionError: Cache operations should not hang when sync listener is present!, failed to complete within 120 seconds , count left: 1
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:761)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:747)
	at com.hazelcast.cache.CacheListenerTest.testSyncListener_shouldNotHang_AfterAction(CacheListenerTest.java:331)
	at com.hazelcast.cache.CacheListenerTest.testSyncListener_shouldNotHang_whenHazelcastInstanceShutdown(CacheListenerTest.java:263)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:622)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:88)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:82)
	at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:334)
	at java.util.concurrent.FutureTask.run(FutureTask.java:166)
	at java.lang.Thread.run(Thread.java:701)
```

[...]

Issue #7285, opened 2016-01-11
NearCacheTest.testGetAll[batchInvalidationEnabled:true]


hasancelik:
```
java.lang.AssertionError: Near cache hits should be > 400 but were 371
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.map.nearcache.NearCacheTest.testGetAll(NearCacheTest.java:437)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/784/testReport/junit/com.hazelcast.map.nearcache/NearCacheTest/testGetAll_batchInvalidationEnabled_true_/

hasancelik:
```
java.lang.AssertionError: Near cache hits should be 481 but were 400 expected:<481> but was:<400>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.map.nearcache.NearCacheTest.testGetAll(NearCacheTest.java:450)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-IbmJDK1.7/com.hazelcast$hazelcast/833/testReport/junit/com.hazelcast.map.nearcache/NearCacheTest/testGetAll_batchInvalidationEnabled_true_/

hasancelik:
```
java.lang.AssertionError: Near cache hits should be 491 but were 428 expected:<491> but was:<428>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.map.nearcache.NearCacheTest.testGetAll(NearCacheTest.java:450)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.maintenance-OracleJDK1.6/com.hazelcast$hazelcast/355/testReport/junit/com.hazelcast.map.nearcache/NearCacheTest/testGetAll_batchInvalidationEnabled_true_/

hasancelik:
```
java.lang.AssertionError: Near cache hits should be 498 but were 440 expected:<498> but was:<440>
[...]

PR #7629, opened 2016-02-29
[DONT-MERGE] Fixes a race when client connection is closed.

### LINKED ISSUE #7582: ClientRegressionWithMockNetworkTest.testClusterShutdown_thenCheckOperationsNotHanging
```
java.lang.AssertionError: Put operations should not hang., failed to complete within 120 seconds , count left: 15
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:761)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:747)
	at com.hazelcast.client.ClientRegressionWithMockNetworkTest.testClusterShutdown_thenCheckOperationsNotHanging(ClientRegressionWithMockNetworkTest.java:861)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x/com.hazelcast$hazelcast-client/4596/testReport/junit/com.hazelcast.client/ClientRegressionWithMockNetworkTest/testClusterShutdown_thenCheckOperationsNotHanging/

sancar:
When connection closed, it was possible to endup with invocations
that are not notified. These invocations was causing infinite
wait. With a sync block race is fixed.

There was a workaround for this in ClientInvcationFuture.get
which is waking up every hearbeatTimeout milliseconds and checking
if connection is still there. This workaround is removed in a
previous pr because when async api is used that workaround is
does not work.

fixes #7582 , #7527.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -106,23 +106,25 @@ abstract class ClientInvocationServiceSupport implements ClientInvocationService
         if (isShutdown) {
             throw new HazelcastClientNotActiveException("Client is shut down");
         }
-        registerInvocation(invocation);
+        synchronized (connection) {
+            registerInvocation(invocation);
 
-        ClientMessage clientMessage = invocation.getClientMessage();
-        if (!isAllowedToSendRequest(connection, invocation) || !writeToConnection(connection, clientMessage)) {
-            final long callId = clientMessage.getCorrelationId();
-            ClientInvocation clientInvocation = deRegisterCallId(callId);
-            if (clientInvocation != null) {
-                callIdSequence.complete();
-                throw new IOException("Packet not send to " + connection.getRemoteEndpoint());
-            } else {
-                if (invocationLogger.isFinestEnabled()) {
-                    invocationLogger.finest("Invocation not found to deregister for call id " + callId);
+            ClientMessage clientMessage = invocation.getClientMessage();
+            if (!isAllowedToSendRequest(connection, invocation) || !writeToConnection(connection, clientMessage)) {
+                final long callId = clientMessage.getCorrelationId();
+                ClientInvocation clientInvocation = deRegisterCallId(callId);
+                if (clientInvocation != null) {
+                    callIdSequence.complete();
+                    throw new IOException("Packet not send to " + connection.getRemoteEndpoint());
+                } else {
+                    if (invocationLogger.isFinestEnabled()) {
+                        invocationLogger.finest("Invocation not found to deregister for call id " + callId);
+                    }
                 }
             }
-        }
 
-        invocation.setSendConnection(connection);
+            invocation.setSendConnection(connection);
+        }
     }
 
     private boolean writeToConnection(ClientConnection connection, ClientMessage clientMessage) {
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -167,13 +158,15 @@ private ClientInvocation deRegisterCallId(long callId) {
...
 
     public void cleanResources(ConstructorFunction<Object, Throwable> responseCtor, ClientConnection connection) {
-        final Iterator<Map.Entry<Long, ClientInvocation>> iter = callIdMap.entrySet().iterator();
-        while (iter.hasNext()) {
-            final Map.Entry<Long, ClientInvocation> entry = iter.next();
-            final ClientInvocation invocation = entry.getValue();
-            if (connection.equals(invocation.getSendConnection())) {
-                iter.remove();
-                invocation.notifyException(responseCtor.createNew(null));
+        synchronized (connection) {
```
gurbuzali: are we sure that these are the only synchronization points ? how about deregister ?

ahmetmircik:
@sancar, can this change cause any performance degradation because of the sync blocks?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: Beside sync points here, deregisterCall id is called when response come from server. And there is no connection related check there so we should not need a sync for that call as far as I see.

gurbuzali:
:+1:

sancar:
verify

sancar:
run-lab-run

serkan-ozal:
might this PR be related also with this issue https://github.com/hazelcast/hazelcast/issues/7285?

sancar:
@ahmetmircik Yes, it can. This fix discussed with 2 other developer. We could not come up with a better one. But, I am open to discussion about alternative fixes with one more persion.

sancar:
@serkan-ozal  Seems unlikely. There needs to be connection closing when an invocation from client has been made. And in  `NearCacheTest.testGetAll`, I don't see any client or server shutdown during the test.

sancar:
@gurbuzali @ihsandemir I made an alternative solution with @ahmetmircik suggestion. 
Can you guys also take a look at that. 
https://github.com/hazelcast/hazelcast/pull/7635

pveentjer:
Before this PR gets merged, a simulator test needs to be done to make sure there is no performance regression. As long as that isn't done, this PR can't be merged.

pveentjer:
If you only care about preventing notification on closed connections, why don't you periodically scan all invocations for closed connections? This is a lot cheaper and you don't polute the happy path (connection is working fine), with exceptional logic.

gurbuzali:
@pveentjer just check the other PR, he already did what you said (almost)

pveentjer:
Then what is the purpose of this pr? Having a synchronized block on a potentially contended object is not what you want to have.

gurbuzali:
@pveentjer you are not following, first he came up with this solution then the other one. and he mentioned this in above comment and put a `don't merge` thing on the title.

pveentjer:
I believe this PR can be closed since the other solution is superior.

pveentjer:
I can't follow every PR every minute.

gurbuzali:
you at least read comments on the PR before writing down something

PR #7635, opened 2016-03-01
Fixes a race when connection is closed.

### LINKED ISSUE #7582: ClientRegressionWithMockNetworkTest.testClusterShutdown_thenCheckOperationsNotHanging

sancar:
When connection closed, it was possible to endup with invocations
that are not notified. These invocations was causing infinite
wait. We let the race happen and check the callIdMap from another
thread every seconds to cleanup unlucky invocations.

There was a workaround for this in ClientInvcationFuture.get
which is waking up every hearbeatTimeout milliseconds and checking
if connection is still there. This workaround is removed in a
previous pr because when async api is used that workaround is
does not work.

fixes #7582 , #7527.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -31,27 +31,25 @@ import com.hazelcast.client.spi.impl.listener.ClientListenerServiceImpl;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.ClassLoaderUtil;
 import com.hazelcast.nio.Connection;
-import com.hazelcast.nio.ConnectionListener;
 import com.hazelcast.spi.exception.TargetDisconnectedException;
-import com.hazelcast.util.ConstructorFunction;
 
 import java.io.IOException;
+import java.util.Collection;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
 
 import static com.hazelcast.client.config.ClientProperty.MAX_CONCURRENT_INVOCATIONS;
 import static com.hazelcast.instance.OutOfMemoryErrorDispatcher.onOutOfMemory;
 
 
-abstract class ClientInvocationServiceSupport implements ClientInvocationService, ConnectionListener
-        , ConnectionHeartbeatListener {
+abstract class ClientInvocationServiceSupport implements ClientInvocationService {
 
-    private static final int WAIT_TIME_FOR_PACKETS_TO_BE_CONSUMED = 10;
     private static final int WAIT_TIME_FOR_PACKETS_TO_BE_CONSUMED_THRESHOLD = 5000;
     protected final HazelcastClientInstanceImpl client;
     protected ClientConnectionManager connectionManager;
@@ -67,7 +65,6 @@ abstract class ClientInvocationServiceSupport implements ClientInvocationService
     private ClientExceptionFactory clientExceptionFactory;
     private volatile boolean isShutdown;
 
-
     public ClientInvocationServiceSupport(HazelcastClientInstanceImpl client) {
         this.client = client;
         int maxAllowedConcurrentInvocations = client.getClientProperties().getInteger(MAX_CONCURRENT_INVOCATIONS);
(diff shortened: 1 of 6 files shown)


sancar:
What should be the value of schedule time to check later ?

gurbuzali:
it is hard to guess a good number. if it is small(<1) you(user) get fast response but the possibility of missing callId increases. if it is big(10) you get late response (have to wait for 10 seconds to retry) but the possibility of missing callId decreases.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -176,6 +178,9 @@ public void cleanResources(ConstructorFunction<Object, Throwable> responseCtor,
                 invocation.notifyException(responseCtor.createNew(null));
             }
         }
+        //Try cleaning again later to compensate the Race between `send` and `cleanResources`
```
pveentjer: Be careful with using a shared thread because you don't know who is using this pool and it could be depleted for some time and then user invocations don't get triggered in case of a closed connection.

I would introduce a dedicated thread that monitors the invocations; the same as the serverside invocation monitor.

pveentjer:
On the serverside we scan every second.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
sancar: This task is scheduled on internal executor which does not make any blocking call. Although it should not be problem,  invocation monitor approach seems cleaner to me, mostly because I don't want to choose a big number here to prevent race and make the user wait longer.

About scanning invocations every second, the data structure that we are using is ConcurrentHashMap here.
This hashmap keeps callId -> invocation and accessed in each invocation and invocation response. 
Do we think that, iterating through this to find dead connections also cause a significant drop in performance.

pveentjer: You won't notice it :)

If you do 100k ops/second.. you are generating many times more garbage. Lets say 300k objects per second. If you have 1000 pending invocations, then at most you will generate a few times X garbage (if any garbage is created at all).

We use the same approach on the server side.. scanning over all invocations see if there are problems like timeout. There is no noticeable performance impact.

pveentjer: Perhaps in the 3.7 release, we get rid of the CHM and use an array. You want to control the number of pending invocations anyway; so for regular invocations it is best to store them in an array. Scanning invocations is then as cheap as scanning an array. 

This approach is going to be used for the improved InvocationRegistry on the server; probably the same approach can be applied on the client. 

It will also immediately solve your back pressure issues.

sancar:
@pveentjer I think there is another problem with scanning invocations. We sort of already discussed this problem with @gurbuzali . 

When this thread finds an invocation that has a closed connection `, it will call `invocation.notifyException`. It could be the case that `cleanResources`  also doing the same.
This two will potentially can cause same invocation to be retried twice. One solution to this is to remove `cleanResources` completely. Is it the case for server side ?

sancar:
run-lab-run

gurbuzali:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -222,48 +167,42 @@ public boolean isShutdown() {
...
-                    break;
+
+                int pendingPacketCount = connection.getPendingPacketCount();
+                if (pendingPacketCount != 0) {
+                    long closedTime = connection.getClosedTime();
+                    long elapsed = System.currentTimeMillis() - closedTime;
+                    if (elapsed < WAIT_TIME_FOR_PACKETS_TO_BE_CONSUMED_THRESHOLD) {
+                        continue;
+                    } else {
+                        invocationLogger.warning("There are packets which are not processed " + pendingPacketCount);
```
ihsandemir: if there are 900 waiting messages for the closed connection, will this print 900 log lines? If so, why not just decrease it to one log line?

sancar: @ihsandemir I have changed the logging accordingly.

PR #7638, opened 2016-03-01
Introducing ability to disable invalidation events for per entry

### LINKED ISSUE #7580: Ability to disable per-entry notifications and enable/disable notification of full-flush events
It would be better to have a way where in per-entry near-cache notifications are disabled, but an ability to enable/disable notification of full-flush events (clear, etc ...).

CC @sertugkaya

serkan-ozal:
… However full-flush invalidation events are still enabled

Closes #7580


emrahkocaman:
:+1:

Issue #7631, opened 2016-02-29
IMap.executeOnKeys() does not support the empty set, and is inconsistent with getAll()


stevebarham:
IMap.executeOnKeys has a fairly irritating wrinkle in 3.6.1; if invoked with an empty set, it throws a misleading NullPointerException. 

I can see from the current master branch that this has been replaced with a more accurate IllegalArgumentException, indicating that an empty collection is not supported:

````java
if (keys.size() == 0) {
  throw new IllegalArgumentException(EMPTY_COLLECTION_IS_NOT_ALLOWED);
}
````

I think this is bad for the following reasons:

1. Checking for .size() == 0 may be inefficient, vs. using .isEmpty(), depending on the incoming set
2. Per many other methods, the expected result of running an entry processor over no keys is not an exception, but an empty map. I agree that passing a null set of keys here represents an error. 
3. The behaviour is inconsistent with other set-receiving methods in IMap, for example getAll():

````java
if (CollectionUtil.isEmpty(keys)) {
  return Collections.emptyMap();
}
````

Switching this to behave more like getAll() seems like a simple fix - please let me know if you agree, and I will send you a PR.

ahmetmircik:
@stevebarham makes sense

jerrinot:
Fixed by 3b4c1c5c2b7c1bc27a37d59cf7597e711f9b8487

PR #7640, opened 2016-03-01
Empty set fix for executeOnKeys() #7631

stevebarham:
Modified ClientMapProxy and MapProxyImpl to accept an empty set of keys over which to execute an entry processor, with early return of an empty map. Added test cases over both classes.


stevebarham:
FYI, I signed the CLA and emailed it to enes.

gurbuzali:
run-lab-run

gurbuzali:
looks good to me, @ahmetmircik can you take a look as well ?

ahmetmircik:
:+1:

PR #7641, opened 2016-03-02
Moved response dispatching into ResponsePacketHandlerImp

pveentjer:
There are a some optimizations that rely on the response to be removed
and this PR is a preparation of these optimizations.


jerrinot:
verify

jerrinot:
:+1:  (on green)

tombujok:
:+1: once green

PR #7642, opened 2016-03-02
Pulled out ThreadLocalRandom

pveentjer:
There are some other pr's pending that rely on this class to be available as
utility.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -358,6 +358,10 @@
     <suppress checks="JavadocVariable" files="com/hazelcast/util/executor/"/>
     <suppress checks="NPathComplexity" files="com/hazelcast/util/executor/DelegatingFuture"/>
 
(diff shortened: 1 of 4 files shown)


jerrinot:
why don't you model the interface after the ThreadLocalRandom from JDK 1.7? 

`public int nextInt(int bound)` etc. 

That would make transition easier once we upgrade our minimal Java requirements.

pveentjer:
Good point. But apart from that?

jerrinot:
well, the alternative is to directly use Doug's code: http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/jsr166y/ThreadLocalRandom.java?revision=1.17&view=markup

It's a backport to JDK6 released into public domain. It potentially offers a bit higher performance (there is  padding to eliminate false sharing) at the expense of some extra memory (8 longs) :)

jerrinot:
:+1:

PR #7643, opened 2016-03-02
Renamed MonitorThread to InvocationMonitorThread

pveentjer:
I renamed it since I'm using it in documentation as a concept and the current
name only has meaning inside of the InvocationMonitor.


jerrinot:
:+1:

PR #7644, opened 2016-03-02
InvocationRegistry implements Iterable

pveentjer:
Instead of exposing the underlying collection, an iterator is now returned.
This way the original collection is shield from the outside. This makes it easier
to change the collection.

This PR is needed for another PR where I used multiple collections (arrays) to store
invocations.


jerrinot:
:+1:

pveentjer:
Since this is a very small change, I'll merge this one.

PR #7646, opened 2016-03-02
merge internal and scheduled executor

sancar:


gurbuzali:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
```diff
@@ -128,35 +110,17 @@ public void execute(Runnable command) {
...
 
     @Override
     public ScheduledFuture<?> scheduleWithRepetition(final Runnable command, long initialDelay, long period, TimeUnit unit) {
         final Runnable decoratedCommand = new SkipOnConcurrentExecutionDecorator(command);
-        return scheduledExecutor.scheduleAtFixedRate(new Runnable() {
-            public void run() {
-                executeInternalSafely(decoratedCommand, failureLoggingExecutionCallback);
-            }
-        }, initialDelay, period, unit);
+        return internalExecutor.scheduleAtFixedRate(decoratedCommand, initialDelay, period, unit);
```
jerrinot: the decorator is not needed here? as it does use the JDK scheduler which makes the single-threaded guarantee on its own.

sancar: @jerrinot I have removed the decorator. I was thinking by leaving that we will make sure the match the semantics. But, after your warning, it occurred to me that decorator is useless here. 
Although I don't see this as problem,  we will end up with slightly different semantics here. 
 `scheduleAtFixedRate ` will run next task immediately in case of the previous one takes longer than the period. But `scheduleWithRepetition` on node side will wait for next period checkpoint.

sancar:
run-lab-run

jerrinot:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
jerrinot: yes. eventually I'd like to have a common scheduler with the same semantic on both client and server. Basic infrastructure should be shared. I know it's not always possible, but it should be the ultimate goal.

PR #7653, opened 2016-03-02
Redesign of Memory Manager

mtopolnik:
1. Split `MemoryStats` into heap and native stats;
2. Split the `MemoryAccessor` interface:
    1.  `MemoryAccessor` itself now supports only basic memory access: get/put all Java primitive types, set memory and copy memory.
    2. `ConcurrentMemoryAccessor` adds support for concurrent access: volatile, ordered, CAS.
    4. `ByteAccessStrategy` is a small interface dedicated to abstracting the logic of endianness-specific operations from the low-level technique of reading/writing individual bytes. This allows a single implementation of this logic in `EndiannesUtil`.
    5. `GlobalMemoryAccessor` adds addressing of Java heap locations in the form (`Object base, long offset`). and provides the contract point for Unsafe-like strategy objects. Specific implementations of this interface take care of memory alignment.


toktarev:
Design looks good

jerrinot:
test failures:
```
com.hazelcast.internal.memory.impl.AlignmentAwareMemoryAccessTest.test_compareAndSwapLong_whenUnaligned
com.hazelcast.internal.memory.impl.AlignmentAwareMemoryAccessTest.test_putOrderedLong_whenUnaligned
com.hazelcast.internal.memory.impl.AlignmentAwareMemoryAccessTest.test_putOrderedInt_whenUnaligned
com.hazelcast.internal.memory.impl.AlignmentAwareMemoryAccessTest.test_putOrderedObject_whenUnaligned
com.hazelcast.internal.memory.impl.AlignmentAwareMemoryAccessTest.test_compareAndSwapObject_whenUnaligned
com.hazelcast.internal.memory.impl.AlignmentAwareMemoryAccessTest.test_putGetObject_whenUnaligned
com.hazelcast.internal.memory.impl.AlignmentAwareMemoryAccessTest.test_compareAndSwapInt_whenUnaligned
```

toktarev:
I see but can't fix it
It is not my branch

mtopolnik:
I already know what they are about and there are more mistakes in the same class that the tests didn't flag. I'll fix and write more tests.

hazelcast/src/main/java/com/hazelcast/internal/memory/GlobalMemoryAccessor.java
```diff
@@ -0,0 +1,38 @@
...
+ * </p><p>
+ * This interface also works as a {@link ByteAccessStrategy} with the semantics analogous to the
+ * semantics of {@code Unsafe}: when an object is supplied as the "resource", access is relative to
+ * that object's base address; when {@code null} is supplied, a base address of zero is implied, i.e.,
+ * the supplied "offset" is treated as the absolute address in CPU's native address space.
+ * </p>
+ */
+@SuppressWarnings("checkstyle:interfaceistype")
+public interface GlobalMemoryAccessor
+        extends ConcurrentMemoryAccessor, HeapMemoryAccessStrategy, ByteAccessStrategy<Object> {
```
serkan-ozal: `GlobalMemoryAccessor` is dependent (extends) to `HeapMemoryAccessStrategy` and also `HeapMemoryAccessStrategy` is dependent to `GlobalMemoryAccessor` over `GlobalMemoryAccessorRegistry.MEM`.
There should not be circular dependency

mtopolnik: In simpler terms, `HeapMemoryAccessStrategy` has an explicit dependency on its implementation. We can resolve this by moving the constants to the implementation class. Maybe the constants shouldn't even be public; the values could be retrieved from the chosen `GlobalMemoryAccessor`.

serkan-ozal: Does it needed `GlobalMemoryAccessor` is a kind of `HeapMemoryAccessStrategy`. I think, there should not be hierarchical relation between `MemoryAccessor` and `MemoryAccessStrategy` but `MemoryAccessor` can use `MemoryAccessStrategy`.

We had already talked about the design and our designs are very similar. I had thought about the design again today and in my design, layers are basicly like this:

```
              ==================
              | MemoryAccessor | => `NativeMemoryAccessor`, `ByteArrayMemoryAccessor`
        ========================
        | MemoryAccessStrategy | => `StandardMemoryAccessStrategy`, `AlignmentAwareMemoryAccessStrategy`
================================
|          MemoryIO            | => Your `EndiannessUtil`
================================
```

`MemoryAccessor`s are the top level abstraction and there is no hierarchical relation between `MemoryAccessor` and `MemoryAccessStrategy`

toktarev: Do we really need MemoryAccessor ?
I think everything can be considered as MemoryAccessStrategy
I don’t see too big difference between these notions
both them have some access to some memory

toktarev: Or everything can be considered as MemoryAccessor without MemoryAccessStrategy.
Two abstractions are redundant here.

We have
ConcurrentMemoryAccessor ObjectMemoryAccessor ByteMemoryAccessor ....

Two abstractions really confuse everybody and overcomplicated design

serkan-ozal: For me, `MemoryAccessStrategy` should be for memory alignment and it takes base address and base object. It is a kind of `Unsafe`. But `MemoryAccessor` itself has wrapped base address/byte[] so the user data structure just uses it like `memoryAccessor.getInt(offset)` and it internally calls `MemoryAccessStrategy`. In here `MemoryAccessStrategy` is for abstraction over platform not for data structure (on-heap / off-heap)

Normally users should not deal with `MemoryAccessStrategy` but `MemoryAccessor`

toktarev: I don't see really big difference among them MemoryAccessStrategy is kind of MemoryAccessor as well.

serkan-ozal: One of the difference is that (which is important for me) you don't need to pass null object as resource to `MemoryAccessStrategy` for native memory accesses or base address which will auto-boxed to `java.lang.Long`.

And also when you implement your `MemoryAccessor` impl for your data structure, you don't need to deal with alignment because this is another layer's responsibility.

 There should be separation of concerns.

toktarev: Sure we will have 2 interfaces but all them will be MemoryAccessors
MemoryAccessor
   ByteMemoryAccessor
   ConcurrentMemoryAccessor
   ResourceMemoryAccessor

toktarev: Strategy is something about behavior
ProcessingStrategy ,  HashingStrategy

But MemoryAccessStrategy just perform some access to Memory like common MemoryAccessor

serkan-ozal: > Sure we will have 2 interfaces but all them will be MemoryAccessors
MemoryAccessor
ByteMemoryAccessor
ConcurrentMemoryAccessor
ResourceMemoryAccessor

All of us are agree about this approach. But my point is that alignment should be handled at another layer not in this `MemoryAccessor` layer. `MemoryAccessor` should be just about where the data is accessed (native memory, byte[] or another resource)

toktarev: But if we have some layers it should be some inheritance.
Now I don't see inheritance, I see two independent abstractions with similar functionality

serkan-ozal: NO, layers don't mean inheritance. Layers can call each other (layer at the below) by delegation.
In here, there is delegation. As I already mentioned above more than once, `MemoryAccessor` uses/delegates `MemoryAccessStrategy` (talking about my design). They are not independent

toktarev: >>alignment

I would understand if we have many implementation of such alignment.
In reality - we will have only 1 Strategy with alignment - UNSAFE-based.

toktarev: I don't see some delegation from MemoryAccessor to MemoryAccessStrategy
could you please show it ?

serkan-ozal: NO, we have 2 different alignment strategy.
STANDARD and ALIGNMENT_AWARE.
If you guarantee that your data structure is aligned, no need to use `ALIGNMENT_AWARE`, just use `STANDARD`. Because there is extra alignment check. But if there is no guarantee, you should use `ALIGNMENT_AWARE`

serkan-ozal: @toktarev PLEASE READ MY COMMENTS ABOVE AGAIN. 

There is `talking about my design`

toktarev: hmm, ok

But I would call it AlignmentMemoryAccessor

toktarev: >> PLEASE READ MY COMMENTS ABOVE AGAIN.

Ok I see that Marko's design is slightly different and discuss about his design

hazelcast/src/main/java/com/hazelcast/memory/JVMMemoryStats.java
```diff
@@ -0,0 +1,48 @@
...
+    MemoryStats getNativeMemoryStats();
+
+    /**
+     * Returns the garbage collector statistics for the JVM
+     *
+     * @return GC statistics
+     */
+    GarbageCollectorStats getGCStats();
+
+    long getMaxMetadata();
```
toktarev: May be we should add javaDoc here as well ??

toktarev:
:+1: looks good for (with one little remark)

mtopolnik:
Failure:



```
testLoads_whenInitialLoaderNodeRemoved(com.hazelcast.map.impl.mapstore.MapLoaderFailoverTest)  Time elapsed: 60.083 sec  <<< ERROR!
10:18:26 org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
10:18:26 	at java.lang.Object.wait(Native Method)
10:18:26 	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.pollResponse(InvocationFuture.java:300)
10:18:26 	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.waitForResponse(InvocationFuture.java:245)
10:18:26 	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.get(InvocationFuture.java:222)
10:18:26 	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.get(InvocationFuture.java:202)
10:18:26 	at com.hazelcast.spi.impl.operationservice.impl.InvokeOnPartitions.retryFailedPartitions(InvokeOnPartitions.java:131)
10:18:26 	at com.hazelcast.spi.impl.operationservice.impl.InvokeOnPartitions.invoke(InvokeOnPartitions.java:67)
10:18:26 	at com.hazelcast.spi.impl.operationservice.impl.OperationServiceImpl.invokeOnAllPartitions(OperationServiceImpl.java:368)
10:18:26 	at com.hazelcast.map.impl.proxy.MapProxySupport.size(MapProxySupport.java:573)
10:18:26 	at com.hazelcast.map.impl.proxy.MapProxyImpl.size(MapProxyImpl.java:80)
10:18:26 	at com.hazelcast.map.impl.mapstore.MapLoaderFailoverTest.testLoads_whenInitialLoaderNodeRemoved(MapLoaderFailoverTest.java:92)
```

mtopolnik:
verify

mtopolnik:
Barrage of spuriously failed testts (don't reproduce locally):

https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/1722/testReport/

mtopolnik:
run-lab-run

mdogan:
:+1:

mdogan:
good job

PR #7656, opened 2016-03-02
Removed deprecated methods from OperationService

pveentjer:


hazelcast/src/main/java/com/hazelcast/spi/OperationService.java
```diff
@@ -112,19 +46,6 @@
...
-     * method, otherwise {@link #executeOperation(Operation)} should be used.
-     *
-     * @param op the operation to check.
-     * @return true if the operation is allowed to run on the calling thread, false otherwise.
-     *
-     * @deprecated since 3.5 since not needed anymore.
-     */
-    @Deprecated
-    boolean isAllowedToRunOnCallingThread(Operation op);
-
```
tombujok: What about the last deprecated method?
`boolean send(Response response, Address target);`

pveentjer: Will be removed in a different pr (removal of response objects).

tombujok:
:+1:

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/OperationService.java
mdogan: btw, just a reminder... `isAllowedToRunOnCallingThread()` is used by EE IMap and ICache.

Donnerbart: Is this a showstopper for this PR? It already has two reviewers, but if we use this method internally, it cannot be removed. Or is there an alternative which should be used instead?

PR #7658, opened 2016-03-03
Renamed WaitSupport to BlockingOperation

pveentjer:
This new name is in line with other Operations we have like BackupOperation,
PartitionAwareOperation, ReadonlyOperation and it immidiately shows it is
an interface that is implemented by Operations. The old naming was not
self explanatory.

WaitSupport is still there to prevent causing incompatibilities, But is deprecated.


jerrinot:
:+1: from me. please do not merge it before it's reviewed by at least 1 reviewer.

hazelcast/src/main/java/com/hazelcast/spi/WaitSupport.java
```diff
@@ -16,19 +16,8 @@
...
-
-    WaitNotifyKey getWaitKey();
-
-    boolean shouldWait();
-
-    long getWaitTimeout();
-
-    void onWaitExpire();
+public interface WaitSupport extends BlockingOperation {
 }
```
tombujok: good idea!

tombujok:
looks good - it's much more intuitive now.
:+1:

PR #7659, opened 2016-03-03
Cleaned up Packet flags logic

pveentjer:
The Packet flags logic has been cleaned up so that all flags can be set in one go
using 
```
FLAGS_OP | FLAGS_URGENT. 
```

The bit shifting logic is the the same as used in e.g.  the SelectionKey where you can register interested ops using 
```
interestedOps(OP_READ | OP_WRITE)
```

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/Packet.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/Packet.java
@@ -36,12 +36,12 @@ public final class Packet extends HeapData implements OutboundFrame {
 
     public static final byte VERSION = 4;
 
-    public static final int FLAG_OP = 0;
-    public static final int FLAG_RESPONSE = 1;
-    public static final int FLAG_EVENT = 2;
-    public static final int FLAG_WAN_REPLICATION = 3;
-    public static final int FLAG_URGENT = 4;
-    public static final int FLAG_BIND = 5;
+    public static final int FLAG_OP = 1 << 0;
+    public static final int FLAG_RESPONSE = 1 << 1;
+    public static final int FLAG_EVENT = 1 << 2;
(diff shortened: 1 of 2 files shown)


pveentjer:
run-lab-run

pveentjer:
verify

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -88,12 +88,35 @@ public void setConn(Connection conn) {
         this.conn = conn;
     }
 
-    public void setFlag(int bit) {
-        flags |= 1 << bit;
+    /**
+     * Sets a particular flag.
```
jerrinot: could you refine the contract a bit? something like: "other flags are left intact"

pveentjer: Imho that is implied by the fact that flags are being used. It would be illogical that other flags would be changed.

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -36,12 +36,12 @@
...
     public static final byte VERSION = 4;
 
-    public static final int FLAG_OP = 0;
-    public static final int FLAG_RESPONSE = 1;
-    public static final int FLAG_EVENT = 2;
-    public static final int FLAG_WAN_REPLICATION = 3;
-    public static final int FLAG_URGENT = 4;
-    public static final int FLAG_BIND = 5;
+    public static final int FLAG_OP = 1;
+    public static final int FLAG_RESPONSE = 1 << 2;
```
jerrinot: one more thing: why the bit-shifting is starting with `<< 2` and not `<<1` ?

PR #7665, opened 2016-03-03
Removed OperationResponseHandler.isLocal

pveentjer:
Also the RemoteOperationResponseHandler is not throwing an exception anymore, but just
logging the fact that the response isn't send. There is no added value in showing
big stack traces here.


pveentjer:
The code around node shutdown is a mess. It is very fragile and unpredictable. We'll need to get it fixed at some point, but not today.
donotshave

PR #7667, opened 2016-03-03
Revert changes in HazelcastInstanceFactory

jerrinot:
HazelcastInstanceFactory is an internal API, but we discovered external projects do use it directly anyway.
We recently did two changes:

1. Renamed HazelcastInstanceFactory into HazelcastInstanceManager (addf2002)
2. Introduced a new abstraction for HazelcastInstanceFactory (2b5ba5ab) and dynamic registration via META-INF

Both changes might break external projects.

The 1st change is easy to revert as it's just naming. Manager actually describes the responsibility better,
but API compatibility has priority. We could introduce a compatibility layer, but it's not worth the extra complexity.

The 2nd change - dynamic loading of HazelcastInstanceFactory was introduce due JET integration,
but they do not use it anymore and the abstraction servers no purposes -> we can remove it too.


pveentjer:
History is being lost in this pr. Can you pull out the reverting of the renaming in 1 pr? Once that is done; rebase this pr and then history won't be lost.

jerrinot:
oh, good catch!

bilalyasar:
:+1:

pveentjer:
:+1:  but needs a rebase

pveentjer:
@jerrinot can you rebase so it can be merged?

PR #7668, opened 2016-03-04
Simplification send response

pveentjer:


pveentjer:
I'll merge this one since there is no difference in logic.

PR #7669, opened 2016-03-04
Removed the Invocation.notify method

pveentjer:
The Invocation.notify method isn't needed.

The content has been copied into the sendResponse method.

And all notify calls that want to notify of an exception, immediately call notifyError.


pveentjer:
verify

jerrinot:
:+1:

PR #7673, opened 2016-03-04
Authentication will not block internal executor

sancar:
Client authentication mechanism is refactored so that there is
no remote call left on internal executor.

related to #7635


sancar:
run-lab-run

gurbuzali:
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -402,35 +438,113 @@ public void addConnectionHeartbeatListener(ConnectionHeartbeatListener connectio
...
                     logger.finest(e);
-                } finally {
+                    callback.onFailure(e);
                     connectionsInProgress.remove(target);
+                    return;
                 }
-
             }
+
+            authenticate(target, connection, asOwner, callback);
```
mdogan: What if `authenticate()` itself fails with an exception? You should remove `target` from `connectionsInProgress` when `authenticate()` fails.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -402,35 +438,113 @@ public void addConnectionHeartbeatListener(ConnectionHeartbeatListener connectio
...
+    private void authenticated(Address target, ClientConnection connection) {
+        if (connections.put(connection.getRemoteEndpoint(), connection) == null) {
+            fireConnectionAddedEvent(connection);
         }
+        connectionsInProgress.remove(target);
     }
+
+    private void failed(Address target, ClientConnection connection, Throwable throwable) {
+        logger.finest(throwable);
+        connection.close(throwable);
```
mdogan: What if connection is already there but authentication `asOwner` fails? Connection will be closed but is it going to be removed from connections map?

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -192,66 +216,90 @@ protected void shutdownSelectors() {
...
-    public ClientConnection getOrTriggerConnect(Address target, Authenticator authenticator) throws IOException {
-        Address remoteAddress = addressTranslator.translate(target);
+    private Connection getOrTriggerConnectInternal(Address target, boolean asOwner,
+                                                   AuthenticationCallback callback) {
+        target = addressTranslator.translate(target);
 
-        if (remoteAddress == null) {
-            throw new IOException("Address is required!");
+        if (target == null) {
+            return null;
```
mdogan: Shouldn't you throw an exception here, instead of returning null? I think passing `target` as null or getting null from `addressTranslator` is an exceptional case.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -402,35 +438,121 @@ public void addConnectionHeartbeatListener(ConnectionHeartbeatListener connectio
...
             }
         }
     }
+
+    private void authenticated(Address target, ClientConnection connection) {
+        ClientConnection oldConnection = connections.put(connection.getRemoteEndpoint(), connection);
+        if (oldConnection == null) {
+            fireConnectionAddedEvent(connection);
+        }
+        assert (connection.equals(oldConnection));
```
mdogan: `assert oldConnection == null || connection.equals(oldConnection)` ?

mdogan:
:+1:

PR #7674, opened 2016-03-04
Authentication will not block internal executor

sancar:
Client authentication mechanism is refactored so that there is
no remote call left on internal executor.

related to #7635


sancar:
run-lab-run

sancar:
Merging since master is merged. https://github.com/hazelcast/hazelcast/pull/7673

PR #7677, opened 2016-03-05
[DONT MERGE] Notifying directly from io threads

pveentjer:


hazelcast/src/main/java/com/hazelcast/internal/serialization/SerializationService.java
```diff
@@ -42,6 +42,8 @@
 
     <T> T toObject(Object data);
 
+    <T> T bytesToObject(byte[] bytes, int dataOffset);
```
toktarev: Can we add also int length here ?
or separate method
+    <T> T bytesToObject(byte[] bytes, int dataOffset, int length);
?

pveentjer: What would be the purpose of the length? The deserialization step determines itself how much data it is going to read.

toktarev: So we always rely on the fact that object is totally presented in byte[] array ?

What is the goal of dataOffset in this case ?

toktarev: I see, it is correct

pveentjer:
run-lab-run

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/serialization/SerializationService.java
toktarev: I am really appreciative for this method
I was going to implement it these days :) Thanks

PR #7678, opened 2016-03-05
Revert changes to MemoryStats

mtopolnik:
The attempt to introduce an abstract notion of statistics for a MemoryManager failed because statistics are tightly coupled to the actual implementation and the abstract notions are of questionable value.

Therefore the `getStatistics` methods was removed from the general `MemoryManager` and the `MemoryStats` interface was reverted to prior state. The only change is the removal of the redundant `Memory` word from `getXNativeMemory` methods.


mtopolnik:
Merging now as these changes are urgent to make the EE side build again.

toktarev:
Could you please explain motivation of this change in details ?
MemoryStats is public API ?

mtopolnik:
This is the undoing of the previous change. MemoryStats wasn't public API before, therefore isn't now, either. In the future maybe we'll manage to properly generalize MemoryStats.

toktarev:
If MemoryStats wasn't public API before, why we can't modify it ?
The change was really valuable and decoupled MemoryStats from memory implementation

mtopolnik:
The immediate concern is insufficient work on a sensible implementation in
HZ memory manager. The API itself will need work to accommodate that. To
make EE buildable again this had to be removed.

PR #7679, opened 2016-03-07
Removed instanceof checks SocketWriter

pveentjer:
Instanceof checks can be very expensive. This is replaced by a simple
frame.getClass()=FrameTask.class since that is the only type of task frame being
used with this change. The actuall runnable is wrapped inside of this
TaskFrame.


jerrinot:
:+1:

PR #7681, opened 2016-03-07
ByteAccessStrategy has been added to the MemoryAccessor interface

toktarev:
Every memory accessor should return some implementation of default ByteAccessStrategy<Void>

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/memory/MemoryAccessor.java
@@ -155,17 +155,24 @@ public interface MemoryAccessor {
      *
      * @param srcAddress  the source address to be copied from
      * @param destAddress the destination address to be copied to
-     * @param lengthBytes       the number of bytes to be copied
+     * @param lengthBytes the number of bytes to be copied
      */
     void copyMemory(long srcAddress, long destAddress, long lengthBytes);
 
     /**
      * Sets memory with given value from specified address as given size.
      *
-     * @param address the start address of the memory region
-     *                which will be set with given value
-     * @param lengthBytes   the number of bytes to be set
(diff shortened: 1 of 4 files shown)


mtopolnik:
The same concern can be served by just defining an implementation of `ByteAccessStrategy<MemoryAccessor>`, and providing a singleton instance as a constant in `EndiannessUtil`.

toktarev:
MemoryAccessor can be always represented as ByteAccessStrategy<Void>
and we should have this method anyway

cangencer:
why is this needed? MemoryAccessor already has .putByte() and .getByte(). If you need to represent it as a ByteAccessStrategy, then you can just create a new class ```MemoryAccessorByteAcessStrategy implements ByteAccessStrategy<MemoryAccessor>```

toktarev:
It is just contract.

Default ByteAccessStrategy &lt; Void &gt;

can be always used by any MemoryAccessor

cangencer:
Seems to me it will be better served by a static method rather than adding to contract as all MemoryAccessors would have the same implemention for this method.

mtopolnik:
Yes, good point. This method doesn't add anything specific to a `MemoryAccessor`.

mtopolnik:
https://github.com/hazelcast/hazelcast/pull/7682

toktarev:
The question here is just about the code:
Or 

EndiannessUtil.readChar(memoryAccessor.asByteAstrategy)
or
EndiannessUtil.readChar(SomeObject.asStrategy(memoryAccessor))

The first one is more preferable to make code pretty for me
In other aspects approaches are the same

mtopolnik:
It's actually 

```
EndiannessUtil.getCharB(CUSTOM_ACCESS, mem, offset);
```

toktarev:
Now I see.
yes good point. I agree with you guys

PR #7683, opened 2016-03-07
Puts getRandomSamples method to storage interface

ahmetmircik:


ahmetmircik:
This is required for simulator runs. Simple change, merging it.

PR #7684, opened 2016-03-07
TCP: Replacing Clock.currentTimeMs by System

pveentjer:
No need for the abstraction here. Just more work for the io thread
to deal with. Fastest call is directly to System.


jerrinot:
:+1:

PR #7685, opened 2016-03-07
Decreases number of serializations in case of getAll

ahmetmircik:
This appeared when solving issue https://github.com/hazelcast/hazelcast-enterprise/issues/788


gurbuzali:
:+1:

PR #7686, opened 2016-03-07
SocketReaders/Writers don't check shutdown any longer

pveentjer:
There is no need to check this. If the connection is going to shut down,
the socket channel is closed eventually and you'll get a
ClosedSocketChannelException in case of reading or writing from/to the SocketChannel. 

So the checking is already done at a lower level. No need to obfuscate the flow with repeated checking.


mdogan:
👍

pveentjer:
This cleanup already has been merged into a different PR.

PR #7690, opened 2016-03-08
Nicer percentage rendering in HealthMonitor

mtopolnik:
A very minor change which presents a percentage with two decimals instead of as an integer.


Donnerbart:
:+1:

PR #7692, opened 2016-03-08
SchedulingExecutorServiceDelegate does not implement j.u.c.ScheduledExecutorService

### LINKED ISSUE #7611: ScheduledExecutorServiceDelegate violates contract of ScheduledExecutorService
Contract of `scheduleAtFixedRate` says: 
```
     * If any execution of this task
     * takes longer than its period, then subsequent executions
     * may start late, but **will not concurrently execute**.
```

However the `ScheduledExecutorServiceDelegate` wraps tasks in `ScheduledTaskRunner` which delegates to a different executor. As a consequence a task can be executed concurrently thus a contract of `ScheduledExecutorService` is violated. 

This leads to surprises. this was seen in a testing environment: 
```java
Exception in thread "cached4" java.lang.IllegalArgumentException: Comparison method violates its general contract!
    at java.util.TimSort.mergeHi(TimSort.java:895)
    at java.util.TimSort.mergeAt(TimSort.java:512)
    at java.util.TimSort.mergeCollapse(TimSort.java:437)
    at java.util.TimSort.sort(TimSort.java:241)
    at java.util.Arrays.sort(Arrays.java:1512)
    at java.util.ArrayList.sort(ArrayList.java:1454)
    at java.util.Collections.sort(Collections.java:175)
    at com.hazelcast.map.impl.eviction.ExpirationManager$ClearExpiredRecordsTask.sortPartitionContainers(ExpirationManager.java:127)
    at com.hazelcast.map.impl.eviction.ExpirationManager$ClearExpiredRecordsTask.run(ExpirationManager.java:121)
    at com.hazelcast.util.executor.CachedExecutorServiceDelegate$Worker.run(CachedExecutorServiceDelegate.java:212)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)
    at com.hazelcast.util.executor.HazelcastManagedThread.executeRun(HazelcastManagedThread.java:76)
    at com.hazelcast.util.executor.HazelcastManagedThread.run(HazelcastManagedThread.java:92)
```

jerrinot:
Fixes #7611

Our scheduler has a different contract hence it should not
implement the interface from JDK.

I renamed it to the SchedulingExecutorServiceDelegate to avoid confusion.
Also I changed the method name from scheduleAtFixedRate to scheduleWithRepetition
to emphasize it has a different semantic.

The scheduleWithFixedDelay was removed as it never worked and it
effectively behaved as the scheduleAtFixedRate anyway.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
@@ -21,6 +21,7 @@ import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.LoggingService;
+import com.hazelcast.spi.impl.executionservice.impl.SkipOnConcurrentExecutionDecorator;
 import com.hazelcast.util.executor.CompletableFutureTask;
 import com.hazelcast.util.executor.PoolExecutorThreadFactory;
 import com.hazelcast.util.executor.SingleExecutorThreadFactory;
@@ -149,19 +150,11 @@ public final class ClientExecutionServiceImpl implements ClientExecutionService 
     }
 
     @Override
-    public ScheduledFuture<?> scheduleAtFixedRate(final Runnable command, long initialDelay, long period, TimeUnit unit) {
+    public ScheduledFuture<?> scheduleWithRepetition(final Runnable command, long initialDelay, long period, TimeUnit unit) {
+        final Runnable decoratedCommand = new SkipOnConcurrentExecutionDecorator(command);
         return scheduledExecutor.scheduleAtFixedRate(new Runnable() {
             public void run() {
-                executeInternalSafely(command, failureLoggingExecutionCallback);
-            }
-        }, initialDelay, period, unit);
-    }
-
-    @Override
-    public ScheduledFuture<?> scheduleWithFixedDelay(final Runnable command, long initialDelay, long period, TimeUnit unit) {
-        return scheduledExecutor.scheduleWithFixedDelay(new Runnable() {
-            public void run() {
-                executeInternalSafely(command, failureLoggingExecutionCallback);
+                executeInternalSafely(decoratedCommand, failureLoggingExecutionCallback);
             }
         }, initialDelay, period, unit);
     }
--- a/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueService.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueService.java
@@ -47,6 +47,7 @@ import com.hazelcast.spi.PartitionReplicationEvent;
 import com.hazelcast.spi.RemoteService;
 import com.hazelcast.spi.StatisticsAwareService;
 import com.hazelcast.spi.TransactionalService;
+import com.hazelcast.spi.TaskScheduler;
 import com.hazelcast.transaction.impl.Transaction;
 import com.hazelcast.util.ConcurrencyUtil;
 import com.hazelcast.util.ConstructorFunction;
@@ -63,7 +64,6 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.logging.Level;
 
 /**
@@ -93,11 +93,10 @@ public class QueueService implements ManagedService, MigrationAwareService, Tran
 
     public QueueService(NodeEngine nodeEngine) {
         this.nodeEngine = nodeEngine;
-        ScheduledExecutorService defaultScheduledExecutor
-                = nodeEngine.getExecutionService().getDefaultScheduledExecutor();
+        TaskScheduler globalScheduler = nodeEngine.getExecutionService().getGlobalTaskScheduler();
         QueueEvictionProcessor entryProcessor = new QueueEvictionProcessor(nodeEngine, this);
         this.queueEvictionScheduler = EntryTaskSchedulerFactory.newScheduler(
-                defaultScheduledExecutor, entryProcessor, ScheduleType.POSTPONE);
+                globalScheduler, entryProcessor, ScheduleType.POSTPONE);
         this.logger = nodeEngine.getLogger(QueueService.class);
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/SkipOnConcurrentExecutionDecorator.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/SkipOnConcurrentExecutionDecorator.java
@@ -16,26 +16,31 @@
 
 package com.hazelcast.spi.impl.executionservice.impl;
 
-import com.hazelcast.util.ExceptionUtil;
+import java.util.concurrent.atomic.AtomicBoolean;
 
-import java.util.concurrent.Executor;
-
-class ScheduledTaskRunner implements Runnable {
-
-    private final Executor executor;
+/**
+ * Decorator to prevent concurrent task execution.
+ *
+ * If this task is running on a thread and another thread calls attempt the execute it concurrently
+ * then the 2nd execution will be skipped.
+ *
+ */
+public class SkipOnConcurrentExecutionDecorator implements Runnable {
+    private final AtomicBoolean isAlreadyRunning = new AtomicBoolean();
     private final Runnable runnable;
 
-    public ScheduledTaskRunner(Runnable runnable, Executor executor) {
-        this.executor = executor;
+    public SkipOnConcurrentExecutionDecorator(Runnable runnable) {
         this.runnable = runnable;
     }
 
     @Override
     public void run() {
-        try {
-            executor.execute(runnable);
-        } catch (Throwable t) {
-            ExceptionUtil.sneakyThrow(t);
+        if (isAlreadyRunning.compareAndSet(false, true)) {
+            try {
+                runnable.run();
+            } finally {
+                isAlreadyRunning.set(false);
+            }
         }
     }
 }
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
@@ -193,8 +193,8 @@ abstract class AbstractClientInternalCacheProxy<K, V>
         }
 
         @Override
-        public ScheduledFuture<?> scheduleWithFixedDelay(Runnable command, long initialDelay, long delay, TimeUnit unit) {
-            return clientExecutionService.scheduleWithFixedDelay(command, initialDelay, delay, unit);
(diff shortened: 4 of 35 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
```diff
@@ -21,6 +21,7 @@
 import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.LoggingService;
+import com.hazelcast.spi.impl.executionservice.impl.SkipOnConcurrentExecutionDecorator;
```
jerrinot: I am not sure about this dependency here.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
```diff
@@ -149,19 +150,11 @@ public void run() {
...
-            }
-        }, initialDelay, period, unit);
-    }
-
-    @Override
-    public ScheduledFuture<?> scheduleWithFixedDelay(final Runnable command, long initialDelay, long period, TimeUnit unit) {
-        return scheduledExecutor.scheduleWithFixedDelay(new Runnable() {
-            public void run() {
-                executeInternalSafely(command, failureLoggingExecutionCallback);
+                executeInternalSafely(decoratedCommand, failureLoggingExecutionCallback);
```
jerrinot: here is a possible issue where the callback will be executed even when execution is skipped. I am not sure if that's a problem is reality.

jerrinot:
It breaks SPI, but that's a good thing given the semantic shift.

pveentjer:
I'm not too happy with the name SchedulingExecutorServiceDelegate since it interface (which is an abstraction of the implementation) clearly explains its implementation. Why not call it Scheduler?

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueService.java
```diff
@@ -93,11 +93,11 @@ public LocalQueueStatsImpl createNew(String key) {
 
     public QueueService(NodeEngine nodeEngine) {
         this.nodeEngine = nodeEngine;
-        ScheduledExecutorService defaultScheduledExecutor
-                = nodeEngine.getExecutionService().getDefaultScheduledExecutor();
+        SchedulingExecutorService defaultSchedulingExecutorService
```
pveentjer: Default scheduler ----> I would rename it to the Global Scheduler. The one that is the working horse for all kinds of stuff.

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/SchedulingExecutorServiceDelegate.java
```diff
@@ -27,12 +29,14 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
-final class ScheduledExecutorServiceDelegate implements ScheduledExecutorService {
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
+public final class SchedulingExecutorServiceDelegate implements SchedulingExecutorService {
```
pveentjer: I would name this class SchedulingExecutorServiceImpl. It really isn't relevant that it is a delegate. Apart from that most services are called FooImpl.

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/SkipOnConcurrentExecutionDecorator.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Decorator to prevent concurrent task execution.
+ *
+ * If this task is running on a thread and another thread calls attempt the execute it concurrently
+ * then the 2nd execution will be skipped.
+ *
+ */
+public class SkipOnConcurrentExecutionDecorator implements Runnable {
+    private final AtomicBoolean isAlreadyRunning = new AtomicBoolean();
```
pveentjer: Use a volatile int field and an AtomicIntegerFieldUpdater. This will reduce the amount of litter.

pveentjer:
Looks good. Minor suggestions regarding naming.

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/SchedulingExecutorServiceDelegate.java
jerrinot: well, it delegates to a given executor service. this is significant implementation aspect, the reason why we have it at all. otherwise we might very well use just the `ScheduledExecutorService` from JDK. 

what about `DelegatingSchedulingExecutorService` ?

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/SkipOnConcurrentExecutionDecorator.java
jerrinot: `AtomicIntegerFieldUpdater` tends to be a tiny bit slower in current HotSpot. (I think the reason is that HotSpot does not fully trust final fields inside `Atomic*FieldUpdater`) 

I would have used the `Atomic*FieldUpdater` if the tasks were created with high rate. However here I assume a component schedules a task once only - during initialization/start. Hence the amount of objects created is not growing over time. 

If you still think it's good to use the Update I can change it, but I see little reasons to do so.

hazelcast/src/main/java/com/hazelcast/spi/SchedulingExecutorService.java
```diff
@@ -0,0 +1,80 @@
...
+ * guarantees a task won't be executed by multiple threads concurrently. The difference is that this service will simply
+ * skip a scheduled execution if another thread is still running the same task instead of postponing its execution.
+ * To emphasize this difference the method is called <code>scheduleWithRepetition</code>
+ * instead of <code>scheduleAtFixedRate</code>
+ *
+ * The other difference is this service does not offer an equivalent of
+ * {@link java.util.concurrent.ScheduledExecutorService#scheduleWithFixedDelay(Runnable, long, long, TimeUnit)}
+ *
+ */
+public interface SchedulingExecutorService extends ExecutorService {
```
ahmetmircik: To me, it is hard to distinguish at first glance between names of `SchedulingExecutorService` and `ScheduledExecutorService`. (I don't have a better name now but want to state the problem)

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
ahmetmircik: Maybe you can add this potential problem as a note on that method.

ahmetmircik:
:+1:  + minor comments

sancar:
Since client does not have delegate executors for different logics, solution can be much simpler. I can convert internal executor to scheduled executor and give the same guarantees of scheduleWithFixDelay and scheduleWithFixRate.  @jerrinot I think this was also your first thought for server but could not do it because server needs `cachedExecutor`.

Client solution
https://github.com/hazelcast/hazelcast/pull/7646

In short, I am in favor of not touching to client in this pr and merge the #7646 because it will be much simpler.

jerrinot:
@sancar: thanks for review! 

The issue with using the regular scheduler from JDK on the client side is that for example the QueryCache requires the a scheduler on both client and member. 

I believe we should offer a scheduler with the same semantic on both clients and members. 
When a client uses a regular JDK scheduler then it's hard to guarantee the same semantic. wdyt?

sancar:
Hmm. QueryCache does not fit to my proposal. Then go ahead and merge this one. In #7646, I can still merge internal and scheduled executor to one scheduled executor for simplicity. But I will leave `scheduleWithRepetiton` in ClientExecutionService and that method will use `SkipOnConcurrentExecutionDecorator` to match the semantics.

sancar:
@jerrinot  Is this going to be backported ?

jerrinot:
@sancar: I think the backport is needed. at very least the behaviour-change should be backported as the current behaviour is causing some bad side-effects - see the original issue. 

I think I'll backport just the `SkipOnConcurrentExecutionDecorator` and use it in inside the `SchedulingExecutorServiceDelegate` and `ClientExService`, but I will not the the rename to maintain  API compatibility. WDYT?

sancar:
@jerrinot Yes, that make sense. 
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
jerrinot: I double checked it and it's not really an issue: the `ClientExecutionService` does not accept custom callback for `scheduleWithRepetition`. 

The service does use the callback internally, but only to logs errors -> not a problem.

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueService.java
jerrinot: done

hazelcast/src/main/java/com/hazelcast/spi/SchedulingExecutorService.java
jerrinot: I know. It's is a pickle. `ScheduledExecutorService` is a pretty good name for this, unfortunately Doug Lea already took it! :) 

I made the naming pattern consistent with `BufferingInputStream` which faced the same dilemma: similar functionality, slightly different contract.

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/SkipOnConcurrentExecutionDecorator.java
pveentjer: On this level I doubt if this tiny bit slower is going to be relevant.

I don't know how often these tasks are being created; if it really isn't often then leave it as it is.

hazelcast/src/main/java/com/hazelcast/spi/SchedulingExecutorService.java
pveentjer: I had to read multiple time to see the difference at all. So yes; please fix it. Else you also introduce a nasty class very everyone using Hazelcast since they will get this name in their popup.

I would go for 'Scheduler'

pveentjer:
:+1: once you come up with a nice name for the scheduler

jerrinot:
@ahmetmircik, @pveentjer: I changed the name to `TaskScheduler`. 

2 people raised the point the current name is too similar to the name from JDK. unambiguity has higher value then explicit indication it's kind of `ExecutorService`.

pveentjer:
Exposing it as an ExecutorService could even be problematic. The ExecutorService exposes lifecycle methods like shutdown etc. These things you typically don't want to expose in an interface.

pveentjer:
:+1:

jerrinot:
@pveentjer: thanks for pointing it out. I understand methods like `shutdown()` should not be exposed to the user code. It should be managed by Hazelcast. I'll take a note.

PR #7694, opened 2016-03-09
NodeEngine.getSerializationService deprecated

pveentjer:
This method is deprecated since SerializationService is not an SPI, but an internal API. Exposing it through the NodeEngine causes a leak of the internal being exposes to the outside.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/NodeEngine.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/NodeEngine.java
@@ -77,6 +77,8 @@ public interface NodeEngine {
      * Gets the SerializationService.
      *
      * @return the SerializationService.
+     * @deprecated since 3.7. Use {@link #toObject(Object)} or {@link #toData(Object)} to (de)serialize. The SerializationService
+     * is an internal API.
      */
     SerializationService getSerializationService();
 
@@ -233,6 +235,7 @@ public interface NodeEngine {
 
     /**
      * Indicates that node is not shutting down or it has not already shut down
+     *
      * @return true if node is not shutting down or it has not already shut down
      */
     boolean isRunning();


mdogan:
Why `SerializationService` is an internal API? Currently most of the SPI services and distributed object proxies are using it directly. At least, basic needed functionality for services should be provided.

toktarev:
JET uses SerializationService as well

mdogan:
An SPI developer should be able to convert an object to `byte[]` and vice versa.

pveentjer:
@mdogan I'm fine with offering that solution, but then we should create an interface we are comfortable with maintaining.

jerrinot:
I agree this should not leak into SPI. However there are 88 production usages in our own code. 

If we are going to deprecate it then we should offer a viable alternative SPI users can use instead of this. I can see it has been exposed 3+ years. It's not that it accidentally slipped through yesterday.

pveentjer:
Offering a viable solution is not part of this PR. This method is broken and we should not expose it to the big public because we keep tying them to our internal API.

pveentjer:
I believe that most of the calls to serializationservice are either toObject or toData.

toktarev:
I would remain
toBytes and fromBytes as public API
no needs to create HeapData if you have only byte[] to deserialize

enesakar:
A solution suggestion:
- Move SerializationService to the SPI because of its production usage.
- Do not deprecate the whole class (SerializationService) and NodeEngine.getSerializationService() method.
- Deprecate the methods that should not be leaked in SerializationService.
- Create a seperate internal class (InternalSerializationService) under internal package
- In 3.8 remove the depracated methods but preserve Serialization service under SPI.

toktarev:
@enesakar +1

sancar:
@enesakar +1

pveentjer:
Sure. I'm not going to do this refactoring; I'll leave it to the team that owns the serialization.

enesakar:
@sancar is on it.

pveentjer:
My suggestion would be to start only with toData and toObject. The more added, the more is exposed and the more needs to be maintained.

PR #7695, opened 2016-03-09
NodeEngine.getService is added

pveentjer:
By adding a getService method, services can be added without needing to add them to the NodeEngine interface. For example WanReplicationService doesn't need to be added as explicit method on NodeEngine, but
```
WanReplicatioService service = nodeEngine.getService(WanReplicationService.SERVICE_NAME)
```
can be used

This is a much more flexible approach and make it a lot easier to add all kinds of new services, without needing to change the HZ core.


Donnerbart:
If I'm not wrong this missing method was on of the issues @noctarius ran into while implementing SnowCast. You cannot access your own services after you registered them. So I think it's a good addition to the SPI.

:+1:

PR #7696, opened 2016-03-09
Deprecated SharedService

pveentjer:
All services can be access using NodeEngine.getService(name). No need for the SharedService marker interface

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapManagedService.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapManagedService.java
@@ -40,7 +40,7 @@ public class MapManagedService implements ManagedService {
 
     @Override
     public void init(NodeEngine nodeEngine, Properties properties) {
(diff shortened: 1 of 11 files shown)


Donnerbart:
I'm not sure about this one. I remember an explanation when to use which interface, I think from a support ticket. But I honestly have no idea if there is a real effect of using the marker interface `SharedService` or not. But since I remember that explanation to an external user, there will surely be implementors of that interface.

pveentjer:
There is currently 1 shared service, the lock service. I fail to see the added value of this shared service. It makes to SPI more complex to use than needs to be. Also any service one should be able to look up; especially his own. His own service should not need to implement 'SharedService' because  it will only be shared with himself.

Donnerbart:
With the deprecation instead of deletion the PR looks good to me. As I said, I expect external implementations, so deprecation should be the way to go.

I added the JavaDoc to the `SharedService` myself in #6265, but can't find the source where I picked up the information. I think it was a support ticket where @eminn or @mdogan explained the usage to a customer. So maybe one of them should be the second reviewer.

:+1:

PR #7699, opened 2016-03-09
Moved WaitNotifyService to impl

pveentjer:
The WaitNotifyService is an implementation detail and is not used
by the SPI user. So therefor it is moved to com.hazelcast.spi.waitnotify 

As part of the moving the InternalWaitNotifyService has been merged into the WaitNotifyService since it doesn't serve a lot of purpose to have 2 internal interfaces.

No Enterprise change needed.


jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/BlockingOperation.java
```diff
@@ -18,9 +18,7 @@
 
 
 /**
- * A interface that can be implemented to participate in the Wait/Notify System.
- *
- * See {@link  com.hazelcast.spi.WaitNotifyService}.
+ * A interface that can be implemented by an operation that can block. For example an ILock.acquire.
```
Donnerbart: Nip picking, but it should be `An interface`.

Donnerbart:
:+1:

pveentjer:
Thanks. I'll fix the typo in a direct push.

Donnerbart:
Already done :)

PR #7700, opened 2016-03-09
Moved the spi partition classes to spi

pveentjer:
All SPI functionlity is part of com.hazelcast.spi to clearly separate the SPI from the API. 

These interfaces are introduced/pulled out in 3.7 so will not cause any incompatibilities.


jerrinot:
:+1:

Donnerbart:
`IPartitionLostEvent` was renamed from `InternalPartitionLostEvent` and annotated with `@PrivateAPI` in 4bd93d4f6752ab0bfe5a303d3b9267fc4cb629d9, so this move should be fine. `IPartitionService` was renamed from `InternalPartitionService`, but not marked as private in the same commit. Nevertheless this is a concern of the first commit.

Note: There was a remark that those `I` names were not meant to be final for the release. Any suggestions for better names?

:+1:

pveentjer:
No suggestions yet. But if times comes and we don't have better name, I'll rename them to Extended... 

The I naming was introduced to subclass an interface like Map/Queue etc. But in this case this naming convention shouldn't be applied here since IPartitionService doesn't extend PartitionService.

PR #7702, opened 2016-03-09
Organized imports in Spring module.

Donnerbart:
I just used the "organize imports" function of IDEA to clean up the imports of Hazelcast. I think that we don't have to discuss the outcome in the Java files much.

But I didn't know that XML files are also cleaned up by this. Do we want those changes?

### DIFF
--- a/hazelcast-spring/src/main/java/com/hazelcast/spring/context/SpringAware.java
+++ b/hazelcast-spring/src/main/java/com/hazelcast/spring/context/SpringAware.java
@@ -16,12 +16,12 @@
 
 package com.hazelcast.spring.context;
 
-import java.lang.annotation.Target;
-import java.lang.annotation.Retention;
-import java.lang.annotation.Inherited;
 import java.lang.annotation.Documented;
-import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
  * Annotates a class for injection of Spring dependencies.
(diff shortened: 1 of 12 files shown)


hazelcast-spring/src/test/resources/com/hazelcast/spring/replicatedmap/replicatedMap-applicationContext-hazelcast.xml
```diff
@@ -18,11 +18,10 @@
 <beans xmlns="http://www.springframework.org/schema/beans"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns:hz="http://www.hazelcast.com/schema/spring"
-       xmlns:cache="http://www.springframework.org/schema/cache"
        xsi:schemaLocation="http://www.springframework.org/schema/beans
 		http://www.springframework.org/schema/beans/spring-beans-2.5.xsd
-		http://www.springframework.org/schema/cache
-		http://www.springframework.org/schema/cache/spring-cache.xsd
+
```
pveentjer: Why these white lines?

Donnerbart: As the guys at IntelliJ. That's the default from their import organizer. I'm not a 100% sure why it touched those XML files anyway, but there was no further option to select files.

pveentjer: Can you at least revert the xml changes? XML != import

Donnerbart: I would bet there is a good reason for this. Maybe those removed beans are unused. But I'll make a separate PR from it and will watch out that no XML files are touched if I do this for the main module.

Donnerbart:
I reverted the changes in the XML files, which need some more investigation. The remaining Java files should not be worth discussing a lot.

pveentjer:
:+1:

PR #7704, opened 2016-03-09
Handled response equal to null case in ClientDeledatingFuture

gurbuzali:
The issue manifests itself when the response equals to null and user calls `future.get` inside the callback.
We guard the `resolving response block` with `response == null`, and this causes to read the `ClientMessage` twice. Fix is assigning an initial `NIL` value (reused `mutex` object ) to response.


gurbuzali:
run-lab-run

gurbuzali:
unrelated test failure
```
com.hazelcast.mapreduce.aggregation.MultiMapBaseAggregationTest.com.hazelcast.mapreduce.aggregation.MultiMapBaseAggregationTest
```
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/2121/

gurbuzali:
run-lab-run

gurbuzali:
run-lab-run

jerrinot:
@gurbuzali: is there a backport for this?

gurbuzali:
I'll prepare it

PR #7711, opened 2016-03-09
[Backport] Node info is logged before `NodeExtension:beforeStart`

kobalski:
Node info is logged before `NodeExtension:beforeStart`  so that if sth goes wrong during `beforeStart` user can see build info.
Also fixes https://github.com/hazelcast/hazelcast-enterprise/issues/714


pveentjer:
:+1:

PR #7716, opened 2016-03-10
Fixes invalidation message sending problem to client near-caches upon…

ahmetmircik:
… member side putAll/loadAll call

closes https://github.com/hazelcast/hazelcast/issues/7715

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
@@ -489,6 +489,35 @@ public class ClientMapNearCacheTest {
         });
     }
 
+
+    @Test
+    public void testMemberLoadAll_invalidates_clientNearCache() {
+        int mapSize = 1000;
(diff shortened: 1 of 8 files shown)


Donnerbart:
Do I understand this fix correctly?
If only clients have a NearCache configured the `isNearCacheEnabled()` returned `false` (since it just checks the member configuration) and no `invalidationKeys` were generated. Therefor you changed this to `isInvalidationEnabled()`, which checks if member invalidation is configured OR a listener for invalidation is registered (which seems to be the clients then).

You also renamed `isNearCacheEnabled()` to `hasMemberNearCache()` to point out what is really checked by that method.

If this is the fix you get a big :+1: from me for the improvement and the new tests :)

jerrinot:
oh, I'd love to have explanations similar the one @Donnerbart provided as a part of changeset. It makes reviewing so much easier and it's also good for history keeping.

ahmetmircik:
@Donnerbart you are absolutely correct.

@jerrinot i thought that it is self-explanatory.

jerrinot:
:+1:

jerrinot:
@ahmetmircik: this PR is rather simple, but @Donnerbart's comment still makes review process easier.

PR #7717, opened 2016-03-10
Fixes invalidation message sending problem to client near-caches upon…

### LINKED ISSUE #7715: IMap doesn't send invalidations to client near-caches upon member side putAll/loadAll

ahmetmircik:
… member side putAll/loadAll call

closes #7715


Donnerbart:
Backport of #7716 which is reviewed and merged. Only difference is the additional tests for the legacy client. I think this can be directly merged once green.

:+1:

Issue #6248, opened 2015-09-18
InvocationFuture long running operations


pveentjer:
The invocation mechanism for blocking operations relies on a periodic timeout so that the operation gets retried. To prevent the calling thread (future.get thread) to wait indefnitely, it will periodically ask the isstillrunning service if the operation got lost. 

To prevent a thread doing this check, the CONTINUE_WAIT is set on the future and the thread unsets it and skips the is still running. The problem is when there are multiple threads doing the future.get. One thread will be lucky and see the CONTINUE_WAIT. The other thread might by unlucky since the other thread has unset the CONTINUE_WAIT value and therefor this thread will continue with the is still running check. Since blocking operations always return false, this thread will conclude that the operation isn't running anymore and therefore it will throw an OperationTimeout.

Effectively it means that an InvocationFuture in Hazelcast can't be shared between multiple threads for blocking operations.

pveentjer:
This issue has been resolved in 3.7

Issue #4931, opened 2015-03-28
Deadmans switch redesign


pveentjer:
The IsStillRunningService checks if a remote operation is still running.

The design is flawed for multiple reasons:
1 you can't see operations pending on a queue.
2 for every thread doing a future.get, there will be potentially a IsStillRunningOperation being executed. So if a cluster is slow and operations don't get response, we'll throw on more isStillRunning operations into the system. Also if there are 10 threads sharing the same future and doing a call on awaitResponse, you get 10 IsStillRunningOperations even though there is only a single operation that needs to be checked.
3 the Invocation.waitForResponse is very complicated an unmaintainable.
4 there is a difference between synchronous calls and asynchronous calls. Synchronous calls do check 'isStillRunning', but asynchronous calls do not. If this logic is pulled outside of the 'awaitResponse', then this part of the difference is fixed.

issue 2 can be fixed by letting the remote side once and a while send a list of all running operations back to the caller
issue 3 is fixed when issue 2 is fixed. Instead of letting the waitForResponse determine when an operation is send, the system just receives a 'packet' containing all call id's and for every callid that doesn't get conformation, we can terminate it. So if there are n calls doing a future.get, instead of having potentially n IsStillRunningOperations, only a single packet with 10 callid's is being returned. So the remote member send this packet periodically to the invoking member.
issue 4 is automatically fixed when the logic is externalized from the future.get.

issue 1 is much more difficult to fix.

pveentjer:
This ticket can be closed since the operation timeout logic is improved in the above described manner.

Issue #4410, opened 2015-01-13
Optimization: IsStillRunning simplification


pveentjer:
Instead of each operation asking if something is still running; let the other system every x seconds return a list of invocations that have been running for more than y seconds. 

The current approach causes additional pressure on a system under pressure since:
- no additional operations need to be executed while a system already is executing operations slowly.
- data can be packet a lot more efficient since a single result could contains just a bunch of longs (the callid's from an invoking machine). So less io/serialization pressure.
- simplification of the invocation logic; currently it has too many responsibilities and the flow is very hard to understand.

So if member1 has 2 operations executing on member 3, and member 2 has 5 operations executing on member 3. And these are all operations that have been running longer than the given limit, then member 3 sends 2 callid's back to member1 and 5 callid's back to member 2. Probably we need to have a dedicated packet id for this.

member 3 can do a periodic check over all operation-threads, and see which operations have been running for a certain amount of time. And to each member that has long running operations, it sends the list of 'their' call id's.

On the invoking side, the system can then timeout operations that were not mentioned in a 'isstillrunning' packet for a too long time.

pveentjer:
This will also help us to get rid of the difference between the future.andThen and the future.get.

pveentjer:
One of the other problems with the current design is that we are only checking what is currently executing, not what is in the pipeline:
- in the connection write-queue (operation and response)
- operation queue.

So it could be that is still running returns false, because the operation has not yet been picked up yet.

jerrinot:
that's the reason I'm sceptical when it comes to usefulness of the while IsStillRunning vehicle. it covers only a very small portion of possibly delays and IMHO it's not worth the extra complexity.

pveentjer:
I would also like to get rid of it completely. We have the heart beat that detect when a node is not usable. So we can lift on that mechanism afaik.

pveentjer:
is still running operation has been removed and replaced by Operation timeout logic cleanup.

PR #7719, opened 2016-03-10
Operation timeout logic cleanup

pveentjer:
The IsStillRunning functionality has been replaced by an Operation heartbeat. The OperationService/WaitNotifyService/DistributedExecutorService are frequently scanned for all executing operations to include the call-id. Then a packet is send back including all callid's of the operation each member is executing on another member. Any operation that fails to receive such an operation heartbeat for an extended period will timeout.

The old approach was problematic:
- very complicated logic
- unreliable since there is a very short timewindow to ask if an operation is still running. If that timewindow expires (e.g. system has a lot of work to do) then the caller can falsely assume the operation isn't running. The new approach doesn't rely on operations. Operation-control-packet is processed outside of the regular operation-threads.
- a future with get doesn't trigger timeout checking
- Also the logic where the 'CONTINUE_WAIT' on the future was set is removed due to its complexity and problems. E.g.if multiple threads do a get in the same future on a blocking operation, then one of the threads will win, the rest will fail with an OperationTimeoutException.

fixes:
https://github.com/hazelcast/hazelcast/issues/6248
https://github.com/hazelcast/hazelcast/issues/4931
https://github.com/hazelcast/hazelcast/issues/4410


pveentjer:
:100:

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -372,18 +388,14 @@ void notifyError(Object error) {
...
-        //when the backups complete, the response will be send by the last backup or backup-timeout-handle mechanism kicks on
+        // if a regular response came and there are backups, we need to wait for the backs.
+        // when the backups complete, the response will be send by the last backup or backup-timeout-handle mechanism kicks on
 
         if (expectedBackups > backupsCompleted) {
             // So the invocation has backups and since not all backups have completed, we need to wait.
             // It could be that backups arrive earlier than the response.
 
-            this.pendingResponseReceivedMillis = Clock.currentTimeMillis();
+            this.pendingResponseReceivedMillis = currentTimeMillis();
```
Donnerbart: I would not use a static import of `System.currentTimeMillis()` here. We also have `Clock.currentTimeMillis()`, so you don't know which clock is used here.

Since you also used `Clock.currentTimeMillis()` in this class: Are you sure the correct clock was used in both places? I have no clue when to use which one, it just looks suspicious to have them both so close together.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -372,18 +388,14 @@ void notifyError(Object error) {
     }
 
     void notifyNormalResponse(Object value, int expectedBackups) {
-        if (value == null) {
-            value = NULL_RESPONSE;
-        }
-
-        //if a regular response came and there are backups, we need to wait for the backs.
-        //when the backups complete, the response will be send by the last backup or backup-timeout-handle mechanism kicks on
+        // if a regular response came and there are backups, we need to wait for the backs.
```
Donnerbart: Minor grammar improvements (hopefully)
```
// if a regular response came and there are backups, we need to wait for them.
// the response will be sent by the last completed backup or when the backup-timeout-handle mechanism kicks in
```

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
```diff
@@ -34,6 +38,8 @@
  */
 public interface OperationExecutor {
 
+
+
```
Donnerbart: I assume you wanted to add something here. Now there is just empty space :)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -522,7 +489,75 @@ private void handleRetry(Object cause) {
...
+        }
+    }
+
+    HeartbeatTimeout detectTimeout(long heartBeatTimeoutMs) {
+        if (pendingResponse != VOID) {
+            // if there is a response, then we won't timeout.
+            return NO__RESPONSE_AVAILABLE;
+        }
+
+        long currentTimeMs = currentTimeMillis();
```
Donnerbart: Same here, it's ambiguous if `System` or `Clock` is used as source (and which one should be used).

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
```diff
@@ -17,52 +17,40 @@
...
 import static com.hazelcast.util.ExceptionUtil.fixAsyncStackTrace;
 import static com.hazelcast.util.Preconditions.isNotNull;
-import static java.lang.Math.min;
 
 /**
  * The InvocationFuture is the {@link com.hazelcast.spi.InternalCompletableFuture} that waits on the completion
  * of a {@link Invocation}. The Invocation executes an operation.
  *
- * @param <E>
+ * m * @param <E>
```
Donnerbart: I think that `m *` was added accidentally.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry_NotifyTest.java
```diff
@@ -191,7 +192,12 @@ public void timeoutResponse() {
         long callId = invocation.op.getCallId();
         invocationRegistry.notifyCallTimeout(callId, null);
 
-        assertNull(invocation.future.join());
+        try {
+            assertNull(invocation.future.join());
+            fail();
+        }catch (OperationTimeoutException expected){
```
Donnerbart: Missing spaces.

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_BlockingTest.java
```diff
@@ -0,0 +1,519 @@
...
+        final ExecutionCallback callback = mock(ExecutionCallback.class);
+        future.andThen(callback);
+
+        // and we eventually expect to fail with an OperationTimeoutException
+        assertFailsEventuallyWithOperationTimeoutException(callback);
+    }
+
+    @Test
+    public void async_whenOperationTimeout() {
+//        Config config = new Config();
```
Donnerbart: Is this test not compiling? If it does I would prefer `@Ignore`, so the test shows up in the statistics and isn't forgotten.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -37,6 +37,7 @@
 import static com.hazelcast.internal.util.counters.SwCounter.newSwCounter;
 import static com.hazelcast.spi.Operation.CALL_ID_LOCAL_SKIPPED;
 import static com.hazelcast.spi.OperationAccessor.setCallId;
+import static java.lang.System.currentTimeMillis;
```
Donnerbart: Same here, I'm not a fan of hiding which clock source is used.

Donnerbart:
I lost a bit track of what's going on after 1/3 of the review. I'm missing a lot of context to do a proper review. I understood the new high level concept, but I'm missing the whole implementation details of the old system to see what is changed and what is still there :( Will give it another try tomorrow.

What about a longer Simulator stress test before we merge this PR? I think we don't have any mysterious operation timeouts at the moment. So a stress test should show if the new system is not producing any false positives.

pveentjer:
@Donnerbart it will run for 24 hours using simulator.

Donnerbart:
Peter explained some details to me today. I still have to look more at the implementation, but the new concept looks very good to me. It's less complex and more scaleable to me (no more `IsStillRunningOperation` flooding).

I think it's worth mentioning that the introduced "operation heartbeat" has nothing to do with the member/node heartbeat (nevertheless the name fits for me). Also this PR doesn't fix that pending operations are not monitored. This could be done later (it's just an additional scanner on the executing side), but it's quite a big change in behavior.

So the concept already gets a thumbs up from me. The implementation will need more review time :)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
```diff
@@ -218,126 +169,76 @@ public E getSafely() {
...
     @Override
-    public E get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-        Object unresolvedResponse = waitForResponse(timeout, unit);
-        return (E) resolveOrThrow(unresolvedResponse);
-    }
-
-    private Object waitForResponse(long time, TimeUnit unit) {
-        if (responseAvailable(response)) {
-            return response;
+    public E get() throws InterruptedException, ExecutionException {
```
jerrinot: why not just delegate to `get(timeToEndOfTheUniverse, ..)` ? 
it's pretty much the same logic.

pveentjer: different logic and different exceptions.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -47,134 +61,178 @@
...
+    }
+
+    @Override
+    public void handle(Packet packet) {
+        scheduler.execute(new ProcessOperationHeartbeatsTask(packet));
     }
 
     /**
-     * The InvocationMonitorThread iterates over all pending invocations and sees what needs to be done
+     * The MonitorTask iterates over all pending invocations and sees what needs to be done
```
jerrinot: could you clarify the contract? "what needs to be done" is a very vague responsibility. 
image

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -47,134 +61,178 @@
...
+        this.slowInvocationThresholdMillis = slowInvocationThresholdMillis();
+
+        this.scheduler = newScheduler();
+
+        nodeEngine.getMetricsRegistry().scanAndRegister(this, "operation.invocations");
+    }
+
+    private ScheduledExecutorService newScheduler() {
+        // the scheduler is configured with a single thread.
+        ScheduledExecutorService scheduler = new ScheduledThreadPoolExecutor(1, new ThreadFactory() {
```
jerrinot: this seems wrong. 
the `coreSize` is 1. but the max size is unlimited. it will create a new thread if all existing threads are busy.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
jerrinot: ok. I see the exception now.

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -41,6 +41,7 @@
     public static final int FLAG_EVENT = 1 << 2;
     public static final int FLAG_URGENT = 1 << 4;
     public static final int FLAG_BIND = 1 << 5;
+    public static final int FLAG_OP_CONTROL = 1 << 6;
```
jerrinot: please specify what's the meaning of this flag.

Donnerbart: If I'm not wrong this is a packet which contains the callIds from running/living operations. Since this concept is called "operation heartbeat" I would call this `FLAG_OP_HEARTBEAT` as well.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
pveentjer: Good one! :)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -396,32 +402,34 @@ void notifyNormalResponse(Object value, int expectedBackups) {
...
             logger.finest("Call timed-out either in operation queue or during wait-notify phase, retrying call: "
                     + toString());
         }
 
-        if (op instanceof BlockingOperation) {
-            // decrement wait-timeout by call-timeout
-            long waitTimeout = op.getWaitTimeout();
-            waitTimeout -= callTimeout;
-            op.setWaitTimeout(waitTimeout);
+        if (!(op instanceof BlockingOperation)) {
```
jerrinot: is this a change from the current behaviour?

pveentjer: Correct. With the old behavior the operation is retried after a call get rejected because it failed to start within the given time window. In the new approach it is end of story; as soon as the remote side has rejected your call because it was too late; it isn't retried. It simplifies the flow + makes core easier to test

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
jerrinot: for future reference: I was wrong here. This is a contract of `ScheduledThreadPoolExecutor`:
```
 * <p>While this class inherits from {@link ThreadPoolExecutor}, a few
 * of the inherited tuning methods are not useful for it. In
 * particular, because it acts as a fixed-sized pool using
 * {@code corePoolSize} threads and an unbounded queue, adjustments
 * to {@code maximumPoolSize} have no useful effect. Additionally, it
 * is almost never a good idea to set {@code corePoolSize} to zero or
 * use {@code allowCoreThreadTimeOut} because this may leave the pool
 * without threads to handle tasks once they become eligible to run.
```

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -494,20 +459,87 @@ private void handleRetry(Object cause) {
...
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    HeartbeatTimeout detectTimeout(long heartbeatTimeoutMillis) {
+        if (pendingResponse != VOID) {
+            // if there is a response, then we won't timeout.
+            return NO__RESPONSE_AVAILABLE;
```
ahmetmircik: This is confusing, even though there is a response (comment says that), method returns `NO__RESPONSE_AVAILABLE`, this lowers code readability and also can cause erroneous cases for the future maintainers due to the naming.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -494,20 +459,87 @@ private void handleRetry(Object cause) {
...
+                : lastHeartbeatMillis + heartbeatTimeoutMillis;
+
+        if (heartBeatExpirationTimeMillis >  Clock.currentTimeMillis()) {
+            return NO__HEARTBEAT_TIMEOUT_NOT_EXPIRED;
+        }
+
+        return YES;
+    }
+
+    enum HeartbeatTimeout {
```
ahmetmircik: This `HeartbeatTimeout ` enum confused me. Normally there should be 2 state YES or NO. Adding causes to their naming makes it hard to understand at first glance. A possible solution, completely remove `HeartbeatTimeout` and just return true or false from the `detectTimeout ` method and use short comments to explain the logic (comments already there in your case).

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_BlockingTest.java
```diff
@@ -0,0 +1,545 @@
...
+
+        // then we register our callback
+        final ExecutionCallback callback = mock(ExecutionCallback.class);
+        future.andThen(callback);
+
+        // and we eventually expect to fail with an OperationTimeoutException
+        assertFailsEventuallyWithOperationTimeoutException(callback);
+    }
+
+//    @Test
```
ahmetmircik: leftover ?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_BlockingTest.java
```diff
@@ -0,0 +1,545 @@
...
+        // then we are going to send the invocation and share the future by many threads.
+        int thisThreadId = 2;
+        final InternalCompletableFuture<Object> future = opService.createInvocationBuilder
+                (null, new LockOperation(new InternalLockNamespace(key), nodeEngine.toData(key), thisThreadId, -1, -1), partitionId)
+                .setCallTimeout(callTimeout)
+                .invoke();
+
+
+//
+//        // lets do a very long wait so that the heartbeat/retrying mechanism have kicked in.
```
ahmetmircik: leftover?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
pveentjer: The problem is then the testing becomes very problematic because you don't know why it has or has not timed out.

pveentjer: Apart from replacing to boolean; do you have any naming suggestions that would prevent confusion?

ahmetmircik: What about this one?
``` 
enum HeartbeatTimeout {
        NO_TIMEOUT__CALL_TIMEOUT_DISABLED,
        NO_TIMEOUT__RESPONSE_AVAILABLE,
        NO_TIMEOUT__HEARTBEAT_TIMEOUT_NOT_EXPIRED,
        NO_TIMEOUT__CALL_TIMEOUT_NOT_EXPIRED,
        TIMEOUT
  }

 ```

gurbuzali: :+1:

ahmetmircik:
:+1: (with minor comments)

pveentjer:
Thanks. I'll fix them now. We still need to wait for @gregrluck before it can be merged.

jerrinot:
:+1:

PR #7720, opened 2016-03-10
Parameterized test improvement

mdogan:
- Removed HazelcastTestRunner which duplicates JUnit's Parameterized
feature.
- Implemented JUnit's ParametersRunnerFactory to create serial
or parallel runner depending on test's category.
- Removed RunParallel annotation. Tests marked as ParallelTest
category are executed with parallel runner.

With this change, now it's possible to execute single parameterized
test on IDEA.

EE counterpart: https://github.com/hazelcast/hazelcast-enterprise/pull/813


gurbuzali:
:+1:

jerrinot:
:+1: excellent work, thanks!

ahmetmircik:
@mdogan i like this, thanks, it was a pain to run single test method.

Donnerbart:
Hmm, I don't know if this is done 100% correctly. We have two types of parallel execution:

- execute test methods in parallel (with the parallel runner)
- execute test class in parallel (in different JVMs)

The annotation `ParallelTest` is a hint for the second one, but you use it to distinct between serial and parallel runner (the first one), don't you?

mdogan:
@Donnerbart: I know `ParallelTest` and `RunParallel` were used for different meanings. But to me this distinction seems complicated, currently none of the tests (or is there any?) need this two distinct functionality. That's why I merged them.

I can revert `RunParallel` annotation removal, if all agree on that.

Donnerbart:
I'm not sure, I just saw this possible issue and wanted to comment on it, before we hit the merge button.

I'd love to get rid of `RunParallel`, but without it I don't see a possibility to run test methods in parallel, without running the whole class in parallel. If we don't need this, I'm totally fine with the change. Maybe we should clarify this in the JavaDoc then, so that with `HazelcastParametersRunnerFactory` you either run the class in isolation (single JVM, serial method execution) or fully parallelized (multiple JVMs, parallel method execution).

eminn:
:+1:

bilalyasar:
i am not sure but our new repos ( wm,hibernate,jclouds..) may need small fix.  ( `If they use HazelcastTestRunner.java` or `RunParallel.java`)

mdogan:
@Donnerbart: added more details to javadoc as you suggested: https://github.com/hazelcast/hazelcast/commit/f6c2c4dcb7d91e0b5fd818de0bc3a5f6bb20c484

We can re-introduce old behaviour easily if it's required some day.

PR #7721, opened 2016-03-10
Fixed wildcard suppression on AbstractClientInternalCacheProxy

Donnerbart:
* Fixed accidental wildcard suppression on `AbstractClientInternalCacheProxy`.
* Cleanup of covered CheckStyle issues.

The suppression.xml uses regex, so the original suppression `checks="...|...|MethodCount|"` ended with an `OR anything`, which resulted in a wildcard suppression. We also had two lines of suppresions for this class.

I removed some warnings related to missing generics usage. I resolved some method length issues by pulling out sync/async methods for too large methods. This also resolved a nested-if-depths issue. The last big issue was the naming of `static final` constants, which should be uppercase.


pveentjer:
:+1:

PR #7724, opened 2016-03-11
Removed ExceptionAction.CONTINUE_WAIT

pveentjer:
It is not being used.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -354,9 +354,6 @@ public abstract class Invocation implements OperationResponseHandler, Runnable {
         }
 
         switch (onException(cause)) {
-            case CONTINUE_WAIT:
-                handleContinueWait();
-                break;
             case THROW_EXCEPTION:
                 notifyNormalResponse(cause, 0);
                 break;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -398,9 +398,6 @@ void notifyError(Object error) {
         }
 
         switch (onException(cause)) {
-            case CONTINUE_WAIT:
-                handleContinueWait();
```
jerrinot: isn't this method now not used too?

pveentjer: This code is not used at all.

jerrinot: so shouldn't you remove the `handleContinueWait()` method too in this PR?

pveentjer: It will be fixed any second. Any other comments?

jerrinot:
what's the impact on SPI users? what's a replacement?

pveentjer:
There is no replacement. If you are using it; bad luck. The mechanism is just not supported and provides no value with the operation timeout logic replacement I'm working on.

jerrinot:
couldn't we deprecate it and treat it as no-op? as that's what it is (once the new timeout is in place)
this way you backward maintain compatibility + avoid code complexity (apart from extra enum)

pveentjer:
We are breaking api regarding operation call timeouts anyway. TraceableOperation for example is removed because we don't trace any operations. So why not get this part fixed as well.

In the SPI we can break stuff and this is rarely used.

jerrinot:
:+1: 
I don't have a strong opinion here. Please wait for a 2nd reviewer before merging the PR.

mtopolnik:
:+1:

PR #7727, opened 2016-03-11
Minor improvement in HashSlotArray API

mtopolnik:
There were a few methods missing which were introduced on the Hot Restart suicide branch.

Also optimized `markAllUnassigned` method by reducing amount of memory written. Now, instead of zeroing out all memory, just the "free slot" markers are written.


jerrinot:
:+1:

mtopolnik:
Minor PR -- merging now.

PR #7730, opened 2016-03-11
Fix bug in HashSlotArrayBase

mtopolnik:
auxAllocateAndAdjustFields read from a free()'d location


mtopolnik:
Merging immediately (build reached the test phase) because the code in question is only exercised from EE tests.

Donnerbart:
So the fix is to read `size()` before the `move()` operation. I don't know the details of the methods, but that looks very reasonable :)

:+1:

mtopolnik:
Yes, `size()` reads from the backing native memory block, and `move()` frees the block after copying it somewhere else.

Donnerbart:
Good finding.

mtopolnik:
Jenkins build found it in the form of a segfault dump file :)

PR #7731, opened 2016-03-11
Two improvements in memory access code

mtopolnik:
1. Add a regression test for the recently fixed segfault bug in HashSlotArrayBase.
2. Add `byte[]`-oriented methods to `MemoryAccessor` to allow the transfer of data between accessor's address space and plain `byte[]`.


toktarev:
:+1: thanks a lot

pveentjer:
:+1:

PR #7734, opened 2016-03-13
Fixes invocation stracktrace problems

### LINKED ISSUE #6247: InvocationFuture race: race on exception
When an InvocationFuture processes a  thrown exception e.g one from the operation.run, it will rewrite the stacktrace so that both local and remote stacktraces are integrated. 

```
 private Object resolveApplicationResponse(Object unresolvedResponse) {
        ...

        if (response instanceof Throwable) {
            Throwable throwable = ((Throwable) response);
            if (invocation.remote) {
                fixRemoteStackTrace((Throwable) response, currentThread().getStackTrace());
            }
            return throwable;
        }

        return response;
    }
```

The problem is that this exception-instance can be shared between multiple threads since the future can be shared between multiple threads. So you have multiple threads modifying a stacktrace of the same exception instance.

Solution: clone the exception first.

Edit: for a remote thrown exception this probably isn't an issue since each thread will get a Data and deserialize that data. Not completely sure about this. But for local exception is could certainly be an issue.

pveentjer:
Fixes stacktrace being modified by multiple threads by cloning the exception.
Fixes that the stacktrace is only rewritten for remote calls; but fails to do so for local calls. So stacktraces are now always rewritten.

fixes #6247

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
@@ -35,7 +35,6 @@ public class MemberLeftException extends ExecutionException implements Retryable
     public MemberLeftException() {
     }
 
-
     public MemberLeftException(String message) {
         super(message);
     }
@@ -47,6 +46,7 @@ public class MemberLeftException extends ExecutionException implements Retryable
 
     /**
      * Returns the member that left the cluster
+     *
      * @return the member that left the cluster
      */
     public Member getMember() {
@@ -56,24 +56,26 @@ public class MemberLeftException extends ExecutionException implements Retryable
     private void writeObject(ObjectOutputStream out) throws IOException {
         out.defaultWriteObject();
 
-        Address address = member.getAddress();
-        String host  = address.getHost();
-        int port = address.getPort();
-
-        out.writeUTF(member.getUuid());
-        out.writeUTF(host);
-        out.writeInt(port);
-        out.writeBoolean(member.isLiteMember());
+        if (member == null) {
+            out.writeBoolean(false);
+        } else {
+            out.writeBoolean(true);
+            out.writeUTF(member.getUuid());
+            out.writeUTF(member.getAddress().getHost());
+            out.writeInt(member.getAddress().getPort());
+            out.writeBoolean(member.isLiteMember());
+        }
     }
 
     private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
         in.defaultReadObject();
 
-        String uuid = in.readUTF();
-        String host = in.readUTF();
-        int port = in.readInt();
-        boolean liteMember = in.readBoolean();
-
-        member = new MemberImpl(new Address(host, port), false, uuid, null, null, liteMember);
+        if (in.readBoolean()) {
+            String uuid = in.readUTF();
+            String host = in.readUTF();
+            int port = in.readInt();
+            boolean liteMember = in.readBoolean();
+            member = new MemberImpl(new Address(host, port), false, uuid, null, null, liteMember);
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


pveentjer:
Can't be backported due to wire incompatibility of the MemberLeftException

hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
```diff
@@ -56,24 +56,26 @@ public Member getMember() {
...
 
-        Address address = member.getAddress();
-        String host  = address.getHost();
-        int port = address.getPort();
-
-        out.writeUTF(member.getUuid());
-        out.writeUTF(host);
-        out.writeInt(port);
-        out.writeBoolean(member.isLiteMember());
+        if (member == null) {
```
jerrinot: I don't understand this part. is this another bug this PR is fixing? NPE waiting to happen?

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
```diff
@@ -28,7 +28,10 @@
  */
 public final class ExceptionUtil {
 
-    private static final String EXCEPTION_SEPARATOR = "------ End remote and begin local stack-trace ------";
+    /**
+     * Separator for local/remote stacktrace.
+     */
+    public static final String EXCEPTION_SEPARATOR = "------ End remote and begin local stack-trace ------";
```
jerrinot: a nitpick: I always have to think hard to tell what part of the stacktrace is the local part and what is the remote part. Perhaps the separator could indicate this?

hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
pveentjer: It is a bug in the serialization that showed up when cloning.

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
pveentjer: Create a ticket and I'll have a look at it today.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
```diff
@@ -380,11 +367,21 @@ private Object resolveApplicationResponse(Object unresolvedResponse) {
...
+                // Exception needs cloning since stacktrace is modified, otherwise same exception shared between threads.
+                Throwable clonedThrowable = serializationService.toObject(serializationService.toData(throwable));
+                fixRemoteStackTrace(clonedThrowable, Thread.currentThread().getStackTrace());
+                return clonedThrowable;
+            } catch (HazelcastSerializationException e) {
+                // It can happen that the exceptions fails cloning, in that case return the original unmodified exception
+                operationService.logger.warning("Failed to clone exception", e);
+                return response;
+            } catch (HazelcastInstanceNotActiveException e) {
+                // In case of a shutdown, lets return the HazelcastInstanceNotActiveException
```
jerrinot: you are losing the original exception response.

hazelcast/src/main/java/com/hazelcast/util/ExceptionUtil.java
jerrinot: https://github.com/hazelcast/hazelcast/issues/7744 thanks!

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
pveentjer: On purpose. Because the stacktrace can't be modified; so the original exception would come out of thin air.

jerrinot: in the previous case (`HazelcastSerializationException`) you are returning the original unmodified exception anyway. is it really that different from this case? in both cases there is a reason why cloning can't be done.

mdogan:
needs rebase

mdogan:
@pveentjer: are you going to rebase this or shall we close the PR?

Issue #7712, opened 2016-03-09
Reducing checktyle suppresions


bontorhumala:
Hi, some friends and I are looking to contribute to Hazelcast as part of university project - see the [twitter account](https://twitter.com/delftswa) and last year [project](http://delftswa.github.io/). We noted that Hazelcast considers checkstyle as technical debt. We are not looking at refactoring codes. Mostly simpler stuff like adding Javadoc, remove modifier redundancy, ordering of variables - constructor - methods, adding a header to files, etc.

Would you guys appreciate this kind of fixes i.e reduce the number of checkstyle error/suppressions? Are there any concerns? Thanks

Donnerbart:
We like all kinds of contribution and reducing technical debt is a valuable task.

The biggest risk I see is that your work could interfere with our release cycle. Normally we do this kind of cleanups early, directly after a release. In this way we don't mix cleanup and feature PR at the end of the development phase.

For our main project this window is closing soon. But we recently pulled out a lot of modules as plugins into their own repositories (http://hazelcast.org/plugins/). I'll check with our Integration Team if we could put you on some of those projects.

May I ask how many people you are? Could your group work together on a single PR or do you need work for several groups?

bontorhumala:
To clear things us, our group is the only one working on Hazelcast as other groups are working in other open source projects. There are 4 of us in this group. We could work together on a single PR. When will the window be closed? Maybe the deadline fits perfectly to our deadline :)

Donnerbart:
There is no exact date yet, but there is already a huge amount of cleanup PRs pending for Hazelcast 3.7 (mostly improvements of what is public and what is private API). I'll bring your offer up in our standup tomorrow. I think we should be able to find a task for your group.

When is your deadline? Since we also need time to review we should pick something suitable for your available time :)

bontorhumala:
The project deadline is the first week of April but now that you are offering specific task (which is even better for us), we would really appreciate it if you can give the task(s) by tomorrow or friday so that we can work on it as soon as possible :) Thanks for bringing it up tomorrow!

The project is mainly about analysing the project, but contributing is a small part of it as a way to get a better feel of the project. We have started working on contribution for the past 2-3 weeks and got 1 merged - https://github.com/hazelcast/hazelcast/pull/7577. Currently, we are looking to give more contributions. Thats why we aim to finish our contributions by Tuesday next week.

Donnerbart:
[...]

PR #7736, opened 2016-03-13
Removing JavaDoc checkstyle suppressions for hazelcast-client

Jorik0:
Hello, I've removed the two javadoc suppresions for hazelcast-client and added the necessary javadoc comments to fix the resulting checkstyle errors. (as discussed in issue #7712 )

I tried to keep the comments in the same style as the rest of the javadoc comments in hazelcast-client. But it could be that they are not in-depth enough. Please tell me if this is the case, as then I will add some more info to them.

I was also a bit confused what to do about the "@ali" tag, for now I just removed it.

PS: Checkstyle seems to be giving some warning now about TransactionException. I don't know what causes this, in any case I haven't changed anything related to it.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -93,9 +93,6 @@
     <suppress checks="MethodCount" files="com/hazelcast/osgi/impl/HazelcastOSGiInstanceImpl"/>
 
     <!-- Client -->
-    <!-- TODO: these JavaDoc issues need to be addressed -->
-    <suppress checks="JavadocMethod" files="com/hazelcast/client/"/>
-    <suppress checks="JavadocType" files="com/hazelcast/client/"/>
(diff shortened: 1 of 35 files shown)


Donnerbart:
The `@ali` tag is most probably a wrong collapse of `@author ali`. I don't know if we have a consensus yet of keeping or removing the `@author` tags in the project, so just keep it as it is for now.

That `TransactionException` thing happens sometimes if there is a CheckStyle error and the compilation is stopped in between. I never investigated this, but this is not related to your work. It's just important that in the end everything compiles again.

Thanks for the PR, I'll try to get someone from the client team to review the content of the docs. Technically the PR seems to be fine, thanks for that :)

sancar:
:+1:

Donnerbart:
:+1:

PR #7737, opened 2016-03-13
Specifying wildcard suppresions for Node.java

bontorhumala:
Hi, I found that you use wildcard to suppress checkstyle error in Node.java. I fixed some checkstyle error:
- private and public variable declaration order
- use a constant THREAD_SLEEP_DURATION_MS to remove MagicNumber error
- fix a line length violation

and I also made the suppresion specific (NPathComplexity, CyclomaticComplexity, ClassFanOutComplexity, etc) and remove the wildcard as discussed in https://github.com/hazelcast/hazelcast/issues/7712. Tested it, also tested it using findbugs and checkstyle.


Donnerbart:
Looks good to me, thanks for the PR!

:+1:

PR #7738, opened 2016-03-13
Fix some checkstyle errors in ConcurrentReferenceHashMap

bontorhumala:
Hi, I fixed some checkstyle errors in ConcurrentReferenceHashMap: 

- constructor declaration order in ConcurrentReferenceHashMap. The constructor for ConcurrentReferenceHashMap (around hundred lines) is previously located after other methods, which is against checkstyle DeclarationOrder rule. I removed them and then moved them just after declaration of instance variables
- remove two inner assignment inside if statement and moved them before the statement
- update suppresions.xml because I have fixed DeclarationOrder

This PR is done as suggested in https://github.com/hazelcast/hazelcast/issues/7712. I ran test against it, also did findbugs and checkstyle test

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -341,7 +341,7 @@
 
     <!-- Util -->
     <!-- written by Doug Lea  -->
-    <suppress checks="MagicNumber|FileLength|DeclarationOrder|RedundantModifier|InnerAssignment|NPath|Cyclomatic"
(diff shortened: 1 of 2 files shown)


pveentjer:
Thanks for you pr. But this is a class we didn't write and therefor we don't want to change it so that we can easily check if there are any changes being made by Doug.

PR #7739, opened 2016-03-13
Fix some checkstyle errors in ThreadLocalRandom

bontorhumala:
Hi, I found that you use wildcard to suppress checkstyle error in ThreadLocalRandom. I fixed some checkstyle error:

- Add brackets to if statements
- Repair some declaration order
- Fix variable naming pattern
- Adding/removing white spaces
- Split multiple variable declarations to several lines

and I also made the suppresion more specific (DeclarationOrder and MagicNumber) and remove the wildcard as discussed in https://github.com/hazelcast/hazelcast/issues/7712. Tested it, also tested it using findbugs and checkstyle.

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -353,7 +353,7 @@
     <suppress checks="NPathComplexity" files="com/hazelcast/util/executor/DelegatingFuture"/>
 
     <!-- written by Doug Lea  -->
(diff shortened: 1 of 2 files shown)


pveentjer:
Thanks for you pr. But this is a class we didn't write and therefor we don't want to change it so that we can easily check if there are any changes being made by Doug.

PR #7740, opened 2016-03-13
InvocationFuture method renaming to align with CompletableFuture

pveentjer:
Renamed InvocationFuture.set to complete
Renamed InternalCompletableFuture.getSafely to join. The getSafely has been deprecated. not removed, for compatibility reason.


jerrinot:
:+1:

PR #7743, opened 2016-03-14
Invocation receives OperationServiceImpl directly

pveentjer:
No need to go through indirection of NodeEngine to retrieve the
OperationServiceImpl


mtopolnik:
:+1:

PR #7745, opened 2016-03-14
Invocation servicename/partitionid/replica-index removal

pveentjer:
All this information is already available on the operation itself. So the operation is used as leading source of this information instead of copied.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -972,10 +972,6 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
         return serializationService.toObject(obj);
     }
 
-    protected Data toData(Object obj) {
-        return serializationService.toData(obj);
-    }
-
     protected Data toData(Object o, PartitioningStrategy partitioningStrategy) {
         return serializationService.toData(o, partitioningStrategy);
     }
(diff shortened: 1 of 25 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -972,10 +972,6 @@ public void executeOnEntriesInternal(EntryProcessor entryProcessor, Predicate pr
         return serializationService.toObject(obj);
     }
 
-    protected Data toData(Object obj) {
```
Donnerbart: Should this be part of this PR. Looks like a non-related cleanup (or I'm missing something).

Donnerbart:
Looks good to me.

:+1:

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/spi/OperationService.java
```diff
@@ -117,6 +117,15 @@
 
     <E> InternalCompletableFuture<E> invokeOnPartition(String serviceName, Operation op, int partitionId);
 
+    /**
+     * Executes an operation on a partition.
+     *
```
tombujok: Just a pickle: since there's no partitionId in the arguments of this method it may be worth mentioning how it know on which partition to invoke the operation on, like 
```Executes an operation on a partition. 
Partition that the operation is invoked on comes from Operation.getPartitoinId()`
```

tombujok:
:+1: (minor comment)

pveentjer:
Thanks tom. I'll fix the comment in a direct push

PR #7750, opened 2016-03-14
Make MemArrayQuickSorter's base address mutable

mtopolnik:
Mutable base address reduces object litter by allowing the same instance to be reused.


serkan-ozal:
:+1:

PR #7751, opened 2016-03-14
Removing wildcard checkstyle suppression for ClientMultiMapProxy

Jorik0:
Hello again, I've refined the suppression for ClientMultiMapProxy and added the param tags to it's javadoc comment.

As asked of us in issue #7712 these PRs are kept small on purpose. But I can imagine this is a bit too small, so if you'd prefer I can tackle all the wildcards under the Client comment in one go. Of course, if you prefer these small PRs I can continue like this.


Donnerbart:
Nice one :)

:+1:

PR #7752, opened 2016-03-14
Introduces custom eviction policy

ahmetmircik:


ahmetmircik:
run-lab-run

ahmetmircik:
run-lab-run

ahmetmircik:
run-lab-run

ahmetmircik:
```
Test Result (1 failure / -3)
com.hazelcast.spi.impl.operationservice.impl.Invocation_BlockingTest.async_whenOperationTimeout```

**Comment by ahmetmircik** [comment]
run-lab-run

File: hazelcast-spring/src/main/java/com/hazelcast/spring/HazelcastConfigBeanDefinitionParser.java [hunk 1]
```diff
@@ -645,11 +647,38 @@ public void handleMap(Node node) {
...
+
+            String implementation = implementationNode != null ? getTextContent(implementationNode) : null;
+            String className = classNameNode != null ? getTextContent(classNameNode) : null;
+
+            if (implementation != null) {
+                mapConfigBuilder.addPropertyReference("customEvictionPolicy", implementation);
+            } else if (className != null) {
+                className = checkHasText(className, "custom-eviction-policy `className` cannot be null or empty");
+                try {
+                    Comparator comparator = ClassLoaderUtil.newInstance(getClass().getClassLoader(), className);
```
serkan-ozal: `CustomEvictionPolicy` instead of `Comparator`?

pom.xml
```diff
@@ -303,6 +303,7 @@
                                         -Dhazelcast.logging.type=none
                                         -Dhazelcast.test.use.network=false
                                         -Dhazelcast.test.multiple.jvm=true
+                                        -ea
```
serkan-ozal: intentionally or left over?

hazelcast/src/main/java/com/hazelcast/map/eviction/CustomEvictionPolicy.java
```diff
@@ -0,0 +1,53 @@
...
+ * @param <V> the type of mapped values
+ *
+ * @see LRUEvictionPolicy
+ * @see LFUEvictionPolicy
+ * @see RandomEvictionPolicy
+ *
+ * @since 3.7
+ *
+ */
+public abstract class CustomEvictionPolicy<K, V> implements Comparator<EntryView<K, V>> {
```
serkan-ozal: * I would prefer interface instead of abstract class.
* It would be better if we use another name without `Custom` term. My point is that our built-in eviction policies (`LRUEvictionPolicy` and `LFUEvictionPolicy`) extend from this one but in fact they are not custom. Therefore, I would use a name without `Custom` prefix.

hazelcast-spring/src/main/java/com/hazelcast/spring/HazelcastConfigBeanDefinitionParser.java
ahmetmircik: fixed.

pom.xml
ahmetmircik: Intentionally added it to parallelTest profile.

hazelcast/src/main/java/com/hazelcast/map/eviction/CustomEvictionPolicy.java
ahmetmircik: How is `MapEvictionPolicy`?

serkan-ozal: I am ok with `MapEvictionPolicy`.
What about making it interface instead of abstract class?

jerrinot: interface is a pickle to evolve -> you can't add a method without breaking backward compatibility. 
with abstract class you can always add a default implementation of the newly added method.

serkan-ozal: @jerrinot regarding to your thoughts, should we prefer abstract class over interface in all cases? So what is the use case for interface in general?

My point is that you can extend from only one abstract class but can implement multiple interfaces.

jerrinot: @serkan-ozal: I understand your objection. I don't think I can make a general rule for all cases. I'd certainly never say "use abstract classes everywhere". It has to be considered on case-by-cases basis. 

When it comes to extension point for users then I'd say intefaces make sense in cases where the contract is not the main concern of implementing classes. Various `FooAware` or perhaps even `BarListener`.

However I'd say in the case of eviciton stratgy an abstract class makes sense. It makes API evolution straightforward. Imagine we'll have this requirement: "I want my strategy to be notified when a new entry is added to a map". It's pretty simple when you have an abstract class. 

The alternative approach is to have 1 method per interface. Then as you are evolving your code  is either cluttered with
``java
if (foo instaceof WeatherAware) {
  WeatherAware wa = (WeatherAware)foo;
  wa.stormStarted();
}
```
or you need to use adapters internally.

It's not uncommon to provide abstract classes as extension points in other projects too. This is quite well-known example: https://github.com/hamcrest/JavaHamcrest/blob/f681d7e29265ad8aed7b46e7d9737c497ad2fd1f/hamcrest-core/src/main/java/org/hamcrest/Matcher.java#L60-L60 :)

File: hazelcast/src/main/java/com/hazelcast/map/eviction/MapEvictionPolicy.java [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+ * @since 3.7
+ *
+ */
+public abstract class MapEvictionPolicy<K, V> implements Comparator<EntryView<K, V>> {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public abstract int compare(EntryView<K, V> entryView1, EntryView<K, V> entryView2);
```
jerrinot: could you add some easy-to-digest eviction-focused description here? 
I always suffer when implementing a comparator. Ascending, descenting, negative...

A simple explanation like: "If you prefer to evict the 1st entry then return positive number, if the 2nd entry is a better candidate then return a negative number" would help.

hazelcast/src/main/java/com/hazelcast/map/eviction/MapEvictionPolicy.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @since 3.7
+ */
+public abstract class MapEvictionPolicy<K, V> implements Comparator<EntryView<K, V>> {
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>
+     *     If you prefer to evict the 1st entry then return positive number, if the 2nd entry is a better candidate
```
jerrinot: "return 0 if both entries are equally good candidates for eviction."

ahmetmircik: done

serkan-ozal:
:+1:

hazelcast/src/main/java/com/hazelcast/map/eviction/MapEvictionPolicy.java
jerrinot: @ahmetmircik: I thought you and @serkan-ozal agreed on defining constants.

jerrinot:
:+1: provided the constants are added (and JavaDoc updated) as discussed.

hasancelik:
run-lab-run

ahmetmircik:
Thanks guys, i will add javaDoc&constants after merge of Serkan's work to make both work consistent.

PR #7755, opened 2016-03-15
More minor invocation logic cleanup

pveentjer:
* Moved getPartition method to PartitionInvocation
* Removed a nodeengine.isAlive check since it is already checked as part of the flow
* Minor renaming of invoke.. we have invoke, doinvoke and invokeinternal. Now simplified to invoke, invoke0 and doInvoke.
* Other minor simplifications


jerrinot:
:+1:

PR #7756, opened 2016-03-15
CompletableFuture cleanup

pveentjer:
The AbstractCompletableFuture and InvocationFuture shared quite a lot of logic. So
this has been pulled into the VeryAbstractCompletableFuture. Mostly the async
execution and registration of callbacks.

More cleanup will follow

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/util/VeryAbstractCompletableFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/util/VeryAbstractCompletableFuture.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.util;
+
+import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.core.ICompletableFuture;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.util.ExceptionUtil;
+
+import java.util.concurrent.Executor;
+
+public abstract class VeryAbstractCompletableFuture<V> implements ICompletableFuture<V> {
+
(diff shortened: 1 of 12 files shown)


hazelcast/src/main/java/com/hazelcast/internal/util/VeryAbstractCompletableFuture.java
```diff
@@ -0,0 +1,110 @@
...
+            try {
+                Object response = resolve(result);
+
+                if (response instanceof Throwable) {
+                    callback.onFailure((Throwable) response);
+                } else {
+                    callback.onResponse((V) response);
+                }
+            } catch (Throwable cause) {
+                logger.severe("Failed asynchronous execution of execution callback: " + callback + "for call " + caller, cause);
```
Donnerbart: Missing space in `"for call "`

Donnerbart:
I'm not a huge fan of the name `VeryAbstractCompletableFuture`, but I don't have a better suggestion (since we already have `AbstractCompletableFuture` and `BasicCompletableFuture`). Maybe something with `Common` or `Base` in the lowest hierarchy?

`AbstractCommonCompletableFuture` -> `AbstractCompletableFuture` -> `BasicCompletableFuture`
`AbstractCommonCompletableFuture` -> `InvocationFuture`

`AbstractBaseCompletableFuture` -> `AbstractCompletableFuture` -> `BasicCompletableFuture`
`AbstractBaseCompletableFuture` -> `InvocationFuture`

I could not find a good pattern for nested abstract classes yet.

:+1: (with a very minor issue)

hazelcast/src/main/java/com/hazelcast/internal/util/VeryAbstractCompletableFuture.java
```diff
@@ -0,0 +1,110 @@
...
+package com.hazelcast.internal.util;
+
+import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.core.ICompletableFuture;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.util.ExceptionUtil;
+
+import java.util.concurrent.Executor;
+
+public abstract class VeryAbstractCompletableFuture<V> implements ICompletableFuture<V> {
```
jerrinot: no JavaDoc? what's responsibility of this class? what it can do for me? when should I chose this one and when the other `AbstractCF`?

pveentjer: I'll fix.

pveentjer:
@Donnerbart probably in the end the InvocationFuture will extend the AbstractCompletableFuture. There are also other futures that need to get fixed like the ClientInvocationFuture. But I want to do it step by step instead of big rewrite of these classes.

Donnerbart:
As a temporary name I'm fine with it. We nevertheless have this issue somewhere else (nested abstract classes), I think somewhere in the test code. A common pattern would be nice to have for those cases.

PR #7757, opened 2016-03-15
Expose a subset of SerializationService through SPI

sancar:
Only four methods of serialization service is included
in SPI.

    <B extends Data> B toData(Object obj);
    <B extends Data> B toData(Object obj,
                              PartitioningStrategy strategy);
    <T> T toObject(Object data);
    ManagedContext getManagedContext();

Query related methods of serialization is intentionally leaved
as internal and not exposed.

All proxies and services are using spi.SerializationService.
Old SerializationService renamed as InternalSerializationService.


pveentjer:
I like the interface. Nice and minimalistic.

hazelcast/src/main/java/com/hazelcast/spi/serialization/SerializationService.java
```diff
@@ -0,0 +1,127 @@
...
+    <T> T toObject(Object data);
+
+    /**
+     * see {@link com.hazelcast.config.Config#setManagedContext(ManagedContext)}
+     *
+     * @return ManagedContext that is set by user in Config
+     */
+    ManagedContext getManagedContext();
+
+    @Deprecated
```
pveentjer: Why are these methods added to a new SPI interface but immediately deprecated? That doesn't sound right to me.

sancar: I am not sure about that also.  If that seems OK to us, I can completely remove them in this pr.

enesakar: ok let's remove them completely.

hazelcast/src/main/java/com/hazelcast/spi/serialization/SerializationService.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.hazelcast.core.ManagedContext;
+import com.hazelcast.core.PartitioningStrategy;
+import com.hazelcast.nio.serialization.Data;
+
+/**
+ * SPI to serialize user objects to {@link Data} and back to Object
+ * {@link Data} is the internal representation of binary data in hazelcast.
+ */
+public interface SerializationService {
```
pveentjer: Can you add a line of comment that this class can't be implemented by an end user since we want to have the freedom to add new methods.

pveentjer:
:+1:

gurbuzali:
:+1:

PR #7759, opened 2016-03-15
Backup should check replica version before executing

mdogan:
If a Backup is received out-of-order, anti-entropy system may do a full sync
before an older backup is applied. In that case, older backup should be ignored
to prevent writing stale data over newer one.


metanet:
:+1:

pveentjer:
:+1:

tombujok:
:+1 (after merge, but I've also had a deep look)

PR #7760, opened 2016-03-15
Backup should check replica version before executing

mdogan:
If a Backup is received out-of-order, anti-entropy system may do a full sync
before an older backup is applied. In that case, older backup should be ignored
to prevent writing stale data over newer one.

Backport of https://github.com/hazelcast/hazelcast/pull/7759


pveentjer:
:+1:

mdogan:
Original PR (#7759) has been merged.

PR #7761, opened 2016-03-15
InvocationFuture stacktrace fixes

pveentjer:
InvocationFuture always fixes stacktrace no matter if remote or local call. The bug is that in case of a local call, e..g iatomiclong.alter; you don't see the local stacktrace; only the stacktrace of the partition thread. 

Renamed the fixstacktrace method to fixasyncstacktrace instead of fixremote

Minor change in the separator message between stiched stacktraces. Remote
and local is removed since it isn't about remote and local


tombujok:
:+1:

Donnerbart:
The description confused me a bit, but the code change looks good :)

:+1:

PR #7763, opened 2016-03-15
More Invocation cleanup

pveentjer:
Removed the callback from the constructor of the Invocation. Instead the 'andThen' is used to register the callback. 

Minor code simplifications in invocation

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -17,7 +17,6 @@
 package com.hazelcast.spi.impl.operationservice.impl;
 
 import com.hazelcast.cluster.ClusterState;
-import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
 import com.hazelcast.core.OperationTimeoutException;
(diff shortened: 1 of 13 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -372,9 +362,7 @@ void notifyError(Object error) {
     }
 
     void notifyNormalResponse(Object value, int expectedBackups) {
-        if (value == null) {
-            value = NULL_RESPONSE;
-        }
+        value = value == null ? NULL_RESPONSE : value;
```
Donnerbart: I don't think this improves readability. In fact I think `value = value == null` is quite hard to read. If I want to write those single-line if-else I mostly use brackets to make give a hint, that there is a bool expr coming: `value = (value == null ? NULL_RESPONSE : value);`.

Donnerbart:
Changes look reasonable to me. I'm not a fan of all rewrites, but that should not be a blocker.

If I'm correct this PR saves some litter by getting rid of the `new ExecutionCallbackNode<E>()` in the `InvocationFuture` constructor. Also most tests are a bit simpler, since the `null` callback is gone. Good one.

Since I'm missing a bit context I cannot reason about the test failures. We should have at least another reviewer who knows these classes well.

:+1:

PR #7767, opened 2016-03-16
Removes unused 'initialLoad' parameter

ahmetmircik:
Removes unused 'initialLoad' parameter from PutAllOperation, it is a left-over.


ahmetmircik:
very simple pr, merging it.

PR #7768, opened 2016-03-16
Fix HeapMemoryManager#reallocate

mtopolnik:
A previous PR merged an incomplete implementation of `reallocate` by accident (its commits were based on this one's commits). Now this PR contributes the corrected implementation.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/internal/memory/impl/HeapMemoryManager.java
+++ b/hazelcast/src/test/java/com/hazelcast/internal/memory/impl/HeapMemoryManager.java
@@ -81,11 +81,7 @@ public class HeapMemoryManager implements MemoryManager {
         @Override
         public long allocate(long size) {
             assertFitsInt(size);
-            final long bytesFree = storage.length - toStorageIndex(heapTop);
-            if (size > bytesFree) {
-                throw new OutOfMemoryError(String.format("Asked to allocate %d, free bytes left %d",
-                        size, bytesFree));
-            }
+            assureEnoughMemoryLeft(size);
             final long addr = heapTop;
             lastAllocatedAddress = heapTop;
             heapTop += size;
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/internal/memory/impl/HeapMemoryManager.java
```diff
@@ -91,12 +96,36 @@ public long allocate(long size) {
...
-            throw new UnsupportedOperationException();
+            assertFitsInt(address);
+            assertFitsInt(currentSize);
+            assertFitsInt(newSize);
+            validateBlock(address, currentSize);
+            final long sizeDelta = newSize - currentSize;
+            final long newAddress;
+            if (address == lastAllocatedAddress) {
+                heapTop += sizeDelta;
+                newAddress = address;
```
serkan-ozal: update old size with new size. address (key) is still same but size (value) is changed.

mtopolnik: Commit amended.

hazelcast/src/test/java/com/hazelcast/internal/memory/impl/HeapMemoryManager.java
```diff
@@ -91,12 +96,36 @@ public long allocate(long size) {
...
+            } else {
+                newAddress = allocate(newSize);
+                System.arraycopy(storage, (int) address, storage, (int) newAddress,
+                        (int) Math.min(currentSize, newSize));
+                allocatedAddrs.remove(address);
+            }
+            usedMemory += sizeDelta;
```
serkan-ozal: In case of reallocated address is not last allocated address, this calculation `usedMemory += sizeDelta;` seems wrong. `allocate` method itself increases the used memory size as new size so the old size must be decreased here not size delta.

mtopolnik: Commit amended.

serkan-ozal:
:+1:

PR #7773, opened 2016-03-17
CacheNotExistException log supression 

pveentjer:
The CacheNotExistException can be thrown as part of normal flow and can potentially be retried. So instead of logging the exception under severse (default behavior), it is logged under debug.


serkan-ozal:
:+1:

PR #7776, opened 2016-03-17
Supress logging on severe in case of prepare failure

pveentjer:
Let the caller deal with the exception; no need to log it on when the operation is executed.


gurbuzali:
:+1:

PR #7777, opened 2016-03-17
Uses recordStore.set instead of recordStore.put in PutAllOperation if…

ahmetmircik:
… there is no map-listener to prevent unnneeded old-value loading of map-loader.

By using `recordStore.set`, we get-rid-of one extra map-loader access.Because `recordStore.put` tries to find previous value by using map-loader, in order to pass it `EntryEvent`. If loading from map-store is expensive, this can lead serious performance degradation. To prevent this potential problem, when there is no map-listener exists, don't use `recordStore.put`

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
@@ -36,8 +36,8 @@ import com.hazelcast.spi.EventFilter;
 import com.hazelcast.spi.EventRegistration;
 import com.hazelcast.spi.EventService;
 import com.hazelcast.spi.NodeEngine;
-import com.hazelcast.spi.serialization.SerializationService;
 import com.hazelcast.spi.impl.eventservice.impl.TrueEventFilter;
+import com.hazelcast.spi.serialization.SerializationService;
 import com.hazelcast.wan.ReplicationEventObject;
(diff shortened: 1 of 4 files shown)


ahmetmircik:
@alparslanavci do we need to backport this?

serkan-ozal:
:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
```diff
@@ -265,6 +265,11 @@ public void hintMapEvent(Address caller, String mapName, EntryEventType eventTyp
         // NOP
     }
 
+    @Override
+    public boolean hasEventListener(String mapName) {
+        return eventService.hasEventRegistration(SERVICE_NAME, mapName);
```
gurbuzali: how about checking if there is a listener which configured with `includeValue=true` ?

ahmetmircik: Good idea but if we go that way, we need to iterate over all registered map-listeners to find if there is any `includeValue=true` registration exists or not.

gurbuzali: we do it anyway while publishing the event ?

gurbuzali:
:+1:

PR #7779, opened 2016-03-17
add new join test

sancar:
We came across this test failure when testing cpp client.
Adding a similar test to nightly tests to catch the bug.


ihsandemir:
:+1:

jerrinot:
@sancar: thanks, much appreciated!

PR #7780, opened 2016-03-17
Removed ExecutionCallback from Invocation

pveentjer:
The ExecutonCallback is part of the InvocationFuture; not of the Invocation. This change pushes
the ExecutionCallback registration into the future using an andThen; just as any normal call.


Donnerbart:
I'm having a déjà vu. Didn't we already reviewed and merged this one? :D

:+1:

pveentjer:
We reviewed but not merged because something broke. So I pulled out this change into a single PR; and I'm currently dealing with the other part

jerrinot:
:+1: on green

pveentjer:
verify

PR #7782, opened 2016-03-18
Fixes inconsistency in naming of performance monitor property

pveentjer:
All performance monitor properties are called 'hazelcast.performance.monitor.something'
but the enabled property is called 'hazelcast.performance.monitoring.enabled' so it is
inconsistent.

This PR fixes that; but still allows the old property to function but shows a warning
when it is detected.


Donnerbart:
Simple change, will merge it directly (since it's backward compatibile, despite the property name change).

:+1:

PR #7784, opened 2016-03-18
Removed OperationService.send(response) from interface

pveentjer:
This method is removed from OperationService since it exposes internal implementation details. The method was deprecated; so safe to remove.

No enterprise change is needed.


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
```diff
@@ -423,7 +423,7 @@ public boolean send(Operation op, Address target) {
         return connectionManager.transmit(packet, connection);
     }
 
-    @Override
+    //@Override
```
Donnerbart: I see dead code :)

Donnerbart:
:+1: (once green)

jerrinot:
:+1:

PR #7787, opened 2016-03-18
Add address validation in HeapMemoryManager

mtopolnik:
Validate each accessed address by checking that it is within the allocated heap region.


pveentjer:
:+1:

PR #7790, opened 2016-03-20
Slow Invocations Plugin

pveentjer:
A new plugin for the performance monitor that shows the invocations that are slow. This was integrated in the InvocationMonitor; but now is moved into its own plugin.


mtopolnik:
:+1: (on green)

PR #7791, opened 2016-03-20
Cleanup of the duplicate handling in the InvocationMonitor

pveentjer:


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -204,12 +178,36 @@ private void scan() {
...
+         * and asks whether invocation is finished or not.
+         * Even if the remote node replies with invocation is timed out,
+         * It can’t be de-registered from the registry because of aforementioned non-matching callId scenario.
+         *
+         * Workaround:
+         * When InvocationMonitor kicks in, it will do a check for invocations that are completed
+         * but their callId's are not matching with their operations. If any invocation found for that type,
+         * it is removed from the invocation registry.
+         */
+        private boolean duplicate(Invocation inv, long callId, Iterator iterator) {
```
Donnerbart: Maybe `isDuplicate()` is better to ease the read flow in `if (isDuplicate())`. Otherwise it sounds like you act on a successful duplication of something.

Donnerbart:
Beside a minor comment :+1 (once green)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -161,31 +159,7 @@ private void scan() {
...
-                *
-                * Workaround:
-                * When InvocationMonitor kicks in, it will do a check for invocations that are completed
-                * but their callId's are not matching with their operations. If any invocation found for that type,
-                * it is removed from the invocation registry.
-                *
-                * */
-                if (!callIdMatches(callId, invocation) && isDone(invocation)) {
+                if (duplicate(invocation, callId, iterator)) {
                     iterator.remove();
```
serkan-ozal: There are two `iterator.remove` in case of `duplicate` returns `true`. One is here and other is inside `duplicate`. One of them should be removed.

Donnerbart: Woah, good catch.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
pveentjer: if it begins with 'is' it is a question; not an action.

PR #7807, opened 2016-03-21
Services are not deleted from ServiceManager on shutdown

pveentjer:
The problem is that code should be able to rely on certain services to always
exist like the LockService. No need to rip out the service and force null checks
all the time.


mdogan:
Availability of a service can be due to configuration. User can enable/disable certain services explicitly. So, either you should exclude `LockService` from this customization or do an availability check for it.

pveentjer:
Hi Mehmet, where can this be configured? The ServiceManagerImpl has explicit registration for all kinds of services; I don't see how a user can interfere with that.

The main change in this PR is that on shutdown, the services are not removed from the ServiceManagerImpl.

mdogan:
User can disable default services and enable only specific services he/she needs in `ServicesConfig`. 

See 
[ServiceManagerImpl.java#L103](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/spi/impl/servicemanager/impl/ServiceManagerImpl.java#L103)
[ServicesConfig.java](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/config/ServicesConfig.java)

pveentjer:
Ok. 

But that shouldn't change anything. What this PR does is to not clear the list of services on shutdown; whatever is configured.

mdogan:
But you can't remove `if (lockService != null) ` checks. `lockService` may be disabled explicitly, on purpose.

pveentjer:
I get it.

Issue #5422, opened 2015-06-02
[TEST-FAILURE]  ExecutionDelayTest.testExecutorRetriesTask_whenOneNodeShutdowns


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OracleJDK8/450/com.hazelcast$hazelcast-client-new/testReport/com.hazelcast.client.executor/ExecutionDelayTest/testExecutorRetriesTask_whenOneNodeShutdowns/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/3265/com.hazelcast$hazelcast-client-new/testReport/com.hazelcast.client.executor/ExecutionDelayTest/testExecutorRetriesTask_whenOneNodeShutdowns/

mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK9-Quality-Outreach/84/com.hazelcast$hazelcast-client-new/testReport/com.hazelcast.client.executor/ExecutionDelayTest/testExecutorRetriesTask_whenOneNodeShutdowns/

toktarev:
The problem is (for 2 nodes):

Node A (owner)
Node B

Client sends Auth Request to node B
Node A dies
Node B checks that A is dead and send AuthenticationException to the client

toktarev:
https://github.com/hazelcast/hazelcast/pull/5447/files

mesutcelik:
fixed by https://github.com/hazelcast/hazelcast/pull/5447

@toktarev ,
Can you please include following keywords in your commit messages? This way issue will be closed when PR is merged.
https://help.github.com/articles/closing-issues-via-commit-messages/

PR #7808, opened 2016-03-21
Changing the exception when client sends request before authentication

sancar:
related to fix done here #5422.
The AuthenticationException send when client sends wrong credentials
should before authentication, it should be retried.
Changing the later case exception to be able to remove
AuthenticationException from exceptions that needs to be retried to
reduce complexity.


ihsandemir:
:+1:

PR #7809, opened 2016-03-21
Replaced GroupProperties and ClientProperties with HazelcastProperties. Propertiy classes moved to SPI.

Donnerbart:
This PR replaces the `GroupProperties` and `ClientProperties` class with a common `HazelcastProperties` class. It also moves all new property classes into the SPI, since they are used in `NodeEngine`, which is already part of the SPI (to prevent to access internal classes from the SPI).

Also `Node` and `HazelcastClientInstanceImpl` have been adapted to create the new `HazelcastProperties` instead of the old classes. The methods to retrieve the properties have been renamed to `getProperties()` which returns `HazelcastProperties`.

This is how the packages will look:

**Package `com.hazelcast.instance`**
instance
Pure backward compatibility, will be gone in next release.

**Package `com.hazelcast.spi.properties`**
spi-member
New SPI property classes for members.

**Package `com.hazelcast.client.spi.properties`**
spi-client
New SPI property classes for clients.


pveentjer:
Node is not part of SPI.

Donnerbart:
Hmm, that would clash badly with `java.util.Properties` and maybe cause a lot of confusion. I would just rename the field to `properties` so the getter could become `getProperties()`, but not the class name.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -124,6 +123,7 @@
 
     private final NodeExtension nodeExtension;
 
+    private final com.hazelcast.internal.properties.HazelcastProperties properties;
```
pveentjer: Do we need to full package name?

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -330,11 +330,11 @@ void start() {
                     hazelcastThreadGroup.getThreadNamePrefix("MulticastThread"));
             multicastServiceThread.start();
         }
-        if (groupProperties.getBoolean(GroupProperty.DISCOVERY_SPI_ENABLED)) {
+        if (properties.getBoolean(GroupProperty.DISCOVERY_SPI_ENABLED)) {
```
pveentjer: Can you transform this to static imports

hazelcast/src/main/java/com/hazelcast/instance/Node.java
Donnerbart: Yes, because we have the deprecated class `HazelcastProperties` in the same package, which can be removed in 3.8.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
Donnerbart: There are over 350 hits on `(GroupProperty.` which matches that pattern. I don't intend to cleanup all of these in this PR.

Donnerbart:
After some discussion I extended the PR and moved the property classes to `com.hazelcast.spi.properties` and `com.hazelcast.client.spi.properties`.

The SPI will be changed so we have `HazelcastProperties getProperties()` in `NodeEngine`. The same signature is used in the internal classes `Node` and `HazelcastClientInstanceImpl`.

hazelcast-client/src/test/java/com/hazelcast/client/internal/properties/HazelcastPropertiesTest.java
```diff
@@ -1,7 +1,8 @@
-package com.hazelcast.client.config;
+package com.hazelcast.client.internal.properties;
```
pveentjer: Do we still need this test? Since the HazelcastProperties is part of core; so is tested there.

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheEventHandler.java
```diff
@@ -57,23 +57,19 @@
 
     CacheEventHandler(NodeEngine nodeEngine) {
         this.nodeEngine = nodeEngine;
-        GroupProperties groupProperties = nodeEngine.getGroupProperties();
-        invalidationMessageBatchEnabled =
-                groupProperties.getBoolean(CACHE_INVALIDATION_MESSAGE_BATCH_ENABLED);
+        HazelcastProperties hazelcastProperties = nodeEngine.getProperties();
```
pveentjer: Can you rename to 'properties'

hazelcast-client/src/test/java/com/hazelcast/client/internal/properties/HazelcastPropertiesTest.java
Donnerbart: Good question, but a good catch, the test seems still to be in the wrong package.

hazelcast/src/main/java/com/hazelcast/spi/EventService.java
```diff
@@ -16,6 +16,8 @@
 
 package com.hazelcast.spi;
 
+import com.hazelcast.spi.properties.GroupProperty;
```
pveentjer: Checkstyle not going to bitch about this since it is only used in javadoc?

pveentjer:
:+1: 

@jerrinot can you have a look

jerrinot:
@Donnerbart: Can you please explain what's the desired final state? 
What's Deprecated and kept for backward compatibility purposes only, what's private, what's part of SPI, etc.

I'm utterly confused by all these changes.

Donnerbart:
The old property classes in the `instance` package are deprecated, partly since 3.6, and should finally be gone in 3.8. All of them.

The new property classes are part of the SPI. We have a single class `HazelcastProperties` for member and client, which contains the logic. We have `GroupProperty` and `ClientProperty` which contain the property definitions.

jerrinot:
:+1: from me. @sancar or @asimarslan should have a look too.

Donnerbart:
**Package `com.hazelcast.instance`**
instance

**Package `com.hazelcast.spi.properties`**
spi-member

**Package `com.hazelcast.client.spi.properties`**
spi-client

sancar:
👍

PR #7819, opened 2016-03-22
Small cleanup of monitors package.

Donnerbart:
Just a minor cleanup without changes in the logic.

There was one unused memory gauge which should be used or removed.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
@@ -49,17 +49,16 @@ import static java.util.concurrent.TimeUnit.SECONDS;
  * {@link GroupProperty#HEALTH_MONITORING_DELAY_SECONDS}
  * Time between printing two logs of health monitor. Default values is 30 seconds.
  * <p/>
- * {@link com.hazelcast.instance.GroupProperty#HEALTH_MONITORING_THRESHOLD_MEMORY_PERCENTAGE}
+ * {@link com.hazelcast.internal.properties.GroupProperty#HEALTH_MONITORING_THRESHOLD_MEMORY_PERCENTAGE}
  * Threshold: Percentage of max memory currently in use
  * <p/>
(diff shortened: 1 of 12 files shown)


hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
```diff
@@ -381,6 +380,8 @@ private void renderHeap() {
                     .append(numberToUnit(runtimeUsedMemory.read())).append(", ");
             sb.append("heap.memory.free=")
                     .append(numberToUnit(runtimeFreeMemory.read())).append(", ");
+            sb.append("heap.memory.available=")
```
Donnerbart: `runtimeAvailableMemory` was unused, so either we add it here or we should remove it.

hazelcast/src/main/java/com/hazelcast/internal/monitors/HealthMonitor.java
```diff
@@ -498,19 +499,19 @@ private void renderOperationService() {
...
      * @return a string of the given number as a format float with two decimal places and a period
      */
-    public static String percentageString(double p) {
+    private static String percentageString(double p) {
         return format("%.2f%%", p);
     }
 
-    public static String numberToUnit(long number) {
-        //CHECKSTYLE:OFF
+    @SuppressWarnings("checkstyle:magicnumber")
```
Donnerbart: No need for a complete suppression on those few lines :)

hazelcast/src/main/java/com/hazelcast/internal/monitors/InvocationPlugin.java
```diff
@@ -45,17 +45,15 @@
      *
      * If set to 0, the plugin is disabled.
      */
-    public static final HazelcastProperty SAMPLE_PERIOD_SECONDS
```
pveentjer: Should stay public.

pveentjer:
All the properties should remain public so I can easily make use of them.

Donnerbart:
None is used outside the package, but I can revert that part. IntelliJ 2016 really got a bit messy with that visibility check enabled by default (even though I switched it to a weak warning) :(

pveentjer:
:+1: 

merge once green

PR #7821, opened 2016-03-22
Added probes to the ClientConnection

pveentjer:


hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -145,14 +145,14 @@ public ClientConnectionManagerImpl(HazelcastClientInstanceImpl client, AddressTr
     protected void initializeSelectors(HazelcastClientInstanceImpl client) {
         inputThread = new NonBlockingIOThread(
                 client.getThreadGroup(),
-                client.getName() + ".ClientInSelector",
+                client.getName() + ".thread-in",
```
pveentjer: The names of the threads are alligned with the names on the server.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientWriteHandler.java
```diff
@@ -91,7 +115,8 @@ private void writeBuffer() throws IOException {
         }
 
         buffer.flip();
-        socketChannel.write(buffer);
+        int bytesWritten = socketChannel.write(buffer);
```
Donnerbart: I'm not a 100% sure, but this could create a "hidden field" issue in SonarQube. In another place you just wrote `messagesRead.inc(builder.onData(buffer));`. I would use the same style in both places, so preferably the one without the "hidden field" issue :)

Donnerbart:
Beside a minor style comment the PR looks good to me. I'm no expert on these classes, but the additions don't look like a big overhead, even if they are in a performance path. I would appreciate a review from the client team though :)

:+1:

pveentjer:
The probes used have very little overhead. The swCounter is a lot cheaper than a volatile write (putOrdered).

pveentjer:
run-lab-run

pveentjer:
run-lab-run

Issue #3395, opened 2014-08-26
Hazelcast node continuosly logs WrongTargetException


agherardi:
Hi,
While running a load test, after a few hours one of the nodes logs thousand of warnings like this:

2014-08-26T06:20:51.898Z] [WARN] [-] [hz._hzInstance_1_vmscore_1.operation.thread-2] [com.hazelcast.spi.impl.WaitNotifyServiceImpl$WaitingOp] - [192.168.240.25]:5701 [vmscore_1] [3.2] Op: com.hazelcast.concurrent.lock.operations.LockOperation@4a642c96, com.hazelcast.spi.exception.WrongTargetException: WrongTarget! this:Address[192.168.240.25]:5701, target:Address[192.168.240.24]:5701, partitionId: 194, replicaIndex: 0, operation: com.hazelcast.spi.impl.WaitNotifyServiceImpl$WaitingOp, service: hz:impl:lockService

We're running 3.2 so I believe we have picked up the patch for 
https://github.com/hazelcast/hazelcast/issues/1815

Any ideas?

SimonTurfer:
We're getting this exception on version 3.3.3 as well, running on EC2 instances. It happens when one instance is being shutdown from the AWS console.

We're also getting spammed with "Retrying[2] packet send operation to...." messages when this happens and it wont stop retrying.

```
[hz._hzInstance_1_dev.cached.thread-22] 2015-01-04 18:03:09,928 [x.x.x.x]:5701 [dev] [3.3.3] Retrying[2] packet send operation to: Address[x.x.x.x]:5701
[hz._hzInstance_1_dev.cached.thread-22] 2015-01-04 18:03:09,932 [x.x.x.x]:5701 [dev] [3.3.3] Retrying[1] packet send operation to: Address[x.x.x.x]:5701
[hz._hzInstance_1_dev.cached.thread-22] 2015-01-04 18:03:09,937 [x.x.x.x]:5701 [dev] [3.3.3] Retrying[2] packet send operation to: Address[x.x.x.x]:5701
[hz._hzInstance_1_dev.cached.thread-22] 2015-01-04 18:03:09,953 [x.x.x.x]:5701 [dev] [3.3.3] Retrying[2] packet send operation to: Address[x.x.x.x]:5701
[...]

PR #7825, opened 2016-03-23
Tentative fix for https://github.com/hazelcast/hazelcast/issues/3395

agherardi:
It looks like the continuous WrongTargetException's are caused by WaitNotifyServiceImpl's ExpirationTask calling invalidate() about once a second on the same WaitingOp. Due to the exception, the WaitingOp never removes itself from the mapWaitingOps queue.

To fix the issue, before calling invalidate(), ExpirationTask  now checks if the target is the local node. If not, it returns a WrongTargetException to the invoker (similar to what we do for partition migration), then remove the WaitingOp from the mapWaitingOps queue.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
@@ -314,11 +314,13 @@ public class WaitNotifyServiceImpl implements WaitNotifyService {
                         return true;
                     }
                     if (waitingOp.isValid() && waitingOp.needsInvalidation()) {
-                        invalidate(waitingOp);
-                        if (!localNodeIsValidTarget(waitingOp)) {
+                        if (localNodeIsValidTarget(waitingOp)) {
+                            invalidate(waitingOp);
+                        } else {
                             logger.warning("Removing waiting operation " + waitingOp +
                                     " from queue since local node is not the target");
                             q.remove(waitingOp);
+                            waitingOp.getOperation().getResponseHandler().sendResponse(new IllegalStateException("Wrong target"));
                         }
                     }
                 }
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
@@ -20,6 +20,7 @@ import com.hazelcast.core.HazelcastInstanceNotActiveException;
 import com.hazelcast.instance.HazelcastThreadGroup;
 import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.instance.Node;
+import com.hazelcast.internal.partition.InternalPartition;
 import com.hazelcast.internal.partition.MigrationInfo;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Address;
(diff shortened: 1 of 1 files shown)


jerrinot:
@agherardi: many thanks for the PR, it's much appreciated! 
It seems it could work, but I'm concerned about the root cause of this. As there already a logic in place to prevent this. 

I'll have a better look in upcoming days!

agherardi:
> It seems it could work, but I'm concerned about the root cause of this. As there already a logic in place to prevent this.

Understood. The issue is pretty easy to repro. If you need a sample, let me know.

hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
```diff
@@ -308,16 +310,35 @@ private boolean doRun() throws Exception {
             }
 
             for (Queue<WaitingOperation> q : mapWaitingOps.values()) {
-                for (WaitingOperation waitingOp : q) {
+                while (it.hasNext()) {
```
jerrinot: I think you messed-up your branch. the is no `it` defined in this scope.

hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
```diff
@@ -308,16 +310,35 @@ private boolean doRun() throws Exception {
...
                 }
             }
             return false;
         }
     }
+
+    private Address getTargetAddress(WaitingOperation waitingOp) {
+        int partitionId = waitingOp.getPartitionId();
+        InternalPartition internalPartition = nodeEngine.getPartitionService().getPartition(partitionId);
+        return internalPartition.getReplicaAddress(waitingOp.getReplicaIndex());
```
jerrinot: this part is tricky: this will give you an address of a primary replica. (=the partition owner). 

However in theory an operation can be targeted to a non-primary partition (=backup). I *think* for now the backup operations are never blocking, but in theory there is nothing preventing blocking backup operations.

You might want to have a look at `Operation::getReplicaIndex`.

hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
agherardi: Yes, sorry - I had missed a commit. It should be fixed now.

hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
agherardi: That's the same logic that throws the WrongTargetException in OperationRunnerImpl:

{quote}
protected void ensureNoPartitionProblems(Operation op) {
        int partitionId = op.getPartitionId();

        if (partitionId < 0) {
            return;
        }

        if (partitionId != getPartitionId()) {
            throw new IllegalStateException("wrong partition, expected: " + getPartitionId() + " but found:" + partitionId);
        }

        if (internalPartition == null) {
            internalPartition = nodeEngine.getPartitionService().getPartition(partitionId);
        }

        if (retryDuringMigration(op) && internalPartition.isMigrating()) {
            throw new PartitionMigratingException(node.getThisAddress(), partitionId,
                    op.getClass().getName(), op.getServiceName());
        }

        Address owner = internalPartition.getReplicaAddress(op.getReplicaIndex());
        if (op.validatesTarget() && !node.getThisAddress().equals(owner)) {
            throw new WrongTargetException(node.getThisAddress(), owner, partitionId, op.getReplicaIndex(),
                    op.getClass().getName(), op.getServiceName());
        }
    }
{quote}

LoneRifle:
This PR appears to be superseded by #8550, which was used to resolve #3395. It should thus be closed.

jerrinot:
@LoneRifle: you are right, thanks!
@agherardi: once more for this effort. It helped great deal to identify the root cause. I'm closing this one as the root cause was addressed by the other PR. Happy Hazelcasting!

PR #7827, opened 2016-03-24
Client package privateapi

pveentjer:
Many classes of the com.hazelcast.client module have been marked as PrivateAPI to prepare them for moving in 3.8.


sancar:
:+1:

PR #7828, opened 2016-03-24
most classes instance package marked as private api

pveentjer:
To prepare them for migration in 3.8.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
@@ -19,10 +19,12 @@ package com.hazelcast.instance;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.spi.annotation.PrivateApi;
 
 /***
  * Default implementation for Hazelcast instance factory;
  */
+@PrivateApi
 public class DefaultHazelcastInstanceFactory implements HazelcastInstanceFactory {
     protected Config wrapConfig(Config config) {
         if (config == null) {
--- a/hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
@@ -18,12 +18,14 @@ package com.hazelcast.instance;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.spi.annotation.PrivateApi;
 
 /***
  * This is interface which provides capability for Hazelcast factories customization;
  * DefaultHazelcastInstanceFactory is used by default, but external service can have;
  * custom implementation which should be specified in META-INF services;
  */
+@PrivateApi
 public interface HazelcastInstanceFactory {
     HazelcastInstance newHazelcastInstance(Config config, String instanceName,
                                            NodeContext nodeContext) throws Exception;
--- a/hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
@@ -26,6 +26,7 @@ import com.hazelcast.nio.tcp.SocketChannelWrapperFactory;
 import com.hazelcast.nio.tcp.TcpIpConnection;
 import com.hazelcast.nio.tcp.WriteHandler;
 import com.hazelcast.security.SecurityContext;
+import com.hazelcast.spi.annotation.PrivateApi;
 
 import java.util.Map;
 
@@ -33,6 +34,7 @@ import java.util.Map;
  * NodeExtension is a <tt>Node</tt> extension mechanism to be able to plug different implementations of
  * some modules, like; <tt>SerializationService</tt>, <tt>SocketChannelWrapperFactory</tt> etc.
  */
+@PrivateApi
 public interface NodeExtension {
 
     /**
--- a/hazelcast/src/main/java/com/hazelcast/instance/NodeExtensionFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/NodeExtensionFactory.java
@@ -16,12 +16,14 @@
 
 package com.hazelcast.instance;
 
+import com.hazelcast.spi.annotation.PrivateApi;
 import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.ServiceLoader;
 
 import java.lang.reflect.Constructor;
 import java.util.Iterator;
 
+@PrivateApi
 public final class NodeExtensionFactory {
 
     private static final String FACTORY_ID = "com.hazelcast.instance.NodeExtension";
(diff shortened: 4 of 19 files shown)


pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
```diff
@@ -18,12 +18,14 @@
...
 import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.spi.annotation.PrivateApi;
 
 /***
  * This is interface which provides capability for Hazelcast factories customization;
  * DefaultHazelcastInstanceFactory is used by default, but external service can have;
  * custom implementation which should be specified in META-INF services;
  */
+@PrivateApi
```
Donnerbart: According to the JavaDoc there can be external implementations. So moving this class could break existing projects?

hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
```diff
@@ -19,10 +19,12 @@
 import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.spi.annotation.PrivateApi;
 
 /***
  * Default implementation for Hazelcast instance factory;
  */
+@PrivateApi
```
Donnerbart: If the interface is meant to be implemented by external services there may also be implementations which extend this non-final class?

hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
```diff
@@ -33,6 +34,7 @@
  * NodeExtension is a <tt>Node</tt> extension mechanism to be able to plug different implementations of
  * some modules, like; <tt>SerializationService</tt>, <tt>SocketChannelWrapperFactory</tt> etc.
  */
+@PrivateApi
 public interface NodeExtension {
```
noctarius: Reading the Javadoc it sounds to me like public API tbh and I know at least one project that uses this, @jerrinot :)

hazelcast/src/main/java/com/hazelcast/instance/NodeExtensionFactory.java
```diff
@@ -16,12 +16,14 @@
...
 
+import com.hazelcast.spi.annotation.PrivateApi;
 import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.ServiceLoader;
 
 import java.lang.reflect.Constructor;
 import java.util.Iterator;
 
+@PrivateApi
 public final class NodeExtensionFactory {
```
noctarius: See comment above

hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
noctarius: Probably yes, don't know any implementation though

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
noctarius: would agree

hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
pveentjer: NodeExtension can't be public API because it is treated as private. We do not care about compatibility etc.

hazelcast/src/main/java/com/hazelcast/instance/NodeExtensionFactory.java
pveentjer: Private API. It should not be used by anyone. If it would be something usable by an end user; we could not add new methods to it; e.g for hot restart.

Also NodeEngine exposes all kinds of implementation specifics like InternalSerializationService, IOService etc.

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
pveentjer: We don't move it; we deprecate it or mark it as don't use. Giving other projects to not use this API.

pveentjer: I'm not sure about this class btw. I always assumed it is private.

pveentjer: It is private since it points to NodeContext and NodeContext points to NodeExtension and NodeExtension is without any question private.

noctarius: Seems like I'm not following, neither why NodeExtension clearly is private, nor why the design prevents it (automatic pickup of NodeExtensionFactory using ServiceLoader).

hazelcast/src/main/java/com/hazelcast/instance/NodeExtensionFactory.java
noctarius: Yes true but it was different in the past :)

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
pveentjer: NodeExtension is private because:

1: if this interface would be 'fixed', then it can't be modified to serve new functionality like e.g. hot restart. NodeExtension is modified without any consideration for compatibility.

2: NodeExtension points to:
```
import com.hazelcast.cluster.ClusterState;
import com.hazelcast.internal.serialization.InternalSerializationService;
import com.hazelcast.nio.IOService;
import com.hazelcast.nio.MemberSocketInterceptor;
import com.hazelcast.nio.tcp.ReadHandler;
import com.hazelcast.nio.tcp.SocketChannelWrapperFactory;
import com.hazelcast.nio.tcp.TcpIpConnection;
import com.hazelcast.nio.tcp.WriteHandler;
```

All these classes are implementation and private api classes. They are not meant for public use. we can modify them to our hearts intent.

hazelcast/src/main/java/com/hazelcast/instance/NodeExtensionFactory.java
pveentjer: In the past there was not a lot of focus on API design.

hazelcast/src/main/java/com/hazelcast/instance/DefaultHazelcastInstanceFactory.java
jerrinot: it will be gone before 3.7 is released.

hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceFactory.java
jerrinot: `NodeExtension` is private. 
It has never been meant to be an extension point. I saw external projects use it for 2 use-cases:
1. to hack in a custom joiner impl. this is no longer needed with Discovery SPI.
2. to hack off-heap store in.

hazelcast/src/main/java/com/hazelcast/instance/NodeExtensionFactory.java
jerrinot: see https://github.com/hazelcast/hazelcast/pull/7828/files#r57309265

pveentjer:
@jerrinot can you review this PR. Since it doesn't break anything; it just slaps some annotations on the classes, there is little risk.

jerrinot:
:+1:

PR #7829, opened 2016-03-24
Invocation.calltimeout should use clusterclock

pveentjer:
The clusterclock is used to determine the invocation time; but the
regular clock was used to determine the timeout. Which isn't consistent.

As agreed with mehmet; the clusterclock should be used here as well.


mdogan:
:+1:

PR #7834, opened 2016-03-25
PerformanceMonitor scheduler task catches exception

pveentjer:
If exception isn't caught and logged; the task will be removed from
execution.


jerrinot:
:+1: it can be merged on green.

PR #7836, opened 2016-03-25
Fixes client shutdown problem

sancar:
Main issue we are trying to fix is to make sure there are
no invocations left after client has shutdown.

The assert(in ClientInvocationServiceSupport) that checks
all invocations are cleared was failing.

One of the reason that an invocation is left there was
clusterService thread. In ClusterService shutdown we will
wait for executor to shutdown completely to make sure, there
is no invocation left triggered by cluster thread.

ConnectionManager close will be done before clusterService
close to make sure, cluster thread can not trigger a new
authentication while trying to shutting down.

It turns out that CleanResources Task logic is indeed not suitable
for using when client is shutting down. Because it was postponing
clearing some invocation for 1 seconds if connection is closed
in last 5 seconds. Much simpler version of clearing invocations
is implemented in place of it.

PR #7838, opened 2016-03-25
Fixes index creation in wrong map-container problem. 

ahmetmircik:
__Bug:__
Upon subsequent destroy and creation of imap, there is a possibility that there can be more than one map-container referenced by different record-stores at the same time. Due to that, indexes can be created in an unexpected map-container and this can lead to return less than expected number of results when we query imap.

__Fix:__
Moved the removal of mapContainer to partition-thread, by making this, i want to be sure that after map#destroy, there is always a new mapContainer instance referenced from a newly created recordstore and that mapContainer instance is different from the removed one.

__Test:__
Testing this defect is so hard. I was able to reproduce it with adding some sleeps. But still trying to find a simple test also.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
@@ -226,18 +226,29 @@ class MapServiceContextImpl implements MapServiceContext {
     }
 
     @Override
-    public void destroyMap(final String mapName) {
-        localMapStatsProvider.destroyLocalMapStatsImpl(mapName);
-        final PartitionContainer[] containers = partitionContainers;
-        final Semaphore semaphore = new Semaphore(0);
+    public void destroyMap(String mapName) {
+        MapContainer mapContainer = mapContainers.get(mapName);
+        if (mapContainer == null) {
+            return;
+        }
+        mapContainer.getMapStoreContext().stop();
+        nearCacheProvider.destroyNearCache(mapName);
+        nodeEngine.getEventService().deregisterAllListeners(SERVICE_NAME, mapName);
+        localMapStatsProvider.destroyLocalMapStatsImpl(mapContainer.getName());
+
+        destroyPartitionsAndMapContainer(mapContainer);
+    }
+
+    private void destroyPartitionsAndMapContainer(MapContainer mapContainer) {
+        Semaphore semaphore = new Semaphore(0);
         InternalOperationService operationService = (InternalOperationService) nodeEngine.getOperationService();
-        for (final PartitionContainer container : containers) {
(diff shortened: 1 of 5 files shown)


ahmetmircik:
run-lab-run

ahmetmircik:
run-lab-run

ahmetmircik:
```	
Test Result (2 failures / -2)
com.hazelcast.client.heartbeat.ClientHeartbeatTest.testHeartbeatStoppedEvent
com.hazelcast.client.heartbeat.ClientHeartbeatTest.com.hazelcast.client.heartbeat.ClientHeartbeatTest
```

ahmetmircik:
run-lab-run

sancar:
@ahmetmircik These client test failures will be fixed by https://github.com/hazelcast/hazelcast/pull/7836

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -284,6 +286,10 @@ public boolean isInvalidationEnabled() {
     public InterceptorRegistry getInterceptorRegistry() {
         return interceptorRegistry;
     }
+
+    public boolean canBeRemoved() {
```
jerrinot: a very small thing: something like `markForRemoval()` could be better. as a developer I do not expect a method starting with `can-` to have a side-effect.

ahmetmircik: Thanks, completely removed that method, relying on `mapContainers#remove(mapName, mapContainer)` should be sufficient.

hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
```diff
@@ -226,18 +226,29 @@ public void flushMaps() {
     }
 
     @Override
-    public void destroyMap(final String mapName) {
-        localMapStatsProvider.destroyLocalMapStatsImpl(mapName);
-        final PartitionContainer[] containers = partitionContainers;
-        final Semaphore semaphore = new Semaphore(0);
+    public void destroyMap(String mapName) {
+        MapContainer mapContainer = mapContainers.get(mapName);
```
jerrinot: which threads can call this method?

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/MapServiceContextImpl.java
ahmetmircik: user + generic operation

ahmetmircik:
verify

jerrinot:
:+1:

eminn:
:+1:

Issue #1686, opened 2014-01-31
Make IndexAwarePredicate generic


torkeld:
It would be nice if the type parameters of Predicate were inherited by the IndexAwarePredicate

LoneRifle:
This issue is now resolved and should be assigned the 3.7 milestone and closed.

jerrinot:
fixed by #7840
@LoneRifle: thanks for the reminder and obviously your contribution too!

PR #7840, opened 2016-03-26
Implement #1686 - Make IndexAwarePredicate generic

LoneRifle:
Add <K, V> generic parameters to IndexAwarePredicate, and propagate to a
minimal number of subclasses. These include:
* All index-aware predicates used for testing;
* FalsePredicate, which implements both IndexAwarePredicate and
Predicate and so needs to supply parameters to both for consistency
(note that TruePredicate does not implement IAP), and;
* AbstractIndexAwarePredicate and by extension, AbstractPredicate.

A previous commit that affected Indexes.query and its sole caller, MapQueryEngineImpl.tryQueryUsingIndexes has been reverted.

Implements #1686

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/IndexAwarePredicate.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/IndexAwarePredicate.java
@@ -23,9 +23,9 @@ import java.util.Set;
 /**
  * This interface is used for create filterable Predicates.
  */
-public interface IndexAwarePredicate extends Predicate {
+public interface IndexAwarePredicate<K, V> extends Predicate<K, V> {
 
-    Set<QueryableEntry> filter(QueryContext queryContext);
+    Set<QueryableEntry<K, V>> filter(QueryContext queryContext);
 
     boolean isIndexed(QueryContext queryContext);
 }
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/FalsePredicate.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/FalsePredicate.java
@@ -35,14 +35,14 @@ import static com.hazelcast.internal.serialization.impl.FactoryIdHelper.PREDICAT
  * when FalsePredicate is going to implement IdentifiedDataSerializable, make sure no new instance
  * is created, but the INSTANCE is returned. No need to create new objects.
  */
-public class FalsePredicate implements IdentifiedDataSerializable, Predicate, IndexAwarePredicate {
+public class FalsePredicate<K, V> implements IdentifiedDataSerializable, Predicate<K, V>, IndexAwarePredicate<K, V> {
     /**
      * An instance of the FalsePredicate.
      */
     public static final FalsePredicate INSTANCE = new FalsePredicate();
 
      @Override
-    public boolean apply(Map.Entry mapEntry) {
+    public boolean apply(Map.Entry<K, V> mapEntry) {
         return false;
     }
 
@@ -52,7 +52,7 @@ public class FalsePredicate implements IdentifiedDataSerializable, Predicate, In
     }
 
     @Override
-    public Set<QueryableEntry> filter(QueryContext queryContext) {
+    public Set<QueryableEntry<K, V>> filter(QueryContext queryContext) {
         return Collections.emptySet();
     }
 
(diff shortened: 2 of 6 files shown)


LoneRifle:
I'm considering whether PagingPredicate (an IndexAwarePredicate) and its biggest dependency, MapQueryEngineImpl, should similarly be made generic. That change will be significant, but will remove a lot of warnings in the build relating to type erasure.

Concerns about use in TransactionalMapProxy via client message tasks should be assuaged by the implicit requirement that operations involving predicates will involve both client and cluster members to have the domain classes involved.

bwzhang2011:
@LoneRifle, that's great for such improvement while using predicate for querying. @jerrinot, would you mind sparing some time and take a look at it ?

jerrinot:
@LoneRifle: many thanks, it looks good to me. :+1: 
@tombujok: can you please have a look too? thanks!

tombujok:
@LoneRifle first of all thanks for your contribution. 

I had a thorough look at this PR but I'm of the opinion that adding generics in this class and in general in similar core classes in the `impl` package does not bring much added value. Having the top-level `Predicate<K, V>` interface generic is not a bad idea, since people who implement their own predicates, may profit from some type-safety for their concrete use-case if they want to. 
Having the in-built `EqualPredicate` generic does not bring any added value, since it uses reflection-based mechanism for extracting values - it would be a misleading use of generics IMHO.

Adding generics to core classes like `MapQueryEngineImpl` or `Indexes` does not add much value either.
Core Hazelcast implementation classes hidden from the user will never take any use of any concrete types (for example that are stored in a map). We do not really care what is store there, thus adding <K,V> everywhere does not really bring anything. We also do not store the key nor the value as Objects passed to the map like an ordinary java.util.Map - there's a couple of formats like Binary or Object, so the objects are transformed on the way, thus these generics would be misleading.

Let's have a look at this example: https://github.com/LoneRifle/hazelcast/blob/d31124c85be4321678fc4044f567b7878e8c1c91/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java#L143
This doesn't bring any type-safety for a couple of reasons:

- key and value of the entry are never touched
- kay and value may be anything, it's just a parametrised method
- QueryResult will never be generic, since the key and value are stored as Data

IMHO adding generics in such classes just brings unnecessary clutter to the code and does not bring much added value apart from just adding generics for the sake of having them...
I know that there was a ticket #1686 so sorry for the confusion - it might not have been well thought through. 

On the downside I would propose to close this PR. On the upside, I'm very happy that you got acquainted with the core hazelcast classes while working on this PR - we will definitely take advantage of your knowledge in the next community-driven issues!

LoneRifle:
I have to disagree with your assessment as I would venture that there are individuals out there who will want to implement their own IndexAwarePredicate, and so, will benefit from the presence of generics. To that end, I propose that at least f1696c3 is merged.

tombujok:
Yep, I agree that it would be good to have just the `IndexAwarePredicate` generic, since it's not in the `impl` package. Would you be interested in sending such a PR?

LoneRifle:
I'll back out d31124c and try to squash all the commits into one so that all you see is one commit. This PR can then be reused.

tombujok:
:+1: thx, looks much better.

sancar:
@tombujok @LoneRifle  Is this one passed the tests ? I think because of this pr we have checkstyle failure on the master. 

see https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/2440/console

```
22:04:36 /appdisk/jenkins/jenkins/workspace/new-lab-fast-pr/hazelcast/src/main/java/com/hazelcast/query/IndexAwarePredicate.java:26: Type Javadoc comment is missing an @param <K> tag.
22:04:36 /appdisk/jenkins/jenkins/workspace/new-lab-fast-pr/hazelcast/src/main/java/com/hazelcast/query/IndexAwarePredicate.java:26: Type Javadoc comment is missing an @param <V> tag.
```

Donnerbart:
run-lab-run

Donnerbart:
Too bad, seems that we cannot convince Jenkins to build this branch again.

Donnerbart:
I fixed the error provisional, so we can create new PRs from master again (I added the same empty JavaDoc for the key types as the `Predicate` has).

LoneRifle:
Thanks @Donnerbart . Sorry about the checkstyle errors - to date I've been unable to get my Windows Eclipse environment to pick up the Hazelcast checkstyle-suppressions.xml so I can't run checkstyle very well. Consequently I tend to be a running joke amongst the devs.

PR #7842, opened 2016-03-26
Removed InvocationBuilder litter for client call

pveentjer:
The InvocationBuilder is needed to configure deserialize; this has not been
made available as option on the InternalCompletableFuture.


pveentjer:
I'll deal with litter from the client in 3.8

PR #7846, opened 2016-03-26
OperationService start cleanup

pveentjer:
Thread startup isn't done in the constructors anymore, but moved into a
start method.


mtopolnik:
:+1:

PR #7847, opened 2016-03-26
Minor cleanup in AsyncResponsePacketHandler.Thread.doRun

pveentjer:


pveentjer:
run-lab-run

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/AsyncResponsePacketHandler.java
```diff
@@ -97,40 +98,26 @@ public ResponseThread(HazelcastThreadGroup threadGroup,
         public void run() {
             try {
                 doRun();
+            } catch (InterruptedException e) {
+                ignore(e);
```
mtopolnik: Maybe an explicit `return` is a better idea? The exception isn't literally "ignored", it's being _handled_ by ending the runnable.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/AsyncResponsePacketHandler.java
```diff
@@ -97,40 +98,26 @@ public ResponseThread(HazelcastThreadGroup threadGroup,
...
-                Packet responsePacket;
-                try {
-                    responsePacket = workQueue.take();
-                } catch (InterruptedException e) {
-                    if (shutdown) {
-                        return;
-                    }
-                    continue;
-                }
+        @SuppressFBWarnings("VO_VOLATILE_INCREMENT")
```
mtopolnik: An explicit `justification=` is a good idea here (even if seemingly obvious).

mtopolnik:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/AsyncResponsePacketHandler.java
pveentjer: The counter can be removed completely

PR #7852, opened 2016-03-28
Moved InvocationRegistry.notifyLogic into ResponseHandler

pveentjer:
Notifyin the invocations is a concern of the response handler; not of the invocation registry. So
the logic has been moved into the ResponseHandler.


jerrinot:
:+1:

PR #7855, opened 2016-03-28
Class cache introduced

jerrinot:
We already have a constructor cache in place. However the constructor cache is not
used in the case of Java serialization. Many classloaders out there have coarse grained
locking and it's causing a contention.

This change generalizes the same map to be used as both constructor cache and class cache.
There is a compatibility switch to disable the new cache should it caused any issues.

This is a backport into 3.4 branch. Master branch should also change the logic how classes are loaded,
but I'm intentionally keeping this changeset as small as possible to minimize a risk of  a regression


noctarius:
:+1: looks good

serkan-ozal:
:+1:

jerrinot:
merged into 3.4.7 branch

PR #7856, opened 2016-03-28
Removes unnecessary full scan of entries to remove indexes during ima…

ahmetmircik:
…p#destroy and node-shutdown


mdogan:
needs rebase

ahmetmircik:
will be addressed later

PR #7859, opened 2016-03-29
Semaphore additions and changes

elFarto:
Hi

For our use of Hazelcast, we needed the ability to change the number of permits available in a semaphore at runtime. This PR adds a few new functions to the ISemaphore class, including increasePermits and reducePermits. In addition, it allows the amount of available permits to become negative, meaning 1 or more releases are required before another acquire will be successful.

Regards
Stephen

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/semaphore/operations/IncreaseOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/semaphore/operations/IncreaseOperation.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.concurrent.semaphore.operations;
+
+import com.hazelcast.concurrent.semaphore.SemaphoreContainer;
+import com.hazelcast.concurrent.semaphore.SemaphoreDataSerializerHook;
+import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.impl.MutatingOperation;
+
+public class IncreaseOperation extends SemaphoreBackupAwareOperation implements MutatingOperation {
+
+    public IncreaseOperation() {
+    }
+
+    public IncreaseOperation(String name, int permitCount) {
+        super(name, permitCount);
+    }
+
+    @Override
+    public void run() throws Exception {
+        SemaphoreContainer semaphoreContainer = getSemaphoreContainer();
+        response = semaphoreContainer.increase(permitCount);
+    }
+
+    @Override
+    public boolean shouldBackup() {
+        return Boolean.TRUE.equals(response);
+    }
+
+    @Override
+    public Operation getBackupOperation() {
+        return new IncreaseBackupOperation(name, permitCount);
+    }
+
+    @Override
+    public int getId() {
+        return SemaphoreDataSerializerHook.INCREASE_OPERATION;
+    }
+}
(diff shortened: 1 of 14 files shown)


gurbuzali:
there is no need to add a new function to increase the permits, `release(int permits)` will do the job
```
    @Test(timeout = 30000)
    public void testIncreasePermits() {
        assertTrue(semaphore.init(10));

        assertEquals(10, semaphore.availablePermits());

        semaphore.release(100);

        assertEquals(110, semaphore.availablePermits());
    }
```

elFarto:
That's almost true, however the release method also changes the amount of permits that node is currently holding. If the node were to crash before those permits were released, the total available permits would not be increased by the amount that you were expecting. This is the reason I added the increasePermits method, as I had originally thought that the release method would be good enough.

pveentjer:
@gurbuzali what should be done with this PR.

elFarto:
Is there any chance of this being merged in, or are there some changes I need to make?

jerrinot:
@elFarto: hello, I apologize for a big delay. The PR looks good to me. 

Can you please sign the Hazelcast Contributor Agreement. This is a prerequisite before accepting a contribution.

elFarto:
Sorry for the delay, but I've finally signed the agreement and been added to the list.

mmedenjak:
Unfortunately this cannot be merged as-is at the moment. The commits need to be squashed and more importantly, rolling upgrade checks need to be added. I will try and make a PR with those in time for 3.9 but I'm not making any guarantees.

Holmistr:
@mmedenjak Any progress here?

mmedenjak:
@elFarto sorry again for the delay. Are you still with us here? Can you do some changes to this PR?

@blazember can you take a look at this PR?

blazember:
@elFarto if you are still with us, could you please do some changes to this PR?

Currently, this change doesn't compile as the code has changed since you sent the PR.

In order to enable rolling upgrades, invoking this new operation is allowed only if the cluster version is >= 3.10. The javadoc of `increasePermits()` needs to mention this. See this example for version verification: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java#L880-L882

`IncreaseOperation` needs to implement `MutatingOperation` marker interface. See `ReduceOperation` as an example.
`IncreaseOperation.getBackupOperation()` currently returns a `ReduceBackupOperation` instance. A new `IncreaseBackupOperation` needs to be introduced along with `IncreaseOperation`.

Additionally, `ClientSemaphoreProxy` also needs to implement `increasePermits()`, which comes with client protocol change. The change is to be done in `SemaphoreCodecTemplate` in `hazelcast-client-protocol` repo in a separate PR. The change should look like the following:

```
    @Request(id = <unused_id>, retryable = false, response = ResponseMessageConst.VOID, partitionIdentifier = "name")
    @Since("1.6")
    void increasePermits(String name, int increase);
```

elFarto:
Yes, I'm still here. I'll see if I can schedule some time to look at this.

elFarto:
Finally found some time to update this. I've also submitted a PR for the client protocol.

hazelcast/src/main/java/com/hazelcast/concurrent/semaphore/operations/IncreaseOperation.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public boolean shouldBackup() {
+        return Boolean.TRUE.equals(response);
+    }
+
+    @Override
+    public Operation getBackupOperation() {
+        return new ReduceBackupOperation(name, permitCount);
```
blazember: This should be `IncreaseBackupOperation`.

hazelcast/src/test/java/com/hazelcast/concurrent/semaphore/SemaphoreBasicTest.java
```diff
@@ -103,6 +103,31 @@ public void testRelease() {
...
+        assertEquals(0, semaphore.availablePermits());
+        
+        semaphore.release(10);
+        
+        assertEquals(5, semaphore.availablePermits());
+    }    
+    
+    
+    @Test(timeout = 30000)
+    public void testIncreasePermits() {
```
blazember: Could you please add a test case to verify that an `IllegalArgumentException` is thrown if the permit is negative?

elFarto:
I've added a test case, and fixed the reference to ReduceBackupOperation. Also fixed-up the SemaphoreDataSerializerHook class that was missing the IncreaseOperation and IncreaseBackupOperation classes.

elFarto:
Checkstyle issues fixed, and all the commits have been flatten into one and pushed.

blazember:
run-lab-run

blazember:
@elFarto our PR builder revealed a few things left to be covered. I created [this commit](https://github.com/blazember/hazelcast/commit/20764ffbd14611c7f8b14780407de10648f4bc42) for the leftovers. Could you please add the missing bits to this PR?

elFarto:
@blazember I've added the changes you've made and pushed them.

blazember:
run-lab-run

blazember:
Thank you @elFarto for your contribution and welcome to the Hazelcast contributors :star2: 
Well done!

PR #7863, opened 2016-03-29
OperationExecutor cleanup

pveentjer:
Some renaming of methods
Simplified logic
Some duplication of methods removed
Some instanceof replace by class checks
Renaming of tests
Additional tests for executeXXX
Additional tests for runOrExecute
Moved executor specific 'probe' registration from opservice to executorservice
Added additional probes


pveentjer:
For enterprise see:
https://github.com/hazelcast/hazelcast-enterprise/pull/848

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
```diff
@@ -136,26 +138,30 @@
      * @throws java.lang.NullPointerException if op is null.
      */
     @Deprecated
-    boolean isAllowedToRunInCurrentThread(Operation op);
+    boolean runAllowed(Operation op);
+
+    /**
+     * Checks if this operation is allowed to be invoked from the current thread. Invoking means that the operation can
+     * be executed on another thread, but that one is going to block for completion using the future.get/join etc.
+     * Blocking for completion can cause problems, e.g. when you hog a partition thread or deadlocks.
```
mtopolnik: I couldn't understand this description.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
```diff
@@ -65,58 +70,44 @@
...
 
-    public void start() {
-        for (Thread t : partitionOperationThreads) {
-            t.start();
-        }
-        for (Thread t : genericOperationThreads) {
-            t.start();
-        }
+        logger.info("Starting with " + genericThreads.length + " generic operation threads and "
+                + partitionThreads.length + " partition operation threads.");
```
mtopolnik: Shouldn't this log statement go into the `start()` method?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
```diff
@@ -410,37 +351,108 @@ OperationRunner getOperationRunner(Operation operation) {
...
+        }
+
+        PartitionOperationThread partitionThread = (PartitionOperationThread) currentThread;
+
+        // so it's a partition operation thread, now we need to make sure that this operation thread is allowed
+        // to execute operations for this particular partitionId
+        return toPartitionThreadIndex(partitionId) == partitionThread.threadId;
+    }
+
+     @Override
```
mtopolnik: Indentation mistake

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
pveentjer: You are right.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
pveentjer: Normally an invocation is done asynchronous; so you execute and you get a invocationfuture. But in most cases you immediately do a future.get so the calling thread is going to block for completion.

This methods checks if blocking for completion is allowed on the current thread for the given operation.

Does this make sense? Lets first make sure you understand it, and then we update the documentation accordingly.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/classic/ClassicOperationExecutor.java
```diff
@@ -65,58 +70,44 @@
...
+        this.partitionThreads = initPartitionThreads(properties, threadGroup, nodeExtension, metricsRegistry);
 
         this.genericOperationRunners = initGenericOperationRunners(properties, operationRunnerFactory);
-        this.genericOperationThreads = initGenericThreads();
+        this.genericThreads = initGenericThreads(threadGroup, nodeExtension, metricsRegistry);
 
-        logger.info("Starting with " + genericOperationThreads.length + " generic operation threads and "
-                + partitionOperationThreads.length + " partition operation threads.");
-    }
+        metricsRegistry.scanAndRegister(this, "operation");
```
Donnerbart: Looks like an escape of the `this` reference during construction.

pveentjer: Last line in the constructor. Similar approach as in other parts of the system.

Donnerbart: Instructions can be re-ordered? This is a huge no-go in our favorite book from Brian Goetz (3.2.1 Safe construction practices). So it maybe wrong in other parts of the system as well.

mtopolnik: In this case there's already the start method so it should be simple.

pveentjer: Instructions in front of the registration, can jump over it.

pveentjer:
If I clean up the registration; I'll do it in 1 go for all other services.

Donnerbart:
I'm not a huge fan of those boolean methods without an `is` prefix, since without that prefix it's not obvious if they change a state / execute something or not. A `if (runAllowed(operation))` could also check if the `operation.run()` was successful.

So I would prefer `isRunAllowed()` and `isInvocationAllowed()` over `runAllowed()` and `invocationAllowed()`. Especially since we use the `isConditionName` pattern in other methods like `isOperationThread()` and variable names like `isAllowed` or `isAsync`.

Beside that it's a :+1: from me (once green).

pveentjer:
Good points.

With the is.. it is implied in the 'allowed'. It asks if something can be done. So adding 'is' to it is effectively duplication and 

if(runAllowed(...)) reads imho better than if(isRunAllowed(...))

So let me think about it.

pveentjer:
Renamed

mtopolnik:
A good reason to rename is that `runAllowed` can be mistaken for "run this task, which is allowed".

pveentjer:
Good points.

They have been renamed

Donnerbart:
run-lab-run

PR #7876, opened 2016-03-30
Exist key operation.

angel9484:
This operation is equal than containsKey but this doesn't trigger the mapLoader.
Useful when you know you are on the owner node and need to fast-query if the value is loaded into the local map instead of iterate over all localKeySet.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -207,6 +207,11 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         return containsKeyInternal(keyData);
     }
 
+    @Override
+    public boolean existKey(Object key) {
+        throw new UnsupportedOperationException("Locality is ambiguous for client!!!");
+    }
+
     protected boolean containsKeyInternal(Data keyData) {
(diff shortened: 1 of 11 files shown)


jerrinot:
Hello @angel9484,
thank you for your PR. I can see usefulness of this, but I am not sure whether this should be exposed via its own method. 

I am a bit worried about API-creep. As in theory you can add a new method which by-pass MapLoader for many existing methods. One option how to prevent this could be adding a method `IMap bypassMapLoader()` (or similar) to the `IMap` interface . It would return another instance of IMap where all calls would bypass configured MapLoader. 

What do you think about this approach?

ahmetmircik:
Instead of `existKey`, as a workaround, `IMap#getEntryView` can be used. It only checks existing in-memory data, doesn't try to load value via map-loader.

enesakar:
Thanks for contribution. My remarks:
`existKey` will create a confusion with `containsKey` in the IMap API.
To bypass mapstore in read operations, you can simply return null from `V load(K key);`
Right?

noctarius:
@enesakar I think the use case is different here. I guess he wants to figure out if a Key was actually loaded into Hazelcast without loading it if not. Your solution would render it to be never loaded at all.

Anyhow I think the method would be better named like `isKeyLoaded(Key)` since I certainly agree that `existKey` is not a good self-explanatory name :)

PR #7877, opened 2016-03-30
MapProvider now is configurable as a wrapper.

angel9484:
Then we can customize the Operations, per example the GetOperation to inject another one.


jerrinot:
Hello @angel9484,
thanks for the PR. Can you please describe a use-case for this?

pveentjer:
@angel9484 do you have an update? In 2 weeks we'll enter feature freeze/

PR #7878, opened 2016-03-31
Renamed ClassicOperationExecutor to OperationExecutorImpl

pveentjer:
It was named classic since a year ago I had another implementation that allows for caller runs optimization. However due to threading limitations in HZ enterprise; this is not a practical model

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
@@ -19,6 +19,7 @@ package com.hazelcast.spi.impl.operationexecutor;
 import com.hazelcast.nio.Packet;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.impl.PartitionSpecificRunnable;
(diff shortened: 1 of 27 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationExecutor.java
```diff
@@ -19,6 +19,7 @@
 import com.hazelcast.nio.Packet;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.impl.PartitionSpecificRunnable;
+import com.hazelcast.spi.impl.operationexecutor.impl.OperationExecutorImpl;
```
Donnerbart: If this import is just used in JavaDoc we will have an "unused import" error somewhere. I think it's one of the tools run by SonarQube, maybe even CheckStyle with it's separate configuration. So which side should I adjust? Should the build fail due to this or should I disable this check it in SonarQube?

pveentjer: I removed the piece of comment so the import is not needed at all.

PR #7879, opened 2016-03-31
OperationService cleanup

pveentjer:
- renamed InternalOpServe.getPartitionOperatonThreadCount to getPartitionThreadCount
- renamed InternalOpServe.getGenericOperationThreadCount to getGenericOperationThreadCount
- added tests for the above methods
- removed Opservice.isAllowedToRunOnCallingThread since it is deprecated
- introduced InOpService.isRunAllowed since it is needed by enterprise
- removed deprecation annotation of OperationExecutor.isRunAllowed since method is still used
- added documentation to some methods on InternalOpService
- renamed AsyncResponsePacketHandler to AsyncResponseHandler to be in line with ResponseHandler
- moved the probe for AsyncResponseHandler from OpService into AsyncResponseHandler
- renamed the probe for asyncresponsepacket handler from response-queue.size to responseQueueSize
- minor testing cleanup

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
@@ -35,6 +35,11 @@ import java.util.List;
  */
 public interface InternalOperationService extends OperationService, PacketHandler {
 
+    /**
+     * Returns the size of the response queue.
+     *
+     * @return the size of the response queue.
+     */
     int getResponseQueueSize();
 
     int getOperationExecutorQueueSize();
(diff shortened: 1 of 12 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
```diff
@@ -35,6 +35,11 @@
  */
 public interface InternalOperationService extends OperationService, PacketHandler {
 
+    /**
+     * Returns the size of the response queue.
+     *
+     * @return the size of the response queue.
```
mtopolnik: Isn't this redundant? I'd only use `@return` in more complicated cases, when there is something further to clarify in addition to the main description.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
```diff
@@ -60,6 +80,16 @@
     boolean isCallTimedOut(Operation op);
 
     /**
+     * Returns true if the given operation is allowed to run on the calling thread, false otherwise.
```
mtopolnik: This sentence already exists in the `@return` part.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/InternalOperationService.java
pveentjer: If I remove it, is it still valid javadoc? And yes, it is repetition.

mtopolnik: It is invalid to have a `@return` tag with no content, but it's fine not to have the tag at all. AFAIK.

PR #7880, opened 2016-03-31
Small cleanup of map proxy classes

Donnerbart:
Small cleanup of map proxy classes (reduced warnings, formatted comments, added generics, renamed single-letter variables).

Changes were done with a Java 6 JDK to ensure that removed casts were not needed. No logic should be changed.


pveentjer:
:+1:

PR #7885, opened 2016-03-31
Smarter PacketDispatcher

pveentjer:
All the dispatching is now done in the PacketDispatcher and removed from the OperationService. A layer less to jump through.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/packetdispatcher/impl/PacketDispatcherImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/packetdispatcher/impl/PacketDispatcherImpl.java
@@ -25,32 +25,40 @@ import static com.hazelcast.instance.OutOfMemoryErrorDispatcher.inspectOutputMem
 import static com.hazelcast.nio.Packet.FLAG_BIND;
 import static com.hazelcast.nio.Packet.FLAG_EVENT;
 import static com.hazelcast.nio.Packet.FLAG_OP;
+import static com.hazelcast.nio.Packet.FLAG_RESPONSE;
 
 /**
  * Default {@link PacketDispatcher} implementation.
  */
-public class PacketDispatcherImpl implements PacketDispatcher {
+public final class PacketDispatcherImpl implements PacketDispatcher {
 
     private final ILogger logger;
     private final PacketHandler eventPacketHandler;
-    private final PacketHandler operationPacketHandler;
+    private final PacketHandler operationExecutor;
     private final PacketHandler connectionPacketHandler;
+    private final PacketHandler responseHandler;
 
     public PacketDispatcherImpl(ILogger logger,
-                                PacketHandler operationPacketHandler,
+                                PacketHandler operationExecutor,
+                                PacketHandler responseHandler,
                                 PacketHandler eventPacketHandler,
                                 PacketHandler connectionPacketHandler) {
         this.logger = logger;
-        this.operationPacketHandler = operationPacketHandler;
+        this.responseHandler = responseHandler;
         this.eventPacketHandler = eventPacketHandler;
         this.connectionPacketHandler = connectionPacketHandler;
+        this.operationExecutor = operationExecutor;
     }
 
     @Override
     public void dispatch(Packet packet) {
         try {
             if (packet.isFlagSet(FLAG_OP)) {
-                operationPacketHandler.handle(packet);
+                if (packet.isFlagSet(FLAG_RESPONSE)) {
+                    responseHandler.handle(packet);
+                } else {
+                    operationExecutor.handle(packet);
+                }
             } else if (packet.isFlagSet(FLAG_EVENT)) {
                 eventPacketHandler.handle(packet);
             } else if (packet.isFlagSet(FLAG_BIND)) {
@@ -60,7 +68,7 @@ public class PacketDispatcherImpl implements PacketDispatcher {
             }
         } catch (Throwable t) {
             inspectOutputMemoryError(t);
-            logger.severe("Failed to process packet:" + packet, t);
+            logger.severe("Failed to process:" + packet, t);
         }
     }
 }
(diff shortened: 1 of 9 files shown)


pveentjer:
run-lab-run

jerrinot:
:+1: good one!

pveentjer:
run-lab-run

pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/spi/impl/packetdispatcher/impl/PacketDispatcherImpl.java
```diff
@@ -25,42 +25,51 @@
...
     }
 
     @Override
     public void dispatch(Packet packet) {
         try {
-            if (packet.isFlagSet(FLAG_OP)) {
-                operationPacketHandler.handle(packet);
-            } else if (packet.isFlagSet(FLAG_EVENT)) {
+            int flags = packet.getFlags();
+            if ((flags & FLAG_OP) != 0) {
```
Donnerbart: Hmm, this looks like a micro-optimization to me. I think `packet.isFlagSet()` is easier to read and should anyway get inlined by the JVM. It's also nearly the same amount of characters and less operators to read.

pveentjer: I cleaned up the flags to be able do smarter bitlogic outside of these operations. Personally I find this easier to read.

pveentjer: In c you could even simplify it to:

```
 @Override
    public void dispatch(Packet packet) {
        try {
            int flags = packet.getFlags();
            if (flags & FLAG_OP) {
                if (flags & FLAG_RESPONSE) {
                    responseHandler.handle(packet);
                } else {
                    operationExecutor.handle(packet);
                }
            } else if (flags & FLAG_EVENT) {
                eventPacketHandler.handle(packet);
            } else if (flags & FLAG_BIND) {
                connectionPacketHandler.handle(packet);
            } else {
                logger.severe("Unknown packet type! Header: " + packet.getFlags());
            }
        } catch (Throwable t) {
            inspectOutputMemoryError(t);
            logger.severe("Failed to process:" + packet, t);
        }
    }
```

mdogan: `packet.isFlagSet()` is just plain English. how come `(flags & FLAG_OP) != 0` is easier to read?

pveentjer: Perhaps customs? In my c days, that is how you normally work with flags.

mdogan: In someone's assembly days, it was done like;

```
movl $0x1, %edx
movl $0x0, %ecx
andl %edx, %ecx
...
```

Donnerbart: I like the idea of combining flags outside of `Packet`, but keeping the bit logic inside the class. So the counterpart of the very nice `packet.setAllFlags(FLAG_A | FLAG_B)` would be something like this:
```
if (packet.hasAllFlags(FLAG_OP | FLAG_RESPONSE)) {
    responseHandler.handle(packet);
} else if (packet.hasAllFlags(FLAG_OP)) {
    operationExecutor.handle(packet);
} else if (packet.hasAllFlags(FLAG_EVENT)) {
```

pveentjer: Personally I really don't see the added value of repeatedly asking the packet for the same data and to check if some bits are set. Perhaps it is a premature optimization, perhaps it is just clean code. So this is a flavor based issue and we'll not reach consensus on this.

pveentjer: I case of doubt I look at Aeron/Agrona:

This comes from Aeron controlled fragment assembler:

```
    public Action onFragment(final DirectBuffer buffer, final int offset, final int length, final Header header)
    {
        final byte flags = header.flags();

        Action action = Action.CONTINUE;

        if ((flags & UNFRAGMENTED) == UNFRAGMENTED)
        {
            action = delegate.onFragment(buffer, offset, length, header);
        }
        else
        {
            if ((flags & BEGIN_FRAG_FLAG) == BEGIN_FRAG_FLAG)
            {
                final BufferBuilder builder = builderBySessionIdMap.computeIfAbsent(header.sessionId(), builderFunc);
                builder.reset().append(buffer, offset, length);
            }
            else
```

pveentjer: What about this one:

```
  @Override
    public void dispatch(Packet packet) {
        try {
            int flags = packet.getFlags() & (FLAG_OP | FLAG_RESPONSE | FLAG_EVENT | FLAG_BIND);
            switch (flags) {
                case FLAG_OP:
                    operationExecutor.handle(packet);
                    break;
                case FLAG_OP | FLAG_RESPONSE:
                    responseHandler.handle(packet);
                    break;
                case FLAG_EVENT:
                    eventPacketHandler.handle(packet);
                    break;
                case FLAG_BIND:
                    connectionPacketHandler.handle(packet);
                    break;
                default:
                    logger.severe("Unknown packet type! Header: " + packet.getFlags());
            }
        } catch (Throwable t) {
            inspectOutputMemoryError(t);
            logger.severe("Failed to process:" + packet, t);
        }
    }
```

noctarius: I personally would agree with Mehmet and David :) Anyhow my solution would be different:

```
enum PacketFlags {
  FLAG_X;

  static boolean flagSet(int flags, PacketFlags flag) {
    return flags & flag == flag;
  }
}
```

pveentjer:
Unless someone has a non opinion based reason to get this pr merged, I'm going to merge this one.

noctarius:
:+1: no other objections :D

Donnerbart:
:+1: Good cleanup :)

PR #7886, opened 2016-03-31
Incorrect protocol version number for response messages

ihsandemir:
Changed the protocol version to 1. It was 0 for the response messages but it needs to be 1.


pveentjer:
:+1:

ihsandemir:
No, it is a correction. The protocol spec that we released was correct. It says:
**Version uint8 Protocol version. Current version is 1.
Hence, it is be an implementation fix.

PR #7888, opened 2016-04-01
Cleanup of PartitionIteratingOperation and ResponseQueueFactory

Donnerbart:
* Cleanup of `PartitionIteratingOperation` and `ResponseQueueFactory`.
* Switched from `HashMaps` to `int[]` and `Object[]` to reduce litter and lookup costs.
* Reused an internal array to reduce litter.
* Pulled out new utility methods to `CollectionUtil`.
* Increased code coverage of `CollectionUtil` and `ResponseQueueFactory`.

I verified the cleanup with JFR and a simple benchmark.

Test code: [PartitionIteratingOperationBenchmark.java](https://gist.github.com/Donnerbart/33e8258d432701c68aea528395a4f5ac#file-partitioniteratingoperationbenchmark-java)
JFR settings: [hz-profiling.zip](https://github.com/hazelcast/hazelcast/files/204521/hz-profiling.zip)

```
Cluster size: 10
Partition count: 2571
Map size: 25710
Iterations: 500000
```
```
500000 operations done in 604276 ms (1208 µs per op) (master)...
vs.
500000 operations done in 587978 ms (1175 µs per op) (cleanup)...
```

| Name | Master | Cleanup |
| --- | ---: | ---: |
| Duration per operation | ~1208 µs | ~1175 µs |
| GC Count | 2206 | 1786 |
| Total GC Time | 5s 30ms | 4s 231ms |
| TLAB count | 13,023,904 | 10,575,857 |
| Allocation Rate for TLAB | 1.16 GB/s | 980.52 MB/s |

**Allocation pressure by class (master vs. cleanup)**
pio-alloc-pressure-by-class-master
pio-alloc-pressure-by-class-cleanup

`HashMap$Node` and `Integer` allocation pressure is down by 14.01%. The added allocation pressure by `int[]` and `Object[]` is 6.37%. But this might be hard to compare, since the whole pressure profile shifted a bit. Nevertheless we saved 96.2% of the `HashMap$Node` and 32.6% of `Integer` allocations, mostly in the `com.hazelcast.spi.impl.operationservice.impl.AsyncResponseHandler$ResponseThread`.

**Garbage collections (master vs. cleanup)**
pio-gc-master
pio-gc-cleanup

This should have a positive effect on all usages of this operation, which should be `IMap.size()`, `IMap.isEmpty()`, `IMap.clear()`, `IMap.contains...()`, `IMap.keySet()`, `IMap.entrySet()`, `IMap.values()`, `IMap.getAll()`, `IMap.executeOnKeys()`,  etc.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/PartitionIteratingOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/PartitionIteratingOperation.java
@@ -32,73 +32,73 @@ import com.hazelcast.spi.impl.operationservice.impl.responses.NormalResponse;
 import com.hazelcast.util.ResponseQueueFactory;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
 
+import static com.hazelcast.util.CollectionUtil.toIntArray;
+
 public final class PartitionIteratingOperation extends AbstractOperation implements IdentifiedDataSerializable {
 
-    private List<Integer> partitions;
     private OperationFactory operationFactory;
-    private Map<Integer, Object> results;
-
-    public PartitionIteratingOperation(List<Integer> partitions, OperationFactory operationFactory) {
-        this.partitions = partitions != null ? partitions : Collections.<Integer>emptyList();
-        this.operationFactory = operationFactory;
-    }
+    private int[] partitions;
+    private Object[] results;
 
     public PartitionIteratingOperation() {
     }
 
+    public PartitionIteratingOperation(OperationFactory operationFactory, List<Integer> partitions) {
+        this.operationFactory = operationFactory;
+        this.partitions = toIntArray(partitions);
+    }
+
     @Override
     public void run() throws Exception {
-        results = new HashMap<Integer, Object>(partitions.size());
         try {
-            Map<Integer, ResponseQueue> responses = executeOperations();
-            getResults(responses);
+            Object[] responses = executeOperations();
+            results = resolveResponses(responses);
         } catch (Exception e) {
             getLogger(getNodeEngine()).severe(e);
         }
     }
 
-    private void getResults(Map<Integer, ResponseQueue> responses) throws InterruptedException {
-        for (Map.Entry<Integer, ResponseQueue> responseQueueEntry : responses.entrySet()) {
-            final ResponseQueue queue = responseQueueEntry.getValue();
-            final Integer key = responseQueueEntry.getKey();
-            final Object result = queue.get();
-            if (result instanceof NormalResponse) {
-                results.put(key, ((NormalResponse) result).getValue());
-            } else {
-                results.put(key, result);
-            }
-        }
-    }
-
-    private Map<Integer, ResponseQueue> executeOperations() {
+    private Object[] executeOperations() {
         NodeEngine nodeEngine = getNodeEngine();
-        Map<Integer, ResponseQueue> responses = new HashMap<Integer, ResponseQueue>(partitions.size());
-        for (final int partitionId : partitions) {
+        Object[] responses = new Object[partitions.length];
+        for (int i = 0; i < partitions.length; i++) {
             ResponseQueue responseQueue = new ResponseQueue();
-            final Operation op = operationFactory.createOperation();
-            op.setNodeEngine(nodeEngine)
-                    .setPartitionId(partitionId)
+            responses[i] = responseQueue;
+
+            Operation operation = operationFactory.createOperation();
+            operation.setNodeEngine(nodeEngine)
+                    .setPartitionId(partitions[i])
                     .setReplicaIndex(getReplicaIndex())
                     .setOperationResponseHandler(responseQueue)
                     .setServiceName(getServiceName())
                     .setService(getService())
                     .setCallerUuid(getCallerUuid());
-            OperationAccessor.setCallerAddress(op, getCallerAddress());
-            responses.put(partitionId, responseQueue);
-            nodeEngine.getOperationService().executeOperation(op);
+            OperationAccessor.setCallerAddress(operation, getCallerAddress());
+            nodeEngine.getOperationService().executeOperation(operation);
         }
         return responses;
     }
 
-    @Override
-    public void afterRun() throws Exception {
+    /**
+     * Replaces the {@link ResponseQueue} entries with its results.
+     *
+     * The responses array is reused to avoid the allocation of a new array.
+     */
+    private Object[] resolveResponses(Object[] responses) throws InterruptedException {
+        for (int i = 0; i < responses.length; i++) {
+            ResponseQueue queue = (ResponseQueue) responses[i];
+            Object result = queue.get();
+            if (result instanceof NormalResponse) {
+                responses[i] = ((NormalResponse) result).getValue();
+            } else {
+                responses[i] = result;
+            }
+        }
+        return responses;
     }
 
     private ILogger getLogger(NodeEngine nodeEngine) {
@@ -106,8 +106,12 @@ public final class PartitionIteratingOperation extends AbstractOperation impleme
     }
 
     @Override
+    public void afterRun() throws Exception {
+    }
+
+    @Override
     public Object getResponse() {
-        return new PartitionResponse(results);
+        return new PartitionResponse(partitions, results);
     }
 
     @Override
@@ -118,17 +122,18 @@ public final class PartitionIteratingOperation extends AbstractOperation impleme
     }
 
     private static class ResponseQueue implements OperationResponseHandler {
-        final BlockingQueue b = ResponseQueueFactory.newResponseQueue();
+
+        private final BlockingQueue<Object> queue = ResponseQueueFactory.newResponseQueue();
 
         @Override
         public void sendResponse(Operation op, Object obj) {
-            if (!b.offer(obj)) {
+            if (!queue.offer(obj)) {
                 throw new HazelcastException("Response could not be queued for transportation");
             }
         }
 
         public Object get() throws InterruptedException {
-            return b.take();
+            return queue.take();
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


pveentjer:
Why not make use of an array?

Donnerbart:
Please read the description:
> I first wanted to use an `Object[]`, but we might not have all partitions as a monotonic sequence (we run on a `List<Integer> partitions`). So we need the partitionIds.

Also `Int2ObjectHashMap` seems to be the wrong candidate, since it cannot handle `null` values (see test failures).

pveentjer:
You can always stick in a NULL marker object.

Donnerbart:
But you also need to resolve those marker objects. In the end you need a map which supports the real `null` value. I'll have a look if this is possible, maybe in another PR.

At least we can safe one `HashMap` and reuse the existing one, although this feels a bit dirty. What do you think about this hack?

Donnerbart:
B̶u̶t̶ ̶y̶o̶u̶ ̶a̶l̶s̶o̶ ̶n̶e̶e̶d̶ ̶t̶o̶ ̶r̶e̶s̶o̶l̶v̶e̶ ̶t̶h̶o̶s̶e̶ ̶m̶a̶r̶k̶e̶r̶ ̶o̶b̶j̶e̶c̶t̶s̶.̶ ̶I̶n̶ ̶t̶h̶e̶ ̶e̶n̶d̶ ̶y̶o̶u̶ ̶n̶e̶e̶d̶ ̶a̶ ̶m̶a̶p̶ ̶w̶h̶i̶c̶h̶ ̶s̶u̶p̶p̶o̶r̶t̶s̶ ̶t̶h̶e̶ ̶r̶e̶a̶l̶ ̶`̶n̶u̶l̶l̶`̶ ̶v̶a̶l̶u̶e̶.̶ ̶I̶'̶l̶l̶ ̶h̶a̶v̶e̶ ̶a̶ ̶l̶o̶o̶k̶ ̶i̶f̶ ̶t̶h̶i̶s̶ ̶i̶s̶ ̶p̶o̶s̶s̶i̶b̶l̶e̶,̶ ̶m̶a̶y̶b̶e̶ ̶i̶n̶ ̶a̶n̶o̶t̶h̶e̶r̶ ̶P̶R̶.̶
̶
̶A̶t̶ ̶l̶e̶a̶s̶t̶ ̶w̶e̶ ̶c̶a̶n̶ ̶s̶a̶f̶e̶ ̶o̶n̶e̶ ̶`̶H̶a̶s̶h̶M̶a̶p̶`̶ ̶a̶n̶d̶ ̶r̶e̶u̶s̶e̶ ̶t̶h̶e̶ ̶e̶x̶i̶s̶t̶i̶n̶g̶ ̶o̶n̶e̶,̶ ̶a̶l̶t̶h̶o̶u̶g̶h̶ ̶t̶h̶i̶s̶ ̶f̶e̶e̶l̶s̶ ̶a̶ ̶b̶i̶t̶ ̶d̶i̶r̶t̶y̶.̶ ̶W̶h̶a̶t̶ ̶d̶o̶ ̶y̶o̶u̶ ̶t̶h̶i̶n̶k̶ ̶a̶b̶o̶u̶t̶ ̶t̶h̶i̶s̶ ̶h̶a̶c̶k̶?̶

I changed the implementation to arrays.

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/PartitionIteratingOperation.java
```diff
@@ -32,73 +32,73 @@
...
+    private Object[] results;
 
-    public PartitionIteratingOperation(List<Integer> partitions, OperationFactory operationFactory) {
-        this.partitions = partitions != null ? partitions : Collections.<Integer>emptyList();
-        this.operationFactory = operationFactory;
+    public PartitionIteratingOperation() {
     }
 
-    public PartitionIteratingOperation() {
+    public PartitionIteratingOperation(OperationFactory operationFactory, List<Integer> partitions) {
```
pveentjer: Why isn't this done by the caller?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/PartitionIteratingOperation.java
```diff
@@ -150,39 +155,45 @@ public int getId() {
     @Override
     protected void writeInternal(ObjectDataOutput out) throws IOException {
         super.writeInternal(out);
-        int pCount = partitions.size();
-        out.writeInt(pCount);
+        out.writeObject(operationFactory);
+        out.writeInt(partitions.length);
```
pveentjer: At the line below there is still a Integer instance creased

hazelcast/src/main/java/com/hazelcast/util/ResponseQueueFactory.java
```diff
@@ -173,5 +176,14 @@ public Object peek() {
...
             }
         }
+
+        /**
+         * Internal method, should be called under lock.
+         */
+        private Object getAndRemoveResponse() {
+            Object value = response;
+            response = null;
+            return (value == NULL) ? null : value;
```
pveentjer: why the extra parenthesis?

hazelcast/src/test/java/com/hazelcast/util/CollectionUtilTest.java
```diff
@@ -129,5 +193,51 @@ public void objectToDataCollection_deserializeBack() {
...
+    public void testToIntArray() {
+        Collection<Integer> list = new ArrayList<Integer>();
+        list.add(42);
+        list.add(23);
+        list.add(Integer.MAX_VALUE);
+
+        int[] intArray = toIntArray(list);
+
+        assertNotNull(intArray);
+        assertEquals(3, intArray.length);
```
pveentjer: Since the value in the list are a bit unpredictable, I would test it like this:

```
assertEquals(list.get(0), intArray[0])
assertEquals(list.get(1), intArray[1);
assertEquals(list.get(2), intArray[2])
```

This makes it easier to read since you don't need to remember these values.

hazelcast/src/main/java/com/hazelcast/util/ResponseQueueFactory.java
Donnerbart: Because it's an expression and it's super helpful for my eyes to give a guidance, that it's not `return value`, but `return (expression which you have to read completely to understand what's going on)`. It's a pattern I always use, although it makes more sense if the first token is quite long.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/PartitionIteratingOperation.java
Donnerbart: Good catch!

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/PartitionIteratingOperation.java
Donnerbart: I followed up the call hierarchy and found no easy way of directly create the partitions as `int[]`, so I left the constructor as it is for now.

The `PartitionIteratingOperation` is created in just a single place, generated by looping over a `Map<Address, List<Integer>> memberPartitions`. I had a brief look how that map is created and found no elegant way to directly create a `Map<Address, int[]>`, since the array lengths are not predictable.

So I could easily pull the conversion up a single level. But then we have to suppress this FindBugs issue without having a lot gained:
```
new com.hazelcast.spi.impl.operationservice.impl.operations.PartitionIteratingOperation(OperationFactory, int[]) may expose internal representation by storing an externally mutable object into PartitionIteratingOperation.partitions [com.hazelcast.spi.impl.operationservice.impl.operations.PartitionIteratingOperation] At PartitionIteratingOperation.java:[line 49]
```
So for now I would leave this as it is, until we find a way to cleanup the whole call hierarchy.

Donnerbart:
I tried to verify the cleanup with JFR today and I think the change is improving the memory usage.

Test code: [PartitionIteratingOperationBenchmark.java](https://gist.github.com/Donnerbart/33e8258d432701c68aea528395a4f5ac#file-partitioniteratingoperationbenchmark-java)
JFR settings: [hz-profiling.zip](https://github.com/hazelcast/hazelcast/files/204521/hz-profiling.zip)

```
Cluster size: 10
Partition count: 2571
Map size: 25710
Iterations: 500000
```

| Name | Master | Cleanup |
| --- | ---: | ---: |
| Duration per operation | ~1208 µs | ~1175 µs |
| GC Count | 2206 | 1786 |
| Total GC Time | 5s 30ms | 4s 231ms |
| TLAB count | 13,023,904 | 10,575,857 |
| Allocation Rate for TLAB | 1.16 GB/s | 980.52 MB/s |

**Allocation pressure by class**
pio-alloc-pressure-by-class-master
pio-alloc-pressure-by-class-cleanup

So we saved 96.2% of the `HashMap$Node` and 32.6% of `Integer` allocations, mostly in the `com.hazelcast.spi.impl.operationservice.impl.AsyncResponseHandler$ResponseThread`.

Donnerbart:
verify

hazelcast/src/main/java/com/hazelcast/util/CollectionUtil.java
```diff
@@ -97,19 +99,60 @@ public static boolean isNotEmpty(Collection collection) {
...
+    /**
+     * Converts a {@link Collection<Integer>} to a primitive {@code int[]} array.
+     *
+     * Returns an empty array if the collection is {@code null}.
+     *
+     * @param collection the given collection
+     * @return a primitive int[] array
+     */
+    public static int[] toIntArray(Collection<Integer> collection) {
+        if (collection == null) {
```
pveentjer: why this null protection?

pveentjer:
:+1:

hazelcast/src/main/java/com/hazelcast/util/CollectionUtil.java
Donnerbart: Because otherwise we'll get a NPE here?
`int[] collectionArray = new int[collection.size()];`

ahmetmircik: i prefer to see NPE, it is more informative than returning an empty array

hazelcast/src/main/java/com/hazelcast/util/CollectionUtil.java
```diff
@@ -97,19 +99,60 @@ public static boolean isNotEmpty(Collection collection) {
...
+     * Converts a {@link Collection<Long>} to a primitive {@code long[]} array.
+     *
+     * Returns an empty array if the collection is {@code null}.
+     *
+     * @param collection the given collection
+     * @return a primitive long[] array
+     */
+    public static long[] toLongArray(Collection<Long> collection) {
+        if (collection == null) {
+            return new long[0];
```
ahmetmircik: same here

ahmetmircik:
👍

PR #7892, opened 2016-04-01
Cleanup of Int2ObjectHashMap and Long2ObjectHashMap.

Donnerbart:
Aligned the format of both classes. Removed superfluous generics. Added note to JavaDoc that both types don't support null keys/values.

Superfluous generics were the main cause for warnings, because their local type `<V>`  hid the type from the main class.

If you compare both classes now, they just differ in their Integer and Long related parts.


Donnerbart:
run-lab-run

pveentjer:
:+1:

PR #7894, opened 2016-04-01
Use String constants in LocalMemoryStatsImpl

mtopolnik:
Duplicated string literals for JSON keys were used, which made it difficult to ascertain that all the literals match between the reading and writing side.


pveentjer:
:+1:

Issue #7810, opened 2016-03-22
NoClassDefFoundError: javax/cache/integration/CacheLoaderException


kuldeep9922:
Hi, 

We are trying to update hazelcast from 3.4 to 3.6.1, but we are repeatedly getting this issue.
`Factory method [public static com.hazelcast.core.HazelcastInstance com.hazelcast.instance.HazelcastInstanceFactory.newHazelcastInstance(com.hazelcast.config.Config)] threw exception; nested exception is java.lang.NoClassDefFoundError: javax/cache/integration/CacheLoaderException`

How can we resolve this issue?

jerrinot:
Hello @kuldeep9922,

thank you for reporting the issue. Can you please describe your environment a bit? Is it a web-application? Is it running in managed environment such as Spring, Java EE or perhaps OSGi? 

What Hazelcast services do you use? IMap? Anything else? What about JCache? 

Are you using client-member topology or member-to-member?

jerrinot:
@kuldeep9922: 
can you check your dependency tree? I believe is an older (non-final) version of JCache API on your classpath. Can you try to exclude it? 

If you are using JCache API then you can try overwrite it with the final version of JCache. 

These are Maven coordinates:
```xml
<dependency>
  <groupId>javax.cache</groupId>
  <artifactId>cache-api</artifactId>
  <version>1.0.0</version>
</dependency>
```

kuldeep9922:
We are working on a web application using Spring 3.1.1, Hibernate 4.1.4 and hazelcast's IMap service. 
It's working fine with hazelcast version 3.4 but throws exception when we change it to 3.6.1

jerrinot:
@kuldeep9922: can you check your classpath the make sure there is no jcache.jar (at any version) available?

kuldeep9922:
@jerrinot No, there is no jcache.jar available.

jerrinot:
@kuldeep9922: can you share full stacktrace of the exception?

kuldeep9922:
```
INFO: [LOCAL] [dev] [3.6.1] Picked Address[127.0.0.1]:35701, using socket ServerSocket[addr=/127.0.0.1,localport=35701], bind any local is false
Mar 22, 2016 12:34:09 PM com.hazelcast.spi.OperationService
INFO: [127.0.0.1]:35701 [dev] [3.6.1] Backpressure is disabled
[...]

PR #7896, opened 2016-04-01
JCache detection improvement

### LINKED ISSUE #7854: Improve JCache detection
We need to improve our JCache detection. 

Right now we are checking if the class `javax.cache.Caching` is available and if so then we create the JCache service and configure client protocol messages. However some older (pre-final) snapshots of JCache JARs have the `Caching` class available, but other classes are missing -> class not found exception on member start.

I think we should:
1. centralize JCache detection - now it's ad-hoc on at least 2 places
2. improve it to make it more reliable (check if the final JCache version in on a classpath)

vbekiaris:
This PR fixes #7854 by introducing a separate utility class `com.hazelcast.cache.impl.JCacheDetector` which handles JCache detection; specifically we make sure we locate a number of classes which are available since JCache 1.0.0-PFD but not in previous versions which caused NoClassDefFound exception as shown in #7810 (therefore when JCache 0.9 or 0.5 is in the classpath, Hazelcast will consider JCache is not available).

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.client.spi.impl;
 
+import com.hazelcast.cache.impl.JCacheDetector;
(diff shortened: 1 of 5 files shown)


jerrinot:
:+1:

serkan-ozal:
:+1:

PR #7903, opened 2016-04-04
Exceptions thrown during isMemberSafe() should go into finest level

jerrinot:
Justification: When there are topology changes then some exceptions are
expected -> the getFutureResult() method should log exceptions
into finest(), not into warn() level.

It's business as usual during partition migration.

Backport of #7902 into 3.6.2 branch.


metanet:
:+1:

ahmetmircik:
:+1:

PR #7906, opened 2016-04-04
InvocationContext introduced

pveentjer:
The goal of the InvocationContext is make the invocaiton a bit less dependant on the OperationService. Currently the OperationService is used as 'context' to get dependencies from and if you want an Invocation, you need an OperationServiceImpl -> NodeEngineImpl -> Node. So you can't use Invocations without having a cluster up and running.

The InvocationContext is the object with all dependencies for the invocations. Apart from reducing the need for anOperationService, a lot of trainwrecks have been removed by adding the dependency directly on the InvocationContext. This used to be a problem because not all dependencies exist when the OperationService is constructed, but the InvocationContext is created after these dependencies are created, so the InvocationContext can get these dependencies.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationContext.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationContext.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.spi.impl.operationservice.impl;
+
+import com.hazelcast.instance.Node;
+import com.hazelcast.internal.cluster.ClusterClock;
+import com.hazelcast.internal.cluster.ClusterService;
+import com.hazelcast.internal.partition.InternalPartitionService;
+import com.hazelcast.internal.serialization.InternalSerializationService;
+import com.hazelcast.internal.util.counters.MwCounter;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.nio.Address;
+import com.hazelcast.nio.ConnectionManager;
+import com.hazelcast.spi.NodeEngine;
+import com.hazelcast.spi.impl.executionservice.InternalExecutionService;
+import com.hazelcast.spi.impl.operationexecutor.OperationExecutor;
+import com.hazelcast.util.executor.ManagedExecutorService;
+
+/**
+ * The InvocationContext contains all 'static' dependencies for an Invocation; dependencies that don't change between
+ * invocations. All invocation specific dependencies/settings are passed in through the constructor of the invocation.
+ *
+ * This object should have no functionality apart from providing dependencies. So no methods should be added to this class.
+ *
+ * The goals of the InvocationContext are:
+ * <ol>
+ * <li>prevent a.b.c.d call in the invocation by pulling all dependencies in the InvocationContext</li>
+ * <li>reduce the need on having a cluster running when testing Invocations.</li>
+ * <lI>removed dependence on NodeEngineImpl. Only NodeEnigne is needed to set on Operation</lI>
+ * <li>reduce coupling to Node. Most calls point now direct to the right dependency, instead of needing
+ * to go through node/NodeEngine. This makes it easier in the future to get rid of Node</li>
+ * </ol>
+ */
+class InvocationContext {
+    ManagedExecutorService asyncExecutor;
+    ClusterClock clusterClock;
+    ClusterService clusterService;
+    ConnectionManager connectionManager;
+    InternalExecutionService executionService;
+    long defaultCallTimeoutMillis;
+    InvocationRegistry invocationRegistry;
+    InvocationMonitor invocationMonitor;
+    String localMemberUuid;
+    ILogger logger;
+    Node node;
+    NodeEngine nodeEngine;
+    InternalPartitionService partitionService;
+    OperationServiceImpl operationService;
+    OperationExecutor operationExecutor;
+    MwCounter retryCount;
+    InternalSerializationService serializationService;
+    Address thisAddress;
+}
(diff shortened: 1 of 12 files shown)


jerrinot:
I see little benefits it its current state. 

If you manage to rid-of the dependency on the `Node` class then yes - very good improvement. But now I see it as another `Node` or `NodeEngine` object. It does not reduce coupling  - `Invocation` still depends on `Node` - and it barely improves cohesion.

pveentjer:
There is a difference between Node/NodeEngine and InvocationContext. The first 2 are god objects, they know about everything and everybody knows about them. With the InvocationContext this is different; only the operationservice.impl know s about it. So therefor you prevent the tight coupling that Node/NodeEngine object cause. Therefor your conclusion about this being another Node/NodeEngine object is questionable.

Dependency to Node can't be prevent due to logic being implement on node like active/state that is not available somewhere else. Cutting out Node is a lot more work and needs to be done at a different part of the system.

pveentjer:
There is no dependence on NodeEngineImpl anymore; just NodeEngine (to inject into the operation). And there is a single dependence on Node ---> getState. There is currently no replacement for that for the time being. Once this has been moved to a more sensible class, it is very easy to remove dependence on Node.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationContext.java
```diff
@@ -0,0 +1,67 @@
...
+ * The goals of the InvocationContext are:
+ * <ol>
+ * <li>prevent a.b.c.d call in the invocation by pulling all dependencies in the InvocationContext</li>
+ * <li>reduce the need on having a cluster running when testing Invocations.</li>
+ * <lI>removed dependence on NodeEngineImpl. Only NodeEnigne is needed to set on Operation</lI>
+ * <li>reduce coupling to Node. Most calls point now direct to the right dependency, instead of needing
+ * to go through node/NodeEngine. This makes it easier in the future to get rid of Node</li>
+ * </ol>
+ */
+class InvocationContext {
```
noctarius: Wouldn't it be better to make all those fields final and provide a constructor to initialize them? Agree no getters necessary but still I think for the sake of code style they should be provided. I know we most probably disagree with this :)

Donnerbart: I'm very unhappy with the style of this class :(
* CheckStyle and SonarQube will create visibility errors for each field
* you'll never know if the context has been constructed completely
* every user of the context can overwrite any field, even by accident!
* everyone can overwrite any field at any time
* did I mention that everyone can overwrite... we should not have this in the SPI!!!
* refactoring will be very painful with this code style (been there, done that with Simulator)
* refactoring is risky on that code style (e.g. you add a new field and no compiler error if you forget to initialize a field in a test, you don't know about)
* this can lead to NPE in production code or tests
* this can lead to silent unexpected behavior if there are null checks and "failover" code paths are used (been there, done that with Simulator)
* the 20 something line construction looks not better than a constructor with all fields
* a constructor with so many parameters will cause a CheckStyle issue as well, but it's a single one for a method instead of one for each field (and the suppression is totally fine for me, since it's a context class)
* IntelliJ will write all the boilerplate code for you, even if you have to change anything
* IntelliJ has shortcuts to change the order of fields, constructor parameters etc.
* InjtelliJ will most probably do everything for you which may lead you to this style
* ...

So `final` for mandatory fields is a must for me. A constructor is a must for me. `package private` is a must for me, `private` and getter are a nice to have for me (although I cannot understand why getters are bad, IntelliJ to the rescue).

pveentjer:
I disagree with a constructor. A class with this many fields is terrible to construct with a constructor since it is difficult to see what the constructor argument means since there are so many. Above 5/6 arguments and constructors become problematic.

pveentjer:
@Donnerbart I disagree with the review comments. I'll address each one of them:

*  CheckStyle and SonarQube will create visibility errors for each field: incorrect. The fields are package friendly; just like the fields on the OperationServiceImpl are.

* you'll never know if the context has been constructed completely. There is only 1 place where this context gets constructed and that is in the OperationServiceImpl. If you have a failure constructing in tests, your test will fail.

* every user of the context can overwrite any field, even by accident!    everyone can overwrite any field at any time. Nonsense argument. I can change any field on any object anyway. Context is not a public object everyone can mess with; it is completely confined within the operationservice.impl package. 

 
*    did I mention that everyone can overwrite... we should not have this in the SPI!!! Nonsense argument. It isn't exposed in the SPI, so there is no point in bringing it up.

*  refactoring will be very painful with this code style (been there, done that with Simulator). Nonsense. Such an object is a lot easier to deal with than a 20 argument constructor. After the object has been constructed, there is no difference between constructor based or just changing the fields.

* refactoring is risky on that code style (e.g. you add a new field and no compiler error if you forget to initialize a field in a test, you don't know about) this can lead to NPE in production code or tests. There is a single place of construction of this object in production code. It isn't an object used all over the code base. 



*    the 20 something line construction looks not better than a constructor with all fields. It is a lot better to read since java is lacking named arguments. 


* a constructor with so many parameters will cause a CheckStyle issue as well, but it's a single one for a method instead of one for each field (and the suppression is totally fine for me, since it's a context class)
    IntelliJ will write all the boilerplate code for you, even if you have to change anything
    IntelliJ has shortcuts to change the order of fields, constructor parameters etc.
    InjtelliJ will most probably do everything for you which may lead you to this style
   
Nonsense.

pveentjer:
I see there is quite some resistance, but once green I'll merge this PR since I see that the code is a lot easier to understand. Since I'm the main caretaker of the operation implementation, I don't care for consensus.

pveentjer:
run-lab-run

jerrinot:
@pveentjer: 
- You had sent a PR. @Donnerbart, @noctarius and my raised concerns. You merged it anyway. Without having a single :+1: This is plain wrong and it sets a very bad example. 
- You are using Aeron as an argument, but it has a different meaning there. It's used as a (media-driver) configuration class. It's even specified in a JavaDoc. As soon as `MediaDriver` is [constructed](https://github.com/jerrinot/Aeron/blob/3676b9f366b6c17e4bb41424df18b96786806f46/aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/MediaDriver.java#L150-L150) then the `Aeron.Context` is fully initialized. No exceptions. 
- Also I don't see any non-private field in `Aeron.Context`.

pveentjer:
We already had discussion. I stick with my choice to get this PR merged.

enesakar:
Following sentence is dangerous:
`Since I'm the main caretaker of the operation service implementation, I don't care for consensus.`
I can not imagine the situation where everyone behaves like this.
Rules should be applied to everyone.

mtopolnik:
I do not find the mutable package-private fields problematic. It isn't ideal, but given their limited scope it doesn't cause too much damage in practice. The same IntelliJ can very quickly reveal any point in the code where a field is getting assigned. I do disagree with a PR being merged in spite of unresolved criticism.

tombujok:
This PR should be now analysed together with the other PR that Peter submitted to address the concerns of the reviewers raised here and to reach a consensus that would satisfy everyone.
https://github.com/hazelcast/hazelcast/pull/8072

Here's my thumbs up for both PRs (as a whole): 👍 (once green)

PR #7907, opened 2016-04-04
Fail faster for client invocations with timeout

sancar:
operation with timeout like, tryPut, tryLock etc do not need to
wait client invocation timeout in case of a retryable failure.

Made a change on client invocation so that users timeout overrides
the client invocation timeout for these kind of APIs.


sancar:
I will close this pr since solution requires a trade-off and we choose the leave the behaviour as-is. 

For example for `tryPut` , my first intention was to return `false` when timeout given by `tryPut` exceeded on client side. If I do that, I will broke the semantics of `tryPut` return type. Before `false` meant there is a lock in that key in that time it is not released. With this change, user can not be sure if lock is there or not. 
Another example, for `Queue.offer(time)`,  `false` means queue is full.

Second option is to throw `OperationTimeoutException` in this case. But that means introducing a new behavior. Our API users can not just check the return value anymore. It also needs to add a `try-catch` block for `OperationTimeoutException`. 

 In our current implementation, as long as there are no cluster changes, user timeout is respected. And when there is cluster changes, there is a possibility that user timeout exceeded. Since it is in rare cases, I think not introducing a new logic and leaving the behaviour as-is  is a better.

Issue #5444, opened 2015-06-03
Lost data if a node is killed during data repartition


elmor:
Hi,

I am evaluating HazelCast 3.42 against Coherence and I experienced a strange behavior with Hazelcast.

 I have a cluster of two nodes(I will name them node1 and node2) and two clients 

I have a map name “toto” (I joined the hazelcast configuration file server.xml) 

1.       I inserted 100000 entries into the map=> all is ok.

2.       I kill the node  node2. All is ok I found the 100000 entries into the node1.

3.       I restart the node2.  While node2 is synchronizing with node1, I kill again node2.

4.       Node1 only have 51919 elements L I lost 48081 entries L (The pb seems to be due to the partitioning)

 

Ho  How can I avoid to lose elements of a node is killed during synchronisation ?

 

I can see the following errors in the log file of my server

Members [1] {

            Member [192.88.65.124]:5701 this

}

> 

####<Jun 2, 2015 3:03:43 PM CEST> <Info> <com.hazelcast.partition.InternalPartitionService> <dcaclx01.dns21.socgen> <Hazel1> <hz._hzInstance_1_dev.migration> <<anonymous>> <> <> <1433250223501> <BEA-000000> <[192.88.65.124]:5701 [dev] [3.4.2] Partition balance is ok, no need to re-partition cluster data... > 

####<Jun 2, 2015 3:05:14 PM CEST> <Info> <Health> <dcaclx01.dns21.socgen> <Hazel1> <weblogic.GCMonitor> <<anonymous>> <> <> <1433250314135> <BEA-310002> <79% of the total memory in the server is free.> 

####<Jun 2, 2015 3:05:41 PM CEST> <Info> <com.hazelcast.nio.tcp.SocketAcceptor> <dcaclx01.dns21.socgen> <Hazel1> <hz._hzInstance_1_dev.IO.thread-Acceptor> <<anonymous>> <> <> <1433250341255> <BEA-000000> <[192.88.65.124]:5701 [dev] [3.4.2] Accepting socket connection from /192.88.65.123:2247> 

[...]

PR #7911, opened 2016-04-05
Improve migration/replication system 

mdogan:
This work improves migration/replication system to makes it safe against data loss. 

For design details please see: https://hazelcast.atlassian.net/wiki/display/COM/Avoid+Data+Loss+on+Migration+-+Solution+Design

_Additional details can be seen in individual commit messages._

Fixes https://github.com/hazelcast/hazelcast/issues/5444

EE counter-part: https://github.com/hazelcast/hazelcast-enterprise/pull/850

Q. E. D.


hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -259,6 +260,13 @@ private void initializeListeners(Config config) {
                 known = true;
             }
 
+            if (listener instanceof InternalMigrationListener) {
+                final InternalPartitionServiceImpl partitionService =
+                        (InternalPartitionServiceImpl) nodeEngine.getPartitionService();
+                partitionService.setInternalMigrationListener((InternalMigrationListener) listener);
```
gurbuzali: what if there are more than one internal listener, latter one will override ?

metanet: This is not user api or spi. It is only done for testing. We use InternalMigrationListener to execute a specific test scenario deterministically within our integration tests so we register only a single internal listener and perform the test with it.

metanet:
This PR has received yet another conflict because of yet another refactoring in the middle of the release...

metanet:
I will not wait for another 8 days for the review. After the conflict is solved, I will merge the PR if there is no review for a couple of following days.

gurbuzali:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/PartitionMigrationEvent.java
```diff
@@ -54,8 +61,33 @@ public int getPartitionId() {
...
+        return currentReplicaIndex;
+    }
+
+    /**
+     * Gets the index of the partition replica that current member will own after migration is committed.
+     * This index will be -1 if partition replica will be moved from current member completely.
+     *
+     * @return index of the partition replica that current member will own after migration is committed
+     */
+    public int getNewReplicaIndex() {
```
jerrinot: could you please describe how the `MigrateAwareService` is supposed to use this information?

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/ExecutionServiceImpl.java
```diff
@@ -167,8 +167,7 @@ private ManagedExecutorService createExecutor(String name, int poolSize, int que
             Node node = nodeEngine.getNode();
             String internalName = name.startsWith("hz:") ? name.substring(BEGIN_INDEX) : name;
             HazelcastThreadGroup hazelcastThreadGroup = node.getHazelcastThreadGroup();
-            PoolExecutorThreadFactory threadFactory = new PoolExecutorThreadFactory(hazelcastThreadGroup,
-                    hazelcastThreadGroup.getThreadPoolNamePrefix(internalName));
+            PoolExecutorThreadFactory threadFactory = new PoolExecutorThreadFactory(hazelcastThreadGroup, internalName);
```
jerrinot: well spotted!

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaStateChecker.java
```diff
@@ -0,0 +1,386 @@
...
+    private Future invoke(Operation operation, int replicaIndex, int partitionId) {
+        OperationService operationService = nodeEngine.getOperationService();
+        return operationService.createInvocationBuilder(SERVICE_NAME, operation, partitionId)
+                .setTryCount(INVOCATION_TRY_COUNT)
+                .setTryPauseMillis(INVOCATION_TRY_PAUSE_MILLIS)
+                .setReplicaIndex(replicaIndex)
+                .invoke();
+    }
+
+    // TODO: VISIBILITY PROBLEM! Replica versions are updated & read only by partition threads!
```
jerrinot: can you please elaborate the problem?

metanet: This will be fixed by graceful shutdown work. Currently, this part reads replica versions from a non-partition thread without guaranteeing thread safety. Normally they are accessed only within partition threads.

hazelcast/src/main/java/com/hazelcast/spi/PartitionMigrationEvent.java
mdogan: `MigrateAwareService` should decide to keep or remove, all or some part of its data, by looking `currentReplicaIndex`, `newReplicaIndex` and `migrationEndpoint`.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaStateChecker.java
jerrinot: just to double-check I understand it properly - right now a given `partitionId` is always access by a same thread -> there is no visibility problem. Is this correct?

The warning serves for other developers to avoid touching this from other threads, am I right?

mdogan: This problem is NOT introduced by this PR, it's already there in the current master. We just spotted it and added a TODO. Also it's already fixed by our graceful shutdown work, as Basri mentioned.

hazelcast/src/main/java/com/hazelcast/spi/PartitionMigrationEvent.java
jerrinot: do you think some sort of guidance help to implementers? 
by looking at our current implementations the logic seem to be quite similar.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaStateChecker.java
jerrinot: that's fine, I'm trying to understand it, that's all.

mdogan: @jerrinot: correct. normally replica-version array is accessed/modified by only it's specific partition thread. But usage here, in the method below the TODO, is not thread-safe. Version array is accessed out side of partition threads.

jerrinot: ✓ thanks!

hazelcast/src/main/java/com/hazelcast/spi/PartitionMigrationEvent.java
mdogan: Logic is similar. But now there are more parameters than before. I can add an example to the class  (`MigrateAwareService`) javadoc.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalMigrationListener.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+package com.hazelcast.internal.partition.impl;
+
+
+import com.hazelcast.internal.partition.MigrationInfo;
+
+import java.util.EventListener;
+
+public abstract class InternalMigrationListener implements EventListener {
```
jerrinot: JavaDoc explaining the purpose (=deterministic testing) of this class would be nice.

hazelcast/src/main/java/com/hazelcast/spi/PartitionMigrationEvent.java
jerrinot: an example would be excellent! 

before the logic was quite simple: a sending side clears its partition on commit, receiving side clears it on rollback. that's very intuitive. 

now you need to take backup count into consideration.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationPlanner.java
```diff
@@ -0,0 +1,281 @@
...
+    private static final boolean TRACE = false;
+
+    interface MigrationDecisionCallback {
+        void migrate(Address source, int sourceCurrentReplicaIndex, int sourceNewReplicaIndex, Address destination,
+                     int destinationCurrentReplicaIndex, int destinationNewReplicaIndex);
+    }
+
+    private final Set<Address> verificationSet = new HashSet<Address>();
+
+    // checkstyle warnings are suppressed intentionally because the algorithm is followed easier within a single coherent method.
```
jerrinot: I understand your motivation behind this decision, but could you include comment on individual cases/branches? 
It's quite hard to follow it unless you have a good understanding of possible migrations cases.

metanet: We have test cases for that

jerrinot: I am aware of that, but having it in-lined in the code would really readability. 

This is usually achieved by having descriptive method names. However you decided for a single method handling all cases - I understand your choice, but one can feel kinda lost. 

A few words would do it. there is already a comment like this. see `// IT IS A SHIFT DOWN`

metanet: I am still not into it but we can add more comments anyway.

Imho, the whole algorithm is not very easy to follow by only checking the code. If someone really wants to grasp the algorithm completely, he should check the test cases and the design document. We have already provided the full description of the algorithm there.

mdogan: I'll add more comments and javadoc. just waiting for review to complete.

jerrinot:
:+1: some minor comments.
It's truly an epic work, hats off to gentlemen!

bwzhang2011:
@mdogan, would you mind sparing some time and take a look at issue https://github.com/hazelcast/hazelcast/issues/7987, could the latest improvement for migration solve such problem ?

PR #7912, opened 2016-04-05
Fix semaphore dead member handling mechanism

mdogan:
- Semaphore cleanup should run on partition thread regardless of
current member being owner or backup of partition.
- If it's owner of partition, it should send backup too.
- This is required to guarantee to execute cleanup even for the cases
where no one owns the partition.


mdogan:
verify

metanet:
👍

PR #7913, opened 2016-04-06
Dedicated priority generic operation threads

pveentjer:
The OperationExecutor is now by default configured with 1 extra priority generic thread. This thread will only process generic priority operations; which still can be picked up by another regular generic thread if the priority thread for whatever reason is busy. But it can't happen that if only regular generic tasks are running that a priority generic task is not going to be picked up.

This should increase reliability since generic threads are used for all kinds of purposes, e.g. launching and syncing on queries. And stuff like heartbeats, cluster operations etc should be processed asap.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/impl/OperationExecutorImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/impl/OperationExecutorImpl.java
@@ -36,10 +36,13 @@ import com.hazelcast.spi.impl.operationexecutor.OperationRunnerFactory;
 import com.hazelcast.spi.impl.operationservice.impl.operations.Backup;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
+import java.util.concurrent.LinkedBlockingQueue;
+
 import static com.hazelcast.internal.metrics.ProbeLevel.MANDATORY;
 import static com.hazelcast.internal.properties.GroupProperty.GENERIC_OPERATION_THREAD_COUNT;
 import static com.hazelcast.internal.properties.GroupProperty.PARTITION_COUNT;
 import static com.hazelcast.internal.properties.GroupProperty.PARTITION_OPERATION_THREAD_COUNT;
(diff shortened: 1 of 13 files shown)


jerrinot:
:+1: on green

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/impl/OperationExecutorImpl.java
```diff
@@ -112,17 +118,19 @@ public OperationExecutorImpl(GroupProperties properties,
...
+        if (threadCount <= 0) {
             // default generic operation thread count
             int coreSize = Runtime.getRuntime().availableProcessors();
-            genericThreadCount = Math.max(2, coreSize / 2);
+            threadCount = Math.max(2, coreSize / 2);
         }
 
-        OperationRunner[] operationRunners = new OperationRunner[genericThreadCount];
+
+        OperationRunner[] operationRunners = new OperationRunner[threadCount + priorityThreadCount];
```
Donnerbart: Should we have a check that `priorityThreadCount` is not negative? We sometimes use `-1` to disable features, which could lead to a silent misconfiguration here.

Donnerbart:
Beside one minor comment :+1: (once green)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/impl/OperationExecutorImpl.java
pveentjer: Good one.

However I would like this to get solved on the property. So when you define a property, you should be able to set a validator on it. In a lot of cases, the same property is being pulled out many times; e.g. partition count. Where should this property be validated?

Donnerbart: It would be a good improvement for the properties, indeed. For now I would just add it after this block:
```
int threadCount = properties.get(...)
if (threadCount <= 0) {
  (...)
}
checkNotNegative("useful text", priorityThreadCount);
```
I think we have something in the Preconditions for this kind of check? So it should be a one-liner for now.

PR #7921, opened 2016-04-06
Cleanup of MapEntries to reduce Map.Entry litter

Donnerbart:
This is an experiment to reduce the litter caused by `java.util.AbstractMap$SimpleImmutableEntry` in `MapEntries`. In many cases we don't need the data as `Map.Entry`, but use key and value separately. So we can save the creation of those wrapper objects until we return something like a map to the user.

Especially the `IMap.putAll()` with backups will profit from this, since we pass around huge `MapEntries` instances between the nodes. Let's have a look where the litter is created. Each creation of `MapEntries` creates an `AbstractMap.SimpleImmutableEntry<Data, Data>` wrapper object for each entry. This happens on the initial creation on the calling node and on each deserialization on the executing nodes. For `IMap.putAll()` there is an additional deserialization for each configured backup:

```
4 members, 4 partitions, 1 sync backup

IMap.putAll() with 500 entries on Member [172.16.18.1]:5702

(local execution on Member [172.16.18.1]:5702, so no deserialization)
PutAllOperation deserialized with 138 entries
PutAllOperation deserialized with 119 entries
PutAllOperation deserialized with 126 entries
PutAllOperation on Member [172.16.18.1]:5703 with 138 entries
PutAllOperation on Member [172.16.18.1]:5704 with 119 entries
PutAllOperation on Member [172.16.18.1]:5702 with 117 entries
PutAllOperation on Member [172.16.18.1]:5701 with 126 entries

PutAllBackupOperation deserialized with 126 entries
PutAllBackupOperation deserialized with 119 entries
PutAllBackupOperation deserialized with 117 entries
PutAllBackupOperation deserialized with 138 entries
PutAllBackupOperation on Member [172.16.18.1]:5702 with 126 entries
PutAllBackupOperation on Member [172.16.18.1]:5703 with 119 entries
PutAllBackupOperation on Member [172.16.18.1]:5704 with 117 entries
PutAllBackupOperation on Member [172.16.18.1]:5701 with 138 entries

=> 1383 wrapper objects
```
```
4 members, 4 partitions, 2 sync backups

IMap.putAll() with 500 entries on Member [172.16.18.1]:5702

(local execution on Member [172.16.18.1]:5702, so no deserialization)
PutAllOperation deserialized with 117 entries
PutAllOperation deserialized with 119 entries
PutAllOperation deserialized with 126 entries
PutAllOperation on Member [172.16.18.1]:5701 with 119 entries
PutAllOperation on Member [172.16.18.1]:5702 with 138 entries
PutAllOperation on Member [172.16.18.1]:5704 with 117 entries
PutAllOperation on Member [172.16.18.1]:5703 with 126 entries

PutAllBackupOperation deserialized with 117 entries
PutAllBackupOperation deserialized with 119 entries
PutAllBackupOperation deserialized with 138 entries
PutAllBackupOperation deserialized with 126 entries
PutAllBackupOperation on Member [172.16.18.1]:5703 with 117 entries
PutAllBackupOperation on Member [172.16.18.1]:5704 with 119 entries
PutAllBackupOperation on Member [172.16.18.1]:5701 with 138 entries
PutAllBackupOperation on Member [172.16.18.1]:5702 with 126 entries

PutAllBackupOperation deserialized with 138 entries
PutAllBackupOperation deserialized with 117 entries
PutAllBackupOperation deserialized with 126 entries
PutAllBackupOperation deserialized with 119 entries
PutAllBackupOperation on Member [172.16.18.1]:5701 with 126 entries
PutAllBackupOperation on Member [172.16.18.1]:5704 with 138 entries
PutAllBackupOperation on Member [172.16.18.1]:5701 with 117 entries
PutAllBackupOperation on Member [172.16.18.1]:5703 with 119 entries

=> 1883 wrapper objects
```
On top of that the `MapEntries` are often created without an initial size, so the internal `ArrayList` is resized frequently, which is quite expensive.

I changed the implementation from `List<Map.Entry<Data, Data>> entries` to two separate `List<Data>` for keys and values. I also added the initial size to the constructor, which is known in a lot of cases! The only headache is the initial creation of the `MapEntries` per partition in the `MapProxySupport.putAllInternal()` method. We don't know the distribution and the entries per partition may vary in size, depending on the user input. We also have a broad range from one to hundreds of thousands of possible entries.

I came up with two different strategies. The first creates a fair amount of litter to create the `MapEntries` instances (one `int[mapSize]` and a `LinkedList$Node` per entry), comparable to the old litter (see `invokePutAllWithAnalysis()`). The second one theoretically creates almost no litter (see `invokePutAllWithEstimation()`), but needs a good estimation about the partition distribution. I solved with a new property, since we cannot know this upfront, so it's tuneable. In all cases the `MapEntries` don't create the `AbstractMap.SimpleImmutableEntry<Data, Data>` litter anymore and are very efficient once being serialized.

| Name | Master | Cleanup |
| --- | ---: | ---: |
| Duration per operation | ~14010 µs | ~13370 µs |
| GC Count | 2918 | 2755 |
| Total GC Time | 3min 15s 268ms | 3min 5s 854ms |
| TLAB count | 6,967,200 | 6,620,614 |
| Allocation Rate for TLAB | 591.53 MB/s | 582.20 MB/s |

```
50000 operations done in  700546 ms (14010 µs per op avg) (master)...
vs.
50000 operations done in  668506 ms (13370 µs per op avg) (cleanup)...
```

There is a big variation in the runtime though. I got several runs on the master branch with 766576 ms (15331 µs per op avg) and also runs on the cleanup that didn't went that well. So I guess the pure throughput is not a good measurement. Maybe the invocation cleanup Peter is working on will help here. Nevertheless the allocation profile and GC times are constantly better with the cleanup, so we will profit from this PR.

**Allocation pressure**
master
cleanup
The `java.util.AbstractMap$SimpleImmutableEntry` allocation pressure of 8.18% is gone.

If I interpret the affected classes correctly we may also benefit when using `EntryProcessors` between members (due to `AbstractMultipleEntryOperation`). When using clients I think there is no change, since I still create the `AbstractMap.SimpleImmutableEntry` instances in the new `MapEntries.putAllToList()` method.

**Question**
Will this break the client protocol in any way? The serialization of `MapEntries`, `PutAllOperation`, `PutAllBackupOperation` etc. have been changed. I hope this is not the case, since I could re-use all task and codec classes as they are. I think we're just changing the member-to-member communication, which should be fine for a minor release.

If we want to merge this, I'll have to prepare an EE PR as well.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -80,12 +80,12 @@ import com.hazelcast.spi.OperationService;
 import com.hazelcast.spi.impl.BinaryOperationFactory;
 import com.hazelcast.spi.partition.IPartition;
 import com.hazelcast.spi.partition.IPartitionService;
+import com.hazelcast.spi.properties.HazelcastProperties;
 import com.hazelcast.spi.serialization.SerializationService;
 import com.hazelcast.util.FutureUtil;
 import com.hazelcast.util.IterableUtil;
 import com.hazelcast.util.ThreadUtil;
 
-import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -93,6 +93,7 @@ import java.util.EventListener;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -103,10 +104,12 @@ import java.util.concurrent.TimeUnit;
 import static com.hazelcast.cluster.memberselector.MemberSelectors.LITE_MEMBER_SELECTOR;
 import static com.hazelcast.cluster.memberselector.MemberSelectors.NON_LOCAL_MEMBER_SELECTOR;
 import static com.hazelcast.map.impl.MapService.SERVICE_NAME;
+import static com.hazelcast.spi.properties.GroupProperty.MAP_PUT_ALL_INITIAL_SIZE_FACTOR;
 import static com.hazelcast.util.ExceptionUtil.rethrow;
 import static com.hazelcast.util.FutureUtil.logAllExceptions;
 import static com.hazelcast.util.IterableUtil.nullToEmpty;
 import static com.hazelcast.util.Preconditions.checkNotNull;
+import static java.lang.Math.ceil;
 import static java.lang.Math.min;
 import static java.util.Collections.singleton;
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -137,9 +140,13 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
     // not final for testing purposes
     protected MapOperationProvider operationProvider;
 
+    private final float putAllInitialSizeFactor;
+
     protected MapProxySupport(String name, MapService service, NodeEngine nodeEngine) {
         super(nodeEngine, service);
         this.name = name;
+
+        HazelcastProperties properties = nodeEngine.getProperties();
 
         this.mapServiceContext = service.getMapServiceContext();
         this.mapContainer = mapServiceContext.getMapContainer(name);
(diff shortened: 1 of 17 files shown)


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -674,6 +671,7 @@ protected void getAllObjectInternal(List<Data> keys, List<Object> resultingKeyVa
      */
     protected void putAllInternal(Map<?, ?> map) {
         int partitionCount = partitionService.getPartitionCount();
+        int initialSize = map.size() / partitionCount;
```
Donnerbart: This is a rude "educated" guess to prevent constant resizing of the arrays in `MapEntries`.

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -673,9 +670,10 @@ protected void getAllObjectInternal(List<Data> keys, List<Object> resultingKeyVa
      * Probably this can be optimized in the future by making use of an PartitionIterating operation.
      */
     protected void putAllInternal(Map<?, ?> map) {
-        int partitionCount = partitionService.getPartitionCount();
-
         try {
+            int partitionCount = partitionService.getPartitionCount();
+            int initialSize = map.size() / partitionCount;
```
Donnerbart: This is an educated guess to create the `MapEntires` with a proper size. If we don't do this we will suffer badly from constant resizing. The `MapEntires` for each partitionId are anyway created lazy, so the worst case would be a very unbalanced input map, where this `initialSize` is not sufficient and we'll get a bunch of resizes. But at least we won't waste memory on partitions, which are not used at all.

Donnerbart:
run-lab-run

PR #7925, opened 2016-04-07
Make PagingPredicate generic

LoneRifle:
Follows up on #7840, allowing users who extend this class to likewise take advantage of now-exposed generics

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapIssueTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapIssueTest.java
@@ -205,7 +205,7 @@ public class ClientMapIssueTest extends HazelcastTestSupport {
             map.put(i, i);
         }
 
-        final PagingPredicate predicate = new PagingPredicate(pageSize);
+        final PagingPredicate<Integer, Integer> predicate = new PagingPredicate<Integer, Integer>(pageSize);
         predicate.nextPage();
 
         final Set<Map.Entry<Integer, Integer>> entries = map.entrySet(predicate);
@@ -229,7 +229,7 @@ public class ClientMapIssueTest extends HazelcastTestSupport {
             map.put(i, i);
         }
 
-        final PagingPredicate predicate = new PagingPredicate(pageSize);
+        final PagingPredicate<Integer, Integer> predicate = new PagingPredicate<Integer, Integer>(pageSize);
         predicate.nextPage();
 
         final Collection<Integer> values = map.values(predicate);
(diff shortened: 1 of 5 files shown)


Donnerbart:
run-lab-run

LoneRifle:
Located checkstyle error with modified suppressions.xml for Windows. Will issue a separate PR for that.

Donnerbart:
run-lab-run

pveentjer:
Seems like an OK change, but I'm not an expert of the paging predicate. @gurbuzali can you have a look at this one.

gurbuzali:
looks fine :+1:

pveentjer:
There is one last concern, backwards compatibility. Just to make sure that we don't run into an unexpected generics incompatibility problems. @mtopolnik @jerrinot WDYT?

mtopolnik:
`Comparator<Map.Entry<K, V>>` is stricter than the previous `Comparator<Map.Entry>>`.

LoneRifle:
That shouldn't be a problem if PagingPredicate is instantiated without type parameters; you can still supply a `Comparator<Map.Entry>`. Or am I mistaken?

pveentjer:
The question is regarding binary compatibility. So can people replace 3.6 by 3.7 without needing to recompile the code.

LoneRifle:
Given generics in Java is implemented using type erasure, I do not foresee any binary compatibility breaks, as type parameters do not affect the binary identity of a method signature.

vbekiaris:
You can always use [sigtest](https://wiki.openjdk.java.net/display/CodeTools/sigtest) to verify binary compatibility. I have already generated a [signature file for Hazelcast 3.6.0 - hazelcast-3.6.sig.zip](https://github.com/hazelcast/hazelcast/files/252394/hazelcast-3.6.sig.zip), you should be able to check for binary compatibility with something like this cmd line:
```bash
$ java -jar sigtestdev.jar Test -Mode bin -B -static -classpath \
$JDK_HOME/jre/lib/rt.jar:$YOUR_PATH_HERE/hazelcast/target/hazelcast-3.7-SNAPSHOT.jar:$HOME/.m2/repository/com/google/code/findbugs/annotations/3.0.0/annotations-3.0.0.jar  \
-filename $YOUR_SIGFILE_PATH_HERE/hazelcast-3.6.sig
```

tombujok:
It's a follow up on the https://github.com/hazelcast/hazelcast/pull/7840
I'll give my thumbs up if the binary compatibility is checked (IMHO it should be fine)

LoneRifle:
Verified at https://gist.github.com/LoneRifle/ae7c883319493f1c86243466cacf873f . Note that two passes had to be done to account for the addition of IMap.setAsync.

tombujok:
👍 
@ahmetmircik could you have a look too?

tombujok:
Actually we have two thumbs-up already. Merging!

PR #7927, opened 2016-04-07
Cleanup of PutAllBackupOperation

Donnerbart:
Another simplification in the code, this time for the `PutAllBackupOperation`.

If we have backups enabled we transfer *all* entries into the `PutAllBackupOperation`. There is no need to copy them one by one into a new list, we can just re-use the existing `MapEntries` instance from the `PutAllOperation`. On the receiving side this will just create a single additional `MapEntries` wrapper object, which we saved on the sending side `List<Map.Entry<Data, Data>>`.

The performance measurement showed a massive improvement, which I don't really believe. I think that was a very bad master vs. a very good cleanup run. But at least there is surely no negative impact :)
```
50000 operations done in  842382 ms (16847 µs per op avg) (master)...
vs.
50000 operations done in  749940 ms (14998 µs per op avg) (cleanup)...
```
Memory profiles in JFR look the same.

I will create the EE counterpart after the review of this PR, but want to have your feedback first.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllBackupOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllBackupOperation.java
@@ -17,10 +17,9 @@
 package com.hazelcast.map.impl.operation;
 
 import com.hazelcast.core.EntryView;
-import com.hazelcast.map.impl.EntryViews;
+import com.hazelcast.map.impl.MapEntries;
 import com.hazelcast.map.impl.record.Record;
 import com.hazelcast.map.impl.record.RecordInfo;
-import com.hazelcast.map.impl.record.Records;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.Data;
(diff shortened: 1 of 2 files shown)


Donnerbart:
verify

Donnerbart:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllBackupOperation.java
```diff
@@ -52,18 +52,18 @@ public PutAllBackupOperation() {
     @Override
     public void run() {
         boolean wanEnabled = mapContainer.isWanReplicationEnabled();
-        for (int i = 0; i < entries.size(); i++) {
-            final RecordInfo recordInfo = recordInfos.get(i);
-            final Map.Entry<Data, Data> entry = entries.get(i);
-            final Record record = recordStore.putBackup(entry.getKey(), entry.getValue());
-            Records.applyRecordInfo(record, recordInfo);
+        int i = 0;
+        for (Map.Entry<Data, Data> entry : entries) {
```
pveentjer: This needs to get addressed as soon as this one is merged:
https://github.com/hazelcast/hazelcast/pull/7921

pveentjer:
Looks good to me. 

:+1:

jerrinot:
:+1:

PR #7935, opened 2016-04-11
Client protocol version 1 compatibility tests

sancar:
ftl templates that is used to generate these tests can be found at
https://github.com/sancar/hazelcast-client-protocol/tree/testGeneratorFiles


mustafaiman:
👍

hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ReferenceObjects.java
```diff
@@ -0,0 +1,123 @@
...
+                if (!isEqual(o1, o2)) {
+                    return false;
+                }
+            }
+            return !(e1.hasNext() || e2.hasNext());
+        }
+        return a.equals(b);
+    }
+
+    public static boolean aBoolean = true;
```
ihsandemir: What about the other primitives listed at https://hazelcast.atlassian.net/wiki/display/HZC/Hazelcast+Open+Binary+Client+Protocol+-+Version+1.0 , e.g. short, float, double, .. Should we add them here?

sancar: I covered all that are used in codecs. Others are defined in protocol but not used apperantly. Since they are not used, I can not add them to these tests. I think they should be added when they are used by a codec.

ihsandemir:
👍  OK after adding the ErrorCodes verification test.

pveentjer:
Since there are 2 thumbs up, this PR can be merged.

PR #7915, opened 2016-04-06
[DONT MERGE] Performance improvements

pveentjer:

PR #7939, opened 2016-04-12
Nonblocking invocationfutures

pveentjer:
The server and client invocation future are now dumb and make use of the AbstractInvocationFuture. The idea behind the AbstractInvocationFuture is:
- it is lock free
- it relies on LockSupport.park/unpark for notification. 
- in case of the ClientInvocationFuture it creates less litter since no LinkedList for callbacks is created. 
- there is less need for lock deflation since the new approach doesn't rely on a intrinsic lock. Therefore you get less gc hiccups (lock deflation is done when a the system does a gc).

This is one of the PRD's for increased performance.

Some performance impressions:
On the old lab using atomiclong.get and pure remote calls; they is roughly a 15% throughput improvement. Using pure local calls, there is roughly a 30% throughput improvement. Also latency goes down a lot; e.g. in case of the local one, max latencies goes from >1s to <30ms. 

Performance improvements on client are less spectacular; +/- 5% increased throughput. 

This is a piece of worked chopped of from:
https://github.com/hazelcast/hazelcast/pull/7915

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
@@ -1,0 +1,406 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.spi.impl;
+
+import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.spi.InternalCompletableFuture;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.concurrent.locks.LockSupport;
+
+import static com.hazelcast.util.ExceptionUtil.rethrow;
+import static com.hazelcast.util.Preconditions.isNotNull;
+import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
+import static java.util.concurrent.locks.LockSupport.park;
+import static java.util.concurrent.locks.LockSupport.parkNanos;
+import static java.util.concurrent.locks.LockSupport.unpark;
+
+/**
+ * todo:
+ * - thread value protection
+ *
+ * The long term goal is that this whole class can be ripped out and replaced by the
+ * {@link java.util.concurrent.java.util.concurrent.CompletableFuture}. So we need to be very careful with adding more
+ * functionality to this class because the more we add, the more difficult the replacement will be.
+ *
+ * @param <V>
+ */
+public abstract class AbstractInvocationFuture<V> implements InternalCompletableFuture<V> {
+
+    static final Object VOID = "VOID";
+
+    // Reduce the risk of rare disastrous classloading in first call to
+    // LockSupport.park: https://bugs.openjdk.java.net/browse/JDK-8074773
+    static {
+        Class<?> ensureLoaded = LockSupport.class;
+    }
+
+    private static final AtomicReferenceFieldUpdater<AbstractInvocationFuture, Object> STATE =
+            newUpdater(AbstractInvocationFuture.class, Object.class, "state");
+
+    protected final Executor defaultExecutor;
+    protected final ILogger logger;
+
+    /**
+     * This field contain the state of the future. Either the future is not complete and the state is:
+     * <ol>
+     * <li>VOID: no response is available.</li>
+     * <li>Thread instance: no response is available and a thread has blocked on completion (e.g. future.get)</li>
+     * <li>{@link ExecutionCallback} instance: no response is available and 1 {@link #andThen(ExecutionCallback)}
+     * was done using the default executor</li>
+     * <li>{@link WaitNode} instance: in case of multiple andThen registrations or future.gets or andThen with custom Executor
+     * </li>
+     * </ol>
+     *
+     * If the state is anything else, it is completed.
+     *
+     * The reason why a single future.get or registered ExecutionCallback doesn't create a WaitNode is that we don't want to cause
+     * additional litter since most of our API calls are a get or a single ExecutionCallback.
+     *
+     * The state field is replaced using a cas, so registration or setting a response is an atomic operation and therefor not
+     * prone to data-races. There is no need to use synchronized blocks.
+     */
+    private volatile Object state = VOID;
+
+    public AbstractInvocationFuture(Executor defaultExecutor, ILogger logger) {
+        this.defaultExecutor = defaultExecutor;
+        this.logger = logger;
+    }
+
+    boolean compareAndSetState(Object oldState, Object newState) {
+        return STATE.compareAndSet(this, oldState, newState);
+    }
+
+    protected final Object getState() {
+        return state;
+    }
+
+    @Override
+    public final boolean isDone() {
+        return isDone(state);
+    }
+
+    private boolean isDone(final Object state) {
+        if (state == null) {
+            return true;
+        }
+
+        return !(state == VOID
+                || state.getClass() == WaitNode.class
+                || state instanceof Thread
+                || state instanceof ExecutionCallback);
+    }
+
+    protected void onInterruptDetected() {
+    }
+
+    @Override
+    public final boolean cancel(boolean mayInterruptIfRunning) {
+        return false;
+    }
+
+    @Override
+    public final boolean isCancelled() {
+        return false;
+    }
+
+    @Override
+    public final V join() {
+        try {
+            // this method is quite inefficient when there is unchecked exception, because it will be wrapped
+            // in a ExecutionException, and then it is unwrapped again.
+            return get();
+        } catch (Throwable throwable) {
+            throw rethrow(throwable);
+        }
+    }
+
+    @Override
+    public final V getSafely() {
+        return join();
+    }
+
+    @Override
+    public final V get() throws InterruptedException, ExecutionException {
+        Object response = registerWaiter(Thread.currentThread(), null);
+        if (response != VOID) {
+            // no registration was done since a value is available.
+            return resolveAndThrow(response);
+        }
+
+        boolean interrupted = false;
+        try {
+            for (; ; ) {
+                park();
+                if (isDone()) {
+                    return resolveAndThrow(state);
+                } else if (Thread.interrupted()) {
+                    interrupted = true;
+                    onInterruptDetected();
+                }
+            }
+        } finally {
+            restoreInterrupt(interrupted);
+        }
+    }
+
+    @Override
+    public final V get(final long timeout, final TimeUnit unit)
+            throws InterruptedException, ExecutionException, TimeoutException {
+        Object response = registerWaiter(Thread.currentThread(), null);
+        if (response != VOID) {
+            return resolveAndThrow(response);
+        }
+
+        long deadlineNanos = System.nanoTime() + unit.toNanos(timeout);
+        boolean interrupted = false;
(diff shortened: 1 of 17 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -445,6 +445,11 @@ void notifyBackupComplete() {
 
         // We are the lucky ones since we just managed to complete the last backup for this invocation and since the
         // pendingResponse is set, we can set it on the future.
+        complete(pendingResponse);
+    }
+
+    private void complete(Object pendingResponse) {
```
Donnerbart: Since you pass different type of objects into the method (response, value, CALL_TIMEOUT, HEARTBEAT_TIMEOUT) I would use the same parameter name as the `AbstractInvocationFuture::complete`, which would be `value`.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -461,7 +466,7 @@ private void handleRetry(Object cause) {
         operationService.invocationRegistry.deregister(this);
```
Donnerbart: Is it okay to call the deregister method multiple times? If `future.interrupted` is `true` two lines below, the `complete(INTERRUPTED)` will call this method again.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -0,0 +1,395 @@
...
+    protected Object resolve(Object value) {
+        return value;
+    }
+
+    protected abstract V resolveAndThrow(Object state) throws ExecutionException, InterruptedException;
+
+    protected abstract TimeoutException newTimeoutException(long timeout, TimeUnit unit);
+
+    /**
+     * Registers a waiter (so a Thread that blocks on the future completion, or a {@link ExecutionCallback} notified
```
Donnerbart: Incomplete sentence in JavaDoc.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -0,0 +1,395 @@
...
+    private Object registerWaiter(Object waiter, Executor executor) {
+        WaitNode waitNode = null;
+        for (; ; ) {
+            final Object oldState = state;
+            if (isDone(oldState)) {
+                return oldState;
+            }
+
+            Object newState;
+            if (oldState == VOID && (executor == null)) {
```
Donnerbart: Superfluous parenthesis around `(executor == null)`.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -0,0 +1,395 @@
...
+                // Nothing is syncing on this future, so instead of creating a WaitNode, we just try
+                // the cas the thread (so no extra litter)
```
Donnerbart: Should this be `to cas the thread`? Otherwise I didn't get the meaning of the comment.

Donnerbart:
I just checked the test failure `ClientInvocation.executionCallback_FailOnShutdown()`, which fails because its `ExecutionCallback::onFailure` checks for a `t.getCause() instanceof HazelcastClientNotActiveException`, but with this PR the `Throwable` itself is the `HazelcastClientNotActiveException`, not its cause.

So it seems that there is an exception encapsulation gone. In general I like that, but it also may break customer code, since they might have the same check like our test does in their projects.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
pveentjer: I'm going to leave it as response since that is the primary naming approach within the invocation. Check responses package.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
pveentjer: It is fine; but good spotting anyway. It was always there, however it wasn't visible since it was done in the future.complete method and I pulled it out the reduce coupling between future and invocation system. I'll move the unregister to the !interrupted branch so there is no duplicate deregistration.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
pveentjer: Done

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
pveentjer: fixed

pveentjer:
Appearently there is a mismatch between exception handling on server and client side which currently gets visible. I need to have a look and I'll fix it so that it reflects the serverside. That is what you get with code duplication.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -0,0 +1,399 @@
...
+import static com.hazelcast.util.ExceptionUtil.rethrow;
+import static com.hazelcast.util.Preconditions.isNotNull;
+import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
+import static java.util.concurrent.locks.LockSupport.park;
+import static java.util.concurrent.locks.LockSupport.parkNanos;
+import static java.util.concurrent.locks.LockSupport.unpark;
+
+/**
+ * todo:
+ * - thread value protection
```
jerrinot: what does this mean?

Donnerbart: If you call `complete()` with a `Thread` as value this will crash, since the algorithm cannot distinct between a waiting caller thread and a thread as result in the state field. A result thread has to be wrapped, which is not yet done.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -0,0 +1,399 @@
...
+    public final V get(final long timeout, final TimeUnit unit)
+            throws InterruptedException, ExecutionException, TimeoutException {
+        Thread currentThread = Thread.currentThread();
+
+        Object response = registerWaiter(currentThread, null);
+        if (response != VOID) {
+            return resolveAndThrow(response);
+        }
+
+        // todo: what about overflow? Imagine someone calls with Long.MAX_VALUE
```
jerrinot: this should be fixed - as I can imagine it's quite possible. 

the same goes for `get(someValue, unit)` where `unit.toNanos(someValue)` causes overflow.

jerrinot:
```java
        long currentTimeNanos = System.nanoTime();
        long deadlineNanos = currentTimeNanos + unit.toNanos(timeout);
        if (deadlineNanos < currentTimeNanos) { //overflow
            deadlineNanos = Long.MAX_VALUE;
        }
```

the overflow in unit conversion will be tricker.

jerrinot: obviously this assumes timeout is a positive number. perhaps we should have a check for this anyway.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -0,0 +1,399 @@
...
+            return true;
+        }
+
+        return !(state == VOID
+                || state.getClass() == WaitNode.class
+                || state instanceof Thread
+                || state instanceof ExecutionCallback);
+    }
+
+    protected void onInterruptDetected() {
```
jerrinot: why we have different implementations for client and members?

jerrinot:
what's the expected behaviour when a thread waiting on `future.get()` is interrupted? 

it appears the client implementation completes with `InterruptedException` immediately(?), but the member implementation is waiting until a next retry?

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -0,0 +1,399 @@
...
+            }
+
+            if (compareAndSetState(oldState, newState)) {
+                // we have successfully registered
+                return VOID;
+            }
+        }
+    }
+
+    void unregisterWaiter(Thread waiter) {
```
jerrinot: is this method needed at all? it's quite complex and it mingles with the (concurrent) waiting list -> it's error-pron

However what are consequences if you do not un-register a waiter at all? a thread with receive a permit at some point in time. but that is absolutely OK as spurious permits/wake-up are normal.

hazelcast/src/test/java/com/hazelcast/spi/impl/AbstractInvocationFuture_AbstractTest.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public TestFuture(Executor executor, ILogger logger) {
+            super(executor, logger);
+        }
+
+        @Override
+        protected void onInterruptDetected() {
+            interruptDetected = true;
+            complete(new InterruptedException());
```
jerrinot: when `complete` return false then the interrupt signal can be lost, right?

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
pveentjer: Exactly

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
pveentjer: The overflow remains obscure to me. Also in the java.util.concurrent I see the same, e.g. from the SynchronousQueue:

```
  final long deadline = timed ? System.nanoTime() + nanos : 0L;     
```

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
pveentjer: Because they behave differently. Because a shared implementation is used, this inconsistency becomes visible. I would like to get this fixed; but not as part of this PR. There are also other inconsistencies e.g. andThen exception handling. On member, the ExecutionExeception envelope is removed, on the client not.

jerrinot: fair enough. can you please open a github issue documenting the inconsistency. to make sure it won't be forgotten.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
pveentjer: The expected behavior due to interruption was and is fragile. It needs to be redesigned in the future; not as part of this PR. This PR maintains the current status.

In short:
- on member the interrupt flag is set; if an operation is retried (blocking op) then the retry is not done but an interruptedexception is set.
- on the client: the thread gets an interruptedexception, but the actual remote operation keeps running. This can cause problems with lock.lock where the client thinks it is interrupted, but on the server the lock is acquired. 

Cleaning this mess up is not part of this PR.

pveentjer: The method is needed, if you do a timed get, you could get a memory leak if you do it repeatedly.

hazelcast/src/test/java/com/hazelcast/spi/impl/AbstractInvocationFuture_AbstractTest.java
pveentjer: Currently the interrupt is not lost because the interrupt flag is always restored on the thread, no matter the exception being thrown. When the InterruptedException is thrown, the flag is also set. If this is a good or a bad thing is a topic in its own; because due to our exception wrapping on join, the actual InterruptedException gets effectively hidden.

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
jerrinot: so what would you suggest to do about it?

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
pveentjer: https://github.com/hazelcast/hazelcast/issues/7963

jerrinot:
:+1:

PR #7944, opened 2016-04-12
Fixes a deadlock on internal executor when listeners are used

sancar:
A similar deadlock was detected and fixed here
hazelcast#7673

Both have the same reason. When remote invocation is waited
indefinetely on internal executor since response also same via
same executor and invocation monitor also runs on same executor,
there is change of deadlock.

As a fix, I have introduced a dedicated single thread executor to
 be used by listener registration logic. This way, internal
 executor will not be blocked.

I have introduced a property to make the test fail constantly.
"hazelcast.client.internal.executor.pool.size" . We should not
need to expose this to user right now. This is just for testing.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/config/ClientProperty.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/config/ClientProperty.java
@@ -83,7 +83,12 @@ public enum ClientProperty implements HazelcastProperty {
      * eventually be removed when the experimental marker is removed.</p>
      * <p>Discovery SPI is <b>disabled</b> by default</p>
      */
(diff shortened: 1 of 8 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
```diff
@@ -70,6 +70,8 @@ public boolean isDone() {
 
     @Override
     public ClientMessage get() throws InterruptedException, ExecutionException {
+        //Client should wait remote invocation indefinitely on internal thread
```
pveentjer: Smells like a hacky assert since you are relying on a threadname.

sancar:
run-lab-run

PR #7945, opened 2016-04-12
Fixes a deadlock on internal executor when listeners are used

sancar:
A similar deadlock was detected and fixed here
hazelcast#7673

Both have the same reason. When remote invocation is waited
indefinetely on internal executor since response also same via
same executor and invocation monitor also runs on same executor,
there is change of deadlock.

As a fix, I have introduced a dedicated single thread executor to
 be used by listener registration logic. This way, internal
 executor will not be blocked.

I have introduced a property to make the test fail constantly.
"hazelcast.client.internal.executor.pool.size" . We should not
need to expose this to user right now. This is just for testing.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
@@ -28,6 +28,7 @@ import com.hazelcast.core.Member;
 import com.hazelcast.core.MemberAttributeEvent;
 import com.hazelcast.core.MembershipEvent;
 import com.hazelcast.nio.Address;
+import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.UuidUtil;
 
 import java.util.Collection;
@@ -36,6 +37,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 
@@ -44,7 +46,6 @@ public class ClientSmartListenerService extends ClientListenerServiceImpl implem
     private final Set<Member> members = new HashSet<Member>();
     private final Map<ClientRegistrationKey, Map<Address, ClientEventRegistration>> registrations
             = new ConcurrentHashMap<ClientRegistrationKey, Map<Address, ClientEventRegistration>>();
-    private final Object listenerRegLock = new Object();
     private String membershipListenerId;
 
     public ClientSmartListenerService(HazelcastClientInstanceImpl client,
@@ -53,21 +54,29 @@ public class ClientSmartListenerService extends ClientListenerServiceImpl implem
     }
 
     @Override
-    public String registerListener(ListenerMessageCodec codec, EventHandler handler) {
-        String userRegistrationId = UuidUtil.newUnsecureUuidString();
-        synchronized (listenerRegLock) {
+    public String registerListener(final ListenerMessageCodec codec, final EventHandler handler) {
+        Future<String> future = registrationExecutor.submit(new Callable<String>() {
+            @Override
+            public String call() {
+                String userRegistrationId = UuidUtil.newUnsecureUuidString();
 
-            ClientRegistrationKey registrationKey = new ClientRegistrationKey(userRegistrationId, handler, codec);
-            registrations.put(registrationKey, new ConcurrentHashMap<Address, ClientEventRegistration>());
-            try {
-                for (Member member : this.members) {
-                    invoke(registrationKey, member.getAddress());
+                ClientRegistrationKey registrationKey = new ClientRegistrationKey(userRegistrationId, handler, codec);
+                registrations.put(registrationKey, new ConcurrentHashMap<Address, ClientEventRegistration>());
+                try {
+                    for (Member member : members) {
+                        invoke(registrationKey, member.getAddress());
+                    }
+                } catch (Exception e) {
+                    deregisterListener(userRegistrationId);
+                    throw new HazelcastException("Listener can not be added", e);
                 }
-            } catch (Exception e) {
-                deregisterListener(userRegistrationId);
-                throw new HazelcastException("Listener can not be added", e);
+                return userRegistrationId;
             }
-            return userRegistrationId;
+        });
+        try {
+            return future.get();
+        } catch (Exception e) {
+            throw ExceptionUtil.rethrow(e);
         }
     }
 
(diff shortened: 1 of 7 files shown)


pveentjer:
This PR is going to lead to problems with this PR:
https://github.com/hazelcast/hazelcast/pull/7939
With the new InvocationFuture, the ClientInvocationFuture will be made very dumb and there is no chance to override get methods. 

Like i said in the other PR; I don't think these asserts are good. What is the purpose of these asserts?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
```diff
@@ -72,6 +72,8 @@ public boolean isDone() {
 
     @Override
     public ClientMessage get() throws InterruptedException, ExecutionException {
+        //Client should wait remote invocation indefinitely on internal thread
```
pveentjer: This comment doesn't make sense. The comments states 'it shuold wait indefinitely on internal thread', but the assert makes sure it isn't an internal thread.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
```diff
@@ -143,22 +138,13 @@ public void shutdown() {
...
-                    members.add(member);
-                    for (ClientRegistrationKey registrationKey : registrations.keySet()) {
-                        try {
-                            invoke(registrationKey, member.getAddress());
-                        } catch (Exception e) {
-                            logger.warning("Listener " + registrationKey + " can not added to new member " + member);
-                        }
-                    }
-                }
+        synchronized (listenerRegLock) {
```
pveentjer: Why is this lock needed? My spider senses go off when I see a global lock with such a huge amount of 'external' code. Would it not be a lot easier to get rid of the lock have a single dedicated thread that takes care of all logic; and actor like thread? This way all modifications go through that object and you don't need to deal that much with synchronization.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
sancar: Comment is wrong. I will add a `not` somewhere in between. Good catch

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
sancar: This is not a new logic. When I was first implementing this one, I start with a single thread. It seemed more complicated to reviewers back then, and I switched to locks.  In short, lock is not related to this pr.

sancar:
@pveentjer I hunted similar bugs last couple of months. Asserts are there for checking we do not do the same mistake again.  But I accept that using thread names are not a good idea,  I will remove the asserts. But I want to put a check somewhere. If there is a good suggestion about checking the same thing any other way, please go ahead.

sancar:
verify

pveentjer:
If you want to reject a certain type of thread, I would do an instanceof check. So your threads could implement a certain interface like OperationHostileThread.

sancar:
@pveentjer Thanks for the suggestion. I will look into it. For this prd, I already see all clients tests are passed with when assert in-place. So, I think, it is safe to postpone this to another pr.

pveentjer:
I'm a bit unclear about the following:

''' When remote invocation is waited
indefinetely on internal executor since response also same via
same executor and invocation monitor also runs on same executor,
there is change of deadlock.'''

1: the responses are send on their own dedicated response thread.
2: why is the invocation monitor not running on its own thread? Pool of threads is nice if you don't care about responsiveness, but can be very problematic otherwise.

sancar:
@pveentjer 
1: Well not exactly response. Problem was with future.andThen part. We are running callbacks of internal invocations in `internal` executor. 

2: We have partition fetching, heartbeat monitor, invocation monitor, establishing connection async, callbacks of internal operations etc... All of these currently running on the internal executor. I think it is simpler to maintain single executor then all services having its own executor/thread. Unless there is a strong reason to make it separate, we did not separate so far. There is one exception to this. We have a single thread `clusterListenerThread` which is responsible for listening memberlist, choosing and establishing owner connection when owner connection disconnected. If this thread closes we are shutting down the client. It is kind of special.

pveentjer:
It perhaps may be simpler for a java code perspective, but you are introducing a lot of risk because if a thread from a threadpool is being 'hogged' you can easily run into a deadlock situation as you are currently running into. Also you can run into all kinds of non responsiveness issues due to depletion. That is why on the serverside threads are being isolated so we can 'guarantee' no funky stuff is happening; unlike with a general purpose executor.

sancar:
This prd solves a critical issue respecting current architecture. Let's
discuss the architecture change outside of this prd. Is that OK ?

On Wed, Apr 13, 2016, 15:39 Peter Veentjer <notifications@github.com> wrote:

> It perhaps may be simpler for a java code perspective, but you are
> introducing a lot of risk because if a thread from a threadpool is being
> 'hogged' you can easily run into a deadlock situation as you are currently
> running into. Also you can run into all kinds of non responsiveness issues
> due to depletion. That is why on the serverside threads are being isolated
> so we can 'guarantee' no funky stuff is happening; unlike with a general
> purpose executor.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/hazelcast/hazelcast/pull/7945#issuecomment-209409650>
>
-- 
Sancar Koyunlu
Software Engineer, Hazelcast
sancar@hazelcast.com

ihsandemir:
:+1:  (After adding correct executor to andThen in ClientSmartListenerService::registerListenerToMember method)

mustafaiman:
👍

ihsandemir:
Note: We discussed the issue with Sancar. To avoid event losing until connection establishment and also letting the client logic simple and easier, it may be preferable to do this solution which handles most of the problem at the server side:

1. On client initiation client connects to all members by default.
2. User sends a registration request to a single server (a new request type).
3. Cluster registers this client to all members as is isLocal flag is set to false.
4. If any member joins the cluster, just transfer all registrations to the new member as well (this may exist for isLocal = false implementation)
5. Just buffer events to the client if client is not yet connected to the member. When connection is setup, deliver the events. Put a queue limit for this buffering.

We may define aPRD for this.

sancar:
Another thing we discussed that locks are not solving the problem, because we are using callbacks Locks are not guarding that part, sine they are handled async. As we discussed, when we first writing this logic before with @ihsandemir @cangencer and as @pveentjer suggested, I will try introducing a single thread and removing locks for solving this issue.

mustafaiman:
👍

ihsandemir:
👍

Issue #1719, opened 2014-02-05
executeOnEntries with predicate doesn't honor index


torkeld:
When executing an entry processor with an index aware predicate the index is never used, instead the predicate is applied to the entire entry set.

wimdeblauwe:
There is a unit test in `com.hazelcast.map.EntryProcessorTest` that references this issue. However, the test is not correct. It does not test for using the index at all. The test will still succeed if you remove the index from the configuration.

The correct test should be like the one in the reproducer at https://github.com/torkeld/hazelcast-issue-reproducer where the unit test checks if the predicate is called or not. If the index is used, the predicate should not be called.

LoneRifle:
Fixed in #8053, with testfix proposed by @wimdeblauwe implemented in 4b37b0e

PR #7946, opened 2016-04-12
Fix #1156/#1719 with @torkeld - Ensure that IndexAwarePredicates actually use indices for IMap.executeOnEntries

### LINKED ISSUE #1156: executeOnEntries in callable ignores index
When a task is submitted to all members, and in the call an executeOnEntries is invokes with a predicate that is based on an index - the index is ignored and a "full scan" of the "local" members is performed.

LoneRifle:
* If the PartitionIteratingOperation has a PartitionWideEntryWithPredicateOperationFactory, try to query the indices associated with the map that the factory is scoped for.
* In the factory itself, if the index service returns keys given a predicate query, retain them and issue a MultipleEntryOperation containing those keys instead of a PartitionWideEntryWithPredicateOperation.

Fixes #1156 and #1719, which are identical.

Credit to @torkeld for the original commit at torkeld@0b15cc1

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionWideEntryWithPredicateOperationFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionWideEntryWithPredicateOperationFactory.java
@@ -17,18 +17,26 @@
 package com.hazelcast.map.impl.operation;
 
 import com.hazelcast.map.EntryProcessor;
+import com.hazelcast.map.impl.MapService;
 import com.hazelcast.nio.ObjectDataInput;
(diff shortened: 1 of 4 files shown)


LoneRifle:
I would like to give credit to @torkeld for the original commit that fixes the issue.

vbekiaris:
Thanks for the awesome contribution! There is some work being done in parallel that overlaps with this PR, so they will be considered in conjunction within the 3.7 release timeframe.

ahmetmircik:
@LoneRifle, thanks for your PR, as @vbekiaris mentioned, we fixed this issue [here](https://github.com/hazelcast/hazelcast/pull/8053). 

Thanks for rising attention to this issue.

PR #7949, opened 2016-04-13
OperationTimeoutException improvement

pveentjer:
Includes the current time in the exception. On multiple occassions I have seen this stacktrace and could
not figure out the current time easily since only this exception is being copied into error message.

So it now includes the Clock time. This makes interpreting the other 'time' values a lot easier.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
@@ -289,19 +289,23 @@ final class InvocationFuture<E> implements InternalCompletableFuture<E> {
             sb.append(invocation.op.getClass().getSimpleName())
                     .append(" invocation failed to complete due to operation-heartbeat-timeout. ");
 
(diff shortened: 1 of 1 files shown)


Donnerbart:
I personally would pull out `Clock.currentTimeMillis()` into a variable. But in this case it will not have a lot of relevance (values are not used for critical business logic), so it should be fine to have consecutive calls.

:+1: (once green)

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationFuture.java
```diff
@@ -289,20 +289,24 @@ private Object newOperationTimeoutException(boolean heartbeatTimeout) {
             sb.append(invocation.op.getClass().getSimpleName())
                     .append(" invocation failed to complete due to operation-heartbeat-timeout. ");
 
+            sb.append("Current time:").append(timeToString(currentTimeMillis())).append(". ");
```
jerrinot: missing space after `:` ?

PR #7950, opened 2016-04-13
InvocationFuture minor improvements

pveentjer:
1: always using System.currentTimeMillis instead of Clock at some parts and System at others
2: improved time toString (a shorter but easier to read time)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/StringUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/StringUtil.java
@@ -17,6 +17,9 @@
 package com.hazelcast.util;
 
 import java.nio.charset.Charset;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 import java.util.Locale;
 
 /**
@@ -32,7 +35,7 @@ public final class StringUtil {
     /**
      * Points to the System property 'line.separator'.
      */
-    public static final String LINE_SEPARATOR =  System.getProperty("line.separator");
+    public static final String LINE_SEPARATOR = System.getProperty("line.separator");
 
     private static final Locale LOCALE_INTERNAL = Locale.ENGLISH;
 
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/util/StringUtil.java
```diff
@@ -127,4 +130,17 @@ public static String lowerCaseInternal(String s) {
...
+    /**
+     * Returns a String representation of the time.
+     *
+     * This method is not particularly efficient since it generates a ton of litter.
+     *
+     * @param timeMillis time in millis
+     * @return the String
+     */
+    public static String timeToString(long timeMillis) {
+        DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
```
Donnerbart: It's a shame that `SimpleDateFormat` is not thread-safe, would save some litter :(

pveentjer: It isn't only the litter of the DateFormat object, but also during the format method, waste is generated. For its current purposes it is fine :)

pveentjer:
run-lab-run

pveentjer:
verify

Donnerbart:
Judging `Clock.currentTimeMillis()` vs. `System.currentTimeMillis()` always gives me a hard time.

The changes in `newOperationTimeoutException()` should be fine, since it should align the timestamps with the ones in the logs, which should make comparisons easier. It's anyway just for log output, so it should not have any critical impact.

The original code in `runAsynchronous()` looks broken anyway, since both clocks were used for the same variable. So with a big offset in `Clock` a `TimeoutException` could be thrown immediately, if I'm not wrong. So the change should fix this. Since this is anyway about a relative time comparison, the `System` clock should be fine.

pveentjer:
In production Clock and System time are exactly the same; however for testing you can tinker with the clock to skew.

Donnerbart:
Okay, I just looked this up. I though `Clock` was the same as `ClusterClock`, but apparently those are two different things.

PR #7952, opened 2016-04-13
Improved OperationTimeoutException

pveentjer:
it includes the operation lastheartbeat of the member. This gives some insight what is happening
with the member communicaton. Imagine a fat packet is jamming the pipeline, then of
course the operation will timeout, but the heartbeat-packet from that member will also fall behind.

So if both member and invocation are very far behind; it is likely some kind of environmental
issue. If the member is up to date and the invocation is not, then appearently something happened
to the operation.

There is also a minor change in the operation heartbeat-system. In the new approach, every member will send to every other member periodically a list of callid's. No matter if this list is empty. This empty list triggers the update of the last heartbeat for that given member. This way we can always rely on the lastheartbeat for a given member; no matter if operations are running or not. This system has nothing to do with the cluster heartbeats.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
@@ -16,9 +16,11 @@
 
 package com.hazelcast.spi.impl.operationservice.impl;
 
+import com.hazelcast.core.Member;
 import com.hazelcast.core.MemberLeftException;
 import com.hazelcast.instance.HazelcastThreadGroup;
 import com.hazelcast.instance.MemberImpl;
+import com.hazelcast.internal.cluster.ClusterService;
 import com.hazelcast.internal.metrics.MetricsProvider;
 import com.hazelcast.internal.metrics.MetricsRegistry;
 import com.hazelcast.internal.metrics.Probe;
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -73,6 +76,7 @@
     private final ILogger logger;
     private final ScheduledExecutorService scheduler;
     private final Address thisAddress;
+    private final ConcurrentMap<Address, AtomicLong> lastHeartbeatPerMember = new ConcurrentHashMap<Address, AtomicLong>();
```
jerrinot: does it have to be concurrent map? it's always run on the same thread, isn't it?

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/DummyOperation.java
```diff
@@ -18,8 +19,17 @@ public DummyOperation(Object value) {
...
     }
 
+    public DummyOperation setDelayMillis(int delayMillis) {
+        this.delayMillis = delayMillis;
+        return this;
+    }
+
     @Override
     public void run() throws Exception {
+        if(delayMillis>0){
```
Donnerbart: You are lucky that CheckStyle isn't running on test files ;)

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
pveentjer: Also other threads; any thread can make an investigation regarding the last heartbeat of that given member (for the invocationfuture error message for example).

Donnerbart:
Test cases seem to match the added code. Can't see any big troubles.

:+1: (once green)

pveentjer:
From Jaromir '''Yes. Sorry, I am AFK. Feel free to merge it'''

PR #7954, opened 2016-04-14
[DONT MERGE] WIP

ahmetmircik:


ahmetmircik:
run-lab-run

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
```diff
@@ -248,6 +241,41 @@ public boolean remove(Data key) {
     }
 
     @Override
+    public Object mapKeyToMarker(Data key) {
+        NearCacheRecord markerRecord = new NearCacheRecord(key, MARKER_VALUE);
+        NearCacheRecord previous = cache.put(key, markerRecord);
```
jerrinot: shouldn't this be CAS too? what if some other thread managed to populate the near cache? then you will remove completely good value.

ahmetmircik:
fix will be handled differently later, no need this PR now.

PR #7960, opened 2016-04-15
Introduction of async atomic long API

### LINKED ISSUE #7957: Introduce async versions of IAtomicLong methods
`IAtomicLong` provides synchronous methods only, even though async versions already exist and are available to hazelcast members via `AsyncAtomicLong` interface. Lack of public async API for `IAtomicLong` is impeding hazelcast integrations e.g. https://github.com/vert-x3/vertx-hazelcast/issues/28.
- Async method versions should be introduced in `IAtomicLong`, using the `Async` suffix convention (as already used in `IMap` & `ICache`). For example `addAndGet`'s async version will be named `addAndGetAsync`.
- `AsyncAtomicLong` interface will be marked as deprecated, to be removed in a future version

vbekiaris:
This PR fixes #7957 by providing async variants of existing methods in `IAtomicLong`. All async methods return `ICompletableFuture`, enabling reactive style invocation of execution callbacks by means of `andThen(...)`. `@Beta` annotation was removed from `ICompletableFuture` which exists since version 3.2. `AsyncAtomicLong` is marked as deprecated and will be removed in a future version.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientAtomicLongProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientAtomicLongProxy.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.client.proxy;
 
+import com.hazelcast.client.impl.ClientMessageDecoder;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.impl.protocol.codec.AtomicLongAddAndGetCodec;
 import com.hazelcast.client.impl.protocol.codec.AtomicLongAlterAndGetCodec;
@@ -31,6 +32,7 @@ import com.hazelcast.client.impl.protocol.codec.AtomicLongGetCodec;
 import com.hazelcast.client.impl.protocol.codec.AtomicLongIncrementAndGetCodec;
 import com.hazelcast.client.impl.protocol.codec.AtomicLongSetCodec;
 import com.hazelcast.core.IAtomicLong;
+import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.IFunction;
 
 import static com.hazelcast.util.Preconditions.isNotNull;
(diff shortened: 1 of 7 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/atomiclong/ClientAtomicLongTest.java
```diff
@@ -77,6 +84,25 @@ public void test() throws Exception {
 
     }
 
+    @Test
```
pveentjer: We definitely need to fix the client testing so that these tests don't need to be duplicated from the server.

hazelcast/src/main/java/com/hazelcast/concurrent/atomiclong/AtomicLongProxy.java
```diff
@@ -66,6 +67,11 @@ public long addAndGet(long delta) {
     }
 
     @Override
```
pveentjer: Could you change the implementation so that the deprecated 'asyncXXX' are forwarding to the new 'XXXAsync'? This way it is a lot easier to get rid of the deprecated methods instead of having more reliance on them.

vbekiaris: Yes, sure.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientAtomicLongProxy.java
```diff
@@ -149,6 +236,88 @@ public void set(long newValue) {
...
+    @Override
+    public ICompletableFuture<Long> decrementAndGetAsync() {
+        ClientMessage request = AtomicLongDecrementAndGetCodec.encodeRequest(name);
+        return invokeOnPartitionAsync(request, DECREMENT_AND_GET_DECODER);
+    }
+
+    @Override
+    public ICompletableFuture<Long> getAsync() {
+        ClientMessage request = AtomicLongGetCodec.encodeRequest(name);
+        return invokeOnPartitionAsync(request, DECREMENT_AND_GET_DECODER);
```
sancar: Should have been GET_ASYNC_DECODER ?

sancar:
👍

vbekiaris:
Thanks for your comments, above issues are now addressed.

pveentjer:
On the serverside there is one 'true' method, the async one. And the sync calls this and then calls join. I love it.

On the client side the logic for sync is copied. Would it not make sense to do on the client the same as on the server.

```
ICompletableFuture fooAsync(){
   .... stuff.
}

void foo(){
    fooAsync().join()
}
```

vbekiaris:
Nice point, one thing to consider is whether `join()` method should be pulled up from `InternalCompletableFuture` to `ICompletableFuture` (which is returned by `IAtomicLong` API). It is a convenient wrapper on `Future.get` and checked exception handling. What do you think?

pveentjer:
I agree, but don't know if we should do it already. Ideally we get rid of the ICompletable/InternalCompletableFuture completely and replace it by the java.util.concurrent.CompletableFuture. 

In this PR the ClientInvocationFuture implements InternalCompletableFuture btw:
https://github.com/hazelcast/hazelcast/pull/7939

vbekiaris:
Great, so I guess we should wait until #7939 is merged, then de-duplicate client code to match server side implementation (implementation in async methods only, sync methods implemented via `*Async.join()`) and then check & merge this PR?

pveentjer:
Lets merge this one and address the duplication issue when my pr is merged.

pveentjer:
:+1:

PR #7962, opened 2016-04-15
Removes record-stats object ref from record object. From now on, Reco…

ahmetmircik:
…rd interface also extends RecordStatistics


hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordStatistics.java
```diff
@@ -16,28 +16,17 @@
 
 package com.hazelcast.map.impl.record;
 
-import com.hazelcast.nio.serialization.DataSerializable;
-
 /**
  * Represents statistics of a {@link Record}.
  */
-public interface RecordStatistics extends DataSerializable {
```
pveentjer: Do we need to keep this interface at all?

hazelcast/src/test/java/com/hazelcast/map/SizeEstimatorTest.java
```diff
@@ -57,7 +57,7 @@ public void testSinglePut() throws InterruptedException {
 
     @Test
     public void testExactHeapCostAfterUpdateWithMultipleBackupNodes() throws InterruptedException {
-        long expectedPerEntryHeapCost = 172L;
+        long expectedPerEntryHeapCost = 156L;
```
pveentjer: Does this not depend on the JVM settings? E.g. if you use compressed or non compressed oops then removing a pointer will reduce size with either 4 or 8 bytes.

pveentjer:
Good work Ahmet. Removing an unwanted object will reduce memory footprint and will also be cache friendlier.

hazelcast/src/main/java/com/hazelcast/map/impl/record/AbstractRecord.java
```diff
@@ -106,10 +105,7 @@ public void setHits(long hits) {
     public long getCost() {
         final int objectReferenceInBytes = 4;
```
pveentjer: This depends on JVM settings. 4 bytes is compressed oops, so if non compressed oops are used, the cost is incorrect.

hazelcast/src/main/java/com/hazelcast/map/impl/record/RecordStatistics.java
ahmetmircik: completely removed it.

hazelcast/src/test/java/com/hazelcast/map/SizeEstimatorTest.java
ahmetmircik: Yes it depends on JVM settings, there is a merged PR to fix that issue for one of 3.5 branches, i will also port it to master.
https://github.com/hazelcast/hazelcast/commit/b90f8dbd9ebac29a02ecfe6c6f44bc1a1e62269d

hazelcast/src/main/java/com/hazelcast/map/impl/record/AbstractRecord.java
ahmetmircik: same with https://github.com/hazelcast/hazelcast/pull/7962#discussion_r60057326

hazelcast/src/test/java/com/hazelcast/map/SizeEstimatorTest.java
pveentjer: How do you dynamically determine the pointer size?

pveentjer:
Is there any update on this PR? It would be nice if we could include it in 3.7

hazelcast/src/test/java/com/hazelcast/map/SizeEstimatorTest.java
ahmetmircik: it is here https://github.com/hazelcast/hazelcast/commit/b90f8dbd9ebac29a02ecfe6c6f44bc1a1e62269d#diff-cddfa764e142a56c228ed23a9802d1a7R35

hazelcast/src/main/java/com/hazelcast/map/impl/record/AbstractRecord.java
pveentjer: Can you put the 4 into e.g Bits? So that we have a single place where this assumption an address is 4 bytes is stored? If we fix it later; we don't need to go over the whole code base.

hazelcast/src/main/java/com/hazelcast/map/impl/record/DataRecordWithStats.java
```diff
@@ -35,11 +35,8 @@
      */
     @Override
     public long getCost() {
-        long cost = super.getCost();
         final int objectReferenceInBytes = 4;
```
pveentjer: constant in bits

pveentjer:
Main comment is about centralizing the reference size so we can fix it once.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
```diff
@@ -93,8 +93,7 @@ public int getType() {
     @Override
     public int getHeapCost() {
         // reference (assuming compressed oops)
-        int objectRef = INT_SIZE_IN_BYTES;
-        return objectRef + (payload != null ? ARRAY_HEADER_SIZE_IN_BYTES + payload.length : 0);
+        return REFERENCE_COST_IN_BYTES + (payload != null ? ARRAY_HEADER_SIZE_IN_BYTES + payload.length : 0);
```
serkan-ozal: More reliable way for `ARRAY_HEADER_SIZE_IN_BYTES` might be retrieved by `arrayBaseOffset`.
Also, the total size might be rounded to `ObjectAlignmentInBytes` (can assume `8` at the moment)

ahmetmircik: Currently, `ARRAY_HEADER_SIZE_IN_BYTES ` calculation is out of scope of this PR. But in a future PR, i am planning to address it.

serkan-ozal:
:+1:

PR #7966, opened 2016-04-18
PagingPredicate anchor update is incorrect

ihsandemir:
There may be cases when the server may return a list of entries larger than the requested page size, in this case the client should not put any anchor into the list that is on a page greater than the requested page.

E.g.
pageSize:5
page:2
Map entries are: (0,0), (1, 1), ... (50, 50)
server returns entries: 5, 6,7,8,9,10,13, 15, 16, 19

Current implementation keeps anchor (19,19) for page 2. This is incorrect.
 
Fixes is just to check for page number when setting the anchor.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapIssueTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapIssueTest.java
@@ -232,8 +232,82 @@ public class ClientMapIssueTest extends HazelcastTestSupport {
         final PagingPredicate predicate = new PagingPredicate(pageSize);
         predicate.nextPage();
 
-        final Collection<Integer> values = map.values(predicate);
+        Collection<Integer> values = map.values(predicate);
         assertEquals(pageSize, values.size());
+
+        /**
+         * There may be cases when the server may return a list of entries larger than the requested page size , in this case
+         * the client should not put any anchor into the list that is on a page greater than the requested page. The case occurs
+         * when multiple members exist in the cluster. E.g.:
+         * pageSize:5
+         * page:2
+         * Map entries are: (0,0), (1, 1), ... (50, 50)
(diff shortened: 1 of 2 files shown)


gurbuzali:
:+1:

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapIssueTest.java
```diff
@@ -232,8 +232,64 @@ public void testMapPagingValues() {
         final PagingPredicate predicate = new PagingPredicate(pageSize);
         predicate.nextPage();
 
-        final Collection<Integer> values = map.values(predicate);
+        Collection<Integer> values = map.values(predicate);
         assertEquals(pageSize, values.size());
+
```
ahmetmircik: this test seems hard to understand to me. Is it possible to add some small explanatory comments on each code block like `this is done for this` or `now we are doing this` ?

ihsandemir: @ahmetmircik I added explanation. Can you check if it is OK now?

sancar:
@ihsandemir backport to maintenance ?

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapIssueTest.java
ahmetmircik: thanks

ihsandemir:
@sancar Backport: https://github.com/hazelcast/hazelcast/pull/7976

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapIssueTest.java
ihsandemir: @ahmetmircik Added more description to the test describing the test.

ihsandemir:
verify

PR #7969, opened 2016-04-18
[DONT MERGE] Invocation retry coordinated from InvocationMonitor

pveentjer:
To reduce threading complexity and resolve the duplicate registration problem, retrying an invocation is done from the InvocationMonitor instead of an arbitrary thread.


pveentjer:
run-lab-run

PR #7975, opened 2016-04-18
SerializationService toObject for InstanceOfPredicate causes NullPointerException

ihsandemir:
When using InstanceOfPredicate, the server side toObject method causes NullPointerException, since the classloader is not set and it is null. This fix uses the default classloader if it is not set by the config to avoid the null pointer exception.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapTest.java
@@ -42,7 +42,9 @@ import com.hazelcast.nio.serialization.Portable;
 import com.hazelcast.nio.serialization.PortableFactory;
 import com.hazelcast.nio.serialization.TestSerializationConstants;
 import com.hazelcast.query.Predicate;
+import com.hazelcast.query.SampleObjects;
 import com.hazelcast.query.SqlPredicate;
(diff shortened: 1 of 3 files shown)


sancar:
@ihsandemir Can you test the same thing also for node ? I took a quick look at the code, and it looks like same test will fail for server too.

hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
```diff
@@ -64,6 +64,9 @@ public InternalSerializationService createSerializationService(byte version) {
         try {
             ClientConfig config = client.getClientConfig();
             ClassLoader configClassLoader = config.getClassLoader();
+            if (null == configClassLoader) {
```
sancar: Using our classLoaderUtil for this seems like correct solution
`klass = ClassLoaderUtil.loadClass(in.getClassLoader(), klassName);`

ihsandemir: @sancar Ok, changed as suggested. Please review again.

ihsandemir:
verify

ihsandemir:
@sancar Added the test.

sancar:
👍

PR #7979, opened 2016-04-19
Minor date to string cleanup in ClusterheartbeatManager

pveentjer:
Now makes use of the StringUtils.timeToString method which gives compact format with ms precision


vbekiaris:
:thumbsup:

Issue #7968, opened 2016-04-18
Configurable path performance log


pveentjer:
The PerformanceLog has a fixed path (user working directory). But this should be configurable.

jerrinot:
fixed by #7981

PR #7981, opened 2016-04-19
PerformanceLog directory configurable

pveentjer:
Makes the directory for outputting the performance log configurable.

Solves #7968


mtopolnik:
:+1:

PR #7983, opened 2016-04-19
Get mapContainer from RecordStore

ahmetmircik:
Reduces contention among partition threads for mapContainer


eminn:
👍

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MapOperation.java
```diff
@@ -126,16 +130,16 @@ protected void evict() {
         recordStore.evictEntries();
     }
 
-    private void getOrCreateRecordStore() {
+    private RecordStore getOrNullRecordStore() {
```
jerrinot: `getRecordStoreOrNull` ?

ahmetmircik: done.

jerrinot:
:+1:

PR #7985, opened 2016-04-19
Constant time map stats

### LINKED ISSUE #4022: IMap stats showing negative values
I'm using hazelcast 2.6.7 and after heavy usage of the cache, the number of misses starts showing up negative. Specifically, it is being reported as the negative of the number of hits (e.g. if hits = 20, misses = -20). I'm getting the map stats as follows:

    LocalMapStats mapStats = HazelcastCache.instance.getMap("myMap").getLocalMapStats();
    long hits = mapStats.getHits();
    long misses = mapStats.getMisses();

muratayan:
`LocalMapStats` logic doesn't change. Complexity improved from per entry O(N) to per partition O(p) = O(1)

fixes #4022

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractRecordStore.java
@@ -65,6 +65,10 @@ abstract class AbstractRecordStore implements RecordStore<Record> {
 
     protected Storage<Data, Record> storage;
 
+    private long hits;
+    private long lastAccess;
+    private long lastUpdate;
+
     protected AbstractRecordStore(MapContainer mapContainer, int partitionId) {
         this.mapContainer = mapContainer;
         this.partitionId = partitionId;
@@ -96,6 +100,7 @@ abstract class AbstractRecordStore implements RecordStore<Record> {
 
         final long maxIdleMillis = calculateMaxIdleMillis(mapConfig);
         setExpirationTime(record, maxIdleMillis);
+        updateStatsOnPut(true, now);
         return record;
     }
 
@@ -124,6 +129,7 @@ abstract class AbstractRecordStore implements RecordStore<Record> {
     }
 
     protected void updateRecord(Data key, Record record, Object value, long now) {
+        updateStatsOnPut(false, now);
         record.onUpdate(now);
         storage.updateRecordValue(key, record, value);
     }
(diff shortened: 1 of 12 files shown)


emrahkocaman:
👍

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractRecordStore.java
```diff
@@ -208,4 +218,48 @@ public void disposeDeferredBlocks() {
...
+        setLastUpdateTime(now);
+
+        if (!newRecord) {
+            updateStatsOnGet(now);
+        }
+    }
+
+    protected void updateStatsOnGet(long now) {
+        setLastAccessTime(now);
+        increaseHits();
```
ahmetmircik: we also need to decrease hits in the event of entry removal.

emrahkocaman: I'm not sure about that. We're trying to reach the total hit count of a map, right? Imagine that a map is cleared, then this means the hit count is 0 now? This should not be the case, IMO.

muratayan:
Updated

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractRecordStore.java
enesakar: agree with @emrahkocaman

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractRecordStore.java
```diff
@@ -208,4 +218,72 @@ public void disposeDeferredBlocks() {
...
+    public void setLastUpdateTime(long time) {
+        this.lastUpdate = Math.max(this.lastUpdate, time);
+    }
+
+
+    protected void updateStatsOnPut(boolean newRecord, long now) {
+        setLastUpdateTime(now);
+
+        if (!newRecord) {
+            updateStatsOnGet(now);
```
enesakar: Yes we can update the access time but I think we should not increase hits on update operation. wdyt? @ahmetmircik @emrahkocaman

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractRecordStore.java
ahmetmircik: To me, if this PR is trying to preserve existing behaviour (it may be correct or not, discussion for  that is out of scope) and is only providing O(1) calculation for stats, `decrease` should be there because in current impl. after `map#clear`, there is no hit.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/AbstractRecordStore.java
ahmetmircik: makes sense

muratayan: Makes sense but won't do it because we are trying to preserve existing behaviour.

ahmetmircik:
👍

PR #7986, opened 2016-04-19
wip

ahmetmircik:


ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
```diff
@@ -255,7 +289,16 @@ protected void getAllObjectInternal(List<Data> keys, List<Object> resultingKeyVa
...
             if (!isOwn(key) || cacheLocalEntries) {
+                long previousCount = counters.get(key);
+                if (invalidationCounter.isStale(key, previousCount)) {
+                    continue;
+                }
+
                 nearCache.put(key, value);
+
+                if (invalidationCounter.isStale(key, previousCount)) {
+                    invalidateCache(key);
```
jerrinot: isn't it enough to just remove the the entry, but not increase the counter?

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
```diff
@@ -255,7 +289,16 @@ protected void getAllObjectInternal(List<Data> keys, List<Object> resultingKeyVa
...
 
             if (!isOwn(key) || cacheLocalEntries) {
+                long previousCount = counters.get(key);
+                if (invalidationCounter.isStale(key, previousCount)) {
+                    continue;
+                }
+
                 nearCache.put(key, value);
+
+                if (invalidationCounter.isStale(key, previousCount)) {
```
jerrinot: This part is tricky and it's worth to think about the best possible sequence of actions. 

An alternative option is to 
- check if your key is stale before calling `map.put()`. 
- if the key is stale then will not touch the near cache at all. 
- if the key is not stale then you do the put and then check for staleness again. 

I believe it could perform better under load - you eliminate unnecessary check and if 1 thread is stale then the other will still have a chance to populate the near cache . The price is the extra check - it will be a tiny bit slower in happy situation. WDYT?

PR #7988, opened 2016-04-20
Add slotHash() to HashSlotArray to enable hashcode caching

mtopolnik:
There is a use case of HashSlotArray where the key stored in the slot is only a pointer to the actual key blob. In such a case it is important for performance to cache the hashcode of the blob. This PR adds a method which can be overridden such that this cached value is used.


toktarev:
👍

PR #7998, opened 2016-04-22
Workaround for unexpected Selector.select behavior

### LINKED ISSUE #7943: Too much cpu used when Hazelcast is idle V 3.6.2
As requested by @noctarius creating new ticket.
Am using 3.6.2 new release still taking high CUP usage.
Verified in java 1.7.0_75 and 1.8.0_77.
Tomcat 7.0.68 (java 1.7),Tomcat 8.0.30 (java 1.8).
OS Windows 7.
Use case - We are using apache shiro as security framework and hazelcast is beeing used to store user session in hazelcast map. We are observing high cpu when no user has logged in for around 3 hours.

[hazelcast.txt](https://github.com/hazelcast/hazelcast/files/215087/hazelcast.txt)

untitled1
untitled

vbekiaris:
This PR introduces a workaround for unexpected `Selector.select` behavior (fixes #7943).
Under unknown circumstances, `Selector.select(timeout)` may enter a state in which it unblocks immediately, without any of the conditions described in its javadoc having been in effect (no keys selected, not woken up, not interrupted and timeout did not pass). Detection is introduced in `NonBlockingIOThread` and `SocketAcceptorThread`, the workaround being to build a new `Selector`.

A new system property `hazelcast.io.selectorMode` is introduced. Possible values are `selectnow` (use `Selector.selectNow` experimental mode) and `selectwithfix` (use default select loop with selector rebuild on bug detection); any other value will default to using the standard select(timeout) loop.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
@@ -21,7 +21,7 @@ import com.hazelcast.internal.metrics.Probe;
 import com.hazelcast.internal.util.counters.SwCounter;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.IOService;
-import com.hazelcast.nio.IOUtil;
+import com.hazelcast.nio.tcp.nonblocking.SelectorMode;
 
 import java.io.IOException;
 import java.nio.channels.ClosedChannelException;
@@ -32,12 +32,15 @@ import java.nio.channels.SocketChannel;
 import java.util.Iterator;
 
 import static com.hazelcast.internal.util.counters.SwCounter.newSwCounter;
+import static com.hazelcast.nio.IOUtil.closeResource;
 import static java.lang.Math.max;
 import static java.lang.System.currentTimeMillis;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class SocketAcceptorThread extends Thread {
     private static final long SHUTDOWN_TIMEOUT_MILLIS = SECONDS.toMillis(10);
+    private static final long SELECT_TIMEOUT_MILLIS = SECONDS.toMillis(60);
+    private static final int SELECT_IDLE_COUNT_THRESHOLD = 10;
 
     private final ServerSocketChannel serverSocketChannel;
     private final TcpIpConnectionManager connectionManager;
@@ -47,7 +50,19 @@ public class SocketAcceptorThread extends Thread {
     private final SwCounter eventCount = newSwCounter();
     @Probe
     private final SwCounter exceptionCount = newSwCounter();
+    // count number of times the selector was recreated (if selectWorkaround is enabled)
+    @Probe
+    private final SwCounter selectorRecreateCount = newSwCounter();
+    // last time select returned
     private volatile long lastSelectTimeMs;
+
+    // When true, enables workaround for bug occuring when SelectorImpl.select returns immediately
+    // with no channels selected, resulting in 100% CPU usage while doing no progress.
+    // See issue: https://github.com/hazelcast/hazelcast/issues/7943
+    private final boolean selectorWorkaround = (SelectorMode.getConfiguredValue() == SelectorMode.SELECT_WITH_FIX);
+
+    private Selector selector;
+    private SelectionKey selectionKey;
 
     public SocketAcceptorThread(
             ThreadGroup threadGroup,
@@ -77,26 +92,28 @@ public class SocketAcceptorThread extends Thread {
             logger.finest("Starting SocketAcceptor on " + serverSocketChannel);
         }
 
-        Selector selector = null;
         try {
             selector = Selector.open();
             serverSocketChannel.configureBlocking(false);
-            serverSocketChannel.register(selector, SelectionKey.OP_ACCEPT);
-            acceptLoop(selector);
+            selectionKey = serverSocketChannel.register(selector, SelectionKey.OP_ACCEPT);
+            if (selectorWorkaround) {
+                acceptLoopWithSelectorFix();
+            } else {
+                acceptLoop();
+            }
         } catch (OutOfMemoryError e) {
             OutOfMemoryErrorDispatcher.onOutOfMemory(e);
         } catch (IOException e) {
             logger.severe(e.getClass().getName() + ": " + e.getMessage(), e);
         } finally {
-            closeSelector(selector);
+            closeSelector();
         }
     }
 
-    private void acceptLoop(Selector selector) throws IOException {
+    private void acceptLoop() throws IOException {
         while (connectionManager.isLive()) {
             // block until new connection or interruption.
             int keyCount = selector.select();
-            lastSelectTimeMs = System.currentTimeMillis();
             if (isInterrupted()) {
                 break;
             }
@@ -104,19 +121,60 @@ public class SocketAcceptorThread extends Thread {
                 continue;
             }
             Iterator<SelectionKey> it = selector.selectedKeys().iterator();
-            while (it.hasNext()) {
-                SelectionKey sk = it.next();
-                it.remove();
-                // of course it is acceptable!
-                if (sk.isValid() && sk.isAcceptable()) {
-                    eventCount.inc();
-                    acceptSocket();
+            handleSelectionKeys(it);
+        }
+    }
+
+    private void acceptLoopWithSelectorFix() throws IOException {
+        int idleCount = 0;
+        while (connectionManager.isLive()) {
+            // block with a timeout until new connection or interruption.
+            long before = currentTimeMillis();
+            int keyCount = selector.select(SELECT_TIMEOUT_MILLIS);
+            if (isInterrupted()) {
+                break;
+            }
+            if (keyCount == 0) {
+                long selectTimeTaken = currentTimeMillis() - before;
+                idleCount = selectTimeTaken < SELECT_TIMEOUT_MILLIS ? idleCount + 1 : 0;
+                // select unblocked before timing out with no keys selected --> bug detected
+                if (idleCount > SELECT_IDLE_COUNT_THRESHOLD) {
+                    // rebuild the selector
+                    rebuildSelector();
+                    idleCount = 0;
                 }
+                continue;
+            }
+            idleCount = 0;
+            Iterator<SelectionKey> it = selector.selectedKeys().iterator();
+            handleSelectionKeys(it);
+        }
+    }
+
+    private void rebuildSelector() throws IOException {
+        selectorRecreateCount.inc();
+        // cancel existing selection key, register new one on the new selector
+        selectionKey.cancel();
+        closeSelector();
+        Selector newSelector = Selector.open();
+        selector = newSelector;
+        selectionKey = serverSocketChannel.register(newSelector, SelectionKey.OP_ACCEPT);
+    }
+
+    private void handleSelectionKeys(Iterator<SelectionKey> it) {
+        lastSelectTimeMs = currentTimeMillis();
+        while (it.hasNext()) {
+            SelectionKey sk = it.next();
+            it.remove();
+            // of course it is acceptable!
+            if (sk.isValid() && sk.isAcceptable()) {
+                eventCount.inc();
+                acceptSocket();
             }
         }
     }
 
-    private void closeSelector(Selector selector) {
+    private void closeSelector() {
         if (selector == null) {
             return;
         }
@@ -188,7 +246,7 @@ public class SocketAcceptorThread extends Thread {
         } catch (Exception e) {
             exceptionCount.inc();
             logger.warning(e.getClass().getName() + ": " + e.getMessage(), e);
-            IOUtil.closeResource(socketChannel);
+            closeResource(socketChannel);
         }
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
@@ -27,7 +27,6 @@ import com.hazelcast.nio.tcp.TcpIpConnectionManager;
 
 import java.io.IOException;
 import java.nio.channels.SelectionKey;
-import java.nio.channels.Selector;
 import java.util.logging.Level;
 
 import static com.hazelcast.internal.metrics.ProbeLevel.DEBUG;
@@ -39,7 +38,6 @@ public abstract class AbstractHandler implements MigratableHandler {
     protected final TcpIpConnection connection;
     protected final TcpIpConnectionManager connectionManager;
     protected final IOService ioService;
-    protected Selector selector;
     protected NonBlockingIOThread ioThread;
     protected SelectionKey selectionKey;
     private final int initialOps;
@@ -56,7 +54,6 @@ public abstract class AbstractHandler implements MigratableHandler {
         this.connection = connection;
         this.ioThread = ioThread;
         this.ioThreadId = ioThread.id;
-        this.selector = ioThread.getSelector();
         this.socketChannel = connection.getSocketChannelWrapper();
         this.connectionManager = connection.getConnectionManager();
         this.ioService = connectionManager.getIoService();
@@ -83,9 +80,13 @@ public abstract class AbstractHandler implements MigratableHandler {
 
     protected SelectionKey getSelectionKey() throws IOException {
         if (selectionKey == null) {
-            selectionKey = socketChannel.register(selector, initialOps, this);
+            selectionKey = socketChannel.register(ioThread.getSelector(), initialOps, this);
         }
         return selectionKey;
+    }
+
+    final void setSelectionKey(SelectionKey selectionKey) {
+        this.selectionKey = selectionKey;
     }
 
     final void registerOp(int operation) throws IOException {
@@ -142,7 +143,6 @@ public abstract class AbstractHandler implements MigratableHandler {
         ioThreadId = ioThread.id;
         selectionKey.cancel();
         selectionKey = null;
-        selector = null;
 
         newOwner.addTaskAndWakeup(new Runnable() {
             @Override
@@ -167,7 +167,6 @@ public abstract class AbstractHandler implements MigratableHandler {
             return;
         }
 
-        selector = newOwner.getSelector();
         selectionKey = getSelectionKey();
         registerOp(initialOps);
     }
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
@@ -22,13 +22,17 @@ import com.hazelcast.internal.metrics.ProbeLevel;
 import com.hazelcast.internal.util.counters.SwCounter;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.spi.impl.operationexecutor.OperationHostileThread;
+import com.hazelcast.util.EmptyStatement;
 
 import java.io.IOException;
 import java.nio.channels.CancelledKeyException;
+import java.nio.channels.ClosedChannelException;
+import java.nio.channels.SelectableChannel;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.util.Iterator;
 import java.util.Queue;
+import java.util.Random;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 import static com.hazelcast.internal.util.counters.SwCounter.newSwCounter;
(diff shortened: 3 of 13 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -56,43 +59,52 @@
...
     @Probe
     private final SwCounter completedTaskCount = newSwCounter();
+    // count number of times the selector was rebuilt (if selectWorkaround is enabled)
+    @Probe
+    private final SwCounter selectorRebuildCount = newSwCounter();
 
     private final ILogger logger;
 
-    private final Selector selector;
+    private volatile Selector selector;
```
pveentjer: Does it need to be volatile? Afaik it isn't needed.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,6 +346,29 @@ public final void shutdown() {
         interrupt();
     }
 
+    // this method is always invoked in this thread
+    // after we have blocked for selector.select in #runSelectLoopWithSelectorFix
+    private void rebuildSelector() {
+        selectorRebuildCount.inc();
+        Selector newSelector = NonBlockingIOThread.newSelector(logger);
```
pveentjer: Can't the 'newSelector method be used?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,6 +346,29 @@ public final void shutdown() {
...
+        Selector selector = this.selector;
```
pveentjer: I would call this one 'oldSelector'.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketReader.java
```diff
@@ -245,6 +245,20 @@ public String toString() {
         return connection + ".socketReader";
     }
 
+    @Override
```
pveentjer: Do we need this method at all? Why can't the AbstractHandler take care of this instead of adding more methods to the subclasses?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
vbekiaris: `NonBlockingIOThread.getSelector()` is invoked from `AbstractHandler.getSelectionKey` and `AbstractHandler.onFailure`. Apart from the former's usage in `NonBlockingSocketReader.init`, I think all other cases in which the `selector` is obtained come from the IO thread itself, so assuming the case of `init` invocation is irrelevant, `volatile` could be removed. Can you confirm?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketReader.java
vbekiaris: Something like this for `AbstractHandler.reschedule`?
```java
    private void reschedule() {
        if (this instanceof Runnable) {
            ioThread.addTaskAndWakeup((Runnable) this);
        }
        else {
            ioThread.addTaskAndWakeup(new Runnable() {
                @Override
                public void run() {
                    try {
                        handle();
                    } catch (Exception e) {
                        onFailure(e);
                    }
                }
            });
        }
    }
```

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
```diff
@@ -39,9 +42,9 @@
     protected final TcpIpConnection connection;
     protected final TcpIpConnectionManager connectionManager;
     protected final IOService ioService;
-    protected Selector selector;
     protected NonBlockingIOThread ioThread;
-    protected SelectionKey selectionKey;
+    // is a Handler shared among threads?
+    protected volatile SelectionKey selectionKey;
```
pveentjer: It will not be touched by multiple threads. So volatile isn't needed.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
```diff
@@ -56,14 +59,15 @@ public AbstractHandler(TcpIpConnection connection, NonBlockingIOThread ioThread,
...
         this.ioThreadId = ioThread.id;
-        this.selector = ioThread.getSelector();
         this.socketChannel = connection.getSocketChannelWrapper();
         this.connectionManager = connection.getConnectionManager();
         this.ioService = connectionManager.getIoService();
         this.logger = ioService.getLogger(this.getClass().getName());
         this.initialOps = initialOps;
     }
 
+    protected abstract void reschedule();
```
pveentjer: Probably this method doesn't need to be implemented at every subclass. We should be able to implement it directly.

I need to think about the implementation, because it could bite with the migration.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
pveentjer: Confirmed.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
```diff
@@ -98,11 +102,33 @@ final void unregisterOp(int operation) throws IOException {
...
     public void onFailure(Throwable e) {
         if (e instanceof OutOfMemoryError) {
             ioService.onOutOfMemory((OutOfMemoryError) e);
         }
+
+        // if our selection key was cancelled due to replacement of the selector
+        // (see NonBlockingIOThread#selectWorkaround), then reschedule this handler for execution
+        if (e instanceof CancelledKeyException && selectionKey != null
+                && selectionKey.selector() != ioThread.getSelector()) {
+            // the selector was replaced, so obtain a new key for the new selector
```
vbekiaris: @pveentjer If a migration is in progress, perhaps we risk overwriting the new `selectionKey` created in the new owner thread executing `completeMigration`. Should this condition be amended as follows to avoid interfering with a migration in progress? In this scenario, `this.ioThread` has already been substituted by `startMigration` with the new owner thread, while `onFailure` would still be executing a task on the previous owner.
```java
if (e instanceof CancelledKeyException && selectionKey != null
                && ioThread == Thread.currentThread()
                && selectionKey.selector() != ioThread.getSelector()) {
```

hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
```diff
@@ -47,8 +49,20 @@
     private final SwCounter eventCount = newSwCounter();
     @Probe
     private final SwCounter exceptionCount = newSwCounter();
+    // count number of times the selector was recreated (if selectWorkaround is enabled)
+    @Probe
+    private final SwCounter selectorRecreateCount = newSwCounter();
+    // last time select returned
     private volatile long lastSelectTimeMs;
 
+    // When true, enables workaround for bug occuring when SelectorImpl.select returns immediately
```
pveentjer: I would pull this logic into a method e..g 

selectorRebuildEnabled and add a log statement when it is true.

hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
```diff
@@ -77,46 +91,89 @@ public void run() {
...
-                if (sk.isValid() && sk.isAcceptable()) {
-                    eventCount.inc();
-                    acceptSocket();
+            handleSelectionKeys(it);
+        }
+    }
+
+    private void acceptLoopWithSelectorFix() throws IOException {
+        int idleCount = 0;
+        while (connectionManager.isLive()) {
```
pveentjer: Why are you using connectionManager.isAlive instead of isInterrupted?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,6 +346,29 @@ public final void shutdown() {
...
+
+        // cancel the old keys
+        for (SelectionKey key: oldSelector.keys()) {
+            key.cancel();
+        }
+
+        try {
+            // close the old selector
+            oldSelector.close();
+
```
pveentjer: Why this white line?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,6 +346,29 @@ public final void shutdown() {
...
+        } catch (Throwable t) {
+            logger.warning("Failed to close selector.", t);
+        }
+        logger.fine("Recreated Selector because of possible java/network stack bug.");
```
pveentjer: Would it make sense to put this on warning because it is a pretty severe case?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketReader.java
pveentjer: we can get rid of the reschedule completely. When you create the new selectionkey with the same interest set as the old key, then on the next selector.select you will be immediately notified since the underlying conditions to trigger the key in the first place, are still there. If data was available for reading, it will still be available for reading. If space is available for writing, it still is available.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
pveentjer: If a migration is in progress, you can't get any events since your selection key is closed.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
```diff
@@ -98,11 +101,33 @@ final void unregisterOp(int operation) throws IOException {
...
+            SelectableChannel ch = selectionKey.channel();
+            int ops = selectionKey.interestOps();
+            Object att = selectionKey.attachment();
+
+            // register the channel with the new selector now
+            try {
+                SelectionKey sk = ch.register(ioThread.getSelector(), ops, att);
+                this.selectionKey = sk;
+            } catch (ClosedChannelException ignore) {
+                EmptyStatement.ignore(ignore);
```
pveentjer: Instead of returning, I would continue with the original flow of the exception. So closing the connection.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,6 +346,29 @@ public final void shutdown() {
...
 
+    // this method is always invoked in this thread
+    // after we have blocked for selector.select in #runSelectLoopWithSelectorFix
+    private void rebuildSelector() {
+        selectorRebuildCount.inc();
+        Selector newSelector = NonBlockingIOThread.newSelector(logger);
+        Selector oldSelector = this.selector;
+        this.selector = newSelector;
+
+        // cancel the old keys
```
pveentjer: This happens automatically when the selector is closed.

pveentjer: One thing just crossed my head. Probably you can rebuild the selection keys here as well because through the selectionkeys you can read all read/write handlers and then force a rebuild of the selection key. No need to rely on the onFailure.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/AbstractHandler.java
pveentjer: This method can be removed completely.

pveentjer:
@vbekiaris is there an update or still enjoying yourself with the deadlock?

vbekiaris:
Back from the deadlock, working on simulating bug detection to test selector rebuild code.

hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
vbekiaris: This is the same as plain acceptLoop above.

vbekiaris:
run-lab-run

hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
vbekiaris: This is now logged on acceptor thread startup: https://github.com/vbekiaris/hazelcast/blob/9e10fffc8d5a3b4c54ad506816d9cfb138df7de8/hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java#L96

vbekiaris:
run-lab-run

hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
```diff
@@ -77,46 +92,93 @@ public void run() {
             logger.finest("Starting SocketAcceptor on " + serverSocketChannel);
         }
 
-        Selector selector = null;
+        if (selectorWorkaround) {
```
pveentjer: I would do this coordinated at a single place. Otherwise you will get 3+3+1 of these log entries.

hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
```diff
@@ -77,46 +92,93 @@ public void run() {
...
+            // of course it is acceptable!
+            if (sk.isValid() && sk.isAcceptable()) {
+                eventCount.inc();
+                acceptSocket();
             }
         }
     }
 
-    private void closeSelector(Selector selector) {
+    private void closeSelector() {
```
pveentjer: Can be replaced by IOUtil.closeResource.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -171,10 +190,15 @@ public final void run() {
...
-                    } else {
-                        runSelectLoop();
+                    switch (selectorMode) {
+                        case CPU_WORKAROUND:
+                            runSelectLoopWithSelectorFix();
+                            break;
+                        case NOW:
+                            runSelectNowLoop();
+                            break;
+                        default:
```
pveentjer: I would not use 'default' for runSelect loop. Add a third 'case' and in the default an exception can be thrown.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -171,10 +190,15 @@ public final void run() {
         try {
             for (; ; ) {
                 try {
-                    if (selectNow) {
-                        runSelectNowLoop();
-                    } else {
-                        runSelectLoop();
+                    switch (selectorMode) {
```
pveentjer: I would rename this to select mode. Because it says something about which select approach is being used; not the selector.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -171,10 +190,15 @@ public final void run() {
...
+                        case CPU_WORKAROUND:
```
pveentjer: I would rename the 3 methods:
selectLoop
selectNowLoop
selectLoopWithFix

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,8 +360,42 @@ public final void shutdown() {
...
+                logger.info("Channel was closed while trying to register with new selector.");
+            } catch (CancelledKeyException e) {
+                // a CancelledKeyException may be thrown in key.interestOps
+                // in this case, since the key is already cancelled, just do nothing
+                EmptyStatement.ignore(e);
+            }
+            key.cancel();
+        }
+
+        IOUtil.closeResource(oldSelector);
```
pveentjer: I would use a static import here.

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThreadingModel.java
```diff
@@ -54,10 +53,16 @@
...
+    // experimental settings; default value is "default" and implies using Selector.select(timeout)
+    // Other possible values:
+    //   "now" -> use Selector.selectNow
+    //   "cpu-workaround" -> use Selector.select(timeout) with workaround for bug occurring when
+    //                       SelectorImpl.select returns immediately with no channels selected,
+    //                       resulting in 100% CPU usage while doing no progress.
+    // See issue: https://github.com/hazelcast/hazelcast/issues/7943
+    private SelectorMode selectorMode;
     private volatile IOBalancer ioBalancer;
+    private boolean selectorWorkaroundTest = Boolean.getBoolean("hazelcast.io.selector.workaround.test");
```
pveentjer: Can you add note we need to convert this to HazelcastProperty. But this requires some deeper changes since in this class we have no access to the HazelcastProperties. So something for 3.8

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/SelectorMode.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.hazelcast.nio.tcp.nonblocking;
+
+import com.hazelcast.util.StringUtil;
+
+/**
+ * Controls the mode in which IO and acceptor thread selectors will be operating
+ */
+public enum SelectorMode {
+    DEFAULT("default"),
```
pveentjer: I would rename this to:
select
selectnow
selectwithfix

hazelcast/src/main/java/com/hazelcast/nio/tcp/SocketAcceptorThread.java
vbekiaris: Logging the selector mode once in `NonBlockingIOThreadingModel.start` should be sufficient, will remove this log entry (https://github.com/vbekiaris/hazelcast/blob/master-nio-workaround/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThreadingModel.java#L125-L125).

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,8 +360,42 @@ public final void shutdown() {
         interrupt();
     }
 
+    // this method is always invoked in this thread
+    // after we have blocked for selector.select in #runSelectLoopWithSelectorFix
+    private void rebuildSelector() {
+        selectorRebuildCount.inc();
+        Selector newSelector = NonBlockingIOThread.newSelector(logger);
```
pveentjer: why do you need NonBLockingIOThread. ?

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingIOThread.java
```diff
@@ -307,8 +360,42 @@ public final void shutdown() {
...
+        Selector oldSelector = this.selector;
+        this.selector = newSelector;
+
+        // reset each handler's selectionKey, cancel the old keys
+        for (SelectionKey key : oldSelector.keys()) {
+            AbstractHandler att = (AbstractHandler) key.attachment();
```
pveentjer: I would call this handler or attachment. att was not immediately clear.

vbekiaris:
run-lab-run

vbekiaris:
run-lab-run

vbekiaris:
run-lab-run

pveentjer:
👍

mdogan:
👍 
good job.

vbekiaris:
run-lab-run

PR #7999, opened 2016-04-22
Unite and generalize MurmurHash functions

mtopolnik:
Current implementation of `MurmurHash3_x86_32_direct` and `MurmurHash3_x64_64_direct` is hardcoded to use the `MEM` accessor. This PR allows the memory accessor to be passed in by the caller.

Since MurmurHash requires little-endian access and a given `MemoryAccessor` could be either big- or little-endian, the `MemoryAccessor` interface was extended with a `boolean isBigEndian()` method.

By introducing the private `LoadStrategy` class this PR unites the on-heap and off-heap hashing functions and thus eliminates code duplication.

Additionally this PR removes the `EndianMemoryAccessor` interface, whose only method was promoted to `MemoryAccessor`. This interface found no usage in the codebase.

Finally, a small potential improvement is introduced into `EndiannessUtil`, where `+` was used to combine bytes into the target type (`int`, `long`, etc.). `|` is the appropriate combining operator.


hazelcast/src/main/java/com/hazelcast/util/HashUtil.java
```diff
@@ -170,23 +175,36 @@ public static int MurmurHash3_x86_32_direct(long address, int offset, int len, i
...
+    public static int MurmurHash3_x86_32_direct(long base, int offset, int len, int seed) {
+        return MurmurHash3_x86_32_direct(MEM, base, offset, len, seed);
+    }
+
+    public static int MurmurHash3_x86_32_direct(MemoryAccessor mem, long base, int offset, int len) {
+        return MurmurHash3_x86_32_direct(mem, base, offset, len, DEFAULT_MURMUR_SEED);
     }
 
+
     public static long MurmurHash3_x64_64(final byte[] data, int offset, int len, final int seed) {
```
toktarev: We have the same algo for 2 different cases
data[] and MemoryAccessor + we have a lot of abstractions how to link them
Can we may be have 1 Murmur calculator + ByteAccessStrategy ?

mtopolnik: I'm afraid of performance issues. Relying on abstractions means going through 2 or three more virtual method calls, so we start depending on the inlining depth etc. In fact I already had deleted some code and reused from EndiannessUtil, but was too afraid to commit it that way.

toktarev: We can run real benchmarks

100000000 Murmurs call with direct byte[] and with ByteAccessStrategy

mtopolnik: An isolated microbenchmark will not give the whole story. The inlining depth in that case will be enough, but in real loops it may well be different. I'm just afraid of performance regressions. But this is definitely something to seriously consider in the future.

toktarev: Strange.

If ByteAccessStrategy kills perfomance, and we rely on it, why do we have it ?
Why we re-used it for Bits ?

Bits is widely used like HashUtil

mtopolnik: The performance issue exists on non-Intel platforms, so it certainly won't show up in our lab. Also, the cost of eight load instructions might buffer the cost of a possible non-inlined function call.

toktarev:
👍

hazelcast/src/main/java/com/hazelcast/util/HashUtil.java
```diff
@@ -53,22 +54,20 @@
...
         int c2 = 0x1b873593;
 
         int h1 = seed;
-        // round down to 4 byte block
-        int roundedEnd = offset + (len & 0xfffffffc);
 
-        for (int i = offset; i < roundedEnd; i += 4) {
+        for (int i = offset; i < tailStart; i += MURMUR32_BLOCK_SIZE) {
             // little endian load order
             int k1 = (data[i] & 0xff)
```
serkan-ozal: we can read integer directly if the underlying platform is little endian and if `offset` is 4 bytes aligned

mtopolnik: I think this is the codepath that should be free of `Unsafe` dependencies (usable when `Unsafe` is not available).

hazelcast/src/main/java/com/hazelcast/util/HashUtil.java
```diff
@@ -111,28 +110,34 @@ public static int MurmurHash3_x86_32(byte[] data, int offset, int len, int seed)
...
-                    : (MEM.getByte(address + i) & 0xff)
-                    | ((MEM.getByte(address + i + 1) & 0xff) << 8)
-                    | ((MEM.getByte(address + i + 2) & 0xff) << 16)
-                    | (MEM.getByte(address + i + 3) << 24);
+        final boolean memIsLittleEndian = !mem.isBigEndian();
+
+        for (long blockAddr = base + offset; blockAddr < tailStart; blockAddr += MURMUR32_BLOCK_SIZE) {
+            // little-endian load order
+            int k1 = memIsLittleEndian
+                    ? mem.getInt(blockAddr)
```
serkan-ozal: If the given `MemoryAccessor` is not alignment aware and if the `offset` (so `blockAddr`) is not 4 bytes aligned, this might cause `SIGBUS` on platforms which don't support unaligned memory access. If we will give this responsibility to the given `MemoryAccessor`, this should be stated in the javadoc.

mtopolnik: Very good point... now I think the best way is to add a `supportsUnalignedAccess` method to `MemoryAccessor` so we don't attempt this when not allowed.

hazelcast/src/main/java/com/hazelcast/util/HashUtil.java
```diff
@@ -279,29 +292,38 @@ public static long MurmurHash3_x64_64(final byte[] data, int offset, int len, fi
...
-        for (int i = 0; i < len / 16; i++) {
-            k1 = MurmurHash3_getBlock_direct(address, (i * 2 * 8) + offset);
-            k2 = MurmurHash3_getBlock_direct(address, ((i * 2 + 1) * 8) + offset);
-
+        long k1;
+        long k2;
+
+        for (long blockAddr = base + offset; blockAddr < tailStart; blockAddr += MURMUR64_BLOCK_SIZE) {
+            if (memIsLittleEndian) {
+                k1 = mem.getLong(blockAddr);
```
serkan-ozal: same comment about unaligned memory access if the given `MemoryAccessor` is not alignment aware and `offset` is not 4 bytes aligned

hazelcast/src/main/java/com/hazelcast/util/HashUtil.java
serkan-ozal: Yes, in fact I was thinking about using `GlobalMemoryAccessor::int getInt(Object base, long offset)` which is a `HeapMemoryAccessor` if it is available, otherwise read from array as it is. But anyway, this is minor thing.

**P.S:** `byte[]` base offset must be added to index (`i`) in this case

mtopolnik: Merging the two implementations by relying on the `MemoryAccessor` abstraction should be the next thing to do here.

serkan-ozal:
👍

Issue #5473, opened 2015-06-09
Execution callback for PartitionClientRequest is registered too late


sertugkaya:
If an operation takes longer than `hazelcast.operation.call.timeout.millis`value, it waits until that operation to finish to throw timeout exception. This may be a general issue besides `PartitionClientRequest` so it needs to be investigated for other operation types too.
Current PartitionClientRequest.java :

````
 @Override
    public final void process() {
        beforeProcess();
        ClientEndpoint endpoint = getEndpoint();
        Operation op = prepareOperation();
        op.setCallerUuid(endpoint.getUuid());
        InvocationBuilder builder = operationService.createInvocationBuilder(getServiceName(), op, getPartition())
                .setReplicaIndex(getReplicaIndex())
                .setResultDeserialized(false);

        ICompletableFuture future = builder.invoke();
        future.andThen(this);
    }
```
If we change like this, it behaves correctly:
````
 @Override
    public final void process() {
        beforeProcess();
        ClientEndpoint endpoint = getEndpoint();
        Operation op = prepareOperation();
        op.setCallerUuid(endpoint.getUuid());
        InvocationBuilder builder = operationService.createInvocationBuilder(getServiceName(), op, getPartition())
                .setReplicaIndex(getReplicaIndex())
                .setExecutionCallback(this)
                .setResultDeserialized(false);

        builder.invoke();
    }
```

pveentjer:
@sertugkaya I know this issue has been posted a long time ago. But can you explain why you think your code changes fixes the problem? Both calls are equal and in 3.7 they are truly equal since the invocationbuilder forwards to 'andThen'. 

```
 @Override
    public InternalCompletableFuture invoke() {
        op.setServiceName(serviceName);

        Invocation invocation;
        if (target == null) {
            op.setPartitionId(partitionId).setReplicaIndex(replicaIndex);
            invocation = new PartitionInvocation(
                    operationService, op, tryCount, tryPauseMillis, callTimeout, resultDeserialized);
        } else {
            invocation = new TargetInvocation(
                    operationService, op, target, tryCount, tryPauseMillis, callTimeout, resultDeserialized);
        }

        InternalCompletableFuture future = invocation.invoke();
        if (executionCallback != null) {
            future.andThen(executionCallback);
        }

        return future;
    }
```

PR #8010, opened 2016-04-25
Performance: client response not async

pveentjer:
The client offloads the serverside result creation always to an async executor. This comes at a performance cost. This PR addresses that problem by only offloading to an async executor, if the completing thread of the invocation future is an OperationHostileThread (e.g. response thread). In other cases, e.g. a readonly operation executed by a partition thread, it will execute the completion handler on the partition thread so 1 handover to another executor is removed.

This will not improvement of updates because the backup ack will most likely be in later, so the backup ack will trigger the completion. Backup ack is executed by the response thread; which is a operation hostile thread and therefor sending the response will be offloaded to the async executor.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractPartitionMessageTask.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractPartitionMessageTask.java
@@ -18,17 +18,21 @@ package com.hazelcast.client.impl.protocol.task;
 
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.core.ExecutionCallback;
+import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.instance.Node;
 import com.hazelcast.nio.Connection;
-import com.hazelcast.spi.InvocationBuilder;
+import com.hazelcast.spi.ExecutionService;
 import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.impl.operationexecutor.impl.PartitionOperationThread;
+
+import java.util.concurrent.Executor;
 
 /**
  * AbstractPartitionMessageTask
  */
 public abstract class AbstractPartitionMessageTask<P>
         extends AbstractMessageTask<P>
-        implements ExecutionCallback {
+        implements ExecutionCallback, Executor {
 
     protected AbstractPartitionMessageTask(ClientMessage clientMessage, Node node, Connection connection) {
         super(clientMessage, node, connection);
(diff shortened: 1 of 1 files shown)


jerrinot:
it appears it effectively revert this: https://github.com/hazelcast/hazelcast/pull/5480

pveentjer:
@jerrinot it doesn't revert the idea behind that pr. That pr converts from one flavor of andThen configuration to another.

pveentjer:
On my standard atomiclong test, 4 member cluster, 40 threads, doing get, performance increased from 622k ops/second to 664k ops/second; that is 6.6% performance improvement. Latencies remain roughly the same.

pveentjer:
A future improvement will be to add some functionality to the invocation/future to transform the response on the partition thread to a client message, so that when the ack comes in, the client message can immediately send to the client without expensive transformations. Then it would not be a problem to let the response thread 'send' the client message since it will be cheap.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractPartitionMessageTask.java
```diff
@@ -57,17 +61,28 @@ public final void processMessage() {
...
 
-        builder.invoke();
+        f.andThen(this, this);
     }
 
     protected abstract Operation prepareOperation();
 
     @Override
+    public void execute(Runnable command) {
+        if (Thread.currentThread() instanceof OperationHostileThread) {
```
jerrinot: isn't it better to change the logic? 

if the current thread is `PartitionThread` then use it directly. otherwise offload to the ASYNC_EXECUTOR. It makes it more explicit when the current thread is used.

jerrinot:
I'm still concerned about the change of the callback style. it's not clear to me why it should matter, but I assume @sertugkaya had his reasons to fill #5473

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractPartitionMessageTask.java
pveentjer: I don't understand your comment. What do you mean 'use it directly'.

pveentjer:
There is no concern about the callback style. There is only one and that is 'future.andThen'. The invocation builder also forwards to this method; so either it is done explicitly or implicit; but in the end it is exactly the same call.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractPartitionMessageTask.java
jerrinot:
```java
if (Thread.currentThread().getClass() == PartitionOperationThread.class) {
  command.run();
else {
  offloadToExecutor();
}
```

jerrinot:
:+1:

PR #8011, opened 2016-04-25
Make StandardMemoryAccessor final

mtopolnik:
Goal: make `AMEM` methods easier to inline by JIT compiler. The motivation is observed performance degradation on some runs of Hot Restart.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/memory/impl/AlignmentAwareMemoryAccessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/memory/impl/AlignmentAwareMemoryAccessor.java
@@ -18,6 +18,10 @@ package com.hazelcast.internal.memory.impl;
 
 import com.hazelcast.internal.memory.MemoryAccessor;
 
+import java.lang.reflect.Field;
+
+import static com.hazelcast.internal.memory.GlobalMemoryAccessorRegistry.AMEM;
+import static com.hazelcast.internal.memory.impl.UnsafeUtil.UNSAFE;
 import static com.hazelcast.internal.memory.impl.UnsafeUtil.UNSAFE_AVAILABLE;
 
 /**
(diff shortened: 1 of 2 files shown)


pveentjer:
Strange that making a class final should make any difference.

serkan-ozal:
I am not sure about this change is the cause (and fix) of the observed performance degradation.

There might be another optimization points. 

For example, I think no need to use `AMEM` (`AlignmentAwareMemoryAccessor`) in `LongLinkedQueue` because in here memory accesses are already aligned. Even though possibly `AlignmentUtil.isXBytesAligned(address)` calls are able to be optimized by CPU branch prediction in this case, wtihout this check, the inlining might be handled easier by JVM. Thefore, using `MEM` (`StandardMemoryAccessor`) seems better choice to be used for such cases.

Possibly, there are another use-cases for redundant `AMEM` usage in our code base. So I think they should be refactored.

mtopolnik:
@pveentjer It helps in borderline cases. With a non-final class there always has to be a type guard instruction for a case which would violate the assumption of a monomorphic call site. With a final class, the site can be monomorphic even without a type profile, as long as the static type of the target is `StandardMemoryAccessor`.

@serkan-ozal It's the opposite: `AMEM` is the `StandardMemoryAccessor` and `MEM` is alignment-aware. The naming was chosen such that `MEM` is the "safe" option and `AMEM` is a special option where the developer guarantees aligned access. `AMEM` was supposed to mean "`MEM` for Aligned access".

serkan-ozal:
@mtopolnik Do you have any benchmark result shows that this change fixes (completely or partially) the observed performance degradation? Or have you verified that this change improves the performance?

mtopolnik:
@serkan-ozal It's not a straightforward situation. In some runs I observed a significant slowdown with the old code, in some runs it was the same as after the change. The code after the change never displayed the degradation. I think it is still inconclusive. The change of this PR is not big so I thought better stay on the safe side.

serkan-ozal:
@mtopolnik I am OK with the PR. I wanted to understand the situation.

PR #8012, opened 2016-04-25
Do not bypass regular operation infrastructure when readBackupData = true

### LINKED ISSUE #7354: Access data is not updated by local reads when reading from backup is enabled
Related to https://github.com/hazelcast/hazelcast/issues/7341

Steps to reproduce:

1. Create a map and enable reading from backups
2. Get local keys by using *localKeySet* method
3. Iterate over those keys and perform a get on each one
4. Take a look at entry view of each key
5. Observe that access timestamps are not updated

jerrinot:
Fixes #7354

We bypass a normal operation infrastructure when reading from a backup.
However we should not bypass it when the current member happens to be owner
and not just a backup of a given partition. As it has an impact on statistics,
it violates a thread model, etc.


muratayan:
I think this should be merged if we are not overlooking something 👍

pveentjer:
Looks good. But this could have an impact on performance.

pveentjer:
I'm missing something:

After the fix:
when you read from the primary, you make use of an partition thread
when you read from a backup, you bypass the partition thread?

So is this the behavior after the fix?

PR #8013, opened 2016-04-25
JCache Custom Eviction Policy Support

serkan-ozal:
This PR introduces `EvictionPolicyComparator` to our internal eviction API and it can be specified through `EvictionConfig`.
Then, on top of these improvements, JCache provides its own `CacheEvictionPolicyComparator` to support custom eviction policy comparators to be used for comparing cache entries to be evicted.


hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -77,6 +85,50 @@ public EvictionConfig(int size, MaxSizePolicy maxSizePolicy, EvictionPolicy evic
...
+         * Do not use setters, because they are overridden in the readonly version of this config and
+         * they cause an "UnsupportedOperationException". Just set directly if the value is valid.
+         */
+
+        // Size cannot be non-positive number
+        if (size > 0) {
+            this.size = size;
+        }
+        // Max-Size policy cannot be null
+        if (maxSizePolicy != null) {
```
ahmetmircik: If we do all checks in `EvictionConfig` class, like `if (maxSizePolicy != null)`, with `Preconditions`, users can be informed truly in case of wrong parameter passing.

serkan-ozal:
run-lab-run

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
serkan-ozal: :+1:

ahmetmircik:
👍

PR #8019, opened 2016-04-26
Replaced cast to ConnectionImpl, now casts to Interface.

kobalski:
Replaced cast to ConnectionImpl, now casts to Interface. This to avoid problems with JDK proxies on TomEE, where these cannot be cast to the implementation, only the interface.

Backport of : https://github.com/hazelcast/hazelcast-ra/pull/2


pveentjer:
:+1:

PR #8020, opened 2016-04-26
InvocationFuture: supress logging in case racy completion

pveentjer:
It can happen that both backup and response handling thread, decide that the invocation can be completed with exactly the same pending value. In this case; don't log a warning, but accept it as part of
the lock free design.

So only in case of a future being completed with a different value than it already completed with, a log warning is made.

This logging is not only irritating for end users since it scares them even though it is accepted behavior. It could also be a performance problem due to excess logging.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
@@ -357,11 +357,12 @@ public abstract class AbstractInvocationFuture<V> implements InternalCompletable
         for (; ; ) {
             final Object oldState = state;
             if (isDone(oldState)) {
-                // it can be that this invocation future already received an answer, e.g. when an invocation already received a
-                // response, but before it cleans up itself, it receives a HazelcastInstanceNotActiveException.
-                logger.warning("The Future.complete(Object value) can only be called once. Request: " + invocationToString()
-                        + ", current response: " + getState() + ", new response: " + value);
-
+                if (oldState != value) {
+                    // it can be that this future already completed, e.g. when an invocation already
+                    // received a response, but before it cleans up itself, it receives a HazelcastInstanceNotActiveException.
+                    logger.warning("Future.complete(Object value) can only be called once. Request: " + invocationToString()
+                            + ", current value: " + oldState + ", offered value: " + value);
+                }
                 return false;
             }
 


hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -357,11 +357,12 @@ public final boolean complete(Object value) {
...
-                // it can be that this invocation future already received an answer, e.g. when an invocation already received a
-                // response, but before it cleans up itself, it receives a HazelcastInstanceNotActiveException.
-                logger.warning("The Future.complete(Object value) can only be called once. Request: " + invocationToString()
-                        + ", current response: " + getState() + ", new response: " + value);
-
+                if (oldState != value) {
+                    // it can be that this future already completed, e.g. when an invocation already
+                    // received a response, but before it cleans up itself, it receives a HazelcastInstanceNotActiveException.
+                    logger.warning("Future.complete(Object value) can only be called once. Request: " + invocationToString()
+                            + ", current value: " + getState() + ", offered value: " + value);
```
Donnerbart: Isn't `oldState` sufficient here? This will safe another read from the `volatile` variable. This would also surely print the value you have compared it to before, just in case there is another concurrent modification going on.

Donnerbart: Isn't `oldState` sufficient here? This will safe another read from the `volatile` variable. This would also surely print the value you have compared it to before, just in case there is another concurrent modification going on.

pveentjer: Volatile reads are cheap and compared to the price of logging, it will hardly make any impact. But it is nicer from a coding point of view since it causes less confusion.

jerrinot: using `oldState` instead of `getState()` would make it slightly easier to reason about. as it's guaranteed not to change under your hands. the current implementation will also not change, but it requires more brain-power when reading it :)

also it will avoid volatile load.

pveentjer: The volatile load is really not worth to mention.

1: it is cheap
2: this branch will only happen in very very rare cases
3: the performance of this branch will be totally dominated by the logging; not by the volatile read.

But I fully agree that it is a readability improvement not to use getState, but use the oldState variable.

jerrinot:
:+1:

Donnerbart:
:+1: (once green)

PR #8022, opened 2016-04-26
Fire UPDATED event even when just the oldValue is matching a query filter

jerrinot:


hasancelik:
run-lab-run

noctarius:
I still think it is important to make this a separate event type as probably nobody tests old / new value again in the listener code. Sure you have both elements but why should I check which one matched my predicate? It might change current code behavior throwing the same type of event again for a new code path.

jerrinot:
I agree. more evaluation is needed here.

PR #8023, opened 2016-04-26
IMap.putAll() performance improvements

Donnerbart:
* Cleanup of `MapEntries`.
* Grouped `putAll()` calls per member.
* Added batch support.

Please see the TDD for details and benchmarks.


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllPerMemberOperation.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.Future;
+
+/**
+ * Inserts the {@link MapEntries} for all partitions of a member via locally invoked {@link PutAllOperation}.
+ * <p/>
+ * Used to reduce the number of remote invocations of an {@link com.hazelcast.core.IMap#putAll(Map)} call.
+ */
+public class PutAllPerMemberOperation extends MapOperation implements DataSerializable {
```
pveentjer: I would make this operation IdentifiedDataSerializable. All operations are automatically DataSerializable; so it isn't needed to repeat it.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllPerMemberOperation.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public void run() throws Exception {
+        NodeEngine nodeEngine = getNodeEngine();
+        Future[] futures = new Future[partitions.length];
+        for (int i = 0; i < partitions.length; i++) {
+            Operation op = new PutAllOperation(name, mapEntries[i]);
+            op.setNodeEngine(nodeEngine)
+                    .setPartitionId(partitions[i])
+                    .setReplicaIndex(getReplicaIndex())
+                    .setServiceName(getServiceName())
```
pveentjer: Map operations  have service name set. So no need to set it (call is ignored anyway).

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllPerMemberOperation.java
```diff
@@ -0,0 +1,100 @@
...
+                    .setService(getService())
+                    .setCallerUuid(getCallerUuid());
+            OperationAccessor.setCallerAddress(op, getCallerAddress());
+            futures[i] = nodeEngine.getOperationService().invokeOnPartition(op);
+        }
+        for (Future future : futures) {
```
pveentjer: Can you add line of comment that eventually this method needs to become non blocking. Currently the generic thread is blocked till all putAll completes.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllPerMemberOperation.java
```diff
@@ -0,0 +1,100 @@
...
+            future.get();
+        }
+    }
+
+    @Override
+    public void afterRun() throws Exception {
+    }
+
+    @Override
+    public Object getResponse() {
```
pveentjer: I would return null since you don't care about the value.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -107,21 +109,55 @@
...
+    /**
+     * Defines the batch size for operations of {@link IMap#putAll(Map)} calls.
+     *
+     * A value of {@code 0} disables the batching and will send a single operation per member with all map entries.
+     *
+     * If you set this value too high, you may ran into OOME or blocked network pipelines due to huge operations.
+     * If you set this value too low, you will lower the performance of the putAll() operation.
+     */
+    private static final HazelcastProperty MAP_PUT_ALL_BATCH_SIZE
+            = new HazelcastProperty("hazelcast.map.put.all.batch.size", 0);
```
pveentjer: would it not make sense to have some other default than 0? Now a user could run into a fat packet and he probably would not quickly figure out what is happening.

Donnerbart: This will most likely kill the performance (see the TDD).

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -107,21 +109,55 @@
...
+     * <ul>
+     * <li>{@code 10} for map sizes about 100 entries</li>
+     * <li>{@code 5} for map sizes between 500 and 5000 entries</li>
+     * <li>{@code 1.5} for map sizes between about 50000 entries</li>
+     * </ul>
+     *
+     * If you set this value too high, you will waste memory.
+     * If you set this value too low, you will suffer from expensive {@link java.util.Arrays#copyOf} calls.
+     */
+    private static final HazelcastProperty MAP_PUT_ALL_INITIAL_SIZE_FACTOR
```
pveentjer: what does 0 mean? How big is the initial array?  Im also confused abou the values. Is this a float or int?

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -107,21 +109,55 @@
...
+     * Defines the initial size of arrays per partition for {@link IMap#putAll(Map)} calls.
+     *
+     * The default value of {@code 0} uses an educated guess, depending on the map size, which is a good overall strategy.
+     * If you insert entries which don't match a normal partition distribution you should configure this factor.
+     * The initial size is calculated by this formula:
+     * {@code initialSize = ceil(MAP_PUT_ALL_INITIAL_SIZE_FACTOR * map.size() / PARTITION_COUNT)}
+     *
+     * As a rule of thumb you can try the following values:
+     * <ul>
+     * <li>{@code 10} for map sizes about 100 entries</li>
```
pveentjer: till?

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -107,21 +109,55 @@
...
+     * <li>{@code 5} for map sizes between 500 and 5000 entries</li>
+     * <li>{@code 1.5} for map sizes between about 50000 entries</li>
```
pveentjer: till?

pveentjer: And what about huge maps?

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -666,48 +708,77 @@ protected void getAllObjectInternal(List<Data> keys, List<Object> resultingKeyVa
...
-
         try {
-            List<Future> futures = new ArrayList<Future>(partitionCount);
-            MapEntries[] entriesPerPartition = new MapEntries[partitionCount];
+            int mapSize = map.size();
+            if (mapSize == 0) {
+                return;
+            }
+
+            boolean useBatching = getPutAllUseBatching(mapSize);
```
pveentjer: is instead of get makes more sense.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
pveentjer: These properties are a bit in a hidden location. Normally user modifiable properties are part of GroupProperties (even though the class is internal). These will be included in the documentation; however your properties are not. This could be good; exposing more than strictly needs to be could lead to problems.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -716,12 +787,69 @@ protected void putAllInternal(Map<?, ?> map) {
...
+    private int getPutAllInitialSize(boolean useBatching, int mapSize, int partitionCount) {
+        if (mapSize == 1) {
+            return 1;
+        }
+        if (useBatching) {
+            return putAllBatchSize;
+        }
+        if (putAllInitialSizeFactor < 1) {
+            // this is an educated guess for the initial size of the entries per partition, depending on the map size
+            return (int) ceil(20f * mapSize / partitionCount / log10(mapSize));
```
pveentjer: should 20f not be a constant?

Donnerbart: Since the value is so arbitrary I thought the formula would give it more context than a name and ripping it out of the formula.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
Donnerbart: I did this on purpose, as mentioned in the TDD. I want to be sure that those properties make sense before we add them "forever" to the canon (and have to provide legacy support for them). Maybe I'll add the `@Beta` annotation to them as well to underline the intent.

Donnerbart: > The default value of {@code 0} uses an educated guess...
Doesn't this explain what `0` means?

The value is float, I made this more clear by changing the "rule of thumb" examples to floats. The constructor uses an int, because we don't have a float constructor.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutAllPerMemberOperation.java
Donnerbart: Changed, but comment won't go away since method line was noted.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
Donnerbart: I just added the `@Beta` annotation to the properties. I also added more JavaDoc:
> {@link IMap#putAll(Map)} splits up the entries of the user input map per partition,
> to eventually send the entries the correct target nodes.
> So the method creates multiple arrays with map entries per partition.
> This value determines how the initial size of these arrays is calculated.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
Donnerbart: I elaborated the formula a bit more in the TDD. Maybe a small summary in the code won't hurt either, if the TDD is approved.

pveentjer:
:+1:

Donnerbart:
run-lab-run

ahmetmircik:
needs rebase

ahmetmircik:
👍

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -716,12 +796,72 @@ protected void putAllInternal(Map<?, ?> map) {
...
+        Future future = createPutAllOperationFuture(name, totalSize, partitions, entries, address);
+        futures.add(future);
+    }
+
+    protected Future createPutAllOperationFuture(final String name, final long size, int[] partitions, MapEntries[] entries,
+                                                 Address address) {
+        MapOperation op = operationProvider.createPutAllPerMemberOperation(name, partitions, entries);
         final long time = System.currentTimeMillis();
-        InternalCompletableFuture<Object> future = operationService.invokeOnPartition(SERVICE_NAME, op, partitionId);
+        InternalCompletableFuture<Object> future = operationService.invokeOnTarget(SERVICE_NAME, op, address);
```
ahmetmircik: Can it be more appropriate to use a logic similar to `invokeOnAllPartitions` here because it internally relies on `InvokeOnPartitions` and it has retry mechanism in case of failures ? Btw, I wrote a small test [here](https://gist.github.com/ahmetmircik/ad727c8ae6158a7a4be3e7fae933467e) and it fails on master and passes on maintenance.

Donnerbart: Will work on this issue in this PR: https://github.com/hazelcast/hazelcast/pull/8275

Issue #8024, opened 2016-04-26
Hazelcast is vulnerable to untrusted deserialization remote code execution


drosenbauer:
I emailed the support address on April 17 and received a response indicating that I should report my findings here. So here we are. 

The Hazelcast cluster join procedure is vulnerable to remote code execution due to Java deserialization. If an attacker can reach a listening Hazelcast instance with a crafted `JoinRequest`, and vulnerable classes are also on the classpath, the attacker can run arbitrary shell commands (among other nefarious things). Hazelcast will blindly deserialize any object it receives in that request stream. Since the `JoinRequest` is what implements authentication, this is necessarily pre-authentication.

This was verified against the latest code from the Git repository, as well as releases 3.7 and 3.2.1.

I have a small whitelist/blacklist filter patch that I can submit a PR for, once I'm cleared to do so. I've already emailed the signed form. Or if you have another solution, feel free! 

If you choose to go the filtering route, the Hazelcast team should probably create a default whitelist for serialization, because blacklists are almost always out of date as soon as they are created.

houey:
any progress on this issue?

kwart:
The issue is not resolved yet. Users using Enterprise edition can avoid the untrusted deserialization in member joining procedure by
* either enabling symmetric encryption;
* or configuring TLS with mutual authentication

If the member discovery mode is UDP multicast, then only the symmetric encryption avoids the abuse.

jerrinot:
Once more thanks for reporting the issue. #12230 is addressing it. It allows to define black/while lists for deserialization.

martinfurmanski:
Could we get this ported to 3.7 release branch?

anuragpathak2608:
@drosenbauer , do you have any data to pentest hazlecast publicly open port?

peterchenadded:
Can someone confirm @kwart comment?

kwart:
@peterchenadded are you asking for attack ready data? No, I don't provide such data.

peterchenadded:
@kwart No I mean does enabling SSL/encryption actually prevent the vulnerability? Can someone confirm?

Also the cve states that's it's vulnerable in hazelcast 3.10.x but wasn't the blacklist/whitelist feature to prevent the issue added in 3.10.0, just that you had to manually enable it.

kwart:
[...]

PR #8025, opened 2016-04-26
Adds a rudimentary serialization filter for issue #8024

drosenbauer:
This will prevent Hazelcast from deserializing objects on the blacklist or objects *not* on the whitelist.

#8024


jerrinot:
Hi @drosenbauer: thanks for you pull request. It's much appreciated. I'll have a closer look in upcoming days, but my first impression is very good.

drosenbauer:
The only change I can think of would be to allow people to use the regular Config objects to set up the blacklist and whitelists, rather than system properties. I'm not familiar enough with the way the internal objects are passed around to do that just yet, though, so if someone would like to do that, feel free!

jerrinot:
My other worry is a potential impact on performance. 

Although this has an impact on Java serialization only and you should not use the Java serialization if you want to have a decent performance anyway.

drosenbauer:
This will probably only down the first initial deserialization, when it populates the lists from the properties. Not much, but you'd see a difference if you did a bunch of "load the list" operations over and over. Once the list is loaded, HashSet.contains and String operations are super fast, so shouldn't slow down the Java deserialization significantly.

If you do go with this option, I would definitely encourage users to use a whitelist in your online documentation as well. Blacklists are better than nothing, but not great. There will always be more deserialization exploits found, and there are [lots of workarounds for blacklists](https://github.com/pwntester/SerialKillerBypassGadgetCollection/tree/master/src/main/java/serialkiller/bypass).

If there's a way to eliminate all Java deserialization until after authentication of the new cluster node (maybe make Credentials not a Java Serializable?), that would be the most ideal solution. That in combination with a whitelist for connected cluster nodes would be a powerful defense.

mdogan:
Needs rebase.

Related issue #8024 was put in _Backlog_.  Moving PR to _Backlog_ too.

Holmistr:
@jerrinot Is this PR still relevant?

PR #8027, opened 2016-04-27
Improved client TargetDisconnectException

pveentjer:
The exception message proves more context of the cause. Also the client connection has been modified to store the cause of the connection close so it can be included in the TargetDisconnectException.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/exception/TargetDisconnectedException.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/exception/TargetDisconnectedException.java
@@ -34,4 +34,8 @@ public class TargetDisconnectedException extends RetryableHazelcastException {
     public TargetDisconnectedException(String message) {
         super(message);
     }
+
+    public TargetDisconnectedException(String message, Throwable cause) {
+        super(message, cause);
+    }
 }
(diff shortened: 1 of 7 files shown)


Donnerbart:
:+1: (once green)

pveentjer:
It is an improvement, but I'm still not terribly happy. One of the causes of an closed connection,  is a real io error. I would like to get this error into the TargetDisconnectException.

pveentjer:
The flow in the client is a bit problematic. When there is a io error, it will claim there is no heartbeat. I'm randomly causing errors in the client connection and the heartbeat failure is even misleading. So this PR needs some additional TLC.

pveentjer:
@Donnerbart can you have another look

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/AbstractClientSelectionHandler.java
```diff
@@ -45,7 +45,7 @@
     private final NonBlockingIOThread ioThread;
     @Probe
     private final int ioThreadId;
-    private volatile SelectionKey sk;
+    protected volatile SelectionKey sk;
```
Donnerbart: Hmm, CheckStyle should not be happy with the field order (`protected` after `private`), but I cannot spot a suppression for this class. Never mind, it's not a blocker anyway.

hazelcast/src/main/java/com/hazelcast/spi/exception/TargetDisconnectedException.java
```diff
@@ -34,4 +34,8 @@ public TargetDisconnectedException(Address address) {
     public TargetDisconnectedException(String message) {
         super(message);
     }
+
+    public TargetDisconnectedException(String message,Throwable cause) {
```
Donnerbart: Missing space after comma (that was the cause for the failed build).

Donnerbart:
:+1: (once green)

Issue #7591, opened 2016-02-24
Revise and improve in-built converters


tombujok:
Currently the in-built converters are sub-optimal when it comes to performance. 
For example, if I want to convert a `Long` to using the `BigInteger` converter - it will first produce a string (litter) and then rely on the `BigInteger` parser to parse this String.
https://github.com/hazelcast/hazelcast/blob/eef987b9920f3f7b4c506e433b0d067cab9ba93c/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java#L163

tombujok:
Thanks @cpiotr, #8034 got merged!
Would you like to review other converters to see if there are any sub-optimal ones?

cpiotr:
@tombujok: sure. Do you have any issues registered around that?

tombujok:
No, just this github issue. I suspect there's similar sub-optimal stuff over there but never looked at all of them.

tombujok:
Hi @cpiotr Did you have any spare time to hack on this ;) ?

cpiotr:
Not yet. I'm currently on vacation. Will get down to it mid-June.

cpiotr:
Hi @tombujok, I took a quick look at the rest of converters. There is a room for improvements in case of `java.sql.{Timestamp, Date}`. Converters for both classes rely on `DateHelper `which uses `SimpleDateFormat `to construct first `java.util.Date` and then instance of target type. 
In both cases, `valueOf `methods exist, which do the same job in shorter time:

Benchmark | Mode | Cnt | Score | Error | Units
--- | --- | --- | --- | --- | --- 
DateHelperTest.benchmarkDateHelper | avgt | 10 | 1.483 | 0.345 | us/op
DateHelperTest.benchmarkDateValueOf | avgt | 10 | 0.214 | 0.033 | us/op
DateHelperTest.benchmarkTimestampHelper | avgt | 10 | 2.047 | 0.517 | us/op
DateHelperTest.benchmarkTimestampValueOf | avgt | 10 | 0.454 | 0.093 | us/op

mmedenjak:
Since we're heavily invested in the new query engine, I don't think it makes sense to put too much time into this.

PR #8034, opened 2016-04-27
BigDecimal and BigInteger converters reviewed and improved #7591

cpiotr:
BigDecimal and BigInteger converters reviewed and improved. Test coverage supplemented.
Issue: #7591 
Work done during Hackergarten in Gdansk by Piotr and Zbyszek
http://www.meetup.com/Trojmiasto-Java-User-Group/events/230290254/

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
@@ -184,13 +184,13 @@ public final class TypeConverters {
             if (value instanceof BigInteger) {
                 return new BigDecimal((BigInteger) value);
             }
-            if (isNotFloatingPointNumber(value)) {
+            if (isIntegralDataType(value)) {
                 Number number = (Number) value;
                 return BigDecimal.valueOf(number.longValue());
             }
-            if (value instanceof Double) {
+            if (isFloatingPointDataType(value)) {
                 Number number = (Number) value;
-                return BigDecimal.valueOf(number.doubleValue());
+                return new BigDecimal(number.doubleValue());
             }
             if (value instanceof Boolean) {
                 return ((Boolean) value) ? BigDecimal.ONE : BigDecimal.ZERO;
@@ -198,7 +198,11 @@ public final class TypeConverters {
             return new BigDecimal(value.toString());
         }
 
-        private boolean isNotFloatingPointNumber(Comparable value) {
+        private boolean isFloatingPointDataType(Comparable value) {
+            return value instanceof Double || value instanceof Float;
+        }
+
+        private boolean isIntegralDataType(Comparable value) {
             return value instanceof Byte || value instanceof Short || value instanceof Integer || value instanceof Long;
         }
     }
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
@@ -169,7 +169,7 @@ public final class TypeConverters {
                 return BigInteger.valueOf(number.longValue());
             }
             if (value instanceof Boolean) {
-                return ((Boolean) value) ? BigInteger.ZERO : BigInteger.ONE;
+                return ((Boolean) value) ? BigInteger.ONE : BigInteger.ZERO;
             }
             return new BigInteger(value.toString());
         }
@@ -193,7 +193,7 @@ public final class TypeConverters {
                 return BigDecimal.valueOf(number.doubleValue());
             }
             if (value instanceof Boolean) {
-                return ((Boolean) value) ? BigDecimal.ZERO : BigDecimal.ONE;
+                return ((Boolean) value) ? BigDecimal.ONE : BigDecimal.ZERO;
             }
             return new BigDecimal(value.toString());
         }
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
@@ -160,6 +160,17 @@ public final class TypeConverters {
             if (value instanceof BigInteger) {
                 return value;
             }
+            if (value instanceof BigDecimal) {
+                BigDecimal decimal = (BigDecimal) value;
+                return decimal.toBigInteger();
+            }
+            if (value instanceof Number) {
+                Number number = (Number) value;
+                return BigInteger.valueOf(number.longValue());
+            }
+            if (value instanceof Boolean) {
+                return ((Boolean) value) ? BigInteger.ZERO : BigInteger.ONE;
+            }
             return new BigInteger(value.toString());
         }
     }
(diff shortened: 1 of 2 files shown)


cpiotr:
run-lab-run

tombujok:
run-lab-run

tombujok:
run-lab-run

hasancelik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
```diff
@@ -160,6 +160,17 @@ Comparable convertInternal(Comparable value) {
...
+                BigDecimal decimal = (BigDecimal) value;
+                return decimal.toBigInteger();
+            }
+            if (value instanceof Number) {
+                Number number = (Number) value;
+                return BigInteger.valueOf(number.longValue());
+            }
+            if (value instanceof Boolean) {
+                return ((Boolean) value) ? BigInteger.ZERO : BigInteger.ONE;
+            }
```
tombujok: This should be inverted ;) ONE in case of true and ZERO otherwise.

hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
```diff
@@ -173,8 +184,23 @@ Comparable convertInternal(Comparable value) {
...
+                Number number = (Number) value;
+                return BigDecimal.valueOf(number.longValue());
+            }
+            if (value instanceof Double) {
+                Number number = (Number) value;
+                return BigDecimal.valueOf(number.doubleValue());
+            }
+            if (value instanceof Boolean) {
+                return ((Boolean) value) ? BigDecimal.ZERO : BigDecimal.ONE;
+            }
```
tombujok: Here too

hazelcast/src/test/java/com/hazelcast/query/impl/TypeConverterTest.java
```diff
@@ -39,4 +53,246 @@ Comparable convertInternal(Comparable value) {
...
+                is(equalTo(expectedBigIntValue))
+        ));
+    }
+
+    @Test
+    public void testBigIntegerConvert_whenPassedBooleanValue_thenConvertToBigInteger() throws Exception {
+        Boolean value = Boolean.TRUE;
+        Comparable expectedBigIntValue = BigInteger.ZERO;   // Boolean TRUE means 0
+                                                            // Boolean FALSE means something non-zero, i.e. 1
+
```
tombujok: Now, I've notice that it's on purpose, that true means zero, but it's non-compatible with the old BooleanConverter which does sth like this to convert a number to boolean
```
          if (value instanceof Number) {
                Number number = (Number) value;
                return number.intValue() != 0;
            }
```

tombujok:
@cpiotr the PR looks very good! There's  just this tiny confusion about how to handle number to boolean conversion. Would be great if you could fix that.

cpiotr:
@tombujok, sure, thanks for the review. Somehow, I got the boolean conversion wrong.

hazelcast/src/test/java/com/hazelcast/query/impl/TypeConverterTest.java
```diff
@@ -39,4 +53,246 @@ Comparable convertInternal(Comparable value) {
...
+                is(instanceOf(BigInteger.class)),
+                is(equalTo(expectedBigIntValue))
+        ));
+    }
+
+    @Test
+    public void testBigIntegerConvert_whenPassedBooleanValue_thenConvertToBigInteger() throws Exception {
+        Boolean value = Boolean.TRUE;
+        Comparable expectedBigIntValue = BigInteger.ZERO;   // Boolean TRUE means 0
+                                                            // Boolean FALSE means something non-zero, i.e. 1
```
tombujok: It would also great to inline the trailing comments. In general that's in the checkstyle rules but they are not evaluated against the test code for now.

cpiotr:
@tombujok Fixed issues related to boolean values.

tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
```diff
@@ -173,8 +184,23 @@ Comparable convertInternal(Comparable value) {
...
                 return new BigDecimal((BigInteger) value);
             }
+            if (isNotFloatingPointNumber(value)) {
+                Number number = (Number) value;
+                return BigDecimal.valueOf(number.longValue());
+            }
+            if (value instanceof Double) {
+                Number number = (Number) value;
+                return BigDecimal.valueOf(number.doubleValue());
+            }
```
tombujok: Just a pickle here, the BigDecimal.valueOf uses the Double.toString(), which makes it slow too. Here's the snippet.
```
public static BigDecimal valueOf(double val) {
        // Reminder: a zero double returns '0.0', so we cannot fastpath
        // to use the constant ZERO.  This might be important enough to
        // justify a factory approach, a cache, or a few private
        // constants, later.
        return new BigDecimal(Double.toString(val));
    }
```
It would be better tom use the `new BigDecimal(double)`.
WDYT?

hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
```diff
@@ -173,8 +184,23 @@ Comparable convertInternal(Comparable value) {
             if (value instanceof BigInteger) {
                 return new BigDecimal((BigInteger) value);
             }
+            if (isNotFloatingPointNumber(value)) {
+                Number number = (Number) value;
+                return BigDecimal.valueOf(number.longValue());
+            }
+            if (value instanceof Double) {
```
tombujok: Maybe we could also add `value instance of Float` and then do `new BigDecimal(float.doubleValue)`

tombujok:
@cpiotr Thx for the changes! Just have two minor comments spotted now. Apart from that looks good.

cpiotr:
@tombujok 
`new BigDecimal(double)` is somewhat unpredicatble. Can accuracy be traded for performance in this case?
From: [javadoc](https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#BigDecimal(double))

> The results of this constructor can be somewhat unpredictable. One might assume that writing new BigDecimal(0.1) in Java creates a BigDecimal which is exactly equal to 0.1 (an unscaled value of 1, with a scale of 1), but it is actually equal to 0.1000000000000000055511151231257827021181583404541015625. This is because 0.1 cannot be represented exactly as a double (or, for that matter, as a binary fraction of any finite length). Thus, the value that is being passed in to the constructor is not exactly equal to 0.1, appearances notwithstanding.
> The String constructor, on the other hand, is perfectly predictable: writing new BigDecimal("0.1") creates a BigDecimal which is exactly equal to 0.1, as one would expect. Therefore, it is generally recommended that the String constructor be used in preference to this one.
> When a double must be used as a source for a BigDecimal, note that this constructor provides an exact conversion; it does not give the same result as converting the double to a String using the Double.toString(double) method and then using the BigDecimal(String) constructor. To get that result, use the static valueOf(double) method.

tombujok:
Yep, I see, but in this case it's about a conversion from double, so it means somebody already has a double, let's say like `0.1000000000000000055511151231257827021181583404541015625`.
Doing `toString` on it and then converting to `BigDecimal` won't make any difference.

If I understand it correctly, what you wrote plays a role if I want to initialise a BigDecimal from scratch and then it's better to use new `BigDecimal.valueOf("0.1")` instead of `new BigDecimal(0.1)`, right?

cpiotr:
Seems valid, I'll get back to that :)

tombujok:
Great. One more formality before I can merge the PR. Could you send us a signed scan of the Hazelcastr Contributor Agreement?

tombujok:
run-lab-run

tombujok:
Thx @cpiotr - looks great! Now I just need the CLA to finish with the merge process. 
Could plz you send it to tom[at]hazelcast.com? I would appreciate it.

tombujok:
👍 (CLA signed)

pveentjer:
:+1:

PR #8035, opened 2016-04-27
[DO NOT MERGE] Caching converted values in predicates. fix #7589

### LINKED ISSUE #7589: Add caching of the converted values specified to the predicates
Currently the predicates do not cache the values specified to them, so if I use the BetweenPredicate for example the `from` and `to` values will be converted for each mapEntry specified to the predicate.
https://github.com/hazelcast/hazelcast/blob/eef987b9920f3f7b4c506e433b0d067cab9ba93c/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/BetweenPredicate.java#L54-L55

The converted values may be cached only if the `attributeValue` passed to the `applyForSingleAttributeValue` method is not null for the first time - otherwise the converter is null too, since the converter is initialised on the basis of the entry's `attributeValue` runtime-type.
https://github.com/hazelcast/hazelcast/blob/eef987b9920f3f7b4c506e433b0d067cab9ba93c/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/BetweenPredicate.java#L50

This will be a significant performance improvement if the type specified to the predicate is of different type than the entry's attribute value.

airborn:
Work done during Hackergarten in Gdansk by Patryk Makuch and Jacek Jackowiak
http://www.meetup.com/Trojmiasto-Java-User-Group/events/230290254/

Fixes: https://github.com/hazelcast/hazelcast/pull/7589

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/BetweenPredicate.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/BetweenPredicate.java
@@ -31,8 +31,10 @@ import java.util.Set;
  */
 public class BetweenPredicate extends AbstractIndexAwarePredicate {
 
-    Comparable to;
-    Comparable from;
+    protected Comparable from;
+    protected Comparable to;
+    private transient volatile Comparable fromConverted;
+    private transient volatile Comparable toConverted;
 
     public BetweenPredicate() {
     }
@@ -51,12 +53,16 @@ public class BetweenPredicate extends AbstractIndexAwarePredicate {
         if (attributeValue == null) {
             return false;
         }
-        Comparable fromConvertedValue = convert(entry, attributeValue, from);
-        Comparable toConvertedValue = convert(entry, attributeValue, to);
-        if (fromConvertedValue == null || toConvertedValue == null) {
+        if (fromConverted == null) {
+            fromConverted = convert(entry, attributeValue, from);
+        }
+        if (toConverted == null) {
+            toConverted = convert(entry, attributeValue, to);
+        }
+        if (fromConverted == null || toConverted == null) {
             return false;
         }
-        return attributeValue.compareTo(fromConvertedValue) >= 0 && attributeValue.compareTo(toConvertedValue) <= 0;
+        return attributeValue.compareTo(fromConverted) >= 0 && attributeValue.compareTo(toConverted) <= 0;
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/EqualPredicate.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/EqualPredicate.java
@@ -34,6 +34,7 @@ import java.util.Set;
 public class EqualPredicate extends AbstractIndexAwarePredicate implements NegatablePredicate {
 
     protected Comparable value;
+    private transient volatile Comparable valueConverted;
 
     public EqualPredicate() {
     }
@@ -57,8 +58,10 @@ public class EqualPredicate extends AbstractIndexAwarePredicate implements Negat
         if (attributeValue == null) {
             return value == null || value == IndexImpl.NULL;
         }
-        value = convert(mapEntry, attributeValue, value);
-        return attributeValue.equals(value);
+        if (valueConverted == null) {
+            valueConverted = convert(mapEntry, attributeValue, value);
+        }
+        return attributeValue.equals(valueConverted);
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


tombujok:
run-lab-run

tombujok:
run-lab-run

bartoszmajsak:
go-gadget-go

tombujok:
@airborn I forgot about one thing. Could you make the *converted fields `transient volatile`?

tombujok:
...and one more formality before I can merge the PR. Could you send us a signed scan of the Hazelcastr Contributor Agreement?
https://hazelcast.atlassian.net/wiki/download/attachments/6357071/Hazelcast%2C%20Inc.%20Individual%20Contrbutor%20License%20Agreement%20v100113.pdf?version=2&modificationDate=1394826512258&api=v2

airborn:
@tombujok done. 
By now you should alsob be abe to find signed copy of CLA in yours inbox.

tombujok:
run-lab-run

tombujok:
verify

tombujok:
run-lab-run

airborn:
@tombujok what is the issue with the tests?

tombujok:
run-lab-run

tombujok:
👍

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/BetweenPredicate.java
```diff
@@ -31,8 +31,10 @@
  */
 public class BetweenPredicate extends AbstractIndexAwarePredicate {
 
-    Comparable to;
-    Comparable from;
+    protected Comparable from;
+    protected Comparable to;
+    private transient volatile Comparable fromConverted;
```
pveentjer: Why do these fields need to be volatile?

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/EqualPredicate.java
```diff
@@ -34,6 +34,7 @@
 public class EqualPredicate extends AbstractIndexAwarePredicate implements NegatablePredicate {
 
     protected Comparable value;
+    private transient volatile Comparable valueConverted;
```
pveentjer: Why volatile?

pveentjer:
Why are the fields volatile? This would suggest that multiple threads are using the same predicate instance. And since they have shared state; I'm worried.

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/BetweenPredicate.java
tombujok: The predicate instance is shared by partitions on each node. Thus if one initialises the other ones should see it properly.

tombujok:
Sure, see my answer aboce. It's safe in this case. The same thing happens already with the predicate type.

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/BetweenPredicate.java
pveentjer: Ok. So effectively it is a caching mechanism; once set it will not change. In that cases volatile makes sense.

tombujok: yep, exactly

pveentjer:
:+1:

pveentjer:
Are you sure that the volatile field content doesn't have non threadsafe state that could potentially cause problems?

jerrinot:
:+1:

tombujok:
That was a great question Peter! I just double-check the implementation of `equals` on `java.util.Date` (since it's mutable) and surprise-surprise the `equals` method is not thread safe...

pveentjer:
In its current state we can't merge. 

How to move forward? One obvious approach to allow for caching, is to clone the predicate so query thread gets its own instance.

tombujok:
A fix is required definitely even for our `default` case now. 
The only fully-safe fix I can think of is to clone the predicate for each thread it's evaluating it.it's serialisable anyway.
As @jerrinot  pointed out there's only one thread for each query by default, so it means no overhead for the default case.
Otherwise people can provide the predicate with their own `Comparable` and we can never be sure if it's a thread-safe class or not.

I'll follow up on it next week.

tombujok:
@airborn OK, we'll be ready to merge this one in 3.8 after this one is merged #8378, just without the volatile keyboard. Could you do this minor change?

airborn:
@tombujok sure, no problem.

tombujok:
Another twist of action here. It will be merge-ready with the `volatile` keyword after #8828 is fixed.

mmedenjak:
@tombujok since https://github.com/hazelcast/hazelcast/issues/7589 and https://github.com/hazelcast/hazelcast/issues/8828 have been moved to 3.10, can this be moved as well?

Holmistr:
@mmedenjak Any progress here?

Holmistr:
@mmedenjak Is there any plan for this PR?

PR #8042, opened 2016-04-28
Deduplicate client AtomicLong code

### LINKED ISSUE #8041: Cleanup ClientAtomicLongProxy implementation of synchronous methods
As discussed in https://github.com/hazelcast/hazelcast/pull/7960#issuecomment-211207677, change implementation of `ClientAtomicLongProxy` sync methods to build on async methods.

vbekiaris:
Fixes #8041 .

Synchronous variants of atomic long methods in `ClientAtomicLongProxy` are now implemented on top of their async counterparts.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientAtomicLongProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientAtomicLongProxy.java
@@ -32,8 +32,8 @@ import com.hazelcast.client.impl.protocol.codec.AtomicLongGetCodec;
 import com.hazelcast.client.impl.protocol.codec.AtomicLongIncrementAndGetCodec;
 import com.hazelcast.client.impl.protocol.codec.AtomicLongSetCodec;
 import com.hazelcast.core.IAtomicLong;
-import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.IFunction;
+import com.hazelcast.spi.InternalCompletableFuture;
 
 import static com.hazelcast.util.Preconditions.isNotNull;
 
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientAtomicLongProxy.java
```diff
@@ -140,185 +140,146 @@ public ClientAtomicLongProxy(String serviceName, String objectId) {
...
     @Override
     public void set(long newValue) {
-        ClientMessage request = AtomicLongSetCodec.encodeRequest(name, newValue);
-        invokeOnPartition(request);
+        setAsync(newValue).join();
     }
 
     @Override
-    public ICompletableFuture<Long> addAndGetAsync(long delta) {
+    public InternalCompletableFuture<Long> addAndGetAsync(long delta) {
```
jerrinot: why `InternalCompletableFuture` ?

vbekiaris: To use `xAsync().join()`  in sync methods and reduce boilerplate of wrapping `Future.get()` within try/catch.

jerrinot: I see. I'm a bit easy about moving `InternalCompletableFuture` "too close" to the user. but it's just "gut feeling". I have no hard arguments. 
@tombujok @pveentjer: WDYT?

vbekiaris: Perhaps a scheme like this?
```java
@Override
public long getAndIncrement() {
  return getAndIncrementAsyncInternal().join();
}

@Override
public ICompletableFuture<V> getAndIncrementAsync() {
 return getAndIncrementAsyncInternal();
}

private InternalCompletableFuture<V> getAndIncrementAsyncInternal();
```

jerrinot: that would work. 
the question is whether it's worth it in the situation when my objections against the current approach are based purely on gut feeling:)

pveentjer:
I don't see any problem using an internal interface in internal code. So I don't have any objections.

:+1:

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientAtomicLongProxy.java
pveentjer: I understand your objection because he has a covariant return type which 'exposes' a stronger interface.

In this case it is perfectly fine since it isn't visible to the end user. 

Then I also understand the 'need' for the internal method. Imho this is not needed and it leads to code noise for no good reason.

pveentjer:
:+1:

PR #8045, opened 2016-04-29
Fix cluster state transaction when a member leaves during commit

mdogan:
When a member leaves during commit of a cluster state tx, after prepare
is completed, tx should be completed successfully. We can assume it's left
after tx completed, because it will be added to the list of members which are
removed while cluster is not active and will be able to rejoin cluster.
Effectively there's no difference between leaving after commit vs during commit.


metanet:
👍

PR #8048, opened 2016-04-29
Read/write-handler cleanup

pveentjer:
Removed writehandler.start since always no-op
Renamed writehandler.shutdown to close
Renamed readhandler.destroy to close
Renamed writehandler.offer to write


Donnerbart:
:+1: (once green)

PR #8051, opened 2016-04-29
Prevent resizing of ArrayList in InternalPartitionServiceImpl

Donnerbart:
Prevent frequent resizing of `ArrayList` in `InternalPartitionServiceImpl.getMemberPartitionsMap()`.

I found this by chance in a profiler run for my `IMap.putAll()` optimizations. We are just missing to initialize the `ArrayList` with a good estimation to prevent a frequent resizing.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
@@ -86,6 +86,7 @@ import java.util.logging.Level;
 import static com.hazelcast.cluster.memberselector.MemberSelectors.DATA_MEMBER_SELECTOR;
 import static com.hazelcast.util.FutureUtil.logAllExceptions;
 import static com.hazelcast.util.FutureUtil.returnWithDeadline;
+import static java.lang.Math.ceil;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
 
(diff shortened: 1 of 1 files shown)


hasancelik:
👍

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -832,15 +833,17 @@ public boolean isPartitionReplicaVersionStale(int partitionId, long[] versions,
 
     @Override
     public Map<Address, List<Integer>> getMemberPartitionsMap() {
-        final Collection<Member> dataMembers = node.getClusterService().getMembers(DATA_MEMBER_SELECTOR);
-        final int dataMembersSize = dataMembers.size();
+        Collection<Member> dataMembers = node.getClusterService().getMembers(DATA_MEMBER_SELECTOR);
+        int dataMembersSize = dataMembers.size();
```
ahmetmircik: Can `dataMembersSize` be zero? If yes, we will have an array initialized with int-max capacity and that means possible OOME.

jerrinot: good one. there should be protection against this. the method should not make assumptions about memberSize > 0. 

@Donnerbart: can you please send a fix.

Donnerbart: Good finding! Actually this can happen if you have a LiteMember only cluster. There is even a test for it, which didn't fail since an exception is thrown in that case. So effectively nothing would have happened, since the exception is thrown before the `ArrayList` is created. But that's risky and so a fix is the better solution: https://github.com/hazelcast/hazelcast/pull/8054

PR #8053, opened 2016-04-29
Adds Index Support For EntryProcessors With A Predicate

ahmetmircik:
Entry-processor can select processable entries by using a predicate. Currently, this selection happens by doing a full-table-scan in a partition. Instead of full-table-scan, indexes can be used if they are available

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -93,6 +93,7 @@ import com.hazelcast.core.IMap;
 import com.hazelcast.core.IMapEvent;
 import com.hazelcast.core.MapEvent;
 import com.hazelcast.core.Member;
+import com.hazelcast.internal.serialization.InternalSerializationService;
 import com.hazelcast.map.EntryProcessor;
 import com.hazelcast.map.MapInterceptor;
 import com.hazelcast.map.MapPartitionLostEvent;
@@ -1071,7 +1072,8 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapEntrySetCodec.ResponseParameters resultParameters = MapEntrySetCodec.decodeResponse(response);
 
         InflatableSet.Builder<Entry<K, V>> setBuilder = InflatableSet.newBuilder(resultParameters.response.size());
-        SerializationService serializationService = getContext().getSerializationService();
+        InternalSerializationService serializationService
+                = ((InternalSerializationService) getContext().getSerializationService());
         for (Entry<Data, Data> row : resultParameters.response) {
             LazyMapEntry entry = new LazyMapEntry(row.getKey(), row.getValue(), serializationService);
             setBuilder.add(entry);
@@ -1128,7 +1130,8 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
         MapEntriesWithPredicateCodec.ResponseParameters resultParameters = MapEntriesWithPredicateCodec.decodeResponse(response);
 
         InflatableSet.Builder<Entry<K, V>> setBuilder = InflatableSet.newBuilder(resultParameters.response.size());
-        SerializationService serializationService = getContext().getSerializationService();
+        InternalSerializationService serializationService
+                = ((InternalSerializationService) getContext().getSerializationService());
         for (Entry<Data, Data> row : resultParameters.response) {
             LazyMapEntry entry = new LazyMapEntry(row.getKey(), row.getValue(), serializationService);
             setBuilder.add(entry);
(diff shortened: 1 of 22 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
```diff
@@ -99,4 +100,12 @@ QueryResult queryLocalPartitions(String name, Predicate predicate, IterationType
...
      */
     Set queryAllPartitionsWithPagingPredicate(String mapName, PagingPredicate pagingPredicate, IterationType iterationType);
+
+
+    /**
+     * Returns {@link QueryOptimizer} for this engine.
+     *
+     * @return {@link QueryOptimizer} for this engine.
+     */
+    QueryOptimizer getQueryOptimizer();
```
jerrinot: I'm not super excited about this. optimization should be an internal concern. is there any other option?

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
ahmetmircik: one option is using MapQueryEngine-impl or another one is putting optimizer to map-service-context

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1071,9 +1072,10 @@ protected void loadAllInternal(boolean replaceExistingValues, Collection<Data> d
         MapEntrySetCodec.ResponseParameters resultParameters = MapEntrySetCodec.decodeResponse(response);
 
         InflatableSet.Builder<Entry<K, V>> setBuilder = InflatableSet.newBuilder(resultParameters.response.size());
-        SerializationService serializationService = getContext().getSerializationService();
+        InternalSerializationService serializationService
+                = ((InternalSerializationService) getContext().getSerializationService());
         for (Entry<Data, Data> row : resultParameters.response) {
-            LazyMapEntry entry = new LazyMapEntry(row.getKey(), row.getValue(), serializationService);
+            LazyMapEntry entry = new LazyMapEntry(row.getKey(), row.getValue(), serializationService, null);
             setBuilder.add(entry);
```
tombujok: Hmm, this has no chance of working with that null. Where is this entry used?

ahmetmircik: it is not doing anything that requires extractors so passing null should be fine or a new constructor can be introduced.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
jerrinot: or expose `optimize()` method in the `MapQueryEngine`? so the optimization code will stay encapsulated inside the `QueryEngine`

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
tombujok: So this entry is never used to extract any attributes from it? The Extractors object does not only provide the ValueExtractors, but it provides the ability to extract value of any property.

ahmetmircik: right, no extraction.

ahmetmircik:
verify

ahmetmircik:
verify

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1071,7 +1072,8 @@ protected void loadAllInternal(boolean replaceExistingValues, Collection<Data> d
         MapEntrySetCodec.ResponseParameters resultParameters = MapEntrySetCodec.decodeResponse(response);
 
         InflatableSet.Builder<Entry<K, V>> setBuilder = InflatableSet.newBuilder(resultParameters.response.size());
-        SerializationService serializationService = getContext().getSerializationService();
+        InternalSerializationService serializationService
```
jerrinot: why you need an `InternalSerializationService` here?

ahmetmircik:
verify

jerrinot:
:+1:

gurbuzali:
👍

ahmetmircik:
verify

tombujok:
I've just reviewed it and the implementation looks fine!
What really bothers me here is that there's not a single test-case for this feature.
Even one integration test would do fine to see if the index is used indeed and that there's no exception thrown, etc. Otherwise we have no control over the possible regression, thus I would restrain from a thumbs-up until a test-case is added.

jerrinot:
Please do not merge this before tests are added.

ahmetmircik:
added tests. @tombujok @jerrinot

PR #8054, opened 2016-04-29
Fixed possible division by zero in InternalPartitionServiceImpl

Donnerbart:
* Fixed possible division by zero in `InternalPartitionServiceImpl.getMemberPartitionsMap()`.
* Small cleanup of `InternalPartitionServiceImpl` test classes.

Thx @ahmetmircik for this finding :)

I first wanted to directly return the empty map, but actually the `getPartitionOwnerOrWait()` method throws a `NoDataMemberInClusterException` if there are no data members in the cluster. I didn't want to change that, so I just fixed the initial size if there are no data members, and let the method throw its exception (there is also a test for this in `InternalPartitionServiceLiteMemberTest`).


pveentjer:
:+1:

PR #8056, opened 2016-04-30
Concurrent operation cleanup

pveentjer:
Renamed 'Base' operation to 'Abstract' since that is the standard naming
strategy used for abstract classes.

Pulled IdentifiedDataSerializable into the Abstract class instead of repeating
it on the subclasses

Extending AbstractOperation instead of Operation to remove more duplication


pveentjer:
run-lab-run

pveentjer:
run-lab-run

vbekiaris:
👍 with a minor note to make `AbstractAtomicLongOperation.getFactoryId` final in https://github.com/hazelcast/hazelcast/pull/8056/files#diff-d222c2ef9d647746a4763419eb9e4807R53

pveentjer:
Thanks for the review. I'll make it final in a direct commit

PR #8057, opened 2016-04-30
AtomicReference.compareAndSet backup error fix

pveentjer:
It backups up when there is no change, and it doesn't backup when there
is a change.


pveentjer:
run-lab-run

jerrinot:
:+1: good finding

pveentjer:
run-lab-run

PR #8059, opened 2016-04-30
OperationRunnerImpl.thisAddress

pveentjer:
Added field to fill in thisAddress. This reduces coupling with node/nodeEngine since only in the
constructor there is a reliance on node to retrieve the thisAddress. This makes it easier in the
future to get rid of Node engine depedency (the thisAddress can easily be injected using the
constructor).


Donnerbart:
:+1: (once green)

pveentjer:
run-lab-run

pveentjer:
run-lab-run

PR #8061, opened 2016-05-02
Inv: Renamed sync backup count to backup ack count

pveentjer:
Acks are more standard. The ack count is the actual number of acknowledgments that need to be received before continuing with the next invocation.

Also the backupresponse writes a byte instead of an int for the backup-ack-acount since number of backups < 127. This shaves 3 bytes from each normal response.


jerrinot:
:+1: the rename makes some memory stale, but the change is quite well isolated and it makes it easier to read -> it's worth it.

Issue #8026, opened 2016-04-27
Invocation_NetworkSplitTest Failures


hasancelik:
**testWaitNotifyService_whenNodeSplitFromCluster**

```
java.lang.AssertionError: expected:<4> but was:<3>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:834)
    at org.junit.Assert.assertEquals(Assert.java:645)
    at org.junit.Assert.assertEquals(Assert.java:631)
    at com.hazelcast.spi.impl.operationservice.impl.Invocation_NetworkSplitTest.testWaitNotifyService_whenNodeSplitFromCluster(Invocation_NetworkSplitTest.java:178)
    at com.hazelcast.spi.impl.operationservice.impl.Invocation_NetworkSplitTest.testWaitNotifyService_whenNodeSplitFromCluster(Invocation_NetworkSplitTest.java:125)
```

**testWaitNotifyService_whenNodePartiallySplitFromCluster**

```
java.lang.AssertionError: expected:<4> but was:<3>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:834)
    at org.junit.Assert.assertEquals(Assert.java:645)
    at org.junit.Assert.assertEquals(Assert.java:631)
    at com.hazelcast.spi.impl.operationservice.impl.Invocation_NetworkSplitTest.testWaitNotifyService_whenNodeSplitFromCluster(Invocation_NetworkSplitTest.java:177)
    at com.hazelcast.spi.impl.operationservice.impl.Invocation_NetworkSplitTest.testWaitNotifyService_whenNodePartiallySplitFromCluster(Invocation_NetworkSplitTest.java:131)
```

**testWaitingInvocations_whenNodePartiallySplitFromCluster_scenario2**

```
ava.lang.AssertionError: Cluster size is not correct. Expected: 3 Actual: 2
    at org.junit.Assert.fail(Assert.java:88)
    at com.hazelcast.test.HazelcastTestSupport.assertClusterSize(HazelcastTestSupport.java:702)
    at com.hazelcast.test.HazelcastTestSupport$10.run(HazelcastTestSupport.java:721)
    at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:857)
    at com.hazelcast.test.HazelcastTestSupport.assertClusterSizeEventually(HazelcastTestSupport.java:717)
[...]

PR #8067, opened 2016-05-02
MockConnectionManager & MockJoiner fixes 

mdogan:
- MockConnectionManager should remove only cluster members while stopping
itself. Otherwise, async member removal task can lead to member split
just after cluster merge.

- MockJoiner should search for master only when there's no master candidate
or found candidate can not be connected.

- Also adding debug level logs for MockJoiner.

Fixes https://github.com/hazelcast/hazelcast/issues/8026


metanet:
👍

PR #8072, opened 2016-05-03
InvocationContext inner class and fields final

pveentjer:
To prevent the InvocationContext being abused for other purposes than Invocation, it has been made a static inner class of the Invocation and renamed to Context.

Also the Context fields are now finally and we get truly amazing test construction of the context.


tombujok:
👍

pveentjer:
run-lab-run

jerrinot:
:+1: on green
in 3.8 we should remove `Node` from the `InvocationContext`

pveentjer:
Probably we can get rid of operationservice as well.

Donnerbart:
:+1: (once green)

PR #8082, opened 2016-05-03
Fix proposal for CachePartitionEventData FindBugs issue

Donnerbart:
@sancar just added `equals()` and `hashCode()` to `CacheEventDataImpl`. So FindBugs doesn't like `CachePartitionEventData` anymore, which extends `CacheEventDataImpl` and has two more fields.

I don't know those classes well, but this is what IDEA creates to extend `equals()` and `hashCode()` for the two additional fields.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheEventDataImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheEventDataImpl.java
@@ -130,7 +130,6 @@ public class CacheEventDataImpl
         }
 
         CacheEventDataImpl that = (CacheEventDataImpl) o;
-
         if (isOldValueAvailable != that.isOldValueAvailable) {
             return false;
         }
@@ -147,7 +146,6 @@ public class CacheEventDataImpl
             return false;
         }
         return dataOldValue != null ? dataOldValue.equals(that.dataOldValue) : that.dataOldValue == null;
-
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Donnerbart:
verify

jerrinot:
:+1:

Donnerbart:
Hasan just fixed the immediate FindBugs issue with this commit: https://github.com/hazelcast/hazelcast/commit/52afb48bc00404f46c4599834ce1bfa207b7b574

But I'm not sure if it's a proper `equals()` and `hashCode()` implementation for `CachePartitionEventData`.

jerrinot:
the commit has made is wrong. It just satisfied FindBugs, but the root cause is still there. 
It should be done your way. CC @hasancelik

hasancelik:
@jerrinot @Donnerbart Guys, sorry for that. My fix is very unhealthy.

Donnerbart:
No problem, at least it stopped the mail spam :) And this one still seems to be mergeable, so no need to worry.

Donnerbart:
@kobalski seems to be the original author of the class. Would you be so kind to have a look as well?

The `equals()` and `hashCode()` methods were generated by IDEA, so I think it should the code itself be okay. It's more the question if both fields should be used to check for equality or e.g. just the `partitionId` (which is the only one added in the `toString()` method).

kobalski:
👍 Thanks @Donnerbart for the fix. `CachePartitionEventData ` is used for Partition Lost Listener for JCache and copied from `MapPartitionEventData`. I am merging the pr.

PR #8086, opened 2016-05-04
Simplify MockJoiner

mdogan:
Simplify MockJoiner mechanism to use present joiner system
as much as possible. Previous code was very complicated and difficult
to track problems.


metanet:
👍

pveentjer:
:+1:

PR #8087, opened 2016-05-04
Added the ConnectionListenable

pveentjer:
For another PR the client/member connectionmanager needs have a single 'listen' interface
so I can listen to connection changes. Using the ConnectionListenable abstraction this is possible.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/ClientConnectionManager.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/ClientConnectionManager.java
@@ -20,14 +20,14 @@ import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.spi.impl.ConnectionHeartbeatListener;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
-import com.hazelcast.nio.ConnectionListener;
+import com.hazelcast.nio.ConnectionListenable;
 
 import java.io.IOException;
 
 /**
  * Responsible for managing {@link com.hazelcast.client.connection.nio.ClientConnection} objects.
  */
-public interface ClientConnectionManager {
+public interface ClientConnectionManager extends ConnectionListenable {
 
     /**
      * Shutdown clientConnectionManager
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/nio/ConnectionObservable.java
```diff
@@ -0,0 +1,36 @@
...
+package com.hazelcast.nio;
+
+import com.hazelcast.spi.annotation.PrivateApi;
+
+/**
+ * Provides connection listen capabilities.
+ */
+@PrivateApi
+public interface ConnectionObservable {
+
```
tombujok: Hmm, `observable` would imply that there's a `addObserver` method.
I would propose renaming it to `ConnectionListenable`

hazelcast/src/main/java/com/hazelcast/nio/ConnectionObservable.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Registers a ConnectionListener.
+     *
+     * If the same listener is registered multiple times, it will be notified multiple times.
+     *
+     * @param listener the ConnectionListener to add.
+     * @throws NullPointerException if listener is null.
+     */
+    void addConnectionListener(ConnectionListener listener);
+}
```
tombujok: `register` would be a better name in this context. I don't know if we may rename it (maybe it has to be backward compatible or sth)

hazelcast/src/main/java/com/hazelcast/nio/ConnectionObservable.java
pveentjer: Sure

hazelcast/src/main/java/com/hazelcast/nio/ConnectionObservable.java
pveentjer: No compatibility since connection manager is private api. But if we call it register, it should be done consistently throughout the platform.

tombujok:
👍 (on green)

pveentjer:
run-lab-run

PR #8088, opened 2016-05-04
Address toString simplification

pveentjer:
Instead of Address[1.2.3.4]:5701, you get [1.2.3.4]:5701. So the 'Address' part has been removed.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/Address.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/Address.java
@@ -186,7 +186,7 @@ public final class Address implements IdentifiedDataSerializable {
 
     @Override
     public String toString() {
-        return "Address[" + getHost() + "]:" + port;
+        return '[' + host + "]:" + port;
     }
 
     private static InetAddress resolve(InetSocketAddress inetSocketAddress) {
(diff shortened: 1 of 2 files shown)


mdogan:
You should keep brackets `[..]`. Otherwise for IPv6 it will be difficult to notice port number;  `72:00:08:c3:c8:61:5701` vs `[72:00:08:c3:c8:61]:5701`

pveentjer:
Thanks for the comment:

What about:
```
   @Override
    public String toString() {
        if(isIPv4()) {
            return host + ':' + port;
        }else{
            return '['+host + "]:" + port;
        }
    }
```

mdogan:
For me, having brackets always makes easier to parse host and port; `192.168.1.3:5701` vs `[192.168.1.3]:5701` or `some.domain.com:5701` vs `[some.domain.com]:5701`. It's a preference... 

But for IPv6 it's by spec, host must be in brackets:  


[RFC 3986](http://www.ietf.org/rfc/rfc3986.txt), section 3.2.2:

>A host identified by an Internet Protocol literal address, version 6 [RFC3513] or later, is distinguished by enclosing the IP literal within square brackets ("[" and "]"). This is the only place where square bracket characters are allowed in the URI syntax. In anticipation of future, as-yet-undefined IP literal address formats, an implementation may use an optional version flag to indicate such a format explicitly rather than rely on heuristic determination.

pveentjer:
Ok. I'm fine with having the brackets. As long as the Address part is gone.

mdogan:
👍

pveentjer:
Reason behind this change: The address is used quite a lot (directly/indirectly) in the performance log. So then you get a lot of 'Address' prefixes that don't add value (or create longer lines than needed).

jerrinot:
:+1:

PR #8089, opened 2016-05-04
Minor changes TcpIpConnection.toString

pveentjer:
including connectionid (useful for grepping)
all info between []

THe reason behind this change is that I'm using the connection toString in the metrics system and performance log. And currently it isn't that easy to grep on. Also the content of the string was a bit messy; some parts in [], other parts after.


tombujok:
👍 looks good

pveentjer:
run-lab-run

PR #8090, opened 2016-05-04
Enhanced logging of TargetDisconnectedException

Donnerbart:
* Enhanced logging of `lastHeartBeatMillis` of `ClientConnection` to "never" instead of "1970-01-01 02:00:00.000".
* Created static methods to create `TargetDisconnectedException` within the class.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -33,7 +33,6 @@ import com.hazelcast.internal.metrics.Probe;
 import com.hazelcast.internal.metrics.ProbeLevel;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Connection;
-import com.hazelcast.spi.exception.TargetDisconnectedException;
 
 import java.io.IOException;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -216,12 +214,8 @@ public void run() {
                 }
 
                 iter.remove();
-
-                Exception ex = new TargetDisconnectedException(format(
-                        "Disconnecting from member %s due to heartbeat problems. Current time: %s. Last heartbeat: %s",
-                        connection.getRemoteEndpoint(),
-                        timeToString(System.currentTimeMillis()),
-                        timeToString(connection.getLastHeartbeatMillis())), connection.getCloseCause());
+                Exception ex = newTargetDisconnectedExceptionCausedByHeartBeat(connection.getRemoteEndpoint(),
```
pveentjer: why don't you pass the connection as argument.

Donnerbart: The fields/methods are just available on `ClientConnection` which is not available in the member SPI.

pveentjer:
* long name which has some broken grammar
* why return an Exception when you are calling this method on the TargetDisconnectException?

Donnerbart:
I was thinking about just adding new constructors, but was unsure since the failure messages are very specific, but the different parameter lists would not reflect that. So I switched to static methods with an explanatory name. The names are so long to give context when static imports are used. I don't see broken grammer though: `newExceptionCausedByReason()`

I return the exception because we need it as an argument in the code. What should be the alternative? Directly throw it? Won't work.

I also thought about moving the exception to the client, but it would be the only one which extends `RetryableHazelcastException` there if I didn't miss anything. There also seems to be no `RetryableException` on the client side yet. Of course this will work, but I was not sure if this is a good move.

pveentjer:
:+1:

Issue #3482, opened 2014-09-02
Client transaction requests are blocking on server side which can block all the generic/client thread pool


mdogan:
Client transaction requests are executed as blocking operations on server side by using transactional proxies directly, which can block all the generic/client thread pool at some point. 

Instead they should use non-blocking/async approach and return response when resource available. 

See `AbstractTxnMapRequest` and `BaseTransactionRequest` for instance...

bwzhang2011:
@mdogan, any update with such issue ?

PR #8092, opened 2016-05-04
[DONT MERGE] Removal of serverside client threadpool for generic ops

pveentjer:


mdogan:
You can't remove client thread-pool without rewriting client side transaction impl. 
See https://github.com/hazelcast/hazelcast/issues/3482

pveentjer:
Ok. So appearently that bug has still not been fixed. Let me see how far I get with the non transaction operations.. Then only for transactions we'll use this threadpool.

pveentjer:
@mdogan I have upgraded the pr to only redirect to the txExecutor in case of transactions; otherwise the generic operation is executed by a generic operation thread. 

There was a relaxation required in the InvokeOnPartitions. This method didn't allow to be executed on an operation thread (no matter if generic or partition). It has been changed to now allow on partition threads.

I have not spend a lot of time on this part of the invocation system and the original reason why it can't be executed on an operation thread is not clear to me.

jerrinot:
@pveentjer: We just had a quick peek with @mdogan and there are other cases where a task is blocking. See https://github.com/jerrinot/hazelcast/blob/eef987b9920f3f7b4c506e433b0d067cab9ba93c/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractCallableMessageTask.java#L26-L26 or https://github.com/jerrinot/hazelcast/blob/92a81f05d4c2882783c1eb3dc3c3bcdcad8a5a0a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java#L50-L50

bwzhang2011:
@pveentjer, any udpate with such issue ?

pveentjer:
No update. The issue remains parked.

PR #8095, opened 2016-05-04
Added missed Spring config elements for maps/services/loginModules etc..

dsukhoroslov:
- class-or-bean-name AttributeGroup introduced, element definitions who use class-name/implementation attributes generalized;
- login-module.implementation handling added
- service.implementation handling added
- map.partition-strategy handling added
- multicast.trusted-interfaces handling added

And the following missed attributes added to hazelcast-spring-3.7 schema:

- multicast.loopbackModeEnabled
- aws.iam-role
- executor-service.statistics-enabled
- queue.statistics-enabled
- multimap.statistics-enabled
- list.statistics-enabled
- set.statistics-enabled


hazelcast-spring/src/main/java/com/hazelcast/spring/HazelcastConfigBeanDefinitionParser.java
```diff
@@ -335,6 +339,18 @@ private AbstractBeanDefinition handleService(Node node) {
...
+            final NamedNodeMap attrs = node.getAttributes();
+            Node classNameNode = attrs.getNamedItem("class-name");
+            if (classNameNode != null) { 
+            	serviceConfigBuilder.addPropertyValue("className", getTextContent(classNameNode));
+            }
+            Node implNode = attrs.getNamedItem("implementation");
+            if (implNode != null) { 
+                serviceConfigBuilder.addPropertyReference("implementation", getTextContent(implNode));
+            }
+            Assert.isTrue(classNameSet || classNameNode != null || implNode != null, "One of 'class-name' or 'implementation' "
```
bilalyasar: i think this logic is wrong. or i am missing sth? @dsukhoroslov

hazelcast-spring/src/main/java/com/hazelcast/spring/HazelcastConfigBeanDefinitionParser.java
```diff
@@ -501,7 +532,7 @@ public void handleTcpIp(Node node, BeanDefinitionBuilder joinConfigBuilder) {
                             "interface", "member", "members");
             final ManagedList members = new ManagedList();
             for (Node n : childElements(node)) {
-                String name = xmlToJavaName(cleanNodeName(n));
+                String name = cleanNodeName(n);
```
bilalyasar: why do you removed  `xmlToJavaName`?

hazelcast/src/main/resources/readme.html
```diff
@@ -1,217 +1,217 @@
-<!doctype html>
```
bilalyasar: couldn't get idea here. What is the change here?

hazelcast-spring/src/main/java/com/hazelcast/spring/HazelcastConfigBeanDefinitionParser.java
dsukhoroslov: that was a bug. here the name is compared with xsd attribute name, not with its java representation. have a look at other similar places.

hazelcast/src/main/resources/readme.html
dsukhoroslov: this is not mine. may be came to my commit after merge with upstream?

hazelcast-spring/src/main/java/com/hazelcast/spring/HazelcastConfigBeanDefinitionParser.java
dsukhoroslov: why do you think it is wrong? it s very similar to all other places where pair of class-name/implementation attributes is processed. the only difference is that I check for null attributes itself and not their string values. but they can not be empty by their type specified in XSD: non-space-string.

bilalyasar: you are checking `classNameSet` variable.. Probably it will true when configuration is valid.

dsukhoroslov:
BTW, there was one more bug in `handleMap` method (of the `HazelcastConfigBeanDefinitionParser` class). At line 672 it was:
`                    mapConfigBuilder.addPropertyValue("quorumName", getTextContent(node));`
it must be taken from `childNode` instead. Probably, this fix should be ported to the version 3.6.3 too, And the other fix discussed above.

hazelcast-spring/src/main/java/com/hazelcast/spring/HazelcastConfigBeanDefinitionParser.java
dsukhoroslov: Ah, I see now. I left the `class-name` element in schema for backward compatibility. It would be nice to specify it as deprecated somewhere in docs, and, probably, in the XSD itself. So if the old `class-name` element is set, the configuration is still valid. as well as `class-name` or `implementation` attribute. The proper way would be to remove the `class-name` element at all, but this will break backward compatibility. But this is a common case for major releases (3.6 -> 3.7), actually.

hazelcast/src/main/resources/readme.html
bilalyasar: @dsukhoroslov can you please squash your commits and extract unrelated changes?

dsukhoroslov: @bilalyasar, I just did a rebase, please have a look

dsukhoroslov:
something went wrong with the PR/rebase. I'd better do a new clean one

PR #8101, opened 2016-05-05
Fixes the Diagnistic property names

pveentjer:
The property names are now following `hazelcast.diagnostics` as prefix instead of `hazelcast.performance.monitor`

A change was made to the HazelcastProperty system to easily deal with deprecated names. 

So this PR introduces the new names, but keeps compatibility on the old names.

Also the metrics level configuration is making use of diagnostics prefix (+ keeping compatibility)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
@@ -19,7 +19,6 @@ package com.hazelcast.spi.properties;
 import com.hazelcast.core.IMap;
 import com.hazelcast.instance.BuildInfo;
 import com.hazelcast.instance.BuildInfoProvider;
-import com.hazelcast.internal.metrics.ProbeLevel;
 import com.hazelcast.internal.diagnostics.HealthMonitorLevel;
 import com.hazelcast.map.QueryResultSizeExceededException;
 import com.hazelcast.map.impl.query.QueryResultSizeLimiter;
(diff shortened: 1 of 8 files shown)


jerrinot:
:+1: 
@Donnerbart: can you also have a look too? since this is your favourite area! :)

pveentjer:
@Donnerbart is not available due to Holidays.

hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
```diff
@@ -122,8 +123,9 @@
      * <p/>
      * By default only mandatory probes are being tracked
      */
-    public static final HazelcastProperty PERFORMANCE_METRICS_LEVEL
-            = new HazelcastProperty("hazelcast.performance.metric.level", ProbeLevel.MANDATORY.name());
+    public static final HazelcastProperty PERFORMANCE_METRICS_LEVEL =
+            new HazelcastProperty(Diagnostics.PREFIX + ".metric.level", ProbeLevel.MANDATORY.name())
```
Donnerbart: I'm not 100% sure about the usage of `Diagnostics.PREFIX` here to construct the name. This is the source for the reference manual, and now it has a redirection to other code. If @Serdaro is fine with it, I have no objection though.

pveentjer: I'm going to move this property to the Diagnostics; like all other Diagnostics properties. So it will not be in the SPI any longer.

Donnerbart:
:+1: (once green)

PR #8104, opened 2016-05-05
Node shutdown fixes

mdogan:
- During graceful shutdown, there was a race between checking if that
node is master and reading master address field. If a node becomes master
in this race window, shutdown process fails with an exception:
`java.lang.IllegalArgumentException: Target is this node!...`

- When node shutdown fails, another thread waiting for shutdown to complete
hangs forever, because node-state never becomes `SHUT_DOWN`. Instead, it should
fail with an exception too.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/Node.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/Node.java
@@ -479,13 +479,16 @@ public class Node {
             return;
         }
         logger.info("Node is already shutting down... Waiting for shutdown process to complete...");
-        while (state != NodeState.SHUT_DOWN) {
+        while (state != NodeState.SHUT_DOWN && shuttingDown.get()) {
             try {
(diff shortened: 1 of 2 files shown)


metanet:
👍  once green

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -479,14 +479,17 @@ private void waitIfAlreadyShuttingDown() {
             return;
         }
         logger.info("Node is already shutting down... Waiting for shutdown process to complete...");
-        while (state != NodeState.SHUT_DOWN) {
+        while (state != NodeState.SHUT_DOWN && shuttingDown.get()) {
```
pveentjer: dumb question: I guess you need the 'shutting down' field because 'state' has no mode for 'shutting down'; only 'shutdown' right?

PR #8123, opened 2016-05-07
Multicast enabled by default in XML 

### LINKED ISSUE #945: hazelcast xml network tag
Hi,

This is a based on discussion here:
https://groups.google.com/forum/#!topic/hazelcast/4xgZNtZKWnw

Default configuration are different between XML config or programmatic config.
The thing is if I want to configure hazelcast (with xml) and just change the group section, i have to set enabled=true for multicast tag.

the XSD specify default="true" and Java specify DEFAULT_ENABLED = true.
But in XML: config.getNetworkConfig().getJoin().getMulticastConfig().setEnabled(false);

So should it be .setEnabled(true) ?

pveentjer:
Multicast was explicitly disabled while parsing.

This is a inconsistency with the xsd; which  states that multicast is enabled by default.
It also is an inconsistency with the programmatic api, where multicast is enabled by default.

This fix resolves these inconsistencies. 

However it could be that xml-users that only configure tcp and have not explicitly disabled multicast, since it was disabled by default, now get an error because multicast needs to be explicitly disabled (you can't have multiple join mechanisms active).

Fix #945

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/XmlConfigBuilder.java
@@ -193,9 +193,7 @@ public class XmlConfigBuilder extends AbstractConfigBuilder implements ConfigBui
 
     @Override
     public Config build() {
-        Config config = new Config();
-        config.getNetworkConfig().getJoin().getMulticastConfig().setEnabled(false);
-        return build(config);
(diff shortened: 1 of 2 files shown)


bilalyasar:
👍

jerrinot:
:+1: it's worth mentioning this in the release notes. CC @Serdaro

PR #8125, opened 2016-05-07
Reduces loglevel of member leaving to info

### LINKED ISSUE #4223: Don't log warnings when a node leaves a cluster gracefully
Log level should be INFO? See http://stackoverflow.com/questions/27240533/how-to-leave-a-hazelcast-cluster-gracefully/

pveentjer:
Logging of spinning and nonblocking handlers is also alligned

Fix #4223


jerrinot:
:+1:

PR #8126, opened 2016-05-08
[DONT MERGE] Initial work on deleting the delegating future.

pveentjer:
The underlying InvocationFuture should be used. There is no reason to wrap
the InvocationFuture and cause a new set of problem
* more litter is created than needed (e..g the wrapping future + the mutex object)
* more litter for 'andThen' since each callback is wrapped in the DelegatingExecutionCallback
* potential performance impact since an additional intrinsic lock is created and this can have performance impact due to lock deflation.
* and potential bugs: see #6325

Also there is cleanup in the map, so that every async version of a sync method,
will acts like the 'master' method and the sync version is nothing else than a
call to the async followed by a join.


pveentjer:
run-lab-run

PR #8130, opened 2016-05-09
Resolved races on setting max in LocalMapStateImpl

### LINKED ISSUE #8128: LocalMapStatsImpl is racy
This class is used by multiple threads, so a method call like this:

```
  public void incrementGets(long latency) {
        GET_COUNT.incrementAndGet(this);
        TOTAL_GET_LATENCIES.addAndGet(this, latency);
        MAX_GET_LATENCY.set(this, Math.max(maxGetLatency, latency));
    }
```

Is racy. If 2 calls come in at the same time with a latency higher than the maxGetLatency, it can be that the highest latency gets overwritten by a lower latency and therefor the highest measured latency can get lost.

pveentjer:
Fix #8128

The same problem was also fixed at LocalReplicatedMapStatsImpl

CacheStatisticsImpl got a cleanup by removing duplicate code for the 'max' functionality.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheStatisticsImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheStatisticsImpl.java
@@ -21,6 +21,8 @@ import com.hazelcast.monitor.NearCacheStats;
 
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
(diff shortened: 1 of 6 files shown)


vbekiaris:
👍 once green

hasancelik:
👍 once green

hasancelik:
run-lab-run

PR #8133, opened 2016-05-09
Cluster management REST URLs are changed

muratayan:
Since we check the URLs via `startsWith()`, it caused lots of trouble before.

@bilalyasar please check if this affects the CLI implementation.


kobalski:
👍

pveentjer:
Looks good. Dumb question; do we have any version compatibility guarantees between HZ and mancenter?

muratayan:
@pveentjer We don't have a guarantee but they are all compatible after `mancenter 3.3`.

pveentjer:
:+1:

PR #8136, opened 2016-05-10
Don't make backups if the EntryOperation was a noop

ChristerF:
Optimization of the EntryProcessor code - if the EP was a noop, don't even ask for an EntryBackupProcessor and don't generate an EntryBackupOperation.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
@@ -60,6 +60,9 @@ public class EntryOperation extends LockAwareOperation implements BackupAwareOpe
     private Object response;
     private transient Object dataValue;
 
+    // Determines whether we should generate a backup operation, by default we should generate a backup operation
+    private transient boolean shouldBackupOperation = true;
+
     public EntryOperation() {
     }
 
@@ -88,6 +91,7 @@ public class EntryOperation extends LockAwareOperation implements BackupAwareOpe
 
         // first call noOp, other if checks below depends on it.
         if (noOp(entry)) {
+            shouldBackupOperation = false;
             return;
         }
         if (entryRemoved(entry, now)) {
@@ -125,13 +129,17 @@ public class EntryOperation extends LockAwareOperation implements BackupAwareOpe
 
     @Override
     public Operation getBackupOperation() {
+
+        // If the EntryProcessor was a noop, then don't generate a backup operation
+        if (!shouldBackupOperation) return null;
+
         EntryBackupProcessor backupProcessor = entryProcessor.getBackupProcessor();
         return backupProcessor != null ? new EntryBackupOperation(name, dataKey, backupProcessor) : null;
     }
 
     @Override
     public boolean shouldBackup() {
-        return entryProcessor.getBackupProcessor() != null;
+        return shouldBackupOperation && entryProcessor.getBackupProcessor() != null;
     }
 
     @Override


ChristerF:
@pveentjer is this something for you to review?

noctarius:
An EntryProcessor could still have side effects on other elements in the same partition. The noop only checks the given entry instance, doesn't it?

ChristerF:
@noctarius like all Java code it could have side effects. This is about the contract for an EP. If the EP doesn't affect the entry there should be nothing to backup. A developer relying on an EntryBackupProcessor to be run on a backup node (side effects in EntryBackupProcessors is a pretty horrible design pattern). The current implementation relies on a fairly cumbersome need to specify whether a backup processor is available and requires unnecessary logic in each EP to track the result of the EP and whether that is a noop or not. This is unnecessary since whether the operation is a noop or not is tracked by the EntryOperation. With this change, you would know that if the EntryBackupProcessor actually gets invoked, the operation was indeed a mutating operation.

ChristerF:
@noctarius if you would be allowed to modify entries in the same partition Hazelcast would have to implement something like Partition Level Transactions in Coherence. And as far as I understand there is no such support nor is it encouraged to modify other entries in the same partition during an EntryProcessor.

noctarius:
Not sure about that. EP is explicitly designed to access other data in the same partition, this is the implicit   partition log.
I agree you should design your EP in a way that you should only make changes in the map you're executing against, anyhow this is not always true.
Anyhow I like the idea of preventing the BEP to actually run, I'm just raising concerns here that this PR will change the current behavior and customers might suffer - especially since this is a hard to find change - when upgrading.

noctarius:
Well it is not encouraged and there is no Partition-Level Transaction but I know I've seen code that does it anyways. If your application is able to correct missing data and bring it back to a consistent state you don't need those transactional behavior.

ChristerF:
We have code that does exactly what you are talking about, but any such code that requires a backup would always call setValue and thus the corresponding backup EP would run. So I don't think that would be an issue if it is truly a mutation.

The whole point about Partition Level Transactions is to support scalability and allow multiple threads to do complex operations against a Partition at the same time (against multiple maps) and then commit those to the primary and the backup as a unit. Since Hazelcast only supports one thread modifying or reading a partition at a time there is no need/use for Partition Level Transactions since there is only one thread working on a map at the same time. 

AFAIU Hazelcast would require a re-design of the Partition handling code to support Partition Level Transactions.

noctarius:
Well there's still another use case for Partition-Level Transactions which is failure. If anything throws an exception and you do multiple mutations, you might end up in a corrupted state - that is why it generally is not recommended to do it and yeah I agree it should basically never be used (beside this very case). Anyhow you might be right about the fact that you most probably will still have a mutation on the main map. I still think I've seen different:

Lock-Map (basically just the map containing the locks) but I think it also was used to run the EP
Different Data Maps, that actually were mutated

I'm still trying to come up with a better approach, unfortunately I don't have any idea yet. You would have to hook into the general system to figure out what happened on other data structures - not a nice thing either. The whole problem will be solved in the future with the PartitionProcessor but for now this is what we have and customers / users do work around the limitation one way or the other :)

Anyhow let's see if someone else has any idea or if they see any problem. I'm not the one to merge it anyways :)

ChristerF:
@noctarius btw, what would be the officially supported Hazelcast way to do what you just described? I.e having an EP modify other maps on the same partition?

noctarius:
Hopefully the PartitionProcessor in the near future. There's also a chance of deadlocking yourself using the above presented, the reason why we generally not recommend to do it. You have to pay lots of attention when doing it.

ChristerF:
The core problem I see with a PartitionProcessor is that it locks you into the notion that a single partition can only have a single thread operating on it. Given read-through, write-through and other similar concepts I think the scalability would be limited if you don't allow multiple threads to do operations simultaneously on a partition (as PLT would in Coherence and other datagrid implementations).

noctarius:
But that would require the probably most fundamental change in the Hazelcast architecture. Don't think this will happen. There are already a few operations that can run side-ways but those are non mutating operations. Supporting multiple writers on a partition will bring it lots of locking and therefore will limit scalability on most other situations.

ChristerF:
@noctarius a fundamental change that would enable a whole new level of functionality for Hazelcast. And as for locking required, yes, but it would enable scalability for many use cases where Hazelcast is a really poor choice today such as cases where read-through or write-through is important.

In terms of a business strategy, you would want to solve harder and harder problems rather than optimizing for the simplest.

bwzhang2011:
@noctarius, would you mind leaving more message for the design of PartitionProcessor especially for the control for multiple threads ? maybe @pveentjer or @mdogan should take a look at it.

noctarius:
Not in this issue, as we're going offtopic

bwzhang2011:
@noctarius, take it easy. as some hazelcast user, just care about any potential issue or detail from hazelcast itself whether it's confirmed or not so it is with the similar comparison with other data grid solutions and also redis (3.2 has been released).

jerrinot:
@ChristerF: As much as I like the idea I am also worried the change breaks existing real-world entry processor implementations. 

Just because EP does not mutate an entry it does not necessary mean it's safe not run it against a replica. Our API even encourages to have a different implementation to be executed on backups. 

Perhaps we should refine EP contract in one version and change it another. I'm moving this to a backlog.

ChristerF:
@jerrinot The current EP contract is simply too undefined to make sense. Which means that the only way of figuring out how it works is to step through the code. As I did that I saw that there was an obvious flaw in the implementation - if nothing has been changed, there shouldn't be a backup. That is a "common sense" interpretation of the EP/BackupEP interfaces which I find it hard to argue against. The current implementation *forces* a quite complicated and non-natural state tracking of whether the EP changed a value or not, and if it did, return a null BackupEP. All other EP implementations in other open source products I've come across wouldn't send a backup if the value was unchanged. I urge you to reconsider this one...

Btw, your JSR-107 implementation of EntryProcessor's is also similarly flawed - it always sends a backup record regardless if the record changed or not from the EP. Seems like a super simple optimization to do. And I would find it very hard to argue against that optimization. cc @gregrluck - you're mr JSR-107, what is your thought on that one?

gregrluck:
For JCache the spec is very clear on this. A Noop is a no-op so skipping the backup is Ok.

ChristerF:
@jerrinot was this the final verdict on this one?

gregrluck:
Christer

Can you create the exact contract you want simply spelled out. 

Regards

Greg Luck

web: http://gregluck.com <http://gregluck.com/>
skype: gregrluck
mobile US: +1 650 924 6244
mobile Australia: +61 408 061 622

Holmistr:
@gregrluck @jerrinot Can we get some decision here? Is there going to be any progress?

gregrluck:
Check with Jaromir but I recall that we thought we needed to move this 4.0 to change the contract.

ChristerF:
JCache EntryProcessor semantics should be quite clear on this one and AFAIU
it has the same lack of optimization.

Holmistr:
@jerrinot Any updates on this or are we waiting for HZ 4.0?

Holmistr:
I'm going to close this PR, since there's no activity on it.
@ChristerF  I'm sorry it got abandoned. Please, feel free to comment if you're willing to pick it up again. We appreciate your effort to contribute! :)

PR #8141, opened 2016-05-10
KillNode rest functionality made safe

muratayan:
- When we do `node.shutdown(false)`, socket which is currently used is closed as well. This was causing REST request not getting an appropriate answer but `java.net.ConnectException: Connection refused`.

- Also, since this is now a REST request, no need for the task polling part of it.


kobalski:
👍

pveentjer:
:+1:

hasancelik:
run-lab-run

PR #8143, opened 2016-05-11
Prioritize COPY and SHIFT UP migrations

metanet:
Prioritize a COPY / SHIFT UP migration against
- a non-conflicting MOVE migration on a hotter index,
- a non-conflicting SHIFT DOWN migration to a colder index.
Non-conflicting migrations have no common participant. Otherwise, order of the migrations should not be changed.

The main motivation of the prioritization is COPY / SHIFT UP migrations increase the available replica count of a partition while a MOVE migration doesn't have an effect on it.


hasancelik:
run-lab-run

mdogan:
👍

PR #8147, opened 2016-05-11
Disable stale read on partition migration

metanet:
Currently, we allow read operations to be performed while a partition is being migrated. It can lead to stale reads on the following scenario:

Lets say Node A is owner of a partition which will be migrated to Node B. When migration starts, mutating operations are blocked but read operations are allowed on Node A.
After Node B (migration destination) commits the migration, it can start performing mutations but Node A (migration source) continues to allow read operations on itself until it commits the migration.

With this flag (`hazelcast.partition.migration.stale.read.disabled`), one can disable stale read possibility on migrations. By default, it is set to false and current behavior is preserved.


mdogan:
👍 

Created ref-manual issue for the new flag: https://github.com/hazelcast/hazelcast-reference-manual/issues/159

jerrinot:
:+1:

PR #8151, opened 2016-05-11
Multiple IAtomicReference issues fixed

### LINKED ISSUE #8149: IAtomicReference#alter does not persist changes
For some reason when I just want to alter an AtomicReference<MyObject> like so:

```
  ref.alter(node -> {
            node.nextNodes[0] = lala;
            node.nextKeys[0] = "lala";
            return node;
        });
```

then nothing changes. A later get will still return the old object. But when i change this to alterAndGet

```
  ref.alterAndGet(node -> {
            node.nextNodes[0] = lala;
            node.nextKeys[0] = "lala";
            return node;
        });
```

then everything works as expected. But in fact I do not want so send the whole object over wire, I just want to modify a property.

jerrinot:
1. getAndAlter test was testing alterAndGet method
2. AlterAndGetOperation optimized - sen backup only when (serialized) new value is different from the original value
3. AlterOperation fixed - use serialized values to make a decision whether to persist change - otherwise a change is lost when the function mutates the original object instead of creating a backup
4. GetAndAlterOperation fixed - the same issues as with AlterOperation

Fixes #8149

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/atomicreference/operations/AlterAndGetOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/atomicreference/operations/AlterAndGetOperation.java
@@ -37,12 +37,16 @@ public class AlterAndGetOperation extends AbstractAlterOperation {
         IFunction f = nodeEngine.toObject(function);
         AtomicReferenceContainer atomicReferenceContainer = getReferenceContainer();
 
-        Object input = nodeEngine.toObject(atomicReferenceContainer.get());
+        Data originalData = atomicReferenceContainer.get();
+        Object input = nodeEngine.toObject(originalData);
         //noinspection unchecked
         Object output = f.apply(input);
-        shouldBackup = true;
-        backup = nodeEngine.toData(output);
-        atomicReferenceContainer.set(backup);
+        Data serializedOutput = nodeEngine.toData(output);
+        shouldBackup = !isEquals(originalData, serializedOutput);
+        if (shouldBackup) {
+            backup = serializedOutput;
+            atomicReferenceContainer.set(backup);
+        }
         response = output;
     }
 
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/concurrent/atomicreference/operations/AlterAndGetOperation.java
```diff
@@ -37,12 +37,16 @@ public void run() throws Exception {
...
-        Object input = nodeEngine.toObject(atomicReferenceContainer.get());
+        Data originalData = atomicReferenceContainer.get();
+        Object input = nodeEngine.toObject(originalData);
         //noinspection unchecked
         Object output = f.apply(input);
-        shouldBackup = true;
-        backup = nodeEngine.toData(output);
-        atomicReferenceContainer.set(backup);
+        Data serializedOutput = nodeEngine.toData(output);
+        shouldBackup = !isEquals(originalData, serializedOutput);
```
jerrinot: this is just an optimization - no need to send backup (and wait for it) if serialized values as the identical.

hazelcast/src/main/java/com/hazelcast/concurrent/atomicreference/operations/AlterAndGetOperation.java
```diff
@@ -37,12 +37,16 @@ public void run() throws Exception {
...
 
-        Object input = nodeEngine.toObject(atomicReferenceContainer.get());
+        Data originalData = atomicReferenceContainer.get();
+        Object input = nodeEngine.toObject(originalData);
         //noinspection unchecked
         Object output = f.apply(input);
-        shouldBackup = true;
-        backup = nodeEngine.toData(output);
-        atomicReferenceContainer.set(backup);
+        Data serializedOutput = nodeEngine.toData(output);
```
pveentjer: I would call serializedOutput somehting like outputData. So it has 'Data' in the name

pveentjer:
Minor review comment. If you can fix the name. Otherwise :+1:

pveentjer:
I'll merge it; you can fix the name with a direct push to prevent needing to wait for another build.

PR #8152, opened 2016-05-11
[BACKPORT] Multiple IAtomicReference issues fixed

### LINKED ISSUE #8149: IAtomicReference#alter does not persist changes

jerrinot:
1. getAndAlter test was testing alterAndGet method
2. AlterAndGetOperation optimized - sen backup only when (serialized) new value is different from the original value
3. AlterOperation fixed - use serialized values to make a decision whether to persist change - otherwise a change is lost when the function mutates the original object instead of creating a backup
4. GetAndAlterOperation fixed - the same issues as with AlterOperation

Fixes #8149, Backport of #8151
(cherry picked from commit 2ad2e7b)


pveentjer:
:+1:

PR #8161, opened 2016-05-13
Reduced binary size of map.get/contains/entryview operations

pveentjer:
Shaves 12 bytes of. TTL and data-value isn't needed. This is done by
making 2 class hierarchies.

1: ReadonlyKeyBasedMapOperation without ttl/data-value
2: MutatingKeyBasedMapOperation with ttl/data-value

No enterprise changes required


jerrinot:
:+1: LGTM

ahmetmircik:
👍

PR #8162, opened 2016-05-13
Shutdown nodes via LifecycleService on cluster shutdown

### LINKED ISSUE #8070: Cluster.shutdown() ignores lifecycle listeners
Expanding from discussion on https://groups.google.com/forum/#!topic/hazelcast/wbI8i2PlkVc

_ClusterServiceImpl_ has method shutdownNodes which calls **node.shutdown(false)**. The same applies for _ShutdownNodeOperation_.

Shouldn't it call _LifecycleService_ instead so the proper listeners are invoked before and after?

Cluster shutdown functionality seems really useful but in our case lot of stuff relies on the listeners for deinitialization.

metanet:
... so that lifecycle listeners are invoked properly on cluster shutdown

Fixes #8070


jerrinot:
:+1:

mdogan:
👍

PR #8163, opened 2016-05-13
Reduced binary Backup size by replica version compression

pveentjer:
By default a long array of 6 elements is send with every backup to store the replicaVersions. So you
get 6*8+4(length)=52 bytes for whole array.

However if there is only a single backup, only the fist version is set
and the other elements are zero. This optimization makes use of that by
only sending the minimum amount of versions.

Also the size of the array has been reduced from int to byte; so even if
all versions are set, you still get a 3 byte space reduction.

In a standard setup with a single backup this leads to a 43 byte size reduction.


pveentjer:
run-lab-run

jerrinot:
:+1: 
a minor thing: an inlined comment would be nice. or perhaps even it's own method.

pveentjer:
run-lab-run

tombujok:
👍

PR #8170, opened 2016-05-13
[DONT MERGE] caller uuid now UUID instead of String

pveentjer:
The Operation.callerUUID is now a UUID instead of String. It has the following advantages:
* it reduces the payload by 24 bytes. The utf encoding is very inefficient and the UUID can be encoded in 16 bytes.
* litter reduction: for a String you need the string object + char array, for a UUID you only need a UUD object. So that is 1 instance per invocation less.


jerrinot:
:+1: please don't forget to create a meaningful commit message.

pveentjer:
Closing this for the time being since we can get rid of the callerUuid for the time being.
https://github.com/hazelcast/hazelcast/pull/8173

PR #8173, opened 2016-05-13
Remove calleruuid from Backup

pveentjer:
This shaves 40 bytes from a BackupOperation and removed 2 object instances per invocation (String/char-array).

There is no point in setting the callerUuid since on the executing side of the Backup operation, the callerUuid will be determined (if missing) based on the caller address.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MutatingKeyBasedMapOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/MutatingKeyBasedMapOperation.java
@@ -16,7 +16,6 @@
 
 package com.hazelcast.map.impl.operation;
 
-import com.hazelcast.map.impl.MapService;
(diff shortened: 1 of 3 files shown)


pveentjer:
verify

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MutatingKeyBasedMapOperation.java
```diff
@@ -62,11 +61,6 @@ protected MutatingKeyBasedMapOperation(String name, Data dataKey, Data dataValue
         this.ttl = ttl;
     }
```
pveentjer: This method isn't needed since it is already available on the superclass.

pveentjer:
verify

vbekiaris:
👍

Donnerbart:
:+1:

PR #8176, opened 2016-05-13
Collection querying in portables - style & error handling improvements

tombujok:


hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DefaultPortableReader.java
```diff
@@ -684,8 +687,11 @@ private PortablePosition findPositionForReading(String path) throws IOException
         }
     }
 
-    @SuppressWarnings({"checkstyle:cyclomaticcomplexity", "checkstyle:returncount"})
+    @SuppressWarnings({"checkstyle:cyclomaticcomplexity", "checkstyle:returncount", "unchecked"})
     private <T> T readSinglePositionFromNonArray(PortablePosition position) throws IOException {
+        if (position.getType() == null) {
+            throw new IllegalArgumentException("Unsupported type read: null");
+        }
         switch (position.getType()) {
```
tombujok: Switch might have resulted with null pointer in unexpected cases.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DefaultPortableReader.java
```diff
@@ -644,8 +644,11 @@ private PortablePosition findPositionForReading(String path) throws IOException
         }
     }
 
-    @SuppressWarnings({"checkstyle:cyclomaticcomplexity", "checkstyle:returncount"})
+    @SuppressWarnings({"checkstyle:cyclomaticcomplexity", "checkstyle:returncount", "unchecked"})
     private <T> T readSinglePositionFromArray(PortablePosition position) throws IOException {
+        if (position.getType() == null) {
+            throw new IllegalArgumentException("Unsupported type read: null");
+        }
```
tombujok: Switch might have resulted with null pointer in unexpected cases.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DefaultPortableReader.java
```diff
@@ -756,7 +762,7 @@ private void validateNotMultiPosition(PortablePosition position) {
     private void validateType(PortablePosition position, FieldType expectedType) {
         FieldType returnedType = position.getType();
         if (position.getIndex() >= 0) {
-            returnedType = returnedType.getSingleType();
+            returnedType = returnedType != null ? returnedType.getSingleType() : null;
         }
```
tombujok: It was a possible null pointer in unexpected cases that should be guarded by the caller, but just in case.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/PortablePositionFactory.java
```diff
@@ -90,16 +89,12 @@ static PortableSinglePosition empty(boolean leaf, boolean any) {
...
 
-    private static void potentiallyNullify(PortableSinglePosition position) {
-        if (position.isEmpty() && !position.isLeaf()) {
-            position.nil = true;
-        } else if (position.isEmpty() && position.getIndex() >= 0) {
-            position.nil = true;
-        }
+    private static boolean isEmptyNil(PortableSinglePosition position) {
+        return position.isEmpty() && (!position.isLeaf() || position.getIndex() >= 0);
     }
```
tombujok: Eliminates potential branch miss - thx @mtopolnik 
Should give a couple of nanos :)

hazelcast/src/main/java/com/hazelcast/query/extractor/ValueReader.java
```diff
@@ -29,29 +29,29 @@
  * <p>
  * The wildcard quantifier may be used a couple of times, like person.leg[any].finger[any] which returns all fingers
  * from all legs.
- *
- * @param <T> Type of the value to read
  */
-public abstract class ValueReader<T> {
+public abstract class ValueReader {
```
tombujok: Didn't make sense to have it here, since a reader may be used to read various paths returning various types.

jerrinot:
+1

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DefaultPortableReader.java
mtopolnik: Why not an `assert`? I think they are appropriate everywhere except at the immediate point of contact with the user.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/PortablePositionFactory.java
mtopolnik: How come there are no caller-side diffs for this change?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DefaultPortableReader.java
tombujok: Yes, would be appropriate too. Don't know what our policy on asserts actually is...

mtopolnik: I already use them everywhere and had a discussion about them. There isn't a definite policy, so it's up to you. Although it may not seem like much difference, all the checks can eventually add up.

Donnerbart: FYI: `assert` seem to lower the branch coverage, if they are not tested (so they seem to behave like normal code regarding code coverage).

mtopolnik: We have explicit support for testing `assert`s: `@RequireAssertEnabled`. Needless to say, my test classes are full of them :)

Donnerbart: Yes, just saying: If you don't trigger the asserted lines, JaCoCo will count this as missing branch coverage. So it's just like a normal `if` which needs a test with your mentioned `@RequireAssertEnabled` annotation.

tombujok: OK, I changed it to assertions since if it's null there it means there's a problem with the code itself, not with the user's input.

tombujok:
verify

mtopolnik:
👍

tombujok:
verify

PR #8177, opened 2016-05-13
Backup target operation never converted to data

pveentjer:
The target operation in the Backup should never be converted to Data. Before this fix, only in case of a single backup the conversion was skipped, but in case of multiple Backups the the target operation is first converted to Data and then stuffed in the Backup.

This is not only wasteful on the sending side, also the receiving side needs to deserialize first to data, and then needs to deserialize to the actual object. 

This fix solves this by never converting to data. No more litter, less complexity, and one byte backup size reduction since a 'is data' flag doesn't need to be set anymore.

There can not be accidental sharing of the target operation because a Backup operation always needs to be send to another member and therefor it is serialized/deserialized so the target always is cloned.


pveentjer:
run-lab-run

PR #8187, opened 2016-05-16
Added more partition group strategy options.

kobalski:
Added `ZoneAwareMemberGroupFactory` and `SPIAwareMemberGroupFactory`.

- `ZoneAwareMemberGroupFactory`  is responsible for creation of partition groups based on metadata provided by Discovery SPI.

-  `SPIAwareMemberGroupFactory`. is responsible calling custom member group impl provided by
 `discoveryStrategy.getPartitionGroupStrategy().getMemberGroups()` method

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/partition/membergroup/SPIAwareMemberGroupFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/partition/membergroup/SPIAwareMemberGroupFactory.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.partition.membergroup;
+
+
+import com.hazelcast.config.InvalidConfigurationException;
+import com.hazelcast.config.properties.ValidationException;
+import com.hazelcast.core.Member;
+import com.hazelcast.internal.partition.impl.PartitionStateManager;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.spi.discovery.DiscoveryStrategy;
+import com.hazelcast.spi.discovery.DiscoveryStrategyFactory;
+import com.hazelcast.spi.discovery.impl.DefaultDiscoveryService;
+import com.hazelcast.spi.discovery.integration.DiscoveryService;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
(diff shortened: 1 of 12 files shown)


noctarius:
Looks good but I would most probably make sure to reuse the DiscoveryStrategy instances created by the DiscoveryService as some strategies might need the start method to be called (for initialization) and some also might not be able to be created twice. We probably have to make DiscoveryService able to return the already created DiscoveryService or collect the PartitionGroupStrategies internally and return those as a set (probably the better way).

kobalski:
Thanks for the feedback @noctarius. You are right we should use already loaded strategies. I am passing `Node.discoveryService` to `MemberGroupFactoryFactory`. DiscoveryService should be the only class responsible for lifecycle of discovery strategies.

kobalski:
verify

noctarius:
We probably also want to change DiscoveryService interface into an abstract class while the API is still beta, to extend it in the future without breaking possible other integrators implementations :)

kobalski:
I agree, but can we do it after code freeze period with another pr.

kobalski:
verify

kobalski:
verify

kobalski:
verify

kobalski:
run-lab-run

hasancelik:
verify

serkan-ozal:
good job 👍

hazelcast/src/main/java/com/hazelcast/partition/membergroup/SPIAwareMemberGroupFactory.java
```diff
@@ -0,0 +1,88 @@
...
+import com.hazelcast.spi.discovery.integration.DiscoveryService;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Set;
+
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
+
```
emrahkocaman: Missing javadoc

hazelcast/src/main/java/com/hazelcast/partition/membergroup/SPIAwareMemberGroupFactory.java
```diff
@@ -0,0 +1,88 @@
...
+public class SPIAwareMemberGroupFactory extends BackupSafeMemberGroupFactory implements MemberGroupFactory {
```
emrahkocaman: Missing javadoc

hazelcast/src/main/java/com/hazelcast/partition/membergroup/SPIAwareMemberGroupFactory.java
```diff
@@ -0,0 +1,88 @@
...
+    protected Set<MemberGroup> createInternalMemberGroups(Collection<? extends Member> allMembers) {
+        Set<MemberGroup> memberGroups = new HashSet<MemberGroup>();
+
+        for (Member member : allMembers) {
+            try {
+                if (member.localMember()) {
+                    DefaultDiscoveryService defaultDiscoveryService = (DefaultDiscoveryService) discoveryService;
+                    // If no discovery strategy is found fail-fast
+                    if (!defaultDiscoveryService.getDiscoveryStrategies().iterator().hasNext()) {
+                        throw new RuntimeException("Failed to configure discovery strategies");
```
emrahkocaman: I'd prefer a more explanatory message

emrahkocaman:
👍 (with only minor comments which can be done later)

kobalski:
verify

kobalski:
verify

kobalski:
verify

PR #8189, opened 2016-05-16
Makes use of total-backup-count(sync+async backups) instead of backup…

ahmetmircik:
…-count(since it is only sync backup count) to fire partition-lost-event

forward port of https://github.com/hazelcast/hazelcast/pull/8188


vbekiaris:
👍

ahmetmircik:
verify

PR #8194, opened 2016-05-17
Introduced Partition Iterator to Cache and Map

eminn:
This PR introduces capability of iterating entries of the single partition to the Cache and Map.

Motivation behind this change to provide ability to load map & cache entries to Apache Spark, partition by partition. 

The iterator will retrieve the keys in batch, and based on the configuration it will either pull values one by one for more recent values or pull the values in batch for better performance.

Client protocol counterpart: https://github.com/hazelcast/hazelcast-client-protocol/pull/7
Enterprise counterpart : https://github.com/hazelcast/hazelcast-enterprise/pull/901

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientClusterWideIterator.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientClusterWideIterator.java
@@ -17,18 +17,18 @@
 package com.hazelcast.client.cache.impl;
 
 import com.hazelcast.cache.impl.AbstractClusterWideIterator;
-import com.hazelcast.cache.impl.CacheKeyIteratorResult;
 import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.impl.protocol.codec.CacheIterateCodec;
+import com.hazelcast.client.impl.protocol.codec.CacheIterateEntriesCodec;
 import com.hazelcast.client.spi.ClientContext;
 import com.hazelcast.client.spi.impl.ClientInvocation;
 import com.hazelcast.client.spi.impl.ClientInvocationFuture;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.util.ExceptionUtil;
-
-import javax.cache.Cache;
 import java.util.Iterator;
+import java.util.List;
+import javax.cache.Cache;
 
 /**
  * Client side cluster-wide iterator for {@link com.hazelcast.cache.ICache}.
@@ -47,28 +47,54 @@ public class ClientClusterWideIterator<K, V>
     private ClientCacheProxy<K, V> cacheProxy;
     private ClientContext context;
 
-    public ClientClusterWideIterator(ClientCacheProxy<K, V> cacheProxy, ClientContext context) {
-        this(cacheProxy, context, DEFAULT_FETCH_SIZE);
+    public ClientClusterWideIterator(ClientCacheProxy<K, V> cacheProxy, ClientContext context, boolean prefetchValues) {
+        this(cacheProxy, context, DEFAULT_FETCH_SIZE, prefetchValues);
     }
 
-    public ClientClusterWideIterator(ClientCacheProxy<K, V> cacheProxy, ClientContext context, int fetchSize) {
-        super(cacheProxy, context.getPartitionService().getPartitionCount(), fetchSize);
+    public ClientClusterWideIterator(ClientCacheProxy<K, V> cacheProxy, ClientContext context,
+                                     int fetchSize, boolean prefetchValues) {
+        super(cacheProxy, context.getPartitionService().getPartitionCount(), fetchSize, prefetchValues);
         this.cacheProxy = cacheProxy;
         this.context = context;
         advance();
     }
 
-    protected CacheKeyIteratorResult fetch() {
-        ClientMessage request = CacheIterateCodec
-                .encodeRequest(cacheProxy.getPrefixedName(), partitionIndex, lastTableIndex, fetchSize);
+    public ClientClusterWideIterator(ClientCacheProxy<K, V> cacheProxy, ClientContext context, int fetchSize,
+                                     int partitionId, boolean prefetchValues) {
+        super(cacheProxy, context.getPartitionService().getPartitionCount(), fetchSize, prefetchValues);
+        this.cacheProxy = cacheProxy;
+        this.context = context;
+        this.partitionIndex = partitionId;
+        advance();
+    }
+
+    protected List fetch() {
         HazelcastClientInstanceImpl client = (HazelcastClientInstanceImpl) context.getHazelcastInstance();
-        try {
-            ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionIndex);
-            ClientInvocationFuture f = clientInvocation.invoke();
-            CacheIterateCodec.ResponseParameters responseParameters = CacheIterateCodec.decodeResponse(f.get());
-            return new CacheKeyIteratorResult(responseParameters.keys, responseParameters.tableIndex);
-        } catch (Exception e) {
-            throw ExceptionUtil.rethrow(e);
+        if (prefetchValues) {
+            ClientMessage request = CacheIterateEntriesCodec
+                    .encodeRequest(cacheProxy.getPrefixedName(), partitionIndex, lastTableIndex, fetchSize);
+            try {
+                ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionIndex);
+                ClientInvocationFuture f = clientInvocation.invoke();
+                CacheIterateEntriesCodec.ResponseParameters responseParameters = CacheIterateEntriesCodec.
+                        decodeResponse(f.get());
+                setLastTableIndex(responseParameters.entries, responseParameters.tableIndex);
+                return responseParameters.entries;
+            } catch (Exception e) {
+                throw ExceptionUtil.rethrow(e);
+            }
+        } else {
+            ClientMessage request = CacheIterateCodec
+                    .encodeRequest(cacheProxy.getPrefixedName(), partitionIndex, lastTableIndex, fetchSize);
+            try {
+                ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionIndex);
+                ClientInvocationFuture f = clientInvocation.invoke();
+                CacheIterateCodec.ResponseParameters responseParameters = CacheIterateCodec.decodeResponse(f.get());
+                setLastTableIndex(responseParameters.keys, responseParameters.tableIndex);
+                return responseParameters.keys;
+            } catch (Exception e) {
+                throw ExceptionUtil.rethrow(e);
+            }
         }
     }
 
--- a/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/ClientMapPartitionIterator.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/ClientMapPartitionIterator.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.map.impl;
+
+import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
+import com.hazelcast.client.impl.protocol.ClientMessage;
+import com.hazelcast.client.impl.protocol.codec.MapFetchEntriesCodec;
+import com.hazelcast.client.impl.protocol.codec.MapFetchKeysCodec;
+import com.hazelcast.client.proxy.ClientMapProxy;
+import com.hazelcast.client.spi.ClientContext;
+import com.hazelcast.client.spi.impl.ClientInvocation;
+import com.hazelcast.client.spi.impl.ClientInvocationFuture;
+import com.hazelcast.map.impl.iterator.AbstractMapPartitionIterator;
+import com.hazelcast.spi.serialization.SerializationService;
+import com.hazelcast.util.ExceptionUtil;
+import java.util.List;
+
+public class ClientMapPartitionIterator<K, V> extends AbstractMapPartitionIterator<K, V> {
+
+    private final ClientMapProxy<K, V> mapProxy;
+    private final ClientContext context;
+
+    public ClientMapPartitionIterator(ClientMapProxy<K, V> mapProxy, ClientContext context, int fetchSize,
+                                      int partitionId, boolean prefetchValues) {
+        super(mapProxy, fetchSize, partitionId, prefetchValues);
+        this.mapProxy = mapProxy;
+        this.context = context;
+    }
+
+    @Override
+    protected List fetch() {
+        HazelcastClientInstanceImpl client = (HazelcastClientInstanceImpl) context.getHazelcastInstance();
+        if (prefetchValues) {
+            ClientMessage request = MapFetchEntriesCodec.encodeRequest(mapProxy.getName(), partitionId, lastTableIndex,
+                    fetchSize);
+            ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionId);
+            try {
+                ClientInvocationFuture f = clientInvocation.invoke();
+                MapFetchEntriesCodec.ResponseParameters responseParameters = MapFetchEntriesCodec.decodeResponse(f.get());
+                setLastTableIndex(responseParameters.entries, responseParameters.tableIndex);
+                return responseParameters.entries;
+            } catch (Exception e) {
+                throw ExceptionUtil.rethrow(e);
+            }
+        } else {
+            ClientMessage request = MapFetchKeysCodec.encodeRequest(mapProxy.getName(), partitionId,
+                    lastTableIndex, fetchSize);
+            ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionId);
+            try {
+                ClientInvocationFuture f = clientInvocation.invoke();
+                MapFetchKeysCodec.ResponseParameters responseParameters = MapFetchKeysCodec.decodeResponse(f.get());
+                setLastTableIndex(responseParameters.keys, responseParameters.tableIndex);
+                return responseParameters.keys;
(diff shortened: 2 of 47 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/iterator/MapKeyIteratorResult.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.nio.serialization.DataSerializable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MapKeyIteratorResult implements DataSerializable {
```
ihsandemir: Should this extend IdentifiedDataSerializable? Non-java languages does not support DataSerializable.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/ClientMapPartitionIterator.java
```diff
@@ -0,0 +1,65 @@
...
+            ClientInvocationFuture f = clientInvocation.invoke();
+            MapIterateCodec.ResponseParameters responseParameters = MapIterateCodec.decodeResponse(f.get());
+            return new MapKeyIteratorResult(responseParameters.keys, responseParameters.tableIndex);
+        } catch (Exception e) {
+            throw ExceptionUtil.rethrow(e);
+        }
+    }
+
+    @Override
+    protected Data toData(Object obj) {
```
ihsandemir: Do we need toData toObject in this class?

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractClusterWideIterator.java
```diff
@@ -83,7 +83,7 @@
     protected final int partitionCount;
     protected int partitionIndex = -1;
 
-    protected int lastTableIndex;
+    protected int lastTableIndex = Integer.MAX_VALUE;
```
ahmetmircik: can you document the reason of this assignment `lastTableIndex = Integer.MAX_VALUE`?

hazelcast/src/main/java/com/hazelcast/map/impl/iterator/AbstractMapPartitionIterator.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.NoSuchElementException;
+
+public abstract class AbstractMapPartitionIterator<K, V> implements Iterator<Map.Entry<K, V>> {
+
+    protected IMap<K, V> map;
+    protected final int fetchSize;
+    protected final int partitionId;
+
+
+    protected int lastTableIndex = Integer.MAX_VALUE;
```
ahmetmircik: can add javaDoc for: what is table and table-index?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapIterateMessageTask.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Collections;
+
+public class MapIterateMessageTask extends AbstractMapPartitionMessageTask<MapIterateCodec.RequestParameters> {
+    public MapIterateMessageTask(ClientMessage clientMessage, Node node, Connection connection) {
+        super(clientMessage, node, connection);
+    }
+
+    @Override
+    protected Operation prepareOperation() {
+        System.out.println("MapIterateMessageTask.prepareOperation");
```
ahmetmircik: system out?

hazelcast/src/main/java/com/hazelcast/map/impl/iterator/AbstractMapPartitionIterator.java
```diff
@@ -0,0 +1,93 @@
...
+    public boolean hasNext() {
+        return (result != null && index < result.getCount()) || advance();
+    }
+
+    public Map.Entry<K, V> next() {
+        while (hasNext()) {
+            currentIndex = index;
+            index++;
+            final Data keyData = result.getKey(currentIndex);
+            final K key = toObject(keyData);
```
ahmetmircik: maybe you can use a lazy-map-entry to defer de-serialization to access time

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/StorageImpl.java
```diff
@@ -148,4 +150,11 @@ public void disposeDeferredBlocks() {
         return records.getRandomSamples(sampleCount);
     }
 
+    @Override
+    public MapKeyIteratorResult fetchNext(int tableIndex, int size) {
+        List<Data> keys = new ArrayList<Data>();
```
ahmetmircik: if possible an initial capacity can be added

ahmetmircik:
>  The iterator will retrieve the keys in batch, and in for each next() call it will get() the corresponding value for that key

I don't know the common-case but i assume entry-iterator and value-iterator can be useful also.

tombujok:
I saw a `wip` commit so I added a "do not merge" prefix so that it's not merged by a sb else (happened to me already:) )

eminn:
@ahmetmircik currently it will pull the keys one by one for getting most up to date versions of the keys, for the next versions I plan to add configuration option to pull all entires in batch for better performance

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
```diff
@@ -62,4 +63,7 @@
      * @return sampled entries.
      */
     Iterable<LazyEntryViewFromRecord> getRandomSamples(int sampleCount);
+
+    MapKeyIteratorResult fetchNext(int tableIndex, int size);
+
```
tombujok: To me `fetch` would be a better name. I know that it returns the next `tableIndex` but it doesn't fetch from the nextIndex, but the given one, so it's a bit misleading to me.

tombujok: This is a low level operation on the `Storage`. Using the name Iterator here is a bit confusing since it is neither an Iterator nor it doesn't know anything about the partitionIteration. Maybe just `MapKeys` or `KeySet` would sound better?

hazelcast/src/main/java/com/hazelcast/map/impl/iterator/MapKeyIteratorResult.java
```diff
@@ -0,0 +1,88 @@
...
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MapKeyIteratorResult implements IdentifiedDataSerializable {
+
+    private int tableIndex;
```
tombujok: Maybe `nextTableIndex` or `nextTableIndexToReadFrom` would be more clear?

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/RecordStore.java
```diff
@@ -188,6 +189,9 @@
      */
     Iterator<Record> iterator(long now, boolean backup);
 
+
+    MapKeyIteratorResult iterate(int tableIndex, int size);
+
```
tombujok: Hmm, it's not really doing an iteration. I mean like a standard  `iterator` that people are used to when they see the word `iterator`.
It's just fetching keys and returning the next table index to read from.
The returning of the nextTableIndex is like  side-effect to me so IMHO I would not concentrate on that while naming this method. Maybe just `fetchKeys` or sth along these lines?

hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
```diff
@@ -52,8 +53,9 @@
     public static final int QUERY_RESULT = 10;
     public static final int EVICT_BACKUP = 11;
     public static final int CONTAINS_KEY = 12;
+    public static final int KEY_ITERATOR = 13;
 
-    private static final int LEN = CONTAINS_KEY + 1;
+    private static final int LEN = KEY_ITERATOR + 1;
```
tombujok: I guess just `14` would be more readable in this case since there's already 13 lines with these numbers.
I was thinking for a while while it's a sum of these two ;)

tombujok: I'm missing some javadocs on the key classes and their key methods (but I guess it's still work in progress):
- AbstractMapPartitionIterator
- RecordStore
- CachePartitionIterator
- etc.

tombujok:
Looks good overall - my comments are mainly related to naming and lack of comments. Good job Emin!

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
eminn: You are right naming is not clear. how about `fetchKeysFromIndex()` and `MapPartialFetchResult` ?

hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
eminn: It's just a convention which was used throughout the serializer hooks, I don't want to be the outlier in this case :) 

For JavaDocs, there will be a followup PR for this change which includes some more edge case tests. I plan to add JavaDocs on that PR.

eminn:
@ahmetmircik @tombujok Guys could you have another look at the PR ?. I've added aforementioned batching of the values to the PR.

eminn:
run-lab-run

eminn:
verify

eminn:
verify

eminn:
verify

eminn:
run-lab-run

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheDataSerializerHook.java
```diff
@@ -324,6 +327,16 @@ public IdentifiedDataSerializable createNew(Integer arg) {
...
             }
         };
+        constructors[ENTRY_ITERATOR] = new ConstructorFunction<Integer, IdentifiedDataSerializable>() {
+            public IdentifiedDataSerializable createNew(Integer arg) {
+                return new CacheEntryIteratorOperation();
+            }
+        };
+        constructors[ENTRY_ITERATION_RESULT] = new ConstructorFunction<Integer, IdentifiedDataSerializable>() {
+            public IdentifiedDataSerializable createNew(Integer arg) {
+                return new CacheEntryIteratorResult();
```
ahmetmircik: Names can be consistent `iterator` or `iteration`:  `ENTRY_ITERATION_RESULT` vs `CacheEntryIteratorResult `?

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientClusterWideIterator.java
```diff
@@ -47,28 +47,54 @@
...
-            return new CacheKeyIteratorResult(responseParameters.keys, responseParameters.tableIndex);
-        } catch (Exception e) {
-            throw ExceptionUtil.rethrow(e);
+        if (prefetchValues) {
+            ClientMessage request = CacheIterateEntriesCodec
+                    .encodeRequest(cacheProxy.getPrefixedName(), partitionIndex, lastTableIndex, fetchSize);
+            try {
+                ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionIndex);
+                ClientInvocationFuture f = clientInvocation.invoke();
+                CacheIterateEntriesCodec.ResponseParameters responseParameters = CacheIterateEntriesCodec.
```
ahmetmircik: `CacheEntryIteratorCodec` ?

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientClusterWideIterator.java
```diff
@@ -47,28 +47,54 @@
...
+            } catch (Exception e) {
+                throw ExceptionUtil.rethrow(e);
+            }
+        } else {
+            ClientMessage request = CacheIterateCodec
+                    .encodeRequest(cacheProxy.getPrefixedName(), partitionIndex, lastTableIndex, fetchSize);
+            try {
+                ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionIndex);
+                ClientInvocationFuture f = clientInvocation.invoke();
+                CacheIterateCodec.ResponseParameters responseParameters = CacheIterateCodec.decodeResponse(f.get());
```
ahmetmircik: `CacheKeyIteratorCodec` ?

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheEntryIteratorResult.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public void readData(ObjectDataInput in)
+            throws IOException {
+        tableIndex = in.readInt();
+        int size = in.readInt();
+        entries = new ArrayList<Map.Entry<Data, Data>>(size);
+        for (int i = 0; i < size; i++) {
+            Data key = in.readData();
+            Data value = in.readData();
+            entries.add(new AbstractMap.SimpleEntry<Data, Data>(key, value));
```
ahmetmircik: do we really need a new meta-object `new AbstractMap.SimpleEntry`? Or a key value list also work?

hazelcast/src/main/java/com/hazelcast/map/impl/iterator/MapEntriesWithCursor.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public void readData(ObjectDataInput in) throws IOException {
+        nextTableIndexToReadFrom = in.readInt();
+        int size = in.readInt();
+        entries = new ArrayList<Map.Entry<Data, Data>>(size);
+        for (int i = 0; i < size; i++) {
+            Data key = in.readData();
+            Data value = in.readData();
+            entries.add(new AbstractMap.SimpleEntry<Data, Data>(key, value));
```
ahmetmircik: same meta-object?

ahmetmircik:
👍 with minor comments

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientClusterWideIterator.java
eminn: It was an old client protocol message which we cannot rename.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientClusterWideIterator.java
eminn: The new version of protocol published as is, so we cannot rename it.

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheEntryIteratorResult.java
eminn: Even if we avoid creating this object in de-serialization phase we are using it as entry list in the code base and in the client response messages. So I prefer not to change this since we just delay the object creation and make code complex than it should be.

pveentjer:
How are you going to deal with a partition migration during iteration? So imagine you have retrieved a batch, a partition migration happens and you want to get the next batch?

checkstyle/suppressions.xml
```diff
@@ -39,9 +39,12 @@
     <!-- Cache -->
     <suppress checks="MethodCount" files="com/hazelcast/config/CacheConfig"/>
     <suppress checks="MethodCount" files="com/hazelcast/cache/ICache"/>
+    <suppress checks="MethodCount" files="com/hazelcast/cache/impl/ICacheRecordStore"/>
```
pveentjer: These changes need to be added to the class itself. We are trying to move away from the supressions.xml

eminn:
Every batch read operation will be executed on the partition owner, so when you read first batch in partition X at location Y, then migration happens, you want to read second batch in partition X at location Z. Since we are making invocations based on the partition id I don't think we need to do anything extra to handle migration. Because all the data will be copied to the new partition owner as is.

checkstyle/suppressions.xml
eminn: I see, I will move them to the appropriate class files.

pveentjer:
But how are you going to continue where you left? So imagine you have a IMap with 100 items in a partitions, you have retrieved 50 items in your first batch, then the migration happens and the map partition has moved to a different box, and then you want to retrieve the second batch of 50 items. How are you going to determine where you continue?

eminn:
Every batch read operation will carry the latest table index of the record store.

pveentjer:
I just had a quick peek and I don't see index code in the RecordStore. So can you point me out where we have such an index?

eminn:
It is the table index of the underlying single bucket `ConcurrentReferenceHashMap`.

pveentjer:
Interesting. Are these guaranteed to be in sync? So when all items have moved, will the table index be the same again?

eminn:
That was a concern @serkan-ozal raised, we are investigating this since some of the cache related record classes might not have `hashCode()` methods overridden. So I assumed it should be, but now, we are trying to verify that.

pveentjer:
Another consistency related question: what if directly after completing the partition migration, a different thread does some modifications, inserts, deletes, updates and then the second batch is retrieved. How does it know where to continue? How to prevent e.g omissions or duplications? Could it be that you miss items or that you get the same item multiple times?

pveentjer:
PS: If this 'index' would really work it could solve a bunch of problems, e.g. map entry processors.

eminn:
You can miss items, but duplications cannot happen. Since we are always reading from higher to lower indexes. Imagine in between two iterations, an item is added and that caused a rehashing. When rehashing happens, an item either stay at it's position, or move up power of two offset. So if we read index `X`, in the next iteration we will read index `X-1`. If rehashing happened, item at index `X` either stayed at it's position `X`or moved to index `X + expansion size`.

hazelcast/src/main/java/com/hazelcast/util/SampleableConcurrentHashMap.java
```diff
@@ -94,6 +95,43 @@ public int fetch(int tableIndex, int size, List<Data> keys) {
...
+    /**
+     * Fetches entries from given <code>tableIndex</code> as <code>size</code>
+     * and puts them into <code>entries</code> list.
+     *
+     * @param tableIndex           Index (checkpoint) for starting point of fetch operation
+     * @param size                 Count of how many entries will be fetched
+     * @param entries              List that fetched entries will be put into
+     * @return the next index (checkpoint) for later fetches
+     */
+    public int fetchEntries(int tableIndex, int size, List<Map.Entry<Data, V>> entries) {
```
serkan-ozal: Seems tha we don't need `Data` type as type of the key. So I think we can use `List<Map.Entry<K, V>> entries` parameter for both of the `fetchEntries` and `fetchKeys` methods.

eminn:
run-lab-run

serkan-ozal:
:+1:

PR #8198, opened 2016-05-17
Remove destroyed cache proxy from owner `CacheManager` on destroy

### LINKED ISSUE #8186: `ICache::destroy` should remove cache itself from the owner `CacheManager`
... otherwise it causes memory leaks because of zombie cache proxies in the `AbstractHazelcastCacheManager::caches`

serkan-ozal:
Fixes #8186

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxy.java
@@ -98,14 +98,11 @@ public abstract class ClientProxy implements DistributedObject {
             context.removeProxy(this);
             try {
                 new ClientInvocation(getClient(), clientMessage).invoke().get();
+                postDestroy();
             } catch (Exception e) {
                 throw ExceptionUtil.rethrow(e);
             }
         }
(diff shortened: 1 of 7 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxy.java
```diff
@@ -98,16 +98,13 @@ public final void destroy() {
             context.removeProxy(this);
             try {
                 new ClientInvocation(getClient(), clientMessage).invoke().get();
+                postDestroy();
```
ahmetmircik: can it be more appropriate to put `postDestroy()` in a `finally` block?

serkan-ozal: I have just wanted it called only when invocation is successful. We may introduce another hook method for error case like `onDestroyError(Exception e)`

emrahkocaman:
👍

ahmetmircik:
👍

serkan-ozal:
verify

Issue #1194, opened 2013-11-22
Optimization: Prevent unneeded cloning


pveentjer:
When an operation is local and data stored in e.g. a map is immutable, then there is no need for serialization/deserialization.

If e.g. a map has OBJECT in-memory-format, and you do a local get, then the actual value can be returned. There is no need to first serialize the object and then deserialize it when the caller receives it. The actual stored object can be returned.

So this optimization should work for types like String, Integer etc. So we need to have some registration option where we can say: it is fine to return the local reference.

bwzhang2011:
@pveentjer, how does this going on ?  any comparison with hz3.3.x and hz3.4.x for such kind of optimization

pveentjer:
Perhaps that in HZ 3.5 we'll add the caller runs optimization as an experimental scheduler. I have it in a separate branch, but there are some things that need to be done.

https://github.com/hazelcast/hazelcast/pull/4513

I don't think we have time to fix the unwanted cloning in 3.5, I hope we can deal with it in 3.6. Because of the local call optimization, things like unwanted serialization are going to have a bigger impact on the performance.

Dretch:
This would be useful for me. I have some tasks running on Hazelcast executor services that I believe would be many times faster with this change.

pveentjer:
Are these task running locally or remote?

Dretch:
@pveentjer 

The tasks are sent to a Hazelcast node with `hazelcastInstance.getExecutorService("default").submitToKeyOwner(...)` and they access data that is local to the Hazelcast node (there is actually only one node in our deployment).

yacota:
I like the idea of this issue but what about using  javax.annotation.concurrent.Immutable to mark immutable java objects and let Hazelcast optimize as much as it can

pveentjer:
I think such implementation details would not be an issue to add. The main thing is the get the new OperationExecutor so this is actually going to matter.

dsukhoroslov:
Hi @pveentjer,

what the status of this issue now, when can we expect it be implemented?

Thanks, Denis.

bokken:
[...]

PR #8203, opened 2016-05-18
[DONT MERGE] Performance: Caller Runs Optimization

pveentjer:
Instead of offloading partition specific operations to partition thread, try to do it yourself.

Very naive benchmark where a single thread calls IAtomicLong.get. 

disabled:
Throughput: 102.234 ops/second
enabled:
Throughput: 1.484.428 op/second

The smaller the cluster, the more impact this optimization will have. Once this one is done, we need to crack  https://github.com/hazelcast/hazelcast/issues/1194 for cheap local access in IMap.

This is an older much more complex approach:
hazelcast/pull/4513


pveentjer:
The same implementation last year did 17.6M ops/second. After reverting simulator to that version, the new implementation using the same test/environment does:

INFO  09:34:22 Performance: 28,846,455.56 ops/s

pveentjer:
Implementation of last year did 17.6M ops/second. Current implementation is at 50M op/second. I hope there will not be too much regression when it is finished.

bwzhang2011:
@pveentjer, any update with such issue ? could be merged into 3.8 branch ?

marsbible:
any update with such issue ? could be merged into 3.8 branch ?

mmedenjak:
I'm going to put this in 3.10 so we merge/close it then

Holmistr:

PR #8206, opened 2016-05-18
Interpret duration time `0` as eternal for cache entry expiration

### LINKED ISSUE #8148: Cache should not expire entites when Duration value is 0 (zero)
According to the JCache API javadoc, when the durationAmount parameter in ` javax.cache.expiry.Duration` constructor is set to 0 (zero), it means cache entites should live eternal. Here is the javadoc:

```
  /**
   * Constructs a duration.
   *
   * @param timeUnit       the unit of time to specify time in. The minimum time unit is milliseconds.
   * @param durationAmount how long, in the specified units, the cache entries should live. 0 means eternal.
   * @throws NullPointerException     if timeUnit is null
   * @throws IllegalArgumentException if durationAmount is less than 0 or a TimeUnit less than milliseconds is specified
   */
  public Duration(TimeUnit timeUnit, long durationAmount) {
```

However the entity directly expires when it is set to zero. See the test below. It fails when executed:

```
    @Test
    public void valueShouldNotBeExpiredWhenDurationIsZero() {
        final int CREATED_EXPIRY_TIME_IN_MSEC = 0;

        Duration duration = new Duration(TimeUnit.MILLISECONDS, CREATED_EXPIRY_TIME_IN_MSEC);
        CacheConfig<Integer, String> cacheConfig = new CacheConfig<Integer, String>();
        cacheConfig.setExpiryPolicyFactory(CreatedExpiryPolicy.factoryOf(duration));

        final Cache<Integer, String> cache = createCache(cacheConfig);
        cache.put(1, "value");

            @Override
                assertNotNull(cache.get(1));
            }
        });
    }
```

serkan-ozal:
Fixes #8148

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
@@ -153,11 +153,9 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
         evictionStrategy = createEvictionStrategy(evictionConfig);
 
         injectDependencies(evictionPolicyEvaluator.getEvictionPolicyComparator());
-
         registerResourceIfItIsClosable(cacheWriter);
         registerResourceIfItIsClosable(cacheLoader);
         registerResourceIfItIsClosable(defaultExpiryPolicy);
-
         init();
     }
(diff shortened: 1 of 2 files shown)


jerrinot:
:+1: it's probably worth mentioning this in the release notes. cc @Serdaro

emrahkocaman:
👍

PR #8208, opened 2016-05-18
Revert changes in HazelcastInstanceFactory

jerrinot:
HazelcastInstanceFactory is an internal API, but we discovered external projects do use it directly anyway.
We recently did two changes:

Renamed HazelcastInstanceFactory into HazelcastInstanceManager (addf200)
Introduced a new abstraction for HazelcastInstanceFactory (2b5ba5a) and dynamic registration via META-INF
Both changes might break external projects.

The 1st change is easy to revert as it's just naming. Manager actually describes the responsibility better,
but API compatibility has priority. We could introduce a compatibility layer, but it's not worth the extra complexity.

The 2nd change - dynamic loading of HazelcastInstanceFactory was introduce due JET integration,
but they do not use it anymore and the abstraction servers no purposes -> we can remove it too.


jerrinot:
this is a replacement for #7667
Github UI shows as the history was lost, but it's actually retained and visible in both `git log` and Intellij IDEA git window.

pveentjer:
👍

pveentjer:
run-lab-run

PR #8220, opened 2016-05-20
adds reserved serialization numbers

mustafaiman:


hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationConstants.java
```diff
@@ -90,6 +90,14 @@
     public static final int JAVA_DEFAULT_TYPE_EXTERNALIZABLE = -101;
 
     // ------------------------------------------------------------
+    // LANGUAGE SPECIFIC SERIALIZERS
+    // USED BY CLIENTS (Not deserialized by server)
+
+    public static final int CSHARP_CLR_SERIALIZATION_TYPE = -110;
```
pveentjer: Do they need to increase by 10?

mustafaiman: I do not know if they needed to increase by 10. However these numbers are already in use. We put them here to prevent somebody picking up these numbers for a different purpose in future.

cangencer: I think I picked -110 initially for C# as -100,-101 were both java specific, so there is some room to add more java specific serialization methods if necessary.  But as @mustafaiman  said they're already in use so any change would break compatibility. Also changing them here would not have any effect anyway.

asimarslan:
👍

cangencer:
👍

PR #8239, opened 2016-05-24
Do not perform migration flow tasks after shut down

metanet:
If a shutting-down node becomes master after it has been notified to complete its shutdown by the old master, it should not perform any migration flow tasks if it has reset its state.

pair-investigated with @mdogan


mdogan:
👍

mdogan:
This is a very minor change, doesn't effect correctness of the system.

PR #8241, opened 2016-05-24
Include member set in MemberAttributeEvent

### LINKED ISSUE #7429: Potential NPE in MembershipEvent.getMembers()
`MemberAttributeEvent` extends `MembershipEvent` and calls its constructor using `null` for `Set<Member> members`, which means that `MembershipEvent.getMembers()` will return `null`. 
If intentional, it's not documented that `getMembers()` can return `null`.

jerrinot:
MemberAttributeEvent extends MembershipEvent hence it should obey its contract.
Fixes #7429

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientMembershipListener.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientMembershipListener.java
@@ -43,6 +43,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
 import static com.hazelcast.spi.exception.TargetDisconnectedException.newTargetDisconnectedExceptionCausedByMemberLeftEvent;
+import static java.util.Collections.unmodifiableSet;
 
 class ClientMembershipListener extends ClientAddMembershipListenerCodec.AbstractEventHandler
         implements EventHandler<ClientMessage> {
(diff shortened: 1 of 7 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientMembershipListener.java
```diff
@@ -112,13 +113,14 @@ public void handle(Collection<Member> initialMembers) {
...
-        for (Member target : members) {
+        Collection<Member> memberList = clusterService.getMemberList();
+        for (Member target : memberList) {
             if (target.getUuid().equals(uuid)) {
                 final MemberAttributeOperationType operationType = MemberAttributeOperationType.getValue(opType);
                 ((AbstractMember) target).updateAttribute(operationType, key, value);
                 MemberAttributeEvent memberAttributeEvent =
-                        new MemberAttributeEvent(client.getCluster(), target, operationType, key, value);
+                        new MemberAttributeEvent(client.getCluster(), target, operationType, key, value,
+                                unmodifiableSet(members));
```
jerrinot: I'm not too sure about a threading model here. However I noticed other events are referring to `members` without any synchronization -> I assume it's ok here too.

jerrinot:
I'm closing this one as it's problematic due client compatibility guarantees.

PR #8202, opened 2016-05-18
Added missing Spring config elements

dsukhoroslov:
fixes for issues #6567, #7363. created instead of PR #8095

PR #8242, opened 2016-05-24
Add `serviceImpl` getter/setter to `ServiceConfig` for backward compatibility

emrahkocaman:
Adds `serviceImpl` getter/setters as deprecated to keep backward compatibility after https://github.com/hazelcast/hazelcast/pull/8202/


jerrinot:
:+1: (a small thing: the new methods should also have `@since` annotation)

emrahkocaman:
run-lab-run

PR #8251, opened 2016-05-25
Fail-fast @RequireAssertEnabled

mtopolnik:
A `@RequireAssertEnabled` test now fails if assert not enabled. Before it just quietly cried in a corner.


mtopolnik:
run-lab-run

Donnerbart:
:+1: This will cause some build failures until we fixed the missing `-ea` in all our builds (PR builder, nightly, Sonar, EE), but this needs to be fixed.

PR #8255, opened 2016-05-26
Fix exception during shutdown when a service depends on LockService

jerrinot:
The  `getSharedService()` is deprecated, however the suggested
alternative (`getService()`) throws an Exception when a service is not found.
This is problematic during shutdown() when other Services are accessing
LockService for clean-up.

Eventually we should clarify life-cycle of services and describe how to deal
with inter-services dependencies.

I was not able to write a proper tests because the exception are caught and
logged. Here is a reproducer (it logs an error without this change on shutdown):

```java
    @Test
    public void testMM() {
        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(1);
        HazelcastInstance instance1 = factory.newHazelcastInstance();

        MultiMap<Object, Object> mm1 = instance1.getMultiMap(randomMapName());
        mm1.put("key", "value");
    }
```

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/servicemanager/impl/ServiceManagerImplTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/servicemanager/impl/ServiceManagerImplTest.java
@@ -46,8 +46,7 @@ public class ServiceManagerImplTest extends HazelcastTestSupport {
 
         HazelcastInstance hz = createHazelcastInstance(config);
         NodeEngineImpl nodeEngine = getNodeEngineImpl(hz);
-        serviceManager = new ServiceManagerImpl(nodeEngine);
-        serviceManager.start();
+        serviceManager = (ServiceManagerImpl) nodeEngine.getServiceManager();
     }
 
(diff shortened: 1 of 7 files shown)


hazelcast/src/test/java/com/hazelcast/spi/impl/servicemanager/impl/ServiceManagerImplTest.java
```diff
@@ -46,8 +46,7 @@ public void setup() {
 
         HazelcastInstance hz = createHazelcastInstance(config);
         NodeEngineImpl nodeEngine = getNodeEngineImpl(hz);
-        serviceManager = new ServiceManagerImpl(nodeEngine);
-        serviceManager.start();
+        serviceManager = (ServiceManagerImpl) nodeEngine.getServiceManager();
```
jerrinot: this is a boy-scout change - it fixes another exception being logged during a test. it's in its own commit, see the commit message for more details.

vbekiaris:
run-lab-run

vbekiaris:
👍

pveentjer:
:+1:

PR #8256, opened 2016-05-26
Only update executor stats when statistics are enabled in ExecutorConfig

### LINKED ISSUE #8223: DistributedExecutorService does not take into account ExecutorConfig.isStatisticsEnabled
Statistics are always gathered but ManCenter doesn't show them if you disable statistics.

vbekiaris:
Fixes #8223

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
+++ b/hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
@@ -55,6 +55,9 @@ public class DistributedExecutorService implements ManagedService, RemoteService
     private static final AtomicReferenceFieldUpdater<CallableProcessor, Boolean> RESPONSE_FLAG =
             AtomicReferenceFieldUpdater.newUpdater(CallableProcessor.class, Boolean.class, "responseFlag");
 
+    // package-local access to allow test to inspect the map's values
+    final ConcurrentMap<String, ExecutorConfig> executorConfigCache = new ConcurrentHashMap<String, ExecutorConfig>();
+
     private NodeEngine nodeEngine;
     private ExecutionService executionService;
     private final ConcurrentMap<String, CallableProcessor> submittedTasks
@@ -69,6 +72,7 @@ public class DistributedExecutorService implements ManagedService, RemoteService
             return new LocalExecutorStatsImpl();
         }
     };
+
     private ILogger logger;
 
     @Override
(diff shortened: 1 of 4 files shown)


jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
```diff
@@ -90,16 +91,21 @@ public void shutdown(boolean terminate) {
     }
 
     public void execute(String name, String uuid, Callable callable, Operation op) {
-        startPending(name);
-        CallableProcessor processor = new CallableProcessor(name, uuid, callable, op);
+        ExecutorConfig cfg = nodeEngine.getConfig().findExecutorConfig(name);
```
gurbuzali: the problem with `findExecutorConfig` is that it can be very costly. We wouldn't want to call it for each execution. So I suggest a map for caching the found config

hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
```diff
@@ -91,7 +95,7 @@ public void shutdown(boolean terminate) {
     }
 
     public void execute(String name, String uuid, Callable callable, Operation op) {
-        ExecutorConfig cfg = nodeEngine.getConfig().findExecutorConfig(name);
+        ExecutorConfig cfg = getOrFindExecutorConfig(name);
```
vbekiaris: Lazily populate the `ExecutorConfig` cache as needed.

hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
```diff
@@ -130,6 +134,7 @@ public boolean cancel(String uuid, boolean interrupt) {
     public void shutdownExecutor(String name) {
         executionService.shutdownExecutor(name);
         shutdownExecutors.add(name);
+        executorConfigCache.remove(name);
```
vbekiaris: Remove `ExecutorConfig` from cache, if one exists, once executor is shutdown.

gurbuzali:
👍

hazelcast/src/main/java/com/hazelcast/executor/impl/DistributedExecutorService.java
```diff
@@ -146,6 +151,7 @@ public void destroyDistributedObject(String name) {
         shutdownExecutors.remove(name);
         executionService.shutdownExecutor(name);
         statsMap.remove(name);
+        executorConfigCache.remove(name);
```
vbekiaris: Remove `ExecutorConfig` from cache, if one exists, when executor proxy is destroyed.

PR #8260, opened 2016-05-26
Make a better test for #1719

LoneRifle:
Don't just track when TestPredicate.apply was invoked, but TestPredicate.filter too.
Replace didApply with isFilteredAndAppliedOnlyOnce.

This has largely been superseded by tests implemented in #8053 but is left here since
it tests if the predicate is actually working by only being applied only to one entry


ahmetmircik:
run-lab-run

ahmetmircik:
👍

PR #8267, opened 2016-05-27
Adds test for LazyEntryViewFromRecord

ahmetmircik:


eminn:
👍

hazelcast/src/main/java/com/hazelcast/map/impl/record/CachedDataRecordWithStats.java
```diff
@@ -52,4 +52,31 @@ public Object getCachedValueUnsafe() {
...
         return CACHED_VALUE.compareAndSet(this, expectedValue, newValue);
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (o == null || getClass() != o.getClass()) {
```
jerrinot: it won't work with subclasses. is it desirable in *Record classes?

ahmetmircik: I don't have a counter example so it seem ok to me.

jerrinot:
:+1:

PR #8282, opened 2016-05-31
Reverse default for thread dump on test failure

mtopolnik:
Replaces the sysprop `hazelcast.test.disableThreadDumpOnFailure` with `hazelcast.test.threadDumpOnFailure`


gurbuzali:
we should add the new property to build runners if this PR gets merged
btw I'm in favor of reverting it 👍

cangencer:
👍 , noise from thread dumps is quite annoying. I didn't know they could be suppressed.

PR #8291, opened 2016-06-01
Wait for proxies to be created before firing a partition lost event

jerrinot:
Partition proxies are created eventually on all members. However
this is an async process driven by eventing system.

If a member does not have a proxy in its registry then it will
not fire `MapPartitionLostListener`.

I also removed the `assertMapContainerCreated()` as it served no purpose -
- `MapServiceContext.getMapContainer()` has getOrCreateSemantic
-> the assert was always true.

This is fixing spurious test failures in `test_mapPartitionLostListener_invoked_fromOtherNode`


metanet:
👍

vbekiaris:
👍

jerrinot:
@metanet, @vbekiaris: thanks for a quick review!

PR #8294, opened 2016-06-02
Suppress findbugs error

vbekiaris:
Suppress finbugs 'expose internal representation' errors in `OrPredicate` and `AndPredicate` for `getPredicates` and `setPredicates` methods which are accessed in `SqlPredicate` while flattening OR/AND predicates.


jerrinot:
:+1:

PR #8300, opened 2016-06-02
NPE in PartitionRuntimeState fixed

jerrinot:
NPE after deserialization

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/partition/PartitionRuntimeState.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/partition/PartitionRuntimeState.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import static com.hazelcast.internal.partition.InternalPartition.MAX_REPLICA_COUNT;
+import static com.hazelcast.util.StringUtil.LINE_SEPARATOR;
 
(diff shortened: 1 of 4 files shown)


mdogan:
👍

vbekiaris:
[`PartitionRuntimeState::writeData`](https://github.com/hazelcast/hazelcast/pull/8300/files#diff-f84ebd51a4ccc6ec495229355f2481d6R183) assumes `addressesToIndexes` will never be null. Isn't it possible to attempt to serialize a `PartitionRuntimeState` that was originally obtained from deserialization and get an NPE?

mdogan:
@vbekiaris: Good point. 
Currently, no, `PartitionRuntimeState` is not sent/serialized again after it's received from master member. But in theory or in future it's possible. We can fix `PartitionRuntimeState::writeData` with the same way too.

jerrinot:
@vbekiaris: very good finding indeed!
I'm going to update this PR.

jerrinot:
PR updated. 
@mdogan: There is also a NPE in `getPartitionTable()` - is it worth fixing too?

mdogan:
@jerrinot: `getPartitionTable()` currently only used after deserialization. But again, someday it can be used without serializing. Good thing is, NPE here will be more visible. It's up to you :)

PR #8301, opened 2016-06-02
InvocationMonitor fix problem with task failure

pveentjer:
Fixes a problem in the InvocationMonitor in case of task error. If a task executed by the scheduler throws an exception, and this task is a periodic executed task like scanning invocations, it will
not be executed again. This is solved by putting a try/catch around all tasks and dropping the uncaughtexceptionhandler from the thread.

Also some logging has been added to see if periodic tasks like scanning invocations is delayed too much. Also a probe has been added to record the number of delays.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
@@ -74,6 +74,7 @@ class InvocationMonitor implements PacketHandler, MetricsProvider {
 
     private static final long ON_MEMBER_LEFT_DELAY_MILLIS = 1111;
     private static final int HEARTBEAT_CALL_TIMEOUT_RATIO = 4;
+    private static final long MAX_DELAY_MILLIS = SECONDS.toMillis(10);
 
     private final NodeEngineImpl nodeEngine;
     private final InternalSerializationService serializationService;
@@ -93,6 +94,8 @@ class InvocationMonitor implements PacketHandler, MetricsProvider {
     @Probe
     private final SwCounter heartbeatPacketsSend = newSwCounter();
     @Probe
+    private final SwCounter delayedExecutionCount = newSwCounter();
+    @Probe
     private final long backupTimeoutMillis;
     @Probe
     private final long invocationTimeoutMillis;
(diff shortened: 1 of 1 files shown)


pveentjer:
I'm still not terribly excited about logging the time. I'm thinking about being able to calculate the next running time of certain tasks and if there is too much of a delay with the expected time and actual time, the task can do some logging.

pveentjer:
The logging has been improved so that a fixed rate task will show if it has been delayed.

pveentjer:
run-lab-run

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -212,6 +214,50 @@ long getLastMemberHeartbeatMillis(Address memberAddress) {
         return heartbeat == null ? 0 : heartbeat.get();
     }
 
+    private abstract class MonitorTask implements Runnable {
+        @Override
+        public void run() {
+            try {
+                run0();
+            } catch (Throwable t) {
+                inspectOutputMemoryError(t);
```
mtopolnik: Maybe this is a good opportunity to fix the name of this method? `inspectOutOfMemoryError`.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -212,6 +214,50 @@ long getLastMemberHeartbeatMillis(Address memberAddress) {
...
+            this.periodMillis = periodMillis;
+        }
+
+        @Override
+        public void run() {
+            long currentTimeMillis = System.currentTimeMillis();
+
+            try {
+                if (expectedNextMillis + MAX_DELAY_MILLIS < currentTimeMillis) {
+                    logger.warning(getClass().getSimpleName() + " delayed " + (currentTimeMillis - expectedNextMillis) + " ms");
```
mtopolnik: Would it contribute anything to triage the delay causes into "last task execution took too long" and "next task failed to be scheduled on time"?

mtopolnik:
👍 (minor comments only)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
pveentjer: What do you suggest?

mtopolnik: I already suggested :)

pveentjer: Ahhh.. I see it.

PR #8307, opened 2016-06-03
Revert partition table initialization check in migration flow tasks

metanet:
They have been introduced to prevent false-positive failures of an assertion but they introduced bugs into the system. So we prefer to revert those changes and remove the assertion since it is very likely to produce false positive failures. It turned out that the assertion is not a very useful one.


metanet:
run-lab-run

mdogan:
👍

mdogan:
merging since this is a trivial change.

PR #8275, opened 2016-05-30
Fixed failure in IMap.putAll() on bouncing members

Donnerbart:
See this comment: https://github.com/hazelcast/hazelcast/pull/8023/files#r64714809

Thanks to @ahmetmircik for the original test. Since the `MapPutAllWithBouncingMemberTest` is running for about a minute, but contributes to the code coverage, I made it a slow test.

PR #8310, opened 2016-06-03
Fixed IMap.putAll() on bouncing members with PutAllPartitionAwareOperationFactory

Donnerbart:
Replaced `PutAllPerMemberOperation` with `PutAllPartitionAwareOperationFactory` to fix failure in `IMap.putAll()` on bouncing members.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/OperationService.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/OperationService.java
@@ -106,6 +106,20 @@ public interface OperationService {
                                             Collection<Integer> partitions) throws Exception;
 
     /**
+     * Invokes an set of operation on selected set of partitions
+     * * <p/>
+     * This method blocks until all operations complete.
+     *
+     * @param serviceName      the name of the service
+     * @param operationFactory the factory responsible for creating operations
+     * @param partitions       the partitions the operation should be executed on.
+     * @return a Map with partitionId as key and the outcome of the operation as value.
+     * @throws Exception
+     */
+    Map<Integer, Object> invokeOnPartitions(String serviceName, OperationFactory operationFactory,
+                                            int[] partitions) throws Exception;
+
+    /**
      * Executes an operation remotely.
      * <p/>
      * It isn't allowed
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
@@ -24,7 +24,6 @@ import com.hazelcast.map.impl.iterator.MapKeysWithCursor;
 import com.hazelcast.map.impl.operation.ContainsKeyOperation;
 import com.hazelcast.map.impl.operation.EvictBackupOperation;
 import com.hazelcast.map.impl.operation.GetOperation;
-import com.hazelcast.map.impl.operation.PutAllPerMemberOperation;
 import com.hazelcast.map.impl.operation.PutBackupOperation;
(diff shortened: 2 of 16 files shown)


Donnerbart:
Either I didn't understand how to use the `PartitionAwareOperationFactory` correctly or it just doesn't match the needs of the putAll() method. With this approach I constantly get exceptions like this during migration:
```
java.lang.IllegalThreadStateException: Thread[hz._hzInstance_2_dev.partition-operation.thread-5,5,_hzInstance_2_dev] cannot make remote call: com.hazelcast.map.impl.operation.PutAllOperation{serviceName='hz:impl:mapService', identityHash=2091219611, partitionId=11, replicaIndex=0, callId=0, invocationTime=-1 (1970-01-01 00:59:59.999), waitTimeout=-1, callTimeout=60000, name=1bcff209-d870-4cf7-8b8b-518d95841f64}
```
I think the cause it that I still create a `PutAllPerMemberOperation` which contains data for several partitions. If nodes join or leave those operations may contain data for remote partitions. The `PartitionAwareOperationFactory` doesn't seem to handle this correctly, since there is no code to re-create a new operation with the correct partitions.

So I think we have to stick with the solution of https://github.com/hazelcast/hazelcast/pull/8275, even though it creates more code (which I think is needed).

ahmetmircik:
@Donnerbart Created a PR based on your commits here https://github.com/hazelcast/hazelcast/pull/8443 and it seems something like this can be a fix https://github.com/hazelcast/hazelcast/pull/8443/commits/b7a317c1bef60f572056574b9b5421e9917091a3

Donnerbart:
Yeah, it seems to work \o/ I also updated this PR with your fix.

Donnerbart:
I will do a comparison with a profiler between master, the other fix approach https://github.com/hazelcast/hazelcast/pull/8275 and this PR tomorrow. Just to be sure we have no performance degradation or other regression with one of the fixes. If both are equally good, I'll prefer this solution to reduce code duplication. I'll have a look at the EE side then.

Donnerbart:
run-lab-run

Donnerbart:
Hmm, I'm a bit baffled with the benchmark results. This approach with the `PartitionAwareOperationFactory` should in theory be slower, since it doesn't offer async methods for the `putAllInternal()` method. So each time an operation is sent to another member, it blocks. So the performance should be especially bad when batching is used (since this increases the number of remote operations).

image

hazelcast/src/main/java/com/hazelcast/spi/OperationService.java
```diff
@@ -106,6 +106,20 @@
...
+     * This method blocks until all operations complete.
+     *
+     * @param serviceName      the name of the service
+     * @param operationFactory the factory responsible for creating operations
+     * @param partitions       the partitions the operation should be executed on.
+     * @return a Map with partitionId as key and the outcome of the operation as value.
+     * @throws Exception
+     */
+    Map<Integer, Object> invokeOnPartitions(String serviceName, OperationFactory operationFactory,
+                                            int[] partitions) throws Exception;
```
ahmetmircik: we already have a similar method in `OperationService` [here] (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/spi/OperationService.java#L105). Cant we use that instead?

Donnerbart: I found no good way to convert an `int[]` to a `Collection<Integer>` and I didn't want to introduce litter again. Maybe we can get rid of the other method eventually, since the array should be more litter friendly.

ahmetmircik:
👍

gurbuzali:
👍

PR #8312, opened 2016-06-04
Introduced FileMetricSet + cleanup

pveentjer:
The FIleMetricSet is introduced that shows some statistics for diskspace
on the partition of the user.home directory.

Also a cleanup was done to not automatically register all metric sets in the
metrics registry; it isn't the responsibility of the registry.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/metrics/metricsets/FileMetricSet.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/metrics/metricsets/FileMetricSet.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


hazelcast/src/main/java/com/hazelcast/internal/metrics/metricsets/FileMetricSet.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Registers all the metrics in this metric pack.
+     *
+     * @param metricsRegistry the MetricsRegistry upon which the metrics are registered.
+     */
+    public static void register(MetricsRegistry metricsRegistry) {
+        checkNotNull(metricsRegistry, "metricsRegistry");
+
+        File file = new File(System.getProperty("user.home"));
```
jerrinot: is `System.getProperty("user.home")` guaranteed to never be null? what if the directly does not exist?

pveentjer: Theoretical case which can be dealt with once we see a single user running into it.

jerrinot:
:+1: I'd still like probes to be as much defensive as possible, but I'm ok with this one too.

PR #8313, opened 2016-06-04
Probe names for the operation thread is between brackets

pveentjer:
We use it with other probes so you get operation.thread[name].someproperty. To put it in line with other properties like tcp.connect[a->b].bla


jerrinot:
:+1:

PR #8269, opened 2016-05-29
Fixes in ICondition

jerrinot:
See #2272 for further discussion.

There are 2 big fixes:
- `ConditionKey` is tight to a particular Waiter.
This is needed as otherwise a thread might eat its own signal.
- BeforeAwaitOperation::getNotifiedKey returns `ConditionKey` when there
is a pending signal.

There are also additional improvements:
- Move getNotifiedKey() up to `LockStoreImpl`
- Map notifiers use this idiom for creating a notifier key:  If there is a pending signal (ConditionKey) then use it as a notifier. Otherwise use a regular LockWaitNotifyKey
- `ConditionInfo` simplification

Fixes #3317, #2272, #3789

PR #8318, opened 2016-06-05
Fix NPE when a lock is removed by LocalLockCleanupOperation

jerrinot:
The `LocalLockCleanupOperation` is executed on the same thread
just before the cleanWaitersAndSignalsFor is called.

The a lock is removed then the `cleanWaitersAndSignalsFor` might
throw NPE.

It's a regression introduced by #8269


mdogan:
👍

vbekiaris:
👍

jerrinot:
@mdogan, @vbekiaris: thanks for a quick review!

PR #8321, opened 2016-06-05
Fix Named-based Data Affinity

### LINKED ISSUE #7947: Client Transactional ISet w/ partition key not working
This fails:
```java
HazelcastInstance client = // client instance
ISet<String> foo = client.getSet("foo@0");
TransactionalTask task = new TransactionalTask() {
    public Object execute(TransactionalTaskContext context) throws TransactionException {
        TransactionalSet<String> foo = context.getSet("foo@0");
        foo.add("Hello");
        return null;
    }
};
client.executeTransaction(task);
assertEquals(1, foo.size());
```
Same code appears to work if executed on a member, _however_ if client submits as task to member, and executes the same on _on the member_, it still doesn't seem to work.

So to summarize, this is probably not exclusive to Set (could have similar issues for other DOs), but it seems isolated to client protocol, transaction, AND partitioned DO names.

jerrinot:
Fix named-based affinity when using structures mapped to a partition by name. 

It also re-uses `SetAbstractTest` on a client side.
Fixes #7947

Partition Strategy on a client side looks quite complicated, It should be reviewed by @asimarslan or @ihsandemir to make sure I didn't forget anything.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
@@ -34,6 +34,7 @@ import com.hazelcast.client.util.ClientDelegatingFuture;
 import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.IFunction;
 import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.partition.strategy.StringPartitioningStrategy;
 import com.hazelcast.ringbuffer.OverflowPolicy;
 import com.hazelcast.ringbuffer.ReadResultSet;
 import com.hazelcast.ringbuffer.Ringbuffer;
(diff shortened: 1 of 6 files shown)


jerrinot:
run-lab-run

jerrinot:
run-lab-run

asimarslan:
👍

ihsandemir:
👍 After the tests passes. I discussed with Jaromir that we need to examine which other structures and APIs use partitioning strategy and if they actually obey this rule. We need to review usages of ClientPartitionService.getPartitionId to make sure that we are not missing a usage. I opened an issue for this: https://github.com/hazelcast/hazelcast/issues/8328

jerrinot:
run-lab-run

jerrinot:
run-lab-run

jerrinot:
@asimarslan, @ihsandemir: Thanks for a review! Shall I send a backport PR?

Issue #7317, opened 2016-01-13
Reliable topic => StaleSequenceException


nilskp:
If a reliable `ITtopic` has a (`RingBuffer` with) small capacity and a `ReliableMessageListener` subscribes with a sequence number smaller than what's available (even when both `isLossTolerant = true` and `isTerminal = false`), it fails:
```
Caused by: com.hazelcast.ringbuffer.StaleSequenceException: sequence:0 is too small. The current headSequence is:2 tailSequence is:4
	at com.hazelcast.ringbuffer.impl.RingbufferContainer.checkBlockableReadSequence(RingbufferContainer.java:177)
	at com.hazelcast.ringbuffer.impl.operations.ReadManyOperation.beforeRun(ReadManyOperation.java:66)
	at com.hazelcast.spi.impl.operationservice.impl.OperationRunnerImpl.run(OperationRunnerImpl.java:166)
	at com.hazelcast.spi.impl.operationexecutor.classic.ClassicOperationExecutor.runOnCallingThread(ClassicOperationExecutor.java:392)
	at com.hazelcast.spi.impl.operationexecutor.classic.ClassicOperationExecutor.runOnCallingThreadIfPossible(ClassicOperationExecutor.java:345)
	at com.hazelcast.spi.impl.operationservice.impl.Invocation.doInvokeLocal(Invocation.java:259)
	at com.hazelcast.spi.impl.operationservice.impl.Invocation.doInvoke(Invocation.java:243)
	at com.hazelcast.spi.impl.operationservice.impl.Invocation.invokeInternal(Invocation.java:210)
	at com.hazelcast.spi.impl.operationservice.impl.Invocation.invoke(Invocation.java:180)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationBuilderImpl.invoke(InvocationBuilderImpl.java:46)
	at com.hazelcast.client.impl.protocol.task.AbstractPartitionMessageTask.processMessage(AbstractPartitionMessageTask.java:65)
	at com.hazelcast.client.impl.protocol.task.AbstractMessageTask.initializeAndProcessMessage(AbstractMessageTask.java:121)
	at com.hazelcast.client.impl.protocol.task.AbstractMessageTask.run(AbstractMessageTask.java:100)
	at com.hazelcast.spi.impl.operationservice.impl.OperationRunnerImpl.run(OperationRunnerImpl.java:127)
[...]

PR #8323, opened 2016-06-05
Include headSeq in StaleSequenceException on a client-side

jerrinot:
It's a similar issue to #7317 - I'm applying the same workaround.

Also I added `toString()` to client proxy implementation
and re-used RingbufferAbstractTest on a client side.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
@@ -37,12 +37,14 @@ import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.ringbuffer.OverflowPolicy;
 import com.hazelcast.ringbuffer.ReadResultSet;
 import com.hazelcast.ringbuffer.Ringbuffer;
+import com.hazelcast.ringbuffer.StaleSequenceException;
 import com.hazelcast.ringbuffer.impl.client.PortableReadResultSet;
 import com.hazelcast.spi.serialization.SerializationService;
 import com.hazelcast.util.CollectionUtil;
 import com.hazelcast.util.ExceptionUtil;
 
 import java.util.Collection;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 import static com.hazelcast.ringbuffer.impl.RingbufferProxy.MAX_BATCH_SIZE;
@@ -50,6 +52,7 @@ import static com.hazelcast.util.Preconditions.checkFalse;
 import static com.hazelcast.util.Preconditions.checkNotNegative;
 import static com.hazelcast.util.Preconditions.checkNotNull;
 import static com.hazelcast.util.Preconditions.checkTrue;
+import static java.lang.String.format;
 
 /**
  * Proxy implementation of {@link Ringbuffer}.
@@ -243,8 +246,20 @@ public class ClientRingbufferProxy<E> extends ClientProxy implements Ringbuffer<
         try {
             final Future future = new ClientInvocation(getClient(), clientMessage, partitionId).invoke();
             return (T) future.get();
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof StaleSequenceException) {
+                StaleSequenceException se = (StaleSequenceException) e.getCause();
+                long l = headSequence();
+                throw new StaleSequenceException(se.getMessage(), l);
+            }
+            throw ExceptionUtil.rethrow(e);
         } catch (Exception e) {
             throw ExceptionUtil.rethrow(e);
         }
+    }
+
+    @Override
+    public String toString() {
+        return format("Ringbuffer{name='%s'}", name);
     }
 }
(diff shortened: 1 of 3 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
```diff
@@ -243,8 +246,20 @@ private static void checkSequence(long sequence) {
         try {
             final Future future = new ClientInvocation(getClient(), clientMessage, partitionId).invoke();
             return (T) future.get();
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof StaleSequenceException) {
+                StaleSequenceException se = (StaleSequenceException) e.getCause();
+                long l = headSequence();
+                throw new StaleSequenceException(se.getMessage(), l);
```
jerrinot: this is ugly as stacktraces are lost. the other option is to set the `headSeq` via reflection....

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferAbstractTest.java
```diff
@@ -557,7 +557,7 @@ public void run() throws Exception {
 
         ReadResultSet<String> resultSet = f.get();
 
-        assertEquals(singletonList("1"), resultSet);
```
jerrinot: this makes assumption about the underlying implementation of `ReadResultSet`

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
pveentjer: Can't the original exception be set as cause? So you get a bit if a funky StaleSequenceException caused by StaleSequenceException.

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferAbstractTest.java
```diff
@@ -557,7 +557,7 @@ public void run() throws Exception {
...
+        assertThat(f.get(), contains("1"));
```
pveentjer: Since when are we using the assertThat approach?

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferAbstractTest.java
```diff
@@ -710,7 +711,7 @@ public void readManyAsync_whenMoreAvailableThanMaximum() throws ExecutionExcepti
         ReadResultSet<String> resultSet = f.get();
 
         assertNotNull(resultSet);
-        assertEquals(asList("1", "2"), resultSet);
+        assertThat(f.get(), contains("1", "2"));
```
pveentjer: contains != equals.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
jerrinot: I'm actually not sure if the original `StaleSequenceException` has any meaningful stack-traces here. Let me check it.

pveentjer: I'm fine with loosing the original stacktrace. It is a 'business' stacktrace; not some kind of internal error.

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferAbstractTest.java
jerrinot: from the contract of `contains()`
```
   * Creates a matcher for {@link Iterable}s that matches when a single pass over the
   * examined {@link Iterable} yields a series of items, each logically equal to the
   * corresponding item in the specified items.  For a positive match, the examined iterable
   * must be of the same length as the number of specified items.
   * <p/>
```

so it will match if the length is same with the same items in the same order. that's exactly what we want, isn't it?

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferAbstractTest.java
jerrinot: from the contract of `contains()`
```
   * Creates a matcher for {@link Iterable}s that matches when a single pass over the
   * examined {@link Iterable} yields a series of items, each logically equal to the
   * corresponding item in the specified items.  For a positive match, the examined iterable
   * must be of the same length as the number of specified items.
```

so it will match if the length is same with the same items in the same order. that's exactly what we want, isn't it?

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientRingbufferProxy.java
jerrinot: so there seems to be a meaningful stacktrace in the original exception. We could log it into `finest`, wdyt? just to make sure it's there for troubleshooting if ever needed.

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferAbstractTest.java
pveentjer: Then contains is a confusing name.

pveentjer:
:+1:

vbekiaris:
👍

PR #8329, opened 2016-06-06
Fix client protocol compilation problem

eminn:


gurbuzali:
👍

gurbuzali:
I'll merge this one since this is just an addition to response-types for the client protocol

PR #8332, opened 2016-06-06
Adds thread interrupt flag check to StoreWorker

ahmetmircik:


hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
```diff
@@ -360,13 +360,16 @@ public void callAfterStoreListeners(Collection<DelayedEntry> entries) {
         boolean result = false;
         Exception exception = null;
         int k = 0;
+        boolean interrupted = false;
         for (; k < RETRY_TIMES_OF_A_FAILED_STORE_OPERATION; k++) {
             try {
                 result = task.run();
+            } catch (InterruptedException ex) {
+                interrupted = true;
```
pveentjer: Why don't you break the loop directly when you run into an interrupt?

pveentjer: You are also loosing the interrupt flag. So it is best to restore it.

ahmetmircik: No, i am breaking the loop, but it cant be seen in PR view here. `loop break` is [here] (https://github.com/ahmetmircik/hazelcast/blob/4155235681cca997dab46bd0db745c672ded7776/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java#L375).

pveentjer: I see you are breaking the loop; but why don't you do it as soon as you catch the exception? What is the advantage of introducing a variable and after the loop check this variable?

Apart from that; you are loosing the interrupt flag. This is a problem because you might have broken free from whatever happened in the run method; any subsequent blocking call won't get interrupted.

ahmetmircik: Flag restoration is already there, will add that immediate break also

ahmetmircik:
verify

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
```diff
@@ -420,8 +424,9 @@ private void sort(List<DelayedEntry> entries) {
 
     private void sleepSeconds(long secs) {
         try {
-            TimeUnit.SECONDS.sleep(secs);
+            SECONDS.sleep(secs);
         } catch (InterruptedException e) {
+            currentThread().interrupt();
             logger.warning(e);
```
pveentjer: Do we want to log it? Because you already have handled id (by breaking the loop and restoring the flag).

ahmetmircik: no need. fixed

pveentjer:
:+1:

PR #8335, opened 2016-06-07
Fixes MBean replicatedmap issue

pveentjer:
1: instead of calling config.getMap, it should be config.getReplicatedMap
2: added toString on replicatedmapconfig


jerrinot:
:+1: a test would be nice.

pveentjer:
There is a test but it was broken as well. Check the files

PR #8340, opened 2016-06-07
[DONT MERGE] Entry listener with predicate should be notified on value entry/exit from predicate value space

vbekiaris:
This PR introduces filtering strategies in map event publisher, to facilitate building a 'query cache' style listener with predicate on `IMap` . Default filtering strategy maintains Hazelcast entry events behavior. A new filtering strategy has been introduced which processes UPDATED entry events and, depending on listener's predicate, may alter the type of event, to match exit/entry of value from predicate value space.


vbekiaris:
[`IssuesTest::testIssue321_3`](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/map/IssuesTest.java#L115) fails due to expectation that an entry event which include values will be published before the same event is published to another listener who excludes values. Since there are no event delivery order guarantees, changing the test to verify one event with value and another event without value are received by the listener, in any order.

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
```diff
@@ -140,75 +158,106 @@ public void publishEvent(Address caller, String mapName, EntryEventType eventTyp
...
-        List<EventRegistration> nullValueRegistrations = null;
+        publishEvent(registrations, caller, mapName, eventType, dataKey, oldValue, value, mergingValue);
+    }
+
+    private void publishEvent(Collection<EventRegistration> registrations, Address caller, String mapName,
+                                     EntryEventType eventType, Data dataKey, Object oldValue, Object value,
+                                     Object mergingValue) {
+
+        Map<Integer, EntryEventData> eventDataIncludingValues = new Int2ObjectHashMap<EntryEventData>(EVENT_DATA_MAP_CAPACITY);
+        Map<Integer, EntryEventData> eventDataExcludingValues = new Int2ObjectHashMap<EntryEventData>(EVENT_DATA_MAP_CAPACITY);
```
ahmetmircik: We can defer `Int2ObjectHashMap` creation at a later point, and can do it when needed, now for ex., if i have `value-included` event registrations and don't have any `value-excluded` ones, there will be always two `Int2ObjectHashMap` instance creation although one is sufficient. This event publishing part is on hot-path and it is better to create possible minimum number of objects.

vbekiaris: Good point, deferred map creation until actually needed.

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
```diff
@@ -140,75 +158,114 @@ public void publishEvent(Address caller, String mapName, EntryEventType eventTyp
...
-        List<EventRegistration> nullValueRegistrations = null;
+        publishEvent(registrations, caller, mapName, eventType, dataKey, oldValue, value, mergingValue);
+    }
+
+    private void publishEvent(Collection<EventRegistration> registrations, Address caller, String mapName,
+                                     EntryEventType eventType, Data dataKey, Object oldValue, Object value,
+                                     Object mergingValue) {
+
+        Map<Integer, EntryEventData> eventDataIncludingValues = null;
+        Map<Integer, EntryEventData> eventDataExcludingValues = null;
```
ahmetmircik: isn't an `ArrayList` cheaper than `Map` to hold event data in terms of number of created objects and calculations needed for hash? Seems an `ArrayList` is a better fit here and no need a Map

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
```diff
@@ -140,75 +158,114 @@ public void publishEvent(Address caller, String mapName, EntryEventType eventTyp
...
-        if (!isEmpty(includeValueRegistrations)) {
-            Data dataOldValue = mapServiceContext.toData(oldValue);
-            Data dataValue = mapServiceContext.toData(value);
-            Data dataMergingValue = mapServiceContext.toData(mergingValue);
-            EntryEventData eventData = createEntryEventData(mapName, caller, dataKey,
-                    dataValue, dataOldValue, dataMergingValue, eventType.getType());
-            int orderKey = pickOrderKey(dataKey);
-            publishEventInternal(includeValueRegistrations, eventData, orderKey);
+        // if events were generated, execute the post-publish hook on each one
+        if (eventDataIncludingValues != null) {
```
ahmetmircik: what about `eventDataExcludingValues`? is it need `postPublishEvent` phase?

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
```diff
@@ -140,75 +158,114 @@ public void publishEvent(Address caller, String mapName, EntryEventType eventTyp
...
+            for (EntryEventData entryEventData : eventDataIncludingValues.values()) {
+                postPublishEvent(entryEventData);
+            }
```
ahmetmircik: To prevent unnecessary iterations, signature can be changed to receive a collection `postPublishEvent(Collection<EntryEventData> eventData)`

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublisherImpl.java
vbekiaris: With current subclass implementation, no it's not needed, however in order to future-proof the implementation (and taking into account the following comment), we could amend `postPublishEvent` as `postPublishEvent(Collection<EntryEventData> eventDataIncludingValues, Collection<EntryEventData> eventDataExcludingValues)`.

ahmetmircik:
Maybe you want to remove some unused methods in `MapEventPublisherImpl`, stumbled them when i get your branch my local,

other than that LGTM 👍

vbekiaris:
Good catch, thanks for pointing that out.

vbekiaris:
The `MapEventPublisherImplTest` committed previously erroneously tests that `EntryEvent` objects received by listeners are the same, however this is not the case: even though a single `EntryEventData` object is used to dispatch the event to listeners, ultimately a new `DataAwareEntryEvent` object is delivered to each listener. Working on replacing this test with another one that actually tests the operation of each `EntryEventDataCache` implementation.

vbekiaris:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/event/DefaultEntryEventFilteringStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+        super(serializationService, mapServiceContext);
+    }
+
+    // This code has been moved from MapEventPublisherImpl.doFilter and
+    // provides the default backwards compatible filtering strategy implementation.
+    @SuppressWarnings("checkstyle:npathcomplexity")
+    @Override
+    public int doFilter(EventFilter filter, Data dataKey, Object dataOldValue, Object dataValue, EntryEventType eventType,
+                        String mapNameOrNull) {
+            if (filter instanceof MapPartitionLostEventFilter) {
```
tombujok: Why can the mapName be null?

hazelcast/src/main/java/com/hazelcast/map/impl/event/QueryCacheNaturalFilteringStrategy.java
```diff
@@ -0,0 +1,250 @@
...
+ *         <td>Mismatch</td>
+ *         <td>Mismatch</td>
+ *         <td>NO MATCH</td>
+ *     </tr>
+ *     <tr>
+ *         <td>Mismatch</td>
+ *         <td>Match</td>
+ *         <td>ADDED</td>
+ *     </tr>
+ * </table>
```
tombujok: Would be great to have such a behaviour-describing table in the `DefaultEntryEventFilteringStrategy` so that one can compare.

hazelcast/src/main/java/com/hazelcast/map/impl/event/QueryCacheNaturalFilteringStrategy.java
```diff
@@ -0,0 +1,250 @@
...
+
+            // decision matrix:
+            // +-----------+-----------+---------+
+            // | Old value | New value | Result  |
+            // +-----------+-----------+---------+
+            // |  Match    | Mismatch  | REMOVED |
+            // |  Match    | Match     | UPDATED |
+            // | Mismatch  | Mismatch  | NO MATCH|
+            // | Mismatch  | Match     |  ADDED  |
+            // +-----------+-----------+---------+
```
tombujok: ok, it's here, but maybe it's better to have it in the javadoc too?

tombujok:
👍 good job (some minor comments)

vbekiaris:
Thanks @ahmetmircik & @tombujok, I pushed updated javadoc & fixed the argument name that was leftover as `mapNameOrNull`. When green, will squash & commit.

hazelcast/src/main/java/com/hazelcast/map/impl/event/AbstractFilteringStrategy.java
```diff
@@ -0,0 +1,69 @@
...
+        QueryEventFilter queryEventFilter = (QueryEventFilter) filter;
+        QueryableEntry entry = new CachedQueryEntry((InternalSerializationService) serializationService,
+                dataKey, testValue, extractors);
+        return queryEventFilter.eval(entry);
+    }
+
+    private Extractors getExtractorsForMapName(String mapName) {
+        if (mapName == null) {
+            return Extractors.empty();
+        }
```
tombujok: Aha, one more thing. Ig the mapName can't be null (since you changed it in the last commit) is this null check here necessary?

vbekiaris: This `null` check was introduced in https://github.com/hazelcast/hazelcast/commit/faa6428c28eba0b8147e38196fe3711c46773ac8 but I don't think there are any chances `mapName` could be `null` as all code paths arrive at this point via one of `publishEvent` variants and the callers are various map operations, which _should_ have a non-null map name.

vbekiaris:
Unrelated test failure `com.hazelcast.durableexecutor.DurableExecutorServiceTest.testStatsIssue2039`:
```
java.util.concurrent.ExecutionException: java.util.concurrent.RejectedExecutionException: Capacity[1] is reached!!! 
	at com.hazelcast.durableexecutor.impl.TaskRingBuffer.findEmptySpot(TaskRingBuffer.java:70)
	at com.hazelcast.durableexecutor.impl.TaskRingBuffer.add(TaskRingBuffer.java:60)
	at com.hazelcast.durableexecutor.impl.DurableExecutorContainer.execute(DurableExecutorContainer.java:59)
	at com.hazelcast.durableexecutor.impl.operations.TaskOperation.run(TaskOperation.java:47)
	at com.hazelcast.spi.impl.operationservice.impl.OperationRunnerImpl.run(OperationRunnerImpl.java:182)
	at com.hazelcast.spi.impl.operationexecutor.impl.OperationThread.process(OperationThread.java:122)
	at com.hazelcast.spi.impl.operationexecutor.impl.OperationThread.run(OperationThread.java:102)
	at ------ submitted from ------.(Unknown Source)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.resolve(InvocationFuture.java:111)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.resolveAndThrow(InvocationFuture.java:74)
	at com.hazelcast.spi.impl.AbstractInvocationFuture.get(AbstractInvocationFuture.java:158)
	at com.hazelcast.durableexecutor.impl.DurableExecutorServiceProxy.submitToPartition(DurableExecutorServiceProxy.java:243)
	at com.hazelcast.durableexecutor.impl.DurableExecutorServiceProxy.submit(DurableExecutorServiceProxy.java:131)
	at com.hazelcast.durableexecutor.DurableExecutorServiceTest.testStatsIssue2039(DurableExecutorServiceTest.java:452)
```

vbekiaris:
run-lab-run

PR #8341, opened 2016-06-08
MemberAttributeEvent doesn't return null members

### LINKED ISSUE #7429: Potential NPE in MembershipEvent.getMembers()

pveentjer:
Instead and EMPTY_SET is returned. This doesn't cause any incompatibilities and prevents
running into an NPE.

fix #7429


jerrinot:
:+1:

PR #8085, opened 2016-05-04
Use return value from TestHazelcastInstanceFactory.initOrCreateConfig

mdogan:
When `null` config is passed to `TestHazelcastInstanceFactory.initOrCreateConfig()`,
it creates a new one returns it, otherwise initializes the config. But if returned
config is not used, a new default config is created while starting a new Hazelcast
instance and that causes failures on build system because multicast is enabled
in default config and that breaks isolation between tests.

Fixes many failures related to mock joiner. Fixes #7463

PR #8344, opened 2016-06-08
[BACKPORT] Use return value from TestHazelcastInstanceFactory.initOrCreateConfig

mdogan:
When `null` config is passed to `TestHazelcastInstanceFactory.initOrCreateConfig()`,
it creates a new one returns it, otherwise initializes the config. But if returned
config is not used, a new default config is created while starting a new Hazelcast
instance and that causes failures on build system because multicast is enabled
in default config and that breaks isolation between tests.

Backport of #8085


jerrinot:
:+1:

PR #8348, opened 2016-06-09
Fixes infinite loop in io system

### LINKED ISSUE #7522: Infinite Loop with NonBlockingSocketReader and ClientReadHandler
- Hazelcast version: 3.6
- Cluster size: 2
- Java version: 7

The "return" statement in the following method can determine an **infinite loop**:

com.hazelcast.nio.tcp.nonblocking.**NonBlockingSocketReader.handle()**

        if (!connection.isAlive()) {
            logger.finest("We are being asked to read, but connection is not live so we won't");
            return;
        }

We had a scenario where killing one member caused  com.hazelcast.nio.tcp.TcpIpConnection.close(Throwable) to be invoked on the other member, but a failure in socketChannel.close() invocation happened.
However, TcpIpConnection "alive" flag was set to false.
Then, com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThread.handleSelectionKey(SelectionKey) continued to be invoked because SelectionKey was still valid, with aforementioned handle() method returning because connection was not alive.
So the loop never ends, because selector is still valid and never closed (a read from the channel would probably return an EOF but the return statement prevents the read).

The same problem happens with:

com.hazelcast.client.connection.nio.**ClientReadHandler.handle()**

        if (!connection.isAlive()) {
                logger.finest("We are being asked to read, but connection is not live so we won't");
            }
            return;
        }

pveentjer:
Fixes #7522

The fix is done by removing the connection check and let the io thread run into
a socketchannel read/write and throw an ioexception.

Also exception logging has been cleaned up. Instead of having logging at 2 places, the connection.close and abstracthandler.onFailure; all logging is moved into the connection.close where one can properly check if the exception is causing the connection to get closed, or if the connection close is causing the exception (cause vs effect).

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
@@ -22,6 +22,7 @@ import com.hazelcast.nio.Connection;
 import com.hazelcast.nio.ConnectionType;
 import com.hazelcast.nio.OutboundFrame;
 
+import java.io.EOFException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Socket;
@@ -40,6 +41,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  *
  * @see IOThreadingModel
  */
+@SuppressWarnings("checkstyle:methodcount")
 public final class TcpIpConnection implements Connection {
 
     private final SocketChannelWrapper socketChannel;
@@ -157,10 +159,6 @@ public final class TcpIpConnection implements Connection {
         return connectionId;
     }
 
-    Object getConnectionAddress() {
-        return (endPoint == null) ? socketChannel.socket().getRemoteSocketAddress() : endPoint;
-    }
-
     public Object getMetricsId() {
         Socket socket = socketChannel.socket();
         SocketAddress localSocketAddress = socket != null ? socket.getLocalSocketAddress() : null;
@@ -230,6 +228,8 @@ public final class TcpIpConnection implements Connection {
         this.closeCause = cause;
         this.closeReason = reason;
 
+        logClose();
+
         try {
             if (socketChannel != null && socketChannel.isOpen()) {
                 socketReader.close();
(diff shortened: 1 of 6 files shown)


pveentjer:
@mdogan can you have a look at this one?

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
```diff
@@ -240,24 +240,38 @@ public void close(String reason, Throwable cause) {
...
         if (cause != null && monitor != null) {
             monitor.onError(cause);
         }
     }
 
+    private void logClose() {
+        String message = toString() + " closed. Reason: ";
+        if (closeReason != null) {
+            message += closeReason;
+        } else if (closeCause != null) {
```
mdogan: Shouldn't we append `closeCause` to the message (when `closeCause ` is non-null)? 
So, instead of `else if (closeCause != null)`, just `if (closeCause != null)`...

For example when `AbstractHandler` calls 
`connection.close("Exception in " + getClass().getSimpleName(), e)`, exception type and message won't be visible in connection log.

pveentjer: The cause already is included in the logging by adding the stacktrace of the closeCause. 

```
  private void logClose() {
        String message = toString() + " closed. Reason: ";
        if (closeReason != null) {
            message += closeReason;
        } else if (closeCause != null) {
            message += closeCause.getClass().getName() + "[" + closeCause.getMessage() + "]";
        } else {
            message += "Socket explicitly closed";
        }

        if (connectionManager.getIoService().isActive()) {
            if (closeCause == null || closeCause instanceof EOFException) {
                logger.info(message);
            } else {
                logger.warning(message, closeCause);<-------
            }
        } else {
            if (closeCause == null) {
                logger.finest(message);
            } else {
                logger.finest(message, closeCause);
            }
        }
    }
```

mdogan: Hmm. I missed `warning` log part... 
Still, having exception in 1st log line may be useful, for instance while doing a simple `grep`.. but not essential...

mdogan:
👍

jerrinot:
:+1:

PR #8357, opened 2016-06-09
Revert "Introduced Partition Iterator to Cache and Map"

eminn:
Reverting the changes since there are unresolved issues related to iteration consistency on the enterprise side.

Reverts hazelcast/hazelcast#8194


eminn:
Reverting the changes since there are unresolved issues related to iteration consistency on the enterprise side.

PR #8358, opened 2016-06-10
Deprecated hazelcast.application.validation.token

pveentjer:
This token was added for a potential customer some years ago, but I doubt it is being used.
So lets deprecate is so we can get rid of it in 3.8.

If people complain, we can always remove the deprecation.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
@@ -50,6 +50,7 @@ public final class GroupProperty {
      * <p/>
      * This validation-token will be checked before member join the cluster.
      */
+    @Deprecated
     public static final HazelcastProperty APPLICATION_VALIDATION_TOKEN
             = new HazelcastProperty("hazelcast.application.validation.token");
 


noctarius:
What was this parameter created for?

pveentjer:
```
  /**
     * Use this property to verify that Hazelcast nodes only join the cluster when their 'application' level configuration is the
     * same.
     * <p/>
     * If you have multiple machines, and you want to make sure that each machine that joins the cluster
     * has exactly the same 'application level' settings (such as settings that are not part of the Hazelcast configuration,
     * maybe some filepath). To prevent these machines with potential different application level configuration from forming
     * a cluster, you can set this property.
     * <p/>
     * You could use actual values, such as string paths, but you can also use an md5 hash. We'll give the guarantee
     * that nodes will form a cluster (become a member) only where the token is an exact match. If this token is different, the
     * member can't be started and therefore you will get the guarantee that all members in the cluster will have exactly the same
     * application validation token.
     * <p/>
     * This validation-token will be checked before member join the cluster.
     */

```

tombujok:
verify

pveentjer:
run-lab-run

Can someone have a look at it?

marco-admyre:
What's the correct way to achieve this functionality without this token? Cluster groups? For example in a Kubernetes cluster where you deploy new versions of apps you can end up in a situation where the old version talks to the new one briefly before the old one dies.

PR #8361, opened 2016-06-10
TCKMBeanServerBuilder moved to test classes

### LINKED ISSUE #3873: com.hazelcast.cache.impl.TCKMBeanServerBuilder should be moved into tests
com.hazelcast.cache.impl.TCKMBeanServerBuilder is not an implementation class. It is only used by TCK tests.

jerrinot:
Fixed #3873 

It's a class used by JSR107 TCK tests only. 
I assume it was in the `main` by accident. Let's see what will break when we move it.


jerrinot:
An idea - perhaps it will break TCK tests when our `test` classes are not on a classpath?

jerrinot:
as expected, this class is required in order to pass JSR107 TCK. 
Thanks to @vbekiaris for showing me this: http://docs.hazelcast.org/docs/3.6/manual/html-single/index.html#testing-for-jcache-specification-compliance

asimarslan:
👍

PR #8362, opened 2016-06-10
Added null checking for TransactionalMap method args

### LINKED ISSUE #7435: TransactionalMap allows null value.
`TransactionalMap` allows null value for `put` method which leads to  inconsistent state of the corresponding `IMap`.

Below is an example which demonstrates this behavior.
```java
public static void main(String[] args)
{
    HazelcastInstance hzInstance = Hazelcast.newHazelcastInstance();
    IMap<String, String> map = hzInstance.getMap("map");

    TransactionContext ctx = hzInstance.newTransactionContext();
    ctx.beginTransaction();

    TransactionalMap<String, String> txnMap = ctx.getMap("map");
    txnMap.put("key", null);

    System.out.println("transactional map size: " + txnMap.size());
    System.out.println("transactional map keys: " + txnMap.keySet());
    System.out.println("transactional map values: " + txnMap.values());

    ctx.commitTransaction();

    System.out.println("map size: " + map.size());
    System.out.println("map keys: " + map.keySet());
    System.out.println("map values: " + map.values());
}
```
In version 3.6, this produces the following output:
```
transactional map size: 0
transactional map keys: []
transactional map values: []
map size: 1
map keys: []
map values: []
```
Similar behavior also occurs on version 3.5.5.

If you put multiple null values, with different keys, into the `TransactionalMap`, the `TransactionalMap` size/keys/values will remain the same, but the size of `IMap` will increase accordingly, with the keys/values remaining empty.

Somewhat related to this issue is that `TransactionalMap` will throw a `NullPointerException` if the key is null. However, it is not due to a `Preconditions` check as is the case when a null key or value is used with `IMap`. I am not sure if this is intended behavior. Below is an example of the stack trace when I put a null key into a `TransactionalMap`.

```
Exception in thread "main" java.lang.NullPointerException
	at com.hazelcast.partition.impl.InternalPartitionServiceImpl.getPartitionId(InternalPartitionServiceImpl.java:1477)
	at com.hazelcast.map.impl.tx.TransactionalMapProxySupport.lockAndGet(TransactionalMapProxySupport.java:273)
	at com.hazelcast.map.impl.tx.TransactionalMapProxySupport.lockAndGet(TransactionalMapProxySupport.java:260)
	at com.hazelcast.map.impl.tx.TransactionalMapProxySupport.putInternal(TransactionalMapProxySupport.java:141)
	at com.hazelcast.map.impl.tx.TransactionalMapProxy.put(TransactionalMapProxy.java:131)
	at MyClass
```

If the transaction occurs on a client, I receive a `NullPointerException` with either a null key or value, with the following stack trace.
```
Exception in thread "main" java.lang.NullPointerException
	at com.hazelcast.client.impl.protocol.util.ParameterUtil.calculateDataSize(ParameterUtil.java:35)
	at com.hazelcast.client.impl.protocol.codec.TransactionalMapPutCodec$RequestParameters.calculateDataSize(TransactionalMapPutCodec.java:38)
	at com.hazelcast.client.impl.protocol.codec.TransactionalMapPutCodec.encodeRequest(TransactionalMapPutCodec.java:52)
	at com.hazelcast.client.proxy.txn.ClientTxnMapProxy.put(ClientTxnMapProxy.java:105)
	at com.hazelcast.client.proxy.txn.ClientTxnMapProxy.put(ClientTxnMapProxy.java:100)
	at MyClass
```

pveentjer:
Also some testing cleanup.

Fixes #7435

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/txn/ClientTxnMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/txn/ClientTxnMapProxy.java
@@ -64,6 +64,8 @@ public class ClientTxnMapProxy<K, V> extends ClientTxnProxy implements Transacti
 
     @Override
     public boolean containsKey(Object key) {
+        checkNotNull(key, "key can't be null");
+
         ClientMessage request = TransactionalMapContainsKeyCodec.encodeRequest(name, getTransactionId(),
                 ThreadUtil.getThreadId(), toData(key));
         ClientMessage response = invoke(request);
@@ -72,6 +74,8 @@ public class ClientTxnMapProxy<K, V> extends ClientTxnProxy implements Transacti
 
     @Override
     public V get(Object key) {
+        checkNotNull(key, "key can't be null");
+
         ClientMessage request = TransactionalMapGetCodec.encodeRequest(name, getTransactionId(),
                 ThreadUtil.getThreadId(), toData(key));
         ClientMessage response = invoke(request);
@@ -80,6 +84,8 @@ public class ClientTxnMapProxy<K, V> extends ClientTxnProxy implements Transacti
 
     @Override
     public V getForUpdate(Object key) {
+        checkNotNull(key, "key can't be null");
+
         ClientMessage request = TransactionalMapGetForUpdateCodec.encodeRequest(name, getTransactionId(),
                 ThreadUtil.getThreadId(), toData(key));
         ClientMessage response = invoke(request);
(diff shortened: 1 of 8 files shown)


jerrinot:
:+1: 
I reckon @asimarslan would like to have a look too.

pveentjer:
The client test should really be fixed so that a single test case can be used in:
local 
remote
client/server


This way I don't need to repeat these tests for the client.

gurbuzali:
👍

asimarslan:
Do we change an undocumented "null constraint" in TransactionalMap.
For the client compatibility we agreed not to fix undocumented API bugs.

asimarslan:
👍  I'm ok on client compatibility

PR #8367, opened 2016-06-13
AsyncAtomicReference deprecated

pveentjer:
The class is deprecated. Methods can be directly integrated into the IAtomicReference.

There is no official published way to get access to the AsyncAtomicReference.


jerrinot:
the order is other way around - first we should add the async methods to `IAtomicReference` only then to deprecate the `AsyncAtomicReference` class.

pveentjer:
There is no official way to get an instance to the IAtomicReference. There is just an interface.

vbekiaris:
@pveentjer if you are OK with #8580 can we close this one?

jerrinot:
I'm closing this one in favour of #8580

PR #8371, opened 2016-06-14
Unwanted responsenotsendexception stacktrace

pveentjer:
When a response can't be send because the member that send the invocation has left the cluster, you get a stacktrace due to the responsenotsendexception. 

Throwing of this exception has been replaced by a simple logging in the RemoteInvocationResponseHandler. Otherwise you need to deal with this exception at a different place. This causes the response sending logic (sending + failure handling) to be smeared over multiple locations.

ResponseNotSentException has been deprecated as well since it doesn't serve a purpose any longer.


pveentjer:
run-lab-run

jerrinot:
:+1:

PR #8377, opened 2016-06-14
NullPointerException If new member suddenly goes down

ihsandemir:
Fix for handling the NullPointerException probability when the listener registration for the failed member (e.g. Member is shutdown just when starting and listener could not be registered, then we can not remove the registration for the member, since the registration for that member could not be added at the first place). 

Note: This PR only aims at eliminating the possibility of NullPointerException and it does not aim at fixing any other possible event system problem.

I could not generate the problem in UIT tests, hence I do not have an additional test for the fix but a user reported such a case.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
@@ -165,7 +165,7 @@ public class ClientSmartListenerService extends ClientListenerServiceImpl implem
                     try {
                         invoke(registrationKey, member.getAddress());
                     } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
```diff
@@ -165,7 +165,7 @@ public void run() {
                     try {
                         invoke(registrationKey, member.getAddress());
                     } catch (Exception e) {
-                        logger.warning("Listener " + registrationKey + " can not added to new member " + member);
+                        logger.warning("Listener " + registrationKey + " can not added to new member " + member + ". Reason:" + e);
```
pveentjer: Please use logger.warning("somemessage", e).

This way you don't loose the stacktrace.

pveentjer:
:+1:

PR #8378, opened 2016-06-14
Fixing Predicate thread-safety

### LINKED ISSUE #8219: Thread-safety issue with Predicates
Since a predicate is reused on each node by all query threads thready-safety issues may arise in the following scenario.
- the user supplies a comparable to a predicate
- there's no guarantee if calling `equals` or `compareTo` is thread safe on the user supplied instance.
- in fact even in-built types like Date can be tricky since `java.util.Date.equals()` method is not thread-safe!

IMHO, In order to solve this issue it would be best to supply each query thread with a dedicated instance of the predicate. It will also allow caching of the converted values which was introduces in this PR #7589 (caching is already used in InPredicate though)

tombujok:
Fixes #8219
Unblocks #7589

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
@@ -219,7 +219,9 @@ public class MapQueryEngineImpl implements MapQueryEngine {
                 = new ArrayList<Future<Collection<QueryableEntry>>>(partitions.size());
 
         for (Integer partitionId : partitions) {
-            QueryPartitionCallable task = new QueryPartitionCallable(name, predicate, partitionId);
+            // Making sure not to use a predicate in a multi-threaded context
+            Predicate predicateCopy = serializationService.copy(predicate);
+            QueryPartitionCallable task = new QueryPartitionCallable(name, predicateCopy, partitionId);
             Future<Collection<QueryableEntry>> future = executor.submit(task);
             futures.add(future);
         }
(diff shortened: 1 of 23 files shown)


tombujok:
run-lab-run

jerrinot:
why you are making a deep copy via Java serialization?

tombujok:
What's the other option? Users may provide their own predicates. Predicate does not extend Cloneable and even if it did we would not be able to verify if it's a deep copy.
Predicate extends Serialisable so serialisation and deserialisation guarantees a deep copy

jerrinot:
We can use serialization service for this: then it's up to the serialization service to chose the best possible strategy.

tombujok:
Switched to SerializationService. Thx for this hint.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -219,7 +219,9 @@ protected QueryResult queryParallel(String name, Predicate predicate, Collection
                 = new ArrayList<Future<Collection<QueryableEntry>>>(partitions.size());
 
         for (Integer partitionId : partitions) {
-            QueryPartitionCallable task = new QueryPartitionCallable(name, predicate, partitionId);
+            // Making sure not to use a predicate in a multi-threaded context
+            Predicate predicateCopy = serializationService.copy(predicate);
```
ahmetmircik: am i wrong to assume this copying is only needed at first pass of predicate from user code? `predicate` must have been copied already before coming to `MapQueryEngineImpl `,  for example in `IMap#localKeySet` method so here we are copying _copy of predicate_?

tombujok:
After some discussions and having seen how many changes this fix requires this PR will be closed.
Instead I'll fix the issue with Date comparison (non thread-safe) and state in the doc that the Predicates should be thread-safe. Otherwise there's a high chance sb will add sth in the future without copying the predicate - it's a very error-prone approach it turned out.

PR #8379, opened 2016-06-15
Removed deprecated Operation.onException

pveentjer:
The method was deprecated in 3.6, so can be removed in the 3.7 release.


jerrinot:
:+1:

PR #8381, opened 2016-06-15
AbstractOperation deprecated

pveentjer:
Instead of the Operation providing abstract methods, it provide a default implementation (which was
the task of the AbstractOperation). Now there is no more need for an AbstractOperation.

AbstractOperation will be fully removed in 3.8.


jerrinot:
:+1:

PR #8382, opened 2016-06-15
Renamed OperationService methods

pveentjer:
executeOperation:execute runOperationOnCallingThread to run

These names are in line with the OperationExecutor. Otherwise we have naming inconsistencies
which make the code more difficult to understand than needs to be.


jerrinot:
:+1:

PR #8388, opened 2016-06-16
Restore com.hazelcast.instance.GroupProperty

pveentjer:


hazelcast/src/main/java/com/hazelcast/instance/GroupProperty.java
```diff
@@ -0,0 +1,655 @@
...
+import static java.lang.String.format;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Defines the name and default value for Hazelcast properties.
+ *
+ * @deprecated since 3.7. Will be removed in 3.8.
+ */
+@PrivateApi
```
mesutcelik: It would be good to have a `@Deprecated` here.

pveentjer: The deprecated is already in the javadoc. 

I never understood the added value of the @Deprecated because with the javadoc version at least one can provide documentation about when it is deprecated.

jerrinot:
:+1:

PR #8399, opened 2016-06-16
Moved multicast discovery to spi

pveentjer:
The multicast functionality should not be put in com.hazelcast.internal since this is private and should  not be known to an end user. However to configure multicast, a user will be pointing to the internal package, which is a clear violation.

the multicast package has been moved to com.hazelcast.spi.discovery.multicast and the implementation has been seprated from the public part by adding a com.hazelcast.spi.discovery.multicast.impl.

### DIFF
--- a/hazelcast/src/test/resources/hazelcast-multicast-plugin.xml
+++ b/hazelcast/src/test/resources/hazelcast-multicast-plugin.xml
@@ -30,7 +30,7 @@
             <tcp-ip enabled="false" />
             <aws enabled="false"/>
             <discovery-strategies>
-                <discovery-strategy enabled="true" class="com.hazelcast.internal.plugin.multicast.MulticastDiscoveryStrategy">
+                <discovery-strategy enabled="true" class="com.hazelcast.spi.discovery.multicast.MulticastDiscoveryStrategy">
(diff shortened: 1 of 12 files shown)


hazelcast/src/test/resources/hazelcast-multicast-plugin.xml
```diff
@@ -30,7 +30,7 @@
             <tcp-ip enabled="false" />
             <aws enabled="false"/>
             <discovery-strategies>
-                <discovery-strategy enabled="true" class="com.hazelcast.internal.plugin.multicast.MulticastDiscoveryStrategy">
+                <discovery-strategy enabled="true" class="com.hazelcast.spi.discovery.multicast.impl.MulticastDiscoveryStrategy">
```
jerrinot: this still does not feel good. an `impl` package in configuration.

pveentjer: Yes. I agree. I'll update the pr

jerrinot:
looks good to me
:+1: perhaps they is documentation/code samples to be updated too? @bilalyasar ?

bilalyasar:
👍

PR #8400, opened 2016-06-16
Fixes race in MulticastDiscoverySender

pveentjer:
stop needs to be volatile. Otherwise there is no happens before relation between
wring stop and reading stop.

A potential bug could be that the JIT decide to move the run check completely
out of the loop and the loop would never end; no matter how many times stop
is being called.


jerrinot:
:+1:

pveentjer:
Don't merge yet. AFAIK the MulticastDiscoveryStrategy needs to be public API as well

PR #8401, opened 2016-06-16
Test fix: serialization compatibility file was all big endian. now it cons…

mustafaiman:
…ists of big endian and little endian objects

Previously `DefaultSerializationServiceBuilder.setConfig` was overriding `.setByteOrder` and `.setAllowUnsafe`. Therefore serialized byte array representations of all objects were in big endian format.
I verified newly generated binary file both with test in here and Node.js client version of the same test (which failed before this change).


asimarslan:
👍

ihsandemir:
👍

Issue #8385, opened 2016-06-15
Fix migration commit vs partition-table update order


mdogan:
Currently `MigrationAwareService.commitMigration()` method is executed after partition-table changes are applied on both source and destination nodes. 

This order is fine for source node, unnecessary resources will be  cleaned up after partition assignment is removed.
But on destination node, this causes a race issue. Partition is assigned to node before `MigrationAwareService`s services completed their commit phase. An operation can be executed between partition assignment and `MigrationAwareService.commitMigration()`. For example, on `IMap` index rebuilding is done during commit phase but an in-fly query operation can come in before indexing is completed.

Expected commit flow order will be;
- Destination `MigrationAwareService.commitMigration()`
- Destination partition-table update 
- Source partition-table update 
- Source `MigrationAwareService.commitMigration()`

Additionally, a new commit hook method can be provided to be called before or after `commitMigration()` depending on source or destination. So, in that case, commit flow will be;

- Destination `MigrationAwareService.commitMigration()` 
- Destination partition-table update 
- Destination `commitMigrationHook()` 
- Source `commitMigrationHook()` 
- Source partition-table update 
- Source `MigrationAwareService.commitMigration()`

mdogan:
PR #9234 guarantees that an operation which is submitted after migration starts, will not be executed before `MigrationAwareService`s `commit()`/`rollback()` completes. That way, partitions will remain consistent before and after a migration. But this doesn't provide any guarantee for accesses outside of partition threads.

mdogan:
After some more thinking, we realize that calling `MigrationAwareService.commitMigration()` in different order on source and destination nodes seems not very possible, at least not without changing internals of migration commit mechanism. 
Reason is, currently, commit of a migration on source node is async and source migration commits can be received in a batch. And a source migration commit can be merged with a promotion commit. 
Promotion commits, which are specialized form of destination commits, are executed synchronously.
When a source commit and a promotion commit are sent together, they will be committed/applied atomically, we cannot split them into two.

Issue #8046, opened 2016-04-29
QueryIndexMigrationTest.testQueryWithIndexesWhileMigrating


hasancelik:
```
java.lang.AssertionError: expected:<30> but was:<29>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:834)
    at org.junit.Assert.assertEquals(Assert.java:645)
    at org.junit.Assert.assertEquals(Assert.java:631)
    at com.hazelcast.map.impl.query.QueryIndexMigrationTest.testQueryWithIndexesWhileMigrating(QueryIndexMigrationTest.java:157)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-IbmJDK1.7/com.hazelcast$hazelcast/892/testReport/junit/com.hazelcast.map.impl.query/QueryIndexMigrationTest/testQueryWithIndexesWhileMigrating/

vbekiaris:
Fixed by #8403

hasancelik:
**maintenance**
```
java.lang.AssertionError: expected:<30> but was:<29>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.map.impl.query.QueryIndexMigrationTest.testQueryWithIndexesWhileMigrating(QueryIndexMigrationTest.java:157)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.maintenance-OracleJDK1.6/com.hazelcast$hazelcast/395/testReport/junit/com.hazelcast.map.impl.query/QueryIndexMigrationTest/testQueryWithIndexesWhileMigrating/

tombujok:
@hasancelik Vassilis fixed it in 3.7, but it's too complex to backport it to 3.6, so it's a won't fix on the maintenance branch.

hasancelik:
```
java.lang.AssertionError: expected:<30> but was:<29>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
[...]

PR #8403, opened 2016-06-16
Query engine should not attempt to use indexes until all migrations are committed

### LINKED ISSUE #8257: PostJoinMapOperationTest.testPostJoinMapOperation_mapWithIndex
```
java.lang.AssertionError: expected:<1> but was:<0>
	at com.hazelcast.map.impl.operation.PostJoinMapOperationTest.testPostJoinMapOperation_mapWithIndex(PostJoinMapOperationTest.java:223)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x/com.hazelcast$hazelcast/5164/testReport/junit/com.hazelcast.map.impl.operation/PostJoinMapOperationTest/testPostJoinMapOperation_mapWithIndex/

vbekiaris:
There is a time window between partition table update and `commitMigration` execution is completed, during which the `QueryEngineImpl` will attempt to use an index, however the index may have not been created yet. In order to avoid this, in `MapMigrationAwareService` we keep track of the count of `beforeMigration` invocations vs `commitMigration`/`rollbackMigration` and only allow the `QueryEngineImpl` to use an index when this counter is 0.
This is a temporary workaround for 3.7, the root cause will be fixed in 3.8 (see #8385).
Fixes #8257 , #8046, #4517.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
@@ -126,7 +126,8 @@ public class MapQueryEngineImpl implements MapQueryEngine {
         // then we try to run using an index, but if that doesn't work, we'll try a full table scan
         // This would be the point where a query-plan should be added. It should determine if a full table scan
         // or an index should be used.
-        QueryResult result = tryQueryUsingIndexes(predicate, initialPartitions, mapContainer, iterationType);
+        QueryResult result = tryQueryUsingIndexes(predicate, initialPartitions, mapContainer, iterationType,
+                initialPartitionStateVersion);
         if (result == null) {
             result = queryUsingFullTableScan(mapName, predicate, initialPartitions, iterationType);
         }
@@ -141,9 +142,13 @@ public class MapQueryEngineImpl implements MapQueryEngine {
     }
 
     protected QueryResult tryQueryUsingIndexes(Predicate predicate, Collection<Integer> partitions, MapContainer mapContainer,
-                                               IterationType iterationType) {
+                                               IterationType iterationType, int initialPartitionStateVersion) {
 
-        if (partitionService.hasOnGoingMigrationLocal()) {
+        // if a migration is in progress, do not attempt to use an index as they may have not been created yet.
+        // MapService.getMigrationsInFlight() returns the number of currently executing migrations (for which
+        // beforeMigration has been executed but commit/rollback is not yet executed).
+        // This is a temporary fix for 3.7, the actual issue will be addressed with an additional migration hook in 3.8.
+        if (mapServiceContext.getService().getOwnerMigrationsInFlight() > 0) {
             return null;
         }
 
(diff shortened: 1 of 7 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -141,9 +142,13 @@ public QueryResult queryLocalPartitions(String mapName, Predicate predicate, Ite
...
     protected QueryResult tryQueryUsingIndexes(Predicate predicate, Collection<Integer> partitions, MapContainer mapContainer,
-                                               IterationType iterationType) {
+                                               IterationType iterationType, int initialPartitionStateVersion) {
 
-        if (partitionService.hasOnGoingMigrationLocal()) {
+        // if a migration is in progress, do not attempt to use an index as they may have not been created yet.
+        // MapService.getMigrationsInFlight() returns the number of currently executing migrations (for which
+        // beforeMigration has been executed but commit/rollback is not yet executed).
+        // This is a temporary fix for 3.7, the actual issue will be addressed with an additional migration hook in 3.8.
+        if (partitionService.hasOnGoingMigrationLocal() || mapServiceContext.getService().getMigrationsInFlight() > 0) {
```
metanet: do we still need this check? you are already tracking the partition owner migrations yourself. 'hasOnGoingMigrationLocal()` returns `true` for both partition owner and backup migrations. If current node is master, it also returns `true` for pending migrations. So it can cause a long delay since index will be disabled for irrelevant partitions.

hazelcast/src/main/java/com/hazelcast/spi/impl/DelegatingMigrationAwareService.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * A {@link MigrationAwareService} that delegates to another {@link MigrationAwareService} and keeps track of the number of
+ * migrations concerning the primary replica (either as current or new replica index) currently in-flight.
+ */
+public class DelegatingMigrationAwareService implements MigrationAwareService {
+
+    private static final int PRIMARY_REPLICA_INDEX = 0;
+
+    private final MigrationAwareService migrationAwareService;
+    private final AtomicInteger migrationsInFlight;
```
metanet: minor comment: can you add a comment or rename the field to specify it tracks only the partition owner migrations?

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
vbekiaris: Thanks for catching this, I will update the condition.

jerrinot:
:+1: good work!

metanet:
👍

vbekiaris:
Thanks @metanet & @jerrinot for your review

pveentjer:
Since there are 2 thumbs up, I'll hit the merge button.

PR #8404, opened 2016-06-16
Adding warning message when __key prefix used without a hash postfix

### LINKED ISSUE #7832: Creating index with wrong key-attribute definition does not throw an error
When creating indexes on keys you can define wrong specs without an error being thrown.

Correct:
```
map.addIndex("__key#login");
```

Wrong:
```
map.addIndex("__key.login");
``` 

Anyhow the latter does neither throw an exception, nor is it automatically corrected. The index is created but empty and will never be used. This results in extremely poor performance of queries without any visible reason (as a full table scan is executed).

tombujok:
Fixes #7832


pveentjer:
:+1:

PR #8406, opened 2016-06-16
HazelcastInstanceLoader respect client instance name

pveentjer:


nilskp:
When it's member and there's an instance name, it uses `Hazelcast.getOrCreateHazelcastInstance(config)`.
No equivalent exists on `HazelcastClient`, so it should probably use some combination of `HazelcastClient.getHazelcastClientByName` with `null` check and `HazelcastClient.newHazelcastClient` with check for `DuplicateInstanceNameException`, to eliminate race condition.

PR #8419, opened 2016-06-20
Fixes QueueAdvancedTest.testTakeInterruption 

pveentjer:
Because the call timeout was configured so small, the invocation didn't receive a heartbeat
in time and therefor it ran into an OperationTimeoutException. 

The call timeout is increased so that the test is less sensitive for delays.


jerrinot:
:+1:

Issue #6688, opened 2015-11-09
ClusterInfoTest.all_nodes_should_have_the_same_cluster_start_time_and_id_after_master_shutdown_and_new_node_join


hasancelik:
```
java.lang.AssertionError: the size of the collection is not correct: found-content:[Member [127.0.0.1]:9806, Member [127.0.0.1]:9808 this, Member [127.0.0.1]:9810] expected:<2> but was:<3>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.test.HazelcastTestSupport$6.run(HazelcastTestSupport.java:643)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:825)
	at com.hazelcast.test.HazelcastTestSupport.assertSizeEventually(HazelcastTestSupport.java:640)
	at com.hazelcast.test.HazelcastTestSupport.assertSizeEventually(HazelcastTestSupport.java:636)
	at com.hazelcast.cluster.ClusterInfoTest.all_nodes_should_have_the_same_cluster_start_time_and_id_after_master_shutdown_and_new_node_join(ClusterInfoTest.java:102)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK8-Quality-Outreach/com.hazelcast$hazelcast/234/testReport/junit/com.hazelcast.cluster/ClusterInfoTest/all_nodes_should_have_the_same_cluster_start_time_and_id_after_master_shutdown_and_new_node_join/

hasancelik:
```
java.lang.AssertionError: the size of the collection is not correct: found-content:[Member [127.0.0.1]:8248, Member [127.0.0.1]:8249 this, Member [127.0.0.1]:8250] expected:<2> but was:<3>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.test.HazelcastTestSupport$6.run(HazelcastTestSupport.java:643)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:825)
	at com.hazelcast.test.HazelcastTestSupport.assertSizeEventually(HazelcastTestSupport.java:640)
[...]

PR #8420, opened 2016-06-20
Fix missing MigrationAwareService.beforeMigration() calls

mdogan:
`MigrationAwareService.beforeMigration()` should be called for each service,
on both source and destination sides and there must be an exactly matching `commitMigration()`
or `rollbackMigration` for each `beforeMigration()` call.

But under some cases, there were missing `beforeMigration()` calls, which breaks this guarantee.
These cases are;
- when another `MigrationAwareService`'s `beforeMigration()` throws an exception
- when a `MigrationAwareService` doesn't provide replication-operation to be
executed on destination side

Now, these missing calls are fixed and current tests are enhanced to check this requirement
for `beforeMigration()`, `commitMigration` and `rollbackMigration()` calls.

Also, fixed a race which causes `AssertPartitionTableTask` to do a false assertion
when multiple members are terminated/crashed.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationManager.java
@@ -260,9 +260,6 @@ public class MigrationManager {
         return false;
     }
 
-    // suppressed intentionally since it is easier the follow the cases in one place.
-    // if this part gets more complicated, method can be divided into multiple methods
-    @SuppressWarnings("checkstyle:cyclomaticcomplexity")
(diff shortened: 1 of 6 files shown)


metanet:
👍

mdogan:
```
java.lang.AssertionError: expected:<1466412868183> but was:<0>
	at com.hazelcast.cluster.ClusterInfoTest.all_nodes_should_have_the_same_cluster_start_time_and_id_after_master_shutdown_and_new_node_join(ClusterInfoTest.java:112)
```
Re-opened issue: https://github.com/hazelcast/hazelcast/issues/6688

mdogan:
run-lab-run

gurbuzali:
👍

PR #8422, opened 2016-06-20
Fix Invalidconfigtest

### LINKED ISSUE #8169: InvalidConfigurationTest depends on system locale setting
The tests check the messages of the InvalidConfigurationExeptions thrown by the XmlConfigBuilder.
But these messages are often extracted from SAXParseExceptions which are localized.
So the maven build is platform dependend and fails.

pveentjer:
Fix #8169 

Bad practice to assert on generated strings by a implementation not controlled by us.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_TimeoutTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_TimeoutTest.java
@@ -298,7 +298,7 @@ public class Invocation_TimeoutTest extends HazelcastTestSupport {
 
     @Test
     public void sync_whenCallTimeout_thenOperationTimeoutException() throws Exception {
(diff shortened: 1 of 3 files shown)


hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_TimeoutTest.java
```diff
@@ -298,7 +298,7 @@ public void async_whenEventuallyHeartbeatTimeout_thenOperationTimeoutException()
 
     @Test
     public void sync_whenCallTimeout_thenOperationTimeoutException() throws Exception {
-        long callTimeoutMs = 10000;
+        long callTimeoutMs = 60000;
```
jerrinot: is this change intentionally in this PR?

tombujok:
@pveentjer do you want to rebase it before code freeze?

mdogan:
needs rebase

PR #8428, opened 2016-06-20
OperationService support  objects get own logger

pveentjer:
Makes it easier to figure out which component is causing the log entry + makes it easier to suppress.


jerrinot:
:+1:

PR #8436, opened 2016-06-22
Makes use of callersUUID in OperationFactoryWrapper

ahmetmircik:
instead of overriding it on node-side


mdogan:
👍

pveentjer:
:+1: once green

Issue #4690, opened 2015-02-27
QuerySlowTest.testRangeIndexSQLPerformance


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast/481/testReport/junit/com.hazelcast.map.query/QuerySlowTest/testRangeIndexSQLPerformance/

mesutcelik:
no recent failure. closing.

Donnerbart:
```
org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
	at java.lang.Object.wait(Native Method)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.pollResponse(InvocationFuture.java:300)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.waitForResponse(InvocationFuture.java:245)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.get(InvocationFuture.java:222)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.get(InvocationFuture.java:202)
	at com.hazelcast.map.impl.proxy.MapProxySupport.invokeOperation(MapProxySupport.java:301)
	at com.hazelcast.map.impl.proxy.MapProxySupport.putInternal(MapProxySupport.java:271)
	at com.hazelcast.map.impl.proxy.MapProxyImpl.put(MapProxyImpl.java:106)
	at com.hazelcast.map.impl.proxy.MapProxyImpl.put(MapProxyImpl.java:96)
	at com.hazelcast.map.impl.query.QuerySlowTest.testRangeIndexSQLPerformance(QuerySlowTest.java:159)
```
https://hazelcast-l337.ci.cloudbees.com/view/Sonar/job/Hazelcast-3.x-sonar/lastSuccessfulBuild/com.hazelcast$hazelcast/testReport/com.hazelcast.map.impl.query/QuerySlowTest/testRangeIndexSQLPerformance/

The whole build took about an hour longer than usual. We also had a ton of PR builders running during the execution. Maybe we just overcommited CloudBees (again).

gurbuzali:
link to the build misleading `lastSuccessfulBuild`, here is the correct one:
https://hazelcast-l337.ci.cloudbees.com/view/Sonar/job/Hazelcast-3.x-sonar/857/
[...]

PR #8439, opened 2016-06-22
Clean up QuerySlowTest and rename to QueryIndexPerformanceTest

mdogan:
QuerySlowTest was meant to test performance of indexed queries against
normal queries. But it was also doing verification of returned result set
and test methods were too complicated.

Now it's renamed to QueryIndexPerformanceTest and tests are simplified to only
measure time taken to return query result.

See https://github.com/hazelcast/hazelcast/issues/4690


gurbuzali:
👍

metanet:
👍

gurbuzali:
run-lab-run

PR #8440, opened 2016-06-22
Added index-usage capability to client-side predicated entryProcessors

ahmetmircik:


tombujok:
The PR looks good. My only comment (not strictly related to this PR) that it would be great to rename the OperationFactoryWrapper to ClientOperationFactory or ClientIdentifiableOperationFactory. 
The name OperationFactoryWrapper is really generic and we also do not really have generic wrapping (since in theory you could wrap a wrapper around another wrapper).
It's also not really a wrapper since we anyway can't treat if in a generic way - we know it's coming from a client to extract the uuid, so we do a down-cast anyway. In this context IMHO it would be better to give it a less generic name to underline why the logic is like-so.

@ahmetmircik do you think you could do this change in this PR?

Apart from that it's a 👍

ahmetmircik:
@tombujok thanks for the review, i tend to see that OperationFactoryWrapper issue in another PR.

PR #8446, opened 2016-06-23
Fixed batching in IMap.putAll()

Donnerbart:
Fixed batching in `IMap.putAll()` by using a `MutableLong` as wrapper, since the shared `Long` instance was overwritten by incrementing it.

I did a similar thing in the first approach and Peter voted for just using a `Long` instead of an additional wrapper class. The problem is that the `Long` instance is put several times into an array, to be reachable via a partitionId. But you cannot change the value of a `Long` instance, so a new one is created by incrementing it. And so each partition had it's own counter instead of a shared counter per member. The whole batching actually didn't work as expected.

This PR fixes this issue by introducing a small wrapper class for a primitive `long`, which can be incremented.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -43,6 +43,7 @@ import com.hazelcast.map.impl.MapEntries;
 import com.hazelcast.map.impl.MapService;
 import com.hazelcast.map.impl.MapServiceContext;
 import com.hazelcast.map.impl.PartitionContainer;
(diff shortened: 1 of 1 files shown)


ahmetmircik:
@Donnerbart Can we use  [MutableInteger](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/util/MutableInteger.java) or [MutableLong](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/util/MutableLong.java) instead of `SimpleCounter `?

Donnerbart:
Ah perfect, yes. Didn't know that we had such a class. Thanks!

ahmetmircik:
👍

PR #8449, opened 2016-06-24
Fix in ReplicatedMapMergePolicyTest

vbekiaris:
Introduced a wait-for-safe-state before splitting the cluster as the test would fail sometimes while waiting on the membership listener's latch (waiting for the member removed event).
Also parametrized the test to deduplicate code.


vbekiaris:
run-lab-run

tombujok:
👍

Donnerbart:
:+1: for the code dedupliation and for the waiting for a safe state (sounds reasonable to me).

Since I don't know the involved subsystems that well, I suggested to check if both tests touch the same code paths. If that is so I have absolutely no objection to merge this one :)

vbekiaris:
Confirmed code coverage is the same in this PR vs existing test.

PR #8456, opened 2016-06-27
`AbstractCacheOperation` cleanup

serkan-ozal:
EE counterpart: https://github.com/hazelcast/hazelcast-enterprise/pull/936


eminn:
I think the motivation behind this change is to support native memory operations on key iteration without making anything special on the EE side. Is this correct ?

serkan-ozal:
yes, same operations can be used for EE side and using interface type (`ICacheRecordStore`) instead of implementation type (`CacheRecordStore`) is better programming practice.

eminn:
👍

PR #8458, opened 2016-06-28
Fix REST shutdownNode test

### LINKED ISSUE #8106: RestTest.testKillNode
```
Error Message

expected:<[OK]> but was:<[{"status":"success"}]>
Stacktrace

org.junit.ComparisonFailure: expected:<[OK]> but was:<[{"status":"success"}]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at com.hazelcast.internal.ascii.RestTest.testKillNode(RestTest.java:269)
```

mdogan:
When node is requested to be shutdown via REST command, response may not be
sent to the caller, if connection is closed before writing response
to the socket channel. In that case, caller will get a connection exception
instead of actual response.

Also, changed shutdown command to use `HazelcastInstance.shutdown()` instead
of `Node.shutdown()`.

Fixes #8106


gurbuzali:
👍

vbekiaris:
👍 once green

bilalyasar:
👍

PR #8461, opened 2016-06-28
Fixes non-thread-safe comparison of java.util.Date in predicates

### LINKED ISSUE #7590: Add java.util.Date conversion to SqlTimestampConverter
Currently if the object stored in a Map has a `java.util.Date` field assigned with a `java.sql.Timestmap` instance and the type specified in a predicate is of java.util.Date type, the query will fail, since the convereter initialised in the `Predicate` will be `SqlTimestampConverter` and it does not handle the `java.util.Date` type.

`java.Sql.Timestamp` extends `java.util.Date` so adding the following section to the `SqlTimestampConverter` would be beneficial:
```
            if (value instanceof java.sql.Date) {
                return value;
            }
```
It's a regression - it's confirmed that it used to work in 3.4.6
Related zendesk ticket: 1453

### LINKED ISSUE #8219: Thread-safety issue with Predicates

tombujok:
Fixes #7590 Fixes #8219

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
@@ -21,8 +21,6 @@ import com.hazelcast.nio.serialization.Portable;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.sql.Timestamp;
-import java.util.Date;
 
 public final class TypeConverters {
     public static final TypeConverter BIG_INTEGER_CONVERTER = new BigIntegerConverter();
@@ -75,14 +73,16 @@ public final class TypeConverters {
         @Override
         Comparable convertInternal(Comparable value) {
             if (value instanceof java.sql.Date) {
-                return value;
+                return ((java.sql.Date) value).getTime();
+            }
+            if (value instanceof java.util.Date) {
+                return ((java.util.Date) value).getTime();
             }
             if (value instanceof String) {
-                return DateHelper.parseSqlDate((String) value);
+                return DateHelper.parseSqlDate((String) value).getTime();
             }
             if (value instanceof Number) {
-                Number number = (Number) value;
-                return new java.sql.Date(number.longValue());
+                return ((Number) value).longValue();
             }
             throw new IllegalArgumentException("Cannot convert [" + value + "] to java.sql.Date");
         }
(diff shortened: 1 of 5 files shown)


gurbuzali:
👍

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/DateHandlingPredicateTest.java
```diff
@@ -0,0 +1,164 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class DateHandlingPredicateTest extends HazelcastTestSupport {
+
+    private IMap<Integer, Customer> map;
+
+    private static long june2016Millis = 1467110170001L;
```
Donnerbart: All three fields should be `final`  and `june2016Millis` should be `JUNE_2016_MILLIS` or `JUNE_2016_TIMESTAMP`, since all are constants (and `CUSTOMER_0` and `CUSTOMER_1` are already following the naming pattern for constants).

hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
```diff
@@ -91,15 +91,20 @@ Comparable convertInternal(Comparable value) {
     static class SqlTimestampConverter extends BaseTypeConverter {
         @Override
         Comparable convertInternal(Comparable value) {
-            if (value instanceof Timestamp) {
-                return value;
+            if (value instanceof java.sql.Timestamp) {
+                return ((java.sql.Timestamp) value).getTime();
```
ahmetmircik: Seems `((java.sql.Timestamp) value).getTime()` is not enough for a true comparison. Please see below method from Timestamp source, it also make use of `nanos` in case of equality. 

```
public int compareTo(Timestamp ts) {
        long thisTime = this.getTime();
        long anotherTime = ts.getTime();
        int i = (thisTime<anotherTime ? -1 :(thisTime==anotherTime?0 :1));
        if (i == 0) {
            if (nanos > ts.nanos) {
                    return 1;
            } else if (nanos < ts.nanos) {
                return -1;
            }
        }
        return i;
    }
```

hazelcast/src/test/java/com/hazelcast/query/impl/predicates/DateHandlingPredicateTest.java
tombujok: Good, catch. Fixed

hazelcast/src/main/java/com/hazelcast/query/impl/TypeConverters.java
tombujok: Good catch too, it's breaking the comparision of two Timestamps. Let me fixed that.

tombujok:
@ahmetmircik @gurbuzali guys, I dug deeper into it and found out that calling getTime is not thread-safe on a Date. It means this problem is currently unsolvable. Here's the reasoning: https://github.com/hazelcast/hazelcast/issues/8219

PR #8464, opened 2016-06-28
SlowOperationDetector improvements

Donnerbart:
Improved `SlowOperationDetector` by moving the start time detection to the `OperationRunner`, which exactly knows when a task was started. This eases especially the unit tests, which were a nightmare to debug before. This also improves the accuracy of the detection, since we don't need to visit a task twice, to measure its runtime.


Donnerbart:
There has been no "logic" moves into the `OperationRunner`. It's like a probe which provides more accurate data for the detection logic. That is a small, but important difference.

pveentjer:
More accurate tracking is not relevant. This mechanism is meant to detect slow operations; operations that execute for a serious amount of time.

If we need accurate tracking, then the start and stop time should be controlled from the OperationRunner and once the operation runner is finished, is should publish this information. This is  a very valid approach, but this isn't sampling based as the current approach is.

This 'instrumentation' based approach could be very useful, but I would suggest doing this by making a custom OperationRunner. It could even be that some redesign is needed in the OperationRunner to make this happen.

PR #8470, opened 2016-06-29
Fixes #7590 Conversion of a Date to Timestamp

### LINKED ISSUE #7590: Add java.util.Date conversion to SqlTimestampConverter

tombujok:
Fixes #7590


ahmetmircik:
👍

vbekiaris:
👍

PR #8474, opened 2016-06-29
ShutdownAll in ManagementService now only deregisters related MBeans

### LINKED ISSUE #7467: All JMX Beans disappear when single Hazelcast instance is shut down in the same JVM
I am running multiple Hazelcast 3.5.5 instances in the same JVM. When i programatically shutdown one instance, all other existing instances are also deregistered as JMX Beans, but they are still alive and running. Creating a new instance programatically then only registers the new instance, all others are no longer available via JMX.

All running Hazelcast Instances should always be visible by JMX (if JMX is enabled).

muratayan:
fixes #7467 

We were previously doing:

```java
Set<ObjectName> entries = mbs.queryNames(new ObjectName(DOMAIN + ":*"), null);
for (ObjectName name : entries) {
    if (mbs.isRegistered(name)) {
        mbs.unregisterMBean(name);
    }
}
```

This approach was causing problems when there was at least 2 ClassLoaders and different clusters in same JVM.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/jmx/ManagementService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/jmx/ManagementService.java
@@ -20,6 +20,7 @@ import com.hazelcast.core.DistributedObject;
 import com.hazelcast.core.DistributedObjectEvent;
 import com.hazelcast.core.DistributedObjectListener;
 import com.hazelcast.instance.HazelcastInstanceImpl;
+import com.hazelcast.instance.HazelcastInstanceProxy;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.spi.properties.GroupProperty;
 
@@ -28,6 +29,7 @@ import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 import java.lang.management.ManagementFactory;
 import java.util.Hashtable;
+import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


ahmetmircik:
👍

hazelcast/src/main/java/com/hazelcast/internal/jmx/ManagementService.java
```diff
@@ -101,10 +103,29 @@ public void destroy() {
         }
     }
 
-    public static void shutdownAll() {
+    public static void shutdownAll(List<HazelcastInstanceProxy> instances) {
+        MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
+        for (HazelcastInstanceProxy instance : instances) {
+            try {
```
emrahkocaman: I think, it's better to change `shutdownAll(HazelcastInstanceImpl instance)` to  `shutdownAll(String instanceName)` and call it inside this for loop.
This way we can get rid of this nested for loop and avoid duplicate code.

hazelcast/src/main/java/com/hazelcast/internal/jmx/ManagementService.java
```diff
@@ -116,6 +137,7 @@ public static void shutdownAll() {
         }
     }
 
+
```
emrahkocaman: Why this new line? :)

muratayan:
Updated

emrahkocaman:
👍

hazelcast/src/main/java/com/hazelcast/internal/jmx/ManagementService.java
```diff
@@ -101,10 +103,17 @@ public void destroy() {
         }
     }
 
-    public static void shutdownAll() {
+    public static void shutdownAll(List<HazelcastInstanceProxy> instances) {
+        for (HazelcastInstanceProxy instance : instances) {
+            shutdown(instance.getName());
```
ahmetmircik: is it possible to return null from `instance.getName()`? What will be the impact in case of null return?

muratayan: I have no idea, can instance name be `null`? In that case `null String` goes here:

```java
    public static String quote(String text) {
        return Pattern.compile("[:\",=*?]")
                .matcher(text)
                .find() ? ObjectName.quote(text) : text;
    }
```

It shouldn't be a problem I guess, also this was the case before, never had any problems apparently.

emrahkocaman: Looks like it already is the case. Beans are registered with the name even it's `null`. (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/jmx/InstanceMBean.java#L116)
This shoudn't be a problem.

emrahkocaman:
verify

emrahkocaman:
run-lab-run

Issue #8471, opened 2016-06-29
PartitionReplicaStateCheckerTest.shouldNotBeSafe_whenUnknownReplicaOwnerPresent_whileNotActive & shouldBeSafe_whenKnownReplicaOwnerPresent_whileNotActive


hasancelik:
```
java.lang.AssertionError: expected:<SAFE> but was:<REPLICA_NOT_SYNC>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:834)
    at org.junit.Assert.assertEquals(Assert.java:118)
    at org.junit.Assert.assertEquals(Assert.java:144)
    at com.hazelcast.internal.partition.impl.PartitionReplicaStateCheckerTest.shouldNotBeSafe_whenUnknownReplicaOwnerPresent_whileNotActive(PartitionReplicaStateCheckerTest.java:157)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-ZuluJDK8/com.hazelcast$hazelcast/250/testReport/junit/com.hazelcast.internal.partition.impl/PartitionReplicaStateCheckerTest/shouldNotBeSafe_whenUnknownReplicaOwnerPresent_whileNotActive/

mdogan:
Improved code to be able to understand real cause of failure (with PR #8476). Closing for now. If it fails again, we'll be able to see which path exactly caused.

hasancelik:
```
java.lang.AssertionError: expected:<SAFE> but was:<REPLICA_NOT_OWNED>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at com.hazelcast.internal.partition.impl.PartitionReplicaStateCheckerTest.shouldNotBeSafe_whenUnknownReplicaOwnerPresent_whileNotActive(PartitionReplicaStateCheckerTest.java:165)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-IbmJDK1.7/com.hazelcast$hazelcast/1018/testReport/junit/com.hazelcast.internal.partition.impl/PartitionReplicaStateCheckerTest/shouldNotBeSafe_whenUnknownReplicaOwnerPresent_whileNotActive/

[...]

PR #8476, opened 2016-06-29
Add new PartitionServiceState.REPLICA_NOT_OWNED

mdogan:
Currently REPLICA_NOT_SYNC is used to denote both missing replica owners
and not-sync backup replicas.

It makes difficult to detect cause of failures like
https://github.com/hazelcast/hazelcast/issues/8471


metanet:
👍

PR #8479, opened 2016-06-29
Tests to cover member & client-side ICacheManagers.

vbekiaris:
Code coverage improvements for `HazelcastInstanceCacheManager` and `ClientICacheManager`

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/impl/ClientICacheManagerTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/impl/ClientICacheManagerTest.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/impl/ClientICacheManagerTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit test ClientICacheManager.getCache exception handling
+ */
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class ClientICacheManagerTest {
+
+    @Test(expected = HazelcastException.class)
```
tombujok: ExpectedException JUnit rule is much better in this context IMHO, but it's part of personal coding style.
Right now you don't know which line throws the exception. With the ExpectedException rule you also don't have to use the `fail` statement.

vbekiaris: 👍 I adopted the `ExpectedException` rule style for both tests. Using `fail` was redundant in the first place, but in any case I agree `ExpectedException` rule style makes for clearer declaration of intent in the code.

tombujok:
👍

Donnerbart:
:+1:

PR #8489, opened 2016-06-30
Runs read operations on replicateIndex==0 on partitionThread when readBackupData==true

### LINKED ISSUE #7341: Reading from backup does not update access data
When **MapConfig** option *"readBackupData"* is set to true then "get" operations do not trigger the update of access timestamps and hit count. This way the entries might get evicted even though they are frequently used. 

Cause of this issue is in *com.hazelcast.map.impl.DefaultRecordStore#readBackupData* method.

### LINKED ISSUE #7354: Access data is not updated by local reads when reading from backup is enabled

tombujok:
Fixes #7341 Fixes #7354 Fix author: @jerrinot
This should be backported to 3.6.5 too

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -131,9 +131,9 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
 
     /**
      * Defines the batch size for operations of {@link IMap#putAll(Map)} calls.
-     *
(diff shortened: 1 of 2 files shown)


tombujok:
This will caused performance regression - since the current readFromBackup does not run of partitionThread if partitionOwner... Closing this one.

PR #8490, opened 2016-07-01
Updates get-count on EntryProcessor.process(entry)

### LINKED ISSUE #7389: no-op multiple entry operations should update `get` and `hit` count
`AbstractMultipleEntryOperation` increments put/remove counts when entries are updated/removed
but if an `EntryProcessor` does only `get` then `get`/`hit` count is not updated.

Stats updated when entry added/updated;
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/operation/AbstractMultipleEntryOperation.java#L157

Stats updated when entry removed;
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/operation/AbstractMultipleEntryOperation.java#L146

tombujok:
Fixes #7389


vbekiaris:
👍 (once green)

Issue #8482, opened 2016-06-30
MapAttributeConfig cannot be loaded in an OSGi environment


chuspb:
Hi,

**Problem:**
I am using Hazelcast 3.6. When I try to use a _MapAttributeConfig_ in an OSGi environment I find that _MapContainer_ receives _name_, _mapconfig_ and _mapServiceContext_:

`public MapContainer(final String name, final MapConfig mapConfig, final MapServiceContext mapServiceContext);`

Then, inside it, the extractors are created like:

`this.extractors = new Extractors(mapConfig.getMapAttributeConfigs());`

And _Extractors_ uses _ExtractorHelpers_, which uses _Class.forName_ to load the _ValueExtractor_ described in each _MapAttributeConfig_. 

Then, I can not load extractors from different bundles, even if I add the bundle classloader to the Hz instance Config classloader.

**Solutions:**

**Option A:** Pass _com.hazelcast.config.Config_ from  _MapServiceContextImpl_ to _MapContainer_ constructor, and then pass it to _Extractors_ constructors. Then try lo load de class using the Config classloader, like: 
`config.getClassLoader().loadClass("com.MyValueExtractor")`. 
If it does not work, then try to load it as you do now with
 `instantiateExtractors(mapAttributeConfigs)`.

**Option B:** Let the _MapAttributeConfig_ have a _ValueExtractor_ instance, so _Extractors_ class just search into _MapAttributeConfig_ for them. The idea is to allow that a bundle can provide a _ValueExtractor_ dinamically.

I think both options can live together.

What do you think? It could be possible? Thanks.

chuspb:
I have made an approach https://github.com/hazelcast/hazelcast/pull/8491.

May be you can review it?

Thanks

tombujok:
@chuspb Thx for the fix. I had to change some minor details, thus I submitted a new PR, but I set you as the author so that it's your commit: #8493
Plz check if that's fine for you.

chuspb:
It is more completed with your changes. Thanks so much!

tombujok:
@chuspb Thx, BTW, did you make any tests that the bundle-class-loader will be propagated to Config.classLoader? or do you have to do it manually? 
IT would be great to have a sample to show that behavior in the samples code-base, sth similar to:
https://github.com/hazelcast/hazelcast-code-samples/blob/master/osgi/src/main/java/com/hazelcast/osgi/examples/OSGiSample.java

PR #8491, opened 2016-07-04
Allow Value Extractors to be loaded in an OSGi environment

chuspb:
IT resolves https://github.com/hazelcast/hazelcast/issues/8482

Basically, ExtractorsHelper can receive the Config instance now. This way, the extractor class of the MapAttributeConfig can be loaded from Config class loader before try it by doing a Class.forName. 

This is very important to OSGi enviromonment, because we can set an special ClassLoader to the Config instance in order to provide Extractors from diferent bundles.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
@@ -16,6 +16,13 @@
 
 package com.hazelcast.map.impl;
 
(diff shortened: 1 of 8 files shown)


tombujok:
@chuspb Thx for the fix. I had to change some minor details, thus I submitted a new PR, but I set you as the author so that it's your commit: https://github.com/hazelcast/hazelcast/pull/8493
Plz check if that's fine for you.

chuspb:
It is more completed with your changes. Thanks so much!

PR #8495, opened 2016-07-04
Improve performance of convertions to java.sql classes

cpiotr:
Converters for `java.sql.{Timestamp, Date, Time}` classes rely on `DateHelper` which uses `SimpleDateFormat` to construct first `java.util.Date` and then instance of target type.
In all three cases, `valueOf` methods exist, which do the same job in shorter time.

Issue: #7591

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/DateHelper.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/DateHelper.java
@@ -44,6 +44,7 @@ final class DateHelper {
 
     static Timestamp parseTimeStamp(final String value) {
         try {
+            // JDK format in Timestamp.valueOf is compatible with TIMESTAMP_FORMAT
             return Timestamp.valueOf(value);
         } catch (IllegalArgumentException e) {
             return throwRuntimeParseException(value, new ParseException(e.getMessage(), 0), TIMESTAMP_FORMAT);
@@ -52,6 +53,7 @@ final class DateHelper {
 
     static java.sql.Date parseSqlDate(final String value) {
         try {
+            // JDK format in Date.valueOf is compatible with DATE_FORMAT
             return java.sql.Date.valueOf(value);
         } catch (IllegalArgumentException e) {
             return throwRuntimeParseException(value, new ParseException(value, 0), SQL_DATE_FORMAT);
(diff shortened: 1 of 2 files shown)


cpiotr:
Hi @tombujok, this PR is a continuation of #7591

tombujok:
Thanks @cpiotr - looks awesome! We're now in the code-freeze period till the release of 3.7 in mid-July.
I'll come back to this PR right after!

noctarius:
I love simple code changes that improve speed! Great contribution! 👍

tombujok:
@cpiotr I had a look at the PR since we may merge it now. The question is whether it's 100% correct. The date/timestamp formats use locale and pattern. ValueOf doesn't take Locale under-consideration, does it? What if the pattern in code changes?

Also, did you test with JMH how much quicker it may be?

Here's a sample JMH test you could reuse:
https://github.com/tombujok/hazelcast/blob/8af891f9d71e09799ae70d1b85cb98aec9e4bd14/hazelcast/src/test/java/com/hazelcast/nio/serialization/impl/DefaultPortableReaderPerformanceTest.java#L32-L32

tombujok:
Ah, I see the perf-tests now ;) Great job!
https://github.com/hazelcast/hazelcast/issues/7591

But anyhow, the other question is valid.

cpiotr:
@tombujok AFAIR patterns used by both methods were the same. I also don't fully understand the question:
> What if the pattern in code changes?
Which pattern? The ones used in `DateHelper` are package-local and scarcely used.

tombujok:
Yeah, I meant if sb changes that internally - we loose that flexibility. But on the other hand id sb wants to change it they may revert to SimpleDateFormat. 

I checked it and you are right. TIMESTAMP_FORMAT, SQL_DATE_FORMAT, SQL_TIME_FORMAT allow digits only and the JDK in-built patterns are compatible with the HZ patterns so we may use it.
Could you maybe just add a comment in each line where valueOf is used that the JDK pattern is compatible with the desired SQL_TIME_FORMAT?

tombujok:
run-lab-run

tombujok:
@cpiotr when you add the comments we can merge this one and then I've got a very interesting issue for you :) #8828

tombujok:
run-lab-run

tombujok:
👍

jerrinot:
:+1: very nice contribution, it's much appreciated!

tombujok:
@cpiotr Thanks once again! If you have some more time feel free to start with this one #8828 :)

PR #8514, opened 2016-07-12
Fix condition await expiration when lock is already held

mdogan:
Currently, condition await expiration is failing when lock is held by
another thread at the time wait-notify system triggers expiration logic.

When await expiration is triggered, it's registered in lock service as expired
if lock is not available at that moment. We can't send failure response directly,
because thread awaiting should be owner of the lock when await() call returns.
UnlockOperation polls expired awaits after lock is released and executes AwaitOperation
using operation-service.

Problem is, operation-service thinks AwaitOperation is timed-out and returns call-timeout
response to await() which causes await to retry instead of expire immediately.

Fix is to assign lock to the caller of await() and send failure response right away after unlock.


ihsandemir:
👍

mdogan:
run-lab-run

metanet:
👍

PR #8524, opened 2016-07-14
Cluster-state should not allowed to change before startup is completed

### LINKED ISSUE #8523: Cluster-state should not allowed to change before startup is completed
Normally `NodeExtension.isStartCompleted()` returns true when node is joined successfully, but when hot-restart is enabled, it returns false until hot-restart process completes on all cluster.

mdogan:
Normally NodeExtension.isStartCompleted() returns true when node is joined successfully, but when hot-restart is enabled, it returns false until hot-restart process completes on all cluster.

Fixes #8523


metanet:
👍

Issue #8535, opened 2016-07-15
Bulk Get with 100 keys fails


wolframite:
A real memcached instance can fetch more than 150 keys at the same time, hazelcast fails at that because the capacity in the `com.hazelcast.nio.ascii.TextReadHandler` is set too low (currently 500).

If this is raised to e.g. 5000, it's possible to fetch 100 keys with a key length of 32 byte. I wrote a unit test, to reproduce the problem: https://github.com/zalora/hazelcast/commit/1c2d4fe14930782d33647cffdd2a0f575b36022c

Is it ok, if I raise the buffer size to 5000 and create a pull request for it? I already signed the contributor agreement form: https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement

noctarius:
Hey Wolfram,

Glad to see you're still working with Hazelcast. Unfortunately you still have to use the memcached API since the PHP is not ready ;-)

As long as you signed the CLA already you're good to send a PR including the fix and a unittest to proof the fix.

Lunatic666:
Oh, I'm not only working with Hazelcast, we're currently rolling it out at work as a replacement for Memcached. That's why I want to make Hazelcast more Memcached compatible. Unfortunately the PHP client has to wait until we're done with the rollout. 

I have some more changes in the pipeline, but I'd like to discuss them first and ask if there are better ways to achieve the goal. I'll send the PR and will mention the issue here...

LoneRifle:
@Lunatic666 I will be happy to help out with implementing the changes you may be proposing. @noctarius should be able to provide you with my contact details, or get hold of me on gitter.

noctarius:
Love the community spirit :) Please guys, go ahead!

Lunatic666:
@LoneRifle Thanks for the offer! I'll catch up with you on Gitter

sancar:
merged to https://github.com/hazelcast/hazelcast/issues/9355

PR #8536, opened 2016-07-15
Fixed: #8535 - Increased buffer size from 500 to 5000

wolframite:
Added unit test to show that the buffer in the TextReadHandler is too small to bulk-get 100 keys and Increased buffer size to 5000

Fix for Issue #8535

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/ascii/TextReadHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/ascii/TextReadHandler.java
@@ -62,7 +62,7 @@ public class TextReadHandler implements ReadHandler {
 
     private static final Map<String, CommandParser> MAP_COMMAND_PARSERS = new HashMap<String, CommandParser>();
 
-    private static final int CAPACITY = 500;
+    private static final int CAPACITY = 5000;
 
     static {
         MAP_COMMAND_PARSERS.put("get", new GetCommandParser());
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/nio/ascii/TextReadHandler.java
```diff
@@ -62,7 +62,7 @@
 
     private static final Map<String, CommandParser> MAP_COMMAND_PARSERS = new HashMap<String, CommandParser>();
 
-    private static final int CAPACITY = 500;
+    private static final int CAPACITY = 5000;
```
mdogan: not sure about having a static capacity. a dynamically growing buffer might be better, I think.

Lunatic666: I thought the ByteBuffer was used for performance reasons. If you say dynamic, would you use a StringBuilder with a start capacity?

mdogan: What I means is, since `TextReadHandler` is single threaded, `ByteBuffer` can be replaced with a bigger one when needed (for example, create with double capacity, copy contents and destroy the old one). Still, a max-threshold will be needed to avoid creating very large buffers.

Or the whole logic can be changed to use a relatively small (1k, 4k whatever) buffer to build command incrementally. I am not sure if it's  really needed to hold all data read from socket in a single large buffer.

sancar:
Hi thanks for the report and fix.  A chosen hardcoded value already proven not to work. The one we choose was not suitable for your usecase. The new one will be a problem for someone else in the future.  Making it working with any data should be the way forward as @mdogan  suggested. I am closing this one and linking the new issue here https://github.com/hazelcast/hazelcast/issues/9355

PR #8539, opened 2016-07-18
Included lastReadTime in TargetDisconnectedException

pveentjer:
to improve debugging.

And last heartbeat is a bit misleading since heartbeats are only send if
there is inactivity on the connection.


metanet:
👍

vbekiaris:
👍

PR #8541, opened 2016-07-18
Added connectionType probe

### LINKED ISSUE #8540: Metrics: connection type in name always NONE
When the connection is made, the connection type is not known yet. But the connection type is used in the metrics id, and the metrics id is used as soon as the connection is made. So it defaults to NONE in the 'getMetricsId' function.

And that is why you see for all connections:

```
   tcp.connection[NONE#/10.144.232.211:60318->/10.231.187.181:5701].in.ioThreadId=1
```

No matter if something a client thread or not.

Probably it would be better to add a probe 'connection type' (integer value) and remove the connection type from the TcpIpConnection.getMetricsId function

This is a minor issue. Just a small debug improvement.

pveentjer:
Added the connection type probe to TcpIpConnection

fixes #8540

This issue is a 'defect' since the printed content is just wrong.
This issue is also an enhancement since we can now see what kind of connection we are dealing with since there are many different types.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.nio.tcp;
 
+import com.hazelcast.internal.metrics.Probe;
 import com.hazelcast.logging.ILogger;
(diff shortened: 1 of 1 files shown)


Donnerbart:
:+1: (once green)

jerrinot:
:+1:

PR #8544, opened 2016-07-18
Increased code coverage of Clock. Fixed useless usage of Clock.

Donnerbart:
* Increased code coverage of `Clock`
* Pulled out `AbstractClockTest` to ease handling of class loader magic to properly change the `Clock` implementation in an isolated environment
* Added `RootCauseMatcher` to assert the root cause of a thrown exception
* Fixed `SecondsBasedEntryTaskSchedulerTest` which tried to override the `Clock` implementation without any effect


hazelcast/src/test/java/com/hazelcast/util/scheduler/SecondsBasedEntryTaskSchedulerTest.java
```diff
@@ -32,151 +34,151 @@
...
-            final SecondsBasedEntryTaskScheduler<Integer, Integer> scheduler =
-                    new SecondsBasedEntryTaskScheduler<Integer, Integer>(executorService, entryProcessor, POSTPONE);
-
-            scheduler.schedule(0, 1, 1);
-            scheduler.schedule(0, 1, 1);
-            assertNotNull(scheduler.get(1));
-            assertEquals(1, scheduler.size());
-        } finally {
-            System.clearProperty(ClockProperties.HAZELCAST_CLOCK_IMPL);
+    public void test_doNotRescheduleEntryWithinSameSecond_postpone() {
```
Donnerbart: This is the fixed test, which tried to change the Clock implementation just in the middle of nowhere. Due to the singleton implementation, this had no effect and just worked by chance. Since this test doesn't use a Hazelcast instance, I changed it without making use of the `AbstractClockTest` class.

cangencer: from what I can see, the second assertion will be ignored if the second jumps when running the test. Wouldn't it be better in some kind of loop? i.e. to try doing the assertion until the start/stop seconds are equal.

cangencer:
👍 except one comment

hazelcast/src/test/java/com/hazelcast/util/scheduler/SecondsBasedEntryTaskSchedulerTest.java
Donnerbart: Oh, that's a good idea, thx.

Donnerbart: I implemented the loop and gave the test a timeout, so we don't have to deal with a max retry for the loop. The whole test class finishes in under 50 ms on my machine, so 10 seconds should be a very generous timeout value for this method.

cangencer: :+1:

pveentjer:
:+1:

PR #8556, opened 2016-07-20
adds Node.js client as a known client

mustafaiman:
Adds Node.js client among known clients. So when a Node.js client (0.4 or greater) connects to server, it will not be shown as UNKNOWN BINARY CLIENT.

Here is an example of how it looks:
```
INFO: [192.168.2.31]:5701 [dev] [3.7-SNAPSHOT] Received auth from Connection[id=1, /127.0.0.1:5701->/127.0.0.1:49587, endpoint=null, alive=true, type=NODEJS_CLIENT], successfully authenticated, principal : ClientPrincipal{uuid='cc4ec7d8-c182-4084-aad7-7161f2e3db55', ownerUuid='7a7188e4-2d5a-4ec5-801c-f3faec60bb37'}, owner connection : true
```

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/ClientTypes.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/ClientTypes.java
@@ -54,6 +54,11 @@ public final class ClientTypes {
      */
     public static final String RUBY = "RBY";
 
+    /**
+     * Node.JS client protocol id
+     */
+    public static final String NODEJS = "NJS";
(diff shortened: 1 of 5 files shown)


Donnerbart:
Changes look plausible :+1:

cangencer:
👍 
in the future the logic for getting connectionType based on client ID should be moved into ClientTypes class, since it's getting rather long now.

Issue #7693, opened 2016-03-08
ClientTransactionalMapQuorumTest.*. [Executing: TransactionOptions{timeoutMillis=120000, durability=1, txType=TWO_PHASE}]


hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 300000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:156)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:811)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:969)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1281)
	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:207)
	at com.hazelcast.client.connection.nio.ClientConnectionManagerImpl$BlockingCallback.get(ClientConnectionManagerImpl.java:259)
	at com.hazelcast.client.connection.nio.ClientConnectionManagerImpl.getOrConnect(ClientConnectionManagerImpl.java:234)
	at com.hazelcast.client.spi.impl.ClientTransactionManagerServiceImpl.connect(ClientTransactionManagerServiceImpl.java:107)
	at com.hazelcast.client.proxy.txn.TransactionContextProxy.<init>(TransactionContextProxy.java:58)
	at com.hazelcast.client.spi.impl.ClientTransactionManagerServiceImpl.newTransactionContext(ClientTransactionManagerServiceImpl.java:57)
	at com.hazelcast.client.impl.HazelcastClientInstanceImpl.newTransactionContext(HazelcastClientInstanceImpl.java:427)
	at com.hazelcast.client.impl.HazelcastClientProxy.newTransactionContext(HazelcastClientProxy.java:169)
	at com.hazelcast.client.quorum.ClientTransactionalMapQuorumTest.testTxReplaceThrowsExceptionWhenQuorumSizeNotMet(ClientTransactionalMapQuorumTest.java:207)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x/com.hazelcast$hazelcast-client/4687/

[...]

PR #8557, opened 2016-07-20
Improve ClientTransactionalMapQuorumTest

cangencer:
Improved tests while investigating #7693. Add tests for positive case,
and made exception handling more sensitive to catching the correct exception.


mustafaiman:
👍 however I find `@Test(expected = Exception.class)` more readable compared to `ExpectedException.expect()`. I guess we do not have any rule for asserting exceptions so it should be fine either way.

cangencer:
The problem with `@Test(expected=Exception.class)` is that it doesn't specify where the exception can happen, so if some other method than the intended throws the exception, the test still passes.

mustafaiman:
I see. Thanks for explanation.

eminn:
👍

PR #8561, opened 2016-07-21
Priority client messages executed on generic operation executor

pveentjer:
Normally all generic client messages are processed by the client executor. However there is no priority,
and tasks in the executor can get processed very slowely. This can cause stability problems e.g. by
a failure to respond to heartbeats in time.

This fix offloads the most important priority generic messages to the operatione executor where there is a
dedicated thread dealing with priority tasks. This makes it a lot less unlikely that the system can't reply
to a heartbeat in time.

Thanks @mdogan for helping out and pinpointing the problem.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
@@ -28,7 +28,11 @@ import com.hazelcast.client.impl.operations.PostJoinClientOperation;
 import com.hazelcast.client.impl.protocol.ClientExceptionFactory;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.impl.protocol.MessageTaskFactory;
+import com.hazelcast.client.impl.protocol.task.AuthenticationCustomCredentialsMessageTask;
+import com.hazelcast.client.impl.protocol.task.AuthenticationMessageTask;
+import com.hazelcast.client.impl.protocol.task.GetPartitionsMessageTask;
 import com.hazelcast.client.impl.protocol.task.MessageTask;
+import com.hazelcast.client.impl.protocol.task.PingMessageTask;
 import com.hazelcast.config.Config;
 import com.hazelcast.core.Client;
 import com.hazelcast.core.ClientListener;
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -162,15 +168,28 @@ public int getClientEndpointCount() {
 
     public void handleClientMessage(ClientMessage clientMessage, Connection connection) {
         int partitionId = clientMessage.getPartitionId();
-        final MessageTask messageTask = messageTaskFactory.create(clientMessage, connection);
+        MessageTask messageTask = messageTaskFactory.create(clientMessage, connection);
+        InternalOperationService operationService = nodeEngine.getOperationService();
         if (partitionId < 0) {
-            executor.execute(messageTask);
+            if (isUrgent(messageTask)) {
+                operationService.execute(new PriorityPartitionSpecificRunnable(messageTask));
```
jerrinot: I don't get it. 
`partitionId < 0` (line 173) means it is NOT a partition specific task, doesn't it? Yet here you are creating `PriorityPartitionSpecificRunnable` 

`/me = confused`

pveentjer: The PriorityPartitionSpecificRunnable is returning the partition id of the message task, which is -1.

pveentjer:
It is is a bit shitty that PartitionSpecificRunnable and PriorityOperation need to be implemented. Would be nicer if it would be a method:

execute(runnable,partitionid, priority). 

We have that method in the OperationExecutor, but it isn't exposed in the OperationService.

jerrinot:
`MessageTask` already extends `PartitionSpecificRunnable` 

The only reason for wrapping is to add the `UrgentSystemOperation` marker? If so wouldn't it be better to add the marker directly to the selected tasks? (and avoid wrapping)

pveentjer:
The task should not be polluted with the wrapping. It already has a priority flag that doesn't work. 

I would rather fix the operation executor so it exposes a method 'execute(task,partitionid,priority)' than to add more dirt throughout the system.

jerrinot:
:+1:

Donnerbart:
It's not super beautiful, but the code looks correct. Would be good if Mehmet would review as well, since he was involved.

:+1:

pveentjer:
It is everything but beautiful :) In 3.8 we need to expose execute(task,partitionid,priority). This way these interface magic isn't needed.

mdogan:
👍 

Do we have an issue in 3.8 for this?

pveentjer:
There is no explicit github ticket for the more powerful met in the operationservice. But there is a ticket to clean up the generic task scheduling from the client on the server.

PR #8563, opened 2016-07-21
Improved ClientConnection.toString

pveentjer:
Includes various times like read/write/close etc. Removed the read/writehandlers so toString
is implemented and it just generates noise.


jerrinot:
:+1:

Donnerbart:
:+1:

PR #8564, opened 2016-07-21
Fixes client aborting invocations too early

pveentjer:
The CleanResourcesTask kills the invocation after the 'connection.heartBeating' is set to false, but before
the client has time to receive the heartbeat, the CleanResourcesTask aborts the client invocation.

So the fix is to only abort invocations after the connection is not alive.

Another minor change is that the exception send to the invocation was misleiding. It was talking about heartbeat problems, but it could very well be that the connection was closed because the master to the client it should drop the connection to the given member.


jerrinot:
2 failing tests:
```
com.hazelcast.client.heartbeat.ClientHeartbeatTest.testAsyncInvocation_whenHeartbeatStopped
com.hazelcast.client.heartbeat.ClientHeartbeatTest.testInvocation_whenHeartbeatStopped
```

PR #8567, opened 2016-07-21
Connection.getCloseReason()

pveentjer:
If the close cause is not explicitly set, the getCloseReason will look at the closeCause.getMessage. This makes it easier to include it in exceptions/logging.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
@@ -267,7 +267,11 @@ public class ClientConnection implements Connection {
 
     @Override
     public String getCloseReason() {
-        return closeReason;
+        if (closeReason == null) {
+            return closeCause == null ? null : closeCause.getMessage();
+        } else {
+            return closeReason;
+        }
     }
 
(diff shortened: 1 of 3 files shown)


Donnerbart:
:+1: (once green)

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientReadHandler.java
```diff
@@ -81,6 +81,10 @@ public void handle() throws Exception {
 
         lastHandle = Clock.currentTimeMillis();
 
+        if(!connection.heartBeating){
```
Donnerbart: CheckStyle error in this block (whitespaces).

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
```diff
@@ -65,7 +65,7 @@
     private final HazelcastClientInstanceImpl client;
 
     private volatile Address remoteEndpoint;
-    private volatile boolean heartBeating = true;
+    public volatile boolean heartBeating = true;
```
Donnerbart: Fields out of order (public -> private).

mdogan:
@pveentjer: needs rebase

PR #8569, opened 2016-07-21
[DONT MERGE] Allow QueryOperation to filter for specific partition IDs

vbekiaris:
WIP
Update query operation with an optional list of "requested for" partition IDs. When these are not null, QueryOperation will return results for those partition IDs owned by the member listed in requested-for partition IDs.


vbekiaris:
Closing this PR in favor of #8570, as this affects query execution even on stable cluster. It may make sense to reconsider in the future, when indexes-per-partition are implemented.

Issue #6471, opened 2015-10-15
QueryIndexMigrationTest.testIndexCleanupOnMigration


hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 60000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:182)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:822)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:980)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1292)
	at java.util.concurrent.FutureTask$Sync.innerGet(FutureTask.java:229)
	at java.util.concurrent.FutureTask.get(FutureTask.java:94)
	at com.hazelcast.map.impl.query.QueryIndexMigrationTest.testIndexCleanupOnMigration(QueryIndexMigrationTest.java:185)

```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-IbmJDK1.6/com.hazelcast$hazelcast/712/testReport/junit/com.hazelcast.map.impl.query/QueryIndexMigrationTest/testIndexCleanupOnMigration/

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK6/com.hazelcast$hazelcast/693/testReport/junit/com.hazelcast.map.impl.query/QueryIndexMigrationTest/testIndexCleanupOnMigration/

hasancelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x/com.hazelcast$hazelcast/3894/testReport/junit/com.hazelcast.map.impl.query/QueryIndexMigrationTest/testIndexCleanupOnMigration/

sancar:
Failed in my pr 
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/16325/console

hasancelik:
[...]

PR #8570, opened 2016-07-21
When adding results from query operations, discard results when one or more partitions have already been processed

vbekiaris:
When aggregating results, do not take into account results for partition IDs which have already been accounted for. This will cause separate per-partition `QueryPartitionOperation`s to be executed

Fixes over-reporting values in #6471

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
@@ -524,9 +524,14 @@ public class MapQueryEngineImpl implements MapQueryEngine {
                 continue;
             }
 
-            Collection<Integer> tmpPartitionIds = queryResult.getPartitionIds();
-            if (tmpPartitionIds != null) {
-                partitionIds.removeAll(tmpPartitionIds);
+            Collection<Integer> queriedPartitionIds = queryResult.getPartitionIds();
+            if (queriedPartitionIds != null) {
+                if (!partitionIds.containsAll(queriedPartitionIds)) {
+                    // results for at least one partition have already been added, so discard these results
+                    // see https://github.com/hazelcast/hazelcast/issues/6471
+                    continue;
+                }
+                partitionIds.removeAll(queriedPartitionIds);
                 for (QueryResultRow row : queryResult.getRows()) {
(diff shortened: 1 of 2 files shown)


vbekiaris:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -526,6 +526,11 @@ protected void addResultsOfPagingPredicate(List<Future<QueryResult>> futures, Co
 
             Collection<Integer> tmpPartitionIds = queryResult.getPartitionIds();
```
tombujok: Maybe queryPartitions? tmpPartitions is kind of generic

hazelcast/src/test/java/com/hazelcast/map/impl/query/QueryIndexMigrationTest.java
```diff
@@ -160,7 +160,7 @@ public void testQueryWithIndexesWhileMigrating() throws Exception {
     /**
      * test for issue #359
      */
-    @Test(timeout = MINUTE)
+    @Test
```
tombujok: That may be risky, maybe let's set 2 minutes?

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
vbekiaris: Yes, that's the variable's name in previous code but it also struck me as not-fitting so will fix it.

hazelcast/src/test/java/com/hazelcast/map/impl/query/QueryIndexMigrationTest.java
vbekiaris: I thought 2 minutes is the default timeout in Hazelcast test runners, or is this not the case?

vbekiaris: After all, this is not the case, added 2 minutes timeout to test definition. Thanks for pointing that out!

tombujok:
:+1: looks good!

ahmetmircik:
👍

PR #8572, opened 2016-07-22
Added tests to durable executor service for full ring-buffer case and newly joined member case

gurbuzali:


hazelcast-client/src/test/java/com/hazelcast/client/executor/durable/ClientDurableExecutorServiceTest.java
```diff
@@ -188,6 +188,30 @@ public void onFailure(Throwable t) {
...
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+                if (t.getCause() instanceof RejectedExecutionException) {
+                    latch.countDown();
+                }
+            }
+        });
+        assertOpenEventually(latch, 10);
```
mdogan: `assertOpenEventually(latch, 10)` can fail under high load, in slow environment. I think better is to just leave default timeout...
`assertOpenEventually(latch)`

mdogan:
👍

hazelcast/src/main/java/com/hazelcast/durableexecutor/impl/operations/RetrieveAndDisposeResultOperation.java
```diff
@@ -26,9 +26,6 @@
 
     private transient Object result;
 
-    public RetrieveAndDisposeResultOperation() {
```
jerrinot: is this operation always running locally? (=it is never (de-)serialized?)

hazelcast-client/src/test/java/com/hazelcast/client/executor/durable/ClientDurableExecutorServiceTest.java
gurbuzali: ok, I will fix it after merge

hazelcast/src/main/java/com/hazelcast/durableexecutor/impl/operations/RetrieveAndDisposeResultOperation.java
gurbuzali: no it can be a remote operation

jerrinot: so shouldn't it have an argument-less contructor?

gurbuzali: it extends `DisposeResultOperation` which has an argument-less contructor

jerrinot:
:+1:

PR #8577, opened 2016-07-22
Backport #7939 - Nonblocking client invocationfutures

ihsandemir:
Backports #7939 client related parts. This will eliminate the non-deterministic object.wait(timeout) problem as well.


pveentjer:
Closing this one. 

I thought we agreed that we should not introduce such risk in a maintenance release.

PR #8580, opened 2016-07-25
Deprecates AsyncAtomicReference. Introduces async methods in IAtomicReference public API.

vbekiaris:
Async methods in `IAtomicReference` follow the `Async`-suffix naming convention; `async`-prefixed methods will be removed once `AsyncAtomicReference` is also removed.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/IAtomicReference.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/IAtomicReference.java
@@ -177,15 +177,6 @@ public interface IAtomicReference<E> extends DistributedObject {
     ICompletableFuture<E> getAndSetAsync(E newValue);
 
     /**
-     * Sets and gets the value.
-     *
-     * @param update the new value
-     * @return  the new value
-     * @deprecated will be removed from Hazelcast 3.4 since it doesn't really serve a purpose.
-     */
-    ICompletableFuture<E> setAndGetAsync(E update);
-
-    /**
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/core/IAtomicReference.java
```diff
@@ -122,4 +142,101 @@
...
+     * @return the old value.
+     */
+    ICompletableFuture<E> getAndSetAsync(E newValue);
+
+    /**
+     * Sets and gets the value.
+     *
+     * @param update the new value
+     * @return  the new value
+     * @deprecated will be removed from Hazelcast 3.4 since it doesn't really serve a purpose.
```
jerrinot: I appreciate it's a copy-n-paste but the note about removal in 3.4 still looks weird.

vbekiaris: Agreed. Should we just remove it from this PR so it doesn't make it to the public API?

jerrinot: yes

vbekiaris: It is implemented with a get-and-set operation and returns the user-supplied `newValue`, so I guess it's not a terribly useful method. /cc @pveentjer ?

pveentjer:
👍

tombujok:
👍

PR #8581, opened 2016-07-25
Heartbeat API, test and logging improvements

cangencer:
- Rename usages of heartBeat/HeartBeat to heartbeat/Heartbeat
- Change heartBeatStopped event to heartbeatResumed for better semantics
- Add logging for last heartbeat received/requested for better diagnostics
- Improve heartbeat tests to cover missing cases.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
@@ -32,6 +32,7 @@ import com.hazelcast.nio.Protocols;
 import com.hazelcast.nio.tcp.SocketChannelWrapper;
 import com.hazelcast.nio.tcp.nonblocking.NonBlockingIOThread;
 import com.hazelcast.spi.properties.GroupProperty;
+import com.hazelcast.util.Clock;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.io.IOException;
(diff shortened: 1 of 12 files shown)


mustafaiman:
👍

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
```diff
@@ -269,21 +271,32 @@ public String getCloseReason() {
...
-    void heartBeatingSucceed() {
-        heartBeating = true;
-        lastHeartbeatMillis = System.currentTimeMillis();
+    void heartbeatResumed() {
+        isHeartBeating = true;
     }
 
-    public long getLastHeartbeatMillis() {
-        return lastHeartbeatMillis;
+    void heartbeatReceived() {
```
pveentjer: this name doesn't explain if it is an 'getter' or 'do something'. Probably using 'onHeartBeatFailed' would be more descriptive.

pveentjer:
Looks good. But this should be moved to 3.8 I guess since we are at the point of closing the 3.7 release. @jerrinot wdyt

jerrinot:
given the nature of changes I am fine with this even in 3.7. 
:+1:

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
cangencer: I've renamed them to "onHeartbeat.." to indicate they're event handlers.

pveentjer:
👍 

merge once green

Donnerbart:
:+1:

PR #8586, opened 2016-07-25
Do not make assertions on messages generated by JDK

### LINKED ISSUE #8169: InvalidConfigurationTest depends on system locale setting

jerrinot:
Fix #8169

Also added `testWhenXmlValidAndPropertiesAreResolved()`
to make sure the basic config skeleton is right.


Donnerbart:
:+1: (once green)

tombujok:
👍

jerrinot:
run-lab-run

PR #8596, opened 2016-07-25
stabilize hazelcast client protocol version at most recent version 1.2.0

mustafaiman:
Should be green once client protocol 1.2 is released.


ihsandemir:
👍  OK, after 1.2.0 is released and it is green.

mustafaiman:
run-lab-run

ihsandemir:
👍

Issue #8593, opened 2016-07-25
Under high load and high ICache.getAsync calls near-cache updates may cause OOM


ihsandemir:
Client updates ICache near cache using the ClientExecutionService userExecutor in an async way for getAsync API. The queue for the userExecutor is unbounded. If the userExecutor threads are slower in consuming these tasks in the executors queue, the queue may grow and this may cause OutOfMemory error. The queue items has reference to Invocation, Request ClientMessage and Response ClientMessage and these may be big objects.

sancar:
Fixed via https://github.com/hazelcast/hazelcast/pull/8649

PR #8602, opened 2016-07-25
Added the near cache executor overloaded property and implementation.

ihsandemir:
Adds the near cache executor overloaded property and implementation. 

Fixes https://github.com/hazelcast/hazelcast/issues/8593


pveentjer:
👎 

The problem with this approach is that as soon as the is a lot of pressure on the executor, you will effectively clean out the near cache. So when the system is under pressure, the performance is made worse because every call is going to be a remote call. This is not a desired situation as we discussed this afternoon.

ihsandemir:
Since maintenance branch changed, this PR is replaced by PR #8669 .

PR #8608, opened 2016-07-26
Locale specific string in MutliLineDiagnosticsLogWriter

vbekiaris:
MutliLineDiagnosticsLogWriter test fails in el_GR.UTF-8 locale (and I guess other locales as well), because MutliLineDiagnosticsLogWriter.STR_LONG_MIN_VALUE is formatted with default platform locale and does not match expected value in MutliLineDiagnosticsLogWriterTest:116

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/diagnostics/MultiLineDiagnosticsLogWriter.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/diagnostics/MultiLineDiagnosticsLogWriter.java
@@ -17,13 +17,14 @@
 package com.hazelcast.internal.diagnostics;
 
 import static com.hazelcast.util.StringUtil.LINE_SEPARATOR;
+import static com.hazelcast.util.StringUtil.LOCALE_INTERNAL;
 
 /**
  * A {@link DiagnosticsLogWriter} that writes over multiple lines. Useful for human reading.
  */
 class MultiLineDiagnosticsLogWriter extends DiagnosticsLogWriter {
 
-    private static final String STR_LONG_MIN_VALUE = String.format("%,d", Long.MIN_VALUE);
+    private static final String STR_LONG_MIN_VALUE = String.format(LOCALE_INTERNAL, "%,d", Long.MIN_VALUE);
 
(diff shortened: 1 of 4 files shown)


tombujok:
👍

hazelcast/src/main/java/com/hazelcast/internal/diagnostics/MultiLineDiagnosticsLogWriter.java
```diff
@@ -16,14 +16,16 @@
...
+
 import static com.hazelcast.util.StringUtil.LINE_SEPARATOR;
 
 /**
  * A {@link DiagnosticsLogWriter} that writes over multiple lines. Useful for human reading.
  */
 class MultiLineDiagnosticsLogWriter extends DiagnosticsLogWriter {
 
-    private static final String STR_LONG_MIN_VALUE = String.format("%,d", Long.MIN_VALUE);
+    private static final String STR_LONG_MIN_VALUE = String.format(Locale.US, "%,d", Long.MIN_VALUE);
```
ahmetmircik: This is also fine but as a minor comment, we have [this one] (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/util/StringUtil.java#L40) to get locale for internal usage. It may be better to get internal-locale from one place.

ahmetmircik:
👍 except one minor comment

hazelcast/src/main/java/com/hazelcast/internal/diagnostics/MultiLineDiagnosticsLogWriter.java
vbekiaris: 👍 amended to use `StringUtil.LOCALE_INTERNAL`

vbekiaris:
I updated `LOCALE_INTERNAL = Locale.US` and reused `LOCALE_INTERNAL` in a couple more places where `Locale.US` was directly referenced. In terms of string methods `toLower/UpperCase` this should be the same as `Locale.ENGLISH` (the language component is the same as previously). @tombujok @ahmetmircik if you are OK with this update, let me know so we can merge.

PR #8565, opened 2016-07-21
Ensure migration thread dies during shutdown

mtopolnik:
As described in #8560, the migration thread is not joined after interrupting, leading to a possible (and actually observed) thread leak. This PR ensures the thread is dead before proceeding with the shutdown procedure.

This PR also includes some non-related code style cleanup in collection classes.

PR #8610, opened 2016-07-27
Fix hanging invocation during shutdown

### LINKED ISSUE #8560: MigrationThread leak
`MigrationThread` can leak after shutting down or even forcefully terminating Hazelcast. It retains the entire `HazelcastInstance`.

`InternalPartitionServiceImpl`:
```java
    @Override
    public void shutdown(boolean terminate) {
        logger.finest("Shutting down the partition service");
        migrationManager.stop();
        reset();
    }
```
`MigrationManager`:
```java
void stop() {
        migrationThread.stopNow();
    }
```
`MigrationThread`:
```java
    void stopNow() {
        queue.clear();
        interrupt();
    }
```
Thread is interrupted, but not joined. In particular, the thread will block forever in some cases, one of which is the following:
`RepairPartitionTableTask#commitPromotionsToDestination`:
```java
                PromotionCommitOperation op = new PromotionCommitOperation(...);
                Future<Boolean> future = nodeEngine.getOperationService()
                        .createInvocationBuilder(SERVICE_NAME, op, destination)
                        .setTryCount(Integer.MAX_VALUE)
                        .setCallTimeout(Long.MAX_VALUE).invoke();
                boolean result = future.get();
```

mdogan:
When an invocation is submitted or retried during node shutdown,
that invocation may not be notified with `HazelcastInstanceNotActiveException`
as expected. Because, there's no mechanism that prevents registering a new invocation
after `InvocationRegistry` is shutdown.

Also, `OperationService` (and all invocation infrastructure) should be shutdown before
any other service, since a service like `InternalPartitionService` may depend on an invocation
to complete (with either success or failure) during shutdown.

Another problem is, currently as an optimization, a local invocation without backup
is not registered in `InvocationRegistry`. That means, we don't have a reference to
that invocation during `OperationService` shutdown and we can't notify/complete it `HazelcastInstanceNotActiveException`.
Since, that invocation won't get notified, it will eventually timeout with `OperationTimeoutException`.
Worse scenario is, that invocation will hang forever if an infinite timeout is given,
like `PromotionCommitOperation` or `MigrationCommitOperation` in migration system.
A user invocation, like `map.get()`, can get an `OperationTimeoutException` if it's invoked
during shutdown.

To fix these issues

- An `alive` flag is introduced in `InvocationRegistry` which is set to false during shutdown.
When a new invocation is requested to be registered, `InvocationRegistry.register()` will return
value of `alive` flag. So, requester should stop there and complete invocation with failure.

- `OperationService.shutdown()` is shifted up in shutdown sequence.

- Optimization of skipping registration of a backup-less & local invocations is removed.

Fixes #8560
Cherry picked @mtopolnik's commit from #8565

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationThread.java
@@ -119,10 +119,28 @@ class MigrationThread extends Thread implements Runnable {
         return activeTask;
     }
 
+    /**
+     * Interrupts the migration thread and joins on it.
+     * <strong>Must not be called on the migration thread itself</strong> because it will result in infinite blocking.
+     */
     void stopNow() {
+        assert currentThread() != this : "stopNow must not be called on the migration thread";
         running = false;
         queue.clear();
         interrupt();
+        boolean currentThreadInterrupted = false;
+        while (true) {
+            try {
+                join();
+            } catch (InterruptedException e) {
+                currentThreadInterrupted = true;
+                continue;
+            }
+            break;
+        }
+        if (currentThreadInterrupted) {
+            currentThread().interrupt();
+        }
     }
 
 }
--- a/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationThread.java
@@ -39,6 +39,7 @@ class MigrationThread extends Thread implements Runnable {
     private final long sleepTime;
 
     private volatile MigrationRunnable activeTask;
+    private volatile boolean running = true;
 
     MigrationThread(MigrationManager migrationManager, HazelcastThreadGroup hazelcastThreadGroup, ILogger logger,
                     MigrationQueue queue) {
@@ -54,7 +55,7 @@ class MigrationThread extends Thread implements Runnable {
     @Override
     public void run() {
         try {
-            while (!isInterrupted()) {
+            while (running) {
                 doRun();
             }
         } catch (InterruptedException e) {
@@ -98,7 +99,7 @@ class MigrationThread extends Thread implements Runnable {
 
     private boolean processTask(MigrationRunnable runnable) {
         try {
-            if (runnable == null || isInterrupted()) {
+            if (runnable == null || !running) {
                 return false;
             }
 
@@ -119,6 +120,7 @@ class MigrationThread extends Thread implements Runnable {
     }
 
     void stopNow() {
+        running = false;
         queue.clear();
         interrupt();
     }
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
@@ -426,10 +426,10 @@ public class NodeEngineImpl implements NodeEngine {
     public void shutdown(final boolean terminate) {
         logger.finest("Shutting down services...");
         waitNotifyService.shutdown();
+        operationService.shutdown();
         proxyService.shutdown();
         serviceManager.shutdown(terminate);
         eventService.shutdown();
-        operationService.shutdown();
         wanReplicationService.shutdown();
         executionService.shutdown();
         metricsRegistry.shutdown();
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -31,6 +31,7 @@ import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
 import com.hazelcast.nio.ConnectionManager;
 import com.hazelcast.partition.NoDataMemberInClusterException;
+import com.hazelcast.spi.BackupAwareOperation;
 import com.hazelcast.spi.BlockingOperation;
 import com.hazelcast.spi.ExceptionAction;
 import com.hazelcast.spi.NodeEngine;
@@ -52,6 +53,7 @@ import com.hazelcast.util.Clock;
 import com.hazelcast.util.executor.ManagedExecutorService;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 import java.util.logging.Level;
@@ -256,7 +258,10 @@ public abstract class Invocation implements OperationResponseHandler {
         }
 
         setInvocationTime(op, context.clusterClock.getClusterTime());
-        context.invocationRegistry.register(this);
+        if (!context.invocationRegistry.register(this)) {
+            return;
+        }
+
         if (remote) {
             doInvokeRemote();
         } else {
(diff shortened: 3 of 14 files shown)


mtopolnik:
👍

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -59,6 +59,8 @@
     private final ILogger logger;
     private final CallIdSequence callIdSequence;
 
+    private final AtomicBoolean alive = new AtomicBoolean(true);
```
jerrinot: why not just `volatile boolean` ?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -487,11 +493,15 @@ private void handleRetry(Object cause) {
...
-                context.executionService.schedule(ASYNC_EXECUTOR, new InvocationRetryTask(), tryPauseMillis, MILLISECONDS);
+            try {
+                if (invokeCount < MAX_FAST_INVOCATION_COUNT) {
+                    // fast retry for the first few invocations
+                    context.asyncExecutor.execute(new InvocationRetryTask());
+                } else {
+                    context.executionService.schedule(ASYNC_EXECUTOR, new InvocationRetryTask(), tryPauseMillis, MILLISECONDS);
+                }
+            } catch (RejectedExecutionException e) {
+                complete(e);
```
jerrinot: do we want to pass this `RejectedExecutionException` to the user?

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationThread.java
```diff
@@ -118,9 +118,27 @@ MigrationRunnable getActiveTask() {
...
+    /**
+     * Interrupts the migration thread and joins on it.
+     * <strong>Must not be called on the migration thread itself</strong> because it will result in infinite blocking.
+     */
     void stopNow() {
+        assert currentThread() != this : "stopNow must not be called on the migration thread";
         queue.clear();
         interrupt();
+        boolean currentThreadInterrupted = false;
+        while (true) {
```
jerrinot: the `MigrationThread` is running "alien" tasks, isn't it? are you sure a task will never ever swallow the interrupt flag?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -89,18 +91,17 @@ long getLastCallId() {
      * Registers an invocation.
      *
      * @param invocation The invocation to register.
+     * @return true if InvocationRegistry is alive when invocation is registered, false otherwise
```
jerrinot: this slightly misleading as technically the invocation is always registered. 
perhaps there is not difference for practical purposes, feel free to ignore this comment.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationThread.java
mtopolnik: Actually it's not, they are from a strictly controlled set of `MigrationRunnable`s. However, given our overall interruptibility status this might be fragile.

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -425,11 +425,11 @@ public void reset() {
 
     public void shutdown(final boolean terminate) {
         logger.finest("Shutting down services...");
+        operationService.shutdown();
```
jerrinot: shutdown ordering changes like this are always somewhat frightening. for example the wait-notify service uses operation-service during shutdown. I assume you know what you are doing ;-)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
mdogan: just preference. `volatile boolean` is fine too.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
mdogan: no, it'll be wrapped by `ExecutionException`.

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
mdogan: I can't see how `WaitNotifyService.shutdown()` uses `OperationService`.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
jerrinot: I see. wouldn't be  `HazelcastNotActiveException` more suitable*? It says what's going on, what's the cause. 

*assuming it can happen only during shutdown.

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
jerrinot: it calls `responseHandler.sendResponse()`. 
The remote response handler impl uses operation service. Currently it's not a problem because the response is sent only for [local invocations](https://github.com/jerrinot/hazelcast/blob/891660f3874acfcfe8cd7c56c73fb7c7ec36e8dd/hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java#L267-L267). but it still feels a bit fragile. 

that being said I don't know any better:)

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
mdogan: you're right, it creates a perception like invocation is not registered when false is returned. 
Maybe we can throw exception instead of returning boolean.

jerrinot: another idea: couldn't the `register()` method itself notify the invocation with the `HazelcastInstanceNotActiveException`  - It's the responsibility of the `InvocationRegistry` to notify invocations on shutdown anyway. 

Then you can simply change the contract to "returns false when InvocationRegistry is not alive and registration is not successful. "

mdogan: make sense.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
jerrinot: ok. I prefer primitive types where possible - no need for extra indirection.

mdogan: ok

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
mdogan: sure. changing to `complete(new HazelcastInstanceNotActiveException(e.getMessage()))`

jerrinot: maybe the original exception could go to `finest` / `trace` log. to have stacktraces available for troubleshooting.

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -426,10 +426,10 @@ public void reset() {
     public void shutdown(final boolean terminate) {
         logger.finest("Shutting down services...");
         waitNotifyService.shutdown();
+        operationService.shutdown();
```
ahmetmircik: After this shift some enterprise-side tests start to throw time-out exceptions. https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-EE-pr/954/console

An example cause is [here](https://github.com/hazelcast/hazelcast-enterprise/blob/master/hazelcast-enterprise/src/main/java/com/hazelcast/map/impl/EnterpriseMapServiceContextImpl.java#L257) an operation is added after operationService shutdown. 

@mdogan what can be done here?

mdogan: If we separate shutdown of `InvocationRegistry` and `OperationService`, I think both problems will be solved. 

See commit: https://github.com/mdogan/hazelcast/commit/1ef4f4081604cc566de435e31a50c69a65ffeb49

ahmetmircik: Thanks, rerunned failing tests, they worked fine with that fix. 

Maybe one another thing to prevent unintended usage of operationService and to make that more visible, operationService can throw an exception if one tries to use it after its' shutdown. I will create an issue for this.

mdogan: https://github.com/hazelcast/hazelcast/pull/8630

PR #8421, opened 2016-06-20
fixed multicast plugin bug and added test

bilalyasar:
- added javadoc
- fixed small bug ( discovery property bug)
- added test

PR #8614, opened 2016-07-28
fixed multicast plugin bug and added test

bilalyasar:
I have rebased and opened new pr from this pr:
https://github.com/hazelcast/hazelcast/pull/8421

- added javadoc
- fixed small bug ( discovery property bug, invalid port check)
- added test

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/spi/discovery/multicast/MemberToMemberDiscoveryTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/spi/discovery/multicast/MemberToMemberDiscoveryTest.java
@@ -18,6 +18,7 @@ package com.hazelcast.spi.discovery.multicast;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
+import com.hazelcast.config.properties.ValidationException;
 import com.hazelcast.core.HazelcastInstance;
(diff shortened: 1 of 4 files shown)


hazelcast/src/test/java/com/hazelcast/spi/discovery/multicast/MemberToMemberDiscoveryTest.java
```diff
@@ -47,7 +48,9 @@ public void setUp() {
 
         System.setProperty(TestEnvironment.HAZELCAST_TEST_USE_NETWORK, "true");
         System.setProperty("java.net.preferIPv4Stack", "true");
-        factory = createHazelcastInstanceFactory(2);
+        if (factory == null) {
```
mesutcelik: Do we really need this? `factory` is already defined in HazelcastTestSupport so all cleanup, null checking etc. looks unnecessary to me. 

I believe we should remove `factory` variable in this class and depend on the one @ HazelcastTestSupport. Good cleanup opportunity for you 👍

mesutcelik:
👍 with minor comment.

Can you please add bug description as well? invalid port fix etc.

bilalyasar:
@mesutcelik reviewed your comments, also added description.

Donnerbart:
:+1:

PR #8616, opened 2016-07-28
Do not pick the node's own address to join in MockJoiner

metanet:
A node registers itself into MockJoiner registry and scans the registry to pick a node to send a join request. It should avoid picking itself.


jerrinot:
:+1:

Donnerbart:
:+1: (once green)

Donnerbart:
run-lab-run

mdogan:
run-lab-run

PR #8623, opened 2016-07-29
[DONT MERGE] Fixed a performance regression in IMap.putAll()

Donnerbart:
Fixed a performance regression in `IMap.putAll()` due to copying the whole data set, which occurred on the client code path.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -124,7 +124,6 @@ import com.hazelcast.query.Predicate;
 import com.hazelcast.spi.impl.UnmodifiableLazyList;
 import com.hazelcast.spi.serialization.SerializationService;
 import com.hazelcast.util.CollectionUtil;
-import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.IterationType;
 import com.hazelcast.util.MapUtil;
 import com.hazelcast.util.Preconditions;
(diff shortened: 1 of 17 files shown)


jerrinot:
@Donnerbart: this PR is aiming to improve performance yet the PR does not present any performance numbers/benchmarks. I'm moving this into 3.7.1.

Donnerbart:
Danny made some benchmarks, but I have a bit trouble of interpreting them (also they were run against 3.7-EA, not 3.6.4).

I'm running some Simulator benchmarks with a similar configuration with 3.6.4, 3.7-SNAPSHOT and this PR at the moment.

Donnerbart:
I cannot really see a performance degradation between `3.6.4` and the `3.7-SNAPSHOT`. There is a single run which is a bit slower, but all others are performing way better:
image
Also this FIX PR doesn't seem to improve the situation.

jerrinot:
@Donnerbart: wasn't the degradation @Danny-Hazelcast measured between 3.7-EA and 3.7-SNAPSHOT?

Donnerbart:
I think so, but `3.7-EA` is not a new major release, it's just an RC. So the important version for me is `3.6.4`. But I can also try a run with `3.7-EA` and my setup.

Donnerbart:
I can't confirm the numbers for `3.7-EA` either, looks quite consistent with the other `3.7` runs.
image

[putAll.zip](https://github.com/hazelcast/hazelcast/files/397024/putAll.zip)

Donnerbart:
Okay, the bad throughput seems to vanish with a higher test duration. The used Simulator version still lacks the warmup feature, so we always have a slower `RUN` phase start. This hurts of course more in "high ops/s scenarios". With a longer `RUN` phase we can level this out:
image

Donnerbart:
I had a look at the benchmark code Danny used, and it creates the input maps on the fly (like Yardstick). So the faster Hazelcast is, the more load and litter the benchmark generates by itself:
https://github.com/Danny-Hazelcast/hzCmd-bench/blob/master/src/main/java/hz/map/PutAll.java

We have a similar variant of this test in Simulator, since I wanted to compare it to Yardstick. And with this the overall throughput is lower and we have a dropdown of about 4% with `3.7-SNAPSHOT`.
image

PR #8624, opened 2016-07-29
Fixed a possible NPE in PartitionWideEntryWithPredicateOperationFactory

Donnerbart:
Fixed a possible NPE in `PartitionWideEntryWithPredicateOperationFactory` on bouncing members.

BTW: There is still no code coverage on the serialization part of `MultipleEntryWithPredicateOperation`, even with the bouncing test.

PR #8626, opened 2016-07-29
Fixed IllegalStateException in PutAllPartitionAwareOperationFactory

Donnerbart:
Fixed `IllegalStateException` in `PutAllPartitionAwareOperationFactory` on bouncing members, without creating thread-safety issues in `PartitionWideEntryWithPredicateOperationFactory`.

This solution is a bit more hacky on the first view, since we make an exception for the `PutAllPartitionAwareOperationFactory`. But @jerrinot pointed out, that the other solution (https://github.com/hazelcast/hazelcast/pull/8624) has at least visibility issues between threads which modify the state of the `PartitionWideEntryWithPredicateOperationFactory`. So this solution is more safe.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvokeOnPartitions.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvokeOnPartitions.java
@@ -16,11 +16,11 @@
 
 package com.hazelcast.spi.impl.operationservice.impl;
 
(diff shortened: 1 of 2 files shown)


jerrinot:
:+1: although this is still a workaround rather than a proper solution.

Donnerbart:
Yeah, I think it's too late for a proper solution for the 3.7 release :(

Donnerbart:
Closing this one in favor of #8627.

PR #8630, opened 2016-08-01
Split OperationService.shutdown into two separate methods

mdogan:
Before `OperationService.shutdown()` was shutting down invocations
and operation executors at once.
In a previous commit, `OperationService.shutdown()` call was shifted up
in shutdown sequence to be able notify all pending invocations while
services were shutting down.
But on EE side, some services are submitting local tasks/operations during
service shutdown (like cleanup tasks) and they started to fail after that change.

As a fix, `OperationService.shutdown()` is split into two methods; first one
will shutdown only invocations (invocation infrastructure), second one will shutdown
operation executors (and operation threads etc).

See PR https://github.com/hazelcast/hazelcast/pull/8610


metanet:
👍

ahmetmircik:
👍

PR #8631, opened 2016-08-01
Implement MemcacheEntry.getKey

LoneRifle:
zalora/zcast@6aedda5 implements a MapInterceptor that partially compresses/decompresses MemcachedEntry payloads on the fly using LZ4 during IMap.get/put operations. It does this by extracting the fields from an intercepted MemcachedEntry and creating a new one with a compressed value and extra lz4 metadata added to flags.

One of the challenges implementing this MapInterceptor is extracting the key from the original MemcachedEntry. @wolframite has hence requested that the key is made available via a getter, which is implemented in this PR. The first and last bytes of the key are determined respectively using dead reckoning, and locating the space character between the key and flags.


noctarius:
Looks good

PR #8632, opened 2016-08-01
TextCommandServiceImpl.ResponseThreadRunnable - Convert to enum

LoneRifle:
Double-checked locking for lazy-init singletons has concurrency problems that require workarounds to ensure that they work [1]. These workarounds tend to add additional overhead - In the case of TextCommandServiceImpl, the field holding the singleton has been made volatile, which is an additional synchronization barrier on top of the synchronized blocks in place in processRequest().

Josh Bloch has long recommended the use of enums as lazy-init singletons [2] as the init synchronization can be easily and effectively handled by the class loader. This PR serves to refactor the ResponseThreadRunnable singleton as an enum, thereby allowing us to avoid the use of double-checked locking.

References:
[1] http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html
[2] http://www.drdobbs.com/jvm/creating-and-destroying-java-objects-par/208403883?pgno=3

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/TextCommandServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/TextCommandServiceImpl.java
@@ -92,8 +92,6 @@ public class TextCommandServiceImpl implements TextCommandService {
     private final AtomicLong decrementMisses = new AtomicLong();
     private final long startTime = Clock.currentTimeMillis();
     private final ILogger logger;
-    private volatile ResponseThreadRunnable responseThreadRunnable;
-    private volatile boolean running = true;
 
     public TextCommandServiceImpl(Node node) {
         this.node = node;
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/internal/ascii/TextCommandServiceImpl.java
```diff
@@ -207,18 +205,11 @@ public long incrementTouchCount() {
...
-                            hazelcastThreadGroup.getInternalThreadGroup(), responseThreadRunnable, threadNamePrefix);
-                    thread.start();
-                }
-            }
-        }
+        HazelcastThreadGroup hazelcastThreadGroup = node.getHazelcastThreadGroup();
+        String threadNamePrefix = hazelcastThreadGroup.getThreadNamePrefix("ascii.service.response");
+        Thread thread = new Thread(
+                hazelcastThreadGroup.getInternalThreadGroup(), ResponseThreadRunnable.INSTANCE, threadNamePrefix);
+        thread.start();
```
LoneRifle: This isn't right - a thread will be spawned with each incoming request. This block ought to be moved into the enum ctor.

LoneRifle:
This will not work at all as we require one ResponseThreadRunnable per Node, not one for all Nodes. Double-checked locking is so far the only way to accomplish lazy-init on a per-instance basis.

PR #8633, opened 2016-08-01
fixed small member attribute bug

bilalyasar:
MemberAttributes are read from Config in [`Node.java` constructor..](https://github.com/hazelcast/hazelcast/blob/c0dd2786291d0a88f54879097d5b3415b107fd40/hazelcast/src/main/java/com/hazelcast/instance/Node.java#L170)  So we need to use `localMember.getAttribute` when we want to access member attributes. Because it can be updated later, so we shouldn't rely old config further.


mdogan:
👍

ahmetmircik:
@bilalyasar seems maintenance also needs this fix: https://github.com/hazelcast/hazelcast/blob/maintenance-3.x/hazelcast/src/main/java/com/hazelcast/instance/Node.java#L563

PR #8639, opened 2016-08-01
Logging exception instead of swallowing it in ExtractorHelper

tombujok:


jerrinot:
:+1: can be merged on green

PR #8649, opened 2016-08-02
Too many client cache Invalidation callbacks for a single ICache.clear call

ihsandemir:
Fix for decreasing the cache clear callbacks for the client for a single clear request. Current implementation can cause up to partition count callbacks for a single clear request. One ICache.clear call causes up to partitioncount callbacks to clientcacheinvalidation listener and the same number of near cache clears. This may really be a high number of callbacks and too many cache invalidations if the partition count is a high number. 

The side affect of eliminating the can invalidation event sending during cache store clear needs to be investigated since AbstractCacheRecordStore.clear is being called from a few other places which may have been affected.

Related to issue #8593

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
@@ -72,8 +72,8 @@ import static com.hazelcast.cache.impl.record.CacheRecordFactory.isExpiredAt;
 public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extends SampleableCacheRecordMap<Data, R>>
         implements ICacheRecordStore, EvictionListener<Data, R> {
 
+    public static final String SOURCE_NOT_AVAILABLE = "<NA>";
     protected static final int DEFAULT_INITIAL_CAPACITY = 256;
-    protected static final String SOURCE_NOT_AVAILABLE = "<NA>";
 
     protected final String name;
     protected final int partitionId;
@@ -1221,7 +1221,6 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
         R record = records.get(key);
         int hitCount = 0;
         boolean removed = false;
-
         try {
             if (recordNotExistOrExpired(record, now)) {
                 if (isStatisticsEnabled()) {
@@ -1278,7 +1277,6 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
         R record = records.get(key);
         Object obj;
         boolean removed = false;
-
         try {
             if (recordNotExistOrExpired(record, now)) {
                 obj = null;
@@ -1455,11 +1453,6 @@ public abstract class AbstractCacheRecordStore<R extends CacheRecord, CRM extend
     @Override
     public void clear() {
         records.clear();
-        onClear();
-    }
-
-    protected void onClear() {
-        invalidateAllEntries();
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractInternalCacheProxy.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractInternalCacheProxy.java
@@ -72,12 +72,13 @@ abstract class AbstractInternalCacheProxy<K, V>
         implements ICacheInternal<K, V>, CacheSyncListenerCompleter {
 
     private static final long MAX_COMPLETION_LATCH_WAIT_TIME = TimeUnit.MINUTES.toMillis(5);
+
     private static final long COMPLETION_LATCH_WAIT_TIME_STEP = TimeUnit.SECONDS.toMillis(1);
-
     private final ConcurrentMap<CacheEntryListenerConfiguration, String> asyncListenerRegistrations;
-    private final ConcurrentMap<CacheEntryListenerConfiguration, String> syncListenerRegistrations;
 
+    private final ConcurrentMap<CacheEntryListenerConfiguration, String> syncListenerRegistrations;
     private final ConcurrentMap<Integer, CountDownLatch> syncLocks;
+
     private final AtomicInteger completionIdCounter = new AtomicInteger();
 
     private HazelcastServerCacheManager cacheManager;
@@ -240,6 +241,9 @@ abstract class AbstractInternalCacheProxy<K, V>
             }
         } catch (Throwable t) {
             throw ExceptionUtil.rethrowAllowedTypeFirst(t, CacheException.class);
+        } finally {
+            // send single invalidation event
+            cacheService.sendInvalidationEvent(nameWithPrefix, null, AbstractCacheRecordStore.SOURCE_NOT_AVAILABLE);
         }
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheSplitBrainHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheSplitBrainHandler.java
@@ -43,7 +43,6 @@ import java.util.concurrent.TimeUnit;
  * Handles split-brain functionality for cache.
  */
 class CacheSplitBrainHandler {
-
     private final NodeEngine nodeEngine;
     private final Map<String, CacheConfig> configs;
     private final CachePartitionSegment[] segments;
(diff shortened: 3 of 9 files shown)


jerrinot:
won't this completely disable invalidations for member-to-member topologies?

mdogan:
What if `cache.clear()` is called from server-side? `CacheClearMessageTask` won't be used at all.

jerrinot:
There are invalidation-related failures as expected.
However I'm surprised the failure are in client tests only?

https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/4037/

```
com.hazelcast.client.cache.nearcache.ClientNearCacheTest.putToCacheAndClearOrDestroyThenCantGetAnyRecordFromClientNearCacheWithBinaryInMemoryFormat
com.hazelcast.client.cache.nearcache.ClientNearCacheTest.putToCacheAndClearOrDestroyThenCantGetAnyRecordFromClientNearCacheWithObjectInMemoryFormat
```

vbekiaris:
Member-side near cache for JCache is not implemented yet.

mdogan:
Regardless of member-size near cache, if `cache.clear()` is called on member-side, clients won't get invalidation messages with this change.

vbekiaris:
Yes, I agree. It was just meant as a clarification on @jerrinot 's comment about client-only failures.

ihsandemir:
@mdogan @jerrinot @vbekiaris I implemented the changes for the server side as well. I also added tests to verify that the received event count is really only one as expected for a clear request. I also merged with the new maintenance 3.7.1.

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
```diff
@@ -1455,22 +1455,22 @@ public String getName() {
     @Override
     public void clear() {
         records.clear();
-        onClear();
     }
 
-    protected void onClear() {
+    public void clearAndSendInvalidation() {
```
vbekiaris: Maybe this method ought to be part of the `ICacheRecordStore` interface? Please also check this in light of `clear` method usage on EE side.

vbekiaris:
I think the PR looks good; please also check for changes required on EE side.

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
ihsandemir: I was reluctant to make change to the interface, but as it is not in SPI and inside impl package I will do this.

hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
```diff
@@ -103,7 +103,8 @@ public void init() {
     public void clear() {
         synchronized (mutex) {
             for (ICacheRecordStore store : recordStores.values()) {
-                store.clear();
+                AbstractCacheRecordStore recordStore = (AbstractCacheRecordStore) store;
```
ahmetmircik: do we need this casting any more?

hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
```diff
@@ -121,7 +122,8 @@ void clearHavingLesserBackupCountThan(int backupCount) {
             for (ICacheRecordStore store : recordStores.values()) {
                 CacheConfig cacheConfig = store.getConfig();
                 if (backupCount > cacheConfig.getTotalBackupCount()) {
-                    store.clear();
+                    AbstractCacheRecordStore recordStore = (AbstractCacheRecordStore) store;
```
ahmetmircik: same here?

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheSplitBrainHandler.java
```diff
@@ -85,6 +86,11 @@ Runnable prepareMergeRunnable() {
                     }
                     // Clear all records either owned or backup
                     cacheRecordStore.clear();
+
+                    // send the cache invalidation event regardless if any actually cleared or not (no need to know how many
+                    // actually cleared)
+                    final CacheService cacheService = nodeEngine.getService(CacheService.SERVICE_NAME);
+                    cacheService.sendInvalidationEvent(cacheName, null, SOURCE_NOT_AVAILABLE);
```
ahmetmircik: why do we need this?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheClearMessageTask.java
```diff
@@ -73,6 +75,11 @@ protected Object reduce(Map<Integer, Object> map) {
                 throw (CacheException) response;
             }
         }
+
+        // send the cache invalidation event regardless if any actually cleared or not (no need to know how many actually cleared)
+        final CacheService cacheService = getService(CacheService.SERVICE_NAME);
+        cacheService.sendInvalidationEvent(getDistributedObjectName(), null, SOURCE_NOT_AVAILABLE);
```
ahmetmircik: seems this should be in `finally` block.

hazelcast/src/test/java/com/hazelcast/cache/CacheClearTest.java
```diff
@@ -165,4 +171,63 @@ public void testClear() {
...
+
+        cache.clear();
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run()
+                    throws Exception {
+                assertEquals(1, counter.get());
+            }
+        }, 2);
```
ahmetmircik: 2 seconds for `assertTrueEventually` timeout is very short? Does it make sense to use default timeout?

hazelcast/src/test/java/com/hazelcast/cache/CacheClearTest.java
```diff
@@ -165,4 +171,63 @@ public void testClear() {
...
+
+        // Make sure that the callback is not called for a while
+        assertTrueAllTheTime(new AssertTask() {
+            @Override
+            public void run()
+                    throws Exception {
+                assertTrue(counter.get() <= 1);
+            }
+        }, 3);
```
ahmetmircik: same here?

hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
ihsandemir: OK, you are right, changed it now.

hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
ihsandemir: OK

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheSplitBrainHandler.java
ihsandemir: I put it so that the old behaviour is not changed. I am not very knowledgable on split brain handling, if it is not needed or we can handle it in a different way, yes I can change it.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheClearMessageTask.java
ihsandemir: OK. Changed it.

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheSplitBrainHandler.java
ahmetmircik: i see thanks

hazelcast/src/test/java/com/hazelcast/cache/CacheClearTest.java
ihsandemir: The test passes with this timeout. The server mostly responds in a shorter time, hence it probably will not fail but yes, we can change to higher timeout. Or we can just wait to see if the test fails in the future due to this reason.

ahmetmircik:
If it is possible to squeeze commits, it would be better, otherwise LGTM 👍

vbekiaris:
👍 (merge once EE counterpart is also ready to merge)

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractInternalCacheProxy.java
```diff
@@ -113,6 +114,11 @@ void setCacheManager(HazelcastServerCacheManager cacheManager) {
     }
 
     @Override
+    protected boolean preDestroy() {
```
ahmetmircik: seems no need to override?

ihsandemir: OK, removed.

PR #8658, opened 2016-08-04
Remove a block of entries from suppressions.xml

mtopolnik:
At some places the CheckStyle complaint was fixed and at others a local `@SuppressWarnings` was added.

Also IntelliJ warnings have been honored about missing `@Override`, and redundant `/** @inheritDoc */` was removed.


Donnerbart:
:+1: (once green)

Nice cleanup!

PR #8659, opened 2016-08-04
Move concurrent array queues from EE Hot Restart module

mtopolnik:
1. Move the classes to OSS.
2. Catch up with Agrona's master, there were several performance touchups.
3. Remove the `Unsafe` dependency, instead code against `AtomicReferenceArray` and `AtomicLongFieldUpdater`.


Donnerbart:
I guess there is not much to review here, since it's 3rd party code.

Did you check your branch with FindBugs, so we don't get those nice build failure emails so close to the weekend?

If FindBugs is not complaning, it's a :+1: from me.

mtopolnik:
Interest for this code migrating to OSS is gone, so it stays in Hot Restart.

PR #8666, opened 2016-08-05
Use cluster time for operation heartbeat timeout checks

metanet:
Invocation is initiated with cluster time. Therefore, all timeout checks should be done using cluster time instead of local clock time. Otherwise, a clock shift that occurs after an invocation is initiated may cause premature operation timeouts.


jerrinot:
👍

pveentjer:
I'm against this. Cluster clock is unreliable and you currently don't give an option to break away from a call. Effectively a call could block much longer than configured by the end user.

PR #8667, opened 2016-08-05
[BACKPORT] Use cluster time for operation heartbeat timeout checks

metanet:
Invocation is initiated with cluster time. Therefore, all timeout checks should be done using cluster time instead of local clock time. Otherwise, a clock shift that occurs after an invocation is initiated may cause premature operation timeouts.


pveentjer:
Against. See:

https://github.com/hazelcast/hazelcast/pull/8666

PR #8668, opened 2016-08-05
Backport of #8649 - Too many client cache Invalidation callbacks for a single ICache.clear call

ihsandemir:
Fix for decreasing the cache clear callbacks for the client for a single clear request. Current implementation can cause up to parition count callbacks for a single clear request.

Added test for proving that the cache callback is only called one time during ICache.clear.

Added code for generating only one clear cache invalidation event when cache.clear is called. Test for ensuring only one clear cache invalidation event is sent for a cache.clear call is added.

Added invalidation event handling for other places where AbstractCacheRecordStore.clear is called.

Fixes issue #8593


ihsandemir:
verify

PR #8669, opened 2016-08-05
[DO NOT MERGE] Add the near cache executor overloaded property and implementation

### LINKED ISSUE #8593: Under high load and high ICache.getAsync calls near-cache updates may cause OOM

ihsandemir:
Adds the near cache executor overloaded property and implementation.

Fixes #8593

Replaces PR #8602


ihsandemir:
verify

PR #8671, opened 2016-08-08
Moved all ConfigXXXReadonly classes into the owning ConfigXXX class

pveentjer:
Also added @Override and moved some config tests into the config package.

This prevent coupling to the implementation and simplifies our api since the Readonly config classes won't show up.


pveentjer:
Closing this one since we can't move them due to non deprecated 'asReadonly' methods

PR #8673, opened 2016-08-08
Replaced AbstractOperation by Operation

pveentjer:
AbstractOperation was deprecated in 3.7 and its responsibilities (providing convenience implementations) has been taken over by the Operation.

For enterprise PR see:
https://github.com/hazelcast/hazelcast-enterprise/pull/983


pveentjer:
run-lab-run

Donnerbart:
We will need at least to patch Simulator 0.8.x to deal with 3.8-SNAPSHOT to deal with this change. I already patched the master branch, but without a maintenance patch this will break the nightly builds again.

Checking the other plugins/repos now.

Donnerbart:
I've updated Simulator and the code samples repository. Both should work after this change. The reference manual seems to contain some `AbstractOperation` references, but this won't break a build. Couldn't find any other usage of that class.

:+1:

PR #8675, opened 2016-08-08
Removed deprecated Properties

pveentjer:
Since these classes have moved to internal, we don't need to offer the public
ones anymore.

No enterprise PR required.


Donnerbart:
run-lab-run

pveentjer:
run-lab-run

jerrinot:
@pveentjer, @Donnerbart: guys, did you check other modules/plugins?

Donnerbart:
I'm quite sure that I ported everything to the new classes before the plugins got ripped out. I'm also not sure if any of the plugins is at 3.8-SNAPSHOT yet. But I'll check them out (I think I should have all of them locally due to the CheckStyle alignments).

As I stated above, the code samples could have a left over, but I'll check that as well.

Donnerbart:
I've checked all relevant repos and they are all fine.

PR #8677, opened 2016-08-08
Diagnostics log file name change

pveentjer:
Customer had problem using '#' in name, so replaced by '_'


Donnerbart:
:+1: (once green)

Issue #8277, opened 2016-05-30
Make ExpirationManager.PERIOD configurable


wildnez:
com.hazelcast.map.impl.eviction.ExpirationManager.PERIOD

Currently its hard coded to 5 seconds. Making it a configurable property will help in tuning the eviction.

PR #8683, opened 2016-08-08
Made ExpirationManager parameters configurable via system property

ahmetmircik:
closes https://github.com/hazelcast/hazelcast/issues/8277

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
@@ -20,47 +20,101 @@ import com.hazelcast.map.impl.MapServiceContext;
 import com.hazelcast.map.impl.PartitionContainer;
 import com.hazelcast.map.impl.operation.ClearExpiredOperation;
 import com.hazelcast.map.impl.recordstore.RecordStore;
+import com.hazelcast.nio.Address;
+import com.hazelcast.spi.ExecutionService;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.Operation;
-import com.hazelcast.spi.OperationService;
 import com.hazelcast.spi.impl.operationservice.InternalOperationService;
 import com.hazelcast.spi.partition.IPartition;
+import com.hazelcast.spi.partition.IPartitionService;
 import com.hazelcast.util.Clock;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.TimeUnit;
 
-// TODO refactor
+import static com.hazelcast.util.CollectionUtil.isEmpty;
+import static com.hazelcast.util.Preconditions.checkPositive;
+import static com.hazelcast.util.Preconditions.checkTrue;
+import static java.lang.Integer.getInteger;
+import static java.lang.Math.min;
+import static java.util.Collections.sort;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 /**
- * Manages expiration operations.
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
```diff
@@ -219,4 +256,19 @@ private void updateLastCleanupTimesBeforeSorting(List<PartitionContainer> partit
             partitionContainer.setLastCleanupTimeCopy(partitionContainer.getLastCleanupTime());
         }
     }
+
+    // used for testing purposes
+    public int getTaskPeriodSeconds() {
```
jerrinot: does it have to `public` ?

hazelcast/src/test/java/com/hazelcast/map/impl/eviction/ExpirationManagerTest.java
```diff
@@ -0,0 +1,125 @@
...
+import static com.hazelcast.map.impl.MapService.SERVICE_NAME;
+import static com.hazelcast.map.impl.eviction.ExpirationManager.SYS_PROP_EXPIRATION_CLEANUP_OPERATION_COUNT;
+import static com.hazelcast.map.impl.eviction.ExpirationManager.SYS_PROP_EXPIRATION_CLEANUP_PERCENTAGE;
+import static com.hazelcast.map.impl.eviction.ExpirationManager.SYS_PROP_EXPIRATION_TASK_PERIOD_SECONDS;
+import static java.lang.String.valueOf;
+import static java.lang.System.clearProperty;
+import static java.lang.System.setProperty;
+import static org.junit.Assert.assertEquals;
+
+@RunWith(HazelcastParallelClassRunner.class)
```
jerrinot: I guess these tests should not be run in parallel - as they are changing global state - the property.

hazelcast/src/test/java/com/hazelcast/map/impl/eviction/ExpirationManagerTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void testTaskPeriodSeconds_set_viaSystemProperty() throws Exception {
+        int expectedPeriodSeconds = 12;
+        setProperty(SYS_PROP_EXPIRATION_TASK_PERIOD_SECONDS, valueOf(expectedPeriodSeconds));
+        int actualTaskPeriodSeconds = new ExpirationManager(getMapServiceContext()).getTaskPeriodSeconds();
+        clearProperty(SYS_PROP_EXPIRATION_TASK_PERIOD_SECONDS);
```
jerrinot: it's safer to store the original property and restore it. just clearing might have unexpected side-effects - it assume the property was not set before the test was executed. also the cleaning should be done in a `finally` block.

jerrinot:
minor things, otherwise +1

eminn:
👍

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
ahmetmircik: fixed.

hazelcast/src/test/java/com/hazelcast/map/impl/eviction/ExpirationManagerTest.java
ahmetmircik: fixed.

hazelcast/src/test/java/com/hazelcast/map/impl/eviction/ExpirationManagerTest.java
ahmetmircik: fixed cleaning.

PR #8690, opened 2016-08-09
Improved SlowOperationDetectorBasicTest

Donnerbart:
* Increased sleep times in `SlowOperationDetectorBasicTest`
* Removed `ParallelTest` category to make the test less timing sensitive

All failures I could find are in the `SlowOperationDetectorBasicTest`. The main difference between the basic and the other `SlowOperationDetector` tests are:
* it's a `QuickTest`
* it's a `ParallelTest`
* it has the shortest sleep times to be a fast `QuickTest`

So I increased the sleep times from 3 to 5 seconds and removed the `ParallelTest` category, so we don't have any contention on this timing sensitive system due to other tests.


Donnerbart:
run-lab-run

PR #8691, opened 2016-08-09
Cleanup of unused private fields

Donnerbart:
I tried not to look left and right in most of the classes. So 99% percent of this PR are the removal of unused private fields. I also checked that EE will not break for any mysterious reasons.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
@@ -23,16 +23,10 @@ import com.hazelcast.nio.ClassLoaderUtil;
 import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.StringUtil;
 
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-
 /**
  * Provider to get any kind ({@link EvictionPolicyType}) of {@link EvictionPolicyEvaluator}.
  */
 public final class EvictionPolicyEvaluatorProvider {
-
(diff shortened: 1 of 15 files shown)


Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
```diff
@@ -63,15 +57,15 @@ public static EvictionPolicyEvaluator getEvictionPolicyEvaluator(EvictionConfigu
...
 
         String evictionPolicyComparatorClassName = evictionConfig.getComparatorClassName();
         if (!StringUtil.isNullOrEmpty(evictionPolicyComparatorClassName)) {
             try {
                 evictionPolicyComparator =
                         ClassLoaderUtil.newInstance(classLoader, evictionPolicyComparatorClassName);
             } catch (Exception e) {
-                ExceptionUtil.rethrow(e);
+                throw ExceptionUtil.rethrow(e);
             }
```
tombujok: This method already throws an exception

Donnerbart: Hmm, without it the `EvictionPolicyComparator evictionPolicyComparator = null;` initialization is needed as well, since the compiler cannot see that the catched exception is rethrown. I'll revert those changes, but maybe we can clean this up in another PR, this bugs me for a long time.

tombujok: ok, thx

tombujok:
👍

vbekiaris:
👍

PR #8692, opened 2016-08-10
Clears member side local near cache after imap#clear call

ahmetmircik:
Problem does not likely happen in common case, because clear-operations also clear local near-caches on partition owners but for example when one sets partition count to 1 on 2 nodes, if member-near-cache is on the map#clear caller node and partition owner(assuming no backups also) is on other node, caller nodes' near cache will not be empty after map#clear call without this fix.


eminn:
As you said even the case is unlikely, nice to have a fix for that👍

PR #8697, opened 2016-08-10
Makes imap#clear and imap#evictAll to send only one clear event to clients near-cache-invalidation-listener

ahmetmircik:


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapClearMessageTask.java
```diff
@@ -46,16 +49,22 @@ protected OperationFactory createOperationFactory() {
...
-        final MapService service = getService(MapService.SERVICE_NAME);
-        final Address thisAddress = service.getMapServiceContext().getNodeEngine().getThisAddress();
-        if (totalAffectedEntries > 0) {
-            service.getMapServiceContext().getMapEventPublisher()
-                    .publishMapEvent(thisAddress, parameters.name, EntryEventType.CLEAR_ALL, totalAffectedEntries);
+
+        MapService service = getService(MapService.SERVICE_NAME);
+        MapServiceContext mapServiceContext = service.getMapServiceContext();
+
+        if (clearedTotal > 0) {
```
ihsandemir: Should we send the event regardless of this number? It is possible that the cache became empty after client called clear before the response is returned to the client. Therefore, it will be safer to send this event regardless.

ahmetmircik: there shouldn't be any problem, one end can send clear event + preserved existing behaviour.

hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
```diff
@@ -983,6 +987,144 @@ public void testNearCache_clearFromClient() {
...
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                assertEquals("Expecting only 1 clear event", 1, handler.getClearEventCount());
+            }
+        });
+    }
+
+    @Test
+    public void receives_one_clearEvent_after_mapLoadAll() throws ExecutionException, InterruptedException {
```
ihsandemir: you can also add memberMap.loadAll test

ihsandemir:
👍

hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
ahmetmircik: done.

gurbuzali:
👍

PR #8714, opened 2016-08-15
Fixes comparison to find missing partitions on client side

ahmetmircik:
As a result of this bug `map#size` and `map#keySet#size` may not match sometimes.


jerrinot:
:+1:

jerrinot:
backports?

ahmetmircik:
@jerrinot will do.

Issue #8715, opened 2016-08-15
Client tasks should be bounced back when a member has not completed start yet


jerrinot:
Clients tasks should not be processed until a member start is completed. "start completed" means (among other things) that loading for HotRestart was completed. 

Serving client requests before loading is completed lead to consistency issues. For example: 
- a member with HotRestart was just started
- entry is stored in the persistence store, but it is not loaded yet into `RAMStore` -> the entry does not exist for a client.  
- loading is finished
- the entry is created "out-of-thin-air" (for the client)

See: `NodeExtension.isStartCompleted()`

Some tasks still have to be processed even when member start is not completed: e.g. the Heartbeat task.

ihsandemir:
The PRs are merged. Closing the issue.

PR #8722, opened 2016-08-15
Client message task should fail if node extension did not start

ihsandemir:
The task checks if the node extension start is finished before processing the message.

Fixes issue https://github.com/hazelcast/hazelcast/issues/8715

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/cluster/ClientNodeExtensionTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/cluster/ClientNodeExtensionTest.java
@@ -1,0 +1,111 @@
+package com.hazelcast.client.cluster;
+
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.client.spi.properties.ClientProperty;
+import com.hazelcast.client.test.TestHazelcastFactory;
+import com.hazelcast.config.Config;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IMap;
+import com.hazelcast.instance.DefaultNodeExtension;
+import com.hazelcast.instance.HazelcastInstanceFactory;
+import com.hazelcast.instance.Node;
+import com.hazelcast.instance.NodeExtension;
+import com.hazelcast.nio.Address;
+import com.hazelcast.test.HazelcastParallelClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
+import com.hazelcast.test.annotation.ParallelTest;
+import com.hazelcast.test.annotation.QuickTest;
+import com.hazelcast.test.mocknetwork.MockNodeContext;
+import com.hazelcast.test.mocknetwork.TestNodeRegistry;
+import org.junit.After;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -109,7 +110,8 @@ protected boolean isAuthenticationMessage() {
     }
 
     private void initializeAndProcessMessage() throws Throwable {
-        if (!node.joined()) {
+        NodeExtension nodeExtension = node.getNodeExtension();
+        if ((null != nodeExtension && !nodeExtension.isStartCompleted()) || (null == nodeExtension && !node.joined())) {
```
jerrinot: can `extension` ever be null?

mdogan: `null` is not expected. no need to check.

ihsandemir: OK, changed.

hazelcast-client/src/test/java/com/hazelcast/client/cluster/ClientNodeExtensionTest.java
```diff
@@ -0,0 +1,96 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.fail;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class ClientNodeExtensionTest {
+    @After
+    public void after()
+    {
+        HazelcastClient.shutdownAll();
```
jerrinot: this does not look right - you are running multiple tests in parallel here.

hazelcast-client/src/test/java/com/hazelcast/client/cluster/ClientNodeExtensionTest.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    @Test (expected = IllegalStateException.class)
+    public void test_canNotConnect_whenNodeExtensionIsNotComplete() {
+        LateStartingExtensionNodeContext nodeContext = new LateStartingExtensionNodeContext(false);
+
+        HazelcastInstanceFactory.newHazelcastInstance(new Config(), randomName(), nodeContext);
+
+        HazelcastClient.newHazelcastClient();
```
jerrinot: you do not want to create real clients in a parallel tests

jerrinot:
impl looks fine to me. the test has to be polished.

jerrinot:
I'm thinking about it again - shouldn't the client attempt to retry?

ihsandemir:
@jerrinot Client is retrying and giving up after the retries fails. I set the invocation timeout to 3 secs in the second test to make sure that it fails in 3 secs for map.get.

jerrinot:
:+1:

PR #8732, opened 2016-08-17
[DONT MERGE] Hold integer ttl

ahmetmircik:


ahmetmircik:
will be addressed later

PR #8737, opened 2016-08-17
Missing client near cache invalidation test for ICache.loadAll 

ihsandemir:
Added test to verify that cache invalidations work for ICache.loadAll as expected.


ahmetmircik:
👍

Issue #8736, opened 2016-08-17
Cluster leave and repartition exception


cstamas:
Similar setup as in https://github.com/hazelcast/hazelcast/issues/8733

Two nodes, `nodeA`, and `nodeB`.

Steps:
* start `nodeA`, forms cluster alone as 72f4c8c4-0ac4-4fba-a5ae-a875b44cd7d5
* start `nodeB`, it joins as 849e65d2-0292-4899-adbe-19bba2019f29
* stop `nodeA`

After `nodeA` started, as in referenced issue, instances of IAtomicLongs and ISemaphores were created and destroyed. During stopping of `nodeA`, `nodeB` reported some exceptions. Relevant logs pasted here:

nodeA:
```
jvm 1    | 2016-08-17 13:30:01,269+0200 INFO  [hz._hzInstance_1_hzgroup.cached.thread-2] *SYSTEM com.hazelcast.nio.tcp.InitConnectionTask - [192.168.5.82]:5701 [hzgroup] [3.7-EA] Connecting to /192.168.5.82:5702, timeout: 0, bind-any: true
jvm 1    | 2016-08-17 13:30:01,270+0200 INFO  [hz._hzInstance_1_hzgroup.cached.thread-4] *SYSTEM com.hazelcast.transaction.TransactionManagerService - [192.168.5.82]:5701 [hzgroup] [3.7-EA] Committing/rolling-back alive transactions of Member [192.168.5.82]:5702 - 12d5c714-6a0b-4556-820d-ebf4d4e361cf, UUID: 12d5c714-6a0b-4556-820d-ebf4d4e361cf
jvm 1    | 2016-08-17 13:30:01,273+0200 INFO  [hz._hzInstance_1_hzgroup.migration] *SYSTEM com.hazelcast.internal.partition.impl.MigrationManager - [192.168.5.82]:5701 [hzgroup] [3.7-EA] Partition balance is ok, no need to re-partition cluster data... 
jvm 1    | 2016-08-17 13:30:02,279+0200 INFO  [hz._hzInstance_1_hzgroup.migration] *SYSTEM com.hazelcast.internal.partition.impl.MigrationThread - [192.168.5.82]:5701 [hzgroup] [3.7-EA] All migration tasks have been completed, queues are empty.
[...]

PR #8738, opened 2016-08-17
Fix a potentially misleading exception message

jerrinot:
Exception can be sent over network hence using "this node"
is not great.

See #8736

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
@@ -196,8 +196,9 @@ class OperationRunnerImpl extends OperationRunner implements MetricsProvider {
             return;
         }
 
+        Address localAddress = node.getThisAddress();
         if (state == NodeState.SHUT_DOWN) {
-            throw new HazelcastInstanceNotActiveException("This node is shut down! Operation: " + op);
+            throw new HazelcastInstanceNotActiveException("Member " + localAddress + " is shut down! Operation: " + op);
         }
 
         if (op instanceof AllowedDuringPassiveState) {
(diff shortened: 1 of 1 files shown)


metanet:
👍

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
```diff
@@ -196,8 +196,9 @@ private void checkNodeState(Operation op) {
             return;
         }
 
+        MemberImpl localMember = node.getLocalMember();
         if (state == NodeState.SHUT_DOWN) {
-            throw new HazelcastInstanceNotActiveException("This node is shut down! Operation: " + op);
+            throw new HazelcastInstanceNotActiveException(localMember + " is shut down! Operation: " + op);
```
ahmetmircik: doesn't this message still contain `this`? See toString [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/instance/AbstractMember.java#L186)

mdogan:
👍

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
jerrinot: good point

jerrinot: perhaps it's ok to use an address here? wdyt?

ahmetmircik: seems better to me.

jerrinot:
@mdogan, @metanet, @ahmetmircik: I changed it to use addresses instead of members. Are you ok with it?

metanet:
yep

PR #8741, opened 2016-08-18
Missing client listener tests on cluster reconnects

ihsandemir:
Added tests to verify the listeners work when client reconnects to the cluster.


ihsandemir:
These tests are already added at AbstractListenersOnReconnectTest.

PR #8742, opened 2016-08-18
Small cleanup of NearCache related classes.

Donnerbart:
* renamed `NearChache.isInvalidateOnChange()` to `NearCache.isInvalidatedOnChange()`
* added missing generics to reduce warnings and make clearer which types are expected in which implementation
* unified naming of Near Cache (according to our documentation)
* fixed and improved some JavaDoc
* improved some comments
* whitespace formatting

The only modified classes which are not in an `impl` or `internal` package are:
* `ClientReplicatedMapProxy`
* `NearCachedClientMapProxy`
* `NearCacheTestSupport`

I didn't change any method interface in those classes, so there should be no compatibility issues.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
@@ -235,7 +235,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         // get updated records from client-2
         for (int i = 0; i < DEFAULT_RECORD_COUNT; i++) {
             final int key = i;
-            // records are stored in the near-cache will be invalidated eventually, since cache records are updated
+            // records are stored in the Near Cache will be invalidated eventually, since cache records are updated
             assertTrueEventually(new AssertTask() {
                 @Override
                 public void run() throws Exception {
@@ -285,7 +285,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
             keyAndValues.put(key, value);
         }
 
-        // verify that records are exist at near-cache of client-1 because `local-update-policy` is `CACHE`
+        // verify that records are exist at Near Cache of client-1 because `local-update-policy` is `CACHE`
         for (Map.Entry<String, String> entry : keyAndValues.entrySet()) {
             String key = entry.getKey();
             String exceptedValue = entry.getValue();
@@ -294,7 +294,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         }
 
         // remove records through client-2 so there will be invalidation events
-        // to send to client to invalidate its near-cache
+        // to send to client to invalidate its Near Cache
         for (Map.Entry<String, String> entry : keyAndValues.entrySet()) {
             nearCacheTestContext2.cache.remove(entry.getKey());
         }
@@ -308,7 +308,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         ((LifecycleServiceImpl) instanceToShutdown.getLifecycleService())
                 .fireLifecycleEvent(LifecycleEvent.LifecycleState.SHUTTING_DOWN);
 
-        // verify that records in the near-cache of client-1 are invalidated eventually when instance shutdown
+        // verify that records in the Near Cache of client-1 are invalidated eventually when instance shutdown
         for (Map.Entry<String, String> entry : keyAndValues.entrySet()) {
             final String key = entry.getKey();
             assertTrueEventually(new AssertTask() {
@@ -354,7 +354,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         // can't get deleted records from client-2
         for (int i = 0; i < DEFAULT_RECORD_COUNT; i++) {
             final int key = i;
-            // records are stored in the near-cache will be invalidated eventually, since cache records are updated.
+            // records are stored in the Near Cache will be invalidated eventually, since cache records are updated.
             assertTrueEventually(new AssertTask() {
                 @Override
                 public void run() throws Exception {
@@ -402,7 +402,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         // can't get replaced keys from client-2
         for (int i : loadKeys) {
             final int key = i;
-            // records are stored in the near-cache will be invalidated eventually, since cache records are updated
+            // records are stored in the Near Cache will be invalidated eventually, since cache records are updated
             assertTrueEventually(new AssertTask() {
                 @Override
                 public void run() throws Exception {
@@ -443,7 +443,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         // can't get expired records from client-2
         for (int i = 0; i < DEFAULT_RECORD_COUNT; i++) {
             final int key = i;
-            // records are stored in the near-cache will be invalidated eventually, since cache records are cleared
+            // records are stored in the Near Cache will be invalidated eventually, since cache records are cleared
             assertTrueEventually(new AssertTask() {
                 @Override
                 public void run() throws Exception {
@@ -463,13 +463,13 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         }
 
         for (int i = 0; i < DEFAULT_RECORD_COUNT; i++) {
-            // get records so they will be stored in near-cache
+            // get records so they will be stored in Near Cache
             nearCacheTestContext.cache.get(i);
         }
 
         for (int i = 0; i < DEFAULT_RECORD_COUNT; i++) {
             Data keyData = nearCacheTestContext.serializationService.toData(i);
-            // check if same reference to verify data coming from near cache
+            // check if same reference to verify data coming from Near Cache
             assertSame(nearCacheTestContext.cache.get(i), nearCacheTestContext.nearCache.get(keyData));
         }
     }
@@ -526,7 +526,7 @@ public abstract class ClientNearCacheTestSupport extends HazelcastTestSupport {
         // can't get expired records from client-2
         for (int i = 0; i < DEFAULT_RECORD_COUNT; i++) {
             final int key = i;
-            // records are stored in the near-cache will be invalidated eventually, since cache records are cleared
+            // records are stored in the Near Cache will be invalidated eventually, since cache records are cleared
             // because we just disable per entry invalidation events, not full-flush events
             assertTrueEventually(new AssertTask() {
                 @Override
--- a/hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
@@ -20,8 +20,9 @@ import com.hazelcast.internal.eviction.impl.comparator.LFUEvictionPolicyComparat
 import com.hazelcast.internal.eviction.impl.comparator.LRUEvictionPolicyComparator;
 import com.hazelcast.internal.eviction.impl.evaluator.DefaultEvictionPolicyEvaluator;
 import com.hazelcast.nio.ClassLoaderUtil;
-import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.StringUtil;
+
+import static com.hazelcast.util.ExceptionUtil.rethrow;
 
 /**
  * Provider to get any kind ({@link EvictionPolicyType}) of {@link EvictionPolicyEvaluator}.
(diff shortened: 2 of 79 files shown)


hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
```diff
@@ -43,27 +44,27 @@ private static EvictionPolicyComparator createEvictionPolicyComparator(EvictionP
...
-        EvictionPolicyComparator evictionPolicyComparator = null;
+        EvictionPolicyComparator evictionPolicyComparator;
 
         String evictionPolicyComparatorClassName = evictionConfig.getComparatorClassName();
         if (!StringUtil.isNullOrEmpty(evictionPolicyComparatorClassName)) {
             try {
                 evictionPolicyComparator = ClassLoaderUtil.newInstance(classLoader, evictionPolicyComparatorClassName);
             } catch (Exception e) {
-                ExceptionUtil.rethrow(e);
+                throw rethrow(e);
```
Donnerbart: I'm not 100% with this extra `throw`, but due to the `rethrow` signature IntelliJ won't stop complaining and cannot see that `evictionPolicyComparator` will always be set if no exception is thrown.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -68,12 +68,12 @@
     private final long timeToLiveMillis;
     private final boolean invalidateOnChange;
     private final EvictionPolicy evictionPolicy;
-    private final AtomicBoolean canCleanUp;
+    private final AtomicBoolean canExpire;
```
pveentjer: Both names are confusing. What about readyToExpire, expireRequired; depending on the situation. It explains better that the system is ready.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -90,21 +90,21 @@ public ClientHeapNearCache(String mapName, ClientContext context, NearCacheConfi
         this.timeToLiveMillis = TimeUnit.SECONDS.toMillis(nearCacheConfig.getTimeToLiveSeconds());
         this.invalidateOnChange = nearCacheConfig.isInvalidateOnChange();
         this.evictionPolicy = EvictionPolicy.valueOf(nearCacheConfig.getEvictionPolicy());
-        this.canCleanUp = new AtomicBoolean(true);
+        this.canExpire = new AtomicBoolean(true);
```
pveentjer: I would push everything that can be done in the 'class constructor' so that this constructor remains short. It also helps to quickly see which fields are just always set independent of the parameters, and which rely on the parameters. 

```
private final AtomicBoolean canExpire = new AtomicBoolean(true);
```

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -90,21 +90,21 @@ public ClientHeapNearCache(String mapName, ClientContext context, NearCacheConfi
...
         this.canEvict = new AtomicBoolean(true);
         this.stats = new NearCacheStatsImpl();
         this.selectedComparator = NearCacheRecord.getComparator(evictionPolicy);
-        this.lastCleanup = Clock.currentTimeMillis();
+        this.lastExpiration = Clock.currentTimeMillis();
```
pveentjer: I would add the time unit. So lastExpirationMillis. This prevents confusion.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
pveentjer: what about renaming this field 'expireInProgress' since it is uses as a lock to prevent concurrent cleanups.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -260,4 +196,71 @@ public NearCacheStatsImpl getNearCacheStats() {
         stats.setOwnedEntryMemoryCost(ownedEntryMemoryCost);
         return stats;
     }
+
+    private void fireCacheEviction() {
```
pveentjer: I would pull out this anonymous innerclass into a regular inner class. The anoymous inner class quickly lead to painful to read code, but also cause problems when analyzing heapdumps since you don't get to see a propper class name.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -260,4 +196,71 @@ public NearCacheStatsImpl getNearCacheStats() {
...
+                            } catch (RejectedExecutionException e) {
+                                canEvict.set(true);
+                            }
+                        }
+                    }
+                });
+            } catch (RejectedExecutionException e) {
+                canEvict.set(true);
+            } catch (Exception e) {
+                throw rethrow(e);
```
pveentjer: Are you not at risk of not releasing the 'canEvict' flag if another exception is thrown?

Perhaps have a single catch which always restores the canEvict flag?

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
```diff
@@ -529,6 +539,116 @@ public void run() throws Exception {
...
+            }
+        });
+    }
+
+    protected NearCacheStats getNearCacheStats(ICache cache) {
+        return cache.getLocalCacheStatistics().getNearCacheStatistics();
+    }
+
+    protected void waitForNearCacheEvictions(ICache cache, int evictionCount) {
+        long started = System.nanoTime();
```
pveentjer: Can't you make use of assertTrueEventually? This prevents repeating the same logic.

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
pveentjer: I have no problem with the; the problem is with the name 'rethrow'.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
```diff
@@ -277,6 +207,79 @@ public NearCacheStatsImpl getNearCacheStats() {
...
+                                stats.incrementEvictions();
+                                updateSizeEstimator(-calculateCost(record));
+                                if (++i > evictSize) {
+                                    break;
+                                }
+                            }
+                        } finally {
+                            canEvict.set(true);
+                        }
+                        if (cache.size() >= maxSize && canEvict.compareAndSet(true, false)) {
```
pveentjer: I know this is moved code, but why release the lock and then acquire it again? It makes the flow counter intuitive.  Perhaps the problem is that the method is just too complex.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
```diff
@@ -277,6 +207,79 @@ public NearCacheStatsImpl getNearCacheStats() {
...
+                            }
+                        } finally {
+                            canExpire.set(true);
+                        }
+                    }
+                });
+            } catch (RejectedExecutionException e) {
+                canExpire.set(true);
+            } catch (Exception e) {
+                throw rethrow(e);
```
pveentjer: You are potentially left with a locked canExpire if an exception is thrown. 

I normally do something like this:
```
boolean unlock = true;
try{
     someaction
    unlock = false
}finally{
   if(unlock) doUnlock()
}
```

It matches the locking idiom so you always know that something gets done.

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
```diff
@@ -127,13 +133,41 @@ public double getRatio() {
     }
 
     @Override
+    public long getEvictions() {
+        return evictions;
+    }
+
+    public void setEvictions(long evictions) {
```
pveentjer: Is this operation not racy? So imagine 2 threads do this at the same time, then you could be suffering from a lost update.

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
```diff
@@ -127,13 +133,41 @@ public double getRatio() {
...
+    public void incrementEvictions() {
+        EVICTIONS.incrementAndGet(this);
+    }
+
+    @Override
+    public long getExpirations() {
+        return expirations;
+    }
+
+    public void setExpirations(long expirations) {
```
pveentjer: Same  here. 

If this isn't an issue, then probably best to document that it isn't an issue so that future readers don't get their internal alarms triggered.

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTestSupport.java
```diff
@@ -0,0 +1,289 @@
...
+    protected void testNearCacheExpiration(final IMap<Integer, Integer> map, final int size, int expireSeconds) {
+        populateMap(map, size);
+        populateNearCache(map, size);
+
+        assertEquals(format("we expected to have all map entries in the Near Cache (%s)", getNearCacheStats(map)),
+                size, getNearCacheStats(map).getOwnedEntryCount());
+
+        final NearCacheStats statsBeforeExpiration = getNearCacheStats(map);
+
+        sleepSeconds(expireSeconds + 1);
```
pveentjer: Are you sure this won't cause problems if there is e.g. a stall or a gc?

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTestSupport.java
```diff
@@ -0,0 +1,289 @@
...
+     * Only needed for testing purposes.
+     */
+    protected void triggerNearCacheEviction(IMap<Integer, Integer> map) {
+        populateMap(map, 1);
+        populateNearCache(map, 1);
+    }
+
+    protected void waitForNearCacheEvictions(IMap map, int evictionCount) {
+        long started = System.nanoTime();
+        while (true) {
```
pveentjer: dejavu?

pveentjer:
Minor comments, nothing big. Looks good

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
Donnerbart: Good one, will be changed.

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
Donnerbart: It's just used in a test scenario. The value is just incremented during runtime.

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
Donnerbart: It's just used in a test scenario. The value is just incremented during runtime.

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTestSupport.java
Donnerbart: We have to wait until entries are expired. If we sleep longer, nothing bad can happen.

Donnerbart:
@pveentjer You commented on a follow-up PR of https://github.com/hazelcast/hazelcast/pull/8769
Don't know if I can keep this PR updated until the other one is merged.

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
Donnerbart: Not scope of this PR.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
Donnerbart: `ClientHeapNearCache` and `NearCacheImpl` are candidates to be removed soon. I just made a little naming adaption to have a better picture of eviction vs. expiration. I won't put much effort in these classes anymore.

PR #8750, opened 2016-08-19
Do not log a warning when a remote node is shutdown while destroying a proxy

jerrinot:
Reasoning:
When a member is shutdown then it's not a big deal when a proxy was
not  destryoed. It's just a noise and it looks scary.

See #8736


metanet:
:+1:

ahmetmircik:
👍

PR #8751, opened 2016-08-19
fixed hibernate client getLocalMembers unsupported exception issue

bilalyasar:


mesutcelik:
👍 merging as this is hibernate only fix... @bilalyasar Can you please send forward port?

PR #8754, opened 2016-08-20
Topic does not serialize message payload when there are no subscribers

### LINKED ISSUE #8262: ITopic.publish should not serialize if there are no subscribers
Class com.hazelcast.topic.impl.TopicProxy#publish serializes the message even if the message is never sent due to no subscribers. This is very inefficient at scale. For example, a producer periodically publishing on 100 topics of which typically only 10 have subscribers.

In the case of non-global-ordering and non-client use, the implementation should defer serialization until it has established that there are subscribers.

jerrinot:
Fixes #8262

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java
+++ b/hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java
@@ -22,9 +22,10 @@ import com.hazelcast.core.Message;
 import com.hazelcast.core.MessageListener;
 import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.internal.cluster.ClusterService;
-import com.hazelcast.logging.ILogger;
 import com.hazelcast.monitor.LocalTopicStats;
 import com.hazelcast.monitor.impl.LocalTopicStatsImpl;
+import com.hazelcast.nio.Address;
+import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.spi.EventPublishingService;
 import com.hazelcast.spi.EventRegistration;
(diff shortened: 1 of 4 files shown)


gurbuzali:
👍

hazelcast/src/main/java/com/hazelcast/topic/impl/TopicService.java
```diff
@@ -146,10 +147,14 @@ public void incrementReceivedMessages(String topicName) {
         getLocalTopicStats(topicName).incrementReceives();
     }
 
-    public void publishEvent(String name, TopicEvent event, boolean multithreaded) {
-        Collection<EventRegistration> registrations = eventService.getRegistrations(TopicService.SERVICE_NAME, name);
-        int partitionId = multithreaded ? counter.incrementAndGet() : name.hashCode();
-        eventService.publishEvent(TopicService.SERVICE_NAME, registrations, event, partitionId);
+    public void publishMessage(String topicName, Object payload, boolean multithreaded) {
+        Collection<EventRegistration> registrations = eventService.getRegistrations(SERVICE_NAME, topicName);
+        if (!registrations.isEmpty()) {
```
tombujok: This is slightly racy, isn't it? I know that there has to be a cut-off point, but still, there' some logic in between, that probably should be minimised?

jerrinot: It's racy only when a registration process is concurrent with a message publication. this is ok and expected.

tombujok:
👍

jerrinot:
@gurbuzali, @tombujok: thanks for a quick review!

PR #8761, opened 2016-08-22
[BACKPORT] Fix wrong client defaults when configured via Spring

### LINKED ISSUE #8746: Spring xml config: the smart routing attribute defaults to false while the docs say its enabled
Hello,

The smart routing attribute defaults to false while the documentation (and the ClientNetworkConfig) defaults to true.

Is there any reasoning behind this choice?

We relied on the "docs default" and now we are experiencing "Wait for owner connection is timed out" errors (which shouldn't be triggered if smart routing is enabled).

`<xs:attribute name="smart-routing" use="optional" type="xs:boolean" default="false"/>`

Note: We are using hazelcast 3.3.3, but the spring config default is the same in the latest release.

Thanks for the info!
Florin.

jerrinot:
Fix #8746, Backport of #8747
It would be great to  just use equals() on the config objects,
however a lot of our *Config(s) do not override equals()/hashcode()
and moreover the config objects are mutated in some cases - e.g.
  https://github.com/hazelcast/hazelcast/blob/eef987b9920f3f7b4c506e433b0d067cab9ba93c/hazelcast-client/src/main/java/com/hazelcast/client/config/ClientNetworkConfig.java#L207-L207

(cherry picked from commit 939c25c)

### DIFF
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.7.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.7.xsd
@@ -2119,9 +2119,9 @@
             <xs:element name="aws" type="aws-client" minOccurs="0" maxOccurs="1"/>
             <xs:element name="discovery-strategies" type="discovery-strategies" minOccurs="0" maxOccurs="1"/>
         </xs:sequence>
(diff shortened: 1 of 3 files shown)


mesutcelik:
👍 

@Serdaro ,
Can you please make sure to include this in release notes?

Serdaro:
👍

PR #8763, opened 2016-08-23
[DO NOT MERGE] Removing static-logging from places where it's unnecessary

### LINKED ISSUE #8760: Fix logging-framework static-context headache
As a user I want to be able to use the logger from a static context, whenever there is no access to NodeEngine without any side-effects to the logging configuration.

**STATE OF THE ART:**
Currently Hazelcast optimises the case when there is more than one HazelcastInstance run on a JVM and they have different logging configurations. Much as I consider this situation feasible and much as I would like to support it I doubt that it's the most common use-case. 

Additionally, it is impossible to always access nodeEngine for logging since there are Hazelcast classes that may be used without having an instance (like Config classes, etc.) 

There is a possible problem in the most common use-case:
- one node in one JVM, or a couple of nodes in one JVM
- the same logging setup configured in the `Config`. 

If somebody uses `Logger.getLogger(HazelcastCachingProvider.class);` the cached `LoggerFactory` gets initialised with the default logging system and will never change. Hazelcast will potentially log to two places

- java.util.logging -> all the loggers accessed from a static context.
- user-specified logging-system from the Config -> all the loggers accessed from the nodeEngine

**FIX:**
As a user:

In the one-node per JVM case:

-  I would prefer the statiticly cached `LoggerFactory` to be overwritten with the one initialised by the `newLoggerFactory(String loggerType)` so that the logging gets eventually unified.
- I would like the statiticly cached `LoggerFactory` to be initialised in the Hazelcast bootstrap sequence, as early as possible in order to reduce the time-spam when the logging is not unified
- I would like to change the HZ manual to put emphasis on the fact that the logging should be primarily configured by the System.property `hazelcast.logging.type` in order not to have anything logged to the default logger before the `LoggerFactory` gets initialised

In the multiple-nodes per JVM case (shared-case):

- I would like to introduce a flag `hazelcast.logging.shared` so that the user may indicate that the statiticly cached `LoggerFactory` does not get overwritten with the by the `newLoggerFactory(String loggerType)` call
- I would like to change the HZ manual to put emphasis on the fact that if the shared-logging is enabled all the statically accessed logging messages will land in the default-logging system 
- I would like to change the HZ manual to say that the logging System.properties (logging.type or logging.class) should not be used at all.

tombujok:
Fixes #8760

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
@@ -97,7 +97,7 @@ public class DefaultClientExtension implements ClientExtension {
 
     @Override
     public SocketInterceptor createSocketInterceptor() {
-        if(logger != null) {
+        if (logger != null) {
             logger.warning("SocketInterceptor feature is only available on Hazelcast Enterprise!");
         }
         return null;
--- a/hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
@@ -32,7 +32,6 @@ import com.hazelcast.internal.serialization.InternalSerializationService;
 import com.hazelcast.internal.serialization.SerializationServiceBuilder;
 import com.hazelcast.internal.serialization.impl.DefaultSerializationServiceBuilder;
 import com.hazelcast.logging.ILogger;
-import com.hazelcast.logging.Logger;
 import com.hazelcast.map.impl.MapService;
 import com.hazelcast.nio.ClassLoaderUtil;
 import com.hazelcast.nio.SocketInterceptor;
@@ -46,13 +45,14 @@ import static com.hazelcast.map.impl.MapConfigValidator.checkNotNative;
 
 public class DefaultClientExtension implements ClientExtension {
 
-    protected static final ILogger LOGGER = Logger.getLogger(ClientExtension.class);
+    protected ILogger logger;
 
     protected volatile HazelcastClientInstanceImpl client;
 
     @Override
     public void beforeStart(HazelcastClientInstanceImpl client) {
         this.client = client;
+        this.logger = client.getLoggingService().getLogger(DefaultClientExtension.class);
     }
 
     @Override
@@ -97,7 +97,9 @@ public class DefaultClientExtension implements ClientExtension {
 
     @Override
     public SocketInterceptor createSocketInterceptor() {
-        LOGGER.warning("SocketInterceptor feature is only available on Hazelcast Enterprise!");
+        if(logger != null) {
+            logger.warning("SocketInterceptor feature is only available on Hazelcast Enterprise!");
+        }
         return null;
     }
 
--- a/hazelcast-client/src/main/java/com/hazelcast/client/util/AddressHelper.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/util/AddressHelper.java
@@ -16,9 +16,9 @@
 
 package com.hazelcast.client.util;
 
-import com.hazelcast.logging.Logger;
 import com.hazelcast.util.AddressUtil;
 import com.hazelcast.util.AddressUtil.AddressHolder;
+import com.hazelcast.util.EmptyStatement;
 
 import java.net.Inet4Address;
 import java.net.Inet6Address;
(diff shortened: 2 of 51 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
```diff
@@ -46,13 +45,14 @@
...
 
-    protected static final ILogger LOGGER = Logger.getLogger(ClientExtension.class);
+    protected ILogger logger;
 
     protected volatile HazelcastClientInstanceImpl client;
 
     @Override
     public void beforeStart(HazelcastClientInstanceImpl client) {
         this.client = client;
+        this.logger = client.getLoggingService().getLogger(DefaultClientExtension.class);
```
jerrinot: is this guaranteed to be initialized when the extension is being created?

hazelcast-client/src/main/java/com/hazelcast/client/util/AddressHelper.java
```diff
@@ -61,7 +61,7 @@ private AddressHelper() {
         try {
             inetAddress = InetAddress.getByName(scopedAddress);
         } catch (UnknownHostException ignored) {
-            Logger.getLogger(AddressHelper.class).finest("Address not available", ignored);
+            EmptyStatement.ignore(ignored);
```
jerrinot: I am not sure about this one. it smells like a bad troubleshooting experience in a distance future :)

hazelcast/src/main/java/com/hazelcast/internal/memory/impl/UnsafeUtil.java
```diff
@@ -55,7 +51,6 @@
             }
         } catch (Throwable t) {
             unsafe = null;
-            LOGGER.warning("Unable to get an instance of Unsafe. Unsafe-based operations will be unavailable", t);
```
jerrinot: I also see this as a problematic. Unsafe may be unavailable due security manager / jigsaw / difference JVM and you wants to know this.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionStateGeneratorImpl.java
```diff
@@ -39,16 +39,21 @@
...
     private static final float RANGE_CHECK_RATIO = 1.1f;
+
     private static final int MAX_RETRY_COUNT = 3;
     private static final int AGGRESSIVE_RETRY_THRESHOLD = 1;
     private static final int AGGRESSIVE_INDEX_THRESHOLD = 3;
     private static final int MIN_AVG_OWNER_DIFF = 3;
 
+    private final ILogger logger;
+
+    public PartitionStateGeneratorImpl(LoggerFactory loggerFactory) {
```
jerrinot: is logging service unavailable yet? as logger factory will not the the fancy stuff with member identification, etc

jerrinot:
There are some controversial logging removals which I am afraid they eventually may bite us back. 

However the bigger concern is passing the `LoggerFactory` - it should be an implementation detail used by LoggingService(-s) only. It should not be used by other services.

hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
tombujok: Yep. Line 201 vs 216
https://github.com/tombujok/hazelcast/blob/2f6ab0358e16847ff86ed363ff0e523834f6aaeb/hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java#L201

hazelcast/src/main/java/com/hazelcast/logging/LoggingService.java
```diff
@@ -18,12 +18,8 @@
 
 import java.util.logging.Level;
 
-public interface LoggingService {
+public interface LoggingService extends LoggerFactory {
```
vbekiaris: In my mental model, `LoggerFactory` abstracts away the logger implementation, so I wouldn't mix `LoggingService` in `LoggerFactory` hierarchy, but that's just my interpretation. wdyt?

tombujok:
You can't have an ice-cream and eat an ice-cream. Your requirements look as follows now:

- don't use static logger
- don't pass logger (or logger-factory) as a parameter
- don't use static context to store loggers
- don't use thread-locals

Sorry, it's impossible to satisfy them. Either you pollute the API with ugly loggingService or you use some globally accessible context.

I'm fine with a minimal PR that initialises the logger on Hazelcast start - #8768 plus #8767
I'm not fine with leaving logging as it is right now.

In this way we may say that:

- try not to log within static context especially if not within a HZ instance (but you can still do it)
- **it is absolutely fine to use static loggers wherever you want within running HZ instance* 

That solves my concern in a minimalistic way.

PR #8765, opened 2016-08-23
Client mocked connections uses the same local Address

ihsandemir:
Modified the mock connection manager so that each new client gets a different local Address. Previously they were getting the same ip port (this confuses the logs and at some places the address is used as key to the connection maps) but their connection id was different, hence this usually did not cause any problems. The fix assigns unique port numbers for each mocked client connection.


asimarslan:
👍

PR #8766, opened 2016-08-23
Split Brain Test Support Improvements

jerrinot:
SplitBrainTestSupport is a convenient class for testing various split-brain situations.
Test author can specify a total cluster size and sizes of each brains during the split-brain
situation. The support class provides callbacks a test can override to be notified when
a split-brain is created and healed.

Use SplitBrainTestSupportTest as an example how to write a simple spit-brain test.


gurbuzali:
👍

vbekiaris:
👍 good work, very convenient

Issue #8767, opened 2016-08-24
Remove logging from Config classes


tombujok:
Remove logging and static loggers from Config classes.
Instead creates a validate method in each config entity that would be called by the HZ instance while the config is passed to it.

LoneRifle:
The logging found in Config/ClientConfig and subclasses are trivial and are limited to reporting to when default fallback configs for distributed objects are used or when HZ is being configured using a certain kind of Config. 

The latter arguably can be removed as the config type is determined by the user when setting up an instance. Knowing when HZ falls back to using default config for distributed objects is useful, so finding an alternate means of having that information will be awkward. @tombujok , do you have any feedback?

Implementing `validate()` appears to be a separate issue, unless its behaviour can be described in more detail in relation to the information conveyed in the logging

mmedenjak:
Agree with @LoneRifle , there are very few remains of usage of static loggers in the entire config package and overall, these are low priority. There is almost no validation that uses static loggers and most of the configuration validation is done at startup time (or distributed object creation time) where it throws a `InvalidConfigurationException`. 

If anyone is interested, you can check out if the remaining usages of static loggers can be removed from the package and, what's more interesting but unrelated to the logging issue - if we can somehow centralise the config validation as now it's rather scattered all over the codebase.

kosuruvarunkumar:
Hi, I would like to pick this up. Can someone guide me through, please?
Thanks

mmedenjak:
@kosuruvarunkumar I'm not sure if this is a good issue to begin with. Things have changed and as we mentioned, we probably don't use static loggers all that much anymore. The remaining usages either can't be removed or are trivial and would require 1 day at most.

The other issue, about centralised validation, is a big task IMHO. There is validation all over the codebase, some is done during config creation time, some during instance instantiation time, some when a feature is about to be used for the first time. I think it'll take a long time to first get a feeling of how it should be done. This being said, I'll remove the "good first issue" label and someone from the Hazelcast team can take a look at a later point to assess how to proceed with this issue. I think we may simply close it and open a new one for the centralised validation part.

PR #8768, opened 2016-08-24
Fix logger initialisation to support custom factory class and not to cache factory if of non-deterministic type

### LINKED ISSUE #8760: Fix logging-framework static-context headache

tombujok:
Initialise the static factory as early as possible. 
Do not cache static logger if of non-deterministic type.
Fixes #8760 
Follow up task is: #8767 

In this way we may say that:

- try not to log within static context especially if not within a HZ instance (but you can still do it)
- **it is absolutely fine to use static loggers wherever you want within running HZ instance**

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/logging/Logger.java
+++ b/hazelcast/src/main/java/com/hazelcast/logging/Logger.java
@@ -17,33 +17,62 @@
 package com.hazelcast.logging;
 
 import com.hazelcast.nio.ClassLoaderUtil;
+import com.hazelcast.util.StringUtil;
 
 public final class Logger {
 
     private static volatile LoggerFactory loggerFactory;
+
     private static final Object FACTORY_LOCK = new Object();
 
     private Logger() {
     }
 
+    /**
+     * @param clazz class to take the logger name from
+     * @return the logger
+     * Use LoggingService instead if possible. Do not log in static context if possible.
+     */
     public static ILogger getLogger(Class clazz) {
         return getLogger(clazz.getName());
     }
 
+    /**
+     * @param name name of the logger
+     * @return the logger
+     * Use LoggingService instead if possible. Do not log in static context if possible.
+     */
     public static ILogger getLogger(String name) {
+        LoggerFactory loggerFactoryToUse = loggerFactory;
         //noinspection DoubleCheckedLocking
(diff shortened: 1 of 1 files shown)


vbekiaris:
👍

Donnerbart:
:+1: (once green)

PR #8769, opened 2016-08-24
Added eviction and expiration count to NearCacheStats

Donnerbart:
* Added eviction and expiration counts to the `NearCacheStats`
* Added the parameter `wasExpired` to `EvictionListener.onEvict()` to distinguish if an entry was evicted due to expiration or a max-size policy eviction, to pick the correct `NearCacheStats`
* Added missing eviction/expiration tests to `NearCacheTest`, `ClientMapNearCacheTest` and `ClientNearCacheTest`
* Fixed `NearCacheTest` and `ClientMapNearCacheTest` regarding tests with `max-size` or `eviction-policy` (which were ignored in extended EE tests)
* Pulled out commonly used `NearCacheTestSupport` for `IMap` Near Cache tests
* Cleanup of `NearCacheImpl` and `ClientHeapNearCache` to distinct better between eviction and expiration

Added evictions and expiration counts to the `NearCacheStats` helped me a lot to verify if the Near Cache eviction or expiration was triggered (and of course to see how many entries were evicted). This will be useful for the Near Cache unification and code samples.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
@@ -26,7 +26,6 @@ import com.hazelcast.map.impl.nearcache.NearCacheRecord;
 import com.hazelcast.monitor.impl.NearCacheStatsImpl;
 import com.hazelcast.spi.serialization.SerializationService;
 import com.hazelcast.util.Clock;
-import com.hazelcast.util.ExceptionUtil;
 
 import java.util.Comparator;
 import java.util.Map;
@@ -39,6 +38,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.hazelcast.config.EvictionPolicy.NONE;
+import static com.hazelcast.util.ExceptionUtil.rethrow;
 
 /**
  * {@link NearCache} implementation for on-heap IMap on clients.
@@ -53,9 +53,9 @@ public class ClientHeapNearCache<K> implements NearCache<K, Object> {
     private static final double EVICTION_FACTOR = 0.2;
 
     /**
-     * TTL Clean up interval
+     * Expiration clean-up interval
      */
-    private static final int TTL_CLEANUP_INTERVAL_MILLS = 5000;
+    private static final int EXPIRATION_CLEANUP_INTERVAL_MILLIS = 5000;
 
     private final ConcurrentMap<K, NearCacheRecord> cache;
     private final String mapName;
@@ -68,12 +68,12 @@ public class ClientHeapNearCache<K> implements NearCache<K, Object> {
     private final long timeToLiveMillis;
     private final boolean invalidateOnChange;
     private final EvictionPolicy evictionPolicy;
-    private final AtomicBoolean canCleanUp;
+    private final AtomicBoolean canExpire;
     private final AtomicBoolean canEvict;
     private final NearCacheStatsImpl stats;
     private final Comparator<NearCacheRecord> selectedComparator;
 
-    private volatile long lastCleanup;
+    private volatile long lastExpiration;
 
     public ClientHeapNearCache(String mapName, ClientContext context, NearCacheConfig nearCacheConfig) {
         this.cache = new ConcurrentHashMap<K, NearCacheRecord>();
@@ -90,21 +90,21 @@ public class ClientHeapNearCache<K> implements NearCache<K, Object> {
         this.timeToLiveMillis = TimeUnit.SECONDS.toMillis(nearCacheConfig.getTimeToLiveSeconds());
         this.invalidateOnChange = nearCacheConfig.isInvalidateOnChange();
         this.evictionPolicy = EvictionPolicy.valueOf(nearCacheConfig.getEvictionPolicy());
-        this.canCleanUp = new AtomicBoolean(true);
+        this.canExpire = new AtomicBoolean(true);
         this.canEvict = new AtomicBoolean(true);
         this.stats = new NearCacheStatsImpl();
         this.selectedComparator = NearCacheRecord.getComparator(evictionPolicy);
-        this.lastCleanup = Clock.currentTimeMillis();
+        this.lastExpiration = Clock.currentTimeMillis();
     }
 
     @Override
     public void put(K key, Object value) {
-        fireTtlCleanup();
+        fireCacheExpiration();
         if (evictionPolicy == NONE && cache.size() >= maxSize) {
             return;
         }
         if (evictionPolicy != NONE && cache.size() >= maxSize) {
-            fireEvictCache();
+            fireCacheEviction();
         }
         Object candidate;
         if (value == null) {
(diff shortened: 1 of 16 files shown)


hazelcast/src/main/java/com/hazelcast/monitor/NearCacheStats.java
```diff
@@ -59,4 +59,11 @@
      * @return hit/miss ratio of the locally owned entries.
      */
     double getRatio();
+
+    /**
+     * Returns the number of evictions of locally owned entries.
```
ahmetmircik: `locally owned entries` seems a blurry explanation. Because, in most cases, nearCache is used for caching entries of remote nodes. An alternative doc can be like this: 

`Returns the number of evictions from this near cache`<-- _but this also doesn't address a question: Are expired entries also included in eviction-count?_
_If yes, why don't we provide a separate `getExpirations` method? Because expiration is not eviction in hazelcast parlance._

Donnerbart: I used the same phrase like in all other methods. I think it makes sense since the statistics all refer to the entries in the local Near Cache, not to the backing data structure (which can be of course remote or partly local).

But you are right about eviction vs. expiration. At the moment there is no difference between that in this PR (eviction, TTL and max-idle should all increase that counter).

Would you be fine with the enhancement if I would split that up?

ahmetmircik: Right `locally owned entries` exists on other javaDocs also. But still think it is blurry and needed to fix for other methods also.

For eviction vs. expiration, adding `getExpirations` makes sense.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -131,6 +131,7 @@ public void run() {
                             int i = 0;
                             for (NearCacheRecord record : records) {
                                 cache.remove(record.getKey());
+                                stats.incrementEvictions();
                                 if (++i > evictSize) {
```
Donnerbart: We might have an off-by-one error here, since we remove one entry more than the calculated `evictSize`. So when there are 100 entries in the Near Cache the `evictSize` will be `20`, but `21` entries will be removed from the Near Cache.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -260,4 +196,71 @@ public NearCacheStatsImpl getNearCacheStats() {
...
+                    public void run() {
+                        try {
+                            Set<NearCacheRecord> records = new TreeSet<NearCacheRecord>(selectedComparator);
+                            records.addAll(cache.values());
+                            int evictSize = (int) (cache.size() * EVICTION_FACTOR);
+                            int i = 0;
+                            for (NearCacheRecord record : records) {
+                                cache.remove(record.getKey());
+                                stats.incrementEvictions();
+                                if (++i > evictSize) {
```
Donnerbart: We might have an off-by-one error here, since we remove one entry more than the calculated `evictSize`. So when there are 100 entries in the Near Cache the `evictSize` will be `20`, but `21` entries will be removed from the Near Cache. I adapted the tests accordingly, but this was a bit confusing to figure out :)

hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
```diff
@@ -1190,49 +1195,6 @@ public void run() {
     }
 
     @Test
-    public void testNearCacheTTLCleanup_triggeredViaPut() {
```
Donnerbart: Expiration should *not* be triggered via put, just eviction. So this test didn't work out anymore with checking the expiration/eviction stats accordingly.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
```diff
@@ -529,6 +538,112 @@ public void run() throws Exception {
...
+        final NearCacheTestContext context = createNearCacheTest(DEFAULT_CACHE_NAME, nearCacheConfig);
+
+        for (int i = 0; i < size; i++) {
+            context.cache.put(i, "value-" + i);
+            context.cache.get(i);
+        }
+
+        final NearCacheStats statsBeforeExpiration = getNearCacheStats(context.cache);
+        assertEquals(format("we expected to have all map entries in the Near Cache (%s)", statsBeforeExpiration),
+                size, statsBeforeExpiration.getOwnedEntryCount() + statsBeforeExpiration.getExpirations());
```
Donnerbart: I added the expiration count here, since there was a test failure that some entries were already removed from the cache.

hazelcast/src/main/java/com/hazelcast/monitor/NearCacheStats.java
```diff
@@ -19,44 +19,58 @@
...
+     * Returns the creation time of this Near Cache on this member.
      *
-     * @return creation time of this NearCache on this member
+     * @return creation time of this Near Cache on this member.
      */
     long getCreationTime();
 
     /**
-     * Returns the number of entries owned by this member.
+     * Returns the number of Near Cache entries owned by this member.
```
vbekiaris: Does "entries owned by this member" make sense in the context of near cache? Perhaps we should amend the javadoc to "entries on this node" (applying to client as well)?

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
vbekiaris: You are right, do you plan to fix this in a future PR?

vbekiaris:
👍 nice & thorough work

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
Donnerbart: I'm already prototyping to replace those classes with `DefaultNearCache`. So I think it's okay to stick with it.

hazelcast/src/main/java/com/hazelcast/monitor/NearCacheStats.java
Donnerbart: That's a good point. Is it okay if I fix it in the pending cleanup PR? I have to rebase that one anyway.

vbekiaris: Sure

Donnerbart:
Peter indirectly reviewed this PR by having a look at https://github.com/hazelcast/hazelcast/pull/8742

I picked up all his comments, beside the ones on implementation details of `NearCacheImpl` and `ClientHeapNearCache`, since they are going to be removed anyway.

So I'm merging this PR now. Thanks for the reviews!

PR #8778, opened 2016-08-25
Fixed eviction of max-idle expired entries in IMap client Near Cache.

Donnerbart:
* Fixed eviction of max-idle expired entries in IMap client Near Cache
* Fixed test which covered that bug

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
@@ -194,7 +194,6 @@ public class ClientHeapNearCache<K> implements NearCache<K, Object> {
         fireTtlCleanup();
         NearCacheRecord record = cache.get(key);
         if (record != null) {
-            record.access();
             if (record.isExpired(maxIdleMillis, timeToLiveMillis)) {
                 cache.remove(key);
                 stats.incrementMisses();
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -194,7 +194,6 @@ public Object get(K key) {
         fireTtlCleanup();
         NearCacheRecord record = cache.get(key);
         if (record != null) {
-            record.access();
```
Donnerbart: This updates the last access timestamp, so an already expired entry will not be evicted.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -205,6 +204,7 @@ public Object get(K key) {
                 return NULL_OBJECT;
             }
             stats.incrementHits();
+            record.access();
```
Donnerbart: This is the position where we call it on the member implementation.

hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
```diff
@@ -869,7 +871,9 @@ public void testNearCacheIdleRecordsEvicted() {
         stats = map.getLocalMapStats().getNearCacheStats();
 
         assertEquals("as the hits are not equal, the entries were not cleared from near cash after MAX_IDLE_SECONDS",
-                hitsBeforeIdleExpire, stats.getHits(), size);
```
Donnerbart: The `size` was acting as `delta` parameter for the assert, so we basically disabled the assert, although the hit counters were different.

vbekiaris:
good finding! 👍

Issue #6991, opened 2015-12-04
ClientRegressionWithMockNetworkTest.*


Donnerbart:
This is a failure from my local runs, so no CloudBees link is available.

```
java.lang.AssertionError
    at org.junit.Assert.fail(Assert.java:86)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at org.junit.Assert.assertTrue(Assert.java:52)
    at com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenClientAndOwnerNodeDiesTogether(ClientRegressionWithMockNetworkTest.java:582)
    at com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenDummyClientAndOwnerNodeDiesTogether(ClientRegressionWithMockNetworkTest.java:563)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
    at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
    at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    at com.hazelcast.test.AbstractHazelcastClassRunner$ThreadDumpAwareRunAfters.evaluate(AbstractHazelcastClassRunner.java:132)
    at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
    at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
    at com.hazelcast.test.HazelcastParallelClassRunner.access$301(HazelcastParallelClassRunner.java:46)
    at com.hazelcast.test.HazelcastParallelClassRunner$TestRunner.run(HazelcastParallelClassRunner.java:139)
    at java.lang.Thread.run(Thread.java:745)
```

https://gist.github.com/Donnerbart/aaed8a5ed47ea92d327f

The failing assert is this one:

```
[...]

PR #8782, opened 2016-08-25
Client Lifecycle listener may not receive events in expected order

ihsandemir:
Fixes lifecycle state listener call ordering. Uses a single threaded executor (clusterExecutor) to satisfy the order. Fixes issue #6991

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
@@ -204,13 +204,13 @@ public class HazelcastClientInstanceImpl implements HazelcastInstance, Serializa
 
         credentials = initCredentials(config);
         threadGroup = new ThreadGroup(instanceName);
-        properties = new HazelcastProperties(config.getProperties());
-        executionService = initExecutionService();
         lifecycleService = new LifecycleServiceImpl(this);
+        properties = new HazelcastProperties(config.getProperties());
 
         metricsRegistry = initMetricsRegistry();
         serializationService = clientExtension.createSerializationService((byte) -1);
         proxyManager = new ProxyManager(this);
+        executionService = initExecutionService();
         loadBalancer = initLoadBalancer(config);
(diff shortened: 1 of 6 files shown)


jerrinot:
@ihsandemir: do you need to offload it to another thread at all? contract of the `LifecycleListener` says: "No blocking calls should be made in this method."

ihsandemir:
@jerrinot This test fails in that case: https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/test/java/com/hazelcast/client/ClientRegressionWithMockNetworkTest.java#L596

The listener is being called from the IO thread and when inside the listener if the user does map.get, since io thread is not working it does not receive the response to the get request and it is blocked indefinitely. 

This is the main reason.

jerrinot:
a user doing `map.get()` from inside the callback is breaking a contract of the listener. should we encourage it? just asking.

ihsandemir:
@jerrinot I would not encourage it but this is not a new test case. Hence, if anybody already used this way and we change the behaviour in implementation we will break them if no thread is used. 

The meaning of "No blocking call" may not really very clear. If our impl were not doing the socket close using the IO thread, then the users map.get call would not block at all.

Hence, I myself also not clear about what to allow and what not to in the callback. May be being able to make a hazelcast api call inside the callback is useful for application writers.

ihsandemir:
I tried if we can utilize the EventService StripedExecutor instead of introducing a new executor for delivering Lifecycle events. It causes problem due to the shutdown order of the services. The Lifecycle service closes after all the other services are closed, but it needs to deliver the SHUTDOWN event as well during LifecycleService shutdown but the EventService is already closed. Hence, the cleanest solution will be introducing the single threaded executor directly as a member of the LifecycleService and using this executor, this will eliminate any dependency problems.

hasancelik:
run-lab-run

asimarslan:
👍

mdogan:
After this change `HiDensityClientNearCacheTest` is failing, on EE side.

Donnerbart:
I can confirm Mehmet's observation, although I have no idea why.

PR #8783, opened 2016-08-25

aqeeph:
String variable initialization.It's "protection" against NullPointerException.


jerrinot:
Hello @aqeeph,

I saw you sent a bunch of similar PRs. May I ask what's the goal of these PRs?  They have very limited description. I assume it's "protection" against NPE? It will not work for us as a `null` / `lack-of-value` has its meaning in many situations.

jerrinot:
Hello, please see an explanation at #8783 (comment) why this is not a good idea.

PR #8789, opened 2016-08-26

aqeeph:


jerrinot:
Hello, please see an explanation at #8783 (comment) why this is not a good idea.

PR #8790, opened 2016-08-26

aqeeph:
String variable initialization.


jerrinot:
Hello, please see an explanation at https://github.com/hazelcast/hazelcast/pull/8783#issuecomment-242641332 why this is not a good idea.

PR #8792, opened 2016-08-26
[DO NOT MERGE] Less aggressive predicate flattening when not all sides are the same

jerrinot:
A candidate fix, do not merge yet.

There must be a backport once a proper fix is created.

PR #8794, opened 2016-08-27
Less aggressive predicate flattening when not all sides are the same

vbekiaris:
Less aggressive predicate flattening when not all sides are the same
Adds SQL predicate tests

(supersedes #8792)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/SqlPredicate.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/SqlPredicate.java
@@ -319,7 +319,7 @@ public class SqlPredicate
         Predicate[] predicates;
         if (klass.isInstance(predicateLeft)) {
             subpredicatesLeft = ((CompoundPredicate) predicateLeft).getPredicates();
-            if (predicateRight instanceof CompoundPredicate) {
+            if (predicateRight.getClass() == klass) {
                 subpredicatesRight = ((CompoundPredicate) predicateRight).getPredicates();
             } else {
                 subpredicatesRight = new Predicate[] {predicateRight};
--- a/hazelcast/src/main/java/com/hazelcast/query/SqlPredicate.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/SqlPredicate.java
@@ -314,22 +314,13 @@ public class SqlPredicate
         // The following could have been achieved with {@link com.hazelcast.query.impl.predicates.FlatteningVisitor},
         // however since we only care for 2-argument flattening, we can avoid constructing a visitor and its internals
         // for each token pass at the cost of the following explicit code.
-        Predicate[] subpredicatesLeft;
-        Predicate[] subpredicatesRight;
         Predicate[] predicates;
-        if (klass.isInstance(predicateLeft)) {
-            subpredicatesLeft = ((CompoundPredicate) predicateLeft).getPredicates();
-            if (predicateRight.getClass() == klass) {
-                subpredicatesRight = ((CompoundPredicate) predicateRight).getPredicates();
-            } else {
-                subpredicatesRight = new Predicate[] {predicateRight};
-            }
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/query/SqlPredicateTest.java
```diff
@@ -67,13 +67,100 @@
...
+            };
+
+    @Test
+    public void testSqlPredicates() {
+        Employee employee = new Employee("Joe", "AUSTIN", 25, true, 0);
+        for (String s : TEST_MATCHING_SQL_PREDICATES) {
+            try {
+                assertSqlTrue(s, employee);
+            }
+            catch (AssertionError assertionError) {
```
jerrinot: I'd do following:
- rename `assertSqlTrue()` to `assertSqlMatching()`.
- amend `assertSqlMatching()` to include the SQL on failure: `assertTrue("Failed matching SQL predicate \"" + s + '"', predicate.apply(entry));`
- remote the try/catch block from here - no need to polute a test with it
- do the same for `assertSqlFalse()`
WDYT?

jerrinot:
It looks good. I made a minor comment regarding a test purity. 

It can be merged and backported into a maintenance branch once it's green. Also please create a GitHub issue and set a milestone to 3.7.1 - so the fix will be visible in the release notes.

jerrinot:
I sent backport #8798. The commits are just cherry-picked from this PR. 
However for master the tests should be cleaned a bit.

vbekiaris:
Thanks for the test cleanup suggestion, looks much better now.

jerrinot:
:+1:

hazelcast/src/main/java/com/hazelcast/query/SqlPredicate.java
```diff
@@ -317,7 +317,7 @@ private void readObject(java.io.ObjectInputStream in)
         Predicate[] predicates;
         if (klass.isInstance(predicateLeft)) {
             subpredicatesLeft = ((CompoundPredicate) predicateLeft).getPredicates();
-            if (predicateRight instanceof CompoundPredicate) {
+            if (predicateRight.getClass() == klass) {
```
pveentjer: I'm not familiar with the problem, so I don't understand the solution. Could you shine some light on it so I can give my thumbs up?

vbekiaris: I intend to add a couple of tests before we get this merged, so I will mark as 'dont merge' this PR for now. The problem was that sub-predicates joined by different logical operator (AND / OR) would be eventually flattened to a single compound predicate, discarding one of the two operators, leading to incorrect representation of the SQL statement. The fixed condition ensures we only flatten when operator is the same.

mmedenjak: This logic seemed a bit complicated to me so I did an attempt at refactoring, might be useful, might not...

```
        Predicate[] predicates;
        if (klass.isInstance(predicateLeft) || klass.isInstance(predicateRight)) {
            Predicate[] left = getSubPredicatesIfClass(predicateLeft, klass);
            Predicate[] right = getSubPredicatesIfClass(predicateRight, klass);
            
            predicates = new Predicate[left.length + right.length];
            ArrayUtils.concat(left, right, predicates);
        } else {
            predicates = new Predicate[] {predicateLeft, predicateRight};
        }
```

And the helper method : 

```
    private static <T extends CompoundPredicate> Predicate[] getSubPredicatesIfClass(Predicate predicate, Class<T> klass){
        if (klass.isInstance(predicate)) {
            return ((CompoundPredicate) predicate).getPredicates();
        } else {
            return new Predicate[]{predicate};
        }
    }
```

vbekiaris:
Updated with refactored code by @mmedenjak and additional tests to verify flattening with combinations of `OR`/`AND` predicates

vbekiaris:
run-lab-run

PR #8798, opened 2016-08-27
[BACKPORT] Less aggressive predicate flattening when not all sides are the same

jerrinot:
Backport of #8794


jerrinot:
I'm merging this one. 
It's a very small change and both me @vbekiaris spent quite some time on it and we have a good understand of what is going on. Also see @8794

PR #8800, opened 2016-08-28
Pass ClassType into a constructor when creating a new instance of Serializer

jerrinot:
When a serializer does not have a constructor accepting `Class` then a non-arg constructor will be used - the same as now. 

Reasoning:
This allows to use the same serializer class to (de-)serialize multiple domain classes. I stumbled upon this while working on the [SubZero](https://github.com/jerrinot/subzero) project.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializerHookLoader.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializerHookLoader.java
@@ -42,6 +42,9 @@ final class SerializerHookLoader {
 
     private static final String FACTORY_ID = "com.hazelcast.SerializerHook";
 
+    private final boolean useDefaultConstructorOnly =
+            Boolean.getBoolean("hazelcast.compat.serializers.use.default.constructor.only");
+
     private final Map<Class, Object> serializers = new HashMap<Class, Object>();
     private final Collection<SerializerConfig> serializerConfigs;
     private final ClassLoader classLoader;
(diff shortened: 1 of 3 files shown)


asimarslan:
Looks good, it's also backward compatible. 👍

pveentjer:
Looks good. 

@asimarslan what should we watch for regarding backwards compatibility? It doesn't change anything that goes over the wire.

asimarslan:
@pveentjer Client compatibility is not just wire data compatibility. It also include behavioural  compatibility. 
Assume a user already has a constructor with class parameter, that user cannot upgrade to a new version without problem.
Moreover we support client only and server only upgrades/downgrades.
We had a long conversation on this subject and it's decided not to change any behaviours even the bugs unless it complies with documentation.
For such cases we can introduce a system parameter which is off by default and enabled if user wants and accept the incompatibility risk

jerrinot:
@asimarslan I added a compatibility switch. However I use the new strategy by default. 

Reasoning:
This will affect a tiny portion of users. It's an intersection of these user categories:
- they have own implementation of `Serializer` 
- the `Serializer` have a constructor accepting `Class` 
- the custom serializer will produce a different bytes when instantiated by the non-empty constructor
- they will not read Release Notes describing the change
- they do not have tests to catch the change before rolling out to PROD

This is such a tiny minority of users that it should not prevent us implementing improvements (enabled by default). Especially when there is a way to re-introduce the old behaviour. It should be enough to describe the change in the release notes. I am not suggesting to backport this into a patch release where changes like this are more problematic.

asimarslan:
👍

PR #8802, opened 2016-08-28
Restore behavior of JCache near-cache invalidation events

### LINKED ISSUE #8801: Behavior of JCache near-cache invalidation events changed
With #8649 the delivery of JCache near-cache invalidation events was changed as follows:
* `Cache.close` now delivers near-cache invalidation events while it used not to
* `Cache.destroy` no longer delivers invalidation events
* `CacheManager.destroyCache` when invoked from a client-side cache manager no longer delivers invalidation events

vbekiaris:
Restores behavior of JCache near-cache invalidation events.

Introduces test (succeeds on v3.7, fails on current `maintenance-3.x` branch) which verifies delivery of invalidation events as needed. The fixes are in a separate commit for clarity.

Fixes #8801


hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
```diff
@@ -95,28 +95,6 @@ protected NearCacheConfig createNearCacheConfig(InMemoryFormat inMemoryFormat) {
                 .setInMemoryFormat(inMemoryFormat);
     }
 
-    protected class NearCacheTestContext {
```
vbekiaris: This class was refactored to top-level class as it is generally useful for near-cache tests.

hazelcast/src/test/java/com/hazelcast/cache/CacheCloseTest.java
```diff
@@ -1,160 +0,0 @@
...
-import java.util.concurrent.atomic.AtomicInteger;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-
-@RunWith(HazelcastParallelClassRunner.class)
-@Category({QuickTest.class, ParallelTest.class})
-public class CacheCloseTest
```
vbekiaris: This test is removed as it is the same as `CacheDestroyTest`, except for the fact that it checks invalidation events received from `Cache.close`, for which no invalidation events should be published in the first place.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
```diff
@@ -0,0 +1,355 @@
+/*
```
ihsandemir: This client part of this test corresponds to the similar test written for cache.clear ClientCacheClearTest. Can we make both tests to be under the same package?

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
```diff
@@ -0,0 +1,355 @@
...
+        assertTrueAllTheTime(new AssertTask() {
+            @Override
+            public void run()
+                    throws Exception {
+                assertEquals(0, counter.get());
+            }
+        }, TIMEOUT);
+    }
+
+    private void registerInvalidationListener(AtomicInteger counter) {
```
ihsandemir: Are we missing the near cache event test for the member side near cache impl?

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
```diff
@@ -0,0 +1,355 @@
...
+        EventHandler handler = new NearCacheInvalidationHandler(counter);
+        ListenerMessageCodec listenerCodec = createInvalidationListenerCodec();
+        try {
+            Field clientContextField = Class.forName("com.hazelcast.client.cache.impl.AbstractClientCacheProxyBase")
```
ihsandemir: Can you check the ClientCacheClearTest.registerInvalidationListener? We have another way of registering client event listener without using reflection.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+    }
+
+    @After
+    public void tearDown() {
+        hazelcastFactory.shutdownAll();
+    }
+
+    @Test
+    public void when_cacheDestroyed_invalidationEventIsReceived() {
```
ihsandemir: Do we need a test for also verifying that there is no more than 1 event received? (was the case for ICache.clear)

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
```diff
@@ -135,7 +113,7 @@ protected NearCacheTestContext createNearCacheTest(String cacheName, NearCacheCo
         NearCache<Data, String> nearCache =
                 nearCacheManager.getNearCache(cacheManager.getCacheNameWithPrefix(cacheName));
 
-        return new NearCacheTestContext(client, cacheManager, nearCacheManager, cache, nearCache);
+        return new NearCacheTestContext(client, null, cacheManager, null, nearCacheManager, cache, null, nearCache);
```
ihsandemir: May be an overloaded constructor (which only takes the non-null parameters) instead of using null for the parameters would look nicer.

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxyBase.java
```diff
@@ -141,9 +141,6 @@ public void close() {
         }
         loadAllTasks.clear();
```
ihsandemir: Not related with this PR but there seems to be a problem with the usage of loadAllTasks in the loop here. Don't we need to work on a copy of the loadAllTasks list for iterator safety?

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
vbekiaris: We only care about the client side in this test, there is no member-side near cache for JCache in 3.7.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
vbekiaris: 👍 thanks, will use that method to register the listener.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
vbekiaris: `ClientCacheClearTest` is also a functional test for `cache.clear`, if you don't mind I'd rather keep this test under the `nearcache` sub-package as it has to do exclusively with near cache functionality.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
vbekiaris: Sounds good, will incorporate this assertion in test cases where an invalidation event is expected to be received.

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
```diff
@@ -288,6 +289,7 @@ public void deleteCache(String name, String callerUuid, boolean destroy) {
         CacheConfig config = deleteCacheConfig(name);
         if (destroy) {
             destroySegments(name);
+            sendInvalidationEvent(name, null, SOURCE_NOT_AVAILABLE);
```
ihsandemir: Do we need to send this event? As I trace this call, it can be called either via reset method (but this method already has invalidation at the end), or ProxyServiceImpl.destroyLocalDistributedObject. Do we need invalidation for ProxyServiceImpl.destroyLocalDistributedObject?

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
vbekiaris: Fixed

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
vbekiaris: Yes, this is necessary, otherwise `Cache.destroy` and `CacheManager.destroyCache`
will not deliver an invalidation event (which used to be the case until 3.7).

The method usage in `reset` invokes `deleteCache` with argument `destroy = false`,
so there will be no duplicate invalidation event.

vbekiaris:
Unrelated failure in `com.hazelcast.durableexecutor.DurableExecutorServiceTest.testLongRunningCallable`

vbekiaris:
run-lab-run

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxyBase.java
vbekiaris: Hmm, I don't see a problem here; we just wait for each `Future` in `loadAllTasks` to complete and then clear the list. The list itself is a `CopyOnWriteArrayList` (so thread-safe) and moreover, in https://github.com/vbekiaris/hazelcast/blob/db5ea7d2fe437fcac8ad1304aeca0925ba8471ee/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheProxyBase.java#L128-L128 we mark the cache as closed so no further tasks can be submitted.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
```diff
@@ -0,0 +1,344 @@
...
+        clear();
+        assertInvalidationEventReceivedExactlyOnce(counter);
+    }
+
+    @Test
+    public void when_cacheClosed_invalidationEventIsNotReceived() {
+        final AtomicInteger counter = new AtomicInteger();
+        registerInvalidationListener(counter);
+        close();
+        assertTrueAllTheTime(new AssertTask() {
```
Donnerbart: I think this code is a good candidate for a `assertInvalidationEventNeverReceived()` method, since we have it three times in the test.

vbekiaris: Fixed -- unfortunately IDEA does a good job hiding these as one-liner lambdas.

hazelcast/src/test/java/com/hazelcast/cache/CacheCloseTest.java
Donnerbart: I was first concerned that we might miss code coverage for the `Cache.close()` method if this test is removed, but it seems that there are some other `testThisAndThat_afterCacheClose`, so we should be fine.

Donnerbart:
The tests look good for me and the code changes seem to fit the new test scenarios. That code is quite new to me, so I hope I didn't miss anything, but so far I'll give my thumbs up.

:+1:

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
```diff
@@ -0,0 +1,336 @@
...
+    // When invoked from a member-side CacheManager, invalidation event is not received.
+    @Test
+    public void when_cacheManagerDestroyCacheInvoked_invalidationEventMayBeReceived() {
+        final AtomicInteger counter = new AtomicInteger();
+        registerInvalidationListener(counter);
+        destroyCacheFromCacheManager();
+        if (invokeCacheOperationsFromMember) {
+            // when from a member-side cache manager, invalidation event cannot be received, as the invalidation listener
+            // registration is removed *before* the invocation to destroy the cache is sent to the member
+            assertInvalidationEventNeverReceived(counter);
```
ihsandemir: Isn't this a problem for working of client near cache? Client near cache will never be invalidated in this case and it will have stale values.

Donnerbart: If the backing cache is destroyed the related client Near Cache should also be destroyed, shouldn't it?

vbekiaris: See https://github.com/vbekiaris/hazelcast/blob/0deb3aed4803a85e1346f34a9a3e45c244e46bba/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java#L235-L235

- near cache invalidation listeners are deregistered
- near cache is destroyed on the client side
- once these are done, we proceed with requesting the distributed object be destroyed https://github.com/vbekiaris/hazelcast/blob/0deb3aed4803a85e1346f34a9a3e45c244e46bba/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxy.java#L96-L96. The member delivers the invalidation events but they cannot be listened-for by the client, however the near cache has already been destroyed so it's not a problem.

In any case, this PR is not about _fixing_ invalidation events (one could argue that they are not needed here), it's about restoring behavior to its original state in 3.7 so we don't have any unexpected regression. We can (and should) review usages of invalidation events for 3.8.

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheService.java
ihsandemir: deleteCache is also being called from https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheDestroyOperation.java#L61 with flag set to true and I guess it will have duplicate events in this case.

I just debugged the test case when_cacheDestroyed_invalidationEventIsReceived test case and I see that the CacheDestroyMessageTask and CacheDestroyOperation is not being called. This looks like another bug. Instead of this, the implementation calls DestroyProxyMessageTask at the server side, this does not look correct.

I think due to this bug, you are seeing a single invalidation, if this operation actually worked it would be multiple such events.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheInvalidationTest.java
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    private HazelcastInstance member;
+    private TestHazelcastFactory hazelcastFactory;
+    private NearCacheTestContext testContext;
+
+    @Before
+    public void setup() {
+        hazelcastFactory = new TestHazelcastFactory();
+        member = hazelcastFactory.newHazelcastInstance(getConfig());
```
ihsandemir: start more than one members and you will see when_cacheDestroyed_invalidationEventIsReceived test fails.

vbekiaris: True, member-count invalidation events are delivered (instead of partition-count events in 3.7).

vbekiaris: Amended test to use 2 members & validate that `Cache.destroy` and `CacheManager.destroyCache` will deliver <= member-count events.

ihsandemir:
👍  Need to open an issue for fixing the invalidation event counts to 1 and check if CacheDestroyMessageTask needs to be used.

PR #8657, opened 2016-08-04
Please don't new byte[3] for every single char in String

nilskp:
Seems a bit unnecessary.

---
This change is [Reviewable](https://reviewable.io/reviews/hazelcast/hazelcast/8657)

PR #8810, opened 2016-08-30
Please don't new byte[3] for every single char in String

nilskp:


LoneRifle:
Just a note to say this is a backport of #8657

Issue #8661, opened 2016-08-04
PartitionServiceSafetyCheckTest.clusterShouldBeSafe_withSingleDataMember


hasancelik:
```
java.lang.AssertionError: null
    at org.junit.Assert.fail(Assert.java:86)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at org.junit.Assert.assertTrue(Assert.java:52)
    at com.hazelcast.internal.partition.PartitionServiceSafetyCheckTest.assertSafe(PartitionServiceSafetyCheckTest.java:204)
    at com.hazelcast.internal.partition.PartitionServiceSafetyCheckTest.clusterShouldBeSafe_withSingleDataMember(PartitionServiceSafetyCheckTest.java:86)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/993/testReport/junit/com.hazelcast.internal.partition/PartitionServiceSafetyCheckTest/clusterShouldBeSafe_withSingleDataMember/

hasancelik:
**maintenance**
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.internal.partition.PartitionServiceSafetyCheckTest.assertSafe(PartitionServiceSafetyCheckTest.java:204)
	at com.hazelcast.internal.partition.PartitionServiceSafetyCheckTest.clusterShouldBeSafe_withSingleDataMember(PartitionServiceSafetyCheckTest.java:86)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast%20Maintenance/job/Hazelcast-3.maintenance-IbmJDK1.8/com.hazelcast$hazelcast/107/testReport/junit/com.hazelcast.internal.partition/PartitionServiceSafetyCheckTest/clusterShouldBeSafe_withSingleDataMember/

hasancelik:
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
[...]

PR #8829, opened 2016-09-02
Log non-safe members in "all instances in safe state" assertion

metanet:
It is for getting a better assertion failure in #8661


metanet:
I am merging this one myself since it is a minor test tweak.

PR #8832, opened 2016-09-02
Destroy non-alive client connections

jerrinot:
- Client heartbeater task can now destroy a connection which is not alive
- getConnection() is now also checking for liveness - non-alive connections are destroyed
- tons of debug logging

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -591,7 +591,9 @@ public class ClientConnectionManagerImpl implements ClientConnectionManager {
             fireConnectionAddedEvent(connection);
         } else {
             if (logger.isFinestEnabled()) {
-                logger.finest("Authentication succeeded for " + connection + " and replaced " + oldConnection);
+                logger.finest("Authentication succeeded for " + connection + " and replaced " + oldConnection
+                    + " Closing the old connection.");
+                destroyConnection(oldConnection, "Replaced by newer connection", null);
             }
         }
         assert oldConnection == null || connection.equals(oldConnection);
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -216,7 +216,15 @@ public class ClientConnectionManagerImpl implements ClientConnectionManager {
         if (target == null) {
             return null;
         }
-        return connections.get(target);
+        ClientConnection connection = connections.get(target);
+        if (connection != null && !connection.isAlive()) {
+            if (logger.isFinestEnabled()) {
+                logger.finest("Get a connection " + connection + " which is not alive. Destroying it.");
+            }
+            destroyConnection(connection, "Not Alive", null);
+            return null;
+        }
+        return connection;
     }
 
     @Override
@@ -227,6 +235,7 @@ public class ClientConnectionManagerImpl implements ClientConnectionManager {
                 if (connection != null) {
                     return connection;
                 }
+
                 AuthenticationFuture firstCallback = triggerConnect(address, asOwner);
                 connection = firstCallback.get(connectionTimeout);
                 if (!asOwner) {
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -368,12 +371,29 @@ protected ClientConnection createSocketConnection(final Address address) throws
     public void destroyConnection(final Connection connection, final String reason, final Throwable cause) {
         Address endpoint = connection.getEndPoint();
         ClientConnection conn = (ClientConnection) connection;
```
asimarslan: I think we can eliminate this "conn" var and just use "connection" instead.

cangencer:
is there a test / scenario which this fixes? Could we explain?

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -216,7 +216,15 @@ public ClientConnection getConnection(Address target) {
         if (target == null) {
             return null;
         }
-        return connections.get(target);
+        ClientConnection connection = connections.get(target);
+        if (connection != null && !connection.isAlive()) {
```
cangencer: for a connection to be not alive, connection must have been closed, and it's only closed via destroyConnection(). In which sequence of events can this case happen?

jerrinot:
@cangencer: this is a possible scenario:
- open a connection client -> member
- send auth message
- receive auth reply -> reply processing is offloaded to an executor
- member closes the connection -> the connection is will never be scheduled for writing
- the executor start processing the auth reply -> it put the connection to the connection map.
```

There is a good chance there are other similar races.

File: hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java (hunk 2 above)
Comment (jerrinot) [inline hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java]: see below

File: hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java (hunk 1 above)
Comment (jerrinot) [inline hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java]: @asimarslan: done

**Comment by cangencer** [comment]
So a non-live connection might be put into the connections map, so it makes sense to check the liveness at the point of retrieval and also sweet it during the Heartbeat check.

👍

**Comment by asimarslan** [comment]
👍

File: hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java [hunk 3]
```diff
@@ -561,14 +584,24 @@ public void run() {
...
         ClientConnection oldConnection = connections.put(connection.getRemoteEndpoint(), connection);
         if (oldConnection == null) {
+            if (logger.isFinestEnabled()) {
+                logger.finest("Authentication succeeded for " + connection
+                        + " and there was no old connection to this end-point");
+            }
             fireConnectionAddedEvent(connection);
+        } else {
+            if (logger.isFinestEnabled()) {
+                logger.finest("Authentication succeeded for " + connection + " and replaced " + oldConnection);
```
ihsandemir: Why not close old connection?

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -216,7 +216,15 @@ public ClientConnection getConnection(Address target) {
         if (target == null) {
             return null;
         }
-        return connections.get(target);
+        ClientConnection connection = connections.get(target);
+        if (connection != null && !connection.isAlive()) {
+            if (logger.isFinestEnabled()) {
+                logger.finest("Get a connection " + connection + " which is not alive. Destroying it.");
+            }
+            destroyConnection(connection, "Not Alive", null);
```
ihsandemir: Connection destroyları make take time notifying the listeners and since it is just for cleanup may be done using e.g. internalExecutor.

ihsandemir: in general, i think that we still need to find the root causes why those connections can stay dangling like that. The solution will eliminate after affects of such a case and may hide the actual problem why they happen at the first place.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
jerrinot: good one. I added a commit.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
jerrinot: I agree. I explained one scenario above, but there might be more. This whole system should be simplified.

PR #8834, opened 2016-09-02
Fix for NoSuchFieldException when running under non-HotSpot JVM

vbekiaris:
In non-Hotspot JVMs `JVMUtil.isCompressedOopsOrNull` will attempt to determine the memory cost of reference by means of field location offset, however `getField` will only succeed if the field is public.

Also, instead of logging at `WARNING` level each exception, in this PR:
* log at `INFO` level in case all attempts to determine memory cost failed and we use default of 4 bytes
* log at `FINE` level the individual exceptions which may have occurred when attempting to determine the memory cost via Hotspot-specific and non-Hotspot methods.

closes https://github.com/hazelcast/hazelcast/issues/8835

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/JVMUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/JVMUtil.java
@@ -36,12 +36,19 @@ public final class JVMUtil {
      * Defines the costs for a reference in Bytes.
      */
     @SuppressWarnings("checkstyle:magicnumber")
-    public static final int REFERENCE_COST_IN_BYTES = isCompressedOops() ? 4 : 8;
+    public static final int REFERENCE_COST_IN_BYTES = is32bitJVM() || isCompressedOops() ? 4 : 8;
 
     private JVMUtil() {
     }
 
-    private static boolean isCompressedOops() {
+    static boolean is32bitJVM() {
+        // sun.arch.data.model is available on Oracle, Zing and (most probably) IBM JVMs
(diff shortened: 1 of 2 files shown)


ahmetmircik:
👍

ahmetmircik:
backport?

hazelcast/src/main/java/com/hazelcast/util/JVMUtil.java
```diff
@@ -36,12 +36,19 @@
      * Defines the costs for a reference in Bytes.
      */
     @SuppressWarnings("checkstyle:magicnumber")
-    public static final int REFERENCE_COST_IN_BYTES = isCompressedOops() ? 4 : 8;
+    public static final int REFERENCE_COST_IN_BYTES = is32bitJVM() || isCompressedOopsOrNull() ? 4 : 8;
```
ahmetmircik: should this be `isCompressedOops` instead of `isCompressedOopsOrNull `? Maybe we can think a better name for `isCompressedOopsOrNull` to prevent confusion.

vbekiaris: Yes, updated so that the name indicates that we use object layout to determine if compressed oops is enabled or not.

vbekiaris:
Updated with 32-bit JVM detection; verified on OpenJDK, Oracle & Azul Zing JVMs.

Donnerbart:
:+1: (once green)

PR #8836, opened 2016-09-03
Fix race in static getLogger()

jerrinot:
See commit messages for details.

It's easier to review it commit-by-commit.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/logging/Logger.java
+++ b/hazelcast/src/main/java/com/hazelcast/logging/Logger.java
@@ -44,9 +44,9 @@ public final class Logger {
      */
     public static ILogger getLogger(String name) {
         LoggerFactory loggerFactoryToUse = loggerFactory;
-        //fast-track when factory is initialized
-        if (loggerFactory != null) {
-            return loggerFactory.getLogger(name);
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/logging/Logger.java
```diff
@@ -44,27 +44,32 @@ public static ILogger getLogger(Class clazz) {
...
+            if (loggerFactory == null) {
+                String loggerType = System.getProperty("hazelcast.logging.type");
+                loggerFactoryToUse = newLoggerFactory(loggerType);
+                if (!StringUtil.isNullOrEmpty(loggerType)) {
+                    loggerFactory = loggerFactoryToUse;
                 }
             }
         }
+        //loggerFactory was initialized -> loggerFactoryToUse was left to null
+        if (loggerFactoryToUse == null) {
```
jerrinot: this is the important bit - if the `loggerFactory` field was initialized before our thread acquired the lock then `loggerFactoryToUse` will be null.

tombujok:
👍

vbekiaris:
👍 good finding!

PR #8842, opened 2016-09-05
Remember cluster state when the cluster is started after graceful cluster shutdown

metanet:
If cluster.shutdown() is invoked when cluster is in the ACTIVE state, the final persisted state on disk will be still ACTIVE. Therefore, nodes will come back with ACTIVE state when they are restarted. The same rule also applies for the FROZEN state. Cluster still moves to PASSIVE state during the shutdown but now it is a transient state which is not persisted to disk.

EE PART: https://github.com/hazelcast/hazelcast-enterprise/pull/1037

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
@@ -104,24 +104,26 @@ public class ClusterStateManager {
                         + initialState);
                 return;
             }
-            this.state = initialState;
-            changeNodeState(initialState);
-            node.getNodeExtension().onClusterStateChange(initialState, false);
+            doSetClusterState(initialState, true);
         } finally {
             clusterServiceLock.unlock();
         }
     }
 
-    void setClusterState(ClusterState newState, boolean persistentChange) {
+    void setClusterState(ClusterState newState, boolean isTransient) {
         clusterServiceLock.lock();
         try {
-            this.state = newState;
-            stateLockRef.set(ClusterStateLock.NOT_LOCKED);
-            changeNodeState(newState);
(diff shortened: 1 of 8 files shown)


mdogan:
👍

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
```diff
@@ -201,6 +203,10 @@ public boolean rollbackClusterState(String txnId) {
     }
 
     public void commitClusterState(ClusterState newState, Address initiator, String txnId) {
```
mmedenjak: Don't really need this method. It is only used from the tests and they can all just set the isTransient to false.

mmedenjak:
Please just fix the merge conflict.

mdogan:
@mmedenjak: Right, `isPersistent` to `isTransient` change is not a necessity. 
But I think, being a `transient` cluster state change is more expressive than a `persistent` one, especially if we don't think this from hot-restart point of view.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
```diff
@@ -104,26 +104,28 @@ void initialClusterState(ClusterState initialState) {
                         + initialState);
                 return;
             }
-            this.state = initialState;
-            changeNodeState(initialState);
-            node.getNodeExtension().onClusterStateChange(initialState, false);
+            doSetClusterState(initialState, true);
```
mmedenjak: Beware that here the method call will additionally call `stateLockRef.set(ClusterStateLock.NOT_LOCKED);` in contrast to the old code.

metanet:
Thank you @mmedenjak for the review. I will rebase the branch and update the pr

mdogan:
run-lab-run

Issue #8847, opened 2016-09-05
Hazelcast observers are lost during disconnects


ihsandemir:
When clients lose connection to the nodes in the cluster, even after the clients reconnected to the cluster, the Entry/MessageListeners never fired again.

When this happens, it happens silently & there are no log messages to give any indication that there is a problem. (You can see the client lose connection & reconnect, but see no sign that the listeners have not been restored)

PR #8848, opened 2016-09-05
Hazelcast listeners are lost during cluster disconnects

ihsandemir:
Fixes the client listener registration if the client owner connection is lost and reconnected (e.g. due to heartbeat or network problems).

Fixes https://github.com/hazelcast/hazelcast/issues/8847


ihsandemir:
Our smart listener mechanism works by client keeping track of all the registered members' registrations. The following cases are possible:

1. Client disconnects from the cluster if network between the owner server and the client is somehow unreliable which causes the client disconnect due to heartbeat timeout. In this case, it is also possible that the server may not close the connection if the server side heartbeat timeout is longer. And later on the client reconnects to the cluster by connecting to another member. 

2. Case 1 happens and the owner member also closes the connection to the client.

3. The owner server is forcefully shutdown(terminate called) or killed with SIGKILL. Owner server will not be able to do any cleanup (in this case remove all he listeners from the cluster on client disconnect).

4. Both the client and the owner member crashes at the same time. The client resources will be left in the cluster forever.

5. It is also observed that if while owner member is shutdown gracefully, it fails to send the ClientDisconnectOperation to other members. This operation does the cleanup of the client resources in the cluster including the listeners and the locks for the client.

We need a solution to solve these problems.

A solution: Add lifecycle listener to client. On CLUSTER_CONNECTED state, the client re-registers all the existing listeners to each member. This solves cases 1, 2, 3(only for listeners), 4(N/A). 5(for the listeners). The problem with this approach is that if ClientDisconnectOperation is run at the member side after the client is reconnected to the cluster and reregistered all the listeners, then the operation will remove all the existing listeners of the client.

ihsandemir:
After applying the solution and adding extra tests, I realised that there is another problem that needs to be fixed. The problem is:

Case 1 (see previous comment): When client reconnects, registers the listeners but then when the connection is restored, the old owner member detects that the client is disconnected and runs the ClientDisconnectedOperation in the whole cluster, removing all the listeners.

Case 2: Just like Case 1, client resources are cleaned fro the cluster. If client reconnects to the cluster after the cleanup complete, reregistering the listeners are ok. But, if client reregisters and cleanup is executed later, then there may be duplicate listeners until the cleanup works and then no listeners after the disconnected operation is run.

In order to solve such cases, a possible solution:

Make the client more intelligent during owner authentication. i.e. Let the owner authentication response return a flag indicating whether all the cluster resources for the client is already removed or not. Then if already removed, reregister the listeners to the whole cluster. This solution also assumes that if only one member ran the disconnected operation but others did not execute it yet, before returning the authentication response, it has to make sure that all the operations are run so that cluster is clean. If the owner member is removed from the cluster, DestroyEndpointTask runs 10 seconds after the member removal. When it is running it should not ClientDisconnectionOperation if the client is already reconnected to the cluster (i.e. the client uuid, owner uuid mapping exists in ownershipMappings and the owner is a different member, this may have a drawback what if only this is true for two out of 5 members in the cluster?).

hasancelik:
run-lab-run

ihsandemir:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

ihsandemir:
I had to make some changes in the solution. The change will be to return the list of members to the client which executed the cleanup operation and let the client re-register to these members. Also for backward compatibility, i just make sure that the cleanup is performed to keep the old behaviour. I may define a new operation that implements AllowedDuringPassiveState interface which makes no partition data change which can be executed in cluster passive state.

ihsandemir:
run-lab-run

ihsandemir:
run-lab-run

ihsandemir:
run-lab-run

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -366,40 +366,62 @@ public void connectionAdded(Connection conn) {
...
 
         private void callDisconnectionOperation(ClientEndpointImpl endpoint) {
             Collection<Member> memberList = nodeEngine.getClusterService().getMembers();
             OperationService operationService = nodeEngine.getOperationService();
-            ClientDisconnectionOperation op = createClientDisconnectionOperation(endpoint.getUuid());
+            String memberUuid = getLocalMember().getUuid();
+            String clientUuid = endpoint.getUuid();
+
+            String ownerMember = ownershipMappings.get(clientUuid);
+            if (memberUuid != ownerMember) {
```
ihsandemir: Member is uniquely identified by the uuid, hence i thought that it is OK to compare uuid strings.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
```diff
@@ -82,6 +84,9 @@
     private int connectedServerVersion = BuildInfo.UNKNOWN_HAZELCAST_VERSION;
     private String connectedServerVersionString;
 
+    private List<Member> clientUnregisteredMembers;
```
ihsandemir: OK, added comment.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
```diff
@@ -16,41 +16,58 @@
...
 import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 public class ClientSmartListenerService extends ClientListenerServiceImpl implements InitialMembershipListener {
-
+    private static final long SMART_LISTENER_MEMBER_ADDED_RESCHEDULE_TIME = 1000;
+    private static final long SMART_LISTENER_CONNECT_ALL_SERVERS_RETRY_WAIT_TIME = 5000;
+    private static final String SMART_LISTENER_SERVICE_CONNECTION_OPENER = "Smart Listener ConnectionOpener";
     private final Set<Member> members = new HashSet<Member>();
```
ihsandemir: OK, added comments.

ihsandemir:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -131,7 +131,8 @@ public ClientConnectionManagerImpl(HazelcastClientInstanceImpl client, AddressTr
         ClientNetworkConfig networkConfig = config.getNetworkConfig();
 
         final int connTimeout = networkConfig.getConnectionTimeout();
-        this.connectionTimeout = connTimeout == 0 ? Integer.MAX_VALUE : connTimeout;
+        final int attemptPeriod = networkConfig.getConnectionAttemptPeriod();
+        this.connectionTimeout = Math.min(connTimeout, attemptPeriod);
```
vbekiaris: Previously, if `connTimeout == 0` then `this.connectionTimeout` would become `Integer.MAX_VALUE`, whereas now, assuming a larger `attemptPeriod` (for example `3000` which is the default), `this.connectionTimeout` will be set to 0. Doesn't this affect behaviour of code that uses `ClientConnectionManagerImpl.connectionTimeout`?

ihsandemir: Ok, reverted this change. Modified the test and added the connection timeout config to the tests client config.

ihsandemir:
run-lab-run

tombujok:
💯 for merging ;)

PR #8853, opened 2016-09-06
Backport #8657: Please don't new byte[3] for every single char in String

LoneRifle:
Credit to @nilskp for the original commit, to be patched over to the maintenance branch.

Supersedes #8810.


Donnerbart:
run-lab-run

PR #8854, opened 2016-09-07
Made EvictionConfig the new default way to configure Near Caches

Donnerbart:
* Cleanup of `NearCacheConfig` and related classes
* Deprecated the old `max-size` and `eviction-policy` configuration
* Added functionality to return the new eviction configuration if the old configuration style was used
* Added fallback for old `max-size` default value if size was not configured explicitly

No EE adaption is needed.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
@@ -19,6 +19,7 @@ package com.hazelcast.client.map.impl.nearcache;
 import com.hazelcast.cache.impl.nearcache.NearCache;
 import com.hazelcast.client.spi.ClientContext;
 import com.hazelcast.client.spi.impl.ClientExecutionServiceImpl;
+import com.hazelcast.config.EvictionConfig;
 import com.hazelcast.config.EvictionPolicy;
 import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.config.NearCacheConfig;
@@ -37,6 +38,7 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import static com.hazelcast.config.EvictionConfigAccessor.initDefaultMaxSize;
 import static com.hazelcast.config.EvictionPolicy.NONE;
 import static com.hazelcast.util.ExceptionUtil.rethrow;
 
@@ -85,11 +87,12 @@ public class ClientHeapNearCache<K> implements NearCache<K, Object> {
         if (inMemoryFormat != InMemoryFormat.BINARY && inMemoryFormat != InMemoryFormat.OBJECT) {
             throw new IllegalArgumentException("Illegal in-memory-format: " + inMemoryFormat);
         }
-        this.maxSize = nearCacheConfig.getMaxSize();
+        EvictionConfig evictionConfig = initDefaultMaxSize(nearCacheConfig.getEvictionConfig());
+        this.maxSize = evictionConfig.getSize();
         this.maxIdleMillis = TimeUnit.SECONDS.toMillis(nearCacheConfig.getMaxIdleSeconds());
         this.timeToLiveMillis = TimeUnit.SECONDS.toMillis(nearCacheConfig.getTimeToLiveSeconds());
         this.invalidateOnChange = nearCacheConfig.isInvalidateOnChange();
-        this.evictionPolicy = EvictionPolicy.valueOf(nearCacheConfig.getEvictionPolicy());
+        this.evictionPolicy = evictionConfig.getEvictionPolicy();
         this.canExpire = new AtomicBoolean(true);
         this.canEvict = new AtomicBoolean(true);
         this.stats = new NearCacheStatsImpl();
--- a/hazelcast-client/src/main/resources/hazelcast-client-config-3.8.xsd
+++ b/hazelcast-client/src/main/resources/hazelcast-client-config-3.8.xsd
@@ -349,10 +349,22 @@
 
     <xs:complexType name="near-cache">
         <xs:all>
-            <xs:element name="max-size" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0"/>
+            <xs:element name="max-size" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0">
+                <xs:annotation>
+                    <xs:documentation>
+                        Deprecated since 3.8, please use &lt;eviction/&gt;
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
             <xs:element name="time-to-live-seconds" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0"/>
             <xs:element name="max-idle-seconds" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0"/>
-            <xs:element name="eviction-policy" type="eviction-policy" minOccurs="0" maxOccurs="1" default="LRU"/>
+            <xs:element name="eviction-policy" type="eviction-policy" minOccurs="0" maxOccurs="1" default="LRU">
+                <xs:annotation>
+                    <xs:documentation>
+                        Deprecated since 3.8, please use &lt;eviction/&gt;
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
             <xs:element name="invalidate-on-change" type="xs:boolean" minOccurs="0" maxOccurs="1" default="true"/>
             <xs:element name="in-memory-format" type="in-memory-format" minOccurs="0" maxOccurs="1" default="BINARY"/>
             <xs:element name="local-update-policy" type="xs:string" default="INVALIDATE" minOccurs="0" maxOccurs="1"/>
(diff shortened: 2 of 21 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/ClientRegressionWithMockNetworkTest.java
```diff
@@ -593,17 +595,27 @@ public void stateChanged(LifecycleEvent event) {
     }
 
     @Test
-    public void testDeadlock_WhenDoingOperationFromLifecycleListener_and_NearCache() {
-        HazelcastInstance instance = hazelcastFactory.newHazelcastInstance();
-        final ClientConfig clientConfig = new ClientConfig();
-        final NearCacheConfig nearCacheConfig = new NearCacheConfig();
```
Donnerbart: I think this `NearCacheConfig` never did anything, since it was not configured with a name. So the client map never picked it up.

hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
```diff
@@ -619,21 +619,6 @@ private void mapStoreConfigXmlGenerator(StringBuilder xml, MapConfig m) {
         }
     }
 
-    private void nearCacheConfigXmlGenerator(StringBuilder xml, NearCacheConfig n) {
```
Donnerbart: Was dead code.

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTestSupport.java
```diff
@@ -117,12 +119,15 @@ public void run() {
...
-                .setMaxSize(size)
-                .setEvictionPolicy(evictionPolicy.name());
+                .setEvictionConfig(new EvictionConfig(size, ENTRY_COUNT, evictionPolicy));
     }
 
     protected NearCacheConfig newNearCacheConfig() {
-        return new NearCacheConfig();
+        NearCacheConfig nearCacheConfig = new NearCacheConfig();
+
+        nearCacheConfig.getEvictionConfig().setSize(NearCacheConfig.DEFAULT_MAX_SIZE);
```
Donnerbart: Had to set the default value from the deprecated `max-size` configuration here, since the new `EvictionConfig` has a way smaller default (`10000` vs. `Integer.MAX_VALUE`). This is a behavior change users might run into!

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -85,11 +86,12 @@ public ClientHeapNearCache(String mapName, ClientContext context, NearCacheConfi
         if (inMemoryFormat != InMemoryFormat.BINARY && inMemoryFormat != InMemoryFormat.OBJECT) {
             throw new IllegalArgumentException("Illegal in-memory-format: " + inMemoryFormat);
         }
-        this.maxSize = nearCacheConfig.getMaxSize();
+        EvictionConfig evictionConfig = nearCacheConfig.getEvictionConfig();
+        this.maxSize = evictionConfig.isSizeConfigured() ? evictionConfig.getSize() : NearCacheConfig.DEFAULT_MAX_SIZE;
```
Donnerbart: This is the fallback to the old `DEFAULT_MAX_SIZE`, if the size was not configured explicitly.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
```diff
@@ -84,12 +85,13 @@ public NearCacheImpl(String mapName, NodeEngine nodeEngine, SizeEstimator<NearCa
 
         Config config = nodeEngine.getConfig();
         NearCacheConfig nearCacheConfig = config.findMapConfig(mapName).getNearCacheConfig();
+        EvictionConfig evictionConfig = nearCacheConfig.getEvictionConfig();
         this.inMemoryFormat = nearCacheConfig.getInMemoryFormat();
-        this.maxSize = nearCacheConfig.getMaxSize() <= 0 ? Integer.MAX_VALUE : nearCacheConfig.getMaxSize();
+        this.maxSize = evictionConfig.isSizeConfigured() ? evictionConfig.getSize() : NearCacheConfig.DEFAULT_MAX_SIZE;
```
Donnerbart: This is the fallback to the old `DEFAULT_MAX_SIZE`, if the size was not configured explicitly.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -85,11 +87,12 @@ public ClientHeapNearCache(String mapName, ClientContext context, NearCacheConfi
         if (inMemoryFormat != InMemoryFormat.BINARY && inMemoryFormat != InMemoryFormat.OBJECT) {
             throw new IllegalArgumentException("Illegal in-memory-format: " + inMemoryFormat);
         }
-        this.maxSize = nearCacheConfig.getMaxSize();
+        EvictionConfig evictionConfig = initDefaultMaxSize(nearCacheConfig.getEvictionConfig());
```
Donnerbart: This initializes the old default max size value, if no size was configured by the user.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheImpl.java
```diff
@@ -84,12 +86,13 @@ public NearCacheImpl(String mapName, NodeEngine nodeEngine, SizeEstimator<NearCa
 
         Config config = nodeEngine.getConfig();
         NearCacheConfig nearCacheConfig = config.findMapConfig(mapName).getNearCacheConfig();
+        EvictionConfig evictionConfig = initDefaultMaxSize(nearCacheConfig.getEvictionConfig());
```
Donnerbart: This initializes the old default max size value, if no size was configured by the user.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -53,6 +52,7 @@
      */
     public static final EvictionPolicy DEFAULT_EVICTION_POLICY = EvictionPolicy.LRU;
 
+    protected boolean isSizeConfigured;
```
Donnerbart: I just added this field and an extra `EvictionConfigAccessor` class, so the interface of the configuration itself didn't change.

Donnerbart:
run-lab-run

hazelcast-client/src/main/resources/hazelcast-client-config-3.8.xsd
```diff
@@ -349,10 +349,22 @@
 
     <xs:complexType name="near-cache">
         <xs:all>
-            <xs:element name="max-size" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0"/>
+            <xs:element name="max-size" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0">
+                <xs:annotation>
+                    <xs:documentation>
+                        Deprecated since 3.8, please use <eviction/>
```
Donnerbart: There seems to be no support for deprecation in the actual XSD standard, so I added it as normal documentation text.

jerrinot: shouldn't the `<eviction/>` be encoded?

hazelcast-client/src/main/resources/hazelcast-client-config-3.8.xsd
```diff
@@ -349,10 +349,22 @@
...
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
             <xs:element name="time-to-live-seconds" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0"/>
             <xs:element name="max-idle-seconds" type="xs:unsignedInt" minOccurs="0" maxOccurs="1" default="0"/>
-            <xs:element name="eviction-policy" type="eviction-policy" minOccurs="0" maxOccurs="1" default="LRU"/>
+            <xs:element name="eviction-policy" type="eviction-policy" minOccurs="0" maxOccurs="1" default="LRU">
+                <xs:annotation>
+                    <xs:documentation>
+                        Deprecated since 3.8, please use <eviction/>
```
jerrinot: same here

hazelcast-client/src/main/resources/hazelcast-client-config-3.8.xsd
Donnerbart: That's a very good question, I have little experience with XSD. A `{@link ...}` would be great, but I doubt that it exists.

Donnerbart: I've encoded it as `&lt;eviction/&gt;` now.

hazelcast-client/src/main/resources/hazelcast-client-config-3.8.xsd
Donnerbart: Done

hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
vbekiaris: I think that we may be actually missing this method's invocation to output the near-cache configuration in `ConfigXmlGenerator.generate`.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
vbekiaris: Tiny one: maybe rename field to `sizeConfigured`?

hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
Donnerbart: I fixed this upfront in https://github.com/hazelcast/hazelcast/pull/8912
Will adapt this PR to check for both configurations after it is merged.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
Donnerbart: Done

Donnerbart:
run-lab-run

PR #8862, opened 2016-09-08
Unified invalidations for member and client near-caches

ahmetmircik:


hazelcast/src/main/java/com/hazelcast/map/impl/operation/ClearNearCacheOperation.java
```diff
@@ -18,6 +18,7 @@
 
 import com.hazelcast.spi.impl.MutatingOperation;
 
+// TODO not used operation remove it.
```
Donnerbart: It's still used in `MapClearNearCacheMessageTask`.

hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheBatchInvalidationTest.java
```diff
@@ -153,7 +153,7 @@ public void testMapClear_shouldClearNearCaches_onOwnerAndBackupNodes() throws Ex
         String mapName = randomMapName();
         Config config = newConfig(mapName);
         config.setProperty(GroupProperty.PARTITION_COUNT.getName(), "1");
-        configureBatching(config, true, Integer.MAX_VALUE, Integer.MAX_VALUE);
+        configureBatching(config, true, 5, 5);
```
Donnerbart: Since these two parameters are the `hazelcast.map.invalidation.batch.size` and `hazelcast.map.invalidation.batchfrequency.seconds` and `Integer.MAX_VALUE` basically should configure this system unusable, I guess it was not used before (due to the distinct operations), and know it is used and needs a more realistic configuration to work.

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
```diff
@@ -117,11 +117,18 @@ public void testBasicUsage() {
         }
 
         map.clear();
-        for (HazelcastInstance instance : instances) {
-            NearCache nearCache = getNearCache(mapName, instance);
-            int size = nearCache.size();
-            assertEquals("Near Cache size should be 0 but was " + size, 0, size);
-        }
+
+        assertTrueEventually(new AssertTask() {
```
Donnerbart: If we need an `assertTrueEventually` after a `map.clear()` now, this looks like a noticeable behavior change to me. Users might actually rely on the fact, that the Near Cache is already invalidated, when `map.clear()` returns. This will at least need a good documentation in the release notes.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/LoadAllOperation.java
```diff
@@ -61,15 +61,6 @@ public void run() throws Exception {
         recordStore.loadAllFromStore(keys);
     }
 
-    @Override
-    public void afterRun() throws Exception {
```
Donnerbart: This scenario is tested in `NearCacheTest` via `testAfterLoadAllWithDefinedKeysNearCacheIsInvalidated()` and `testAfterLoadAllNearCacheIsInvalidated()`.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -960,18 +957,12 @@ protected boolean evictInternal(Data keyData) {
     public void evictAll() {
         ClientMessage request = MapEvictAllCodec.encodeRequest(name);
         invoke(request);
-
-        clearNearCachesOnLiteMembers();
```
Donnerbart: Should be tested via `ClientMapNearCacheLiteMemberTest.testEvictAll()`.

:+1:

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -960,18 +957,12 @@ protected boolean evictInternal(Data keyData) {
...
     }
 
     @Override
     public void loadAll(boolean replaceExistingValues) {
         ClientMessage request = MapLoadAllCodec.encodeRequest(name, replaceExistingValues);
         invoke(request);
-
-        if (replaceExistingValues) {
```
Donnerbart: Could not find a test for this, so I wrote one: https://github.com/hazelcast/hazelcast/pull/8869/files#diff-ec815c78e843f777f47642c555a0d9b2R140

:+1:

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -983,10 +974,6 @@ public void loadAll(Set<K> keys, boolean replaceExistingValues) {
 
         Collection<Data> dataKeys = CollectionUtil.objectToDataCollection(keys, getSerializationService());
         loadAllInternal(replaceExistingValues, dataKeys);
-
-        if (replaceExistingValues) {
```
Donnerbart: Could not find a test for this, so I wrote one: https://github.com/hazelcast/hazelcast/pull/8869/files#diff-ec815c78e843f777f47642c555a0d9b2R147

:+1:

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1428,22 +1415,12 @@ protected void putAllInternal(Map<Integer, List<Map.Entry<Data, Data>>> entryMap
     public void clear() {
         ClientMessage request = MapClearCodec.encodeRequest(name);
         invoke(request);
-
-        clearNearCachesOnLiteMembers();
```
Donnerbart: Should be tested via `ClientMapNearCacheLiteMemberTest.testClear()`.

:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -79,9 +80,8 @@ public Data apply(Object input) {
         }
     };
     protected final ConstructorFunction<Void, RecordFactory> recordFactoryConstructor;
-    protected final boolean memberNearCacheInvalidationEnabled;
     /**
-     * Holds number of registered {@link com.hazelcast.map.impl.nearcache.InvalidationListener} from clients.
+     * Holds number of registered {@link InvalidationListener} from clients.
```
Donnerbart: We should put this `com.hazelcast.map.impl.nearcache.InvalidationListener` import back into the JavaDoc, it's not used otherwise.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EvictAllOperation.java
```diff
@@ -46,10 +46,6 @@ public EvictAllOperation(String name) {
 
     @Override
     public void run() throws Exception {
-
-        // TODO this also clears locked keys from near cache which should be preserved.
-        clearLocalNearCache();
```
Donnerbart: I couldn't find a test for this scenario, so I wrote one: https://github.com/hazelcast/hazelcast/pull/8869/files#diff-1255ed0d931aed11a9aa83ddb3864b8fR799

:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EvictAllBackupOperation.java
```diff
@@ -37,8 +37,6 @@ public EvictAllBackupOperation(String name) {
 
     @Override
     public void run() throws Exception {
-        clearLocalNearCache();
```
Donnerbart: I couldn't find a test for this scenario, so I wrote one: https://github.com/hazelcast/hazelcast/pull/8869/files#diff-1255ed0d931aed11a9aa83ddb3864b8fR799

:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/operation/ClearOperation.java
```diff
@@ -41,10 +41,6 @@ public ClearOperation(String name) {
 
     @Override
     public void run() {
-        // near-cache clear will be called multiple times by each clear operation,
```
Donnerbart: Should be tested via `NearCacheTest.testMapClear_clears_localNearCache`.

:+1:

hazelcast/src/main/java/com/hazelcast/map/impl/operation/ClearBackupOperation.java
```diff
@@ -33,10 +33,6 @@ public ClearBackupOperation(String name) {
 
     @Override
     public void run() {
-        // clear near-cache also on this backup operation, there is a possibility that no owner partition exists on this
-        // node but a near-cache exists.
-        clearLocalNearCache();
```
Donnerbart: Should be tested via `NearCacheTest.testMapClear_clears_localNearCache`.

:+1:

Donnerbart:
In general I really like this PR. The missing test cases were simple to write and seem to pass.

I'm just a bit concerned about the `map.clear()` requiring an `assertTrueEventually()` now, which might be an unexpected change in behavior.

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
ahmetmircik: For `assertTrueEventually`: Waiting is needed to see remote node near-caches are also cleared, local near-cache will be cleared already when `map.clear()` returns. means no difference from caller sides' perspective. Also, as of now, member-near-caches' behaviour is inlined with existing client side near-caches.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/ClearNearCacheOperation.java
ahmetmircik: Yes it will not be used any more but we will keep it for client protocol backward compatibility reasons.

hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
Donnerbart: Yes, but it was not needed before, was it? And that was my point. This looks like a behavioral change to me, so we need at least to document this in the release notes.

ahmetmircik: i see your point, it is ok to have it on release notes, will work on it with @Serdaro

PR #8865, opened 2016-09-09
Fixed wrong calculation of Near Cache owned memory costs

Donnerbart:
* Added an additional tests for Near Cache memory cost calculation
* Fixed wrong calculation of Near Cache owned memory costs

All Near Caches implementations which are based on the `AbstractNearCacheRecordStore` suffer from a wrong calculation of owned memory costs. Whenever a record in the Near Cache is updated, we upfront add the memory costs for its key + value:
```
BaseHeapNearCacheRecordStore extends AbstractNearCacheRecordStore

@Override
protected R putRecord(K key, R record) {
    R oldRecord = records.put(key, record);
    nearCacheStats.incrementOwnedEntryMemoryCost(getTotalStorageMemoryCost(key, record));
    return oldRecord;
}
```
If `oldRecord` exists, we detract just the costs for the value, but not the key:
```
AbstractNearCacheRecordStore

oldRecord = putRecord(key, record);
if (oldRecord == null) {
    nearCacheStats.incrementOwnedEntryCount();
} else {
    long oldRecordMemoryCost = getRecordStorageMemoryCost(oldRecord);
    nearCacheStats.decrementOwnedEntryMemoryCost(oldRecordMemoryCost);
}
```
So in the end each record update piles up the costs of a key.

The new tests were failing like this before the fix was applied:
```
java.lang.AssertionError: The Near Cache is empty, there should be no owned entry memory costs 
Expected :0
Actual   :3696
```
The easiest fix is to detract the costs for key + value if an old record was found. So the upfront added key costs are removed again. We could also skip adding the key costs, but then we would have to duplicate this logic in all record store implementations (OS + EE). With the actual approach, it's just one line in the abstract base class.

There is no EE adaption needed, since the tests work for all scenarios.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTest.java
@@ -161,4 +161,24 @@ public class ClientNearCacheTest extends ClientNearCacheTestSupport {
     public void testNearCacheIdleRecordsExpired_withBinaryInMemoryFormat() {
         testNearCacheExpiration_withMaxIdle(InMemoryFormat.BINARY);
     }
+
+    @Test
+    public void testNearCacheMemoryCostCalculation_withObjectInMemoryFormat() {
+        testNearCacheMemoryCostCalculation(InMemoryFormat.OBJECT, 1);
+    }
+
+    @Test
+    public void testNearCacheMemoryCostCalculation_withConcurrentCacheMisses_withObjectInMemoryFormat() {
+        testNearCacheMemoryCostCalculation(InMemoryFormat.OBJECT, 10);
+    }
+
+    @Test
+    public void testNearCacheMemoryCostCalculation_withBinaryInMemoryFormat() {
+        testNearCacheMemoryCostCalculation(InMemoryFormat.BINARY, 1);
+    }
+
+    @Test
+    public void testNearCacheMemoryCostCalculation_withConcurrentCacheMisses_withBinaryInMemoryFormat() {
+        testNearCacheMemoryCostCalculation(InMemoryFormat.BINARY, 10);
+    }
 }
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/AbstractNearCacheRecordStore.java
```diff
@@ -286,7 +286,7 @@ public void put(K key, V value) {
             if (oldRecord == null) {
                 nearCacheStats.incrementOwnedEntryCount();
             } else {
-                long oldRecordMemoryCost = getRecordStorageMemoryCost(oldRecord);
+                long oldRecordMemoryCost = getTotalStorageMemoryCost(key, oldRecord);
```
Donnerbart: This is the fix.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
```diff
@@ -629,6 +631,46 @@ public void run() {
         });
     }
 
+    protected void testNearCacheMemoryCostCalculation(InMemoryFormat inMemoryFormat, int threadCount) {
```
vbekiaris: Since `OBJECT` format does not perform memory cost calculation, why not return immediately in this case?

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTest.java
```diff
@@ -161,4 +161,24 @@ public void testNearCacheIdleRecordsExpired_withObjectInMemoryFormat() {
     public void testNearCacheIdleRecordsExpired_withBinaryInMemoryFormat() {
         testNearCacheExpiration_withMaxIdle(InMemoryFormat.BINARY);
     }
+
+    @Test
+    public void testNearCacheMemoryCostCalculation_withObjectInMemoryFormat() {
```
vbekiaris: This test class could be a good candidate for deduplication using parametrization (perhaps in another PR?).

vbekiaris:
👍 good finding!

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
Donnerbart: That a good point, but I think we should check for it. I've added an explicit check that there are no memory costs being calculated for `OBJECT` in-memory format.

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTest.java
Donnerbart: Good idea!

Issue #8868, opened 2016-09-12
JUnit dependency in com.hazelcast.test.mocknetwork.MockJoiner


rvanheest:
I came across the testing/mocking framework that you provided in the hazelcast library. So I just started to use `new TestHazelcastInstanceFactory(1).newHazelcastInstance(config)` in my tests rather than spinning up a real instance. One improvement is that running the corresponding tests is now 4 times as fast!

However, as I use Scala with ScalaTest as my testing framework, I found I also have to import the JUnit library. The reason for this is because `com.hazelcast.test.mocknetwork.MockJoiner` has a dependency on `org.junit.Assert`. When the JUnit dependency is not included by hand (e.g. in my own `pom.xml`), I get a `java.lang.NoClassDefFoundError: org/junit/Assert` while the test instance is spinning up:

```
Sep 12, 2016 9:13:28 AM com.hazelcast.system
INFO: [127.0.0.1]:5001 [dev] [3.6.4] Hazelcast 3.6.4 (20160701 - 5b94d9f) starting at Address[127.0.0.1]:5001
Sep 12, 2016 9:13:28 AM com.hazelcast.system
INFO: [127.0.0.1]:5001 [dev] [3.6.4] Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
Sep 12, 2016 9:13:28 AM com.hazelcast.system
INFO: [127.0.0.1]:5001 [dev] [3.6.4] Configured Hazelcast Serialization version : 1
Sep 12, 2016 9:13:28 AM com.hazelcast.spi.OperationService
INFO: [127.0.0.1]:5001 [dev] [3.6.4] Backpressure is disabled
Sep 12, 2016 9:13:28 AM com.hazelcast.spi.impl.operationexecutor.classic.ClassicOperationExecutor
INFO: [127.0.0.1]:5001 [dev] [3.6.4] Starting with 4 generic operation threads and 8 partition operation threads.
Sep 12, 2016 9:13:28 AM com.hazelcast.core.LifecycleService
INFO: [127.0.0.1]:5001 [dev] [3.6.4] Address[127.0.0.1]:5001 is STARTING
Sep 12, 2016 9:13:28 AM com.hazelcast.instance.Node
SEVERE: [127.0.0.1]:5001 [dev] [3.6.4] Error while joining the cluster!
java.lang.NoClassDefFoundError: org/junit/Assert
[...]

PR #8872, opened 2016-09-13
JUnit dependencies in MockJoiner and TestNodeRegistry

rvanheest:
This PR is based on issue #8868. It removes the dependency on JUnit for the `MockJoiner` and `TestNodeRegistry` classes.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/mocknetwork/TestNodeRegistry.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/mocknetwork/TestNodeRegistry.java
@@ -23,6 +23,7 @@ import com.hazelcast.instance.Node;
 import com.hazelcast.instance.NodeContext;
 import com.hazelcast.instance.NodeState;
 import com.hazelcast.nio.Address;
+import com.hazelcast.test.HazelcastTestSupport;
 
 import java.util.Collection;
 import java.util.Collections;
@@ -32,11 +33,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
-public final class TestNodeRegistry {
-
-    private static void verifyInvariant(boolean check, String msg) {
-        if (!check) throw new AssertionError(msg);
-    }
+public final class TestNodeRegistry extends HazelcastTestSupport {
 
     private final ConcurrentMap<Address, Node> nodes = new ConcurrentHashMap<Address, Node>(10);
     private final Collection<Address> joinAddresses;
--- a/hazelcast/src/test/java/com/hazelcast/test/mocknetwork/TestNodeRegistry.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/mocknetwork/TestNodeRegistry.java
@@ -23,7 +23,6 @@ import com.hazelcast.instance.Node;
 import com.hazelcast.instance.NodeContext;
 import com.hazelcast.instance.NodeState;
 import com.hazelcast.nio.Address;
-import com.hazelcast.test.HazelcastTestSupport;
 
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


Donnerbart:
run-lab-run

jerrinot:
Hi @rvanheest,

thanks for this contribution. Relying on JVM-level assertions being enabled is a tricky business. I would rather go with an explicit check and throwing `AssertionError`. WDYT?

rvanheest:
@jerrinot indeed this is a little tricky. However, if you want to through these `AssertionError`s instead, you need to do the if-check there as well, which is already done for you in the `assert`. Being more of a Scala guy myself, I know there are special functions such as [`assert()`](http://www.scala-lang.org/api/current/index.html#scala.Predef$@assert(assertion:Boolean,message:=>Any):Unit) in that language that would do the trick. I could create something similar for these changes instead if you want... but if you think an if-check by hand would suffice, I'd be more than happy to change it in that way!

Let me know!

jerrinot:
@rvanheest: I understand. The problem is the `assert`  keyword is checked only when a JVM is started with the `-ea` flag. I don't think we should rely on it. 

We have had a quite few issues with the `MockJoiner` in the past - they can be pretty nasty and hard-to-find -> We should make sure invariants are always checked, no matter what flags were used to start the JVM.

rvanheest:
absolutely agree on that! Those invariants should be checked at all time, no matter the flags. I'll change the `assert` keywords to 'normal' checks and `AssertionError`s.

rvanheest:
@jerrinot See this new commit (ffaf1c6). I created a static method called `verifyInvariant(boolean check, String msg)` that throws an `AssertionError` with the `msg` if `check` is false. For now I included this as a private function to both files, as you probably don't want to expose this function in the API. However, I realize that this is code duplication, which is probably not desirable. Any suggestions as to where I could best put this function?

rvanheest:
(your response in [the issue](https://github.com/hazelcast/hazelcast/issues/8868#issuecomment-246659026) 😉 )

Missed that `HazelcastTestSupport` class. Nice trick to have a abstract class for this! I added `verifyInvariant` there and did an `extend` in `TestNodeRegistry`. However, as `MockJoiner` already extends another class I wasn't able to do so there. Instead I used static importing there.

hazelcast/src/test/java/com/hazelcast/test/mocknetwork/TestNodeRegistry.java
```diff
@@ -32,9 +33,7 @@
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
-import static org.junit.Assert.assertEquals;
-
-public final class TestNodeRegistry {
+public final class TestNodeRegistry extends HazelcastTestSupport {
```
jerrinot: it's better to use static imports here as well.

jerrinot:
:+1:

hazelcast/src/test/java/com/hazelcast/test/mocknetwork/TestNodeRegistry.java
rvanheest: Ah, I see. I saw this construct in some other places, so I thought I might copy it. But this is better indeed.

jerrinot:
@rvanheest: looks good to me. Have you signed Hazelcast Contributor Agreement? Your name rings same bell, but I don't see you at https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement

rvanheest:
Just sent it.

Donnerbart:
run-lab-run

Donnerbart:
Hmm, `HazelcastTestSupport` has dependencies on `org.junit.Assert` again. Don't we introduce the dependency the changed classes again?

jerrinot:
we don't. it's fine as long as you don't call a method depending on it. (or do stuff like proxying, etc..)

rvanheest:
Nope, it's not fine! I'm looking into it, but I overlooked the whole purpose of this PR for a second. Let me look into it a bit.

Donnerbart:
:+1: (once green)

rvanheest:
OK, figured it out: since we do a static import, `XXX()` is on compile time rewritten to `HazelcastTestSupport.XXX()`. There it still requires all imports of the `HazelcastTestSupport` class, which indeed contain JUnit imports...

I now rewrote it to a private static method in both classes again. I think that is the safest way for now. I considered creating a new class for this, but it didn't feel appropriate to have a class with only one static method in it next to the `HazelcastSupport` class. Would be too risky for future development to put this function back in the other class...

Donnerbart:
run-lab-run

rvanheest:
Hi, I was wondering what's up with this PR. Do I still need to do something before it can be merged?

jerrinot:
@rvanheest: thanks for the kick!

PR #8873, opened 2016-09-13
Fixed NearCacheTest.testMapContainsKey_withNearCache()

Donnerbart:
Fixed wrong mapName being used in test, so no Near Cache was configured.

Found this due to missing code coverage in the `NearCachedMapProxyImpl`.


hazelcast/src/test/java/com/hazelcast/map/nearcache/NearCacheTest.java
```diff
@@ -302,7 +302,7 @@ public void testMapContainsKey_withNearCache() {
         config.getMapConfig(mapName).setNearCacheConfig(newNearCacheConfig().setInvalidateOnChange(true));
         HazelcastInstance instance = createHazelcastInstanceFactory(clusterSize).newInstances(config)[0];
 
-        IMap<String, String> map = instance.getMap("mapName");
+        IMap<String, String> map = instance.getMap(mapName);
```
Donnerbart: The Near Cache was configured for the map `mapName`, which is a String variable with a random content, but the map `"mapName"` with a literal String was used.

pveentjer:
:+1:

PR #8874, opened 2016-09-13
Cardinality estimate service 

tkountis:
Support for estimating cardinalities of unique entities in big data sets with constant memory space. 
This new data structure is a direct implementation of the well known algorithm HyperLogLog by **Philippe Flajolet** et al and HyperLogLog++ by **Stefan Heule** et al. 

http://algo.inria.fr/flajolet/Publications/FlFuGaMe07.pdf
http://static.googleusercontent.com/media/research.google.com/en//pubs/archive/40671.pdf


hazelcast/src/main/java/com/hazelcast/core/HazelcastInstance.java
```diff
@@ -416,6 +416,8 @@
      */
     ICacheManager getCacheManager();
 
+    ICardinalityEstimator getCardinalityEstimator(String name);
```
jerrinot: this should have a really good JavaDoc. This is the place where you need to "sell" the data structure! :)

jerrinot:
do you think a hard-coded precision is ok for the first version?

hazelcast/src/main/java/com/hazelcast/core/ICardinalityEstimator.java
```diff
@@ -0,0 +1,336 @@
...
+ * ICardinalityEstimator is a redundant and highly available distributed data-structure used
+ * for probabilistic cardinality estimation purposes, on unique items, in significantly sized data cultures.
+ *
+ * ICardinalityEstimator is internally based on HyperLogLog data-structure,
+ * and uses P^2 byte registers for storage and computation. (Default P = 14)
+ */
+public interface ICardinalityEstimator extends DistributedObject {
+
+    /**
+     * Aggregates the given hash which can result in a new estimation being available or not.
```
jerrinot: when is the estimation not available? you should explain it.

tkountis:
@jerrinot The hardcoded precision is indeed a problem, but the rational behind the decision was that 14 appears to have a very good balance between space usage and std error on the estimation. On the other hand, exposing it through the API would make the CardinalityEstimator interface bound to HyperLogLog detail which is an unwanted side-effect. We could expose this as configuration detail, but then again I preferred to hide it until we see demand for this from users.

jerrinot:
I understand. I'm fine with having it hard-coded for the first release. Adding new configuration options is easy :)

hazelcast/src/main/java/com/hazelcast/cardinality/hyperloglog/impl/HyperLogLogEncType.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.hazelcast.cardinality.hyperloglog.IHyperLogLog;
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+
+import java.io.IOException;
+
+public enum HyperLogLogEncType {
+
+    SPARSE {
```
Donnerbart: Could you add some JavaDoc here to explain briefly the different types? I guess they differ in precision vs. memory usage (and maybe runtime).

hazelcast/src/main/java/com/hazelcast/cardinality/hyperloglog/impl/HyperLogLogSparseStore.java
```diff
@@ -0,0 +1,190 @@
...
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * [1] http://static.googleusercontent.com/media/research.google.com/en//pubs/archive/40671.pdf
```
Donnerbart: This JavaDoc could be also be improved, e.g. that [1] looks like a citation reference, which makes no sense in this context.

hazelcast/src/main/java/com/hazelcast/cardinality/operations/AggregateAndEstimateOperation.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public int getId() {
+        return CardinalityEstimatorDataSerializerHook.AGGREGATE_AND_ESTIMATE;
+    }
+
+    @Override
+    public void run() throws Exception {
+        CardinalityEstimatorContainer est = getCardinalityEstimatorContainer();
```
Donnerbart: I would name this `container` like I saw it in other operations, to prevent confusion with the field `estimate`.

hazelcast/src/main/java/com/hazelcast/cardinality/operations/BatchAggregateAndEstimateOperation.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public int getId() {
+        return CardinalityEstimatorDataSerializerHook.BATCH_AGGREGATE_AND_ESTIMATE;
+    }
+
+    @Override
+    public void run() throws Exception {
+        CardinalityEstimatorContainer est = getCardinalityEstimatorContainer();
```
Donnerbart: Same here, would name the variable `container`.

hazelcast/src/test/java/com/hazelcast/cardinality/hyperloglog/HyperLogLogAbstractTest.java
```diff
@@ -0,0 +1,78 @@
...
+        assertEquals(true, store.aggregateAll(new long[] { 1L, 2000L, 3000, 40000L }));
+        assertEquals(4L, store.estimate());
+    }
+
+    @Test
+    public void aggregateBigRange() {
+        int sampleStep = 1000;
+        ByteBuffer bb = ByteBuffer.allocate(4);
+
+        for (int i=1; i<=10000000; i++) {
```
Donnerbart: Just nitpicking, but some whitespaces whould make this easier to read :)

jerrinot:
Do we need the `*all()` versions in the `CardinalityEstimator`? the API surface seems rather broad for a first version, for my taste anyway. 

I'd rather see something like `aggregate(Object o)`. We could serialize the object (by using Hazelcast `SerializationService`) and calculate 64 bits hash. Have you tried to consider? I can imagine it may yield bigger error due constant part in the serialized blog, but a good hashing function should deal with this?

jerrinot:
I really like the implementation/design as it is now. so the public API is the last remaining bit to be solved.

tkountis:
@jerrinot I initially designed with Object in mind, and was using the internal serializer to serialize to bytes and then get the hash of them. However, I wanted to always transmit over the wire, hash values, rather than big byte arrays, I believe it wont be as straight forward to do so on the client side, if I have Object params. Will give it a try, thanks for the feedback.

jerrinot:
@tkountis: both client & member sides have SerializationService available. see https://github.com/jerrinot/hazelcast/blob/0deb3aed4803a85e1346f34a9a3e45c244e46bba/hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxy.java#L178-L178

tkountis:
@jerrinot Yep saw that ;)

checkstyle/suppressions.xml
```diff
@@ -56,6 +56,11 @@
     <suppress checks="MethodCount" files="com/hazelcast/cache/impl/CacheStatisticsImpl"/>
     <suppress checks="ClassDataAbstractionCoupling" files="com/hazelcast/cache/impl/DefaultOperationProvider"/>
 
+    <!-- Cardinality -->
+    <suppress checks="JavadocMethod" files="com/hazelcast/cardinality/"/>
```
mmedenjak: I believe that currently there was a non-formal decision to create new suppressions as annotations and not add them in the checkstyle xml. The problem with package-wide suppressions is that they automatically suppress warnings on new classes. David Sondermann is the authority on code analysis so we can contact him and reach a common agreement.

hazelcast/src/main/java/com/hazelcast/cardinality/impl/CardinalityEstimatorContainer.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.hazelcast.cardinality.impl.hyperloglog.HyperLogLog;
+import com.hazelcast.cardinality.impl.hyperloglog.impl.HyperLogLogImpl;
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.DataSerializable;
+
+import java.io.IOException;
+
+public class CardinalityEstimatorContainer implements DataSerializable {
```
mmedenjak: The class seems a bit thin. Other Container classes either implement the semantics of the underlying data structure directly or provide additional behaviour on top of the behaviour of the data structure. On the other hand, the separation of concerns seems as a valid reason to keep this as-is.

checkstyle/suppressions.xml
pveentjer: Correct. We are moving the suppressions to the source and away from the suppression file.

hazelcast-client/src/test/java/com/hazelcast/client/cardinality/ClientCardinalityEstimatorTest.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        assertEquals(4, estimator.estimate());
+        assertEquals(true, estimator.aggregate("Test"));
+        assertEquals(5, estimator.estimate());
+    }
+
+    @Test
+    public void testAsync() throws Exception {
+        ICompletableFuture<Long> f1 = estimator.estimateAsync();
```
pveentjer: I'm not that crazy about these do/verify/do/verify tests. 

Normally best to split them up to individual test methods. If they need to have an assertion. Otherwise one quickly gets dirty test since it isn't obvious if a particular case is tested, so it is being repeated all over.

hazelcast-client/src/test/java/com/hazelcast/client/cardinality/ClientCardinalityEstimatorTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+    @Test
+    public void aggregate() {
+        CardinalityEstimator estimator = client.getCardinalityEstimator("aggregate");
+        assertEquals(true, estimator.aggregate(1L));
+        assertEquals(1L, estimator.estimate());
+    }
+
+    @Test
+    public void aggregateAsync()
```
pveentjer: I just throw Exception to get rid of all the exception noise. But that is just flavor.

hazelcast/src/main/java/com/hazelcast/cardinality/CardinalityEstimator.java
```diff
@@ -0,0 +1,119 @@
...
+import com.hazelcast.core.ICompletableFuture;
+
+/**
+ * CardinalityEstimator is a redundant and highly available distributed data-structure used
+ * for probabilistic cardinality estimation purposes, on unique items, in significantly sized data cultures.
+ *
+ * CardinalityEstimator is internally based on HyperLogLog data-structure,
+ * and uses P^2 byte registers for storage and computation. (Default P = 14)
+ */
+public interface CardinalityEstimator extends DistributedObject {
```
pveentjer: Probably good to slap @Beta on top.

hazelcast/src/main/java/com/hazelcast/cardinality/CardinalityEstimator.java
```diff
@@ -0,0 +1,119 @@
...
+     *              // handle failure
+     *          }
+     *     });
+     * </pre>
+     * </p>
+     * @param obj the serializable object to aggregate in the estimate.
+     * @return {@link ICompletableFuture} bearing the response, True, when a new estimate can be computed.
+     * @since 3.8
+     */
+    ICompletableFuture<Boolean> aggregateAsync(Object obj);
```
pveentjer: Can it be called with null?

checkstyle/suppressions.xml
Donnerbart: Those package wide Javadoc suppressions are fine, but we shouldn't have them for public packages. In public packages we should solve them. All implementation details, where we are allowed to omit some JavaDoc should be in packages which contain `internal` or `impl`. For those we already have a project wide matching suppression.

So actually this one should not be needed and get fixed.

mmedenjak: I agree, some minor suppressions for internal classes should be allowed to be added package-wide but I would say that other code smell warnings for both private and public classes should be defined with annotations.

hazelcast/src/main/java/com/hazelcast/cardinality/impl/CardinalityEstimatorContainer.java
tkountis: Thanks @mmedenjak. This is indeed a thin wrapper around the actual implementation, the idea is that there are more than one cardinality estimators out there, so I wanted to keep this generic enough that in the future we could potentially have configuration to choose the implementation. Therefore this separation helps.

hazelcast/src/main/java/com/hazelcast/cardinality/CardinalityEstimator.java
```diff
@@ -0,0 +1,135 @@
...
+public interface CardinalityEstimator extends DistributedObject {
+
+    /**
+     * Consider the 64bit hash value, produced by the serialized form of the given object,
+     * in the cardinality estimation. The implementation is free to choose whether this hash
+     * will be used towards the estimation, therefore, this operation may or may not affect
+     * the current estimation.
+     *
+     * Input object, unless a boxed primitive, must have a
+     * {@link com.hazelcast.nio.serialization.StreamSerializer} implementation, registered in the Hazelcast
```
jerrinot: This is confusing/misleading. 
The object also can just implement `j.i.Serializable`, it can implement `DataSerializable`, etc. The only condition is this: "Hazelcast has to know how to serialize it". The exact serialization strategy is irrelevant. I would completely remove this as this is true for (almost) all other Hazelcast calls. Even the most basic - `map.put()`, etc.

hazelcast/src/main/java/com/hazelcast/cardinality/CardinalityEstimator.java
```diff
@@ -0,0 +1,135 @@
...
+ * CardinalityEstimator is internally based on a HyperLogLog++ data-structure,
+ * and uses P^2 byte registers for storage and computation. (Default P = 14)
+ */
+@Beta
+public interface CardinalityEstimator extends DistributedObject {
+
+    /**
+     * Consider the 64bit hash value, produced by the serialized form of the given object,
+     * in the cardinality estimation. The implementation is free to choose whether this hash
+     * will be used towards the estimation, therefore, this operation may or may not affect
```
jerrinot: I am not sure if this detail is so important to have it in the first paragraph. 

This is the most important method of the whole interface. It should have a clear API which states: "Use this to add a new object into your estimation". Something along these lines. It should be the first sentence - for super-quick discoverability. Everything else is just a detail/clarification.

tkountis: Followed your suggestions for the Javadoc. 👍

hazelcast/src/main/java/com/hazelcast/cardinality/CardinalityEstimator.java
```diff
@@ -0,0 +1,127 @@
...
+     *     // do something else, then read the result
+     *     Boolean result = future.get(); // this method will block until the result is available
+     * </pre>
+     * </p>
+     * <p>
+     * <pre>
+     *     ICompletableFuture&lt;Void&gt; future = estimator.addAsync();
+     *     future.andThen(new ExecutionCallback&lt;Void&gt;() {
+     *          void onResponse(Void response) {
+     *              // do something with the result
```
jerrinot: is there any result at all? :) it's more like "at this point you know the object was added to the set"

jerrinot:
I like the new API. :+1: from me. 
This assume the Tom's comment about using `IdentifiedDataSerializable` will be addressed.

tkountis:
Regarding Tom's suggestion, I confirmed that all 'root' serializable are indeed IdentifiedDataSerializable. The remaining are serialized/deserialized through these root operations, so they don't require ID info.

hazelcast/src/main/java/com/hazelcast/cardinality/CardinalityEstimator.java
tkountis: duh

PR #8887, opened 2016-09-15
Parameterized ClientNearCacheTest to deduplicate test code

Donnerbart:
Thanks @vbekiaris for the hint to parameterize this test class.


hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCacheTestSupport.java
```diff
@@ -468,8 +468,11 @@ public void run() throws Exception {
         }
     }
 
-    protected void doTestGetAllReturnsFromNearCache() {
-        NearCacheConfig nearCacheConfig = createNearCacheConfig(InMemoryFormat.OBJECT);
+    protected void doTestGetAllReturnsFromNearCache(InMemoryFormat inMemoryFormat) {
```
Donnerbart: I parameterized this test as well, so it isn't executed twice with the same configuration (and burning CPU time).

PR #8889, opened 2016-09-16
Serialization enhancements in the IMap

tombujok:


hazelcast/src/main/java/com/hazelcast/map/impl/operation/ClearBackupOperation.java
```diff
@@ -16,11 +16,13 @@
 
 package com.hazelcast.map.impl.operation;
 
-import com.hazelcast.nio.serialization.DataSerializable;
+import com.hazelcast.map.impl.MapDataSerializerHook;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
 import com.hazelcast.spi.BackupOperation;
 import com.hazelcast.spi.impl.MutatingOperation;
 
-public class ClearBackupOperation extends MapOperation implements BackupOperation, MutatingOperation, DataSerializable {
```
pveentjer: Isn't it easier to impose it on MapOperation? This way you know you are going to catch all map operations.

pveentjer:
It looks good. Please have a look at the comment regarding moving the logic into the MapOperation.

Can you also squash the commits?

pveentjer:
If IdentifiedDataSerializable is not moved into MapOperation that is ok with me. However for the rolling upgrade feature it would be smart to convert every DataSerializable implementation we have (not only operations) to IdentifiedDataSerializable since the latter reduces coupling with the implementation since the class name is not part of the byte-stream.

tombujok:
I'll do it on the MapOperation level

tombujok:
@pveentjer all operations are IdentifiedDataSerializable now. Will be squashed at merge. Could you have a look?

pveentjer:
:+1:

pveentjer:
Since we have 2 :+1: , I'm merging this one

PR #8891, opened 2016-09-16
Diagnostics framework uses 24 hours format in timestamps

jerrinot:
Otherwise it's hard to interpret the logs, especially around mid-day.


Donnerbart:
:+1: (once green)

Issue #8890, opened 2016-09-16
Change serialisation in classes using java.io.Serializable


tombujok:
HazelcastExpiryPolicy 
CacheClearResponse
CacheEventListenerAdaptor
CacheEntryListener
CacheStatisticsMXBeanImpl 
EvictionConfig 
MaxSizeConfig
NearCacheConfig
WanReplicationRef
SimulateLoadTask 
LazyMapEntry (?)
MulticastMemberInfo

tombujok:
These one were identified implementing both -> java.io.Serializable and the HZ alternative (DataSerializable or IdentifiedDataSerializable):
HazelcastExpiryPolicy 
CacheClearResponse 
EvictionConfig 
MaxSizeConfig
NearCacheConfig
WanReplicationRef

CacheStatisticsMXBeanImpl ->  JMX bean should probably  java.io.Serializable. Anyhow this object won't be versioned among the nodes.

FIXED:
SimulateLoadTask (DONE)
MulticastMemberInfo (DONE) 
CacheEventListenerAdaptor (DONE to both)
CacheEntryListener (DONE to both)

tombujok:
[DONE] Marked with BinaryInterface:
-----------------------------------------------
HazelcastExpiryPolicy
EvictionConfig
MaxSizeConfig
NearCacheConfig
WanReplicationRef

[DONE] Implementing both Serializable and IDS:
-----------------------------------------------
CacheClearResponse (isn't it a BinaryInterface class?)
LazyMapEntry

TODO:
-----------------------------------------------
CacheEventListenerAdaptor
CacheEntryListener
CacheStatisticsMXBeanImpl
SimulateLoadTask
MulticastMemberInfo

@vbekiaris we still need a test to guard classes implementing Serializable only. Let's discuss on Tuesday. Thx!

PR #8892, opened 2016-09-16
Making classes use HZ serialization instead of java.io.Serialization

tombujok:
Solves some of the #8890


pveentjer:
It looks good, but I don't know if this can cause any incompatibilities.

PR #8899, opened 2016-09-19
PartitionIteratingOperation & Backup more detail in diagnostics

### LINKED ISSUE #8855: PartitionIteratingInfo more detail diagnostics
Currently in the invocation pluggin you see something like:

```
com.hazelcast.spi.impl.operationservice.impl.operations.PartitionIteratingOperation samples=11867
```

Which is not very helpful since it isn't clear what type of operation is executed? Size? Something else?

The same goes for backup operation.

pveentjer:
If a PartitionIteratingOperation or a Backup is found in the invocations or
in the pending connections, instead of just mentioning the outer classname,
the diagnostics tries to look inside so one can see e.g.
PartitionIteratingOperation(MapSizeFactory) as name.

This gives better insight of the type of operations being encountered.

fix #8855


pveentjer:
Merging this one since it is just a debug/diagnostics logging change.

PR #8901, opened 2016-09-19
Counter for failed backup operations

jerrinot:


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
```diff
@@ -89,6 +90,9 @@
     private final Address thisAddress;
     private final boolean staleReadOnMigrationEnabled;
 
+    @Probe(name = "failedBackups")
+    private final Counter failedBackupsCounter = newSwCounter();
```
pveentjer: Do you want to track this per partition operation runner?

jerrinot: I am indecisive. it's makes it harder to interpret it, but it gives finer-grained view. wdyt?

pveentjer: Personally I don't care of this information on a partition level, however if you want to have it, go nuts.

However the probes on non partition operationrunnerimpl should not be created. Because they will always be null and therefor cause noise.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationexecutor/OperationRunnerFactory.java
```diff
@@ -28,14 +30,14 @@
...
-    OperationRunner createPartitionRunner(int partitionId);
+    OperationRunner createPartitionRunner(int partitionId, Counter failedBackupsCounter);
 
     /**
      * Creates an OperationRunner to execute generic Operations.
      *
      * @return the created OperationRunner
      */
-    OperationRunner createGenericRunner();
+    OperationRunner createGenericRunner(Counter failedBackupsCounter);
```
pveentjer: No counter needed; generic and ad hoc will never have backups.

jerrinot: silly me!

pveentjer:
Since it is not a risky change, I'm merging this one.

PR #8902, opened 2016-09-19
ReplicatedMap and DurableExecutor IdentifiedSerializable

tombujok:


tombujok:
Will be squashed on merge

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/PutAllOperation.java
```diff
@@ -36,7 +34,7 @@
 /**
  * Puts a set of records to the replicated map.
  */
-public class PutAllOperation extends Operation implements IdentifiedDataSerializable {
+public class PutAllOperation extends AbstractSerializableOperation {
```
pveentjer: Should this operation not extend from AbstractReplicatedMapOperation?

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/ReplicatedMapDataSerializerHook.java
```diff
@@ -47,8 +47,19 @@
     public static final int OP_MERGE = 9;
     public static final int VERSION_RESPONSE_PAIR = 10;
     public static final int OP_GET = 11;
+    public static final int OP_CHECK_REPLICA_VERSION = 12;
```
pveentjer: I would fix these constant names. Most other data-structures don't use 'OP_' prefix.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/VersionResponsePair.java
```diff
@@ -28,6 +28,7 @@
  * Contains response and partition version for update operations on replicated map.
  */
 public class VersionResponsePair implements DataSerializable, IdentifiedDataSerializable {
```
pveentjer: This interface only needs to implement IdentifiedDataserializable, since that extends DataSerializable.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/PutAllOperation.java
tombujok: I wanted to have it like this, but the AbstractReplicatedMapOperation has some fields that are not used everywhere, so actually AbstractReplicatedMapOperation should extend AbstractSerializableOperation.

tombujok: Fixed

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/ReplicatedMapDataSerializerHook.java
tombujok: Thx. Fixed

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/operation/VersionResponsePair.java
tombujok: Thx. Fixed

pveentjer:
:+1:

PR #8905, opened 2016-09-19
Lock module IdentifiedDataSerializable

pveentjer:


hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockDataSerializerHook.java
```diff
@@ -42,20 +42,23 @@
...
+    public static final int GET_LOCK_COUNT = 7;
+    public static final int GET_REMAINING_LEASETIME = 8;
+    public static final int IS_LOCKED = 9;
+    public static final int LOCK_BACKUP = 10;
+    public static final int LOCK = 11;
+    public static final int LOCK_REPLICATION = 12;
+    public static final int SIGNAL_BACKUP = 13;
+    public static final int SIGNAL = 14;
+    public static final int UNLOCK_BACKUP = 15;
+    public static final int UNLOCK = 16;
```
tombujok: Hmm, I guess we should not change the IDs? 
I mean in theory there won't be 3.7 talking to 3.8 but still it's kind of shaky... WDYT?

pveentjer: I think we can change the id's when needed. It doesn't make the system any more or less reliable since the container is now causing an incompatibility (it went from DS to IDS).

tombujok: Fair enough.

PR #8906, opened 2016-09-19
IMap.tryRemove javadoc fix

### LINKED ISSUE #8884: IMap:tryRemove(K key, long timeout, TimeUnit timeunit) Javadoc is not correct
I was looking for a "remove" method on IMap that doesn't return the previous value (in order to avoid deserialization). I found "tryRemove" and here is his Javadoc : 

http://docs.hazelcast.org/docs/3.7/javadoc/com/hazelcast/core/IMap.html#tryRemove(K,%20long,%20java.util.concurrent.TimeUnit)

I think that the "Warning-2:" isn't correct because this method does return only a boolean. It's a little bit confusing in here.

Am I correct ?

Have a good day :smile: 

Edit : by the way, the correct method for what I expected was "delete"

pveentjer:
fix #8884


pveentjer:
Merging this one since it is just removal of incorrect javadoc.

PR #8907, opened 2016-09-19
Gets expiration paramaters from config object

ahmetmircik:


hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
```diff
@@ -234,15 +242,13 @@ private boolean notHaveAnyExpirableRecord(PartitionContainer partitionContainer)
     }
 
     private Operation createExpirationOperation(int expirationPercentage, int partitionId) {
-        ClearExpiredOperation clearExpiredOperation = new ClearExpiredOperation(expirationPercentage);
-        NodeEngine nodeEngine = mapServiceContext.getNodeEngine();
-        clearExpiredOperation
-                .setNodeEngine(nodeEngine)
+        Operation operation = new ClearExpiredOperation(expirationPercentage);
```
pveentjer: Why not use:
```
return new ClearOperation()
   .setblabla()
   .setFoofoo();
```

ahmetmircik: done.

PR #8916, opened 2016-09-20
JCache backup operations do not propagate CacheNotExistsException

jerrinot:
Reasoning:
1. I am a partition owner. You are a backup replica. I execute PUT and send you BACKUP
2. before you process the BACKUP someone else will tell you to destroy the cache
3. when you start processing the BACKUP then the cache no longer exists -> exception

New cache creation is blocking a caller until the cache is created on all members
->
if backup operation does not find the cache then it's because it was already destroyed.
not because it wasn't created yet.
->
Skipping the backup when a cache is not found will not lose any item.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheRemoveBackupOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheRemoveBackupOperation.java
@@ -29,7 +29,7 @@ import java.io.IOException;
  * Backup operation used by remove operations.
  */
 public class CacheRemoveBackupOperation
-        extends AbstractCacheOperation
+        extends AbstractBackupCacheOperation
         implements BackupOperation, MutatingOperation {
 
     private boolean wanOriginated;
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CacheRemoveBackupOperation.java
```diff
@@ -47,17 +47,13 @@ public CacheRemoveBackupOperation(String name, Data key, boolean wanOriginated)
...
     @Override
-    public void run()
+    public void runInternal()
             throws Exception {
-        if (cache != null) {
-            response = cache.removeRecord(key);
-        } else {
-            response = Boolean.FALSE;
-        }
+        cache.removeRecord(key);
```
jerrinot: there was a copy&paste error I fixed - response is irrelevant for backup operations.

Donnerbart:
run-lab-run

Donnerbart:
I still see a lot of duplicated code. Can't we make more operations use the new abstract Operation and put this piece of code in there?
```
        ICacheService service = getService();
        try {
             cache = service.getOrCreateRecordStore(name, getPartitionId());
         } catch (CacheNotExistsException e) {
             getLogger().finest("Error while getting a cache", e);
         }
```

jerrinot:
@Donnerbart: good one, I'll have a look

PR #8920, opened 2016-09-20
Cleanup of DefaultNearCache classes

Donnerbart:
* Removed `DefaultNearCache::init()`, which just prevented clean `final` fields and was not needed for the EE extension, since we already initialize new fields in the overloaded EE constructors, so we don't have to duplicate this in `init()`
* Small cleanup of `DefaultNearCache$ExpirationTask` creation
* Moved superfluous constructor of `NearCacheContext`, which was just used in EE (and did nothing but set a parameter to `null`)

Requires an EE change.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
@@ -39,58 +39,26 @@ public class DefaultNearCache<K, V> implements NearCache<K, V> {
     protected final NearCacheConfig nearCacheConfig;
     protected final NearCacheContext nearCacheContext;
     protected final SerializationService serializationService;
-    protected NearCacheRecordStore<K, V> nearCacheRecordStore;
-    protected ScheduledFuture expirationTaskFuture;
+    protected final NearCacheRecordStore<K, V> nearCacheRecordStore;
+    protected final ScheduledFuture expirationTaskFuture;
 
-    public DefaultNearCache(String name, NearCacheConfig nearCacheConfig,
-                            NearCacheContext nearCacheContext) {
+    public DefaultNearCache(String name, NearCacheConfig nearCacheConfig, NearCacheContext nearCacheContext) {
         this.name = name;
         this.nearCacheConfig = nearCacheConfig;
         this.nearCacheContext = nearCacheContext;
         this.serializationService = nearCacheContext.getSerializationService();
         this.nearCacheRecordStore = createNearCacheRecordStore(nearCacheConfig, nearCacheContext);
-        init();
+        this.expirationTaskFuture = createAndScheduleExpirationTask();
     }
 
-    public DefaultNearCache(String name, NearCacheConfig nearCacheConfig,
-                            NearCacheContext nearCacheContext,
+    public DefaultNearCache(String name, NearCacheConfig nearCacheConfig, NearCacheContext nearCacheContext,
                             NearCacheRecordStore<K, V> nearCacheRecordStore) {
         this.name = name;
         this.nearCacheConfig = nearCacheConfig;
         this.nearCacheContext = nearCacheContext;
         this.serializationService = nearCacheContext.getSerializationService();
         this.nearCacheRecordStore = nearCacheRecordStore;
-        init();
-    }
-
-    protected void init() {
-        startExpirationTask();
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
```diff
@@ -39,58 +39,26 @@
...
     protected final NearCacheContext nearCacheContext;
     protected final SerializationService serializationService;
-    protected NearCacheRecordStore<K, V> nearCacheRecordStore;
-    protected ScheduledFuture expirationTaskFuture;
+    protected final NearCacheRecordStore<K, V> nearCacheRecordStore;
+    protected final ScheduledFuture expirationTaskFuture;
 
-    public DefaultNearCache(String name, NearCacheConfig nearCacheConfig,
-                            NearCacheContext nearCacheContext) {
+    public DefaultNearCache(String name, NearCacheConfig nearCacheConfig, NearCacheContext nearCacheContext) {
```
pveentjer: Can't you route all constructor traffic through a single master constructor?

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
```diff
@@ -109,6 +77,14 @@ protected ScheduledFuture scheduleExpirationTask(NearCacheExecutor nearCacheExec
         }
     }
 
+    private ScheduledFuture createExpirationTask() {
```
pveentjer: I would call this 'initExpirationTask' since it does more than just creating; it also takes care of scheduling it on an executor. But... not that important.

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
```diff
@@ -175,9 +151,9 @@ public int size() {
         return nearCacheRecordStore.size();
     }
 
-    protected class ExpirationTask implements Runnable {
+    private class ExpirationTask implements Runnable {
 
-        protected AtomicBoolean expirationInProgress = new AtomicBoolean(false);
+        private AtomicBoolean expirationInProgress = new AtomicBoolean(false);
```
pveentjer: Can it be made final?

pveentjer:
Minor comments. Nothing preventing from merging. :+1:

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
Donnerbart: If I try to replace the first constructor with `this(name, nearCacheConfig, nearCacheContext, nearCacheContext.getSerializationService(), createNearCacheRecordStore(nearCacheConfig, nearCacheContext), createExpirationTask());` I get the compile error `cannot reference this before supertype constructor has been called`, so I don't know how to achieve this.

Beside the size of the methods which don't make them good candidates for inlineing, we need to override them in EE. So maybe we could go with `newInstance` methods to achieve this, but I don't know if this is a better approach than having two small constructors.

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
Donnerbart: Done

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
Donnerbart: I renamed it to `createAndScheduleExpirationTask`, so it matches the `create...` schema of the other methods.

pveentjer: :+1:

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
pveentjer: Ok. No problem. Sometimes there is no way out

PR #8924, opened 2016-09-20
Initial work on removing ClientDelegatingFuture

pveentjer:
This class is heavily used on the client and requires quite some work. This is the first of a series of PR that eventually completely removed the ClientDelegatingFuture and lets the ClientInvocationFuture take care of the work.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/PartitionSpecificClientProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/PartitionSpecificClientProxy.java
@@ -18,12 +18,10 @@ package com.hazelcast.client.proxy;
 
 import com.hazelcast.client.impl.ClientMessageDecoder;
 import com.hazelcast.client.impl.protocol.ClientMessage;
+import com.hazelcast.client.spi.ClientInvocationService;
 import com.hazelcast.client.spi.ClientProxy;
-import com.hazelcast.client.spi.impl.ClientInvocation;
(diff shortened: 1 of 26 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/PartitionSpecificClientProxy.java
```diff
@@ -43,21 +41,28 @@ protected void onInitialize() {
         partitionId = getContext().getPartitionService().getPartitionId(partitionKey);
     }
 
-    protected ClientMessage invokeOnPartition(ClientMessage req) {
-        return invokeOnPartition(req, partitionId);
+    protected ClientMessage invokeOnPartition(ClientMessage request) {
+        return invokeOnPartition(request, partitionId);
+    }
+
+    protected <T> T invokeOnPartition(ClientMessage request, ClientMessageDecoder decoder) {
```
LoneRifle: I've always been puzzled when reading Hazelcast's codebase by duplicate logic in both sync and async implementations of the same method. Is the reason why this is the case so that we avoid an extra method call incurred if sync delegates to async, assuming the implementation is simple enough?

pveentjer: It requires further cleaning up. The method call isn't the issue; but the exception handling could be a hornets nest.

My goal is to get rid of client side duplication

Issue #8923, opened 2016-09-20
DelegatingFuture broken


pveentjer:
We need to get rid of this class for the same reasons this ClientDelegatingFuture needs to be removed.

See
https://github.com/hazelcast/hazelcast/issues/8127

PR #8925, opened 2016-09-20
Clean up DelegatingFuture

pveentjer:
gets rid of the locking and most of the complex logic.

fixes https://github.com/hazelcast/hazelcast/issues/8923

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/executor/DelegatingFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/executor/DelegatingFuture.java
@@ -20,103 +20,109 @@ import com.hazelcast.core.ExecutionCallback;
 import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.internal.serialization.InternalSerializationService;
 import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.spi.InternalCompletableFuture;
 import com.hazelcast.spi.serialization.SerializationService;
-import com.hazelcast.util.ExceptionUtil;
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
-public class DelegatingFuture<V> implements ICompletableFuture<V> {
+/**
+ * A {@link InternalCompletableFuture} implementation that delegates the real logic to an underlying
+ * {@link InternalCompletableFuture} and decorates it with additional behavior:
+ * <ol>
+ * <li>change the returned value by setting the result</li>
+ * <li>always deserializing the content</li>
+ * <li>caching the deserialized content so that a deserialization only happens once. This should be used with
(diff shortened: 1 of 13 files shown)


hazelcast/src/test/java/com/hazelcast/util/executor/CancellableDelegatingFutureTest.java
```diff
@@ -35,17 +34,15 @@ public void testInnerFutureThrowsCancellationExceptionWhenOuterFutureIsCancelled
             try {
                 future.getFuture().get();
                 fail();
-            } catch (ExecutionException expected) {
-                assertTrue(expected.getCause() instanceof CancellationException);
+            } catch (CancellationException expected) {
```
vbekiaris: A minor thing, one could also use the `ExpectedException` rule style here.

hazelcast/src/main/java/com/hazelcast/util/executor/DelegatingFuture.java
```diff
@@ -20,103 +20,109 @@
...
 
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+
+/**
+ * A {@link InternalCompletableFuture} implementation that implement the real logic to an underlying
```
vbekiaris: "delegates" instead of "implement"?

hazelcast/src/test/java/com/hazelcast/util/executor/CancellableDelegatingFutureTest.java
pveentjer: Can you point to an example? I never did something with the rules.

pveentjer: I'm failing to see how this improves readability.

pveentjer: Ok.. it is a bit shorter:

```
  @Test
    public void testInnerFutureThrowsCancellationExceptionWhenOuterFutureIsCancelled() throws Exception {
        final TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(1);
        final HazelcastInstance instance = factory.newHazelcastInstance();
        IExecutorService executorService = instance.getExecutorService(randomString());
        final CompletesOnInterruptionCallable callable = new CompletesOnInterruptionCallable();
        final DelegatingFuture<Boolean> future = (DelegatingFuture<Boolean>) executorService.submit(callable);

        if (future.cancel(true)) {
            expected.expect(CancellationException.class);
            future.getFuture().get();
        }
    }
```

Donnerbart: You can e.g. match for the root cause if you have nested exceptions (just search for the usage of  `RootCauseMatcher`). Or you can match on exception messages (which is not nice, but if it's necessary you can do it).

vbekiaris: Right, plus the check is localized with the invocation that should throw the exception instead of a blanket `@Test(expected = Exception.class)` (this is not the case here, but in terms of general usefulness).

PR #8927, opened 2016-09-21
Fix expected memory costs in tests

vbekiaris:
The expected memory costs in test code were previously calculated assuming compressed oops is enabled (or a 32-bit JVM). These have now been updated to take into account the JVM reference cost in bytes as estimated by `JVMUtil`.

Fixes #8908 and #8909

PR #8928, opened 2016-09-21
[BACKPORT] Fix expected memory costs in tests

vbekiaris:
Backport of #8927 , updates expected memory costs taking into account the reference cost in bytes in the current JVM.


hazelcast/src/test/java/com/hazelcast/map/SizeEstimatorTest.java
```diff
@@ -48,7 +49,8 @@ public void smoke() throws InterruptedException {
 
     @Test
     public void testSinglePut() throws InterruptedException {
-        long expectedPerEntryHeapCost = 156L;
```
mmedenjak: Backport the explanations for the magic numbers, otherwise it's ok

PR #8929, opened 2016-09-21
Renamed WaitNotifyService to OperationParker

pveentjer:
It is renamed to OperationParker

The method await was very misleading since it doesn't 'await'; it 'park' the
operation and then continues whatever it was doing.

The method 'notify' is confusing since we already have an Object.notify. So
this has been made symetrical with the park ---> unpark.


pveentjer:
There are no EE deps.

PR #8930, opened 2016-09-21
Replaced NearCacheImpl with DefaultNearCache

Donnerbart:
* Replaced NearCacheImpl with DefaultCache
* Replaced NearCacheSizeEstimator with memory costs from DefaultNearCache
* Added EvictionPolicy.RANDOM and EvictionPolicy.NONE


hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/MapNearCacheInvalidationFromClientTest.java
```diff
@@ -192,7 +192,7 @@ public void run() {
         assertTrueEventually(new AssertTask() {
```
pveentjer: If you get a lot of such methods, you could make a method 'assertNullEventually' or assertEqualsEventually.

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/AbstractNearCacheRecordStore.java
```diff
@@ -231,7 +234,8 @@ protected void onExpire(K key, R record) {
     }
 
     protected boolean isEvictionEnabled() {
-        return evictionStrategy != null && evictionPolicyEvaluator != null;
+        return evictionStrategy != null && evictionPolicyEvaluator != null
```
pveentjer: To improve the readability of such expressions I often do something like 

```
return evictionStrategy != null
     && evictionPolicyEvaluator != null
     && evictionPolicyType != null
     && !evictionPolicyType.equals(EvictionPolicyType.NONE)
```
Then it becomes easier to see the structure

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyType.java
```diff
@@ -29,8 +29,17 @@
     /**
```
pveentjer: Please add a few lines of explanation for the feeble minded people like me what the purpose is. I guess random picks random items and none doesn't evict at all.

hazelcast/src/test/java/com/hazelcast/cache/nearcache/NearCacheRecordStoreTest.java
```diff
@@ -196,13 +196,13 @@ public void evictionTriggeredAndHandledSuccessfullyWithEntryCountMaxSizePolicyAn
     }
 
     @Test
-    public void evictionTriggeredAndHandledSuccessfullyWithEntryCountMaxSizePolicyAndDefaultEvictionPolicyOnNearCacheObjectRecordStore() {
-        doEvictionWithEntryCountMaxSizePolicy(InMemoryFormat.OBJECT, null);
+    public void evictionTriggeredAndHandledSuccessfullyWithEntryCountMaxSizePolicyAndRandomEvictionPolicyNearCacheObjectRecordStore() {
```
pveentjer: If methods become so long I would pull out a seperate test: 

NearCacheRecordStore_EvictionTest

Just a suggestion..

pveentjer:
It looks good, but my knowledge of near cache is not sufficient to put a :+1: on it.

hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/MapNearCacheInvalidationFromClientTest.java
Donnerbart: I was just fixing test failures due to the replacement. The whole Near Cache tests are in a quite bad state and will need more time to unify and polish. I spoke to Jaromir about this and we'll have a look that we do this after the upcoming Near Cache performance optimizations.

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/AbstractNearCacheRecordStore.java
Donnerbart: Done

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyType.java
Donnerbart: Done

hazelcast/src/test/java/com/hazelcast/cache/nearcache/NearCacheRecordStoreTest.java
Donnerbart: As I said before, just fixing bugs here. TLC for Near Cache tests is another huge task which we'll plan for this sprint, if there is time left.

pveentjer:
@Donnerbart there is a conflict. Can you rebase?

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyType.java
```diff
@@ -29,8 +29,17 @@
...
-    // TODO Maybe another "CUSTOM" type for user defined eviction policies
+    /**
+     * Picks a random entry
+     */
+    RANDOM,
+
+    /**
+     * Doesn't evict entries (will not add new entries to the Near Cache when it's full)
+     */
+    NONE,
```
Donnerbart: I think I managed to add the old behavior of `NONE` eviction which was:
```
public void put(Data key, Object value) {
    fireCacheExpiration();
    if (evictionPolicy == NONE && cache.size() >= maxSize) {
        return;
    }
```
So if there is no expiration, no new entries should be added to the Near Cache, since no eviction will be triggered.

hazelcast/src/main/java/com/hazelcast/internal/eviction/impl/comparator/RANDOMEvictionPolicyComparator.java
```diff
@@ -0,0 +1,36 @@
...
+ * {@link com.hazelcast.config.EvictionPolicy#RANDOM} policy based {@link EvictionPolicyComparator}.
+ */
+@SuppressFBWarnings(
+        value = "SE_COMPARATOR_SHOULD_BE_SERIALIZABLE",
+        justification = "No need to serializable since its instance is not serialized")
+public class RANDOMEvictionPolicyComparator extends EvictionPolicyComparator {
+
+    @Override
+    public int compare(EvictableEntryView e1, EvictableEntryView e2) {
+        return QuickMath.compareIntegers(e1.getKey().hashCode(), e2.getKey().hashCode());
```
Donnerbart: I took that code from the `NearCacheRecord`:
```
    private static final Comparator<NearCacheRecord> DEFAULT_COMPARATOR = new Comparator<NearCacheRecord>() {
        public int compare(NearCacheRecord o1, NearCacheRecord o2) {
            return QuickMath.compareIntegers(o1.key.hashCode(), o2.key.hashCode());
        }
    };
```
But you're right, it shouldn't make a difference and changing the behavior of "random" should not have an effect on a user.

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionPolicyEvaluatorProvider.java
```diff
@@ -38,6 +39,10 @@ private static EvictionPolicyComparator createEvictionPolicyComparator(EvictionP
                 return new LRUEvictionPolicyComparator();
             case LFU:
                 return new LFUEvictionPolicyComparator();
+            case RANDOM:
+                return new RANDOMEvictionPolicyComparator();
```
Donnerbart: Done

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/NearCacheProvider.java
```diff
@@ -79,46 +70,47 @@ private boolean isBatchingEnabled() {
...
-        return getOrPutIfAbsent(nearCacheMap, mapName, nearCacheConstructor);
+    public NearCache getOrCreateNearCache(String mapName) {
+        NearCacheConfig nearCacheConfig = nodeEngine.getConfig().getMapConfig(mapName).getNearCacheConfig();
+        EvictionConfigAccessor.initDefaultMaxSize(nearCacheConfig.getEvictionConfig());
+
+        NearCacheContext nearCacheContext = new NearCacheContext(
+                nearCacheManager,
+                nodeEngine.getSerializationService(),
+                new MemberNearCacheExecutor(nodeEngine.getExecutionService()),
+                null);
```
Donnerbart: I copied this from the `AbstractClientInternalCacheProxy`. Actually the only usage of that parameter seems to be in Enterprise.

hazelcast/src/test/java/com/hazelcast/cache/nearcache/NearCacheRecordStoreTest.java
vbekiaris: 👍  this test class would benefit from parametrization & renaming.

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/store/AbstractNearCacheRecordStore.java
```diff
@@ -270,6 +275,10 @@ public V get(K key) {
     public void put(K key, V value) {
         checkAvailable();
 
+        if (!isEvictionEnabled() && maxSizeChecker != null && maxSizeChecker.isReachedToMaxSize()) {
+            return;
```
Donnerbart: I don't think it is. The old implementation tries to expire values, but if the eviction policy is `NONE` and the Near Cache is full, it does nothing an just returns:
```
public void put(Data key, Object value) {
    fireCacheExpiration();
    if (evictionPolicy == NONE && cache.size() >= maxSize) {
        return;
    }
    if (evictionPolicy != NONE && cache.size() >= maxSize) {
        fireCacheEviction();
    }
    ...
}
```
The expiration is done regularly by the `ExpirationTask` in the `DefaultNearCache`, so we don't need to trigger it before checking the eviction policy.

Donnerbart: I just saw that I can simplify the check to:
```
        if (!isEvictionEnabled() && evictionChecker.isEvictionRequired()) {
            return;
        }
```
Which will call the same code from within the `MaxSizeEvictionChecker` class. So it's a little less duplication.

Donnerbart: We just discussed this. Of course, with an _existing_ key we'll produce a stale value here. Looks like a bug in the other implementations though, which I copied.

That's a very good finding!

I'll write a test for this and fix this.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

vbekiaris:
👍

PR #8935, opened 2016-09-21
[BACKPORT] Destroy non-alive client connections

jerrinot:
forward of #8832 from the 3.7.1 release branch into maintanance-3.x

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -591,7 +591,9 @@ public class ClientConnectionManagerImpl implements ClientConnectionManager {
             fireConnectionAddedEvent(connection);
         } else {
             if (logger.isFinestEnabled()) {
-                logger.finest("Authentication succeeded for " + connection + " and replaced " + oldConnection);
+                logger.finest("Authentication succeeded for " + connection + " and replaced " + oldConnection
+                    + " Closing the old connection.");
+                destroyConnection(oldConnection, "Replaced by newer connection", null);
             }
         }
         assert oldConnection == null || connection.equals(oldConnection);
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -216,7 +216,15 @@ public ClientConnection getConnection(Address target) {
         if (target == null) {
             return null;
         }
-        return connections.get(target);
+        ClientConnection connection = connections.get(target);
```
pveentjer: I guess this is a temporary hack? Because doing cleanup as part of doing a get sounds like it :)

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -561,14 +584,26 @@ public void run() {
...
+                destroyConnection(oldConnection, "Replaced by newer connection", null);
+            }
         }
         assert oldConnection == null || connection.equals(oldConnection);
         connectionsInProgress.remove(target);
     }
 
-    private void failed(Address target, ClientConnection connection, Throwable cause) {
-        logger.finest(cause);
+    private void authenticationFailed(Address target, ClientConnection connection, Throwable cause) {
```
pveentjer: I would call this method 'onAuthenticationFailed' because it indicated that you want to do something as a consequence of a failed authentication.

pveentjer:
Are we going to do a proper cleanup for 3.8? 

For a bug fix I'm fine with it. So :+1:

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
jerrinot: temporary until the "Simpler Client IO - Re-architecture" work is finished

pveentjer: can you add a line of comment on the master branch? This way we know you have not turned to the dark side.

PR #8937, opened 2016-09-21
Removed InternalCompletableFuture.getSafely

pveentjer:
replaced by join which was added in 3.7 as replacement for getSafely


pveentjer:
Thanks guys. I like to keep the code clean :)

PR #8938, opened 2016-09-21
Added Near Cache key pre-loader

Donnerbart:


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
```diff
@@ -41,6 +41,16 @@
             newUpdater(NearCacheStatsImpl.class, "evictions");
     private static final AtomicLongFieldUpdater<NearCacheStatsImpl> EXPIRATIONS =
             newUpdater(NearCacheStatsImpl.class, "expirations");
+    private static final AtomicLongFieldUpdater<NearCacheStatsImpl> PERSISTENCE_COUNT =
```
mmedenjak: Are these field updater instances needed? I can see they are only used for `set`.

hazelcast-client/src/main/java/com/hazelcast/client/map/impl/nearcache/ClientHeapNearCache.java
```diff
@@ -200,6 +209,10 @@ public NearCacheStatsImpl getNearCacheStats() {
         return stats;
     }
 
+    @Override
+    public void preload(DataStructureAdapter adapter) {
```
Donnerbart: This implementation is going to be removed, just had to add the method until this PR is merged: https://github.com/hazelcast/hazelcast/pull/8997

hazelcast/src/main/java/com/hazelcast/cache/impl/nearcache/impl/DefaultNearCache.java
```diff
@@ -128,6 +140,9 @@ public void destroy() {
         if (expirationTaskFuture != null) {
             expirationTaskFuture.cancel(true);
         }
+        if (storageTaskFuture != null) {
+            storageTaskFuture.cancel(true);
```
Donnerbart: Yes, I was thinking about this. Have to check how to implement this properly. I hope the renaming solution will be atomic enough until then.

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
Donnerbart: That's a good one. I just copied the actual pattern, but in this case we should have a single writer and the multiple fields are anyway not guarded by a lock (it's just statistics). Maybe I can directly set the volatile fields in this case, yes.

hazelcast/src/main/java/com/hazelcast/util/BufferingInputStream.java
```diff
@@ -0,0 +1,81 @@
+/*
```
Donnerbart: Yes, that was my plan. This is why I added the `BufferingInputStream` in that common package and just removed the fixed `BUFFER_SIZE` from it. If this get's merged, we should be able to re-use the class in HotRestart. I have to look at the other methods which I borrowed from the other HotRestart classes though.

Donnerbart: I just added `readFullyOrNothing()` and `readFully()` to `IOUtil`, so we can re-use all this code for the Near Cache pre-loader and HotRestart. Maybe I'll pull that part out, since the code is tested separately and to shrink this PR.

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
Donnerbart: I just used a lock to guard the update of all those values. So it's atomic and just a single memory barrier.

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
mmedenjak: Hm, I could be wrong but aren't the stats updated by only one thread at a time? I guess this is the periodic store run by `DefaultNearCache.StorageTask#run` which already guards against simultaneous runs.
Also, since the fields are not volatile anymore and the reads are not guarded by the same lock, it means that the reads from an another thread could return stale values.

Donnerbart: Ah, darn, I thought the synchronized will to the trick, but yes, missing the read barrier in the getters... Will switch to volatile again. Stupid me.

tkountis:
👍 looks good

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
@vbekiaris Could you give me a hint what I did wrong here?
```
The following classes are DataSerializable while they should be IdentifiedDataSerializable:
class com.hazelcast.config.NearCachePreloaderConfig$NearCachePreloaderConfigReadOnly
```
No other config class implements `IdentifiedDataSerializable`, is this a false positive due to the inner class? I think this is going to be the new approach for all read-only config classes in HZ 3.9.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -0,0 +1,244 @@
...
+            Data dataKey = serializationService.toData(key);
+            if (dataKey != null) {
+                int dataSize = dataKey.totalSize();
+
+                ensureBufHasRoom(fos, INT_SIZE_IN_BYTES);
+                buf.putInt(dataSize);
+
+                int position = 0;
+                int remaining = dataSize;
+                while (remaining > 0) {
```
mmedenjak: In cases where `remaining > BUFFER_SIZE` we can skip writing to the `ByteBuffer`, just as in `ChunkFileOut`.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -0,0 +1,244 @@
...
+        }
+        return "nearcache-" + toFileName(nearCacheName) + ".store";
+    }
+
+    private static long getElapsedMillis(long started) {
+        return NANOSECONDS.toMillis(System.nanoTime() - started);
+    }
+
+    @SuppressWarnings({"checkstyle:magicnumber", "checkstyle:booleanexpressioncomplexity"})
+    private static int toInt(byte[] bytes) {
```
mmedenjak: There is also the `EndiannessUtil#readIntL` method

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
mmedenjak: Are all fields that have FieldUpdater instances accessed concurrently and are the FieldUpdater instances needed? If they are only accessed by one thread at a time, they can just be volatile. E.g. I believe that `PERSISTENCE_COUNT` is accessed by only one thread. But I could be wrong.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
mmedenjak: And I didn't know about that one :)

hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientNearCachePreloaderTest.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+
+    private static long getPersistenceCount(NearCacheTestContext context) {
+        return context.nearCache.getNearCacheStats().getPersistenceCount();
+    }
+
+    private static void waitForNearCachePersistence(NearCacheTestContext context, int persistenceCount) {
+        long oldPersistenceCount = getPersistenceCount(context);
+        while (getPersistenceCount(context) <= oldPersistenceCount + persistenceCount) {
```
Donnerbart: I changed this and also transformed the tests to use the new unified Near Cache test approach with the new context. I was actually missing a test for the `IMap` scenario, just the `JCache` path was tested. With the unified approach this was easy to add.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
Donnerbart: I actually changed the write part as well to use the counterpart `Bits#writeIntB`, otherwise I depend on the `ByteBuffer` implementation if little or big endian is being used. I hope it's okay to stick with big endian for this, otherwise I have to re-create the binary storage files which are used in the load test.

hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
Donnerbart: Hmm, I had concurrent writers in the first approach, but had to change this to a single writer, due to Jaromir's concern about concurrent disk writes. Actually the only new field with a FieldUpdater is `PERSISTENCE_COUNT` and I used it to have the convenient `PERSISTENCE_COUNT.incrementAndGet(this)` for an atomic increment.

Donnerbart:
Since the last review from Matko I changed the following things due to the TDD review from Peter:
* I simplified the names in the config
* I added a file header with magic bytes and a file format, so we can change / extend this in the future
* I added more tests for invalid header values and an empty store file

What I didn't change is the concurrency style in the `NearCacheStatsImpl`, since I'd like to keep it that way for now (it's a very little overhead and not called frequently, so I like to keep it in the style of the class). I think that is the last remaining comment from Matko's PR and Peter's TDD review.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/config/NearCachePreloaderConfig.java
```diff
@@ -0,0 +1,197 @@
...
+    private int storeIntervalSeconds = DEFAULT_STORE_INTERVAL_SECONDS;
+
+    private NearCachePreloaderConfig readOnly;
+
+    public NearCachePreloaderConfig() {
+    }
+
+    public NearCachePreloaderConfig(NearCachePreloaderConfig nearCachePreloaderConfig) {
+        /**
+         * ===== NOTE =====
```
pveentjer: All comments are notes. The ==== NOTE ==== only causes noise

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/DefaultNearCacheManager.java
```diff
@@ -107,5 +132,96 @@ public void destroyAllNearCaches() {
...
+
+        @Override
+        public void run() {
+            nearCache.preload(adapter);
+            if (scheduledFuture != null) {
+                preloadTaskFutures.remove(scheduledFuture);
+            }
+        }
+
+        void setScheduledFuture(ScheduledFuture<?> scheduledFuture) {
```
pveentjer: What is the added value of this setter? Why not directly write to the scheduledFuture field?

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/DefaultNearCacheManager.java
```diff
@@ -107,5 +132,96 @@ public void destroyAllNearCaches() {
...
+                        storageInProgress.set(false);
+                    }
+                }
+            }
+        }
+
+        private boolean isScheduled(NearCache nearCache, long now) {
+            NearCacheStats nearCacheStats = nearCache.getNearCacheStats();
+            NearCachePreloaderConfig preloaderConfig = nearCache.getPreloaderConfig();
+            if (nearCacheStats.getLastPersistenceTime() == 0) {
```
pveentjer: getLastWriteTime?

hazelcast/src/main/java/com/hazelcast/config/NearCachePreloaderConfig.java
mmedenjak: Also, that's javadoc syntax. If just adding comments, you can prefer block comments.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -0,0 +1,303 @@
...
+
+            int loadedKeys = loadKeySet(bis, adapter);
+
+            long elapsedMillis = getElapsedMillis(started);
+            logger.info(format("Loaded %d keys of Near Cache %s in %d ms", loadedKeys, nearCacheName, elapsedMillis));
+        } catch (Exception e) {
+            logger.warning(format("Could not pre-load Near Cache %s (%s): [%s] %s", nearCacheName, storeFile.getAbsolutePath(),
+                    e.getClass().getSimpleName(), e.getMessage()));
+        } finally {
+            tmpBytes = null;
```
pveentjer: why are you nulling tmpBytes?

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -0,0 +1,303 @@
...
+        this.tmpStoreFile = new File(fileName + "~");
+    }
+
+    /**
+     * Loads the values via a stored key file into the supplied {@link DataStructureAdapter}.
+     *
+     * @param adapter the {@link DataStructureAdapter} to load the values from
+     */
+    public void loadKeys(DataStructureAdapter<Data, ?> adapter) {
+        long started = System.nanoTime();
```
pveentjer: I would name this startedNanos. This is because in a single method you make use of 2 time units; nanos; millis.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -0,0 +1,303 @@
...
+
+            fos.flush();
+            rename(tmpStoreFile, storeFile);
+
+            persistenceStats(started);
+        } catch (Exception e) {
+            logger.warning(format("Could not store keys of Near Cache %s (%s): [%s] %s", nearCacheName,
+                    storeFile.getAbsolutePath(), e.getClass().getSimpleName(), e.getMessage()));
+        } finally {
+            tmpBytes = null;
```
pveentjer: why nulling tmpBytes?

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -0,0 +1,303 @@
...
+        if (builder.size() > 0) {
+            adapter.getAll(builder.build());
+        }
+        return loadedKeys;
+    }
+
+    private void writeKeySet(FileOutputStream fos, FileChannel outChannel, Set<KS> keySet) throws IOException {
+        for (KS key : keySet) {
+            Data dataKey = serializationService.toData(key);
+            if (dataKey != null) {
```
pveentjer: Flavor: to reduce the amount of indenting, I normally do a

```
if(dataKey==null)continue;
```

Depends of course if the flushLocalBuffer(channel) needs to be done or not.

pveentjer:
Looks good. Only some superficial comments

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
Donnerbart: I have no test case for this (since it's not so easy to reproduce), so we would have an untested code branch in the code. In those cases I sometimes write down the "happy path" to have the lines covered.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
Donnerbart: I changed that code part several times. But you're right, I'll make it a final field. I think we can afford a few bytes being held forever.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/DefaultNearCacheManager.java
```diff
@@ -107,5 +132,96 @@ public void destroyAllNearCaches() {
...
+        private final AtomicBoolean storageInProgress = new AtomicBoolean(false);
+
+        @Override
+        public void run() {
+            long now = System.currentTimeMillis();
+            for (NearCache nearCache : nearCacheMap.values()) {
+                if (!isScheduled(nearCache, now)) {
+                    continue;
+                }
+                if (storageInProgress.compareAndSet(false, true)) {
```
Donnerbart: That a leftover from the first multi-threaded implementation. Good catch.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/DefaultNearCacheManager.java
Donnerbart: All of the related stats are named `getPersistenceCount` or `getLastPersistence...`. And `write` could be a write to the Near Cache itself, not the disk. Maybe something with `store`, because we use that term for the `MapStore` (which is mostly a persistent layer as well).

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/DefaultNearCacheManager.java
Donnerbart: Was an external class before and those tasks were handled by the Near Cache itself, not the manager. Removed the setter.

hazelcast/src/main/java/com/hazelcast/config/NearCachePreloaderConfig.java
Donnerbart: I copied that style from all other config classes. I don't like it because it's a dangling JavaDoc comment, which is actually not allowed at that position. But I would prefer to have all config classes in the same style. So I would merge it like this an clean up all config classes at once, if we can agree on a better style.

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/config/NearCachePreloaderConfig.java
pveentjer: :+1:

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
pveentjer: I will not quickly allow a test to define how my method body should look like :)

PR #8948, opened 2016-09-22
Stops background expiration task when cluster state is PASSIVE

ahmetmircik:
Also adds `ClusterStateListener` for internal usages

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
@@ -22,6 +22,7 @@ import com.hazelcast.cluster.ClusterState;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.SerializationConfig;
 import com.hazelcast.core.PartitioningStrategy;
+import com.hazelcast.internal.cluster.ClusterStateListener;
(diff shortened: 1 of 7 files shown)


ahmetmircik:
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/4645/
```
Test Result (1 failure / +1)
com.hazelcast.client.listeners.ListItemListenerOnReconnectTest.testListenersNonSmartRoutingMultipleServer
```

ahmetmircik:
run-lab-run

jerrinot:
Wouldn't be better to introduce a cluster state listeners? Hooking a high-level (map) service into `DefaultNodeExtension` does not feel right. 

@mdogan: WDYT? 

Moreoever I believe the PR as it is now might throw NPE when MapService is disabled.

mdogan:
We initially introduced `NodeExtension.onClusterStateChange(ClusterState newState, boolean persistentChange)` hook to be able to persist cluster-state via hot-restart. Note the additional `persistentChange` flag, this custom flag is solely there for hot-restart. That's why we didn't implement this via listener or a marker service interface. 

I agree that if general services want to be notified about cluster-state changes, it's better to do that with a generic solution, like listeners. But I'm not sure whether that solution should be part of SPI or stay as internal (at least for now).

PR #8950, opened 2016-09-22
Added javaDoc for ExpirationManager

### LINKED ISSUE #8894: TTL callback is invoked with significant delay
Dear Hazelcast folks,

I'm using Hazelcast 3.7.1 and trying to use TTL + eviction listener.
In general eviction listener called with significant delay.
I configured setProperty("hazelcast.internal.map.expiration.task.period.seconds", "1"); as it was suggested in the #7722 issue.

I inserted 100 elements to the map and set TTL to 5. Here the [gist with logs](https://gist.github.com/alexterman/cc39288cbb8e01cda3bcdfd618de114d)
As I can see from the prints my eviction callback called in batches of around 15 items 7 -8 times and the max delay is stands for 12 seconds. 

I was expecting 1 batch of all items.
Should I configure more stuff to get it work?

Thanks,
Alex

ahmetmircik:
fixes #8894

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
@@ -45,13 +45,38 @@ import static java.util.Collections.sort;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 /**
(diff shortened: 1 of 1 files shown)


ahmetmircik:
run-lab-run

ahmetmircik:
Merging this javaDoc update pr.

PR #8954, opened 2016-09-22
OperationServiceImpl.isCallTimedOut cleanup

pveentjer:
This method should not rely on 'returnsResponse' if something is timed out. It should solely rely
on the call timeout.

By default each Operation is created with a Long.MAX_VALUE call timeout so unless this is explicitly changed, an Operation will not timeout. And if configured otherwise it is better to obey the behavior to timeout and not to ignore the timeout because it doesn't return a response.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
@@ -339,10 +339,9 @@ public final class OperationServiceImpl implements InternalOperationService, Met
 
     @Override
     public boolean isCallTimedOut(Operation op) {
-        // Join operations should not be checked for timeout
-        // because caller is not member of this cluster
+        // Join operations should not be checked for timeout because caller is not member of this cluster
         // and can have a different clock.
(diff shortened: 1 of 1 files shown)


pveentjer:
@mdogan can you have a look at this one? I'm trying to cleanup the usage of returnsResponse so we can add more complex operations like batching or 'async' operations in the system in a more natural way.

jerrinot:
it's been there [since 2012](https://github.com/jerrinot/hazelcast/commit/a14d98db070a99d2974c5f9dc1c1b337bc306da4#diff-4156b4cb13226c1e85343044b67c15a4R168). I guess we really need Mehmet to review this.

mdogan:
We were checking `returnsResponse()` because when it returns false, operation's response-handler was null, afaik. I don't know this part exactly anymore, many things changed.
Also what happens when a timeout response is returned when there's no invocation is registered on caller side?

pveentjer:
We can easily deal with the null response handler. Either add a null check, if just always make sure there is an empty response handler set. 

Only in case of a callId > 0, a remote response handler is set. So if there is some kind of system call without an invocation and it runs into a operation timeout ----> it will run into an empty response handler and therefor no response is send to the caller. So this should not lead to any problems.

PR #8961, opened 2016-09-22
SPI & Client Operations & all OperationFactories made IdentifiedSerializable

tombujok:
EE counterpart submitted


hazelcast/src/main/java/com/hazelcast/client/impl/ClientDataSerializerHook.java
```diff
@@ -0,0 +1,70 @@
...
+                switch (typeId) {
+                    case CLIENT_DISCONNECT:
+                        return new ClientDisconnectionOperation();
+                    case RE_AUTH:
+                        return new ClientReAuthOperation();
+                    case GET_CONNECTED_CLIENTS:
+                        return new GetConnectedClientsOperation();
+                    case POST_JOIN:
+                        return new PostJoinClientOperation();
+                    case OP_FACTORY_WRAPPER:
```
pveentjer: I'm confused. We serialize an operation factory (wrapper?)

hazelcast/src/main/java/com/hazelcast/client/impl/operations/ClientDisconnectionOperation.java
```diff
@@ -30,7 +32,7 @@
 import java.util.Collection;
 import java.util.Set;
 
-public class ClientDisconnectionOperation extends Operation implements UrgentSystemOperation {
+public class ClientDisconnectionOperation extends Operation implements UrgentSystemOperation, IdentifiedDataSerializable {
```
pveentjer: Isn't it more convenient to have a single super class 'ClientOperation' that takes care of the IdentifiedDataSerializable? This we can get rid of the duplication and create a bit more cohesion.

hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
```diff
@@ -98,82 +118,100 @@
...
-    public static final int TXN_PREPARE_BACKUP = 68;
-    public static final int TXN_ROLLBACK = 69;
-    public static final int TXN_ROLLBACK_BACKUP = 70;
-    public static final int TXN_SET = 71;
-    public static final int TXN_UNLOCK = 72;
-    public static final int TXN_UNLOCK_BACKUP = 73;
+    public static final MutableInteger ID = new MutableInteger();
 
-    private static final int LEN = TXN_UNLOCK_BACKUP + 1;
+    public static final int PUT = ID.value++;
```
pveentjer: This is very dangerous. I think the existing approach is more solid since you hard code a id to a type of class and therefor it is easier to see which id is mapped to which class + it more difficult to f*ck up. If someone removes e.g. ENTRIES, then all id's are now magically going to shift. With the existing approach that would not happen.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/Invalidation.java
```diff
@@ -30,7 +30,7 @@
 /**
```
pveentjer: We need to get confirmation from e.g. the client or near cache team to see if this class is used on the client.

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/UuidFilter.java
```diff
@@ -27,7 +28,7 @@
  * Compares supplied uuid with this filters' uuid to prevent unneeded delivery of an invalidation event to operation caller.
  * Operation caller invalidates its own local near-cache, no need to send an extra invalidation from remote.
  */
-public class UuidFilter implements EventFilter, DataSerializable {
+public class UuidFilter implements EventFilter, IdentifiedDataSerializable {
```
pveentjer: Client dependence?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/AddIndexOperationFactory.java
```diff
@@ -16,6 +16,7 @@
 
 package com.hazelcast.map.impl.operation;
```
pveentjer: Since there are so many map specific OperationFactories, I think it would be practical to introduce an AbstractMapOperationFactory that already takes care of the IdentifiedDataSerializable implementation and getFactoryId. This way we keep the noise down.

hazelcast/src/main/java/com/hazelcast/multimap/impl/MultiMapDataSerializerHook.java
```diff
@@ -101,6 +102,7 @@
     public static final int TXN_REMOVE_ALL_BACKUP = 38;
     public static final int TXN_ROLLBACK = 39;
     public static final int TXN_ROLLBACK_BACKUP = 40;
+    public static final int MULTIMAP_OP_FACTORY = 41;
```
pveentjer: Next time when you do such a conversion PR, please split them up in logical units. E.g. map conversion PR and multimap conversion PR. I have burned my fingers so many times on a PR that is bigger than it needs to be and the effect is that quite a lot of time (and patience) is being wasted.

hazelcast/src/main/java/com/hazelcast/multimap/impl/operations/MultiMapOperationFactory.java
```diff
@@ -24,8 +25,7 @@
 
 import java.io.IOException;
 
-public class MultiMapOperationFactory
-        implements OperationFactory {
```
pveentjer: For the same reasons as the Abstract Map OperationFactory, please introduce an AbstractMultiMapOperationFactory that can take care of the repeating noise.

hazelcast/src/main/java/com/hazelcast/spi/OperationFactory.java
```diff
@@ -26,9 +26,8 @@
  * {@link OperationService} uses this factory to create {@link Operation}s for each partition by calling
  * {@link #createOperation()}
  * </p>
- *
  */
-public interface OperationFactory extends DataSerializable {
```
pveentjer: Hmmm. In theory we can make this change since we only provide a best effort guarantee in the SPI module, but now we are going to break every implementation that is not making use of IdentifiedDataSerializable. I have no problem with the change by itself; but it should have been done at the beginning. 

@jerrinot WDYT

hazelcast/src/main/java/com/hazelcast/spi/impl/eventservice/impl/TrueEventFilter.java
```diff
@@ -18,15 +18,16 @@
...
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
 import com.hazelcast.spi.EventFilter;
+import com.hazelcast.spi.impl.SpiDataSerializerHook;
 
 import java.io.IOException;
 
 /**
  * An event filter which does not filter any event and always evaluates {@code true} results.
  */
-public final class TrueEventFilter implements EventFilter, DataSerializable {
```
pveentjer: I don't know if the client relies on these classes.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationparker/impl/ParkedOperation.java
```diff
@@ -34,15 +36,18 @@
 
 import static com.hazelcast.util.EmptyStatement.ignore;
 
-class ParkedOperation extends Operation implements Delayed, PartitionAwareOperation {
-    final Queue<ParkedOperation> queue;
-    final Operation op;
-    final BlockingOperation blockingOperation;
-    final long expirationTime;
+public class ParkedOperation extends Operation implements Delayed, PartitionAwareOperation, IdentifiedDataSerializable {
```
pveentjer: ParkedOperation is a local operation and will never be serialized. So it doesn't need to implement any form of serialization. This should be added to the documentation to prevent confusion

hazelcast/src/main/java/com/hazelcast/spi/impl/proxyservice/impl/DistributedObjectEventPacket.java
```diff
@@ -18,13 +18,14 @@
...
-import com.hazelcast.nio.serialization.DataSerializable;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+import com.hazelcast.spi.impl.SpiDataSerializerHook;
 
 import java.io.IOException;
 
 import static com.hazelcast.core.DistributedObjectEvent.EventType;
 
-public final class DistributedObjectEventPacket implements DataSerializable {
+public final class DistributedObjectEventPacket implements IdentifiedDataSerializable {
```
pveentjer: I don't know if there is any client reliance

hazelcast/src/main/java/com/hazelcast/client/impl/operations/ClientDisconnectionOperation.java
tombujok: In general yes, there were just 4 client operations though. I changed it as suggested.

hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
tombujok: Noted. I put it on the TODO task-stack to unify all serialization hooks, so plz let me fix it in a separate PR.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/AddIndexOperationFactory.java
tombujok: Probably a long discussion to decide if it's worth it or not. It clutters the inheritance tree just for one bolierplate method. See `PartitionAwareOperationFactory` but I changed it as suggested since should cause no harm in the modified classes - at least for now.

hazelcast/src/main/java/com/hazelcast/multimap/impl/operations/MultiMapOperationFactory.java
tombujok: That's a premature optimization. There's just one Operation factory for multimap.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationparker/impl/ParkedOperation.java
tombujok: Thx. Fixed

hazelcast/src/main/java/com/hazelcast/multimap/impl/MultiMapDataSerializerHook.java
tombujok: Sure. When I started didn't seem as big as it turned out to be...

hazelcast/src/main/java/com/hazelcast/spi/OperationFactory.java
tombujok: I would say that it's better to break & fix it now. Otherwise we're stucked with the package and naming forever.
It would be great to have a policy banning the usage of DS in HZ classes guarded by a checkstyle/unittest rule run in each build. Otherwise it will be constant struggle.

tombujok:
@pveentjer Thx for the review! Vassilis is working on a tool to inspect all client dependencies. 
Let's wait for his output for the 3 points you mentioned

hazelcast/src/main/java/com/hazelcast/spi/impl/proxyservice/impl/DistributedObjectEventPacket.java
tombujok: Verified, no client reliance

hazelcast/src/main/java/com/hazelcast/client/impl/ClientDataSerializerHook.java
tombujok: Yes, it's the case. It's out of scope of this PR though.

hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
tombujok: Fixed

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/UuidFilter.java
tombujok: Class removed from this PR.

hazelcast/src/main/java/com/hazelcast/spi/impl/eventservice/impl/TrueEventFilter.java
tombujok: Class removed from this PR.

hazelcast/src/main/java/com/hazelcast/spi/impl/proxyservice/impl/DistributedObjectEventPacket.java
tombujok: Class removed from this PR.

tombujok:
All classes about which there was doubt if there's a dependence from the client side has been removed from this PR. It's just Operations and OperationFactories now.

pveentjer:
:+1:  once green and all commits squashed

hazelcast/src/main/java/com/hazelcast/spi/OperationFactory.java
jerrinot: "[...] Otherwise we're stucked with the package and naming forever." <-- I don't see how. 

It's only important for the actual factory implementations to implement the `IdentifiedDataSerializable` interface. Serilization strategy declared by the interface is good as a reminder / way to force the implementer to use a specific strategy. However I don't see any impact on Rolling Upgrade.

tombujok: Yeah, I meant we're stuck with the implementations. Didn't notice that Peter's comment was on the top-level class.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionWideEntryWithPredicateOperationFactory.java
```diff
@@ -155,4 +156,19 @@ public void readData(ObjectDataInput in) throws IOException {
         }
         return partitionToKeys;
     }
+
+    @Override
+    public Operation createOperation() {
```
vbekiaris: Why do we override this method? Since it's a `PartitionAwareOperationFactory`, shouldn't its entry point be `createPartitionOperation(int partition)` and `createOperation` should be left as-is in the parent (throwing `UnsupportedOperationException`)?

tombujok: It's for the deserialiser to create an operation and then populate its fields from the byte stream

PR #8964, opened 2016-09-22
Cluster Operations transformed to IdentifiedDataSerializable

tombujok:
Does not break the EE build.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
@@ -17,28 +17,70 @@
 package com.hazelcast.internal.cluster.impl;
 
 import com.hazelcast.instance.MemberImpl;
+import com.hazelcast.internal.cluster.impl.operations.AuthenticationFailureOperation;
+import com.hazelcast.internal.cluster.impl.operations.AuthorizationOperation;
+import com.hazelcast.internal.cluster.impl.operations.BeforeJoinCheckFailureOperation;
+import com.hazelcast.internal.cluster.impl.operations.ChangeClusterStateOperation;
+import com.hazelcast.internal.cluster.impl.operations.ConfigMismatchOperation;
 import com.hazelcast.internal.cluster.impl.operations.FinalizeJoinOperation;
+import com.hazelcast.internal.cluster.impl.operations.GroupMismatchOperation;
 import com.hazelcast.internal.cluster.impl.operations.HeartbeatOperation;
+import com.hazelcast.internal.cluster.impl.operations.JoinCheckOperation;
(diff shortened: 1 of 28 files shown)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
```diff
@@ -47,31 +91,150 @@ public int getFactoryId() {
 
     @Override
     public DataSerializableFactory createFactory() {
-        return FACTORY;
-    }
+        ConstructorFunction<Integer, IdentifiedDataSerializable>[] constructors = new ConstructorFunction[LEN];
```
tombujok: All non-Operations have been already here in the Hook, see the changes below. I've only added operations,.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/AbstractJoinOperation.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.cluster.impl.operations;
+
+import com.hazelcast.internal.cluster.impl.ClusterDataSerializerHook;
+import com.hazelcast.spi.Operation;
+
+public abstract class AbstractJoinOperation extends Operation implements JoinOperation {
```
vbekiaris: Not sure if it's worth introducing this separate abstract class, its 4 subclasses could as well extend `AbstractClusterOperation` and override `returnsResponse` to return `true`. wdyt?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/AbstractClusterOperation.java
```diff
@@ -16,10 +16,12 @@
 
 package com.hazelcast.internal.cluster.impl.operations;
 
+import com.hazelcast.internal.cluster.impl.ClusterDataSerializerHook;
 import com.hazelcast.internal.cluster.impl.ClusterServiceImpl;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
 import com.hazelcast.spi.Operation;
 
-abstract class AbstractClusterOperation extends Operation implements JoinOperation {
+abstract class AbstractClusterOperation extends Operation implements JoinOperation, IdentifiedDataSerializable {
```
tombujok: Fixed

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/AbstractJoinOperation.java
tombujok: Good point, maybe for readibility let's keep it? I mean if there's more maybe someone will forget to override the returnsResponse. But thx for the comment

PR #8965, opened 2016-09-23
Removed copy paste code OperationServiceImpl.invokeOnPartitions

pveentjer:
The method was copied to deal with int array and collection<Integer>. Code has
been cleaned up so that the int array converts to collection and then forwards
to the main method.


pveentjer:
Since this is a very low risk change, I'm merging this one.

PR #8967, opened 2016-09-23
OperationRunnerImpl sendBackup moved

pveentjer:
this method has been moved to the OperationBackupHandler since it is a more of a concern
for the latter than the former. Else the logic is spread over multiple classes.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
@@ -128,7 +128,7 @@ public final class OperationServiceImpl implements InternalOperationService, Met
     final NodeEngineImpl nodeEngine;
     final Node node;
     final ILogger logger;
-    final OperationBackupHandler operationBackupHandler;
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
```diff
@@ -392,7 +392,20 @@ public boolean isCallTimedOut(Operation op) {
     @Override
     public Map<Integer, Object> invokeOnPartitions(String serviceName, OperationFactory operationFactory, int[] partitions)
             throws Exception {
-        return invokeOnPartitions(serviceName, operationFactory, toIntegerList(partitions));
+        // todo: copy paste logic from the above code.
+        Map<Address, List<Integer>> memberPartitions = new HashMap<Address, List<Integer>>(3);
```
jerrinot: does this belong to this PR?

pveentjer: Let me see what happened here. Perhaps I didn't start from the right branch since this is from another cleanup

PR #8968, opened 2016-09-23
OperationRunnerImpl.handleOperationError remove returnsResponse check

pveentjer:
If there is a response handler set, just send it.

The problem is that a method might decide as part of its run that it does or does not have a response (yet). So it could start out with an initial value 'returnsResponse=false', but if there is an exception, we need to send back an error response no matter of the 'returnsResponse' status because the real response is likely never to be send since the run failed. 

This is partly a cleanup but also it resolves an issue with exceptions not being returned.

There is another important change. In case of a Backup, the callId is set on the backup. This causes the wrong response handler to be set on the Backup; this ResponseHandler is a remote responde handler that returns to the primary; and not the caller. This is potentially dangerous because the response could be matched with an arbitrary invocation on the primary; instead of on the caller. 

Instead of setting the call id on Backup, it is set on the contained BackupOperation. This causes the Backup to be initiated with an Empty ResponseHandler once it gets deserialized.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
@@ -43,6 +43,10 @@ import java.util.Arrays;
 import static com.hazelcast.spi.impl.OperationResponseHandlerFactory.createEmptyResponseHandler;
 import static com.hazelcast.spi.partition.IPartition.MAX_BACKUP_COUNT;
 
+/**
+ * The Backup is a container around the operation that takes care of the actual backup. Backups
+ * are sent from the primary and don't have a call'id set.
+ */
 public final class Backup extends Operation implements BackupOperation, IdentifiedDataSerializable {
 
     private Address originalCaller;
@@ -148,12 +152,12 @@ public final class Backup extends Operation implements BackupOperation, Identifi
 
     @Override
     public void afterRun() throws Exception {
-        if (!valid || !sync || getCallId() == 0 || originalCaller == null) {
+        if (!valid || !sync || backupOp.getCallId() == 0 || originalCaller == null) {
             return;
         }
 
         NodeEngineImpl nodeEngine = (NodeEngineImpl) getNodeEngine();
-        long callId = getCallId();
+        long callId = backupOp.getCallId();
         OperationServiceImpl operationService = (OperationServiceImpl) nodeEngine.getOperationService();
 
         if (nodeEngine.getThisAddress().equals(originalCaller)) {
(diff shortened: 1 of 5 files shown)


pveentjer:
@Donnerbart can you rereview? I cleaned up a AbstractCallableTaskOperation to not do its own error logic any longer since that isn't needed with this fix.

pveentjer:
run-lab-run

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
```diff
@@ -44,6 +44,10 @@
 import static com.hazelcast.spi.partition.IPartition.MAX_BACKUP_COUNT;
 import static java.util.concurrent.atomic.AtomicReferenceFieldUpdater.newUpdater;
 
+/**
+ * The Backup is a container around the operation that takes care of the actual backup. Backups
+ * are send from the primary and don't have a call'id set.
```
Donnerbart: "Backups are sen**t**..."

pveentjer:
run-lab-run

pveentjer:
run-lab-run

pveentjer:
run-lab-run

pveentjer:
@mdogan can you also have a look

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/operations/Backup.java
```diff
@@ -154,12 +158,12 @@ public void run() throws Exception {
 
     @Override
     public void afterRun() throws Exception {
-        if (!valid || !sync || getCallId() == 0 || originalCaller == null) {
+        if (!valid || !sync || backupOp.getCallId() == 0 || originalCaller == null) {
```
mdogan: What if `backupOp` is `null`? According to the `run()` method, that seems possible.

```java
if (backupOp == null && backupOpData != null) {
    backupOpUpdater.lazySet(this, (Operation) nodeEngine.getSerializationService().toObject(backupOpData));
}

// I THINK AT THIS POINT backupOp SHOULD NOT ALLOWED TO BE null.

if (backupOp != null) {
    ensureBackupOperationInitialized();

    backupOp.beforeRun();
    backupOp.run();
    backupOp.afterRun();
}
```

pveentjer: Backup op can't be null. I'll have another look at the method. Thanks for the review.

jerrinot:
run-lab-run

pveentjer:
Closing this one in favor of: https://github.com/hazelcast/hazelcast/pull/11282

PR #8974, opened 2016-09-25
Removed OperationResponseHandler.isLocal

pveentjer:
The OperationResponseHandler.isLocal call is removed. A response handler should handle the response, it should not provide information about where to respond to.

This question can be answered by the Operation; so an additional method has been introduced to check if the operation is running locally or not.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
@@ -80,6 +80,10 @@ public abstract class Operation implements DataSerializable {
         setFlag(true, BITMASK_CALL_TIMEOUT_64_BIT);
     }
 
+    public boolean executedLocally() {
+        return nodeEngine.getThisAddress().equals(callerAddress);
+    }
+
     public boolean isUrgent() {
         return this instanceof UrgentSystemOperation;
     }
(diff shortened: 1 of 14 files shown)


hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -80,6 +80,10 @@ public Operation() {
         setFlag(true, BITMASK_CALL_TIMEOUT_64_BIT);
     }
 
+    public boolean executedLocally() {
```
jerrinot: shouldn't this be overwritten by `Backup` & co.?

pveentjer: No. A backup is always executed on a different node then were it originated (primary). So it is always a remote call; and that is exactly what this method will return since callerAddress (primary) != this address.

Donnerbart: Aren't backups in a one-node cluster executed on the same node? Will this solution work for that scenario?

pveentjer: No backups in 1 node cluster :)

PR #8977, opened 2016-09-26
Fix heartbeat problem for BlockingOperations

pveentjer:
The problem is that the ParkedOperation; the container around the BlockingOperation; is
asked for callid/calleraddress. But this container object doesn't have any sensible
information, so the heartbeat is not constructed correctly for a blocked Operation.


hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_BlockingTest.java
```diff
@@ -263,12 +264,31 @@ public void run() throws Exception {
...
+        q.take();
+    }
+
     /**
      * Tests if the future on a blocking operation can be shared by multiple threads. This tests fails in 3.6 because
      * only 1 thread will be able to swap out CONTINUE_WAIT and all other threads will fail with an OperationTimeoutExcepyion
      */
     @Test
     public void sync_whenManyGettersAndLotsOfWaiting() throws Exception {
+        setLoggingLog4j();
```
jerrinot: left-over?

pveentjer: Fixed.

vbekiaris:
👍

PR #8979, opened 2016-09-26
Fix BlockingOperation heartbeat problem

pveentjer:
The problem is that the WaitingOperation; the container around the BlockingOperation; is
asked for callid/calleraddress. But this container object doesn't have any sensible
information, so the heartbeat is not constructed correctly for a blocked Operation.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
@@ -24,11 +24,11 @@ import com.hazelcast.internal.partition.MigrationInfo;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Address;
 import com.hazelcast.spi.BlockingOperation;
+import com.hazelcast.spi.LiveOperations;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/waitnotifyservice/impl/WaitNotifyServiceImpl.java
```diff
@@ -86,8 +86,10 @@ public WaitNotifyServiceImpl(final NodeEngineImpl nodeEngine) {
     @Override
     public void populate(LiveOperations liveOperations) {
         for (Queue<WaitingOperation> queue : mapWaitingOps.values()) {
-            for (WaitingOperation op : queue) {
-                liveOperations.add(op.getCallerAddress(), op.getCallId());
+            for (WaitingOperation waitingOperation : queue) {
+                // we need to read out the data from the BlockedOperation; not from the ParkerOperation-container.
```
jerrinot: a tiny one - the `ParkerOperation-container.` has a different name in the maintanance branch.

pveentjer: let me do that through a direct push if there are not other comments.

PR #8984, opened 2016-09-26
Improve HyperLogLog coverage & Fix bug in Sparse to Dense conversion.

tkountis:


tkountis:
run-lab-run

hazelcast/src/test/java/com/hazelcast/cardinality/impl/hyperloglog/HyperLogLogEncoderAbstractTest.java
```diff
@@ -52,33 +60,39 @@ public void add() {
...
     @Test
-    public void addAll() {
-        boolean changed = false;
-        for (long hash : new long[] { 1L, 1L, 2000L, 3000, 40000L }) {
-            changed |= encoder.add(hash);
-        }
-        assertEquals(true, changed);
-        assertEquals(4L, encoder.estimate());
-    }
+    public void addBigRange() throws FileNotFoundException {
```
jerrinot: can you describe this test a bit? it's not clear what scenario is it testing.

tkountis: Adding up-to `runLength()` random numbers on both a Set and a HyperLogLog encoder.
Sampling the `estimate` and the `actual` count every 100 operations.
Compare to find the error rate, of the measurements and store them in a histogram.
At the end of the test, we verify that the 99th percentile of the histogram is less than the expected `max` error which is the result of `std error` (1.04 / √m) + 3%. (2% is the typical accuracy, but tests on the implementation showed up rare occurrences of 3%)

jerrinot: can you add this as a comment to the test?

pveentjer:
:+1: once green

PR #8986, opened 2016-09-27
MetricsRegistry weakreference support

pveentjer:
Instead of needing to deregister, lets the gc take care of it. This is done by wrapping
the source in a WeakReference.

The connection logic has been cleaned up:
- no more metrics registry in the constructor
- no more deregister; but rely on weakreference instead


pveentjer:
Going for a different approach that doesn't rely on gc.

PR #8989, opened 2016-09-27
Alligned ClientRead/Write-Handler naming with server

pveentjer:
Names are now alligned with server. This is preparation for SSL
multiple threads issue for 3.8


Donnerbart:
The member class names are `NonBlockingSocketWriter` and `NonBlockingSocketReader` (just to save the lookup for other reviewers).

PR #8991, opened 2016-09-27
Drop client io thread

pveentjer:
Dropped the ClientNonBlockingOutputThread. 

This step is required to let the client run on the same IOThreadingModel as on the server.


pveentjer:
There will be even more re-use once the threading model is merged. The schedule method and related functionality is currently copy paste, but I'm looking for a better approach.

pveentjer:
Thanks guys.

PR #8992, opened 2016-09-27
Added clear near-cache invalidation

ahmetmircik:
Added `ClearNearCacheInvalidation` and removed special case for `SingleNearCacheInvalidation` with null key. Before this PR we were accepting a null key-ed invalidation as a clearing invalidation., this was the special case for `SingleNearCacheInvalidation`. Now every invalidation has its own type.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/BatchInvalidator.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/BatchInvalidator.java
@@ -20,11 +20,10 @@ import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.LifecycleEvent;
 import com.hazelcast.core.LifecycleListener;
 import com.hazelcast.core.LifecycleService;
-import com.hazelcast.map.impl.MapServiceContext;
-import com.hazelcast.map.impl.nearcache.NearCacheProvider;
-import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.spi.EventFilter;
 import com.hazelcast.spi.EventRegistration;
 import com.hazelcast.spi.ExecutionService;
+import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.util.ConstructorFunction;
 
 import java.util.ArrayList;
@@ -42,8 +41,10 @@ import java.util.concurrent.atomic.AtomicInteger;
 import static com.hazelcast.map.impl.MapService.SERVICE_NAME;
 import static com.hazelcast.spi.properties.GroupProperty.MAP_INVALIDATION_MESSAGE_BATCH_FREQUENCY_SECONDS;
 import static com.hazelcast.spi.properties.GroupProperty.MAP_INVALIDATION_MESSAGE_BATCH_SIZE;
+import static com.hazelcast.util.CollectionUtil.isEmpty;
 import static com.hazelcast.util.ConcurrencyUtil.getOrPutIfAbsent;
-import static java.util.Collections.singletonList;
+import static java.lang.Math.min;
+import static java.util.Collections.emptyList;
 
 /**
  * Sends invalidations to Near Cache in batches.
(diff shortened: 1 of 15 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/BatchInvalidator.java
```diff
@@ -110,40 +119,30 @@ private void invalidate(InvalidationQueue invalidationQueue, boolean offloadEven
...
-                continue;
-            }
-
-            if (invalidation.getKey() == null) {
-                foundClearEvent = true;
-                invalidations = singletonList(invalidation);
-            } else {
-                invalidations.add(invalidation);
-            }
+            invalidations.add(invalidation);
```
Donnerbart: Shouldn't we keep the optimization to just send a singleton list if there is a `ClearNearCacheInvalidation`?
```
if (invalidation instanceof ClearNearCacheInvalidation) {
    return singletonList(invalidation);
}
```

ahmetmircik: For now it makes further improvements harder that is why i removed it, planning to re-add it later (fyi: 3.7 doesn't have this optimization)

ahmetmircik:
@Donnerbart removed `equals()` and `hashCode()`

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/BatchNearCacheInvalidation.java
```diff
@@ -48,11 +44,15 @@ public void consumedBy(InvalidationHandler invalidationHandler) {
         invalidationHandler.handle(this);
     }
 
-    public void add(SingleNearCacheInvalidation invalidation) {
+    public void add(Invalidation invalidation) {
```
mmedenjak: Do we even need this add method? It looks like it is only being used from the `BatchNearCacheInvalidationTest#equals_itself_after_deserialization` and the test could very well construct the `BatchNearCacheInvalidation` with the list. The invalidations field default `EMPTY_LIST` can also then be removed since it is always assigned in the constructor.

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
```diff
@@ -443,23 +444,22 @@ public void onInvalidate(Invalidation invalidation) {
...
         @Override
-        public void handle(BatchNearCacheInvalidation batchNearCacheInvalidation) {
-            List<SingleNearCacheInvalidation> invalidations = batchNearCacheInvalidation.getInvalidations();
+        public void handle(BatchNearCacheInvalidation batch) {
+            List<Invalidation> invalidations = batch.getInvalidations();
 
-            for (SingleNearCacheInvalidation invalidation : invalidations) {
-                handle(invalidation);
+            for (Invalidation invalidation : invalidations) {
+                invalidation.consumedBy(this);
```
mmedenjak: Could we perhaps remove this level of indirection? Seems like all `Invalidation` classes just call `invalidationHandler.handle(this)` and we just return the control back to the handler.

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/invalidation/BatchNearCacheInvalidation.java
ahmetmircik: changed it, not using `EMPTY_LIST ` any more.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
ahmetmircik: The main reason for using `invalidationHandler` is to avoid `instanceOf` checks since it is fragile here. By doing it this way, `invalidationHandler` won't be compiled when there is a new `invalidation` type introduced and it will force us to do required changes on relevant code parts. And yes that can cost indirections.

ahmetmircik:
run-lab-run

Donnerbart:
run-lab-run

ahmetmircik:
thanks @Donnerbart

PR #8995, opened 2016-09-27
Ring buffer store and RingbufferContainer refactoring

mmedenjak:
This PR is consisted of two parts - adding a storing mechanism for the ring buffer and refactoring the `RingbufferContainer` class.

The ring buffer storing mechanism now allows reading sequences older than the head sequence thus reading items which are no longer in the ring buffer. Also, upon initialization the ring buffer will ask the store for the newest sequence ID and start from there. This allows the client to replay older messages in a reliable topic via the `ReliableMessageListener` even though the underlying ring buffer has overwritten the items.

The refactoring was made because the `RingbufferContainer` class contained logic for the actual ring buffer structure, the TTL expiration logic, the ring buffer store logic and the key for the wait/notify blocking ops. Some parts are now factored out of this class and can be reused (e.g. the `ArrayRingbuffer`) in other parts of the code.

Also, added some extra Javadoc :)


pveentjer:
If there are 2 parts of a PR; my experience is that it is best to offer 2 PR's. If one if the PR's gets stuck or rejected, the other one can flow through. I have burned my fingers many times on this.

hazelcast/src/main/java/com/hazelcast/config/RingbufferStoreConfig.java
```diff
@@ -0,0 +1,120 @@
...
+import com.hazelcast.core.RingbufferStoreFactory;
+
+import java.util.Properties;
+
+import static com.hazelcast.util.Preconditions.isNotNull;
+
+/**
+ * Configuration for the {@link RingbufferStore}.
+ */
+public class RingbufferStoreConfig {
```
pveentjer: Is there a ReadonlyRingbufferStoreConfig? Normally there is some kind of immutable wrapper around the config.

hazelcast/src/main/java/com/hazelcast/core/RingbufferStore.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param data     the value of the data to store
+     */
+    void store(long sequence, T data);
+
+    /**
+     * Stores multiple entries. Implementation of this method can optimize the store operation.
+     *
+     * @param map map of entries to store
+     */
+    void storeAll(Map<Long, T> map);
```
pveentjer: Is there a reason there needs to be a map? Are the items not a sequence of items? So you would only need an array an an initial sequence for the first item.

hazelcast/src/main/java/com/hazelcast/core/RingbufferStore.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Loads the value of a given sequence. Null value means that the item was not found.
+     *
+     * @param sequence the sequence of the requested item
+     * @return requested item, null if not found
+     */
+    Object load(long sequence);
```
pveentjer: There is no batched version for this method?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -108,197 +137,257 @@ public RingbufferConfig getConfig() {
...
+        return ringbuffer.getCapacity();
     }
 
     public long size() {
-        return tailSequence - headSequence + 1;
+        return ringbuffer.size();
     }
 
     public boolean isEmpty() {
         return size() == 0;
```
pveentjer: I would forward this method to the ringbuffer.

pveentjer:
Minor comments. The code looks very good.

hazelcast/src/main/java/com/hazelcast/config/RingbufferStoreConfig.java
mmedenjak: Added.

hazelcast/src/main/java/com/hazelcast/core/RingbufferStore.java
mmedenjak: Changed.

hazelcast/src/main/java/com/hazelcast/core/RingbufferStore.java
mmedenjak: Not yet. I am thinking of a use case - perhaps when a user requests a sequence and that sequence is before the head sequence, the container could load up to bulk_size items from the store and keep them in memory so that the subsequent loads are faster. The problem is, though, that other subscribers could be too slow and they could request some other sequence ID, thus loading some other sequence range. If we have two slow subscribers, we could end up by alternating loading ranges of items. Or keep them both in memory. But then I would suggest dropping them as they are read so we don't keep them indefinitely. And that has issues too if we have two slow subscribers. 
The queue store doesn't have this problem because it bulk loads for the ID's in the queue and no other ID's.
So I believe the bulk load could be beneficial but I'm not sure yet how to implement it correctly. Any suggestions? We could always add it later if necessary...

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mmedenjak: Done.

mmedenjak:
I will close this PR and open a new one, seems like the branch got corrupted while rebasing

mmedenjak:
I have opened a new PR here since I couldn't reopen this one after a forced push - https://github.com/hazelcast/hazelcast/pull/9012. The changes are the same, incorporated every suggestion from Peter except the batch load since I am not sure if we can do it in a good way.

Issue #8944, opened 2016-09-22
Hazelcast client outside AWS cannot connect to hazelcast node inside AWS : Regression


rajeshnair:
We have just upgraded from 3.3.4 to 3.6.2 
and looks like the below Client program broken

````  
      AwsConfig config = new AwsConfig();
        config.setAccessKey("***********");
        config.setSecretKey("*******");
        config.setHostHeader("ec2.us-east-1.amazonaws.com");
        config.setSecurityGroupName("*****");
        config.setTagKey("hazelcast_service");
        config.setTagValue("true");
        config.setEnabled(true);

         AWSClient aclient = new AWSClient(config);
        for(String s : aclient.getAddresses().keySet()) {
            System.out.print(s+ "  "+aclient.getAddresses().get(s));
        }
       

        ClientConfig clientConfig = new ClientConfig();
        ClientAwsConfig clientAwsConfig = new ClientAwsConfig();
        clientAwsConfig.setInsideAws( false );
        clientAwsConfig.setAccessKey(config.getAccessKey());
        clientAwsConfig.setSecretKey(config.getSecretKey());
        String region = config.getRegion();
        clientAwsConfig.setRegion(region);
        clientAwsConfig.setHostHeader( "ec2."+region+".amazonaws.com" );
        clientAwsConfig.setSecurityGroupName(config.getSecurityGroupName());
        clientAwsConfig.setEnabled(true);
        clientAwsConfig.setConnectionTimeoutSeconds(10000);

        clientConfig.getNetworkConfig().setAwsConfig(clientAwsConfig);
        clientConfig.getNetworkConfig().setConnectionTimeout(10000);

        String clusterName = "myhzcluster";
        clientConfig.getGroupConfig().setName(clusterName);
        String password = "hzdummypwd";
        clientConfig.getGroupConfig().setPassword(password);
        HazelcastClient.newHazelcastClient(clientConfig);
````

The same program connects fine with hazelcast 3.3.4 
but fails to connect with hazelcast 3.6.2 with following error

````
Exception in thread "main" java.lang.IllegalStateException: Unable to connect to any address in the config! The following addresses were tried:[/172.31.31.121:5703, localhost/127.0.0.1:5703, /172.31.31.121:5702, /172.31.31.121:5701, localhost/127.0.0.1:5702, localhost/127.0.0.1:5701]
	at com.hazelcast.client.spi.impl.ClusterListenerSupport.connectToCluster(ClusterListenerSupport.java:174)
[...]

PR #8996, opened 2016-09-27
fixed aws address translator public ip issue

bilalyasar:
fixes https://github.com/hazelcast/hazelcast/issues/8944
Scenario: 
hazelcast member inside aws
hazelcast client outside aws
When client tries to initial connection, it should translate address ( from private to public ). Otherwise it couldn't connect. ( because it tries private)

Also in AwsAddressTranslator, if address is already translated, we shouldn't retranslate it. I added small enhancement/fix.

I couldn't add tests, since it aws related issue. I have manually tested many times.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/AwsAddressTranslator.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/AwsAddressTranslator.java
@@ -24,7 +24,7 @@ import com.hazelcast.logging.LoggingService;
 import com.hazelcast.nio.Address;
 
 import java.net.UnknownHostException;
-import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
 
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/AwsAddressTranslator.java
```diff
@@ -35,7 +36,7 @@
 public class AwsAddressTranslator implements AddressTranslator {
 
     private final ILogger logger;
-    private volatile Map<String, String> privateToPublic;
+    private volatile Map<String, String> privateToPublic = new HashMap<String, String>();
```
mesutcelik: do we still need following checking after this initialization?
`return table != null ? table : Collections.<String, String>emptyMap();`

bilalyasar: thanks, updated.

emrahkocaman:
👍

PR #9005, opened 2016-09-28
[3.7.2] Increase threshold for local wall-clock drift detection

jerrinot:
Too small threshold means a system can be fooled by a lengthy GC pause

(cherry picked from commit 0701c30)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterHeartbeatManager.java
```diff
@@ -61,7 +61,7 @@
  */
 public class ClusterHeartbeatManager {
 
-    private static final long CLOCK_JUMP_THRESHOLD = 10000L;
```
pveentjer: Can you make a TimeUnit.SECONDS.to.. so it is a bit easier to digest?

pveentjer:
It looks alright. @metanet @mdogan can you have a look as well since you know most about this part of the system.

pveentjer:
Ok.. I'll give my thumbs up.

metanet:
well, I am aware that we cannot have an optimal value here but now it is going up to 5 mins from 10 secs. I am fine with increasing it but 5 mins seems a too big number to me.

jerrinot:
@metanet: fair enough. it's 2 minutes now. that was what Mehmet originally suggested.

metanet:
👍

PR #9009, opened 2016-09-28
Used animal sniffer plugin 1.15

ahmetmircik:
Ran into this issue: https://github.com/mojohaus/animal-sniffer/issues/8 and it seems fixed in 1.15


pveentjer:
:+1: once green

PR #9011, opened 2016-09-29
Fixes char array growing issue in Diagnostics

pveentjer:
If a plugin has a lot of data to write, there are 2 buffers that are going to grow and
keep that size ever after; a stringbuffer and char array in the diagnostics sytem. This
is unacceptable since it can lead to unwanted memory usage.

This fix addresses the issue by immediately writing to a PrintWriter instead of
first to a strinbuffer then to an array and then to file.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/diagnostics/DiagnosticsLogFile.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/diagnostics/DiagnosticsLogFile.java
@@ -25,6 +25,7 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetEncoder;
 
@@ -42,7 +43,6 @@ import static java.lang.String.format;
 final class DiagnosticsLogFile {
 
     private static final int ONE_MB = 1024 * 1024;
-    private static final int INITIAL_CHAR_BUFF_SIZE = 4 * 1024;
 
     // points to the file where the log content is written to
     volatile File file;
@@ -51,14 +51,11 @@ final class DiagnosticsLogFile {
     private final ILogger logger;
     private final String fileName;
 
-    private char[] charBuff = new char[INITIAL_CHAR_BUFF_SIZE];
     private int index;
-    private BufferedWriter bufferedWriter;
+    private PrintWriter printWriter;
     private int maxRollingFileCount;
     private int maxRollingFileSizeBytes;
     private final DiagnosticsLogWriter logWriter;
-    // calling File.length generates a lot of litter; so we'll track it ourselves
-    private long fileLength;
 
     DiagnosticsLogFile(Diagnostics diagnostics) {
         this.diagnostics = diagnostics;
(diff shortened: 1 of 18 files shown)


jerrinot:
@pveentjer: I just skimmed it, but do I get it right you flush after with each write? 
also there are some WIP commits

hazelcast/src/main/java/com/hazelcast/internal/diagnostics/DiagnosticsLogFile.java
```diff
@@ -107,36 +104,22 @@ private void renderStaticPlugins() throws IOException {
...
-        fileLength += actualSize;
+        plugin.run(logWriter);
     }
 
-    private BufferedWriter newWriter() throws FileNotFoundException {
+    private PrintWriter newWriter() throws FileNotFoundException {
         FileOutputStream fos = new FileOutputStream(file, true);
         CharsetEncoder encoder = Charset.forName("UTF-8").newEncoder();
-        return new BufferedWriter(new OutputStreamWriter(fos, encoder));
+        return new PrintWriter(new BufferedWriter(new OutputStreamWriter(fos, encoder), Short.MAX_VALUE));
```
Donnerbart: I always get a headache from those stream/writer classes and we have four nested instances here.

Do I get it right that this is a file writer construct with a fixed buffer of 32 MB? And every time that (internal) buffer is full, it's written to disk?

And it's enough to close the outer `PrintWriter` to release all inner streams and writers? The `BufferedWriter` doesn't seem to call close() on its `OutputStreamWriter`. Maybe it's not needed at all, but I'm always a bit careful with resource handling.

pveentjer:
What do you mean with flushing after each write? Each character being written?

hazelcast/src/test/java/com/hazelcast/internal/diagnostics/MetricsPluginTest.java
```diff
@@ -52,7 +52,7 @@ public long get(Object source) throws Exception {
             }
         });
 
-        logWriter.clean();
+        //logWriter.init();
```
Donnerbart: Left-over

jerrinot:
each `write()` being called. plugin generates 1000 short records (lines) then you will flush stuff to a disk 1000x ?

hazelcast/src/main/java/com/hazelcast/internal/diagnostics/SingleLineDiagnosticsLogWriter.java
```diff
@@ -29,66 +31,67 @@
     public void startSection(String name) {
         if (sectionLevel == -1) {
             appendDateTime();
-            sb.append(' ');
+
```
Donnerbart: Superfluous newline

pveentjer:
I flush once per plugin-execution.

pveentjer:
```
private PrintWriter newWriter() throws FileNotFoundException {
         FileOutputStream fos = new FileOutputStream(file, true);
         FileOutputStream fos = new FileOutputStream(file, true);
         CharsetEncoder encoder = Charset.forName("UTF-8").newEncoder();
         CharsetEncoder encoder = Charset.forName("UTF-8").newEncoder();
-        return new BufferedWriter(new OutputStreamWriter(fos, encoder));
+        return new PrintWriter(new BufferedWriter(new OutputStreamWriter(fos, encoder), Short.MAX_VALUE));
     }
     }
```

hazelcast/src/main/java/com/hazelcast/internal/diagnostics/DiagnosticsLogFile.java
pveentjer: 32 kb.

When the buffer is full it gets flushed, or unless it is explicitly flushed (e.g when the plugin has completed rendering its content)

Closing: yes.. that is how it works. So you can close the 'first' one and it automatically gets propagates; the same goes for the flush.

Donnerbart: Okay, that's fine.

For the `close()` propagation you are right for the `PrintWriter` and the `OutputStreamWriter` (which closes its encoder which closes the `FileOutputStream`). But as I already said, the `BufferedWriter` doesn't seem to propagate the `close()` call, it just flushes and sets the reference to `null`. Do we need to do a manual close to release the `OutputStreamWriter` correctly?

Donnerbart: We solved this, it's using try(...) which I didn't see because I wasn't expecting Java 7+ (but it's in the JDK). So it's fine to just close the outer resource. Thanks for that lesson, will help me in my Near Cache pre-loader PR.

PR #9012, opened 2016-09-29
Ring buffer store and RingbufferContainer refactoring

mmedenjak:
This is a recreated PR because the original PR couldn't be reopened after a forced git push. The original PR : https://github.com/hazelcast/hazelcast/pull/8995

This PR is consisted of two parts - adding a storing mechanism for the ring buffer and refactoring the RingbufferContainer class.

The ring buffer storing mechanism now allows reading sequences older than the head sequence thus reading items which are no longer in the ring buffer. Also, upon initialization the ring buffer will ask the store for the newest sequence ID and start from there. This allows the client to replay older messages in a reliable topic via the ReliableMessageListener even though the underlying ring buffer has overwritten the items.

The refactoring was made because the RingbufferContainer class contained logic for the actual ring buffer structure, the TTL expiration logic, the ring buffer store logic and the key for the wait/notify blocking ops. Some parts are now factored out of this class and can be reused (e.g. the ArrayRingbuffer) in other parts of the code.

Also, added some extra Javadoc :)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
@@ -18,12 +18,12 @@ package com.hazelcast.ringbuffer.impl;
 
 import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.config.RingbufferConfig;
+import com.hazelcast.core.HazelcastException;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.nio.serialization.DataSerializable;
 import com.hazelcast.ringbuffer.StaleSequenceException;
-import com.hazelcast.spi.NodeEngine;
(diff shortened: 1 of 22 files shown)


pveentjer:
:+1:

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -108,197 +135,260 @@ public RingbufferConfig getConfig() {
...
+    @SuppressWarnings("unchecked")
+    public void cleanup() {
+        if (expirationPolicy != null) {
+            expirationPolicy.cleanup(ringbuffer);
+        }
+    }
+
     /**
-     * Cleans up the ringbuffer by deleting all expired items.
+     * Check if the sequence can be read from the ring buffer or if the sequence is of the next item to be added into the
```
jerrinot: this is not very clear to me

pveentjer:
Do the javadoc fix in a direct push. I'm going to merge this PR.

mmedenjak:
I've actually fixed the javadoc in this PR. Tried to make it clearer, hope it is now.

PR #9022, opened 2016-09-29
Don't fail when registering a duplicate PortableFactory.

### LINKED ISSUE #2395: Hazelcast 3.2.1: PortableFactory[-23] is already registered! with custom class loader
We are using a custom class loader in our application. After upgrading to Hazelcast 3.2.1 this causes Hazelcast to throw an IllegalArgumentException: PortableFactory[-23] is already registered! in PortableHookLoader.register().

This is caused by the ServiceLoader returning a set of ServiceDefinitions with duplicate class name entries. This in turn seems to caused by a change in revision 88b8e4d (2014-04-14) where the ServiceDefinition.equals() and ServiceDefinition.hashCode() was changed to include the classLoader as well as the className.

Was this change to ServiceDefinition intended?

ejain:
Resolves #2395.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/PortableHookLoader.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/PortableHookLoader.java
@@ -80,13 +80,7 @@ final class PortableHookLoader {
     private void register(int factoryId, PortableFactory factory) {
         final PortableFactory current = factories.get(factoryId);
         if (current != null) {
-            if (current.equals(factory)) {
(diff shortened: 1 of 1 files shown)


jerrinot:
Hello @ejain
thanks for your PR. I appreciate your effort, however I don't think it's the right solution. 

It'll hide the original problem: The same PortableFactory is loaded by 2 different classloaders. It smells like a gate for hard-to-find `ClassCastException`s and failing fast is the right thing to do. What we should do instead is to add extra logging to make the real root cause analysis easier. I can imagine we could include classloader of both factories in the exception. wdyt?

ejain:
Regarding the root cause, see https://github.com/hazelcast/hazelcast/issues/2395#issuecomment-245461396.

PR #9026, opened 2016-09-30
Decoupling IOThreadingModel from server

pveentjer:
Decouples the IOThreadingModel from the serverside TcpIpConnection and related classes. This is needed to let the client make use of the same threading model.


pveentjer:
run-lab-run

jerrinot:
what's the end-goal here? I see the interface no longer depends on the `TcpConnection`, but the implementations do.

pveentjer:
The end goal is that the NonBlockingIOThreadingModel can be shared between client and server. But I'm going step by step here

PR #9032, opened 2016-10-03
Fixes problem in the IORebalancing

pveentjer:
The error was simple; the eventCount field was copied into the AbstractHandler
but it was not removed from the NonBlockingSocketWriter (subclass). So we have one eventCount
hiding another. The one in the AbstractHandler was used for migratation load
detection, but the one on NonBlockingSocketWriter gets updated by the io system.


metanet:
👍

pveentjer:
Thanks

PR #9038, opened 2016-10-03
Fixing queue load bug 

mmedenjak:
The bug surfaced when the bulkLoad parameter is greater than 1 (the bulk load is enabled) and the load method was trying to load data for items after the bulk load parameter. The load method would then collect the IDs for the first bulkLoad items in the queue and omit the ID of the item actually requested, thus never loading the data for the item.

Also added and fixed some javadoc.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/collection/impl/queue/QueueStoreTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/collection/impl/queue/QueueStoreTest.java
@@ -79,6 +79,30 @@ public class QueueStoreTest extends HazelcastTestSupport {
     }
 
     @Test
+    public void testRemoveAll() {
+        int maxSize = 2000;
+        final Config config = new Config();
+        final QueueStoreConfig queueStoreConfig = new QueueStoreConfig()
+                .setStoreImplementation(new TestQueueStore())
+                .setProperty("bulk-load", String.valueOf(200));
+        config.getQueueConfig("testQueueStore")
+                .setMaxSize(maxSize)
+                .setQueueStoreConfig(queueStoreConfig);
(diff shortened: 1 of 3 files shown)


hazelcast/src/test/java/com/hazelcast/collection/impl/queue/QueueStoreTest.java
```diff
@@ -79,6 +80,28 @@ public void testQueueStoreLoadMoreThanMaxSize() {
...
+        final HazelcastInstance instance = createHazelcastInstance(config);
+        final IQueue<Object> queue = instance.getQueue("testQueueStore");
+
+        final ArrayList<Integer> data = new ArrayList<Integer>();
+        for (int i = 0; i < maxSize; i++) {
+            queue.add(i);
+            data.add(i);
+        }
+        queue.removeAll(data);
+        assertEquals(0, queue.size());
```
pveentjer: Should data not be verified? Because if removeAll would just drop the items and not put them in 'data', this test would not fail.

mmedenjak: I am not quite sure what should have been verified but I have rewritten the test to a simpler version which fails on the existing codebase with a NPE and passes with the proposed changes. Please clarify if it needs more changes.

pveentjer: Something like this:

```
boolean result =  queue.removeAll(data);

assertTrue(result);
assertEquals(0, queue.size());
assertEquals(asList(1,2,3,4), data)
```
So that the content of 'data' is verified and also the return value.

PR #9039, opened 2016-10-03
Fixed NPE in AlignmentUtil when Unsafe is not available.

pveentjer:
The fix is adding a check if Unsafe available and defaulting if not.


jerrinot:
@pveentjer: back-port?

sertugkaya:
@pveentjer repeating the @jerrinot's question above as a reminder.

PR #9045, opened 2016-10-04
Added missing generics to CacheConfig setters

Donnerbart:


Donnerbart:
run-lab-run

pveentjer:
Could this cause subtle backwards compatibility issues? @mtopolnik @jerrinot @vbekiaris WDYT?

Apart from that it looks fine.

mtopolnik:
I can't think of any problems. Before this change, some methods already had the type params and some didn't.

pveentjer:
I'm no expert, but can there be potential binary incompatibility?

Donnerbart:
I think in the worst case this kind of incompatibility is accepted in minor releases, but not patch releases, right?

pveentjer:
Also not in minor releases; it should be a drop in replacement.

If there is truly no binary compatibility issue; I would be more than happy to give thumbs up.

But my knowledge is lacking here.

PR #9048, opened 2016-10-04
Fixed setting of default max size in EvictionConfig

Donnerbart:
Fixed the setting of the default max size in the `EvictionConfig` for on-heap `IMap`s by doing this once in the `Config` and `ClientConfig` class. We actually have to do this before we transform the configuration to a read-only version.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/config/ClientConfig.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/config/ClientConfig.java
@@ -37,6 +37,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 
+import static com.hazelcast.config.NearCacheConfigAccessor.initDefaultMaxSizeForOnHeapMaps;
 import static com.hazelcast.util.Preconditions.checkFalse;
 
 /**
@@ -138,7 +139,7 @@ public class ClientConfig {
         return this;
     }
 
-     /**
+    /**
      * Gets {@link java.util.Properties} object
      *
      * @return {@link java.util.Properties} object
@@ -288,7 +289,13 @@ public class ClientConfig {
         NearCacheConfig nearCacheConfig = lookupByPattern(nearCacheConfigMap, name);
         if (nearCacheConfig == null) {
             nearCacheConfig = nearCacheConfigMap.get("default");
+            if (nearCacheConfig != null) {
+                // if there is a default config we have to clone it,
+                // otherwise you will modify the same instances via different Near Cache names
+                nearCacheConfig = new NearCacheConfig(nearCacheConfig);
+            }
         }
+        initDefaultMaxSizeForOnHeapMaps(nearCacheConfig);
         return nearCacheConfig;
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
@@ -57,7 +57,6 @@ public class EvictionConfig implements EvictionConfiguration, DataSerializable, 
      */
     public static final EvictionPolicy DEFAULT_EVICTION_POLICY = EvictionPolicy.LRU;
 
-    protected boolean sizeConfigured;
     protected int size = DEFAULT_MAX_ENTRY_COUNT;
     protected MaxSizePolicy maxSizePolicy = DEFAULT_MAX_SIZE_POLICY;
     protected EvictionPolicy evictionPolicy = DEFAULT_EVICTION_POLICY;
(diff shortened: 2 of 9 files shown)


hazelcast/src/main/java/com/hazelcast/config/NearCacheConfigAccessor.java
```diff
@@ -19,18 +19,23 @@
...
-    public static EvictionConfig initDefaultMaxSize(EvictionConfig evictionConfig) {
-        if (!evictionConfig.sizeConfigured) {
-            evictionConfig.setSizeInternal(EvictionConfig.DEFAULT_MAX_ENTRY_COUNT_FOR_ON_HEAP_MAP);
+    public static void initDefaultMaxSizeForOnHeapMaps(NearCacheConfig nearCacheConfig) {
+        if (nearCacheConfig == null) {
+            return;
+        }
+
+        EvictionConfig evictionConfig = nearCacheConfig.getEvictionConfig();
+        if (nearCacheConfig.getInMemoryFormat() != InMemoryFormat.NATIVE && !evictionConfig.sizeConfigured) {
```
Donnerbart: Added a missing check for `NATIVE`, so we don't change the value for off-heap maps.

hazelcast/src/main/java/com/hazelcast/config/NearCacheConfigReadOnly.java
```diff
@@ -81,9 +81,4 @@ public NearCacheConfig setLocalUpdatePolicy(LocalUpdatePolicy localUpdatePolicy)
     public NearCacheConfig setEvictionConfig(EvictionConfig evictionConfig) {
         throw new UnsupportedOperationException("This config is read-only");
     }
-
-    @Override
-    public EvictionConfig getEvictionConfig() {
-        return super.getEvictionConfig().getAsReadOnly();
```
Donnerbart: This was causing a call of `checkEvictionConfig()` which doesn't work for some Near Cache configurations. Need to fix this separately. But this causes some EE tests to fail at the moment.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -176,13 +176,6 @@ public int getSize() {
     }
 
     public EvictionConfig setSize(int size) {
-        return setSizeInternal(size);
-    }
-
-    /**
-     * Package private method for {@link EvictionConfigAccessor}.
-     */
-    EvictionConfig setSizeInternal(int size) {
```
Donnerbart: Not needed anymore, since we work on the modifiable configuration, before we make it a read-only one.

hazelcast-client/src/main/java/com/hazelcast/client/config/ClientConfig.java
```diff
@@ -287,8 +288,9 @@ public ClientConfig addProxyFactoryConfig(ProxyFactoryConfig proxyFactoryConfig)
     public NearCacheConfig getNearCacheConfig(String name) {
         NearCacheConfig nearCacheConfig = lookupByPattern(nearCacheConfigMap, name);
         if (nearCacheConfig == null) {
-            nearCacheConfig = nearCacheConfigMap.get("default");
+            nearCacheConfig = new NearCacheConfig(nearCacheConfigMap.get("default"));
```
Donnerbart: Peter notices that it's dangerous to use the `default` configuration without cloning, since we would return the same instance for different names. This should fix that (good finding!).

hazelcast-client/src/main/java/com/hazelcast/client/config/ClientConfig.java
```diff
@@ -288,7 +289,13 @@ public NearCacheConfig getNearCacheConfig(String name) {
         NearCacheConfig nearCacheConfig = lookupByPattern(nearCacheConfigMap, name);
         if (nearCacheConfig == null) {
             nearCacheConfig = nearCacheConfigMap.get("default");
+            if (nearCacheConfig != null) {
+                // if there is a default config we have to clone it,
+                // otherwise you will modify the same instances via different Near Cache names
+                nearCacheConfig = new NearCacheConfig(nearCacheConfig);
```
Donnerbart: Peter noticed that it's dangerous to use the `default` configuration without cloning, since we would return the same instance for different names. This should fix that (good finding!).

hazelcast/src/main/java/com/hazelcast/config/NearCacheConfigReadOnly.java
mmedenjak: Beware that now the getEvictionConfig on the NearCacheConfigReadOnly returns a mutable EvictionConfig which differs from other *ReadOnly config classes with nested instances.

Donnerbart: I just added that getter two days ago, so this is not a huge drawback. I still have to figure out a solution for this "the `EvictionConfig` needs different sanity checks depending on the context" issue.

PR #9049, opened 2016-10-05
Pulled out AbstractNonBlockingSocketReader

pveentjer:
this class can be shared between client and server implementations by subclassing.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketReader.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketReader.java
@@ -16,11 +16,8 @@
 
 package com.hazelcast.nio.tcp.nonblocking;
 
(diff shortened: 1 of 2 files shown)


pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/nio/tcp/nonblocking/NonBlockingSocketReader.java
```diff
@@ -53,121 +47,24 @@
  * {@link ReadHandler} to get processed.
  */
 public final class NonBlockingSocketReader
```
jerrinot: don't you want to use a name as `MemberNonBlockingSockerReader` ? something what makes this subclass unique.

pveentjer: The subclass is unique. The client is called ClientNonBlockingSocketReader.

PR #9056, opened 2016-10-06
SocketWriterInitializer

pveentjer:
Intead of using subclassing approach for SocketReader/SocketWriter (this would lead to 4 subclasses for readers and 4 for writes) the SocketWriterInitializer is used that gets injected as a strategy.


pveentjer:
This PR causes a problem in enterprise SSL. I'm looking at the it.

PR #9057, opened 2016-10-06
SocketReaderInitializer

pveentjer:
Introduces SocketReaderInitializer. 

Needed to make the io threading model sharable between client and server.

Depends on this PR:
https://github.com/hazelcast/hazelcast/pull/9056

Also resolves a potential bug. If not all protocol bytes are received, the protocol information could get lost in case of a NoneBlockingSocketReader because it will immediately throw an Exception instead of waiting for the missing bytes.


pveentjer:
run-lab-run

PR #9063, opened 2016-10-07
Packet stateless

pveentjer:
The Packet doesn't track any state for reading/writing on io. So the same
packet instance can be send to multiple connections if needed, reducing
pressure on the heap.

No further changes have been made in the system to make use of this.


pveentjer:
Closing this one. Going for a different approach where the Packet is completely removed on the write path.

PR #9065, opened 2016-10-07
IOThreading models not dependent on server

pveentjer:
Disconnects the io threadmodels from the serverside logic.
Changed client io so it fully depends on the threadmodel logic.
Currently threadcount on client is unchanged; modifications to these settings is a part of a different pr.

On the serverside we have an IOThreadModel which makes it possible to play with different thread models.. like nonblocking using selectors (like the client is using), but also spinning approach. This is abstracted with this change.. this iothreadmodel is not serverside specific anymore.. but a general purpose something.. which can now be used by the client so both client and server can use the same iothreadmodel abstraction

On the client there is currently 1 input and 1 output thread; on the serverside we have 3+3..  for the ssl improvements we need to have 3+3 for the client as well. Changing the threadcounts is something for a different PR. I have integrated the IOThreadModel in the client; but its original settings are preserved


pveentjer:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientMembershipListener.java
```diff
@@ -162,9 +162,7 @@ private void memberRemoved(Member member) {
         logger.info(membersString());
         final Connection connection = connectionManager.getConnection(member.getAddress());
         if (connection != null) {
-            connectionManager.destroyConnection(connection,
-                    null,
-                    newTargetDisconnectedExceptionCausedByMemberLeftEvent(member, connection.toString()));
+            connection.close(null, newTargetDisconnectedExceptionCausedByMemberLeftEvent(member, connection.toString()));
```
ihsandemir: Destroy also removes from the connections map. Why did you change this?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientMembershipListener.java
```diff
@@ -196,8 +194,7 @@ private void fireMembershipEvent(List<MembershipEvent> events) {
             if (clusterService.getMember(address) == null) {
                 Connection connection = connectionManager.getConnection(address);
                 if (connection != null) {
-                    connectionManager.destroyConnection(connection, null,
-                            newTargetDisconnectedExceptionCausedByMemberLeftEvent(member, connection.toString()));
+                    connection.close(null, newTargetDisconnectedExceptionCausedByMemberLeftEvent(member, connection.toString()));
```
ihsandemir: why change?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
```diff
@@ -257,7 +257,7 @@ public void heartbeatStopped(Connection connection) {
                     clientConnection.getLastHeartbeatReceivedMillis(),
                     clientConnection.lastReadTimeMillis(),
                     clientConnection.getCloseCause());
-            connectionManager.destroyConnection(connection, null, ex);
+            connection.close(null, ex);
```
ihsandemir: why change?

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
```diff
@@ -147,7 +147,7 @@ public void stateChanged(LifecycleEvent event) {
             }
         });
 
-        connectionManager.destroyConnection(connectionToServer, null, null);
+        connectionToServer.close(null, null);
```
ihsandemir: why change?

hazelcast-client/src/test/java/com/hazelcast/client/ClientConnectionTest.java
```diff
@@ -157,7 +157,7 @@ public void run()
...
```
ihsandemir: why change?

hazelcast-client/src/test/java/com/hazelcast/client/test/TestClientRegistry.java
```diff
@@ -358,9 +358,7 @@ public boolean equals(Object o) {
         @Override
         public void close(String reason, Throwable cause) {
             super.close(reason, cause);
-            ClientConnectionManager connectionManager = responseConnection.getConnectionManager();
-            connectionManager.destroyConnection(responseConnection, reason,
-                    new TargetDisconnectedException("Mocked Remote socket closed"));
+            responseConnection.close(reason, new TargetDisconnectedException("Mocked Remote socket closed"));
```
ihsandemir: why change from destroy?

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -118,50 +119,52 @@ public void handle(OutOfMemoryError error) {
...
-                loggingService.getLogger(NonBlockingIOThread.class),
-                outOfMemoryHandler);
-        client.getMetricsRegistry().scanAndRegister(inputThread, "tcp." + inputThread.getName());
-
-        outputThread = new NonBlockingIOThread(
-                client.getThreadGroup(),
-                client.getName() + ".thread-out",
-                loggingService.getLogger(NonBlockingIOThread.class),
-                outOfMemoryHandler);
-        client.getMetricsRegistry().scanAndRegister(outputThread, "tcp." + outputThread.getName());
```
pveentjer: What do you mean with covered? They are being sucked into the metrics registry and since 3.7 the client can also make use of that. The client even has client side diagnostics which can write to file.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientMembershipListener.java
pveentjer: There is not a change. The close of the connection is always done on the connection; the connection then will do a callback to the connection manager to remove whatever needs to be removed. Otherwise you have 2 api's to close a connection; this has been simplified to a single API.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientMembershipListener.java
pveentjer: The connection.close is leading. It will do a callback to the connection manager to cleanup whatever needs to be cleaned up.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
pveentjer: I get your comment now. Yes. The IOThreadModel takes care of the metrics on the threads and whatever else is required.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -364,19 +363,24 @@ protected ClientConnection createSocketConnection(final Address address) throws
...
             logger.info("Removed connection to endpoint: " + endPoint + ", connection: " + connection);
 
-            conn.close(reason, cause);
-            for (ConnectionListener connectionListener : connectionListeners) {
-                connectionListener.connectionRemoved(conn);
+            for (ConnectionListener listener : connectionListeners) {
+                listener.connectionRemoved(connection);
             }
-        } else {
-            connection.close(reason, cause);
```
pveentjer: Nobody else than the connection will call the 'onClose'. So the connection is being closed.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -364,19 +363,24 @@ protected ClientConnection createSocketConnection(final Address address) throws
...
+    }
+
     @Override
-    public void destroyConnection(final Connection connection, final String reason, final Throwable cause) {
+    public void onClose(Connection connection) {
         Address endPoint = connection.getEndPoint();
-        ClientConnection conn = (ClientConnection) connection;
-        if (endPoint != null && connections.remove(endPoint, conn)) {
+
+        if (endPoint != null && connections.remove(endPoint, connection)) {
```
ihsandemir: Is there a possibility here where the listeners may not be notified in this case: a new connection to the same address is constructed and put into the map before this line is executed. Then the connection shall be removed but the listeners won't be notified, is this possible?

pveentjer: If that could happen, it is already a bug in the existing code.

ihsandemir: we can open an issue to investigate this possibility.

Issue #9006, opened 2016-09-28
3.6.x client cannot obtain cache proxy from 3.7.x cluster


vbekiaris:
To reproduce, have a 3.6 client connect on a 3.7 cluster and execute:
```
HazelcastInstance hz = HazelcastClient.newHazelcastClient();
CachingProvider cachingProvider = HazelcastClientCachingProvider.createCachingProvider(hz);
CacheManager cacheManager = cachingProvider.getCacheManager();
Cache<String,String> cache = cacheManager.getCache("test", String.class, String.class);
```
Client fails with:

```
Exception in thread "main" javax.cache.CacheException: Error opening URI [hazelcast]
	at com.hazelcast.cache.impl.AbstractHazelcastCachingProvider.getCacheManager(AbstractHazelcastCachingProvider.java:97)
	at com.hazelcast.cache.impl.AbstractHazelcastCachingProvider.getCacheManager(AbstractHazelcastCachingProvider.java:126)
	at CacheClientBootstrap.main(CacheClientBootstrap.java:33)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
Caused by: com.hazelcast.cache.CacheNotExistsException: Couldn't find cache config with name pRef
	at com.hazelcast.cache.impl.AbstractCacheService.checkCacheSimpleConfig(AbstractCacheService.java:197)
	at com.hazelcast.cache.impl.AbstractCacheService.createDistributedObject(AbstractCacheService.java:177)
	at com.hazelcast.spi.impl.proxyservice.impl.ProxyRegistry.doCreateProxy(ProxyRegistry.java:177)
	at com.hazelcast.spi.impl.proxyservice.impl.ProxyRegistry.createProxy(ProxyRegistry.java:170)
	at com.hazelcast.spi.impl.proxyservice.impl.ProxyServiceImpl.initializeDistributedObject(ProxyServiceImpl.java:133)
	at com.hazelcast.spi.impl.proxyservice.impl.operations.InitializeDistributedObjectOperation.run(InitializeDistributedObjectOperation.java:42)
[...]

Issue #9004, opened 2016-09-28
Hazelcast 3.7.x client cannot obtain cache proxy from 3.6.x cluster


vbekiaris:
Steps to reproduce:
- Start a 3.6.x cluster with a `Cache` configured
- Start a 3.7.x client, then obtain a reference to the configured cache:

```
HazelcastInstance hz = HazelcastClient.newHazelcastClient();
CachingProvider cachingProvider = HazelcastClientCachingProvider.createCachingProvider(hz);
CacheManager cacheManager = cachingProvider.getCacheManager();
Cache<String,String> cache = cacheManager.getCache("test", String.class, String.class);
```

Fails with exception:

```
Exception in thread "main" com.hazelcast.nio.serialization.HazelcastSerializationException: Problem while reading DataSerializable, namespace: 0, id: 0, class: 'com.hazelcast.config.CacheEvictionConfig', exception: java.lang.Class cannot be cast to com.hazelcast.internal.eviction.EvictionPolicyComparator
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.rethrowReadException(DataSerializableSerializer.java:141)
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:130)
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:52)
    at com.hazelcast.internal.serialization.impl.StreamSerializerAdapter.read(StreamSerializerAdapter.java:46)
    at com.hazelcast.internal.serialization.impl.AbstractSerializationService.readObject(AbstractSerializationService.java:216)
    at com.hazelcast.internal.serialization.impl.ByteArrayObjectDataInput.readObject(ByteArrayObjectDataInput.java:600)
    at com.hazelcast.config.CacheConfig.readData(CacheConfig.java:566)
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:127)
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:52)
    at com.hazelcast.internal.serialization.impl.StreamSerializerAdapter.read(StreamSerializerAdapter.java:46)
    at com.hazelcast.internal.serialization.impl.AbstractSerializationService.toObject(AbstractSerializationService.java:172)
[...]

PR #9066, opened 2016-10-07
Cache getconfig and create config does not work for 3.6.x client connecting to the 3.8-SNAPSHOT server

ihsandemir:
Adds `toObject(Data, Class)` interface to `SerializationService`. This interface allows deserialization using a known object type.

Adds the capability to deserialize a `DataSerializable` using a custom class name in the deserialized data rather than the actual object class name (e.g. the derived class name).

Adds client and server version exchange during connection authentication.

Adds compatibility fixes for Cache createConfig and getConfig requests and responses.

Adds a 3.6.x compatibility switch to the server and client. Namely, the following properties: `hazelcast.compatibility.3.6_client`, `hazelcast.compatibility.3.6_server`. By default, this property is not set and 3.7.3, 3.8-SNAPSHOT will work with 3.7.x but not with 3.6.x. To make it work with 3.6.x, then the switch should be used.

Fixes https://github.com/hazelcast/hazelcast/issues/9006 and https://github.com/hazelcast/hazelcast/issues/9004


ihsandemir:
run-lab-run

vbekiaris:
run-lab-run

ihsandemir:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientClusterService.java
```diff
@@ -114,4 +114,9 @@
      */
     Address getOwnerConnectionAddress();
 
+    /**
+     *
+     * @return The connected server hazelcast version as calculated by BuilInfo class.
+     */
+    int getHazelcastServerVersion();
```
ihsandemir: Ok, I will review the design according to this.

ihsandemir: @vbekiaris Can you check the latest commit, how does it look?

ihsandemir:
run-lab-run

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheCreateConfigMessageTask.java
```diff
@@ -56,7 +76,7 @@ protected Operation prepareOperation() {
     protected ClientMessage encodeResponse(Object response) {
         CacheService service = getService(getServiceName());
         final Data responseData = service.toData(response);
-        return CacheCreateConfigCodec.encodeResponse(serializationService.toData(responseData));
+        return CacheCreateConfigCodec.encodeResponse(responseData);
```
vbekiaris: Client 3.6.5 with member 3.8-SNAPSHOT breaks for me here; we are missing the compatibility flag check to respond with the appropriate `CacheConfig` type.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEndpointImpl.java
```diff
@@ -65,6 +67,7 @@ public ClientEndpointImpl(ClientEngineImpl clientEngine, Connection conn) {
         } else {
             socketAddress = null;
         }
+        this.clientVersion = -1;
```
vbekiaris: `UNKNOWN_HAZELCAST_VERSION`?

hazelcast/src/main/java/com/hazelcast/config/CacheConfig.java
```diff
@@ -75,7 +75,7 @@
      * Disables invalidation events for per entry but full-flush invalidation events are still enabled.
      * Full-flush invalidation means the invalidation of events for all entries when clear is called.
      */
-    private boolean disablePerEntryInvalidationEvents;
+    private boolean disablePerEntryInvalidationEvents = true;
```
vbekiaris: Isn't this changing the default `Cache` configuration? What's the reasoning for this change?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DataSerializableSerializer.java
```diff
@@ -131,6 +133,45 @@ public DataSerializable read(ObjectDataInput in) throws IOException {
         }
     }
 
+    @Override
+    public DataSerializable read(ObjectDataInput in, Class aClass)
```
vbekiaris: This method shares most code with the existing `DataSerializable read(ObjectDataInput in)` method, it would be nice to deduplicate (eg have both public methods delegate to an internal method). The same is true for most changes under `internal/serialization` package.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
```diff
@@ -80,6 +90,15 @@ private Connection getConnection(Address target) throws IOException {
         return connection;
     }
 
+    private Connection getOrConnect(Address target) throws IOException {
```
vbekiaris: Is this method any different from `getOrTriggerConnect`?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
```diff
@@ -67,6 +67,8 @@
     private volatile Address ownerConnectionAddress;
     private volatile ClientPrincipal principal;
 
+    private int hazelcastServerVersion;
```
vbekiaris: Leftover?

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEndpointImpl.java
ihsandemir: Ok, Done

hazelcast/src/main/java/com/hazelcast/config/CacheConfig.java
ihsandemir: OK, corrected this.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DataSerializableSerializer.java
vbekiaris: Looks good!

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
ihsandemir: OK, yes, removed it.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
ihsandemir: Yes, triggerConnect does not wait for connection establishment but just triggers a connect but this method waits for the connection establishment.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheHelper.java
```diff
@@ -43,6 +50,7 @@
  * Helper class for some client cache related stuff.
  */
 final class ClientCacheHelper {
+    private static final ILogger LOGGER = Logger.getLogger(ClientCacheHelper.class);
```
ihsandemir: OK, removed.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientClusterService.java
vbekiaris: IMHO instead of exposing `Connection` in `ClientInvocationService`, I would rather keep `ClientClusterService.getClusterVersion()` as I think it more accurately describes it's intended meaning: we want to decide which type to use based on the cluster version.
In any case, since there is currently no notion of cluster version, we could as well defer this change (if needed) to a further PR in the context of rolling-upgrades, I am just worried that if we do introduce a new method in SPI in this PR we may be stuck with it. wdyt?

ahmetmircik:
May there be a problem when working with different client types? In a scenario like c#, node.js and java clients connecting to same cluster together... what kind of things can happen? which parts are we expecting to work or not to work?

ahmetmircik:
One other point, is there a plan to prevent similar failures in next releases?

ihsandemir:
@ahmetmircik I forgot to mention that non-java clients do not implement ICache at all since it is a JCache standart, a java specific structure. Hence, we will not have the same problem with non-java clients.

ahmetmircik:
i would like to see this PR separated into multiple PR-s. For example, one is adding version, one is adding toObject etc. Now it seems hard to review. 
Also it will be hard for the reviewer to catch potential latent errors. Another thing, if it is possible keeping the code of this fix isolated, as much as we can do, will be beneficial for the future enhancement.

vbekiaris:
@ahmetmircik wrt plan for avoiding similar compatibility issues, there is some work underway to document classes which are used implicitly in client-member communication (transferred as opaque `Data`); see #9102. 
This by itself is not a plan, but it should enable us to automate compatibility testing targetting these classes (for example, see https://github.com/vbekiaris/hazelcast/commit/d6b4e52760f30e122fd7dd10f4bd8747ef1d5ad1 wherein we determine whether all `DataSerializable` and non-`@ClientProtocol` classes are also `IdentifiedDataSerializable` in the context of rolling upgrades work).

ihsandemir:
run-lab-run

ihsandemir:
@ahmetmircik I understand your concern about separating these changes into separate PR, but then what if we merge the serialization change PR and when some reviewer in this PR says that we should do it another way and we do not need the serialization change, then we need to revert that merged PR. what do you think? Hence, I understand it being complicated with two things in one PR but they are related and based on the solution which is being reviewed and it is subject to change based on the reviews.

asimarslan:
@ahmetmircik @ihsandemir I don't think we should split this PR. The only reason of the serialization change is to support a requirement of this JCache fix. I think the changes caused by same reason should be kept in the same PR.

PR #9067, opened 2016-10-07
Logs exception in DelegatingTaskDecorator

ahmetmircik:
Logging helps to see what went wrong, otherwise scheduled executor stops runnable silently.

According to [javaDocs:](http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleAtFixedRate%28java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit%29) `If any execution of the task encounters an exception, subsequent executions are suppressed`

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/DelegatingTaskDecorator.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/DelegatingTaskDecorator.java
@@ -16,8 +16,6 @@
 
 package com.hazelcast.spi.impl.executionservice.impl;
 
-import com.hazelcast.util.ExceptionUtil;
-
 import java.util.concurrent.Executor;
 
 /**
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/StoreWorker.java
```diff
@@ -262,4 +262,8 @@ private static int getWriteDelaySeconds(MapStoreContext mapStoreContext) {
         return mapStoreConfig.getWriteDelaySeconds();
     }
 
+    @Override
+    public String toString() {
+        return "StoreWorker{" + "mapName='" + mapName + "\'}";
```
Donnerbart: I think that backslash `\'` is superfluous and will be printed.

ahmetmircik: fixed

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/DelegatingTaskDecorator.java
```diff
@@ -45,6 +47,7 @@ public void run() {
         try {
             executor.execute(runnable);
         } catch (Throwable t) {
+            getLogger(getClass()).severe("Cannot execute " + runnable, t);
             ExceptionUtil.sneakyThrow(t);
```
pveentjer: So is this sneakyThrow going to remain or not?

Imho the catch should be removed completely. The 'executor' should be configured with an exception logger.

ahmetmircik:
handled in https://github.com/hazelcast/hazelcast/pull/9107

PR #9072, opened 2016-10-10
Fixed cloning constructor of RingbufferConfig

Donnerbart:
* Fixed cloning constructor of RingbufferConfig, which didn't copy the internal RingbufferStoreConfig field
* Increased the code coverage of RingbufferConfig

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
@@ -88,6 +88,9 @@ public class RingbufferConfig {
         this.asyncBackupCount = config.asyncBackupCount;
         this.timeToLiveSeconds = config.timeToLiveSeconds;
         this.inMemoryFormat = config.inMemoryFormat;
+        this.ringbufferStoreConfig = config.ringbufferStoreConfig == null
+                ? null
+                : new RingbufferStoreConfig(config.ringbufferStoreConfig);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/config/RingbufferConfig.java
```diff
@@ -88,6 +88,7 @@ public RingbufferConfig(RingbufferConfig config) {
         this.asyncBackupCount = config.asyncBackupCount;
         this.timeToLiveSeconds = config.timeToLiveSeconds;
         this.inMemoryFormat = config.inMemoryFormat;
+        this.ringbufferStoreConfig = config.ringbufferStoreConfig;
```
pveentjer: Should the ringbufferStoreConfig not be fully cloned like this:

```
 this.ringbufferStoreConfig = config.ringbufferStoreConfig
 == null? null: new RingbufferStoreConfig(config.ringbufferStoreConfig)
```

Donnerbart: That's a good one. Didn't think about it, since we normally just use that constructor to create the read-only configuration. But it's public, so we should make it a proper clone.

pveentjer:
run-lab-run

Donnerbart:
The problem with the fix it, that it breaks a test. I assume that the test was never working, due to the missing cloning. That clone constructor is used to create the read-only configuration, which we eventually use in the code. So that read-only config was always missing the `RingbufferStoreConfig`. This is now fixed, so we very well might discover hidden bugs with an enabled `RingbufferStore`.

In this case I get a `NPE` in `IdCheckerRingbufferStore.getLargestSequence()`, which is called during the init of the `RingbufferContainer`. The reason is that the `Long lastKey` is `null`, which throws the NPE when being casted to `long`.

I would kindly ask @mmedenjak to have a look how to fix this properly, after this PR is merged. I ignored the failing test so the PR builder is green now.

PR #9074, opened 2016-10-10
Master build fails due to extended Authentication message

ihsandemir:
Client protocol is extended to include the client and server versions exchange and since the hazelcast master depends on the client protocol snapshot the builds started to fail. This fix adds client and server version exchange during connection authentication.


ihsandemir:
The protocol compatbility tests (e.g. ClientCompatibilityTest) are auto generated tests.

hazelcast/src/main/java/com/hazelcast/instance/BuildInfo.java
```diff
@@ -59,6 +72,37 @@ public byte getSerializationVersion() {
...
+                return calculatedVersion;
+            } catch (Exception e) {
+                Logger.getLogger(BuildInfo.class).warning("Failed to calculate version using version string " + version, e);
+            }
+        }
+
+        return UNKNOWN_HAZELCAST_VERSION;
+    }
+
+    public static int getUnknownHazelcastVersion() {
```
Donnerbart: Do we need a getter for a public constant?

hazelcast/src/test/java/com/hazelcast/instance/BuildInfoProviderTest.java
```diff
@@ -74,6 +74,22 @@ public void testReadValues() {
...
+        assertEquals(-1, BuildInfo.calculateVersion("a.3.7.5"));
+        assertEquals(-1, BuildInfo.calculateVersion("3.a.5"));
+        assertEquals(-1, BuildInfo.calculateVersion("3,7.5"));
+        assertEquals(-1, BuildInfo.calculateVersion("3.7,5"));
+        assertEquals(30700, BuildInfo.calculateVersion("3.7-SNAPSHOT"));
+        assertEquals(30702, BuildInfo.calculateVersion("3.7.2"));
+        assertEquals(30702, BuildInfo.calculateVersion("3.7.2-SNAPSHOT"));
+        assertEquals(109902, BuildInfo.calculateVersion("10.99.2-SNAPSHOT"));
+        assertEquals(19930, BuildInfo.calculateVersion("1.99.30"));
+        assertEquals(109930, BuildInfo.calculateVersion("10.99.30-SNAPSHOT"));
```
Donnerbart: We should have a test for `RC` versions here as well, just to be sure the regex is working properly.

ihsandemir: OK, just added the test.

hazelcast/src/main/java/com/hazelcast/instance/BuildInfo.java
Donnerbart: Ah, that may be one of the visibility checks, okay. I can have a look at this once the PR is merged. This is surely no blocker :)

ihsandemir:
run-lab-run

asimarslan:
👍

PR #9076, opened 2016-10-10
Make write-behind task reschedule itself

ahmetmircik:
Otherwise executors queue can grow continuously and task can be suppressed due to the retry of failed store operations.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/StoreWorker.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/StoreWorker.java
@@ -28,6 +28,7 @@ import com.hazelcast.map.impl.recordstore.RecordStore;
 import com.hazelcast.nio.Address;
 import com.hazelcast.partition.InternalPartition;
 import com.hazelcast.partition.InternalPartitionService;
+import com.hazelcast.spi.ExecutionService;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.util.Clock;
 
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/StoreWorker.java
```diff
@@ -83,11 +82,31 @@ public StoreWorker(MapStoreContext mapStoreContext, WriteBehindProcessor writeBe
         NodeEngine nodeEngine = mapServiceContext.getNodeEngine();
         InternalPartitionService partitionService = nodeEngine.getPartitionService();
         this.partitionCount = partitionService.getPartitionCount();
+        this.executionService = nodeEngine.getExecutionService();
     }
 
 
+    public synchronized void start() {
+        stopped = false;
+        run();
```
jerrinot: shouldn't this just schedule an execution instead of running on the calling thread?

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/StoreWorker.java
```diff
@@ -71,6 +69,7 @@
      * @see #calculateHighestStoreTime
      */
     private long lastHighestStoreTime;
+    private boolean stopped;
```
jerrinot: this should be volatile. also something like `stopRequested` would be more precise.

ahmetmircik: no need to make it `volatile` it is under `synchronized` blocks.
changed the name.

Donnerbart: I prefer to call these variables `running`, because we mostly use them like `while (runnning) { ... }`. So we don't have to negate them all them time.

PR #9084, opened 2016-10-11
Make write-behind task reschedule itself

ahmetmircik:
forward port of #9076


ahmetmircik:
already reviewed here: #9076

Issue #115, opened 2012-04-06
Add support for ScheduledExecutorService


mlaccetti:
Would be delightful to be able to use a distributed ScheduledExecutorService that could guarantee execution of something across a cluster, on each node, etc, etc.

isaeedm:
+1
Is this feature going to be released with Hazelcast 3.5? if yes when will 3.5 be released?

crowdconnected:
+1

tazio:
+1, I need it

bwzhang2011:
@gurbuzali , how does this going on ? could it be integrated in 3.5 and also with support for cron expression.

bmcustodio:
@enesakar any news/progress on this?

pveentjer:
Will be added in 3.6.

bmcustodio:
@pveentjer is there no way for it to make to 3.5?

pveentjer:
I doubt it because 3.5 isn't that far away and we still have a lot of stuff to do.

tmazzaro:
@pveentjer we are evaluating using hazelcast vs other alternatives but we do need this feature. can you please let me know when this will be available in which version? many thanks tom

pveentjer:
It is on the radar for 3.6. We are already working on gathering the requirements for this particular feature.

tmazzaro:
Couldn't find in the website when 3.6 is planned for pls?

bwzhang2011:
@pveentjer, any update with the issue ? hope it could be provided in hz3.5 or else we should do more integration with quartz to make the schedule distributed.

bwzhang2011:
@pveentjer, any update with such issue ?

jamesmmchugh:
+1

sertugkaya:
+1

Petikoch:
+1

edoparo:
+1

logemann:
it was on the radar a year ago and now its dropped?

sertugkaya:
@logemann it has a milestone set for 3.7

marcolattarulo:
+1

olivergg:
+1

drshriveer:
+1

mikemainguy:
+1

danfink:
+1

JAC2703:
+1

wneild:
+1

tommagowan:
+1

cristcost:
Hi all,
[...]

PR #9087, opened 2016-10-11
Distributed scheduled executor service - Beta

tkountis:
Beta Distributed & durable implementation of j.u.c.ScheduledExecutor (fixing #115) provides similar API to the ScheduledExecutorService with some exceptions but also additional methods like scheduling tasks on a specific member, on a member who is owner of a specific key, executing a tasks on multiple members etc.

- Durable tasks that get re-scheduled on replica nodes in the event of node loss.
- Support for retrieving a task if the original reference was lost
- Support for PartitionAware tasks
- Support for Stateful jobs
- Support for Named jobs
- Support for managed context (HazecastInstanceAware)
- Relying on Durable executors to run tasks


bwzhang2011:
nice work hope it would be tested and verified and merged into master when 3.8 released.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorServiceProxy.java
```diff
@@ -0,0 +1,475 @@
...
+        return new RunnableAdapter<T>(command);
+    }
+
+    //TODO tkountis - Add support for PartitionAware tasks
+    private int getKeyPartitionId(Object key) {
+        return getNodeEngine().getPartitionService().getPartitionId(key);
+    }
+
+    private <V> IScheduledFuture<V> submitAsync(String taskName, Operation op) {
+        int partitionId = partitionRandom.nextInt(partitionCount);
```
tkountis: Thanks @gurbuzali, already noticed that and fixed in my dev. 👍

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorServiceProxy.java
```diff
@@ -0,0 +1,475 @@
...
+
+        ScheduledFutureProxy futureProxy = new ScheduledFutureProxy(
+                ScheduledTaskHandler.of(partitionId, getName(), taskName));
+        attachHazelcastInstance(futureProxy);
+        return futureProxy;
+    }
+
+    private <V> IScheduledFuture<V> submitOnMemberAsync(String taskName, Operation op, Member member) {
+        Address address = member.getAddress();
+        op.setPartitionId(-1); // Make partition agnostic
```
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/scheduleexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,304 @@
...
+                ((StatefulRunnable) original).saveState(state);
+                publishStateToReplicas();
+            }
+        }
+
+        private void publishStateToReplicas() {
+            OperationService operationService = nodeEngine.getOperationService();
+            IPartition partition = nodeEngine.getPartitionService().getPartition(partitionId);
+            for (int i = 1; i < getDurability() + 1; i++ ) {
+                Address address = partition.getReplicaAddress(i);
```
tkountis: Hmm, indeed, didn't think of that, good catch. Thanks. 👍

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
```diff
@@ -0,0 +1,144 @@
...
+import com.hazelcast.spi.PartitionReplicationEvent;
+import com.hazelcast.spi.RemoteService;
+import com.hazelcast.spi.partition.MigrationEndpoint;
+
+import java.util.Collections;
+import java.util.Properties;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class DistributedScheduledExecutorService
```
gurbuzali: there are 2 `DistributedScheduledExecutorService`. one is in `scheduledexecutor` other one is in `scheduleexecutor` package (mind the missing `d` in package-name)

with a second glance to the packages, looks like there was a renaming to the package but the old one lingers still

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,332 @@
...
+
+        while (taskEntries.hasNext()) {
+            Map.Entry<String, ScheduledTaskDescriptor> entry = taskEntries.next();
+            ScheduledTaskDescriptor descriptor = entry.getValue();
+            ScheduledFuture<?> future = descriptor.getScheduledFuture();
+
+            // Stop primary one, so we can safely capture latest state.
+            // However, may be incorrect state due to interruption.
+            while (!future.isCancelled() && !future.isDone()) {
+                future.cancel(true);
```
gurbuzali: we should not stop the primary one if we cannot resume it afterwards. this replication does not mean we are moving the task, it can happen anytime when a backup partition goes out of sync from the primary. 
so replication should not affect the running task on primary

hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private int durability = DEFAULT_DURABILITY;
+
+    private com.hazelcast.config.ScheduledExecutorConfig.ScheduledExecutorConfigReadOnly readOnly;
+
+    public ScheduledExecutorConfig() {
+    }
+
+    public ScheduledExecutorConfig(String name) {
+        this.name = name;
```
pveentjer: Do the arguments need to be checked? Most config have some form of arg checking to provide feedback to user when illegal configuration is made

hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Override
+    public String toString() {
+        return "ScheduledExecutorConfig{"
+                + "name='" + name + '\''
+                + ", durability=" + durability
+                + '}';
+    }
+
+    private static class ScheduledExecutorConfigReadOnly extends com.hazelcast.config.ScheduledExecutorConfig {
```
pveentjer: I guess checkstyle wants to have this final.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,395 @@
...
+ * exceptions but also additional methods like scheduling tasks on a specific member, on a member who is owner of a specific key,
+ * executing a tasks on multiple members etc.
+ *
+ *</p>
+ *
+ * Tasks (<tt>Runnable</tt> and/or <tt>Callable</tt>) scheduled on any partition through an <tt>IScheduledExecutorService</tt>,
+ * yield some durability characteristics.
+ * <ul>
+ *     <li>When a node goes down (up to <tt>durability</tt> config), the scheduled task will resume
+ *     on a replica node, by starting over.</li>
```
pveentjer: 'starting over' or 'restarting'

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
```diff
@@ -0,0 +1,149 @@
...
+            if (partitions[partitionId] != null) {
+                partitions[partitionId].destroy();
+            }
+            partitions[partitionId] = new ScheduledExecutorPartition(nodeEngine, partitionId);
+        }
+    }
+
+    @Override
+    public void shutdown(boolean terminate) {
+        reset();
```
gurbuzali: shutdown recreates the partitions, can we have a destroy only version here

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,426 @@
...
+
+    private final NodeEngine nodeEngine;
+
+    private final InternalExecutionService executionService;
+
+    private final int partitionId;
+
+    private final int durability;
+
+    private final Map<String, ScheduledTaskDescriptor> tasks;
```
gurbuzali: This map can be accessed(read) out of partitionOperationThread in `GetAllScheduledOperation`. It should be a ConcurrentMap

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,426 @@
...
+
+    private final Map<String, BackupTaskDescriptor> backups;
```
gurbuzali: since we have a `member-partition-container` it means it is accessed from generic threads. and this means we need to have ConcurrentMap over here too.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,426 @@
...
+        this.nodeEngine = nodeEngine;
+        this.executionService = (InternalExecutionService) nodeEngine.getExecutionService();
+        this.partitionId = partitionId;
+        this.durability = durability;
+        this.tasks = new HashMap<String, ScheduledTaskDescriptor>();
+        this.backups = backups;
+    }
+
+    public <V> ScheduledFuture<V> schedule(TaskDefinition definition) {
+        checkNotDuplicateTask(definition.getName());
```
gurbuzali: Looks like this `member-partition` will cause more problem. Imagine 2 clients try to schedule a named task on a specific member at the same time. They both send `ScheduleTaskOperation`, they run on generic threads in parallel. They do the name checking at the same time and pass.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/DuplicateTaskException.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.scheduledexecutor;
+
+/**
+ * An exception thrown when a task's name is already used before for another (or the same, if re-attempted) schedule.
+ * Tasks under a scheduler must have unique names.
+ */
+public class DuplicateTaskException
```
pveentjer: This should probably extend from HazelcastException or perhaps IllegalArgumentException?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,395 @@
...
+     *
+     * @param command the task to execute
+     * @param delay the time from now to delay execution
+     * @param unit the time unit of the delay parameter
+     * @return a ScheduledFuture representing pending completion of
+     *         the task and whose {@code get()} method will return
+     *         {@code null} upon completion
+     * @throws RejectedExecutionException if the task cannot be
+     *         scheduled for execution
+     * @throws NullPointerException if command is null
```
pveentjer: what happens in case of 0 or negative timeout?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,395 @@
...
+     * @param period the period between successive executions
+     * @param unit the time unit of the initialDelay and period parameters
+     * @return a ScheduledFuture representing pending completion of
+     *         the task, and whose {@code get()} method will throw an
+     *         exception upon cancellation
+     * @throws RejectedExecutionException if the task cannot be
+     *         scheduled for execution
+     * @throws NullPointerException if command is null
+     */
+    IScheduledFuture<?> scheduleWithRepetition(Runnable command, long initialDelay,
```
pveentjer: Why not use scheduleAtFixedRate or scheduleWithFixedDelay. These are the methods from the ScheduledExecutor

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,395 @@
...
+     *         exception upon cancellation
+     * @throws RejectedExecutionException if the task cannot be
+     *         scheduled for execution
+     * @throws NullPointerException if command is null
+     */
+    IScheduledFuture<?> scheduleOnMemberWithRepetition(Runnable command, Member member,
+                                                       long initialDelay, long period, TimeUnit unit);
+
+    /**
+     * Creates and executes a one-shot action that becomes enabled
```
pveentjer: What happens when the partition is migrated before the task is started, or during task execution?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,395 @@
...
+     *         scheduled for execution
+     * @throws NullPointerException if command is null
+     */
+    Map<Member, IScheduledFuture<?>> scheduleOnAllMembersWithRepetition(Runnable command, long initialDelay,
+                                                                        long period, TimeUnit unit);
+
+    /**
+     * Creates and executes a one-shot action that becomes enabled
+     * after the given delay on all {@link Member}s given.
+     *
```
pveentjer: what happens when new members are added or existing members are dropped?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,395 @@
...
+    /**
+     * Creates a new {@link IScheduledFuture} from the given handler.
+     * This is useful in case your member node or client from which the original
+     * scheduling happened, went down, and now you want to access the <tt>ScheduledFuture</tt> again.
+     *
+     * @param handler The handler of the task as found from {@link IScheduledFuture#getHandler()}
+     * @param <V> The return type of callable tasks
+     * @return A new {@link IScheduledFuture} from the given handler.
+     */
+    <V> IScheduledFuture<V> getScheduled(ScheduledTaskHandler handler);
```
pveentjer: Flavor: perhaps call this 'getScheduledFuture'? getScheduled doesn't say anything to me without looking at the documentation.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/ScheduledTaskStatistics.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Returns how many times the task was ran/called.
+     *
+     * @return The numbers of runs
+     */
+    long getTotalRuns();
+
+    /**
+     * @return The monotonic creation time the task.
```
pveentjer: What unit is this?
Probably good to refer to System.currentMillis 'monotonic creation time' is not revealing to me. Perhaps I'm too stupid :)

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/ScheduledTaskStatistics.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Returns the duration of the task's last execution.
+     *
+     * @param unit The time unit to return the duration in.
+     * @return The total duration of the task's last execution.
+     */
+    long getLastRunDuration(TimeUnit unit);
+
+    /**
+     * Returns the last period of time the task was idle, waiting to get scheduled in.
```
pveentjer: I believe you can remove the last 'in.'

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/ScheduledTaskStatistics.java
```diff
@@ -0,0 +1,89 @@
...
+     */
+    long getLastRunEnd();
+
+    /**
+     * Returns the duration of the task's last execution.
+     *
+     * @param unit The time unit to return the duration in.
+     * @return The total duration of the task's last execution.
+     */
+    long getLastRunDuration(TimeUnit unit);
```
pveentjer: Why does this method have a unit as parameter and the others not. I'm fine with both approaches; but currently it isn't consistent.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/StatefulTask.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * An interface to provide means for saving & loading state for {@link Runnable} and {@link java.util.concurrent.Callable}
+ * tasks scheduled with an {@link IScheduledExecutorService}. When task implements this interface, the Scheduled Executor
+ * will be able to handle state of the task among the replicas in an event of Migration or Node failure.
+ *
+ * Example:
+ * <pre>
+ * public class CleanUpTask implements Runnable, StatefulTask<String, Integer>, HazelcastInstanceAware {
+ *
+ *      transient HazelcastInstance instance;
```
pveentjer: I would make this private to prevent people thinking they need to make it package friendly.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/TaskHelper.java
```diff
@@ -0,0 +1,55 @@
...
+import com.hazelcast.scheduledexecutor.impl.NamedTaskDecorator;
+import com.hazelcast.spi.annotation.Beta;
+
+import java.util.concurrent.Callable;
+
+/**
+ * A helper class with utilities to act upon {@link Runnable} and/or {@link Callable} tasks.
+ */
+@Beta
+public final class TaskHelper {
```
pveentjer: Why don't we call this class TaskUtils? To be in line with the other utility classes.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/AmendableScheduledTaskStatistics.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.scheduledexecutor.impl;
+
+import com.hazelcast.scheduledexecutor.ScheduledTaskStatistics;
+
+/**
+ * Created by Thomas Kountis.
```
pveentjer: We can see in the git history who created this class; no need to add your name.

Can you add a line of comment what the purpose of this class is? Amendable.... doesn't say anything to me.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public static final String SERVICE_NAME = "hz:impl:scheduledExecutorService";
+
+    public static final int MEMBER_PARTITION = -1;
+
+    private NodeEngine nodeEngine;
+
+    private ScheduledExecutorPartition[] partitions;
+
+    private ScheduledExecutorPartition memberPartition;
```
tkountis: When scheduling on Members directly, rather than on a KeyOwner/Partition.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/ScheduledTaskStatistics.java
tkountis: The methods having a unit as an arg return duration back, so you can receive it in the form of your preference. The ones without an argument, return a single point in time.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,395 @@
...
+    /**
+     * Initiates an orderly shutdown in which previously submitted
+     * tasks are executed, but no new tasks will be accepted.
+     * Invocation has no additional effect if already shut down.
+     *
+     * <p>This method does not wait for previously submitted tasks to
+     * complete execution.
+     */
+    void shutdown();
+
```
tkountis: Indeed will do, just had to take this out of this release, time constraints.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
tkountis: Added further doc.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
tkountis: Same behaviour to j.u.c.ScheduledExecutorService -> delay < 0 = 0

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
tkountis: There is class Javadoc for the difference

> The difference is that this service will skip a scheduled execution if another thread is still running the same task, instead of postponing its execution.
> To emphasize this difference the method is called <code>scheduleWithRepetition</code>

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,456 @@
...
+        this.executionService = (InternalExecutionService) nodeEngine.getExecutionService();
+        this.partitionId = partitionId;
+        this.durability = durability;
+        this.tasks = new ConcurrentHashMap<String, ScheduledTaskDescriptor>();
+        this.backups = backups;
+    }
+
+    public <V> ScheduledFuture<V> schedule(TaskDefinition definition) {
+        try {
+            acquireMemberPartitionLockIfNeeded();
```
gurbuzali: I'm ok with this solution but I think creating a separate class for member-container would be better. We would decouple partition and member logic. 
getting rid off these `if else` blocks would be the extra, we had them all over the class
```
if member-partition 
...
else
...
```

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorDataSerializerHook.java
```diff
@@ -0,0 +1,136 @@
...
+                    case GET_ALL_SCHEDULED:
+                        return new GetAllScheduledOperation();
+                    case GET_RESULT:
+                        return new GetResultOperation();
+                    case PUBLISH_RESULT:
+                        return new ResultReadyNotifyOperation();
+                    case SHUTDOWN:
+                        return new ShutdownOperation();
+                    default:
+                        return null;
```
gurbuzali: throwing exception if `typeId` is not known would be better, `IllegalArgumentException` maybe ?

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/ExecutionServiceImpl.java
```diff
@@ -243,6 +248,16 @@ public void executeDurable(String name, Runnable command) {
     }
 
     @Override
+    public ScheduledFuture<?> scheduleDurable(String name, Runnable command, long delay, TimeUnit unit) {
+        return getDurableTaskScheduler(name).schedule(command, delay, unit);
```
gurbuzali: So we are delegating the tasks to durable-executor. Imagine this scenario:
user obtains durable-executor and scheduled-executor with the same name
```
instance.getDurableExecutorService("foo");
instance.getScheduledExecutorService("foo");
```
and uses both of them accordingly. at some time s/he decides to shutdown the scheduled-executor, which will shutdown the underlying durable-executor.

Won't this be a problem ? 

As a solution, we can add some prefix to the name like `hz:impl:scheduledExecutorService` while obtaining the durable-executor.

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceTest.java
```diff
@@ -0,0 +1,618 @@
...
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class ScheduledExecutorServiceTest extends HazelcastTestSupport {
+
+    private TestHazelcastInstanceFactory factory;
+
+    @After
+    public void teardown() {
+        if (factory != null) {
```
gurbuzali: you don't have to terminate instances, it is handled in `HazelcastTestSupport` class. Check `shutdownNodeFactory` method.

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceTest.java
```diff
@@ -0,0 +1,618 @@
...
+        assertEquals(members.size(), allScheduled.size());
+
+        for (Member member : members) {
+            assertEquals(1, allScheduled.get(member).size());
+            assertEquals(25.0, allScheduled.get(member).get(0).get(), 0);
+        }
+    }
+
+    public HazelcastInstance[] createClusterWithCount(int count) {
+        HazelcastInstance[] instances = new HazelcastInstance[count];
```
gurbuzali: you don't have to create instance array explicitly, check `factory.newInstances()` method

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceTest.java
```diff
@@ -0,0 +1,618 @@
...
+
+    static class ICountdownLatchRunnableTask implements Runnable, Serializable, HazelcastInstanceAware {
+
+        final String latchName;
+
+        transient HazelcastInstance instance;
+
+        ICountdownLatchRunnableTask(String latchName, HazelcastInstance instance) {
+            this.latchName = latchName;
+            this.instance = instance;
```
gurbuzali: you shouldn't pass instance, it will be set via `setHazelcastInstance`

tkountis: This is not the case for local-member invocation, it is when its a remote call, and the OP has to get deserialized, but not when local

hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java
```diff
@@ -0,0 +1,126 @@
...
+
+import static com.hazelcast.util.Preconditions.checkNotNegative;
+
+/**
+ * Configuration options for the {@link com.hazelcast.scheduledexecutor.IScheduledExecutorService}
+ */
+public class ScheduledExecutorConfig {
+
+    /**
+     * Scheduled of Executor
```
mmedenjak: Fix javadoc

hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java
```diff
@@ -0,0 +1,126 @@
...
+    public ScheduledExecutorConfig(String name, int durability) {
+        this.name = name;
+        this.durability = durability;
+    }
+
+    public ScheduledExecutorConfig(com.hazelcast.config.ScheduledExecutorConfig config) {
+        this(config.getName(), config.getDurability());
+    }
+
+    public com.hazelcast.config.ScheduledExecutorConfig.ScheduledExecutorConfigReadOnly getAsReadOnly() {
```
mmedenjak: You can remove all `com.hazelcast.config.` like these in the file

hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java
```diff
@@ -0,0 +1,126 @@
...
+     * Gets the durability of the executor
+     *
+     * @return the durability of the executor
+     */
+    public int getDurability() {
+        return durability;
+    }
+
+    /**
+     * Sets the durability of the executor
```
mmedenjak: Can you add some short description of what is durability? I know that `DurableExecutorConfig#setDurability` doesn't describe it but some do, e.g. `TransactionOptions#setDurability`

hazelcast/src/main/java/com/hazelcast/core/HazelcastInstance.java
```diff
@@ -430,6 +431,21 @@
     CardinalityEstimator getCardinalityEstimator(String name);
 
     /**
+     * Obtain the {@link IScheduledExecutorService} that provides
```
mmedenjak: Fix javadoc

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/ReplicationOperation.java
```diff
@@ -0,0 +1,89 @@
...
+import com.hazelcast.scheduledexecutor.impl.ScheduledExecutorPartition;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ReplicationOperation
+        extends AbstractSchedulerOperation {
+
+    private Map<String, Map<String, BackupTaskDescriptor>> map;
```
tkountis: Talked offline, and agreed that this is not needed, since its only used for Partition Owned tasks, there is no backups for members, so this is never accessed outside the Partition thread.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/ReplicationOperation.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    protected void readInternal(ObjectDataInput in) throws IOException {
+        int size = in.readInt();
+        map = new HashMap<String, Map<String, BackupTaskDescriptor>>(size);
+        for (int i = 0; i < size; i++) {
+            String key = in.readUTF();
+            int subSize = in.readInt();
+            Map<String, BackupTaskDescriptor> subMap = new HashMap<String, BackupTaskDescriptor>(subSize);
```
tkountis: Same here.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
tkountis: Agreed, will change.

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceTest.java
tkountis: 👍

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceTest.java
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/core/HazelcastInstance.java
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/BackupTaskDescriptor.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private Map masterState;
+
+    public BackupTaskDescriptor() {
+    }
+
+    public BackupTaskDescriptor(TaskDefinition definition) {
+        this.definition = definition;
+        this.masterStats = new ScheduledTaskStatisticsImpl();
+        this.masterState = new ConcurrentHashMap();
```
mmedenjak: Does this need to be a `ConcurrentHashMap`? Is the mutation from different threads? This should then be made clear by changing the parameter and field types. For instance, in `SyncStateOperation#readInternal` a regular `HashMap` is instantiated on deserialization.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/BackupTaskDescriptor.java
```diff
@@ -0,0 +1,89 @@
...
+
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class BackupTaskDescriptor implements IdentifiedDataSerializable {
```
mmedenjak: Maybe some javadoc for future developers? It can be only class-level if the class is internal.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
```diff
@@ -0,0 +1,157 @@
...
+import com.hazelcast.spi.PartitionReplicationEvent;
+import com.hazelcast.spi.RemoteService;
+import com.hazelcast.spi.impl.executionservice.InternalExecutionService;
+import com.hazelcast.spi.partition.MigrationEndpoint;
+
+import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class DistributedScheduledExecutorService
```
mmedenjak: Same as BackupTaskDescriptor, perhaps some class-level javadoc, especially since the class sounds like it is an important part.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorPartition.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import static com.hazelcast.util.ConcurrencyUtil.getOrPutIfAbsent;
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
+public class ScheduledExecutorPartition {
```
mmedenjak: Javadoc since it is not a trivial class

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Override
+    public void shutdown(boolean terminate) {
+        shutdownExecutors.clear();
+
+        if (memberPartition != null) {
+            memberPartition.destroy();
+        }
+
+        for (int partitionId = 0; partitionId < partitions.length; partitionId++) {
```
mmedenjak: Can replace with for-each

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorServiceProxy.java
```diff
@@ -0,0 +1,359 @@
...
+    @Override
+    public void shutdown() {
+        NodeEngine nodeEngine = getNodeEngine();
+        Collection<Member> members = nodeEngine.getClusterService().getMembers();
+        OperationService operationService = nodeEngine.getOperationService();
+        Collection<Future> calls = new LinkedList<Future>();
+
+        for (Member member : members) {
+            if (member.localMember()) {
+                getService().shutdownExecutor(name);
```
mmedenjak: If you want to add the current member shutdown to the Future list, you can safely call the same `operationService.invokeOnTarget` method since it will internally check if you are invoking on the local member : com/hazelcast/spi/impl/operationservice/impl/Invocation.java:510

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorPartition.java
```diff
@@ -0,0 +1,128 @@
...
+    void disposeObsoleteReplicas(int thresholdReplicaIndex) {
+        if (thresholdReplicaIndex < 0) {
+            containers.clear();
+        }
+
+        Iterator<ScheduledExecutorContainer> iterator = containers.values().iterator();
+        while (iterator.hasNext()) {
+            ScheduledExecutorContainer container = iterator.next();
+            if (thresholdReplicaIndex > container.getDurability()) {
+                iterator.remove();
```
mmedenjak: Do you need to dispose of some resources in ScheduledExecutorContainer here?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Override
+    public DistributedObject createDistributedObject(String name) {
+        return new ScheduledExecutorServiceProxy(name, nodeEngine, this);
+    }
+
+    @Override
+    public void destroyDistributedObject(String name) {
+        if (shutdownExecutors.remove(name) != null) {
```
mmedenjak: So `InternalExecutionService#shutdownDurableExecutor` is called only if previously `DistributedScheduledExecutorService#shutdownExecutor` has been called?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,416 @@
...
+     */
+    <V> Map<Member, List<IScheduledFuture<V>>> getAllScheduledFutures();
+
+    /**
+     * Initiates an orderly shutdown in which previously submitted
+     * tasks are executed, but no new tasks will be accepted.
+     * Invocation has no additional effect if already shut down.
+     *
+     * <p>This method does not wait for previously submitted tasks to
+     * complete execution.
```
tkountis: Indeed, similar to j.u.c.Executor this is an async shutdown, work keeps going on, and this doesn't block until its done. I could modify the comment to make it more clear, but I think keeping it same to j.u.c makes it more trivial in terms of consistency.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,456 @@
...
+            Map.Entry<String, ScheduledTaskDescriptor> entry = iterator.next();
+            ScheduledTaskDescriptor descriptor = entry.getValue();
+
+            try {
+                BackupTaskDescriptor replica = new BackupTaskDescriptor(descriptor.getDefinition());
+                replica.setMasterState(descriptor.getState().get());
+                replica.setMasterStats(descriptor.getStats());
+                replicas.put(entry.getKey(), replica);
+            } finally {
+                descriptor.getLock().set(false);
```
tkountis: Yes, single place.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/DistributedScheduledExecutorService.java
tkountis: Well as per idea behind the design yes :), but in practise no, the Durable Executor service can shutdown the executor too, since I am delegating my work to the same executors, therefore as @gurbuzali suggested this needs fix.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
mmedenjak: You mean only in this PR? But the lock will be set in other places?

tkountis:
@mmedenjak regarding the lock, this is a leftover, i just removed. Good catch, I will address the other comments too

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/ShutdownOperation.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public int getId() {
+        return ScheduledExecutorDataSerializerHook.SHUTDOWN;
+    }
+
+    @Override
+    protected void writeInternal(ObjectDataOutput out)
+            throws IOException {
+        super.writeInternal(out);
```
mmedenjak: No need for these overrides since they just delegate.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/GetAllScheduledOperation.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public int getId() {
+        return ScheduledExecutorDataSerializerHook.GET_ALL_SCHEDULED;
+    }
+
+    @Override
+    protected void writeInternal(ObjectDataOutput out)
+            throws IOException {
+        super.writeInternal(out);
```
mmedenjak: NOOP

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/SyncStateOperation.java
```diff
@@ -0,0 +1,80 @@
...
+        return ScheduledExecutorDataSerializerHook.SYNC_STATE_OP;
+    }
+
+    @Override
+    protected void writeInternal(ObjectDataOutput out)
+            throws IOException {
+        super.writeInternal(out);
+        out.writeUTF(taskName);
+        out.writeInt(state.size());
+        Iterator<Map.Entry> it = state.entrySet().iterator();
```
mmedenjak: For-each

hazelcast/src/main/java/com/hazelcast/spi/TaskScheduler.java
```diff
@@ -54,6 +55,8 @@
      */
     ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit);
 
+    <V> ScheduledFuture<V> schedule(Callable<V> command, long delay, TimeUnit unit);
```
mmedenjak: Add javadoc

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/ExecutionServiceImpl.java
```diff
@@ -190,7 +190,9 @@ private ManagedExecutorService createExecutor(String name, int poolSize, int que
             pool.allowCoreThreadTimeOut(true);
             executor = pool;
         } else {
-            throw new IllegalArgumentException("Unknown executor type: " + type);
+            throw new IllegalArgumentException("Unknown e"
```
mmedenjak: Fix string concatenation

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/ExecutionServiceImpl.java
```diff
@@ -200,6 +202,10 @@ public ManagedExecutorService getExecutor(String name) {
         return ConcurrencyUtil.getOrPutIfAbsent(executors, name, constructor);
     }
 
+    public ManagedExecutorService getDurable(String name) {
```
mmedenjak: Add @Override

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/ExecutionServiceImpl.java
```diff
@@ -263,6 +284,10 @@ public TaskScheduler getTaskScheduler(String name) {
         return new DelegatingTaskScheduler(scheduledExecutorService, getExecutor(name));
     }
 
+    public TaskScheduler getDurableTaskScheduler(String name) {
```
mmedenjak: Can be private

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
tkountis: This was a left-over. It is now changed to a different approach as discussed with @metanet

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorPartition.java
tkountis: Yes, caught that, good finding.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorMemberBin.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import static com.hazelcast.util.ConcurrencyUtil.getOrPutIfAbsent;
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
+public class ScheduledExecutorMemberBin implements ScheduledExecutorContainerHolder {
```
tkountis: The duplication is minimal, for the shake of not introducing yet another class, I will keep it as is. If this grows in the future, we can abstract it then. WDYT ?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/ScheduledTaskStatistics.java
```diff
@@ -0,0 +1,91 @@
...
+     * @return The numbers of runs
+     */
+    long getTotalRuns();
+
+    /**
+     * @return The creation time the task
+     *
+     * @see {@link System#nanoTime()}.
+     */
+    long getCreatedAt();
```
ahmetmircik: seems its more readable if method name contains returned time-unit something like this: `getCreatedAtNanos`

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
```diff
@@ -0,0 +1,408 @@
...
+
+    private final InternalExecutionService executionService;
+
+    private final int partitionId;
+
+    private final int durability;
+
+    private final ConcurrentMap<String, ScheduledTaskDescriptor> tasks;
+
+    private final Map<String, BackupTaskDescriptor> backups;
```
ahmetmircik: Here one map(`tasks `) is concurrent and other one(`backups`) is not. Is this ok? Can `backups` be accessed concurrently?

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/InternalExecutionService.java
```diff
@@ -18,14 +18,27 @@
...
 
+    ExecutorService getDurable(String name, boolean scheduled);
 
     void executeDurable(String name, Runnable command);
 
+    ScheduledFuture<?> scheduleDurable(String name, Runnable command, long delay, TimeUnit unit);
+
+    <V> ScheduledFuture<V> scheduleDurable(String name, Callable<V> command, long delay, TimeUnit unit);
+
+    ScheduledFuture<?> scheduleDurableWithRepetition(String name, Runnable command,
```
ahmetmircik: Seems `scheduleDurableWithRepetition` may cause some problems. For example: imagine  i have a task which is scheduled to run in every 1 second and my thread-pool for scheduler has only 1 thread. 
But my task run takes longer than 1 second say 10 seconds. In this case, even my task is not finished, scheduler will continue to add new tasks to task queue. Eventually in time whole queue will be occupied and there will be no place for new tasks.

An example of queue size changes after 10 second periods will be something like this:
1 (1)
1 + 10 (11)
1 + 9 + 10 (20)
1 + 8 + 10 + 10 (29)
...

Will this be a problem for the end user WDYT?

tkountis: Yes, this is a problem in the case of scheduling so frequent tasks, if we have more than one workers its no as much of a problem then since we skip execution if the task is running so we can drain the queue fast enough. But to be honest, if someone schedules a task to run so frequently that takes that much to complete, the queue size will be the least of his worries. However, I do see the point, so if we notice this in production we can address later.  WDYT ?

ahmetmircik: AFAIK default queue capacity is integer max for now and it is not configurable for this scheduled executor. So at least memory leaks can be seen if there is a slow db interaction for a task for example. Not sure what to do here. First thing comes, why do we have this `scheduleDurableWithRepetition` method instead of regular `scheduleAtFixedRate` or `scheduleWithFixedDelay` as in `java.util.concurrent.ScheduledExecutorService`? Maybe we can completely remove it or fix its implementation because this doesn't provide similar expected behaviour with `java.util.concurrent.ScheduledExecutorService` from this very perspective.

tkountis: Regarding the reasons behind the method selection, is that the implementation was already supported so it was much more trivial to rely on that. I am afraid, that changing either at this stage is a much bigger change.  One trivial fix which I don't know if it will have any side effects, would be to change the order of the decorators before scheduling the task, ie. 
`Runnable decoratedTask = new SkipOnConcurrentExecutionDecorator(new DelegatingTaskDecorator(command, executor));` thus avoiding further queuing of the task if it is still running.

@jerrinot any opinions ?

jerrinot: This is the contract of `scheduleWithRepetition()`: 
`If any execution of this task takes longer than its period, then subsequent execution will be skipped.`

This `skip-on-overlap` strategy is the very reason why it has such unusual name: To indicate it has a different contract from `scheduleAtFixedRate` or `scheduleWithFixedDelay`.

ahmetmircik: Do you also think it can cause memory leaks?

tkountis: Yes, it can, if you schedule sth to run so frequently.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,416 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Distributed & durable implementation of {@link ScheduledExecutorService}.
```
ahmetmircik: > Distributed & durable implementation of {@link ScheduledExecutorService}.

`IScheduledExecutorService ` is not extending `ScheduledExecutorService` hence this is not an implementation of it. Doc may be a left over?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/IScheduledExecutorService.java
```diff
@@ -0,0 +1,416 @@
...
+ * guarantees a task won't be executed by multiple threads concurrently. The difference is that this service will
+ * skip a scheduled execution if another thread is still running the same task, instead of postponing its execution.
+ * To emphasize this difference the method is called <code>scheduleAtFixedRate</code>
+ * instead of <code>scheduleAtFixedRate</code>
+ * <br/><br/>
+ * The other difference is this service does not offer an equivalent of
+ * {@link java.util.concurrent.ScheduledExecutorService#scheduleWithFixedDelay(Runnable, long, long, TimeUnit)}
+ */
+@Beta
+public interface IScheduledExecutorService extends DistributedObject {
```
ahmetmircik: what was the reason that we didn't extend `ScheduledExecutorService`? In other distributed objects if there is an interface available we are extending it, putting some examples here:

`public interface ILock extends Lock, DistributedObject`
`public interface IMap<K, V> extends ConcurrentMap`
`public interface ICondition extends Condition `
`public interface ICollection<E> extends Collection<E>, DistributedObject`

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/BackupTaskDescriptor.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private TaskDefinition definition;
+
+    private ScheduledTaskStatisticsImpl masterStats;
+
+    private Map masterState;
```
tkountis: I believe these to be old review comments.
BackupTaskDescriptor is not used anymore.

hazelcast/src/main/java/com/hazelcast/spi/impl/executionservice/impl/DelegateAndSkipOnConcurrentExecutionDecorator.java
```diff
@@ -16,39 +16,61 @@
...
+        DelegateDecorator(Runnable runnable) {
+            this.runnable = runnable;
+        }
+
+        @Override
+        public void run() {
+            try {
+                runnable.run();
+            } finally {
+                isAlreadyRunning.set(false);
```
ahmetmircik: 👍

PR #9088, opened 2016-10-11
Cache recordsStores should be cleaned up during service shutdown

mdogan:
Otherwise during a force-start during hot-restart, these recordstores
will leak closed hot-restart store refs.


metanet:
:+1:

PR #9089, opened 2016-10-11
Cache recordsStores should be cleaned up during service shutdown

mdogan:
Otherwise during a force-start during hot-restart, these recordstores
will leak closed hot-restart store refs.

Backport of https://github.com/hazelcast/hazelcast/pull/9088


mdogan:
Original PR is merged.

PR #9092, opened 2016-10-11
Backport PR #9066 and PR #9145 - 3.6 client does not work with 3.7.x server when using Cache

ihsandemir:
Cache compatibility fix

Backport of PR #9066 and PR #9145


ihsandemir:
verify

PR #9093, opened 2016-10-11
Increased code coverage of QueryCacheConfig

Donnerbart:
I could not find any reason for `equals()` and `hashCode()` methods for config classes. They are just adding a massive block of hard to test code. They also are error prone, since they are easily forgotten whenever a new field is added. If they are just needed for tests it's always better to do a manual equality check in the test code instead of adding those methods to the production code.


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
verify

PR #9102, opened 2016-10-12
Document classes used as Data in Hazelcast client-member protocol

vbekiaris:
Annotates opaque classes used in client-server communication with `@ClientProtocol`.

Currently this only serves as documentation, however will also enable compatibility tests automation.


hazelcast/src/main/java/com/hazelcast/core/IFunction.java
```diff
@@ -27,6 +29,8 @@
  * @param <R> a result
  * @since 3.2
  */
+
+@ClientProtocol
 public interface IFunction<T, R> extends Serializable {
```
tombujok: I was wondering whether we should annotate interfaces, I'm not sure myself.
Maybe it's good to do so, since the user will know that they should not chance their implementations.
On the other hand the interface does not have any serialization logic. 
But still it's probably better to annotate them.

tombujok:
👍

tombujok:
@vbekiaris @jerrinot @Donnerbart I'm still thinking about the name. 
`@ClientProtocol` means to me that it's a class that constitutes the the client protocol but that's misleading. On the other hand I could not come up with a better name either.
The exact semantics of that annotation is that the class annotated with it is `referenced by the client`. 
Maybe `@ClientReferenced` or `@ClientDependency`?

The other reasoning comes even further, since it's important what it means that it is `referenced by the client`. It means that this class should not change when it comes to its binary format, so maybe `@BinaryInterface` or `@BinaryDependency` ?

Feedback appreciated.

tombujok:
run-lab-run

Donnerbart:
I'm not very good at naming, but I see your point. I still like to refer the client protocol, since we also have the "Member protocol" as reason for binary compatibility. Maybe `@ClientProtocolDependency` is a good hint? Independent of the annotation name we still have the annotation's JavaDoc which elaborates its meaning.

jerrinot:
`@ClientDependency` sounds good to me.

tombujok:
It's not a part of the protocol. It's the part of the client data send using the protocol. it's oblivious to the protocol what's being sent.

vbekiaris:
I think that `BinaryInterface` describes quite accurately what we do care about: changing the binary representation of the annotated class might result in compatibility issues. There are also classes which for one reason or the other cannot be converted (easily) to `IdentifiedDataSerializable` while not necessarily being part of client-member communication. So I would go for `BinaryInterface`.
Conversion is still in progress, so classes are still being annotated and added to this PR. Tests will be also included in the final PR to ensure:
* `DataSerializable`s are either `@ClientProtocol` (or whatever the final name is) or `IdentifiedDataSerializable`s, except for classes in public packages which are not `@PrivateApi`
* `IdentifiedDataSerializable`'s {factoryId, id} are unique.

tombujok:
@vbekiaris don't know for which name you finaly went, but @BinaryInterface looks good indeed. Feel free to pick any you like ;)

vbekiaris:
Thanks everyone, I finally went with `@BinaryInterface`.

The PR is now organised in two commits. First one includes annotating classes with `@BinaryInterface` and fixing `IdentifiedDataSerializable` issues (eg missing constructor functions). The second commit adds tests for `DataSerializable` conventions (which actually helped locating the issues fixed in this PR). @tombujok & @Donnerbart could you please have another look?

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueContainer.java
```diff
@@ -64,8 +64,8 @@
     private static final int ID_PROMOTION_OFFSET = 100000;
     private final Map<Long, TxQueueItem> txMap = new HashMap<Long, TxQueueItem>();
     private final Map<Long, Data> dataMap = new HashMap<Long, Data>();
-    private final QueueWaitNotifyKey pollWaitNotifyKey;
-    private final QueueWaitNotifyKey offerWaitNotifyKey;
+    private QueueWaitNotifyKey pollWaitNotifyKey;
```
vbekiaris: It's not possible to keep those fields as `final`, `name` is not yet known and we wouldn't be able to create those wait/notify keys at deserialization time in `readData`, based on the `name` that is deserialized.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/operations/LocalLockCleanupOperation.java
```diff
@@ -35,6 +35,14 @@
 
     private final String uuid;
 
+    /**
+     * This constructor should not be used to obtain an instance of this class; it exists to fulfill IdentifiedDataSerializable
+     * coding conventions.
+     */
+    public LocalLockCleanupOperation() {
+        uuid = "";
```
Donnerbart: Can't we use a pattern like
```
    public LocalLockCleanupOperation() {
        throw new UnsupportedOperationException("LocalLockCleanupOperation is local only.");
    }
```
for these classes? Your test could accept this as a valid implementation and there is no risk that the constructor is accidentally used (at least not in a test). The only issue could be that you won't be able to construct an instance to call `getId()`, right?

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/QueueContainer.java
Donnerbart: Ah, forget it. I didn't see the variable, just the "poll" and "offer". Sure, it's done properly your way.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/operations/LocalLockCleanupOperation.java
vbekiaris: You're right, what we lose with throwing `UnsupportedOperationException` in the constructor is the opportunity to verify that the factoryId-id combination is either throwing `UnsupportedOperationException` or unique. Maybe the utility of failing fast and hard instead of relying on reading the javadoc in a not-supposed-to-be-used constructor is greater than being able to instantiate every `IdentifiedDataSerializable` class. @tombujok wdyt?

vbekiaris:
@ihsandemir thanks for the review! All interfaces you mention above apart from `ExpiryPolicy` which comes from external JSR107 dependency are already annotated (including `CombinerFactory` which I suppose is the last one, instead of `ReducerFactory`). When an interface or class is annotated, test code considers the annotation present in all its implementations/subclasses so in terms of tests this won't make a difference. Still, I agree it would be a good idea to have the annotation visible so here's a summary of what I changed:
* added the annotation to `HazelcastExpiryPolicy` which is the sole implementation of `ExpiryPolicy` in our production codebase.
* added the annotation to `UsernamePasswordCredentials` & `AbstractCredentials`
* added the annotation to `Predicate` implementations.
* `AbstractEntryProcessor` is now annotated
* `ReducerFactory` & `CombinerFactory` implementations are now annotated (lots of them)
* a few more
I committed the above in a separate commit so they are distinguishable from the rest of changes.

vbekiaris:
run-lab-run

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * Tests for classes which are used as Data in the client protocol
+ */
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class})
+public class DataSerializableConventionsTest {
+
+    // verify that any class which is DataSerializable and is not annotated with @BinaryInterface
+    // is also an IdentifiedDataSerializable
+    // ignored until the conversion to IdentifiedDataSerializable is done
```
ihsandemir: how is it ignored?

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
```diff
@@ -0,0 +1,222 @@
...
+                        instanceFromFactory.getClass().equals(instance.getClass()));
+            } catch (UnsupportedOperationException e) {
+                // expected from local operation classes not meant for serialization
+            }
+        }
+    }
+
+    /**
+     * Returns all {@code IdentifiedDataSerializable} concrete classes from production classpath of the currently tested module
+     * only, when argument is {@code true}, otherwise from the JVM's classpath.
```
ihsandemir: Where is the argument?

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
```diff
@@ -0,0 +1,222 @@
...
+                    classesWithInstantiationProblems.add(klass.getName() + " failed with " + e.getMessage());
+                }
+            }
+        }
+
+        if (!classesThrowingUnsupportedOperationException.isEmpty()) {
+            System.out.println("INFO: " + classesThrowingUnsupportedOperationException.size() + " classes threw "
+                    + "UnsupportedOperationException in getFactoryId/getId invocation:");
+            for (String className : classesThrowingUnsupportedOperationException) {
+                System.out.println(className);
```
ihsandemir: Can not we do a test here? Usually, nobody will go and check these print lines in the future.

hazelcast/src/test/java/com/hazelcast/cache/impl/CachePartitionEventDataTest.java
```diff
@@ -0,0 +1,63 @@
+/*
```
ihsandemir: There is already a test called CacheSerializationTest, why not add this to that test?

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
vbekiaris: An `IdentifiedDataSerializable` implementation throwing `UnsupportedOperationException` from `getFactoryId` and/or `getId` is expected (eg local-only operations such as `FinalizeMigrationOperation` which are not supposed to be deserialized/serialized over the network). So this is intended to be only reported for information on the test's output, it should not fail the test.

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
vbekiaris: Comment was left over from previous implementation, fixed.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/operations/LocalLockCleanupOperation.java
tombujok: As discussed with Vassilis, it's probably better to have the getId/factoryId to throw since:

- if I subclass a "non-local" operation to make it local and do not override the getId/factoryI they will return "wrong" values.
- if it's a local class it may use the non-arg constructor for sth else

PR #9104, opened 2016-10-12
HiDensityClientNearCacheTest is failing after  PR https://github.com/hazelcast/hazelcast/pull/8782, on EE side

ihsandemir:
The PR https://github.com/hazelcast/hazelcast/pull/8782 made the enterprise tests fail. The memory manager should not be accessed after it is closed. Since when using the executor for delivering the SHUTTING_DOWN event to AbstractHazelcastCacheManager lifecycle listener, it is possible that the listener is called after the memory manager destruction. We also realized that there is no need for cache manager close while the client is shutting down. Hence, the solution is to remove cachemanager close for client but just leave it for the server side (and server side does not use executor for lifecycle event deliveries it just works fine).


ihsandemir:
@mdogan Yes, both the server and client jcache tests of EE pass in my local run.

PR #9106, opened 2016-10-13
Specific commit id versions can not compile due to protocol 1.3.0-snapshot dependency

ihsandemir:
Changed the protocol dependency to a specific snapshot version (dated version).


asimarslan:
👍

PR #9109, opened 2016-10-13
Introduced LoggingScheduledExecutor

ahmetmircik:
Logging helps to see what went wrong, otherwise scheduled executor stops runnable silently.

According to [javaDocs:](http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleAtFixedRate%28java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit%29) `If any execution of the task encounters an exception, subsequent executions are suppressed`

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/executor/CompletableFutureTask.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/executor/CompletableFutureTask.java
@@ -89,4 +89,11 @@ public class CompletableFutureTask<V> extends AbstractCompletableFuture<V> imple
         }
     }
 
+    @Override
+    public String toString() {
+        return "CompletableFutureTask{"
+                + "callable=" + callable
+                + ", runner=" + runner
+                + '}';
+    }
 }
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
@@ -23,6 +23,7 @@ import com.hazelcast.logging.LoggingService;
 import com.hazelcast.spi.properties.HazelcastProperties;
 import com.hazelcast.spi.properties.HazelcastProperty;
 import com.hazelcast.util.executor.CompletableFutureTask;
+import com.hazelcast.util.executor.LoggingScheduledExecutor;
 import com.hazelcast.util.executor.PoolExecutorThreadFactory;
 
 import java.util.concurrent.Callable;
@@ -32,7 +33,6 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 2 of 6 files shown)


pveentjer:
Can you use a different name than 'diagnostics'? Since we already have a diagnostics module which has its own executor.

hazelcast/src/main/java/com/hazelcast/util/executor/DiagnosticScheduledExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * {@link java.util.concurrent.ScheduledExecutorService} stops a task silently by default if there is an exception.
+ * This class adds logging behaviour for exceptional cases by overriding {@link ScheduledThreadPoolExecutor#afterExecute}
+ * and {@link ScheduledThreadPoolExecutor#decorateTask} methods.
+ *
+ * @see {@link java.util.concurrent.ScheduledExecutorService#scheduleWithFixedDelay}
+ */
+public class DiagnosticScheduledExecutor extends ScheduledThreadPoolExecutor {
+
+    private final ILogger logger = getLogger(getClass());
```
pveentjer: I would love to see this being the way to obtain a logger in Hazelcast, but can't this cause problems? Normally a logger needs to be obtained through the LoggingService.

hazelcast/src/main/java/com/hazelcast/util/executor/DiagnosticScheduledExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public DiagnosticScheduledExecutor(int corePoolSize, ThreadFactory threadFactory) {
+        super(corePoolSize, threadFactory);
+    }
+
+    @Override
+    protected <V> RunnableScheduledFuture<V> decorateTask(Runnable runnable, RunnableScheduledFuture<V> task) {
```
pveentjer: Why do the tasks need to be decorated with something that only provides a toString implementation?

hazelcast/src/main/java/com/hazelcast/util/executor/DiagnosticScheduledExecutor.java
```diff
@@ -0,0 +1,143 @@
...
+            } catch (ExecutionException e) {
+                t = e.getCause();
+            } catch (InterruptedException i) {
+                t = i;
+                currentThread().interrupt();
+            }
+        }
+
+        if (t != null) {
+            logger.severe("Cannot execute " + r, t);
```
pveentjer: I would rename this to 'failed to execute'.

hazelcast/src/main/java/com/hazelcast/util/executor/DiagnosticScheduledExecutor.java
ahmetmircik: it is not straightforward to get running task, that is the reason.

hazelcast/src/main/java/com/hazelcast/util/executor/DiagnosticScheduledExecutor.java
ahmetmircik: fixed

hazelcast/src/main/java/com/hazelcast/util/executor/CompletableFutureTask.java
```diff
@@ -89,4 +89,11 @@ protected void cancelled(boolean mayInterruptIfRunning) {
         }
     }
 
+    @Override
+    public String toString() {
+        return "CompletableFutureTask{"
+                + "callable=" + callable
```
pveentjer: Can you clean up the toString so that either callable or runnable is included? No need to include both, since one of them is never set.

pveentjer:
What kind of tasks get executed by this executor? Do we really want to wrap each task with another object; increasing pressure on the gc. There are way like reflection to get the original callable.

ahmetmircik:
@pveentjer Used nearly for 10 tasks like imap expiration, SyncReplicaVersionTask, phoneHome etc. For gc pressure, it will add only one wrapper for these 10 task, that will not be a repetitive wrapping for every execution so only 10 extra object. For the reflection case, it may not work under security manager or may not be portable, found this way safer than it and did like this.

pveentjer:
Can you add a line of comment to the javadoc that the task is wrapped to provide decent toString. Then I'll give green light. Imho this part of the ScheduledExecutorService is suboptimal; they should have provided a decent toString or a hook to get the actual implementation.

ahmetmircik:
@pveentjer done

PR #9111, opened 2016-10-13
Cleanup of Near Cache related classes

Donnerbart:
* Made `ClientExecutionService` implement `ExecutionService`
* Removed obsolete `NearCacheExecutor`
* Unified `initNearCache()` method names
* Moved `NearCacheContext` creation to `NearCacheProvider` constructor

Please review together with the linked Hazelcast Enterprise PR.


hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientNearCacheExecutor.java
```diff
@@ -24,14 +24,14 @@
 
 public final class ClientNearCacheExecutor implements NearCacheExecutor {
```
pveentjer: Imho this class is just noise. Define a single ScheduledExecutor interface with scheduleWithRepetition and offer it on a 'util' level. Then we don't need to make all these wrapper classes.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
```diff
@@ -79,10 +79,10 @@ public NearCachedClientMapProxy(String serviceName, String name) {
     protected void onInitialize() {
         super.onInitialize();
 
-        init();
```
pveentjer: What is the extra value of calling this 'initNearCache'?

hazelcast/src/main/java/com/hazelcast/map/impl/nearcache/MemberNearCacheExecutor.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.hazelcast.map.impl.nearcache;
+
+import com.hazelcast.cache.impl.nearcache.NearCacheExecutor;
+import com.hazelcast.spi.ExecutionService;
+
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+public final class MemberNearCacheExecutor implements NearCacheExecutor {
```
pveentjer: Another piece of noise.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
Donnerbart: It's overridden in Hazelcast EE.

Donnerbart:
I extended the cleanup to get rid of the `NearCacheExecutor` by adding `ExecutionService` to the client implementation. Maybe the client and member implementation could be unified or more abstracted, but this is beyond the scope of this PR. This PR is limited to cleanup the usage in the Near Cache, which is already improved a lot by getting rid of the useless wrapper class and instances.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientNearCacheExecutor.java
```diff
@@ -1,37 +0,0 @@
...
-
-package com.hazelcast.client.cache.impl;
-
-import com.hazelcast.cache.impl.nearcache.NearCacheExecutor;
-import com.hazelcast.client.spi.ClientExecutionService;
-
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
-
-public final class ClientNearCacheExecutor implements NearCacheExecutor {
```
pveentjer: Deleting code always make me happy.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
```diff
@@ -20,14 +20,18 @@
 import com.hazelcast.core.ICompletableFuture;
```
pveentjer: Not part of this PR; but is this class needed at all? 

So why do we need all these wrapper executors in the first place.

pveentjer:
:+1: 

But another pair of eyes is needed

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
Donnerbart: Don't know exactly yet. The client also implements `Executor` and has an "internal" and "external" executor, the member has some stuff to register and retrieve named executors. So there are some extensions, but maybe we can unify those classes in the end, even if some functionality is just used in either member or clients.

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientExecutionService.java
```diff
@@ -17,19 +17,18 @@
...
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Executor service for Hazelcast clients.
  *
  * Allows asynchronous execution and scheduling of {@link Runnable} and {@link Callable} commands.
  */
-public interface ClientExecutionService extends Executor {
+public interface ClientExecutionService extends ExecutionService, Executor {
```
ahmetmircik: 👍

PR #9112, opened 2016-10-13
Fixed EvictionConfig check and read-only creation

Donnerbart:
* Fixed read-only creation of `EvictionConfig` from `NearCacheReadOnlyConfig`
* Fixed check of `EvictionConfig` which needs a context, since we allow different options for Near Cache and Map/Cache
* Moved checks from `EvictionConfigHelper` to `ConfigValidator`


hazelcast/src/main/java/com/hazelcast/config/NearCacheConfigReadOnly.java
```diff
@@ -81,4 +81,9 @@ public NearCacheConfig setLocalUpdatePolicy(LocalUpdatePolicy localUpdatePolicy)
     public NearCacheConfig setEvictionConfig(EvictionConfig evictionConfig) {
         throw new UnsupportedOperationException("This config is read-only");
     }
+
+    @Override
+    public EvictionConfig getEvictionConfig() {
+        return super.getEvictionConfig().getAsReadOnly();
```
Donnerbart: This broke the EE build before, because it triggered a check of the `EvictionConfig`, which failed in the context of some Near Cache tests.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -137,7 +136,6 @@ public EvictionConfig(EvictionConfig config) {
         if (config.comparator != null) {
             this.comparator = config.comparator;
         }
-        EvictionConfigHelper.checkEvictionConfig(evictionPolicy, comparatorClassName, comparator);
```
Donnerbart: We cannot check the `EvictionConfig`, since we are missing the context here.

hazelcast-spring/src/main/java/com/hazelcast/spring/AbstractHazelcastBeanDefinitionParser.java
```diff
@@ -398,7 +398,7 @@ protected BeanDefinition getEvictionConfig(final Node node) {
             }
 
             try {
-                EvictionConfigHelper.checkEvictionConfig(evictionPolicyValue, comparatorClassNameValue, comparatorBean);
+                checkEvictionConfig(evictionPolicyValue, comparatorClassNameValue, comparatorBean, false);
```
pveentjer: This method has now lost a bit of a readability. The false argument makes no sense when reading the invocation? 

Can't you add an method checkNearCacheEvictionConfig method which puts in false for this argument?

hazelcast/src/main/java/com/hazelcast/internal/config/ConfigValidator.java
```diff
@@ -64,6 +71,60 @@ public static void checkNearCacheConfig(NearCacheConfig nearCacheConfig, boolean
...
+     * Checks if parameters for an {@link EvictionConfig} are valid in their context.
+     *
+     * @param evictionPolicy      the {@link EvictionPolicy} for the {@link EvictionConfig}
+     * @param comparatorClassName the comparator class name for the {@link EvictionConfig}
+     * @param comparator          the comparator implementation for the {@link EvictionConfig}
+     * @param isNearCache         {@code true} if the config is for a Near Cache, {@code false} otherwise
+     */
+    public static void checkEvictionConfig(EvictionPolicy evictionPolicy,
+                                           String comparatorClassName,
+                                           Object comparator,
```
pveentjer: Object as argument?

hazelcast-spring/src/main/java/com/hazelcast/spring/AbstractHazelcastBeanDefinitionParser.java
pveentjer: When I read the method call.. I have difficulties with true and false in lot of cases.

hazelcast/src/main/java/com/hazelcast/internal/config/ConfigValidator.java
pveentjer: :+1:

hazelcast-spring/src/main/java/com/hazelcast/spring/AbstractHazelcastBeanDefinitionParser.java
Donnerbart: I know what you mean, but the obvious decision path for the `EvictionConfig` check is "is it a config for Near Cache or (`IMap`|`ICache`|I think there was a third one)". So there is not really a nice way to describe the "not a Near Cache". This is why we have this `false` parameter in this one place.

I could make two wrapper methods like `checkEvictionConfigForNearCache()` and `checkEvictionConfigForNonNearCache()`, but that looks a bit too much for the very few calls we have for those methods.

Donnerbart:
run-lab-run

PR #9113, opened 2016-10-14
Convert Cache DataSerializable classes to IdentifiedDataSerializable

vbekiaris:
These `cache` classes are not used in client-member communication and can be safely converted to `IdentifiedDataSerializable`. No EE changes are required.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheDataSerializerHook.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/CacheDataSerializerHook.java
@@ -19,6 +19,8 @@ package com.hazelcast.cache.impl;
 import com.hazelcast.cache.HazelcastExpiryPolicy;
 import com.hazelcast.cache.impl.client.CacheBatchInvalidationMessage;
 import com.hazelcast.cache.impl.client.CacheSingleInvalidationMessage;
+import com.hazelcast.cache.impl.event.CachePartitionLostEventFilter;
+import com.hazelcast.cache.impl.merge.entry.DefaultCacheEntryView;
 import com.hazelcast.cache.impl.operation.CacheBackupEntryProcessorOperation;
 import com.hazelcast.cache.impl.operation.CacheClearBackupOperation;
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/CacheDataSerializerHook.java
```diff
@@ -337,6 +349,36 @@ public IdentifiedDataSerializable createNew(Integer arg) {
                 return new CacheEntryIterationResult();
             }
         };
+        constructors[CACHE_PARTITION_LOST_EVENT_FILTER] = new ConstructorFunction<Integer, IdentifiedDataSerializable>() {
+            public IdentifiedDataSerializable createNew(Integer arg) {
+                return new CachePartitionLostEventFilter();
+            }
```
tombujok: not touched by client?

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheDataSerializerHook.java
```diff
@@ -337,6 +349,36 @@ public IdentifiedDataSerializable createNew(Integer arg) {
...
+        };
+        constructors[DEFAULT_CACHE_ENTRY_VIEW] = new ConstructorFunction<Integer, IdentifiedDataSerializable>() {
+            public IdentifiedDataSerializable createNew(Integer arg) {
+                return new DefaultCacheEntryView();
+            }
```
tombujok: not touched by client?

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheDataSerializerHook.java
vbekiaris: Only used in member-side `CacheAddPartitionLostListenerMessageTask.call`, as far as I can tell this is not transferred over the network between client-member, only among members during member-side listener registration.
Client-side listener registration registers a member-side `InternalCachePartitionLostListenerAdapter` listener which then delivers `CachePartitionLostEvent` (a plain old `Serializable`) to the client.

PR #9118, opened 2016-10-17
ParkedOperation IDS for the sake of static code validation

tombujok:


Donnerbart:
Instead of creating dead code (which hurts the code coverage, since I don't see a test for this), we could just suppress the failure, whatever it is (I assume FindBugs).

tombujok:
I agree but it's not dead code. This way no-one will every serialize it.

tombujok:
BTW, later on I'll introduce a abstract LocalOperation in a separate PR and most of this code will be gone. There's more operations like this in OS and EE.

PR #9119, opened 2016-10-17
Diagnostics Store Latency plugin

pveentjer:
Added a plugin for the the different store like MapStore to provide information about latencies. All kinds of different statistics are available like count, max, avg, and latency distribution.

Example output:

```
17-9-2016 13:12:34 MapStoreLatency[
                          map[
                                  loadAllKeys[
                                          count=1
                                          totalTime(us)=8
                                          avg(us)=8
                                          max(us)=8
                                          latency-distribution[
                                                  0..99us=1]]
                                  load[
                                          count=100
                                          totalTime(us)=4,632,190
                                          avg(us)=46,321
                                          max(us)=99,178
                                          latency-distribution[
                                                  0..99us=1
                                                  1600..3199us=3
                                                  3200..6399us=3
                                                  6400..12799us=7
                                                  12800..25599us=13
                                                  25600..51199us=32
                                                  51200..102399us=41]]]]
17-9-2016 13:12:34 RingbufferStoreLatency[
                          ring[
                                  load[
                                          count=47
                                          totalTime(us)=2,366,188
                                          avg(us)=50,344
                                          max(us)=96,176
                                          latency-distribution[
                                                  1600..3199us=1
                                                  6400..12799us=7
                                                  12800..25599us=3
                                                  25600..51199us=12
                                                  51200..102399us=24]]
                                  getLargestSequenceProbe[
                                          count=1
                                          totalTime(us)=38,186
                                          avg(us)=38,186
                                          max(us)=38,186
                                          latency-distribution[
                                                  25600..51199us=1]]]]

```

The approach used is decorate the Store/Loader with a Latency tracking decorator. This proxy is dynamically wrapped around the original Store/Loader. If the LatencyStorePlugin is disabled, the decorator is not added; and therefor there is no performance penalty.

The main purpose of this plugin is to provide solid information if e.g. a database is causing problems. We want to see which Store/Loader was running into problems and we want to have all kinds of statistics to determine what is happening.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/LatencyTrackingQueueStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/LatencyTrackingQueueStore.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.collection.impl.queue;
+
+import com.hazelcast.core.QueueStore;
+import com.hazelcast.internal.diagnostics.StoreLatencyPlugin;
+import com.hazelcast.internal.diagnostics.StoreLatencyPlugin.LatencyProbe;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+
+public class LatencyTrackingQueueStore<T> implements QueueStore<T> {
+    static final String KEY = "QueueStoreLatency";
+
+    private final LatencyProbe loadProbe;
+    private final LatencyProbe loadAllKeysProbe;
+    private final LatencyProbe loadAllProbe;
+    private final LatencyProbe deleteProbe;
+    private final LatencyProbe deleteAllProbe;
+    private final LatencyProbe storeProbe;
+    private final LatencyProbe storeAllProbe;
+    private final QueueStore<T> delegate;
+
+    public LatencyTrackingQueueStore(QueueStore<T> delegate, StoreLatencyPlugin plugin, String queueName) {
+        this.delegate = delegate;
+        this.loadProbe = plugin.newProbe(KEY, queueName, "load");
+        this.loadAllProbe = plugin.newProbe(KEY, queueName, "loadAll");
+        this.loadAllKeysProbe = plugin.newProbe(KEY, queueName, "loadAllKeys");
+        this.deleteProbe = plugin.newProbe(KEY, queueName, "delete");
+        this.deleteAllProbe = plugin.newProbe(KEY, queueName, "deleteAll");
+        this.storeProbe = plugin.newProbe(KEY, queueName, "store");
+        this.storeAllProbe = plugin.newProbe(KEY, queueName, "storeAll");
+    }
+
+    @Override
+    public void store(Long key, T value) {
+        long startNanos = System.nanoTime();
+        try {
+            delegate.store(key, value);
+        } finally {
+            storeProbe.recordValue(System.nanoTime() - startNanos);
+        }
+    }
+
+    @Override
+    public void storeAll(Map<Long, T> map) {
+        long startNanos = System.nanoTime();
+        try {
+            delegate.storeAll(map);
+        } finally {
+            storeAllProbe.recordValue(System.nanoTime() - startNanos);
+        }
+    }
+
+    @Override
+    public void delete(Long key) {
+        long startNanos = System.nanoTime();
+        try {
+            delegate.delete(key);
+        } finally {
+            deleteProbe.recordValue(System.nanoTime() - startNanos);
+        }
+    }
+
+    @Override
+    public void deleteAll(Collection<Long> keys) {
+        long startNanos = System.nanoTime();
+        try {
+            delegate.deleteAll(keys);
+        } finally {
+            deleteAllProbe.recordValue(System.nanoTime() - startNanos);
+        }
+    }
+
+    @Override
+    public T load(Long key) {
+        long startNanos = System.nanoTime();
+        try {
+            return delegate.load(key);
+        } finally {
+            loadProbe.recordValue(System.nanoTime() - startNanos);
+        }
+    }
+
+    @Override
+    public Map<Long, T> loadAll(Collection<Long> keys) {
+        long startNanos = System.nanoTime();
+        try {
+            return delegate.loadAll(keys);
+        } finally {
+            loadAllProbe.recordValue(System.nanoTime() - startNanos);
+        }
+    }
+
+    @Override
+    public Set<Long> loadAllKeys() {
+        long startNanos = System.nanoTime();
+        try {
+            return delegate.loadAllKeys();
+        } finally {
+            loadAllKeysProbe.recordValue(System.nanoTime() - startNanos);
+        }
+    }
+}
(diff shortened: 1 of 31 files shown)


hazelcast/src/main/java/com/hazelcast/Main.java
```diff
@@ -0,0 +1,151 @@
+package com.hazelcast;
```
pveentjer: Will be removed from the final PR.

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/LatencyTrackingCacheLoader.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class LatencyTrackingCacheLoader<K, V> implements CacheLoader<K, V> {
+
+    private static final String KEY = "CacheLoaderLatency";
+
+    private final CacheLoader<K, V> delegate;
+    private final StoreLatencyPlugin.LatencyProbe loadProbe;
+    private final StoreLatencyPlugin.LatencyProbe loadAllProbe;
+
+    public LatencyTrackingCacheLoader(CacheLoader<K, V> delegate, StoreLatencyPlugin plugin, String cacheName) {
```
mmedenjak: How about if the LatencyTracking* classes don't get the plugin but rather have fields annotated with `@Probe`. The decorators would be instantiated not if the plugin is not-null but rather by configuration. Additionally, the `com.hazelcast.internal.metrics.impl.FieldProbe` class would need to be expanded to support a new kind of field (LatencyProbe). Just a suggestion so the decorators don't need to know about plugins.

hazelcast/src/main/java/com/hazelcast/internal/diagnostics/Diagnostics.java
```diff
@@ -115,6 +117,8 @@
     private final boolean enabled;
     private ScheduledExecutorService scheduler;
     private final HazelcastThreadGroup hzThreadGroup;
+    private final ConcurrentMap<Class<? extends DiagnosticsPlugin>, DiagnosticsPlugin> pluginsMap
```
mmedenjak: If we had `@Probe` annotated latency measuring fields we wouldn't need this.

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/LatencyTrackingQueueStore.java
```diff
@@ -0,0 +1,118 @@
...
+    private final StoreLatencyPlugin.LatencyProbe loadAllProbe;
+    private final StoreLatencyPlugin.LatencyProbe deleteProbe;
+    private final StoreLatencyPlugin.LatencyProbe deleteAllProbe;
+    private final StoreLatencyPlugin.LatencyProbe storeProbe;
+    private final StoreLatencyPlugin.LatencyProbe storeAllProbe;
+    private final QueueStore<T> delegate;
+
+    public LatencyTrackingQueueStore(String queueName, QueueStore<T> delegate, StoreLatencyPlugin plugin) {
+        this.delegate = delegate;
+        this.loadProbe = plugin.newProbe(KEY, queueName, "load");
```
mmedenjak: The probes would be then detected by reflection

hazelcast/src/main/java/com/hazelcast/internal/diagnostics/Diagnostics.java
pveentjer: I know. 

I'm still not sure how the final design should look like for latency tracking. Currently we have a cheap way to track some counters, so I'm fine with having them enabled even if diagnostics is disabled. But I'm not looking forward to sprinkle the code with latency measuring code even though it perhaps isn't used.

So I'm not yet sure how the metrics api should be changed. Another issue is that the metrics system, dumps all metrics. But in this case, if we would make use of metrics, I don't want the metrics writer do dump all metrics; some metrics have their own renderer (in this case StoreLatencyPlugin). 

So I understand that the current design is not optimal. But I don't have a clear vision how it should look like before embarking on a big metrics change. Eventually this latency tracking stuff will be merged in metrics..  but currently I don't see a clear way how to do it. 

Another issue is that I have 5 days for this task; adding this stuff into metrics was not part of that estimation.

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/cache/impl/ICacheRecordStore.java
```diff
@@ -448,4 +449,6 @@
      * @return <tt>true</tt> if wan replication is enabled for this record store, <tt>false</tt> otherwise
      */
     boolean isWanReplicationEnabled();
+
+    void instrument(NodeEngine nodeEngine);
```
mmedenjak: How about removing this from the interface and keeping this on the instance? Seems like something that not all implementations would want to have.

pveentjer: Good spotting.

hazelcast/src/main/java/com/hazelcast/map/impl/LatencyTrackingMapLoader.java
```diff
@@ -0,0 +1,71 @@
...
+import com.hazelcast.core.MapLoader;
+import com.hazelcast.internal.diagnostics.StoreLatencyPlugin;
+import com.hazelcast.internal.diagnostics.StoreLatencyPlugin.LatencyProbe;
+
+import java.util.Collection;
+import java.util.Map;
+
+public class LatencyTrackingMapLoader<K, V> implements MapLoader<K, V> {
+
+    static final String KEY = "MapStoreLatency";
```
ahmetmircik: `MapLoaderLatency` ?

pveentjer: I thought about this. However..  if a user would implement a MapStore,then part of its information is in the MapStoreLatency subtree and the other part in the MapLoaderLatency section.

So the information tree above.

hazelcast/src/test/java/com/hazelcast/map/impl/StoreLatencyPlugin_MapIntegrationTest.java
```diff
@@ -0,0 +1,123 @@
...
+        for (int k = 0; k < 100; k++) {
+            map.get(k);
+        }
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                File file = getNodeEngineImpl(hz).getDiagnostics().currentFile();
+                String content = fileAsText(file);
+                System.out.println(content);
```
ahmetmircik: `System.out.println` left-over?

pveentjer: Indeed. I'll get rid of it. Good spotting

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/StoreLatencyPlugin_RingbufferIntegrationTest.java
```diff
@@ -0,0 +1,102 @@
...
+        for (long k = 0; k <= rb.tailSequence(); k++) {
+            rb.readOne(k);
+        }
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                File file = getNodeEngineImpl(hz).getDiagnostics().currentFile();
+                String content = fileAsText(file);
+                System.out.println(content);
```
pveentjer: I'll remove it

ahmetmircik:
For the example output:
```17-9-2016 13:12:34 MapStoreLatency[
                          map[
                                  loadAllKeys[
                                          count=1
                                          totalTime(us)=8
                                          avg(us)=8
                                          max(us)=8
                                          latency-distribution[
                                                  0..99us=1]]
                                  load[
                                          count=100
                                          totalTime(us)=4,632,190
                                          avg(us)=46,321
                                          max(us)=99,178
                                          latency-distribution[
                                                  0..99us=1
                                                  1600..3199us=3
                                                  3200..6399us=3
                                                  6400..12799us=7
                                                  12800..25599us=13
                                                  25600..51199us=32
                                                  51200..102399us=41]]]]```

How do i know for which map this output belongs to? Does it print map-name in the output?

**Comment by pveentjer** [comment]
The name I chose for the map is 'map'. See the example below. If I would have called it 'employees' it would have been a better example

```
17-9-2016 13:12:34 MapStoreLatency[
                          map[<---------------- HERE IS THE MAP NAME
                                  loadAllKeys[
                                          count=1
                                          totalTime(us)=8
                                          avg(us)=8
                                          max(us)=8
                                          latency-distribution[
                                                  0..99us=1]]
                                  load[
                                          count=100
                                          totalTime(us)=4,632,190
                                          avg(us)=46,321
                                          max(us)=99,178
                                          latency-distribution[
                                                  0..99us=1
                                                  1600..3199us=3
                                                  3200..6399us=3
                                                  6400..12799us=7
                                                  12800..25599us=13
                                                  25600..51199us=32
                                                  51200..102399us=41]]]]
```

PR #9120, opened 2016-10-17
ManagementOperation converted to IdentifiedDataSerializable

tombujok:


vbekiaris:
`ThreadDumpOperation`, `GetMapConfigOperation` and `ChangeWanStateOperation` are effectively only used locally (via `ManagementCenterService.callOnThis`). In fact, if `GetMapConfigOperation` is invoked remotely, then it will throw an exception as its `MapConfig` response is not serializable at all. Should we perhaps throw `UnsupportedOperationException` for attempts to serialize/deserialize these 3 operations via serializer hooks?

tombujok:
@vbekiaris True, on the other hand I wanted not to change any logic in these transformations. In theory you can serialize this classes prior to this PR.

PR #9126, opened 2016-10-18
Added PartitionPredicate

pveentjer:
The PartitionPredicate restricts the execution of its target predicate to a single partition.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
@@ -47,6 +47,7 @@ import com.hazelcast.mapreduce.aggregation.Aggregation;
 import com.hazelcast.mapreduce.aggregation.Supplier;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.query.PagingPredicate;
+import com.hazelcast.query.PartitionPredicate;
 import com.hazelcast.query.Predicate;
 import com.hazelcast.query.TruePredicate;
 import com.hazelcast.spi.InitializingObject;
@@ -607,16 +608,7 @@ public class MapProxyImpl<K, V> extends MapProxySupport implements IMap<K, V>, I
     @Override
     @SuppressWarnings("unchecked")
     public Set<K> keySet(Predicate predicate) {
-        checkNotNull(predicate, NULL_PREDICATE_IS_NOT_ALLOWED);
-
-        MapQueryEngine queryEngine = getMapQueryEngine();
-        if (predicate instanceof PagingPredicate) {
-            return queryEngine.queryAllPartitionsWithPagingPredicate(name, (PagingPredicate) predicate, IterationType.KEY);
-        } else {
-            QueryResult result = queryEngine.invokeQueryAllPartitions(name, predicate, IterationType.KEY);
-            return new QueryResultCollection<K>(
-                    getNodeEngine().getSerializationService(), IterationType.KEY, false, true, result);
-        }
+        return executePredicate(predicate, IterationType.KEY, true);
     }
 
     @Override
@@ -626,16 +618,7 @@ public class MapProxyImpl<K, V> extends MapProxySupport implements IMap<K, V>, I
 
     @Override
     public Set entrySet(Predicate predicate) {
-        checkNotNull(predicate, NULL_PREDICATE_IS_NOT_ALLOWED);
-
-        MapQueryEngine queryEngine = getMapQueryEngine();
-        if (predicate instanceof PagingPredicate) {
-            return queryEngine.queryAllPartitionsWithPagingPredicate(name, (PagingPredicate) predicate, IterationType.ENTRY);
-        } else {
-            QueryResult result = queryEngine.invokeQueryAllPartitions(name, predicate, IterationType.ENTRY);
-            return new QueryResultCollection<Map.Entry<K, V>>(
-                    getNodeEngine().getSerializationService(), IterationType.ENTRY, false, true, result);
-        }
+        return executePredicate(predicate, IterationType.ENTRY, true);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


ahmetmircik:
@pveentjer can you also point the reasoning behind this addition?

jerrinot:
I am not convinced there are good reasons to expose `partitionId` to query users. 
Other places use `keyOwner` as to abstract end-users away from partitions.

pveentjer:
@ahmetmircik  The most important reason is a performance optimization because you don't need to access a single partition instead of all partitions.

ahmetmircik:
@pveentjer thanks for the explanation, it makes sense. Only worry is `partitionId` seems more internal at this level, can we make it transparent to the user? Maybe user can only give an object and we calculate its partition?

hazelcast/src/main/java/com/hazelcast/query/PartitionPredicate.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * A {@link Predicate} that restricts the execution of a Predicate to a single Partition. This can help to speed up query
+ * execution since only 1 instead of all partitions needs to be queried.
+ *
+ * This predicate can only be used as an outer predicate.
+ *
+ * @param <K>
+ * @param <V>
+ */
+public class PartitionPredicate<K, V> implements Predicate<K, V>, DataSerializable {
```
ahmetmircik: `IdentifiedDataSerializable`?

pveentjer:
Code has been converted to partitionKey (Object) instead of Partition-id

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
```diff
@@ -613,7 +614,15 @@ public void clearMapOnly() {
         if (predicate instanceof PagingPredicate) {
             return queryEngine.queryAllPartitionsWithPagingPredicate(name, (PagingPredicate) predicate, IterationType.KEY);
         } else {
```
tombujok: Why couldn't the user page per single-partition?

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
```diff
@@ -652,7 +669,16 @@ public Set entrySet(Predicate predicate) {
...
+            if (predicate instanceof PartitionPredicate) {
+                PartitionPredicate partitionPredicate = (PartitionPredicate) predicate;
+                int partitionId = getNodeEngine().getPartitionService().getPartitionId(partitionPredicate.getPartitionKey());
+                result = queryEngine.invokeQuerySinglePartition(
+                        name, partitionPredicate.getTarget(), IterationType.VALUE, partitionId);
+            } else {
+                result = queryEngine.invokeQueryAllPartitions(name, predicate, IterationType.VALUE);
+            }
+
             return new QueryResultCollection<V>(
```
tombujok: The only difference between the pasted snippets is the IterationType. An abstraction to a method would do good here.

hazelcast/src/main/java/com/hazelcast/query/PartitionPredicate.java
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @return the target Predicate.
+     */
+    public Predicate<K, V> getTarget() {
+        return target;
+    }
+
+    @Override
+    public boolean apply(Map.Entry<K, V> mapEntry) {
+        throw new UnsupportedOperationException();
```
ahmetmircik: what about returning `return target.apply(mapEntry)` instead of `throw new UnsupportedOperationException()`? 
2 quick benefit: we don't need to remember `getTarget` call and `getTarget` method can be removed.

hazelcast/src/main/java/com/hazelcast/query/PartitionPredicate.java
```diff
@@ -0,0 +1,92 @@
...
+ * A {@link Predicate} that restricts the execution of a Predicate to a single Partition. This can help to speed up query
+ * execution since only 1 instead of all partitions needs to be queried.
+ *
+ * This predicate can only be used as an outer predicate.
+ *
+ * @param <K>
+ * @param <V>
+ */
+public class PartitionPredicate<K, V> implements Predicate<K, V>, DataSerializable {
+
```
tombujok: This is mixing concerns. I understand it was easier to implement it this way that you just wrap a predicate around, since the existing implementation didn't have to be changed at all, but why should even a user know about this class at all? This is just an implementation detail.
It you want to expose it like this, the PagingPredicate  should be a private class hidden behind the Predicates factory, exposed through the Predicate Interface only.
Like this:
```
Predicate query = Predicates.equals("age", 40)
Predicate query = Predicates.onPartition(query, partitionKey)
map.values(query)
```

To me, however, adding a logic on how many partitions a query should be executed should not be passed by a predicate. I know it's just a wrapper, but it's not an healthy way how to grow an API in a sustainable way. What if I want to add one or two similar flags, how many times will I wrap the predicate around?

Don't get me wrong, I like the idea, but not how it's exposed to the end-user in the public API.
Predicates may be used in multiple places. e.g. ReplicatedMap, TransactionalMap, etc. 
What will happen if the user passes a PartitionPredicate there?
Even in the IMap you changed `keySet`, `entrySet` and `values` only. 
There are other methods like
`localKeySet` , `executeOnEntries`  that do not support this type of Predicate.

hazelcast/src/main/java/com/hazelcast/query/PartitionPredicate.java
pveentjer: The PartitionPredicate is meant as a control mechanism for the QueryEngine to schedule the Query. This PartitionPredicate is then stripped away from the actual predicate and that gets executed.

If this PartitionPredicate does allow for execution, it would be possible to do
```
not(partitionpredicate(1,true))
```
But this would not be valid, since it would not be executing on partition one. So it is best to fail the PartitionPredicate if it does get executed. This would follow the fail fast principle.

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
```diff
@@ -646,16 +629,31 @@ public Set entrySet(Predicate predicate) {
     @Override
     @SuppressWarnings("unchecked")
     public Collection<V> values(Predicate predicate) {
+        return executePredicate(predicate, IterationType.VALUE, false);
+    }
+
+    // todo: this logic should be moved to the query engine because interpreting the query is not a task
+    // of the MapProxy.
```
vbekiaris: 👍 Testament to the fact: will you also implement this in the tasks supporting client-side map proxy?

pveentjer: I have no such plans. The query engine is a pure a server side matter and if the client code needs a cleanup dealing with predicates, it should be done in a different task.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
tombujok: OK, thx. Let's not move it now, since I'm in the middle of the refactoring of the queryEngine.

tombujok:
As I said in one of the unanswered comments adding a predicate which is not really a predicate only for a limited subset of methods in Map that use predicates (ignoring that ReplicatedMap, TransactionalMap also use Predicates) just to pass a hint to a queryEngine is a suboptimal-decision to me.

Throwing an UnsupportedOperationException in all other places that do not support PartitionPredicate is the best possible API design (even in IMap there's `localKeySet` and `executeOnEntries`).

Since you only care  about `keySet`, `entrySet` and `values` methods I would overload them with signature `values(Predicate predicate, Object partitionKeyToFetchFrom)` to make the API clean and extensible.

hazelcast/src/main/java/com/hazelcast/query/PartitionPredicate.java
tombujok: Peter, I don't know if you saw this comment. It's shown as outdated.

pveentjer:
The current design is in line with other wrapping approaches like paging predicate.

This task has grown from a 10 minute job into a multi hour effort and the more people join, the more opinions are being added. I'm failing to see the added value of this.

pveentjer:
I'm reopening it. I believe Tom figured out that adding an overloaded method for keyset,entryset,values with an additional argument, will rapidly lead to a method explosion since now we already need to add 6 methods (3+3 without predicate); but if we would continue with this approach and he would add his transformer argument, there would already be 12 methods exposed for executing on a particular query. 

In HZ 4 we should introduce a query object so we can add control settings to that. Where we don't suffer from method explosion, but also don't suffer from accidentally getting the PartitionPredicate not being an outer predicate.

So something like this:

```
Query.of("where age>40")
    .maxCount(25)
    .maxPage(30)
    .partitionKey("007")
    .resultTransformer(propertyExtractor)
    .build();
```

tombujok:
Still very unhappy with the design of this PR. I hold on to your word to fix it in 4.0

gurbuzali:
currently we don't support `PartitionPredicate` for clients, do we have plans to do so ?

burakcelebi:
We need this for .Net Client too.

PR #9137, opened 2016-10-19
Queue quorum

mmedenjak:
Adding current quorum support to queue actions - https://hazelcast.atlassian.net/wiki/pages/viewpage.action?pageId=105447518.

Also, doing some minor refactoring and adding javadoc.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/operations/QueueOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/collection/impl/queue/operations/QueueOperation.java
@@ -23,28 +23,24 @@ import com.hazelcast.collection.impl.queue.QueueEventFilter;
 import com.hazelcast.collection.impl.queue.QueueService;
 import com.hazelcast.core.ItemEventType;
 import com.hazelcast.nio.Address;
-import com.hazelcast.nio.ObjectDataInput;
-import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
 import com.hazelcast.spi.BackupOperation;
 import com.hazelcast.spi.EventRegistration;
 import com.hazelcast.spi.EventService;
-import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.NamedOperation;
 import com.hazelcast.spi.PartitionAwareOperation;
 import com.hazelcast.spi.exception.RetryableHazelcastException;
+import com.hazelcast.spi.impl.AbstractNamedOperation;
 
-import java.io.IOException;
 import java.util.Collection;
 
 /**
  * This class contains methods for Queue operations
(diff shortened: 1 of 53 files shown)


mmedenjak:
run-lab-run

pveentjer:
After burning my fingers many times myself; keep pr's as clean as possible. If a PR has multiple things; it is going to bite you one way or another. Either the pr gets reviewed late; if there is a problem.. it is difficult to figure out what is relevant in the given pr. It also makes reviewing more problematic since it is more difficult to figure out what is the feature. At the end.. the amount of suffering caused by splitting up the PR is larger than the amount of suffering after sending a fat pr.

hazelcast/src/main/java/com/hazelcast/collection/impl/queue/operations/QueueOperation.java
```diff
@@ -126,11 +121,11 @@ public int getFactoryId() {
 
     @Override
```
pveentjer: Can these methods not be fully deleted?

hazelcast/src/main/java/com/hazelcast/quorum/impl/QuorumServiceImpl.java
```diff
@@ -106,25 +107,41 @@ public void addQuorumListener(String name, QuorumListener listener) {
...
+     * @return the quorum
+     */
     private QuorumImpl findQuorum(Operation op) {
         if (!isNamedOperation(op) || !isQuorumAware(op)) {
             return null;
         }
-        QuorumAwareService service = op.getService();
-        String name = ((NamedOperation) op).getName();
-        String quorumName = service.getQuorumName(name);
+        final QuorumAwareService service = op.getService();
```
pveentjer: bad boy

pveentjer:
Looks good. Minor comments.

hazelcast/src/main/java/com/hazelcast/quorum/impl/QuorumServiceImpl.java
```diff
@@ -106,25 +107,41 @@ public void addQuorumListener(String name, QuorumListener listener) {
...
+     * @param op the operation for which the quorum should be present
+     * @throws QuorumException if the operation requires a quorum and the quorum is not present
+     */
     public void ensureQuorumPresent(Operation op) {
         if (inactive) {
             return;
         }
 
-        QuorumImpl quorum = findQuorum(op);
+        final QuorumImpl quorum = findQuorum(op);
```
mmedenjak: Ah yes. This could not be the place for discussion but working on Marko's code for some time I've come to like them - makes the variables more visible due to syntax highlighting and makes it obvious that the value will not change. I guess I'll keep adding them, maybe just contain it to the new code that I write. I realise that it creates noise when I include this on existing code in a PR.

mmedenjak:
@pveentjer I agree on the overzealous usage of `final`. Toned it down a bit and containing it to new code.

pveentjer:
👍 

Please find another reviewer since this is not a small change.

PR #9142, opened 2016-10-19
Convert TransactionLogRecord hierarchy to IdentifiedDataSerializable

vbekiaris:
TransactionLogRecord and subclasses converted to `IdentifiedDataSerializable`

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/tx/MapTransactionStressTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/tx/MapTransactionStressTest.java
@@ -368,6 +368,16 @@ public class MapTransactionStressTest extends HazelcastTestSupport {
         public String toString() {
             return "SleepyTransactionLogRecord{}";
         }
+
+        @Override
+        public int getFactoryId() {
+            return 0;
+        }
+
+        @Override
+        public int getId() {
+            return 0;
+        }
     }
 
     static class TxnIncrementor implements Runnable {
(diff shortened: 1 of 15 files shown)


hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/FactoryIdHelper.java
```diff
@@ -112,6 +112,11 @@
 
     public static final String MANAGEMENT_DS_FACTORY = "hazelcast.serialization.ds.management";
     public static final int MANAGEMENT_DS_FACTORY_ID = -36;
+
+    // DataSerializerHook for test IdentifiedDataSerializable classes
+    public static final String TEST_DS_FACTORY = "hazelcast.test.ds.management";
+    public static final int TEST_DS_FACTORY_ID = -37;
+
```
tombujok: Hmm, this is a test code in production code-base.
What I do in tests with IDS classes is I just register the test-factory in the serialization config. 
Maybe it would be a better way to go her too?

hazelcast/src/test/java/com/hazelcast/map/impl/tx/MapTransactionStressTest.java
```diff
@@ -368,6 +368,16 @@ public void readData(ObjectDataInput in) throws IOException {
         public String toString() {
             return "SleepyTransactionLogRecord{}";
         }
+
+        @Override
+        public int getFactoryId() {
+            return 0;
```
gurbuzali: I don't what is the test logic but I guess this won't get serialized ?

vbekiaris: Yes, this class does not get serialized during the test.

PR #9146, opened 2016-10-20
Internal classes converted to IdentifiedDataSerializable

vbekiaris:
Introduces new `TextProtocolsDataSerializerHook` for REST & Memcache protocol support classes.
Converts `com.hazelcast.internal` classes from `DataSerializable` to `IdentifiedDataSerializable`.
After merge, requires an EE-side counterpart to register the new `DataSerializerHook`.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
@@ -17,6 +17,7 @@
 package com.hazelcast.internal.cluster.impl;
 
 import com.hazelcast.instance.MemberImpl;
+import com.hazelcast.internal.cluster.MemberInfo;
 import com.hazelcast.internal.cluster.impl.operations.AuthenticationFailureOperation;
 import com.hazelcast.internal.cluster.impl.operations.AuthorizationOperation;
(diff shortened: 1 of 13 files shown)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
```diff
@@ -66,7 +68,7 @@
     public static final int CONFIG_MISMATCH = 11;
     public static final int GROUP_MISMATCH = 12;
     public static final int JOIN_CHECK = 13;
-    public static final int JOIN_REQUEST = 14;
+    public static final int JOIN_REQUEST_OP = 14;
```
vbekiaris: Renamed to disambiguate `JoinRequestOperation` vs `JoinRequest` class added below.

pveentjer:
It looks good. 

My concern is that this could potentially impact client like MemberInfo. It always is a bit obscure to much which classes the client relies on.

vbekiaris:
`MemberInfo` is only used in cluster operations (`JoinRequest`, `FinalizeJoinOperation`, `MemberInfoUpdateOperation`); client makes use of `com.hazelcast.client.impl.MemberImpl` & `AbstractMember` instead.

pveentjer:
:+1:

PR #9147, opened 2016-10-20
Fix QueueTransactionLogRecord type ID

vbekiaris:
QueueTransactionLogRecord ID slipped through being the same with CollectionTransactionLogRecord one.
A test to verify `IdentifiedDataSerializable`s have unique {factoryId, id} combinations will be provided in #9102 along with other related tests once conversion to `IdentifiedDataSerializable` is done.


pveentjer:
:+1:

Issue #9023, opened 2016-09-29
3.6.5 to 3.7.1 Upgrade breaks Spring cache


we64:
I have a Spring 3.2.17 project that has been using Hazelcast 3.6.5 fine with Spring cache utilizing com.hazelcast.spring.cache.HazelcastCacheManager directly.

When I change the Hazelcast version to 3.7.1 in my pom, calls to Spring cache will generate the following exception:
`java.lang.ClassNotFoundException: org.springframework.cache.Cache$ValueRetrievalException
	org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1332)
	org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1166)
	com.hazelcast.spring.cache.HazelcastCacheManager.getCache(HazelcastCacheManager.java:51)
	org.springframework.cache.interceptor.CacheAspectSupport.getCaches(CacheAspectSupport.java:164)
	org.springframework.cache.interceptor.CacheAspectSupport$CacheOperationContext.&lt;init&gt;(CacheAspectSupport.java:411)
	org.springframework.cache.interceptor.CacheAspectSupport.getOperationContext(CacheAspectSupport.java:176)
	org.springframework.cache.interceptor.CacheAspectSupport.createOperationContext(CacheAspectSupport.java:366)
	org.springframework.cache.interceptor.CacheAspectSupport.execute(CacheAspectSupport.java:195)
	org.springframework.cache.interceptor.CacheInterceptor.invoke(CacheInterceptor.java:66)
	org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:172)
	org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:633)
	com.brilent.ga.parser.util.ApplicantMatchingResultUtil$$EnhancerBySpringCGLIB$$72abbb9a.clearAllApplicantMatchingResultCache(&lt;generated&gt;)
	com.brilent.ga.parser.dao.ParsingDAO.processMatchingTasks(ParsingDAO.java:4643)
	com.brilent.ga.parser.dao.ParsingDAO.matchJobActualWork(ParsingDAO.java:4105)
	com.brilent.ga.parser.dao.ParsingDAO.matchJobDescription(ParsingDAO.java:4068)
[...]

PR #9148, opened 2016-10-20
Delegate resolution of ValueRetrievalException in inner delegate

snicoll:
This commit fixes a `ClassNotFoundException` on
`Cache$ValueRetrievalException` when using the cache abstraction with
Spring Framework prior to 4.3.

`ValueRetrievalException` is an inner exception class introduced in
Spring Framework 4.3. Throwing that exception in the body of a method
should have been working fine but for some reason the body of the method
is resolved upfront when the class is loaded.

This commits moves the resolution to an inner class to enforce that such
type is not going to be loaded unless the class is invoked explicitly and
this only happens if the `sync` attribute is used (new in 4.3).

Closes gh-9023

### DIFF
--- a/hazelcast-spring/src/main/java/com/hazelcast/spring/cache/HazelcastCache.java
+++ b/hazelcast-spring/src/main/java/com/hazelcast/spring/cache/HazelcastCache.java
@@ -91,7 +91,7 @@ public class HazelcastCache implements Cache {
         try {
             value = valueLoader.call();
         } catch (Exception ex) {
-            throw new ValueRetrievalException(key, valueLoader, ex);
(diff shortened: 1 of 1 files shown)


mesutcelik:
run-lub-run

mesutcelik:
verify

mesutcelik:
Thanks @snicoll ! Can we have a forward-port to master branch?

snicoll:
@mesutcelik what do you mean? Another PR with a cherry-pick on master?

mesutcelik:
yes.

snicoll:
@mesutcelik it looks like @emrahkocaman already did that...

mesutcelik:
yep. Thanks to both of you...

PR #9149, opened 2016-10-20
Refactoring of the MapQueryEngine required by the fast-aggregations and projections

tombujok:
The problem:
In the previous approach the MapQueryEngine contained the following:
- logic how to run a query (full-partition-scan or index query)
- logic scheduling query operations 
- logic orchestrating queries (first query thread, later partition thread)

So, the proxies like a MapProxy used MapQueryEngine to schedule a query. In this call the MapQueryEngine dispatched operations which called (surprise, surprise) the MapQueryEngine to run the query logic. This was a mess and a lot of spaghetti...

There was also a lot of inconsistencies how difference scenarios of queries were run, e.g. query-optimizations, or queries with or without a PagingPredicate.
Finally, there was a big mess when it comes to the method names, to name some fancy ones: queryLocationPartition, queryTheLocalPartition, queryLocalPartitions.

Currently there are 3 entities separating main concerns:
- MapQueryRunner -> (local and parallel) runs the query logic in the caller thread.
- MapQueryDispatcher -> dispatches query operations (QueryOperation and QueryPartitionOperation). Does not know about the queryRunner and queryDispatcher.
- MapQueryEngine -> orchestrates the query by dispatching the query operations using the MapQueryDispatcher and merges thee result. (does not know about the queryRunner)

In this way there's no cycles and spaghetti bindings. It's also possible to read the code...

Furthermore:
- in the new approach there's no special path for PagingPredicate and non-paging predicate. Most of the stuff has been abstracted away and generalized.
- It's been fixed that the queries executed on the partitionThreads are optimized now (earlier the optimisation was not executed)

Important:

- This refactoring is the first one of a series.  It's not meant to fix all the problems. In my opinion it brings the query engine code to a next level.
- The result of this refactoring is consistent and does not limit us in any way.
- It opens possibilities to refactor further. 

I'll be working on fast-aggregations and projections next, thus the MapQueryEngine will be refactored further. Test coverage will also significantly increase.

The commits will be squashed at merge.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
@@ -39,40 +39,40 @@ import static com.hazelcast.util.SortingUtil.getSortedSubList;
 /**
  * Specialization of the {@link MapLocalQueryRunner} - the only difference is that
  * the query evaluation per partition is run in a PARALLEL fashion.
- *
+ * <p>
  * Runs query operations in the calling thread (thus blucking it)
  * Query evaluation per partition is run in a PARALLEL fashion.
  * <p>
  * Used by query operations only: QueryOperation & QueryPartitionOperation
  * Should not be used by proxies or any other query related objects.
  */
-public class MapLocalParallerQueryRunner extends MapLocalQueryRunner {
+public class MapLocalParallelQueryRunner extends MapLocalQueryRunner {
 
-    protected static final int QUERY_EXECUTION_TIMEOUT_MINUTES = 5;
+    private static final int QUERY_EXECUTION_TIMEOUT_MINUTES = 5;
 
     private final ManagedExecutorService executor;
     private final int timeoutInMinutes;
 
-    public MapLocalParallerQueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer,
+    public MapLocalParallelQueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer,
                                        ManagedExecutorService executor, int timeoutInMinutes) {
         super(mapServiceContext, optimizer);
         this.executor = executor;
         this.timeoutInMinutes = timeoutInMinutes;
     }
 
-    public MapLocalParallerQueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer,
+    public MapLocalParallelQueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer,
                                        ManagedExecutorService executor) {
         this(mapServiceContext, optimizer, executor, QUERY_EXECUTION_TIMEOUT_MINUTES);
     }
 
     @Override
-    protected Collection<QueryableEntry> runPartitionScanQuery(
+    protected Collection<QueryableEntry> runUsingPartitionScan(
             String mapName, Predicate predicate, Collection<Integer> partitions) {
         try {
             if (predicate instanceof PagingPredicate) {
-                return runFullPartitionScanQueryWithPaging(mapName, (PagingPredicate) predicate, partitions);
+                return runUsingPartitionScanWithPaging(mapName, (PagingPredicate) predicate, partitions);
             } else {
-                return runFullPartitionScanQueryWithoutPaging(mapName, predicate, partitions);
+                return runUsingPartitionScanWithoutPaging(mapName, predicate, partitions);
             }
         } catch (InterruptedException e) {
             throw new HazelcastException(e.getMessage(), e);
@@ -81,16 +81,16 @@ public class MapLocalParallerQueryRunner extends MapLocalQueryRunner {
         }
     }
 
-    private List<QueryableEntry> runFullPartitionScanQueryWithPaging(
+    private List<QueryableEntry> runUsingPartitionScanWithPaging(
             String name, PagingPredicate predicate, Collection<Integer> partitions)
             throws InterruptedException, ExecutionException {
 
-        List<QueryableEntry> result = runFullPartitionScanQueryWithoutPaging(name, predicate, partitions);
+        List<QueryableEntry> result = runUsingPartitionScanWithoutPaging(name, predicate, partitions);
         Map.Entry<Integer, Map.Entry> nearestAnchorEntry = getNearestAnchorEntry(predicate);
         return getSortedSubList(result, predicate, nearestAnchorEntry);
     }
 
-    private List<QueryableEntry> runFullPartitionScanQueryWithoutPaging(
+    private List<QueryableEntry> runUsingPartitionScanWithoutPaging(
             String name, Predicate predicate, Collection<Integer> partitions)
             throws InterruptedException, ExecutionException {
 
@@ -123,7 +123,7 @@ public class MapLocalParallerQueryRunner extends MapLocalQueryRunner {
 
         @Override
         public Collection<QueryableEntry> call() throws Exception {
-            return runPartitionScanQueryOnSinglePartition(name, predicate, partition);
+            return runUsingPartitionScanOnSinglePartition(name, predicate, partition);
         }
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
@@ -48,7 +48,7 @@ import static com.hazelcast.util.SortingUtil.getSortedSubList;
  */
 public class MapLocalParallelQueryRunner extends MapLocalQueryRunner {
 
-    private static final int QUERY_EXECUTION_TIMEOUT_MINUTES = 5;
+    protected static final int QUERY_EXECUTION_TIMEOUT_MINUTES = 5;
 
     private final ManagedExecutorService executor;
     private final int timeoutInMinutes;
@@ -81,7 +81,7 @@ public class MapLocalParallelQueryRunner extends MapLocalQueryRunner {
         }
     }
 
-    private List<QueryableEntry> runUsingPartitionScanWithPaging(
+    protected List<QueryableEntry> runUsingPartitionScanWithPaging(
             String name, PagingPredicate predicate, Collection<Integer> partitions)
             throws InterruptedException, ExecutionException {
 
@@ -90,7 +90,7 @@ public class MapLocalParallelQueryRunner extends MapLocalQueryRunner {
         return getSortedSubList(result, predicate, nearestAnchorEntry);
     }
 
-    private List<QueryableEntry> runUsingPartitionScanWithoutPaging(
+    protected List<QueryableEntry> runUsingPartitionScanWithoutPaging(
             String name, Predicate predicate, Collection<Integer> partitions)
             throws InterruptedException, ExecutionException {
 
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
@@ -40,7 +40,7 @@ import static com.hazelcast.util.SortingUtil.getSortedSubList;
  * Specialization of the {@link MapLocalQueryRunner} - the only difference is that
  * the query evaluation per partition is run in a PARALLEL fashion.
  * <p>
- * Runs query operations in the calling thread (thus blucking it)
+ * Runs query operations in the calling thread (thus blocking it)
  * Query evaluation per partition is run in a PARALLEL fashion.
  * <p>
  * Used by query operations only: QueryOperation & QueryPartitionOperation
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.map.impl.query;
+
+import com.hazelcast.core.HazelcastException;
+import com.hazelcast.map.impl.MapServiceContext;
+import com.hazelcast.query.PagingPredicate;
+import com.hazelcast.query.Predicate;
+import com.hazelcast.query.impl.QueryableEntry;
+import com.hazelcast.query.impl.predicates.QueryOptimizer;
+import com.hazelcast.util.executor.ManagedExecutorService;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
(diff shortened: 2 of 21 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    protected Collection<QueryableEntry> runPartitionScanQuery(
+            String mapName, Predicate predicate, Collection<Integer> partitions) {
+        try {
+            if (predicate instanceof PagingPredicate) {
+                return runParallelFullPartitionScanQueryWithPaging(mapName, (PagingPredicate) predicate, partitions);
+            } else {
+                return runParallelFullPartitionScanQueryWithoutPaging(mapName, predicate, partitions);
```
pveentjer: Do we still need the 'runParallel' prefix? Since this is already the parallel runner.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
```diff
@@ -0,0 +1,317 @@
...
+        this.queryResultSizeLimiter = new QueryResultSizeLimiter(mapServiceContext, logger);
+        this.queryOptimizer = optimizer;
+        this.operationService = nodeEngine.getOperationService();
+        this.clusterService = nodeEngine.getClusterService();
+        this.localMapStatsProvider = mapServiceContext.getLocalMapStatsProvider();
+        this.executor = nodeEngine.getExecutionService().getExecutor(QUERY_EXECUTOR);
+    }
+
+    // full query = index query (if possible), then partition-scan query
+    public QueryResult runFullQuery(String mapName, Predicate predicate, IterationType iterationType)
```
pveentjer: Would it not be easier if we introduce a Query object here? Passing these parameters make the whole thing more complex; especially if new parameters need to be added.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
```diff
@@ -16,122 +16,21 @@
...
-     *
-     * @param mapName   the name of the map
-     * @param predicate the predicate
-     * @return the QueryResult
-     * @throws ExecutionException
-     * @throws InterruptedException
-     */
-    QueryResult queryLocalPartitions(String mapName, Predicate predicate, IterationType iterationType)
-            throws ExecutionException, InterruptedException;
+    Set runQueryOnAllPartitions(String mapName, Predicate predicate, IterationType iterationType, boolean uniqueResult);
```
pveentjer: Would it not be easier if:
1: introduce a query object which contains what you actually want to do. Like single partition, multiple partitions etc. Also when new parameters are added; with a query the interface won't break that quick
2: make it async. This way we don't block the generic thread.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
tombujok: Agreed.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
tombujok: Probably so. I'll be adding fast-aggregations and projections in the next couple of days, thus I would not like to optimize the API prematurely. Adding those will be a great validation of the API and will definitely result in its changes. Other hints appreciated.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngine.java
tombujok: Ad.1 Plz see comment below. Will be addressed soon.
Ad.2 Plz see the general comment. I agree it should be async, but this will be addresses after the projections and fast-aggregations.

tombujok:
The commits will be squashed at merge.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
```diff
@@ -0,0 +1,130 @@
...
+ * Specialization of the {@link MapLocalQueryRunner} - the only difference is that
+ * the query evaluation per partition is run in a PARALLEL fashion.
+ *
+ * Runs query operations in the calling thread (thus blucking it)
+ * Query evaluation per partition is run in a PARALLEL fashion.
+ * <p>
+ * Used by query operations only: QueryOperation & QueryPartitionOperation
+ * Should not be used by proxies or any other query related objects.
+ */
+public class MapLocalParallerQueryRunner extends MapLocalQueryRunner {
```
pveentjer: A few notes:

- This class takes care of assigning a thread to the actual query so it can get processed. So it is comparable to the OperationExecutor and not the OperationRunner.
- Both OperationRunner and OperationExecutor imply it is local. 

So should this class not be called ParallelMapQueryExecutor?

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
```diff
@@ -0,0 +1,130 @@
...
+        this.timeoutInMinutes = timeoutInMinutes;
+    }
+
+    public MapLocalParallerQueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer,
+                                       ManagedExecutorService executor) {
+        this(mapServiceContext, optimizer, executor, QUERY_EXECUTION_TIMEOUT_MINUTES);
+    }
+
+    @Override
+    protected Collection<QueryableEntry> runPartitionScanQuery(
```
pveentjer: I think we can drop 'query' from the name of the method since query execution is already implied by the name of the class.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
```diff
@@ -0,0 +1,317 @@
...
+        }
+        // else: if fallback to full table scan also failed to return any results due to migrations,
+        // then return empty result set without any partition IDs set (so that it is ignored by callers).
+
+        updateStatistics(mapContainer);
+
+        return result;
+    }
+
+    protected Collection<QueryableEntry> runIndexQuerySafely(
```
pveentjer: I would drop 'query' from all the method names here.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
```diff
@@ -0,0 +1,317 @@
...
+        // Also make sure there are no long migrations in flight which may have started after starting the query
+        // but not completed yet.
+        if (isResultSafe(initialPartitionStateVersion)) {
+            return entries;
+        } else {
+            return null;
+        }
+    }
+
+    protected Collection<QueryableEntry> runIndexQuery(Predicate predicate, MapContainer mapContainer) {
```
pveentjer: runUsingIndex e.g.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
```diff
@@ -0,0 +1,317 @@
...
+        // Also make sure there are no long migrations in flight which may have started after starting the query
+        // but not completed yet.
+        if (isResultSafe(initialPartitionStateVersion)) {
+            return entries;
+        } else {
+            return null;
+        }
+    }
+
+    protected Collection<QueryableEntry> runPartitionScanQuery(
```
pveentjer: runUsingPartitionScan e.g.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
```diff
@@ -0,0 +1,317 @@
...
+    protected final NodeEngine nodeEngine;
+    protected final ILogger logger;
+    protected final QueryResultSizeLimiter queryResultSizeLimiter;
+    protected final InternalSerializationService serializationService;
+    protected final IPartitionService partitionService;
+    protected final QueryOptimizer queryOptimizer;
+    protected final OperationService operationService;
+    protected final ClusterService clusterService;
+    protected final LocalMapStatsProvider localMapStatsProvider;
+    protected final ManagedExecutorService executor;
```
pveentjer: Perhaps I'm missing something, but why does this class have an executor which is only used by its subclass?

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -16,707 +16,215 @@
...
-        this.parallelEvaluation = nodeEngine.getProperties().getBoolean(QUERY_PREDICATE_PARALLEL_EVALUATION);
         this.executor = nodeEngine.getExecutionService().getExecutor(QUERY_EXECUTOR);
+        this.queryDispatcher = new MapQueryDispatcher(mapServiceContext);
     }
 
-    QueryResultSizeLimiter getQueryResultSizeLimiter() {
-        return queryResultSizeLimiter;
-    }
-
+    // invoked from proxy layer
```
pveentjer: Is this comment not implied by the purpose of the NodeEngine? It being the class called from the proxies.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
pveentjer: After studying the sources a bit longer; I think there are 2 concerns conflated into a single class hierarchy.

You have the concerns:
-  Query Running; so actually chewing through the data with or without index and all that.
-  Query Threading; assigning thread to a query to get processed.

In the OperationService these 2 concerns are separated into 2 classes:
- OperationExecutor
- OperationRunner

Perhaps I'm missing something since processing is quite complex, but would it not make sense to do something similar here.

pveentjer:
It looks good. Good to see the QueryEngine is getting some overdue TLC. Since  this PR isn't the final state of the QueryEngine, but an intermediate state, I'm fine with the changes. Please have a careful look at the review.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
tombujok: Good catch.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryDispatcher.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.concurrent.Future;
+
+import static com.hazelcast.cluster.memberselector.MemberSelectors.DATA_MEMBER_SELECTOR;
+import static com.hazelcast.map.impl.query.MapQueryEngineUtils.shouldSkipPartitionsQuery;
+import static com.hazelcast.spi.ExecutionService.QUERY_EXECUTOR;
+import static com.hazelcast.util.ExceptionUtil.rethrow;
+import static java.util.Collections.singletonList;
+
+/**
+ * Dispatches query operations.
```
tombujok: Added

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
tombujok: Thx. That's a good tip indeed.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
tombujok: Done here too.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
tombujok: Done

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
tombujok: Done

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalQueryRunner.java
tombujok: Managed to remove the same field from the MapQueryEngine too.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallerQueryRunner.java
tombujok: I agree there's two concerns there but I think you slightly missed the point by saying `This class takes care of assigning a thread to the actual query`. The runner does not assign a thread to run a query. The query consists of a couple of phases:

- run index query
- run full-partition scan

In the MapLocalQueryRunner all phases are run within the calling thread.
In the MapLocalParallelQueryRunner most of the stuff hapens in the calling thread, apart from the full-partitionScan predicate evaluation - this is delegated to the executor. However, this is encapsulated and the caller does not know about it since the call is blocking anyway.

In order to do what you are suggesting I would have to create the following classes and then use these classes from within the LocalQueryRunner.

- PartitionScanRunner (just have the logic of partition scanning)
- PartitionScanExecutor (runs the partitions can logic - either in the same thread or in the executor)

This comment is appreciated and I see your point, but it looks a bit like an overkill for now since there's only two implementations and their code is very short. 
Till now everything happened in one class, so I guess we can postpone this discussion and create even more fine-grained design on the way.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
tombujok: Don't get what you mean,but the purpose of this comment was to say that this class should be only used from the top-level proxy layer (like MapProxy, etc.) I've adjusted the javadoc

pveentjer:
@tombujok is this going to cause any EE problems? Otherwise it should just be merged

tombujok:
@pveentjer Thx for the review! Yeah, working on EE version. The build will fail, so plz wait for the EE PR.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
```diff
@@ -0,0 +1,130 @@
...
+
+import static com.hazelcast.map.impl.query.MapQueryEngineUtils.waitForResult;
+import static com.hazelcast.query.PagingPredicateAccessor.getNearestAnchorEntry;
+import static com.hazelcast.util.SortingUtil.getSortedSubList;
+
+/**
+ * Specialization of the {@link MapLocalQueryRunner} - the only difference is that
+ * the query evaluation per partition is run in a PARALLEL fashion.
+ * <p>
+ * Runs query operations in the calling thread (thus blucking it)
```
jerrinot: typo. `blucking`

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapQueryEngineImpl.java
```diff
@@ -16,615 +16,158 @@
...
+        QueryResult result = doRunQueryOnQueryThreads(
+                mapName, predicate, iterationType, mutablePartitionIds, ALL_MEMBERS);
+        if (isResultFromAnyPartitionMissing(mutablePartitionIds)) {
+            doRunQueryOnPartitionThreads(mapName, predicate, iterationType, mutablePartitionIds, result);
         }
 
         return result;
     }
 
+    // partition thread ONLY (for now)
```
jerrinot: what does it mean?

tombujok: It means that this operation runs queries on partition threads only, bypassing the query threads.

hazelcast/src/main/java/com/hazelcast/map/impl/query/MapLocalParallelQueryRunner.java
tombujok: Fixed

PR #9151, opened 2016-10-21
Set other exceptions than execution-exceptions to fine level

ahmetmircik:
Otherwise there will be so much noise, especially when running tests. Our main goal was logging of execution exceptions, only keeping them severe is sufficient.


ahmetmircik:
tiny change, merging it.

PR #9155, opened 2016-10-24
Classes converted to IdentifiedDataSerializable

vbekiaris:
Remaining classes converted to IdentifiedDataSerializable.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/WriteBehindStateHolder.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/WriteBehindStateHolder.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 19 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/operation/WriteBehindStateHolder.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+
+/**
+ * Holder for write-behind-specific state.
+ */
+public class WriteBehindStateHolder implements IdentifiedDataSerializable {
```
pveentjer: Why don't we rename this class to WriteBehindState?

hazelcast/src/main/java/com/hazelcast/query/impl/IndexImpl.java
```diff
@@ -165,7 +166,7 @@ public boolean isOrdered() {
         return indexStore.getRecordMap(indexValue);
     }
 
-    public static final class NullObject implements Comparable, DataSerializable {
+    public static final class NullObject implements Comparable, IdentifiedDataSerializable {
```
pveentjer: I don't know how frequently this object is being send, but perhaps we could make a cached version in the corresponding factory?

pveentjer:
:+1: 

My comments are not directly related to your code since the issues were already there. I just spotted them due to the changes

PR #9157, opened 2016-10-24
Use Clock.currentTimeMillis() in logged timeout messages

vbekiaris:
Use `Clock.currentTimeMillis()` in messages logged in case of call timeout

Since the cluster clock does not follow system clock, currently we may
produce confusing output as the logged total elapsed time may appear
to be significantly different than expected.


pveentjer:
:+1: 

I would be happier if we just get rid of the Clock abstraction completely

PR #9159, opened 2016-10-24
Add support for handling Jet messages

cangencer:
Introduce a hook and a packet flag for dispatching Jet protocol
packets.

This will enable Jet to use Hazelcast connections, instead of creating its own separate network stack.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
@@ -98,6 +98,8 @@ import static java.lang.System.currentTimeMillis;
 @SuppressWarnings("checkstyle:classdataabstractioncoupling")
 public class NodeEngineImpl implements NodeEngine {
 
+    private static final String JET_SERVICE_NAME = "hz:impl:jetService";
+
     private final Node node;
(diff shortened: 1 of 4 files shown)


pveentjer:
Minor comment. Looks good

pveentjer:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/NodeEngineImpl.java
```diff
@@ -135,14 +137,35 @@ public NodeEngineImpl(final Node node) {
...
+        return new PacketHandler() {
+
+            private volatile PacketHandler handler;
+
+            @Override
+            public void handle(Packet packet) throws Exception {
+                if (handler == null) {
+                    handler = serviceManager.getService(JET_SERVICE_NAME);
+                    if (handler == null) {
+                        throw new UnsupportedOperationException("Jet is not registered on this node");
```
cangencer: it should be caught and handled by `PacketDispatcherImpl.dispatch()`

pveentjer: at the moment jet packets enter the system and you have no appropriate handler.. what do you want to do otherwise?

PR #9162, opened 2016-10-24
Optimized Near Cache initialization in ClientReplicatedMapProxy

Donnerbart:
- Moved initialization of Near Cache to `onInitialize()` method, instead of checking this on each `get()`

The goal of this PR is to have the Near Cache for `ReplicatedMap` on Hazelcast clients being created when `hazelcastInstance.getReplicatedMap("myMap")` is called, not lazy on the first `replicatedMap.get(0)` call.

If there is no good reason to init the Near Cache lazy, I would like to align this behavior with the other proxies. Otherwise I will have a hard time to fetch the Near Cache upfront for the unified Near Cache tests, before I start the test itself. I also cannot just call something like `replicatedMap.get(0)`, since this will affect the Near Cache size and stats, which are checked in some tests.

The only benefit I can imagine is if you are just putting data into a `ReplicatedMap` via a Hazelcast client. In this case the old implementation would never create the Near Cache instance and would safe some KB of memory for the instances and empty data structure. On the other hand we might experience a very late failure scenario, when the Near Cache is created (maybe after hours of usage) on the first `get()` operation. So I think we should sacrifice a little bit of memory to fail fast and have the same initialization pattern in all Near Cache proxies.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientReplicatedMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientReplicatedMapProxy.java
@@ -65,7 +65,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.hazelcast.cache.impl.nearcache.NearCache.NULL_OBJECT;
 import static com.hazelcast.util.Preconditions.checkNotNull;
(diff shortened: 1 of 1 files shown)


jerrinot:
looks good to me. the proxy creation might require extra network hop now: to register invalidation listener. I am not sure if what's a problem or not.

Donnerbart:
run-lab-run

Donnerbart:
In my opinion it's also better to have this additional lag in the `getReplicatedMap()` call before you actually use the map. This may otherwise cause a small latency peak on the first `get()` operation.

But you might give a good hint for a more severe issue: since we have an invalidation listener, this change may cause more events being sent to the client now, since the Near Cache is directly created including the listener. If the replicated map is anyway used to `get()` data from the cluster, there will be no big difference for the users, but if the most common use case is to just `put()` data into the cluster, this might cause some overhead.

PR #9168, opened 2016-10-25
[BACKPORT] Increase threshold for local wall-clock drift detection

jerrinot:
Too small threshold means a system can be fooled by a lengthy GC pause
Port of f651b11fe0 from 3.7.2 release branch into maintanance


pveentjer:
Not part of this pr. Just some brainfarting.

Can't we determine clock drift based on nano time instead of System.currentTimeMillis because nanotime will at least not be prone to changes in the system clock. 

For example when there is a 5 minute jump in the System clock, but the nano time has not jumped that much, then you know there is a drift. But if there would be a 5 minute gc, then both system time and nanotime would have increased 5 minutes.

pveentjer:
:+1:

PR #9171, opened 2016-10-25
Cleanup of NearCacheContext creation on Hazelcast clients

Donnerbart:
* cached `NearCacheContext` for Hazelcast clients in `HazelcastClientInstanceImpl`
* used cached `NearCacheContext` in `AbstractClientInternalCacheProxy`, `ClientReplicatedMapProxy` and `NearCachedClientMapProxy`
* replaced `NearCacheManager` with `NearCacheContext` in `ClientContext`

This aligns the behavior on Hazelcast clients with Hazelcast members, which use the same `NearCacheContext` instance from the `NearCacheProvider` for all Near Caches. This also ensures that the `ClassLoader` from the `ClientConfig` is passed into the `NearCacheContext` in all usage scenarios.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.client.impl;
 
+import com.hazelcast.cache.impl.nearcache.NearCacheContext;
 import com.hazelcast.cache.impl.nearcache.NearCacheManager;
 import com.hazelcast.cardinality.impl.CardinalityEstimatorService;
 import com.hazelcast.client.ClientExtension;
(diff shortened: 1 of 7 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
```diff
@@ -222,6 +224,8 @@ public HazelcastClientInstanceImpl(ClientConfig config,
         listenerService = initListenerService();
         userContext = new ConcurrentHashMap<String, Object>();
         nearCacheManager = clientExtension.createNearCacheManager();
+        nearCacheContext = new NearCacheContext(serializationService, executionService, nearCacheManager,
+                config.getClassLoader());
```
Donnerbart: I've added the `ClassLoader` here, since I could not see any negative effect of doing so. We have dealt with a `null` instance here before, so the critical change would be if the `ClassLoader` is actually set.

pveentjer:
:+1:

PR #9173, opened 2016-10-25
Add toString into PriorityPartitionSpecificRunnable

jerrinot:
This is useful for troubleshooting as otherwise slow operation detector
can see just `com.hazelcast.client.impl.ClientEngineImpl$PriorityPartitionSpecificRunnable`


pveentjer:
:+1:

PR #9175, opened 2016-10-26
Padding option for InternalSerializationService

pveentjer:
The padding option can be used to reserve extra bytes in front of the actual payload. This
space can be used for putting e.g. header information.

This functionality is required to get rid of all kinds of data wrappers like Response/Packet. Instead of creating these wrapper objects with their own fields, a byte-array can be reserved with sufficient space so the Response/Packet information can be written in front of the actual payload.


asimarslan:
@pveentjer Is this PR change binary compatibility of serialized data? Can an old client handle it?

pveentjer:
There is no binary change. It is purely an optimization; it will not change what goes over the wire.

If you get some bytes from a Packet; and then from an byte-array, or if you immediately get them from a byte-array.. its is the same bytes :)

asimarslan:
👍

Issue #9172, opened 2016-10-25
putToCacheAndGetInvalidationEventWhenNodeShutdown test randomly fails


ihsandemir:
putToCacheAndGetInvalidationEventWhenNodeShutdown test simulates member shutting down using the code:
((LifecycleServiceImpl) instanceToShutdown.getLifecycleService())
                .fireLifecycleEvent(LifecycleEvent.LifecycleState.SHUTTING_DOWN);

The cache invalidation listener is CacheEventHandler. The Connection manager also listens for this event. The order in which these listeners are called changes the result of the test.

Fail case: Connection manager listener called first closing the connections, and then the CacheEventHandler listener is called but the cache invalidation can not be delivered since the connection was closed.

Pass case: CacheEventHandler listener called first and then the Connection manager listener.

A possible solution is to invalidate near caches at the client side when CLIENT_DISCONNECTED (us client lifecycle listener).

For the maintenance-3.x, the corresponding tests are:
putToCacheAndGetInvalidationEventWhenNodeShutdownWithBinaryInMemoryFormat
putToCacheAndGetInvalidationEventWhenNodeShutdownWithObjectInMemoryFormat

ihsandemir:
The test is put to ignore with PR https://github.com/hazelcast/hazelcast/pull/8848 . Hence, lets not forget to reopen the test with the fix PR.

sancar:
Fixed by https://github.com/hazelcast/hazelcast/commit/68b6547aedcce5d32883f7acfeb132e0d66e2990

PR #9177, opened 2016-10-26
The client may not invalidate near caches on cluster disconnection - putToCacheAndGetInvalidationEventWhenNodeShutdown test randomly fails

ihsandemir:
Fix for issue https://github.com/hazelcast/hazelcast/issues/9172

Invalidates the near caches when the client is disconnected from the cluster to avoid the usage of stale near cache data.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
@@ -225,6 +225,7 @@ public abstract class ClusterListenerSupport implements ConnectionListener, Conn
                     @Override
                     public void run() {
                         try {
+                            client.getNearCacheManager().clearAllNearCaches();
                             connectToCluster();
                         } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


pveentjer:
If I understand it correctly.

Imagine a client has e.g. 10gb of data in its near cache and its connected to a group of members and the connection to one of these members fails, it will wipe out the complete 10gb of nearcache data?

jerrinot:
Is this needed? AFAIK the changes @ahmetmircik is working on will deal with this situation (in a more granular form)

asimarslan:
👍

asimarslan:
Client continues to operate from near cache in case of a disconnect from cluster but this fix might change that .
+1 @jerrinot

pveentjer:
Wiping out the near cache has other consequences. If multiple clients experience a disconnect at roughly the same time, and eventually the connection is restored; they cause a huge load on the cluster.

ihsandemir:
Decided to ignore the test for now (at this PR: https://github.com/hazelcast/hazelcast/pull/8848), until a better fix for the issue is submitted.

PR #9179, opened 2016-10-26
OSS changes for Hot Restart IP address restriction removal

mdogan:
Changes in join mechanism and partition service required to remove IP address restriction in Hot Restart.

See design doc for details:

https://hazelcast.atlassian.net/wiki/display/EN/Copy+HRS+Data+from+a+Source+to+a+Target+Cluster+Design

EE counterpart: https://github.com/hazelcast/hazelcast-enterprise/pull/1101

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
@@ -17,6 +17,7 @@
 package com.hazelcast.internal.cluster.impl;
 
 import com.hazelcast.cluster.ClusterState;
+import com.hazelcast.core.Member;
 import com.hazelcast.instance.BuildInfo;
 import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.instance.Node;
@@ -211,7 +212,7 @@ public class ClusterJoinManager {
             }
 
             final Address target = joinRequest.getAddress();
-            if (checkClusterStateBeforeJoin(target)) {
+            if (checkClusterStateBeforeJoin(target, joinRequest.getUuid())) {
                 return;
             }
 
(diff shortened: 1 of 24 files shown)


hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -666,6 +659,11 @@ private void finalizeOrRollbackMigration(PartitionRuntimeState partitionState) {
         }
 
         updateAllPartitions(partitionTable);
+        partitionStateManager.setVersion(partitionState.getVersion());
+        if (!partitionStateManager.setInitialized()) {
```
mmedenjak: Is there a chance that the partition state manager is not initialized? Can we not just call `node.getNodeExtension().onPartitionStateChange();`?

mdogan: `partitionStateManager.setInitialized()` already calls `node.getNodeExtension().onPartitionStateChange()` if it sets `initialized=true`. This is just to remove 2nd redundant `node.getNodeExtension().onPartitionStateChange()` call. But I can remove `if` check completely. It doesn't make so much diff performance-wise.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionStateManager.java
```diff
@@ -307,4 +315,30 @@ void reset() {
...
+        }
+        int count = 0;
+        for (InternalPartitionImpl partition : partitions) {
+            if (partition.replaceAddress(oldAddress, newAddress) > -1) {
+                count++;
+            }
+        }
+        node.getNodeExtension().onPartitionStateChange();
+        if (count > 0) {
+            node.getNodeExtension().onPartitionStateChange();
```
mdogan: good catch 👍

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
```diff
@@ -233,25 +235,46 @@ private void executeJoinRequest(JoinRequest joinRequest, Connection connection)
...
+                        + " a known passive member.");
+
                 return true;
             }
+
+            return false;
         }
-        return false;
+
+        if (clusterService.isMemberRemovedWhileClusterIsNotActive(uuid)) {
```
metanet: Yes, a member should leave the cluster and join back with the same uuid but a completely new address. 2 members cannot switch their addresses and join back. It is too much fantasy :)

mdogan: - When cluster is active, we allow anyone to join.
- When cluster is not active, we only allow only known members (members those are removed while not active) to join. A member will be known if;
    * either its address is known, then we expect that;
        - either its uuid is unknown 
        - or address mapped to that uuid is equal to its address, means its returning using the same address and uuid
    * or only its uuid is known

For example, assume `[A, Uuid-1]`, `[B, Uuid-2]` are known when cluster is not active. Then we allow;
- `[A, Uuid-1]` 
- `[B, Uuid-3]`
- `[C, Uuid-1]`

to join. But we don't allow;

- `[A, Uuid-2]`
- `[B, Uuid-1]` 
to join.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
mmedenjak: Ok. I was actually wondering if at this point where we are applying new partition state the partition state manager could still be uninitialized or is the partition state manager initialized from earlier in the node lifecycle.

mdogan: `partitionStateManager` may be uninitialized at this point on non-master nodes. That means this is the initial partition assignment.

mdogan:
Most probably unrelated failure:
`com.hazelcast.client.listeners.QueueItemListenerOnReconnectTest.testListenersNonSmartRoutingMultipleServer`

Also, the same test failed on a different PR build (https://github.com/hazelcast/hazelcast/pull/9234);
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/5653/
`com.hazelcast.client.listeners.SetItemListenerOnReconnectTest.testListenersNonSmartRoutingMultipleServer`

run-lab-run

mdogan:
run-lab-run

mmedenjak:
@mdogan @metanet Can we get this merged? So that https://github.com/hazelcast/hazelcast/pull/9203 can get rebased

PR #9180, opened 2016-10-26
Networking package

pveentjer:
Moved all shared networking logic between client and server into the com.hazelcast.internal.networking package. This way it is a lot easier to understand what is member specific and what is generic.

This PR doesn't contain any other change than moving of classes to different packages.


Donnerbart:
I think it will be useful to set this checkbox for commits like this:
screenshot from 2016-10-26 15-13-42
It will just optimize the imports of the changed files, nothing more. Otherwise we'll have that screwed up import order in a huge number of classes.

If you prefer to commit from the command line, you can trigger an optimize just for changed classes per module via the context menu and this checkbox:
screenshot from 2016-10-26 15-17-44

pveentjer:
Imports optimized.

PR #9181, opened 2016-10-26
Fixed Near Cache usage in TransactionalMapProxySupport

Donnerbart:
Added proper support for the Near Cache to `getInternal()` and put fetched values into the Near Cache (including stale read prevention).

When extending my unified Near Cache tests to the `TransactionalMap` I realised that the proxy is missing some functionality in the `get()` code path. Actually this is the only code path with uses the Near Cache at all, but nevertheless it should support all Near Cache features. So I copied the logic from the `IMap` proxy.

I think good reviewers would be @ahmetmircik, @vbekiaris or @gurbuzali.


Donnerbart:
Hmm, I think for now I will not try to add the full test-suite for the `TransactionalMap`, so we don't need to add a full support in that proxy. I'll just try to test the "fetch data from the Near Cache" part.

PR #9183, opened 2016-10-26
Hazelcast listeners are lost during cluster disconnects - Backport PR #8848

ihsandemir:
Fixes the client listener registration if the client owner connection is lost and reconnected (e.g. due to heartbeat or network problems).

Authentication response returns a list of members at which the client is unregistered so that the listener service acts upon this list. The smart listener service re-registers the listeners to these unregistered members. The non-smart listener service re-registers the listeners only if the reconnected member is not the same as the previous member or the previous member client resources were cleaned up.

The ClientDisconnectionOperation also acts smart so that it only cleans client resources if the dead member uuid is the current owner member uuid of the client and there is no alive connection endpoint to the client.

For compatibility, when client is connected to earlier versions the smart listener service re-registers all existing listeners.


ihsandemir:
run-lab-run

ihsandemir:
run-lab-run

alparslanavci:
Backport of #8848. Just mentioning here to connect the PRs.

hazelcast/src/main/java/com/hazelcast/client/impl/operations/ClientReAuthOperation.java
```diff
@@ -55,7 +55,7 @@ public void run() throws Exception {
 
     @Override
     public boolean returnsResponse() {
-        return Boolean.TRUE;
+        return Boolean.FALSE;
```
vbekiaris: just nitpicking here, this is unnecessary unboxing, we could just return the primitive `false` directly.

ihsandemir: OK, corrected.

hazelcast/src/main/java/com/hazelcast/util/FutureUtil.java
```diff
@@ -92,6 +92,23 @@ public void handleException(Throwable throwable) {
         }
     };
 
+    /**
+     * This ExceptionHandler rethrows {@link java.util.concurrent.ExecutionException}s and logs
+     * {@link com.hazelcast.core.MemberLeftException}s to the log.
+     */
+    public static final ExceptionHandler RETHROW_ALL_EXCEPT_MEMBER_LEFT = new ExceptionHandler() {
```
vbekiaris: This exception handler doesn't seem to have any usages, maybe remove it from the final PR?

hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ClientCompatibilityNullTest_1_0.java
```diff
@@ -5751,3 +5705,6 @@ public void handle(java.util.Collection<com.hazelcast.map.impl.querycache.event.
     }
 }
 
+
```
vbekiaris: newlines leftover

ihsandemir: This is a generated test, i need to remove them in its ftl file in client-protocol. It does not look urgent, hence i will just open an issue for this and correct ftl later.

ihsandemir:
run-lab-run

ihsandemir:
run-lab-run

ihsandemir:
run-lab-run

ihsandemir:
Some more details:

When the client for some reason loses the connection the member which is the owner for the client, it becomes CLIENT_DISCONNECTED and client reconnects to another member and then its lifecycle state becomes CLIENT_CONNECTED again. When this reconnection happen, it does re-register all the listeners to the members.

It is different when losing connection to non-owner member. If non owner connection is lost, no action needs to be taken in terms of listeners.

The listeners are registered to any new members which join the cluster.

Hence, you do not need to reregister any existing listeners since we are taking care of it at the background.

Issue #9081, opened 2016-10-11
Multicast socket exception if no network present.


neilstevenson:
If networking in the hazelcast.xml is set as,

```
    <network>
        <port>7000</port>
        <interfaces>
            <interface>127.0.0.1</interface>
        </interfaces>
    </network>
```

and there is only 127.0.0.1 available, on Hazelcast 3.7.2 the following exception appears

```
SEVERE: [127.0.0.1]:7000 [dev-configured-from-xml] [3.7.1] Can't assign requested address

java.net.SocketException: Can't assign requested address

at java.net.PlainDatagramSocketImpl.join(Native Method)

at java.net.AbstractPlainDatagramSocketImpl.join(AbstractPlainDatagramSocketImpl.java:178)

at java.net.MulticastSocket.joinGroup(MulticastSocket.java:323)

at com.hazelcast.internal.cluster.impl.MulticastService.createMulticastService(MulticastService.java:110)

at com.hazelcast.instance.Node.<init>(Node.java:191)

at com.hazelcast.instance.HazelcastInstanceImpl.createNode(HazelcastInstanceImpl.java:155)

at com.hazelcast.instance.HazelcastInstanceImpl.<init>(HazelcastInstanceImpl.java:126)

at com.hazelcast.instance.HazelcastInstanceFactory.constructHazelcastInstance(HazelcastInstanceFactory.java:218)

at com.hazelcast.instance.HazelcastInstanceFactory.newHazelcastInstance(HazelcastInstanceFactory.java:176)

at com.hazelcast.instance.HazelcastInstanceFactory.newHazelcastInstance(HazelcastInstanceFactory.java:126)

at com.hazelcast.core.Hazelcast.newHazelcastInstance(Hazelcast.java:87)

at XMLMain.main(XMLMain.java:7)
```

although the server does actually start.

Multicast without network isn't a real scenario, but it's the kind of thing you might use when doing development while travelling, and the exception is misleading.

It may IPV6 related, as using `System.setProperty("java.net.preferIPv4Stack", "true");` stops it occuring.

jerrinot:
@neilstevenson: what's your OS?

neilstevenson:
I'm on a Macbook, macOS Sierra, Java 1.8.0_101. This scenario comes up on training courses, which use 3.7.1 (my text says 3.7.2, typo).

jerrinot:
We use ipv4 as a default multicast group. This blows up when the only non-loopback interface is ipv6

[...]

PR #9190, opened 2016-10-27
Warn user when on localhost but loopback mode disabled.

tkountis:
Attempt to warn the end user, for potential auto-discovery issues as described in #9081, when the bind address is on the loopback interface, but loopback mode is disabled.

This is just a best effort attempt to early catch potential issues with mostly dev environments when not internet connected. As seen in #9081 even when on loopback mode, it could fail later when attempting to send a datagram.


tkountis:
Thanks @vbekiaris for reviewing. I double checked the code, and I believe the tense is correct, since the server socket is already bind. The multicast-service comes later in the flow.

tkountis:
@vbekiaris thanks for noticing the wrong tense.

tkountis:
run-lab-run

pveentjer:
:+1:

PR #9191, opened 2016-10-28
Adds option to reset StoreLatencyPlugin statistics

pveentjer:
An automatic reset option was added to the latency plugin that resets the statistics. 

There is no sliding window implementation; the current implementation is suboptimal but requested by Alparslan to give them some feedback what happened in a given time window. 

Due to the implementation it can happen that the reset period is 5 minutes and at 4:59 a very latency is measured of e..g 15 minutes.. then this 15 minutes is rendered once... and then it is gone. This is a known limitation of this approach since there is no sliding window.


pveentjer:
run-lab-run

PR #9192, opened 2016-10-28
Add counter for sync delivery failure

jerrinot:


pveentjer:
:+1:

hazelcast/src/main/java/com/hazelcast/spi/impl/eventservice/impl/EventServiceImpl.java
```diff
@@ -394,8 +396,11 @@ private void sendEvent(Address subscriber, EventEnvelope eventEnvelope, int orde
                     .setTryCount(SEND_RETRY_COUNT).invoke();
             try {
                 f.get(SEND_EVENT_TIMEOUT_SECONDS, TimeUnit.SECONDS);
-            } catch (Exception ignored) {
-                ignore(ignored);
+            } catch (Exception e) {
+                syncDeliveryFailureCount.inc();
```
pveentjer: Would it be to add another count; so the number of ones send?

jerrinot: you mean a counter of sync event?

pveentjer: A counter for the number of operations send. This way the failureCount gets a bit more context.

PR #9195, opened 2016-10-29
Minor InvocationRegistry cleanup

pveentjer:
The calculation of the concurrency level is now moved into the InvocationRegistry. It
is no longer a concern of the OperationServiceImpl. 

This makes the OperationServiceImpl a bit cleaner since it doesn't need to know about implementation details of the InvocationRegistry.

I believe this is an called an improved cohesion since letting the operationservice take care of a calculation that isn't its primary concern is causing low cohesion. And by moving it to the InvocationRegistry it increased cohesion since the OperationServiceImpl doesn't do something it doesn't care about; and it is moved to a place where it is important.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
@@ -99,9 +99,6 @@ import static java.util.concurrent.TimeUnit.SECONDS;
  */
 public final class OperationServiceImpl implements InternalOperationService, MetricsProvider, LiveOperationsTracker {
 
-    private static final int CORE_SIZE_CHECK = 8;
-    private static final int CORE_SIZE_FACTOR = 4;
-    private static final int CONCURRENCY_LEVEL = 16;
     private static final int ASYNC_QUEUE_CAPACITY = 100000;
     private static final long TERMINATION_TIMEOUT_MILLIS = SECONDS.toMillis(10);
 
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
```diff
@@ -148,22 +145,17 @@ public OperationServiceImpl(NodeEngineImpl nodeEngine) {
         this.nodeEngine = nodeEngine;
         this.node = nodeEngine.getNode();
         this.thisAddress = node.getThisAddress();
-        this.logger = node.getLogger(OperationService.class);
+        this.logger = node.loggingService.getLogger(OperationService.class);
```
mmedenjak: `Node#getLogger(java.lang.Class)` will do the same thing, forward the method call to `LoggingServiceImpl#getLogger(java.lang.Class)`. Is this change deliberate then? Should other calls to `Node#getLogger(java.lang.Class)` in this class be changed too?

pveentjer: I believe it should be changed. Node isn't responsible for providing an API to create a logger; that is already the concern of the LoggingService.

pveentjer: I'll remove it from this PR so it remains clean.

pveentjer:
Minor comment regarding logger has been resolved by reverting the log chance. So merging this one

PR #9197, opened 2016-10-30
TcpIpConnectionManager shutdown scheduler

### LINKED ISSUE #9187: scheduler in TcpIpConnectionManager not finished after instance.shutdown() called
scheduler in TcpIpConnectionManager not finished after instance.shutdown() called
private final ScheduledExecutorService scheduler = new ScheduledThreadPoolExecutor(4);

which causes problem in our spring boot environment - those threads run indefinitely and our aplication is not finished

how to reproduce with sample app:
1) run 4 instances java -jar hazelcast-test-1.0-SNAPSHOT.jar
2) every member is in clusetr
3) 1st instace - watch threads in jvisualvm
4) 2nd-4th kill them - (-9, no ctrl+c)
5) start another 3 instances

repeat 4) and 5) until pool-2-thread-1, pool-2-thread-2, pool-2-thread-3 and pool-2-thread-4 are created

now hit ctrl + c in console of 1st instace, all threads are finished except pool-2-thread-1, pool-2-thread-2, pool-2-thread-3 and pool-2-thread-4
they are eventually finised too, but only in this simplified example (after 60 seconds)

sample: HzlTest.java in https://github.com/vloo/hazelcast-test repo

pveentjer:
The internal scheduler was not shutdown on stop. Fix #9187

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
@@ -531,6 +531,7 @@ public class TcpIpConnectionManager implements ConnectionManager, PacketHandler 
         shutdownAcceptorThread();
         closeServerSocket();
         stop();
+        scheduler.shutdownNow();
         connectionListeners.clear();
     }
 


hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
```diff
@@ -512,6 +512,7 @@ public synchronized void stop() {
         connectionsMap.clear();
         monitors.clear();
         activeConnections.clear();
+        scheduler.shutdownNow();
```
mdogan: I think `scheduler` should be shutdown on `ConnectionManager.shutdown()`, not on `ConnectionManager.stop()`. Otherwise it should be restarted on `ConnectionManager.start()`.

pveentjer: I believe that currently there is already an issue with the threading model since it is shutdown in the stop method and it can't be restarted.

PR #9201, opened 2016-10-31
IO service simplification

pveentjer:
Removed IOService.getLogger, replaced by getLoggingService
Removed IOService.getThreadPrefix/getThreadGroup, replaced by getHazelcastThreadGoup
Removed IOService.toData/toObject. Not used

No Enterprise change needed.


pveentjer:
Trainwreck is simplified.

PR #9203, opened 2016-10-31
Implement join and UUID changes related to the partial-start feature

metanet:
* If Node A receives a join request from Node B, it accepts it only when both of these nodes are not excluded.
   - If Node B is excluded, it will fail with a BeforeJoinFailureOperation.
   - If Node A is excluded, it will silently ignore join request of Node B.

* Allow a node to change its UUID before the node start is completed. Therefore, Node.localMember reference is not final anymore. Still, It is cache-able after NodeExtension.isStartCompleted() returns true.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
@@ -221,6 +221,11 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
     }
 
     @Override
+    public String getThisUuid() {
+        return node.getThisUuid();
+    }
+
+    @Override
     public MemberImpl getLocalMember() {
         return node.getLocalMember();
     }
@@ -375,7 +380,7 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
                     return;
                 }
 
-                String localMemberUuid = node.getLocalMember().getUuid();
+                String localMemberUuid = node.getThisUuid();
                 String ownerUuid = endpoint.getPrincipal().getOwnerUuid();
                 if (localMemberUuid.equals(ownerUuid)) {
                     callDisconnectionOperation(endpoint);
--- a/hazelcast/src/main/java/com/hazelcast/config/HotRestartPersistenceConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/HotRestartPersistenceConfig.java
@@ -58,6 +58,8 @@ public class HotRestartPersistenceConfig {
     private int parallelism = DEFAULT_PARALLELISM;
     private int validationTimeoutSeconds = DEFAULT_VALIDATION_TIMEOUT;
     private int dataLoadTimeoutSeconds = DEFAULT_DATA_LOAD_TIMEOUT;
+    private HotRestartClusterDataRecoveryPolicy clusterDataRecoveryPolicy
+            = HotRestartClusterDataRecoveryPolicy.FULL_RECOVERY_ONLY;
 
     /**
      * Returns whether hot restart enabled on this member.
@@ -75,6 +77,28 @@ public class HotRestartPersistenceConfig {
      */
     public HotRestartPersistenceConfig setEnabled(boolean enabled) {
         this.enabled = enabled;
+        return this;
+    }
+
+    /**
+     * Returns the policy to be used when the cluster is started
+     *
+     * @return the policy to be used when the cluster is started
+     */
+    public HotRestartClusterDataRecoveryPolicy getClusterDataRecoveryPolicy() {
+        return clusterDataRecoveryPolicy;
+    }
+
+    /**
+     * Sets the policy to be used when the cluster is started
+     *
+     * @param clusterDataRecoveryPolicy the policy to be used when the cluster is started
+     *
+     * @return HotRestartConfig
+     */
+    public HotRestartPersistenceConfig setClusterDataRecoveryPolicy(HotRestartClusterDataRecoveryPolicy
+                                                                            clusterDataRecoveryPolicy) {
+        this.clusterDataRecoveryPolicy = clusterDataRecoveryPolicy;
         return this;
     }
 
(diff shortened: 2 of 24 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -375,7 +380,7 @@ public void connectionRemoved(Connection connection) {
                     return;
                 }
 
-                String localMemberUuid = node.getLocalMember().getUuid();
+                String localMemberUuid = node.getThisUuid();
```
metanet: Do you mean the `localMemberUuid` reference?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
```diff
@@ -460,9 +460,23 @@ public void provideMetrics(MetricsRegistry registry) {
...
-                ASYNC_QUEUE_CAPACITY, ExecutorType.CONCRETE);
+        setInvocationContext();
+
+        invocationMonitor.start();
+        operationExecutor.start();
+        asyncResponseHandler.start();
+        slowOperationDetector.start();
+    }
+
+    public void setInvocationContext() {
```
pveentjer: Can this be renamed to 'initInvocationContext' since this method isn't a setter.

pveentjer: And why is it public?

metanet: Sure, I will rename it. It is public because HotRestartService accesses it during force start.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
```diff
@@ -460,9 +460,23 @@ public void provideMetrics(MetricsRegistry registry) {
...
+        ManagedExecutorService asyncExecutor;
+        if (this.invocationContext != null) {
```
pveentjer: Is initInvocationContext called more than once? Because there is a null check I don't understand.

If it is called more than once, there could be 2 instances of the invocationContext which doesn't make a lot of sense.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
pveentjer: Ok

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
pveentjer: So if I understand it correctly, the InvocationContext will get a different content? I guess the nodeEngine.getLocalMember().getUuid() is updated?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
pveentjer: Can you add a few lines what the purpose is of this method? I believe I  understand it is done to update one of the properties of the InvocationContext (e.g. uuid change).

metanet: Sure, I will add a comment line. 

This is happening before the node start is completed. Therefore, there cannot be any invocations yet. After node is started, uuid is static.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
metanet: Yes, if a node is doing force-start, it will change it is uuid.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
pveentjer: Ok

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationServiceImpl.java
pveentjer: Ok. I understand this part. Thanks for the explanation

metanet: Great. You are welcome.

pveentjer:
I have only looked at the operation/invocation section and I don't see anything **here** to prevent this PR from being merged. This is not a full thumbs up because my understanding of the clustering section is not sufficient to say anything sensible about that.

metanet:
Thank you for your review. I guess there will be more reviews on other parts of the pr.

hazelcast/src/main/java/com/hazelcast/config/HotRestartPersistenceConfig.java
```diff
@@ -58,6 +58,8 @@
     private int parallelism = DEFAULT_PARALLELISM;
     private int validationTimeoutSeconds = DEFAULT_VALIDATION_TIMEOUT;
     private int dataLoadTimeoutSeconds = DEFAULT_DATA_LOAD_TIMEOUT;
+    private HotRestartClusterDataRecoveryPolicy clusterDataRecoveryPolicy
+            = HotRestartClusterDataRecoveryPolicy.FULL_RECOVERY_ONLY;
```
mdogan: Reminder: we should apply this change for XML config too.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -139,7 +130,6 @@ public ClusterServiceImpl(Node node) {
         clusterClock = new ClusterClockImpl(logger);
 
         thisAddress = node.getThisAddress();
-        thisMember = node.getLocalMember();
```
metanet: Local member information can be changed before the start is completed. Therefore, it is not safe to cache local member here.

PR #9204, opened 2016-10-31
Added support to remove import-package instructions while merging dep…

noctarius:
…endencies


jerrinot:
@noctarius can you describe bit a bit? what's the intention here, use-case, etc. 
thanks!

noctarius:
See the changes in the references Hazelcast Enterprise PR :) The main purpose is to remove wrong imports from Hazelcast modules in our hazelcas(-enterprise)-all.jar files that prevent them from starting up in OSGi environments. You can either mark them all as optional, however some of them seem to be not required at all (except some very narrow use cases), like javax.awt.* or javax.swing.* or they're just test dependencies (like org.mockito.*).

PR #9205, opened 2016-10-31
Fix partition replica version handling when backups are reordered

mdogan:
When backups are reordered and a backup with a version greater than `(localVersion + 1)`
is received, backup is applied but partition replica versions are not updated. After that point, until replica is repaired, succeeding stale backups cannot be detected
anymore. This breaks monotonicity of backups.

Now, partition replica versions are always updated when `(newVersion > localVersion)` but
partition is marked as dirty if `(newVersion > localVersion + 1)`. Anti-entropy system will take
this dirty flag into account too.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaVersions.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaVersions.java
@@ -22,10 +22,15 @@ import java.util.Arrays;
 
 import static java.lang.System.arraycopy;
 
+// read and updated only by partition threads
 final class PartitionReplicaVersions {
-    final int partitionId;
-    // read and updated only by operation/partition threads
-    final long[] versions = new long[InternalPartition.MAX_BACKUP_COUNT];
+    private final int partitionId;
+    private final long[] versions = new long[InternalPartition.MAX_BACKUP_COUNT];
+    /**
+     * Shows whether partition has missing backups somewhere between the last applied backup
+     * and the last incremental backup received.
+     */
+    private boolean dirty;
 
     PartitionReplicaVersions(int partitionId) {
         this.partitionId = partitionId;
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaVersions.java
```diff
@@ -22,10 +22,11 @@
...
 import static java.lang.System.arraycopy;
 
+// read and updated only by partition threads
 final class PartitionReplicaVersions {
-    final int partitionId;
-    // read and updated only by operation/partition threads
-    final long[] versions = new long[InternalPartition.MAX_BACKUP_COUNT];
+    private final int partitionId;
+    private final long[] versions = new long[InternalPartition.MAX_BACKUP_COUNT];
+    private boolean dirty;
```
pveentjer: Can you add a few lines of comment what 'dirty' means? It has a change which isn't replicated?

hazelcast/src/test/java/com/hazelcast/nio/tcp/PacketFilter.java
```diff
@@ -25,6 +25,6 @@
  */
 public interface PacketFilter {
 
-    boolean allow(Packet packet, Address endpoint);
+    boolean filter(Packet packet, Address endpoint);
```
pveentjer: What does filter mean? True means the packet is OK or true means the packet should be dropped?

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaVersions.java
mdogan: added comments.

hazelcast/src/test/java/com/hazelcast/nio/tcp/PacketFilter.java
mdogan: added comments.

PR #9212, opened 2016-11-02
Add buildNumber in MemberImpl

vbekiaris:
- Introduces `MemberImpl.getBuildNumber`, populated from joining members' join requests.
- `BuildInfoProvider.getBuildInfo` does not parse runtime properties on each invocation, instead returns a cached static `BuildInfo` object.

reasoning: `buildNumber` was already sent over the wire in `JoinMessage` & `JoinRequest`, however the information was not available in `MemberImpl` objects returned by `Cluster.getMembers`.


vbekiaris:
`buildNumber` information would not be propagated to all members of the cluster so closing this PR.

PR #9225, opened 2016-11-07
getEventExecutor API missing error

ihsandemir:
Restoring getEventExecutor which is required by EE.

PR #9226, opened 2016-11-07
getEventExecutor API missing error

ihsandemir:
Restoring getEventExecutor which is required by enterprise.


ihsandemir:
Ported from https://github.com/hazelcast/hazelcast/pull/9225

PR #9228, opened 2016-11-07
EntryProcessor sample for non-java client tests

asimarslan:
A simple IdentifiedDataserializable EntryProcessor and its factory for non java clients to test EntryProcessor.


hazelcast/src/test/java/com/hazelcast/client/test/IdentifiedEntryProcessor.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public void readData(ObjectDataInput in) throws IOException {
+        value = in.readUTF();
+    }
+
+    @Override
+    public Object process(Map.Entry<String, String> entry) {
+        entry.setValue(value);
```
ihsandemir: It may be better to set this returned value different from what was deserialized just to prevent suppressing any error in case the value we get is not from the result of process method. E.g. you can just return value + " result from process"

PR #9229, opened 2016-11-07
Versioned serialization

tombujok:
Prepares the ground for the versioned serialization implemented in the EE.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/VersionedObjectDataInput.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/VersionedObjectDataInput.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.serialization.impl;
+
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.VersionAware;
+import com.hazelcast.version.Version;
+
+import java.io.InputStream;
+
+/**
+ * Base class for ObjectDataInput that is VersionAware and allows mutating the version.
+ * What the version means it's up to the Serializer/Deserializer.
+ * If the serializer supports versioning it may set the version to use for the serialization on this object.
+ */
+abstract class VersionedObjectDataInput extends InputStream implements ObjectDataInput, VersionAware {
+
+    protected Version version = Version.UNKNOWN;
+
+    /**
+     * If the serializer supports versioning it may set the version to use for the serialization on this object.
+     *
+     * @param version
+     */
+    public void setVersion(Version version) {
+        this.version = version;
+    }
+
+    /**
+     * If the serializer supports versioning it may set the version to use for the serialization on this object.
+     * This method makes the version available for the user.
+     *
+     * @return the version of Version.UNKNOWN if the version is unknown to the object.
+     */
+    @Override
+    public Version getVersion() {
+        return version;
+    }
+
+}
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/VersionedObjectDataOutput.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/VersionedObjectDataOutput.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.serialization.impl;
+
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.VersionAware;
+import com.hazelcast.version.Version;
+
+import java.io.OutputStream;
+
+/**
+ * Base class for ObjectDataInput that is VersionAware and allows mutating the version.
+ * What the version means it's up to the Serializer/Deserializer.
+ * If the serializer supports versioning it may set the version to use for the serialization on this object.
+ */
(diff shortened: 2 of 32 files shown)


tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
```diff
@@ -53,7 +54,7 @@
      * Removes the specified MembershipListener.
      * <p/>
      * If the same MembershipListener is registered multiple times, it needs to be removed multiple times.
-     *
+     * <p>
```
pveentjer: You can disable this in Idea :)

hazelcast/src/main/java/com/hazelcast/internal/serialization/SerializationServiceBuilder.java
```diff
@@ -63,5 +64,7 @@
 
     SerializationServiceBuilder setInitialOutputBufferSize(int initialOutputBufferSize);
 
+    SerializationServiceBuilder setClusterService(ClusterService clusterService);
```
tombujok: yeah, I know, it's not perfect. I need a clusterVersion on the write/read call.
we can also change the signatures, but it's used sometimes only

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
tombujok: yeah, I have to finally do it actually :)

hazelcast/src/main/java/com/hazelcast/internal/serialization/SerializationServiceBuilder.java
tombujok: Aha, and some serializers may actually not need it at all, so that was the reason why I didn't change the `StreamSerializer` API.

pveentjer: To reduce coupling, the Cluster version (String?) could be used instead of ClusterService.

tombujok: The thing is that the clusterVersion changes in runtime so I need the `newest` version while serialising

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DataSerializableSerializerOld.java
```diff
@@ -0,0 +1,198 @@
+/*
```
tombujok: will be deleted from the PR

hazelcast/src/main/java/com/hazelcast/nio/serialization/Versioned.java
```diff
@@ -0,0 +1,29 @@
+/*
```
tombujok: will be in impl package.

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
```diff
@@ -216,4 +217,6 @@
      * @since 3.6
      */
     void shutdown(TransactionOptions transactionOptions);
+
+    Version getClusterVersion();
```
mdogan: Is `Version` a public api?

tombujok: yep. it will be used to do an upgrade by the user.

tombujok: ok, nicely spot! I'll double check those

tombujok: Fixed

hazelcast/src/main/java/com/hazelcast/internal/serialization/SerializationServiceBuilder.java
pveentjer: I understand this. But I feel hesitant introducing a dependency on a big object like ClusterService for the sake of obtaining a version. Of course it is the easiest thing to do in the short term, but it increases coupling between the 2 services.

vbekiaris: What about registering the `SerializationService` as a cluster version listener? (not part of this PR); it would require some listener support code on the `ClusterStateManager` side, but should reduce coupling.

tombujok: Sure, I understand this.
I need the version though. Two options I see is to change the service's APIe to pass the version on the serialization call (it seems ugly). 
The other option is to have some coupling, we can introduce a new interface (a sub-interface of the ClusterService).
A listener sounds good too - don't know how feasible that is.

tombujok: Fixed by introducing a versionListener in the EE.

hazelcast/src/main/java/com/hazelcast/internal/cluster/ClusterVersionListener.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Listeners interested on acting upon cluster version update should implement this interface.
+ * TODO: only for services? (like ClusterStateListener?)
+ * TODO: can clients listen for cluster version change?
+ *
+ * @see Cluster#getClusterVersion()
+ * @see Cluster#changeClusterVersion(Version)
+ */
+public interface ClusterVersionListener {
+
```
tombujok: This will come from Vassiliss' PR

pveentjer:
Looks good

hazelcast/src/main/java/com/hazelcast/version/Version.java
```diff
@@ -0,0 +1,166 @@
...
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.DataSerializable;
+
+import java.io.IOException;
+
+/**
+ * Determines Hazelcast version.
+ * Immutable.
+ */
+public final class Version implements DataSerializable {
```
jerrinot: also constructors should not be public? given the class itself is part of public API

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/VersionedObjectDataOutput.java
```diff
@@ -0,0 +1,41 @@
...
+import com.hazelcast.version.Version;
+
+import java.io.OutputStream;
+
+/**
+ * Base class for ObjectDataOutput that is VersionAware and allows mutating the version
+ */
+abstract class VersionedObjectDataOutput extends OutputStream implements ObjectDataOutput, VersionAware {
+
+    protected Version version = Version.UNKNOWN;
```
jerrinot: what's the semantic of this? under what circumstances it can return `UKNOWN` version?

hazelcast/src/main/java/com/hazelcast/version/Version.java
tombujok: Class removed, since provided in another PR.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/VersionedObjectDataInput.java
```diff
@@ -0,0 +1,41 @@
...
+abstract class VersionedObjectDataInput extends InputStream implements ObjectDataInput, VersionAware {
+
+    protected Version version = Version.UNKNOWN;
+
+    public void setVersion(Version version) {
+        this.version = version;
+    }
+
+    @Override
+    public Version getVersion() {
```
tombujok: Fixed. More doc will be in EE that actually uses it.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/VersionedObjectDataOutput.java
tombujok: Fixed. More doc will be in EE that actually uses it.

tombujok:
run-lab-run

PR #9230, opened 2016-11-09
Move concurrent queues from HotRestart module

mtopolnik:
The concurrent array queues are now used in Jet as well.

Also:

1. Move `MPSCQueue` to the same `...concurrent` package
2. Restore the deleted method `getNearCache`, which is used from Simulator code.

Followup PR on EE: https://github.com/hazelcast/hazelcast-enterprise/pull/1117

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -26,6 +26,9 @@
     <suppress checks="Javadoc(Method|Type|Variable)" files="com/hazelcast/internal/"/>
     <suppress checks="Javadoc(Method|Type|Variable)" files="com/hazelcast/util/executor/"/>
 
+    <!-- Concurrent queue composed of many parts for padding that avoids false sharing -->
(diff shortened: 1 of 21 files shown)


checkstyle/suppressions.xml
```diff
@@ -26,6 +26,9 @@
     <suppress checks="Javadoc(Method|Type|Variable)" files="com/hazelcast/internal/"/>
     <suppress checks="Javadoc(Method|Type|Variable)" files="com/hazelcast/util/executor/"/>
 
+    <!-- Concurrent queue composed of many parts for padding that avoids false sharing -->
+    <suppress checks="OuterTypeNumber" files="AbstractConcurrentArrayQueue\.java"/>
```
pveentjer: You don't want to add supress in the java code?

mtopolnik: Impossible because it's on file level.

pveentjer:
Since this has already been reviewed since it was EE code; and the rest is just moving classes around... +1

PR #9231, opened 2016-11-09
Hot backup

mmedenjak:
Hot Restart snapshot feature. Base for https://github.com/hazelcast/hazelcast-enterprise/pull/1118

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/hotrestart/HotRestartBackupService.java
+++ b/hazelcast/src/main/java/com/hazelcast/hotrestart/HotRestartBackupService.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.hotrestart;
+
+/**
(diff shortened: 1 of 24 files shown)


hazelcast/src/main/java/com/hazelcast/nio/IOUtil.java
```diff
@@ -388,6 +391,89 @@ public static File getFileFromResources(String resourceFileName) {
...
+    public static void copy(File source, File target) {
+        if (!source.exists()) {
+            throw new IllegalArgumentException("Source does not exist");
+        }
+        if (source.isDirectory()) {
+            copyDirectory(source, target);
+        } else {
+            copyFile(source, target, -1);
+        }
+    }
```
tombujok: I know it's a pain in the neck, but these would need some integration tests to make sure it works.

mmedenjak: Agree, added.

hazelcast/src/main/java/com/hazelcast/hotrestart/HotRestartBackupService.java
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.hotrestart;
+
+/**
+ * Service for starting cluster-wide hot restart data backups, determining the local backup state and interrupting a currently
+ * running local hot restart backup.
+ */
+public interface HotRestartBackupService {
```
mdogan: Are these classes part of public API? If not then it's better to move all to an `internal` package?

hazelcast/src/test/java/com/hazelcast/internal/cluster/impl/LockGuardTest.java
```diff
@@ -16,6 +16,7 @@
 
 package com.hazelcast.internal.cluster.impl;
```
mdogan: Shouldn't you move the test under `com.hazelcast.internal.util`?

hazelcast/src/main/java/com/hazelcast/hotrestart/HotRestartBackupService.java
mmedenjak: This will be public API, the user will be able to trigger backup via API or management center.

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

PR #9233, opened 2016-11-09
Cache event listeners triggers multiple events when the listener is configured using the xml

ihsandemir:
Fixes the issue of duplicating the cache config listener registration configurations when xml based listener is defined for a Cache. A new test is added as well.


pveentjer:
What was the cause of the problem and how is it fixed?

ihsandemir:
Fixes issue: https://github.com/hazelcast/hazelcast/issues/9315

ihsandemir:
@pveentjer The problem was that the listener config was being added into the config all the time, where it should not be added when the config is retrieved by the client from the member.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
```diff
@@ -340,6 +342,30 @@ public void registerCacheEntryListener(CacheEntryListenerConfiguration<K, V> cac
         }
     }
 
+    @Override
+    public void registerCacheEntryListener(CacheEntryListenerConfiguration cacheEntryListenerConfiguration, boolean addToConfig) {
```
vbekiaris: You could deduplicate code and implement `registerCacheEntryListener(CacheEntryListenerConfiguration<K, V> cacheEntryListenerConfiguration)` by invoking  `registerCacheEntryListener(cacheEntryListenerConfiguration, true)`.

hazelcast/src/main/java/com/hazelcast/cache/impl/InternalCacheEntryListenerRegisterer.java
```diff
@@ -0,0 +1,34 @@
...
+package com.hazelcast.cache.impl;
+
+import javax.cache.configuration.CacheEntryListenerConfiguration;
+
+/**
+ * Used internally to register the cache entry listener with adding the configuration to the cache config or not.
+ * @param <K>
+ * @param <V>
+ */
+public interface InternalCacheEntryListenerRegisterer<K, V> {
```
ahmetmircik: Seems `registerCacheEntryListener ` can be added to `ICacheInternal` instead of introducing `InternalCacheEntryListenerRegisterer`

vbekiaris: 👍

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
ihsandemir: Yes, that was the original idea, which i forgot to refactor later. Corrected now.

ihsandemir:
run-lab-run

PR #9234, opened 2016-11-10
Partitions should remain unavailable until services commit/rollback

mdogan:
Partitions should be unavailable/inaccessible until `MigrationAwareService`s
complete commit/rollback during migration and/or promotion.
Each partition has an `isMigrating` flag which is set before migration starts
and cleared after migration completes. When a partition is marked as migrating,
partition operations will be rejected immediately.

This fix provides safety by guaranteeing an operation which is submitted after
migration starts, won't be executed before MigrationAwareService commit/rollback returns.

This is a partial fix for issue https://github.com/hazelcast/hazelcast/issues/8385.
There's no ordering for observers outside of partition operation threads.


metanet:
👍

mdogan:
run-lab-run

PR #9239, opened 2016-11-11
Minor Invocation.MIN_TIMEOUT cleanup

pveentjer:
Renamed it to MIN_TIMEOUT_MILLIS to include the time unit. This should prevent any timeunit confusion

Instead of hard coding it to 10000 is is now rewritten to SECONDS.toMillis(10). This is easier
to read.


pveentjer:
run-lab-run

PR #9241, opened 2016-11-11
Fix Operation.waitTimeout update on retrying

### LINKED ISSUE #9250: Problems with timeout in blocking calls
In case of retrying a BlockingOperation, the waitTimeout is made smaller so that
a timed blocking operation eventually times out.

However there are 2 bugs:

 * it can happen that a positive timeout, becomes negative due to subtraction. A negative
    timeout is considered to be an infinite timeout. So this can cause problems since the bound waiting transforms into an infinite wait.

  * it can happen that a blocking call doesn't wait as long as specified because instead of subtracting
    the elapsed time (good), the call timeout is subtracted (bad). This is a problem if the elapsed time is
    smaller than the call timeout. This can happen because the OperationParker periodically sweeps
    through all operations to expire them. This could happen 1 second after the operation
    get executed, but it will still get a 60 second (default timeout) being deducted from the
    remaining timeout. So it could be that you call lock.tryLock(60,seconds), but instead of waiting 60 seconds for the lock to come available, after 0 seconds the lock.tryLock could return false.

pveentjer:
In case of retrying a BlockingOperation, the waitTimeout is made smaller so that
a timed blocking operation eventually times out.

However there are 2 bugs:
- it can happen that a positive timeout, becomes negative due to subtraction. A negative
timeout is considered to be an infinite timeout. So this can cause problems since the bound waiting transforms into an infinite wait.

- it can happen that a blocking call doesn't wait as long as specified because instead of subtracting
the elapsed time (good), the call timeout is subtracted (bad). This is a problem if the elapsed time is
smaller than the call timeout. This can happen because the OperationParker periodically sweeps
through all operations to expire them. This could happen 1 second after the operation
get executed, but it will still get a 60 second (default timeout) being deducted from the
remaining timeout. So it could be that you call lock.tryLock(60,seconds), but instead of waiting 60 seconds for the lock to come available, after 0 seconds the lock.tryLock could return false.

todo: need to add some tests

fixes #9250

@mdogan @jerrinot can you have a look at this PR

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -82,6 +82,8 @@ import static com.hazelcast.util.ExceptionUtil.rethrow;
 import static com.hazelcast.util.StringUtil.timeToString;
 import static java.lang.Boolean.FALSE;
 import static java.lang.Boolean.TRUE;
+import static java.lang.Math.max;
+import static java.lang.Math.min;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.logging.Level.FINEST;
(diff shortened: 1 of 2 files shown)


pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -426,15 +428,30 @@ void notifyCallTimeout() {
...
+            // We need to subtract the elapsed time so that the waitTimeout gets smaller on every retry.
+
+            // first we determine elapsed time. To prevent of the elapsed time being negative due to cluster-clock reset, and
+            // the call timeout increasing instead of decreasing, the elapsedTime will be at least 0.
+            // For elapsed time we rely on cluster-clock, since op.invocationTime is also based on it.
+            long elapsedTime = max(0, context.clusterClock.getClusterTime() - op.getInvocationTime());
+
+            // We need to take care of not running into a negative wait-timeout, because it will be interpreted as infinite.
+            // That why the max with 0, so that 0 is going to be the smallest remaining timeout
+            newWaitTimeout = max(0, oldWaitTimeout - elapsedTime);
```
jerrinot: Do we have a well-defined behaviour when `waitTimeout == 0` ?

pveentjer:
```
  /**
     * Returns the wait timeout in millis. -1 means infinite wait, and 0 means no waiting at all.
     *
     * The wait timeout is the amount of time a {@link BlockingOperation} is allowed to parked in the
     * {@link com.hazelcast.spi.impl.operationparker.OperationParker}.
     *
     * Examples:
     * <ol>
     *     <li>in case of ILock.tryLock(10, ms), the wait timeout is 10 ms</li>
     *     <li>in case of ILock.lock(), the wait timeout is -1</li>
     *     <li>in case of ILock.tryLock(), the wait timeout is 0.</li>
     * </ol>
     *
     * The waitTimeout only has meaning for blocking operations. For non blocking operations the value is undefined.
     *
     * @return the wait timeout.
     */
    public final long getWaitTimeout() {
        return waitTimeout;
    }
```

jerrinot: splendid!

jerrinot:
looks good. add some tests and I am sold! :)

PR #9243, opened 2016-11-11
Adds cluster version and node-cluster version validation.

vbekiaris:
Adds the notion of cluster version to `Cluster` and codebase version to `Node`. Main changes:
* Added cluster version & node version
* During node join, node version is validated for compatibility with current cluster version

Requires an EE counterpart.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/Cluster.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/Cluster.java
@@ -20,6 +20,7 @@ import com.hazelcast.cluster.ClusterState;
 import com.hazelcast.spi.properties.GroupProperty;
 import com.hazelcast.transaction.TransactionException;
 import com.hazelcast.transaction.TransactionOptions;
+import com.hazelcast.version.Version;
 
 import java.util.Set;
 
@@ -167,6 +168,29 @@ public interface Cluster {
      * @since 3.6
      */
     void changeClusterState(ClusterState newState, TransactionOptions transactionOptions);
+
+    /**
+     * The cluster version indicates the operating version of the cluster. It is separate from each node's codebase version,
+     * as it may be acceptable for a node to operate at a different compatibility version than its codebase version. This method
+     * may return {@code null} if invoked after the ClusterService is constructed and before the node forms a cluster, either
+     * by joining existing members or becoming master of its standalone cluster if it is the first node on the cluster.
+     * Importantly, this is the time during which a lifecycle event with state
+     * {@link com.hazelcast.core.LifecycleEvent.LifecycleState#STARTING} is triggered.
+     *
+     * For example, consider a cluster comprised of nodes running on {@code hazelcast-3.8.jar}. Each node's codebase version
+     * is 3.8.0 and on startup the cluster version is also 3.8.0. After a while, another node joins, running on
+     * {@code hazelcast-3.9.jar}; this node's codebase version is 3.9.0. If deemed compatible, it is allowed to join the cluster.
+     * At this point, the cluster version is still 3.8.0, the 3.9.0 member should be able to adapt its behaviour to be compatible
+     * with other the 3.8.0 members. Once all 3.8.0 members have been shutdown and replaced by other members on codebase
+     * version 3.9.0, still the cluster version will be 3.8.0. At this point, it is possible to update the cluster version to
+     * 3.9.0, since all cluster members will be compatible with the new cluster version. Once cluster version
+     * is updated to 3.9.0, further communication among members will take place in 3.9.0 and all new features and functionality
+     * of version 3.9.0 will be available.
+     *
+     * @return the version at which this cluster operates.
+     * @since 3.8
+     */
+    Version getClusterVersion();
 
     /**
      * Changes state of the cluster to the {@link ClusterState#PASSIVE} transactionally,
--- a/hazelcast/src/main/java/com/hazelcast/core/Member.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/Member.java
@@ -18,6 +18,7 @@ package com.hazelcast.core;
 
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.DataSerializable;
+import com.hazelcast.version.Version;
 
 import java.net.InetSocketAddress;
 import java.util.Map;
@@ -257,5 +258,16 @@ public interface Member extends DataSerializable, Endpoint {
      * @since 3.2
      */
     void removeAttribute(String key);
+
+    /**
+     * Returns the Hazelcast codebase version of this member; this may or may not be different from the version reported by
+     * {@link Cluster#getClusterVersion()}, for example when a node with a different codebase version is added to an
+     * existing cluster. See the documentation for {@link Cluster#getClusterVersion()} for a more thorough discussion
+     * of {@code Cluster} and {@code Member} / {@code Node} version.
+     *
+     * @return the {@link Version} of this member.
+     * @since 3.8
+     */
+    Version getVersion();
 
 }
--- a/hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
@@ -19,6 +19,7 @@ package com.hazelcast.core;
 import com.hazelcast.instance.MemberImpl;
 import com.hazelcast.nio.Address;
 import com.hazelcast.spi.exception.RetryableException;
+import com.hazelcast.version.Version;
 
 import java.io.IOException;
 import java.io.ObjectInputStream;
@@ -57,13 +58,14 @@ public class MemberLeftException extends ExecutionException implements Retryable
         out.defaultWriteObject();
 
         Address address = member.getAddress();
-        String host  = address.getHost();
+        String host = address.getHost();
         int port = address.getPort();
 
         out.writeUTF(member.getUuid());
         out.writeUTF(host);
         out.writeInt(port);
         out.writeBoolean(member.isLiteMember());
+        out.writeObject(member.getVersion());
     }
 
     private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
@@ -73,7 +75,8 @@ public class MemberLeftException extends ExecutionException implements Retryable
         String host = in.readUTF();
         int port = in.readInt();
         boolean liteMember = in.readBoolean();
+        Version version = (Version) in.readObject();
 
-        member = new MemberImpl(new Address(host, port), false, uuid, null, null, liteMember);
+        member = new MemberImpl(new Address(host, port), version, false, uuid, null, null, liteMember);
     }
 }
--- a/hazelcast/src/main/java/com/hazelcast/instance/AbstractMember.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/AbstractMember.java
@@ -16,7 +16,6 @@
 
 package com.hazelcast.instance;
 
-import com.hazelcast.nio.serialization.impl.BinaryInterface;
 import com.hazelcast.cluster.MemberAttributeOperationType;
 import com.hazelcast.core.Member;
 import com.hazelcast.logging.ILogger;
@@ -24,7 +23,9 @@ import com.hazelcast.nio.Address;
 import com.hazelcast.nio.IOUtil;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.impl.BinaryInterface;
 import com.hazelcast.spi.annotation.PrivateApi;
+import com.hazelcast.version.Version;
 
 import java.io.IOException;
 import java.net.InetAddress;
@@ -43,24 +44,26 @@ public abstract class AbstractMember implements Member {
     protected Address address;
     protected String uuid;
     protected boolean liteMember;
+    protected Version version;
 
     protected AbstractMember() {
     }
 
-    protected AbstractMember(Address address) {
-        this(address, null, null);
+    protected AbstractMember(Address address, Version version) {
+        this(address, version, null, null);
     }
 
-    protected AbstractMember(Address address, String uuid) {
-        this(address, uuid, null);
+    protected AbstractMember(Address address, Version version, String uuid) {
+        this(address, version, uuid, null);
     }
 
-    protected AbstractMember(Address address, String uuid, Map<String, Object> attributes) {
-        this(address, uuid, attributes, false);
+    protected AbstractMember(Address address, Version version, String uuid, Map<String, Object> attributes) {
+        this(address, version, uuid, attributes, false);
     }
 
-    protected AbstractMember(Address address, String uuid, Map<String, Object> attributes, boolean liteMember) {
+    protected AbstractMember(Address address, Version version, String uuid, Map<String, Object> attributes, boolean liteMember) {
         this.address = address;
+        this.version = version;
         this.uuid = uuid;
         if (attributes != null) {
             this.attributes.putAll(attributes);
@@ -70,6 +73,7 @@ public abstract class AbstractMember implements Member {
 
     protected AbstractMember(AbstractMember member) {
         this.address = member.address;
+        this.version = member.version;
         this.uuid = member.uuid;
         this.attributes.putAll(member.attributes);
         this.liteMember = member.liteMember;
@@ -150,11 +154,17 @@ public abstract class AbstractMember implements Member {
     }
 
     @Override
+    public Version getVersion() {
+        return version;
+    }
+
+    @Override
     public void readData(ObjectDataInput in) throws IOException {
         address = new Address();
         address.readData(in);
         uuid = in.readUTF();
         liteMember = in.readBoolean();
+        version = in.readObject();
         int size = in.readInt();
         for (int i = 0; i < size; i++) {
             String key = in.readUTF();
@@ -168,6 +178,7 @@ public abstract class AbstractMember implements Member {
         address.writeData(out);
         out.writeUTF(uuid);
         out.writeBoolean(liteMember);
+        out.writeObject(version);
         Map<String, Object> attributes = new HashMap<String, Object>(this.attributes);
         out.writeInt(attributes.size());
         for (Map.Entry<String, Object> entry : attributes.entrySet()) {
--- a/hazelcast/src/main/java/com/hazelcast/instance/BuildInfoProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/BuildInfoProvider.java
@@ -16,6 +16,8 @@
 
 package com.hazelcast.instance;
 
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.Logger;
 import com.hazelcast.util.EmptyStatement;
 
 import java.io.InputStream;
@@ -28,9 +30,29 @@ import static com.hazelcast.nio.IOUtil.closeResource;
  */
 public final class BuildInfoProvider {
 
+    /**
+     * Use this in production code to obtain the BuildInfo already parsed when this class was first loaded.
+     * Its properties will not change at runtime.
+     */
+    public static final BuildInfo BUILD_INFO;
+
+    private static final ILogger LOGGER;
+
+    static {
+        LOGGER = Logger.getLogger(BuildInfoProvider.class);
+        BUILD_INFO = getBuildInfo();
+    }
+
     private BuildInfoProvider() {
     }
 
+    /**
+     * Parses {@code hazelcast-runtime.properties} for {@code BuildInfo}; also checks for overrides in System.properties.
+     * Use this method to obtain and cache a {@code BuildInfo} object or from test code that needs to re-parse properties
+     * on each invocation.
+     *
+     * @return the parsed BuildInfo
+     */
     public static BuildInfo getBuildInfo() {
         final InputStream inRuntimeProperties =
                 BuildInfoProvider.class.getClassLoader().getResourceAsStream("hazelcast-runtime.properties");
@@ -62,6 +84,13 @@ public final class BuildInfoProvider {
             build = String.valueOf(hazelcastBuild);
         }
         int buildNumber = Integer.parseInt(build);
+
+        // override version with a system property
+        String overridingVersion = System.getProperty("hazelcast.version");
+        if (overridingVersion != null) {
+            LOGGER.info("Overriding hazelcast version with system property value " + overridingVersion);
+            version = overridingVersion;
+        }
 
         String sv = runtimeProperties.getProperty("hazelcast.serialization.version");
         byte serialVersion = Byte.parseByte(sv);
--- a/hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
@@ -21,9 +21,16 @@ import com.hazelcast.cache.impl.ICacheService;
 import com.hazelcast.cluster.ClusterState;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.SerializationConfig;
+import com.hazelcast.core.HazelcastInstanceAware;
 import com.hazelcast.core.PartitioningStrategy;
 import com.hazelcast.internal.cluster.ClusterStateListener;
+import com.hazelcast.internal.cluster.ClusterVersionListener;
+import com.hazelcast.internal.cluster.impl.JoinMessage;
+import com.hazelcast.internal.cluster.impl.VersionMismatchException;
 import com.hazelcast.internal.management.dto.ClusterHotRestartStatusDTO;
+import com.hazelcast.internal.networking.ReadHandler;
+import com.hazelcast.internal.networking.SocketChannelWrapperFactory;
+import com.hazelcast.internal.networking.WriteHandler;
 import com.hazelcast.internal.serialization.InternalSerializationService;
 import com.hazelcast.internal.serialization.SerializationServiceBuilder;
 import com.hazelcast.internal.serialization.impl.DefaultSerializationServiceBuilder;
@@ -38,10 +45,7 @@ import com.hazelcast.nio.MemberSocketInterceptor;
 import com.hazelcast.nio.tcp.DefaultSocketChannelWrapperFactory;
 import com.hazelcast.nio.tcp.MemberReadHandler;
 import com.hazelcast.nio.tcp.MemberWriteHandler;
-import com.hazelcast.internal.networking.ReadHandler;
-import com.hazelcast.internal.networking.SocketChannelWrapperFactory;
 import com.hazelcast.nio.tcp.TcpIpConnection;
-import com.hazelcast.internal.networking.WriteHandler;
 import com.hazelcast.partition.strategy.DefaultPartitioningStrategy;
 import com.hazelcast.security.SecurityContext;
 import com.hazelcast.spi.NodeEngine;
@@ -51,7 +55,9 @@ import com.hazelcast.spi.impl.servicemanager.ServiceManager;
 import com.hazelcast.spi.properties.GroupProperty;
 import com.hazelcast.util.ConstructorFunction;
 import com.hazelcast.util.ExceptionUtil;
+import com.hazelcast.util.Preconditions;
 import com.hazelcast.util.UuidUtil;
+import com.hazelcast.version.Version;
 import com.hazelcast.wan.WanReplicationService;
 import com.hazelcast.wan.impl.WanReplicationServiceImpl;
 
@@ -59,16 +65,19 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import static com.hazelcast.map.impl.MapServiceConstructor.getDefaultMapServiceConstructor;
+import static com.hazelcast.version.Version.MAJOR_MINOR_VERSION_COMPARATOR;
 
 @PrivateApi
-@SuppressWarnings({"checkstyle:methodcount"})
+@SuppressWarnings({"checkstyle:methodcount", "checkstyle:classfanoutcomplexity"})
 public class DefaultNodeExtension implements NodeExtension {
 
     protected final Node node;
     protected final ILogger logger;
     protected final ILogger systemLogger;
+    protected final List<ClusterVersionListener> clusterVersionListeners = new CopyOnWriteArrayList<ClusterVersionListener>();
 
     private final MemoryStats memoryStats = new DefaultMemoryStats();
 
@@ -221,13 +230,17 @@ public class DefaultNodeExtension implements NodeExtension {
     }
 
     @Override
-    public void validateJoinRequest() {
+    public void validateJoinRequest(JoinMessage joinMessage) {
+        // check joining member's major.minor version is same as current cluster version's major.minor numbers
+        if (MAJOR_MINOR_VERSION_COMPARATOR.compare(joinMessage.getVersion(), node.getClusterService().getClusterVersion()) != 0) {
+            throw new VersionMismatchException("Joining node's version " + joinMessage.getVersion() + " is not compatible with "
+                    + node.getVersion());
+        }
     }
 
     @Override
     public void onClusterStateChange(ClusterState newState, boolean isTransient) {
-        NodeEngineImpl nodeEngine = node.getNodeEngine();
-        ServiceManager serviceManager = nodeEngine.getServiceManager();
+        ServiceManager serviceManager = node.getNodeEngine().getServiceManager();
         List<ClusterStateListener> listeners = serviceManager.getServices(ClusterStateListener.class);
         for (ClusterStateListener listener : listeners) {
             listener.onClusterStateChange(newState);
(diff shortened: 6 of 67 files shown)


hazelcast/src/main/java/com/hazelcast/core/Cluster.java
```diff
@@ -169,6 +170,16 @@
     void changeClusterState(ClusterState newState, TransactionOptions transactionOptions);
 
     /**
+     *
+     * @return the version at which this cluster operates.
```
pveentjer: Can you add a line regarding null? Can it ever be null?

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
```diff
@@ -169,6 +170,16 @@
...
+     */
+    Version getClusterVersion();
+
+    void changeClusterVersion(Version version);
```
pveentjer: Javadoc needed. And what about calling this 'setCusterVersion' ? What is the added value of calling this 'change'?

hazelcast/src/main/java/com/hazelcast/core/Member.java
```diff
@@ -258,4 +260,12 @@
      */
     void removeAttribute(String key);
 
+    /**
+     * Returns the Hazelcast version of this member; this may or may not be different from the version reported by
```
pveentjer: Under which conditions can this be different?

hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
```diff
@@ -57,13 +58,14 @@ private void writeObject(ObjectOutputStream out) throws IOException {
         out.defaultWriteObject();
```
pveentjer: Are we sure this exception isn't picked up by the client? If so, there could be a compatibility issue.

hazelcast/src/main/java/com/hazelcast/instance/AbstractMember.java
```diff
@@ -16,15 +16,16 @@
 
 package com.hazelcast.instance;
 
-import com.hazelcast.nio.serialization.impl.BinaryInterface;
 import com.hazelcast.cluster.MemberAttributeOperationType;
 import com.hazelcast.core.Member;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Address;
```
pveentjer: This class is used as a parent class on the client side. Are we sure this isn't going to cause wire incompatibilities?

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -230,7 +244,39 @@ public void onClusterStateChange(ClusterState newState, boolean isTransient) {
...
     @Override
+    public void onClusterVersionChange(Version newVersion) {
+        NodeEngineImpl nodeEngine = node.getNodeEngine();
+        ServiceManager serviceManager = nodeEngine.getServiceManager();
+        List<ClusterVersionListener> listeners = serviceManager.getServices(ClusterVersionListener.class);
+        for (ClusterVersionListener listener : listeners) {
+            listener.onClusterVersionChange(newVersion);
+        }
+        // also trigger cluster version change on explicitly registered listeners
+        if (!clusterVersionListeners.isEmpty()) {
```
pveentjer: I would get of the if. This logic isn't going to be called frequently.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -230,7 +244,39 @@ public void onClusterStateChange(ClusterState newState, boolean isTransient) {
...
+            for (ClusterVersionListener listener : clusterVersionListeners) {
+                listener.onClusterVersionChange(newVersion);
+            }
+        }
+    }
+
+    @Override
+    public boolean isNodeVersionCompatibleWith(Version clusterVersion) {
```
pveentjer: Should this not be a function implemented on the version itself? 

E.g. if(node.getVersion().isCompatible(clusterVersion))...

This way you keep the NodeExtension a bit cleaner. The code becomes a bit more cohesive since the version related functions are bundled in the version.

hazelcast/src/main/java/com/hazelcast/version/Version.java
```diff
@@ -0,0 +1,162 @@
...
+        return ClusterDataSerializerHook.VERSION;
+    }
+
+    @Override
+    @SuppressWarnings("checkstyle:booleanexpressioncomplexity")
+    public int compareTo(Version otherVersion) {
+        int thisVersion = (major << 24 & 0xff0000) | ((minor << 16) & 0xff00) | (patch & 0xff);
+        int thatVersion = (otherVersion.major << 24 & 0xff0000) | (otherVersion.minor << 16 & 0xff00)
+                | (otherVersion.patch & 0xff);
+        return (thisVersion > thatVersion) ? 1 : (thisVersion == thatVersion ? 0 : -1);
```
pveentjer: Nested ternary operations are a pain to read. Can you convert the outer one to an if else?

hazelcast/src/test/java/com/hazelcast/client/impl/MemberImplTest.java
```diff
@@ -41,15 +42,15 @@ public void testConstructor() {
 
     @Test
     public void testConstructor_withAddress() {
-        MemberImpl member = new MemberImpl(address);
+        MemberImpl member = new MemberImpl(address, Version.of("3.8.0"));
```
pveentjer: Is this not going to cause problems as soon as we move to the development of 3.9? So do we really want to have a hardcoded 3.8.0 in the system?

hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
vbekiaris: Client-side exceptions are instantiated on the client-side by `ClientExceptionFactory` that only delivers the exception type, message and stack trace elements, not using java serialization.

hazelcast/src/main/java/com/hazelcast/instance/AbstractMember.java
vbekiaris: `Member` class in client protocol is serialized/deserialized using the dedicated `com.hazelcast.client.impl.protocol.codec.MemberCodec`, the serialized form implemented in `AbstractMember` is only relevant for member-to-member communication.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
vbekiaris: Done

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
vbekiaris: Yes, both validations introduced in `NodeExtension` are overridden in EE side.

hazelcast/src/test/java/com/hazelcast/client/impl/MemberImplTest.java
vbekiaris: In the particular test we need a fixed version in the constructed member objects to validate. In any case I have reverted to using a `Version` object for the current codebase version in other tests which do not need a fixed version to validate, to avoid unpleasant surprises in the future.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -230,7 +244,39 @@ public void onClusterStateChange(ClusterState newState, boolean isTransient) {
...
+        Preconditions.checkNotNull(clusterVersion);
+        return MAJOR_MINOR_VERSION_COMPARATOR.compare(node.getVersion(), clusterVersion) == 0;
+    }
+
+    @Override
     public boolean registerListener(Object listener) {
```
vbekiaris: Thanks, it's in the EE-side PR.

hazelcast/src/main/java/com/hazelcast/core/MemberLeftException.java
pveentjer: Ok. good to know.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
pveentjer: Ok. then it makes sense to do it like this.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
```diff
@@ -119,33 +131,62 @@ void setClusterState(ClusterState newState, boolean isTransient) {
...
+    }
+
     private void doSetClusterState(ClusterState newState, boolean isTransient) {
         this.state = newState;
         stateLockRef.set(ClusterStateLock.NOT_LOCKED);
         changeNodeState(newState);
         node.getNodeExtension().onClusterStateChange(newState, isTransient);
     }
 
+    private void doSetClusterVersion(Version newVersion) {
```
pveentjer: We should also capture it in the diagnostics so we can see when a cluster change has happened.

vbekiaris: @pveentjer I added a cluster version change listener in `SystemLogPlugin`. Test to verify output will be on the EE-side PR.

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
```diff
@@ -169,6 +170,68 @@
...
+     * If an invalid version transition is requested, for example changing to a different major version, an
+     * {@link IllegalArgumentException} is thrown.
+     *
+     * @param version new version of the cluster
+     * @since 3.8
+     */
+    void changeClusterVersion(Version version);
+
+    /**
+     * Changes the cluster version transactionally, with the transaction options provided. Internally this method uses the same
```
vbekiaris: Once all nodes of the cluster have been upgraded, cluster version needs to be also upgraded via `cluster.sh` or management center. A cluster version upgrade is executed transactionally on the cluster and in the process each node validates it's compatible with the requested cluster version change. I think it's a valid entry point for `Cluster` API, wdyt?

vbekiaris: The invoker sets the requested version number, so it is possible one could invoke it with any version number, valid or not. Without the `Version` argument the method's utility would be limited, as it would only allow for one-way change (upgrade) to the "next" version, where "next"'s meaning would be essentially hard-coded in the method's implementation.

mdogan: I see.. but method's utility is already limited, right? At least for now... It cannot set cluster version to any number, it's only for upgrade (and downgrade?). 
What if we don't expose it in public api but only via REST, shell script and management center?

hazelcast/src/main/java/com/hazelcast/internal/cluster/ClusterService.java
```diff
@@ -123,4 +126,41 @@
...
+     * If a node of the cluster is not compatible with the given cluster {@code version}, as implemented by
+     * {@link com.hazelcast.instance.NodeExtension#isNodeVersionCompatibleWith(Version)}, then a
+     * {@link com.hazelcast.internal.cluster.impl.VersionMismatchException} is thrown.
+     *
+     * If an invalid version transition is requested, for example changing to a different major version, an
+     * {@link IllegalArgumentException} is thrown.
+     *
+     * @param version new version of the cluster
+     * @since 3.8
+     */
```
tombujok: It's important to say here, that while the version is being changes the clusterState is changed to IN_TRANSITION and thus other nodes may not join. Did you verify this behavior?

hazelcast/src/main/java/com/hazelcast/internal/cluster/ClusterVersionListener.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.hazelcast.internal.cluster;
+
+import com.hazelcast.core.Cluster;
+import com.hazelcast.version.Version;
+
+/**
+ * Listeners interested on acting upon cluster version update should implement this interface.
+ * Services registered with the Hazelcast {@link com.hazelcast.spi.impl.servicemanager.ServiceManager} which implement this
```
tombujok: just a spelling: `interested in`

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -933,11 +954,11 @@ private void shutdownNodes() {
         hazelcastInstance.getLifecycleService().shutdown();
     }
 
-    public void initialClusterState(ClusterState clusterState) {
+    public void initialClusterState(ClusterState clusterState, Version version) {
         if (node.joined()) {
```
tombujok: When is this invoked? Aha, probably on JoinFinalize, but if there's no join in a single-node cluster?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/JoinMessage.java
```diff
@@ -20,16 +20,23 @@
...
 public class JoinMessage implements IdentifiedDataSerializable {
 
     protected byte packetVersion;
     protected int buildNumber;
+    /**
+     * this is populated with the codebase version of the node trying to join the cluster
+     * (ie {@link com.hazelcast.instance.Node#getVersion()}).
+     */
+    protected Version version;
     protected Address address;
```
tombujok: Could we describe somewhere the flow of JoinMessages and joinRequests, and responses, also including what happenes with the version? It will be easier to understand it for people who know nothing about it.

It's still a bit confusing to me what the difference is between `JoinMessage` and `JoinRequest`

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublishingService.java
```diff
@@ -124,7 +125,7 @@ private void dispatchEntryEventData(EntryEventData entryEventData, ListenerAdapt
     private Member getMember(EventData eventData) {
         Member member = nodeEngine.getClusterService().getMember(eventData.getCaller());
         if (member == null) {
-            member = new MemberImpl(eventData.getCaller(), false);
+            member = new MemberImpl(eventData.getCaller(), Version.UNKNOWN, false);
         }
```
tombujok: What is this for? and why do we publish unknown?

hazelcast/src/main/java/com/hazelcast/version/Version.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Override
+    public int getId() {
+        return ClusterDataSerializerHook.VERSION;
+    }
+
+    @Override
+    @SuppressWarnings("checkstyle:booleanexpressioncomplexity")
+    public int compareTo(Version otherVersion) {
+        int thisVersion = (major << 16 & 0xff0000) | ((minor << 8) & 0xff00) | (patch & 0xff);
```
tombujok: Can you add some javadoc here ;) ? It's tough to read those binary operators.

hazelcast/src/main/java/com/hazelcast/version/MajorMinorVersionComparator.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package com.hazelcast.version;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+/**
+ * Version comparator that disregards patch version.
+ */
```
tombujok: Maybe it's good to add : `and compares the major and minor versions only`

hazelcast/src/main/java/com/hazelcast/version/Version.java
```diff
@@ -0,0 +1,180 @@
...
+import java.io.Serializable;
+import java.util.Comparator;
+
+/**
+ * Determines Hazelcast version.
+ */
+@SuppressWarnings("checkstyle:magicnumber")
+public final class Version implements IdentifiedDataSerializable, Comparable<Version>, Serializable {
+
+
```
tombujok: Jaromir had a couple of comments to this class in my PR (but I see it's a bit different now) 
Here's the link: https://github.com/hazelcast/hazelcast/pull/9229

hazelcast/src/test/java/com/hazelcast/version/VersionTest.java
```diff
@@ -0,0 +1,71 @@
...
+    public void testMajorMinorVersionComparator() {
+        assertEquals(0, MAJOR_MINOR_VERSION_COMPARATOR.compare(VERSION_3_8, VERSION_3_8_1));
+        assertEquals(0, MAJOR_MINOR_VERSION_COMPARATOR.compare(VERSION_3_8, VERSION_3_8_2));
+        assertTrue(MAJOR_MINOR_VERSION_COMPARATOR.compare(VERSION_3_9, VERSION_3_8) > 0);
+        assertTrue(MAJOR_MINOR_VERSION_COMPARATOR.compare(VERSION_3_8, VERSION_3_9) < 0);
+        assertTrue(MAJOR_MINOR_VERSION_COMPARATOR.compare(VERSION_3_9, VERSION_3_8_1) > 0);
+        assertTrue(MAJOR_MINOR_VERSION_COMPARATOR.compare(VERSION_3_8_1, VERSION_3_9) < 0);
+    }
+
+}
```
tombujok: Did you get rid of the test verifying the race between joiner and cluster-version change?

hazelcast/src/main/java/com/hazelcast/instance/BuildInfoProvider.java
```diff
@@ -63,6 +76,12 @@ public static BuildInfo getBuildInfo() {
         }
         int buildNumber = Integer.parseInt(build);
 
+        // override version with a system property
+        String overridingVersion = System.getProperty("hazelcast.version");
+        if (overridingVersion != null) {
+            version = overridingVersion;
+        }
+
```
vbekiaris: 👍 for logging a message if used and keeping undocumented; having means to override the version is very useful for tests.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -218,7 +227,14 @@ public void shutdown() {
     }
 
     @Override
-    public void validateJoinRequest() {
+    public void validateJoinRequest(JoinRequest joinRequest) {
+        // check joining member's major.minor version is same as current cluster version's major.minor numbers
+        if (node.getClusterService().getClusterVersion() != null
+            && MAJOR_MINOR_VERSION_COMPARATOR.compare(joinRequest.getVersion(),
```
vbekiaris: Null check was redundant, removed it

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -236,7 +252,38 @@ public void onPartitionStateChange() {
...
+    @Override
     public boolean registerListener(Object listener) {
+        if (listener instanceof HazelcastInstanceAware) {
+            ((HazelcastInstanceAware) listener).setHazelcastInstance(node.hazelcastInstance);
+        }
+        if (listener instanceof ClusterVersionListener) {
+            ClusterVersionListener clusterVersionListener = (ClusterVersionListener) listener;
+            clusterVersionListeners.add(clusterVersionListener);
+            // on registration, invoke once the listening method so version is properly initialized on the listener
+            clusterVersionListener.onClusterVersionChange(getClusterOrNodeVersion());
```
vbekiaris: There are listener tests in `DefaultNodeExtensionTest` which verify the listener is invoked on registration.

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
```diff
@@ -169,6 +170,31 @@
     void changeClusterState(ClusterState newState, TransactionOptions transactionOptions);
 
     /**
+     * The cluster version indicates the operating version of the cluster. It is separate from each node's codebase version,
+     * as it may be acceptable for a node to operate at a different compatibility version than its codebase version. This method
+     * may return {@code null} if invoked after the ClusterService is constructed and before the node forms a cluster, either
+     * by joining existing members or becoming master of its standalone cluster if it is the first node on the cluster.
```
vbekiaris: This happens in `Node.setAsMaster()`, invoked during `Node.join()`.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
tombujok: Great!

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -250,4 +297,13 @@ public boolean triggerForceStart() {
     public String createMemberUuid(Address address) {
         return UuidUtil.createMemberUuid(address);
     }
+
+    // obtain cluster version, if already initialized, otherwise this node's version
+    private Version getClusterOrNodeVersion() {
+        if (node.getClusterService() != null && node.getClusterService().getClusterVersion() != null) {
```
vbekiaris: This is only used when delivering current cluster version to a freshly registered listener with the first callback to `onClusterVersionChange`.

In specific case of serialization service, on node startup cluster version will be `null`, so I am not sure how you would like to handle this on SS side; my original thought was that SS listener shouldn't have to deal with `null` cluster version but feel free to propose any improvement.

I guess other cluster version listeners are only registered after node has joined, this is only relevant to cluster version listeners registered before node join (so currently this is only SS).

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
vbekiaris: Invoked from `FinalizeJoinOperation` which is sent from master to joining member (so the joining member will initialize its cluster version here). Single-node cluster is taken care of in `Node.setAsMaster` which is invoked from `Node.join` when no joiner is configured or if no other members are found (so current node self-appoints itself master).

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublishingService.java
vbekiaris: good catch, the member that publishes the event has version `node.version`.

hazelcast/src/main/java/com/hazelcast/internal/cluster/ClusterService.java
vbekiaris: I will update the javadoc with this information and restore the missing test.

hazelcast/src/test/java/com/hazelcast/version/VersionTest.java
vbekiaris: good one, I will restore it

tombujok: Cool, probably it should be in EE though.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/JoinMessage.java
vbekiaris: `JoinMessage` is used as:
* payload in `MasterDiscoveryOperation`. A not joined node configured with `TcpIpJoiner` sends this out to a list of possible member addresses (deducted from tcp-ip configuration) to discover who is master and then join.
* response from a cluster's master node to a node sending a `JoinRequest` when using the multicast joiner; this is the way a not-joined node becomes aware of the cluster master and then sends out a `JoinRequestOperation` to actually join the cluster.
* request & response payload in `JoinCheckOperation` when searching for other clusters in the context of split-brain handling.

`JoinRequest` is used to join a cluster from either multicast or TCP/IP joiner, as well as from multicast joiner when searching for other clusters.

I have already added javadoc to important methods in `ClusterJoinManager` in an attempt to keep track of what each method does. It's not complete documentation of the discovery & join process, but should help.

hazelcast/src/main/java/com/hazelcast/version/Version.java
vbekiaris: @jerrinot `Version` is IDS & has public constructors in this PR.

With regards to semantics of `Version.UNKNOWN`, in the context of this PR it is used to populate the `version` field in client-side `MemberImpl` as client is not aware of the version of each member.

In `Version` class `patch` field is considered to be always available: if the `Version` instance is initialized via parsing a `String` which lacks the `patch` component, it is initialized to 0. `Version.equals` requires all 3 components to be equal. A separate `major.minor`-only comparator is also provided and used where appropriate (e.g. when testing node codebase version vs cluster version compatibility in this PR, we require major.minor equality, allowing for patch-level compatibility only).

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/JoinMessage.java
tombujok: Thx. Good write-up.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -219,13 +229,17 @@ public void shutdown() {
     }
 
     @Override
-    public void validateJoinRequest() {
+    public void validateJoinRequest(JoinMessage joinMessage) {
+        // check joining member's major.minor version is same as current cluster version's major.minor numbers
+        if (MAJOR_MINOR_VERSION_COMPARATOR.compare(joinMessage.getVersion(), node.getClusterService().getClusterVersion()) != 0) {
+            throw new VersionMismatchException("Joining node's version " + joinMessage.getVersion() + " is not compatible with "
+                    + node.getVersion());
+        }
```
tombujok: This constraint is something that we're introducing right now. Earlier, if by any chance two minor releases were compatible, the users might have tried to do a rolling-upgrade and it might have worked.
Do we want to have this constraint considering that the versioned serialization won't work in OS anyway?
CC: @jerrinot
(posting again so that it's not seen in outdated view)

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
tombujok: Hmm, but now I'm thinking that if it returns a null the operation serialized to join a cluster do not send any version. I think it was better to have it set to Node's version first, since there has to be a version all the time.
Wdyt?

hazelcast/src/main/java/com/hazelcast/core/Member.java
```diff
@@ -258,4 +260,15 @@
      */
     void removeAttribute(String key);
 
+    /**
+     * Returns the Hazelcast codebase version of this member; this may or may not be different from the version reported by
+     * {@link ClusterService#getClusterVersion()}, for example when a node with a different codebase version is added to an
```
tombujok: Ah, now I get it. So the Member.version is the version of Member X on stored on another member.

hazelcast/src/test/java/com/hazelcast/instance/DefaultNodeExtensionTest.java
```diff
@@ -0,0 +1,152 @@
...
+        ClusterVersionListener listener = new ClusterVersionListener() {
+            @Override
+            public void onClusterVersionChange(Version newVersion) {
+                latch.countDown();
+            }
+        };
+        assertTrue(nodeExtension.registerListener(listener));
+        assertOpenEventually(latch);
+    }
+
```
tombujok: Could we have a test to guard the the listener will get node's codebase version if cluster.getVersion is null?
Otherwise the versioned serialization will be screwed.

hazelcast/src/main/java/com/hazelcast/version/Version.java
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Immutable UNKNOWN version
+     */
+    public static final Version UNKNOWN = new Version(0, 0, 0);
+    /**
+     * Version comparator that takes into account only major & minor version, disregarding patch version number.
+     */
+    public static final transient Comparator<Version> MAJOR_MINOR_VERSION_COMPARATOR = new MajorMinorVersionComparator();
+
+    //    private static final int UNKNOWN_VERSION_COMPONENT = 0;
```
tombujok: can be removed

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
vbekiaris: Since the serialization service registers its own cluster version listener, it will never receive a `null` cluster version; when `Cluster.getClusterVersion()==null` it will initialize the listener with the node's codebase version.

hazelcast/src/test/java/com/hazelcast/instance/DefaultNodeExtensionTest.java
vbekiaris: Added in `DefaultNodeExtensionTest`

hazelcast/src/main/java/com/hazelcast/version/Version.java
vbekiaris: Done

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
tombujok: As discussed, it's fine to leave it here.

hazelcast/src/test/java/com/hazelcast/instance/DefaultNodeExtensionTest.java
tombujok: Great. thx

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
tombujok: OK, as discussed there will be a property to set the version that should be used before the proper cluster version is set. It's for the sake of backward compatibility, to properly serialize the join messages.

vbekiaris: Done

hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
```diff
@@ -626,6 +626,15 @@
             = new HazelcastProperty("hazelcast.serialization.version",
             BuildInfoProvider.getBuildInfo().getSerializationVersion());
 
+    /**
+     * Override cluster version to use while node is not yet member of a cluster. The cluster version assumed before joining
+     * a cluster may affect the serialization format of cluster discovery & join operations and its compatibility with members
+     * of a cluster running on different Hazelcast codebase versions. The default is to use the node's codebase version. You may
+     * need to override it for your node to join a cluster running on a previous cluster version.
+     */
+    public static final HazelcastProperty INIT_CLUSTER_VERSION
```
tombujok: Looks good. Nice description.

hazelcast/src/main/java/com/hazelcast/core/Member.java
```diff
@@ -258,4 +260,15 @@
...
+     * existing cluster. See the documentation for {@link ClusterService#getClusterVersion()} for a more thorough discussion
+     * of {@code Cluster} and {@code Member} / {@code Node} version.
```
mdogan: Can you change `{@link ClusterService#getClusterVersion()}` to `{@link Cluster#getClusterVersion()}`? This is a public class and I don't think it's desired to refer/import an internal class here.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -695,6 +704,7 @@ public void setAsMaster() {
         setJoined();
```
mdogan: I think `setJoined()` should be the last statement, once node is observed as `joined`, cluster version (and other props) should be visible.

hazelcast/src/main/java/com/hazelcast/map/impl/event/MapEventPublishingService.java
```diff
@@ -124,7 +124,7 @@ private void dispatchEntryEventData(EntryEventData entryEventData, ListenerAdapt
     private Member getMember(EventData eventData) {
         Member member = nodeEngine.getClusterService().getMember(eventData.getCaller());
         if (member == null) {
-            member = new MemberImpl(eventData.getCaller(), false);
+            member = new MemberImpl(eventData.getCaller(), nodeEngine.getNode().getVersion(), false);
```
mdogan: Shouldn't node's codebase version be available on `NodeEngine` interface?  `NodeEngine.getVersion()`? So, no need to cast impl...

hazelcast/src/test/java/com/hazelcast/cluster/ClusterUpgradeTest.java
```diff
@@ -0,0 +1,119 @@
...
+    static final Version VERSION_2_1_1 = Version.of(2, 1, 1);
+    static final Version VERSION_2_2_0 = Version.of(2, 2, 0);
+    static final Version VERSION_3_0_0 = Version.of(3, 0, 0);
+
+    static final int CLUSTER_MEMBERS_COUNT = 3;
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    private TestHazelcastInstanceFactory factory = new TestHazelcastInstanceFactory(10);
```
mdogan: minor note:
When `private TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(10)` is used,  `factory` will be tear down automatically by `HazelcastTestSupport.shutdownNodeFactory()`.

hazelcast/src/test/java/com/hazelcast/instance/DefaultNodeExtensionTest.java
```diff
@@ -0,0 +1,211 @@
...
+import static com.hazelcast.instance.BuildInfoProvider.BUILD_INFO;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test DefaultNodeExtension behavior
+ */
+public class DefaultNodeExtensionTest extends HazelcastTestSupport {
```
mdogan: No `@Category` is provided.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
vbekiaris: well spotted; actually `changeClusterVersion` kicks-off the cluster state change to apply cluster version on each member of the cluster. I think this is not necessary at this point, as it is enough to just set the local cluster version property, if not already set; subsequent `FinalizeJoinOperation`s will apply the cluster version to each member joining the cluster.

vbekiaris:
@mdogan rebased & amended according to your comments.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -692,9 +701,13 @@ Joiner createJoiner() {
     public void setAsMaster() {
         logger.finest("This node is being set as the master");
         masterAddress = address;
+        if (getClusterService().getClusterVersion() == null) {
+            getClusterService().getClusterStateManager().setClusterVersion(version);
+        }
         setJoined();
         getClusterService().getClusterClock().setClusterStartTime(Clock.currentTimeMillis());
         getClusterService().setClusterId(UuidUtil.createClusterUuid());
+//        getClusterService().changeClusterVersion(version);
```
mdogan: minor: left over, commented out.

hazelcast/src/main/java/com/hazelcast/replicatedmap/impl/ReplicatedMapEventPublishingService.java
```diff
@@ -160,7 +160,7 @@ public void fireMapClearedEvent(int deletedEntrySize, String name) {
     private Member getMember(EventData eventData) {
         Member member = replicatedMapService.getNodeEngine().getClusterService().getMember(eventData.getCaller());
         if (member == null) {
-            member = new MemberImpl(eventData.getCaller(), false);
+            member = new MemberImpl(eventData.getCaller(), nodeEngine.getNode().getVersion(), false);
```
mdogan: minor: `nodeEngine.getVersion()`

vbekiaris:
Thanks @pveentjer , @mdogan , @tombujok , @jerrinot for the reviews & comments.

PR #9244, opened 2016-11-11
[WIP] Bug in Invocation timeout mechanism due to old heartbeats

pveentjer:
In case of retries, it can be that an invocation still has a lastHeartbeatReceived from a previous attempt. When the invocation mechanism kicks in, it checks if the lasthearbeatReceved is 0; if so... then it works based on the invocation time. When the lasthearbeatReceved is set, it will determine based on this lasthearbeatReceved the heartbeat timeout. But this causes problems when a old heartbeat is set and therefor not 0.

To fix the problem. instead of checking lastHeartbeatReceived = 0, it checks if lastHeartbeatReceived < operation.invocationTime.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -570,21 +570,30 @@ public abstract class Invocation implements OperationResponseHandler {
             return NO_TIMEOUT__CALL_TIMEOUT_DISABLED;
         }
 
+        long invocationTime = op.getInvocationTime();
+
         // a call is always allowed to execute as long as its own call timeout
-        long deadlineMillis = op.getInvocationTime() + callTimeoutMillis;
-        if (deadlineMillis > context.clusterClock.getClusterTime()) {
+        if (invocationTime + callTimeoutMillis > context.clusterClock.getClusterTime()) {
             return NO_TIMEOUT__CALL_TIMEOUT_NOT_EXPIRED;
         }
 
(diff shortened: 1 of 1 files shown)


pveentjer:
@mdogan @jerrinot can you have a look.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -570,21 +570,30 @@ HeartbeatTimeout detectTimeout(long heartbeatTimeoutMillis) {
...
+        // It can happen that the lastHeartbeatMillis is from a previous retry. So instead of doing a 0 check,
+        // we check if the invocationTime is past the lastHeartbeatMillis
+        long heartbeatExpirationMillis;
+        if (invocationTime > lastHeartbeatMillis) {
+            // no sensible heartbeat was received.
+
+            // on top of its own call timeout, it is allowed to execute until there is a heartbeat timeout;
+            // so if the callTimeout is five minutes, and the heartbeatTimeout is one minute, then the operation is allowed
+            // to execute for at least six minutes before it is timing out
+            heartbeatExpirationMillis = invocationTime + callTimeoutMillis + heartbeatTimeoutMillis;
```
pveentjer: I think there could be another problem here. InvocationTime is based on 'clusterClock', but we now use invocationTime in the heartbeatExpirationMillis; where it is compared with a regular Clock. This was already in the cde; not part of these changes.

pveentjer:
The problem is a bit more complex. The lastHeartbeatMillis is reset to 0 in case of a retry.

pveentjer:
Closing this one since we went for a simpler approach: resetting the lastHeartbeatMillis once a retry is done.

PR #9245, opened 2016-11-11
Error in ClientHearbeatMonitor connection.close message

pveentjer:
Now should display current time; but is displaying lastTimePacketReceived


hazelcast/src/main/java/com/hazelcast/client/impl/ClientHeartbeatMonitor.java
```diff
@@ -86,14 +86,14 @@ private void monitor(String memberUuid, ClientEndpointImpl clientEndpoint) {
...
+        final long lastTimePacketReceived = connection.lastReadTimeMillis();
         final long timeoutInMillis = SECONDS.toMillis(heartbeatTimeoutSeconds);
         final long currentTimeMillis = Clock.currentTimeMillis();
-        if (lastTimePackageReceived + timeoutInMillis < currentTimeMillis) {
+        if (lastTimePacketReceived + timeoutInMillis < currentTimeMillis) {
             if (memberUuid.equals(clientEndpoint.getPrincipal().getOwnerUuid())) {
                 String message = "Client heartbeat is timed out, closing connection to " + connection
-                        + ". Now: " + timeToString(lastTimePackageReceived)
-                        + ". LastTimePacketReceived: " + timeToString(lastTimePackageReceived);
+                        + ". Now: " + timeToString(Clock.currentTimeMillis())
```
ahmetmircik: This may not be important but just for my curiosity what is the true `now` here? Is it `Clock.currentTimeMillis()` or previously calculated `currentTimeMillis `?

pveentjer: Very good spotting

PR #9246, opened 2016-11-11
Error in ClientHearbeatMonitor connection.close message

pveentjer:
The problem was that the lastPacketReceived time was used for now
instead of current time millis.


jerrinot:
backport of https://github.com/hazelcast/hazelcast/pull/9245

PR #9249, opened 2016-11-14
Configures ssl client with 3 threads

pveentjer:
By default nothig changes.

The client has now option to explicitely configure number of input threads/output threads. This defaults
to 1 if SSL is not enabled, and 3 if SSL enabled (to allign with server).

The reason why 3+3 threads are used for SSL, is that encryption/decryption of the stream is done on the io threads. And client is lot slower than server when SSL is used.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -38,11 +38,17 @@ import com.hazelcast.client.spi.impl.ClientInvocation;
 import com.hazelcast.client.spi.impl.ClientInvocationFuture;
 import com.hazelcast.client.spi.impl.ConnectionHeartbeatListener;
 import com.hazelcast.client.spi.impl.listener.ClientListenerServiceImpl;
+import com.hazelcast.client.spi.properties.ClientProperty;
+import com.hazelcast.config.SSLConfig;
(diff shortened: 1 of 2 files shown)


pveentjer:
It is already tested as part of the serverside logic.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -150,16 +154,37 @@ public ClientConnectionManagerImpl(HazelcastClientInstanceImpl client, AddressTr
...
                 client.getLoggingService(),
                 client.getMetricsRegistry(),
                 new HazelcastThreadGroup(client.getName(), logger, client.getClientConfig().getClassLoader()),
                 outOfMemoryHandler,
-                1,
-                1,
-                0,
+                inputThreads,
+                outputThreads,
+                properties.getInteger(ClientProperty.IO_BALANCER_INTERVAL_SECONDS),
```
pveentjer: If there is only a single thread in the io system, the balancer disabled itself (nothing to do). The 20 seconds are the same as on the server side.

No validation check,but we don't do on the server side:

```
  @Override
    public int getBalancerIntervalSeconds() {
        return node.getProperties().getSeconds(GroupProperty.IO_BALANCER_INTERVAL_SECONDS);
    }
```

pveentjer:
I have merged this PR too fast. I need to get a :+1: from Ihsan; otherwise I'll undo the merge

ihsandemir:
👍

Issue #9255, opened 2016-11-14
MapLoader.loadAll(true) doesn't reload after reloading with new keys


gAmUssA:
Reproducer: https://github.com/amitgsoni/hazelcast

1. Start cache server (Don't load any map) 
2. Load map first time --> Using `com.hazelcast.client.HazelcastClient.getMap().size()` 
3. Reload Update key(s) of given map --> Using `com.hazelcast.client.HazelcastClient getMap().loadAll(keys, true);`

-- Once we do operation of step-3 the getMap().loadAll(true) is not working as expected

ZD 2149

jerrinot:
@gAmUssA: I'm looking at the reproducer. Can you please elaborate the "Once we do operation of step-3 the getMap().loadAll(true) is not working as expected" statement? 

What is supposed to happen and what's actually happening?

gAmUssA:
use case 1

- map has maploader
- hz.getMap invokes loadAllKeys, loadAll
- hz.getMap.loadAll(true) invokes loadAll

all good here

use case 2

- map has maploader
- hz.getMap invokes loadAllKeys, loadAll
- new set of keys
- hz.getMap.loadAdd(keys, true) invokes loadAll for new set of keys
- hz.getMap.loadAll(true) doesn't work.

Thank you

jerrinot:
a further clarification from @gAmUssA: "hz.getMap.loadAll(true) doesn't work." == Map Loader is not invoked

gAmUssA:
@jerriot let me know if you need any further clarification

jerrinot:
I did a little bit of debugging and when `hz.getMap.loadAdd(keys, true)` is called then the subsequent `loadAll(true)` will be no-op because the [state here](https://github.com/jerrinot/hazelcast/blob/4d0231e20ed8d923bf4f666f0ff279cdf8cdeacc/hazelcast/src/main/java/com/hazelcast/map/impl/MapKeyLoader.java#L229) will be still `LOADING`.

jerrinot:
assigning to Tom as discussed.

PR #9259, opened 2016-11-15
[DO NOT MERGE] Do not track key loading process when calling map.loadAll(keys)

jerrinot:
Reasoning:
`loadAll(keys)` does not use coordinator as each member receives all keys
directly from the caller.

A candidate fix for #9255


hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
```diff
@@ -139,8 +139,6 @@ public void loadAllFromStore(List<Data> keys) {
             Future f = recordStoreLoader.loadValues(keys);
             loadingFutures.add(f);
         }
-
-        keyLoader.trackLoading(false, null);
     }
```
jerrinot: I thought it would be too easy! thanks! what would you suggest to do? to split these paths?

PR #9260, opened 2016-11-16
[BACKPORT] loadAll(keys) has to notify RecordStores about the end of loading

jerrinot:
When a RecordStore process LoadAllOperation it expects to receive
LoadStatusOperation eventually. The LoadStatusOperation indicates
there are no more keys to be loaded. Without this the record store never
change its state into the `LOADED` state -> subsequent loadAll() calls are ignored

A candidate for #9255

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapLoadAllTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapLoadAllTest.java
@@ -2,10 +2,13 @@ package com.hazelcast.client.map;
 
 import com.hazelcast.client.test.TestHazelcastFactory;
 import com.hazelcast.config.Config;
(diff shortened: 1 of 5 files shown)


tombujok:
The PR looks good, the only thing that is lacking to me is some documentation/description why we needed to add it, etc. I can imagine we will forget it in a couple of months.

tombujok:
One more comment is that there's interference between loadAll and loadAll(keys) with this fix.
The worst thing that could happen is:

- you do (A.) loadAll which will load all keys (like 10M)
- while the abovementioned call is in progress you do (B.) loadAll (with 1000 keys)
- when B finishes it updates the loading status on all partitions so it means (A.) will think that the load has finished, even though (A.) may be still in progress for a long time.

jerrinot:
closed in favour of https://github.com/hazelcast/hazelcast/pull/9262

PR #9261, opened 2016-11-16
Lock quorum

mmedenjak:
Adding quorum to lock actions. We are reusing the existing quorum support which means that this does not guarantee correctness, just best-effort detection of quorum. This will be emphasised in the  documentation and in the TDD.

The lock configuration element/class was added since lock didn't yet have configuration.

Note to keep in mind when reviewing - adding quorum for locks affects locking methods on map and multimap as well.

### DIFF
--- a/hazelcast/src/main/resources/hazelcast-config-3.8.xsd
+++ b/hazelcast/src/main/resources/hazelcast-config-3.8.xsd
@@ -52,6 +52,7 @@
                 <xs:element name="reliable-topic" type="reliable-topic" minOccurs="0" maxOccurs="unbounded"/>
                 <xs:element name="jobtracker" type="jobtracker" minOccurs="0" maxOccurs="unbounded"/>
                 <xs:element name="semaphore" type="semaphore" minOccurs="0" maxOccurs="unbounded"/>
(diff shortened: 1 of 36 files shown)


hazelcast/src/main/resources/hazelcast-config-3.8.xsd
```diff
@@ -1172,6 +1173,23 @@
             </xs:simpleType>
         </xs:attribute>
     </xs:complexType>
+
+    <xs:complexType name="lock">
```
vbekiaris: Consider adding support for serializing the new elements in `ConfigXmlGenerator` (but feel free to add it later with a separate PR, it's not crucial).

mmedenjak: Right. There's the spring config as well, I will create a new PR for that.

mmedenjak: I've added lock config to config xml generator and spring config. Also fixed queue and ringbuffer config in xml generator.

eminn:
Apart from the comments @vbekiaris mentioned, looks good to be merged 👍

mmedenjak:
@eminn Thanks for the review!

PR #9262, opened 2016-11-16
[BACKPORT] Do not change keyLoaderState when loading keys supplied by user

### LINKED ISSUE #9255: MapLoader.loadAll(true) doesn't reload after reloading with new keys

jerrinot:
The LoadAllOperation is used in 2 contexts:
1. loadAll() - when all entries are reloaded
2. loadAll(keys) - when reloading of specific keys is requested by user

The first case uses a complex key-loading coordination and state transitions,
while the 2nd case is fairly simple - every member receives a set of keys
directly from the user.

This changeset fix the behaviour in the 2nd case where the
state was changed by accident and this led to a buggy
behaviour.

Fixes #9255 
EE counter-part: https://github.com/hazelcast/hazelcast-enterprise/pull/1134


jerrinot:
merging myself as a the same PR into 3.7.3 is already reviewed&merged.

PR #9263, opened 2016-11-16
[3.7.3] Do not change keyLoaderState when loading keys supplied by user

### LINKED ISSUE #9255: MapLoader.loadAll(true) doesn't reload after reloading with new keys

jerrinot:
The LoadAllOperation is used in 2 contexts:
1. loadAll() - when all entries are reloaded
2. loadAll(keys) - when reload of specific keys is requested by user

The first case uses a complex key-loading coordination and state transitions,
while the 2nd case is fairly simple - every member receives a set of keys
directly from the user.

This changeset fix the behaviour in the 2nd case where the
state was changed by accident and this led to a buggy
behaviour.

Fixes #9255
EE-part: https://github.com/hazelcast/hazelcast-enterprise/pull/1135

Port of #9262 into the 3.7.3 release branch
(cherry picked from commit 734509d)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
@@ -215,8 +215,9 @@ public class DefaultMapOperationProvider implements MapOperationProvider {
     }
 
     @Override
-    public MapOperation createLoadAllOperation(String name, List<Data> keys, boolean replaceExistingValues) {
-        return new LoadAllOperation(name, keys, replaceExistingValues);
+    public MapOperation createLoadAllOperation(String name, List<Data> keys, boolean replaceExistingValues,
+                                               boolean withUserSuppliedKeys) {
+        return new LoadAllOperation(name, keys, replaceExistingValues, withUserSuppliedKeys);
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/LoadAllOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/LoadAllOperation.java
@@ -39,15 +39,20 @@ public class LoadAllOperation extends MapOperation implements PartitionAwareOper
     private List<Data> keys;
 
     private boolean replaceExistingValues;
+    private boolean withUserSuppliedKeys;
 
     public LoadAllOperation() {
         keys = Collections.emptyList();
     }
 
-    public LoadAllOperation(String name, List<Data> keys, boolean replaceExistingValues) {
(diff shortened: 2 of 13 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
```diff
@@ -215,8 +215,9 @@ public MapOperation createGetOperation(String name, Data dataKey) {
     }
 
     @Override
-    public MapOperation createLoadAllOperation(String name, List<Data> keys, boolean replaceExistingValues) {
-        return new LoadAllOperation(name, keys, replaceExistingValues);
+    public MapOperation createLoadAllOperation(String name, List<Data> keys, boolean replaceExistingValues,
+                                               boolean localOnlyLoad) {
+        return new LoadAllOperation(name, keys, replaceExistingValues, localOnlyLoad);
```
ahmetmircik: maybe we can use `withUserSuppliedKeys` instead of `localOnlyLoad`. `localOnlyLoad` seemed not clear

hazelcast/src/main/java/com/hazelcast/map/impl/operation/LoadAllOperation.java
```diff
@@ -39,15 +39,20 @@
...
 
-    public LoadAllOperation(String name, List<Data> keys, boolean replaceExistingValues) {
+    public LoadAllOperation(String name, List<Data> keys, boolean replaceExistingValues, boolean localOnlyLoad) {
         super(name);
         this.keys = keys;
         this.replaceExistingValues = replaceExistingValues;
+        this.localOnlyLoad = localOnlyLoad;
+        // the localOnlyLoad field piggy-backs on Operation flags for serialization
+        // as the field was introduced in a patch release
+        setFlag(localOnlyLoad, BITMASK_LOAD_ALL_LOCAL_ONLY);
```
ahmetmircik: nice hack

hazelcast/src/main/java/com/hazelcast/map/impl/operation/DefaultMapOperationProvider.java
jerrinot: can do. @tombujok: WDYT?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/LoadAllOperation.java
jerrinot: I blame Tom for this evil genius-kind of hack! :)

tombujok: Thx Jaromir, let's say it's our joint work ;)

PR #9267, opened 2016-11-17
[DO NOT MERGE] Composite PR generally about operation cancellation and retrying

mtopolnik:


mtopolnik:
run-lab-run

pveentjer:
Looks very good!

cangencer:
verify

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -178,37 +180,84 @@ public final Operation setReplicaIndex(int replicaIndex) {
...
+     *     <li>When an Invocation of the operation is created, call ID is assigned a positive value.</li>
+     *     <li>When the invocation ends, the operation is {@link #deactivate()}d, but the call ID is preserved.</li>
+     *     <li>The same operation may be involved in a further invocation (retrying); this will assign a
+     *     new call ID and reactivate the operation.</li>
+     * </ul>
+     * @return the call ID
      */
     public final long getCallId() {
-        return callId;
+        return Math.abs(callId);
```
pveentjer: Why the abs? call id can't be negative.

pveentjer: I see it. You are using a negative call id for a special purposes.

mtopolnik: Please review https://github.com/hazelcast/hazelcast/pull/9267/commits/a33dd8dc77062ce189583dc7e7283498447cce4d in isolation, this will make things clear.

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -178,37 +180,84 @@ public final Operation setReplicaIndex(int replicaIndex) {
...
+            throw new IllegalArgumentException("Attempted to set non-positive call ID " + newId);
+        }
+        final long c = this.callId;
+        if (c > 0) {
+            throw new IllegalStateException(String.format(
+                    "Attempt to overwrite the call ID of an active operation: current %d, requested %d",
+                    this.callId, newId));
+        }
+        if (!CALL_ID.compareAndSet(this, c, newId)) {
+            throw new IllegalStateException(String.format("Concurrent modification of call ID. Initially observed %d,"
```
pveentjer: Can you include to toString of the operation in the exception?

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -106,13 +106,13 @@ public final boolean isDone() {
...
 
-    private boolean isDone(final Object state) {
+    private static boolean isDone(final Object state) {
         if (state == null) {
             return true;
         }
 
         return !(state == VOID
-                || state.getClass() == WaitNode.class
+                || state instanceof WaitNode
```
pveentjer: why the instanceof check? WaitNode final internal class.

pveentjer: Not terribly important.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/operations/AbstractLockOperation.java
```diff
@@ -28,19 +28,23 @@
...
 
 import java.io.IOException;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
 public abstract class AbstractLockOperation extends Operation
         implements PartitionAwareOperation, IdentifiedDataSerializable {
 
     public static final int ANY_THREAD = 0;
 
+    private static final AtomicLongFieldUpdater<AbstractLockOperation> REFERENCE_CALL_ID =
```
pveentjer: Ok. Makes sense.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationMonitor.java
```diff
@@ -317,38 +307,6 @@ public void run0() {
             log(invocationCount, backupTimeouts, normalTimeouts);
         }
 
-        /**
```
pveentjer: Nice to see this hack getting resolved

hazelcast/src/main/java/com/hazelcast/spi/OperationAccessor.java
```diff
@@ -50,6 +50,15 @@ public static void setCallId(Operation op, long callId) {
     }
 
     /**
+     * Resets the callId for the Operation to zero. For details see {@link Operation#resetCallId()}
+     *
+     * @param op the Operation on which to reset the callId
+     */
```
pveentjer: Can you add the @returns as well?

hazelcast/src/main/java/com/hazelcast/spi/CallsPerMember.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Makes sure that a list of counters is created for a member.
+     * <p/>
+     * This method exists to ensure that an operation-heartbeat is always sent, even if there are no running operations.
+     *
+     * @param address the address of the member.
+     */
+    public void ensureMember(Address address) {
```
cangencer: seems we have two methods doing the same thing?

hazelcast/src/main/java/com/hazelcast/spi/CallsPerMember.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private static long[] toArray(List<Long> longs) {
+        long[] array = new long[longs.size()];
+        for (int k = 0; k < array.length; k++) {
+            array[k] = longs.get(k);
+        }
+        return array;
+    }
+
+    static final class CategorizedCallIds {
```
cangencer: class can be made private

hazelcast/src/main/java/com/hazelcast/spi/LiveOperations.java
```diff
@@ -18,83 +18,12 @@
...
-        }
-
-        return array;
-    }
+public interface LiveOperations {
 
-    public void clear() {
-        callIdsByMember.clear();
-    }
+    void add(Address address, long callId);
```
cangencer: some javadoc might be useful here - Saying that this is the caller's address and call id of the operation which is to be tracked.

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -178,37 +180,85 @@ public final Operation setReplicaIndex(int replicaIndex) {
...
+     */
+    final boolean deactivate() {
+        long c = callId;
+        if (c <= 0) {
+            return false;
+        }
+        if (CALL_ID.compareAndSet(this, c, -c)) {
+            return true;
+        }
+        if (callId > 0) {
```
cangencer: what if callId is set back to 0? or is that impossible? If not, maybe that should fail as well?

mtopolnik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -178,37 +180,85 @@ public final Operation setReplicaIndex(int replicaIndex) {
...
-        return callId;
+        return Math.abs(callId);
+    }
+
+    /**
+     * Tells whether this operation is involved in an ongoing invocation. Such an operation always has a
+     * positive call ID.
+     * @return {@code true} if the operation's invocation is active; {@code false} otherwise
+     */
+    final boolean hasActiveInvocation() {
```
cangencer: might be better to call the method isActive - to match better with "deactivate"

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
```diff
@@ -131,30 +139,32 @@ public long getLastCallId() {
...
         @Override
         public int getMaxConcurrentInvocations() {
             return maxConcurrentInvocations;
         }
 
         @Override
         public long next(Invocation invocation) {
-            assert invocation.op.getCallId() == 0 : "callId should be null:" + invocation;
+            final Operation op = invocation.op;
+            assert !invocation.isActive() : "Invocation is already in progress:" + invocation;
```
cangencer: this check seems redundant, since setCallId will already fail later on with an active invocation. The method can be then simplified as `long next(boolean isUrgent)`, furthermore, it's probably better to make it throw a checked exception and handle the exception at the caller site.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -637,6 +433,180 @@ boolean detectAndHandleBackupTimeout(long timeoutMillis) {
...
+        return allowed;
+    }
+
+    private void invoke0(boolean isAsync) {
+        if (invokeCount > 0) {
+            throw new IllegalStateException("This invocation is already in progress");
+        } else if (isActive()) {
+            throw new IllegalStateException(
+                    "Attempt to reuse the same operation in multiple invocations. Operation is " + op);
+        }
```
cangencer: won't these checks be covered by the setCallId() check later on?

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -637,6 +433,180 @@ boolean detectAndHandleBackupTimeout(long timeoutMillis) {
...
+
+        invokeCount++;
+
+        // register method assumes this method has run before it is being called so that remote is set correctly
+        if (!initInvocationTarget()) {
+            return;
+        }
+
+        setInvocationTime(op, context.clusterClock.getClusterTime());
+        if (!context.invocationRegistry.register(this)) {
```
cangencer: what happens to invokeCount here? looks like it'll never be decremented?

hazelcast/src/main/java/com/hazelcast/spi/CallsPerMember.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    public Set<Address> addresses() {
+        return callIdsByMember.keySet();
+    }
+
+    public OperationControl toOpControl(Address address) {
+        CategorizedCallIds callIds = callIdsByMember.get(address);
+        if (callIds == null) {
+            throw new IllegalArgumentException("unknown address");
```
mdogan: _minor:_ Can you append `address` to the exception message?

hazelcast/src/main/java/com/hazelcast/spi/CallsPerMember.java
mtopolnik: For clarity. I don't want to call `getOrCreateCallIdsForMember`, ignoring the return value, when all i want is to ensure the address is added to the map.

hazelcast/src/main/java/com/hazelcast/spi/CanCancelOperations.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.hazelcast.spi;
+
+import com.hazelcast.nio.Address;
+
+/**
+ * Implemented by a service that can cancel its operations.
+ */
+public interface CanCancelOperations {
```
mtopolnik: This interface is actually symmetrical to `LiveOperationsTracker` and should be implemented by the same classes that implement that one. However these implementations are out of scope in this PR. The primary motivation is cancellation of Jet jobs and we can gradually add more cancelability to main HZ operations.

hazelcast/src/main/java/com/hazelcast/spi/CallsPerMember.java
mtopolnik: Done.

hazelcast/src/main/java/com/hazelcast/spi/LiveOperations.java
mtopolnik: Doen.

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
mtopolnik: Originally it was called `isActive` but was renamed to make it more clear. But perhaps I can rename it back because there is an `isActive` method on Invocation now.

hazelcast/src/main/java/com/hazelcast/spi/Operation.java
mtopolnik: Impossible. Zero is only the starting point and in fact `callId` should never assume the same value twice. The sequence is `0` -> `id1` -> -`id1` -> `id2` -> `-id2` ...

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/CallIdSequence.java
mtopolnik: These are good ideas, I'll implement it that way. About the check, it's  an `assert`, not a business logic-level check. However, after all the changes it does indeed seem redundant.

hazelcast/src/main/java/com/hazelcast/spi/CanCancelOperations.java
mdogan: I see. 
Currently `InvocationFuture.cancel()` is returning false and doing nothing. With your change, future will be completed with `CancellationException` and by default cancellation (until proper impls are provided) still will be a no-op.
I'm not exactly sure, but won't this be an issue? `future.cancel()` returns `true` but actually does nothing.

mtopolnik: Actually the current state is broken with respect to `Future` contract: 
> After this method returns, subsequent calls to `isDone()` will always return `true`.
The new state is correct because `Future` is not job-control API; by `cancel`ing it you simply waive your interest in the eventual return value. Taking no action to percipitate task completion is by the book.

hazelcast/src/main/java/com/hazelcast/spi/CallsPerMember.java
mtopolnik: Done

hazelcast/src/main/java/com/hazelcast/spi/CanCancelOperations.java
mdogan: also contract says:
>  If successful, and this task has not started when cancel is called, this task should never run.

Of course this depends on definition of _this task has not started_ ...

mtopolnik: Yes, since the task is starting concurrently, this can't be anything more than a hint to the implementation.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
mtopolnik: `register` can also throw a `HazelcastOverloadException`. That case is also unhandled.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
cangencer: OK, I'm considering this out of scope for this PR.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
cangencer: same as previous comment - out of scope of this PR..

PR #9270, opened 2016-11-17
Reserve serializer constants for Jet

cangencer:


pveentjer:
Why is a JET addition required? Isn't it simplest to run on top of IdentifiedDataSerializable?

pveentjer:
👍

cangencer:
It is for some very common types like Map.Entry - where we want to have the minimum possible overhead

PR #9272, opened 2016-11-18
Add Management Center integration for Hot Restart partial start

mdogan:
EE counterpart: https://github.com/hazelcast/hazelcast-enterprise/pull/1137


hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -271,4 +273,9 @@ public boolean isMemberExcluded(Address memberAddress, String memberUuid) {
     @Override
     public void handleExcludedMemberUuids(Address sender, Set<String> excludedMemberUuids) {
     }
+
+    @Override
+    public ClusterHotRestartStatusDTO getCurrentClusterHotRestartStatus() {
```
pveentjer: Just nitpicking; does the word 'current' added value? so what about getClusterHotRestartStatus, or just clusterHotRestartStatus without the get. Anyhow.. not that important.

pveentjer:
:+1:

PR #9273, opened 2016-11-18
Client endpoint may have a different connection than it was initially constructed

ihsandemir:
Reverted some of the the changes that was done as part of listener registration fix in terms of endpoint management. I previously changed to reuse the same endpoint for the reconnecting client in order to make the old existing listener be able to deliver the event messages to the same client that was reconnected (Doing this in AbstractBaseTask.sendClientMessage). But this required that the endpoint connection can be changed at some time in the future, when the client reconnected, but then the endpoint properties may become different from what the initial endpoint connection represented. This does not cause any functionality problem but it is clearer this way to have a constant connection in the endpoint.

One other problem with the previous change was that the server was actively closing an existing client connection during authentication (AuthenticationBaseMessageTask.checkExistingEndpoint()) which not the expected behaviour for the client. It was causing unnecessary disconnects and this was causing the c++ client tests fail.

This change allowed to revert back some of the interface changes as well.

Also, added a reason to be printed when the endpoint is being destroyed. Since, this is an internal interface, I am hoping that this won't be a problem.

Please review the change at AbstractMessageTask.sendClientMessage and AbstractMessageTask.findSendConnection. I overwrite findSendConnection for PingMessageTask since the ping should always use the same connection on the response. The reason for findSendConnection is for event existing handlers to be able to deliver an event for a client which actually changed its connection and reconnected to the same member. In this case, the original connection of the event handler at the message task (e.g. AbstractMapAddEntryListenerMessageTask.ClientMapListener) is closed, it needs to send the messages over a new live connection if possible.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/ClientEndpointManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/ClientEndpointManager.java
@@ -47,13 +47,6 @@ public interface ClientEndpointManager {
     ClientEndpoint getEndpoint(Connection connection);
 
     /**
-     * Gets the endpoint of a client with a provided client uuid.
-     * @param clientUuid The uuid of the client for which you want the endpoint.
-     * @return The found endpoint or null of no endpoint was found.
-     */
-     ClientEndpoint getEndpoint(String clientUuid);
-
-    /**
      * Gets all the endpoints for a given client.
      *
      * @param clientUuid the uuid of the client
@@ -93,10 +86,11 @@ public interface ClientEndpointManager {
      * todo: what happens when the endpoint was never registered
      *
      * @param endpoint the endpoint to remove.
+     * @param reason The reason why the endpoint is being removed
      * @throws java.lang.NullPointerException if endpoint is null.
-     * @see #removeEndpoint(ClientEndpoint, boolean)
+     * @see #removeEndpoint(ClientEndpoint, boolean, String)
      */
-    void removeEndpoint(ClientEndpoint endpoint);
+    void removeEndpoint(ClientEndpoint endpoint, String reason);
 
     /**
      * Removes an endpoint and optionally closes it immediately.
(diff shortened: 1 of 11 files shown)


ihsandemir:
run-lab-run

Donnerbart:
Will this also solve the issue with this test? https://github.com/hazelcast/hazelcast/pull/8848/files#diff-dcbe5133f1970761105f659dbabcf8a6R86

I think you ignored it when you introduced some of the changes you are reverting now.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEndpointImpl.java
```diff
@@ -89,7 +84,15 @@ public String getUuid() {
 
     @Override
     public boolean isAlive() {
```
vbekiaris: So a `ClientEndpoint` is alive as long as its connection or another connection from this member to the same client UUID is alive? Isn't it risky that one might do something like:
```
if (endpoint.isAlive()) {
  endpoint.getConnection().write(frame);
}
```
and fail as `endpoint.getConnection().isAlive()` may be `false` even though `endpoint.isAlive()==true`?

ihsandemir: The client does not use connection.write but it always sends the message to the client using AbstractMessageTask.sendClientMessage which does this: final Connection endpointConnection = findSendConnection(); 

We need this kind of logic since for the listeners the connection to the client may change during the client lifetime when the client reconnects to the cluster before it is cleaned up and findSendConnection ensures to use the correct connection.

hazelcast/src/main/java/com/hazelcast/client/ClientEndpointManager.java
```diff
@@ -93,21 +86,30 @@
...
      *
-     * @param endpoint the endpoint to remove.
+     * @param ce the endpoint to remove.
      * @param closeImmediately if the endpoint is immediately closed.
+     * @param reason The reason why the endpoint is being removed.
      * @throws java.lang.NullPointerException if endpoint is null.
-     * @see #removeEndpoint(ClientEndpoint)
+     * @see #removeEndpoint(ClientEndpoint, String)
+     */
+    void removeEndpoint(ClientEndpoint ce, boolean closeImmediately, String reason);
```
ihsandemir: it was ce originally but I now changed it to clientEndpoint.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -44,8 +44,8 @@
 
     protected final ClientMessage clientMessage;
 
-    protected final Connection connection;
-    protected ClientEndpoint endpoint;
+    protected volatile Connection connection;
```
ihsandemir: I did a final commit now where I update the connection during findSendConnection method, this was the original reason for it which I later forgot to update. I updated it in the latest commit now.

vbekiaris:
needs rebase

ihsandemir:
run-lab-run

PR #9277, opened 2016-11-21
WAN enhancements

emrahkocaman:
- Config changes for new WANQueueFulBehavior
- New REST method to allow clearing WRQs.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
@@ -64,6 +64,8 @@ public class HttpPostCommandProcessor extends HttpCommandProcessor<HttpPostComma
                 handleShutdownNode(command);
             } else if (uri.startsWith(URI_WAN_SYNC_MAP)) {
                 handleWanSyncMap(command);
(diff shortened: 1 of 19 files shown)


emrahkocaman:
verify

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
```diff
@@ -323,6 +325,25 @@ private void handleWanSyncMap(HttpPostCommand command) throws UnsupportedEncodin
         textCommandService.sendResponse(command);
     }
 
+    private void handleWanClearQueues(HttpPostCommand command) throws UnsupportedEncodingException {
```
mesutcelik: That must be called after WAN Rep Stopped via ManCenter. right?

mesutcelik:
👍 with one minor comment

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
emrahkocaman: Yes but WAN replication is paused/resumed automatically when clear queues command is in progress.

PR #9278, opened 2016-11-21
Shade/relocate tests jar too

mdogan:
com.eclipsesource.minimal-json artifact is shaded our release jar
and `com.eclipsesource.json` package is relocated to
`com.hazelcast.com.eclipsesource.json` to avoid conflicts with user classpath.

But entries in test jar remain untouched, which makes impossible to test
Management Center requests on EE side.

With this change references to `com.eclipsesource.json` in test classes are
relocated as `com.hazelcast.com.eclipsesource.json` too.

See EE PR https://github.com/hazelcast/hazelcast-enterprise/pull/1137


pveentjer:
:+1:

PR #9281, opened 2016-11-21
Split brain handling: only allow clusters with same major.minor version to merge

vbekiaris:
A separate `JoinMessage` subclass is introduced, specifically for split-brain handling.
In `JoinCheckOperation` and `AbstractJoiner.shouldMerge`, we ensure clusters' `major.minor` versions match before merging.
Introduces extension hooks in `SplitBrainTestSupport` to allow for tests which result in separate clusters.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/SplitBrainJoinMessage.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/SplitBrainJoinMessage.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.cluster.impl;
+
+import com.hazelcast.nio.Address;
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.version.Version;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * A {@code JoinMessage} issued by the master node of a subcluster to the master of another subcluster
+ * while searching for other clusters for split brain recovery.
+ */
+public class SplitBrainJoinMessage extends JoinMessage {
+
+    protected Version clusterVersion;
+
+    public SplitBrainJoinMessage() {
+    }
+
+    public SplitBrainJoinMessage(byte packetVersion, int buildNumber, Version version, Address address, String uuid,
+                                 boolean liteMember, ConfigCheck configCheck, Collection<Address> memberAddresses,
+                                 int dataMemberCount, Version clusterVersion) {
+        super(packetVersion, buildNumber, version, address, uuid, liteMember, configCheck, memberAddresses, dataMemberCount);
+        this.clusterVersion = clusterVersion;
+    }
+
+    @Override
+    public void readData(ObjectDataInput in)
+            throws IOException {
+        super.readData(in);
+        clusterVersion = in.readObject();
+    }
+
+    @Override
+    public void writeData(ObjectDataOutput out)
+            throws IOException {
+        super.writeData(out);
+        out.writeObject(clusterVersion);
+    }
+
+    @Override
+    public String toString() {
+        return "SplitBrainJoinMessage{"
+                + "packetVersion=" + packetVersion
+                + ", buildNumber=" + buildNumber
+                + ", codebaseVersion=" + version
+                + ", clusterVersion=" + clusterVersion
+                + ", address=" + address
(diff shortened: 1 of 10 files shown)


vbekiaris:
run-lab-run

vbekiaris:
Unrelated test failure `testClusterReconnectDueToHeartbeatSmartRouting`

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/SplitBrainJoinMessage.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Collection;
+
+/**
+ * A {@code JoinMessage} issued by the master node of a subcluster to the master of another subcluster
+ * while searching for other clusters for split brain recovery.
+ */
+public class SplitBrainJoinMessage extends JoinMessage {
+
+    protected Version clusterVersion;
+
```
tombujok: Shouldn't we also send the master node's codebase version?

vbekiaris: It is sent in `JoinMessage.version`, however when attempting to locate clusters for merge, the master node's codebase version is irrelevant in this implementation.

tombujok: OK. My only concern is that the master knows the codebase version of the other master. If that's the case I'm fine with it.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/JoinCheckOperation.java
```diff
@@ -61,12 +66,13 @@ public void run() {
             ILogger logger = getLogger();
             try {
                 if (service.getClusterJoinManager().validateJoinMessage(request)) {
-                    try {
-                        // also validate other node's version is able to join our cluster
-                        node.getNodeExtension().validateJoinRequest(request);
+                    // also validate other cluster's major.minor version is same as this cluster
```
mmedenjak: Isn't this what `node.getNodeExtension().validateJoinRequest(request)` will do anyway - check the cluster version? Why was that method call removed? Because it throws an exception instead of just returning false?

vbekiaris: Not really, `NodeExtension.validateJoinRequest(request)` validates `JoinMessage.version` (ie. the joining node's codebase version) is OK to join the cluster. In the specific case, it would validate that the subcluster's master node codebase version is OK to join the other subcluster, but this does not provide any indication about our ability to merge one cluster to another. That is why it is replaced with validation of cluster-to-cluster version `major.minor` equality.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/JoinCheckOperation.java
```diff
@@ -29,17 +29,22 @@
 
 import java.io.IOException;
 
+import static com.hazelcast.version.Version.MAJOR_MINOR_VERSION_COMPARATOR;
+
+/**
+ * Validate whether clusters may merge to recover from a split brain, based on configuration & cluster version.
+ */
 public class JoinCheckOperation extends AbstractJoinOperation {
```
tombujok: Can't we name this operation SplitBrainMergeCheckOperation?
It's confusing, I thought it's invoked on normal join.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/JoinCheckOperation.java
tombujok: Could we add a comment, why they have to be the same, e.g. it will lead to sneaky-downgrade or sneaky-upgrade?
We may simply forget it in the future...

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/JoinCheckOperation.java
vbekiaris: 👍  renamed to `SplitBrainMergeValidationOperation`

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/JoinCheckOperation.java
vbekiaris: Done

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/SplitBrainMergeValidationOperation.java
```diff
@@ -61,12 +66,25 @@ public void run() {
...
+                    // unified cluster but some members might be kicked out due to incompatibility. For example
+                    // assuming a 3.8.0 cluster with 2 nodes at codebase versions 3.8.0 & 3.9.0 (master) and another
+                    // cluster with 3x3.9.0 nodes at cluster version 3.9.0: if we only validated on master nodes' codebase
+                    // version, we would find them to be compatible and let the smaller cluster merge towards the bigger one.
+                    // However we would end up with a cluster at cluster version 3.9.0 (including 4x3.9.0 nodes) and the
+                    // 3.8.0-codebase node would be kicked out of the cluster. To enable the kicked-out node to join again
+                    // the cluster, the user would be forced to upgrade the member to 3.9.0 codebase version.
+                    // The implicit change of cluster version ("sneaky upgrade") and the change in membership would be a
+                    // surprise to users and may cause unexpected issues.
+                    if (MAJOR_MINOR_VERSION_COMPARATOR.compare(service.getClusterVersion(), request.getClusterVersion()) == 0) {
```
tombujok: Great description!

PR #9282, opened 2016-11-21
Backport https://github.com/hazelcast/hazelcast/pull/9273

ihsandemir:
Backports https://github.com/hazelcast/hazelcast/pull/9273

Reverted some of the the changes that was done as part of listener registration fix in terms of endpoint management. I previously changed to reuse the same endpoint for the reconnecting client in order to make the old existing listener be able to deliver the event messages to the same client that was reconnected (Doing this in AbstractBaseTask.sendClientMessage).

Also, added a reason to be printed when the endpoint is being destroyed. Since, this is an internal interface, I am hoping that this won't be a problem.


ihsandemir:
run-lab-run

ihsandemir:
run-lab-run

ihsandemir:
run-lab-run

PR #9285, opened 2016-11-21
Reset hot restart data before join

metanet:
Joiner impls check if the node needs to reset its hot restart data. If that it is the case, the node resets its hot restart data to be able to join to the cluster.

This change pollutes join mechanism because of hot restart mechanics. However, without it, users have to clear hot restart data manually if they start excluded nodes after the cluster is started.

EE counterpart: https://github.com/hazelcast/hazelcast-enterprise/pull/1141


mdogan:
run-lab-run

Issue #9251, opened 2016-11-14
Problems in timeout due to old heartbeats


pveentjer:
In case of retries, it can be that an invocation still has a lastHeartbeatReceived from a previous attempt. When the invocation mechanism kicks in, it checks if the lasthearbeatReceved is 0; if so... then it works based on the invocation time. When the lasthearbeatReceved is set, it will determine based on this lasthearbeatReceved the heartbeat timeout. But this causes problems when a old heartbeat is set and therefor not 0.

To fix the problem. instead of checking lastHeartbeatReceived = 0, it checks if lastHeartbeatReceived < operation.invocationTime.

See https://github.com/hazelcast/hazelcast/pull/9286

PR #9286, opened 2016-11-21
Invocation lastHeartbeatMillis reset in case of retry

pveentjer:
In case of a retrying operation, an old heartbeat could still be stored
on the invocation. If the invocation monitor sees this; it currently doesn't
understand that this from an old invocation; it will just conclude that the
invocation has not received any recent heartbeats and will abort the invocation
with an OperationTimeoutException.

This pr fixes that by resetting the lastHeartbeat when the invocation is retried.

Fix https://github.com/hazelcast/hazelcast/issues/9251


jerrinot:
@pveentjer: backport?

pveentjer:
Yes. I'll create a backport

PR #9287, opened 2016-11-21
Invocation lastHeartbeatMillis reset in case of retry

pveentjer:
In case of a retrying operation, an old heartbeat could still be stored
on the invocation. If the invocation monitor sees this; it currently doesn't
understand that this from an old invocation; it will just conclude that the
invocation has not received any recent heartbeats and will abort the invocation
with an OperationTimeoutException.

This pr fixes that by resetting the lastHeartbeat when the invocation is retried.

backport of https://github.com/hazelcast/hazelcast/pull/9286


Donnerbart:
What is the second file in this PR? An empty PID file?

pveentjer:
I'll remove it. No idea what it is. Apart from this file; you are fine with it?

Donnerbart:
Yes, otherwise it seems to be a proper 1:1 backport.

PR #9290, opened 2016-11-22
fixed aws address translator public ip issue

bilalyasar:
this is back port of #8996


AjitDas:
@bilalyasar - am having problem with doing get() put() delete() operation with HazelcastClient AWS APIs where as value() and clear() works, more information you can find issue #9419

PR #9292, opened 2016-11-22
small fix for aws address translator

bilalyasar:
This modification doesn't affect default translator.

In Aws, due to public/private terms, we need to convert ip-address sometimes.

This fix is tested from me and from an user.


mesutcelik:
@asimarslan @ihsandemir 

Can you please have a look at this? 
👍 from my side...

In case of [DefaultAddressTranslator](https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/DefaultAddressTranslator.java#L29), there will be no logic but [AwsAddressTranslator](https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/AwsAddressTranslator.java) applies some logic to get the right address.  It is a small change but required to have hz-client to work with AWS environment.

asimarslan:
👍

PR #9298, opened 2016-11-23
Retain call ID after invocation completes

mtopolnik:
* Operation.getCallId() on an inactive operation now keeps returning the call ID of the last invocation, until another call ID is set.
* Operation.setCallId() prevents overwriting the call ID on an active operation.
* Operation.deactivate() must now be used to deactivate the operation.
* Operation.hasActiveInvocation() can be used to check activation state of operation.


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -346,7 +346,7 @@ private void notifyWithExceptionWhenTargetIsNull() {
 
     @Override
     public void sendResponse(Operation op, Object response) {
-        if (!RESPONSE_RECEIVED.compareAndSet(this, FALSE, TRUE)) {
+        if (RESPONSE_RECEIVED.getAndSet(this, TRUE)) {
```
pveentjer: why did you change this? The compareAndSet is more clear since it is more obvious you get a one time shot.

mtopolnik: I'll revert this.

pveentjer:
:+1:

PR #9299, opened 2016-11-23
[FORWARDPORT] WAN sync enhancements

emrahkocaman:
- Config changes for new WANQueueFulBehavior
- New REST method to allow clearing WRQs.
- New REST method to allow sync all maps

Forwardport of https://github.com/hazelcast/hazelcast/pull/9277

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/internal/management/ClearWanQueuesTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/internal/management/ClearWanQueuesTest.java
@@ -1,0 +1,56 @@
+package com.hazelcast.internal.management;
+
+import com.eclipsesource.json.JsonObject;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.instance.Node;
+import com.hazelcast.internal.management.operation.ClearWanQueuesOperation;
+import com.hazelcast.internal.management.request.ChangeWanStateRequest;
+import com.hazelcast.internal.management.request.ClearWanQueuesRequest;
+import com.hazelcast.test.HazelcastParallelClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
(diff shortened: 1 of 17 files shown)


hazelcast/src/test/java/com/hazelcast/internal/management/ClearWanQueuesTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Before
+    public void setUp() {
+        HazelcastInstance hz = createHazelcastInstance();
+        Node node = getNode(hz);
+        managementCenterService = node.getManagementCenterService();
+    }
+
+    @Test
+    public void testOperationDefaultConstructor() {
+        ClearWanQueuesOperation operation = new ClearWanQueuesOperation();
```
pveentjer: Testing a default constructor isn't providing a lot of value since it is designed to do nothing.

Also no verification is done on the state (which would be pointless anyway). But verifying of the created object really is created, is not proving much value since otherwise we would have a major JVM bug at our hands.

I would remove this test since it doesn't provide any value. And in the testSerialization you are already testing the default constructor due to the serialization.

hazelcast/src/test/java/com/hazelcast/internal/management/ClearWanQueuesTest.java
```diff
@@ -0,0 +1,62 @@
...
+        assertNotNull(operation);
+    }
+
```
pveentjer: I don't see a test where the actual behavior is being tested. So you you send a ClearWanQueuesRequest, but you don't verify that the wan queue actually has been cleared.

emrahkocaman: It's test is on EE PR.

emrahkocaman:
Merging this PR as it's backport is already reviewed.

PR #9303, opened 2016-11-23
Retry the invocation only if managed to deregister

mtopolnik:
Avoids a race between two retrying threads which could result in a leak inside the Invocation Registry. If two threads initiate the retrying of the same operation, each thread will create a new call ID, but only the last call ID update will be preserved by the Operation instance. Upon completion, the overwritten call ID will fail to be removed from the Invocation Registry.

If the operation was never registered, every thread is allowed to retry it. This is because invocation can fail even before registration and then deregistration always fails. Allowing retrying to all threads seems safe because there is no risk of concurrent retrying of an operation that didn't even make it to registration.

The race is still there that can lead to multiple concurrent operation retries. Only one retry invocation will be found in the registry; others will be unaccounted for.

Also, if the invocation was registered before, but then a retry fails before re-registering, this will lead to the loss of any progress in retrying. It is currently thought that failing before re-registering won't ever happen.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -174,7 +174,7 @@ public abstract class Invocation implements OperationResponseHandler {
 
     @Override
     public void sendResponse(Operation op, Object response) {
-        if (RESPONSE_RECEIVED.getAndSet(this, TRUE)) {
+        if (!RESPONSE_RECEIVED.compareAndSet(this, FALSE, TRUE)) {
             throw new ResponseAlreadySentException("NormalResponse already responseReceived for callback: " + this
                     + ", current-response: : " + response);
         }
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -580,9 +579,7 @@ private void handleRetry(Object cause) {
 
         if (future.interrupted) {
             complete(INTERRUPTED);
-        } else {
-            context.invocationRegistry.deregister(this);
-
+        } else if (op.getCallId() == 0 || context.invocationRegistry.deregister(this)) {
```
mdogan: `op.getCallId() == 0` is for the case invocation failed even before registering, right?

mtopolnik: Yes. It happens when `invTarget == null`.

PR #9306, opened 2016-11-23
NodeState support for Management Center

tombujok:


pveentjer:
Looks good. One comment about NodeState being an interface.

hazelcast/src/main/java/com/hazelcast/monitor/NodeState.java
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.monitor;
+
+import com.hazelcast.cluster.ClusterState;
+import com.hazelcast.internal.management.JsonSerializable;
+import com.hazelcast.version.Version;
+
+public interface NodeState extends JsonSerializable {
```
tombujok: That's a good observation. I just followed the convention they is use there - all of the classes are divided into interface and impl there:
https://github.com/hazelcast/hazelcast/tree/master/hazelcast/src/main/java/com/hazelcast/monitor

pveentjer: Ok. Then it is fine with me.

pveentjer:
:+1:

PR #9307, opened 2016-11-23
Invocation cleanup

mtopolnik:
* fix call ID sequence leak in `InvocationRegistry.register()`
* add a convenience method `Invocation.isActive()`
* improve diagnostic messages in exceptions
* add more Javadoc and code comments
* reduce the coupling of `CallIdSequence.next()`: instead of the whole `Invocation` instance, now it just takes a boolean argument. This dramatically simplifies unit testing.
* reuse `BackoffIdleStrategy` in `CallIdSequence#waitForSpace` 
* add tests for the improved call ID logic on Operation
* silence the `warn` logging on double completion of invocation futures: don't complain when the double completion involves a cancelation.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
@@ -17,6 +17,7 @@
 package com.hazelcast.spi.impl.operationservice.impl;
 
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
+import com.hazelcast.core.HazelcastOverloadException;
 import com.hazelcast.core.MemberLeftException;
 import com.hazelcast.internal.metrics.MetricsProvider;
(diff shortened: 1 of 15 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -102,8 +104,13 @@ long getLastCallId() {
      * @return false when InvocationRegistry is not alive and registration is not successful, true otherwise
      */
     public boolean register(Invocation invocation) {
-        long callId = callIdSequence.next(invocation);
-        // Fails if the operation is already active
+        final long callId;
+        try {
+            callId = callIdSequence.next(invocation.op.isUrgent());
+        } catch (TimeoutException e) {
+            throw new HazelcastOverloadException("Failed to start invocation due to overload: " + invocation);
```
pveentjer: maybe good to include the cause?

pveentjer:
👍

PR #9310, opened 2016-11-24
Fix racy MockConnectionManager.stop()

mdogan:
In `MockConnectionManager.stop()`, other (possible) cluster members
are instructed to remove shutting down local member.
There's a branch skipping that when target member doesn't know local member.
But if a new member is already joining at that time, it may not be able to
notice shutting down member.
Removed that skipping branch, since `clusterService.removeAddress()` checks
both address and uuid of a member under cluster service lock.

Fixes https://github.com/hazelcast/hazelcast-enterprise/issues/879


pveentjer:
:+1:

PR #9312, opened 2016-11-24
Mock connection manager processes the buffered incoming messages late

ihsandemir:
Makes sure that the incoming buffered messages waiting in the buffer  are immediately processed while switching from blocked to unblocked state.


sancar:
run-lab-run

PR #9314, opened 2016-11-24
Threads with name 'pool' fixes

### LINKED ISSUE #9210: thread names assigned to all threads created by Hazelcast
it would be nice to have names assigned to all threads created by Hazelcast (and not default ones eg: pool-2-thread-1, pool-2-thread-2)
I have found 2 case where Hazelcast creates ThreadPool - but without name pattern
TcpIpConnectionManager:
private final ScheduledExecutorService scheduler = new ScheduledThreadPoolExecutor(4);
and
MetricsRegistryImpl
private final ScheduledExecutorService scheduledExecutorService = new ScheduledThreadPoolExecutor(2);

Have nice day (-:

pveentjer:
MetricsRegistryImpl and TcpIpConnectionManager create a bunch of threads
but the created threads don't have an clear name. This pr fixes that.

Fixes #9210

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
@@ -25,6 +25,7 @@ import com.hazelcast.internal.networking.SocketChannelWrapper;
 import com.hazelcast.internal.networking.SocketChannelWrapperFactory;
 import com.hazelcast.internal.networking.nonblocking.NonBlockingIOThreadingModel;
 import com.hazelcast.internal.networking.nonblocking.iobalancer.IOBalancer;
+import com.hazelcast.internal.util.concurrent.ThreadFactoryImpl;
 import com.hazelcast.internal.util.counters.MwCounter;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.logging.LoggingService;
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
```diff
@@ -129,7 +130,8 @@ public TcpIpConnectionMonitor createNew(Address endpoint) {
     @Probe
     private final MwCounter closedCount = newMwCounter();
 
-    private final ScheduledExecutorService scheduler = new ScheduledThreadPoolExecutor(4);
+    private final ScheduledExecutorService scheduler
+            = new ScheduledThreadPoolExecutor(4, new ThreadFactoryImpl("TcpIpConnectionManager-thread-"));
```
pveentjer: I think I can do that

pveentjer: There will be a quite a big set of changes required because now there is going to be a reliance of the HazelcastThreadGroup. Not only on the server, but also on the client.

pveentjer:
Too much work. I'm going to close the ticket and let the users deal with their pool threads.

pveentjer:
I'm going to re-open this.

My suggestion is the following. I'll create a new github ticket where we add the naming of cluster group. The current naming (without the clustergroup) is already an improvement because at least it is clear who created the threads. Also in most cases people have only a single hz instance

PR #9318, opened 2016-11-25
Disambiguates usage of version information

vbekiaris:
Version information is broken down to 3 separate entities, according to its usage:
* Version (in nio package) is used in serialization to determine the serialization version.
* MemberVersion (major.minor.patch) determines the current node's codebase version.
* ClusterVersion is the version at which the cluster operates (major.minor only).

Even though these entities share some parts, they do not share any hierarchy on purpose to ensure clean separation.
Requires an EE counterpart (https://github.com/hazelcast/hazelcast-enterprise/pull/1154).


mdogan:
I'm not sure about `NodeVersion` name. We don't use `node` word in public api as far as I remember. `Member.getVersion()` returns `NodeVersion`...
Since this is codebase version, as major.minor.patch, why don't we call it just `Version`?

vbekiaris:
Plain `Version` is used in versioned serialization API (in `com.hazelcast.nio.Version`), so it might create confusion when one needs to import both. I think we should choose something like:
(1) `com.hazelcast.nio.Version` -> `com.hazelcast.nio.SerializationVersion` and `com.hazelcast.version.NodeVersion` -> `com.hazelcast.version.Version`
or
(2) `com.hazelcast.version.NodeVersion` -> `com.hazelcast.version.MemberVersion`

mdogan:
@vbekiaris: `MemberVersion` sounds better :)

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
```diff
@@ -165,15 +165,15 @@ private void handleChangeClusterVersion(HttpPostCommand command) throws Unsuppor
...
+                    version = ClusterVersion.of(versionParam);
                     clusterService.changeClusterVersion(version);
                     res = res.replace("${STATUS}", "success");
-                    res = res.replace("${VERSION}", clusterService.getClusterVersion().asString());
+                    res = res.replace("${VERSION}", clusterService.getClusterVersion().toString());
                 } catch (Exception ex) {
                     res = res.replace("${STATUS}", "fail");
-                    res = res.replace("${VERSION}", clusterService.getClusterVersion().asString());
+                    res = res.replace("${VERSION}", clusterService.getClusterVersion().toString());
                 }
```
tombujok: This was on purpose asString, since toString may be more like a debug info or sth changed by someone in the future, not being aware it's being shown to the user in GUI. but feel free to decide by yourself.

hazelcast/src/main/java/com/hazelcast/nio/VersionAware.java
```diff
@@ -16,16 +16,14 @@
...
 /**
  * Enables getting the version from the implementing object.
  * It may be any Version (cluster version, node version, custom version, etc.) It's up to the implementer.
  */
 public interface VersionAware {
 
     /**
-     * @return the version or Version.UNKNOWN if version is unknown to the object.
+     * @return the version or {@code null} if version is unknown to the object.
      */
```
tombujok: It will require the users to do a null check in the conditional-deserialzation. I don't know what's better though...

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
tombujok: 👍

hazelcast/src/main/java/com/hazelcast/nio/VersionAware.java
tombujok: 👍

PR #9319, opened 2016-11-25
Register invocation before setting target

mtopolnik:
This avoids the hack where double retrying is allowed if call ID is 0 on an invocation.


mtopolnik:
verify

mtopolnik:
I'm merging this now because there are some ongoing random test failures. This may or may not help but is a low-risk change.

PR #9322, opened 2016-11-25
Clusters fail to merge when hashcode of both sides' address are same

mdogan:
When both sub-clusters have the same number of cluster data members
then we determine merging side with a pure function that was comparing hashcodes
of addresses of both sides. If hashcodes are same then both sides expect
other one to be merging side and merge never happens.

Replaced this function with one that's using `String.compareTo()` with
`address:port`. Since it's guaranteed that addresses will not be the same,
this function will always produce a non-zero result.


mdogan:
run-lab-run

mdogan:
run-lab-run

mdogan:
Original is merged already: https://github.com/hazelcast/hazelcast/pull/9323

PR #9324, opened 2016-11-25
Fix issue with MockConnectionManager.stop when nodes stopped concurrently

mdogan:
When nodes terminates concurrently, `clusterService.removeAddress` can fail because
target node is already shutdown.

This issue is introduced by https://github.com/hazelcast/hazelcast/pull/9311


mdogan:
run-lab-run

mdogan:
run-lab-run

mdogan:
Original is merged already: #9323

Issue #8728, opened 2016-08-16
SecondsBasedEntryTaskSchedulerTest.test_doNotRescheduleEntryWithinSameSecond_postpone


hasancelik:
```
java.lang.AssertionError: First schedule() call should always be successful
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at com.hazelcast.util.scheduler.SecondsBasedEntryTaskSchedulerTest.test_doNotRescheduleEntryWithinSameSecond_postpone(SecondsBasedEntryTaskSchedulerTest.java:85)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-ZuluJDK7/com.hazelcast$hazelcast/299/testReport/junit/com.hazelcast.util.scheduler/SecondsBasedEntryTaskSchedulerTest/test_doNotRescheduleEntryWithinSameSecond_postpone/

Donnerbart:
```
java.lang.AssertionError: First schedule() call should always be successful
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.util.scheduler.SecondsBasedEntryTaskSchedulerTest.test_doNotRescheduleEntryWithinSameSecond_postpone(SecondsBasedEntryTaskSchedulerTest.java:85)
```
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/4330/testReport/junit/com.hazelcast.util.scheduler/SecondsBasedEntryTaskSchedulerTest/test_doNotRescheduleEntryWithinSameSecond_postpone/

Donnerbart:
```
java.lang.AssertionError: First schedule() call should always be successful
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.util.scheduler.SecondsBasedEntryTaskSchedulerTest.test_doNotRescheduleEntryWithinSameSecond_postpone(SecondsBasedEntryTaskSchedulerTest.java:85)
```
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/4561/testReport/junit/com.hazelcast.util.scheduler/SecondsBasedEntryTaskSchedulerTest/test_doNotRescheduleEntryWithinSameSecond_postpone/

Donnerbart:
[...]

PR #9328, opened 2016-11-26
Instantiate and register split brain messages listener once for the lifetime of a node

vbekiaris:
This PR promotes the existing anonymous listener in `searchForOtherClusters` to top-level class and instantiates & registers the listener just once, during `MulticastJoiner` construction.

Reasoning:
* do not miss any split-brain join messages sent while listener is not registered
* do not produce litter on each execution of split brain handler

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastJoiner.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastJoiner.java
@@ -24,8 +24,8 @@ import com.hazelcast.util.Clock;
 import com.hazelcast.util.EmptyStatement;
 import com.hazelcast.util.RandomPicker;
 
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.BlockingDeque;
+import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -40,9 +40,14 @@ public class MulticastJoiner extends AbstractJoiner {
     private final AtomicInteger currentTryCount = new AtomicInteger(0);
     private final AtomicInteger maxTryCount;
 
+    // this deque is used as a stack, the SplitBrainMulticastListener adds to its head and the periodic split brain handler job
+    // also polls from its head.
+    private final BlockingDeque<SplitBrainJoinMessage> splitBrainJoinMessages = new LinkedBlockingDeque<SplitBrainJoinMessage>();
+
     public MulticastJoiner(Node node) {
         super(node);
         maxTryCount = new AtomicInteger(calculateTryCount());
+        node.multicastService.addMulticastListener(new SplitBrainMulticastListener(node, splitBrainJoinMessages));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


vbekiaris:
PR builder reports compilation errors in aggregations test:
```
02:41:54 [ERROR] /home/jenkins/jenkins/workspace/new-lab-fast-pr/hazelcast/src/test/java/com/hazelcast/aggregation/SumAggregationTest.java:[159,49] type parameters of <T>T cannot be determined; no unique maximal instance exists for type variable T with upper bounds long,java.lang.Number
02:41:54 [ERROR] /home/jenkins/jenkins/workspace/new-lab-fast-pr/hazelcast/src/test/java/com/hazelcast/aggregation/SumAggregationTest.java:[187,49] type parameters of <T>T cannot be determined; no unique maximal instance exists for type variable T with upper bounds long,java.lang.Number
02:41:54 [ERROR] /home/jenkins/jenkins/workspace/new-lab-fast-pr/hazelcast/src/test/java/com/hazelcast/aggregation/SumAggregationTest.java:[219,51] type parameters of <T>T cannot be determined; no unique maximal instance exists for type variable T with upper bounds double,java.lang.Number
02:41:54 [ERROR] /home/jenkins/jenkins/workspace/new-lab-fast-pr/hazelcast/src/test/java/com/hazelcast/aggregation/SumAggregationTest.java:[251,51] type parameters of <T>T cannot be determined; no unique maximal instance exists for type variable T with upper bounds double,java.lang.Number
```

tombujok:
run-lab-run

tombujok:
@Donnerbart @vbekiaris There have been some changes in the aggregation tests with a direct commit and no review? 
See the history: https://github.com/hazelcast/hazelcast/commits/master/hazelcast/src/test/java/com/hazelcast/aggregation/SumAggregationTest.java

Donnerbart:
run-lab-run

vbekiaris:
Thanks @tombujok @Donnerbart .
Updated to use a `BlockingDeque` as a stack so the split brain join message processed will always be the latest received: the listener adds each message received to the deque's head, split brain job polls from the head; each time split brain job execution ends, the deque is cleared.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastJoiner.java
```diff
@@ -148,7 +141,7 @@ public void onMessage(Object msg) {
                 logger.warning(e);
             }
         } finally {
-            node.multicastService.removeMulticastListener(listener);
+            splitBrainJoinMessages.clear();
```
tombujok: I'm just thinking what if there's more brains than two, for example five? 
With this logic their join requests will be discarded. 
That is the reason I was thinking that we maybe need some message timestamp and expiration duration?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/SplitBrainMulticastListener.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public SplitBrainMulticastListener(Node node, BlockingDeque<SplitBrainJoinMessage> deque) {
+        this.node = node;
+        this.deque = deque;
+    }
+
+    public void onMessage(Object msg) {
+        if (msg != null && msg instanceof SplitBrainJoinMessage) {
+            SplitBrainJoinMessage joinRequest = (SplitBrainJoinMessage) msg;
+            if (node.getThisAddress() != null && !node.getThisAddress().equals(joinRequest.getAddress())) {
```
mdogan: `node.getThisAddress() != null` is redundant, `thisAddress` cannot be null.

mdogan: `msg != null` is also redundant. `msg instanceof SplitBrainJoinMessage` already implies that.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastJoiner.java
mdogan: As I understand, having `splitBrainJoinMessages.clear()` at the end of each cycle, join messages will be alive at most one cycle of split discovery.

mdogan: I cannot think any problem with processing each message. There are already checks/validations for merge process. This is only discovery part.

vbekiaris: I think expiration is not required. If an old split brain join message is processed, then either the clusters should still merge (win), or if some other changes have occurred in the meanwhile (eg already merged so sending address is not still the master of other cluster, cluster version changed to incompatible version on one cluster) then there will be no merge.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/SplitBrainMulticastListener.java
vbekiaris: Thanks, fixed

vbekiaris:
Updated to process all split brain join messages in the stack on each cyle of the split brain handler job.

vbekiaris:
Unrelated test failure in `SecondsBasedEntryTaskSchedulerTest.test_doNotRescheduleEntryWithinSameSecond_postpone` ( #8728 ).

vbekiaris:
run-lab-run

vbekiaris:
Thanks @mdogan & @tombujok for your feedback

PR #9331, opened 2016-11-28
Distributed Classloading

jerrinot:
Allow to configure a member to load locally unavailable classes
from other cluster members. This can massively simplifies Hazelcast
deployment.

It's released as Beta - it's a feature small in LoC, but big
in impossible impact. Beta feateure means both API and behaviour
can change without a notice.

Clients cannot provide classes in the first version, Lite Members
can be a workaround for some use-cases.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/Node.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/Node.java
@@ -23,6 +23,7 @@ import com.hazelcast.cluster.impl.TcpIpJoiner;
 import com.hazelcast.cluster.memberselector.MemberSelectors;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.DiscoveryConfig;
+import com.hazelcast.config.DistributedClassloadingConfig;
 import com.hazelcast.config.JoinConfig;
 import com.hazelcast.config.ListenerConfig;
 import com.hazelcast.config.MemberAttributeConfig;
@@ -40,6 +41,7 @@ import com.hazelcast.internal.cluster.impl.DiscoveryJoiner;
 import com.hazelcast.internal.cluster.impl.JoinRequest;
 import com.hazelcast.internal.cluster.impl.MulticastJoiner;
 import com.hazelcast.internal.cluster.impl.MulticastService;
+import com.hazelcast.internal.distributedclassloading.DistributedClassLoader;
 import com.hazelcast.internal.cluster.impl.SplitBrainJoinMessage;
 import com.hazelcast.internal.management.ManagementCenterService;
 import com.hazelcast.internal.partition.InternalPartitionService;
@@ -74,6 +76,7 @@ import com.hazelcast.version.MemberVersion;
 
 import java.lang.reflect.Constructor;
 import java.nio.channels.ServerSocketChannel;
+import java.security.PrivilegedAction;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
@@ -89,6 +92,7 @@ import static com.hazelcast.spi.properties.GroupProperty.LOGGING_TYPE;
 import static com.hazelcast.spi.properties.GroupProperty.MAX_JOIN_SECONDS;
 import static com.hazelcast.spi.properties.GroupProperty.SHUTDOWNHOOK_ENABLED;
 import static com.hazelcast.spi.properties.GroupProperty.SHUTDOWNHOOK_POLICY;
+import static java.security.AccessController.doPrivileged;
 
 @SuppressWarnings({"checkstyle:methodcount", "checkstyle:visibilitymodifier", "checkstyle:classdataabstractioncoupling",
         "checkstyle:classfanoutcomplexity"})
@@ -155,12 +159,12 @@ public class Node {
 
     private volatile Address masterAddress;
 
-    @SuppressWarnings("checkstyle:executablestatementcount")
+    @SuppressWarnings({"checkstyle:executablestatementcount", "checkstyle:methodlength"})
     public Node(HazelcastInstanceImpl hazelcastInstance, Config config, NodeContext nodeContext) {
         this.hazelcastInstance = hazelcastInstance;
         this.config = config;
         this.liteMember = config.isLiteMember();
-        this.configClassLoader = config.getClassLoader();
+        this.configClassLoader = getConfigClassloader(config);
         this.properties = new HazelcastProperties(config);
 
         String policy = properties.getString(SHUTDOWNHOOK_POLICY);
@@ -212,6 +216,24 @@ public class Node {
             }
             throw ExceptionUtil.rethrow(e);
         }
+    }
+
+    private ClassLoader getConfigClassloader(Config config) {
+        DistributedClassloadingConfig distributedClassloadingConfig = config.getDistributedClassloadingConfig();
+        ClassLoader classLoader;
+        if (distributedClassloadingConfig.isEnabled()) {
+            ClassLoader parent = config.getClassLoader();
+            final ClassLoader theParent = parent == null ? Node.class.getClassLoader() : parent;
+            classLoader = doPrivileged(new PrivilegedAction<DistributedClassLoader>() {
+                @Override
+                public DistributedClassLoader run() {
+                    return new DistributedClassLoader(theParent);
+                }
+            });
+        } else {
+            classLoader = config.getClassLoader();
+        }
+        return classLoader;
     }
 
     public HazelcastThreadGroup getHazelcastThreadGroup() {
--- a/hazelcast/src/main/java/com/hazelcast/nio/IOUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/IOUtil.java
@@ -468,6 +468,21 @@ public final class IOUtil {
         }
     }
 
+    public static byte[] toByteArray(InputStream is) throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        drainTo(is, baos);
+        return baos.toByteArray();
+    }
+
+    public static void drainTo(InputStream input, OutputStream output) throws IOException {
+        byte[] buffer = new byte[1024];
+        int n;
+        while (-1 != (n = input.read(buffer))) {
+            output.write(buffer, 0, n);
+        }
+    }
+
+
     private static final class ClassLoaderAwareObjectInputStream extends ObjectInputStream {
 
         private final ClassLoader classLoader;
--- a/hazelcast/src/main/java/com/hazelcast/config/Config.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/Config.java
@@ -128,6 +128,8 @@ public class Config {
 
     private HotRestartPersistenceConfig hotRestartPersistenceConfig = new HotRestartPersistenceConfig();
 
+    private DistributedClassloadingConfig distributedClassloadingConfig = new DistributedClassloadingConfig();
+
     private String licenseKey;
 
     private boolean liteMember;
@@ -1259,6 +1261,28 @@ public class Config {
      */
     public Config setLiteMember(boolean liteMember) {
         this.liteMember = liteMember;
+        return this;
+    }
+
+    /**
+     * Get current configuration of distributed classloading.
+     *
+     * @since 3.8
+     *
+     */
+    public DistributedClassloadingConfig getDistributedClassloadingConfig() {
+        return distributedClassloadingConfig;
+    }
+
+    /**
+     * Set distributed classloading configuration
+     *
+     * @param distributedClassloadingConfig
+     * @since 3.8
+     * @return distributed classloading configuration
+     */
+    public Config setDistributedClassloadingConfig(DistributedClassloadingConfig distributedClassloadingConfig) {
+        this.distributedClassloadingConfig = distributedClassloadingConfig;
         return this;
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/config/DistributedClassloadingConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/DistributedClassloadingConfig.java
@@ -1,0 +1,264 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.config;
+
+import com.hazelcast.spi.annotation.Beta;
+
+/**
+ * Configuration of distributed classloading. When enabled it allows members to load classes from other members.
+ * This simplifies deployment as you do not have to deploy your domain classes into classpath of all cluster members.
+ *
+ */
+
+@Beta
+public class DistributedClassloadingConfig {
+
+    /**
+     * Controls caching of caches loaded from remote members
+     */
+    public enum ClassCacheMode {
+        /**
+         * Never caches loaded classes. This is suitable for loading runnables, callables, entry processors, etc.
+         *
+         */
+        OFF,
+
+        /**
+         * Cache indefinitely. This is suitable when you load long-living objects, such as domain objects stored
+         * in a map.
+         *
+         */
+        ETERNAL,
+    }
+
+    /**
+     * Controls how to react on receiving a classloading request from a remote member
+     *
+     */
+    public enum ProviderMode {
+        /**
+         * Never serves classes to other members. This member will never server classes to remote members.
+          */
+        OFF,
+
+        /**
+         * Serve classes from local classpath only. Classes loaded from other members will be used locally,
+         * but they won't be served to other members.
+         *
+         */
+        LOCAL_CLASSES_ONLY,
+
+        /**
+         * Serve classes loaded from both local classpath and from other members.
+         *
+         */
+        LOCAL_AND_CACHED_CLASSES
+    }
+
+    private ClassCacheMode classCacheMode = ClassCacheMode.ETERNAL;
+    private ProviderMode providerMode = ProviderMode.LOCAL_AND_CACHED_CLASSES;
+    private String blacklistedPrefixes;
+    private String whitelistedPrefixes;
+    private String providerFilter;
+    private boolean enabled;
+
+    /**
+     * Enable or disable distributed classloading. Default: {{@code false}}
+     *
+     * @param enabled
+     * @return
+     */
+    public DistributedClassloadingConfig setEnabled(boolean enabled) {
+        this.enabled = enabled;
+        return this;
+    }
+
+    /**
+     * Returns true when distributed classloading is enabled
+     *
+     * @return {{@code true}} when distributed classloading is enabled
+     */
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    /**
+     * Filter to constraint members to be used for classloading request when a class
+     * is not available locally.
+     *
+     * Filter format: <code>HAS_ATTRIBUTE:foo</code> this will send classloading requests
+     * only to members which has a member attribute <code>foo</code> set. Value is ignored,
+     * it can be any type. A present of the attribute is sufficient.
+     *
+     * This facility allows to have a control on classloading. You can e.g. start Hazelcast lite
+     * members dedicated for class-serving.
+     *
+     * Example usage:
+     * This member will load classes only from members with the <code>class-provider</code> attribute set.
+     * It won't ask any other member to provide a locally unavailable class:
+     * <code>
+     *     Config hazelcastConfig = new Config();
+     *
+     *     DistributedClassloadingConfig distributedClassloadingConfig = hazelcastConfig.getDistributedClassloadingConfig();
+     *     distributedClassloadingConfig.setProviderFilter(HAS_ATTRIBUTE:class-provider);
+     *
+     *     HazecastInstance instance = Hazelcast.newHazelcastInstance(hazelcastConfig);
+     * </code>
+     *
+     *
+     * This member is marked with the <code>class-provider</code> attribute - the member configured above may him
+     * it to provide a class which is not locally available:
+     * <code>
+     *     Config hazelcastConfig = new Config();
+     *
+     *     MemberAttributeConfig memberAttributes = hazelcastConfig.getMemberAttributeConfig();
+     *     memberAttributes.setAttribute("class-provider", "true");
+     *
+     *     HazecastInstance instance = Hazelcast.newHazelcastInstance(hazelcastConfig);
+     * </code>
+     *
+     *
+     * Setting the filter to null will allow to load classes from all members.
+     *
+     * Default: {{@code null}}
+     *
+     *
+     * @see {@link com.hazelcast.core.Member#setStringAttribute(String, String)}
+     * @param providerFilter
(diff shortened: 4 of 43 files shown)


hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassLoader.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    protected Class<?> loadClass(String name, boolean resolve)
+            throws ClassNotFoundException {
+        Class<?> clazz = classloadingService.findLoadedClass(name);
+        if (clazz == null) {
+            try {
+                return super.loadClass(name, resolve);
+            } catch (ClassNotFoundException e) {
+                if (classloadingService == null) {
```
mdogan: If `classloadingService` can be null, then `Class<?> clazz = classloadingService.findLoadedClass(name)` will throw NPE.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassLoader.java
```diff
@@ -0,0 +1,57 @@
...
+    private static final ILogger LOG = Logger.getLogger(DynamicClassLoader.class);
+
+    private ClassloadingService classloadingService;
+
+    public DynamicClassLoader(ClassLoader parent) {
+        super(parent);
+    }
+
+    public void setClassloadingService(ClassloadingService service) {
+        this.classloadingService = service;
```
mdogan: Is `classloadingService` set before every thread that may call `loadClass()` starts? If not, isn't it possible `classloadingService` to be observed as null in `loadClass`?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/ClassData.java
```diff
@@ -0,0 +1,67 @@
...
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+import java.io.IOException;
+
+/**
+ * Bytecode of class
+ *
+ */
+public class ClassData implements IdentifiedDataSerializable {
```
mdogan: why a wrapper around `byte[]` is needed?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassLoader.java
jerrinot: good one. the `classloadingService.findLoadedClass(name);` was added later. I'll check this.

Donnerbart:
run-lab-run

jerrinot:
run-lab-run

hazelcast/src/main/java/com/hazelcast/config/DynamicClassloadingConfig.java
```diff
@@ -0,0 +1,235 @@
...
+ * Configuration of dynamic classloading. When enabled it allows members to load classes from other members.
+ * This simplifies deployment as you do not have to deploy your domain classes into classpath of all cluster members.
+ *
+ */
+public class DynamicClassloadingConfig {
+
+    /**
+     * Controls caching of caches loaded from remote members
+     */
+    public enum ClassCacheMode {
```
jerrinot: I am not sure whether this to keep inner or move to the top-level.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/ClassLocator.java
```diff
@@ -0,0 +1,202 @@
...
+        if (clazz != null) {
+            return clazz;
+        }
+        return tryToGetClassFromRemote(name);
+    }
+
+    private Class<?> tryToGetClassFromRemote(String name) throws ClassNotFoundException {
+        ContextMutexFactory.Mutex classMutex = mutexFactory.mutexFor(name);
+        try {
+            synchronized (classMutex) {
```
tkountis: I am guessing that this is needed because you are not accessing it through the partition threads, can you document that here just to help the next one touching this code

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/filter/AndFilter.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package com.hazelcast.internal.dynamicclassloading.impl.filter;
+
+/**
+ * Filter matching only when both sub-filters are matching
+ *
+ * @param <T>
+ */
+public final class AndFilter<T> implements Filter<T> {
```
tkountis: Wondering whether we can have these under util to allow other futures to re-use.

hazelcast/src/main/java/com/hazelcast/config/DynamicClassloadingConfig.java
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Comma separated list of prefixes of classes which will never be loaded remotely.
+     * A prefix can be a package name or a classname.
+     *
+     * For example setting a blacklist prefix to <code>com.foo</code> will disable remote loading of all classes
+     * from the <code>com.foo</code> package, but also classes from all sub-packages won't be loaded.
+     * Eg. <code>com.foo.bar.MyClass</code> will be black-listed too.
+     *
+     * When you set the blacklist to <code>com.foo.Class</code> then the Class will obviously be black-listed,
```
jerrinot: You can't. The `.class` prefix is an internal representation of classes on a filesystem. 
However here you are specifiying logical classname prefixes. 

I understand your point, perhaps in the future we could add "strict-match-mode" instead prefix-match.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/ClassLocator.java
jerrinot: I added explanation + Java6 aware mutexes

hazelcast/src/main/java/com/hazelcast/config/DynamicClassloadingConfig.java
```diff
@@ -0,0 +1,235 @@
...
+
+        /**
+         * Serve classes from local classpath only. Classes loaded from other members will be used locally,
+         * but they won't be served to other members.
+         *
+         */
+        LOCAL_CLASSES_ONLY,
+
+        /**
+         * Server classes loaded from both local classpath and from other members.
```
jerrinot: thanks

hazelcast/src/main/java/com/hazelcast/config/DynamicClassloadingConfig.java
```diff
@@ -0,0 +1,239 @@
...
+     */
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    /**
+     * Filter to constraint members to be used for classloading request when a class
+     * is not available locally.
+     *
+     * Filter format: <code>HAS_ATTRIBUTE:foo</code> this will send classloading requests
```
mdogan: Can you also add a more clear example to use this filter? like;

```java
MemberAttributeConfig memberAttributes = config.getMemberAttributeConfig();
memberAttributes.setAttribute("class-provider", "true");

DynamicClassloadingConfig classloadingConfig = config.getDynamicClassloadingConfig();
classloadingConfig.setProviderFilter("HAS_ATTRIBUTE:class-provider");
[...]
```

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -210,6 +214,24 @@ public Node(HazelcastInstanceImpl hazelcastInstance, Config config, NodeContext
         }
     }
 
+    private ClassLoader getConfigClassloader(Config config) {
+        DynamicClassloadingConfig dynamicClassloadingConfig = config.getDynamicClassloadingConfig();
+        ClassLoader classLoader;
+        if (dynamicClassloadingConfig.isEnabled()) {
+            ClassLoader parent = config.getClassLoader();
+            final ClassLoader theParent = parent == null ? Node.class.getClassLoader() : parent;
+            classLoader = doPrivileged(new PrivilegedAction<DynamicClassLoader>() {
```
mdogan: What is the exact reason of using `doPrivileged()`? We don't have access control checks anywhere in code. Why is it required here? You can also explain above the method.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassLoader.java
```diff
@@ -0,0 +1,66 @@
...
+            // -> we can use it directly without consulting the parent classloader
+            //when the class is not found then we will consult the parent classloader and eventually the classloading
+            //service
+            clazz = dynamicClassloadingService.findLoadedClass(name);
+        }
+        if (clazz == null) {
+            try {
+                return super.loadClass(name, resolve);
+            } catch (ClassNotFoundException e) {
+                if (dynamicClassloadingService == null) {
```
mdogan: How do you guarantee that `dynamicClassloadingService` set by `setDynamicClassloadingService()` method will be eventually visible? 
What are the possible threads that will try to load classes via `DynamicClassLoader`? Can a user thread (non Hazelcast thread) use `DynamicClassLoader `?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import static com.hazelcast.internal.dynamicclassloading.impl.filter.ClassNameFilterParser.parseClassNameFilters;
+import static com.hazelcast.internal.dynamicclassloading.impl.filter.MemberProviderFilterParser.parseMemberFilter;
+
+public final class DynamicClassloadingService implements ManagedService {
+    public static final String SERVICE_NAME = "cl-service";
```
mdogan: Can you give a clear name?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
```diff
@@ -0,0 +1,101 @@
...
+    private boolean enabled;
+
+    private ClassDataProvider provider;
+    private ClassLocator locator;
+
+    @Override
+    public void init(NodeEngine nodeEngine, Properties properties) {
+        DynamicClassloadingConfig config = nodeEngine.getConfig().getDynamicClassloadingConfig();
+        if (!config.isEnabled()) {
+            //TODO: Fail-fast when some attributes are configured but the service is disabled
```
mdogan: Not sure about this comment. Why fail fast? What if one disables the dynamic classloading intentionally? to test, to debug etc.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
```diff
@@ -0,0 +1,101 @@
...
+            throws ClassNotFoundException {
+        if (!enabled) {
+            throw new ClassNotFoundException("Dynamic Classloading is not enabled.");
+        }
+        return locator.handleClassNotFoundException(name);
+    }
+
+    //called by distributed classloader on this member
+    public Class<?> findLoadedClass(String name) {
+        if (!enabled) {
```
mdogan: Is `enabled` guaranteed to be visible to all threads those are able to load classes?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
```diff
@@ -0,0 +1,101 @@
...
+            return null;
+        }
+        return locator.findLoadedClass(name);
+    }
+
+    @Override
+    public void reset() {
```
mdogan: Shouldn't cached data be cleared here?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/ClassLocator.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private Class<?> tryToGetClassFromRemote(String name) throws ClassNotFoundException {
+        //we need to acquire a classloading lock before defining a class
+        //Java 7+ can use locks with per-class granularity while Java 6
+        //has to use a single lock.
+        //mutexFactory abstract these differences away
+        Closeable classMutex = mutexFactory.getMutexForClass(name);
+        try {
+            synchronized (classMutex) {
```
mdogan: Is this guaranteed to be deadlock free for java 7+?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/ClassSource.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public Class<?> define() {
+        clazz = defineClass(name, bytecode, 0, bytecode.length);
+        return clazz;
+    }
+
+    @Override
+    protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+        if (name.equals(this.name)) {
+            return clazz;
```
mdogan: `ClassSource` will be accessed only after retrieving it from `classSourceMap`, right? Is it shared in any way other than `classSourceMap`?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/filter/Filter.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.hazelcast.internal.dynamicclassloading.impl.filter;
+
+/**
+ * Decide whether an object matches a condition or not.
+ *
+ * @param <T>
+ */
+public interface Filter<T> {
```
mdogan: Looks like a good candidate for a util package.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/operation/ClassDataFinderOperation.java
```diff
@@ -0,0 +1,87 @@
...
+        this.className = className;
+    }
+
+    public ClassDataFinderOperation() {
+    }
+
+    @Override
+    public void beforeRun() throws Exception {
+        NodeEngine nodeEngine = getNodeEngine();
+        logger = nodeEngine.getLogger(ClassDataFinderOperation.class);
```
mdogan: it seems `logger` is not used.

hazelcast/src/test/java/com/hazelcast/internal/dynamicclassloading/impl/filter/MemberProviderFilterParserTest.java
```diff
@@ -0,0 +1,64 @@
...
+import com.hazelcast.test.HazelcastTestSupport;
+import org.junit.Test;
+
+import java.util.Map;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class MemberProviderFilterParserTest extends HazelcastTestSupport {
```
mdogan: no `@Category`.

hazelcast/src/test/java/other/packge/MyEP.java
```diff
@@ -0,0 +1,23 @@
+package other.packge;
```
mdogan: I prefer a clear package name, like `dynamicclassloading.blabla...`

hazelcast/src/test/java/other/packge/MyEP.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Map;
+
+/**
+ * This test class is intentionally in its own package
+ * as Hazelcast as a special rules for loading classes
+ * from com.hazelcast.* package.
+ *
+ */
+public class MyEP extends AbstractEntryProcessor<Integer, Integer> {
```
mdogan: Do you really need a logic inside this `EntryProcessor`? If yes, can you give a more suitable name?

hazelcast/src/test/java/com/hazelcast/internal/dynamicclassloading/impl/ClassDataProviderTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import static com.hazelcast.config.DynamicClassloadingConfig.ProviderMode.LOCAL_AND_CACHED_CLASSES;
+import static com.hazelcast.config.DynamicClassloadingConfig.ProviderMode.LOCAL_CLASSES_ONLY;
+import static com.hazelcast.config.DynamicClassloadingConfig.ProviderMode.OFF;
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.mock;
+
+public class ClassDataProviderTest {
```
mdogan: No `@Category`.

hazelcast/src/main/java/com/hazelcast/nio/IOUtil.java
```diff
@@ -468,6 +468,21 @@ private static void copyDirectory(File source, File target) {
         }
     }
 
+    public static byte[] toArray(InputStream is) throws IOException {
```
mdogan: `toByteArray`?

hazelcast/src/main/java/com/hazelcast/nio/IOUtil.java
```diff
@@ -468,6 +468,21 @@ private static void copyDirectory(File source, File target) {
...
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        copy(is, baos);
+        return baos.toByteArray();
+    }
+
+    public static void copy(InputStream input, OutputStream output) throws IOException {
```
mdogan: I'm not sure `copy` is a good name. what about `drainTo` or `transferTo`?

hazelcast/src/main/java/com/hazelcast/config/DynamicClassloadingConfig.java
jerrinot: more filters are definitely an option. I was even considering a custom filter impl. `CUSTOM:com.bar.FooFilter`, but not for the 1st version. 

I'll add a better example, that's a very good hint.

hazelcast/src/main/java/com/hazelcast/config/DynamicClassloadingConfig.java
```diff
@@ -0,0 +1,239 @@
...
+     *
+     * @see #setBlacklistedPrefixes(String)
+     * @return currently configured blacklist prefixes
+     */
+    public String getBlacklistedPrefixes() {
+        return blacklistedPrefixes;
+    }
+
+    /**
+     * Command separated list of prefixes of classes which will be loaded remotely.
```
jerrinot: thanks!

hazelcast/src/main/java/com/hazelcast/instance/Node.java
jerrinot: well, findbugs was complaining. I could either add a suppression or do this. 
what do you see as a better option?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
jerrinot: the TODO is to introduce fail-fast when you e.g. configure blacklisting (=an intend to use dynamic classloading), but the service itself is disabled.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
jerrinot: good one. it should be `volatile`

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
jerrinot: I am not sure. The smaller brain might have some unique classes?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/filter/Filter.java
jerrinot: @tkountis already suggested this. so `com.hazelcast.internal.util.Filter` ?

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/DynamicClassloadingService.java
mdogan: But what if I disable the service intentionally, for example during testing, bug hunting? Will you force me to remove whole classloading config?

hazelcast/src/main/java/com/hazelcast/nio/IOUtil.java
jerrinot: `drainTo` sounds good.

hazelcast/src/test/java/com/hazelcast/internal/dynamicclassloading/impl/ClassDataProviderTest.java
jerrinot: thanks!

hazelcast/src/test/java/other/packge/MyEP.java
jerrinot: yeah, I want to check it was actually executed and it did the job. I will rename it though.

hazelcast/src/test/java/other/packge/MyEP.java
jerrinot: ✓

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/operation/ClassDataFinderOperation.java
jerrinot: good one. the operation used to do more but then I moved its logic outside.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/ClassSource.java
jerrinot: it's shared via the classSourceMap + it can be also put into the `ThreadLocalClassCache`.

hazelcast/src/main/java/com/hazelcast/internal/dynamicclassloading/impl/ClassLocator.java
jerrinot: It's a mutex per classname (in Java 7+)
I guess you could come-up with a scenario where dead-lock can occur, but can't think of any realistic one. It's all circle dependencies where depended classes are spread over multiple members.

hazelcast/src/main/java/com/hazelcast/instance/Node.java
jerrinot: yes

hazelcast/src/main/java/com/hazelcast/config/DynamicClassloadingConfig.java
jerrinot: I'm still thinking about it. Having it an internal class is somewhat clumsy/versbose. It's not too different from e.g. https://github.com/jerrinot/hazelcast/blob/eef987b9920f3f7b4c506e433b0d067cab9ba93c/hazelcast/src/main/java/com/hazelcast/config/EvictionPolicy.java#L21-L21 @mdogan WDYT?

mrkk:
is there a web project example code that uses this feature?
two war files that I can deploy? 1 acting as client and another as cache server?

PR #9336, opened 2016-11-28
Fixed NearCacheConfig

Donnerbart:
* fixed JavaDoc
* deprecated constructors with deprecated fields as parameters
* added missing field to serialization
* increased code coverage

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/NearCacheConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/NearCacheConfig.java
@@ -16,10 +16,10 @@
 
 package com.hazelcast.config;
 
-import com.hazelcast.nio.serialization.impl.BinaryInterface;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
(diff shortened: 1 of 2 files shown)


pveentjer:
:+1:

Donnerbart:
run-lab-run

jerrinot:
if a serialized blob with this class is not send to a client (or from a client) then you should remove the `@BinaryInterface` annotation. If it is sent between a member and a client then you cannot add fields.

However keeping the `@BinaryInterface` and adding new fields does not compute :)
cc @vbekiaris

Donnerbart:
I didn't add the annotation for the config classes. But I think those are only used between members and MC to see the configuration of a member. And I also think that the annotation is needed for the automatic test @vbekiaris wrote to check the uniqueness of the IDs etc.

Donnerbart:
run-lab-run

vbekiaris:
`NearCacheConfig` looks like it's not used in client-member communication, so we should remove the `@BinaryInterface` annotation and turn into `IdentifiedDataSerializable`. @Donnerbart I'll do that in a separate PR.

PR #9346, opened 2016-11-29
Adds get-cluster-version and change-cluster-version commands in cluster.sh

vbekiaris:


tombujok:
It doesn't need to be green since it's a shell script change.

PR #9349, opened 2016-12-01
Add "on task done" callback to InvocationFuture

mtopolnik:
After implementing `cancel` on Invocation's future, the future's completion status no longer matches the completion of the underlying computational task. As soon as the future is canceled, the user loses interest in the result of the computation, but internally nothing changes for us until the task actually stops executing and releases its resources. We now need a separate signal and an associated callback that runs when the task is done.

This PR introduces a small and low-impact addition to `Invocation`, but is crucial for Jet.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
@@ -241,6 +241,13 @@ public abstract class InvocationBuilder {
         return executionCallback;
     }
 
+    /**
+     * Sets a callback that will respond to the "task done" event for the invocation this builder is about to create.
+     * It occurs upon the release of computational and other resources used by the task underlying the invocation.
+     * The user loses interest in the computation as soon as the invocation's future is canceled, but our
+     * internal concern is keeping track of resource usage. Therefore we need a lifecycle event independent
+     * of the regular future completion/cancelation.
+     */
     public InvocationBuilder setDoneCallback(Runnable doneCallback) {
         this.doneCallback = doneCallback;
         return this;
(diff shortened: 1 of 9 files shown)


mdogan:
Are you going to add a test that ensures `doneCallback` is executed even when future is cancelled?

mtopolnik:
Indeed, just writing it :)

hazelcast/src/main/java/com/hazelcast/spi/InvocationBuilder.java
```diff
@@ -240,5 +241,10 @@ protected ExecutionCallback getTargetExecutionCallback() {
         return executionCallback;
     }
 
+    public InvocationBuilder setDoneCallback(Runnable doneCallback) {
```
mtopolnik: I already had that description on Invocation Future, but then it was lost in a refactoring step. It is not re-introduced on this method, plus a reference to it from `Invocation.taskDoneCallback`.

mtopolnik:
```java
/**
 * Sets a callback that will respond to the "task done" event for the invocation this builder is about to create.
 * It occurs upon the release of computational and other resources used by the task underlying the invocation.
 * The user loses interest in the computation as soon as the invocation's future is canceled, but our
 * internal concern is keeping track of resource usage. Therefore we need a lifecycle event independent
 * of the regular future completion/cancelation.
 */
```

hazelcast/src/test/java/com/hazelcast/spi/impl/operationservice/impl/Invocation_TaskDoneTest.java
```diff
@@ -0,0 +1,112 @@
...
+            @Override
+            public void onResponse(Object response) { }
+        });
+
+        // When
+        fut.cancel(true);
+
+        // Then
+        assertTrue(canceledLatch.await(1, SECONDS));
+        assertFalse(cb.done);
```
mdogan: Isn't this fragile? What if `doneCallback` is executed before this assertion?

PR #9352, opened 2016-12-01
Introduced map#removeAll with predicate

ahmetmircik:
Adds `map#removeAll(predicate)` method. It can be used to remove all entries which match with the supplied predicate. If a map has index, matching entries will be found via index search, otherwise they will be found by full-scan.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/EntryRemovingProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/EntryRemovingProcessor.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.map.impl;
+
+import com.hazelcast.map.AbstractEntryProcessor;
+
+import java.util.Map;
+
+public class EntryRemovingProcessor extends AbstractEntryProcessor {
+
+    public static final EntryRemovingProcessor ENTRY_REMOVING_PROCESSOR = new EntryRemovingProcessor();
+
+    public EntryRemovingProcessor() {
+    }
+
+    public Object process(Map.Entry entry) {
+        ((LazyMapEntry) entry).remove();
+        return null;
+    }
+}
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
@@ -262,6 +262,12 @@ public class NearCachedMapProxyImpl<K, V> extends MapProxyImpl<K, V> {
     }
 
     @Override
+    protected void removeAllInternal(Predicate predicate) {
+        super.removeAllInternal(predicate);
+        nearCache.clear();
+    }
+
+    @Override
     protected void deleteInternal(Data key) {
         super.deleteInternal(key);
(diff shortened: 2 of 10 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/EntryRemovingProcessor.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+
+package com.hazelcast.map.impl;
+
+import com.hazelcast.map.AbstractEntryProcessor;
+
+import java.util.Map;
+
+public class EntryRemovingProcessor extends AbstractEntryProcessor<Integer, Integer> {
```
ahmetmircik: seems a left over. fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
```diff
@@ -262,6 +262,12 @@ protected Data removeInternal(Data key) {
     }
 
     @Override
+    protected void removeAllInternal(Predicate predicate) {
+        super.removeAllInternal(predicate);
+        nearCache.clear();
```
ahmetmircik: i did also think on that pain point and will be happy to hear better alternatives.

mmedenjak: Maybe a silly suggestion but what about this (of course this needs some refactoring): 

```
    @Override
    protected void removeAllInternal(Predicate predicate) {
        final AbstractEntryProcessor evictingProcessor = new AbstractEntryProcessor() {
            @Override
            public Object process(Entry entry) {
                nearCache.remove(toData(entry.getKey(), partitionStrategy));
                ((LazyMapEntry) entry).remove();
                return null;
            }
        };
        
        OperationFactory operation = operationProvider.createPartitionWideEntryWithPredicateOperationFactory(name,
                evictingProcessor, predicate);
        try {
            operationService.invokeOnAllPartitions(SERVICE_NAME, operation);
        } catch (Throwable t) {
            throw rethrow(t);
        }
    }
```

jerrinot: can we evaluate the predicate locally (on entries in the near-cache)? is it too silly? :)

ahmetmircik: sync or async? it is an O(n) operation

jerrinot: You cannot return from the `remote(predicate)` call until the relevant entries are removed from near-cache. Is this true? 

Cannot the entry processor return keys it processed?

ahmetmircik: > when keysCount < X then EP returns the keys else remove everything ?

What is a nice X?

ahmetmircik: yes it can be no problem with it, hd near-cache has a different imp then hd imap

vbekiaris: Here are the options we discussed with @ahmetmircik :
(1) execute near cache full scan asynchronously to locate entries matching the predicate and remove them. Pros: accurate, does not throw away valid items from your near cache, Cons: may cause stale reads until the async full scan is done - so it’s different from `remove`‘s behavior
(2) return the removed keys from the `EntryProcessor` and only remove these from the near cache. Pros: as above, Cons: additional serialization/network costs, places burden on opposite use case in which user’s predicate matches a large number of keys
(3) make it “adaptive”: the `removeAll` operations return the number of keys removed; on the local member, if number of entries removed > (some threshold percentage) then just clear all otherwise do the full scan & check against the predicate. Making this threshold overridable by a property would allow users to tune it according to their use case (a small threshold to avoid full near cache scan and quickly turn to full `clear` at the cost of emptying near cache or a large threshold that would increase latency of `removeAll` and keep their near cache warm).

Of the options above, I am leaning more towards (3) but it may pose some implementation challenges.

ahmetmircik:
@vbekiaris will do after merge, thanks

ahmetmircik:
@vbekiaris javaDoc update is here: https://github.com/hazelcast/hazelcast/commit/aff62af35f3278caabe040be68b67c1f77a51438

PR #9356, opened 2016-12-02
Removing a code piece that is not needed. #9351

sancar:
This change is related to to only mocked connections. The part added
claims to letting client connection to be closed when server
is closed. But the MockConnectionManager already closes all
connections when shutting down.

backport of #9351


sancar:
Merging this, since https://github.com/hazelcast/hazelcast/pull/9351 forward port is reviewed and merged

PR #9359, opened 2016-12-02
Add configurable shutdown hook policy

### LINKED ISSUE #4648: Hazelcast Shutdown Hook calls terminate() instead of shutdown()
The default behaviour of a Hazelcast instance is to register a JVM shutdown hook that calls terminate()  on that instance.

Calling terminate() causes Hazelcast to forcibly shutdown the instance without ensuring that the data on the node being shutdown has finished syncing to the remaining nodes in the cluster.

I would argue that the default behaviour should be to call shutdown() which cleanly shuts down the Hazelcast instance and waits until all data has finished syncing to other nodes.

I am more than happy to submit a patch for this issue.

jerrinot:
Fixes #4648


jerrinot:
@mdogan, @vbekiaris: thanks for a quick review!

PR #9365, opened 2016-12-05
destroyConnection is called inside log.isFinestEnabled check

sancar:
This bug first introduced in 3.7.1
see https://github.com/hazelcast/hazelcast/pull/8832/files#diff-6b832face5ab9f5e23687eaeb478627eR596

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -604,10 +604,11 @@ public class ClientConnectionManagerImpl implements ClientConnectionManager {
             }
             fireConnectionAddedEvent(connection);
         } else {
+            //This part is hit when client sends authenticateion request again to an already
+            //connected server to upgrade itself to owner connection. In this case old new connection is same
+            assert connection.equals(oldConnection);
             if (logger.isFinestEnabled()) {
-                logger.finest("Authentication succeeded for " + connection + " and replaced " + oldConnection
-                    + " Closing the old connection.");
-                destroyConnection(oldConnection, "Replaced by newer connection", null);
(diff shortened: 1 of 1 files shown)


sancar:
Well isn't this interesting ?  @jerrinot same test also fails in my local.  One more thing to investigate.

jerrinot:
@sancar are you telling me the PR worked only thanks to the bug?

sancar:
@jerrinot yes it seems so.

sancar:
@jerrinot I understand the mistake we are doing. The place that we are trying to add `destroyConnection(oldConnection, "Replaced by newer connection", null);` is accessed only when there is a connection that member. And it re-authenticates to upgrade itself to owner connection. Both old and new connection is same in this case. That is why there was no `destroyConnection` there before. And we should not introduce it now. You accidentally make the right thing by adding it inside `isFinestEnabled` check. Bad news is from 3.7.1 to 3.7.5, if someone runs it with finest logs are opened, the client will fail. Since running on production with debug logs are opened not likely, I think this is not a huge problem. I will update this pr accordingly.

sancar:
@jerrinot 
I had a assert there to make sure that old and new connection is same when old is not null. 
https://github.com/hazelcast/hazelcast/pull/8832/files#diff-6b832face5ab9f5e23687eaeb478627eR599
Interestingly that assert is removed later by @ihsandemir here
https://github.com/hazelcast/hazelcast/commit/f27eda4583d62a1f6a8d22ce4342bb3289b1e49a#diff-6b832face5ab9f5e23687eaeb478627eL601

I wonder if this rule is still applicable or not. @ihsandemir do you remember why did we delete it ?

ihsandemir:
@sancar The rule should still be applicable. No specific reason for removing the assert (except that we will not handle the problem if it happens at production). It can be put back. 

There is already an info level log printed, why not append the log details (which is in the finest message) into this existing info log so that we can get all the information even when finest log is not open.

PR #9370, opened 2016-12-05
Add scheduled executor and cardinality estimator permission policies

mmedenjak:
Add cardinality estimator and scheduled executor to client permission policies


mmedenjak:
@jerrinot @tkountis I've noticed inconsistencies for durable executor config as well. Now hazelcast.xml, hazelcast-spring.xml, xml generator should all have :
- all necessary client permissions (for durable and scheduled executor, cardinality estimator)
- durable executor support
- scheduled executor support
Please check the PR again if possible.

PR #9371, opened 2016-12-05
Jet flow-control packet + Packet.flags cleanup

mtopolnik:
* redefine the three packet type flags (bits 0, 2, 5) as an encoded integer discriminating up to 7 types. Define the new `JET` type as type 3.
* add new `Packet.FLAG_JET_FLOW_CONTROL` flag
* rename flag-related methods to make more sense
* fix Javadoc to reflect the actual behavior of flag methods
* use the `char` type instead of `short` to store the flags bitfield (without this, using the bit 15 of flags would break `isFlagSet()` due to sign extension)
* rename `FLAG_RESPONSE` to `FLAG_OP_RESPONSE`
* improve documentation of flags and group them according to how they are acted upon by the receiver

Additional cleanup:
* clean up logic in `createInputOutputFactory()`
* delete identical overriding method in `StandardMemoryAccessor`

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/Packet.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/Packet.java
@@ -87,7 +87,8 @@ public final class Packet extends HeapData implements OutboundFrame {
 
     private static final int HEADER_SIZE = BYTE_SIZE_IN_BYTES + SHORT_SIZE_IN_BYTES + INT_SIZE_IN_BYTES + INT_SIZE_IN_BYTES;
 
-    private short flags;
+    // char is a 16-bit unsigned integer. Here we use it as a bitfield.
+    private char flags;
     private int partitionId;
     private transient Connection conn;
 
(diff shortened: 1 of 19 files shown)


hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -38,26 +38,56 @@
...
-    /**
-     * A flag to indicate this is a special control packet for the operation system like invocation-heartbeats.
-     */
+
+    // 3. Type-specific flags. Same bits can be reused within each type
+
+    // 3.a Operation packet flags
+
+    /** Marks an Operation packet as Response */
+    public static final int FLAG_OP_RESPONSE = 1 << 1;
```
jerrinot: does it have to be an operation? right now operation is the only type which can generate a response, but there is nothing preventing you to introduce a new packet type yielding a response.

mtopolnik: My idea is for each type to redefine the meaning of packet type-specific flags (bits 1 and 6). Each type that needs a "response" bit can add a constant `FLAG_<X>_RESPONSE` with the same value.

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -38,33 +39,61 @@
...
+
+
+    //            END OF HEADER FLAG SECTION
+
 
 
     private static final int HEADER_SIZE = BYTE_SIZE_IN_BYTES + SHORT_SIZE_IN_BYTES + INT_SIZE_IN_BYTES + INT_SIZE_IN_BYTES;
 
-    private short flags;
+    private char flags;
```
jerrinot: this deserves a comment

mtopolnik: You mean... `short` is obviously a bitfield, but `char` needs to be explained?

jerrinot: yes. it's an unusual choice. it's not immediately clear why `char` (my guess: to not bother with signs)

mtopolnik: `// char is a 16-bit unsigned integer. Here we use it as a bitfield.`

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -38,33 +39,61 @@
...
 
-    /**
-     * Flag to indicate this will be a Jet packet
-     */
-    public static final int FLAG_JET = 1 << 7;
+
+    // 3.b Jet packet flags
+
+    /** Marks a Jet packet as Flow control */
+    public static final int FLAG_JET_FLOW_CONTROL = 1 << 1;
```
jerrinot: so this is a JET-packet-specific flag, but there is no JET packet definition. I guess it would deserve an explanation too.

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
jerrinot: good enough.

hazelcast/src/main/java/com/hazelcast/nio/Packet.java
```diff
@@ -331,13 +357,108 @@ public int hashCode() {
...
+         * Represents "missing packet type", consists of all zeros. A zeroed-out packet header would
+         * resolve to this type.
+         * <p>
+         * {@code ordinal = 0}
+         */
+        NULL,
+        /**
+         * The type of an Operation packet.
+         * <p>
+         * {@code ordinal = 0}
```
vbekiaris: ordinal here is 1

PR #9375, opened 2016-12-06
Expose hot restart service on Cluster service

mmedenjak:
Expose Hot restart service via API on Cluster service. This needs some cleanup in a separate PR since there is an existing `HotRestartService` in EE.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientClusterProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientClusterProxy.java
@@ -21,6 +21,7 @@ import com.hazelcast.cluster.ClusterState;
 import com.hazelcast.core.Cluster;
 import com.hazelcast.core.Member;
 import com.hazelcast.core.MembershipListener;
+import com.hazelcast.hotrestart.HotRestartService;
 import com.hazelcast.transaction.TransactionOptions;
 import com.hazelcast.version.ClusterVersion;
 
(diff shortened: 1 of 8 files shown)


jerrinot:
is there any other way to expose it? this is a very specific functionality to have a privileged place inside `HazelcastInstance` @mdogan: WDYT?

mmedenjak:
@jerrinot @mdogan It does feel weird to have it in HazelcastInstance. I'm open to suggestions.

metanet:
what about `com.hazelcast.core.Cluster` ?

jerrinot:
@metanet: That's a good idea. It triggers a cluster-wide snapshot and there is already `shutdown()` for cluster-wide shutdown in the `Cluster` interface. It feels somewhat similar. 

Do we have to expose everything? What about adding just the `backup(prefix)` (or `snapshot(prefix)` ?) method directly to the `Cluster` interface.

mmedenjak:
@jerrinot @metanet I can expose the service or four methods - backup(), backup(seq), interrupt() and getBackupTaskStatus(). Seems like it's better if it is contained in a service. WDYT?

mmedenjak:
Migrated the exposed service to Cluster service

mmedenjak:
@jerrinot @metanet I have changed the name of the exposed service. The refactoring for the methods and `NodeExtension` will be in an another PR. Is the change ok, can we merge this?

metanet:
@mmedenjak  good to go

Issue #9373, opened 2016-12-05
GetAll not updating localMapStats (hz 3.7.3)


fredang:
Hi,

I did a small test where I call Get/GetAll/Put/PutAll multiple times on a distributed map and then output the local map stats.
The Get/Put/PutAll are showing properly in the local map stats.
However this is not the case for GetAll (seems to be excluded from the stats).

To reproduce the issue:
```
        HazelcastInstance instance = Hazelcast.newHazelcastInstance();

        IMap<String, Integer> map = instance.getMap("test");
        for(int i = 0 ; i < 100 ; i++) {
            Map<String, Integer> putMap = new HashMap<>();
            String key1 = "key1-" + i;
            String key2 = "key2-" + i;
            putMap.put("key1-" + i, i);
            putMap.put("key2-" + i , i);
            map.putAll(putMap);


            Set<String> keys = new HashSet<>();
            keys.add(key1);
            keys.add(key2);
            map.getAll(keys);
        }
        System.out.println(map.getLocalMapStats());

```

It outputs the following stats
```
LocalMapStatsImpl{lastAccessTime=1480968873116, lastUpdateTime=1480968873116, hits=200, 
numberOfOtherOperations=0, numberOfEvents=0, getCount=0, putCount=200, removeCount=0,
totalGetLatencies=0, totalPutLatencies=120, totalRemoveLatencies=0, ownedEntryCount=200,
backupEntryCount=0, backupCount=1, ownedEntryMemoryCost=31780, backupEntryMemoryCost=0,
creationTime=1480968872891, lockedEntryCount=0, dirtyEntryCount=0, heapCost=31780}
```

jerrinot:
fixed by #9387

jerrinot:
:clap: for @fredang !

PR #9387, opened 2016-12-06
Small fix for #9373 (GetAll not updating localMapStats)

fredang:
Made GetAll calls update the Get Stats
Added 2 tests to check that GetAll and PutAll are updating the stats

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -672,6 +672,7 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
         if (keys.isEmpty()) {
             return;
         }
+        long time = System.currentTimeMillis();
         Collection<Integer> partitions = getPartitionsForKeys(keys);
         Map<Integer, Object> responses;
         try {
@@ -684,6 +685,7 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
                     resultingKeyValuePairs.add(toObject(entries.getValue(i)));
                 }
             }
+            localMapStats.incrementGets(keys.size(), System.currentTimeMillis() - time);
         } catch (Exception e) {
             throw rethrow(e);
         }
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -672,6 +672,7 @@ protected void getAllObjectInternal(List<Data> keys, List<Object> resultingKeyVa
         if (keys.isEmpty()) {
             return;
         }
+        long time = System.currentTimeMillis();
```
gaganis: @fredang Good job, mate! Started to look into this myself, but you beat me to it! 

The only thing I did differently was that I have inserted the statement recording the start time after the creating the operationFactory. This makes it more consistent to how the respective existing code works for the put operation. 
eg

```
            OperationFactory operationFactory = operationProvider.createGetAllOperationFactory(name, keys);
            final long time = System.currentTimeMillis();
            responses = operationService.invokeOnPartitions(SERVICE_NAME, operationFactory, partitions);
```

It could be debated that your position better because it also takes into account the factory creation. Going down this road would mean that to keep the two behaviours consistent it would also need to be changed for putAll. But this would mean two changes instead of one. 

Best,
Giorgos

jerrinot:
@fredang hi, many thanks for the fix, it's very much appreciated!
The fix looks very good to me. 

Can you please sign the [Hazelcast Contributor Agreement](https://hazelcast.atlassian.net/wiki/display/COM/Hazelcast+Contributor+Agreement). This is a prerequisite before accepting a contribution.

fredang:
Thanks @gaganis and @jerrinot for the code review. I have updated the code accordingly to not include the factory creation time.

I have just signed and emailed the signed contributor agreement.

jerrinot:
@fredang: many thanks for this contribution! 

image

jerrinot:
@fredang fancy creating a PR into the [maintenance branch](https://github.com/hazelcast/hazelcast/tree/maintenance-3.x) ?

fredang:
Thanks @ahmetmircik and @jerrinot.
Sorry @jerrinot, will be more careful next time when creating the PR.

PR #9388, opened 2016-12-06
client protocol version upgraded to 1.3.3

asimarslan:
client protocol is updated to support old .net clients that have a bug prevent connecting to client 3.7.3+


asimarslan:
run-lab-run

PR #9390, opened 2016-12-07
Put eventExecutor.shutdown back to client

sancar:
It seems it is removed when it should not.
To be able to find affected versiopns easily later,  here are the commits deletes this
It is deleted in 3.8 here
https://github.com/hazelcast/hazelcast/pull/8782/files#diff-4785c8d5812cf0e37314bb1f280640c9L92
It is deleted in 3.7.3 here
https://github.com/hazelcast/hazelcast/pull/9097/files#diff-4785c8d5812cf0e37314bb1f280640c9L92


jerrinot:
great finding! this explains the hanging Hazelcast instances @Donnerbart reported?

cangencer:
@jerrinot Yes I noticed this when I was testing Jet in the lab - that the client wouldn't shut down properly.

PR #9394, opened 2016-12-07
Clear local records before cache replication data is applied to local

metanet:
Before applying replication data, local cache data must be cleared so that there is no left-over stale data on backup.


jerrinot:
good one! interestingly enough it's already done in HD: https://github.com/hazelcast/hazelcast-enterprise/commit/97faedc2154c3eb9b3f651ca0e376d12ecbf0f95#diff-5fa4c9fd934af8f17977d31030f0d226R102

metanet:
Yes, it is missing only in heap cache. All other variations are just fine.

metanet:
unrelated test failure

metanet:
run-lab-run

PR #9399, opened 2016-12-08
Listeners may not be registered correctly if ReAuthOperation fails for some member

ihsandemir:
During authentication, AuthenticationMessageTask sends ReAuthOperations to all members and gets responses from them using a timeout value of 10 seconds. Some of these invocations may fail due to:

1. member is shutting down and the cluster member list update is still not received: We try to get an invocation response from a shutting down member.
2. member somehow fails to respond within 10 seconds (may be a network problem as well).

In these cases, we need to handle clients for which the ReAuthOperation actually succeded, they already changed their client ownership map. There may be two options here:

1. Reset back the original state of these affected members.
2. When a second ReAuthOperation comes to this member, it does run the operation but when returning whether client had any previous resources (such as listeners) or not also by checking the listener map as well.

This fix does the second option. So, if at a client no cleanup is run at all or it is run but the previous AuthenticationMessageTask failed (i.e. there is an entry at the client ownership map) and a new AuthenticationMessageTask is executed, then also check the existence resources of this client.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/operations/ClientReAuthOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/operations/ClientReAuthOperation.java
@@ -33,7 +33,7 @@ public class ClientReAuthOperation
         implements UrgentSystemOperation, AllowedDuringPassiveState {
 
     private String clientUuid;
-    private boolean clientDisconnectOperationRun;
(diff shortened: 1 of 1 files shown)


ihsandemir:
This PR probably fixed this issue: https://github.com/hazelcast/hazelcast/pull/9475 Hence, closing the PR without any merge.

PR #8543, opened 2016-07-18
Fix for issue #8261

ihsandemir:
Fixes #8261 . Makes sure that the CleanupResourcesTask only sets the future value after the connection is closed.

PR #9401, opened 2016-12-08
Cleanup for changes made in  #8543

sancar:
In CleanResourceTask invocation.notifyException called twice.
, removing one of them. Could not add tests for this one
since AbstractInvocationFuture.complete can be called multiple
times. Calling twice not causing bug, but it was wrong.

Two static methods related to client is removed from
TargetDisconnectedException since it is also used in node and
methods specific to client should not be there

Detailed heartbeart related logging inside
newTargetDisconnectedExceptionCausedByHeartbeat is removed,
since these are already available in ClientConnection.toString.

ClientConnection.isClosed is removed since it nearly means !isAlive.
One caveat related to removing is race between closedTime and isAlive
boolean. To remove this race I also removed isAlive boolean.
isAlive means closedTime == 0 from now on.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
@@ -49,7 +49,6 @@ import java.util.concurrent.TimeUnit;
 
 import static com.hazelcast.client.spi.properties.ClientProperty.MAX_CONCURRENT_INVOCATIONS;
 import static com.hazelcast.instance.OutOfMemoryErrorDispatcher.onOutOfMemory;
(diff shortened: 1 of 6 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationServiceSupport.java
```diff
@@ -236,38 +223,24 @@ public void run() {
...
-                        connection.getCloseCause());
-                invocation.notifyException(ex);
+                notifyException(invocation, connection);
             }
             if (expiredConnections != null) {
                 logExpiredConnections(expiredConnections);
             }
         }
 
-        private void notifyException(ClientInvocation invocation, ClientConnection connection, boolean isAlive) {
```
ihsandemir: what happens if isAlive changes from true to false while executing this method? I think it is better we pass the parameter in.

PR #9407, opened 2016-12-08
Map.aggregate not available when in-memory format is NATIVE

vbekiaris:
Addresses comment https://github.com/hazelcast/hazelcast-enterprise/pull/1184#discussion_r91517975


vbekiaris:
Thanks for the quick review @ahmetmircik & @tombujok

PR #9408, opened 2016-12-09
Heartbeat callback should run on internal executor

sancar:
Small fix changes made in https://github.com/hazelcast/hazelcast/pull/8581

Default executor is reserved for user invocations callbacks
and user listeners. Everything else we used internally should
run on internal executor and should be non-blocking.


sancar:
verify

hasancelik:
run-lab-run

PR #9409, opened 2016-12-09
Heartbeat callback should run on internal executor

sancar:
Small fix change made in https://github.com/hazelcast/hazelcast/pull/8581

Default executor is reserver for user invocations callbacks
and user listeners. Everything else we used internally should
run on internal executor and should be non-blocking.

backport of https://github.com/hazelcast/hazelcast/pull/9408


sancar:
verify

sancar:
run-lab-run

PR #9415, opened 2016-12-09
Fixes nested queries on VersionedPortables with different versions.	

### LINKED ISSUE #8846: Can't do nested queries on VersionedPortables with different versions. Hazelcast throws IllegalArgumentException: Not a registered Portable field
Given an object that looks like this:

``` java
public class TestObjectPortable implements VersionedPortable {
    private TestObjectMetadataPortable objectMetadataPortable;

    public TestObjectPortable(TestObjectMetadataPortable objectMetadata) {
        this.objectMetadataPortable = objectMetadata;
    }

    public int getClassVersion() {
        return 3;
    }
    ...
```

and the metadata object: 

``` java
public class TestObjectMetadataPortable implements VersionedPortable {
    String metaData;
    String reference;

    public TestObjectMetadataPortable(String metaData, String reference) {
        this.metaData = metaData;
        this.reference = reference;
    }

    public int getClassVersion() { return 1; }
    ...
```

If I try to do the following query: 

``` java
map.values(Predicates.equal("objectMetadataPortable.reference", "ref2"));
```

Hazelcast would throw an IllegalArgumentException: Not a registered Portable field.

However if I change the class version of TestObjectMetadataPortable to 3 as well everything works as expected.

Project to proof the issue: 
[HazelCastBug.zip](https://github.com/hazelcast/hazelcast/files/455251/HazelCastBug.zip)

tombujok:
Fixes #8846

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/FieldDefinitionImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/FieldDefinitionImpl.java
@@ -26,18 +26,19 @@ public class FieldDefinitionImpl implements FieldDefinition {
     FieldType type;
     int classId;
     int factoryId;
+    int version;
 
-    public FieldDefinitionImpl(int index, String fieldName, FieldType type) {
-        this(index, fieldName, type, 0, 0);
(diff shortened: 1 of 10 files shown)


tombujok:
run-lab-run

PR #9416, opened 2016-12-09
Fix error in DiagnosticsLogWriter month 1 off

pveentjer:
The month is one off because Date.MONTH is zero based.


hazelcast/src/main/java/com/hazelcast/internal/diagnostics/DiagnosticsLogWriter.java
```diff
@@ -120,7 +120,7 @@ protected void appendDateTime() {
     private void appendDate() {
         write(calendar.get(DAY_OF_MONTH));
         write('-');
-        write(calendar.get(MONTH));
+        write(calendar.get(MONTH + 1));
```
mtopolnik: This adds one to the `MONTH` constant and will get the `WEEK_OF_YEAR` field instead.

Correct code is `write(calendar.get(MONTH) + 1);`

pveentjer: doh

PR #9427, opened 2016-12-12
Log ignored member info update operation

metanet:
Additionally, do not update the partition table if MemberInfoUpdateOperation decides to ignore the member list update.
Another minor change: Reduce periodic member list publish interval in migration tests. It can happen that a joining node can miss a published member list if there are concurrent joins. Since periodic member list publish is 5 mins by default, tests can timeout before that node receives the up-to-date member list.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/MemberInfoUpdateOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/MemberInfoUpdateOperation.java
@@ -68,9 +68,10 @@ public class MemberInfoUpdateOperation extends AbstractClusterOperation implemen
             }
 
             return;
-        } else if (!node.getThisUuid().equals(targetUuid)) {
-            String msg = "targetUuid: " + targetUuid + " is different than this node's uuid: " + node.getThisUuid();
-            throw new IllegalStateException(msg);
+        }
+
+        if (!checkValid()) {
+            return;
         }
 
         processMemberUpdate();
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/MemberInfoUpdateOperation.java
```diff
@@ -96,13 +98,24 @@ final void processPartitionState() {
...
+
+        if (isLocal) {
+            return true;
+        }
+
+        Address endpoint = conn.getEndPoint();
+        Address masterAddress = clusterService.getMasterAddress();
+        boolean valid = (endpoint != null && endpoint.equals(masterAddress));
+        if (!valid) {
+            getLogger().warning("Ignoring operation because sender: " + endpoint + " is not known master: " + masterAddress);
```
jerrinot: do we need a warning here? it can happen what a master disconnected while having in-flight operations = business as usual?

metanet: I can set it to `debug` level

jerrinot: +1

I saw users wondering [about this](https://github.com/jerrinot/hazelcast/blob/dee6be7311a9beeaf135470846872c24f71008bc/hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java#L621-L621)

This PR seems me to be similar.

PR #9440, opened 2016-12-13
Inject Hazelcast instance into quorum function

### LINKED ISSUE #7499: Support HazelcastInstanceAware for QuorumFunction

jerrinot:
Fixes #7499

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/quorum/impl/QuorumImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/quorum/impl/QuorumImpl.java
@@ -17,6 +17,8 @@
 package com.hazelcast.quorum.impl;
 
 import com.hazelcast.config.QuorumConfig;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.HazelcastInstanceAware;
 import com.hazelcast.core.Member;
 import com.hazelcast.nio.ClassLoaderUtil;
 import com.hazelcast.quorum.Quorum;
@@ -59,7 +61,7 @@ public class QuorumImpl implements Quorum {
         this.config = config;
         this.quorumName = config.getName();
         this.size = config.getSize();
-        initializeQuorumFunction();
+        initializeQuorumFunction(nodeEngine.getHazelcastInstance());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


hazelcast/src/test/java/com/hazelcast/quorum/QuorumTest.java
```diff
@@ -227,6 +229,49 @@ public boolean apply(Collection<Member> members) {
         }
     }
 
+    @Test
+    public void givenQuorumFunctionConfigured_whenImplementsHazelcastInstanceAware_thenHazelcastInjectsItsInstance() throws Exception {
+        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(1);
```
Donnerbart: A HZ instance factory in a local variable looks suspicious to me. Are those instances shutdown properly?

hazelcast/src/test/java/com/hazelcast/quorum/QuorumTest.java
```diff
@@ -227,6 +229,49 @@ public boolean apply(Collection<Member> members) {
...
+            this.instance = instance;
+        }
+
+        @Override
+        public boolean apply(Collection<Member> members) {
+            return false;
+        }
+    }
+
+
```
Donnerbart: If you have to touch this class, this superfluous newline could also be removed :)

hazelcast/src/test/java/com/hazelcast/quorum/QuorumTest.java
```diff
@@ -227,6 +229,49 @@ public boolean apply(Collection<Member> members) {
         }
     }
 
+    @Test
+    public void givenQuorumFunctionConfigured_whenImplementsHazelcastInstanceAware_thenHazelcastInjectsItsInstance() throws Exception {
```
Donnerbart: I can't see it on GitHub, but if the tests don't need it you could remove the `throws Exception` to shorten the lines a bit (the test names are already quite long).

hazelcast/src/main/java/com/hazelcast/quorum/impl/QuorumImpl.java
```diff
@@ -200,6 +202,9 @@ private void initializeQuorumFunction() {
                 throw ExceptionUtil.rethrow(e);
             }
         }
+        if (quorumFunction instanceof HazelcastInstanceAware) {
```
ahmetmircik: maybe doing this check in the end of the method makes it less fragile

Issue #8825, opened 2016-09-01
Native memory config is not included in the result while generating an XML string from config


alparslanavci:
While generating the xml string from a config object in `ConfigXmlGenerator.generate()` method, `native-memory` configuration is not appended to the result string. This causes that `native-memory` config for members does not seen in management center.

PR #9441, opened 2016-12-13
added nativeMemoryConfig xml generator

bilalyasar:
fixes https://github.com/hazelcast/hazelcast/issues/8825

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
@@ -134,6 +134,8 @@ public class ConfigXmlGenerator {
 
         liteMemberXmlGenerator(xml, config);
 
+        nativeMemoryXmlGenerator(xml, config);
+
         hotRestartXmlGenerator(xml, config);
 
         xml.append("</hazelcast>");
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
```diff
@@ -825,6 +827,29 @@ private void hotRestartXmlGenerator(StringBuilder xml, Config config) {
...
+                .append("\"")
+                .append(">");
+        xml.append("<size")
+                .append(" unit=\"").append(nativeMemoryConfig.getSize().getUnit()).append("\"")
+                .append(" value=\"").append(nativeMemoryConfig.getSize().getValue()).append("\"")
+                .append("/>");
+
+        appendNode(xml, "min-block-size", nativeMemoryConfig.getMinBlockSize());
+        appendNode(xml, "page-size", nativeMemoryConfig.getPageSize());
+        appendNode(xml, "metadata-space-percentage", nativeMemoryConfig.getMetadataSpacePercentage());
```
mdogan: same as `Float.toString()`.

jerrinot: ok. so it won't use the notation with mantissa & exponent. 
I wasn't sure about it, thanks!

PR #9444, opened 2016-12-13
client.getDistributedObjects considers removed proxies too.

### LINKED ISSUE #9423: HazelcastClient.getDistributedObjects is not updated after distributed object removal from other instance
HazelcastClient.getDistributedObjects() returns the proxies already created. But the proxy list is not updated after a distributed object removal from other instance. Below is the reproducer test for the issue: 

```    
    @Test
    public void testGetDistributedObjectsAfterRemove(){
        HazelcastInstance server = hazelcastFactory.newHazelcastInstance();
        IMap<Object, Object> firstMap = server.getMap("firstMap");
        server.getMap("secondMap");

        HazelcastInstance client = hazelcastFactory.newHazelcastClient();
        assertEquals(2, client.getDistributedObjects().size());

        firstMap.destroy();

        assertEquals(1, client.getDistributedObjects().size());
    }
```

sancar:
Client does not listen distributed objects unless user specifically
adds a listener. When a proxy removed from a remote client/server,
client has no idea about it. When getDistributedObjects called,
now it compares the local proxies and removes the ones that removed
in the remote.

fixes #9423

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
@@ -144,8 +144,10 @@ import com.hazelcast.util.ExceptionUtil;
 import com.hazelcast.util.ServiceLoader;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Future;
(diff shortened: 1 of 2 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
```diff
@@ -580,10 +582,21 @@ public IScheduledExecutorService getScheduledExecutorService(String name) {
...
+            }
+
+            Collection<DistributedObjectInfo> newDistributedObjectInfo = resultParameters.response;
+            for (DistributedObjectInfo distributedObjectInfo : newDistributedObjectInfo) {
+                localDistributedObjects.remove(distributedObjectInfo);
                 getDistributedObject(distributedObjectInfo.getServiceName(), distributedObjectInfo.getName());
             }
+
+            for (DistributedObjectInfo distributedObjectInfo : localDistributedObjects) {
+                proxyManager.removeProxy(distributedObjectInfo.getServiceName(), distributedObjectInfo.getName());
```
ihsandemir: It would be nice to have a log about this implicit removal.

hazelcast-client/src/test/java/com/hazelcast/client/proxy/DistributedObjectListenerTest.java
```diff
@@ -79,4 +82,19 @@ public void run() throws Exception {
...
+    @Test
+    public void testGetDistributedObjectsAfterRemove() {
+        HazelcastInstance server = hazelcastFactory.newHazelcastInstance();
+        IMap firstMap = server.getMap("firstMap");
+        server.getMap("secondMap");
+
+        HazelcastInstance client = hazelcastFactory.newHazelcastClient();
+        assertEquals(2, client.getDistributedObjects().size());
+
+        firstMap.destroy();
```
ihsandemir: we can assume this has the same affect as a second client destroying the proxy but since the reported error was generated by using the clients I would recommend adding that test here as well. i.e. just get a second client verify it has 2 objects initially, then let other client destroy a proxy and verify that this client sees this removal as expected.

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
sancar: We also have implicit creation 4 lines above of this. Do you think we should log it also or not ?
Do you want the log for us or  for the user ?
For example, user is able to listen create and remove events via listener.

hazelcast-client/src/test/java/com/hazelcast/client/proxy/DistributedObjectListenerTest.java
sancar: I am adding one more test.

sancar:
run-lab-run

PR #9446, opened 2016-12-13
Initialize HashMap/HashSet sizes. Allow reads to be done without defensive copies.

bokken:
Size HashSet and HashMap (and Linked variants) instances correctly.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
@@ -18,8 +18,6 @@ package com.hazelcast.internal.nearcache.impl.preloader;
 
 import com.hazelcast.config.NearCachePreloaderConfig;
 import com.hazelcast.internal.adapter.DataStructureAdapter;
-import com.hazelcast.internal.nearcache.NearCacheRecord;
-import com.hazelcast.internal.nearcache.impl.NearCacheRecordMap;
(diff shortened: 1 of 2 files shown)


bokken:
I now have a signed CLA submitted.

bokken:
This at least partially addresses https://github.com/hazelcast/hazelcast/issues/1194 by allowing a new map configuration of forceDefensiveCopy which defaults to true. If the in memory format is OBJECT and forceDefensiveCopy is set to false, then get and getAll operations no longer force a defensive copy for a local read, including read from backups.

jerrinot:
hello @bokken,

many thanks for your effort! Can you please split this into 2 separated PRs? The initialization with size seems like a no-brainer to me - it can be merged quickly, even in the current state of 3.8. (we are in a Feature Freeze)

However the 2nd part might require more thinking, at very least there multiple option for end-user configuration: I can imagine a new `IN-MEMORY-FORMAT` option instead of a brand new configuration attribute.

bokken:
@jerrinot, I created a new PR which just has the hash based collection initialization changes:
https://github.com/hazelcast/hazelcast/pull/9582

bokken:
@jerrinot, I had originally considered a new in memory format option, and that certainly could be done. There appeared to be some other existing issues and suggestions that pointed towards this approach. 

It seems there could be some challenges with using the in memory format as that would appear to impact more than just the Map and the concept of defensive copy does not appear to be applicable to all data structures.

bokken:
I have now created a separate PR which only has the changes to avoid defensive copies when getting out of an IMap.

https://github.com/hazelcast/hazelcast/pull/9620

PR #9447, opened 2016-12-14
Fix potential NPE in member operating in 3.6 client compatibility mode

vbekiaris:
When a member operates with the 3.6 client compatibility flag set, `CacheConfig` objects are wrapped in `LegacyCacheConfig` for serialization without taking into account the possibility of `CacheConfig` being `null`. For example, when a 3.6 client `CacheManager` tries to obtain a `Cache` with `getCache` which is not yet configured on the member side, then the member will fail to encode its response to the client with a `NullPointerException`.


vbekiaris:
Unrelated test failure `com.hazelcast.client.scheduledexecutor.ClientScheduledExecutorServiceTest.schedule_whenShutdown`

vbekiaris:
run-lab-run

PR #9456, opened 2016-12-14
Do not require cache writer/loader on client classpath

### LINKED ISSUE #9453: JCache client needs access to CacheLoader implementation to work
Hello.
I am using hazelcast 3.7.4 on a server node, as well as a hazelcast client node.
I have configured a cache using XML on the server node similar to, using read-through and specifying a cache loader factory.

On the client side, I am receiving a `com.hazelcast.nio.serialization.HazelcastSerializationException: java.lang.ClassNotFoundException` for my CacheLoader's implementation class, since it is not available in the client's classpath.

While searching, I've found a similar _closed_ issue: #6676
Looking at the tests, I found [one test](https://github.com/hazelcast/hazelcast/blob/9d2eb04228da9c121161435410d90125a1ab597a/hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientCacheReadWriteThroughTest.java#L79) that seems to test the exact issue I am experiencing, however I believe the test is incorrect and passes while it shouldn't, since it doesn't actually use a Hazelcast _Client_ caching provider.

I gave it a go to write a test for the issue I'm trying to describe. While not _quite_ good a test, I had some feedback that the CacheLoader is required on the client, it fails with an IllegalStateException on Factory's writeExternal method.

    package com.hazelcast.client.cache;

    import com.hazelcast.cache.impl.HazelcastServerCachingProvider;
    import com.hazelcast.client.cache.impl.HazelcastClientCachingProvider;
    import com.hazelcast.client.config.ClientConfig;
    import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
    import com.hazelcast.config.CacheConfig;
    import com.hazelcast.core.HazelcastInstanceAware;
    import com.hazelcast.instance.HazelcastInstanceImpl;
    import com.hazelcast.internal.util.ThreadLocalRandom;
    import org.junit.Test;

    import javax.cache.Cache;
    import javax.cache.configuration.CompleteConfiguration;
    import javax.cache.configuration.Factory;
    import javax.cache.integration.CacheLoader;
    import javax.cache.integration.CacheLoaderException;
    import javax.cache.spi.CachingProvider;
    import java.io.Externalizable;
    import java.io.IOException;
    import java.io.ObjectInput;
    import java.io.ObjectOutput;
    import java.util.HashMap;
    import java.util.Map;
    import java.util.UUID;

    public class ClientCacheReadThroughNoCacheLoader {
        @Test
        public void clientDoesNotNeedCacheLoader() throws Exception {
            final String serverName = UUID.randomUUID().toString();
            final Config serverConfig = new Config(serverName);
            final int port = ThreadLocalRandom.current().nextInt(5000, 15000);
            serverConfig.getNetworkConfig().setPort(port).setPublicAddress("localhost");

            final HazelcastInstance server = Hazelcast.newHazelcastInstance(serverConfig);
            CachingProvider serverProvider = HazelcastServerCachingProvider.createCachingProvider(server);
            CompleteConfiguration<Integer, String> cacheConfig =
                    new CacheConfig<Integer, String>()
                            .setTypes(Integer.class, String.class)
                            .setReadThrough(true)
                            .setCacheLoaderFactory(new ServerSideCacheLoaderFactory());

            final String cacheName = UUID.randomUUID().toString();
            serverProvider.getCacheManager().createCache(cacheName, cacheConfig);

            System.out.println("Creating client now");

            final String clientName = UUID.randomUUID().toString();
            final ClientConfig clientConfig = new ClientConfig();
            clientConfig.setInstanceName(clientName);
            clientConfig.getNetworkConfig().addAddress("localhost:"+port);

            final HazelcastInstance client = HazelcastClient.newHazelcastClient(clientConfig);
            CachingProvider clientCachingProvider = HazelcastClientCachingProvider.createCachingProvider(client);
            Cache<Integer, String> clientCache = clientCachingProvider.getCacheManager()
                    .getCache(cacheName, Integer.class, String.class);
        }

        public static class ServerSideCacheLoaderFactory
                implements Factory<ServerSideCacheLoader>, HazelcastInstanceAware, Externalizable {

            private transient HazelcastInstance hazelcastInstance;
            public ServerSideCacheLoaderFactory() {
                System.out.println("initialized with instance: " + hazelcastInstance);
            }

            @Override
            public void setHazelcastInstance(HazelcastInstance hazelcastInstance) {
                this.hazelcastInstance = hazelcastInstance;
            }

            @Override
            public ServerSideCacheLoader create() {
                System.out.println("create called");
                if (hazelcastInstance instanceof HazelcastInstanceImpl) {
                    return new ServerSideCacheLoader();
                } else {
                    throw new IllegalStateException("This factory can only be used at server side!");
                }
            }

            @Override
            public void writeExternal(ObjectOutput out) throws IOException {
                if(hazelcastInstance != null) {
                    if(hazelcastInstance instanceof HazelcastClientInstanceImpl) {
                        throw new IllegalStateException("This factory can only be used at server side!");
                    }
                }
                System.out.println("write with instance: " + hazelcastInstance);
            }

            @Override
            public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
                if(hazelcastInstance != null) {
                    if(hazelcastInstance instanceof HazelcastClientInstanceImpl) {
                        throw new IllegalStateException("This factory can only be used at server side!");
                    }
                }
                System.out.println("read with instance: " + hazelcastInstance);
            }
        }

        private static class ServerSideCacheLoader implements CacheLoader<Integer, String> {

            static String valueOf(Integer key) {
                return "value-of-" + key;
            }

            @Override
            public String load(Integer key) {
                return valueOf(key);
            }

            @Override
            public Map<Integer, String> loadAll(Iterable<? extends Integer> keys) throws CacheLoaderException {
                Map<Integer, String> result = new HashMap<Integer, String>();
                for (Integer key : keys) {
                    String value = load(key);
                    result.put(key, value);
                }
                return result;
            }

        }
    }

jerrinot:
Fixes #9453

Previously we forced users to provide a factory have it on classpath of clients.
This is unnecessary as clients obviously do not use Writer/Loader.

It also makes declarative configuration simpler to use. Instead of forcing
users to implement a factory it's not possible to directly specify a class to use.

So instead of
<cache-loader-factory class-name="com.foo.MyCacheLoaderFactory" />
it's now possible to use:
<cache-loader class-name="com.package.YourMapLoaderImpl" />

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/config/CacheSimpleConfigTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/config/CacheSimpleConfigTest.java
@@ -1,0 +1,55 @@
+package com.hazelcast.config;
+
+import com.hazelcast.test.HazelcastParallelClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
+import com.hazelcast.test.annotation.ParallelTest;
+import com.hazelcast.test.annotation.QuickTest;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+
+@RunWith(HazelcastParallelClassRunner.class)
+@Category({QuickTest.class, ParallelTest.class})
+public class CacheSimpleConfigTest extends HazelcastTestSupport {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void givenCacheLoaderIsConfigured_whenConfigureCacheLoaderFactory_thenThrowIllegalStateException() {
+        CacheSimpleConfig config = new CacheSimpleConfig();
+        config.setCacheLoader("foo");
+
+        expectedException.expect(IllegalStateException.class);
+        config.setCacheLoaderFactory("bar");
+    }
+
+    @Test
+    public void givenCacheLoaderFactoryIsConfigured_whenConfigureCacheLoader_thenThrowIllegalStateException() {
+        CacheSimpleConfig config = new CacheSimpleConfig();
+        config.setCacheLoaderFactory("bar");
+
+        expectedException.expect(IllegalStateException.class);
+        config.setCacheLoader("foo");
+    }
+
+    @Test
+    public void givenCacheWriterIsConfigured_whenConfigureCacheWriterFactory_thenThrowIllegalStateException() {
+        CacheSimpleConfig config = new CacheSimpleConfig();
+        config.setCacheWriter("foo");
+
+        expectedException.expect(IllegalStateException.class);
+        config.setCacheWriterFactory("bar");
+    }
+
+    @Test
+    public void givenCacheWriterFactoryIsConfigured_whenConfigureCacheWriter_thenThrowIllegalStateException() {
+        CacheSimpleConfig config = new CacheSimpleConfig();
+        config.setCacheWriterFactory("bar");
+
+        expectedException.expect(IllegalStateException.class);
+        config.setCacheWriter("foo");
+    }
+}
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.8.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.8.xsd
@@ -880,10 +880,24 @@
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
+                                <xs:attribute name="cache-loader" type="xs:string" use="optional">
+                                    <xs:annotation>
+                                        <xs:documentation>
+                                            Defines the cache loader class name.
+                                        </xs:documentation>
+                                    </xs:annotation>
+                                </xs:attribute>
                                 <xs:attribute name="cache-writer-factory" type="xs:string" use="optional">
                                     <xs:annotation>
                                         <xs:documentation>
                                             Defines the cache writer factory class name.
+                                        </xs:documentation>
+                                    </xs:annotation>
+                                </xs:attribute>
+                                <xs:attribute name="cache-writer" type="xs:string" use="optional">
+                                    <xs:annotation>
+                                        <xs:documentation>
+                                            Defines the cache writer class name.
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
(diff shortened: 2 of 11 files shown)


jerrinot:
it's a border-line PR when it come to feature the on-going feature freeze. Technically it's a new feature, but the fact we require the writer/loader factories on client can be seen as a bug. there is no good reason for this.

jerrinot:
run-lab-run

hazelcast/src/test/java/com/hazelcast/config/CacheSimpleConfigTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class CacheSimpleConfigTest extends HazelcastTestSupport {
+
+    @Test
+    public void givenCacheLoaderIsConfigured_whenConfigureCacheLoaderFactory_thenThrowIllegalStateException() {
+        CacheSimpleConfig config = new CacheSimpleConfig();
+        config.setCacheLoader("foo");
+
+        try {
+            config.setCacheLoaderFactory("bar");
+            fail("Should throw IllegalStateException");
```
vbekiaris: just a matter of style, you could also use a `@Rule ExpectedException` and rewrite as:
```
expectedException.expect(IllegalStateException.class);
config.setCacheLoaderFactory("bar");
```
IMHO makes the expected outcome more clear

jerrinot:
@vbekiaris: comments addressed.

ahmetmircik:
Can you please separate commits like this: actual fix, spring conf. fix and other ones... now it is hard to see actual fix.

Secondly, I ran the test provided in the issue [here](https://github.com/hazelcast/hazelcast/issues/9453#issue-195496816) but it failed . Should that test be passed? Is my expectation wrong that test should pass?

jerrinot:
@ahmetmircik: this test fails because the factory is not on a client classpath. We cannot change client deserialization due client protocol compatibility.

This PR tries to tackle the issue from the different angle: It does remove the need to define custom factories. The extra configuration `cache-loader` and `cache-writer` uses a standard factories defined in the  JCache API: it's guarantee to be on a client. Hence I don't see a wait to split this - the improvements in configuration parsers are the fix.

ahmetmircik:
thanks for the explanation

PR #9459, opened 2016-12-15
Allow member without JCache in classpath to join cluster when no Cache configurations are defined

### LINKED ISSUE #9301: Do not fail when node without JCache service joins cluster where JCache service is enabled but not used
Use case: a cluster runs with some members in a container where JCache API is available but is not used in Hazelcast so no `Cache`s have been created but JCache service is enabled in those members. A standalone member without JCache API in classpath will not be able to join the cluster because it will have to execute a `PostJoinCacheOperation` however the JCache service will not be available on that node.
Enhancement: instead of failing, when JCache service is not available and no cache configs exist in `PostJoinCacheOperation` log a warning that the node is joining a cluster where JCache is enabled but does not have JCache API in its classpath. If a `Cache` is created at a later point on another member, it will fail anyway.

vbekiaris:
Allows a member without JCache API in its classpath to join a cluster whose members do have the JCache API, when JCache is not used (no `CacheConfig`s are already defined) and log a warning that if JCache API will be used, it will fail.
Also prints an informative `SEVERE` message in case `CacheConfig`s are delivered in the `PostJoinCacheOperation` but JCache API artifact is missing from the classpath (currently, a service not found exception is logged that indicates that `CacheService` is not available but does not provide information towards resolving the issue).

Fixes #9301

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/PostJoinCacheOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/operation/PostJoinCacheOperation.java
@@ -31,7 +31,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import static com.hazelcast.cache.impl.ICacheService.SERVICE_NAME;
-import static com.hazelcast.cache.impl.JCacheDetector.isJcacheAvailable;
+import static com.hazelcast.cache.impl.JCacheDetector.isJCacheAvailable;
 
 public class PostJoinCacheOperation extends Operation implements IdentifiedDataSerializable {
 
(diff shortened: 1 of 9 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/operation/PostJoinCacheOperation.java
```diff
@@ -38,14 +43,31 @@ public void addCacheConfig(CacheConfig cacheConfig) {
...
-        return ICacheService.SERVICE_NAME;
+        return SERVICE_NAME;
     }
 
     @Override
     public void run() throws Exception {
-        ICacheService cacheService = getService();
-        for (CacheConfig cacheConfig : configs) {
-            cacheService.putCacheConfigIfAbsent(cacheConfig);
+        if (isJcacheAvailable(getNodeEngine().getConfigClassLoader())) {
```
ahmetmircik: `isJCacheAvailable` with capital C (it is not introduced here but just wanted to note)

vbekiaris: Thanks @ahmetmircik , I renamed the methods in `JCacheDetector` with a capital C in a separate commit. Once green will squash & merge.

PR #9464, opened 2016-12-15
HttpPostCommandProcessor cleanup

mmedenjak:
The exceptions thrown by the service methods were swallowed up by the command processor. Now logging is added and the exception message is returned by response. 

Also, refactored the methods to remove code duplication.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
@@ -29,6 +29,7 @@ import com.hazelcast.internal.cluster.ClusterService;
 import com.hazelcast.internal.management.ManagementCenterService;
 import com.hazelcast.internal.management.dto.WanReplicationConfigDTO;
 import com.hazelcast.internal.management.operation.AddWanConfigOperation;
+import com.hazelcast.logging.ILogger;
 import com.hazelcast.spi.InternalCompletableFuture;
 import com.hazelcast.spi.OperationService;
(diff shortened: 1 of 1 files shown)


hasancelik:
good job 👍

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
```diff
@@ -128,40 +130,31 @@ private void handleChangeClusterState(HttpPostCommand command) throws Unsupporte
...
                 } else {
-                    res = res.replace("${STATUS}", "fail");
-                    res = res.replace("${STATE}", state.toString().toLowerCase());
+                    res = response(ResponseType.FAIL, "state", state.toString().toLowerCase());
                 }
             }
         } catch (Throwable throwable) {
-            res = res.replace("${STATUS}", "fail");
-            res = res.replace("${STATE}", "null");
+            logger.severe("Error occurred while changing cluster state", throwable);
```
pveentjer: Do we want to have this on severe level? Normally I use severe for some kind of internal system error. In this case exceptions could be thrown because the user did something 'stupid'.

pveentjer:
Looks good; apart from the logging level issue.

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
mmedenjak: Good point. I will also consider if some actions should be logged at a higher level than others.

mmedenjak:
run-lab-run

pveentjer:
:+1:

PR #9465, opened 2016-12-15
EE query fix - CachedQueryEntry is serialized with NATIVE memory

tombujok:


hazelcast/src/main/java/com/hazelcast/map/impl/MapDataSerializerHook.java
```diff
@@ -277,6 +278,7 @@
     public static final int READ_AND_RESET_ACCUMULATOR = 127;
     public static final int SET_READ_CURSOR = 128;
     public static final int ACCUMULATOR_CONSUMER = 129;
+    public static final int CACHED_QUERY_ENTRY = 130;
 
     private static final int LEN = ACCUMULATOR_CONSUMER + 1;
```
vbekiaris: should be `CACHED_QUERY_ENTRY + 1`

tombujok: 👍

vbekiaris:
```
java.lang.AssertionError: Factory with ID -10 instantiated an object of class com.hazelcast.query.impl.CachedQueryEntry while expected type was class com.hazelcast.map.impl.LazyMapEntry
```
Should override IDS method `getId` and create a separate type ID for subclass `LazyMapEntry`?

tombujok:
run-lab-run

pveentjer:
:+1:

PR #9468, opened 2016-12-15
Hot restart services refactoring

mmedenjak:
This is the first step in a small cleanup of hot restart classes and methods. 
- Hot restart related methods were moved from the NodeExtension into the InternalHotRestartService interface. 
- Hot restart service implementations were renamed.
The next step would be to refactor the InternalHotRestartServiceImpl so that separate functionalities are in separate classes.
Once we start adding non-hot-backup related public methods, we can create a new public implementation class and just delegate to the Hot backup service for backup related methods. For now all public methods are hot-backup related hence no need for a separate public service.

EE PR : https://github.com/hazelcast/hazelcast-enterprise/pull/1209

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/hotrestart/InternalHotRestartService.java
+++ b/hazelcast/src/main/java/com/hazelcast/hotrestart/InternalHotRestartService.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.hotrestart;
+
+import com.hazelcast.config.HotRestartPersistenceConfig;
+import com.hazelcast.internal.management.dto.ClusterHotRestartStatusDTO;
+import com.hazelcast.nio.Address;
+
+import java.util.Set;
+
+/**
+ * Internal service for interacting with hot restart related functionalities (e.g. force and partial start)
+ */
+public interface InternalHotRestartService {
+    /**
+     * Forces node to start by skipping hot-restart completely and removing all hot-restart data
+     * even if node is still on validation phase or loading hot-restart data.
+     *
+     * @return true if force start is triggered successfully. force start cannot be triggered if hot restart is disabled or
+     * the master is not known yet
+     */
+    boolean triggerForceStart();
+
+    /**
+     * Triggers partial start if the cluster cannot be started with full recovery and
+     * {@link HotRestartPersistenceConfig#clusterDataRecoveryPolicy} is set accordingly.
+     *
+     * @return true if partial start is triggered.
+     */
+    boolean triggerPartialStart();
+
+    /**
+     * Checks if the given member has been excluded during the cluster start or not.
+     * If returns true, it means that the given member is not allowed to join to the cluster.
+     *
+     * @param memberAddress address of the member to check
+     * @param memberUuid    uuid of the member to check
+     * @return true if the member has been excluded on cluster start.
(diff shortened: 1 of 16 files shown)


hazelcast/src/main/java/com/hazelcast/hotrestart/InternalHotRestartService.java
```diff
@@ -0,0 +1,82 @@
...
+import com.hazelcast.config.HotRestartPersistenceConfig;
+import com.hazelcast.internal.management.dto.ClusterHotRestartStatusDTO;
+import com.hazelcast.nio.Address;
+
+import java.util.Set;
+
+/**
+ * Internal service for interacting with hot restart related functionalities (e.g. force and partial start)
+ */
+public interface InternalHotRestartService {
```
pveentjer: Is this class exposed to the public? If not, would it not make more sense to move it to com.hazelcast.internal.hotrestart? This way the com.hazelcast module remains clean + we get a good separation between public and private

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
```diff
@@ -233,7 +235,8 @@ private void handlePartialStart(HttpPostCommand command) throws UnsupportedEncod
             if (!(groupConfig.getName().equals(groupName) && groupConfig.getPassword().equals(groupPass))) {
                 res = res.replace("${STATUS}", "forbidden");
             } else {
-                boolean success = node.getNodeExtension().triggerPartialStart();
+                final InternalHotRestartService hotRestartService = node.getNodeExtension().getInternalHotRestartService();
+                boolean success = hotRestartService != null && hotRestartService.triggerPartialStart();
```
pveentjer: Would it not be an idea to always have a hotRestartService but in case of not being enabled, that a NullHotRestartService is used? This way you never need to do null checks.

http://refactoring.com/catalog/introduceNullObject.html

pveentjer:
Looks good so far. Keep up the good work!

hazelcast/src/main/java/com/hazelcast/instance/Node.java
```diff
@@ -666,10 +669,13 @@ public SplitBrainJoinMessage createSplitBrainJoinMessage() {
     public JoinRequest createJoinRequest(boolean withCredentials) {
         final Credentials credentials = (withCredentials && securityContext != null)
                 ? securityContext.getCredentialsFactory().newCredentials() : null;
+        final InternalHotRestartService hotRestartService = nodeExtension.getInternalHotRestartService();
+        final Set<String> excludedMemberUuids = hotRestartService != null
```
mdogan: Instead of `null` checking everywhere, can we have a nop implementation of `InternalHotRestartService` in OSS? 
So `nodeExtension.getInternalHotRestartService()` never returns `null`...

hazelcast/src/main/java/com/hazelcast/hotrestart/InternalHotRestartService.java
mdogan: What about `HotRestartServiceIntegrationService`? Since this is not related to implementing Hot Restart functionality, just to interact with actual Hot Restart impl.

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
metanet: I guess we don't have many of these checks before so it was ok. For now, I am ok with both approaches. We can introduce a dummy hot restart service.

hazelcast/src/main/java/com/hazelcast/hotrestart/InternalHotRestartService.java
mmedenjak: That's right. Thus we would end up moving the entire package to `internal`. Perhaps that's too big of a change.

metanet: I think current package structure is fine for now. We can discuss it later on and prepare a separate pr for class moves if we decide that way

mmedenjak:
@pveentjer @mdogan @metanet I've :
- changed the class name `InternalHotRestartService` to `HotRestartIntegrationService`
- changed the class name `InternalHotBackupService` to `HotBackupService`
- added noop implementations of these classes

Most Hot restart functionality is in the `com.hazelcast.spi.hotrestart` package and most of it is private. Although the package name could be changed to internal we will postpone this for an another PR.

I think this addresses most of the comments. If noone has any objections today, I will merge.

Issue #5364, opened 2015-05-24
BasicMapTest.testMapTryPut


bilalyasar:
```
java.lang.AssertionError: expected:<0> but was:<2>
    at org.junit.Assert.fail(Assert.java:88)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-OracleJDK1.6/com.hazelcast$hazelcast/246/testReport/junit/com.hazelcast.map/BasicMapTest/testMapTryPut/

ahmetmircik:
closed via https://github.com/hazelcast/hazelcast/commit/cba5069687cbeacd197393dba096a6d7bb30a0be

hasancelik:
```
java.lang.AssertionError: CountDownLatch failed to complete within 120 seconds , count left: 1
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:812)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:805)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:797)
	at com.hazelcast.map.BasicMapTest.testMapTryPut(BasicMapTest.java:738)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-IbmJDK1.8/com.hazelcast$hazelcast/402/testReport/junit/com.hazelcast.map/BasicMapTest/testMapTryPut/

vbekiaris:
Failed build log seems to indicate that the [`tryPut` invocation](https://github.com/hazelcast/hazelcast/blob/c0dd2786291d0a88f54879097d5b3415b107fd40/hazelcast/src/test/java/com/hazelcast/map/BasicMapTest.java#L718) waited for 5 seconds before being timed out, so for some unknown reason the main test thread did not manage to unlock the key in this period of time (perhaps due to load in the build environment?).

PR #9472, opened 2016-12-15
Improve logging in BasicMapTest

vbekiaris:
Add failure message that indicates in which step the thread that tries to put stopped making progress (for #5364 ). Also, if `tryPut` times out, fail the test immediately instead of waiting for timeout.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/BasicMapTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/BasicMapTest.java
@@ -702,6 +702,7 @@ public class BasicMapTest extends HazelcastTestSupport {
         final AtomicInteger counter = new AtomicInteger(6);
         final CountDownLatch latch = new CountDownLatch(1);
         final CountDownLatch waitTryPutFailure = new CountDownLatch(1);
+        final StringBuilder failureMessageBuilder = new StringBuilder();
         Thread thread = new Thread(new Runnable() {
             public void run() {
                 try {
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/map/BasicMapTest.java
```diff
@@ -742,8 +748,17 @@ public void run() {
         assertOpenEventually(waitTryPutFailure);
 
         map.unlock("key1");
-        assertOpenEventually(latch);
+        try {
+            boolean completed = latch.await(ASSERT_TRUE_EVENTUALLY_TIMEOUT, TimeUnit.SECONDS);
+            assertTrue("tryPut operations were not completed in time, counter was " + counter.get() + " and failure"
```
vbekiaris: Unwrapped `assertOpenEventually` to evaluate `counter.get()` and `failureMessageBuilder.toString()` at time of (potential) failure.

pveentjer: I'm never crazy about doing asserts on string content due to fragility.

pveentjer:
:+1:

hazelcast/src/test/java/com/hazelcast/map/BasicMapTest.java
vbekiaris: This is actually `assertTrue(longFailureMessage, booleanToAssert)`. Put the blame on the new side-by-side github review layout :-)

PR #9473, opened 2016-12-16
Used same immutable event data for all listeners

ahmetmircik:
Used same immutable event data for all listeners

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/BatchInvalidator.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/BatchInvalidator.java
@@ -133,10 +133,11 @@ public class BatchInvalidator extends Invalidator {
     }
 
     private void sendInvalidations(String mapName, List<Invalidation> invalidations) {
+        Invalidation invalidation = new BatchNearCacheInvalidation(mapName, invalidations);
+
         Collection<EventRegistration> registrations = eventService.getRegistrations(serviceName, mapName);
         for (EventRegistration registration : registrations) {
             if (eventFilter.apply(registration)) {
-                Invalidation invalidation = new BatchNearCacheInvalidation(mapName, invalidations);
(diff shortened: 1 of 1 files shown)


ahmetmircik:
Failed test:
`ClientReliableTopicOverloadTest.whenBlock_whenNoSpace`: https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/6303/

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/BatchInvalidator.java
```diff
@@ -133,10 +133,11 @@ private void createAndSendInvalidations(String mapName, InvalidationQueue invali
     }
 
     private void sendInvalidations(String mapName, List<Invalidation> invalidations) {
+        Invalidation invalidation = new BatchNearCacheInvalidation(mapName, invalidations);
```
pveentjer: Good work.

One thing that could be a potential problem (I don't enough about how this is being used), but in the old code, if there are no listener, no BatchNearCacheInvalidation is created, unlike the new code. One way to fix this is to create the BatchNearCacheInvalidation lazily.

ahmetmircik: yes there will always be one instance.

pveentjer: Then I would say: good work and :+1: 

Can you add a line of comment explaining the assumption? For future generations..

ahmetmircik: comment added here: https://github.com/hazelcast/hazelcast/commit/15a3ad94d2603f37d05342941d963f993b4cec2e

PR #9475, opened 2016-12-16
AuthBaseMessageTask should not block generic threads

sancar:
Detailed bug descripton from @metanet
all generic operation threads are occupied by AuthenticationBaseMessageTask﻿⁠⁠⁠⁠
instances. Within an operation thread, it tries to send ﻿⁠⁠⁠⁠ClientReAuthOperation﻿⁠⁠⁠⁠ to all members
and blocks until it receives the responses. These ﻿⁠⁠⁠⁠ClientReAuthOperation﻿⁠⁠⁠⁠s are also going to
be executed on generic operation threads. When there are a lot of concurrent ﻿⁠⁠⁠⁠
AuthenticationBaseMessageTask﻿⁠⁠⁠⁠ instances, they will occupy all of the operation
threads and ﻿⁠⁠⁠⁠ClientReAuthOperation﻿⁠⁠⁠⁠ won’t be executed ==> ﻿⁠⁠⁠⁠AuthenticationBaseMessageTask﻿⁠⁠⁠⁠
instances will block until timeout

Solution:
Responses of ReauthOperatoin are redirected to client via callbacks rather than future.get


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -314,21 +286,67 @@ public Permission getRequiredPermission() {
         return null;
     }
 
-    private class OperationInfo {
-        private final Member member;
-        private final Future future;
+    private final class MultiTargetCallback {
```
mdogan: `MultiTargetCallback` and `SingleTargetCallback` are very similar to ones in `AbstractMultiTargetMessageTask`. Can you extract these classes and reuse?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -85,23 +88,50 @@ private void handleEndpointNotCreatedConnectionNotAlive() {
...
-                return handleUnauthenticated();
+                sendClientMessage(prepareUnauthenticatedClientMessage());
+                break;
             case SERIALIZATION_VERSION_MISMATCH:
-                return handleSerializationVersionMismatch();
+                sendClientMessage(prepareSerializationVersionMismatchClientMessage());
+                break;
             default:
-                throw new IllegalStateException("Unsupported authentication status :" + authenticationStatus);
+                sendClientMessage(new IllegalStateException("Unsupported authentication status :" + authenticationStatus));
```
pveentjer: is 'sendResponse' not more informative?

sancar:
run-lab-run

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
pveentjer: It is pretty confusing that MultiTargetCallback is not a callback implementation at all.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -314,21 +286,67 @@ public Permission getRequiredPermission() {
...
+        public void notify(Member member, Object result) {
+            if (targets.remove(member)) {
+                results.put(member, result);
+            } else {
+                if (results.containsKey(member)) {
+                    logger.warning("Duplicate response from -> " + member);
+                }
+                logger.warning("Unknown target! -> " + member);
+            }
+            if (!targets.isEmpty()) {
```
pveentjer: There is a potential race here. If 2 responses are returned at the same time, it could be that both responses trigger the removal of the item on target, but one of the responses has not put the item yet in the 'results'. So a partially filled 'results' could be send to the client.

sancar: Then we have this problem in other places too similar implementation in `AbstractMultiTargetMessageTask`. I will try to merge this MultiTargetCallback and SingleTargetCallback with the ones in AbstractMultiTargetMessageTask and remove the race so that it is fixed there too.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
sancar: @mdogan I admit that I gave up on idea of reusing the code fast in this case. I will take another look  to see if it is possible.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMultiTargetMessageTask.java
```diff
@@ -37,57 +37,54 @@
...
     protected AbstractMultiTargetMessageTask(ClientMessage clientMessage, Node node, Connection connection) {
         super(clientMessage, node, connection);
     }
 
     @Override
     protected void processMessage() throws Throwable {
         ClientEndpoint endpoint = getEndpoint();
-        OperationFactory operationFactory = createOperationFactory();
-        Collection<Address> targets = getTargets();
+        Supplier<Operation> operationFactory = createOperationFactory();
```
pveentjer: since the OperationFactory has been replaced by supplied, it makes sense to have the local variable 'operationFactory' renamed to 'operationSupplier' or something along those lines.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMultiTargetMessageTask.java
```diff
@@ -37,57 +37,54 @@
...
-        final Collection<Address> targets;
-        final Map<Address, Object> results;
+        final Collection<Member> targets;
+        final Map<Member, Object> results;
 
-        private MultiTargetCallback(Collection<Address> targets) {
-            this.targets = synchronizedSet(new HashSet<Address>(targets));
-            this.results = new ConcurrentHashMap<Address, Object>(targets.size());
+        private MultiTargetCallback(Collection<Member> targets) {
+            this.targets = synchronizedSet(new HashSet<Member>(targets));
```
pveentjer: Do these still need to be sychronized since the notify is synchronized?

pveentjer:
Looks good. Minor comments. Nothing obstructing a 👍

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMultiTargetMessageTask.java
sancar: You are right. I missed that. I am removing the extra synchronization.

PR #9479, opened 2016-12-16
Make use of a portion of max-queue-capacity for each worker in StripedExecutor

ahmetmircik:
Max event queue capacity is determined via system property. Each worker in StripedExecutor should consume only a portion of that capacity. Otherwise we will have `threadCount * system-defined-capacity` instead of `system-defined-capacity`. See [GroupProperty#EVENT_QUEUE_CAPACITY](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java#L105).

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/executor/StripedExecutor.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/executor/StripedExecutor.java
@@ -29,13 +29,15 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
 import static com.hazelcast.util.HashUtil.hashToIndex;
+import static com.hazelcast.util.Preconditions.checkPositive;
+import static java.lang.Math.ceil;
 
 /**
  * The StripedExecutor internally uses a stripe of queues and each queue has its own private worker-thread.
  * When a task is 'executed' on the StripedExecutor, the task is checked if it is a StripedRunnable. If it
  * is, the right worker is looked up and the task put in the queue of that worker. If the task is not a
  * StripedRunnable, a random worker is looked up.
- * <p/>
+ *
  * If the queue is full and the runnable implements TimeoutRunnable, then a configurable amount of blocking is
  * done on the queue. If the runnable doesn't implement TimeoutRunnable or when the blocking times out,
  * then the task is rejected and a RejectedExecutionException is thrown.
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/util/executor/StripedExecutor.java
```diff
@@ -204,5 +223,10 @@ private void process(Runnable task) {
                 logger.severe(getName() + " caught an exception while processing task:" + task, e);
             }
         }
+
+        // used in tests.
```
pveentjer: I would make the field just package friendly.

pveentjer:
Minor comments. So :+1:

hazelcast/src/main/java/com/hazelcast/util/executor/StripedExecutor.java
```diff
@@ -45,20 +47,27 @@
...
+                           int threadCount, int maximumQueueCapacity) {
+        checkPositive(threadCount, "threadCount should be positive but found " + threadCount);
+        checkPositive(maximumQueueCapacity, "maximumQueueCapacity should be positive but found " + maximumQueueCapacity);
+
         this.logger = logger;
-        this.maximumQueueSize = maximumQueueSize;
         this.size = threadCount;
         this.workers = new Worker[threadCount];
+
+        // Max event queue capacity is determined via system property. Each worker in this executor should consume
```
mdogan: I don't think having *event queue* specific comment in `StripedExecutor ` is a good idea. `StripedExecutor` is not specific to event system only.

ahmetmircik: fixed.

hazelcast/src/test/java/com/hazelcast/util/executor/StripedExecutorTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void throws_illegalArgumentException_whenThreadCount_isNotPositive() throws Exception {
+        expectedException.expect(IllegalArgumentException.class);
+
+        new StripedExecutor(getLogger(getClass()), "", null,
+                0, 0);
```
ahmetmircik: fixed.

PR #9481, opened 2016-12-16
Increase code coverage for HttpPostCommandProcessor

mmedenjak:
Increased code coverage and refactored test helper methods.


pveentjer:
:+1:

PR #9482, opened 2016-12-16
Test Hiccup/Jitter Monitor

jerrinot:
The current impl starts a monitoring thread and never stop it. I am not sure if that's a problem as I assume tests are executed in its own JVM anyway?

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/jitter/JitterRule.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/jitter/JitterRule.java
@@ -1,15 +1,11 @@
 package com.hazelcast.test.jitter;
 
-import com.hazelcast.util.StringUtil;
 import org.junit.rules.TestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
-import java.lang.management.GarbageCollectorMXBean;
-import java.lang.management.ManagementFactory;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
-import java.util.List;
 
 import static com.hazelcast.util.QuickMath.nextPowerOfTwo;
 import static com.hazelcast.util.StringUtil.LINE_SEPARATOR;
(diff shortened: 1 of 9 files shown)


mdogan:
Good and helpful work! 
I haven't really look at `JitterRecorder` impl.  I'm assuming you know what you do there :)

Do you know is it possible to attach junit rule to test runner? Currently only tests extending `HazelcastTestSupport` are measured.

jerrinot:
@mdogan it is actually implemented as a rule. the `HazelcastTestSupport` just [adds this](https://github.com/hazelcast/hazelcast/pull/9482/files#diff-1f0f61976faa2b7e01b0bb185336b895R92) rule by default. However nothing is preventing you to use the rule in arbitrary tests.

mdogan:
@jerrinot: I see, thanks. 

So what's preventing us from merging this?

jerrinot:
Currently it starts a [monitoring thread](https://github.com/hazelcast/hazelcast/pull/9482/files#diff-484693c50114046bbb9f978b3aa0b473R57) and never stops it. 

I don't know if that's a problem. I assume tests are always executed in a new JVM, but I am not 100% sure.

mdogan:
What if we make jitter-rule disabled by default and enable it via system property? We can enable it on jenkins always.

jerrinot:
I added Jenkins auto-detection + a property for explicit control. let's see whether it works at all.

hasancelik:
run-lab-run

jerrinot:
previous failures were intentional - to test reporting on a test failure.

hazelcast/src/test/java/com/hazelcast/test/jitter/JitterRule.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                long startTime = System.currentTimeMillis();
+                try {
+                    base.evaluate();
+                } catch (Throwable t) {
+                    long endTime = System.currentTimeMillis();
```
mmedenjak: We should possibly log the jitter no matter if the test failed or succeed. A rule that is executed after the JitterRule can cause the test to fail, depending on the order in which rules are applied.

PR #9483, opened 2016-12-16
Fixing serializability of the LazyMapEntry

tombujok:


hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
```diff
@@ -43,8 +44,7 @@
  * @see com.hazelcast.map.impl.operation.EntryOperation#createMapEntry(Data, Object)
  */
 
-public class LazyMapEntry extends CachedQueryEntry implements Serializable, IdentifiedDataSerializable {
-    private static final long serialVersionUID = 0L;
+public class LazyMapEntry extends CachedQueryEntry implements IdentifiedDataSerializable {
```
pveentjer: I see that Serializable has been removed. Are you sure that can't cause problems?

hazelcast/src/main/java/com/hazelcast/map/impl/LazyMapEntry.java
```diff
@@ -44,7 +46,6 @@
  */
 
 public class LazyMapEntry extends CachedQueryEntry implements Serializable, IdentifiedDataSerializable {
-    private static final long serialVersionUID = 0L;
```
tombujok: Fixed.

tombujok:
I'm merging since it breaks JET build. I'll look into annotating it with BinaryInterface tomorrow with Vassilis.

Issue #5453, opened 2015-06-05
MapLoaderTest.testMapCanBeLoaded_whenLoadAllKeysThrowsExceptionFirstTime


mesutcelik:
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-OpenJDK7/558/artifact/hazelcast/target/surefire-reports/com.hazelcast.map.mapstore.MapLoaderTest-output.txt

It is weird that it tries to connect Man Center

```
http://127.0.0.1:8090/mancenter
04:24:25,085 DEBUG [ManagementCenterService] hz._hzInstance_2628_dev.MC.State.Sender - [127.0.0.1]:7675 [dev] [3.5-EA2-SNAPSHOT] Opening collector connection:http://127.0.0.1:8090/mancenter/collector.do
04:24:25,085 DEBUG [ManagementCenterService] hz._hzInstance_2628_dev.MC.Task.Poller - [127.0.0.1]:7675 [dev] [3.5-EA2-SNAPSHOT] Opening getTask connection:http://127.0.0.1:8090/mancenter/getTask.do?member=127.0.0.1:7675&cluster=dev
04:24:25,086 DEBUG [ManagementCenterService] hz._hzInstance_2628_dev.MC.Task.Poller - [127.0.0.1]:7675 [dev] [3.5-EA2-SNAPSHOT] Failed to connect to management center
java.net.ConnectException: Connection refused
    at java.net.PlainSocketImpl.socketConnect(Native Method)
    at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:339)
    at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:200)
    at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:182)
    at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
    at java.net.Socket.connect(Socket.java:579)
    at java.net.Socket.connect(Socket.java:528)
    at sun.net.NetworkClient.doConnect(NetworkClient.java:180)
    at sun.net.www.http.HttpClient.openServer(HttpClient.java:432)
    at sun.net.www.http.HttpClient.openServer(HttpClient.java:527)
    at sun.net.www.http.HttpClient.<init>(HttpClient.java:211)
    at sun.net.www.http.HttpClient.New(HttpClient.java:308)
[...]

PR #9496, opened 2016-12-19
Fixes MapLoader #5453

tombujok:
Fixes MapLoader #5453


tombujok:
Tested for 20h on Zulu, IBM-JDK, etc. 
Without a fix it used to fail within an hour.

PR #9497, opened 2016-12-19
Fixes #9255. MapLoader.loadAll(true) with keys

### LINKED ISSUE #9255: MapLoader.loadAll(true) doesn't reload after reloading with new keys

tombujok:
Fixes #9255. 
Tests by @jerrinot


tombujok:
Final statement: I analysed it a couple of times and this solution is fully correct.

PR #9500, opened 2016-12-20
Sent all existing enqueued invalidations

ahmetmircik:
- otherwise invalidations queue can grow very fast under high load
- also this was the old behaviour, this pr restores it.
- fix candidate for https://github.com/hazelcast/hazelcast-enterprise/issues/1203


hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/BatchInvalidator.java
```diff
@@ -114,10 +112,7 @@ private void pollAndSendInvalidations(String mapName, InvalidationQueue invalida
     }
 
     private List<Invalidation> pollInvalidations(InvalidationQueue invalidationQueue) {
```
ahmetmircik: we can only call this method if queue size greater than batchSize here: https://github.com/ahmetmircik/hazelcast/blob/28c512d56e8e0339de2f84a4339d3ae583e10947/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/invalidation/BatchInvalidator.java#L91

PR #9504, opened 2016-12-20
Created operations from partition-aware-factory when OperationFactory…

ahmetmircik:
…Wrapper is a wrapper for it

fixes https://github.com/hazelcast/hazelcast-enterprise/issues/1213

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvokeOnPartitions.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvokeOnPartitions.java
@@ -31,6 +31,8 @@ import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
+import static com.hazelcast.spi.impl.operationservice.impl.operations.PartitionAwareFactoryAccessor.extractPartitionAware;
+
 /**
  * Executes an operation on a set of partitions.
  */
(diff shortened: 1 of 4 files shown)


ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvokeOnPartitions.java
```diff
@@ -124,7 +126,7 @@ private void retryFailedPartitions() throws InterruptedException, ExecutionExcep
 
         for (Integer failedPartition : failedPartitions) {
             Operation operation;
-            if (operationFactory instanceof PartitionAwareOperationFactory) {
+            if (isPartitionAwareFactory(operationFactory)) {
```
pveentjer: Why not add a method 'isPartitionAware' to the OperationFactory interface?

ahmetmircik: Yes good point, here my goal is to fix this issue: hazelcast/hazelcast-enterprise#1213 on what we have already and porting it to maintenance. `PartitionAwareOperationFactory` and `OperationFactory` relation can be handled in a separate issue. Does it make sense?

ahmetmircik: created an issue for this https://github.com/hazelcast/hazelcast/issues/9555

pveentjer:
:+1:  But we should make a note somewhere the design should be cleaned up.

PR #9506, opened 2016-12-20
Fixes problem caused by concurrent running authentications

sancar:
We have a timeout for establishing connection. If owner connection
authentication takes longer than connection, client tries to
authenticate to another node. When first tried authentication message
runs later than the second one, it messes up states at both client
and cluster related to ownership of client.

As a solution,
1. ClientReauthOperation only if request has bigger correlation id
then it is already seen.
2. Server returns exception to client if there is a collusion with
correlation id's so that client does not apply an old auth.
3. When servers return successfully authenticated client applies
only if it has same correlation id with last send authentication message.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
@@ -19,12 +19,16 @@ package com.hazelcast.client.heartbeat;
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.client.connection.ClientConnectionManager;
 import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
+import com.hazelcast.client.spi.impl.ClusterListenerSupport;
 import com.hazelcast.client.spi.impl.ConnectionHeartbeatListener;
 import com.hazelcast.client.spi.properties.ClientProperty;
 import com.hazelcast.client.test.ClientTestSupport;
 import com.hazelcast.client.test.TestHazelcastFactory;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IExecutorService;
 import com.hazelcast.core.IMap;
+import com.hazelcast.core.LifecycleEvent;
+import com.hazelcast.core.LifecycleListener;
 import com.hazelcast.core.Member;
 import com.hazelcast.core.Partition;
 import com.hazelcast.nio.Connection;
@@ -43,6 +47,7 @@ import org.junit.runner.RunWith;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 
 import static org.hamcrest.Matchers.containsString;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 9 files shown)


sancar:
run-lab-run

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -102,14 +105,22 @@
...
+            };
     private final Node node;
     private final NodeEngineImpl nodeEngine;
     private final Executor executor;
 
     private final SerializationService serializationService;
     // client uuid -> member uuid
     private final ConcurrentMap<String, String> ownershipMappings = new ConcurrentHashMap<String, String>();
+    //// client uuid -> last authentication call id
+    private final ConcurrentMap<String, AtomicLong> lastAuthenticationCallIds = new ConcurrentHashMap<String, AtomicLong>();
```
mdogan: Can't you keep auth call-id in client endpoint?

hazelcast/src/main/java/com/hazelcast/client/impl/ReAuthenticationOperationSupplier.java
```diff
@@ -23,13 +23,15 @@
 public class ReAuthenticationOperationSupplier implements Supplier<Operation> {
 
     private final String uuid;
+    private final long correlationIdOfAuthMessage;
```
mdogan: Is this `callId` or `correlationId`? Can you use the same naming everywhere?

hazelcast/src/main/java/com/hazelcast/client/impl/operations/ClientReAuthOperation.java
```diff
@@ -33,19 +33,26 @@
         implements UrgentSystemOperation, AllowedDuringPassiveState {
 
     private String clientUuid;
+    private long clientCallID;
```
mdogan: Is this `callId` or `correlationId`? Can you use the same naming everywhere?

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
sancar: There can be multiple endpoints to same client uuid. But I need to keep this per client uuid. It looks like we can't.

hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
```diff
@@ -211,4 +218,66 @@ private static ClientConfig getClientConfig() {
...
+        //trigger incoming messages from instance2
+        IExecutorService executorService = client.getExecutorService("s");
+        Future f = executorService.submitToMember(new SerializableClass(), instance2.getCluster().getLocalMember());
+        f.get();
+
+        //late authentication response from instance2 should not be able to change clients state
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                assertEquals(instance3.getLocalEndpoint().getUuid(), clientClusterService.getPrincipal().getOwnerUuid());
```
ihsandemir: can you check that the owner membership map has the correct owner at both instance2 and instance3 for the the client.

hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
```diff
@@ -211,4 +218,66 @@ private static ClientConfig getClientConfig() {
...
+        final CountDownLatch countDownLatch = new CountDownLatch(2);
+        client.getLifecycleService().addLifecycleListener(new LifecycleListener() {
+            @Override
+            public void stateChanged(LifecycleEvent event) {
+                countDownLatch.countDown();
+            }
+        });
+
+        HazelcastInstance instance2 = hazelcastFactory.newHazelcastInstance();
+        blockMessagesFromInstance(instance2, client);
```
ihsandemir: this will cause the non-owner authentication request processed for the client but the response will not be returned. But we also want to test the case where authentication task for owner request is not run at all but it needs to be run later. can we test this case as well?

sancar: I could not found a way to test it

hazelcast-client/src/test/java/com/hazelcast/client/heartbeat/ClientHeartbeatTest.java
sancar: I added the checks

ihsandemir:
OK 👍 , after https://github.com/hazelcast/hazelcast/pull/9506#pullrequestreview-14399392 change.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -347,11 +359,22 @@ public void shutdown(boolean terminate) {
         ownershipMappings.clear();
     }
 
+    public boolean checkAndSetGreaterThanLastAuthenticationCorrelationId(String clientUuid, long newCorrelationId) {
```
mdogan: minor: `setIfGreaterThanLastAuthenticationCorrelationId(...)`, to make naming symmetric.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -102,14 +105,23 @@
 
     private static final int EXECUTOR_QUEUE_CAPACITY_PER_CORE = 100000;
     private static final int THREADS_PER_CORE = 20;
-
+    private static final ConstructorFunction<String, AtomicLong> CONSTRUCTOR_FUNC =
```
mdogan: minor: something like `LAST_AUTH_CORRELATION_ID_CONSTRUCTOR_FUNC`...

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -134,6 +134,12 @@ public void processMessage() throws Throwable {
             return;
         }
         String uuid = getUuid();
+        if (clientMessage.getCorrelationId() < clientEngine.getLastAuthenticationCorrelationId(uuid)) {
+            String message = "Server already processed authentication from client with uuid " + uuid
+                    + ". Not applying requested ownership change to " + clientEngine.getThisUuid();
+            logger.info(message);
```
mdogan: Previous one was warning but this is info.  https://github.com/hazelcast/hazelcast/pull/9506/files#diff-6bae815677da05c4ca3cecc2f364fe2bR55

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -134,6 +134,12 @@ public void processMessage() throws Throwable {
             return;
         }
         String uuid = getUuid();
+        if (clientMessage.getCorrelationId() < clientEngine.getLastAuthenticationCorrelationId(uuid)) {
```
mdogan: Isn't this already checked by `ClientReAuthOperation`? Is this a requirement or optimization? This one duplicates the logic in `clientEngine`.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
mdogan: or maybe even just `trySetLastAuthenticationCorrelationId()`

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
sancar: I was not hundred percent sure about this either. In the first version it was not implemented. I added after review comments. This is not requirement, just optimisation.
If correlation id is already updated, this prevents sending operations to cluster.  My thinking is that since authentication is not something common, doing optimisation here not that important. And more code always means more bug.

Issue #8721, opened 2016-08-15
Client disconnect leads to stucked clusterExecutor shutdown (3.6.2)


uladzimirFilipchanka:
I have two apps deployed on different VM's. First app connects to another as Hazelcast Client. 
As soon second app goes down, client starts reconnection as expeced. Finally the client says 

> "Could not re-connect to cluster shutting down the client"

and goes to `ClusterListenerSupport:233` to shutdown LifecycleService. 

After some components shutdowns, called `ClusterListenerSupport.shutdown`
And the component tries to shutdown clusterExecutor and stucked for `TERMINATE_TIMEOUT_SECONDS` because the task is launched inside this clusterExecutor. 

Here is screen from debug view
![b2008d0b19](https://cloud.githubusercontent.com/assets/1476017/17664601/f59af734-62fd-11e6-9279-64f2a158ef30.jpg)

Log says:

> 13:34:41,955 INFO  [com.hazelcast.core.LifecycleService](hz.client_50_ustaxi-shift.cluster-1) HazelcastClient[hz.client_50_ustaxi-shift][3.6.2] is SHUTTING_DOWN
> 13:35:11,957 WARNING [com.hazelcast.client.spi.impl.ClusterListenerSupport](hz.client_50_ustaxi-shift.cluster-1) cluster executor awaitTermination could not completed in 30 seconds
> 13:35:11,960 INFO  [com.hazelcast.core.LifecycleService](hz.client_50_ustaxi-shift.cluster-1) HazelcastClient[hz.client_50_ustaxi-shift][3.6.2] is SHUTDOWN

This issue reproduces permanently and only after target instance goes down.

PR #9508, opened 2016-12-21
Client cluster thread trying to shutdown itself

sancar:
Launch a new thread when client needs to be closed from cluster
thread.

fixes https://github.com/hazelcast/hazelcast/issues/8721

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
@@ -51,7 +51,7 @@ import static com.hazelcast.client.spi.properties.ClientProperty.SHUFFLE_MEMBER_
 
 public abstract class ClusterListenerSupport implements ConnectionListener, ConnectionHeartbeatListener, ClientClusterService {
 
-    private static final long TERMINATE_TIMEOUT_SECONDS = 30;
+    public static final long TERMINATE_TIMEOUT_SECONDS = 30;
 
     protected final HazelcastClientInstanceImpl client;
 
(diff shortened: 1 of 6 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
```diff
@@ -233,7 +233,12 @@ public void run() {
                             connectToCluster();
                         } catch (Exception e) {
                             logger.warning("Could not re-connect to cluster shutting down the client", e);
-                            client.getLifecycleService().shutdown();
+                            new Thread(new Runnable() {
```
pveentjer: Can you also add a name to the thread? This way it become easier to determine what the purpose of this thread is when doing some debugging.

sancar: I added the name.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
```diff
@@ -233,7 +233,16 @@ public void run() {
...
```
ihsandemir: is it possible to utilize the executor of com.hazelcast.client.impl.LifecycleServiceImpl instead of a new Thread creation here?

sancar: No, it is not possible. If we do that we end up with same problem but with different executor.
When `client.getLifecycleService().shutdown()`  returns, it has to make sure all executors are closed. We should not call `client.getLifecycleService().shutdown()`  inside a thread that it needs to close. 
Similar usage on server side 
https://github.com/hazelcast/hazelcast/pull/9508/files#diff-b84f3952c1015997c1cb46310f7af2a1R43

PR #9509, opened 2016-12-21
Client cluster thread trying to shutdown itself

sancar:
Launch a new thread when client needs to be closed from cluster
thread.

fixes https://github.com/hazelcast/hazelcast/issues/8721


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
```diff
@@ -230,7 +230,12 @@ public void run() {
                             connectToCluster();
                         } catch (Exception e) {
                             logger.warning("Could not re-connect to cluster shutting down the client", e);
-                            client.getLifecycleService().shutdown();
+                            new Thread(new Runnable() {
+                                @Override
+                                public void run() {
+                                    client.getLifecycleService().shutdown();
```
pveentjer: I would slap a try/catch-log on this so exceptions aren't lost.

pveentjer:
Looks good. Minor comment.

Issue #8655, opened 2016-08-03
Invocation_RetryTest.invocationShouldComplete_whenRetried_DuringShutdown


hasancelik:
```
java.lang.AssertionError: RetryingOperation failed to complete within 2 MINUTES. Invocation{op=com.hazelcast.spi.impl.operationservice.impl.Invocation_RetryTest$RetryingOperation{serviceName='null', identityHash=218560542, partitionId=175, replicaIndex=0, callId=0, invocationTime=1470167694759 (2016-08-02 22:54:54.759), waitTimeout=-1, callTimeout=9223372036854775807}, tryCount=2147483647, tryPauseMillis=500, invokeCount=7, callTimeoutMillis=9223372036854775807, firstInvocationTimeMs=1470167693754, firstInvocationTime='2016-08-02 22:54:53.754', lastHeartbeatMillis=0, lastHeartbeatTime='1970-01-01 02:00:00.000', target=null, pendingResponse={VOID}, backupsAcksExpected=0, backupsAcksReceived=0, connection=null}
    at org.junit.Assert.fail(Assert.java:88)
    at com.hazelcast.spi.impl.operationservice.impl.Invocation_RetryTest.invocationShouldComplete_whenRetried_DuringShutdown(Invocation_RetryTest.java:149)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-sonar/com.hazelcast$hazelcast/1152/testReport/junit/com.hazelcast.spi.impl.operationservice.impl/Invocation_RetryTest/invocationShouldComplete_whenRetried_DuringShutdown/

hasancelik:
```
[...]

PR #9510, opened 2016-12-21
Fix invocation retry mechanism on node shutdown

metanet:
When the node is shutting down, pending retries may be lost and invocation futures may not be notified. It is because we schedule retries via ExecutionService and we don't keep track of them via invocation registry etc. During shutdown, ExecutionService does not invoke pending tasks.

Here, we move retry tasks to invocation monitor's scheduler. By this way, we guarantee that all of them will run during shutdown. At any time, there can be at most 1 retry task running and we don't apply back pressure to it when it is put into the invocation registry.

Fixes https://github.com/hazelcast/hazelcast/issues/8655

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
@@ -61,7 +61,6 @@ import java.util.logging.Level;
 import static com.hazelcast.cluster.ClusterState.FROZEN;
 import static com.hazelcast.cluster.ClusterState.PASSIVE;
 import static com.hazelcast.cluster.memberselector.MemberSelectors.DATA_MEMBER_SELECTOR;
-import static com.hazelcast.spi.ExecutionService.ASYNC_EXECUTOR;
 import static com.hazelcast.spi.OperationAccessor.hasActiveInvocation;
 import static com.hazelcast.spi.OperationAccessor.setCallTimeout;
 import static com.hazelcast.spi.OperationAccessor.setCallerAddress;
@@ -84,7 +83,6 @@ import static java.lang.Boolean.FALSE;
 import static java.lang.Boolean.TRUE;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static java.util.logging.Level.FINEST;
 import static java.util.logging.Level.WARNING;
@@ -214,6 +212,10 @@ public abstract class Invocation implements OperationResponseHandler {
 
     boolean isActive() {
         return hasActiveInvocation(op);
+    }
+
+    boolean isRetryCandidate() {
+        return op.getCallId() != 0;
     }
 
     /**
@@ -608,13 +610,14 @@ public abstract class Invocation implements OperationResponseHandler {
 
         if (future.interrupted) {
             complete(INTERRUPTED);
-        } else if (context.invocationRegistry.deregister(this)) {
+        } else {
             try {
+                InvocationRetryTask retryTask = new InvocationRetryTask();
                 if (invokeCount < MAX_FAST_INVOCATION_COUNT) {
                     // fast retry for the first few invocations
-                    context.asyncExecutor.execute(new InvocationRetryTask());
+                    context.invocationMonitor.execute(retryTask);
                 } else {
-                    context.executionService.schedule(ASYNC_EXECUTOR, new InvocationRetryTask(), tryPauseMillis, MILLISECONDS);
+                    context.invocationMonitor.schedule(retryTask, tryPauseMillis);
                 }
             } catch (RejectedExecutionException e) {
                 completeWhenRetryRejected(e);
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -106,7 +106,8 @@ long getLastCallId() {
     public boolean register(Invocation invocation) {
         final long callId;
         try {
-            callId = callIdSequence.next(invocation.op.isUrgent());
+            boolean force = invocation.op.isUrgent() || invocation.isRetryCandidate();
```
pveentjer: When an invocation is retried; first it is unregistered and therefor the call id is set to 0. So the invocation.isRetryCandidate AFAIK will always return false.

metanet: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/spi/Operation.java#L216

pveentjer: The other problem is that as part of the deregister, the tail sequence of the id generator is incremented; so any invocation that wants to start can start. This means that it is possible to get more in flight invocations than intended. Even though it is allowed to have a few more in flight invocations than allowed; I'm worried that we could get a bubble building up especially if async invocations are executed.

metanet: I am not sure how your scenario is possible. Non-retry operations are still amenable to max concurrent operations limitation. There can be at most (max concurrent non-retry operations + 1 retry) at any time. Can you elaborate your scenario?

mdogan: @pveentjer: Call id is never set to 0 once it's set to a non-zero value. It's part of Marko's recent change and guarantees that only and only one caller will be able to deregister the invocation.

mdogan: Only executor thread(s) in InvocationMonitor is allowed to submit retries. So, once registry is full, it can be exceed only by number of retrying threads. It's 1 right now.

metanet: here is a scenario to show at most + 1 retry can be executed. 

```
occupied slots: N

retry queue: R1, R2, R3

1) R1 deregister: N -> N - 1
2) new non-retry invocation: N - 1 -> N
3) R1 register: N -> N + 1
4) R2 deregister: N + 1 -> N. A new non-retry invocation cannot start since all slots are occupied. 
5) R2 register: N -> N + 1
```

pveentjer: My main problem, and it could very well I'm missing something, is that the callIdSequence.complete is called when doing an deregister for the sake of a retry.

```
 public boolean deregister(Invocation invocation) {
        if (!deactivate(invocation.op)) {
            return false;
        }
        invocations.remove(invocation.op.getCallId());
        callIdSequence.complete();
        return true;
    }
```

So imagine there are 1000 in flight operations (perhaps a lot of async calls) on the same partition, and lets assume we have max 1000 invocations (back pressure). So we have clallidsequence.head=1000 and callidsequence.tail=0. 

So currently any new invocation is blocking; because the maximum number of in-flight operations have been reached. 

Now we get a partition migration and all these 1000 invocations are going to be bounced and retried. The first thing done is the deregistration of the invocation. So eventually the clallidsequence.head=1000 and callidsequence.tail=1000. And also we have now 1000 invocation scheduled for retrying.

The user is still hammering the system with async calls and the user is now allowed to have another 1000 extra invocations so that eventually clallidsequence.head=2000 and callidsequence.tail=1000.

So at this point in time we have 1000 retrying invocations and 1000 invocation doing their first attempt. Agree?

mdogan: @pveentjer: you're right. somehow we were thinking that register & deregister will be done by the same thread (InvocationMonitor.scheduler).
But deregister is (mostly) done by response thread. So, yes, registry can grow unbounded.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -696,7 +702,7 @@ public void run() {
             // When retrying, we must reset lastHeartbeat, otherwise InvocationMonitor will see the old value
             // and falsely conclude that nothing has been done about this operation for a long time.
             lastHeartbeatMillis = 0;
-            doInvoke(false);
+            doInvoke(true);
```
pveentjer: Why has this been changed to true?

mdogan: Generic operations can be executed on calling thread directly, if not async. But invocation monitor scheduler can't execute operations.

pveentjer:
:+1: good work Basri.

Issue #8038, opened 2016-04-28
RingbufferMigrationTest.test


Donnerbart:
**Error Message**

```
expected:<0> but was:<864>
```

**Stacktrace**

```
java.lang.AssertionError: expected:<0> but was:<864>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:834)
    at org.junit.Assert.assertEquals(Assert.java:645)
    at org.junit.Assert.assertEquals(Assert.java:631)
    at com.hazelcast.ringbuffer.impl.RingbufferMigrationTest.test(RingbufferMigrationTest.java:46)
```

https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/2792/testReport/junit/com.hazelcast.ringbuffer.impl/RingbufferMigrationTest/test/

gurbuzali:
I've tried to reproduce this on my local but couldn't. I think it is safe to assume all those enhancements/fixes related with node shutdown and migration may fixed this issue too.

hasancelik:
```
java.lang.AssertionError: expected:<0> but was:<1000>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.ringbuffer.impl.RingbufferMigrationTest.test(RingbufferMigrationTest.java:46)
```

https://hazelcast-l337.ci.cloudbees.com/view/Official%20Builds/job/Hazelcast-3.x-IbmJDK1.7/com.hazelcast$hazelcast/1010/testReport/junit/com.hazelcast.ringbuffer.impl/RingbufferMigrationTest/test/

hasancelik:
```
java.lang.AssertionError: expected:<0> but was:<1000>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.ringbuffer.impl.RingbufferMigrationTest.test(RingbufferMigrationTest.java:46)
```

[...]

PR #9512, opened 2016-12-21
Disable TTL on ringbuffer migration test

mmedenjak:
The tests can fail if the TTL passes before the assertions are run. This disables the TTL for this particular test since TTL is not being tested here.
Fixes : https://github.com/hazelcast/hazelcast/issues/8038

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferMigrationTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferMigrationTest.java
@@ -1,5 +1,7 @@
 package com.hazelcast.ringbuffer.impl;
 
+import com.hazelcast.config.Config;
+import com.hazelcast.config.RingbufferConfig;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferMigrationTest.java
```diff
@@ -27,22 +29,25 @@ public void setup() {
...
+        long oldTailSeq = hz1.getRingbuffer(ringbufferName).tailSequence();
+        long oldHeadSeq = hz1.getRingbuffer(ringbufferName).headSequence();
 
-        HazelcastInstance hz2 = instanceFactory.newHazelcastInstance();
-        HazelcastInstance hz3 = instanceFactory.newHazelcastInstance();
+        HazelcastInstance hz2 = instanceFactory.newHazelcastInstance(config);
+        HazelcastInstance hz3 = instanceFactory.newHazelcastInstance(config);
         hz1.shutdown();
 
         sleepSeconds(5);
```
vbekiaris: Instead of relying on sleeping an arbitrary number of seconds, maybe remove the sleep statement and use `assertEqualsEventually`?

mmedenjak: I will look into it.

jerrinot:
run-lab-run

hazelcast/src/test/java/com/hazelcast/ringbuffer/impl/RingbufferMigrationTest.java
mmedenjak: Updated the test. The test even lasts less now and it is much clearer. Good suggestion!

PR #9513, opened 2016-12-21
Cleanup invalidations and invalidators

ahmetmircik:


ahmetmircik:
tiny cleanups merging it.

PR #9514, opened 2016-12-22
Wrong uuid set to ReAuthOperation caller uuid.

sancar:
Added ownership tests for client

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/ClientOwnershipTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/ClientOwnershipTest.java
@@ -1,0 +1,183 @@
+package com.hazelcast.client;
+
+import com.hazelcast.client.config.ClientConfig;
+import com.hazelcast.client.impl.ClientEngineImpl;
+import com.hazelcast.client.test.TestHazelcastFactory;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IExecutorService;
+import com.hazelcast.test.AssertTask;
+import com.hazelcast.test.HazelcastParallelClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
+import com.hazelcast.test.annotation.ParallelTest;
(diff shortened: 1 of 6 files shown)


sancar:
run-lab-run

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
```diff
@@ -218,6 +213,7 @@ protected Connection findSendConnection() {
             return connection;
         }
 
+        String clientUuid = endpoint.getUuid();
```
ihsandemir: after this change it seems that you can also remove the cached clientUuid.

hazelcast-client/src/test/java/com/hazelcast/client/ClientOwnershipTest.java
```diff
@@ -0,0 +1,150 @@
+package com.hazelcast.client;
```
ihsandemir: may be you can add this case: client connected, cluster shutdown or all members shutdown, then start the cluster (or one member) again and check owner.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMessageTask.java
sancar: Do you mean this https://github.com/hazelcast/hazelcast/pull/9514/files#diff-62dfc5312563d71974e936132c44bd73L70 ?

ihsandemir: ok, already removed :+1

hazelcast-client/src/test/java/com/hazelcast/client/ClientOwnershipTest.java
sancar: I added the test just now

PR #9515, opened 2016-12-22
Adding client ownership related tests

sancar:
backport of #9514

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMultiTargetMessageTask.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMultiTargetMessageTask.java
@@ -16,7 +16,6 @@
 
 package com.hazelcast.client.impl.protocol.task;
 
-import com.hazelcast.client.ClientEndpoint;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.instance.Node;
 import com.hazelcast.nio.Address;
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AbstractMultiTargetMessageTask.java
```diff
@@ -56,7 +54,6 @@ protected void processMessage() throws Throwable {
         MultiTargetCallback callback = new MultiTargetCallback(targets);
         for (Address target : targets) {
             Operation op = operationFactory.createOperation();
-            op.setCallerUuid(endpoint.getUuid());
```
Donnerbart: Can you explain why this is not needed? This is the only part of the PR I don't understand.

Donnerbart: Ah, this seems to be a backport of https://github.com/hazelcast/hazelcast/pull/9514

sancar: It is not the case it not needed. It is wrong to set that uuid. Endpoint.getUuid is the uuid of client. Caller uuid is used only in one place which is AuthBaseMessageTask(this is true for only master). In maintenance this uuid is not used at all anywhere. And for all the operations generated from client only this one sets the uuid of client. I removed it to make it same for all operations and same with master.

sancar:
Merging this since forward port #9514 confirmed and merged.

PR #9516, opened 2016-12-22
Increase code coverage

mmedenjak:
Increased code coverage for hot backup feature and added preconditions, equals and hashcode to ClusterHotRestartStatusDTO

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/dto/ClusterHotRestartStatusDTO.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/dto/ClusterHotRestartStatusDTO.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import static com.hazelcast.config.HotRestartClusterDataRecoveryPolicy.FULL_RECOVERY_ONLY;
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/internal/management/dto/ClusterHotRestartStatusDTO.java
```diff
@@ -125,4 +131,26 @@ public void fromJson(JsonObject root) {
             memberHotRestartStatusMap.put(member, status);
         }
     }
+
+    @Override
+    @SuppressWarnings("checkstyle:innerassignment")
+    public boolean equals(Object obj) {
```
Donnerbart: If you don't need `equals()` and `hashCode()` in production, just for an `assertEquals()` it's mostly easier to write a specialized assert method in the test than to implement those methods. The main reason is that they normally decrease the code coverage due to the missing branch coverage (especially of the `equals()` method).

An example for this are some of the config class tests, which just compare the fields instead of implementing `equals()` and `hashCode()`.

mmedenjak: Updated.

Issue #8811, opened 2016-08-31
MapTransactionTest.testPutTTL


hasancelik:
```
java.lang.AssertionError: expected:<value> but was:<null>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:834)
    at org.junit.Assert.assertEquals(Assert.java:118)
    at org.junit.Assert.assertEquals(Assert.java:144)
    at com.hazelcast.map.impl.tx.MapTransactionTest.testPutTTL(MapTransactionTest.java:412)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-IbmJDK1.7/com.hazelcast$hazelcast/1015/testReport/junit/com.hazelcast.map.impl.tx/MapTransactionTest/testPutTTL/

PR #9520, opened 2016-12-23
Fix map ttl test

mmedenjak:
The test could fail if the assertion that the entry was in the map was executed too late. Since this is an expiration test, the assertion was changed to just wait until the entry was null.

Fixes: https://github.com/hazelcast/hazelcast/issues/8811

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/tx/MapTransactionTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/tx/MapTransactionTest.java
@@ -42,6 +42,7 @@ import com.hazelcast.query.SampleObjects;
 import com.hazelcast.query.SampleObjects.Employee;
 import com.hazelcast.query.SqlPredicate;
 import com.hazelcast.spi.properties.GroupProperty;
+import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.ExpectedRuntimeException;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/map/impl/tx/MapTransactionTest.java
```diff
@@ -408,14 +413,19 @@ public Boolean execute(TransactionalTaskContext context) throws TransactionExcep
         });
         assertTrue(b);
 
-        IMap map1 = h2.getMap("putWithTTL");
-        assertEquals("value", map1.get("1"));
+        final Object result = map2.get("1");
+        if (deletedCounter.get() > 0) {
+            System.out.println("Value was already removed from map, quitting test");
```
eminn: In the test case we are expecting the value to be evicted eventually, so we don't need to test against value insertion but value eviction.

So we can refactor the remaining test to an `assertTrueEventually` block which checks the `deletedCounter.get()>0` and `assertNull(map2.get("1"))`. This way we can also get rid of the `sleep`

mmedenjak:
run-lab-run

hazelcast/src/test/java/com/hazelcast/map/impl/tx/MapTransactionTest.java
mmedenjak: This was changed. Also, the atomic counter was removed due to the fact that the interceptor can be serialized and invoked on an another instance, thus failing to pass assertion. As the counter was irrelevant and we already have one assertion that the map contains the value, we just need to assert eventually that the map does not contain the value anymore. WDYT?

mmedenjak:
run-lab-run

mmedenjak:
@eminn can you check if it looks ok now?

jerrinot:
run-lab-run

Issue #8840, opened 2016-09-05
NearCacheLiteMemberTest.testPutAll


hasancelik:
```
java.lang.AssertionError: NearCache size should be > 0 but was 0
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at com.hazelcast.map.nearcache.NearCacheLiteMemberTest.assertLiteMemberNearCacheNonEmpty(NearCacheLiteMemberTest.java:407)
    at com.hazelcast.map.nearcache.NearCacheLiteMemberTest.testPutAll(NearCacheLiteMemberTest.java:169)
    at com.hazelcast.map.nearcache.NearCacheLiteMemberTest.testPutAll(NearCacheLiteMemberTest.java:69)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-ZuluJDK8/com.hazelcast$hazelcast/316/testReport/junit/com.hazelcast.map.nearcache/NearCacheLiteMemberTest/testPutAll/

jerrinot:
this looks like a fragile test - near cache invalidations can arrive after the near-cache is populated -> the near-cache is emptied -> assertion failure.

mmedenjak:
I have managed to reproduce the issue and have been thinking about a solution to this. We can use the key state markers to see if the near cache was emptied after the put operations. The problem with that approach is that there are only 271 markers and by observing that the markers changed we cannot be sure that all keys have been emptied on the near cache. So a not pretty approach is filling the near cache before the put operations and waiting after the put operations until the near cache is cleared. It fixes the issue but doesn't look nice. Comments are welcome.

```
public static void testPutAll(HazelcastInstance instance, HazelcastInstance lite, String mapName) {
        IMap<Object, Object> map = instance.getMap(mapName);
        IMap<Object, Object> liteMap = lite.getMap(mapName);
        NearCachedMapProxyImpl proxy = (NearCachedMapProxyImpl) liteMap;
        NearCache liteNearCache = proxy.getNearCache();
        final InternalSerializationService serializationService = getNode(instance).getSerializationService();
        int count = 100;

        // fill the near cache with the same data as below so we can detect when it is emptied
        for (int i = 0; i < count; i++) {
            liteNearCache.put(serializationService.toData(i), i);
        }
        final NearCacheStats stats = liteNearCache.getNearCacheStats();
[...]

PR #9521, opened 2016-12-23
Fix near cache lite member test

mmedenjak:
The test first sends invalidation events to the lite member and then fills the near cache. If the invalidation events arrive after the near cache is filled, the test fails. This fix first fills the near cache and waits for the invalidation events to arrive before filling the near cache again.

Fixes : https://github.com/hazelcast/hazelcast/issues/8840


mmedenjak:
run-lab-run

Issue #7028, opened 2015-12-08
Improve generics in EntryProcessors related methods in IMap


jerrinot:
`IMap` has methods as `Object executeOnKey(K key, EntryProcessor entryProcessor);` instead of `Object executeOnKey(K key, EntryProcessor<K, V> entryProcessor);`

It can lead to bugs where an entry processor accidentally changes a type of the value stored -> `ClassCastException`. What's worst the exception will likely be thrown later during a different operation -> a very hard to find class bugs.

Furthermore the `EntryProcessor` should probably use a generic type as its return value. So
instead of the current:

``` java
public interface EntryProcessor<K, V> extends Serializable {
    Object process(Map.Entry<K, V> entry);
[...]
```

we should have:

``` java
public interface EntryProcessor<K, V, R> extends Serializable {
    R process(Map.Entry<K, V> entry);
[...]
```

Attention must be paid to possible compile-time compatibility issues when introducing these changes.

LoneRifle:
@tombujok @jerrinot I plan to submit a PR in relation to this issue, unless @tombujok has already started work on this?

PR #9527, opened 2016-12-27
Declare and impl TypedEntryProcessor and supporting cast

LoneRifle:
* Declare a new interface - TypedEntryProcessor - that extends EntryProcessor, but allows us to return objects whose type is determined by a third generic param

* Provide overloaded equivalents of methods on IMap interface that have a EntryProcessor argument so that it specifically handles TypedEntryProcessor, declaring a generic return type bound to TypedEntryProcessor.

This allows us to retain binary compatibility with existing EntryProcessor implementations in the wild, while still giving users an option to have the means to get results from entry processing of a specific type without casting.

This PR attempts to address the second part of #7028


LoneRifle:
Closed on discussion with devs; this will add to the clutter in the IMap API, and given we are so many revisions into the 3.x series, the changes proposed may be deprecated very quickly.

PR #9528, opened 2016-12-27
Introduces test for Serializable classes conventions

### LINKED ISSUE #8890: Change serialisation in classes using java.io.Serializable

vbekiaris:
Annotates `Serializable` classes in which serializable form changes may break compatibility with `@BinaryInterface`.
- `Aggregator` & `Projection` is used in client-member communication
- `EvictionPolicyComparator` is used in `CacheEvictionConfig` which is itself a @BinaryInterface
- `MulticastMemberInfo` is used in `MulticastDiscoveryStrategy`

Additionally:
- `CacheEventListenerAdaptor` converted to IdentifiedDataSerializable
- Removed `Serializable` from `CacheStatisticsMXBeanImpl`, its only field is marked transient and it's in an internal package. The only use case I see could be if registered with a remote MBeanServer, but this use case is already broken (`statistics` field is `transient` so the deserialized object would not return any values).
- `CacheEntryListener` should not implement `Serializable`, I don't see it not serialized/deserialized anywhere.

Fixes #8890

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
@@ -31,6 +31,7 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
+import java.io.Serializable;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.util.HashMap;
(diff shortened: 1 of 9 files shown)


vbekiaris:
run-lab-run

hazelcast/src/test/java/com/hazelcast/internal/serialization/impl/DataSerializableConventionsTest.java
```diff
@@ -99,6 +99,54 @@ public void test_dataSerializableClasses_areIdentifiedDataSerializable() {
     }
 
     /**
+     * Verifies that any class which is {@link DataSerializable} and is not annotated with {@link BinaryInterface}
+     * is also an {@link IdentifiedDataSerializable}.
```
tombujok: Just a comment typo -> Should be Serializable instead of DataSerializable

vbekiaris: thanks, fixed.

vbekiaris:
run-lab-run

PR #9531, opened 2016-12-28
Fix URL encoding problem while getting tasks from Management Center

emre-aydin:
Fixes https://github.com/hazelcast/management-center/issues/299


mesutcelik:
👍

PR #9534, opened 2016-12-28
Apply join mechanism fixes

metanet:
* Update joined flag and master address within cluster service lock so that master address cannot be reset in racy cases after the node is already joined.
* Verify master address within cluster service lock to eliminate races when there are in-flight member list update operations and master change.


metanet:
run-lab-run

PR #9541, opened 2016-12-28
Improvements of Near Cache pre-loader

Donnerbart:
* Added `NearCachePreloaderLock` to ensure that only one client writes to the same storage file.
* Added test with two clients and same storage file.
* Added `lastPersistenceFailure` to `NearCacheStats` to ease debugging.
* Changed configuration parameter name from `file-name` to `filename`.

This fix prevents obscure exceptions like this one:
```
Could not store keys of Near Cache /hz/defaultNearCache (/home/donnerbart/IdeaProjects/hazelcast/nearCache-defaultNearCache.store)
com.hazelcast.core.HazelcastException: Failed to rename /home/donnerbart/IdeaProjects/hazelcast/nearCache-defaultNearCache.store~ to /home/donnerbart/IdeaProjects/hazelcast/nearCache-defaultNearCache.store because /home/donnerbart/IdeaProjects/hazelcast/nearCache-defaultNearCache.store~ doesn't exist.
```

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
@@ -93,6 +93,7 @@ public class NearCachePreloader<KS> {
     private final NearCacheStatsImpl nearCacheStats;
     private final SerializationService serializationService;
 
+    private final NearCachePreloaderLock lock;
     private final File storeFile;
     private final File tmpStoreFile;
 
@@ -105,9 +106,14 @@ public class NearCachePreloader<KS> {
         this.nearCacheStats = nearCacheStats;
         this.serializationService = serializationService;
 
-        String fileName = getFileName(preloaderConfig.getFileName(), nearCacheName);
-        this.storeFile = new File(fileName);
-        this.tmpStoreFile = new File(fileName + "~");
+        String filename = getFileName(preloaderConfig.getFilename(), nearCacheName);
+        this.lock = new NearCachePreloaderLock(logger, filename + ".lock");
+        this.storeFile = new File(filename);
+        this.tmpStoreFile = new File(filename + "~");
+    }
+
+    public void destroy() {
+        lock.release();
     }
 
     /**
@@ -197,6 +203,8 @@ public class NearCachePreloader<KS> {
             updatePersistenceStats(startedNanos);
         } catch (Exception e) {
             logger.warning(format("Could not store keys of Near Cache %s (%s)", nearCacheName, storeFile.getAbsolutePath()), e);
+
+            nearCacheStats.addPersistenceFailure(e);
         } finally {
             deleteQuietly(tmpStoreFile);
             closeResource(fos);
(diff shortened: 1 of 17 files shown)


hazelcast/src/main/java/com/hazelcast/monitor/impl/NearCacheStatsImpl.java
```diff
@@ -208,6 +227,7 @@ public JsonObject toJson() {
         root.add("lastPersistenceDuration", lastPersistenceDuration);
         root.add("lastPersistenceWrittenBytes", lastPersistenceWrittenBytes);
         root.add("lastPersistenceKeyCount", lastPersistenceKeyCount);
+        root.add("lastPersistenceFailure", lastPersistenceFailure == null ? NULL_MARKER : lastPersistenceFailure);
```
Donnerbart: It seems not supported to store `null` in a `JsonObject`, at least I got exceptions when I did this. This is the reason for this stupid `NULL_MARKER`...

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -107,7 +108,7 @@ public NearCachePreloader(String nearCacheName, NearCachePreloaderConfig preload
 
         String fileName = getFileName(preloaderConfig.getFileName(), nearCacheName);
         this.storeFile = new File(fileName);
-        this.tmpStoreFile = new File(fileName + "~");
+        this.tmpStoreFile = new File(fileName + "-" + newUnsecureUUID());
```
Donnerbart: It's not easy to pass the real member/client UUID down to the Near Cache pre-loader. So I just use a random UUID. Since it's a temporary file the benefit of having the real UUID seems to be small. If there is an exception there should be the filename in the message anyway.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloaderLock.java
```diff
@@ -0,0 +1,86 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.RandomAccessFile;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileLock;
+import java.nio.channels.OverlappingFileLockException;
+
+import static com.hazelcast.nio.IOUtil.closeResource;
+
+class NearCachePreloaderLock {
```
Donnerbart: I notice two differences: `HazelcastException` instead of `HotRestartException` and I do a `lockFile.deleteOnExit();` in the `finally` block. This might either break some tests (different exceptions) or some behavior (lock file being deleted). Beside that we surely can use the same code base for both features.

Donnerbart: And of course I don't have the HotRestart home in the log messages, but that might be a minor issue.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Issue #9333, opened 2016-11-28
BasicClientReplicatedMapNearCacheTest Failures


hasancelik:
**testNearCacheMemoryCostCalculation**
```
java.lang.AssertionError: The Near Cache is empty, there should be no owned entry memory costs expected:<0> but was:<81>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.internal.nearcache.AbstractBasicNearCacheTest.testNearCacheMemoryCostCalculation(AbstractBasicNearCacheTest.java:386)
	at com.hazelcast.internal.nearcache.AbstractBasicNearCacheTest.testNearCacheMemoryCostCalculation(AbstractBasicNearCacheTest.java:335)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:88)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:82)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.lang.Thread.run(Thread.java:745)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-ZuluJDK7/com.hazelcast$hazelcast-client/401/testReport/junit/com.hazelcast.client.replicatedmap.nearcache/BasicClientReplicatedMapNearCacheTest/testNearCacheMemoryCostCalculation_format_BINARY_/

[...]

PR #9544, opened 2016-12-28
Fix BasicClientReplicatedMapNearCacheTest failures

mmedenjak:
The tests would fail because the assertion passed if the owned count was > 0 and later the cost was asserted to be equal to 0.
Fixes:https://github.com/hazelcast/hazelcast/issues/9333


sancar:
I did not get the fix. Now you are checking for something more restricted than before. This change leads to more failures if nothing else. 
Equal or greater to  only equal ?

mmedenjak:
@sancar the test was removing entries from the near cache and at the end of the test asserting that the entry count was >= 0 and, if so, the entry cost was 0. In this case, the near cache doesn't need to be emptied yet, the owned entry count can be e.g. 4 and the cost can be 81 for the first assertion to pass and the second assertion to fail.

pveentjer:
run-lab-run

PR #9547, opened 2016-12-29
Fire STARTING lifecycle event after listeners are registered

### LINKED ISSUE #9189: LifecycleState events not received
I am using the official Docker image for hazelcast (tried both versions 3.7.1 and 3.7.2)
I have registered a lifecycle listener through the xml configuration file.
However I cannot see it receiving the _STARTING_ lifecycle event, or the _SHUTTING_DOWN_ and _SHUTDOWN_ events, only the _STARTED_ event.

After going through the source code, I have traced it to [firing lifecycle STARTING event](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceImpl.java#L132) before [lifecycle listeners are added from the configuration](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceImpl.java#L134)

With regards to the _SHUTTING_DOWN_ and _SHUTDOWN_ events, I am not quite sure why this happens, I request a docker container instance to stop using `docker stop _INSTANCE_`, but neither event is received from the listener.

metanet:
Fixes #9189

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceImpl.java
@@ -89,7 +89,6 @@ import java.util.Collection;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
-import static com.hazelcast.core.LifecycleEvent.LifecycleState.STARTING;
 import static com.hazelcast.util.Preconditions.checkNotNull;
 
 @PrivateApi
@@ -132,9 +131,9 @@ public class HazelcastInstanceImpl implements HazelcastInstance {
 
         try {
             logger = node.getLogger(getClass().getName());
-            lifecycleService.fireLifecycleEvent(STARTING);
 
             node.start();
+
             if (!node.isRunning()) {
                     throw new IllegalStateException("Node failed to start!");
             }
--- a/hazelcast/src/main/java/com/hazelcast/instance/Node.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/Node.java
@@ -30,6 +30,7 @@ import com.hazelcast.config.MemberAttributeConfig;
 import com.hazelcast.core.ClientListener;
 import com.hazelcast.core.DistributedObjectListener;
 import com.hazelcast.core.HazelcastInstanceAware;
+import com.hazelcast.core.LifecycleEvent.LifecycleState;
 import com.hazelcast.core.LifecycleListener;
 import com.hazelcast.core.MembershipListener;
 import com.hazelcast.core.MigrationListener;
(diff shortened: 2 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/instance/HazelcastInstanceImpl.java
```diff
@@ -132,9 +132,10 @@ protected HazelcastInstanceImpl(String name, Config config, NodeContext nodeCont
 
         try {
             logger = node.getLogger(getClass().getName());
-            lifecycleService.fireLifecycleEvent(STARTING);
```
mdogan: Shouldn't `STARTING ` be fired before `node.start()`?

metanet: `Node.start()` is initializing internal stuff. I am not sure if it makes a difference to the user to get the `STARTING` event before internal initialization starts or not. If you think it matters, we can register the listeners first and invoke the lifecycle listener before `start()` is invoked.

mdogan: When `node.start()` returns, node becomes started and joined to the cluster. To me `STARTING ` means the state before join, `STARTED` means node is ready to operate. 

Not sure but in this case, I don't see any significant difference between `STARTING` and `STARTED`.

metanet: Then, what about invoking it within `Node.start()` like this:

```
void start() {
        nodeEngine.start();
        initializeListeners(config);

        hazelcastInstance.lifecycleService.fireLifecycleEvent(STARTING);

        connectionManager.start();
        ...
```

jerrinot: Imagine you have a `HazelcastInstanceAware` listener. 

I don't see much usage for the `STARTING`. It says something: "We have initiated `HazelcastInstance` start, but it's not quite ready yet".

metanet: Then, we should have another rule for `HazelcastInstanceAware` interface to state that the system provides you the `HazelcastInstance` object after the start up is completed.

jerrinot: yes, that would be ideal. at least this is [documented](http://docs.hazelcast.org/docs/3.7/manual/html-single/index.html#implementing-hazelcastinstanceaware): 
`[...] The reason is that when HazelcastInstance is injected for a HazelcastInstanceAware implementation, it may not be up and running at the injection time.`

metanet:
run-lab-run

PR #9552, opened 2016-12-29
Clean up quorum implementation

### LINKED ISSUE #8820: QuorumService inconsistent state
Team,

from discussion with @metanet: 

> When cluster service starts during node start, it registers itself as the only member and publishes the membership event.  `QuorumService` is a `MembershipAwareService`, therefore it is expected to update quorums when it receives these membership events. If I don’t miss anything, the problem is as follows: services are not registered to ServiceManager yet when cluster service publishes this event on node start. So, it seems that QuorumService is missing it and failing to update its internal quorum status (initial value is “true” but I think it should be sth like “not-decided”).

Test to reproduce.

```
@Test
public void test() {
        Config config = new Config();
        QuorumConfig quorumConfig = new QuorumConfig().setName(randomString()).setEnabled(true);
        quorumConfig.setSize(3);
        quorumConfig.setType(QuorumType.READ_WRITE);
        config.addQuorumConfig(quorumConfig);
        HazelcastInstance hazelcastInstance = createHazelcastInstance(config);
        System.out.println(hazelcastInstance.getQuorumService().getQuorum("q").isPresent());
}
```

returns `true`.

[ZD ticket `1961`](https://hazelcast.zendesk.com/agent/tickets/1961)

metanet:
Please see commit messages for further details.

Fixes #8820

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/Node.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/Node.java
@@ -373,6 +373,7 @@ public class Node {
         nodeEngine.start();
         initializeListeners(config);
         hazelcastInstance.lifecycleService.fireLifecycleEvent(LifecycleState.STARTING);
+        clusterService.sendLocalMembershipEvent();
         connectionManager.start();
         if (config.getNetworkConfig().getJoin().getMulticastConfig().isEnabled()) {
             final Thread multicastServiceThread = new Thread(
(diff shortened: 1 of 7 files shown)


metanet:
run-lab-run

metanet:
run-lab-run

pveentjer:
Since there are 2 thumbs up, I'm hitting the merge button.

PR #9563, opened 2016-12-30
Fixed JVM crash in Near Cache preloader for NATIVE in-memory-format.

Donnerbart:
Changed the `NearCachePreloader` to accept an key-set `Iterator` instead of a number of `NearCacheRecordMap` instances. So the `NearCachePreloader` doesn't know anymore from how many sources the keys are coming from, which makes it an implementation detail of EE to deal with the segmented `NearCacheRecordStore`.

The actual fix is in the Enterprise PR, not in this one.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
@@ -18,8 +18,6 @@ package com.hazelcast.internal.nearcache.impl.preloader;
 
 import com.hazelcast.config.NearCachePreloaderConfig;
 import com.hazelcast.internal.adapter.DataStructureAdapter;
-import com.hazelcast.internal.nearcache.NearCacheRecord;
-import com.hazelcast.internal.nearcache.impl.NearCacheRecordMap;
 import com.hazelcast.internal.serialization.impl.HeapData;
 import com.hazelcast.internal.util.BufferingInputStream;
 import com.hazelcast.logging.ILogger;
@@ -37,7 +35,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
-import java.util.Set;
+import java.util.Iterator;
 
 import static com.hazelcast.nio.Bits.INT_SIZE_IN_BYTES;
 import static com.hazelcast.nio.Bits.readIntB;
@@ -55,7 +53,7 @@ import static java.util.concurrent.TimeUnit.NANOSECONDS;
 /**
  * Loads and stores the keys from a Near Cache into a file.
  *
- * @param <KS> type of the {@link NearCacheRecord} keys
+ * @param <KS> type of the {@link com.hazelcast.internal.nearcache.NearCacheRecord} keys
  */
 public class NearCachePreloader<KS> {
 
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -157,16 +153,11 @@ private boolean checkHeader(BufferingInputStream bis) throws IOException {
...
-     * {@link com.hazelcast.internal.nearcache.NearCacheRecordStore} which contains multiple records maps.
+     * Stores the Near Cache keys from the supplied iterator.
      *
-     * @param records the {@link NearCacheRecordMap} instances to retrieve the keys from
-     * @param <R>     the {@link NearCacheRecord} type
-     * @param <NCRM>  the {@link NearCacheRecordMap} type
+     * @param keySetIterator the key set to store
      */
-    public <R extends NearCacheRecord, NCRM extends NearCacheRecordMap<KS, R>> void storeKeys(NCRM... records) {
+    public void storeKeys(Iterator keySetIterator) {
```
pveentjer: Why does keySetIterator have a raw-type?

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -180,9 +171,7 @@ private boolean checkHeader(BufferingInputStream bis) throws IOException {
             writeInt(fos, FileFormat.INTERLEAVED_LENGTH_FIELD.ordinal());
 
             // writer keys
```
pveentjer: The method 'writeKeySet' already is very self explanatory.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -234,9 +223,9 @@ private int loadKeySet(BufferingInputStream bis, DataStructureAdapter<Data, ?> a
         return loadedKeys;
     }
 
-    private void writeKeySet(FileOutputStream fos, FileChannel outChannel, Set<KS> keySet) throws IOException {
-        for (KS key : keySet) {
-            Data dataKey = serializationService.toData(key);
+    private void writeKeySet(FileOutputStream fos, FileChannel outChannel, Iterator keySetIterator) throws IOException {
```
pveentjer: I'm confused about the content of the keySetIterator. Apart from not having a type; the name gives the impression that it is an iterator that returns key-sets; not keys. But when I look at the while loop; it gives the impression that the keySetIterator is returning keys. So perhaps the name of the keySetIterator should be 'keyIterator'?

pveentjer:
Looks good. Minor comments.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
Donnerbart: Because of overly complex generics in the abstractions of the `NearCacheRecordStore` implementations: `AbstractNearCacheRecordStore<K, V, KS, R extends NearCacheRecord, NCRM extends NearCacheRecordMap<KS, R>>`. I eventually gave up to provide the correct type here in the scope of this PR.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
Donnerbart: It's an iterator over the keySet of a data structure.
```
map.keySet().iterator()
map.values().iterator()
map.entrySet().iterator()
```
We want the `keySet().iterator()` here, therefore the name.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
Donnerbart: Okay, I think I figured it out. At least I don't get compile errors, runtime errors and the test seems to run.

PR #9571, opened 2017-01-03
removed Lifecylelistener from mancenter service and fixed ManagementC…

hasancelik:
…enterServiceIntegrationTest.testTimedMemberStateNotNull

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
@@ -20,9 +20,6 @@ import com.eclipsesource.json.JsonObject;
 import com.eclipsesource.json.JsonValue;
 import com.hazelcast.config.GroupConfig;
 import com.hazelcast.config.ManagementCenterConfig;
-import com.hazelcast.core.LifecycleEvent;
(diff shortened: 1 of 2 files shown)


hasancelik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
```diff
@@ -164,7 +157,7 @@ static String cleanupUrl(String url) {
         return url.endsWith("/") ? url : url + '/';
     }
 
-    public void start() {
+    private void start() {
```
mdogan: I think you should check whether management center is enabled.

```java
        if (!managementCenterConfig.isEnabled()) {
            return;
        }
```

mdogan: Also you shouldn't register membership listener if not enabled.

hasancelik: Yes, you are right. fixed https://github.com/hazelcast/hazelcast/pull/9571/commits/f75fc2f44a0778c910e3f1f9e0f404368446cde3#diff-3850ee7b4bd9df3672e6c4b18d1ec70cR131

PR #9572, opened 2017-01-03
Validate root xml config element name

### LINKED ISSUE #7342: Hazelcast client silently fails when loading server side config
If a server side config is supplied to the client xml config builder using:

``` java
        ClientConfig config = new XmlClientConfigBuilder( "client1.xml" ).build();
        client = HazelcastClient.newHazelcastClient( config );
```

Where the client config is:

```
<?xml version="1.0" encoding="UTF-8"?>
<hazelcast xsi:schemaLocation="http://www.hazelcast.com/schema/config hazelcast-config-3.6.xsd"
           xmlns="http://www.hazelcast.com/schema/config"
           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

    <network>
        <port auto-increment="true">5705</port>

        <join>
            <multicast enabled="false"/>
            <tcp-ip enabled="true">
                <member-list>
                    <member>127.0.0.1:5705</member>
                    <member>127.0.0.1:5706</member>
                </member-list>
            </tcp-ip>
        </join>

        <interfaces enabled="true">
            <interface>127.0.0.1</interface>
        </interfaces>

    </network>

</hazelcast>
```

this silently passes without error, but the cluster member list is not processed (because the XML is incorrect for the client).

Perhaps the client side XML config builder should fail if it does not find a `<hazelcast-client>` root node.  It would also be nice if the client warned about the usage of the server side element `<member-list>` when the `<cluster-members>` element should be used (perhaps it should warn about all server-side only elements encountered).

Or make the client side XML process the same XML elements as the server side XML and drop the client specific XML.

mdogan:
Fixes #7342

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/config/XmlClientConfigBuilder.java
@@ -76,6 +76,7 @@ import static com.hazelcast.util.StringUtil.upperCaseInternal;
 /**
  * Loads the {@link com.hazelcast.client.config.ClientConfig} using XML.
  */
+@SuppressWarnings("checkstyle:methodcount")
 public class XmlClientConfigBuilder extends AbstractConfigBuilder {
 
     private static final ILogger LOGGER = Logger.getLogger(XmlClientConfigBuilder.class);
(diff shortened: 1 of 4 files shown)


pveentjer:
Dumb questions: can't this be verified by the XSD?

mdogan:
@pveentjer: I really don't know about xsd validation. I just tried to solve the simplest way I can.

pveentjer:
👍

PR #9576, opened 2017-01-04
Fixes delivering MemberLeftException on callback in AbstractInvocationFuture

sancar:
Since MemberLeftException itself is ExecutionException it is
unwrapped. And It has always null cause.

Note: MemberLeftException is the only exception extending
ExecutionException in hazelcast.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
@@ -39,11 +39,11 @@ import static java.util.concurrent.locks.LockSupport.unpark;
 
 /**
  * Custom implementation of {@link java.util.concurrent.CompletableFuture}.
- *
+ * <p>
  * The long term goal is that this whole class can be ripped out and replaced by {@link java.util.concurrent.CompletableFuture}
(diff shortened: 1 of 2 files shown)


sancar:
Not sure if best way to solve this, since I am not sure about its side-effects on node side. I am open to suggestions.

Danny-Hazelcast:
https://github.com/hazelcast/hazelcast/issues/9584

hazelcast/src/main/java/com/hazelcast/spi/impl/AbstractInvocationFuture.java
```diff
@@ -261,6 +262,9 @@ public void run() {
 
     // this method should not be needed; but there is a difference between client and server how it handles async throwables
     protected Throwable unwrap(Throwable throwable) {
+        if (throwable instanceof MemberLeftException) {
+            return throwable;
+        }
         if (throwable instanceof ExecutionException) {
             return throwable.getCause();
```
jerrinot: wouldn't be better to have `if (throwable instanceof ExecutionException && throwable.getCause() != null))` instead?

sancar: Yes, I think that would be better. Changing it now.

sancar:
run-lab-run

PR #9577, opened 2017-01-04
Allow invokeAllPartitions from generic operation thread.

### LINKED ISSUE #8492: IllegalThreadStateException when creating Map with configured indexes
An `IllegalThreadStateException` may be thrown sometimes on a member of a cluster with 2 or more members when a hazelcast client creates a new map with configured indexes. This exception is not thrown when indexes are created programmatically.
Thanks @noctarius for reporting this with the following reproducer:

```
import com.hazelcast.config.MapConfig;
import com.hazelcast.config.MapIndexConfig;
import com.hazelcast.core.IMap;
import java.util.List;

public class Test {

        MapConfig mapConfig = config.getMapConfig("test");
        List<MapIndexConfig> mapIndexConfigs = mapConfig.getMapIndexConfigs();

        MapIndexConfig mapIndexConfig = new MapIndexConfig();
        mapIndexConfig.setAttribute("foo");
        mapIndexConfig.setOrdered(true);
        mapIndexConfigs.add(mapIndexConfig);

        Hazelcast.newHazelcastInstance(config);
        Hazelcast.newHazelcastInstance(config);
        HazelcastInstance hazelcastInstance = HazelcastClient.newHazelcastClient();
        IMap<String, String> test = hazelcastInstance.getMap("test");

        test.put("foo", "bar");
    }
}
```

vbekiaris:
Since #4889 `PartitionIteratingOperation` is non blocking, so
generic operation threads should be able to invoke on all partitions.

Fixes #8492 in which map proxy creation would fail when
client targets proxy creation on member other than client
connection owner member.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapWithIndexCreationTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapWithIndexCreationTest.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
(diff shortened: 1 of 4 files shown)


vbekiaris:
Fixes #9578

hazelcast-client/src/test/java/com/hazelcast/client/map/ClientMapWithIndexCreationTest.java
```diff
@@ -0,0 +1,100 @@
...
+                        hz2.getCluster().getLocalMember().getAddress().getPort());
+
+        HazelcastInstance client = factory.newHazelcastClient(clientConfig);
+
+        try {
+            IMap<String, SampleObjects.Employee> test = client.getMap("test");
+            test.put("foo", new SampleObjects.Employee(1, "name", "age", 32, true, 230));
+        } finally {
+            client.shutdown();
+            factory.shutdownAll();
```
mdogan: minor: just to make test method clean, you can shutdown factory in `@After` method. Also explicitly shutting down the client is no needed, factory handles that.

vbekiaris: Thanks, amended as suggested.

PR #9579, opened 2017-01-04
Improved query cache accumulators

Donnerbart:
* Removed unused methods from Accumulator interface and implementations.
* Improved some generics to reduce warnings.
* Fixed a duplicated logger.isFinestEnabled() in CoalescingPublisherAccumulator.
* Made classes and constructors package private where ever possible.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/querycache/QueryCacheCoalescingTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/querycache/QueryCacheCoalescingTest.java
@@ -1,7 +1,6 @@
 package com.hazelcast.map.impl.querycache;
 
 import com.hazelcast.config.Config;
-import com.hazelcast.config.MapConfig;
 import com.hazelcast.config.QueryCacheConfig;
 import com.hazelcast.core.EntryEvent;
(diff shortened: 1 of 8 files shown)


hazelcast/src/test/java/com/hazelcast/map/impl/querycache/QueryCacheCoalescingTest.java
```diff
@@ -32,6 +33,17 @@
     @SuppressWarnings("unchecked")
     private static final Predicate<Integer, Integer> TRUE_PREDICATE = TruePredicate.INSTANCE;
 
+    @Before
```
vbekiaris: Do we really need `DEBUG` logging for this test?

Donnerbart: The `CoalescingPublisherAccumulator` has two `logger.isFinestEnabled()` blocks which are pushing the code coverage below the threshold. This should enable those code paths.

vbekiaris: 👍 thanks

PR #9580, opened 2017-01-04
Fix broken scheduled task replication when completed

tkountis:
When cancelling a task, the replication algorithm is still promoting the task in the new node causing further scheduling. This misbehaviour was noted during tests, where after cancelling, the node is shutting down, and before asserting the stats the new owner already restarts the task and increments the counters.

This should fix all tests failing with counters greater than the expected value, eg. expected 5 actual 6.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
+++ b/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
@@ -23,27 +23,36 @@ import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Metadata holder for scheduled tasks.
- * Scheduled ones have a reference in their future in {@link #scheduledFuture}.
+ * Scheduled ones have a reference in their future in {@link #future}.
  * Stashed ones have this reference null.
  */
 public class ScheduledTaskDescriptor implements IdentifiedDataSerializable {
 
     private TaskDefinition definition;
 
-    private ScheduledFuture<?> scheduledFuture;
+    private ScheduledFuture<?> future;
 
     /**
-     * Stats are written from a single thread and read by many outside the partition threads. (see. Member owned tasks)
+     * SPMC (see. Member owned tasks)
      */
-    private AtomicReference<ScheduledTaskStatisticsImpl> stats = new AtomicReference<ScheduledTaskStatisticsImpl>(null);
+    private volatile ScheduledTaskStatisticsImpl stats;
 
     /**
-     * State is only write/read only through the partition threads.
+     * MPMC (Multiple Producers Multiple Concumers)
+     * MP when cancelling, due to member owned tasks, all other writes are SP and through partition threads.
+     * Reads are MP for member owned tasks.
+     */
+    private AtomicReference<ScheduledTaskResult> resultRef = new AtomicReference<ScheduledTaskResult>(null);
+
+    /**
+     * SPMC
      */
     private Map<?, ?> state;
 
@@ -53,13 +62,16 @@ public class ScheduledTaskDescriptor implements IdentifiedDataSerializable {
     public ScheduledTaskDescriptor(TaskDefinition definition) {
         this.definition = definition;
         this.state = new HashMap();
-        this.stats.set(new ScheduledTaskStatisticsImpl());
+        this.stats = new ScheduledTaskStatisticsImpl();
     }
+
     public ScheduledTaskDescriptor(TaskDefinition definition,
-                                   Map<?, ?> state, ScheduledTaskStatisticsImpl stats) {
+                                   Map<?, ?> state, ScheduledTaskStatisticsImpl stats,
+                                   ScheduledTaskResult result) {
         this.definition = definition;
-        this.stats.set(stats);
+        this.stats = stats;
         this.state = state;
+        this.resultRef.set(result);
     }
 
     public TaskDefinition getDefinition() {
@@ -67,15 +79,19 @@ public class ScheduledTaskDescriptor implements IdentifiedDataSerializable {
     }
 
     ScheduledTaskStatisticsImpl getStatsSnapshot() {
-        return stats.get().snapshot();
-    }
-
-    void setStats(ScheduledTaskStatisticsImpl stats) {
-        this.stats.set(stats);
+        return stats.snapshot();
     }
 
     Map<?, ?> getStateSnapshot() {
-        return new HashMap(state);
+        return state;
+    }
+
+    ScheduledTaskResult getTaskResult() {
+        return resultRef.get();
+    }
+
+    void setStats(ScheduledTaskStatisticsImpl stats) {
+        this.stats = stats;
     }
 
     void setState(Map<?, ?> snapshot) {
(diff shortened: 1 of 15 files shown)


tkountis:
run-lab-run

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
```diff
@@ -90,6 +117,79 @@ void setScheduledFuture(ScheduledFuture<?> future) {
...
+
+    boolean cancel(boolean mayInterrupt)
+            throws ExecutionException, InterruptedException {
+
+        if (!cancelled.compareAndSet(false, true) || scheduledFuture == null) {
+            return false;
+        }
+
+        // Unpeel Executor's Future (in case of Callable, see. DelegatingCallableTaskDecorator)
+        Future delegatedFuture = null;
```
mmedenjak: How about this : 
```
final boolean cancelled = unpeelFutureOrRoot().cancel(mayInterrupt);
        if (AT_FIXED_RATE.equals(definition.getType())) {
            scheduledFuture.cancel(mayInterrupt);
        }
        return cancelled;
```

mmedenjak: Also, this approach has the problem that for a single-run callable to get cancelled, it will first wait for it to get scheduled before cancelling it. Here is one possible approach to fix this :

```
final boolean cancelled = scheduledFuture.cancel(mayInterrupt);
        if (SINGLE_RUN.equals(definition.getType())) {
            try {
                return ((Future) scheduledFuture.get()).cancel(mayInterrupt);
            } catch (CancellationException e) {
                // SCHEDULING WAS ALREADY CANCELLED, ABORT
            }
        }
        return cancelled;
```

mmedenjak:
Also, rebase the commits and try to make shorter the first line of the commit. The details can come in a new paragraph along with bullet points. This is a good guide that I try to adhere to - http://chris.beams.io/posts/git-commit/

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
```diff
@@ -90,6 +117,79 @@ void setScheduledFuture(ScheduledFuture<?> future) {
...
+            throws ExecutionException, InterruptedException {
+
+        boolean finished = this.result.get() != null;
+        boolean cancelled = this.cancelled.get();
+        return finished || cancelled || (scheduledFuture != null && unpeelFutureOrRoot().isDone());
+    }
+
+    boolean shouldSchedule() {
+        // Backup tasks (stashed) that aren't Callable's that finished and successfully synced, neither cancelled on another node.
+        return scheduledFuture == null && this.result.get() == null && !this.cancelled.get();
```
mmedenjak: What if the user scheduled a callable that completed and returned `null`? Does that mean that the callable get rescheduled? Or when the task threw an exception? Related, check other places where `result.get()` is compared to `null`.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
```diff
@@ -90,6 +117,79 @@ void setScheduledFuture(ScheduledFuture<?> future) {
...
     }
 
+    void setResult(Object result) {
+        this.result.set(result);
+    }
+
+    Object getResult()
+            throws ExecutionException, InterruptedException {
+
+        return unpeelFutureOrRoot().get();
```
mmedenjak: Should check if the result has already been calculated. Currently the result is moved to the other replica only to check the task state.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
```diff
@@ -38,11 +44,24 @@
     private ScheduledFuture<?> scheduledFuture;
 
     /**
-     * Stats are written from a single thread and read by many outside the partition threads. (see. Member owned tasks)
+     * Stats are written from a single thread and read by many outside the partition threads.
```
tkountis: I consider the Atomic* API to be cleaner and easier to notice during refactoring, and since it comes _almost_ cost-free (ignoring the ref indirection :) ) I would prefer to keep it this way, unless you have a strong argument against it.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
tkountis: I chose to do my approach to avoid a leaky race-condition between cancelling the task first, only to get it re-scheduled then, before canceling the scheduler one. Not a big deal, just easier to reason about. However, you do have a good point in the second part, which I forgot when I did the refactoring and moved the lines around, I will revert, and handle the exception!!

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
mmedenjak: I don't have a strong argument. The Atomic API might be clean but this was a surprise to me - I see the Atomic field and expect multiple writers only to realise later that that wasn't the case.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
```diff
@@ -53,41 +62,100 @@ public ScheduledTaskDescriptor() {
...
+    }
+
+    void setTaskResult(ScheduledTaskResult result) {
+        this.resultRef.set(result);
+    }
+
+    Object get()
+            throws ExecutionException, InterruptedException {
+
+        ScheduledTaskResult result = resultRef.get();
```
mmedenjak: Should we also check if this we already have a result here in `resultRef` (this was a one-time callable which has migrated to a different node)?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
```diff
@@ -53,41 +62,100 @@ public ScheduledTaskDescriptor() {
...
     }
 
     ScheduledTaskStatisticsImpl getStatsSnapshot() {
-        return stats.get().snapshot();
+        return stats.snapshot();
     }
 
-    void setStats(ScheduledTaskStatisticsImpl stats) {
-        this.stats.set(stats);
+    Map<?, ?> getStateSnapshot() {
```
mmedenjak: I guess this can be renamed to `getState` since we are no longer creating a new map.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskResult.java
```diff
@@ -0,0 +1,127 @@
...
+
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+
+import java.io.IOException;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+
+public class ScheduledTaskResult
```
mmedenjak: Just an idea, you don't need to use it if you don't like it - instead of having a Future and a resultRef in the `ScheduledTaskDescriptor`, maybe on reconstruction we can just put a future which has already completed - something like `java.util.concurrent.CompletableFuture#completedFuture` or like Guava's `Futures.immediate*`.

tkountis: That was the idea originally, the TaskResult was going to be a CompletedFuture and the API would be uniform, but the problem is with serialization, it adds a bit more complexity on handling the default Future which isn't serializable, and the virtual one which would be for our needs.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
tkountis: 👍

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
tkountis: Missed that one, done and added a test for the case.

PR #9581, opened 2017-01-04
Throttling of queries executed from a HZ client. 

### LINKED ISSUE #9568: OOME &  full-gc's in aggregations heavy-load simulator test

tombujok:
The number of thread in the client executor is too big for full-table scan queries. If there's too many of them executed at once the client executor holds to much data and throws OOME eventually.

Fixes #9568

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
@@ -33,6 +33,7 @@ import com.hazelcast.client.impl.protocol.task.AuthenticationMessageTask;
 import com.hazelcast.client.impl.protocol.task.GetPartitionsMessageTask;
 import com.hazelcast.client.impl.protocol.task.MessageTask;
 import com.hazelcast.client.impl.protocol.task.PingMessageTask;
+import com.hazelcast.client.impl.protocol.task.map.AbstractMapQueryMessageTask;
 import com.hazelcast.config.Config;
 import com.hazelcast.core.Client;
 import com.hazelcast.core.ClientListener;
@@ -104,6 +105,7 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
 
     private static final int EXECUTOR_QUEUE_CAPACITY_PER_CORE = 100000;
     private static final int THREADS_PER_CORE = 20;
+    private static final int QUERY_THREADS_PER_CORE = 1;
     private static final ConstructorFunction<String, AtomicLong> LAST_AUTH_CORRELATION_ID_CONSTRUCTOR_FUNC =
             new ConstructorFunction<String, AtomicLong>() {
                 @Override
@@ -114,6 +116,7 @@ public class ClientEngineImpl implements ClientEngine, CoreService, PostJoinAwar
     private final Node node;
     private final NodeEngineImpl nodeEngine;
     private final Executor executor;
+    private final Executor queryExecutor;
 
     private final SerializationService serializationService;
     // client uuid -> member uuid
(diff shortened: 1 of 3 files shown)


tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -163,6 +167,22 @@ private Executor newExecutor() {
                 ExecutorType.CONCRETE);
     }
 
+    private Executor newClientQueryExecutor() {
```
pveentjer: Why is a new threadpool introduced? Can't it be processed by the generic threadpool?

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanRunner.java
```diff
@@ -72,7 +72,7 @@ public PartitionScanRunner(MapServiceContext mapServiceContext) {
     @SuppressWarnings("unchecked")
     public Collection<QueryableEntry> run(String mapName, Predicate predicate, int partitionId) {
         PagingPredicate pagingPredicate = predicate instanceof PagingPredicate ? (PagingPredicate) predicate : null;
-        List<QueryableEntry> resultList = new LinkedList<QueryableEntry>();
+        List<QueryableEntry> resultList = new ArrayList<QueryableEntry>();
```
pveentjer: Why change to an ArrayList? Perhaps there is a good reason for it.

tombujok: It was just a test.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
tombujok: We figured out with @jerrinot that it may lead to deadlocks since the node may execute operations on other nodes.

tombujok:
@sancar @pveentjer @jerrinot Could you review this PR? Thx!

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -149,20 +153,36 @@ private ClientExceptionFactory initClientExceptionFactory() {
...
-    private Executor newExecutor() {
+    private Executor newClientExecutor() {
         final ExecutionService executionService = nodeEngine.getExecutionService();
         int coreSize = Runtime.getRuntime().availableProcessors();
 
         int threadCount = node.getProperties().getInteger(GroupProperty.CLIENT_ENGINE_THREAD_COUNT);
         if (threadCount <= 0) {
             threadCount = coreSize * THREADS_PER_CORE;
         }
+        logger.info("Creating new client executor with threadCount=" + threadCount);
```
jerrinot: I'd move this to `fine` or even `finest`.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -149,20 +153,36 @@ private ClientExceptionFactory initClientExceptionFactory() {
...
 
+    private Executor newClientQueryExecutor() {
+        final ExecutionService executionService = nodeEngine.getExecutionService();
+        int coreSize = Runtime.getRuntime().availableProcessors();
+
+        int threadCount = node.getProperties().getInteger(GroupProperty.CLIENT_ENGINE_QUERY_THREAD_COUNT);
+        if (threadCount <= 0) {
+            threadCount = coreSize * QUERY_THREADS_PER_CORE;
+        }
+        logger.info("Creating new client query executor with threadCount=" + threadCount);
```
jerrinot: same here. this is just noisy during a normal usage.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
tombujok: Hmm, really? It's just two lines on boot, not per query or sth. 
To me If I set the number of threads I want to be sure that the option is picked up and I don't have a typo or sth.
WDYT?

jerrinot: I prefer to be less noisy by default. We print thread count just for operational threads.

tombujok: Done

PR #9589, opened 2017-01-05
added jetVersion to BuildInfo

gurbuzali:


hazelcast/src/main/java/com/hazelcast/instance/BuildInfo.java
```diff
@@ -69,6 +76,18 @@ public byte getSerializationVersion() {
         return serializationVersion;
     }
 
+    public String getJetVersion() {
```
pveentjer: Can these values ever be null? What if someone isn't using JET?

gurbuzali: These values will be null if someone is not using JET

pveentjer: Could you add this to the javadoc?

pveentjer:
Apart from the javadoc and not yet having a green build; 👍

hazelcast/src/main/java/com/hazelcast/instance/BuildInfo.java
```diff
@@ -35,14 +35,21 @@
     private final boolean enterprise;
     private final byte serializationVersion;
 
+    private final String jetVersion;
+    private final String jetBuild;
+    private final String jetRevision;
```
ahmetmircik: adding these fields to `toString` method can be helpful for diagnostic purposes.

pveentjer: Good spotting @ahmetmircik

hazelcast/src/main/java/com/hazelcast/util/PhoneHome.java
```diff
@@ -180,6 +182,18 @@ public String convertToLetter(int size) {
         return parameterCreator.getParameters();
     }
 
+    private void addJetParameters(PhoneHomeParameterCreator parameterCreator, BuildInfo buildInfo) {
+        if (buildInfo.getJetVersion() == null) {
+            parameterCreator.addParam("jetv", "");
+            parameterCreator.addParam("jetb", "");
+            parameterCreator.addParam("jetr", "");
```
hasancelik: Could you update [PhoneHomeTest](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/util/PhoneHomeTest.java#L44) according to these changes? Default values are enough to test.

PR #9591, opened 2017-01-05
Introduces test for Map merge after split brain

vbekiaris:
Adds a test to verify that `MapMergePolicy.merge` is invoked for all entries being merged after split-brain is healed (related to #9358).

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/MapSplitBrainTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/MapSplitBrainTest.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/map/impl/MapSplitBrainTest.java
```diff
@@ -0,0 +1,149 @@
...
+ * Given:
+ *   3-members cluster, map configured with 2 backups (so data are complete on each member after split-brain)
+ *   and custom merge policy that counts number of merge invocations
+ * When:
+ *   cluster splits in two subclusters with {1, 2} members respectively, then merges
+ * Then:
+ *   custom merge policy's merge method is invoked for all entries of the map
+ */
+@RunWith(HazelcastSerialClassRunner.class)
+@Category({QuickTest.class})
```
jerrinot: It should have the `ParallelTest` category. 
It does not use a real network -> it can be executed in parallel with other tests.

jerrinot:
the test is checking a merge policy was invoked. However there is no check the value returned by the policy was actually applied -> It's still possible to have a broken merge policy and this test would pass.

I'd suggest to add a test checking effects of the merge policy instead of checking whether it was invoked. 

I can imagine this:
1. have `AddingMergePolicy`
2. the first brain add entries with keys `0...1999` and value `1`
3. the 2nd brain add entries with keys `1000.2999` andwith value `2`
3. heal the split-brain
4. assert
- keys `0...999` have value `1`
- keys `1000...1999` have values `3`
- keys `2000....2999` have value `2`

Or something along those lines.

vbekiaris:
Good spotting, I initially thought merging functionality was already tested in `MapMergePolicyQuickTest` and `com.hazelcast.map.merge.*` tests but these do not actually simulate a split brain, rather they work at the `RecordStore` level or unit test the `MapMergePolicy` implementation itself. I will improve the test as suggested.

vbekiaris:
thanks for the quick review @jerrinot @ahmetmircik

PR #9593, opened 2017-01-06
[back-port]removed Lifecylelistener from mancenter service and fixed ManagementC…

hasancelik:
…nterServiceIntegrationTest.testTimedMemberStateNotNull

back port of #9571

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
@@ -122,8 +122,8 @@ public class ManagementCenterService {
         this.timedMemberStateFactory = new TimedMemberStateFactory(instance);
         this.identifier = newManagementCenterIdentifier();
 
-        if (this.managementCenterConfig.isEnabled()) {
-            this.instance.getCluster().addMembershipListener(new ManagementCenterService.MemberListenerImpl());
+        if (managementCenterConfig.isEnabled()) {
+            this.instance.getCluster().addMembershipListener(new MemberListenerImpl());
             start();
         }
     }
--- a/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
@@ -20,9 +20,6 @@ import com.eclipsesource.json.JsonObject;
 import com.eclipsesource.json.JsonValue;
 import com.hazelcast.config.GroupConfig;
 import com.hazelcast.config.ManagementCenterConfig;
-import com.hazelcast.core.LifecycleEvent;
-import com.hazelcast.core.LifecycleEvent.LifecycleState;
-import com.hazelcast.core.LifecycleListener;
 import com.hazelcast.core.Member;
 import com.hazelcast.core.MemberAttributeEvent;
 import com.hazelcast.core.MembershipEvent;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
```diff
@@ -125,21 +122,16 @@ public ManagementCenterService(HazelcastInstanceImpl instance) {
         this.timedMemberStateFactory = new TimedMemberStateFactory(instance);
         this.identifier = newManagementCenterIdentifier();
 
-        registerListeners();
+        if (this.managementCenterConfig.isEnabled()) {
```
pveentjer: 'this.' isn't needed. Only when it is ambiguous, e.g. a setter/constructor, this is recommended.

hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
```diff
@@ -125,21 +122,16 @@ public ManagementCenterService(HazelcastInstanceImpl instance) {
...
+            this.instance.getCluster().addMembershipListener(new ManagementCenterService.MemberListenerImpl());
+            start();
+        }
     }
 
     private String getManagementCenterUrl() {
         return managementCenterConfig.getUrl();
     }
```
pveentjer: Why did you inline the function?

hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
```diff
@@ -125,21 +122,16 @@ public ManagementCenterService(HazelcastInstanceImpl instance) {
         this.timedMemberStateFactory = new TimedMemberStateFactory(instance);
         this.identifier = newManagementCenterIdentifier();
 
-        registerListeners();
+        if (this.managementCenterConfig.isEnabled()) {
+            this.instance.getCluster().addMembershipListener(new ManagementCenterService.MemberListenerImpl());
```
pveentjer: You don't need to add the prefix 'ManagementCenterService.' since you are already in this class.

hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
```diff
@@ -125,21 +122,16 @@ public ManagementCenterService(HazelcastInstanceImpl instance) {
...
+            start();
```
pveentjer: So if I understand it correctly:

old: start mancenterservice once the HZ instance is started.

new: start mancenterservice before/during HZ instance started.

Can't this cause problems? If the mancenter is being exposed before the HZ instance is fully started; can't this lead to running into partially initialised data-structures?

hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
hasancelik: I failed to notice while copy-pasting the code  👍

hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
hasancelik: same copy paste problem 😃

hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
hasancelik: Actually start() method triggers all thread that are getting statistics and member state. And [TimedMemberStateFactory](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/management/TimedMemberStateFactory.java#L105) checks member state.

I deleted lifecycle listener and moved start() method into constructor because when we trigger partial or force start via rest call to man center, node can not start man center due to this listener. When partial start is proceeding, node state is STARTING not STARTED.

hasancelik:
run-lab-run

PR #9600, opened 2017-01-07
Fix 9599 Map.asyncSet return type problems

### LINKED ISSUE #9599: imap.setAsync throws an erroneous java.lang.ClassCastException inside the returned iCompletableFuture
We are using Hazeclast version 3.7.4. 
Java(TM) SE Runtime Environment (build 1.8.0_60-b27)

We use Hazelcast with Scala, but I don't believe the issue has anything to do with that.  Here is a simple bit of code to replicate the problem. 


```scala
import com.hazelcast.core.Hazelcast
import com.hazelcast.core.HazelcastInstance
import com.hazelcast.config._
import com.hazelcast.core._

object HazelcastTest {
  
  val hz = Hazelcast.newHazelcastInstance()
  val hazelcastMap: com.hazelcast.core.IMap[String, Int] = hz.getMap("mymap")
  val f: ICompletableFuture[Void] = hazelcastMap.setAsync("a", 5)
   //throws an Exception
  f.andThen(new ExecutionCallback[Void]() { override def onResponse(a: Void): Unit = println("HELLO WORLD"); 
  override def onFailure(t: Throwable): Unit = println("failure = " + t);
  }) 

  val fp: ICompletableFuture[Int] = hazelcastMap.putAsync("a", 5)
  //works as expected
  fp.andThen(new ExecutionCallback[Int]() { override def onResponse(a: Int): Unit = println("HELLO WORLD i =" + a); 
  override def onFailure(t: Throwable): Unit = println("failure = " + t);
  })
}
```

setAsync's onResponse is called but before the callback can be run the following Exception is thrown:

```
java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.Void                                                                                                                                                                                                              
at $sess.cmd9$$anon$1.onResponse(cmd9.sc:1)                                                                                                                                                                                                                                           
at com.hazelcast.util.executor.DelegatingFuture$DelegatingExecutionCallback.onResponse(DelegatingFuture.java:172)                                                                                                                                                                     
at com.hazelcast.spi.impl.AbstractInvocationFuture$1.run(AbstractInvocationFuture.java:251)                                                                                                                                                                                           
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)                                                                                                                                                                                                    
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)                                                                                                                                                                                                    at java.lang.Thread.run(Thread.java:745)                                                                                                                                                                                                                                              
at com.hazelcast.util.executor.HazelcastManagedThread.executeRun(HazelcastManagedThread.java:76)                                                                                                                                                                                      
at com.hazelcast.util.executor.HazelcastManagedThread.run(HazelcastManagedThread.java:92)         
```

However, the set did succeed and I can see my values inside the map.

pveentjer:
Fix #9599

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/SetOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/SetOperation.java
@@ -46,11 +46,6 @@ public class SetOperation extends BasePutOperation {
     }
 
     @Override
-    public Object getResponse() {
-        return newRecord;
-    }
-
-    @Override
     public int getId() {
         return MapDataSerializerHook.SET;
     }
--- a/hazelcast/src/test/java/com/hazelcast/map/AsyncTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/AsyncTest.java
@@ -105,6 +105,16 @@ public class AsyncTest extends HazelcastTestSupport {
     }
 
     @Test
+    public void testSetAsync_issue9599() throws Exception {
+        IMap<String, String> map = instance.getMap(randomString());
+        Future<Void> f = map.setAsync(key, value1);
+
+        // the return value was not of type Void, but Boolean. So assignment to Void would fail.
+        Void v = f.get();
+        assertNull(v);
+    }
+
+    @Test
     public void testSetAsyncWithTtl() throws Exception {
         IMap<String, String> map = instance.getMap(randomString());
 


pveentjer:
run-lab-run

PR #9601, opened 2017-01-08
WIP Further fix related to #9273 and #9271

sancar:
Related pull requests

In #9271, connections with resources are destroyed. I have converted
it so that every endpoint is removed. The purpose of heartbeat
was to clean resources of a client when heartbeat is timeout in the
first place.

With the fix above, the addition of findLiveConnection is not necessary 
any more. . When endpoint destroyed listeners
are removed. This means there can not be any listener task that its connection
is not live but it is trying to send events to client.


sancar:
run-lab-run

PR #9604, opened 2017-01-08
Fix wrong estimation due to wrong encoding engine.

tkountis:
When using an estimator to constantly add new elements and not checking the estimate
then encoding engine is not upgraded to the Dense, which leads to two problems.
a. the default engine keeps consuming more and more resources and it is fundamentally very hungry and slow
b. upon calling estimate at the very end, the conversion gives very wrong results

The problem with the delayed conversion, between the encoders, is that the actual condition check, whether to upgrade, takes place during the `estimate()` call, rather than the `add()`.

This patch, moves the check to the `add()` instead.

Credit goes to @pveentjer for reporting the issue.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/cardinality/CardinalityEstimatorStressTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/cardinality/CardinalityEstimatorStressTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pveentjer:
Can you also add a reproducer to the test code that failed before the change and passed after the change is applied?

I'll try it in with the simulator test.

pveentjer:
I just verified this fix with Simulator and it seems to be running fine. The error is small and the throughput remains stable.

pveentjer:
👍 

Thanks for the quick fix!

hazelcast/src/test/java/com/hazelcast/cardinality/CardinalityEstimatorStressTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(HazelcastSerialClassRunner.class)
+@Category(NightlyTest.class)
+public class CardinalityEstimatorStressTest extends HazelcastTestSupport {
+
```
tombujok: Nice test!

PR #9606, opened 2017-01-09
Partition initialization fixes

mdogan:
- Partitions should not be initialized until startup is completed
- Add a new variant of getMemberPartitions() which is not triggering initialization


hasancelik:
run-lab-run

PR #9608, opened 2017-01-09
Fixes ClientDisconnectionOperation removes the client endpoint twice

sancar:
ClientDisconnectionOperation removes the client endpoint twice when
client reconnected to same node. Reason was only check we had was client
and member uuid not changed. In this scenario, both of them do not
change. We are addinitially checking last authentication correlation
id of client to prevent second endpoint destroy.

A group property is added to make similar tests to take less time.
"hazelcast.client.endpoint.remove.delay.seconds". This value was
hardcoded as 10 seconds and causing all related tests to wait for
at least 10 seconds.


sancar:
run-lab-run

PR #9612, opened 2017-01-09
Use different property file for Hazelcast Enterprise build info

jerrinot:
When hazelcast-enterprise-runtime.properties are found then
we treat hazelcast-runtime.properties as an upstream source.

Hazelcast Enterprise can see what (Open Source) Hazelcast
was used  to build that particular version of Enterprise.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/BuildInfo.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/BuildInfo.java
@@ -34,15 +34,22 @@ public class BuildInfo {
     private final int buildNumber;
     private final boolean enterprise;
     private final byte serializationVersion;
+    private final BuildInfo upstreamBuildInfo;
 
     public BuildInfo(String version, String build, String revision, int buildNumber, boolean enterprise,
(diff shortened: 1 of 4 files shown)


mesutcelik:
👍

hazelcast/src/main/java/com/hazelcast/instance/BuildInfo.java
```diff
@@ -105,6 +116,7 @@ public String toString() {
                 + ", revision=" + revision
                 + ", enterprise=" + enterprise
                 + ", serializationVersion=" + serializationVersion
+                + (upstreamBuildInfo == null ? "" : ", upstream=" + serializationVersion)
```
vbekiaris: Accidentally writing `serializationVersion` instead of `upstreamBuildInfo` ?

jerrinot: good one, thanks!

jerrinot:
@mesutcelik, @gurbuzali, @vbekiaris: thanks for quick reviews!

PR #9616, opened 2017-01-10
Increased code coverage of HazelcastManifestTransformer

Donnerbart:
A first bunch of tests for the `HazelcastManifestTransformer` and its inner classes.

I'm a bit worried since the `HazelcastManifestTransformer` seems to use a JDK 7 method:
```
jarOutputStream.flush();
```

I don't know if this should break our builds if somewhere if JDK 6 is used.

Is there also a reason why `edu.emory.mathcs.backport.java.util` is used instead of the normal `java.util` package? This leads to some generic errors, since those `Arrays.asList()` and `Collections.emptySet()` are using raw types.


hazelcast-build-utils/src/main/java/com/hazelcast/buildutils/HazelcastManifestTransformer.java
```diff
@@ -203,8 +195,7 @@ public void modifyOutputStream(JarOutputStream jarOutputStream)
         jarOutputStream.flush();
```
Donnerbart: Can we use `IOUtil.close(jarOutputStream)` instead?

Donnerbart:
I just made a test by throwing an exception in `HazelcastManifestTransformer.modifyOutputStream()`, to see if it's used in a regular `mvn clean install`. And it seems it is!
```
Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:2.2:shade (shade-artifacts) on project hazelcast-all: Error creating shaded jar: boooooooooooooooooooooooooooooooooooooooooooooooooooooooom -> [Help 1]
```
So I'm really asking myself why we don't have issues with the `jarOutputStream.flush()` in our JDK 6 builds. Nevertheless we should fix this to get rid of this:
screenshot from 2017-01-11 11-46-13

Donnerbart:
Okay, it looks like that the `flush()` method has been there in JDK 6 as well. Maybe there are some exotic JDKs which are missing it before Java 7, but this explains why we don't see any failures in our builds.

I suppressed the error, so we don't have a red marker in IDEA.

Donnerbart:
run-lab-run

noctarius:
`edu.emory.mathcs.backport.java.util` is probably used for no specific reason. The original code was copied, so I guess it was already used somewhere else. Feel free to test the original `java.util` methods however these backports often add features which don't exist in Java 6, so not sure about this myself.

`flush()` should be around forever and I'm pretty sure it wasn't added in Java 7, no idea where taht message comes from:

Copyied from OutputStream.java:
```
    /**
     * Flushes this output stream and forces any buffered output bytes
     * to be written out. The general contract of <code>flush</code> is
     * that calling it is an indication that, if any bytes previously
     * written have been buffered by the implementation of the output
     * stream, such bytes should immediately be written to their
     * intended destination.
     * <p>
     * If the intended destination of this stream is an abstraction provided by
     * the underlying operating system, for example a file, then flushing the
     * stream guarantees only that bytes previously written to the stream are
     * passed to the operating system for writing; it does not guarantee that
     * they are actually written to a physical device such as a disk drive.
     * <p>
     * The <code>flush</code> method of <code>OutputStream</code> does nothing.
     *
     * @exception  IOException  if an I/O error occurs.
     */
    public void flush() throws IOException {
    }
```

No marker about @since java7

PR #9618, opened 2017-01-10
WIP ClientDisconnectionOperation removes the client endpoint twice

sancar:
ClientDisconnectionOperation removes the client endpoint twice when client reconnected to same node. Reason was only check we had was client
and member uuid not changed. In this scenario, both of them do not
change. We are addinitially checking last authentication correlation
id of client to prevent second endpoint destroy.

A group property is added to make similar tests to take less time.
"hazelcast.client.endpoint.remove.delay.seconds". This value was
hardcoded as 10 seconds and causing all related tests to wait for
at least 10 seconds.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
@@ -181,7 +181,8 @@ public abstract class AuthenticationBaseMessageTask<P> extends AbstractCallableM
         Connection connection = endpoint.getConnection();
         ILogger logger = clientEngine.getLogger(getClass());
 
-        endpoint.authenticated(principal, credentials, isOwnerConnection(), clientVersion);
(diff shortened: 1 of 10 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/AuthenticationBaseMessageTask.java
```diff
@@ -181,7 +181,8 @@ private ClientMessage handleAuthenticated() {
         Connection connection = endpoint.getConnection();
         ILogger logger = clientEngine.getLogger(getClass());
 
-        endpoint.authenticated(principal, credentials, isOwnerConnection(), clientVersion);
+        clientEngine.trySetLastAuthenticationCorrelationId(principal.getUuid(), clientMessage.getCorrelationId());
```
ihsandemir: Should we check the boolean result of this call and throw exception? (similar as done at master)

sancar: That was a fix for another issue. We can not backport it because we can not change operation in maintenance. Throwing exception here does not fix the issue by itself.

PR #9619, opened 2017-01-10
Fixing multiple bugs in ResultSets; Increasing test coverage

### LINKED ISSUE #9614: size() doesn't match the actual element number of Set<QueryableEntry>
Set<QueryableEntry> set = ((IndexAwarePredicate) predicate).filter(queryContext);
int size = set.size();
    
int act_size = 0;
for(QueryableEntry queryableEntry : set) {
     act_size++;
}

assert size==act_size;    =====> **Here will throws AssertionError**
__________________________________________________________________________________
The issue is in com.hazelcast.query.impl.AndResultSet, size() will return the size of setSmallest,
but iterator will return the intersection size of setSmallest,otherIndexedResults,lsNoIndexPredicates.
So the inconsistence occurs.  

See the following code:

  ```
public int size() {
     return setSmallest.size();
  }
    
  public boolean hasNext() {
		if (currentEntry != null) {
			return true;
		}

		for (; it.hasNext();) {
			QueryableEntry entry = it.next();

			if (checkOtherIndexedResults(entry) && checkNoIndexPredicates(entry)) {
				currentEntry = entry;
				return true;
			}
		}

		return false;
}  
```

### PR DESCRIPTION
**Pull request by tombujok**
Fixes #9614 (among other things)

### CODE REVIEW

File: hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java [hunk 1]
```diff
@@ -161,6 +150,14 @@ public void remove() {
 
     @Override
     public int size() {
-        return setSmallest.size();
+        if (size == -1) {
```
jerrinot: this could be problematic performance-wise. as it's used during evaluation. I see you have some sort of caching, but I am not sure about possible consequences. what's the worst (performance) case you can imagine here?

hazelcast/src/test/java/com/hazelcast/query/impl/OrResultSetTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+        OrResultSet resultSet = new OrResultSet(indexedResults);
+
+        int sizeMethod = resultSet.size();
+        int sizeIterator = 0;
+        for (QueryableEntry queryableEntry : resultSet) {
+            sizeIterator++;
+        }
+
+        assertEquals(sizeIterator, 2 * size);
```
Donnerbart: I think this should be `assertEquals(2 * size, sizeIterator);`, since the order is `expected` then `actual` value. Same for the next assertion.

hazelcast/src/test/java/com/hazelcast/query/impl/OrResultSetTest.java
```diff
@@ -0,0 +1,109 @@
...
+        public Data getKeyData() {
+            return null;
+        }
+
+        @Override
+        public Data getValueData() {
+            return null;
+        }
+    }
+
```
Donnerbart: Unnecessary newline here, but there is a newline missing at the end of the file :)

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -103,7 +92,7 @@ public boolean hasNext() {
                 return true;
             }
 
-            for (; it.hasNext();) {
+            for (; it.hasNext(); ) {
```
tombujok: True

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -161,6 +150,14 @@ public void remove() {
...
+        if (size == -1) {
+            int calculatedSize = 0;
+            for (Iterator<QueryableEntry> it = iterator(); it.hasNext(); it.next()) {
+                calculatedSize++;
+            }
+            size = calculatedSize;
+        }
+        return size;
     }
+
```
tombujok: I like it so :)

hazelcast/src/test/java/com/hazelcast/query/impl/OrResultSetTest.java
tombujok: True, fixed

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
tombujok: It's just iterating over the set once, then it's cached. The only place where it's used instantiated is here:
https://github.com/tombujok/hazelcast/blob/c2472d25552d6ad830858365cc142ef73b4c4a41/hazelcast/src/main/java/com/hazelcast/query/impl/predicates/AndPredicate.java#L99-L99

What do you mean precisely when you say that it's used during evaluation.

tombujok: You were right. I've identified edge cases where the overhead of the size() method was significant (i.e. slowing down the query 5x times). 
Fixed.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -161,6 +150,28 @@ public void remove() {
...
+            }
+            size = calculatedSize;
+        }
+        return size;
+    }
+
+    /**
+     * @return returns estimated size without calculating the full result set in full-result scan.
+     */
+    public int estimatedSize() {
```
jerrinot: does it have to be public?

tombujok: Yes, it unfortunately does, but the users get the result as the top-level Set interface

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/AndPredicate.java
```diff
@@ -99,6 +100,16 @@ public Predicate accept(Visitor visitor, Indexes indexes) {
         return new AndResultSet(smallestIndexedResult, otherIndexedResults, lsNoIndexPredicates);
     }
 
+    private int size(Set<QueryableEntry> result) {
+        // In case of AndResultSet and OrResultSet calling size() may be very expensive so quickerEstimated() size is used
```
jerrinot: `quickerEstimated()` does not exist.

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/AndPredicate.java
```diff
@@ -99,6 +100,16 @@ public Predicate accept(Visitor visitor, Indexes indexes) {
...
+        if (result instanceof AndResultSet) {
```
jerrinot: the question is whether to create an interface `WithSizeEstimator` or something. probably an overkill.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -28,37 +28,26 @@
...
     private final List<Set<QueryableEntry>> otherIndexedResults;
     private final List<Predicate> lsNoIndexPredicates;
+    private int size;
 
     public AndResultSet(Set<QueryableEntry> setSmallest, List<Set<QueryableEntry>> otherIndexedResults,
                         List<Predicate> lsNoIndexPredicates) {
         this.setSmallest = isNotNull(setSmallest, "setSmallest");
         this.otherIndexedResults = otherIndexedResults;
         this.lsNoIndexPredicates = lsNoIndexPredicates;
+        this.size = -1;
```
jerrinot: `cachedSize` ?

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/AndPredicate.java
tombujok: I was thinking about it, but kind of it's just two classes. But thanks for the hint.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
tombujok: Fixed

hazelcast/src/main/java/com/hazelcast/query/impl/predicates/AndPredicate.java
tombujok: Fixed.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -28,37 +28,26 @@
...
     }
 
     public byte[] toByteArray(ObjectDataOutput out) throws IOException {
```
mdogan: I can't see where this `toByteArray()` method is used.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -28,37 +28,26 @@
...
 
     public AndResultSet(Set<QueryableEntry> setSmallest, List<Set<QueryableEntry>> otherIndexedResults,
                         List<Predicate> lsNoIndexPredicates) {
         this.setSmallest = isNotNull(setSmallest, "setSmallest");
         this.otherIndexedResults = otherIndexedResults;
         this.lsNoIndexPredicates = lsNoIndexPredicates;
+        this.cachedSize = -1;
     }
 
     public byte[] toByteArray(ObjectDataOutput out) throws IOException {
```
ahmetmircik: what was the need for this `toByteArray` method?

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -28,37 +28,26 @@
...
     private final List<Set<QueryableEntry>> otherIndexedResults;
     private final List<Predicate> lsNoIndexPredicates;
+    private int cachedSize;
 
     public AndResultSet(Set<QueryableEntry> setSmallest, List<Set<QueryableEntry>> otherIndexedResults,
                         List<Predicate> lsNoIndexPredicates) {
         this.setSmallest = isNotNull(setSmallest, "setSmallest");
         this.otherIndexedResults = otherIndexedResults;
         this.lsNoIndexPredicates = lsNoIndexPredicates;
+        this.cachedSize = -1;
```
ahmetmircik: using a static final variable instead of `-1` can increase readability.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
tombujok: That's a good hint. First I thought it overrides a method in higher-levelled interface. Now I see it's not used.
Should we delete it? Wdyt also @jerrinot @mdogan?

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
tombujok: Thx. Fixed.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
jerrinot: I reckon it was originally meant to be used for serialization? I can see the method was removed from `OrResultSet` [quite some time ago](https://github.com/jerrinot/hazelcast/commit/ee3455b2c8c256df86b2917ef8e7af0c6292df5b#diff-773b532d77d82620bf8102c731d1f76bL31).

I would just remove it from `AndResultSet` (In a PR into master only)

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
```diff
@@ -161,6 +153,28 @@ public void remove() {
...
+            }
+            cachedSize = calculatedSize;
+        }
+        return cachedSize;
     }
+
+    /**
+     * @return returns estimated size without calculating the full result set in full-result scan.
+     */
+    public int estimatedSize() {
```
ahmetmircik: if `estimatedSize()` is called on this result set before `size() `,  is there a possibility that returned value from `size()` is wrong because of the _size-caching_?

tombujok: Good question, but the answer is `no` since the `estimatedSize()` won't cache anything.
It uses the `cachedSize` if it is initialized (since it delivers the most accurate result), but if there's no `cachedSize` it will return an estimated size without caching it anywhere.

tombujok:
@ahmetmircik toByteArray removed.

hazelcast/src/main/java/com/hazelcast/query/impl/AndResultSet.java
ahmetmircik: i see now, thanks

PR #9597, opened 2017-01-06
Sizing hash collections take 3

bokken:
This is a replacement for https://github.com/hazelcast/hazelcast/pull/9582 which finally has white space issues and import order resolved.

Fixes : https://github.com/hazelcast/hazelcast/issues/9445

PR #9620, opened 2017-01-10
Add option for InMemoryFormat of object to skip defensive copies.

bokken:
Allow skipping defensive copies when retrieving from an IMap and the InMemoryFormat.OBJECT is being used.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapEntries.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapEntries.java
@@ -35,31 +35,139 @@ import java.util.Map;
  */
 public final class MapEntries implements IdentifiedDataSerializable {
 
+    private static interface ValueWrapper {
+        public Object getValue();
+
+        public Data getValueAsData();
+    }
+
+    private static final ValueWrapper NULL_VALUE_WRAPPER = new ValueWrapper() {
+
+        @Override
+        public Data getValueAsData() {
+            return null;
+        }
+
+        @Override
+        public Object getValue() {
+            return null;
+        }
+    };
+
+    private static final class DataValueWrapper implements ValueWrapper {
+        private final Data value;
+
+        /**
+         * @param value
+         */
+        private DataValueWrapper(Data value) {
+            this.value = value;
+        }
+
+        public static ValueWrapper forData(Data value) {
+            return value != null ? new DataValueWrapper(value) : NULL_VALUE_WRAPPER;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Object getValue() {
+            return value;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Data getValueAsData() {
+            return value;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public String toString() {
+            return "DataValueWrapper [value=" + this.value + "]";
+        }
+    }
+
+    private static final class SerializableObjectValueWrapper implements ValueWrapper {
+        private final Object value;
+        private final SerializationService serializationService;
+
+        /**
+         * @param value
+         * @param serializationService
+         */
+        private SerializableObjectValueWrapper(Object value, SerializationService serializationService) {
+            super();
+            this.value = value;
+            this.serializationService = serializationService;
+        }
+
+        public static ValueWrapper forValue(Object value, SerializationService serializationService) {
+            return value != null ? new SerializableObjectValueWrapper(value, serializationService) : NULL_VALUE_WRAPPER;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Object getValue() {
+            return value;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public Data getValueAsData() {
+            return serializationService.toData(value);
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public String toString() {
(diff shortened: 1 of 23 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/MapEntries.java
```diff
@@ -35,31 +35,139 @@
  */
 public final class MapEntries implements IdentifiedDataSerializable {
 
+    private static interface ValueWrapper {
```
pveentjer: Having these wrappers is not the way forward. 

The problem is that it will:
- increase the amount of memory used per map entry since an additional object + object reference.
- it will making walking through memory more complex due to the extra indirection
- it will put more pressure on the gc in case of an update due to additional litter.

bokken: What about changing the wrapper to also include the key? This would allow collapsing the keys and values lists into a single list of entries. It also makes the putAllTo* methods more efficient as only a single Iterator needs to be created/managed.
Do you have some other specific change in mind?
Do you have benchmarks to measure the impacts you are concerned about?

My guess is that changes to initialize collections to correct sizes, such as done https://github.com/hazelcast/hazelcast/pull/9597 and in this PR reduce more object creation than what is introduced here.

I have added changes to use a single Entry class in MapEntries, which avoids polymorphism.

LoneRifle: I, like @pveentjer , don't think that we should be using wrappers here; these don't appear to have any performance impact when you're dealing with small numbers of entries, but once you get up to thousands or millions (which is what hazelcast is aiming for), these will present a problem. As it is, you are adding one more reference per entry held.

Iterator overhead is trivial relative to per-entry overhead, particularly when the number of entries is large.

In any case, the goal of your PR is to add an option to skip defensive copies, and I struggle to see how this particular change is relevant.

LoneRifle: I, like @pveentjer , don't think that we should be using wrappers here; these don't appear to have any performance impact when you're dealing with small numbers of entries, but once you get up to thousands or millions (which is what hazelcast is aiming for), these will present a problem. As it is, you are adding one more reference per entry held.

Iterator overhead is trivial relative to per-entry overhead, particularly when the number of entries is large.

In any case, the goal of your PR is to add an option to skip defensive copies, and I struggle to see how this particular change is relevant.

bokken: @LoneRifle, the changes here are required to avoid creating defensive copies as a result of a getAll method call. The current implementation of MapEntries requires the value to be represented as Data, which forces a defensive copy to populate MapEntries. The proposed changes defer the serialization to Data until necessary and leaves as Object as long as possible.

Do you have an alternative suggestion on how to avoid defensive copies on getAll operations?

Do you have benchmarks to measure the impacts you are concerned about?

LoneRifle: I leave it to @pveentjer to provide the benchmarks, given the hazelcast team would have seen far more where this is a problem, relative to my own experiences as a user. 

I suggest as an alternative retaining two `Lists` within MapEntries, but changing the type param of values to Object, then adjusting the rest of the PR as needed. I realize that this isn't as clean as you would like, but this may well be the compromise that would _both_ avoid GC litter _and_ defer serialization.

bokken:
@pveentjer and @LoneRifle, I have updated the PR to no longer have a wrapper. I left only 1 list with interleaved key/value pairs as this appeared to offer some potential advantages for MapProxySupport's internal support for MapProxyImpl.executeOnEntries.

LoneRifle:
What shall we do with this PR? If it has been abandoned, I would be happy to have a look and rebase/rewrite parts of it, giving credit where due.

bokken:
It is not my intention to abandon, but I do not currently have the time to constantly keep up to date.
If you have time/desire to update and incorporate that would be great.
If there is interest in getting this included, I can update in the next week or so.

Holmistr:
@LoneRifle @bokken Do you think you will be picking up the work?

LoneRifle:
It's been a while since I've worked on hazelcast, but if memory serves, the one remaining thing that held this back was a massive refactor of the class hierarchy. I wouldn't mind putting together a PR that supersedes this one which embodies the gist of the desired changes

Holmistr:
We should probably first agree that it's the way to go. @mmedenjak or @pveentjer, could you please take a look at this approach?

Holmistr:
I'm going to close this PR, since there's no activity on it.
@bokken @LoneRifle  I'm sorry it got abandoned. Please, feel free to comment if you're willing to pick it up again. We appreciate your effort to contribute! :)

mmedenjak:
Also, because of some other reasons, we will try to introduce this gradually in 4.1 and onwards, using a marker interface instead of configuration.

bokken:
@mmedenjak was something similar to this included in 4.1? Or is there a similar change in process?

mmedenjak:
@bokken interesting that you ask. Actually, I'm assigned to work on this in 4.2. On the other hand, I've been involved in a bunch of other stuff so I hope to be able to finish it. But this time, our PM scheduled it as an addition to IMDG and I'm being held accountable so I guess the pressure is on :) 

Here's an initial design document about how I wanted to implement it - https://github.com/mmedenjak/hazelcast/blob/4.2-design-documents/docs/design/performance/01-avoid-local-copy-on-read-for-immutable-objects.md

bokken:
@mmedenjak let me know if there are ways I can help move this along.

PR #9622, opened 2017-01-11
Moved eviction calls to recordStore

ahmetmircik:
This refactoring is helpful to control/see all eviction calls from one place. And will be used for later fixes.


jerrinot:
run-lab-run

jerrinot:
there is a suspicious test failure:
image

ahmetmircik:
will be handled later with a different PR

PR #9630, opened 2017-01-12
Fixes disappering map stats

### LINKED ISSUE #9590: Map stats are disappeared after member restart
see https://groups.google.com/d/msgid/hazelcast/5be14630-9d9b-4cb5-a250-a24a9c1ccb57%40googlegroups.com?utm_medium=email&utm_source=footer

ahmetmircik:
__problem:__ If there is no map proxy, stats are not generated. This can be happen if one restarts member and if there is an ongoing update from previously connected client. In this case, since client proxy is not re-created upon member restart, stats collection over map proxy cannot be done.

__fix:__ Stats are collected over recordStores instead of map proxy.

fixes #9590

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/LocalMapStatsProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/LocalMapStatsProvider.java
@@ -21,10 +21,10 @@ import com.hazelcast.internal.cluster.ClusterService;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.map.impl.nearcache.NearCacheProvider;
 import com.hazelcast.map.impl.recordstore.RecordStore;
+import com.hazelcast.monitor.LocalMapStats;
 import com.hazelcast.monitor.NearCacheStats;
 import com.hazelcast.monitor.impl.LocalMapStatsImpl;
 import com.hazelcast.nio.Address;
-import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.partition.IPartition;
 import com.hazelcast.spi.partition.IPartitionService;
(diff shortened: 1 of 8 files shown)


hazelcast/src/test/java/com/hazelcast/map/impl/MapStatisticsAwareServiceTest.java
```diff
@@ -39,11 +39,11 @@
...
     @Test
-    public void getStats() {
+    public void getStats_whenMapHasData() {
         HazelcastInstance hz = createHazelcastInstance();
         warmUpPartitions(hz);
         // when one map exists
-        hz.getMap("map");
+        hz.getMap("map").put(1, 1);
         MapService mapService = getNodeEngineImpl(hz).getService(MapService.SERVICE_NAME);
         Map<String, LocalMapStats> mapStats = mapService.getStats();
```
vbekiaris: Add another test for members who have a `MapContainer` but no `MapProxy`? something like:
```
    @Test
    public void getStats_whenMapProxyDoesNotExist() {
        HazelcastInstance hz1 = factory.newHazelcastInstance();
        hz1.getMap("map").put(1,1);
        // on the second member the map container will be created but no map proxy
        HazelcastInstance hz2 = factory.newHazelcastInstance();
        waitAllForSafeState();

        MapService mapService = getNodeEngineImpl(hz2).getService(MapService.SERVICE_NAME);
        Map<String, LocalMapStats> mapStats = mapService.getStats();

        // then we obtain 1 local map stats instance
        assertEquals(1, mapStats.size());
        assertNotNull(mapStats.get("map"));
    }
```

vbekiaris: I just realized the proxy is already created on the joining side by the post join operation, so please disregard the comment above.

hazelcast/src/main/java/com/hazelcast/map/impl/LocalMapStatsProvider.java
```diff
@@ -78,47 +85,84 @@ public void destroyLocalMapStatsImpl(String name) {
...
 
-        for (int partitionId = 0; partitionId < partitionService.getPartitionCount(); partitionId++) {
-            IPartition partition = partitionService.getPartition(partitionId);
-            Address owner = partition.getOwnerOrNull();
-            if (owner == null) {
-                //no-op because no owner is set yet. Therefore we don't know anything about the map
-                continue;
-            }
+    public Map<String, LocalMapStats> createAllLocalMapStats() {
+        Map statsMap = new HashMap();
```
ahmetmircik: fixed.

PR #9634, opened 2017-01-12
Synchronize default config values and fix xml generator

mmedenjak:
Here we move the old config compatibility checker to the test code since it is no longer used in production. We also expand on this compatibility checker to include as many configuration as possible and use it to check if the default values in the XML config, the Java config and hazelcast-default.xml are the same. The XML generator tests were already using this compatibility checker so expanding it uncovered some missing config values. The changes :

- fix the "default" value in the hazelcast-config.xsd for the networkConfig/joinConfig/multicastConfig/loopbackModeEnabled from true to false
- fix documentation for the some default values
- add fluent API for generating XML
- add missing tags to xml generator

### DIFF
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.8.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.8.xsd
@@ -261,7 +261,7 @@
                                             Number of items in the ringbuffer. If no time-to-live-seconds is set, the size will
                                             always
                                             be equal to capacity after the head completed the first loop around the ring. This is
-                                            because no items are getting retired.
+                                            because no items are getting retired. The default value is 10000.
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
@@ -1206,7 +1206,7 @@
                                     <xs:annotation>
                                         <xs:documentation>
                                             The queue-size defines the maximum number of tasks are able to wait to be
-                                            processed. A value of 0 means unbounded queue.
+                                            processed. A value of 0 means number of partitions * 2.
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
@@ -1376,7 +1376,8 @@
                             maxOccurs="unbounded">
                     <xs:annotation>
                         <xs:documentation>
-                            DataSerializableFactory class to be registered.
+                            Custom classes implementing com.hazelcast.nio.serialization.DataSerializableFactory to be registered.
+                            These can be used to speed up serialization/deserialization of objects.
                         </xs:documentation>
                     </xs:annotation>
                 </xs:element>
--- a/hazelcast/src/main/java/com/hazelcast/config/Config.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/Config.java
@@ -17,7 +17,6 @@
 package com.hazelcast.config;
 
 import com.hazelcast.config.matcher.MatchingPointConfigPatternMatcher;
-import com.hazelcast.core.HazelcastException;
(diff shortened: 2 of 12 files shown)


hazelcast/src/test/java/com/hazelcast/config/ConfigCompatibilityChecker.java
```diff
@@ -0,0 +1,852 @@
...
+            throw new HazelcastException("Incompatible group password");
+        }
+        checkPartitionGroupConfigCompatible(c1, c2);
+        checkSerializationConfigCompatible(c1, c2);
+        checkServicesConfigCompatible(c1, c2);
+        checkManagementCenterConfigCompatible(c1, c2);
+        checkHotRestartConfigCompatible(c1, c2);
+
+        checkNetworkConfigCompatible(c1, c2);
+        checkCompatibleConfigs("map", c1, c2, c1.getMapConfigs(), c2.getMapConfigs(), new MapConfigChecker());
```
vbekiaris: Nice & arduous job. Would it be possible to homogenize all config checks to be invoked with `checkCompatibleConfigs`? (as far as I could tell, all compatibility checks are done by `ConfigChecker`s)

hazelcast/src/main/java/com/hazelcast/config/Config.java
```diff
@@ -1286,7 +1282,7 @@ public Config setDistributedClassloadingConfig(DistributedClassloadingConfig dis
         return this;
     }
 
-    private <T> T lookupByPattern(Map<String, T> configPatterns, String itemName) {
+    <T> T lookupByPattern(Map<String, T> configPatterns, String itemName) {
```
mmedenjak: It's used in the `com.hazelcast.config.ConfigCompatibilityChecker#checkCompatibleConfigs`. Perhaps we can do this differently, haven't thought about it.

hazelcast-spring/src/main/resources/hazelcast-spring-3.8.xsd
```diff
@@ -222,7 +222,7 @@
                                             Number of items in the ringbuffer. If no time-to-live-seconds is set, the size will
                                             always
                                             be equal to capacity after the head completed the first loop around the ring. This is
-                                            because no items are getting retired.
+                                            because no items are getting retired. The default value is 10000.
                                         </xs:documentation>
                                     </xs:annotation>
                                 </xs:attribute>
```
mmedenjak: I would. Can you suggest something, I am not sure if I can write something meaningful. Something like : 
> Custom classes implementing com.hazelcast.nio.serialization.DataSerializableFactory to be registered. These can be used to speed up serialization/deserialization of objects.

hazelcast/src/main/resources/hazelcast-default.xml
```diff
@@ -221,7 +221,7 @@
 
     <jobtracker name="default">
         <max-thread-size>0</max-thread-size>
-        <!-- Queue size 0 means number of partitions * 2 -->
+        <!-- Queue size 0 means the number of processors available to the Java virtual machine -->
```
mmedenjak: Ah, I mistakenly thought the comment applied to `max-thread-size`. Will fix.

mmedenjak:
@vbekiaris added more missing configs, refactored the xml builder to use a fluent xml generator and addressed the comments. Will rebase after there are no more needed changes.

jerrinot:
@vbekiaris, @bilalyasar, @hasancelik: guys, are you happy with this PR now?

mmedenjak:
run-lab-run

PR #9642, opened 2017-01-13
Memcache: BulkGet implemented as a single command

### LINKED ISSUE #9637: MemcacheTest.testBulkSetGet_withManyKeys
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/6683/testReport/junit/com.hazelcast.internal.ascii/MemcacheTest/testBulkSetGet_withManyKeys/

Regression

Timeout waiting for bulk values: waited 216,000 ms. Node status: Connection Status { /127.0.0.1:5703 active: true, authed: true, last read: 215,997 ms ago }
Stacktrace

net.spy.memcached.OperationTimeoutException: Timeout waiting for bulk values: waited 216,000 ms. Node status: Connection Status { /127.0.0.1:5703 active: true, authed: true, last read: 215,997 ms ago }
	at net.spy.memcached.internal.BulkGetFuture.get(BulkGetFuture.java:126)
	at net.spy.memcached.internal.BulkGetFuture.get(BulkGetFuture.java:51)
	at net.spy.memcached.MemcachedClient.getBulk(MemcachedClient.java:1465)
	at net.spy.memcached.MemcachedClient.getBulk(MemcachedClient.java:1509)
	at net.spy.memcached.MemcachedClient.getBulk(MemcachedClient.java:1523)

jerrinot:
Fixes #9637

The old implementation generated PartialGetCommand per each key
and EndCommand. Commands are executed on a multi-threaded
executor service: PartialGetCommand can be re-ordered with EndCommand.

The new implementation has 2 advantages
1. Uses map.getAll() <- it gives better performance than calling get() per each key
2. Having a single command eliminates races.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/memcache/EntryConverter.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/memcache/EntryConverter.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.ascii.memcache;
+
+import com.hazelcast.internal.ascii.TextCommandService;
+import com.hazelcast.logging.ILogger;
+
+import static com.hazelcast.util.StringUtil.stringToBytes;
+
+/**
+ * Produced {@link MemcacheEntry} from key and values
+ *
+ */
+public final class EntryConverter {
+    private final TextCommandService textCommandService;
+    private final ILogger logger;
+
+    public EntryConverter(TextCommandService textCommandService, ILogger logger) {
+        this.textCommandService = textCommandService;
+        this.logger = logger;
+    }
+
+    public MemcacheEntry toEntry(String key, Object value) {
+        if (value == null) {
+            return null;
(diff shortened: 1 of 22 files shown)


hazelcast/src/main/java/com/hazelcast/internal/ascii/memcache/EntryConverter.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        MemcacheEntry entry = null;
+        if (value instanceof MemcacheEntry) {
+            entry = (MemcacheEntry) value;
+        } else if (value instanceof byte[]) {
+            entry = new MemcacheEntry(key, ((byte[]) value), 0);
+        } else if (value instanceof String) {
+            entry = new MemcacheEntry(key, stringToBytes((String) value), 0);
+        } else {
```
tombujok: Can't the value be Data?

jerrinot: it should not. it uses a regular IMap proxy and proxy never exposes Data to users.

hazelcast/src/test/java/com/hazelcast/util/collection/InternalSetMultiMapTest.java
```diff
@@ -0,0 +1,70 @@
...
+        assertThat(results, hasSize(1));
+        assertThat(results, contains("value"));
+    }
+
+    @Test
+    public void entrySet_whenEmpty_thenReturnEmptySet() {
+        Set<?> entries = multiMap.entrySet();
+        assertThat(entries, hasSize(0));
+    }
+}
```
tombujok: Maybe a separate integration test doing a round-trip from a client wouldn't be a bad idea? :)

jerrinot: we already have an [integration test](https://github.com/jerrinot/hazelcast/blob/f87f2b5989350fd466fd5818f6d6619ae4166324/hazelcast/src/test/java/com/hazelcast/internal/ascii/MemcacheTest.java#L58-L58) - that's how this all started - @mdogan had added new tests and they failed occasional.

tombujok: ok 👍

PR #9646, opened 2017-01-13
Added removeAll codec to make implementation more client protocol fri…

ahmetmircik:
…endly

depends on https://github.com/hazelcast/hazelcast-client-protocol/pull/55

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -61,6 +61,7 @@ import com.hazelcast.client.impl.protocol.codec.MapPutAllCodec;
 import com.hazelcast.client.impl.protocol.codec.MapPutCodec;
 import com.hazelcast.client.impl.protocol.codec.MapPutIfAbsentCodec;
 import com.hazelcast.client.impl.protocol.codec.MapPutTransientCodec;
+import com.hazelcast.client.impl.protocol.codec.MapRemoveAllCodec;
 import com.hazelcast.client.impl.protocol.codec.MapRemoveCodec;
 import com.hazelcast.client.impl.protocol.codec.MapRemoveEntryListenerCodec;
 import com.hazelcast.client.impl.protocol.codec.MapRemoveIfSameCodec;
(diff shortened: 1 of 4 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1576,14 +1582,6 @@ public String toString() {
         return "IMap{" + "name='" + name + '\'' + '}';
     }
 
-    @Override
-    protected void onInitialize() {
-        super.onInitialize();
```
ihsandemir: why delete this?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapRemoveAllMessageTask.java
```diff
@@ -0,0 +1,89 @@
...
+import com.hazelcast.security.permission.MapPermission;
+import com.hazelcast.spi.OperationFactory;
+
+import java.security.Permission;
+import java.util.Map;
+
+import static com.hazelcast.map.impl.EntryRemovingProcessor.ENTRY_REMOVING_PROCESSOR;
+import static com.hazelcast.map.impl.MapService.SERVICE_NAME;
+
+public class MapRemoveAllMessageTask extends AbstractMapAllPartitionsMessageTask<MapRemoveAllCodec.RequestParameters> {
```
ihsandemir: Why is it all partitions task? all partitions task may be problematic in terms of response waiting which may block an operation thread. From the operation, I understand that it may be specific to a partition hence can it be AbstractPartitionMessageTask?

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
ahmetmircik: no delete only moved it up https://github.com/hazelcast/hazelcast/pull/9646/files/21d7af8bc84afa61c7e99b6f967ceb7f62b4aad8#diff-3540ff2664ebe82892440b80a7293680R250

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapRemoveAllMessageTask.java
ahmetmircik: i need `createOperationFactory` that was the reason.

ahmetmircik:
run-lab-run

Issue #9650, opened 2017-01-16
Classloader Leakage in ClockTest (PermGen Space OOME)


Donnerbart:
We seem to have an issue with the `ClockTest`, which sometimes seems to produce a PermGen OOME failure:
```
11:45:29 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test (default-test) on project hazelcast: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test failed: There was an error in the forked process
11:45:29 
[ERROR] java.lang.OutOfMemoryError: PermGen space
```

This could be a production issue with a real leak or just a test issue with the `FilteringClassloader` which is used in that test to create multiple instances of the `Clock` singleton. Another path to explore would be the `ClassloaderUtil` which supports caching. Maybe our test classes are stacked up there for no good reason.

At least it makes me suspicious that the `ClockTest` (which doesn't start any HZ instance) takes about 3 seconds to execute `test_whenConfiguringClockOffset_thenSystemOffsetClockIsCreated` (which is the actual test to create a new `Clock` singleton with the second internal implementation).

Donnerbart:
![screenshot from 2017-01-16 14-51-55](https://cloud.githubusercontent.com/assets/4196298/21985510/5e675aac-dbfb-11e6-93a5-c1223d51e24b.png)
This is the allocation profile from just a single `startIsolatedNode()` and the shutdown of it. This doesn't look very healthy to me (84 MB in about 3 seconds, just for class loading related code). If this is polluting the PermGen Space I can image why this is causing troubles. I'll have a look now what's going on there in detail.

Donnerbart:
I ran the tests with Java 6, since Java 8 has no PermGen Space anymore. Unfortunately JFC is not available in JDK 6, so I used VisualVM for a quick peek into the PermGen space.

I modified the test to call `startIsolatedNode();` and `shutdownIsolatedNode();` in a loop until it ran into the OOME:
![screenshot from 2017-01-17 15-16-22](https://cloud.githubusercontent.com/assets/4196298/22023831/1fdf23d2-dcc8-11e6-9d83-8f2e0fb33ab6.png)
[...]

PR #9652, opened 2017-01-16
Split ClockTest into a QuickTest and SlowTest

Donnerbart:
`ClockTest` seems to crash our CI builds quite often. The test method
`test_whenConfiguringClockOffset_thenSystemOffsetClockIsCreated()` takes
over 3 seconds to complete. The `SystemClockChangeTest` which uses the
same classloader magic takes up to 3 minutes (it's a `NightlyTest`).

This commit splits up the `ClockTest` into a really quick unit test of all
single components and skips the classloading magic for the singleton
part of the `Clock` implementation. This should fix the immediate issue
with our builds, but needs further investigation of the root cause.

Part of https://github.com/hazelcast/hazelcast/issues/9650

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/util/Clock.java
+++ b/hazelcast/src/main/java/com/hazelcast/util/Clock.java
@@ -18,6 +18,8 @@ package com.hazelcast.util;
 
 import com.hazelcast.nio.ClassLoaderUtil;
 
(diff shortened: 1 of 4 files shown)


Donnerbart:
I've added some more tests to the `QuickTest`. We should have about 85% line coverage without any classloader magic now. Will still investigate the classloading test with a profiler, to see why it takes so long and what happens under the hood.

PR #9626, opened 2017-01-11
Fixes disappering map stats

ahmetmircik:
forward port of https://github.com/hazelcast/hazelcast/pull/9630

PR #9653, opened 2017-01-16
Provided map stats for the maps that only have a proxy but not have a…

ahmetmircik:
…ny data

This was a regression caused by https://github.com/hazelcast/hazelcast/pull/9626

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/MapStatisticsAwareServiceTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/MapStatisticsAwareServiceTest.java
@@ -17,11 +17,13 @@
 package com.hazelcast.map.impl;
 
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.IMap;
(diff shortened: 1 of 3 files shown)


hazelcast/src/test/java/com/hazelcast/map/impl/MapStatisticsAwareServiceTest.java
```diff
@@ -43,7 +43,7 @@ public void getStats_whenMapHasData() {
         HazelcastInstance hz = createHazelcastInstance();
         warmUpPartitions(hz);
         // when one map exists
-        hz.getMap("map").put(1, 1);
+        hz.getMap("map");
```
vbekiaris: maybe keep this test as is to match its goal (`getStats_whenMapHasData`) and add a separate to test the case where proxy is created but no data have been stored yet?

ahmetmircik: done.

PR #9657, opened 2017-01-16
Provided map stats for the maps which only have a proxy but don't hav…

ahmetmircik:
…e any data

backport of https://github.com/hazelcast/hazelcast/pull/9653

PR #9658, opened 2017-01-16
Provided map stats for the maps which only have a proxy but don't hav…

ahmetmircik:
…e any data

backport of #9653


ahmetmircik:
see https://github.com/hazelcast/hazelcast/pull/9657

PR #9668, opened 2017-01-17
Use Log4j2 in tests

jerrinot:
## Still work in progress. 

Steps to be done:
- [x] Tests should not poke global logging setting.
- [x] Configure test runners to use log4j2 when no other logging is explicitly configured
- [x] Configure log4j2 to have the same feature as our current logging. This includes appending test name to each logging line. 
- [x] Tests should use a specialized logging factory to cover branches guarderded behind `isFooEnabled()` checks. 
- [ ] Check EE tests do not depend on log4j

Why to upgrade?
- log4j1 is no longer maintained
- I suspect it has a dead-lock causing occasional test-failures.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
@@ -61,9 +61,10 @@ public abstract class AbstractHazelcastClassRunner extends AbstractParameterized
     private static final boolean THREAD_CONTENTION_INFO_AVAILABLE;
 
     static {
-        String logging = "hazelcast.logging.type";
-        if (System.getProperty(logging) == null) {
-            System.setProperty(logging, "log4j2");
+        String loggingType = "hazelcast.logging.type";
+        String loggingClass = "hazelcast.logging.class";
+        if (System.getProperty(loggingType) == null && System.getProperty(loggingClass) == null) {
+            System.setProperty(loggingClass, TestLoggerFactory.class.getName());
             System.setProperty("isThreadContextMapInheritable", "true");
         }
         if (System.getProperty(TestEnvironment.HAZELCAST_TEST_USE_NETWORK) == null) {
--- a/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
@@ -60,12 +60,14 @@ public abstract class AbstractHazelcastClassRunner extends AbstractParameterized
     private static final boolean THREAD_CPU_TIME_INFO_AVAILABLE;
     private static final boolean THREAD_CONTENTION_INFO_AVAILABLE;
 
+    private static String LOGGING_TYPE_PROP_NAME = "hazelcast.logging.type";
+    private static String LOGGING_CLASS_PROP_NAME = "hazelcast.logging.class";
+
     static {
-        String loggingType = "hazelcast.logging.type";
-        String loggingClass = "hazelcast.logging.class";
-        if (System.getProperty(loggingType) == null && System.getProperty(loggingClass) == null) {
-            System.setProperty(loggingClass, TestLoggerFactory.class.getName());
+        if (shouldForceTestLoggingFactory()) {
+            System.setProperty(LOGGING_CLASS_PROP_NAME, TestLoggerFactory.class.getName());
             System.setProperty("isThreadContextMapInheritable", "true");
+            System.clearProperty(LOGGING_TYPE_PROP_NAME);
         }
         if (System.getProperty(TestEnvironment.HAZELCAST_TEST_USE_NETWORK) == null) {
             System.setProperty(TestEnvironment.HAZELCAST_TEST_USE_NETWORK, "false");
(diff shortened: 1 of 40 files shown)


Donnerbart:
Please consider that we use the "enable log4" and "set a fine/finest log level" to trigger the execution of logging code blocks. If you remove this log level setting from the tests, we might lose the coverage of those code blocks. This is an issue for the code coverage QA stage and may hide bugs in those code blocks, e.g.:
```
MyClass foo = getFoo();
  logger.finest("MyClass value: " + foo.getValue());
}
```
And this is just the simplest bug type I can think of.

I totally agree that tests should not fiddle with global log settings. We also cannot set FINEST as default, since this will make our log size explode. This is why I came up with the idea of having a `TestLogger` implementation which fakes it outer status to "yeah, all levels are enabled", but just delegates INFO and higher to the real logger. FINE and FINEST could be saved in a small RingBuffer, so they are accessible from the test when needed. Something like this.

jerrinot:
@Donnerbart: https://github.com/hazelcast/hazelcast/pull/9668/commits/03a29d22c16ca7c77b2829c92b43b68512f3ce6a we happy?

hazelcast/src/test/java/com/hazelcast/test/TestLoggerFactory.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private Log4j2Factory log4j2Factory = new Log4j2Factory();
+
+    @Override
+    public ILogger getLogger(String name) {
+        ILogger logger = log4j2Factory.getLogger(name);
+        return new DelegatingTestLogger(logger);
+    }
+
+    private static class DelegatingTestLogger implements ILogger {
```
mdogan: nice workaround! +1

hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
```diff
@@ -262,46 +256,8 @@ public static void printAllStackTraces() {
...
-    private static boolean isLog4jLoaded() {
-        setLoggingLog4j();
-        try {
-            Class.forName("org.apache.log4j.Logger");
-            Class.forName("org.apache.log4j.Level");
-            return true;
-        } catch (Throwable ignored) {
-            return false;
-        }
+    public static void setLoggingLog4j2() {
```
mdogan: `setLoggingLog4j2()` is not used at all. Do we need it somewhere? I think we should avoid explicitly/programmatically setting/changing logging in tests. In future we can decide to switch another logging framework.

Donnerbart: Can't we enable logging for all tests using the new delegating logger? Or is even the INFO level too much output?

hazelcast/src/test/resources/log4j.properties
```diff
@@ -29,3 +29,6 @@ log4j.logger.com.hazelcast.spi.hotrestart.cluster=debug
 log4j.logger.com.hazelcast.test.mocknetwork=debug
 
 log4j.rootLogger=info, stdout
+
+# WARNING: This config file is no longer used in Hazelcast tests
```
Donnerbart: Can't we remove the whole file then?

jerrinot: I don't know. there is a config file for each logging type we support.

hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
jerrinot: the new logger is used for all tests using ran via the `AbstractHazelcastClassRunner` (unless you configure logging explicitly)

I agree this method can be removed.

hazelcast/src/test/resources/log4j.properties
Donnerbart: Hmm, maybe they are or were used to easily test all logging frameworks. Then let's keep it as long as we support log4j.

jerrinot:
There is still something switching logging to log4j1.

jerrinot:
ok, it's part of the Jenkins build setup:
image

hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
```diff
@@ -64,9 +65,10 @@
     static {
         String loggingType = "hazelcast.logging.type";
         String loggingClass = "hazelcast.logging.class";
-        if (System.getProperty(loggingType) == null && System.getProperty(loggingClass) == null) {
+        if (JenkinsDetector.isOnJenkins()) {
```
mdogan: what about `JenkinsDetector.isOnJenkins() || (System.getProperty(loggingType) == null && System.getProperty(loggingClass) == null)` ?

I'd like to see same logging while running tests on my local, instead of defining logging type each time.

jerrinot: Good idea. 

I see the `JenkinsDetector.isOnJenkins()` as a migration measure. Once/if the code is merged then we should remove the logging configuration from Jenkins builds. It's duplicated on way too many places (in all builds)

jerrinot:
run-lab-run

hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
```diff
@@ -20,7 +20,7 @@
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.test.annotation.Repeat;
 import com.hazelcast.util.EmptyStatement;
-import org.apache.log4j.MDC;
+import org.apache.logging.log4j.ThreadContext;
```
mdogan: not used?

jerrinot: removed, thanks!

Donnerbart:
(assuming the last test failure is fixed)

PR #9672, opened 2017-01-17
Dump threads on test failures on Jenkins by default

mmedenjak:
If the system property is set, use that configuration value. Otherwise, enable thread dumps on test failures if running on Jenkins.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
@@ -44,7 +44,6 @@ import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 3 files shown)


hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
```diff
@@ -19,6 +19,7 @@
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.test.annotation.Repeat;
+import com.hazelcast.test.jitter.JenkinsDetector;
```
mdogan: You can even move `JenkinsDetector` to `com.hazelcast.test` package.

mmedenjak: Moved.

Issue #9374, opened 2016-12-06
[lock] [entry-processor] Transaction couldn't obtain lock for the key


dmitrymz:
we are sporadically experiencing TransactionException: Transaction couldn't obtain lock for the key, while calling TransactionalMap.getForUpdate().

What is sad that the cluster itself looks normal, no exceptions in the logs, and on the client side we have no idea how to handle this exception. Is this partition corruption, data issue...
How as a client we can rely on datastore throwing such generic exceptions...

could you please explain what conditions are leading to this exception? what should be done / best practices dealing with transactions to avoid this situation? Should we retry the request?
how else we can handle this exception?

from the source of TransactionalMapProxySupport:270
```
            versionedValue = future.get();
            if (versionedValue == null) {
                throw new TransactionException("Transaction couldn't obtain lock for the key: " + toObjectIfNeeded(key));
            }
```
what does this check for null mean? was the value deleted while trying to obtain the lock?

```
Caused by: com.hazelcast.transaction.TransactionException: Transaction couldn't obtain lock for the key: HeapData{type=-11, hashCode=-1540458485, partitionHash=-1540458485, totalSize=48, dataSize=40, heapCost=68}
	at com.hazelcast.map.impl.tx.TransactionalMapProxySupport.lockAndGet(TransactionalMapProxySupport.java:270) ~[hazelcast-3.6.5.jar:3.6.5]
	at com.hazelcast.map.impl.tx.TransactionalMapProxySupport.getForUpdateInternal(TransactionalMapProxySupport.java:117) ~[hazelcast-3.6.5.jar:3.6.5]
	at com.hazelcast.map.impl.tx.TransactionalMapProxy.getForUpdate(TransactionalMapProxy.java:124) ~[hazelcast-3.6.5.jar:3.6.5]
	at com.hazelcast.client.impl.protocol.task.transactionalmap.TransactionalMapGetForUpdateMessageTask.innerCall(TransactionalMapGetForUpdateMessageTask.java:43) ~[hazelcast-3.6.5.jar:3.6.5]
	at com.hazelcast.client.impl.protocol.task.AbstractTransactionalMessageTask.call(AbstractTransactionalMessageTask.java:34) ~[hazelcast-3.6.5.jar:3.6.5]
	at com.hazelcast.client.impl.protocol.task.AbstractCallableMessageTask.processMessage(AbstractCallableMessageTask.java:34) ~[hazelcast-3.6.5.jar:3.6.5]
[...]

PR #9676, opened 2017-01-17
Throw TransactionTimedOutException when unable to lock key for update

mmedenjak:
Previously TransactionTimeout was thrown, now throw TransactionTimedOutException. Update the javadoc to point out that this exception can be thrown from the methods that do mutation.

Affected issue : https://github.com/hazelcast/hazelcast/issues/9374


jerrinot:
@mmedenjak: do you know if this exception is also propagated to client proxies?

mmedenjak:
@jerrinot just checked, the client gets the same `TransactionTimedOutException`

Issue #9386, opened 2016-12-06
Scheduled Executor Tests are too slow


jerrinot:
from Jenkins log:
`17:37:05 Tests run: 33, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 403.775 sec - in com.hazelcast.client.scheduledexecutor.ClientScheduledExecutorServiceTest`

This makes the PR builder slow.

jerrinot:
You can either make the tests faster or split them into quick tests and slow nightly tests.

PR #9678, opened 2017-01-17
Inject HazelcastInstance in HazelcastInstanceAware commands

### LINKED ISSUE #9675: IScheduledExecutorService appears not to use HazelcastInstanceAware on local operations
Version: 3.8 EA

I've been using the IScheduledExecutorService which was added in Hazelcast 3.8. It appears that when operations are scheduled and executed on a single local node, that the HazelcastInstance does not get injected. An instance is injected if the task is submitted and executed on a remote node.

Sample reproduction case; console contains:

```
Testing via ExecutorService
Injected state: true
Testing via ScheduledExecutorService
Injected state: false
```

Expected:

```
Testing via ExecutorService
Injected state: true
Testing via ScheduledExecutorService
Injected state: true
```

Test case:

```java
import com.hazelcast.core.HazelcastInstanceAware;
import org.junit.After;
import org.junit.Before;
import org.junit.Test;

import java.io.Serializable;

public class HazelcastInstanceAwareTest {
    HazelcastInstance instance;

    @Before
    public void setup() {
        instance = Hazelcast.newHazelcastInstance();
    }

    @After
    public void teardown() throws Exception {
        instance.shutdown();
    }

    @Test
    public void test() throws ExecutionException, InterruptedException {
        System.err.println("Testing via ExecutorService");
        assertTrue(instance.getExecutorService("foo").submit((Callable<Boolean>) new AwareRunnable()).get());

        System.err.println("Testing via ScheduledExecutorService");
        instance.getScheduledExecutorService("bar").scheduleAtFixedRate(new AwareRunnable(), 0, 5, TimeUnit.SECONDS);

        Thread.sleep(TimeUnit.SECONDS.toMillis(10));
    }

    static class AwareRunnable implements Runnable, Callable<Boolean>, HazelcastInstanceAware, Serializable {
        private volatile HazelcastInstance instance;

        @Override
        public void setHazelcastInstance(final HazelcastInstance instance) {
            this.instance = instance;
        }

        @Override
        public void run() {
            call();
        }

        @Override
        public Boolean call() {
            final boolean injected = (instance != null);
            System.err.println("Injected state: " + injected);
            return injected;
        }
    }
}
```

vbekiaris:
When a `HazelcastInstanceAware` command is passed as argument to one of the `schedule*` methods in `IScheduledExecutorService`, inject the `HazelcastInstance`.

In case the command is deserialized on another member, `HazelcastInstance` is already injected by https://github.com/vbekiaris/hazelcast/blob/d2694664aaf0d0e4bc9a8b792e300672cb0e48a2/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/AbstractSerializationService.java#L268-L268 . However when the command is scheduled on the local member, `HazelcastInstance` was not injected at all.

Fixes #9675

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientScheduledFutureProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientScheduledFutureProxy.java
@@ -186,7 +186,7 @@ public class ClientScheduledFutureProxy<V>
     @Override
     public V get()
             throws InterruptedException, ExecutionException {
-        return this.get0().join();
+        return this.get0().get();
     }
 
     @Override
--- a/hazelcast-client/src/test/java/com/hazelcast/client/scheduledexecutor/ClientScheduledExecutorServiceBasicTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/scheduledexecutor/ClientScheduledExecutorServiceBasicTest.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.scheduledexecutor;
+
+import com.hazelcast.client.test.TestHazelcastFactory;
+import com.hazelcast.config.Config;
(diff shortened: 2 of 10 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientScheduledFutureProxy.java
```diff
@@ -186,7 +186,7 @@ public boolean isDone() {
     @Override
     public V get()
             throws InterruptedException, ExecutionException {
-        return this.get0().join();
+        return this.get0().get();
```
vbekiaris: `join` peels exceptions wrapped in `ExecutionException`s, so it would unwrap the exception in the received `ExecutionException`.

vbekiaris:
Pushed an additional commit that:
* Splits `ScheduledExecutorServiceTest` to separate quick & slow tests. Quick test finishes (on my machine) in 20seconds. (fixes #9386)
* Removes `HazelcastInstance` constructor argument from `Callable`s/`Runnable`s that implement `HazelcastInstanceAware`
* Uses `InvocationFuture.get` instead of `join` to avoid peeling `ExecutionException`s in `IScheduledFuture.get` (fixes #9685)
* Enables two client-side ignored tests & introduces some minor clean ups in tests.

PR #9679, opened 2017-01-17
[DONT MERGE] Fixed a classloader leak in ThreadLocalRandom

Donnerbart:
`ThreadLocalRandom` puts itself into the `ThreadLocal` storage. Since it
holds a reference to its classloader the whole class hierarchy of
Hazelcast is still reachable and cannot be garbage collected.

This leads to a PermGen Space OOME if the `FilteringClassLoader` is used
to start a whole Hazelcast instance for several times.

This is a very hacky fix, but it seems to work in my scenario.
But I'm not sure if it's a good solution and if it's the correct place to hook in.
Let's first see if the PR builder shows any unwanted side effects.

Possible fix for https://github.com/hazelcast/hazelcast/issues/9650

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
@@ -16,25 +16,19 @@
 
 package com.hazelcast.internal.util;
 
-/*
- * Written by Doug Lea with assistance from members of JCP JSR-166
- * Expert Group and released to the public domain, as explained at
- * http://creativecommons.org/publicdomain/zero/1.0/
- */
-
 import java.util.Random;
 
 /**
- * A random number generator isolated to the current thread.  Like the
+ * A random number generator isolated to the current thread. Like the
  * global {@link java.util.Random} generator used by the {@link
  * java.lang.Math} class, a {@code ThreadLocalRandom} is initialized
  * with an internally generated seed that may not otherwise be
  * modified. When applicable, use of {@code ThreadLocalRandom} rather
  * than shared {@code Random} objects in concurrent programs will
- * typically encounter much less overhead and contention.  Use of
+ * typically encounter much less overhead and contention. Use of
  * {@code ThreadLocalRandom} is particularly appropriate when multiple
- * tasks (for example, each a {@link java.util.concurrent.ForkJoinTask}) use random numbers
- * in parallel in thread pools.
+ * tasks (for example, each a {@code java.util.concurrent.ForkJoinTask})
+ * use random numbers in parallel in thread pools.
  *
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
```diff
@@ -45,14 +39,31 @@
...
  *
- * @since 1.7
+ * Written by Doug Lea with assistance from members of JCP JSR-166
+ * Expert Group and released to the public domain, as explained at
+ * http://creativecommons.org/publicdomain/zero/1.0/
+ *
  * @author Doug Lea
  */
-public class ThreadLocalRandom extends Random {
+public final class ThreadLocalRandom extends Random {
```
mdogan: Please do not modify this class. It should be identical to the one in JSR-166 repo.

mdogan:
Can you explain how/when `ThreadLocalRandom` class is loaded on those tests?

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
Donnerbart: I had to add a new method, and then I think it's better to integrate the class into our code base. I think we did this with other classes before. Nevertheless, this is just an experiment for now, so this might not be the final solution anyway.

Donnerbart:
I don't know exactly why or when it is loaded. I just see two use cases in the code at all, and they both should not be invoked in that test (no client, no backpressure). I assume it's done via the class loading of the `FilteringClassLoader` which loads all `com.hazelcast` classes. I think this triggers the static initializer of the `ThreadLocalRandom` class.

As soon as the `localRandom` or `LOCAL_RANDOM` field is set, we have a reference to it in the `ThreadLocal` of the test thread. And I understood that this one holds the classloader reference which has again references to all other HZ classes. So none of them can be GC collected.

Donnerbart:
I was wrong, the `BackPressureRegulator` is used by default and uses the `ThreadLocalRandom`, so this might be the source of the leakage.

mdogan:
But it's loaded when only `randomSyncDelay()` is called. and `BackPressureRegulator` invokes `randomSyncDelay()` when only backpressure is enabled.

mdogan:
ok, it's also called in c'tor;

```
        for (int partitionId = 0; partitionId < partitionCount; partitionId++) {
            syncDelays[partitionId * INTS_PER_CACHE_LINE] = randomSyncDelay();
        }
```

Donnerbart:
Yes, I need to do a bit more research. Jaromir pointed me to an article about a general `ThreadLocal` cleanup. I'll see if we have this issue with our other `ThreadLocal` usages as well for other classloaders than our `FilteringClassLoader`.

PR #9680, opened 2017-01-17
SizeEstimator refactoring to better fit the needs of the HD entry cost

tkountis:
Refactored SizeEstimator to be consistent with other API around entry cost
Added one additional reference cost in the CachedDataRecordWithStats which was missing
Improved API where needed to allow support of HD entry cost for EE

TODO

- [x] Verify changes with Management Centre


hazelcast/src/main/java/com/hazelcast/map/impl/OwnedEntryCostEstimator.java
```diff
@@ -17,17 +17,20 @@
...
 /**
- * Size estimator general contract.
+ * Owned entry cost estimator general contract.
  *
- * @param <T> the type of object which's size going to be estimated.
+ * @param <K> the type of key which's size going to be estimated.
+ * @param <V> the type of key which's size going to be estimated.
  */
-public interface SizeEstimator<T> {
+public interface OwnedEntryCostEstimator<K, V> {
```
ahmetmircik: `owned` in the name can be confusing. Because size estimator can also calculate backup entry costs or near cached entry costs. Maybe naming it as `MemoryCostEstimator` fits better?

hazelcast/src/main/java/com/hazelcast/map/impl/OwnedEntryCostEstimator.java
```diff
@@ -17,17 +17,20 @@
...
 
-    long getSize();
+    long getEstimate();
 
-    void add(long size);
+    void adjustEstimateBy(long adjustment);
 
-    long calculateSize(T object);
+    long calculateCost(V value);
```
ahmetmircik: `calculateValueCost`? 

Also can you please add javaDoc for the methods?

hazelcast/src/main/java/com/hazelcast/map/impl/OwnedEntryCostEstimator.java
mdogan: Agree with @ahmetmircik, `EntryCostEstimator` fits better, owned or not, it can estimate cost of an entry.

PR #9686, opened 2017-01-18
Suppressing excessive logging on serialization failure during multicast join

### LINKED ISSUE #8867: Excessive logging of 3.6 and 3.7 cluster multicast discovery join failure
[As mentioned on Gitter](https://gitter.im/hazelcast/hazelcast?at=57d27fcef33967fc643f191c):

I have an existing Hazelcast 3.6 cluster running. Then I start up a cluster with a **different group name** using Hazelcast 3.7 on the same network. They are both using the same multicast IP and port.

**Expected**
They do not join each others' cluster because the group names differ, and **there is no log output** (at INFO level or above).

**Actual**
They do not join each others' cluster, but there is a lot of log output about failing to join, which is confusing and floods the logs. For example, there are logs like the following:

```
2016-09-09 16:56:56.728  WARN 12687 --- [hz._hzInstance_1_test-dev.MulticastThread] c.h.i.cluster.impl.MulticastService      : [1xx.xx.xx.xxx]:xxxxx [test-dev] [3.7] Received data format is invalid. (An old version of Hazelcast may be running here.)

com.hazelcast.nio.serialization.HazelcastSerializationException: Problem while reading DataSerializable, namespace: 0, id: 0, class: 'com.hazelcast.cluster.impl.JoinRequest', exception: com.hazelcast.cluster.impl.JoinRequest
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.rethrowReadException(DataSerializableSerializer.java:141) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:130) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:52) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.serialization.impl.StreamSerializerAdapter.read(StreamSerializerAdapter.java:46) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.serialization.impl.AbstractSerializationService.readObject(AbstractSerializationService.java:216) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.serialization.impl.ByteArrayObjectDataInput.readObject(ByteArrayObjectDataInput.java:600) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.cluster.impl.MulticastService.receive(MulticastService.java:207) [hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.cluster.impl.MulticastService.run(MulticastService.java:165) [hazelcast-3.7.jar:3.7]
    at java.lang.Thread.run(Thread.java:745) [na:1.8.0_72]
Caused by: java.lang.ClassNotFoundException: com.hazelcast.cluster.impl.JoinRequest
    at java.net.URLClassLoader.findClass(URLClassLoader.java:381) ~[na:1.8.0_72]
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424) ~[na:1.8.0_72]
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331) ~[na:1.8.0_72]
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357) ~[na:1.8.0_72]
    at com.hazelcast.nio.ClassLoaderUtil.tryLoadClass(ClassLoaderUtil.java:151) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.nio.ClassLoaderUtil.loadClass(ClassLoaderUtil.java:120) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.nio.ClassLoaderUtil.newInstance(ClassLoaderUtil.java:73) ~[hazelcast-3.7.jar:3.7]
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:125) ~[hazelcast-3.7.jar:3.7]
    ... 7 common frames omitted
```

**Workaround**
Use a different multicast IP or port.

Other than logging output, I think there is no actual effect, but it certainly caused concern in our DEV environment when starting up the new 3.7 cluster. We can work around it by changing the multicast IP or port for now, as having our logs flooded with those stack traces is not a reasonable option for us and I'm not sure it is smart for us to silence logs from `MulticastService`. However, this means we need to worry about the multicast IP and port of all other Hazelcast instances (that also vary in minor version) on the network and may have this happen again in the future. We'd rather not manage the port/IP and instead have a policy of unique group name.

Note that I tested and this is not an issue if both clusters are Hazelcast 3.6. If the info so far is not enough, I can put together a repository on GitHub to reproduce the issue.

tombujok:
Fixes #8867 (or at least makes it less verbose and cumbersome)

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastService.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastService.java
@@ -47,6 +47,7 @@ public final class MulticastService implements Runnable {
     private static final int SOCKET_BUFFER_SIZE = 64 * 1024;
     private static final int SOCKET_TIMEOUT = 1000;
     private static final int SHUTDOWN_TIMEOUT_SECONDS = 5;
+    private static final int JOIN_SERIALIZATION_ERROR_SUPPRESSION_MILLIS = 60000;
 
     private final List<MulticastListener> listeners = new CopyOnWriteArrayList<MulticastListener>();
     private final Object sendLock = new Object();
@@ -59,7 +60,9 @@ public final class MulticastService implements Runnable {
     private final DatagramPacket datagramPacketSend;
     private final DatagramPacket datagramPacketReceive;
 
+    private long lastLoggedJoinSerializationFailure;
     private volatile boolean running = true;
+
 
     private MulticastService(Node node, MulticastSocket multicastSocket) throws Exception {
         this.logger = node.getLogger(MulticastService.class.getName());
@@ -217,7 +220,14 @@ public final class MulticastService implements Runnable {
                 }
             } catch (Exception e) {
                 if (e instanceof EOFException || e instanceof HazelcastSerializationException) {
-                    logger.warning("Received data format is invalid. (An old version of Hazelcast may be running here.)", e);
+                    long now = System.currentTimeMillis();
+                    if (now - lastLoggedJoinSerializationFailure > JOIN_SERIALIZATION_ERROR_SUPPRESSION_MILLIS) {
+                        lastLoggedJoinSerializationFailure = now;
+                        logger.warning("Received a JoinRequest with an incompatible binary-format. "
+                                + "An old version of Hazelcast may be using the same multicast discovery port. "
+                                + "Are you running multiple Hazelcast clusters on this host? "
+                                + "(This message will be suppressed for 60 seconds). ");
+                    }
                 } else {
                     throw e;
                 }


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastService.java
```diff
@@ -59,8 +60,10 @@
     private final DatagramPacket datagramPacketSend;
     private final DatagramPacket datagramPacketReceive;
 
+    private volatile long lastLoggedJoinSerializationFailure;
```
vbekiaris: Does it have to be `volatile`? As far as I can tell, this field is read & updated from a single thread ([`multicastServiceThread`](https://github.com/vbekiaris/hazelcast/blob/0f62712f53ae6e207e7909f785f5f3c78e960393/hazelcast/src/main/java/com/hazelcast/instance/Node.java#L379-L379)) and its default value `0` will be visible to the subsequently started thread.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastService.java
```diff
@@ -72,6 +75,7 @@ private MulticastService(Node node, MulticastSocket multicastSocket) throws Exce
         this.datagramPacketSend = new DatagramPacket(new byte[0], 0, InetAddress.getByName(multicastConfig.getMulticastGroup()),
                 multicastConfig.getMulticastPort());
         this.datagramPacketReceive = new DatagramPacket(new byte[DATAGRAM_BUFFER_SIZE], DATAGRAM_BUFFER_SIZE);
+        this.lastLoggedJoinSerializationFailure = 0;
```
vbekiaris: Redundant initialization

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastService.java
tombujok: Thx. You're right

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastService.java
tombujok: Yep, true, but implicit initialisation is less readable. I inlined it though.

Issue #9674, opened 2017-01-17
Near cache compatibility with 3.7 client is broken for TTL expirations


ihsandemir:
The following client test fails if the client if 3.7.4 and when it is run against 3.8-SNAPSHOT server. It passes when both client and server are 3.8-SNAPSHOT or 3.7.4.
```
package com.hazelcast.client.map.impl.nearcache;

import com.hazelcast.client.HazelcastClient;
import com.hazelcast.client.config.ClientConfig;
import com.hazelcast.config.NearCacheConfig;
import com.hazelcast.core.Hazelcast;
import com.hazelcast.core.HazelcastInstance;
import com.hazelcast.core.IMap;
import com.hazelcast.test.AssertTask;
import com.hazelcast.test.HazelcastTestSupport;

import java.util.concurrent.TimeUnit;

import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertNull;

/**
 * Created by ihsan on 17/01/17.
 */
public class TTLTest {
    public static void main(String[] args) {
        int ttlTime = 2000; //msecs
        String mapName = "SomeMap";
        ClientConfig config = new ClientConfig();
        config.addNearCacheConfig(new NearCacheConfig(mapName));
        HazelcastInstance client = HazelcastClient.newHazelcastClient(config);
        final IMap<Object, Object> map = client.getMap(mapName);
        map.put(1, "value1", ttlTime, TimeUnit.MILLISECONDS);

        Object firstValue = map.get(1);
        System.out.println("First value:" + firstValue);
        assertEquals("value1", firstValue);

        try {
            Thread.sleep(ttlTime);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }

        // trigger eviction
        Object secondValue = map.get(1);
        System.out.println("Second value:" + secondValue);

        HazelcastTestSupport.assertTrueAllTheTime(new AssertTask() {
            @Override
            public void run()
                    throws Exception {
                Object value = map.get(1);
                System.out.println("Value is " + value);
                assertEquals("value1", value);
            }
        }, 20);

        System.out.println("Test finished");

        HazelcastClient.shutdownAll();
    }
}
```

ihsandemir:
More info #9669

jerrinot:
hi @ihsandemir, do you know the root cause? Are you working on this?

PR #9687, opened 2017-01-18
Added connected server version aware near cache invalidation listener…

ahmetmircik:
… to deal with client compatibility issues

Fixes two issues related with invalidation listeners backward compatibility:
- 3.7 client should receive invalidations from 3.8 member as if they are coming from 3.7 member. To fix this [canSendInvalidation method](https://github.com/hazelcast/hazelcast/pull/9687/files#diff-5b97be3d065d7c5265d7afb77c8b5b06R132) is added.
- 3.8 client should attach invalidation listeners to 3.7 member as if it is a 3.7 client.    To fix this `ConnectedMemberVersionAwareNearCacheEventHandler` is introduced.

fixes https://github.com/hazelcast/hazelcast/issues/9674


ahmetmircik:
closing this PR, it will be handled in a different way.

PR #9690, opened 2017-01-18
Remove ThreadLocalRandom from ClientReplicatedMapProxy

Donnerbart:
We use the random generator a single time, so it's not worth to
initialize the ThreadLocal object here. It's also reducing the
possibility for a classloader leakage until we properly fix this.

Part of https://github.com/hazelcast/hazelcast/issues/9650


Donnerbart:
The fix should not be necessary anymore after https://github.com/hazelcast/hazelcast/pull/9691 is merged, but I nevertheless don't think we need a `ThreadLocal` random generator here. This still causes more overhead than a local variable, unless you create hundreds of replicated maps.

Donnerbart:
After a discussion with Jaromir we decided to not change this code. Since `UuidUtil` will be using the same `Random` instance as the `ThreadLocalRandom` now, we might even safe the a new `Random` instance here.

PR #9691, opened 2017-01-18
Fixed classloader leakage through ThreadLocal<>

Donnerbart:
Replaced use of `ThreadLocal<Random>` subclass in `ThreadLocalRandom` (used
to override `initialValue()`) with a method that explicitly gets the
thread local value and sets it if it's null.

This avoids some nasty classloader issues that show up particularly in
web applications, where this anonymous subclass is holding a reference
to the entire web app classloader, which is not released until the
thread exits.

This basically replaces `ThreadLocalRandom` as backport of JDK 7 with a
simple holder class for leakage safe `ThreadLocal<Random>` and
`ThreadLocal<SecureRandom>` fields.

Made `UuidUtil` use of the new `ThreadLocalRandom` instead of its own
`ThreadLocal` fields.

See https://jira.mongodb.org/browse/JAVA-1164
Fixes https://github.com/hazelcast/hazelcast/issues/9650


Donnerbart:
Added fixes for `HazelcastParallelClassRunner` and `SampleableConcurrentHashMap`.

hazelcast/src/test/java/com/hazelcast/test/HazelcastParallelClassRunner.java
```diff
@@ -149,12 +149,7 @@ public void run() {
 
         private final Properties globalProperties;
 
-        private final ThreadLocal<Properties> localProperties = new InheritableThreadLocal<Properties>() {
-            @Override
-            protected Properties initialValue() {
-                return init(new Properties());
-            }
-        };
+        private final ThreadLocal<Properties> localProperties = new ThreadLocal<Properties>();
```
jerrinot: I will have a deeper look tomorrow, but there is nothing wrong with the TL here. it's a test code only, it's not something you deploy & undeploy. so here we should prefer simplicity.

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
```diff
@@ -16,199 +16,60 @@
...
-            int bits = next(2);
-            long half = n >>> 1;
-            long nextn = ((bits & 2) == 0) ? half : n - half;
-            if ((bits & 1) == 0)
-                offset += n - nextn;
-            n = nextn;
+    public static Random current() {
+        Random random = THREAD_LOCAL_RANDOM.get();
+        if (random == null) {
+            random = new Random(seedUniquifier() ^ System.nanoTime());
```
mdogan: why not just `new Random()`? why do we care the implementation behind?

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
```diff
@@ -16,199 +16,60 @@
...
- * @since 1.7
- * @author Doug Lea
+ * A random number generator isolated to the current thread.
  */
-public class ThreadLocalRandom extends Random {
-    // same constants as Random, but must be redeclared because private
-    private static final long multiplier = 0x5DEECE66DL;
-    private static final long addend = 0xBL;
-    private static final long mask = (1L << 48) - 1;
+public final class ThreadLocalRandom {
```
mdogan: This is not a `Random` class anymore. I think this need a better name to reflect it.

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
Donnerbart: I just moved this from the `UuidUtil`, so we would have to ask whoever wrote that originally. Since I re-use the new `ThreadLocalRandom` in the `UuidUtil`, I didn't want to change the Random initialization.

I think the explanation was that some JDKs initialize the random with just `System.currentMillis()`, not `System.nanoTime()`. This is still a kind of backport of ThreadLocalRandom, so maybe it's valid to enforce the initialization we prefer.

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
Donnerbart: I agree, but I don't really had a better name coming to my mind. I'd like to rename it as well, since we don't get a nice auto-import with the name duplication with the JDK class.

hazelcast/src/test/java/com/hazelcast/test/HazelcastParallelClassRunner.java
Donnerbart: Hmm, you might be right that we don't need it here. But I don't like to have this pattern in the code anymore, so no one can pick it up accidentally.

Donnerbart: Actually this change seems to break the `SystemPropertiesPluginTest`, so it seems to break the `SystemProperties` handling for tests.

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
mdogan: I think that's the JDK issue if one is creating low quality random.

hazelcast/src/test/java/com/hazelcast/test/HazelcastParallelClassRunner.java
Donnerbart: I changed `new InheritableThreadLocal()` to `new ThreadLocal()` by accident, that caused the test failure.

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
Donnerbart: I added the original comment to the new class, so it's at least clear what the intent is.

hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandom.java
Donnerbart: I renamed the class to `ThreadLocalRandomProvider` and the methods to `get()` and `getSecure()`.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientReplicatedMapProxy.java
```diff
@@ -93,7 +93,7 @@ public ClientReplicatedMapProxy(String serviceName, String objectName) {
     @Override
     protected void onInitialize() {
         int partitionCount = getContext().getPartitionService().getPartitionCount();
-        targetPartitionId = ThreadLocalRandom.current().nextInt(partitionCount);
+        targetPartitionId = ThreadLocalRandomProvider.get().nextInt(partitionCount);
```
mdogan: I know your intention is not to change current behaviour, but `ClientReplicatedMapProxy.onInitialize()` doesn't need a thread local random at all. It's just called once.

mdogan: hmm you've already a PR for that: https://github.com/hazelcast/hazelcast/pull/9690/files

Donnerbart: I just closed the PR which would have removed that :D https://github.com/hazelcast/hazelcast/pull/9690

Jaromir and I just discussed this with the following outcome: Since `UuidUtil` uses the same `Random` instance now, we might profit from keeping `ThreadLocalRandomProvider`, since the `Random` instance might already be there (or maybe re-used later by a UUID creation). Does this make sense to you or did we miss something?

mdogan: makes sense. both fine with me.

PR #9699, opened 2017-01-19
[backport] added nativeMemoryConfig xml generator

### LINKED ISSUE #8825: Native memory config is not included in the result while generating an XML string from config

hasancelik:
bakcport of #9441 
fixes #8825


mesutcelik:
👍

Issue #9677, opened 2017-01-17
Client is not updating listener registrations in case of member restart with a new version 


ahmetmircik:
__Scenario:__
One 3.8 client is working with one 3.8 member. Then member is closed, 3.8 member jar is replaced with 3.7.4 member jar and member restarted. Client re-connects but listeners are not re-registered here: https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java#L136

__Maybe Solution:__
Compare members according to uuids instead of addresses here: https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java#L122

Issue #4290, opened 2014-12-12
ClientSplitBrainTest.testClientListeners_InSplitBrain


bilalyasar:
```
java.lang.AssertionError: id 0 expected:<3> but was:<0>
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.failNotEquals(Assert.java:743)
    at org.junit.Assert.assertEquals(Assert.java:118)
    at org.junit.Assert.assertEquals(Assert.java:555)
    at com.hazelcast.client.ClientSplitBrainTest$1.run(ClientSplitBrainTest.java:93)
    at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:239)
    at com.hazelcast.client.ClientSplitBrainTest.checkEventsEventually(ClientSplitBrainTest.java:90)
    at com.hazelcast.client.ClientSplitBrainTest.testClientListeners_InSplitBrain(ClientSplitBrainTest.java:84)
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-OpenJDK6/com.hazelcast$hazelcast-client/83/testReport/junit/com.hazelcast.client/ClientSplitBrainTest/testClientListeners_InSplitBrain/

bilalyasar:
again failed: https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast-client/465/testReport/junit/com.hazelcast.client/ClientSplitBrainTest/testClientListeners_InSplitBrain/

bilalyasar:
failed,  same exception
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/470/com.hazelcast$hazelcast-client/testReport/com.hazelcast.client/ClientSplitBrainTest/testClientListeners_InSplitBrain/

bilalyasar:
```
Running com.hazelcast.client.ClientSplitBrainTest
Build timed out (after 360 minutes). Marking the build as aborted.
```
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-nightly/472/console

hasancelik:
```
java.lang.AssertionError: listener id 2
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
[...]

PR #9700, opened 2017-01-19
Listeners session is bound to connection lifecycle

sancar:
In following pr's listener lifecycle diverged so that eventough
client connection breaks server trying to find another connection
to same client address to be able to continue to send events.

https://github.com/hazelcast/hazelcast/pull/8848

This pr restores that logic. Listeners are alive as long as the
connection is alive. When connection is dropped, listeners registaration
needs to be done to that member again. This will be done on client
by ClientListenerService.

This will simplify ClientListenerService further. It does not need to
look at the unregisteredMembers that is coming with authentication response.

To preserve one bug fix in https://github.com/hazelcast/hazelcast/pull/8848.
SmartListenerService converted to listen for connections rather than Members.
Other than that ClientSmartListenerService and ClientNonSmartListenerService
reverted to changes before https://github.com/hazelcast/hazelcast/pull/8848.

With that change server do not need to find another alive connection
when returning response or sending event to client. The connection
that request came is the only one that server needs to use. If it is
dead, response/event should be dropped. `findSendConnection` removed.

endpoint.removeDestroyAction(parameters.registrationId) added
unnecessarily to RemoveDistributedObjectListenerMessageTask. Base class
already handles that. Removed back.
It was added here https://github.com/hazelcast/hazelcast/pull/8848

endpointManager.removeEndpoint(connection) usages changed with
connection.close. Instead, connection.close will trigger
connectionRemoved, Necessary cleanups will be done in
ClientEngineImpl.connectionRemoved

endpointManager.removeEndpoint with delay is removed since the delay
logic is already moved to ClientEngineImpl.

isAlive boolean added to mock connections to prevent multiple close
attempt to same connection. The original ones already have that flag.

MockConnectionManager.destoryConnection renamed to onClose to make it
similar with ConnectionManager.

fixes https://github.com/hazelcast/hazelcast/issues/9714 
fixes https://github.com/hazelcast/hazelcast/issues/9677
fixes https://github.com/hazelcast/hazelcast/issues/4290
fixes https://github.com/hazelcast/hazelcast/issues/9803


sancar:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
```diff
@@ -88,364 +80,131 @@ public String call() {
...
             public void run() {
-                registrationExecutor.submit(new Runnable() {
-                    @Override
-                    public void run() {
-                        ensureConnectionsToAllServers();
-                    }
-                });
+                Collection<Member> memberList = clientClusterService.getMemberList();
+                for (Member member : memberList) {
+                    clientConnectionManager.getOrTriggerConnect(member.getAddress(), false);
```
sancar: There is a race of we call it with `true`. Only ClusterListenerSupport should call it with ownerConnection=true. In this case, there is no problem. If connection is established as owner to this address, it will return immediately without making a remote call, since it is already connected.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
```diff
@@ -88,364 +80,131 @@ public String call() {
...
-                        logger.warning("Listener " + registrationKey + " can not be added to new member " + member
-                                        + " rescheduling the registration in " + SMART_LISTENER_MEMBER_ADDED_RESCHEDULE_TIME
-                                + " msecs", e);
-                        client.getClientExecutionService().schedule(new Runnable() {
-                            @Override
-                            public void run() {
-                                memberAdded(membershipEvent);
-                            }
-                        }, SMART_LISTENER_MEMBER_ADDED_RESCHEDULE_TIME, TimeUnit.MILLISECONDS);
+                        logger.warning("Listener " + registrationKey + " can not be added to new connection: "
```
sancar: When connection is disconnected and reconnected, it will be handled in connectionAdded. 
But invocation cancelled because heartbeat timed out is a problem here.  I left that out, because I was thinking it was not handled already. But now looking at the code again, there was a retry. I need to work on this further. 

In your implementation, ClientNonSmartService does not handle this case but ClientSmartService handles it. This is why I got confused. I will try to find a solution for both. And note that all tests are passing. which means there is no test that reproduces that scenario. Given that, it might be better to handle this in a follow-up pr, instead of this one.

hazelcast-client/src/test/java/com/hazelcast/client/listeners/AbstractListenersOnReconnectTest.java
```diff
@@ -393,14 +484,18 @@ public void run() throws Exception {
                 if (smartRouting) {
                     Collection<Member> members = clientInstanceImpl.getClientClusterService().getMemberList();
                     for (ClientEventRegistration registration : registrations) {
-                        Member registeredSubscriber = registration.getSubscriber();
+                        Connection registeredSubscriber = registration.getSubscriber();
+                        boolean contains = false;
+                        for (Member member : members) {
+                            contains |= registeredSubscriber.getEndPoint().equals(member.getAddress());
```
sancar: There is not change in test behaviour. 
Since `registration.getSubscriber()` returning Connection instead of Member now, I had to change the verification code.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
sancar: Updated the pr so that Heartbeat failures are handled in ClientSmartListenerService.
For ClientNonSmartListenerService I skipped it as it was before, because when heartbeat failed for single connection it will be closed anyway. And when connection is added back it will register invocation.

PR #9701, opened 2017-01-19
Listeners session is bound to connection lifecycle

sancar:
In following pr's listener lifecycle diverged so that eventough
client connection breaks server trying to find another connection
to same client address to be able to continue to send events.

https://github.com/hazelcast/hazelcast/pull/8848

This pr restores that logic. Listeners are alive as long as the
connection is alive. When connection is dropped, listeners registaration
needs to be done to that member again. This will be done on client
by ClientListenerService.

This will simplify ClientListenerService further. It does not need to
look at the unregisteredMembers that is coming with authentication response.

To preserve one bug fix in https://github.com/hazelcast/hazelcast/pull/8848.
SmartListenerService converted to listen for connections rather than Members.
Other than that ClientSmartListenerService and ClientNonSmartListenerService
reverted to changes before https://github.com/hazelcast/hazelcast/pull/8848.

With that change server do not need to find another alive connection
when returning response or sending event to client. The connection
that request came is the only one that server needs to use. If it is
dead, response/event should be dropped. `findSendConnection` removed.

endpoint.removeDestroyAction(parameters.registrationId) added
unnecessarily to RemoveDistributedObjectListenerMessageTask. Base class
already handles that. Removed back.
It was added here https://github.com/hazelcast/hazelcast/pull/8848

endpointManager.removeEndpoint(connection) usages changed with
connection.close. Instead, connection.close will trigger
connectionRemoved, Necessary cleanups will be done in
ClientEngineImpl.connectionRemoved

endpointManager.removeEndpoint with delay is removed since the delay
logic is already moved to ClientEngineImpl.

isAlive boolean added to mock connections to prevent multiple close
attempt to same connection. The original ones already have that flag.

MockConnectionManager.destoryConnection renamed to onClose to make it
similar with ConnectionManager.

fixes https://github.com/hazelcast/hazelcast/issues/9714 
fixes https://github.com/hazelcast/hazelcast/issues/9677
fixes https://github.com/hazelcast/hazelcast/issues/4290
fixes https://github.com/hazelcast/hazelcast/issues/9803


sancar:
run-lab-run

hasancelik:
run-lab-run

sancar:
ClientNearCacheBatchInvalidationTest.testHigherBatchSize_shouldNotCauseAnyInvalidation_onClient

```
java.lang.AssertionError: expected:<1000> but was:<999>
	at com.hazelcast.client.map.impl.nearcache.ClientNearCacheBatchInvalidationTest$1.run(ClientNearCacheBatchInvalidationTest.java:218)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:967)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:984)
	at com.hazelcast.client.map.impl.nearcache.ClientNearCacheBatchInvalidationTest.assertNearCacheSizeEventually(ClientNearCacheBatchInvalidationTest.java:213)
	at com.hazelcast.client.map.impl.nearcache.ClientNearCacheBatchInvalidationTest.testHigherBatchSize_shouldNotCauseAnyInvalidation_onClient(ClientNearCacheBatchInvalidationTest.java:123)
```

The failing test seems unrelated to me although I am not hundred percent sure. Leaving it here as reference.

sancar:
run-lab-run

hazelcast-client/src/test/java/com/hazelcast/client/ClientServiceTest.java
```diff
@@ -436,4 +436,102 @@ public void stateChanged(LifecycleEvent event) {
...
+                latch.countDown();
+            }
+        });
+
+        config.addListenerConfig(listenerConfig);
+
+        HazelcastInstance instance = hazelcastFactory.newHazelcastInstance(config);
+        HazelcastInstance client = hazelcastFactory.newHazelcastClient();
+        client.shutdown();
+        assertOpenEventually(latch);
```
ihsandemir: It may be better to use different latches for connect and disconnect. Assume there is a mistake in code and clientConnected is fired twice and clientDisconnected is not fired at all. The test will pass but this is not what we want. you may use the event count logic as you used in the following tests.

hazelcast/src/main/java/com/hazelcast/core/ClientListener.java
```diff
@@ -19,22 +19,22 @@
...
+ * The ClientListener provides the ability to listen to clients connecting and disconnecting from the member.
  *
  * @see Client
  * @see ClientService#addClientListener(ClientListener)
  */
 public interface ClientListener extends EventListener {
 
     /**
-     * Invoked when a new client is connected.
+     * Invoked when a client is connected.
```
ihsandemir: Will it be one if client uses 3 connections to the member? Do we need this level of documentation?

ihsandemir: It may worth mentioning this API re-phrasing as a fix in the release notes.

PR #9704, opened 2017-01-19
Fix wrong scheduled task count caused by backup replicas

### LINKED ISSUE #9694: Wrong getAllScheduledCount on Scheduled Executor Service
@Danny-Hazelcast reported the following, which I managed to reproduce.

When scheduling on random partitions and trying to get all scheduled tasks from the service, the number of tasks is wrong. The returned figure seems to match num_of_tasks * backup_count.

Reproduce
```
@Test
    public void scheduleRandomPartitions_getAllScheduled()

        HazelcastInstance[] instances = createClusterWithCount(2);
        IScheduledExecutorService s = getScheduledExecutor(instances, "s");

        int expectedTotal = 11;
        for (int i=0; i < expectedTotal; i++) {
            s.schedule(new PlainCallableTask(), 0, SECONDS);
        }

        Map<Member, List<IScheduledFuture<Double>>> allScheduled = s.getAllScheduledFutures();

        int actualTotal = 0;
        for (Member member : allScheduled.keySet()) {
            actualTotal += allScheduled.get(member).size();
        }

        assertEquals(expectedTotal, actualTotal, 0);
    }
```

The above fails with:

```
java.lang.AssertionError: 
Expected :11.0
Actual   :22.0
 <Click to see difference>


	at org.junit.Assert.assertEquals(Assert.java:575)
	at org.junit.Assert.assertEquals(Assert.java:700)
	at com.hazelcast.scheduledexecutor.ScheduledExecutorServiceBasicTest.scheduleRandomPartitions_getAllScheduled(ScheduledExecutorServiceBasicTest.java:785)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
```

tkountis:
When scheduling tasks on random partitions, the getAllScheduled call
returns twice as many futures from the expected result. The reason
seems to be the backup replicas which are also returned, even though they are
just stashed placeholders having no future scheduled. To address this, we introduced
a flag, marking the main copy of the task as master upon initital scheduling or promotions
and using that when building the list of tasks to return to the API call.

Fixes #9694

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
+++ b/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
@@ -41,12 +41,12 @@ public class ScheduledTaskDescriptor implements IdentifiedDataSerializable {
 
     /**
      * Only accessed through a member lock or partition threads
-     * Used to identify which replica of the task is the master, to only return that instance
+     * Used to identify which replica of the task is the owner, to only return that instance
      * when {@link com.hazelcast.scheduledexecutor.impl.operations.GetAllScheduledOperation} operation is triggered.
-     * This flag is set to true only on ititial scheduling of a task, and on after a promotion (stashed or migration),
+     * This flag is set to true only on initial scheduling of a task, and on after a promotion (stashed or migration),
      * in the latter case the other replicas get disposed.
      */
-    private transient boolean masterReplica;
+    private transient boolean isTaskOwner;
 
     /**
      * SPMC (see. Member owned tasks)
@@ -87,12 +87,12 @@ public class ScheduledTaskDescriptor implements IdentifiedDataSerializable {
         return definition;
     }
 
-    public boolean isMasterReplica() {
-        return masterReplica;
+    public boolean isTaskOwner() {
+        return isTaskOwner;
     }
 
-    public void setMasterReplica(boolean masterReplica) {
-        this.masterReplica = masterReplica;
+    void setTaskOwner(boolean taskOwner) {
+        this.isTaskOwner = taskOwner;
     }
 
     ScheduledTaskStatisticsImpl getStatsSnapshot() {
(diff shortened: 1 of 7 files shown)


jerrinot:
When there is a migration rolllback then the source side will always set itself as a master replica. is this the right thing to do?

tkountis:
@jerrinot thanks for the comment, but to be honest I can't say I understand the concern :/ 
In a migration rollback, the source side will take ownership of the task and re-schedule it once again, meaning it is going to be the master replica. Am I understanding something wrong from your comment ? What would you expect the correct behaviour to be ?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskDescriptor.java
```diff
@@ -40,6 +40,15 @@
     private ScheduledFuture<?> future;
 
     /**
+     * Only accessed through a member lock or partition threads
+     * Used to identify which replica of the task is the master, to only return that instance
+     * when {@link com.hazelcast.scheduledexecutor.impl.operations.GetAllScheduledOperation} operation is triggered.
+     * This flag is set to true only on ititial scheduling of a task, and on after a promotion (stashed or migration),
+     * in the latter case the other replicas get disposed.
+     */
+    private transient boolean masterReplica;
```
mmedenjak: Perhaps for clarity rename to isPartitionOwner since "master" is a different thing in the cluster.

tkountis:
@mmedenjak addressed both comments ;)

PR #9705, opened 2017-01-19
Added ThreadLocalLeakTest to test Hazelcast for ThreadLocal leaks

Donnerbart:
For some reason the `UserCodeDeploymentClassLoader` doesn't seem to create the leakage. I didn't have a look yet why, just disabled this one in the test for now.

Test for https://github.com/hazelcast/hazelcast/issues/9650

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
@@ -16,6 +16,7 @@
 
 package com.hazelcast.test;
 
+import classloading.ThreadLocalLeakTestUtils;
 import com.hazelcast.client.impl.ClientEngineImpl;
 import com.hazelcast.cluster.ClusterState;
 import com.hazelcast.config.Config;
@@ -217,6 +218,10 @@ public abstract class HazelcastTestSupport {
     // #####################################
     // ########## generic utility ##########
     // #####################################
+
+    protected void checkThreadLocalsForLeaks() throws Exception {
+        ThreadLocalLeakTestUtils.checkThreadLocalsForLeaks(getClass().getClassLoader());
+    }
 
     public static void ignore(Throwable ignored) {
     }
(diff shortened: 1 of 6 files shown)


Donnerbart:
For some reason the `OWN` classloader doesn't work on Jenkins, since there are some Mockito leftovers. But I don't use Mockito in that test. Maybe it's from another test of that JVM? For now I will stick with the `FILTERING` classloader which was anyway the original cause of the issue.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
I fixed the `OWN`  classloader by allowing `org.mockito.configuration.DefaultMockitoConfiguration` and
`org.mockito.internal.progress.MockingProgressImpl` to remain in `ThreadLocal` in the test. Maybe I can automate this by scanning the `ThreadLocal` map before the application is started, but the static approach was faster to implement. The bonus is a test with a real-world classloader.

hazelcast/src/test/java/com/hazelcast/test/HazelcastTestSupport.java
```diff
@@ -218,6 +219,10 @@ public static Packet toPacket(HazelcastInstance local, HazelcastInstance remote,
     // ########## generic utility ##########
     // #####################################
 
+    protected void checkThreadLocalsForLeaks() throws Exception {
+        ThreadLocalLeakTestUtils.checkThreadLocalsForLeaks(getClass().getClassLoader());
```
Donnerbart: Is this the correct classloader to use in normal tests?

Donnerbart:
After a short discussion it is expected that the `UserCodeDeploymentClassLoader` will not create a detectable leak in this scenario. It's due to a very conservative checking that the leakage was produced by the exact same classloader which is given to the `checkThreadLocalsForLeaks()` method. But the `UserCodeDeploymentClassLoader` doesn't define classes on its own, it just delegates. So there is no detectable leakage. I just removed it from the test for now.

Donnerbart:
run-lab-run

PR #9711, opened 2017-01-20
ClientInvocation minor simplification

pveentjer:
Instead of accesing the properties; looking up a value by key. And then
parsing the value to get some constant; for every invocation. This is
now done once and read as a constant.


pveentjer:
I'm fine if we move this cleanup/performance improvement to 3.9.

mdogan:
run-lab-run

PR #9713, opened 2017-01-20
Ability to register user exception types to ClientExceptionFactory

viliam-durina:
* Make `ClientExceptionFactory.register()` method public
* Removed duplicately added exception classes under different codes and added checks to prevent that in future
* Moved `ClientExceptionFactory` creation from `ClientInvocationServiceSupport` to `HazelcastClientInstanceImpl`

Functionality is needed for Jet so that our exception is not repported as `UndefinedErrorCodeException`

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/client/protocol/ClientExceptionFactoryTestSimple.java
+++ b/hazelcast/src/test/java/com/hazelcast/client/protocol/ClientExceptionFactoryTestSimple.java
@@ -1,0 +1,64 @@
+package com.hazelcast.client.protocol;
+
+import com.hazelcast.client.impl.protocol.ClientExceptionFactory;
+import com.hazelcast.client.impl.protocol.ClientExceptionFactory.ExceptionFactory;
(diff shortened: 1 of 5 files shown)


hazelcast/src/test/java/com/hazelcast/client/protocol/ClientExceptionFactoryTestSimple.java
```diff
@@ -0,0 +1,78 @@
...
+public class ClientExceptionFactoryTestSimple {
+
+    private ClientExceptionFactory exceptionFactory = new ClientExceptionFactory(false);
+
+    private static class MyException extends Exception {}
+
+    @Test
+    public void testDuplicateErrorCode() {
+        try {
+            exceptionFactory.register(ClientProtocolErrorCodes.ARRAY_INDEX_OUT_OF_BOUNDS,
```
cangencer: it would be better to use `ExpectedException` `Rule` here and in the other tests, rather than try catch blocks

sancar:
This is actually fixing a defect since it is removing the duplicates. Added Defect Label.

Issue #9712, opened 2017-01-20
Cache-local-entries option is invalid for client-side NearCache configuration


vbekiaris:
An `IllegalArgumentException` should be thrown when a hazelcast client starts with `cache-local-entries` option set in its configuration.

PR #8970, opened 2016-09-23
Deprecated <cache-local-entries/> in client config

Donnerbart:
Deprecated `<cache-local-entries/>` in client config, since it has no effect on clients

* Renamed `MapConfigValidator` to `ConfigValidator`
* Moved `ConfigValidator` to a more general internal package
* Added deprecation to `hazelcast-client-config-3.8.xsd`
* Added `NearCacheConfig` validation to `ConfigValidator`

PR #9716, opened 2017-01-20
Log warning on invalid client-side near cache config

vbekiaris:
Log warning when client-side near-cache config includes invalid option cache-local-entries option 

Reasoning: `cache-local-entries` option in client-side `NearCacheConfig` was silently ignored in previous releases. Throwing an exception was introduced in #8970 thus an application upgraded from 3.7 or earlier to 3.8 with an offending client-side configuration would fail on startup with an `IllegalArgumentException` whereas it was previously silently working. With this PR a warning will be logged in 3.8 to indicate erroneous usage of this configuration option and fail-fast behavior will be implemented in 3.9 (#9712).

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/config/ConfigValidator.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/config/ConfigValidator.java
@@ -66,8 +66,8 @@ public final class ConfigValidator {
         checkEvictionConfig(nearCacheConfig.getEvictionConfig(), true);
 
         if (isClient && nearCacheConfig.isCacheLocalEntries()) {
-            throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


Donnerbart:
Hmm, that test looks a bit fragile. I think we could just `@Ignore` the test until we restore the exception (we could mention that in the ignore verification text).

mdogan:
@vbekiaris: Can you explain in description what is the reasoning behind this change? Why warn logging instead of throwing exception?

vbekiaris:
@mdogan I edited PR description to include reasoning for this change

vbekiaris:
@Donnerbart restored the original test and `@Ignore`d it until #9712 is fixed

vbekiaris:
run-lab-run

PR #9737, opened 2017-01-23
Add WAN javadoc and fix wan sync counters

mmedenjak:
- the AbstractWanPublisher.counterMap was being set after putting the wan events in the staging queue. The striped executor might have already processed some events causing a wrong count or NPE on AbstractWanPublisher.removeReplicationEvent. The fix first sets the counters before adding items into the staging queue, thus preventing the executor from observing events that are not included in the counterMap
- while checking if a backup event should be dropped in AbstractWanPublisher.isEventDroppingNeeded, the if statement allowed a backup event to be dropped even if the backup counter hasn't yet reached the queue size but the currentElementCount counter has. The if statement has been fixed to drop the backup event only if the backup counter has reached the queue size, regardless of the currentElementCount
- add javadoc for WAN related functionalities

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/WanPublisherConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/WanPublisherConfig.java
@@ -48,10 +48,27 @@ public class WanPublisherConfig implements DataSerializable {
         return this;
     }
 
+    /**
+     * Get the capacity of the queue for WAN replication events. IMap, ICache, normal and backup events count against
+     * the queue capacity separately. When the queue capacity is reached, backup events are dropped while normal
+     * replication events behave as determined by the {@link #getQueueFullBehavior()}.
+     * The default queue size for replication queues is {@value #DEFAULT_QUEUE_CAPACITY}.
+     *
+     * @return the queue capacity
+     */
     public int getQueueCapacity() {
         return queueCapacity;
     }
 
+    /**
+     * Set the capacity of the queue for WAN replication events. IMap, ICache, normal and backup events count against
+     * the queue capacity separately. When the queue capacity is reached, backup events are dropped while normal
+     * replication events behave as determined by the {@link #getQueueFullBehavior()}.
+     * The default queue size for replication queues is {@value #DEFAULT_QUEUE_CAPACITY}.
+     *
+     * @param queueCapacity the queue capacity
+     * @return this configuration
+     */
     public WanPublisherConfig setQueueCapacity(int queueCapacity) {
         this.queueCapacity = queueCapacity;
         return this;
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/config/WanPublisherConfig.java
```diff
@@ -48,10 +48,27 @@ public WanPublisherConfig setGroupName(String groupName) {
         return this;
     }
 
+    /**
+     * Set the capacity of the queue for WAN replication events. IMap, ICache, normal and backup events count against
```
emrahkocaman: Should be "Get the capacity...".

hazelcast/src/main/java/com/hazelcast/config/WanReplicationRef.java
```diff
@@ -81,15 +81,35 @@ public WanReplicationRef setMergePolicy(String mergePolicy) {
         return this;
     }
 
+    /**
+     * Add class name implementing the CacheWanEventFilter for filtering WAN replication events.
```
emrahkocaman: I think it's better like "... CacheWanEventFilter or MapWanEventFilter..."
Same for `getFilters` and `setFilters` methods' javadoc.

mmedenjak: Changed.

hazelcast/src/main/java/com/hazelcast/config/WanPublisherConfig.java
mmedenjak: Changed.

hazelcast/src/main/java/com/hazelcast/config/WanPublisherConfig.java
```diff
@@ -48,10 +48,27 @@ public WanPublisherConfig setGroupName(String groupName) {
         return this;
     }
 
+    /**
+     * Get the capacity of the queue for WAN replication events. IMap, ICache, normal and backup events count against
+     * the queue capacity separately. When the queue capacity is reached, backup events are dropped while normal
+     * replication events behave as determined by the {@link #getQueueFullBehavior()}.
+     * The default queue size for replication queues is 10000.
```
Donnerbart: You could use `{@value #DEFAULT_QUEUE_CAPACITY}` here, so you don't have a copy&paste value of the source code in the JavaDoc.

hazelcast/src/main/java/com/hazelcast/config/WanPublisherConfig.java
```diff
@@ -48,10 +48,27 @@ public WanPublisherConfig setGroupName(String groupName) {
...
+     */
     public int getQueueCapacity() {
         return queueCapacity;
     }
 
+    /**
+     * Set the capacity of the queue for WAN replication events. IMap, ICache, normal and backup events count against
+     * the queue capacity separately. When the queue capacity is reached, backup events are dropped while normal
+     * replication events behave as determined by the {@link #getQueueFullBehavior()}.
+     * The default queue size for replication queues is 10000.
```
Donnerbart: You could use `{@value #DEFAULT_QUEUE_CAPACITY}` here, so you don't have a copy&paste value of the source code in the JavaDoc.

PR #9738, opened 2017-01-23
DefaultWriteBehindProcessor improvement for partial failures at storeAll/deleteAll

### LINKED ISSUE #9733: DefaultWriteBehindProcessor: do not add whole batchMap in failureList in case of failure
when we use storeAll/deleteAll methods in MapStore interface, it is possible that underlying implementation has stored/deleted some entries from the supplied batch and then gets an exception on some particular entry in the middle of the batch. In order to handle this scenario properly the MapStore could remove from the supplied batch entries which were properly stored in the underlying persistent store. The only thing to improve in the DefaultWiteBehindProcessor is to properly synchronise the map passed to processBatch method with the initial batchMap (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java#L243):

```
            @Override
            public boolean run() throws Exception {
                callBeforeStoreListeners(batchMap.values());
                final Map map = convertToObject(batchMap);
                try {
                    final boolean result = operationType.processBatch(map, mapStore);
                } catch (Exception ex) {
                    // leave in batchMap only entries which were left in the map...
                   throw ex;
                }
                callAfterStoreListeners(batchMap.values());
                return result;
            }
```
I can provide PR for this, if you wish.

Thanks, Denis.

dsukhoroslov:
- added agreed piece of code in the DefaultWriteBihindProcessor;
- added a test case in the WriteBehindFailAndRetryTest;

Fixes #9733

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
@@ -249,8 +249,8 @@ class DefaultWriteBehindProcessor extends AbstractWriteBehindProcessor<DelayedEn
                     result = operationType.processBatch(map, mapStore);
                 } catch (Exception ex) {
                     Iterator<Object> keys = batchMap.keySet().iterator();
-                    for (Object key = null; keys.hasNext(); key = keys.next()) {
-                        if (!map.containsKey(key)) {
+                    while (keys.hasNext()) {
+                        if (!map.containsKey(DefaultWriteBehindProcessor.this.toObject(keys.next()))) {
                             keys.remove();
                         }
                     }
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
@@ -25,6 +25,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -243,7 +244,18 @@ class DefaultWriteBehindProcessor extends AbstractWriteBehindProcessor<DelayedEn
             public boolean run() throws Exception {
                 callBeforeStoreListeners(batchMap.values());
                 final Map map = convertToObject(batchMap);
-                final boolean result = operationType.processBatch(map, mapStore);
+                boolean result;
+                try {
+                    result = operationType.processBatch(map, mapStore);
+                } catch (Exception ex) {
+                    Iterator<Object> keys = batchMap.keySet().iterator();
+                    for (Object key = null; keys.hasNext(); key = keys.next()) {
+                        if (!map.containsKey(key)) {
+                            keys.remove();
+                        }
+                    }
+                    throw ex;
+                }
                 callAfterStoreListeners(batchMap.values());
                 return result;
             }
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
@@ -260,9 +260,6 @@ class DefaultWriteBehindProcessor extends AbstractWriteBehindProcessor<DelayedEn
                 return result;
             }
 
-            /**
-             * Call when store failed.
-             */
             @Override
             public List<DelayedEntry> failureList() {
                 failedDelayedEntries = new ArrayList<DelayedEntry>(batchMap.values().size());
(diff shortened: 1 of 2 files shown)


mdogan:
run-lab-run

mdogan:
@dsukhoroslov:
There's checkstyle failure;

>hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java:240:26: Anonymous inner class length is 33 lines (max allowed is 30).

You can test checkstyle on local. 
`mvn clean validate -P checkstyle`
See https://github.com/hazelcast/hazelcast#checkstyle-and-findbugs

dsukhoroslov:
@mdogan, I run it. The output is:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.15:checkstyle (default) on project hazelcast: An error has occurred in Checkstyle report generation. 
Failed during checkstyle execution: There are 6038 errors reported by Checkstyle 6.1.1 with C:\Work\HZ\git\hazelcast/checkstyle/checkstyle.xml ruleset. 
```

mdogan:
@dsukhoroslov: interesting, checkstyle itself failed. 
Anyway, only failure is the one above. anonymous inner class length issue. You can either refactor anonymous class to a named inner class or somehow reduce its length.

dsukhoroslov:
@mdogan, I see two possible fixes for this checkstyle failure:
1) introduce new static class implementing RetryTask in the same file, move code at lines 240..272 in it;
2) just remove unnecessary comment on failureList method (remove lines 263..265);

which one do you prefer? (I prefer the second one as it keeps close RetryTask implementations for single and batch store operations)

mdogan:
@dsukhoroslov, fine with me. that comment looks redundant.

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
```diff
@@ -243,14 +244,22 @@ private Map convertToObject(Map<Object, DelayedEntry> batchMap) {
...
                 final Map map = convertToObject(batchMap);
-                final boolean result = operationType.processBatch(map, mapStore);
+                boolean result;
+                try {
+                    result = operationType.processBatch(map, mapStore);
+                } catch (Exception ex) {
+                    Iterator<Object> keys = batchMap.keySet().iterator();
+                    for (Object key = null; keys.hasNext(); key = keys.next()) {
+                        if (!map.containsKey(key)) {
+                            keys.remove();
```
ahmetmircik: Seems `keys.next()` is not called before `keys.remove()` for the first iteration.  In that case key is null.

dsukhoroslov: hmm. Your observation looks correct. But then it should throw NPE at containsKey check, afaik. Ok, I'll trace it.

dsukhoroslov:
@ahmetmircik, fixed. The weird thing is that the test I added for this case has passed successfully from command line (mvn -Dtest=WriteBehindFailAndRetryTest test). Only when debugged test in IDE I found the batchMap contains binary data, and the test has failed properly..

ahmetmircik:
thanks it seems ok to me now, one last thing, can you please squash the commits into one and send a PR for master branch?

dsukhoroslov:
done, see #9747

ahmetmircik:
can you also please squash the commits of this PR as its master port? After that i will merge this PR.

ahmetmircik:
run-lab-run

ahmetmircik:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
```diff
@@ -243,14 +244,22 @@ private Map convertToObject(Map<Object, DelayedEntry> batchMap) {
             public boolean run() throws Exception {
                 callBeforeStoreListeners(batchMap.values());
                 final Map map = convertToObject(batchMap);
-                final boolean result = operationType.processBatch(map, mapStore);
+                boolean result;
+                try {
+                    result = operationType.processBatch(map, mapStore);
+                } catch (Exception ex) {
+                    Iterator<Object> keys = batchMap.keySet().iterator();
+                    while (keys.hasNext()) {
```
mdogan: This change might be risky for a maintenance release if somehow a user clears the batch `map` (or removes entries unintentionally) that's passed to `MapStore.storeAll()/deleteAll()`. At the moment in javadoc, we don't mention anything about modification of batch. For instance, I cannot think a reason why someone implements it this way but for following `storeAll()` implementation;

```java
public void storeAll(Map<K, V> map) {
   try {
      [...]
   } catch (Exception e) {
      [...]       
   } finally {
      map.clear();
   }
}
```

@jerrinot: what do you think?

jerrinot: I agree. It's a behavior change and we should avoid this in a maintenance release unless the old behavior is a clear bug. 
the [forward port is OK to me](https://github.com/hazelcast/hazelcast/pull/9747/files), assuming it also described the behavior in the contract of MapStore. 

@dsukhoroslov: the 3.8 is going to be out soon.

jerrinot:
@dsukhoroslov: I appreciate your effort, however it's unacceptable to change behavior like this in a maintanance release. 

Let's concentrate on the [PR into master](https://github.com/hazelcast/hazelcast/pull/9747/files).  That looks good to me, I believe this only missing bit is to describe the behaviour in the MapStore JavaDoc.

PR #9744, opened 2017-01-24
Fixes 3.7 client - 3.8 member invalidation compatibility 

ahmetmircik:
fixes https://github.com/hazelcast/hazelcast/issues/9674

This PR fixes two issues related with invalidation listeners backward compatibility:
- 3.7 client should receive invalidations from 3.8 member as if they are coming from 3.7 member. To fix this `canSendInvalidation` method.
- 3.8 client should attach invalidation listeners to 3.7 member as if it is a 3.7 client. To fix this `ConnectedServerVersionAwareNearCacheEventHandler` is introduced.

depends on https://github.com/hazelcast/hazelcast-client-protocol/pull/57


ahmetmircik:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientInternalCacheProxy.java
```diff
@@ -912,9 +925,66 @@ protected ICompletableFuture createCompletedFuture(Object value) {
...
+
+        @Override
+        public void beforeListenerRegister() {
+            legacyHandler.beforeListenerRegister();
+            repairingHandler.beforeListenerRegister();
+        }
+
+        @Override
+        public void onListenerRegister() {
+            supportsRepairableNearCache = supportsRepairableNearCache();
```
ahmetmircik: this handles listener re-registration in case of server version change. For ex. 3.8 client connected 3.7 server and later server version is changed to 3.8 or vice versa. In these cases appropriate listener should be registered.

PR #9747, opened 2017-01-24
DefaultWriteBehindProcessor improvement for partial failures at storeAll/deleteAll

### LINKED ISSUE #9733: DefaultWriteBehindProcessor: do not add whole batchMap in failureList in case of failure

dsukhoroslov:
Fixes #9733

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/MapStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/MapStore.java
@@ -51,6 +51,11 @@ public interface MapStore<K, V> extends MapLoader<K, V> {
      * storeAll is used when writeDelaySeconds is positive (write-behind).
      * If an exception is thrown, the entries will try to be stored one by one using the store() method.
      *
+     * Note: on the retry phase only entries left in the map will be stored one-by-one. In this way a
+     * MapStore implementation can handle partial storeAll cases when some entries were stored successfully
+     * before failure happens. Entries removed from the map will be not passed to subsequent call to
+     * store() method any more.
+     *
      * @param map map of entries to store
      */
     void storeAll(Map<K, V> map);
@@ -67,6 +72,11 @@ public interface MapStore<K, V> extends MapLoader<K, V> {
      * Deletes multiple entries from the store.
      * If an exception is thrown the entries will try to be deleted one by one using the delete() method.
      *
+     * Note: on the retry phase only entries left in the keys will be deleted one-by-one. In this way a
+     * MapStore implementation can handle partial deleteAll cases when some entries were deleted successfully
+     * before failure happens. Entries removed from the keys will be not passed to subsequent call to
+     * delete() method any more.
+
      * @param keys the keys of the entries to delete.
      */
     void deleteAll(Collection<K> keys);
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
@@ -25,6 +25,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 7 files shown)


ahmetmircik:
run-lab-run

ahmetmircik:
```	Test Result (1 failure / +1)
com.hazelcast.client.heartbeat.ClientHeartbeatTest.testClientEndpointsDelaySeconds_whenHeartbeatResumed```

**Comment by ahmetmircik** [comment]
run-lab-run

**Comment by dsukhoroslov** [comment]
@jerrinot, is the following javadoc ok?

```
    /**
     * Stores multiple entries. Implementation of this method can optimize the
     * store operation by storing all entries in one database connection.
     * storeAll is used when writeDelaySeconds is positive (write-behind).
     * If an exception is thrown, the entries will try to be stored one by one using the store() method.
     * 
     * Note: on the retry phase only entries left in the map will be storied one-by-one. In this way a 
     * MapStore implementation can handle partial storeAll cases when some entries were stored successfully
     * before failure happens. Entries removed from the map will be not passed to subsequent call to 
     * store() method any more.
     *
     * @param map map of entries to store
     */
    void storeAll(Map<K, V> map);

    /**
     * Deletes multiple entries from the store.
     * If an exception is thrown the entries will try to be deleted one by one using the delete() method.
     *
     * Note: on the retry phase only entries left in the map will be deleted one-by-one. In this way a 
     * MapStore implementation can handle partial deleteAll cases when some entries were deleted successfully
     * before failure happens. Entries removed from the map will be not passed to subsequent call to 
     * delete() method any more.

     * @param keys the keys of the entries to delete.
     */
    void deleteAll(Collection<K> keys);
```

**Comment by jerrinot** [comment]
@dsukhoroslov: looks good to me, many thanks!

**Comment by vbekiaris** [comment]
run-lab-run

File: hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java [hunk 1]
```diff
@@ -243,14 +244,22 @@ private Map convertToObject(Map<Object, DelayedEntry> batchMap) {
...
                 callBeforeStoreListeners(batchMap.values());
                 final Map map = convertToObject(batchMap);
-                final boolean result = operationType.processBatch(map, mapStore);
+                boolean result;
+                try {
+                    result = operationType.processBatch(map, mapStore);
+                } catch (Exception ex) {
+                    Iterator<Object> keys = batchMap.keySet().iterator();
+                    while (keys.hasNext()) {
+                        if (!map.containsKey(DefaultWriteBehindProcessor.this.toObject(keys.next()))) {
```
vbekiaris: `DefaultWriteBehindProcessor.this.` is redundant here

hazelcast/src/main/java/com/hazelcast/core/MapStore.java
```diff
@@ -50,6 +50,11 @@
      * store operation by storing all entries in one database connection.
      * storeAll is used when writeDelaySeconds is positive (write-behind).
      * If an exception is thrown, the entries will try to be stored one by one using the store() method.
+     * 
+     * Note: on the retry phase only entries left in the map will be storied one-by-one. In this way a
```
vbekiaris: typo: "stored" instead of "storied" (same below)

dsukhoroslov: fixed

hazelcast/src/main/java/com/hazelcast/map/impl/mapstore/writebehind/DefaultWriteBehindProcessor.java
dsukhoroslov: In my IDE (Eclipse 4.5.1) remove of 'this' reference causes error: Cannot make a static reference to the non-static method toObject(Object) from the type AbstractWriteBehindProcessor<DelayedEntry>

dsukhoroslov: fixed

vbekiaris:
run-lab-run

vbekiaris:
run-lab-run

vbekiaris:
looks good now, can you please squash your commits and update the PR?

jerrinot:
@dsukhoroslov: I can still see 6 distinct commits. 
GitHub offers to squash commit on merging, but I guess it would remove you as an author and I don't want to do that.

dsukhoroslov:
@jerrinot could you merge it as it is? I don't care about the author name, but this squash op.. don't want waste time with it any more, sorry.

jerrinot:
I did not squash it, but rebased on the master -> your are still the author of the changeset, but the ugly merge commit is gone. see https://github.com/hazelcast/hazelcast/commits/master

thanks again for this contribution!

dsukhoroslov:
thanks :)

Issue #9037, opened 2016-10-03
IndexSplitBrainTest.testIndexDoesNotReturnStaleResultsAfterSplit


hasancelik:
```
java.lang.AssertionError: entry should be in map2 before split
    at org.junit.Assert.fail(Assert.java:88)
    at org.junit.Assert.assertTrue(Assert.java:41)
    at org.junit.Assert.assertNotNull(Assert.java:712)
    at com.hazelcast.query.impl.IndexSplitBrainTest.testIndexDoesNotReturnStaleResultsAfterSplit(IndexSplitBrainTest.java:74)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/1057/testReport/junit/com.hazelcast.query.impl/IndexSplitBrainTest/testIndexDoesNotReturnStaleResultsAfterSplit/

hasancelik:
```
java.lang.AssertionError: Cluster size is not correct. Expected: 2 Actual: 1
	at org.junit.Assert.fail(Assert.java:88)
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSize(HazelcastTestSupport.java:750)
	at com.hazelcast.test.HazelcastTestSupport$10.run(HazelcastTestSupport.java:769)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:905)
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSizeEventually(HazelcastTestSupport.java:765)
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSizeEventually(HazelcastTestSupport.java:760)
	at com.hazelcast.query.impl.IndexSplitBrainTest.testIndexDoesNotReturnStaleResultsAfterSplit(IndexSplitBrainTest.java:90)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-OracleJDK1.6/com.hazelcast$hazelcast/1083/testReport/junit/com.hazelcast.query.impl/IndexSplitBrainTest/testIndexDoesNotReturnStaleResultsAfterSplit/

hasancelik:
```
java.lang.AssertionError: CountDownLatch failed to complete within 120 seconds , count left: 1
	at org.junit.Assert.fail(Assert.java:88)
[...]

PR #9751, opened 2017-01-24
Implement IndexSplitBrainTest on top of SplitBrainTestSupport

vbekiaris:
* Implement test on top of `SplitBrainTestSupport` infrastructure
* Some assertions have been migrated to `assertTrueEventually` to avoid time sensitivity: as shown in #9037 due to loaded test environment assertions could fail in case migrations were not completed in time.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/query/impl/IndexSplitBrainTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/query/impl/IndexSplitBrainTest.java
@@ -10,14 +10,11 @@ import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.DataSerializable;
 import com.hazelcast.query.Predicate;
 import com.hazelcast.query.Predicates;
-import com.hazelcast.spi.properties.GroupProperty;
 import com.hazelcast.test.AssertTask;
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/query/impl/IndexSplitBrainTest.java
```diff
@@ -30,45 +27,65 @@
...
-
-        // create split 
-        closeConnectionBetween(h1, h2);
-        assertClusterSizeEventually(1, h1);
-        assertClusterSizeEventually(1, h2);
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run()
+                    throws Exception {
+                assertNotNull("Entry should exist in map2 before split", map2.get(key));
```
mdogan: `map.get()` after `map.put()` should return the put value immediately.

vbekiaris: 👍 when [this assertion failed](https://github.com/hazelcast/hazelcast/issues/9037#issue-180588154) there was no initial partition warmup in test code, so maybe the failure was due to initial partition assignments not being completed on the second instance in the time between `get` and `put`

Issue #9719, opened 2017-01-20
ClientHeartbeatTest testClientEndpointsDelaySeconds_whenHeartbeatResumed


jerrinot:
```
Error Message

expected:<2> but was:<4>
Stacktrace

java.lang.AssertionError: expected:<2> but was:<4>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.heartbeat.ClientHeartbeatTest$6.run(ClientHeartbeatTest.java:264)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueAllTheTime(HazelcastTestSupport.java:895)
	at com.hazelcast.client.heartbeat.ClientHeartbeatTest.testClientEndpointsDelaySeconds_whenHeartbeatResumed(ClientHeartbeatTest.java:261)
Standard Output

Finished Running Test: testInvocation_whenHeartbeatStopped in 7.244 seconds.
```

https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.maintenance-IbmJDK1.8/com.hazelcast$hazelcast-client/167/testReport/junit/com.hazelcast.client.heartbeat/ClientHeartbeatTest/testClientEndpointsDelaySeconds_whenHeartbeatResumed/

sancar:
@jerrinot I am moving the milestone to 3.7.6 since this failure is in maintenance

hasancelik:
@sancar failure from **master**
```
java.lang.AssertionError: expected:<2> but was:<3>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.heartbeat.ClientHeartbeatTest$9.run(ClientHeartbeatTest.java:321)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueAllTheTime(HazelcastTestSupport.java:951)
	at com.hazelcast.client.heartbeat.ClientHeartbeatTest.testClientEndpointsDelaySeconds_whenHeartbeatResumed(ClientHeartbeatTest.java:318)
```
[...]

PR #9754, opened 2017-01-25
Test fix testClientEndpointsDelaySeconds_whenHeartbeatResumed

sancar:
Test meant to wait  CLIENT_DISCONNECTED and CLIENT_CONNECTED.
But because events fired on an executor, in an unlucky run,
we got CLIENT_STARTED and CLIENT_CONNECTED before expected
events come. Although these events totally healthy, they
were incrementing the count value used for testing.

I changed the test so that the count value used for testing
is incremented only after first CLIENT_DISCONNECTED fired.

fixes https://github.com/hazelcast/hazelcast/issues/9719


sancar:
backport of https://github.com/hazelcast/hazelcast/pull/9755

PR #9759, opened 2017-01-25
added isHotBackupEnabled to HotRestartState

hasancelik:
after #9756 , ManCenter needs these changes to read *isHotBackupEnabled*

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
@@ -28,16 +28,24 @@ public class HotRestartStateImpl implements HotRestartState {
 
     private BackupTaskStatus backupTaskStatus;
 
+    private boolean isHotBackupEnabled;
+
     public HotRestartStateImpl() {
     }
 
-    public HotRestartStateImpl(BackupTaskStatus backupTaskStatus) {
+    public HotRestartStateImpl(BackupTaskStatus backupTaskStatus, boolean isHotBackupEnabled) {
         this.backupTaskStatus = backupTaskStatus;
+        this.isHotBackupEnabled = isHotBackupEnabled;
     }
 
     @Override
     public BackupTaskStatus getBackupTaskStatus() {
         return backupTaskStatus;
+    }
+
+    @Override
+    public boolean isHotBackupEnabled() {
+        return this.isHotBackupEnabled;
     }
 
     @Override
@@ -47,6 +55,7 @@ public class HotRestartStateImpl implements HotRestartState {
             root.add("backupTaskState", backupTaskStatus.getState().name());
             root.add("backupTaskCompleted", backupTaskStatus.getCompleted());
             root.add("backupTaskTotal", backupTaskStatus.getTotal());
+            root.add("isHotBackupEnabled", isHotBackupEnabled);
         }
         return root;
     }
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
```diff
@@ -28,25 +28,38 @@
 
     private BackupTaskStatus backupTaskStatus;
 
+    private boolean isHotBackupEnabled;
+
     public HotRestartStateImpl() {
     }
 
     public HotRestartStateImpl(BackupTaskStatus backupTaskStatus) {
```
mmedenjak: You can remove this constructor since I guess it is only used in tests now. Change the tests to use the new one.

hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
```diff
@@ -56,12 +69,14 @@ public void fromJson(JsonObject json) {
...
+        final boolean jsonIsHotBackupEnabled = JsonUtil.getBoolean(json, "isHotBackupEnabled");
         backupTaskStatus = jsonBackupTaskState != null ? new BackupTaskStatus(BackupTaskState.valueOf(jsonBackupTaskState),
                 jsonBackupTaskCompleted, jsonBackupTaskTotal) : null;
+        isHotBackupEnabled = jsonIsHotBackupEnabled;
     }
 
     @Override
     public String toString() {
-        return "HotRestartStateImpl{backupTaskStatus=" + backupTaskStatus + '}';
+        return "HotRestartStateImpl{backupTaskStatus=" + backupTaskStatus + "isHotBackupEnabled" + isHotBackupEnabled + '}';
```
mmedenjak: Just add a separator : `", isHotBackupEnabled"`

hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
```diff
@@ -56,12 +69,14 @@ public void fromJson(JsonObject json) {
         final String jsonBackupTaskState = getString(json, "backupTaskState", null);
         final int jsonBackupTaskCompleted = JsonUtil.getInt(json, "backupTaskCompleted", 0);
         final int jsonBackupTaskTotal = JsonUtil.getInt(json, "backupTaskTotal", 0);
+        final boolean jsonIsHotBackupEnabled = JsonUtil.getBoolean(json, "isHotBackupEnabled");
```
emrahkocaman: We need to use overloaded version of `JsonUtil#getBoolean` which takes a default value for the cases where "isHotBackupEnabled" is not present in JSON. This is needed to provide backward compatibility between newer versions of mancenter and older versions of hz.

hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
hasancelik: 👍

hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
hasancelik: 👍

hazelcast/src/main/java/com/hazelcast/monitor/impl/HotRestartStateImpl.java
```diff
@@ -56,12 +65,16 @@ public void fromJson(JsonObject json) {
         final String jsonBackupTaskState = getString(json, "backupTaskState", null);
         final int jsonBackupTaskCompleted = JsonUtil.getInt(json, "backupTaskCompleted", 0);
         final int jsonBackupTaskTotal = JsonUtil.getInt(json, "backupTaskTotal", 0);
+        final boolean jsonIsHotBackupEnabled = JsonUtil.getBoolean(json, "isHotBackupEnabled", false);
         backupTaskStatus = jsonBackupTaskState != null ? new BackupTaskStatus(BackupTaskState.valueOf(jsonBackupTaskState),
                 jsonBackupTaskCompleted, jsonBackupTaskTotal) : null;
+        isHotBackupEnabled = jsonIsHotBackupEnabled;
```
emrahkocaman: Why do we need `jsonIsHotBackupEnabled`, looks like it's not used anywhere? I think, we can get rid of `jsonIsHotBackupEnabled` and assign the parsed boolean directly to `isHotBackupEnabled`.

hasancelik: Done :+1:

PR #9741, opened 2017-01-24
Print a warning when Discovery SPI enabled but not configured on client

mdogan:
Printing a warning log instead of throwing exception to make similar
behavior with server side.

Fixes #9722

Issue #9722, opened 2017-01-21
Client side discovery assumes DiscoveryStrategyConfigs


neilstevenson:
If `hazelcast.discovery.enabled` on the client side, code assumes DiscoveryStrategyConfigs

1. If neither DiscoveryServiceProvider or DiscoveryStrategyConfigs is specified, an NPE is thrown.
A more helpful exception should be provided.

2. If DiscoveryServiceProvider is specified and DiscoveryStrategyConfigs is not, an NPE is still thrown.

Reproducer for 3.8-EA:
```
    	try {
            ClientConfig clientConfig1 = new ClientConfig();
            clientConfig1.setProperty("hazelcast.discovery.enabled", Boolean.TRUE.toString());
            HazelcastClient.newHazelcastClient(clientConfig1);
    	} catch (Exception e1) {
    	    // An NPE isn't the most helpful
    	    e1.printStackTrace(System.err);
    	}

    	try {
            ClientConfig clientConfig2 = new ClientConfig();
            clientConfig2.setProperty("hazelcast.discovery.enabled", Boolean.TRUE.toString());
            DiscoveryServiceProvider discoveryServiceProvider = new DiscoveryServiceProvider() {
    			public DiscoveryService newDiscoveryService(DiscoveryServiceSettings arg0) {
    				return (DiscoveryService) new Object();
    			}
    		};
            clientConfig2.getNetworkConfig().getDiscoveryConfig().setDiscoveryServiceProvider(discoveryServiceProvider);
            HazelcastClient.newHazelcastClient(clientConfig2);
    	} catch (Exception e2) {
    	    // DiscoveryServiceProvider is only used if DiscoveryStrategyConfigs is not empty
    	    e2.printStackTrace(System.err);
    	}
```

PR #9762, opened 2017-01-26
Custom DiscoveryService fixes

mdogan:
- `DiscoveryService.discoverNodes()` should not return null,
added preconditions to places it's called.
- `DiscoveryServiceProvider` & `DiscoveryService` should be called
even when no DiscoveryStrategy exists.
- `ClientNetworkConfig.setDiscoveryConfig()` shouldn't accept null
similar to server side `JoinConfig.setDiscoveryConfig()`.

Follow up fix for #9722 & #9741

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/discovery/DiscoveryAddressProvider.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/discovery/DiscoveryAddressProvider.java
@@ -28,6 +28,8 @@ import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collection;
 
+import static com.hazelcast.util.Preconditions.checkNotNull;
+
 public class DiscoveryAddressProvider
         implements AddressProvider {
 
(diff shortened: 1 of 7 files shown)


tombujok:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/discovery/DiscoveryAddressProvider.java
```diff
@@ -41,7 +43,8 @@ public DiscoveryAddressProvider(DiscoveryService discoveryService, LoggingServic
 
     @Override
     public Collection<InetSocketAddress> loadAddresses() {
-        Iterable<DiscoveryNode> discoveredNodes = discoveryService.discoverNodes();
+        Iterable<DiscoveryNode> discoveredNodes = checkNotNull(discoveryService.discoverNodes(),
```
bilalyasar: after this pr, we may return empty list rather than `null` https://github.com/hazelcast/hazelcast-azure/blob/master/src/main/java/com/hazelcast/azure/AzureDiscoveryStrategy.java#L144
// cc @mesutcelik @emre-aydin

mesutcelik:
👍

PR #9767, opened 2017-01-26
Change Near Cache preloader's filename to a directory preference

tkountis:
A Near Cache config can be shared across multiple different instances if the * wildcard is used in its name, which by inheritence applies to the pre-loader configuration too. The pre-loader config, however, has a filename option which allows the user to hard code the name of the file store that will be used for the pre-loader. In this shared scenario above, this causes exceptions in rare concurrent conditions, when multiple maps/caches get created at the same time, where we try to acquire the same file lock due to the name clash. 

This change, makes the preference a directory rather than a filename. The user is still able to choose a file location for the pre-loader, but the actual filename is the Near Cache's name itself.
Fix hazelcast/hazelcast-enterprise#1252


jerrinot:
I created https://github.com/hazelcast/hazelcast/issues/9771

Issue #9752, opened 2017-01-25
3.8-SNAPSHOT  Verbose stack trace logged in client for maxConcurrentInvocations, reached


Danny-Hazelcast:
verbose stack trace and Server level warning, for a case, which is expected, and really just a msg confirming that all is working as expected.

its not really a SEVER issue, with a stack trace.  as a user has explicitly configured back pressure and the prop. so the msg is to verbose.

```
 -Dhazelcast.client.max.concurrent.invocations=100
```
also the user will have to catch the com.hazelcast.core.HazelcastOverloadException
in the user code so the user is all ready directly dealing with a com.hazelcast.core.HazelcastOverloadException making the client log msg redundant. 

```
Jan 25, 2017 7:00:53 AM com.hazelcast.client.spi.ClientExecutionService
SEVERE: hz.client_0 [HZ] [3.8-SNAPSHOT] Failed to execute LoggingDelegatingFuture{task=RepairingTask{}}
com.hazelcast.core.HazelcastOverloadException: maxConcurrentInvocations : 100 is reached
	at com.hazelcast.client.spi.impl.CallIdSequence$CallIdSequenceFailFast.next(CallIdSequence.java:69)
	at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport.registerInvocation(ClientInvocationServiceSupport.java:160)
	at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport.send(ClientInvocationServiceSupport.java:114)
	at com.hazelcast.client.spi.impl.ClientSmartInvocationServiceImpl.invokeOnConnection(ClientSmartInvocationServiceImpl.java:119)
	at com.hazelcast.client.spi.impl.ClientSmartInvocationServiceImpl.invokeOnTarget(ClientSmartInvocationServiceImpl.java:81)
	at com.hazelcast.client.spi.impl.ClientInvocation.invokeOnSelection(ClientInvocation.java:135)
	at com.hazelcast.client.spi.impl.ClientInvocation.invoke(ClientInvocation.java:114)
	at com.hazelcast.client.cache.impl.nearcache.invalidation.ClientCacheMetaDataFetcher.scanMembers(ClientCacheMetaDataFetcher.java:72)
	at com.hazelcast.internal.nearcache.impl.invalidation.MetaDataFetcher.fetchMetadata(MetaDataFetcher.java:62)
[...]

PR #9768, opened 2017-01-27
Reduce excessive logging when meta data fetcher invocation fail.

sancar:
fixes https://github.com/hazelcast/hazelcast/issues/9752


jerrinot:
is there a concept of urgent messages on a client? can't this (exhausted slots) happen to something more critical? for example the heartbeat?

sancar:
@jerrinot There is a way to bypass this check. Authentication, Ping , GetPartition and Membership listener registration uses that. See usages of `ClientInvocation.invokeUrgent`

PR #9772, opened 2017-01-27
Making EP operation partitionAware

tombujok:


mdogan:
@tombujok: what is the reasoning behind this change? since `PartitionAwareOperation` doesn't have any meaning in runtime. It's not used at all.

tombujok:
@mdogan It's for documenting purposes and static-code analysis. I'm fixing the bug related to record-store creation outside of the partition-thread. I can imagine we could add some validation in the future, whether partitionAware operations are actually run on partitionThreads. IMHO there's need for some more consistent convention. Otherwise it's tough to apply any reasoning on the code and each place where an operation is invoked has to be tracked...

mdogan:
@tombujok: I see, thanks. 
`PartitionAwareOperation` was being used as marker interface, all partition specific operations had to be marked as `PartitionAwareOperation`. This constraint was removed later...

PR #9783, opened 2017-01-30
Move mem cost calculation to support change in EE

tkountis:
Moved memory cost calculation one level deeper, to support a change in EE.
hazelcast/hazelcast-enterprise#1276


hazelcast/src/main/java/com/hazelcast/monitor/impl/LocalMapStatsImpl.java
```diff
@@ -390,6 +390,7 @@ public String toString() {
                 + ", lockedEntryCount=" + lockedEntryCount
                 + ", dirtyEntryCount=" + dirtyEntryCount
                 + ", heapCost=" + heapCost
+                + ", nearCacheStats=" + nearCacheStats
```
ahmetmircik: since `nearCacheStats` can be `null` when there is no near cache enabled, 
doing `nearCacheStats == null ? "" :  ", nearCacheStats=" + nearCacheStats` can be more appropriate

PR #9785, opened 2017-01-30
Fix wrong task count when having multiple scheduled executors

### LINKED ISSUE #9784: Scheduled Executor wrong number of tasks returned on getAllScheduled
When scheduling tasks with ***multiple*** executors, calling `getAllScheduled` on one of them returns tasks scheduled per member but for ***all*** executors rather than for the one we called the `getAllScheduled` from.

Reproducer:
```
@Test
public void wrongTaskCount()

    String runsCounterName = "runs";
    HazelcastInstance instance = Hazelcast.newHazelcastInstance(null);
    ICountDownLatch runsLatch = instance.getCountDownLatch(runsCounterName);

    int numOfSchedulers = 10;
    int numOfTasks = 10;
    int expectedTotal = numOfSchedulers * numOfTasks;

    runsLatch.trySetCount(expectedTotal);

    for (int i = 0; i < numOfSchedulers; i++) {
        IScheduledExecutorService s = instance.getScheduledExecutorService("scheduler_" + i);

        for (int k = 0; k < numOfTasks; k++) {
            s.scheduleAtFixedRate(new ICountdownLatchRunnableTask(runsCounterName), 0, 2, SECONDS);
        }
    }

    runsLatch.await(10, SECONDS);

    int actualTotal = 0;
    for (int i = 0; i < numOfSchedulers; i++) {
        actualTotal += countScheduledTasksOn(instance.getScheduledExecutorService("scheduler_" + i));
    }

    assertEquals(expectedTotal, actualTotal, 0);
}
```

tkountis:
When scheduling tasks with multiple executors, calling getAllScheduled on one of them returns tasks scheduled per member but for all executors rather than for the one we called the getAllScheduled from.

This change is patching the operation to only work on a specific container given its name, and also introduces minor class hierarchy improvements, and proper resetting of a container when its proxy gets destroyed.
Fix #9784

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/AbstractScheduledExecutorContainerHolder.java
+++ b/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/AbstractScheduledExecutorContainerHolder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


tkountis:
run-lab-run

Issue #9781, opened 2017-01-30
Flood in log when quorum is not present


vmnikulin:
Hello!

When quorum is not present operations like `MapSizeOperation` produces 2 * 271 (542) log records on every client request. So when quorum is broken log made by `MapSizeOperation` quick fill my disk. 
 
Log file looks like:
`
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-5] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-0] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-4] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-0] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-5] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-1] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-7] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
2017-01-30 17:35:14,676 [hz._hzInstance_1_dev.partition-operation.thread-1] WARN [MapSizeOperation] [192.168.1.151]:5701 [dev] [3.6.7] Cluster quorum failed, quorum minimum size: 2, current size: 1
[...]

PR #9790, opened 2017-01-31
Do not log business exceptions

### LINKED ISSUE #9051: DurableExecutorService exceptions
If `DurableExecutorService.retrieveResult(long)` is run from a client for a task that doesn't exist, two exceptions appear on 3.7.1.

In the server logs `StaleTaskIdException`. This shouldn't be logged to the server if re-thrown to the client.

On the client, `ExecutionException`. This should be a `StaleTaskIdException`

jerrinot:
Exceptions such as StaleSequenceException are part of a normal
flow. They are propagated to user code and there is no reason
why Hazelcast should log them by default.

Fixes #9051, #9781

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/Operation.java
@@ -25,7 +25,7 @@ import com.hazelcast.nio.Connection;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.DataSerializable;
-import com.hazelcast.quorum.QuorumException;
+import com.hazelcast.spi.exception.SilentException;
 import com.hazelcast.spi.exception.RetryableException;
 import com.hazelcast.spi.properties.GroupProperty;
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/spi/Operation.java
```diff
@@ -486,7 +487,9 @@ public void onExecutionFailure(Throwable e) {
      */
     public void logError(Throwable e) {
         final ILogger logger = getLogger();
-        if (e instanceof RetryableException) {
+        if (e instanceof SilentException) {
+            logger.finest(e.getMessage(), e);
+        } else if (e instanceof RetryableException) {
```
vbekiaris: One could mark an `Exception` with both `SilentException` & `RetryableException`; in this case it will not be retried at all, just silently ignored. I cannot think why one would want a silent & retryable exception, but should we add a related warning in `SilentException`'s javadoc?
Also, since `QuorumException` becomes a `SilentException` with this PR, the `instanceof QuorumException` else-if branch below can be removed.

vbekiaris: Please ignore the remark about `RetryableException`, I now see it's just the logging that is being done here.

PR #9793, opened 2017-02-01
[WIP] Rolling upgrade cleanup & utils

tombujok:
Version & ClusterVersion may not be null anymore (UNKNOWN or value)
Enables doing assertions on Version and ClusterVersion
Introduces Version & ClusterVersion constants
Introduces IllegalVersionException throw in case the assertions fails

```
assertThat(clusterVersion).isLessThan(CLUSTER_VERSION_3_8);
assertThat(version).isLessThan(VERSION_9);
```

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
@@ -68,7 +68,7 @@ public class ClusterStateManager {
     private static final long LOCK_LEASE_EXTENSION_MILLIS = TimeUnit.SECONDS.toMillis(20);
 
     // this is the version at which the cluster operates. see Cluster#getClusterVersion
-    volatile ClusterVersion clusterVersion;
+    private volatile ClusterVersion clusterVersion = ClusterVersion.UNKNOWN;
 
     private final Node node;
     private final ILogger logger;
@@ -166,16 +166,16 @@ public class ClusterStateManager {
         clusterServiceLock.lock();
         try {
             state = ClusterState.ACTIVE;
-            // not notifying cluster version listeners about change to null. consider for example the following scenario:
+            // not notifying cluster version listeners about change to UNKNOWN. consider for example the following scenario:
             // - node starts with codebase version 3.9, overrides init cluster version via group property to 3.8
             // - node joins an existing 3.8 cluster which is undergoing rolling-upgrade to 3.9
             // - once all cluster members are on 3.9, cluster version is upgraded to 3.9.0
-            // - clusterVersion is reset to null
+            // - clusterVersion is reset to UNKNOWN
             // - if cluster version listener is notified of null cluster version, it should receive the overridden one (3.8)
             // - if 3.8 discovery & join messages are incompatible, node will not be able to join 3.9 cluster
             // Instead, not notifying cluster version listeners will let the node use its last set cluster version for discovery &
             // join messages and join the cluster.
-            clusterVersion = null;
+            clusterVersion = ClusterVersion.UNKNOWN;
             stateLockRef.set(LockGuard.NOT_LOCKED);
         } finally {
             clusterServiceLock.unlock();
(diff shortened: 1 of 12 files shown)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
```diff
@@ -238,7 +238,7 @@ private void validateNodeCompatibleWith(ClusterVersion clusterVersion) {
 
     // validate transition from current to newClusterVersion is allowed
     private void validateClusterVersionChange(ClusterVersion newClusterVersion) {
-        if (clusterVersion != null && clusterVersion.getMajor() != newClusterVersion.getMajor()) {
+        if (clusterVersion != null && !clusterVersion.isUnknown() && clusterVersion.getMajor() != newClusterVersion.getMajor()) {
```
Donnerbart: The `clusterVersion != null` is obsolete now, isn't it?

hazelcast/src/main/java/com/hazelcast/version/ClusterVersionAssertion.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    public void isNotUnknown() {
+        if (actual.isUnknown()) {
+            fail("unknown");
+        }
+    }
+
+    private void fail(String msg, String version) {
```
Donnerbart: You could use `ClusterVersion version` instead of `String version`, then you can save all those `version.toString()` in the class.

EDIT: I just saw the `fail(String msg)`, so this might not be an option.

hazelcast/src/main/java/com/hazelcast/version/ClusterVersionAssertion.java
```diff
@@ -0,0 +1,84 @@
...
+        throw new IllegalVersionException(
+                String.format("Actual cluster version %s is %s %s", actual.toString(), msg, version));
```
Donnerbart: * you could add a static import for `String.format()`
* I think the `toString()` is implicitly called when `%s` is used
* so even with `ClusterVersion` instead of `String` everything might fit into a single line of code:
```
throw new IllegalVersionException(format("Actual cluster version %s is %s %s", actual, msg, version));
```

hazelcast/src/main/java/com/hazelcast/version/IllegalVersionException.java
```diff
@@ -0,0 +1,41 @@
...
+public class IllegalVersionException extends HazelcastException {
+
+    public IllegalVersionException() {
+    }
+
+    public IllegalVersionException(final String message) {
+        super(message);
+    }
+
+    public IllegalVersionException(final String message, final Throwable cause) {
```
Donnerbart: I didn't check everything, but unused constructors will hurt the code coverage. I also think Peter has the rule of thumb to not have them, regardless of the code coverage. So in the end you might want to remove the unused ones.

hazelcast/src/test/java/com/hazelcast/version/ClusterVersionAssertionTest.java
```diff
@@ -0,0 +1,131 @@
...
+    public void isNotUnknown_fail_explicit() throws Exception {
+        assertThat(of("0.0")).isNotUnknown();
+    }
+
+    @Test
+    public void isNotUnknown() throws Exception {
+        assertThat(of("3.8")).isNotUnknown();
+    }
+
+}
```
Donnerbart: Missing newline (would be a CheckStyle error if we would include test classes)

hazelcast/src/test/java/com/hazelcast/version/VersionAssertionTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void isNotUnknown_fail_explicit() throws Exception {
+        assertThat(of(-1)).isNotUnknown();
+    }
+
+    @Test
+    public void isNotUnknown() throws Exception {
+        assertThat(of(8)).isNotUnknown();
+    }
+
+}
```
Donnerbart: Missing newline (would be a CheckStyle error if we would include test classes)

tombujok:
@vbekiaris I'll need to rebase this after the version unification ;)

PR #9797, opened 2017-02-03
Merge ClusterVersion to Version

vbekiaris:
Expands `Version` to `major.minor` instead of a single byte version value.
Unify `ClusterVersion` to `Version` to simplify development for backwards compatibility.


vbekiaris:
run-lab-run

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -234,7 +234,7 @@ public void shutdown() {
     @Override
     public void validateJoinRequest(JoinMessage joinMessage) {
         // check joining member's major.minor version is same as current cluster version's major.minor numbers
-        if (!joinMessage.getVersion().asClusterVersion().equals(node.getClusterService().getClusterVersion())) {
+        if (!joinMessage.getVersion().asVersion().equals(node.getClusterService().getClusterVersion())) {
             throw new VersionMismatchException("Joining node's version " + joinMessage.getVersion() + " is not compatible with "
```
tombujok: Hmm, do we need this asVersion()?

tombujok: Aha, shouldn't it be getMemberVersion().asVersion() to make it less confusing?

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -268,9 +268,9 @@ public void onClusterVersionChange(ClusterVersion newVersion) {
     }
 
     @Override
-    public boolean isNodeVersionCompatibleWith(ClusterVersion clusterVersion) {
+    public boolean isNodeVersionCompatibleWith(Version clusterVersion) {
         Preconditions.checkNotNull(clusterVersion);
-        return node.getVersion().asClusterVersion().equals(clusterVersion);
+        return node.getVersion().asVersion().equals(clusterVersion);
     }
```
tombujok: Same here? There's probably more places like this.

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
vbekiaris: 👍 just pushed another commit that renames `JoinMessage.version` to `memberVersion`

PR #9800, opened 2017-02-04
[WIP] Add instance prefix to hazelcast threads

### LINKED ISSUE #8542: add group name to all HZ threads
There are a number of HZ threads with name like "cached16", "scheduled" or "pool-1-thread-2". Do you mind to add the thread's HZ cluster name to the thread name, like it is done for partition operation threads, event threads, etc? In my case I have several HZ instances working in the same JVM, so it is not clear to which HZ instance the "cached16" thread belongs.

Thanks, Denis

jerrinot:
Fixed #8542


jerrinot:
I am closing it as there were quite some changes recently (ThreadGroup abandoned, etc)

PR #9806, opened 2017-02-06
Bouncing member test fixes

vbekiaris:
* When using `BounceMemberRule.test`, members bouncing thread would stop without waiting for actual test tasks to complete.
* Reverts QueryBounceTest to be executed in isolation (use serial class runner, not marked as `ParallelTest`, as was the case for `QueryDuringMigrationsStressTest` it replaced).
* Applies `JitterRule` for hiccups monitoring in `QueryBounceTest`


vbekiaris:
thanks @mdogan for the quick review

PR #9807, opened 2017-02-06
Fixes for Near Cache configuration

Donnerbart:
* Unified order of parameters in XSD and created XML
* Added cache-local-entries to ConfigXmlGenerator
* Added missing conversion of maxSize to valid value for EvictionConfig
* Updated documentation for cache-local-entries in XSD

Part of https://github.com/hazelcast/hazelcast-reference-manual/issues/232

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/NearCacheConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/NearCacheConfig.java
@@ -159,7 +159,7 @@ public class NearCacheConfig implements DataSerializable, Serializable {
         if (evictionConfig != null) {
             this.evictionConfig = evictionConfig;
         } else {
-            this.evictionConfig.setSize(maxSize);
+            this.evictionConfig.setSize(calculateMaxSize(maxSize));
             this.evictionConfig.setEvictionPolicy(EvictionPolicy.valueOf(evictionPolicy));
             this.evictionConfig.setMaximumSizePolicy(ENTRY_COUNT);
         }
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/config/NearCacheConfig.java
```diff
@@ -159,7 +159,7 @@ public NearCacheConfig(int timeToLiveSeconds, int maxSize, String evictionPolicy
         if (evictionConfig != null) {
             this.evictionConfig = evictionConfig;
         } else {
-            this.evictionConfig.setSize(maxSize);
+            this.evictionConfig.setSize(calculateMaxSize(maxSize));
```
Donnerbart: This should convert the `0` from the deprecated config to `Integer.MAX_VALUE` for the new configuration via `EvictionConfig`, which doesn't allow `0` as size.

hazelcast/src/main/resources/hazelcast-config-3.8.xsd
```diff
@@ -2438,35 +2446,27 @@
...
+
+                        Deprecated since 3.8, please use &lt;eviction/&gt;
                     </xs:documentation>
                 </xs:annotation>
             </xs:element>
             <xs:element name="cache-local-entries" type="xs:boolean" minOccurs="0" maxOccurs="1" default="false">
                 <xs:annotation>
                     <xs:documentation>
-                        True to cache local entries also.
+                        True to cache local entries, which belong to the member itself.
```
Donnerbart: This should be the only content change in the XSD files, everything else is just re-ordering.

hazelcast/src/test/java/com/hazelcast/config/ConfigXmlGeneratorTest.java
```diff
@@ -135,7 +136,8 @@ public void testMapNearCacheConfig() {
                 .setInMemoryFormat(InMemoryFormat.NATIVE)
                 .setMaxSize(23)
                 .setEvictionPolicy("LRU")
-                .setMaxIdleSeconds(42);
+                .setMaxIdleSeconds(42)
+                .setCacheLocalEntries(true);
```
Donnerbart: This settings is member-only and was missing in the generated XML.

PR #9809, opened 2017-02-06
Do not deserialize Portable objects when storing into index

### LINKED ISSUE #9808: Client Map addIndex fail for portable only values without registering portable classes on member
A client cannot add index to map using Portable values without registering Portable classes on member.

Below test fails

```
    @Test
    public void test_addIndexFromClient_portable() {
        Config config = new Config();
        HazelcastInstance hz1 = factory.newHazelcastInstance(config);

        ClientConfig clientConfig = new ClientConfig();
        clientConfig.getSerializationConfig()
                    .addPortableFactory(TestSerializationConstants.PORTABLE_FACTORY_ID, new PortableFactory() {
                        public Portable create(int classId) {
                            return new NamedPortable();
                        }

                    });
        HazelcastInstance client = factory.newHazelcastClient(clientConfig);
        IMap<String, NamedPortable> test = client.getMap("test");

        test.put("key1", new NamedPortable("name-1", 1));
        test.addIndex("name", true);
    }

```

jerrinot:
fixes #9808

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/standalone/ClientMapStandaloneTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/standalone/ClientMapStandaloneTest.java
@@ -45,6 +45,7 @@ import java.util.concurrent.CountDownLatch;
 
 import static com.hazelcast.query.Predicates.equal;
 import static com.hazelcast.query.Predicates.in;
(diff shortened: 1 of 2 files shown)


jerrinot:
thanks for reviews, I am going to send a backport.

PR #9812, opened 2017-02-07
Do not deserialize Portable objects when storing into index

jerrinot:
backport of #9809
(cherry picked from commit 3e07678)


jerrinot:
forward port is already merged

Issue #9798, opened 2017-02-03
Predicate returns wrong data


carlo-activia:
When using EntryProcessor and index on a field the EntryProcessor uses I get wrong data when executing a Predicate. Here is the code:

  @Test
  public void issue() throws Exception {
    Config config = new Config();
    MapConfig testMapConfig = config.getMapConfig("testMap");
    testMapConfig.setInMemoryFormat(InMemoryFormat.OBJECT);
    testMapConfig.getMapIndexConfigs().add(new MapIndexConfig("lastValue", true));
    HazelcastInstance instance = Hazelcast.newHazelcastInstance(config);

    try {
      IMap<Object, Object> testMap = instance.getMap("testMap");

      testMap.set(1L, new MyData(10));

      testMap.executeOnKey(1L, new MyProcessor(15));

      Predicate betweenPredicate = Predicates.between("lastValue", 0, 11);
      Collection<Object> values = testMap.values(betweenPredicate);
      Assert.assertEquals(0, values.size());
    } finally {
      instance.shutdown();
    }
  }

Here is the EntryProcessor:

  static class MyProcessor extends AbstractEntryProcessor<Long, MyData> {
    private long newValue;

    public MyProcessor(long newValue) {
      this.newValue = newValue;
    }

    @Override
    public Object process(Map.Entry<Long, MyData> entry) {
      MyData data = entry.getValue();
      data.setLastValue(newValue);
      entry.setValue(data);
      return null;
    }
  }

I expect the assertion to work, but it fails I get 1 record. If I change OBJECT to BINARY or if I comment the line "testMapConfig.getMapIndexConfigs().add(new MapIndexConfig("lastValue", true));" then it works.

Any ideas? I am using Hazelcast 3.7.5.

Thanks in advance.

tombujok:
Thx for reporting this.
Everything will work if you change your EntryProcessor to create a new entry with the new value.
```
static class MyProcessor extends AbstractEntryProcessor<Long, MyData> {
        private long newValue;

        public MyProcessor(long newValue) {
            this.newValue = newValue;
        }

        @Override
        public Object process(Map.Entry<Long, MyData> entry) {
            MyData newData = new MyData(newValue);
            entry.setValue(newData);
            return newValue;
        }
    }
```

In order to modify the index we need the old `entry.getValue()`. If you modify it the index won't find the old value in the Object since it has been modified.
With BINARY the object is deserialised so the oldValue is preserved.

[...]

PR #9818, opened 2017-02-07
Clones object for EntryProcessing if an index exists 

tombujok:
Solves #9798

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
@@ -50,6 +50,7 @@ import com.hazelcast.wan.WanReplicationService;
 
 import java.util.concurrent.atomic.AtomicInteger;
 
+import static com.hazelcast.config.InMemoryFormat.OBJECT;
 import static com.hazelcast.map.impl.eviction.Evictor.NULL_EVICTOR;
 import static com.hazelcast.map.impl.mapstore.MapStoreContextFactory.createMapStoreContext;
 import static java.lang.System.getProperty;
@@ -284,6 +285,10 @@ public class MapContainer {
 
     // callback called when the MapContainer is de-registered from MapService and destroyed - basically on map-destroy
     public void onDestroy() {
+    }
+
+    public boolean shouldCloneOnEntryProcessing() {
+        return getIndexes().hasIndex() & OBJECT.equals(mapConfig.getInMemoryFormat());
     }
 }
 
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -285,6 +286,10 @@ public InterceptorRegistry getInterceptorRegistry() {
     // callback called when the MapContainer is de-registered from MapService and destroyed - basically on map-destroy
     public void onDestroy() {
     }
+
+    public boolean shouldCloneOnEntryProcessing() {
+        return getIndexes().hasIndex() & OBJECT.equals(mapConfig.getInMemoryFormat());
+    }
 }
```
tombujok: I know that this is an imperfect location for this method but it's the only common context for all the calls that exists right now. I could also create an EntryProcessorUtils class, but seemed an overkill for now. Feel free to chime in.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionWideEntryOperation.java
```diff
@@ -61,29 +61,33 @@ public void innerBeforeRun() throws Exception {
...
-            Map.Entry entry = createMapEntry(dataKey, oldValue);
+            Map.Entry entry = createMapEntry(dataKey, value);
             Data response = process(entry);
             if (response != null) {
                 responses.add(dataKey, response);
             }
 
             // first call noOp, other if checks below depends on it.
-            if (noOp(entry, oldValue, now)) {
+            if (noOp(entry, value, now)) {
```
mmedenjak: Shouldn't this be `oldValue`? Same with `entryRemoved`

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionWideEntryBackupOperation.java
```diff
@@ -42,21 +43,24 @@ public PartitionWideEntryBackupOperation(String name, EntryBackupProcessor backu
...
+            if (!applyPredicate(dataKey, value)) {
                 continue;
             }
 
-            Map.Entry entry = createMapEntry(dataKey, oldValue);
+            Map.Entry entry = createMapEntry(dataKey, value);
             processBackup(entry);
 
-            if (noOp(entry, oldValue)) {
+            if (noOp(entry, value)) {
```
mmedenjak: `oldValue`?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionWideEntryOperation.java
tombujok: Thx, fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PartitionWideEntryBackupOperation.java
tombujok: Thx, fixed.

jerrinot:
+1 from me.

PR #9824, opened 2017-02-07
Enable logging in default profile

Donnerbart:
* Enable logging in default profile and removed `LOCAL` profile
* added newlines in `groups` and `excludedGroups` to increase readability

It was quite tricky to find out where the `-Dhazelcast.logging.type=none` was coming from, if no Maven profile was selected. After some discussion Jaromir and me agreed to try out removing the logging setting from the default profile.

If we remove this we should theoretically be able to remove the whole `LOCAL` profile, if I'm not wrong. At least logging works now out of the box, without knowing that you have to select the `LOCAL` profile to activate it. The `groups` and `excludedGroups` are anyway not working with the IDEA runner, so they should not provide any value.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,7 +252,6 @@
                         -Xms128m -Xmx1G -XX:MaxPermSize=128M
                         -Dhazelcast.phone.home.enabled=false
                         -Dhazelcast.mancenter.enabled=false
-                        -Dhazelcast.logging.type=none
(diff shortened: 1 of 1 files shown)


mesutcelik:
👍

Donnerbart:
I left the `LOCAL` profile for now, some people may use it on the command line with the Maven binary. I think a cleaner and maybe easier Maven setup is something for HZ 4.

PR #9827, opened 2017-02-07
Harden getAllScheduled to be more resiliant when members go down

### LINKED ISSUE #9788: Scheduled Executor wrong number of tasks returned on getAllScheduled Kill Random Member
this tests shows getAllScheduled() returns incorrect count while cluster member are killed at random.

```
10:41:42 jvm{id=HzMember2HZ running=false pid=0 lib=$HOME/hz-root/hz-lib/3.8-SNAPSHOT dir=hz-root/HzMember2HZ ip=Box{user=ec2-user, pub=54.91.94.164, pri=10.0.0.153}}
10:41:52 HzClient2HZ validate_schedule hz.executor.scheduled.validate.ScheduledFuturesCount threadId=0 global.AssertionException: total Scheduled Futures Count 9979 != 10000 
10:41:52 HzClient3HZ validate_schedule hz.executor.scheduled.validate.ScheduledFuturesCount threadId=0 global.AssertionException: total Scheduled Futures Count 9979 != 10000 
10:41:52 HzClient4HZ validate_schedule hz.executor.scheduled.validate.ScheduledFuturesCount threadId=0 global.AssertionException: total Scheduled Futures Count 9975 != 10000 
10:41:52 HzClient1HZ validate_schedule hz.executor.scheduled.validate.ScheduledFuturesCount threadId=0 global.AssertionException: total Scheduled Futures Count 9983 != 10000 
```

tkountis:
`GetAllScheduled` is called on each member and constructs the accumulated map of tasks per member. When a member goes down, we then have incorrect count of tasks, since we are not accessing that information through the partition-threads. Eventually the numbers will be correct, but  we can improve this doing the same operation per partition rather than per member. 

However, fetching all tasks per partition, will miss tasks owned by members (`scheduleOnMember*`) which we still have to use the old approach to collect.

Fixes #9788


tkountis:
run-lab-run

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/InvokeOnMembers.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+
+/**
+ * Executes an operation on a set of targets.
+ * Similar to {@link com.hazelcast.spi.impl.operationservice.impl.InvokeOnPartitions} but for members.
+ */
+@PrivateApi
+public final class InvokeOnMembers {
```
mmedenjak: Some duplication with the `InvokeOnPartitions`. Perhaps we can move this to `com.hazelcast.spi.impl.operationservice.impl` and have a common base class

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/GetAllScheduledOperationFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.scheduledexecutor.impl.ScheduledExecutorDataSerializerHook;
+import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.OperationFactory;
+
+import java.io.IOException;
+
+public class GetAllScheduledOperationFactory implements OperationFactory {
```
mmedenjak: Can this be merged with `GetAllScheduledOnMemberOperationFactory`?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorServiceProxy.java
```diff
@@ -345,6 +320,60 @@ public void shutdown() {
         waitWithDeadline(calls, SHUTDOWN_TIMEOUT, TimeUnit.SECONDS, WHILE_SHUTDOWN_EXCEPTION_HANDLER);
     }
 
+    private <V> void retrieveAllMemberOwnedScheduled(Map<Member, List<IScheduledFuture<V>>> accumulator) {
+        try {
+            InvokeOnMembers invokeOnMembers = new InvokeOnMembers(getNodeEngine(), getServiceName(),
+                    new GetAllScheduledOnMemberOperationFactory(name),
+                    getNodeEngine().getClusterService().getMembers());
```
mmedenjak: Do you include lite members?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/InvokeOnMembers.java
tkountis: That was the idea, but due to 3.8 I didn't want to change the public API of `OperationService` at this stage. We can do so for next version.

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/InvokeOnMembers.java
```diff
@@ -0,0 +1,135 @@
...
+        for (Map.Entry<Member, Future> responseEntry : futures.entrySet()) {
+            try {
+                Future future = responseEntry.getValue();
+                results.put(responseEntry.getKey(), serializationService.toObject(future.get()));
+            } catch (Throwable t) {
+                if (logger.isFinestEnabled()) {
+                    logger.finest(t);
+                } else {
+                    logger.warning(t.getMessage());
+                }
```
tkountis: Kept it consistent with the partitions version, I also thought was a bit convoluted. I can look at this again once we merge the logic, wdyt ?

hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorServiceProxy.java
tkountis: Yes, @jerrinot brought this up also, but this is consistent with the ExecutorService. This is a computation service, being an executor and handling tasks and such, however, the members (including lite) do store some data, the task definition in this case.

sancar:
One think I have problem with that futures are waited on the operation thread.
You can make use of ` AbstractMultiTargetMessageTask` to avoid that.  Basically it waits responses via callback and send response to client when all responses came. 

I don't want to force you to do this we already have similar problems elsewhere. 
see `AbstractAllPartitionsMessageTask` and `AbstractMultiPartitionMessageTask`.
We know that they are harmful. And we have a prd to replace them all in one swipe with `callbacks` instead of `future.gets`

sancar:
@tkountis I have included your commit in my pr. We can close this one without merging.
This commit will be merged in https://github.com/hazelcast/hazelcast/pull/9862

PR #9832, opened 2017-02-08
Convert built-in Projections and Aggregations to IdentifiedDataSerializable

tombujok:
Convert built-in Projections and Aggregations to IdentifiedDataSerializable

EE counterpart: https://github.com/hazelcast/hazelcast-enterprise/pull/1289

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/aggregation/Aggregator.java
+++ b/hazelcast/src/main/java/com/hazelcast/aggregation/Aggregator.java
@@ -16,8 +16,6 @@
 
 package com.hazelcast.aggregation;
 
-import com.hazelcast.nio.serialization.impl.BinaryInterface;
-
 import java.io.Serializable;
 
 /**
@@ -36,7 +34,6 @@ import java.io.Serializable;
  * @param <R> result type
  * @since 3.8
  */
-@BinaryInterface
 public abstract class Aggregator<I, R> implements Serializable {
 
     /**
--- a/hazelcast/src/main/java/com/hazelcast/aggregation/impl/AggregatorDataSerializerHook.java
+++ b/hazelcast/src/main/java/com/hazelcast/aggregation/impl/AggregatorDataSerializerHook.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.aggregation.impl;
+
+import com.hazelcast.internal.serialization.DataSerializerHook;
+import com.hazelcast.internal.serialization.impl.ArrayDataSerializableFactory;
+import com.hazelcast.internal.serialization.impl.FactoryIdHelper;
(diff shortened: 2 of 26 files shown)


hazelcast/src/main/java/com/hazelcast/aggregation/Aggregator.java
```diff
@@ -36,7 +34,6 @@
  * @param <R> result type
  * @since 3.8
  */
-@BinaryInterface
```
asimarslan: Can we keep this annotation for now, I'll send a separate PR for them.

hazelcast/src/main/java/com/hazelcast/aggregation/impl/AggregatorDataSerializerHook.java
```diff
@@ -0,0 +1,150 @@
...
+    public int getFactoryId() {
+        return F_ID;
+    }
+
+    @Override
+    public DataSerializableFactory createFactory() {
+        ConstructorFunction<Integer, IdentifiedDataSerializable>[] constructors = new ConstructorFunction[LEN];
+
+        constructors[BIG_DECIMAL_AVG] = new ConstructorFunction<Integer, IdentifiedDataSerializable>() {
+            public IdentifiedDataSerializable createNew(Integer arg) {
```
ihsandemir: I wished that there was a ConstructorFunction that returns a value type and I do not need a constructor arg, since the arg is an unused argument.

hazelcast/src/main/java/com/hazelcast/aggregation/impl/BigDecimalAverageAggregator.java
```diff
@@ -57,4 +62,27 @@ public BigDecimal aggregate() {
...
+
+    @Override
+    public int getId() {
+        return AggregatorDataSerializerHook.BIG_DECIMAL_AVG;
+    }
+
+    @Override
+    public void writeData(ObjectDataOutput out) throws IOException {
+        out.writeUTF(attributePath);
+        out.writeObject(sum);
```
ihsandemir: BigDecimal object needs to be one of our supported serialization types. I see that it is java serializable at the moment. You may need to change this.

hazelcast/src/main/java/com/hazelcast/aggregation/impl/DistinctValuesAggregator.java
```diff
@@ -48,4 +52,34 @@ public void combine(Aggregator aggregator) {
...
+    public int getId() {
+        return AggregatorDataSerializerHook.DISTINCT_VALUES;
+    }
+
+    @Override
+    public void writeData(ObjectDataOutput out) throws IOException {
+        out.writeUTF(attributePath);
+        out.writeInt(values.size());
+        for (Object value : values) {
+            out.writeObject(value);
```
ihsandemir: Type of the value object can only be one of IDS, Portable, or custom but the I can not restrict this in the API. Is there a way to indicate that R should implement one of these interfaces?

sancar: @ihsandemir
As far as I understand,  These are customer objects. This values can be anytype. No need to restrict them.
This is no different than for datas on `map.put( ? , ? )`

hazelcast/src/main/java/com/hazelcast/aggregation/impl/BigDecimalAverageAggregator.java
sancar: Yes, this big decimal seems like problem for non-java clients. 
Or may be non-java client dont even need to support this class ever ?

sancar: @tombujok Sorry. BigDecimal is supported in SerializationService and we can support in other languages also.
 So there is no problem.

hazelcast/src/main/java/com/hazelcast/aggregation/impl/AggregatorDataSerializerHook.java
tombujok: Would be good, but out of scope for this PR. Thx for the feedback though.

PR #9833, opened 2017-02-08
Added near cache serialization count tests

ahmetmircik:


Donnerbart:
I see new duplicated test code for JCache and IMap again. Can't we use the new framework for those new tests? It should be quite easy to add the code once to `AbstractBasicNearCacheTest` and see if there are failures in any of the Near Cache supported scenarios.

ahmetmircik:
@Donnerbart test is a serial test which relies on static variables, seems not a good fit for `AbstractBasicNearCacheTest`

Donnerbart:
I just saw that, yes. Hmm, I think you could make those counters normal fields or even local variables and hand them over as parameters to `SerializationCountingData` and the assert method. Then the tests have no dependency on each other and can also be run parallel. What do you think?

Issue #9735, opened 2017-01-23
InvalidationMemberAddRemoveTest.ensure_nearCachedClient_and_member_data_sync_eventually


hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 300000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:349)
	at com.hazelcast.spi.impl.AbstractInvocationFuture.get(AbstractInvocationFuture.java:179)
	at com.hazelcast.client.util.ClientDelegatingFuture.get(ClientDelegatingFuture.java:127)
	at com.hazelcast.client.util.ClientDelegatingFuture.get(ClientDelegatingFuture.java:113)
	at com.hazelcast.client.cache.impl.AbstractClientCacheProxy.getInternal(AbstractClientCacheProxy.java:137)
	at com.hazelcast.client.cache.impl.AbstractClientCacheProxy.get(AbstractClientCacheProxy.java:248)
	at com.hazelcast.client.cache.impl.ClientCacheProxy.get(ClientCacheProxy.java:79)
	at com.hazelcast.client.cache.impl.ClientCacheProxy.get(ClientCacheProxy.java:91)
	at com.hazelcast.client.cache.impl.nearcache.invalidation.InvalidationMemberAddRemoveTest$5.run(InvalidationMemberAddRemoveTest.java:182)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:967)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:984)
	at com.hazelcast.client.cache.impl.nearcache.invalidation.InvalidationMemberAddRemoveTest.ensure_nearCachedClient_and_member_data_sync_eventually(InvalidationMemberAddRemoveTest.java:177)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/1211/com.hazelcast$hazelcast-client/testReport/com.hazelcast.client.cache.impl.nearcache.invalidation/InvalidationMemberAddRemoveTest/ensure_nearCachedClient_and_member_data_sync_eventually/

[...]

Issue #9490, opened 2016-12-19
InvalidationMemberAddRemoveTest.ensure_nearCached_and_actual_data_sync_eventually


hasancelik:
```
java.lang.IllegalThreadStateException: null
	at java.lang.Thread.start(Thread.java:705)
	at com.hazelcast.map.impl.nearcache.invalidation.InvalidationMemberAddRemoveTest.ensure_nearCached_and_actual_data_sync_eventually(InvalidationMemberAddRemoveTest.java:156)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast/1174/testReport/junit/com.hazelcast.map.impl.nearcache.invalidation/InvalidationMemberAddRemoveTest/ensure_nearCached_and_actual_data_sync_eventually/

hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 300000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:315)
	at com.hazelcast.spi.impl.AbstractInvocationFuture.get(AbstractInvocationFuture.java:153)
	at com.hazelcast.map.impl.proxy.MapProxySupport.invokeOperation(MapProxySupport.java:371)
	at com.hazelcast.map.impl.proxy.MapProxySupport.getInternal(MapProxySupport.java:302)
	at com.hazelcast.map.impl.proxy.MapProxyImpl.get(MapProxyImpl.java:114)
	at com.hazelcast.map.impl.nearcache.invalidation.InvalidationMemberAddRemoveTest$5.run(InvalidationMemberAddRemoveTest.java:164)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:967)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:984)
	at com.hazelcast.map.impl.nearcache.invalidation.InvalidationMemberAddRemoveTest.ensure_nearCached_and_actual_data_sync_eventually(InvalidationMemberAddRemoveTest.java:160)
```

[...]

PR #9841, opened 2017-02-09
 Reset destination uuid if it is not primary

ahmetmircik:
_Main Subject:_
This PR is related to eventually consistent near cache improvements.

_Problem:_
Issue was observed when closing a member with `terminate()` method, `shutdown()` is ok.

_Fix:_
 If migration destination is not primary replica, partition uuid should be reset because we are only trying to keep metadata on primaries in case of migrations. Metadata don't have any backup.

And also refactored `InvalidationMemberAddRemoveTest`s.

Additionally, this pr can be seen a fix attempt for the issues:
closes https://github.com/hazelcast/hazelcast/issues/9490
closes https://github.com/hazelcast/hazelcast/issues/9735

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/MapMigrationAwareService.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/MapMigrationAwareService.java
@@ -77,9 +77,11 @@ class MapMigrationAwareService implements MigrationAwareService {
 
         if (SOURCE == event.getMigrationEndpoint()) {
             clearMapsHavingLesserBackupCountThan(event.getPartitionId(), event.getNewReplicaIndex());
-            getMetaDataGenerator().resetMetadata(event.getPartitionId());
+            getMetaDataGenerator().removeUuidAndSequence(event.getPartitionId());
         } else if (DESTINATION == event.getMigrationEndpoint()) {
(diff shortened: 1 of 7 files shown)


jerrinot:
this should fix #9490, right?

ahmetmircik:
Not sure the relation with  #9490, this pr is fixing issue with node `terminate()`

hazelcast/src/main/java/com/hazelcast/map/impl/MapMigrationAwareService.java
```diff
@@ -77,9 +77,11 @@ public void commitMigration(PartitionMigrationEvent event) {
 
         if (SOURCE == event.getMigrationEndpoint()) {
             clearMapsHavingLesserBackupCountThan(event.getPartitionId(), event.getNewReplicaIndex());
-            getMetaDataGenerator().resetMetadata(event.getPartitionId());
+            getMetaDataGenerator().removeUuidAndSequence(event.getPartitionId());
         } else if (DESTINATION == event.getMigrationEndpoint()) {
-            getMetaDataGenerator().getOrCreateUuid(event.getPartitionId());
+            if (event.getNewReplicaIndex() != 0) {
```
ahmetmircik: new puts will create uuid on promoted partition, it shouldn't be an issue for this reason.

PR #9842, opened 2017-02-09
Fix lock eviction scheduling on migration when lock expired

mmedenjak:
Previously on lock migration the lock was scheduled for eviction with difference between the current cluster clock and the expiration time. If the migration lasted long enough and lock has already expired it could schedule expiration with a negative value. This would cause the lock to be expired after a period which is equal to the time elapsed from the actual expiration time - the more the migration is delayed, the more will the lease be extended.

Added some javadoc for lock classes, fixed some generics and simplified a couple of if-statements.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStore.java
@@ -20,33 +20,140 @@ import com.hazelcast.nio.serialization.Data;
 
 import java.util.Set;
 
+/**
+ * A container for multiple locks. Each lock is differentiated by a different {@code key}.
+ */
 public interface LockStore {
 
+    /**
+     * Lock a specific key.
+     *
+     * @param key         the key to lock
+     * @param caller      the identifier for the caller
+     * @param threadId    the identifier for the thread on the caller
+     * @param referenceId the identifier for the invocation of the caller (e.g. operation call ID)
+     * @param leaseTime   the lease duration in milliseconds
+     * @return if the lock was successfully acquired
+     */
     boolean lock(Data key, String caller, long threadId, long referenceId, long leaseTime);
 
+    /**
+     * Lock a specific key for use inside a transaction.
+     *
+     * @param key         the key to lock
+     * @param caller      the identifier for the caller
+     * @param threadId    the identifier for the thread on the caller
+     * @param referenceId the identifier for the invocation of the caller (e.g. operation call ID)
+     * @param leaseTime   the lease duration in milliseconds
+     * @param blockReads  whether reads for the key should be blocked (e.g. for map and multimap)
+     * @return if the lock was successfully acquired
+     */
     boolean txnLock(Data key, String caller, long threadId, long referenceId, long leaseTime, boolean blockReads);
 
+    /**
+     * Extend the lease time for an already locked key.
+     *
+     * @param key       the locked key
+     * @param caller    the identifier for the caller
+     * @param threadId  the identifier for the thread on the caller
+     * @param leaseTime time in milliseconds for the lease to be extended
+     * @return if the lease was extended
+     */
     boolean extendLeaseTime(Data key, String caller, long threadId, long leaseTime);
 
+    /**
+     * Unlock a specific key.
+     *
+     * @param key         the locked key
+     * @param caller      the identifier for the caller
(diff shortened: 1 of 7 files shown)


hazelcast/src/test/java/com/hazelcast/concurrent/lock/LockBounceMemberTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Runs a cluster of {@value MEMBER_COUNT + 1} and calls lease lock from {@value DRIVER_COUNT} drivers. During the test bounces
+ * {@value MEMBER_COUNT} cluster members (shuts down and restarts). The remaining 1 member is stable (never bounced) but no
+ * locks are owned by this member.
+ * Asserts that all locks are unlocked eventually.
+ */
+@RunWith(HazelcastSerialClassRunner.class)
+@Category(SlowTest.class)
+public class LockBounceMemberTest {
```
mmedenjak: This is a useful test and simulates a scenario similar to the one we are experiencing issues with - https://hazelcast-l337.ci.cloudbees.com/view/rollup/job/rollup-lease-lock/. The flip side is that it actually didn't uncover the cause of the issue (lease never expiring) so I'm not sure if we should keep it.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStore.java
```diff
@@ -20,33 +20,140 @@
...
+     * @param caller      the identifier for the caller
+     * @param threadId    the identifier for the thread on the caller
+     * @param referenceId the identifier for the invocation of the caller (e.g. operation call ID)
+     * @param leaseTime   the lease duration in milliseconds
+     * @return if the lock was successfully acquired
+     */
     boolean lock(Data key, String caller, long threadId, long referenceId, long leaseTime);
 
+    /**
+     * Lock a specific key.
```
mdogan: This method marks the lock as transactional. It's the main difference between this and above plain `lock()`.

hazelcast/src/test/java/com/hazelcast/concurrent/lock/LockBounceMemberTest.java
mdogan: Can you rename the test to make it clear `lock with lease` is being tested? Something like `LockWithLeaseBounceMemberTest`, `LeaseLockBounceMemberTest` ... etc

hazelcast/src/test/java/com/hazelcast/concurrent/lock/LockBounceMemberTest.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final int MEMBER_COUNT = 4;
+    private static final int DRIVER_COUNT = 4;
+    private static final int LEASE_DURATION_SECONDS = 10;
```
mdogan: Maybe reducing lease duration to something like 1sec or a few 100ms helps to reveal the issue?

hazelcast/src/test/java/com/hazelcast/concurrent/lock/LockLeaseMemberBounceTest.java
```diff
@@ -0,0 +1,136 @@
...
+        bounceMemberRule.testRepeatedly(runnables, TEST_DURATION_SECONDS);
+
+        assertTrueEventually(new AssertTask() {
+            @Override
+            public void run() throws Exception {
+                for (Driver driver : drivers) {
+                    driver.doAssert();
+                }
+            }
+        }, MILLISECONDS.toSeconds(LEASE_DURATION_MILLIS) + 10);
```
mdogan: I think timeout here is too small, it can cause accidental failures. Also note that, `assertTrueEventually(...)` runs `AssertTask` in 200ms intervals.

mmedenjak: Hm, but that's the lease duration in seconds (floored) + 10 seconds. Isn't that enough?

mdogan: Ah sorry. You're right. I misinterpreted that. 
(What I read in my mind is `MILLISECONDS.toSeconds(LEASE_DURATION_MILLIS + 10)`)

PR #9847, opened 2017-02-09
Updates in Serializable & DataSerializable conventions test

vbekiaris:
* `DataSerializableConventionsTest` has been updated to reflect the following conventions:
 - classes in public packages are now checked for compliance (while they were previously excluded)
 - only concrete classes are checked for compliance. It is OK to annotate interfaces or abstract classes with `@BinaryInterface` as an indication to implementers, however each implementation has to be considered separately.
 - as a consequence of the above `@BinaryInterface` is not considered as inherited.
* `@BinaryInterface` has been extended with a `reason` method to indicate why a class is annotated as such

This PR's first commit includes the changes in the test itself and updates `DataSerializable` classes (either annotated or converted to `IdentifiedDataSerializable`); the second commit updates `Serializable` classes. The third commit moves `BinaryInterface` from `com.hazelcast.nio.serialization.impl` package to `com.hazelcast.nio.serialization`.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/CacheMergePolicy.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/CacheMergePolicy.java
@@ -16,6 +16,8 @@
 
 package com.hazelcast.cache;
 
+import com.hazelcast.nio.serialization.impl.BinaryInterface;
+
 import java.io.Serializable;
 
 /**
@@ -29,6 +31,7 @@ import java.io.Serializable;
  * (s)he should use {@link StorageTypeAwareCacheMergePolicy} which is sub-type of this interface.
  * </p>
  */
+@BinaryInterface
 public interface CacheMergePolicy extends Serializable {
 
     /**
--- a/hazelcast/src/main/java/com/hazelcast/cache/CacheMergePolicy.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/CacheMergePolicy.java
@@ -16,7 +16,7 @@
 
 package com.hazelcast.cache;
 
-import com.hazelcast.nio.serialization.impl.BinaryInterface;
+import com.hazelcast.nio.serialization.BinaryInterface;
 
 import java.io.Serializable;
 
--- a/hazelcast/src/main/java/com/hazelcast/cache/CacheNotExistsException.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/CacheNotExistsException.java
@@ -16,6 +16,10 @@
 
 package com.hazelcast.cache;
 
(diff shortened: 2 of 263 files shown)


vbekiaris:
`@BinaryInterface` has been extended with a `reason` to explain why a class cannot be migrated to `IdentifiedDataSerializable`. The default reason is that it is used in client-member communication, however I think most of the classes in this PR are not converted due to `OTHER_CONVENTION` (see below excerpt from `BinaryInterface.Reason` javadoc for an explanation). In any case, if you identify false positives in this PR where reason is the default (ie no value = `CLIENT_COMPATIBILITY`) please let me know and I will amend the PR.
```
        /**
         * Class is used in client-member communication, therefore changing its serialized form will break
         * client-member compatibility.
         */
        CLIENT_COMPATIBILITY,
        /**
         * Class is not used in client-member communication, however is part of public API so cannot be
         * migrated to {@code IdentifiedDataSerializable}.
         */
        PUBLIC_API,
        /**
         * Class may or may not be used in serialized form in the context of Hazelcast however is {@code Serializable}
         * due to conventions and cannot be converted to {@code IdentifiedDataSerializable}. Examples:
         * <ul>
         *     <li>inheritance from a class external to Hazelcast e.g. {@code CacheEntryEvent}, which imposes
         *     serializability even when not required</li>
         *     <li>inheritance from a class external to Hazelcast, serializability may be desired, however
         *     {@code IdentifiedDataSerializable} requires a default no-args constructor and non-final fields which
         *     is not always possible (e.g. subclasses of {@code java.util.EventObject} & {@code java.lang.Exception})</li>
         *     <li>a {@code Comparator} which should by convention also implement {@code Serializable}.</li>
         * </ul>
         */
        OTHER_CONVENTION,
```

sancar:
Ok. BinaryInterface does not used just protocol anymore. 

And If we are marking every serilizable than there is not much meaning left to BinaryInterface annotation. We can just follow Serilizable classes for this purpose ? 

I think description needs to be more precise.
```Class is not used in client-member communication, however is part of public API so cannot migrated to {@code IdentifiedDataSerializable}.```
Just cannot be migrated to IdentifiedDataSerializable . Or can not change binary representation. 
What if I make it Portable ? 
What if I add a field  ?
And can I change them between major, minor or patches ?

It still says 
applicable to 3.x
 * - NEVER CHANGE THEM
 * - NEVER MAKE THEM IMPLEMENT THE VERSIONED INTERFACE
on the interface javadoc ?

vbekiaris:
Very good point @sancar , "not migratable to IDS" is a separate concern from "class is used in client-member communication".

PR #9849, opened 2017-02-09
Don't cache null values in jcache near cache

ahmetmircik:
This was a regression only in 3.8, fixing it.


ahmetmircik:
tiny fix merging it.

PR #9855, opened 2017-02-10
Added convenience methods to ILogger

Donnerbart:
I saw that we were missing some convenience methods in `ILogger`, so we had to use `logger.log(Level.XXX, ...)` and `logger.isLoggable(Level.XXX)` in a lot of places.

Less imports, shorted code, happy developers :)


mmedenjak:
@Donnerbart I was thinking about adding some lazy-evaluation methods like what we already have in `com.hazelcast.spi.hotrestart.impl.gc.GcLogger`. WDYT?

Donnerbart:
Uuuh, I like it, especially with the built-in `String.format()`. Should we merge this first? Then we know quite good which methods are used and needed.

Issue #9816, opened 2017-02-07
No DataSerializerFactory registered for namespace: -301858816 ClientReliableTopicProxy,  client compatibility 


Danny-Hazelcast:
```
SEVERE: hz.client_0 [HZ] [3.7] Failed asynchronous execution of execution callback: com.hazelcast.client.util.ClientDelegatingFuture$DelegatingExecutionCallback@2d7e12edfor call Cli
entMessage{length=62, correlationId=41465, messageType=190a, partitionId=54, isComplete=true, isRetryable=false, isEvent=false, writeOffset=0}
com.hazelcast.nio.serialization.HazelcastSerializationException: No DataSerializerFactory registered for namespace: -301858816
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:114)
    at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:52)
    at com.hazelcast.internal.serialization.impl.StreamSerializerAdapter.read(StreamSerializerAdapter.java:46)
    at com.hazelcast.internal.serialization.impl.AbstractSerializationService.toObject(AbstractSerializationService.java:172)
    at com.hazelcast.ringbuffer.impl.client.PortableReadResultSet.iterator(PortableReadResultSet.java:63)
    at com.hazelcast.client.proxy.ClientReliableTopicProxy$MessageRunner.onResponse(ClientReliableTopicProxy.java:231)
    at com.hazelcast.client.proxy.ClientReliableTopicProxy$MessageRunner.onResponse(ClientReliableTopicProxy.java:197)
    at com.hazelcast.client.util.ClientDelegatingFuture$DelegatingExecutionCallback.onResponse(ClientDelegatingFuture.java:227)
    at com.hazelcast.client.util.ClientDelegatingFuture$DelegatingExecutionCallback.onResponse(ClientDelegatingFuture.java:203)
    at com.hazelcast.spi.impl.AbstractInvocationFuture$1.run(AbstractInvocationFuture.java:251)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)
    at com.hazelcast.util.executor.HazelcastManagedThread.executeRun(HazelcastManagedThread.java:76)
[...]

PR #9859, opened 2017-02-10
Removes support for compressed IDS header

vbekiaris:
See https://github.com/hazelcast/hazelcast-enterprise/issues/1295

requires EE counterpart


vbekiaris:
Fixes #9816

PR #9862, opened 2017-02-11
Fix/scheduled executor/master

### LINKED ISSUE #9788: Scheduled Executor wrong number of tasks returned on getAllScheduled Kill Random Member

### LINKED ISSUE #9846: TaskDefinition should be removed from client protocol
`com.hazelcast.scheduledexecutor.impl.TaskDefinition` is used in serialized form in client protocol. It should blow into fields and put into method definition in protocol.

sancar:
@tkountis 
`GetAllScheduled` is called on each member and constructs the accumulated map of tasks per member. When a member goes down, we then have incorrect count of tasks, since we are not accessing that information through the partition-threads. Eventually the numbers will be correct, but  we can improve this doing the same operation per partition rather than per member. 

However, fetching all tasks per partition, will miss tasks owned by members (`scheduleOnMember*`) which we still have to use the old approach to collect.

Fixes #9788 
@sancar 
TaskDefinition was used as serialized to a single string(urn) in client protocol. This pr https://github.com/hazelcast/hazelcast-client-protocol/pull/61  expands those usages as different parameter.
Besides expanding parameters, messageTasks going to specifically to address or partition are separated.

fixes #9846


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/scheduledexecutor/ScheduledExecutorTaskCancelFromAddressMessageTask.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public String getMethodName() {
+        return "cancel";
+    }
+
+    @Override
+    public Object[] getParameters() {
+        return new Object[] { parameters.mayInterruptIfRunning };
```
tkountis: SchedulerName, TaskName ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/scheduledexecutor/ScheduledExecutorTaskCancelFromPartitionMessageTask.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public String getMethodName() {
+        return "cancel";
+    }
+
+    @Override
+    public Object[] getParameters() {
+        return new Object[] {parameters.mayInterruptIfRunning };
```
tkountis: Same here

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/scheduledexecutor/ScheduledExecutorTaskDisposeFromAddressMessageTask.java
```diff
@@ -91,6 +84,6 @@ public String getMethodName() {
 
     @Override
     public Object[] getParameters() {
-        return new Object[] { parameters.handlerUrn };
+        return null;
```
tkountis: Same

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/scheduledexecutor/ScheduledExecutorTaskDisposeFromPartitionMessageTask.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public String getMethodName() {
+        return "dispose";
+    }
+
+    @Override
+    public Object[] getParameters() {
+        return null;
```
tkountis: same, please check other places too

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/scheduledexecutor/ScheduledExecutorTaskCancelFromAddressMessageTask.java
sancar: This parameters are not parameters of protocol. These are parameters passed by user in the API. 
```
boolean cancel(boolean mayInterruptIfRunning);
```

For example, look at
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapPutMessageTask.java#L69
Eventough map name is in parameters, it is not used in `getParameters`

tkountis: ok, but these parameters are used for the security interceptor, and even though one can get the name of the scheduler (distributedObject), the action name (eg. cancel) and the mayInterrupt flag, they can't see which task this is targeted to. So **unless I am missing sth**, we are missing the taskName here at the very least ?
Also, [this](https://github.com/hazelcast/hazelcast/pull/9862/commits/1a3cf3bf7ba1201418a300865f3b178dd754a4c4#diff-2f3f5d0869f5bd6f5adcd16b6339bf9cR93) needs to adhere to the same pattern.

asimarslan:
This PR moved in to #9878

PR #9868, opened 2017-02-13
Fire MERGE_FAILED when node fails to merge to discovered cluster

### LINKED ISSUE #9856: Adding `MERGE_FAILED` to LifeCycleState's
Currently there's only `MERGING, MERGED` but no event fired for a failed merge. This would be informative and further give an idea to user to understand a reason of a `SHUTTING_DOWN` event since a failed merge is followed by a shutdown.

This can be added to both 3.7.x and 3.8.x branches.

mdogan:
Currently there are only `MERGING`, `MERGED` events, but no event fired for a failed merge.

Fixes #9856

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/LifecycleEvent.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/LifecycleEvent.java
@@ -42,6 +42,7 @@ public final class LifecycleEvent {
         SHUTDOWN,
         MERGING,
         MERGED,
+        MERGE_FAILED,
         CLIENT_CONNECTED,
         CLIENT_DISCONNECTED
     }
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/core/LifecycleEvent.java
```diff
@@ -42,6 +42,7 @@
         SHUTDOWN,
         MERGING,
         MERGED,
+        MERGE_FAILED,
```
jerrinot: this change ordinal value of the `CLIENT_`. is it ok?

jerrinot: I guess some serialization strategies might break

mdogan: `LifecycleEvent` is not serialized. It should be completely local event. Relying ordinals unless explicitly defined should be considered as bug, I think.

jerrinot: mr. Bloch agrees with you:

PR #9874, opened 2017-02-13
Updates in Serializable & DataSerializable conventions test

vbekiaris:
DataSerializableConventionsTest has been updated to reflect the following conventions:
- classes in public packages are now checked for compliance (while they were previously excluded)
- only concrete classes are checked for compliance. It is OK to annotate interfaces or abstract classes with @BinaryInterface as an indication to implementers, however each implementation has to be considered separately.
- as a consequence of the above `@BinaryInterface` is not considered as inherited.
- introduces a separate `@SerializableByConvention` annotation to mark classes which implement `Serializable` or `DataSerializable` and:
    * are not used in serialized form, however extend from a `Serializable` class
    * are used in serialized form, however cannot be converted due to extending from a superclass with no default constructor or fields which cannot be set after construction  (e.g. subclasses of `Throwable`, `java.util.EventObject`, `java.security.Permission`)

Due to the large number of classes affected, this PR is split in several commits, see commit messages for details.
An EE counterpart is required

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java
@@ -23,7 +23,7 @@ import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
- * Indicates that the annotated class cannot be converted to {@code IdentifiedDataSerializable} due to other conventions.
+ * Indicates that the annotated class cannot be converted to {@link IdentifiedDataSerializable} due to other conventions.
  * The annotation's {@link #value()} provides reasoning why class cannot be converted. Classes annotated with
  * {@code SerializableByConvention} are excluded from compliance tests by {@code DataSerializableConventionsTest}.
  */
@@ -34,17 +34,19 @@ public @interface SerializableByConvention {
 
     enum Reason {
         /**
-         * Class definition is part of public API so cannot be migrated to {@code IdentifiedDataSerializable}.
+         * Class definition is part of public API so cannot be migrated to {@link IdentifiedDataSerializable} in
+         * minor or patch releases.
          */
         PUBLIC_API,
         /**
          * Class is {@code Serializable} due to conventions and cannot be converted to {@code IdentifiedDataSerializable}.
          * Examples:
          * <ul>
-         *     <li>inheritance from a class external to Hazelcast e.g. {@code CacheEntryEvent}, which imposes
-         *     serializability even when not required</li>
-         *     <li>inheritance from a class external to Hazelcast, however {@code IdentifiedDataSerializable} requires a
-         *     default no-args constructor and non-final fields which  is not always possible (e.g. subclasses of
+         *     <li>inheritance from a class external to Hazelcast e.g. {@link javax.cache.event.CacheEntryEvent} imposes
+         *     serializability to its implementation in {@link com.hazelcast.cache.impl.CacheEntryEventImpl} even though
+         *     the latter is never serialized</li>
+         *     <li>inheritance from a class external to Hazelcast, however {@link IdentifiedDataSerializable} requires a
+         *     default no-args constructor and non-final fields which is not always possible (e.g. subclasses of
          *     {@code java.util.EventObject} & {@code java.lang.Exception})</li>
          *     <li>a {@code Comparator} which should by convention also implement {@code Serializable}.</li>
          * </ul>
--- a/hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java
@@ -39,17 +39,23 @@ public @interface SerializableByConvention {
          */
         PUBLIC_API,
         /**
-         * Class is {@code Serializable} due to conventions and cannot be converted to {@code IdentifiedDataSerializable}.
+         * Class is {@code Serializable} due to inheritance or other coding conventions and cannot or is not desirable
+         * to have it converted to {@link IdentifiedDataSerializable}.
          * Examples:
          * <ul>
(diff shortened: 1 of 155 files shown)


vbekiaris:
run-lab-run

vbekiaris:
run-lab-run

Donnerbart:
> as a consequence of the above `@BinaryInterface` is not considered as inherited
This sounds like fun with hierarchies. One idea: we could throw an AssertionError for all unannotated classes, once a `@BinaryInterface` has been found. Then we either annotate them or use `@SuppressWarnings("binaryinterface")` to mark them accordingly.

Donnerbart:
Do I get it right that `@SerializableByConvention` annotated classes **should** be `@IdentifiedDataSerializable`, but can't be due to e.g. client compatibility. And that we should convert them in Hazelcast 4.0?

vbekiaris:
Those `@SerializableByConvention` with value `PUBLIC_API` can be migrated to `IdentifiedDataSerializable` in 4.0. But `Exception`, `EventObject` etc subclasses (with the default value `INHERITANCE`) cannot be migrated.
Client compatibility is always indicated with `@BinaryInterface`.

Donnerbart:
Did I understand it correctly that there is a double meaning of `SerializableByConvention#Reason.INHERITANCE`? It always means that a class is inheriting its `@Serializable` from a class where we cannot change it (since its base class is external). But in some cases
* we don't care (the **serialization is unused** case)
* we might want to change it, but it's not possible (the **serialization is used** case)

I don't think it's worth to reflect this difference via distinct `Reason` enums though. For example a "we don't care class" could be used on the wire in the future, then the Reason would need to be changed. But we anyway can't change it, so this is just additional work.

But I see that it's a lot of classes which need the annotation with `Reason.INHERITANCE`. And this could get annoying, e.g. for new `Exception`s which will most probably will always be forgotten to be annotated until the PR builder runs. So I'm asking myself if it would be a good idea (and possible) to replace this reason with a whitelist of well known superclasses. Then `SerializableByConvention` would be a pure TODO list for the next major version. And we would just have a single place where all external classes are collected.

vbekiaris:
Thanks @Donnerbart for the review. I agree that having a whitelist of well-known `Serializable` superclasses makes sense and will make for a more pleasant development experience, without surprises in the PR builder. I have addressed your review comments and added  `java.lang.Throwable`,  `java.util.EventObject`, `java.security.Permission` & `java.security.PermissionCollection` in the whitelist.

vbekiaris:
run-lab-run

tombujok:
Looks good. I was just wondering that there are some classes annotated without giving a reason in the annotation. Does it mean a default reason?

vbekiaris:
Yes, default reason for `@SerializableByConvention` is [`INHERITANCE`](https://github.com/vbekiaris/hazelcast/blob/9ac71597a32f62f8c8fc000c428ea6ec9110cd6c/hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java#L33-L33)

hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java
```diff
@@ -0,0 +1,56 @@
...
+         */
+        PUBLIC_API,
+        /**
+         * Class is {@code Serializable} due to conventions and cannot be converted to {@code IdentifiedDataSerializable}.
+         * Examples:
+         * <ul>
+         *     <li>inheritance from a class external to Hazelcast e.g. {@link javax.cache.event.CacheEntryEvent} imposes
+         *     serializability to its implementation in {@link com.hazelcast.cache.impl.CacheEntryEventImpl} even though
+         *     the latter is never serialized</li>
+         *     <li>inheritance from a class external to Hazelcast, however {@link IdentifiedDataSerializable} requires a
```
Donnerbart: This example is outdated, since those classes are now whitelisted. And I think in general superclasses without default constructor should be whitelisted in the future. So maybe we should mention or `{@link}` the whitelist here?

hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java
```diff
@@ -0,0 +1,56 @@
...
+         *     default no-args constructor and non-final fields which is not always possible (e.g. subclasses of
+         *     {@code java.util.EventObject} & {@code java.lang.Exception})</li>
```
Donnerbart: Just nitpicking, but `&` will be a forbidden character in Java 8 JavaDocs. So either `&amp;` or `and` should be used :)

hazelcast/src/main/java/com/hazelcast/nio/serialization/SerializableByConvention.java
vbekiaris: Updated javadoc as follows:
```
        /**
         * Class is {@code Serializable} due to inheritance or other coding conventions and cannot or is not desirable
         * to have it converted to {@link IdentifiedDataSerializable}.
         * Examples:
         * <ul>
         *     <li>inheritance from a {@code Serializable} class external to Hazelcast e.g.
         *     {@link javax.cache.event.CacheEntryEvent} is serializable however its implementation in
         *     {@link com.hazelcast.cache.impl.CacheEntryEventImpl} is never used in serialized form, so there
         *     is no interest in converting this class to {@link IdentifiedDataSerializable}</li>
         *     <li>a {@code Comparator} which should by convention also implement {@code Serializable}.</li>
         * </ul>
         *
         * Classes whose superclasses are {@link java.io.Serializable} without a default constructor or with private
         * final fields, such as {@link Throwable} or {@link java.util.EventObject}, cannot be converted to
         * {@link IdentifiedDataSerializable}, so these are not annotated with {@code SerializableByConvention}.
         * Instead these classes are white-listed in
         * {@code com.hazelcast.internal.serialization.impl.DataSerializableConventionsTest#SERIALIZABLE_WHITE_LIST}
         * and are excluded from conventions testing.
         */
```

(cannot make a `@link` here to test code)

Issue #9869, opened 2017-02-13
[lock] LockLeaseMemberBounceTest.leaseShouldExpireWhenMemberBouncing


hasancelik:
```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at com.hazelcast.concurrent.lock.LockLeaseMemberBounceTest$Driver.doAssert(LockLeaseMemberBounceTest.java:122)
	at com.hazelcast.concurrent.lock.LockLeaseMemberBounceTest$1.run(LockLeaseMemberBounceTest.java:80)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:967)
	at com.hazelcast.concurrent.lock.LockLeaseMemberBounceTest.leaseShouldExpireWhenMemberBouncing(LockLeaseMemberBounceTest.java:76)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast/1238/testReport/junit/com.hazelcast.concurrent.lock/LockLeaseMemberBounceTest/leaseShouldExpireWhenMemberBouncing/

https://hazelcast-l337.ci.cloudbees.com/view/Sonar/job/Hazelcast-3.x-sonar/com.hazelcast$hazelcast/1403/testReport/junit/com.hazelcast.concurrent.lock/LockLeaseMemberBounceTest/leaseShouldExpireWhenMemberBouncing/

hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 300000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
	at java.util.concurrent.FutureTask.awaitDone(FutureTask.java:425)
	at java.util.concurrent.FutureTask.get(FutureTask.java:187)
	at com.hazelcast.test.bounce.BounceMemberRule.waitForFutures(BounceMemberRule.java:375)
	at com.hazelcast.test.bounce.BounceMemberRule.testWithDuration(BounceMemberRule.java:239)
[...]

Issue #9256, opened 2016-11-15
[lock] lease lock not released under cluster shutdown test,  3.8-SANPSHOT


Danny-Hazelcast:
nightly builds 

https://hazelcast-l337.ci.cloudbees.com/view/shutdown/job/shutdown-lease-lock/
https://hazelcast-l337.ci.cloudbees.com/view/rollup/job/rollup-lease-lock/

some times fail, and the lease locks are not released.
members are being shutdown and restarted.

https://hazelcast-l337.ci.cloudbees.com/view/rollup/job/rollup-lease-lock/8/console
http://54.87.52.100/~jenkins/workspace/rollup-lease-lock/3.7.3/2016_11_15-03_08_36

```
HzMember1HZ validate hz.lock.validate.LeaseLockAssert threadId=0 global.AssertionException: lease-lock292.isLocked()==true 
```

however this fail is rare,

pveentjer:
Can you provide a better title because the current title doesn't provide any insights what the problem is

jerrinot:
@Danny-Hazelcast can you link the test scenario please?

mdogan:
Base class: https://github.com/Danny-Hazelcast/hzCmd-bench/blob/master/src/main/java/hz/lock/validate/base/LockValidate.java

Lease Lock: https://github.com/Danny-Hazelcast/hzCmd-bench/blob/master/src/main/java/hz/lock/validate/LeaseLock.java

Assertion: https://github.com/Danny-Hazelcast/hzCmd-bench/blob/master/src/main/java/hz/lock/validate/LockAssert.java

Main timestep function is:
```java
    public void timeStep() {

        int i = random.nextInt(count);
        ILock lock = getLock(i);

        lock.lock(seconds, TimeUnit.SECONDS);
        incMapIdx(i);
        increments[i]++;
    }
```

Assertion function is:
```java
    public void timeStep() throws AssertionException {
        ...
        for (int i=0; i<count; i++) {
            if( getIncrementFor(i) != getMapIdx(i) ){
                throw new AssertionException(getInfoFor(i));
            }
        }

        for (int i=0; i<count; i++) {
            ILock lock = getLock(i);
            if(lock.isLocked()){
                throw new AssertionException(lock.getName()+" isLocked()==true");
            }
        }
        setRunning(false);
    }
```

As far as I see, `incMapIdx(i)` can be called concurrently if locking thread is de-scheduled long enough to lock lease expire. 
```java
[...]

PR #9875, opened 2017-02-13
Retry UnlockIfLeaseExpiredOperation if exception is retryable

mmedenjak:
When the UnlockIfLeaseExpiredOperation operation fails the lock will remain locked forever, unless the lock is migrated causing the operation to be scheduled again. This can happen for instance if the partition is migrating because a member is joining or leaving the cluster. Since some exceptions can allow the operation to be retried, we will retry the operation in a similar way to how the invocation system retries the operation.

State before the fix : 
1. When the lock was replicated or migrated the unlock would be scheduled on every replica. E.g. on migration - https://github.com/mmedenjak/hazelcast/blob/b47842101e65c5227f6711aff6b9a9ee4e9a1016/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java#L256-L256
2. When the schedule is about to be processed, every replica would create an operation and execute it on its partition thread, thus unlocking all replicas
3. On the partition owner, the `UnlockIfLeaseExpiredOperation#shouldBackup` method would return true, causing the invocation system to send the `UnlockBackupOperation` to the replicas. This would again unlock the already unlocked lock

State after the fix : 
1. When the lock was replicated or migrated the unlock would be scheduled on every replica. E.g. on migration - https://github.com/mmedenjak/hazelcast/blob/b47842101e65c5227f6711aff6b9a9ee4e9a1016/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java#L256-L256
2. When the schedule is about to be processed, every replica would create an operation and invoke it on the partition owner. 
3. On the partition owner, the `UnlockIfLeaseExpiredOperation#shouldBackup` method would return true, causing the invocation system to send the `UnlockBackupOperation` to the replicas. This would again unlock the already unlocked lock
4. The partition owner receives as many `UnlockIfLeaseExpiredOperation` operations as there are replicas. This is ok since these operations are idempotent (on `lock` and `extendLeaseTime` the version is increased causing the unlock to fail if a different operation locked the lock in the meantime)

We could improve on this by not scheduling the operation on every replica.

Fixes :
https://github.com/hazelcast/hazelcast/issues/9869
https://github.com/hazelcast/hazelcast/issues/9256

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockEvictionProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockEvictionProcessor.java
@@ -23,7 +23,6 @@ import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.ObjectNamespace;
 import com.hazelcast.spi.Operation;
-import com.hazelcast.spi.OperationAccessor;
 import com.hazelcast.spi.OperationResponseHandler;
 import com.hazelcast.spi.OperationService;
 import com.hazelcast.spi.exception.RetryableException;
@@ -70,16 +69,12 @@ public final class LockEvictionProcessor implements ScheduledEntryProcessor<Data
     private void submit(UnlockOperation operation, Data key) {
         int partitionId = nodeEngine.getPartitionService().getPartitionId(key);
         OperationService operationService = nodeEngine.getOperationService();
-        operation.setNodeEngine(nodeEngine);
-        operation.setServiceName(SERVICE_NAME);
         operation.setPartitionId(partitionId);
-        OperationAccessor.setCallerAddress(operation, nodeEngine.getThisAddress());
-        operation.setCallerUuid(nodeEngine.getLocalMember().getUuid());
         operation.setOperationResponseHandler(unlockResponseHandler);
         operation.setValidateTarget(false);
         operation.setAsyncBackup(true);
 
-        operationService.execute(operation);
+        operationService.invokeOnTarget(SERVICE_NAME, operation, nodeEngine.getThisAddress());
     }
 
     private class UnlockResponseHandler implements OperationResponseHandler {
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/operations/UnlockIfLeaseExpiredOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/operations/UnlockIfLeaseExpiredOperation.java
@@ -34,12 +34,7 @@ public final class UnlockIfLeaseExpiredOperation extends UnlockOperation {
 
     private int version;
 
-    /**
-     * This constructor should not be used to obtain an instance of this class; it exists to fulfill IdentifiedDataSerializable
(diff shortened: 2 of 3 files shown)


mdogan:
As far as I see, this change fixes an issue but creates a new one :)

Before: It may happen that `UnlockIfLeaseExpiredOperation` execution fails because of a partition migration. That makes lock remain forever.

After: It may happen that primary replica executes the `UnlockIfLeaseExpiredOperation` successfully but fails to send backups. And then primary crashes. In that case, lock will remain forever.

mdogan:
I think we can fix both issues using 

```java
    private void submit(UnlockOperation operation, Data key) {
        int partitionId = nodeEngine.getPartitionService().getPartitionId(key);
        operation.setNodeEngine(nodeEngine);
        operation.setServiceName(SERVICE_NAME);
        operation.setPartitionId(partitionId);
        OperationAccessor.setCallerAddress(operation, nodeEngine.getThisAddress());
        operation.setCallerUuid(nodeEngine.getLocalMember().getUuid());
        operation.setOperationResponseHandler(unlockResponseHandler);
        operation.setValidateTarget(false);
        operation.setAsyncBackup(true);

        nodeEngine.getOperationService().invokeOnTarget(SERVICE_NAME, operation, nodeEngine.getThisAddress());
    }
```

Invocation with `invokeOnTarget()`  on local member will handle retry cases. And since this invocation will be performed on all replicas locally, each replica will eventually unlock.

mmedenjak:
@mdogan I agree, that is an easy fix. I am not sure if that can still lead to some being unlocked and some not. But regarding to the idea that the partition owner sends the unlock to the replicas - I'm guessing that option would also include scheduling the eviction on the backup replica when it is promoted, since the owner might have died before, during or after unlocking.

mmedenjak:
@jerrinot @mdogan updated the solution to use Mehmet's suggestion, please review.

mmedenjak:
@jerrinot @mdogan can you please review? I think the change is not that big and should fix the solution without changing the unlock mechanism (every replica unlocks only itself).

mdogan:
@mmedenjak: Sorry, I forgot to approve earlier.

hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockEvictionProcessor.java
```diff
@@ -70,16 +69,12 @@ private void sendUnlockOperation(Data key, int version) {
...
         operation.setPartitionId(partitionId);
-        OperationAccessor.setCallerAddress(operation, nodeEngine.getThisAddress());
-        operation.setCallerUuid(nodeEngine.getLocalMember().getUuid());
         operation.setOperationResponseHandler(unlockResponseHandler);
         operation.setValidateTarget(false);
         operation.setAsyncBackup(true);
 
-        operationService.execute(operation);
+        operationService.invokeOnTarget(SERVICE_NAME, operation, nodeEngine.getThisAddress());
     }
```
tombujok: Looks good. The other way of doing it is having a callback with execute. The invokeOnTarget can be back-pressured. But it's fine.

PR #9887, opened 2017-02-14
Fixing wrong map usage

sancar:
Since putting to map when iterating over it may cause undefined
behaviour that usage is chaged to a safe one.


sancar:
run-lab-run

PR #9898, opened 2017-02-14
Made deprecated read-only config classes inner private classes

Donnerbart:
* made all read-only config classes private static inner classes
* made all `getAsReadOnly()` methods package private if possible
* made all `getAsReadOnly()` methods which are used in code `@PrivateApi`
* made all `getAsReadOnly()` methods return the parent class
* renamed `MemberAttributeConfig.asReadOnly()` to `getAsReadOnly()`

Some `getAsReadOnly()` methods are used in our code. I didn't fix this but left those few methods `public` and annotated them with `@PrivateApi`. Since all `getAsReadOnly()` return the parent class now, we don't leak the read-only implementation anymore.

I checked with @vbekiaris that the classes with `@BinaryInterface` can be moved and made private. I didn't change any content or serialization methods. `DataSerializableConventionsTest` and `EnterpriseDataSerializableConventionsTest` are still green. EE still compiles.

I just adjusted the tests to compile again where necessary. There is some more cleanup needed. Also some code coverage will be missing for some classes. This will be fixed in a separate PR (this one is already big enough).


Donnerbart:
run-lab-run

Donnerbart:
@enesakar Do we really have to wait for HZ 4.0 to clean this up?

enesakar:
Yes, no harm to keep it deprecated.

Donnerbart:
I have a different opinion on this. We have a bunch of deprecated classes and `getAsReadOnly()` methods which as still in use and just deprecated because we want to move those classes from public to private. They are not going to be deleted, just moved. So we have a lot of "deprecated xxx still used" warnings in the code, which won't go away without this cleanup. This is a harm to code readability.

The deprecation is out there since 3.7 and we just postponed the cleanup since the `getAsReadOnly()` methods were not deprecated together with the config classes. So this was a "let's be super nice to our users", add the missing deprecation in 3.8 and do the cleanup in 3.9. So the read-only config classes are deprecated since about a year now. And I doubt that there are anyway users of them.

I would really like to see this cleanup being done in the 3.x versions.

enesakar:
I did not fell very safe, once we told they will be removed in 4.0. but I understand your points too. 
@jerrinot wdyt?

jerrinot:
If a class is clearly public API (e.g. `CacheConfig`) and has a public method then you cannot make this method private just by annotating it with a custom annotation. 

I see little value to do this change in Hazelcast 3.9. I see risks and additional confusion. In Hazelcast 4.0 we should reconsider this `asReadOnly()`. If it's not part of public API then it should not be visible and this should be enforced by a compiler.

pveentjer:
I disagree. We have a deprecation policy of 1 version and we should get rid of code as soon as possible. As long as it remains it keeps causing confusion.

Donnerbart:
I think part of the confusion Jaromir meant is the `@PrivateAPI` on the few `getAsReadOnly()` methods which are still public. They need to be public since they are used in our own code (and it's not in the scope of this PR to clean that up). But this should be a separate issue than the removal of deprecated code.

Since the `getAsReadOnly()` methods return the normal Config class now, and not the deprecated ReadOnly class anymore, it doesn't hurt that much to have those few public methods left - with or without `@PrivateAPI` annotation.

PR #9909, opened 2017-02-14
Removed onPutError

ahmetmircik:


hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/store/AbstractNearCacheRecordStore.java
```diff
@@ -468,19 +468,12 @@ protected R updateReservedRecordInternal(K key, V value, R reservedRecord, long
...
 
-        try {
-            nearCacheStats.decrementOwnedEntryMemoryCost(getTotalStorageMemoryCost(key, reservedRecord));
-
-            updateRecordValue(reservedRecord, value);
-            reservedRecord.casRecordState(UPDATE_STARTED, READ_PERMITTED);
+        nearCacheStats.decrementOwnedEntryMemoryCost(getTotalStorageMemoryCost(key, reservedRecord));
 
-            nearCacheStats.incrementOwnedEntryMemoryCost(getTotalStorageMemoryCost(key, reservedRecord));
-            onPut(key, value, reservedRecord, null);
```
Donnerbart: You are also removing the `onPut()` call. If I get it right this had no effect, since the forth parameter is `null`, so it would never do anything in the EE implementation. Is that correct?

ahmetmircik: correct no need on ee

PR #9916, opened 2017-02-15
adding identified predicate builder (CompositePredicate)

sancar:
Adding identified predicate builder (CompositePredicate) for non java clients to be able to support predicate builder. 
PredicateBuilder was only DataSerializable before. Our other clients(c++ , nodejs , csharp ..) do not support DataSerializable.


sancar:
Decided not to merge.

PR #9929, opened 2017-02-15
Making clusterVersion non-null at all times

tombujok:


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateManager.java
```diff
@@ -237,7 +237,7 @@ private void validateNodeCompatibleWith(Version clusterVersion) {
 
     // validate transition from current to newClusterVersion is allowed
     private void validateClusterVersionChange(Version newClusterVersion) {
-        if (clusterVersion != null && clusterVersion.getMajor() != newClusterVersion.getMajor()) {
+        if (clusterVersion != null && !clusterVersion.isUnknown() && clusterVersion.getMajor() != newClusterVersion.getMajor()) {
```
vbekiaris: `clusterVersion != null` condition could be removed

tombujok: Thx, fixed!

vbekiaris:
In `ClusterStateChange::validate` cluster version was already guarded against going back to undefined by the `(newState == null)` condition. We should a new condition to ensure that when the cluster state change is of type `ClusterVersion`, requested new cluster version cannot be `UNKNOWN`.

tombujok:
@vbekiaris great finding! I added a fix for that. Could you have a look?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterStateChange.java
```diff
@@ -57,6 +58,12 @@ public void validate() {
             throw new IllegalArgumentException("Invalid null state");
         }
 
+        if (isOfType(Version.class)) {
+            if (((Version) newState).isUnknown()) {
+                throw new IllegalArgumentException("Cannot change Version to UNKNOWN!");
+            }
+        }
+
```
tombujok: Here's the fix for UNKNOWN version

vbekiaris:
👍 looks good!

PR #9951, opened 2017-02-20
Metric Registry: Auto convert getter name into a property name

jerrinot:
Example
```java
@Probe
public int getFoo() {
 [...]
}
```

Now generates a probe named `foo` instead of `getFoo`

This increases usefulness of implicit probe names and does not force
you to use explicit names only to remove the get- prefix.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/MethodProbe.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/MethodProbe.java
@@ -37,6 +37,7 @@ import static com.hazelcast.internal.metrics.impl.ProbeUtils.TYPE_PRIMITIVE_LONG
 import static com.hazelcast.internal.metrics.impl.ProbeUtils.TYPE_SEMAPHORE;
 import static com.hazelcast.internal.metrics.impl.ProbeUtils.getType;
 import static com.hazelcast.internal.metrics.impl.ProbeUtils.isDouble;
+import static com.hazelcast.util.StringUtil.getterIntoProperty;
 import static java.lang.String.format;
 
 /**
@@ -63,8 +64,10 @@ abstract class MethodProbe implements ProbeFunction {
     }
 
     private String getName(String namePrefix) {
-        String name = method.getName();
-        if (!probe.name().equals("")) {
+        String name;
+        if (probe.name().equals("")) {
+            name = getterIntoProperty(method.getName());
+        } else {
             name = probe.name();
         }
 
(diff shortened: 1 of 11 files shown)


hazelcast/src/main/java/com/hazelcast/util/StringUtil.java
```diff
@@ -247,4 +251,38 @@ public static int lastIndexOf(String input, char ch) {
...
+     * It's written defensively, when output is not a getter then it
+     * return the original name.
+     *
+     * It converters name starting with the get- prefix only. When a getter
+     * starts with is- prefix (=boolean) then it does not convert it.
+     *
+     * @param getterName
+     * @return property matching the given getter
+     */
+    public static String getterIntoProperty(String getterName) {
```
jerrinot: this is where the magic happens

hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/MethodProbe.java
```diff
@@ -63,8 +64,10 @@ void register(MetricsRegistryImpl metricsRegistry, Object source, String namePre
     }
 
     private String getName(String namePrefix) {
-        String name = method.getName();
```
pveentjer: we need to make sure that we are not going to generate litter when the diagnostics writes to e.g. file. I don't have the logic of this part in my head at the moment.

jerrinot: `getName()` is private and is used during a probe registration only.

hazelcast/src/main/java/com/hazelcast/util/StringUtil.java
```diff
@@ -247,4 +253,38 @@ public static int lastIndexOf(String input, char ch) {
...
```
pveentjer: The name is not revealing the intent; I can't guess based on the name of this method what the purpose is.

jerrinot: my original choice was something like `stringGetterPrefix()`. but there is more going on - e.g. conversion to lower case. 

bear in mind this is located in a class `StringUtil`. do you have a better name?

pveentjer: toPropertyName? And the argument I would call 'getMethodName'

jerrinot: could do

pveentjer:
:+1:

PR #9973, opened 2017-02-23
Use URI as config location or instance name in CachingProvider.getCacheManager

### LINKED ISSUE #9958: JCache bootstrap should have a way to reconciliate existing `HazelcastInstance`
This is an issue raised on the Spring Boot tracker for which I think there's nothing we can do. Spring Boot auto-configures hazelcast since it's not specific to caching. It turns you might use hazelcast for that as well.

If you chose to use JCache, then we bootstrap via the standard `JCache` API (obviously) and doing so leads to Hazelcast starting again. I think there should be a way for Hazelcast to figure out that an `HazelcastInstance` has already been started on this `ClassLoader` and that configuration URI and reuse that rather than creating a fresh one. As a user of the `JCache` API there's not much we can do.

We already have a [hazelcast specific hack](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/JCacheCacheConfiguration.java#L126) because Hazelcast' doesn't honour the URI parameter of the `CacheProvider#getCacheManager` properly.

You can easily reproduce using [this sample](https://github.com/snicoll-scratches/test-hazelcast-jcache): uncomment the `spring.cache.type` property in `application.properties` and you'll see two members starting in the cluster.

vbekiaris:
When failing to locate or create a `HazelcastInstance` via Properties, use URI as config location or instance name in CachingProvider.getCacheManager as described in [reference manual](http://docs.hazelcast.org/docs/3.8/manual/html-single/index.html#scoping-to-join-clusters)

Fixes #9958

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
@@ -16,7 +16,6 @@
 
 package com.hazelcast.cache.impl;
 
-import com.hazelcast.cache.HazelcastCachingProvider;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
 import com.hazelcast.core.Hazelcast;
@@ -29,6 +28,11 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Properties;
+
+import static com.hazelcast.cache.HazelcastCachingProvider.HAZELCAST_CONFIG_LOCATION;
+import static com.hazelcast.cache.HazelcastCachingProvider.HAZELCAST_INSTANCE_ITSELF;
+import static com.hazelcast.cache.HazelcastCachingProvider.HAZELCAST_INSTANCE_NAME;
+import static com.hazelcast.util.StringUtil.isNullOrEmptyAfterTrim;
 
 /**
  * Provides server cachingProvider implementation. <p>This implementation is used by {@link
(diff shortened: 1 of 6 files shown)


snicoll:
Just to make sure I get this right, here is a scenario in Spring Boot:

1. Hazelcast is started via the general auto-configuration. An `HazelcastInstance` is created using the default configuration file (`hazelcast.xml`) at the root of the classpath
2. A `javax.cache.CacheManager` is requested after, using Hazelcast as the provider. With this fix, Hazelcast will realize that no URI = default and will check for the presence of an `HazelcastInstance`that match those criteria, and therefore will not start another instance

Am I getting this properly? 

If I am, the scenario above can be tuned with an URI to a configuration file (The hazelcast instance is created with `foo-bar-hazelcast.xml` and if we call `CacheProvider.getCacheManager("foo-bar-hazelcast.xml", ...)` it should also realize such a URI has been processed and reused the existing `HazelcastInstance`.

If I got this wrong, please let me know as Spring Boot exposes those limitations at the moment.

vbekiaris:
Hi @snicoll , this PR does not work around the fact that `instance-name` must be set in the configuration in order to match an existing `HazelcastInstance`. This limitation stems from the fact that it is possible to create as many `HazelcastInstance`s as you want from a given configuration file that does not specify an instance name -> how should we pick one `HazelcastInstance` of many?

snicoll:
I think that makes sense and there is this getOrCreate that does exactly what I want. But it does not for the automatic detection of the configuration file. I've created #10007 to track that. I think this would fix most of the issues we have in Boot.

hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
```diff
@@ -56,83 +61,100 @@ public static HazelcastServerCachingProvider createCachingProvider(HazelcastInst
...
-            } catch (Exception e) {
-                throw ExceptionUtil.rethrow(e);
-            }
+            Config config = getConfigFromLocation(location, classLoader, instanceName);
+            return HazelcastInstanceFactory.getOrCreateHazelcastInstance(config);
         }
 
         // If instance name is specified via properties, get instance through it.
         if (instanceName != null) {
             return Hazelcast.getHazelcastInstanceByName(instanceName);
```
jerrinot: out of the scope for this PR, but I believe this should use `getOrCreate()`

right now there is an inconsistency: when you explicitly pass a configuration then it will always use it and it will use `getOrCreate()`

I believe when you pass just the `instanceName`, but no explicit configuration then it should use default configuration.

hazelcast/src/main/java/com/hazelcast/cache/impl/HazelcastServerCachingProvider.java
```diff
@@ -56,83 +61,100 @@ public static HazelcastServerCachingProvider createCachingProvider(HazelcastInst
...
-        return getInstanceThroughDefaultInstanceIfItIsDefault(isDefaultURI);
+        // resolving HazelcastInstance via properties failed, try with URI as XML configuration file location
+        final boolean isDefaultURI = (uri == null || uri.equals(getDefaultURI()));
+        if (!isDefaultURI) {
+            try {
+                // try locating a Hazelcast config at CacheManager URI
+                Config config = getConfigFromLocation(uri, classLoader, null);
+                return HazelcastInstanceFactory.getOrCreateHazelcastInstance(config);
+            } catch (Exception e) {
+                ignore(e);
```
jerrinot: this makes troubleshooting hard. `LOGGER.finest()` is a better option.

vbekiaris:
Test failure `com.hazelcast.client.map.querycache.ClientQueryCacheBasicTest.entryRemoved_whenValueMatchesPredicate[includeValues: false, useQueryCacheFilteringStrategy: false, nearCache: false]`

vbekiaris:
run-lab-run

PR #9976, opened 2017-02-24
[WIP] Map.get with projection offloading

tombujok:


hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
```diff
@@ -0,0 +1,88 @@
...
+    public GetWithProjectionOperation(String name, Data dataKey, Projection projection) {
+        super(name, dataKey);
+
+        this.dataKey = dataKey;
+        this.projection = projection;
+    }
+
+    @Override
+    public void run() {
+        Object value = recordStore.get(dataKey, false);
```
pveentjer: You might run into a data race here if the object is stored in OBJECT format. If you don't clone, the projection thread(s)! and partition thread (modifying EP) could be accessing the same object instance,

hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
```diff
@@ -0,0 +1,88 @@
...
+        runOffloadedProjection(value, (OperationServiceImpl) getNodeEngine().getOperationService());
+    }
+
+    private void runOffloadedProjection(final Object value, final OperationServiceImpl ops) {
+        ops.onStartAsyncOperation(this);
+        getNodeEngine().getExecutionService().execute(ASYNC_EXECUTOR, new Runnable() {
```
pveentjer: For the time being it is ok, but in the future having control on the executor is the better choice.

hazelcast/src/main/java/com/hazelcast/core/IMap.java
```diff
@@ -80,6 +80,18 @@
...
+     * Gets the object stored under the given key, applies the given projection to it and returns output of the projection.
+     * The object fetching will happen on the partition-thread, the projection will be applied on a generic thread.
+     * Return a clone of the value.
+     *
+     * @param k   key to fetch the value from
+     * @param p   projection to apply
+     * @param <O> type of the output object
+     * @return result of the projection that is applied to the object fetch from the given k.
+     */
+    <O> O get(Object k, Projection<V, O> p);
```
pveentjer: Should k be of type object or K?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
tombujok: Thx. fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
tombujok: Added GroupProperties

hazelcast/src/main/java/com/hazelcast/core/IMap.java
tombujok: Map's contract somehow defines it as Object even though there's K

tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
pveentjer: Eventually I would coordinate this on a call level. So each call should be able to configure the executor it runs on. This way the caller has fine grained control on what gets run in parallel and what gets queued.If you have a single global property; the caller has not a lot of control.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public void run() {
+        Object value = recordStore.get(dataKey, false);
+
+        boolean offload = getNodeEngine().getProperties().getBoolean(GET_PROJECTION_OFFLOAD);
+        SerializationService ss = getNodeEngine().getSerializationService();
+        if (offload) {
+            value = isObjectFormat() ? ss.toObject(ss.toData(value)) : ss.toObject(value);
```
pveentjer: I would move the toObject into the offloading. So you convert object -> data on the partition thread, but the offloading thread can take care of deserializing data to object. This removes more pressure for the partition thread and will allow for more parallel operations on the same partition.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
tombujok: Sure, the GroupProperty was added for the sake of PoC for now

hazelcast/src/main/java/com/hazelcast/map/impl/operation/GetWithProjectionOperation.java
tombujok: Fixed

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -286,14 +288,25 @@ public V get(Object key) {
...
+        Data projectionData = toData(projection);
+        return toObject(getInternalWithProjection(keyData, projectionData));
     }
 
-    protected Object getInternal(Data keyData) {
-        ClientMessage request = MapGetCodec.encodeRequest(name, keyData, getThreadId());
+    protected Object getInternalWithProjection(Data keyData, Data projection) {
+        ClientMessage request = MapGetWithProjectionCodec.encodeRequest(name, keyData, projection, getThreadId());
         ClientMessage response = invoke(request, keyData);
         MapGetCodec.ResponseParameters resultParameters = MapGetCodec.decodeResponse(response);
```
sancar: Need to replace with
`MapGetWithProjectionCodec.ResponseParameters resultParameters = MapGetWithProjectionCodec.decodeResponse(response);`

tombujok: Thx @sancar Fixed!

PR #9977, opened 2017-02-24
Introduces convenience methods for Version comparison

vbekiaris:
Also adds serialization tests for Versioned DataSerializable


hazelcast/src/test/java/com/hazelcast/version/VersionTest.java
```diff
@@ -40,6 +56,14 @@ public void isGreaterThan() throws Exception {
     }
 
     @Test
+    public void isUnknownOrGreaterThan() throws Exception {
```
pveentjer: What about patch releases?

vbekiaris: `Version` models cluster version which is `major.minor`, no patch level information is maintained.

pveentjer:
:+1:

PR #9993, opened 2017-02-28
[BACKPORT][DO NOT MERGE] Map.get with projection WIP

tombujok:


hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
```diff
@@ -82,11 +90,95 @@ public void innerBeforeRun() throws Exception {
...
+        final Data finalDataKey = dataKey;
+        final String finalCaller = getCallerUuid();
+        final long finalThreadId = threadId;
+        final long finalCallId = getCallId();
+
+        ops.onStartAsyncOperation(this);
+        executorName = executorName.equals("") ? ASYNC_EXECUTOR : executorName;
+
+        // lock key - has to do it still on the partition-thread
+        recordStore.lock(finalDataKey, finalCaller, finalThreadId, finalCallId, Long.MAX_VALUE);
```
pveentjer: This is because the call is pessimistic?

pveentjer: Are we sure that even if the record is locked, a get won't be obstructed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
```diff
@@ -82,11 +90,95 @@ public void innerBeforeRun() throws Exception {
...
+                    final Map.Entry entry = createMapEntry(dataKey, value);
+                    final Data result = process(entry);
+                    if (!noOp(entry)) {
+
+                        // only case when we need to synchronize back with the
+                        OperationService ops = getNodeEngine().getOperationService();
+                        Data newValue = toData(entry.getValue());
+                        EntrySetOperation updateOperation = new EntrySetOperation(name, finalDataKey, newValue, -1);
+                        updateOperation.setPartitionId(getPartitionId());
+                        ops.invokeOnPartition(updateOperation).andThen(new ExecutionCallback<Object>() {
```
pveentjer: Can this be pulled into a non anonymous inner class?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
```diff
@@ -82,11 +90,95 @@ public void innerBeforeRun() throws Exception {
...
+                            @Override
+                            public void onResponse(Object response) {
+                                try {
+                                    recordStore.unlock(finalDataKey, finalCaller, finalThreadId, finalCallId);
```
pveentjer: Afaik this onResponse is called from an arbitrary thread. Does the recordStore like this? Should this not be done from the appropriate operation thread (and take into consideration that the partition could have been migrated).

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
```diff
@@ -82,11 +90,95 @@ public void innerBeforeRun() throws Exception {
...
+            public void run() {
+                try {
+                    final Map.Entry entry = createMapEntry(dataKey, value);
+                    final Data result = process(entry);
+                    if (!noOp(entry)) {
+
+                        // only case when we need to synchronize back with the
+                        OperationService ops = getNodeEngine().getOperationService();
+                        Data newValue = toData(entry.getValue());
+                        EntrySetOperation updateOperation = new EntrySetOperation(name, finalDataKey, newValue, -1);
```
pveentjer: Would be nice if we could have a setAndUnlock operation so you can write and unlock the record in 1 go.

PR #9995, opened 2017-03-01
Fix re-execution of completed task when killing the owner node

### LINKED ISSUE #9965: DurableExecutorService re-executes completed tasks in case of node failure.
Hello,
I tried the following use case. I have some questions regarding behaviour of re-execution of tasks in  case of node failure.

[https://github.com/shalakhansidmul/hazelcast_test](https://github.com/shalakhansidmul/hazelcast_test)
In the provided example, there are 4 tasks:
SumTask, PutSumInAMapTask, SquareTask, PrintTheSquareTask.
All are HazelcastInstanceAware.
Order of execution:

Node1 will submit Sumtask to the executor.
Just before returning, Sumtask will submit PutSumInAMapTask to the executor.
Just before returning, PutSumInAMapTask will submit SquareTask to the executor.
Just before returning, SquareTask will submit PrintThe SquareTask to the executor.
There are 4 nodes in my Hazelcast cluster.
Steps:
Start the MasterMember(Node1) and three SlaveMembers(Node2,Node3,Node4) .
Observation:
SumTask submitted from node1. It executes on Node2
It submits PutSumInAMapTask and returns.
Randomly, PutSumInAMapTask also executes on Node2.
While the task is still running, I kill Node2.
The Hazelcast cluster repartitions the data and re-executes SumTask and PutSumInAMapTask on Node3.
Why does it execute SumTask even when it is completed?
It should ideally just resume/restart PutSumInAMapTask , right?

tkountis:
When a task is submitted to a key owner, and the task completes, we replace the runnable
with the result in the internal ring buffer. However, there is no backup-operation to sync
the result with the backup nodes. If the node shuts down gracefully then the replication
will properly handle it, but when the node is killed, the backup nodes still have the task
in their ring-buffer rather than the result, thus, we re-execute it.

Fix #9965


tkountis:
run-lab-run

hazelcast/src/main/java/com/hazelcast/durableexecutor/impl/operations/PutResultOperation.java
```diff
@@ -58,6 +65,17 @@ public WaitNotifyKey getNotifiedKey() {
...
 
     @Override
+    public boolean shouldBackup() {
+        Version version = getNodeEngine().getClusterService().getClusterVersion();
+        return version.isGreaterOrEqual(V3_9);
+    }
+
+    @Override
+    public Operation getBackupOperation() {
+        return new PutResultBackupOperation(name, sequence, result);
```
vbekiaris: In relation to #10010 , shouldn't we send either a `PutResult` or `PutResultBackup` operation depending on cluster version?

hazelcast/src/main/java/com/hazelcast/durableexecutor/impl/operations/PutResultOperation.java
```diff
@@ -58,6 +65,17 @@ public WaitNotifyKey getNotifiedKey() {
     }
 
     @Override
+    public boolean shouldBackup() {
+        Version version = getNodeEngine().getClusterService().getClusterVersion();
+        return version.isGreaterOrEqual(V3_9);
```
vbekiaris: If #10010 is merged, `3.8` should be also able to handle a backup (see comment below). wdyt?

PR #10002, opened 2017-03-02
Object input/output should always return current member version

vbekiaris:


hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/DataSerializableSerializer.java
```diff
@@ -53,6 +54,7 @@
     public static final byte EE_FLAG = 1 << 1;
 
     private static final String FACTORY_ID = "com.hazelcast.DataSerializerHook";
+    private static final Version VERSION = Version.of(BUILD_INFO.getVersion());
```
mdogan: Should we use `BuildInfoProvider.getBuildInfo()` instead? Since version can be overridden via system property.

vbekiaris: `BUILD_INFO` is initialized with `BuildInfoProvider.getBuildInfo()` which takes into account the overriding property. We can switch it to a non-final field & use `getBuildInfo()` when/if we need to re-evaluate properties for testing.

PR #10004, opened 2017-03-02
WIP DONT MERGE Replacing ClientExecutionService with ExecutionService 

sancar:
Removed ClientExecutionService
Executor creations in client side used to be scattered among
services. Each was dealing with its own create ,shutdown.
All replaced with ExecutionService.registerExecutor

This is an alternative solution to https://github.com/hazelcast/hazelcast/pull/10003


sancar:
I have decided to park these changes because ExecutionServiceImpl needs work to be able to used in client side. 
Main problem is that there is a possibility of getting exception from executor when scheduling and this exception is not propogated. This causes some client operations to hang when node is shutting down.

PR #10016, opened 2017-03-03
Small cleanup of Transactional related proxy classes

Donnerbart:
* The `TransactionalQueueProxySupport` already implements methods from the `TransactionalQueue` interface, so it should implement it, so we can have proper `@Override` annotations in that class.
* The `TransactionalMultiMapProxySupport` already implements methods from the `TransactionalMap` interface, so it should implement it, so we can have proper `@Override` annotations in that class.
* Made internal methods package private where possible
* Added some static imports
* Pulled out fields for `OperationService` and `IPartitionService` where they are used frequently in a class

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/multimap/impl/txn/TransactionalMultiMapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/multimap/impl/txn/TransactionalMultiMapProxySupport.java
@@ -17,6 +17,7 @@
 package com.hazelcast.multimap.impl.txn;
 
 import com.hazelcast.config.MultiMapConfig;
+import com.hazelcast.core.TransactionalMultiMap;
 import com.hazelcast.multimap.impl.MultiMapRecord;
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/multimap/impl/txn/TransactionalMultiMapProxySupport.java
```diff
@@ -138,12 +180,12 @@ protected boolean removeInternal(Data key, Data value) {
         if (recordId != -1) {
             TxnRemoveOperation operation = new TxnRemoveOperation(name, key, recordId, value);
             logRecord.addOperation(operation);
-            return recordId != -1;
+            return true;
```
Donnerbart: This is always `true` (see the same check in line 180).

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

PR #10034, opened 2017-03-07
Fix racy scheduled executor, syncState when task is cancelled.

### LINKED ISSUE #9984: ClientScheduledExecutorServiceBasicTest.cancelledAndDone_durable
```
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.scheduledexecutor.ScheduledExecutorServiceBasicTest.cancelledAndDone_durable(ScheduledExecutorServiceBasicTest.java:529)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
	at java.lang.reflect.Method.invoke(Method.java:597)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:89)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:81)
	at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:303)
	at java.util.concurrent.FutureTask.run(FutureTask.java:138)
	at java.lang.Thread.run(Thread.java:662)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast%20Maintenance/job/Hazelcast-3.maintenance-OracleJDK1.6-freestyle/10/testReport/junit/com.hazelcast.client.scheduledexecutor/ClientScheduledExecutorServiceBasicTest/cancelledAndDone_durable/

tkountis:
Cancelling an in-flight task (not handling interrupts properly), results in a cancel task state, which is racy with the syncState when the task finally completes execution. This is *only* documenting itself when replication happens, where the result of the task is used as decision mechanism to re-schedule it or not on the new partition owner. 

The patch is gating the result assignment to check whether it was set before. 
If so, the new result is ignored.

Fix #9984

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
+++ b/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java
@@ -201,7 +201,17 @@ public class ScheduledExecutorContainer {
 
         descriptor.setState(newState);
         descriptor.setStats(stats);
-        descriptor.setTaskResult(resolution);
+
+        if (descriptor.getTaskResult() != null) {
+            // Task result previously populated - ie. through cancel
+            // Ignore all subsequent results
+            if (logger.isFineEnabled()) {
+                logger.fine(String.format("[Scheduler: " + name + "][Partition: " + partitionId + "] syncState result skipped. "
+                                        + "Current: %s New: %s", descriptor.getTaskResult(), resolution));
+            }
+        } else {
+            descriptor.setTaskResult(resolution);
+        }
     }
 
     public boolean shouldParkGetResult(String taskName)
--- a/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/SyncStateOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/operations/SyncStateOperation.java
@@ -41,7 +41,7 @@ public class SyncStateOperation
 
     protected ScheduledTaskResult result;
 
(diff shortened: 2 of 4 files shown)


tkountis:
run-lab-run

tkountis:
Thanks Matko for going the extra mile checking on this one. I was also curious how this would behave for single-run tasks and how j.u.c does it. So i tried the Callable with j.u.c and even though the task finishes later, the cancelation takes precedent and on get() i always get the CancellationException, even though the task was completed. Thats mainly the reason I didn't make any more optimisations on this one to keep it aligned with j.u.c. 
As for the optimisation, maybe you can log this as a separate issue, and we can do this in 3.9 I don't want to back-port than what we absolutely need to.

mmedenjak:
That is a curious case and I agree it is a good choice to keep it aligned with j.u.c. It's just that results from the Callable and a cancellation can get reordered on backup replicas and they could end up in different states.

PR #10044, opened 2017-03-07
Offloading EP to non-partition threads for reads & writes

tombujok:
Updated on 17.03.2017

### Overview

Implementation of the the Offloadable contract for the EntryProcessor execution on a single key.

Allows off-loading the processing unit implementing this interface to the specified or default Executor. Currently supported in:

  * IMap.executeOnKey(Object, EntryProcessor)
  * IMap.submitToKey(Object, EntryProcessoor)
  * IMap.submitToKey(Object, EntryProcessor, ExecutionCallback)

### Offloadable (for reading & writing)

If the EntryProcessor implements the Offloadable interface the processing will be offloaded to the given ExecutorService allowing unblocking the partition-thread. The key will be locked for the time-span of the processing in order to not generate a write-conflict.

 If the EntryProcessor implements Offloadable the invocation scenario looks as follows:
 - EntryOperation fetches the entry and locks the given key on partition-thread
 - Then the processing is offloaded to the given executor
 - When the processing finishes
    * if there is a change to the entry, a EntryOffloadableSetUnlockOperation is spawned
      which sets the new value and unlocks the given key on partition-thread
    * if there is no change to the entry, a EntryOffloadableSetUnlockOperation is spawned, which just unlocks the given key on partition thread

 There will not be a conflict on a write due to the pessimistic locking of the key.
 The threading looks as follows:

1. partition-thread (fetch & lock)
2. execution-thread (process)
3. partition-thread (set & unlock, or just unlock if no changes)

### Offloadable (for reading only)

If the EntryProcessor implements the Offloadable and ReadOnly interfaces the processing will be offloaded to the given ExecutorService allowing unblocking the partition-thread. Since the EntryProcessor is not supposed to do any changes to the Entry the key will NOT be locked for the time-span of the processing.

 If the EntryProcessor implements Offloadable and ReadOnly the invocation scenario looks as follows:
 - EntryOperation fetches the entry and DOES NOT lock the given key on partition-thread
 - Then the processing is offloaded to the given executor
 - When the processing finishes
    * if there is a change to the entry -> exception is thrown
    * if there is no change to the entry -> the result is returned to the user from the executor-thread.

In the read-only case the threading looks as follows:

1. partition-thread (fetch)
2. execution-thread (process)

### Primary partition - main actors

- EntryOperation (ordinary & offloaded cases)
- EntryOffloadableSetUnlockOperation

 ### Backup partitions

Offloading will not be applied to backup partitions. It is possible to initialize the EntryBackupProcessor with some input provided by the EntryProcessor in the EntryProcessor.getBackupProcessor() method.
The input allows providing context to the EntryBackupProcessor - for example the "delta" so that the EntryBackupProcessor does not have to calculate the "delta" but it may just apply it, so that it does not hog the partition-thread.

### Locking

The locking takes place only locally. If a partition locked by an off-loaded task gets migrated, the lock will not be migrated. In this situation the off-loaded task "relying" on the lock will fail on the unlock operation, since it will notice that there is no such a lock and therefore the processing for the key will get retried. The reason behind is that the off-loadable backup-processing does not use locking there cannot be any transfer of `off-loadable locks` from the primary replica to backup replicas and the other-way around.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
@@ -101,7 +101,9 @@ import com.hazelcast.core.IMap;
 import com.hazelcast.core.IMapEvent;
 import com.hazelcast.core.MapEvent;
 import com.hazelcast.core.Member;
+import com.hazelcast.core.ReadOnly;
 import com.hazelcast.internal.serialization.InternalSerializationService;
+import com.hazelcast.map.EntryBackupProcessor;
 import com.hazelcast.map.EntryProcessor;
 import com.hazelcast.map.MapInterceptor;
 import com.hazelcast.map.MapPartitionLostEvent;
@@ -1267,10 +1269,21 @@ public class ClientMapProxy<K, V> extends ClientProxy implements IMap<K, V> {
     }
 
     public Object executeOnKeyInternal(Data keyData, EntryProcessor entryProcessor) {
+        validateEntryProcessorForSingleKeyProcessing(entryProcessor);
         ClientMessage request = MapExecuteOnKeyCodec.encodeRequest(name, toData(entryProcessor), keyData, getThreadId());
         ClientMessage response = invoke(request, keyData);
         MapExecuteOnKeyCodec.ResponseParameters resultParameters = MapExecuteOnKeyCodec.decodeResponse(response);
         return toObject(resultParameters.response);
+    }
+
+    private static void validateEntryProcessorForSingleKeyProcessing(EntryProcessor entryProcessor) {
+        if (entryProcessor instanceof ReadOnly) {
+            EntryBackupProcessor backupProcessor = entryProcessor.getBackupProcessor();
+            if (backupProcessor != null) {
+                throw new IllegalArgumentException(
+                        "EntryProcessor.getBackupProcessor() should be null for a ReadOnly EntryProcessor");
+            }
+        }
     }
 
     @Override
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java
@@ -257,8 +257,17 @@ public final class LockServiceImpl implements LockService, ManagedService, Remot
         if (event.getMigrationEndpoint() == MigrationEndpoint.SOURCE) {
             clearLockStoresHavingLesserBackupCountThan(event.getPartitionId(), event.getNewReplicaIndex());
         } else {
-            int partitionId = event.getPartitionId();
-            scheduleEvictions(partitionId);
+            scheduleEvictions(event.getPartitionId());
+        }
+        // Local locks are local to the partition and replicaIndex where they have been acquired.
+        // That is the reason they are removed on any partition event on the destination.
+        removeLocalLocks(event.getPartitionId());
+    }
+
+    private void removeLocalLocks(int partitionId) {
+        LockStoreContainer container = containers[partitionId];
+        for (LockStoreImpl lockStore : container.getLockStores()) {
+            lockStore.removeLocalLocks();
         }
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStore.java
@@ -38,6 +38,18 @@ public interface LockStore {
     boolean lock(Data key, String caller, long threadId, long referenceId, long leaseTime);
 
     /**
+     * Lock a specific key on a local partition only
+     *
+     * @param key         the key to lock
+     * @param caller      the identifier for the caller
+     * @param threadId    the identifier for the thread on the caller
+     * @param referenceId the identifier for the invocation of the caller (e.g. operation call ID)
+     * @param leaseTime   the lease duration in milliseconds
+     * @return if the lock was successfully acquired
+     */
+    boolean localLock(Data key, String caller, long threadId, long referenceId, long leaseTime);
+
+    /**
      * Lock a specific key for use inside a transaction.
      *
      * @param key         the key to lock
--- a/hazelcast/src/main/java/com/hazelcast/core/IMap.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/IMap.java
@@ -1549,6 +1549,38 @@ public interface IMap<K, V> extends ConcurrentMap<K, V>, LegacyAsyncMap<K, V> {
      * Returns the the object which is result of the process() method of EntryProcessor.
      * <p/>
      *
+     * The EntryProcessor may implement the Offloadable and ReadOnly interfaces.
+     *
+     * If the EntryProcessor implements the Offloadable interface the processing will be offloaded to the given
+     * ExecutorService allowing unblocking the partition-thread, which means that other partition-operations
+     * may proceed. The key will be locked for the time-span of the processing in order to not generate a write-conflict.
+     * In this case the threading looks as follows:
+     *  1.) partition-thread (fetch & lock)
+     *  2.) execution-thread (process)
+     *  3.) partition-thread (set & unlock, or just unlock if no changes)
+     *
+     * If the EntryProcessor implements the Offloadable and ReadOnly interfaces the processing will be offloaded to the
+     * given ExecutorService allowing unblocking the partition-thread. Since the EntryProcessor is not supposed to do
+     * any changes to the Entry the key will NOT be locked for the time-span of the processing. In this case the threading
+     * looks as follows:
+     *  1.) partition-thread (fetch & lock)
+     *  2.) execution-thread (process)
+     * In this case the EntryProcessor.getBackupProcessor() has to return null; otherwise an IllegalArgumentException
+     * exception is thrown.
+     *
+     * If the EntryProcessor implements only ReadOnly without implementing Offloadable the processing unit will not
+     * be offloaded, however, the EntryProcessor will not wait for the lock to be acquired, since the EP will not
+     * do any modifications.
+     *
+     * Using offloading is useful if the EntryProcessor encompasses heavy logic that may stall the partition-thread.
+     *
+     * Offloading will not be applied to backup partitions. It is possible to initialize the EntryBackupProcessor
+     * with some input provided by the EntryProcessor in the EntryProcessor.getBackupProcessor() method.
+     * The input allows providing context to the EntryBackupProcessor - for example the "delta"
+     * so that the EntryBackupProcessor does not have to calculate the "delta" but it may just apply it.
+     *
+     * @see Offloadable
+     * @see ReadOnly
      * @return result of entry process.
      * @throws NullPointerException if the specified key is null
      */
@@ -1569,6 +1601,38 @@ public interface IMap<K, V> extends ConcurrentMap<K, V>, LegacyAsyncMap<K, V> {
      * Applies the user defined EntryProcessor to the entry mapped by the key with
      * specified ExecutionCallback to listen event status and returns immediately.
      *
+     * The EntryProcessor may implement the Offloadable and ReadOnly interfaces.
+     *
+     * If the EntryProcessor implements the Offloadable interface the processing will be offloaded to the given
+     * ExecutorService allowing unblocking the partition-thread, which means that other partition-operations
+     * may proceed. The key will be locked for the time-span of the processing in order to not generate a write-conflict.
+     * In this case the threading looks as follows:
+     *  1.) partition-thread (fetch & lock)
+     *  2.) execution-thread (process)
+     *  3.) partition-thread (set & unlock, or just unlock if no changes)
+     *
+     * If the EntryProcessor implements the Offloadable and ReadOnly interfaces the processing will be offloaded to the
+     * given ExecutorService allowing unblocking the partition-thread. Since the EntryProcessor is not supposed to do
+     * any changes to the Entry the key will NOT be locked for the time-span of the processing. In this case the threading
+     * looks as follows:
+     *  1.) partition-thread (fetch & lock)
+     *  2.) execution-thread (process)
+     * In this case the EntryProcessor.getBackupProcessor() has to return null; otherwise an IllegalArgumentException
+     * exception is thrown.
+     *
+     * If the EntryProcessor implements only ReadOnly without implementing Offloadable the processing unit will not
+     * be offloaded, however, the EntryProcessor will not wait for the lock to be acquired, since the EP will not
+     * do any modifications.
+     *
+     * Using offloading is useful if the EntryProcessor encompasses heavy logic that may stall the partition-thread.
+     *
+     * Offloading will not be applied to backup partitions. It is possible to initialize the EntryBackupProcessor
+     * with some input provided by the EntryProcessor in the EntryProcessor.getBackupProcessor() method.
+     * The input allows providing context to the EntryBackupProcessor - for example the "delta"
+     * so that the EntryBackupProcessor does not have to calculate the "delta" but it may just apply it.
+     *
+     * @see Offloadable
+     * @see ReadOnly
      * @param key            key to be processed.
      * @param entryProcessor processor to process the key.
      * @param callback       to listen whether operation is finished or not.
@@ -1582,6 +1646,38 @@ public interface IMap<K, V> extends ConcurrentMap<K, V>, LegacyAsyncMap<K, V> {
      * EntryProcessor is not cancellable, so calling ICompletableFuture.cancel() method
      * won't cancel the operation of EntryProcessor.
      *
+     * The EntryProcessor may implement the Offloadable and ReadOnly interfaces.
+     *
+     * If the EntryProcessor implements the Offloadable interface the processing will be offloaded to the given
+     * ExecutorService allowing unblocking the partition-thread, which means that other partition-operations
+     * may proceed. The key will be locked for the time-span of the processing in order to not generate a write-conflict.
+     * In this case the threading looks as follows:
+     *  1.) partition-thread (fetch & lock)
+     *  2.) execution-thread (process)
+     *  3.) partition-thread (set & unlock, or just unlock if no changes)
+     *
+     * If the EntryProcessor implements the Offloadable and ReadOnly interfaces the processing will be offloaded to the
+     * given ExecutorService allowing unblocking the partition-thread. Since the EntryProcessor is not supposed to do
+     * any changes to the Entry the key will NOT be locked for the time-span of the processing. In this case the threading
+     * looks as follows:
+     *  1.) partition-thread (fetch & lock)
+     *  2.) execution-thread (process)
+     * In this case the EntryProcessor.getBackupProcessor() has to return null; otherwise an IllegalArgumentException
+     * exception is thrown.
+     *
+     * If the EntryProcessor implements only ReadOnly without implementing Offloadable the processing unit will not
+     * be offloaded, however, the EntryProcessor will not wait for the lock to be acquired, since the EP will not
+     * do any modifications.
+     *
+     * Using offloading is useful if the EntryProcessor encompasses heavy logic that may stall the partition-thread.
+     *
+     * Offloading will not be applied to backup partitions. It is possible to initialize the EntryBackupProcessor
+     * with some input provided by the EntryProcessor in the EntryProcessor.getBackupProcessor() method.
+     * The input allows providing context to the EntryBackupProcessor - for example the "delta"
+     * so that the EntryBackupProcessor does not have to calculate the "delta" but it may just apply it.
+     *
+     * @see Offloadable
+     * @see ReadOnly
      * @param key            key to be processed
      * @param entryProcessor processor to process the key
      * @return ICompletableFuture from which the result of the operation can be retrieved.
--- a/hazelcast/src/main/java/com/hazelcast/core/Offloadable.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/Offloadable.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.core;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.config.ExecutorConfig;
+import com.hazelcast.map.EntryProcessor;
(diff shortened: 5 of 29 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockBackupOperation.java
```diff
@@ -0,0 +1,175 @@
...
+import static com.hazelcast.map.impl.EntryViews.createSimpleEntryView;
+import static com.hazelcast.map.impl.MapService.SERVICE_NAME;
+
+/**
+ * Set & Unlock processing for the EntryOffloadableBackupOperation
+ *
+ * See the javadoc on {@link EntryOffloadableOperation}
+ */
+public class EntryOffloadableSetUnlockBackupOperation extends UnlockBackupOperation {
+
```
tombujok: TODO: This operation needs to be blocking.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1267,12 +1270,30 @@ public Object executeOnKey(K key, EntryProcessor entryProcessor) {
...
         ClientMessage request = MapExecuteOnKeyCodec.encodeRequest(name, toData(entryProcessor), keyData, getThreadId());
         ClientMessage response = invoke(request, keyData);
         MapExecuteOnKeyCodec.ResponseParameters resultParameters = MapExecuteOnKeyCodec.decodeResponse(response);
         return toObject(resultParameters.response);
     }
 
+    private static void validateEntryProcessorForSingleKeyProcessing(EntryProcessor entryProcessor) {
+        if (entryProcessor instanceof Offloadable) {
+            EntryBackupProcessor backupProcessor = entryProcessor.getBackupProcessor();
+            if (!(backupProcessor instanceof Offloadable)) {
```
mdogan: Isn't `if (!(backupProcessor instanceof Offloadable))` conflicting with the `ReadOnly` check below? When EP is `ReadOnly`, you expect  `getBackupProcessor()` to return `null` but  `(null instanceof Offloadable)` returns `false`. 

It looks like you should swap the places of `if (entryProcessor instanceof Offloadable) {...` and `if (entryProcessor instanceof ReadOnly) {...` blocks.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,315 @@
...
+
+        if (!readOnly) {
+            recordStore.lock(finalDataKey, finalCaller, finalThreadId, finalCallId, Long.MAX_VALUE);
+        }
+
+        // execute for key
+        getNodeEngine().getExecutionService().execute(executorName, new Runnable() {
+            @Override
+            public void run() {
+                if (readOnly) {
```
mdogan: Better to create two different `Runnable`s for readonly and write cases... They look like completely separate.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationBackupHandler.java
```diff
@@ -253,7 +254,9 @@ private Operation getBackupOperation(BackupAwareOperation backupAwareOp) {
         // set service name of backup operation.
         // if getServiceName() method is overridden to return the same name
         // then this will have no effect.
-        backupOp.setServiceName(op.getServiceName());
+        if (!(op instanceof EntryOffloadableSetUnlockOperation)) {
```
tombujok: TODO: this is still a hack.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,315 @@
...
+        final long finalThreadId = threadId;
+        final long finalCallId = getCallId();
+
+        ops.onStartAsyncOperation(this);
+        executorName = executorName.equals(Offloadable.DEFAULT_ASYNC_EXECUTOR) ? ASYNC_EXECUTOR : executorName;
+
+        final ObjectNamespace namespace = new DefaultObjectNamespace(MapService.SERVICE_NAME, name);
+
+        if (!readOnly) {
+            recordStore.lock(finalDataKey, finalCaller, finalThreadId, finalCallId, Long.MAX_VALUE);
```
mdogan: Partition can get migrated during offloaded execution and unlock invocation after the processing can fail and you may not learn the result of unlock.
So, I think it's better to acquire the lock with a lease.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,315 @@
...
+        final MapServiceContext mapServiceContext = mapService.getMapServiceContext();
+        return mapServiceContext.toData(obj);
+    }
+
+    /**
+     * noOp in two cases:
+     * - setValue not called on entry
+     * - or entry does not exist and no add operation is done.
+     */
+    private static boolean noOp(Map.Entry entry, Object oldValue) {
```
mdogan: This `noOp` logic is present in many classes. Can it be reused?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,315 @@
...
+        operation.setPartitionId(getPartitionId());
+        getNodeEngine().getOperationService().invokeOnPartition(operation).andThen(new ExecutionCallback<Object>() {
+            @Override
+            public void onResponse(Object response) {
+                getOperationResponseHandler().sendResponse(EntryOffloadableOperation.this, result);
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+                // TODO What should we do in this case?
```
mdogan: My suggestion is to acquire the lock with a lease...

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableBackupOperation.java
```diff
@@ -0,0 +1,211 @@
...
+        final long finalThreadId = threadId;
+        final long finalCallId = getCallId();
+
+        ops.onStartAsyncOperation(this);
+        executorName = executorName.equals(Offloadable.DEFAULT_ASYNC_EXECUTOR) ? ASYNC_EXECUTOR : executorName;
+
+        final ObjectNamespace namespace = new DefaultObjectNamespace(MapService.SERVICE_NAME, name);
+
+        // TODO hope this will lock the backup on a Backup Operation?
+        recordStore.lock(finalDataKey, finalCaller, finalThreadId, finalCallId, Long.MAX_VALUE);
```
mdogan: Partition can get migrated during offloaded execution and unlock invocation after the processing can fail (either unlock fails or timeouts and you can't learn the result). 

Even record may be unlocked on primary but remain locked on backup. And when backup is promoted, it will remain locked forever.

So, I think it's better to acquire the lock with a lease.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockBackupOperation.java
mdogan: What do you mean by blocking? `BlockingOperation` ?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockBackupOperation.java
```diff
@@ -0,0 +1,175 @@
...
+            recordStore.putBackup(key, value);
+        } else {
+            throw new IllegalArgumentException("Unsupported event type " + modificationType);
+        }
+    }
+
+    private void unlockKey() {
+        LockStoreImpl lockStore = getLockStore();
+        response = lockStore.unlock(key, originalCallerUuid, threadId, getReferenceCallId());
+        lockStore.pollExpiredAwaitOp(key);
```
mdogan: `pollExpiredAwaitOp()` is for `ICondition`. `IMap` doesn't support conditions. So, you should just call `recordStore.unlock(...)` instead of extending `c.h.c.lock.operations.UnlockBackupOperation`. Similar to `TxnUnlockBackupOperation`.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockOperation.java
```diff
@@ -0,0 +1,262 @@
...
+            getLocalMapStats().incrementRemoves(getLatencyFrom(begin));
+        } else if (modificationType == ADDED || modificationType == UPDATED) {
+            recordStore.set(key, value, DEFAULT_TTL);
+            getLocalMapStats().incrementPuts(getLatencyFrom(begin));
+        } else {
+            throw new IllegalArgumentException("Unsupported event type " + modificationType);
+        }
+    }
+
+    private void unlockKey() {
```
mdogan: I think you should only call `recordStore.unlock(...)` similar to `TxnUnlockOperation`. No need to extend `c.h.c.lock.operations.UnlockOperation`.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -1010,6 +1014,23 @@ public Data executeOnKeyInternal(Data key, EntryProcessor entryProcessor) {
         }
     }
 
+    private static void validateEntryProcessorForSingleKeyProcessing(EntryProcessor entryProcessor) {
+        if (entryProcessor instanceof Offloadable) {
```
mdogan: I think you should swap the places of if (entryProcessor instanceof Offloadable) {... and if (entryProcessor instanceof ReadOnly) {... blocks.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationBackupHandler.java
mdogan: Is this needed because `EntryOffloadableSetUnlockOperation` extends `UnlockOperation `?

tombujok: Yes, fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockOperation.java
tombujok: Thx. My concern was the duplication of the notified, but having implemented the change it looks better. Fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockBackupOperation.java
tombujok: It was a wrong TODO. I deleted it. Fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockBackupOperation.java
tombujok: Thx. Fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
tombujok: Fixed. Along with some other convenience methods.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
tombujok: Fixed.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
tombujok: Good catch :)

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
tombujok: Fixed here too. BTW, couldn't find a good common place for this method. Now it's duplicated in the client and server proxy.

tombujok:
Seems that the pessimistic approach is almost impossible to implement due to backup handling.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockOperation.java
```diff
@@ -0,0 +1,266 @@
...
+    @Override
+    public void afterRun() throws Exception {
+        super.afterRun();
+        if (modificationType == null) {
+            return;
+        }
+
+        mapServiceContext.interceptAfterPut(name, value);
+        if (isPostProcessing(recordStore)) {
+            Record record = recordStore.getRecord(value);
```
mmedenjak: Should this be `getRecord(dataKey)`?

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockOperation.java
```diff
@@ -0,0 +1,266 @@
...
+            value = record == null ? null : getNodeEngine().getSerializationService().toData(record.getValue());
+        }
+        invalidateNearCache(dataKey);
+        publishEntryEvent();
+        publishWanReplicationEvent();
+        evict(dataKey);
+    }
+
+    protected void evict(Data excludedKey) {
```
mmedenjak: Same as super.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/ClientMapProxy.java
```diff
@@ -1267,12 +1269,23 @@ public Object executeOnKey(K key, EntryProcessor entryProcessor) {
...
         ClientMessage response = invoke(request, keyData);
         MapExecuteOnKeyCodec.ResponseParameters resultParameters = MapExecuteOnKeyCodec.decodeResponse(response);
         return toObject(resultParameters.response);
     }
 
+    private static void validateEntryProcessorForSingleKeyProcessing(EntryProcessor entryProcessor) {
+        if (entryProcessor instanceof ReadOnly) {
+            EntryBackupProcessor backupProcessor = entryProcessor.getBackupProcessor();
+            if (backupProcessor != null) {
+                throw new IllegalArgumentException(
```
tombujok: Thx, added!

hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockStore.java
```diff
@@ -38,6 +38,18 @@
...
+     * Lock a specific key on a local partition only
+     *
+     * @param key         the key to lock
+     * @param caller      the identifier for the caller
+     * @param threadId    the identifier for the thread on the caller
+     * @param referenceId the identifier for the invocation of the caller (e.g. operation call ID)
+     * @param leaseTime   the lease duration in milliseconds
+     * @return if the lock was successfully acquired
+     */
+    boolean localLock(Data key, String caller, long threadId, long referenceId, long leaseTime);
```
tombujok: 👍 :)

hazelcast/src/main/java/com/hazelcast/core/IMap.java
```diff
@@ -1549,6 +1549,29 @@ boolean tryLock(K key, long time, TimeUnit timeunit, long leaseTime, TimeUnit le
      * Returns the the object which is result of the process() method of EntryProcessor.
      * <p/>
      *
+     * The EntryProcessor may implement the Offloadable and ReadOnly interfaces.
+     *
+     * If the EntryProcessor implements the Offloadable interface the processing will be offloaded to the given
+     * ExecutorService allowing unblocking the partition-thread. The key will be locked for the time-span of the
```
tombujok: Yes, I think they do know, but I added a sentence similar to the one suggested.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,298 @@
...
+        getNodeEngine().getExecutionService().execute(executorName, new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    final Map.Entry entry = createMapEntry(dataKey, previousValue);
+                    final Data result = process(entry);
+                    if (!noOp(entry, previousValue)) {
+                        throw new HazelcastException("It is not allowed to modify an entry in a ReadOnly EntryProcessor");
+                    }
+                    getOperationResponseHandler().sendResponse(EntryOffloadableOperation.this, result);
```
tombujok: It has a default signature only. Need to use the other one.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,298 @@
...
+                } catch (Throwable t) {
+                    getOperationResponseHandler().sendResponse(EntryOffloadableOperation.this, t);
+                } finally {
+                    ops.onCompletionAsyncOperation(EntryOffloadableOperation.this);
```
tombujok: I think you meant that in the read-only path it was not started but completed? I fixed that, thx.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,298 @@
...
+                }
+            }
+        });
+    }
+
+    @SuppressWarnings({"unchecked", "checkstyle:anoninnerlength"})
+    private void runOffloadedModifyingEntryProcessor(final Object previousValue, String executorName) {
```
tombujok: In the EE project. OSS does not deal with the NATIVE at all.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableOperation.java
```diff
@@ -0,0 +1,298 @@
...
+        ops.invokeOnPartition(updateOperation).andThen(new ExecutionCallback<Object>() {
+            @Override
+            public void onResponse(Object response) {
+                getOperationResponseHandler().sendResponse(EntryOffloadableOperation.this, result);
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+                // TODO: verify that if t instanceof RetryableException -> the operation root-operation get retried
+                getOperationResponseHandler().sendResponse(EntryOffloadableOperation.this, t);
```
tombujok: I'm aware of that. It holds if the Operation *throws* this exception. I wanted to verify if it is true if the Operation *sends a response* that is a `RetryableException`, as suggested by Peter, since it's this deferred-response case.
I've just verified it and it works fine only because we are doing sendResponse manually.
If an exception was thrown from within the operation, it would not have been retried, since the returnsResponse method returns `false`. Tricky...

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockOperation.java
tombujok: Good catch!

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockOperation.java
tombujok: Removed

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOffloadableSetUnlockOperation.java
```diff
@@ -0,0 +1,266 @@
...
+        } else {
+            final Record record = recordStore.getRecord(key);
+            if (record != null) {
+                final EntryView entryView = createSimpleEntryView(key, value, record);
+                mapEventPublisher.publishWanReplicationUpdate(name, entryView);
+            }
+        }
+    }
+
+    protected boolean isPostProcessing(RecordStore recordStore) {
```
tombujok: Removed.

tombujok:
run-lab-run

tombujok:
run-lab-run

tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/concurrent/lock/LockServiceImpl.java
```diff
@@ -258,10 +258,20 @@ public void commitMigration(PartitionMigrationEvent event) {
             clearLockStoresHavingLesserBackupCountThan(event.getPartitionId(), event.getNewReplicaIndex());
         } else {
             int partitionId = event.getPartitionId();
+            // Local locks are local to the partition and replicaIndex where they have been acquired.
+            // That is the reason they are removed on any partition event on the destination.
+            removeLocalLocks(partitionId);
```
mdogan: Shouldn't `removeLocalLocks()` be called on `SOURCE`? They won't be replicated to destination, right?

tombujok: Hmm, it's just for local migrations (replicaIndex changes) -> so it a replica gets downgraded from 0 -> to 1, the locks should be removed.

if it gets upgraded 1->0 there won't be locks anyway

mdogan: If a replica is downgraded `0 -> 1` (in general `0 -> X` where `X > 0`) then that means 0th replica is migrated to somewhere else. So, downgrade from 0 to  X (where `X > 0`) only happens on source, not on destination.

This is very similar to `clearLockStoresHavingLesserBackupCountThan(...)`. Even you can call `removeLocalLocks()` in `clearLockStoresHavingLesserBackupCountThan()`.

mdogan: You should also add a test to verify local locks are not migrated to destination and removed on source after migration.

tombujok:
run-lab-run

tombujok:
run-lab-run

tombujok:
@mdogan Thx for review! I'll add it in a separate PR.

hazelcast/src/main/java/com/hazelcast/core/Offloadable.java
```diff
@@ -0,0 +1,67 @@
...
+ *     <li>{@link IMap#submitToKey(Object, EntryProcessor)} </li>
+ *     <li>{@link IMap#submitToKey(Object, EntryProcessor, ExecutionCallback)} </li>
+ * </ul>
+ */
+public interface Offloadable {
+
+    /**
+     * Constant meaning that there will be no off-loading if returned from the {@link #getExecutorName()} method.
+     */
+    String NO_OFFLOADING = "a2b7e85f-17ce-4fff-8833-4516efeb0b74";
```
jerrinot: what's rationale behind using UUID's here?

tombujok: Enes had this pretty sensible comment that it is possible to send a null or an empty string by coincidence, just forgetting to initialize that.

PR #9529, opened 2016-12-27
spring xsd xs:all fix

bilalyasar:
fixes https://github.com/hazelcast/hazelcast/issues/9289

PR #10054, opened 2017-03-09
Allow RANDOM eviction policy in XML configuration

### LINKED ISSUE #10053: XML configuration does not allow RANDOM eviction eviction-policy
`RANDOM` is a valid [eviction policy](https://github.com/jerrinot/hazelcast/blob/77ee1969666d385b1bcbc6a6294911f398b7e9fd/hazelcast/src/main/java/com/hazelcast/config/EvictionPolicy.java#L37-L37), but it appears it's not permitted [in the XSD](https://github.com/jerrinot/hazelcast/blob/77ee1969666d385b1bcbc6a6294911f398b7e9fd/hazelcast/src/main/resources/hazelcast-config-3.8.xsd#L2792-L2792)

The XSD contraint should be relaxed to allow `RANDOM` policy. Also Spring configuration should be checked.

jerrinot:
Also fixing [a regression](https://github.com/hazelcast/hazelcast/pull/9529) where only a single client near-cache could be configured in Spring XML.
Fixes #10053
This should be backported.

### DIFF
--- a/hazelcast-client/src/main/resources/hazelcast-client-config-3.9.xsd
+++ b/hazelcast-client/src/main/resources/hazelcast-client-config-3.9.xsd
@@ -385,6 +385,7 @@
             <xs:enumeration value="NONE"/>
             <xs:enumeration value="LRU"/>
             <xs:enumeration value="LFU"/>
+            <xs:enumeration value="RANDOM"/>
         </xs:restriction>
     </xs:simpleType>
 
--- a/hazelcast-spring/src/main/resources/hazelcast-spring-3.9.xsd
+++ b/hazelcast-spring/src/main/resources/hazelcast-spring-3.9.xsd
@@ -1471,7 +1471,7 @@
         <xs:complexType>
             <xs:complexContent>
                 <xs:extension base="hazelcast-bean">
-                    <xs:all>
+                    <xs:choice minOccurs="0" maxOccurs="unbounded">
                         <xs:element name="spring-aware" type="xs:string" minOccurs="0" maxOccurs="1"/>
                         <xs:element name="group" type="group" minOccurs="0" maxOccurs="1"/>
                         <xs:element name="properties" type="properties" minOccurs="0" maxOccurs="1"/>
@@ -1481,9 +1481,9 @@
                         <xs:element name="serialization" type="serialization" minOccurs="0" maxOccurs="1"/>
                         <xs:element name="proxy-factories" type="proxy-factories" minOccurs="0" maxOccurs="1"/>
                         <xs:element name="load-balancer" type="load-balancer" minOccurs="0" maxOccurs="1"/>
-                        <xs:element name="near-cache" type="near-cache-client" minOccurs="0" maxOccurs="1"/>
+                        <xs:element name="near-cache" type="near-cache-client" minOccurs="0" maxOccurs="unbounded"/>
                         <xs:element name="query-caches" type="query-caches-client" minOccurs="0" maxOccurs="1"/>
-                    </xs:all>
+                    </xs:choice>
                     <xs:attribute name="executor-pool-size" type="xs:int" use="optional"/>
                     <xs:attribute name="credentials-ref" type="xs:string" use="optional"/>
                 </xs:extension>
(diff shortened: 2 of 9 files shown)


hazelcast-spring/src/main/resources/hazelcast-spring-3.9.xsd
```diff
@@ -1471,7 +1471,7 @@
         <xs:complexType>
             <xs:complexContent>
                 <xs:extension base="hazelcast-bean">
-                    <xs:all>
+                    <xs:choice minOccurs="1" maxOccurs="unbounded">
```
jerrinot: this is THE important change.

hazelcast-client/src/main/resources/hazelcast-client-config-3.9.xsd
```diff
@@ -385,6 +385,7 @@
             <xs:enumeration value="NONE"/>
             <xs:enumeration value="LRU"/>
             <xs:enumeration value="LFU"/>
+            <xs:enumeration value="RANDOM"/>
```
jerrinot: adding this line was my original intend.

hazelcast-spring/src/main/resources/hazelcast-spring-3.9.xsd
```diff
@@ -1481,9 +1481,9 @@
                         <xs:element name="serialization" type="serialization" minOccurs="0" maxOccurs="1"/>
                         <xs:element name="proxy-factories" type="proxy-factories" minOccurs="0" maxOccurs="1"/>
                         <xs:element name="load-balancer" type="load-balancer" minOccurs="0" maxOccurs="1"/>
-                        <xs:element name="near-cache" type="near-cache-client" minOccurs="0" maxOccurs="1"/>
+                        <xs:element name="near-cache" type="near-cache-client" minOccurs="0" maxOccurs="unbounded"/>
```
jerrinot: so being nested inside `<xs:choice>` means we can have multiple `<near-cache>` elements.

hazelcast-spring/src/main/resources/hazelcast-spring-3.9.xsd
Donnerbart: I had to look up that XSD stuff, since I don't know it by heart:
* `<xs:all>` means that all elements can appear in any order, but all elements MUST occur once and only once (which prevents multiple Near Caches)
https://www.w3schools.com/xml/el_all.asp
* `<xs:choice minOccurs="1" maxOccurs="unbounded">` means that all elements can appear in any order, but all elements MUST occur once (and can be there unlimited time, which allows multiple Near Caches)
https://www.w3schools.com/xml/el_choice.asp

What I don't understand is the `minOccurs="1"`. What if I don't have a Near Cache? This sounds wrong to me.

Donnerbart: Okay, each child element defines its own `minOccurs="0" maxOccurs="1"` etc., so this seems to override the `<xs:choice minOccurs="1" maxOccurs="unbounded">`. But I still don't understand the meaning of the `minOccurs` then.

jerrinot: good one. I copied it from [here](https://github.com/jerrinot/hazelcast/blob/3bf3ffbf6661f8ca4d7bc079d95ac14091299df0/hazelcast-client/src/main/resources/hazelcast-client-config-3.9.xsd#L26-L26) Perhaps one can have a very basic configuration where everything is on default.

jerrinot: actually it works just fine. the rules are quite complicated, see this: http://stackoverflow.com/questions/5214555/understanding-xsdchoice-and-minoccurs

Donnerbart: Hmm, so basically the `<xs choice minOccurs="1"` has no effects, since a e.g. even an absent Near Cache configuration (which is allowed by its own `minOccurs="0"`) counts to the choice. So we cannot enforce that "at least one child element needs to be defined" if there are optional elements like the Near Cache.

So this works fine, but it still looks like it should do something else :) Wouldn't be `<xs choice `minOccurs="0"` be more logical then? Even now a total empty block should work fine, doesn't it?

jerrinot: changed

mesutcelik:
👍 THE change is the change to `choice` which is well discussed i believe and i see there is a test for multiple definition of client near-cache configuration.

jerrinot:
@Donnerbart, @mesutcelik, @sancar: thanks for a quick review!

Issue #9665, opened 2017-01-17
client side OOME with Async operations 3.8-SNAPSHOT


Danny-Hazelcast:
while making async operations test,  I got a client side GC Overhead limit exceeded,  OOME http://54.87.52.100/~jenkins/workspace/temp/HzClient460HZ-hprof.zip

One instance of "com.hazelcast.util.executor.LoggingScheduledExecutor" loaded by "sun.misc.Launcher$AppClassLoader @ 0xe6fae438" occupies 327,143,344 (97.65%) bytes. The memory is accumulated in one instance of "java.util.concurrent.RunnableScheduledFuture[]" loaded by "<system class loader>".`

so it looks like in hat master,  we have a client side OOME,

member side seams ok.

i can only get this result with a full on 2000 client 3 member,

Danny-Hazelcast:
async near cache test result in client size hprof.  and hs_err jvm crash and core

http://54.87.52.100/~jenkins/workspace/temp/HzClient74HZ-nearCache.hprof.zip

and http://54.87.52.100/~jenkins/workspace/temp/nearCacheAplStyle-HzClient60HZ-core-hsErr-hprof.zip

Danny-Hazelcast:
setup
https://github.com/Danny-Hazelcast/hzCmd-bench/blob/zeta/lab/hz/apl/cacheOps/go

async ops
https://github.com/Danny-Hazelcast/hzCmd-bench/blob/master/lab/hz/apl/cacheOps/ops

Danny-Hazelcast:
i see -Dhazelcast.client.max.concurrent.invocations=100 is working and that 

com.hazelcast.core.HazelcastOverloadException is thrown

and handled by the client test code before making another async call

Danny-Hazelcast:
Asım Arslan 
My initial findings show that `-Dhazelcast.client.max.concurrent.invocations=100` not working properly. I can reproduce backpressure is not working and causing OOME.

asimarslan:
After some analysis we can reproduce that client backpressure is broken. I'll link this issue to original issue and continue from there. 

Serialization, near cache and jcache statistics make problem apparent. 

#8568

bwzhang2011:
@asimarslan, any update with such issure ?

Issue #8568, opened 2016-07-21
Client fail-fast (hazelcast.client.max.concurrent.invocations) not working for async operations causing OOM


ihsandemir:
Client has a configurable hazelcast.client.max.concurrent.invocations property which limits outstanding client requests. We have observed at some cases that for async calls it is possible that this limit may not work as expected and this may cause outstanding requests to grow which may cause OOM. 

When examining the issue we observe a lot of client messages in heap. We see that both the request and response messages are in the heap. This is most probably caused by slow Callback executions. Here is a test case for generation of the issue:
https://github.com/ihsandemir/hazelcast/blob/maxInvocationFix/hazelcast-client/src/test/java/com/hazelcast/client/ClientMaxAllowedInvocationTest.java#L93

ihsandemir:
Observation: The client controls number of outstanding invocations using the correlation id. It increases it when it registers a request to be sent and the number is decreased when the response is received from the tcp channel for that request, but it is decremented before notifying the future (at ResponseThread.handleClientMessage ). Hence, this control is not including the part including and after the invocation notify. Hence, the client shall continue sending new requests because the counter is decreased and this may lead a lot of responses (more than the configured overload limit) being processed at the invocation.notify stage. This is especially true for async calls where they have andThen logic.

PR #10064, opened 2017-03-13
Include internal callbacks to backpressure

sancar:
CallIdSequence will be completed to accept next invocation only when
the callbacks running on internal executors are completed.

A second change made to achieve back pressure safely. If response
is already available when andThen is called with an internal callback.
Then internal callback runs on calling thread instead of executor.
Since it is already not permitted to do any blocking call in internal
threads, this will achieve a natural backpressure.

fixes https://github.com/hazelcast/hazelcast/issues/9665
fixes https://github.com/hazelcast/hazelcast/issues/8568


sancar:
run-lab-run

PR #10066, opened 2017-03-13
[BACKPORT] Fix spurious crash in ClientExceptionFactory

viliam-durina:
The ExceptionUtilsTest in jet crashed spuriously with

```java.lang.ArrayIndexOutOfBoundsException
    at java.lang.System.arraycopy(Native Method)
    at com.hazelcast.client.impl.protocol.ClientExceptionFactory.createExceptionMessage(ClientExceptionFactory.java:708)
    ...
```

We couldn't trace down the reason (because after adding some log messages it 
did not crash). Seems like the stacktraces or causes of the Throwables changed
under our hands, however unlikely or unbelievable that sounds. We catch
`ExecutionException` in the test, probably it was changed after being thrown.

I've just rewritten the stuff to call the `getStackTrace` and `getCause` just 
once, and the problem disappeared (at least does not fail after 1000 repeats, 
while before it failed after 40 trials or so). This implementation is 
resilient to this.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientExceptionFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientExceptionFactory.java
@@ -75,10 +75,12 @@ import java.io.UTFDataFormatException;
 import java.net.SocketException;
 import java.net.URISyntaxException;
 import java.security.AccessControlException;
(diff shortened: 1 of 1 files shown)


sancar:
@viliam-durina We are gonna have a forward port right?

viliam-durina:
@sancar yes.

Issue #9967, opened 2017-02-22
heath check url accepts garbage after `health` command


gAmUssA:
When health check is enabled, user can send garbage request like

`http://<your member's host IP>:5701/hazelcast/healthqqq` and similar which will return correct response.

Thank you

PR #10073, opened 2017-03-13
Fix health check url accepting garbase

gaganis:
by checking that no extra characters exist in the URL.

Resolves: #9967

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/internal/ascii/HTTPCommunicator.java
+++ b/hazelcast/src/test/java/com/hazelcast/internal/ascii/HTTPCommunicator.java
@@ -73,6 +73,12 @@ public class HTTPCommunicator {
         return doGet(url);
     }
 
+    public String getFailingClusterHealthWithTrailingGarbage() throws IOException {
+        String baseAddress = instance.getCluster().getLocalMember().getSocketAddress().toString();
+        String url = "http:/" + baseAddress + HttpCommandProcessor.URI_HEALTH_URL + "garbage";
+        return doGet(url);
+    }
+
     public String getClusterHealth() throws IOException {
         String baseAddress = instance.getCluster().getLocalMember().getSocketAddress().toString();
         String url = "http:/" + baseAddress + HttpCommandProcessor.URI_HEALTH_URL;
--- a/hazelcast/src/test/java/com/hazelcast/internal/ascii/RestClusterTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/internal/ascii/RestClusterTest.java
@@ -255,4 +255,11 @@ public class RestClusterTest extends HazelcastTestSupport {
         HTTPCommunicator communicator = new HTTPCommunicator(instance);
         communicator.getClusterHealth();
     }
+
+    @Test(expected = IOException.class)
+    public void fail_on_healthcheck_url_with_garbage() throws Exception {
+        HazelcastInstance instance = Hazelcast.newHazelcastInstance(config);
+        HTTPCommunicator communicator = new HTTPCommunicator(instance);
+        communicator.getFailingClusterHealthWithTrailingGarbage();
+    }
 }
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
@@ -47,7 +47,7 @@ public class HttpGetCommandProcessor extends HttpCommandProcessor<HttpGetCommand
             handleQueue(command, uri);
         } else if (uri.startsWith(URI_CLUSTER)) {
             handleCluster(command);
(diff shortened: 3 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/nio/ascii/TextReadHandler.java
```diff
@@ -210,6 +210,9 @@ private boolean isCommandTypeEnabled(TextCommand command) {
                 boolean isMancenterRequest = uri.startsWith(HttpCommandProcessor.URI_MANCENTER_CHANGE_URL);
                 boolean isClusterManagementRequest = uri.startsWith(HttpCommandProcessor.URI_CLUSTER_MANAGEMENT_BASE_URL);
                 boolean isHealthCheck = healthcheckEnabled && uri.startsWith(HttpCommandProcessor.URI_HEALTH_URL);
```
tkountis: I would recommend handling the BAD_REQUEST inside the Processor see. `HttpGetCommandProcessor` rather than in the TextReadHandler; which is a very abstract protocol input handler. I know we already have some form of validation here, but I recon its better to keep it from getting more populated. Also, I would just change **there**, the `.startsWith` with `.equals` and let the default 400 propagate.

hazelcast/src/test/java/com/hazelcast/internal/ascii/HTTPCommunicator.java
```diff
@@ -74,8 +74,13 @@ public String getClusterInfo() throws IOException {
     }
 
     public String getClusterHealth() throws IOException {
+        String suffix ="";
+        return getClusterHealthWithSuffix(suffix);
+    }
+
+    public String getClusterHealthWithSuffix(String suffix) throws IOException {
```
tkountis: Minor comment, can't help but thinking that this is a valid API, since I now have available a call which accepts suffix, but if I provide anything but empty string it will fail. So, I would suggest for testing purposes just hit directly the `doGet` rather than having this extra method.

hazelcast/src/test/java/com/hazelcast/internal/ascii/RestClusterTest.java
```diff
@@ -255,4 +255,12 @@ public void fail_with_deactivatedHealthCheck() throws Exception {
         HTTPCommunicator communicator = new HTTPCommunicator(instance);
         communicator.getClusterHealth();
     }
+
+    @Test(expected = SocketException.class)
+    public void fail_on_healthcheck_with_garbage() throws Exception {
+        HazelcastInstance instance = Hazelcast.newHazelcastInstance(config);
+        HTTPCommunicator communicator = new HTTPCommunicator(instance);
+        String result = communicator.getClusterHealthWithSuffix("qqq");
+        System.out.println("result = " + result);
```
tkountis: Does this print something after the exception happens ?

gaganis:
Thanks a lot @tkountis for your review! 

May I ask what is your push policy for pull requests? Do you prefer amended commit or additional ones?

hazelcast/src/main/java/com/hazelcast/nio/ascii/TextReadHandler.java
gaganis: >  I know we already have some form of validation here

Yes that is what led me to also put my check here. Your suggestion makes it a much cleaner change though! 

Thanks!

hazelcast/src/test/java/com/hazelcast/internal/ascii/HTTPCommunicator.java
gaganis: I think I can see your point here, and it is a very interesting one. I am not 100% sure I get the why  so my understanding is the following:
 _When we see this method's name, it is not directly apparent that this method leads to failure and furthermore we cannot see how and when it fails. Which could confuse future users._

Am I correct?

hazelcast/src/test/java/com/hazelcast/internal/ascii/RestClusterTest.java
gaganis: Dead code :(. Will remove.

hazelcast/src/test/java/com/hazelcast/internal/ascii/HTTPCommunicator.java
tkountis: Absolutely 🙂

gaganis:
I have pushed an additional commit implementing the suggested changes.

I would prefer to squash these to one commit and force push them to the PR. Please let me know if that is OK. 
Thanks!

Donnerbart:
We prefer squashing as well :)

Donnerbart:
run-lab-run

gaganis:
Just pushed squashed commit :)

tkountis:
Looks good to me ! Merge it 👍

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
Thanks for the contribution!

PR #10078, opened 2017-03-14
ThreadLocalRandomProvide does not rely on non-arg constructor of Random

### LINKED ISSUE #9985: ClientTxnMultiMapTest.testConcrruentTxnPut
```
java.lang.AssertionError: expected null, but was:<com.hazelcast.transaction.TransactionNotActiveException: Transaction is not active>
	at org.junit.Assert.failNotNull(Assert.java:755)
	at org.junit.Assert.assertNull(Assert.java:737)
	at org.junit.Assert.assertNull(Assert.java:747)
	at com.hazelcast.client.txn.ClientTxnMultiMapTest.testConcrruentTxnPut(ClientTxnMultiMapTest.java:144)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast%20Maintenance/job/Hazelcast-3.maintenance-ZuluJDK6-freestyle/10/testReport/junit/com.hazelcast.client.txn/ClientTxnMultiMapTest/testConcrruentTxnPut/

jerrinot:
some JDK implementations do not have thread-safe  seed generation -> 2 concurrent calls
could create Random instances initialized with the same seed. As a result UUIDUtil could
generate the same UUID.

fixed #9985

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandomProvider.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandomProvider.java
@@ -24,6 +24,8 @@ import java.util.Random;
  */
 public final class ThreadLocalRandomProvider {
 
+    private static final Random SEED_GENERATOR = new Random();
+
     private static final ThreadLocal<Random> THREAD_LOCAL_RANDOM = new ThreadLocal<Random>();
     private static final ThreadLocal<SecureRandom> THREAD_LOCAL_SECURE_RANDOM = new ThreadLocal<SecureRandom>();
 
@@ -38,7 +40,8 @@ public final class ThreadLocalRandomProvider {
     public static Random get() {
         Random random = THREAD_LOCAL_RANDOM.get();
         if (random == null) {
-            random = new Random();
+            long seed = SEED_GENERATOR.nextLong();
+            random = new Random(seed);
             THREAD_LOCAL_RANDOM.set(random);
         }
         return random;


hazelcast/src/main/java/com/hazelcast/internal/util/ThreadLocalRandomProvider.java
```diff
@@ -38,7 +40,8 @@ private ThreadLocalRandomProvider() {
     public static Random get() {
         Random random = THREAD_LOCAL_RANDOM.get();
         if (random == null) {
-            random = new Random();
+            long seed = SEED_GENERATOR.nextLong();
```
Donnerbart: Random is just thread-safe since Java 7, isn't it? May this cause issues when this method is called from different threads?

jerrinot: "thread-safe since Java 7" <-- really?

noctarius: Yeah, on Java 6 we'll need to fallback on CAS operations

jerrinot: in Sun Java 6 the seed generator is not thread-safe: 2 threads can create 2 instances of `Random` initialized with the same seed. 

However when you have a single `Random` instance then 2 threads calling `getLong()` will get different results. The CAS loop is inside `Random`. It has been like this since at least Java 5. See http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6362070

noctarius: Oh yeah, you're right. Missed you're using a generator now. Sorry :)

Donnerbart: Okay, so after some discussions and readings this is my actual state of knowledge:
* using the same `Random` instance from multiple threads should be fine since Java 5
* this might have an impact on performance, since there is a CAS loop inside of e.g. `Random::nextLong()`
* so for performance reasons we want to use multiple `Random` instances (here via `ThreadLocal`)
* there can still be an issue with generating multiple `Random` instances from multiple threads -> they can have same random seed
* this is the reason for the `SEED_GENERATOR`, which should be thread-safe due to the same `Random` instance being used

Donnerbart:
I think we need a backport for 3.8.1 as well, don't we?

jerrinot:
run-lab-run

jerrinot:
run-lab-run

jerrinot:
run-lab-run

jerrinot:
run-lab-run

hasancelik:
run-lab-run

hasancelik:
run-lab-run

PR #10079, opened 2017-03-15
[DO NOT MERGE] Add sslEnabled to TimedMemberState

emre-aydin:
We need this so that Management Center knows if the cluster is SSL enabled, and if so, uses an `https` prefix when accessing cluster's REST API.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/monitor/TimedMemberState.java
+++ b/hazelcast/src/main/java/com/hazelcast/monitor/TimedMemberState.java
@@ -40,8 +40,9 @@ public final class TimedMemberState implements Cloneable, JsonSerializable {
     MemberStateImpl memberState;
     Set<String> instanceNames;
     List<String> memberList;
-    Boolean master;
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/monitor/TimedMemberState.java
```diff
@@ -143,6 +154,7 @@ public void fromJson(JsonObject json) {
         final JsonObject jsonMemberState = getObject(json, "memberState");
         memberState = new MemberStateImpl();
         memberState.fromJson(jsonMemberState);
+        sslEnabled = getBoolean(json, "sslEnabled");
```
emrahkocaman: I think it's better to use  the overloaded version of `getBoolean` that takes a default value.

emre-aydin: Good suggestion, done.

PR #10081, opened 2017-03-15
Fired updated event after update via putTransient

ahmetmircik:
closes https://github.com/hazelcast/hazelcast/issues/10077

Issue #80, opened 2012-03-23
Update map without update ttl


ghost:
Currenty, when update an object in map. it's ttl is updated to.
Please give us an option to update map w/o update ttl.<p>Migrated from http://code.google.com/p/hazelcast/issues/detail?id=788</p>

PR #10082, opened 2017-03-15
Fired updated event after update via putTransient

ahmetmircik:
backport of https://github.com/hazelcast/hazelcast/pull/10081

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutTransientOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/operation/PutTransientOperation.java
@@ -30,7 +30,7 @@ public class PutTransientOperation extends BasePutOperation {
 
     @Override
     public void run() {
-        recordStore.putTransient(dataKey, dataValue, ttl);
+        dataOldValue = mapServiceContext.toData(recordStore.putTransient(dataKey, dataValue, ttl));
         putTransient = true;
     }
 
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
@@ -169,8 +169,8 @@ public class NearCachedMapProxyImpl<K, V> extends MapProxyImpl<K, V> {
 
     @Override
     protected void putTransientInternal(Data key, Data value, long ttl, TimeUnit timeunit) {
-        invalidateCache(key);
         super.putTransientInternal(key, value, ttl, timeunit);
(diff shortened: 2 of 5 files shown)


neilstevenson:
@ahmetmircik 

Ref http://stackoverflow.com/questions/43659013/why-does-hazelcast-imap-puttransient-fire-the-entryaddedlistener

If you try the code
```
    String str = "world";
    String key = "hello";
    map.putTransient(key, str, 1000, TimeUnit.MILLISECONDS);
    map.putTransient(key, str, 1000, TimeUnit.MILLISECONDS);
    map.putTransient(key, str, 1000, TimeUnit.MILLISECONDS);
    map.putTransient(key, str, 1000, TimeUnit.MILLISECONDS);

    map.put(key, str);
    map.put(key, str);
    map.put(key, str);
    map.put(key, str);
```
it doesn't work correctly.

There are 4 `putTransient` then 4 `put` for the same key.

Expected sequence should be ADDED, UPDATED, UPDATED, UPDATED, UPDATED, UPDATED, UPDATED, UPDATED.

Actual sequence is ADDED, UPDATED, UPDATED, UPDATED, UPDATED, UPDATED, UPDATED, UPDATED, EVICTED, EXPIRED.

If the last action is a `put` not a `putTransient`, the entry should have no TTL and so never generate an expiry event

ahmetmircik:
Hi @neilstevenson,

It is implemented in this way since early days of hazelcast, updates with `put` don't change life span of an entry. If you want to set new ttl, you should use other `ttl-ed put` methods.
For example, to set an eternal ttl, you should set it to 0: `map.put(key, str, 0, TimeUnit.MILLISECONDS)`.

neilstevenson:
@ahmetmircik Updates with `put(k,v)` should set the TTL to 0.
Current behaviour is inconsistent, a `put(k,v)` may or may not result in the entry being removed, depending on what has gone before. The TTL isn't available to the caller, and it shouldn't need to care.

ahmetmircik:
@neilstevenson In that case max-idle-seconds can be used instead of TTL. So a `put(k,v)` can shift expiration time.

But i can see that this is a complicated issue. For example, this is an old ticket asking for update without changing TTL https://github.com/hazelcast/hazelcast/issues/80. I believe current behaviour was based on this kind of requests.     

Also besides per-entry TTL, one can configure per-map TTL, in that case setting TTL to 0 with `put(k,v)` will break current behaviour.

I'm in favour of not changing it.

neilstevenson:
A `put(k,v)` without a timeout should result in the same behaviour whether or not the key already exists in the map or not.
Meaning, it is removed if the map has a timeout and not removed if the map has no timeout.

Issue #10083, opened 2017-03-15
BackupEntryProcessor returned from AbstractEntryProcessor does not support HazelcastInstanceAware


mmedenjak:
If the backup entry processor implements `HazelcastInstanceAware`, the instance is not injected. 
Reproducer:

```
@RunWith(HazelcastSerialClassRunner.class)
@Category({QuickTest.class, ParallelTest.class})
public class BackupEntryProcessorInstanceAware extends HazelcastTestSupport {

    public static final String MAP_NAME = "EntryProcessorTest";

    @Override
    public Config getConfig() {
        Config config = super.getConfig();
        MapConfig mapConfig = new MapConfig(MAP_NAME);
        mapConfig.setInMemoryFormat(BINARY);
        config.addMapConfig(mapConfig);
        return config;
    }


    @Test
    public void test() {
        final Config cfg = getConfig();
        final TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(2);
        final HazelcastInstance i1 = factory.newHazelcastInstance(cfg);
        final HazelcastInstance i2 = factory.newHazelcastInstance(cfg);

        try {
            final IMap<Object, Object> map = i1.getMap(MAP_NAME);
            map.put("a", 1);
            map.put("b", 2);
            map.executeOnEntries(new PartitionAwareTestEntryProcessor(MAP_NAME));
        } catch (ClassCastException e) {
            e.printStackTrace();
            fail("ClassCastException must not happen!");
        } finally {
            i1.shutdown();
        }
    }


    private static class PartitionAwareTestEntryProcessor extends AbstractEntryProcessor<Object, Object>
            implements HazelcastInstanceAware {

        private String name;
        private transient HazelcastInstance hz;

        private PartitionAwareTestEntryProcessor(String name) {
            this.name = name;
        }

        @Override
        public Object process(Map.Entry<Object, Object> entry) {
            hz.getMap(name).put(entry.getKey(), entry.getValue());
            return null;
        }

        @Override
        public void setHazelcastInstance(HazelcastInstance hazelcastInstance) {
            this.hz = hazelcastInstance;
        }
    }

}
```

PR #10087, opened 2017-03-15
Inject HazelcastInstance in backup entry processor

mmedenjak:
If the supplied entry procesor extended AbstractEntryProcessor and implemented HazelcastInstanceAware, the backup processor would never get the hazelcast instance injected and the process method could fail with NPE. The fix checks if the entry processor implements HazelcastInstanceAware and supplies a different backup processor in that case.

Fixes : https://github.com/hazelcast/hazelcast/issues/10083

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
@@ -16,6 +16,8 @@
 
 package com.hazelcast.map;
 
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.HazelcastInstanceAware;
 import com.hazelcast.nio.serialization.impl.BinaryInterface;
 
 import java.util.Map;
@@ -65,10 +67,26 @@ public abstract class AbstractEntryProcessor<K, V> implements EntryProcessor<K, 
         return entryBackupProcessor;
     }
 
-    private class EntryBackupProcessorImpl implements EntryBackupProcessor<K, V> {
+    private class EntryBackupProcessorImpl implements EntryBackupProcessor<K, V>, HazelcastInstanceAware {
+        // to fix https://github.com/hazelcast/hazelcast/issues/10083 we need to add the HazelcastInstanceAware
+        // interface on the EntryBackupProcessorImpl. Unfortunately this changes the generated serialVersionUID
+        // so sending this to a member which doesn't contain the fix would cause a deserialization exception.
+        // That is why we will set the serialVersionUID here to the same generated value for a
+        // EntryBackupProcessorImpl without the implemented interface. The value is calculated by a specification
+        // based on types and fields so it should be the same on all JVM implementations
+        static final long serialVersionUID = -5081502753526394129L;
+
         @Override
         public void processBackup(Map.Entry<K, V> entry) {
             process(entry);
+        }
+
+        @Override
+        public void setHazelcastInstance(HazelcastInstance hazelcastInstance) {
+            final AbstractEntryProcessor<K, V> outer = AbstractEntryProcessor.this;
+            if (outer instanceof HazelcastInstanceAware) {
+                ((HazelcastInstanceAware) outer).setHazelcastInstance(hazelcastInstance);
+            }
         }
     }
 }
--- a/hazelcast/src/test/java/com/hazelcast/map/BackupEntryProcessorInstanceAwareTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/BackupEntryProcessorInstanceAwareTest.java
@@ -1,0 +1,81 @@
+package com.hazelcast.map;
+
+import com.hazelcast.config.Config;
+import com.hazelcast.config.MapConfig;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.core.HazelcastInstanceAware;
+import com.hazelcast.core.IMap;
+import com.hazelcast.test.AssertTask;
+import com.hazelcast.test.HazelcastSerialClassRunner;
+import com.hazelcast.test.HazelcastTestSupport;
+import com.hazelcast.test.TestHazelcastInstanceFactory;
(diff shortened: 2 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
```diff
@@ -71,4 +75,16 @@ public void processBackup(Map.Entry<K, V> entry) {
...
         }
     }
+
+    private class HazelcastInstanceAwareEntryBackupProcessorImpl extends EntryBackupProcessorImpl
+            implements HazelcastInstanceAware {
+
+        @Override
+        public void setHazelcastInstance(HazelcastInstance hazelcastInstance) {
+            final AbstractEntryProcessor<K, V> outer = AbstractEntryProcessor.this;
+            if (outer instanceof HazelcastInstanceAware) {
```
jerrinot: is this check needed?

mmedenjak: Not really. Defensive programming I guess. I'll remove it.

jerrinot:
I wonder if you can make `EntryBackupProcessorImpl` to implement `HazelcastInstanceAware`. 
The question is this: will this break serialization?

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
tombujok: This check is obligatory, isn't it? Since the class that extends the `AbstractEntryProcessor` may not implement the `HazelcastInstanceAware`, so the cast may fail.

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
```diff
@@ -65,10 +67,21 @@ public AbstractEntryProcessor(boolean applyOnBackup) {
         return entryBackupProcessor;
     }
 
-    private class EntryBackupProcessorImpl implements EntryBackupProcessor<K, V> {
+    private class EntryBackupProcessorImpl implements EntryBackupProcessor<K, V>, HazelcastInstanceAware {
```
tombujok: This will change the `serialVersionUID` of the EntryBackupProcessorImpl, so it won't deserialize in "emulated-mode" of a rolling upgrade, on a 3.8.x node.

What we could do is to have two implementations of the EntryBackupProcessorImpl, and check the clusterVersion in the EntryOperaiton, and do sth like this:
```
@Override
    public Operation getBackupOperation() {
       // ...
        EntryBackupProcessor backupProcessor = null;
        if(entryProcessor instanceOf AbstractEntryProcessor) {
                if(clusterVersion smaller than 3.9) {      
                         backupProcessor = entryProcessor.getBackupProcessor(); 
                } else {
                         // this class will be present on 3.9+ nodes
                         backupProcessor = new HazelcastAwareBackupEntryProcessorWrapper(entryProcessor);
                }
        }
        return backupProcessor != null ? new EntryBackupOperation(name, dataKey, backupProcessor) : null;
    }
```

What protects us is that the following AbstractEntryProcess's method is final:
```
public final EntryBackupProcessor<K, V> getBackupProcessor() {
        return entryBackupProcessor;
    }
```

tombujok:
Well spotted Matko!

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
jerrinot: "This will change the serialVersionUID" <-- that's the reason this PR adds an explicit `serialVersionUID`. which happens to be the same as the generated `serialVersionUID` which does not implement the `HazelcastInstananceAware` interface.

tombujok: That's a nice hack, but to me it begs for an explicit comment that it's the same.

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
mmedenjak: The comment is because of a previous version where `AbstractEntryProcessor` implemented `HazelcastInstanceAware` so this check was not needed. Later we figured out that we can't change the type hierarchy of `AbstractEntryProcessor` nor set it's `serialVersionUID`.

Now the check is mandatory since the outer object no longer necessarily implements `HazelcastInstanceAware`.

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
mmedenjak: Added a more detailed explanation, please check.

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
mmedenjak: Actually I remember from the commit message, the first version checked if the AbstractEntryProcessor and returned a different EntryBackupProcessorImpl. So the check in the "HazelcastInstanceAwareEntryBackupProcessorImpl" didn't need to check the outer type again.

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
tombujok: Looks good!

hazelcast/src/main/java/com/hazelcast/map/AbstractEntryProcessor.java
tombujok: 👍

mmedenjak:
@jerrinot Is the PR ok now?

jerrinot:
run-lab-run

jerrinot:
run-lab-run

tombujok:
run-lab-run

mmedenjak:
run-lab-run

PR #10118, opened 2017-03-21
Fixes unncessary client context creation per proxy

### LINKED ISSUE #10117: No need to create ClientContext for every proxy
Causes unneeded resource allocations.
see https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/spi/ProxyManager.java#L348

sancar:
fixes #10117


sancar:
run-lab-run

Issue #10124, opened 2017-03-21
ServiceLoader loads resources in parent classloader regardless of classloader filtering and/or child-first policy


dfeist:
Given the implementation of  `findHighestReachableClassLoader` in `ServiceLoader` it is not possible to use a different version of Hazelcast in a app-server environment where an application either uses a child-first policy or explicitly configures that `com.hazelcast` should be loaded from the application classpath.

While shading might be a good way around this in some circumstances, it shouldn't really be required.

The resulting exception is:
```
Root Exception stack trace:
java.lang.ClassCastException: Cannot cast com.hazelcast.collection.impl.queue.QueuePortableHook to com.hazelcast.internal.serialization.PortableHook
	at java.lang.Class.cast(Class.java:3369)
	at com.hazelcast.util.ServiceLoader$NewInstanceIterator.next(ServiceLoader.java:336)
	at com.hazelcast.internal.serialization.impl.PortableHookLoader.load(PortableHookLoader.java:51)
	at com.hazelcast.internal.serialization.impl.PortableHookLoader.<init>(PortableHookLoader.java:44)
	at com.hazelcast.internal.serialization.impl.SerializationServiceV1.<init>(SerializationServiceV1.java:88)
	at com.hazelcast.internal.serialization.impl.DefaultSerializationServiceBuilder.createSerializationService(DefaultSerializationServiceBuilder.java:264)
	at com.hazelcast.internal.serialization.impl.DefaultSerializationServiceBuilder.build(DefaultSerializationServiceBuilder.java:218)
	at com.hazelcast.client.impl.DefaultClientExtension.createSerializationService(DefaultClientExtension.java:77)
	at com.hazelcast.client.impl.HazelcastClientInstanceImpl.<init>(HazelcastClientInstanceImpl.java:164)
	at com.hazelcast.client.HazelcastClient.newHazelcastClient(HazelcastClient.java:84)
```

See: 
https://github.com/hazelcast/hazelcast/issues/2395
https://github.com/hazelcast/hazelcast/issues/9353

vbekiaris:
Fixed by #10130 & #10139

PR #10130, opened 2017-03-22
[BACKPORT] Skip hooks implementing an expected interface, but loaded by a different classloader

jerrinot:
Fixing #10124
Backport of https://github.com/hazelcast/hazelcast/pull/10139


hazelcast/src/main/java/com/hazelcast/instance/BuildInfoProvider.java
```diff
@@ -81,20 +84,42 @@ static void setJetProperties(Properties properties, BuildInfo buildInfo) {
...
         } catch (Exception ignored) {
             ignore(ignored);
         } finally {
-            closeResource(properties);
+            closeResource(propertyStream);
         }
         return runtimeProperties;
     }
 
+    private static URL findLastResourceURL(String resourceName, ClassLoader classLoader) {
```
jerrinot: this is somewhat fragile. it assumes the classloader with obey the recommended order of resource iterating. 
```
    * <p> This method will first search the parent class loader for the
     * resource; if the parent is <tt>null</tt> the path of the class loader
     * built-in to the virtual machine is searched.  That failing, this method
     * will invoke {@link #findResource(String)} to find the resource.  </p>
     *
     * @param  name
     *         The resource name
     *
     * @return  A <tt>URL</tt> object for reading the resource, or
     *          <tt>null</tt> if the resource could not be found or the invoker
     *          doesn't have adequate  privileges to get the resource.
     *
     * @since  1.1
     */
    public URL getResource(String name) {
```

hazelcast/src/main/java/com/hazelcast/util/ServiceLoader.java
```diff
@@ -338,30 +339,75 @@ public void remove() {
...
         @Override
         public boolean hasNext() {
-            return iterator.hasNext();
+            if (nextClass != null) {
+                return true;
+            }
+            return advance();
+        }
+
+        private boolean advance() {
```
jerrinot: this is where the magic happens. testes are yet to bet written.

hazelcast/src/main/java/com/hazelcast/instance/BuildInfoProvider.java
dfeist: While this method uses this order, it's only a recommendation that implementations overriding this do the same.  Safest approach IMO is to use `findResource` via reflection.

```
@apiNote When overriding this method it is recommended that an
     * implementation ensures that any delegation is consistent with the {@link
     * #getResources(java.lang.String) getResources(String)} method.
```

jerrinot: replaces by using a generated class instead of a plain property file.

hazelcast/src/main/java/com/hazelcast/nio/ClassLoaderUtil.java
```diff
@@ -192,6 +192,27 @@ public static boolean isClassDefined(String className) {
...
+     * classloader and hence the class is not assignable into it.
+     *
+     * This is useful for logging purposes.
+     *
+     * @param clazz
+     * @param iface
+     * @return
+     */
+    public static boolean implementsIntefaceWithSameName(Class<?> clazz, Class<?> iface) {
+        Class<?>[] interfaces = clazz.getInterfaces();
```
jerrinot: excellent catch and today I learned something new! thanks!

mdogan: Afaik, even visiting superclasses is not enough. You should also obtain interfaces extended by its own interfaces and superclasses' interfaces.

mdogan: something like;

```java
static Collection<Class> getAllInterfaces(Class clazz) {
    Collection<Class> interfaces = new HashSet<>();
    addInterfaces(clazz, interfaces);

    Class<?> superClass = clazz.getSuperclass();
    while (superClass != null) {
        addInterfaces(superClass, interfaces);
        superClass = superClass.getSuperclass();
    }
    return interfaces;
}

static void addInterfaces(Class clazz, Collection<Class> allInterfaces) {
    Class[] interfaces = clazz.getInterfaces();
    Collections.addAll(allInterfaces, interfaces);
    for (Class cl : interfaces) {
        addInterfaces(cl, allInterfaces);
    }
}
```

PR #10131, opened 2017-03-22
Fixes [any] operator handling in built-in aggregators for null/empty collections/arrays

### LINKED ISSUE #10126: Fast Aggregations on [any] operator on empty array does not work
For  builtin aggregations like Count, LongSum etc.
If we used them with an `attributePath` on list/array field like;

```
class Person implements Portable {
...
    int[] valueArray;
...
}
```

if we initialize valueArray with an empty array`valueArray = new int[]{};`

```
map.aggregate(Aggregators.<Map.Entry<Integer, Person>>integerSum.integerSum("valueArray[any]"));
```
throws NullPointerException instead of returning 0, 

```
map.aggregate(Aggregators.<Map.Entry<Integer, Person>>count("valueArray[any]"));
```
returns 1 instead of 0.

Other builtin aggregators may also fail, they should be tested.

tombujok:
Fixes #10126

Allows differentiating whether a null in the MultiResult is a null value because the value of the field was null or whether the null means that the targetObject was null.

For query evaluation the difference is not important - and we need to return null in both cases there. (see [any] corner cases in the reference manual)
For aggregations it makes a difference and we need this context knowledge there.

MultiResult is a structure returned if an extraction contains an `[any]` operator, since a single read of a `person.hands[any].length` expression may return more than one result.

Here's more info about `[any]`: http://docs.hazelcast.org/docs/3.8/manual/html-single/index.html#querying-in-collections-and-arrays

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/getters/GetterFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/getters/GetterFactory.java
@@ -26,6 +26,8 @@ import static com.hazelcast.query.impl.getters.AbstractMultiValueGetter.validate
 
 public final class GetterFactory {
 
+    private static final String ANY_POSTFIX = "[any]";
+
     private GetterFactory() {
     }
 
@@ -47,6 +49,9 @@ public final class GetterFactory {
                 returnType = getCollectionType(collection);
             }
             if (returnType == null) {
+                if (modifierSuffix.equals(ANY_POSTFIX)) {
+                    return NullMultiValueGetter.NULL_MULTIVALUE_GETTER;
+                }
                 return NullGetter.NULL_GETTER;
             }
         } else if (isExtractingFromArray(fieldType, modifierSuffix)) {
@@ -92,6 +97,9 @@ public final class GetterFactory {
                 returnType = getCollectionType(collection);
             }
             if (returnType == null) {
+                if (modifierSuffix.equals(ANY_POSTFIX)) {
+                    return NullMultiValueGetter.NULL_MULTIVALUE_GETTER;
+                }
                 return NullGetter.NULL_GETTER;
             }
         } else if (isExtractingFromArray(methodReturnType, modifierSuffix)) {
--- a/hazelcast/src/main/java/com/hazelcast/query/impl/getters/GetterFactory.java
+++ b/hazelcast/src/main/java/com/hazelcast/query/impl/getters/GetterFactory.java
@@ -23,6 +23,8 @@ import java.lang.reflect.Method;
 import java.util.Collection;
 
 import static com.hazelcast.query.impl.getters.AbstractMultiValueGetter.validateModifier;
+import static com.hazelcast.query.impl.getters.NullGetter.NULL_GETTER;
+import static com.hazelcast.query.impl.getters.NullMultiValueGetter.NULL_MULTIVALUE_GETTER;
 
 public final class GetterFactory {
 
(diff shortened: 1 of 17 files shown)


tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
```diff
@@ -69,6 +69,7 @@
 public class MultiResult<T> {
 
     private List<T> results;
+    private boolean targetNullOrEmpty;
```
ahmetmircik: an explanatory comment can be added to code.

ahmetmircik:
is this pr backportable?

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
tombujok: Ok, sure. I'll add it.

tombujok:
@ahmetmircik yep, it's back-portable.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
tombujok: Added

tombujok:
Do not merge for now. I still need to verify sth here.

tombujok:
Two issues before this can get merged:
- it depends on this https://github.com/hazelcast/hazelcast/pull/10143
- there is still no portable support

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
```diff
@@ -70,6 +70,8 @@
 
     private List<T> results;
 
+    private boolean nullOrEmptyTarget;
```
ahmetmircik: can we have this kind of comment back? https://github.com/hazelcast/hazelcast/pull/10131/commits/9ded149c1085d829813626046f47206b8b375522#diff-08181c4c9677f20321b81b482ff1d796R75

hazelcast/src/main/java/com/hazelcast/query/impl/getters/GetterFactory.java
```diff
@@ -47,6 +49,9 @@ public static Getter newFieldGetter(Object object, Getter parentGetter, Field fi
                 returnType = getCollectionType(collection);
             }
             if (returnType == null) {
+                if (modifierSuffix.equals(ANY_POSTFIX)) {
+                    return NullMultiValueGetter.NULL_MULTIVALUE_GETTER;
```
ahmetmircik: `NullMultiValueGetter.INSTANCE`?
 or can be imported statically, current form can make reads hard

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
tombujok: hmm weird, I didn't delete it on purpose, somehow it disappeared... Sure I will add it.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/GetterFactory.java
tombujok: Done.

hazelcast/src/main/java/com/hazelcast/query/impl/getters/MultiResult.java
tombujok: Done.

tombujok:
run-lab-run

Issue #9219, opened 2016-11-04
[docker] Docker: "This node is not requested endpoint" issues


bitsofinfo:
Since the original issue is closed, I'm re-posting this as we are re-visiting this now and still having issues: https://github.com/hazelcast/hazelcast/issues/4537#issuecomment-234265809

Perhaps hazelcast can re-open the old issue or look into addressing it as it is sort of a blocker at this point (still need to try testing w/ overlay network), but regardless:

What is described below is reproducible with this sample project: 
https://github.com/bmudda/hazelcast-docker-test

Referenced issues: 
https://github.com/hazelcast/hazelcast/issues/4537
https://github.com/hazelcast/hazelcast-docker/issues/10

a) clone, build and create a docker image for the sample project above (see the README)

b) Edit the `docs/hazelcast.xml` to reflect the "members" as they will appear as mapped to the docker host when running as containers (i.e. [dockerhostip]:[mapped-5701-port])

```
...
      <tcp-ip enabled="true">
        <member>[DOCKER_HOST_IP]:40001</member>
        <member>[DOCKER_HOST_IP]:40002</member>
      </tcp-ip>
...
```

c) In terminal one, launch node1 (mapped to 40001)

`docker run --rm=true -v /path/to/hazelcast-docker-test/docs:/config -p 40001:5701  [IMAGE_ID] java -Dhazelcast.config=/config/hazelcast.xml -jar /hzdocker/hazelcast-docker-test.jar`

d) In terminal two, launch node2 (mapped to port 40002)

`docker run --rm=true -v /path/to/hazelcast-docker-test/docs:/config -p 40002:5701  [IMAGE_ID] java -Dhazelcast.config=/config/hazelcast.xml -jar /hzdocker/hazelcast-docker-test.jar`

One both up you will see these kind of messages and each container trys to connect to the other to form the cluster

```
....
Nov 04, 2016 9:05:01 PM com.hazelcast.nio.tcp.TcpIpConnectionManager
WARNING: [172.17.0.3]:5701 [hazelcast-docker-test] [3.7.2] Wrong bind request from [172.17.0.2]:5701! This node is not requested endpoint: [192.168.0.148]:40001

...
Nov 04, 2016 9:15:44 PM com.hazelcast.nio.tcp.TcpIpConnectionManager
[...]

PR #10132, opened 2017-03-22
New group property, hazelcast.socket.allow.any.public.address

emrahkocaman:
I've added a new property `hazelcast.socket.allow.any.public.address`  to provide a solution to the infamous error message "Wrong bind request! ... This node is not requested endpoint".

This property will be useful especially in containerized environments. (Docker in specific). 
When Hazelcast starts in a container it can bind itself to a local address but other containers can connect to that container only through its public IP address which causes the aforementioned error.

One solution is to set public IP address in `NetworkConfig`  but this approach is not scalable/applicable where IP addresses changes dynamically (imagine you're creating new containers using Swarm, Kubernetes etc).

So I've introduced this new property which skips the local address control and also members send it's public/private addresses in the `BindMessage` for connection mapping.

Below are the issues for the need;

https://github.com/hazelcast/hazelcast/issues/9219
https://github.com/hazelcast/hazelcast/issues/4537

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/BindMessage.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/BindMessage.java
@@ -27,14 +27,16 @@ public class BindMessage implements IdentifiedDataSerializable {
 
     private Address localAddress;
     private Address targetAddress;
+    private Address translatedAddress;
     private boolean reply;
 
     public BindMessage() {
     }
 
-    public BindMessage(Address localAddress, Address targetAddress, boolean reply) {
+    public BindMessage(Address localAddress, Address targetAddress, Address translatedAddress, boolean reply) {
         this.localAddress = localAddress;
         this.targetAddress = targetAddress;
+        this.translatedAddress = translatedAddress;
         this.reply = reply;
     }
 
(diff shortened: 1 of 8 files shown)


bitsofinfo:
+1

peloncano:
+1

nfedyk:
+1

bmudda:
+1

mesutcelik:
@tombujok Can someone have a look at this? 
@emrahkocaman Can you please rebase?

bitsofinfo:
Lets get this in!!

jerrinot:
@emrahkocaman: I am afraid this breaks Rolling Upgrade compatibility.

kozervar:
+1

bmudda:
+1

bitsofinfo:
This is so badly needed!

emrahkocaman:
Closing this PR as this solution uses an address translation map and ignores the fact that multiple members might have the same translated address. (ie: members in different boxes).
We need to come up with a better solution.

bitsofinfo:
When will some of the issues around interfaces, binding etc actually be addressed in hazelcast for people trying to run in dynamically allocated container environments? There are numerous issues tied to this.

PR #10035, opened 2017-03-07
Separate proxy for jcache near cache

ahmetmircik:
Introduced `NearCachedClientCacheProxy` as an extension of `ClientCacheProxy`. All near cache related logic is moved to this new proxy class.

PR #10133, opened 2017-03-22
Fixed dependency to javax.cache in ClientICacheManager

Donnerbart:
* fixed an unwanted dependency on java.cache in `ClientICacheManager`
* added tests which exclude `javax.cache` via the `FilteringClassloader`

Without this fix I get the following error when starting a client with just an IMap and Near Cache configuration:
```
BuildInfo{version='3.9-SNAPSHOT', build='20170322', buildNumber=20170322, revision=ccd594b, enterprise=false, serializationVersion=1}
java.lang.NoClassDefFoundError: javax/cache/Cache
    at java.lang.ClassLoader.defineClass1(Native Method)
    at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
    at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
    at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
    at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
    at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
    at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
    at java.security.AccessController.doPrivileged(Native Method)
    at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
    at com.hazelcast.client.impl.HazelcastClientInstanceImpl.<init>(HazelcastClientInstanceImpl.java:241)
    at com.hazelcast.client.HazelcastClient$1.createHazelcastInstanceClient(HazelcastClient.java:55)
    at com.hazelcast.client.HazelcastClientManager.newHazelcastClient(HazelcastClientManager.java:74)
    at com.hazelcast.client.HazelcastClient.newHazelcastClient(HazelcastClient.java:72)
```

This reverts a tiny part of https://github.com/hazelcast/hazelcast/pull/10035

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/HazelcastClientCacheManager.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/HazelcastClientCacheManager.java
@@ -113,7 +113,7 @@ public final class HazelcastClientCacheManager extends AbstractHazelcastCacheMan
         try {
             ClientICacheManager cacheManager = client.getCacheManager();
             String nameWithPrefix = cacheConfig.getNameWithPrefix();
-            ICacheInternal cache = cacheManager.getCacheByFullName(nameWithPrefix);
+            ICacheInternal<K, V> cache = (ICacheInternal<K, V>) cacheManager.getCacheByFullName(nameWithPrefix);
             cache.setCacheManager(this);
             return cache;
         } catch (Throwable t) {
--- a/hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientICacheManager.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/impl/ClientICacheManager.java
@@ -17,7 +17,7 @@
 package com.hazelcast.client.impl;
 
 import com.hazelcast.cache.HazelcastCacheManager;
-import com.hazelcast.cache.impl.ICacheInternal;
+import com.hazelcast.cache.ICache;
 import com.hazelcast.cache.impl.ICacheService;
 import com.hazelcast.client.spi.impl.ClientServiceNotFoundException;
 import com.hazelcast.core.HazelcastException;
@@ -42,12 +42,12 @@ public class ClientICacheManager implements ICacheManager {
     }
 
     @Override
-    public <K, V> ICacheInternal<K, V> getCache(String name) {
+    public <K, V> ICache<K, V> getCache(String name) {
         checkNotNull(name, "Retrieving a cache instance with a null name is not allowed!");
         return getCacheByFullName(HazelcastCacheManager.CACHE_MANAGER_PREFIX + name);
     }
 
-    public <K, V> ICacheInternal<K, V> getCacheByFullName(String fullName) {
+    public <K, V> ICache<K, V> getCacheByFullName(String fullName) {
         checkNotNull(fullName, "Retrieving a cache instance with a null name is not allowed!");
         try {
             return instance.getDistributedObject(ICacheService.SERVICE_NAME, fullName);
--- a/hazelcast-client/src/test/java/classloading/JavaXCacheDependencyClientTest.java
+++ b/hazelcast-client/src/test/java/classloading/JavaXCacheDependencyClientTest.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package classloading;
+
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.test.HazelcastSerialClassRunner;
+import com.hazelcast.test.annotation.QuickTest;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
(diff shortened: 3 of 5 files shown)


Donnerbart:
run-lab-run

vbekiaris:
It would be great to add a test that creates a new client with a `ClassLoader` that fails loading any `javax.cache` classes, so we can ensure the dependency does not creep in again. It seems that when using the narrower `ICacheInternal` return type, `ClientICacheManager` class loading triggers eager resolution of `ICache extends Cache`, leading to the error; perhaps it's due to the additional synthetic bridge methods?
Other than that 👍

vbekiaris:
More evidence we need a test to verify `HazelcastClient` creation without JCache in classpath from [JVM specification - 5.4 Linking](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-5.html#jvms-5.4):

> For example, a Java Virtual Machine implementation may choose to resolve each symbolic reference in a class or interface individually when it is used ("lazy" or "late" resolution), or to resolve them all at once when the class is being verified ("eager" or "static" resolution). This means that the resolution process may continue, in some implementations, after a class or interface has been initialized. Whichever strategy is followed, any error detected during resolution must be thrown at a point in the program that (directly or indirectly) uses a symbolic reference to the class or interface.

My understanding is it is perfectly OK for JVMs to eagerly load all classes / interfaces referenced in a class' constant pool --> `HazelcastClient` would not initialize without `JCache` in the classpath, should a JVM apply eager resolution.

Donnerbart:
I've added tests for `HazelcastInstance` and `HazelcastClientInstance` with a default configuration, which should not need `javax.cache`. I used the `FilteringClassLoader` to exclude `javax.cache`, so we get a `ClassNotFoundException` if we add a regression like this again.

Those tests fail on the current master and passed locally on this branch.

Donnerbart:
As far as I understood this will not resolve the "a JVM can decide to eagerly load all classes in the constant pool", but at least we should see a hard dependency on `javax.cache` now. For those "hidden" dependencies we might need an `ImportControl` CheckStyle rule which forbids the usage of `javax.cache' outside of the cache packages or something like this (assuming that classes from those packages are just used when a Cache is configured).

vbekiaris:
Adding the `ImportControl` may or may not help (in the current issue no `javax.cache` classes were directly referenced). I think we are covered with the new tests to get an early warning.

ahmetmircik:
Nice finding @Donnerbart. 

As an alternative fix, can we introduce on demand creation of `ClientICacheManager`? This sort of approach can mitigate code fragility and prevent it from breaking after simple touches, as in my case.

Donnerbart:
I don't know if the import of `ClientICacheManager` can already trigger the eager loading @vbekiaris  described, even if it's not created / used. I guess it depends on what a "symbolic reference" is exactly in terms of the JLS.

If an import is safe this would surely be a better fix. I just did the revert since it was the easiest way to restore a working state.

ahmetmircik:
Afaik importing doesn't require loading. But at least, as of my testing, on demand creation is fixing the regression.

Donnerbart:
Okay, I'll see that I change the fix then.

Donnerbart:
I advanced the test to see "how far" we can get without `javax.cache` and trying to create a JCache. With the old code member and client would both start and you can even invoke `getCacheManager()`. If we just switch to a lazy initialization for that field (and keep the `ICacheInternal`) this behavior would break on the client side. We would also introduce the only non-final field in the `HazelcastClientInstanceImpl`. And since `getCacheManager()` is a public API method, it can be called by multiple threads. So we would need proper synchronization on its non-final field to prevent multiple instances. Or just make it `volatile` and rely on the fact that `ClientICacheManager`, which also might change in the future.

So I think it's really better to revert the code to the old state. The new tests should help us a little to not fall into this trap again.

ahmetmircik:
>I advanced the test to see "how far" we can get without javax.cache and trying to create a JCache. With the old code member and client would both start and you can even invoke getCacheManager()

@Donnerbart What was the reason of not throwing `java.lang.NoClassDefFoundError: javax/cache/Cache` upon `getCacheManager()` call? Seems throwing exception completely expected at first sight.

It will be better if we manage to mitigate code fragility besides fixing the issue.

Donnerbart:
I just tried out the old version with a real code sample and it's completely fine to retrieve the `ICacheManager`:
```
ICacheManager memberCacheManager = member.getCacheManager();
memberCacheManager.getCache("test");
```
```
ICacheManager clientCacheManager = client.getCacheManager();
clientCacheManager.getCache("test");
```
Both code snippets just fail on the `getCache()` call and actually fail with a custom exception, not a `ClassNotFoundException`:
```
java.lang.IllegalStateException: There is no valid JCache API library at classpath. Please be sure that there is a JCache API library in your classpath and it is newer than `0.x` and `1.0.0-PFD` versions!
	at com.hazelcast.instance.HazelcastInstanceCacheManager.getCacheByFullName(HazelcastInstanceCacheManager.java:54)
	at com.hazelcast.instance.HazelcastInstanceCacheManager.getCache(HazelcastInstanceCacheManager.java:45)
	at ...
```
```
java.lang.IllegalStateException: At client, There is no valid JCache API library at classpath. Please be sure that there is a JCache API library in your classpath and it is newer than `0.x` and `1.0.0-PFD` versions!
	at com.hazelcast.client.impl.ClientICacheManager.getCacheByFullName(ClientICacheManager.java:56)
	at com.hazelcast.client.impl.ClientICacheManager.getCache(ClientICacheManager.java:47)
	at ...
```
I think we should not change this in the 3.x version, since projects may rely on this, e.g. shared application code which fetches the `ICacheManager` in all scenarios, but just calls `getCache()` in deployments with a JCache configuration.

Actually it's a pity that `cacheManagerClazz.getDeclaredMethod("getCache", String.class)` in my test triggers a `ClassNotFoundException`, so we cannot check for our custom exception when `javax.cache` is not on the classpath.

ahmetmircik:
Makes sense to not to change this for 3.x, we can provide a proper fix in later versions for fragility issue.

Donnerbart:
Little code change, big impact :) Thanks for the reviews and the discussion.

PR #10137, opened 2017-03-23
Cluster Service Consistency Improvements

### LINKED ISSUE #9486: Cluster member-list update operations are not ordered
Cluster member-list update operations are not ordered, new member addition and removal operations can get reordered on receiving/processing side. Also periodic member-list publish operation has no order with other member adding/removing operations. That can cause having different member lists on different members. 

For example assume cluster has [A, B, C] members;  
- B shuts down
- A publishes [A, B, C] with a periodic member-list publish operation
- B is removed from both A and C. Member-list seen by;
   * A -> [A, C]
   * C -> [A, C]
- Periodic member-list publish operation is processed by C.  Member-list seen by;
   * A -> [A, C]
   * C -> [A, B, C]

After that point all membership updates will be rejected by C with following error (see `ClusterServiceImpl.shouldProcessMemberUpdate()`);
```
Received an inconsistent member update which contains new members 
and removes some of the current members! Ignoring and requesting a new member update...
```

This issue can be fixed by adding an incremental sequence in form `(master-uuid, sequence)` to all member update operations.

### LINKED ISSUE #9501: Latest member-list may not be received when FinalizeJoinOperation invocation timeouts
When `FinalizeJoinOperation` invocation during join timeouts, master can continue processing new join requests and publish new member lists. 
But if a member receives/executes `FinalizeJoinOperation` late, `MemberInfoUpdateOperation`s received before that are all rejected. In this situation, a member may not learn the latest member list until a periodic member list is published (default 5min) or a new member joins.

For example, assume A is master and a single node cluster;
- B starts up and sends a join request to A
- A handles the join request, updates member list as [A, B] and sends `FinalizeJoinOperation` to B
- A's `FinalizeJoinOperation` invocation timeouts, B has not processed `FinalizeJoinOperation` yet
- C starts up and sends a join request to A
- A handles the join request, updates member list as [A, B, C] and sends `FinalizeJoinOperation` to C and `MemberInfoUpdateOperation` to B
- B rejects `MemberInfoUpdateOperation` since it's not joined yet
- C executes `FinalizeJoinOperation` and updates member list as [A, B, C]
- B executes `FinalizeJoinOperation` that was sent earlier and updates member list as [A, B]

At that point, B cannot learn latest member list [A, B, C] until a periodic member list is published or a new member joins.

Related to https://github.com/hazelcast/hazelcast/issues/9486

mdogan:
Fixes consistency problems in membership management, member list replication and mastership claim mechanisms. 

- Membership update operations should be processed in the same order on all members as they are published by master.
- A later membership update should be able to tolerate a missing intermediate update. Each membership update should contain all the latest membership state.
- New master member should gather knowns member lists from cluster members and decide the most recent member list.

Fixes #9486
Fixes #9501

For detailed design see: https://hazelcast.atlassian.net/wiki/display/EN/Cluster+Service+Consistency+Improvements+Design

Some ideas are borrowed from _Group Membership and View Synchrony in Partitionable Asynchronous Distributed Systems: Specifications_ paper: http://dl.acm.org/citation.cfm?id=250010


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MembershipManager.java
```diff
@@ -0,0 +1,895 @@
...
+        } finally {
+            clusterServiceLock.unlock();
+        }
+    }
+
+    void handleExplicitSuspicion(MembersViewMetadata expectedMembersViewMetadata, Address suspectedAddress) {
+        clusterServiceLock.lock();
+        try {
+            MembersViewMetadata localMembersViewMetadata = createLocalMembersViewMetadata();
+            if (!localMembersViewMetadata.equals(expectedMembersViewMetadata)) {
```
mmedenjak: How come we ignore the request if the metadata isn't the same? Doesn't explicit suspicion occur when 2 members have different views - e.g. one member thinks it's master and sends a finalizeJoin operation or a member sends a heartbeat operation to the master but the sender is no longer a part of the cluster?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -192,125 +166,173 @@ public void init(NodeEngine nodeEngine, Properties properties) {
...
+                    logger.warning(endpoint + " has sent MasterConfirmation with " + membersViewMetadata
+                            + ", but it is not a member of this cluster!");
+                    // This guy knows me as its master but I am not. I should explicitly tell it to remove me from its cluster.
+                    // It should suspect me so that it can move on.
+                    // IMPORTANT: I should not tell it to remove me from cluster while I am trying to claim my mastership.
+
+                    sendExplicitSuspicion(membersViewMetadata);
+
+                    for (Member m : getMembers(NON_LOCAL_MEMBER_SELECTOR)) {
+                        sendExplicitSuspicionTrigger(m.getAddress(), membersViewMetadata);
```
mmedenjak: Just checking - this means that all members in the cluster (that master knows of) will send an explicit suspicion to the faulty node (the one that sent the master confirmation to this node)? So that the faulty node will suspect all members of this cluster?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -192,125 +166,173 @@ public void init(NodeEngine nodeEngine, Properties properties) {
...
-            if (destroyConnection && conn != null) {
-                conn.close(reason, null);
-            }
-            MemberImpl deadMember = getMember(deadAddress);
-            if (deadMember != null) {
-                removeMember(deadMember);
-                logger.info(membersString());
+
+            if (!membershipManager.clearMemberSuspicion(candidateAddress, "Mastership claim")) {
+                throw new IllegalStateException("Cannot accept mastership claim of " + candidateAddress + ". "
```
mmedenjak: Checking - the node that is claiming mastership will then become master only of those nodes that don't have a younger master already - possibly only a master of only itself?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -192,125 +166,173 @@ public void init(NodeEngine nodeEngine, Properties properties) {
...
+            MemberImpl masterCandidate = membershipManager.getMember(candidateAddress, candidateUuid);
+            checkTrue(masterCandidate != null ,
+                    candidateAddress + " claims mastership but it is not a member!");
+
+            MemberMap memberMap = membershipManager.getMemberMap();
+            if (!shouldAcceptMastership(memberMap, masterCandidate)) {
+                String message = "Cannot accept mastership claim of " + candidateAddress
+                        + " at the moment. There are more suitable master candidates in the member list.";
+                logger.fine(message);
+                throw new RetryableHazelcastException(message);
```
mmedenjak: What would happen if every node other than the one claiming mastership know that an older node is alive? E.g. We have [A,B,C,D,E]. Only B starts suspecting A and starts mastership claim. Other nodes will return `RetryableHazelcastException` to B since they know A is actually alive. B then receives a heartbeat from A but it will continue retrying the mastership claim.
Will B in the end become master only of itself - when the retry count is exceeded for all members? And then it will rejoin the cluster [A,C,D,E] by split brain handling?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MembershipManager.java
metanet: We ignore it because if the local state doesn't match with the expected state, it means that the local member has done something else so the current explicit suspicion response may be already stale. Additionally, if the explicit suspicion case is still valid, it will be triggered again eventually.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MembershipManager.java
```diff
@@ -0,0 +1,895 @@
...
+        return isNewMemberPresent;
+    }
+
+    private Future<MembersView> invokeFetchMembersViewOp(Address target, String targetUuid) {
+        Operation op = new FetchMembersViewOp(targetUuid).setCallerUuid(node.getThisUuid());
+
+        return nodeEngine.getOperationService()
+                .createInvocationBuilder(SERVICE_NAME, op, target)
+                .setTryCount(mastershipClaimTimeoutSeconds)
+                .setCallTimeout(TimeUnit.SECONDS.toMillis(mastershipClaimTimeoutSeconds)).invoke();
```
metanet: Yes, it is a trade-off between availability and correctness. If it is shorter, the cluster will become available earlier on master failure but it can kick out some slow members during the process.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
metanet: Yes. By that way, it will split quickly and merge back.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
metanet: Exactly. It will try to claim its mastership for some time but no one will respond. So it will split alone and merge back eventually.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
metanet: Members: [A, B, C, D]
B suspects A, C suspects [A, B], D suspects [A, B]. B and C start the mastership claim process. D first receives mastership claim of C and accepts it since it is a valid one. Then, D receives a mastership claim from B. At that point, it will reject it because it is too late to accept mastership claim of B since D has already accepted a younger node as its master: which is C.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MembershipManager.java
```diff
@@ -0,0 +1,895 @@
...
+        for (MemberImpl member : removedMembers) {
+            handleMemberRemove(memberMapRef.get(), member);
+        }
+
+        sendMembershipEvents(currentMemberMap.getMembers(), addedMembers);
+
+        MemberMap membersRemovedInNotActiveState = membersRemovedInNotActiveStateRef.get();
+        membersRemovedInNotActiveStateRef.set(MemberMap.cloneExcluding(membersRemovedInNotActiveState, members));
+
+        clusterHeartbeatManager.heartbeat();
```
metanet: I couldn't understand your question. Can you elaborate?

mmedenjak: The heartbeat method calculates the clock drift based on the last invocation - https://github.com/hazelcast/hazelcast/blob/111a9deb0636fa6d87327f714c8ca2676f592fd6/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterHeartbeatManager.java#L387.
So if it called in between periodic invocations, the absolute clock jump could be up to `heartbeatIntervalMillis/2`. I might be wrong here though.

metanet: Here, `clusterHeartbeatManager.heartbeat();` is done for the new members to set their initial heartbeat timestamps.

mmedenjak: Isn't the `MembershipManager#updateMembers` method also called from : 
- on the master when a new node is joining (`ClusterJoinManager#startJoin` -> `ClusterServiceImpl#updateMembers` -> `MembershipManager#updateMembers` -> `ClusterHeartbeatManager#heartbeat`). Here it is irrelevant of the new member.
- on all cluster members when a new node is joining (`ClusterJoinManager#startJoin` -> `MembersUpdateOp#run` -> `ClusterServiceImpl#updateMembers` -> `MembershipManager#updateMembers` -> `ClusterHeartbeatManager#heartbeat`
`ClusterServiceImpl#updateMembers`)
- on any member as a result of `TriggerMemberListPublishOp`
- on any member when publishing member list - `MembershipManager#publishMemberList`, `MembershipManager#removeMember`, `DecideNewMembersViewTask`
- on master when updating local member list on membership claim - `DecideNewMembersViewTask`
It seems like this method can be called when updating local member list in various situations.

metanet: Yes, it is called from the places you have pointed out. However, `clusterHeartbeatManager.heartbeat();` is called for the members which are absent in the current local member list but present in the new member list.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
```diff
@@ -702,42 +683,41 @@ private void startJoin() {
...
-                    Operation memberInfoUpdateOperation = new MemberInfoUpdateOperation(member.getUuid(), memberInfos, time,
+                    Operation op = new MembersUpdateOp(member.getUuid(), newMembersView, time,
                             partitionRuntimeState, true);
-                    calls.add(invokeClusterOperation(memberInfoUpdateOperation, member.getAddress()));
+                    op.setCallerUuid(node.getThisUuid());
+                    invokeClusterOp(op, member.getAddress());
                 }
 
-                int timeout = Math.min(calls.size() * FINALIZE_JOIN_TIMEOUT_FACTOR, FINALIZE_JOIN_MAX_TIMEOUT);
-                waitWithDeadline(calls, timeout, TimeUnit.SECONDS, whileFinalizeJoinsExceptionHandler);
```
metanet: Because we reduced periodic membership update duration. Membership publishes can be also triggered via heartbeats etc.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/MemberRemoveOperation.java
```diff
@@ -18,15 +18,17 @@
...
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
-import com.hazelcast.spi.impl.AllowedDuringPassiveState;
 
 import java.io.IOException;
 
-public class MemberRemoveOperation extends AbstractClusterOperation implements AllowedDuringPassiveState {
+@Deprecated
+// not used on 3.9+
```
metanet: We already do it in `com.hazelcast.internal.cluster.impl.MembershipManagerCompat#removeMember()` which is called within this operation.

metanet:
run-lab-run

metanet:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/AuthenticationFailureOp.java
```diff
@@ -21,7 +21,10 @@
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.spi.impl.NodeEngineImpl;
 
-public class AuthenticationFailureOperation extends AbstractClusterOperation {
+public class AuthenticationFailureOp extends AbstractClusterOperation {
+
+    public AuthenticationFailureOp() {
```
vbekiaris: Overriding the default `Operation` constructor will not set the validate target & 64-bit call timeout flags for this operation (https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/spi/Operation.java#L81-L82) -- is this really desirable?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/FetchMembersViewOp.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public int getId() {
+        return ClusterDataSerializerHook.FETCH_MEMBER_LIST_STATE;
+    }
+
+    @Override
+    protected void writeInternal(ObjectDataOutput out) throws IOException {
+        super.writeInternal(out);
```
vbekiaris: `super.read/writeInternal` are no-ops from `Operation`, could be removed

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/AuthenticationFailureOp.java
mdogan: Constructors cannot be overridden. `super()` will be implicitly called regardless of defining empty c'tor. Most of the operations already have empty c'tors.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/FetchMembersViewOp.java
mdogan: sure it can be removed.

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
```diff
@@ -90,7 +90,7 @@ private void handleGetClusterVersion(HttpGetCommand command) {
 
     private void handleCluster(HttpGetCommand command) {
         Node node = textCommandService.getNode();
-        StringBuilder res = new StringBuilder(node.getClusterService().membersString());
+        StringBuilder res = new StringBuilder(node.getClusterService().memberListString());
```
vbekiaris: The string format in the response to will be always the new one, no matter the value of `hazelcast.legacy.memberlist.format.enabled`, right?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/operations/AuthenticationFailureOp.java
vbekiaris: doh, please ignore!

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpGetCommandProcessor.java
mdogan: You're right. Member list format switch should affect the rest response too. will fix that.

PR #10138, opened 2017-03-23
Introduce NO_MIGRATION cluster state

mdogan:
`NO_MIGRATION` state stands between `ACTIVE` and `FROZEN` cluster states.
Like `ACTIVE` state, it allows new members to join but also doesn't allow
repartitioning/rebalancing and creating new backups similar to `FROZEN`.
Only promotions for missing primaries are performed which are already local,
no data movement is involved.

Once state becomes `ACTIVE`, repartitioning is triggered and missing backups
are replicated.

See [Gigantic Cache Migration Enhancements](https://hazelcast.atlassian.net/wiki/display/PM/Gigantic+Cache+Migration+Enhancements) requirement 3.

>Add manual control to use the new member added to the cluster so that the user can choose when to incur the degradation:
> 1. Once the cluster is started, Man Center can declare a NO_MIGRATION mode.
> 2. When a node is lost existing backup partitions get promoted. Avoid creating new backup partitions to prevent data migration in this mode.
> 3. If primary and backups are missing create new primary, empty partitions and also backup partitions as this does not cause any data copying.
> 4. When a node is added it has no partitions until Ops turns ACTIVE mode back on.
> 5. When this is done, partitions are added to the new node, backups are created and migrations performed. 

See [Gigantic Cache Migration Enhancements Design](https://hazelcast.atlassian.net/wiki/display/EN/Gigantic+Cache+Migration+Enhancements+Design) for more details.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
@@ -41,7 +41,9 @@ import java.util.List;
 import java.util.Set;
 
 import static com.hazelcast.util.StringUtil.bytesToString;
+import static com.hazelcast.util.StringUtil.lowerCaseInternal;
 import static com.hazelcast.util.StringUtil.stringToBytes;
+import static com.hazelcast.util.StringUtil.upperCaseInternal;
 
 public class HttpPostCommandProcessor extends HttpCommandProcessor<HttpPostCommand> {
     private static final byte[] QUEUE_SIMPLE_VALUE_CONTENT_TYPE = stringToBytes("text/plain");
@@ -116,16 +118,7 @@ public class HttpPostCommandProcessor extends HttpCommandProcessor<HttpPostComma
             if (!(groupConfig.getName().equals(groupName) && groupConfig.getPassword().equals(groupPass))) {
                 res = response(ResponseType.FORBIDDEN);
             } else {
-                ClusterState state = clusterService.getClusterState();
-                if (stateParam.equals("frozen")) {
-                    state = ClusterState.FROZEN;
-                }
-                if (stateParam.equals("active")) {
-                    state = ClusterState.ACTIVE;
-                }
-                if (stateParam.equals("passive")) {
-                    state = ClusterState.PASSIVE;
-                }
+                ClusterState state = ClusterState.valueOf(upperCaseInternal(stateParam));
                 if (!state.equals(clusterService.getClusterState())) {
                     clusterService.changeClusterState(state);
                     res = response(ResponseType.SUCCESS, "state", state.toString().toLowerCase());
@@ -148,7 +141,8 @@ public class HttpPostCommandProcessor extends HttpCommandProcessor<HttpPostComma
             if (!checkCredentials(command)) {
                 res = response(ResponseType.FORBIDDEN);
             } else {
-                res = response(ResponseType.SUCCESS, "state", clusterService.getClusterState().toString().toLowerCase());
+                ClusterState clusterState = clusterService.getClusterState();
+                res = response(ResponseType.SUCCESS, "state", lowerCaseInternal(clusterState.toString()));
             }
         } catch (Throwable throwable) {
             logger.warning("Error occurred while getting cluster state", throwable);
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
@@ -237,7 +237,7 @@ public abstract class AbstractJoiner implements Joiner {
             return false;
         }
 
-        if (!checkClusterStateActiveBeforeMerge(joinMessage)) {
+        if (!checkClusterStateAllowsJoinBeforeMerge(joinMessage)) {
             return false;
         }
 
@@ -311,9 +311,9 @@ public abstract class AbstractJoiner implements Joiner {
         return true;
     }
 
-    private boolean checkClusterStateActiveBeforeMerge(SplitBrainJoinMessage joinMessage) {
+    private boolean checkClusterStateAllowsJoinBeforeMerge(SplitBrainJoinMessage joinMessage) {
         ClusterState clusterState = clusterService.getClusterState();
-        if (clusterState != ClusterState.ACTIVE) {
+        if (!clusterState.isJoinAllowed()) {
             if (logger.isFineEnabled()) {
                 logger.fine("Should not merge to " + joinMessage.getAddress() + ", because this cluster is in "
                         + clusterState + " state.");
@@ -443,7 +443,7 @@ public abstract class AbstractJoiner implements Joiner {
 
     /**
      * Prepares the cluster state for cluster merge by changing it to {@link ClusterState#FROZEN}. It expects the current
-     * cluster state to be {@link ClusterState#ACTIVE}.
+     * cluster state to be {@link ClusterState#ACTIVE} or {@link ClusterState#NO_MIGRATION}.
      * The method will keep trying to change the cluster state until {@link GroupProperty#MERGE_NEXT_RUN_DELAY_SECONDS} elapses
      * or until the sleep period between two attempts has been interrupted.
      *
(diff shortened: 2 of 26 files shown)


mdogan:
run-lab-run

mdogan:
run-lab-run

mdogan:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
```diff
@@ -580,7 +578,7 @@ private void handleInvocationException(Exception e) {
 
     private void notifyWithExceptionWhenTargetIsNull() {
         ClusterState clusterState = context.clusterService.getClusterState();
-        if (clusterState == FROZEN || clusterState == PASSIVE) {
+        if (!clusterState.isMigrationAllowed()) {
             notifyError(new IllegalStateException("Partitions can't be assigned since cluster-state: " + clusterState));
```
mmedenjak: Note : the check now includes `IN_TRANSITION` and `NO_MIGRATION`. 
On a side note, I'm confused with the exception message. "Partitions can't be assigned..." - but isn't this just an operation invocation where the target happened to be `null`?

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionStateManager.java
```diff
@@ -149,7 +149,7 @@ boolean initializePartitionAssignments(Set<Address> excludedAddresses) {
         // if it's started and not locked the state yet.
         stateVersion.incrementAndGet();
         ClusterState clusterState = node.getClusterService().getClusterState();
-        if (clusterState != ClusterState.ACTIVE) {
+        if (!clusterState.isMigrationAllowed()) {
```
mmedenjak: Please update the javadoc. Same for the method below.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaStateChecker.java
```diff
@@ -162,7 +161,7 @@ private boolean hasMissingReplicaOwners() {
                 }
 
                 if (clusterService.getMember(address) == null
-                        && (!isClusterNotActive || !clusterService.isMemberRemovedWhileClusterIsNotActive(address))) {
+                        && (clusterState.isJoinAllowed() || !clusterService.isMemberRemovedWhileClusterIsNotActive(address))) {
```
mmedenjak: `IN_TRANSITION` doesn't pass this check anymore.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaStateChecker.java
```diff
@@ -259,7 +257,7 @@ private int invokeReplicaSyncOperations(int maxBackupCount, Semaphore semaphore,
                     continue;
                 }
 
-                if (!isClusterActive && clusterService.isMemberRemovedWhileClusterIsNotActive(replicaAddress)) {
+                if (!clusterState.isJoinAllowed() && clusterService.isMemberRemovedWhileClusterIsNotActive(replicaAddress)) {
```
mmedenjak: `IN_TRANSITION` is now allowed.

hazelcast/src/main/java/com/hazelcast/internal/ascii/rest/HttpPostCommandProcessor.java
```diff
@@ -116,16 +118,7 @@ private void handleChangeClusterState(HttpPostCommand command) throws Unsupporte
...
-                if (stateParam.equals("frozen")) {
-                    state = ClusterState.FROZEN;
-                }
-                if (stateParam.equals("active")) {
-                    state = ClusterState.ACTIVE;
-                }
-                if (stateParam.equals("passive")) {
-                    state = ClusterState.PASSIVE;
-                }
+                ClusterState state = ClusterState.valueOf(upperCaseInternal(stateParam));
```
mdogan: Yes, previous code silently fails when an illegal `stateParam` is received. I think making failure visible is better.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
```diff
@@ -313,7 +313,7 @@ private boolean checkMergeTargetIsNotMember(SplitBrainJoinMessage joinMessage) {
 
     private boolean checkClusterStateActiveBeforeMerge(SplitBrainJoinMessage joinMessage) {
         ClusterState clusterState = clusterService.getClusterState();
-        if (clusterState != ClusterState.ACTIVE) {
+        if (!clusterState.isJoinAllowed()) {
```
mdogan: Changing `checkClusterStateActivenBeforeMerge()` to `checkClusterStateAllowsJoinBeforeMerge()`.

mdogan: For `membersRemovedInNotActiveState` name, I tried to change it before, but couldn't find a good candidate. There are many methods referring the same name. 
I can change it to `membersRemovedInNotJoinableState`.

mmedenjak: I agree and as I've said it's not that big of a deal and will probably affect lots of code. We can do this later.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
```diff
@@ -456,19 +456,23 @@ private boolean prepareClusterState(ClusterServiceImpl clusterService) {
         }
 
         long until = Clock.currentTimeMillis() + mergeNextRunDelayMs;
-        while (clusterService.getClusterState() == ClusterState.ACTIVE) {
-            try {
-                clusterService.changeClusterState(ClusterState.FROZEN);
+        while (Clock.currentTimeMillis() < until) {
+            ClusterState clusterState = clusterService.getClusterState();
+            if (!clusterState.isMigrationAllowed() && !clusterState.isJoinAllowed()
```
mdogan: Yes you're right. Some other process can change cluster state concurrently. Merge process is not atomic. I would not even try to make it :) Existing merge mechanism is mostly kept as is.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -318,7 +320,7 @@ public boolean isMemberAllowedToJoin(Address address) {
         lock.lock();
         try {
             ClusterState clusterState = node.getClusterService().getClusterState();
-            if (clusterState == ClusterState.FROZEN || clusterState == ClusterState.PASSIVE) {
+            if (!clusterState.isMigrationAllowed()) {
```
mdogan: yes, partition service just decides by looking migration/partitioning aspect. Decides `isMemberAllowedToJoin()` based on partition service perspective.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/MigrationManager.java
```diff
@@ -490,7 +490,7 @@ void onShutdownRequest(Address address) {
         }
 
         ClusterState clusterState = node.getClusterService().getClusterState();
-        if (clusterState == ClusterState.FROZEN || clusterState == ClusterState.PASSIVE) {
+        if (!clusterState.isMigrationAllowed()) {
```
mdogan: `NO_MIGRATION` should be included. 

But `IN_TRANSITION` should be excluded, right. Perfect finding! Will fix that.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
mdogan: Still, I will add `clusterState != ClusterState.IN_TRANSITION`  check here too. To make it more visible.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/Invocation.java
mdogan: When migration is not allowed, even in `IN_TRANSITION` state, target may be null. Because partitions won't be assigned. 

Exception message is a little misleading. I'll change that to 

```java
new IllegalStateException("Target of invocation cannot be found! Partition owner is null "
                    + "but partitions can't be assigned in cluster-state: " + clusterState);
```

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionStateManager.java
mdogan: will do.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaStateChecker.java
mdogan: Actually checking `IN_TRANSITION` is redundant here, because to be able to change cluster state, we ensure that there are no ongoing/pending migrations.

mdogan: Btw I can write a comment above the cluster state check.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/PartitionReplicaStateChecker.java
mdogan: same as https://github.com/hazelcast/hazelcast/pull/10138#discussion_r112666414

PR #10139, opened 2017-03-23
Skip hooks implementing an expected interface, but loaded by a different classloader

jerrinot:
Fixing to #10124

Problem Statement:
the Mule container has Hazelcast 3.6 embedded. A business application deployed to the container has Hazelcast 3.8 inside. Mule can be configured to prefer application classes -> so an application can initiate Hazelcast 3.8 start. However then the service loading mechanism will find hooks from Hazecast 3.6 -> boom.

This PR is trying to prevent it

Goals:

Everything what was working before should work now. Regression in a patch release is a big deal.
Avoid introducing a new flag / configuration option
The idea is simple:
When a hook implements an interface with the expected name, but it still isn't assignable to the interface then we can assume it was loaded by a different classloader and we can skip it.

When the hook does not implement an interface with the expected name then it's likely a bug -> fail-fast.


hazelcast/src/main/java/com/hazelcast/instance/BuildInfoProvider.java
```diff
@@ -56,14 +58,16 @@ private BuildInfoProvider() {
...
+        BuildInfo buildInfo = readBuildPropertiesClass(BuildProperties.class, null);
+        try {
+            Class<?> enterpriseClass = BuildInfoProvider.class.getClassLoader()
+                    .loadClass("com.hazelcast.instance.EnterpriseBuildProperties");
+            buildInfo = readBuildPropertiesClass(enterpriseClass, buildInfo);
+        } catch (ClassNotFoundException e) {
+            ignore(e);
         }
+
+        Properties jetProperties = loadPropertiesFromResource("jet-runtime.properties");
```
mdogan: Someday, we're going to do the same for Jet, right?

hazelcast/src/main/java/com/hazelcast/util/ServiceLoader.java
```diff
@@ -338,32 +342,106 @@ public void remove() {
...
                 throw new HazelcastException(e);
             }
         }
 
+        private void onNonAssignableClass(String className, Class candidate) {
+            if (expectedType.isInterface()) {
+                if (ClassLoaderUtil.implementsInterfaceWithSameName(candidate, expectedType)) {
+                    // this can happen in application containers - different Hazelcast JARs are loaded
+                    // by different classloaders.
+                    LOGGER.fine("There appears to be a classloading conflict. "
```
mdogan: Should this logged more visible? like info or warning?

hazelcast/src/main/java/com/hazelcast/instance/BuildInfoProvider.java
jerrinot: correct. but we can do it only after Jet expose the class.

hazelcast/src/main/java/com/hazelcast/util/ServiceLoader.java
jerrinot: I am not sure. it's probably a business as usual when deployed an app. container. 
it provides no value unless you are troubleshooting it.

jerrinot:
run-lab-run

PR #10142, opened 2017-03-24
Add capability to promote a lite member to a normal member

mdogan:
Added a new method `Cluster.promoteLocalLiteMember()` which promotes
local member to a normal data member if it's a lite member. After promotion,
partition rebalancing is triggered if cluster state allows. Eventually
promoted member will own some portion of partitions.

See [Gigantic Cache Migration Enhancements](https://hazelcast.atlassian.net/wiki/display/PM/Gigantic+Cache+Migration+Enhancements) requirement 6.

See [Gigantic Cache Migration Enhancements Design](https://hazelcast.atlassian.net/wiki/display/EN/Gigantic+Cache+Migration+Enhancements+Design) for more details.

Depends on https://github.com/hazelcast/hazelcast/pull/10137

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/Cluster.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/Cluster.java
@@ -84,6 +84,18 @@ public interface Cluster {
     Member getLocalMember();
 
     /**
+     * Promotes local lite member to data member.
+     * When this method returns both {@link #getLocalMember()} and {@link #getMembers()}
+     * reflects the promotion.
+     *
+     * @throws IllegalStateException if member is not a lite member
+     * @throws IllegalStateException if mastership claim in progress
+     * @throws IllegalArgumentException if local member cannot be identified as a member of the cluster
+     * @since 3.9
+     */
+    void promoteLocalLiteMember();
+
+    /**
      * Returns the cluster-wide time in milliseconds.
      * <p/>
      * Cluster tries to keep a cluster-wide time which might be different than the member's own system time.
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
@@ -45,6 +45,7 @@ import com.hazelcast.internal.cluster.impl.operations.ShutdownNodeOp;
 import com.hazelcast.internal.cluster.impl.operations.SplitBrainMergeValidationOp;
 import com.hazelcast.internal.cluster.impl.operations.TriggerExplicitSuspicionOp;
 import com.hazelcast.internal.cluster.impl.operations.TriggerMemberListPublishOp;
+import com.hazelcast.internal.cluster.impl.operations.PromoteLiteMemberOp;
 import com.hazelcast.internal.partition.MigrationInfo;
 import com.hazelcast.internal.serialization.DataSerializerHook;
 import com.hazelcast.internal.serialization.impl.ArrayDataSerializableFactory;
@@ -102,8 +103,9 @@ public final class ClusterDataSerializerHook implements DataSerializerHook {
     public static final int TRIGGER_EXPLICIT_SUSPICION = 39;
     public static final int MEMBERS_VIEW_METADATA = 40;
     public static final int HEARTBEAT_COMPLAINT = 41;
+    public static final int PROMOTE_LITE_MEMBER = 42;
 
-    public static final int LEN = HEARTBEAT_COMPLAINT + 1;
+    static final int LEN = PROMOTE_LITE_MEMBER + 1;
 
     @Override
     public int getFactoryId() {
@@ -331,7 +333,11 @@ public final class ClusterDataSerializerHook implements DataSerializerHook {
                 return new HeartbeatComplaintOp();
             }
         };
-
+        constructors[PROMOTE_LITE_MEMBER] = new ConstructorFunction<Integer, IdentifiedDataSerializable>() {
+            public IdentifiedDataSerializable createNew(Integer arg) {
+                return new PromoteLiteMemberOp();
+            }
+        };
         return new ArrayDataSerializableFactory(constructors);
     }
 }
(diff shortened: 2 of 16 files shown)


mdogan:
run-lab-run

mdogan:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterDataSerializerHook.java
```diff
@@ -102,8 +103,9 @@
     public static final int TRIGGER_EXPLICIT_SUSPICION = 39;
     public static final int MEMBERS_VIEW_METADATA = 40;
     public static final int HEARTBEAT_COMPLAINT = 41;
+    public static final int PROMOTE_LITE_MEMBER = 42;
 
-    public static final int LEN = HEARTBEAT_COMPLAINT + 1;
```
pveentjer: Will this cause any minor version incompatibilities or is the system smart enough not to upgrade an 'old' lite member?

mdogan: We're checking cluster version before invoking the api method, see:

```java
    @Override
    public void promoteLocalLiteMember() {
        if (getClusterVersion().isLessThan(Versions.V3_9)) {
            throw new UnsupportedOperationException("Lite member promotion is not available!");
        }
    ...
```

pveentjer: ok

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
```diff
@@ -84,6 +84,18 @@
     Member getLocalMember();
 
     /**
+     * Promotes local lite member to data member.
+     * When this method returns both {@link #getLocalMember()} and {@link #getMembers()}
+     * reflects the promotion.
+     *
+     * @throws IllegalStateException if member is not a lite member
+     * @throws IllegalStateException if mastership claim in progress
+     * @throws IllegalArgumentException if local member cannot be identified as a member of the cluster
```
mdogan: hmm... I can change `IllegalArgumentException ` to `IllegalStateException `.  `IllegalArgumentException` is thrown by `MembershipManager#promoteToNormalMember` and that exception is delegated to the caller.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -345,16 +352,28 @@ public void merge(Address newTargetAddress) {
...
             resetClusterId();
+            clearInternalState();
         } finally {
             lock.unlock();
         }
     }
 
+    private void resetLocalMemberUuid() {
+        assert lock.isHeldByCurrentThread() : "Called without holding cluster service lock!";
+        assert !isJoined() : "Cannot reset local member uuid when joined.";
```
mdogan: `resetLocalMemberUuid()` is already a private method and it's called in a controlled flow. Assertion is mostly to document the behaviour, not for correctness of the method.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterServiceImpl.java
```diff
@@ -1020,6 +1061,60 @@ public MembershipManagerCompat getMembershipManagerCompat() {
...
+        PromoteLiteMemberOp op = new PromoteLiteMemberOp();
+        op.setCallerUuid(member.getUuid());
+
+        InternalCompletableFuture<MembersView> future =
+                nodeEngine.getOperationService().invokeOnTarget(SERVICE_NAME, op, master.getAddress());
+        MembersView view = future.join();
+
+        lock.lock();
+        try {
+            if (!member.getAddress().equals(master.getAddress())) {
```
mdogan: yes, possible. `MembershipManager#promoteToNormalMember()` applies the changes locally.

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
```diff
@@ -338,9 +338,9 @@ public void memberAdded(MemberImpl member) {
         logger.fine("Adding " + member);
         lock.lock();
         try {
-            if (!member.localMember()) {
+//            if (!member.localMember()) {
```
mdogan: yes, local member check removed for the case lite member is master case. I forgot to remove commented section.

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
mmedenjak: I meant to ask about this line - https://github.com/hazelcast/hazelcast/pull/10142/files#diff-49146434024bde944cb0d360926213efR1090. 
It would mean that it applied a new MemberView but it still was a lite member. But looking at the `MembershipManager#promoteToNormalMember` I can see only 2 outcomes - either it is promoted and the new member view is sent OR and exception is thrown (which will occur on the caller).

hazelcast/src/main/java/com/hazelcast/internal/partition/impl/InternalPartitionServiceImpl.java
mdogan: I think you're right about cluster state. Lite member promotion should not be allowed when migrations are not allowed by cluster state. I'll send a new PR to fix that.

hazelcast/src/main/java/com/hazelcast/core/Cluster.java
mdogan: Aha.. that case happens when master changes during lite member promotion. 

Suppose lite member sens request to master (Member-A). 
- Member-A processes the request, sends the response back but then crashes without being able to publish new member list.
- Member-B becomes the new master and lite member accepts its mastership. 

In this case, `updateMembers(view, master.getAddress(), master.getUuid())` call on lite member will fail, because master is changed. So, even though it receives a successful response to its request from the former master, its promotion is not applied to member list. That's why it fails there.

PR #10148, opened 2017-03-26
Experimental settings for io backoff and response backoff

pveentjer:
In case of the io system using the selectnow, a configurable
backoff policy is available instead of always spinning.

The same goes for response queue for member and client.

It can be configured using
```
somesystemproperty=selectnow,20,50,1,100
somesystemproperty==backoff,20,50,11,100
```
20=spins
50=yields
1 is min park nanos
100 is max park nanos

These experimental settings aren't set by default, so the user
explicitly needs to set either selectnow or configure the
response queue to get any behavior different than the default
behavior.

Under certain conditions doing a backoff instead of blocking
can lead to an interesting performance gain. These settings have
been added to do further experiments and case specific tuning.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
@@ -24,6 +24,7 @@ import com.hazelcast.internal.util.counters.SwCounter;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.spi.impl.operationexecutor.OperationHostileThread;
 import com.hazelcast.util.EmptyStatement;
+import com.hazelcast.util.concurrent.IdleStrategy;
 
 import java.io.IOException;
 import java.nio.channels.CancelledKeyException;
@@ -81,8 +82,13 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
 
     private final SelectorMode selectMode;
 
+    private final IdleStrategy idleStrategy;
+
     // last time select unblocked with some keys selected
     private volatile long lastSelectTimeMs;
+
+    private volatile boolean stop;
+
     // set to true while testing
     private boolean selectorWorkaroundTest;
 
@@ -90,15 +96,7 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
                                String threadName,
                                ILogger logger,
                                IOOutOfMemoryHandler oomeHandler) {
-        this(threadGroup, threadName, logger, oomeHandler, SelectorMode.SELECT);
-    }
-
-    public NonBlockingIOThread(ThreadGroup threadGroup,
-                               String threadName,
-                               ILogger logger,
-                               IOOutOfMemoryHandler oomeHandler,
-                               SelectorMode selectMode) {
-        this(threadGroup, threadName, logger, oomeHandler, selectMode, newSelector(logger));
+        this(threadGroup, threadName, logger, oomeHandler, SelectorMode.SELECT, null);
     }
 
     public NonBlockingIOThread(ThreadGroup threadGroup,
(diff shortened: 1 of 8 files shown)


hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
```diff
@@ -245,7 +254,7 @@ private void coolDown() {
     }
 
     private void selectLoop() throws IOException {
-        while (!isInterrupted()) {
+        while (!stop) {
```
Donnerbart: I guess we used `interrupt()` in `shutdown()` to interrupt the `taskQueue.poll()` and finish this loop with a single state. After this change we might not be able to leave the loop, if anything else than `shutdown()` can interrupt this thread (since the `stop` flag will not be set). This might also effect the normal `selectLoop()` methods, not just your experimental `selectNowLoop()`.

Do we need a
```
while (!stop) {
    try {
        ...
        Thread.currentThread.interrupt();
        stop = true; // or break / return
    }
}
```
to handle interruption correctly?

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThreadingModel.java
```diff
@@ -72,6 +75,9 @@
     private volatile IOBalancer ioBalancer;
     private boolean selectorWorkaroundTest = Boolean.getBoolean("hazelcast.io.selector.workaround.test");
 
+    private BackoffIdleStrategy idleStrategy;
+
+
```
Donnerbart: Superfluous newline

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThreadingModel.java
```diff
@@ -97,6 +103,11 @@ public NonBlockingIOThreadingModel(
     private SelectorMode getSelectorMode() {
         if (selectorMode == null) {
             selectorMode = SelectorMode.getConfiguredValue();
+
+            String selectorModeString = SelectorMode.getConfiguredString();
+            if (selectorModeString.startsWith("selectnow,")) {
```
Donnerbart: Can't we move this string magic to `SelectorMode` instead of spreading this across the code?

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/SelectorMode.java
```diff
@@ -27,19 +27,22 @@
...
-        String valueToCheck = StringUtil.isNullOrEmptyAfterTrim(value) ? null : value.trim().toLowerCase();
-        if (valueToCheck == null) {
+        if (value.equals("select")) {
             return SELECT;
-        } else if (valueToCheck.equals("selectnow")) {
-            return SELECT_NOW;
-        } else if (valueToCheck.equals("selectwithfix")) {
+        } else if (value.equals("selectwithfix")) {
             return SELECT_WITH_FIX;
+        } else if (value.equals("selectnow") || value.startsWith("selectnow,")) {
```
Donnerbart: `SELECT_NOW` is just used in 6 places. I would vote for a new constant `SELECT_WITH_IDLE_STRATEGY` instead of this hacky way. It doesn't look hard to implement.

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThreadingModel.java
pveentjer: How is this any clearer?

```
public enum SelectorMode {
    SELECT,
    SELECT_NOW,
    SELECT_WITH_FIX;

    public static final String SELECT_STRING = "select";
    public static final String SELECTWITHFIX_STRING = "selectwithfix";
    public static final String SELECT_NOW_STRING = "selectnow";

    public static SelectorMode getConfiguredValue() {
        return fromString(getConfiguredString());
    }

    public static String getConfiguredString() {
        return System.getProperty("hazelcast.io.selectorMode", SELECT_STRING).trim().toLowerCase();
    }

    public static SelectorMode fromString(String value) {
        if (value.equals(SELECT_STRING)) {
            return SELECT;
        } else if (value.equals(SELECTWITHFIX_STRING)) {
            return SELECT_WITH_FIX;
        } else if (value.equals(SELECT_NOW_STRING) || value.startsWith(SELECT_NOW_STRING + ",")) {
            return SELECT_NOW;
        } else {
            throw new IllegalArgumentException(format("Unrecognized selectorMode [%s]", value));
        }
    }
}
```

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/SelectorMode.java
pveentjer: I question the readability improvement of the suggested change. I doubt it is making code any easier to read.

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
pveentjer: Noting else than shutdown should interrupt a thread. Interrupt is not a general purpose method everybody can invoke and see what happens.

PR #10150, opened 2017-03-27
[DONT MERGE] Near cache prototype fixed stale read detector

Donnerbart:


hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/record/AbstractNearCacheRecord.java
```diff
@@ -42,8 +43,9 @@
             AtomicLongFieldUpdater.newUpdater(AbstractNearCacheRecord.class, "recordState");
 
     protected long creationTime = TIME_NOT_SET;
-    protected long sequence;
+    protected MetaDataContainer metaDataContainer;
```
Donnerbart: This is the price for the speedup: one reference to the `MetaDataContainer` per `NearCacheRecord`

ahmetmircik: what was wrong with `partitionId` as in this [commit](https://github.com/hazelcast/hazelcast/pull/10150/commits/8d4f8573ed29a29498856e3b8f872f93de453153)? I think it was also HD friendly.

PR #10173, opened 2017-03-29
Cache.getAsync on a near-cached client cache should increase stats.cacheGets by one

vbekiaris:
The stats maintenance callback should be registered for execution once.
Adds tests for asserting `ICache` stats when async methods are used. Parametrizes client-side cache statistics test to execute on plain and near-cached client-side cache proxy.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
@@ -145,8 +145,8 @@ public class NearCachedClientCacheProxy<K, V> extends ClientCacheProxy {
     protected ClientDelegatingFuture getAsyncInternal(Data dataKey, ExpiryPolicy expiryPolicy, ExecutionCallback callback) {
         try {
             long reservationId = nearCache.tryReserveForUpdate(dataKey);
-            ClientDelegatingFuture future = super.getAsyncInternal(dataKey, expiryPolicy, callback);
-            future.andThenInternal(new GetAsyncCallback(dataKey, reservationId, callback), true);
+            GetAsyncCallback getAsyncCallback = new GetAsyncCallback(dataKey, reservationId, callback);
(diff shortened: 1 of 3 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -146,7 +146,7 @@ protected ClientDelegatingFuture getAsyncInternal(Data dataKey, ExpiryPolicy exp
         try {
             long reservationId = nearCache.tryReserveForUpdate(dataKey);
             ClientDelegatingFuture future = super.getAsyncInternal(dataKey, expiryPolicy, callback);
-            future.andThenInternal(new GetAsyncCallback(dataKey, reservationId, callback), true);
+            future.andThenInternal(new GetAsyncCallback(dataKey, reservationId), true);
```
vbekiaris: This is the fix: the statistics maintenance callback is already registered with `andThenInternal` in above statement, no need to execute it again within `GetAsyncCallback`.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -146,7 +146,7 @@ protected ClientDelegatingFuture getAsyncInternal(Data dataKey, ExpiryPolicy exp
         try {
             long reservationId = nearCache.tryReserveForUpdate(dataKey);
             ClientDelegatingFuture future = super.getAsyncInternal(dataKey, expiryPolicy, callback);
```
ahmetmircik: Shouldn't it be something like this? So we only have 1 execution callback instead of 2.
```
GetAsyncCallback getAsyncCallback = new GetAsyncCallback(dataKey, reservationId, callback);
ClientDelegatingFuture future = super.getAsyncInternal(dataKey, expiryPolicy, getAsyncCallback);
```

vbekiaris: thanks, updated as suggested

Issue #10127, opened 2017-03-22
Increase test coverage for NearCachedClientCacheProxy


ahmetmircik:
see https://github.com/hazelcast/hazelcast/pull/10035#discussion_r107198910

PR #10175, opened 2017-03-29
Returned values from near cache on getAsync call

ahmetmircik:

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/internal/nearcache/AbstractNearCacheBasicTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/internal/nearcache/AbstractNearCacheBasicTest.java
@@ -328,15 +328,20 @@ public abstract class AbstractNearCacheBasicTest<NK, NV> extends HazelcastTestSu
      */
     @Test
     public void testGetAsyncPopulatesNearCache() throws Exception {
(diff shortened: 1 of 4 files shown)


hazelcast/src/test/java/com/hazelcast/internal/nearcache/AbstractNearCacheBasicTest.java
```diff
@@ -335,7 +335,8 @@ public void testGetAsyncPopulatesNearCache() throws Exception {
 
         // generate Near Cache hits
         populateNearCacheAsync(context);
-
+        assertEquals("Not equals to expected Near Cache hit count",
```
Donnerbart: If you can assert the misses as well then `assertNearCacheStats(NearCacheTestContext<?, ?, ?, ?> context, long expectedOwnedEntryCount, long expectedHits, long expectedMisses)` would be a better match. Then you could also get rid of the `assertNearCacheSize()` call.

Donnerbart: If not then you can use `context.stats` instead of `context.nearCache.getNearCacheStats()`

ahmetmircik: done.

PR #10176, opened 2017-03-29
Fix typography 2

viliam-durina:
- "text : text" should be "text: text" and few more like that
- some grammar fixed


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientSmartInvocationServiceImpl.java
```diff
@@ -75,7 +75,7 @@ public void invokeOnTarget(ClientInvocation invocation, Address target)
             throw new NullPointerException("Target can not be null");
         }
         if (!isMember(target)) {
-            throw new TargetNotMemberException("Target :  " + target + " is not member. ");
+            throw new TargetNotMemberException("Target \"" + target + "\" is not a member.");
```
Donnerbart: I'm not sure if we normally use `"` or `'` for quoting variables in logs. I think in the `toString()` it's normally `'` due to the default of IDEA auto-creation. I'm not sure if we are consistent in the code with this though.

viliam-durina: I replaced with `'`.

PR #10178, opened 2017-03-29
Added annotation MethodNotAvailable to ease unified Near Cache testing

Donnerbart:
* added annotation `MethodNotAvailable` to mark methods in implementations
  of `DataStructureAdapter` which are not available in that data structure
* added `MethodAvailableException` to be thrown by unavailable methods
  to have a clear indication that a test used an unsupported method
* added `MethodAvailableMatcher` to easily test if a method is available
* added method `assumeThatMethodIsAvailable()` to `NearCacheTestUtils` to
  automatically skip tests which require a specific method
* changed all `DataStructureAdapter` implementations to make use of the
  `MethodNotAvailable` annotation and `MethodNotAvailableException`
* changed `AbstractNearCacheBasicTest` to check if required methods are
  available before using them

The motivation behind this change is a growing complexity of the unified Near Cache tests due to some methods not being supported by all data structures. A good example is the `ReplicatedMap` which doesn't implement a lot of asynchronous methods which `IMap`  and `ICache`  have.

There are several drawbacks with the actual solution:
* Some async methods are faked by using the sync method and wrapping it in an artificial `SimpleCompletedFuture` instance. A failure on such wrapped methods is misleading.
* Some methods just differ in their return type, e.g. a method from `IMap` returns the value and the same method from `JCache` returns just a `boolean`. This is also faked in the `DataStructureAdapter`, but this leads to incomplete tests.
* Some methods cannot be faked, so they throw an `UnsupportedMethodException` which has to be handled by overriding the unified tests which use that method in the test class.
* Some methods are just implemented by a single data structure, so it's likely that they are not tested at all.

All in all this leads to not all methods being tested and some methods being tested multiple times. This is not helping the code coverage and just wastes CPU time.

In addition some methods are not working properly, e.g. some invalidations in the `ReplicatedMap` are broken. Since we wrapped the broken `remove()` method to fake `removeAsync()` we also had to ignore those async tests for the `ReplicatedMap`, although that method doesn't even exist. This leads to a wrong impression of what is tested and which feature combinations are really working.

To solve this issue I came up with the `@MethodNotAvailable`  annotation to mark methods which are not implemented in a specific `DataStructureAdapter`. With the helper method `assumeThatMethodIsAvailable()` we can easily test for methods which are not implemented in all scenarios. With `assumeThat()` unavailable combinations are automatically skipped without any manual effort. So with this PR the implementation of the `DataStructureAdapter`  defines if a test will be executed or not.

With this test framework addition we can write unified tests once and have them executed just in the scenarios which actually support that method. We still have a TODO list in form of ignored tests for methods which are implemented but broken (see `ClientReplicatedMapNearCacheBasicTest`). Also if we would implement missing methods to a data structure in the future, we just have to implement them in its `DataStructureAdapter` and the existing tests will automatically pick them up.


hazelcast-client/src/test/java/com/hazelcast/client/replicatedmap/nearcache/ClientReplicatedMapNearCacheBasicTest.java
```diff
@@ -120,30 +120,6 @@ public void whenCacheIsFull_thenPutOnSameKeyShouldUpdateValue_withUpdateOnNearCa
     public void whenPutAllIsUsed_thenNearCacheShouldBeInvalidated_withUpdateOnNearCacheAdapter() {
     }
 
-    @Test(expected = UnsupportedOperationException.class)
-    @Override
-    public void whenReplaceIsUsed_thenNearCacheShouldBeInvalidated_withUpdateOnNearCacheAdapter() {
```
Donnerbart: These kind of `@Override` methods are not needed anymore. They are not indicating that an existing method is broken, just that a method is not available in this scenario.

hazelcast-client/src/test/java/com/hazelcast/client/replicatedmap/nearcache/ClientReplicatedMapNearCacheBasicTest.java
```diff
@@ -155,16 +131,4 @@ public void whenRemoveIsUsed_thenNearCacheShouldBeInvalidated_withUpdateOnNearCa
     @Ignore(value = "The ClientReplicatedMapProxy is missing `invalidateNearCache(keyData)` calls")
     public void whenRemoveIsUsed_thenNearCacheShouldBeInvalidated_withUpdateOnDataAdapter() {
     }
-
-    @Test
-    @Override
-    @Ignore(value = "The ClientReplicatedMapProxy is missing `invalidateNearCache(keyData)` calls")
-    public void whenRemoveAsyncIsUsed_thenNearCacheShouldBeInvalidated_withUpdateOnNearCacheAdapter() {
```
Donnerbart: These kind of methods are actually worse, because they indicate that an async method is broken, but it doesn't exist at all. In this case the sync method is broken, which is internally used by the `DataStructureAdapter`. With this PR this test is just skipped, because the async method doesn't exist in this scenario.

hazelcast/src/main/java/com/hazelcast/internal/adapter/DataStructureAdapter.java
```diff
@@ -56,4 +56,43 @@
     LocalMapStats getLocalMapStats();
 
     boolean containsKey(K key);
+
+    /**
+     * Contains all methods of {@link DataStructureAdapter}.
+     */
+    enum DataStructureMethods implements DataStructureAdapterMethod {
```
Donnerbart: This is the only place where we need to add additional information in the future. For every new method in the `DataStructureAdapter` we need to add the signature information here, so the `assumeThatMethodIsAvailable()` method can check for the `@MethodIsUnavailable` annotation.

hazelcast/src/main/java/com/hazelcast/internal/adapter/DataStructureAdapterMethod.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.adapter;
+
+/**
+ * Defines the signature of a method in {@link DataStructureAdapter} to search for available methods.
+ */
+public interface DataStructureAdapterMethod {
```
Donnerbart: I pulled out this interface so I can test the `MethodAvailableMatcher`  without adding garbage to the real code.

hazelcast/src/main/java/com/hazelcast/internal/adapter/IMapDataStructureAdapter.java
```diff
@@ -52,8 +52,9 @@ public boolean putIfAbsent(K key, V value) {
     }
 
     @Override
+    @MethodNotAvailable
     public ICompletableFuture<Boolean> putIfAbsentAsync(K key, V value) {
-        return new SimpleCompletedFuture<Boolean>(map.putIfAbsent(key, value) == null);
+        throw new MethodNotAvailableException();
```
Donnerbart: This is an example of a formerly faked async method, which is now properly marked as "not available".

hazelcast/src/test/java/com/hazelcast/internal/nearcache/AbstractNearCacheBasicTest.java
```diff
@@ -259,139 +261,139 @@ private void whenPutAllIsUsed_thenNearCacheShouldBeInvalidated(boolean useNearCa
...
      */
     @Test
     public void whenReplaceWithOldValueIsUsed_thenNearCacheShouldBeInvalidated_withUpdateOnDataAdapter() {
         nearCacheConfig.setInvalidateOnChange(true);
-        whenReplaceIsUsed_thenNearCacheShouldBeInvalidated(false, true);
+        whenReplaceIsUsed_thenNearCacheShouldBeInvalidated(false, DataStructureMethods.REPLACE_WITH_OLD_VALUE);
     }
 
-    private void whenReplaceIsUsed_thenNearCacheShouldBeInvalidated(boolean useNearCacheAdapter, boolean useOldValue) {
+    private void whenReplaceIsUsed_thenNearCacheShouldBeInvalidated(boolean useNearCacheAdapter, DataStructureMethods method) {
```
Donnerbart: We have less chained `boolean` parameters now, so the test code is easier to read and less error prone to write.

hazelcast/src/test/java/com/hazelcast/internal/nearcache/AbstractNearCacheBasicTest.java
```diff
@@ -259,139 +261,139 @@ private void whenPutAllIsUsed_thenNearCacheShouldBeInvalidated(boolean useNearCa
...
         NearCacheTestContext<Integer, String, NK, NV> context = createContext();
+        DataStructureAdapter<Integer, String> adapter = useNearCacheAdapter ? context.nearCacheAdapter : context.dataAdapter;
+        assumeThatMethodIsAvailable(adapter, method);
```
Donnerbart: This is all we need to do to check if the required method is available in the executed scenario. If the method is not available `assumeThat()` will stop the test without a failure.

hazelcast/src/test/java/com/hazelcast/internal/nearcache/AbstractNearCacheBasicTest.java
```diff
@@ -259,139 +261,139 @@ private void whenPutAllIsUsed_thenNearCacheShouldBeInvalidated(boolean useNearCa
...
+            if (method == DataStructureMethods.REPLACE_WITH_OLD_VALUE) {
                 assertTrue(adapter.replace(i, "value-" + i, "newValue-" + i));
             } else {
                 assertEquals("value-" + i, adapter.replace(i, "newValue-" + i));
             }
         }
 
         int expectedNearCacheSize = useNearCacheAdapter && isCacheOnUpdate(nearCacheConfig) ? DEFAULT_RECORD_COUNT : 0;
-        assertNearCacheSizeEventually(context, expectedNearCacheSize, "Invalidation is not working on replace()");
+        String message = format("Invalidation is not working on %s()", method.getMethodName());
```
Donnerbart: A convenient way to create useful assert messages.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
The spurious failures with the owned entry count should be unrelated. I managed to reproduce it once on my local machine with the actual master. I've created an issue for this, so we can fix it seperately: https://github.com/hazelcast/hazelcast/issues/10193

hazelcast/src/main/java/com/hazelcast/internal/adapter/DataStructureAdapter.java
Donnerbart: Yeah, I tried some reflection magic as well, but had ran into the same issue. I'll try to come up with a test which doesn't need too much maintenance on it's own.

hazelcast/src/main/java/com/hazelcast/internal/adapter/DataStructureAdapter.java
```diff
@@ -56,4 +56,43 @@
...
+        CLEAR("clear"),
+        SET("set", Object.class, Object.class),
+        PUT("put", Object.class, Object.class),
+        PUT_IF_ABSENT("putIfAbsent", Object.class, Object.class),
+        PUT_IF_ABSENT_ASYNC("putIfAbsentAsync", Object.class, Object.class),
+        REPLACE("replace", Object.class, Object.class),
+        REPLACE_WITH_OLD_VALUE("replace", Object.class, Object.class),
```
Donnerbart: Very good finding and a good test case for the missing test :)

hazelcast/src/main/java/com/hazelcast/internal/adapter/DataStructureAdapter.java
Donnerbart: I've added a small and simple `DataStructureMethodsTest` which creates a runtime implemention of `DataStructureAdapter` via ByteBuddy (was already used in the project). So there is no additional test implementation which needs to be maintained when we add new methods.

I added two tests:
* all referenced methods have to be found
* all referenced methods have to have a unique signature

The second one is catching the error you've found manually:
```
java.lang.AssertionError: DataStructureAdapter::replace(Object, Object) is referenced by DataStructureMethods.REPLACE_WITH_OLD_VALUE, but was already referenced by DataStructureMethods.REPLACE
```
The first one would look like this:
```
java.lang.AssertionError: DataStructureAdapter::get(String) is referenced by DataStructureMethods.GET, but could not be found!
```

Donnerbart: NOTE: It's still not a constraint that all methods of `DataStructureAdapter` have to be defined in `DataStructureMethods`. I think that's okay, e.g. if they are anyway dependent on another method which is already defined.

If that check is needed it should be easy to add to the existing test (get all methods from `DataStructureAdapter` and check if they are in the `knownMethods` map of the test.

Donnerbart:
run-lab-run

hazelcast/src/test/java/com/hazelcast/internal/adapter/DataStructureMethodsTest.java
```diff
@@ -0,0 +1,57 @@
...
+        Class<?> dataStructureAdapterImpl = new ByteBuddy()
+                .subclass(DATA_STRUCTURE_ADAPTER)
+                .make()
+                .load(getClass().getClassLoader(), ClassLoadingStrategy.Default.WRAPPER)
+                .getLoaded();
+
+        Map<Method, DataStructureMethods> knownMethods = new HashMap<Method, DataStructureMethods>();
+        for (DataStructureMethods method : DataStructureMethods.values()) {
+            try {
+                Method adapterMethod = dataStructureAdapterImpl.getMethod(method.getMethodName(), method.getParameterTypes());
```
vbekiaris: There is no need to create an implementation class, could use directly `DataStructureAdapter.class.getMethod...`

Donnerbart: Stupid me... I first tried to get an implementation whose methods could be invoked, so I had a look at Byte Buddy to have a real instance. But for the signature checking this was not needed, but I left Byte Buddy in the code. I simplified the test now. Very good finding, thanks!

PR #10180, opened 2017-03-29
Add WAN publisher support for discovery SPI

mmedenjak:
Change WAN connection manager, publisher configuration and implementation  to support discovery SPI. The current mechanism of defining endpoints is turned into an implementation of the discovery SPI. The connection manager now reruns discovery on a periodic basis and adds the discovered nodes to the available list. The size of the list can be limited.

Notes:
- the striped executor in the WanBatchReplication is sized according to the number of target endpoints when sending the first event and does not resize afterwards
- the AbstractWanReplication implementation of isConnected has changed - previously it checked that there were no failed connections, now it checks if there are any discovered endpoints
- the periodic discovery task does not attempt to connect to the target endpoint before adding it to the list

EE part : https://github.com/hazelcast/hazelcast-enterprise/pull/1390


mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

Issue #10107, opened 2017-03-20
[query] ClientQueryBounceTest.testQuery


hasancelik:
**master**
```
org.junit.runners.model.TestTimedOutException: test timed out after 300000 milliseconds
	at java.lang.Thread.sleep(Native Method)
	at java.lang.Thread.sleep(Thread.java:340)
	at java.util.concurrent.TimeUnit.sleep(TimeUnit.java:360)
	at com.hazelcast.test.HazelcastTestSupport.sleepSeconds(HazelcastTestSupport.java:269)
	at com.hazelcast.test.bounce.BounceMemberRule.testWithDuration(BounceMemberRule.java:236)
	at com.hazelcast.test.bounce.BounceMemberRule.testRepeatedly(BounceMemberRule.java:203)
	at com.hazelcast.map.QueryBounceTest.prepareAndRunQueryTasks(QueryBounceTest.java:92)
	at com.hazelcast.map.QueryBounceTest.testQuery(QueryBounceTest.java:75)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:105)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:97)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.lang.Thread.run(Thread.java:745)
```

[...]

Issue #9870, opened 2017-02-13
[query] QueryBounceTest.testQueryWithIndexes


hasancelik:
```
java.lang.AssertionError: Obtained 995 results for query 'age >= 44814 AND age < 45814' expected:<1000> but was:<995>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.map.QueryBounceTest$QueryRunnable.run(QueryBounceTest.java:135)
	at com.hazelcast.test.bounce.BounceMemberRule$TestTaskRunable.run(BounceMemberRule.java:505)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast/1239/testReport/junit/com.hazelcast.map/QueryBounceTest/testQueryWithIndexes/

hasancelik:
**ClientQueryBounceTest.testQueryWithIndexes**
```
java.lang.AssertionError: Obtained 1003 results for query 'age >= 24492 AND age < 25492' expected:<1000> but was:<1003>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.map.QueryBounceTest$QueryRunnable.run(QueryBounceTest.java:135)
	at com.hazelcast.test.bounce.BounceMemberRule$TestTaskRunable.run(BounceMemberRule.java:505)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
[...]

PR #10181, opened 2017-03-30
BounceMemberRule logging & fixes, smaller cluster in QueryBounceTest

vbekiaris:
- Smaller cluster and test drivers size (4 member & 4 test drivers instead of 6 & 5 respectively) in QueryBounceTest to reduce probability of timeout. In local debugging, sometimes initial cluster was formed after 1 minute which does count towards the timeout defined in `@Test(timeout=...)`.
- Added logging in BounceMemberRule for easier troubleshooting
- Bounced members and test drivers are now accessed with volatile semantics, as they are used by several threads:
  - Cluster setup is executed in the thread that initializes the `AbstractHazelcastClassRunner` (eg when running within the IDE this is the `main` thread)
  - The `FailOnTimeoutStatement` that wraps every test to enforce the timeout spawns a separate thread that invokes the test method. `BounceMemberRule.test*` will be invoked from this thread and will typically need to obtain at least a test driver `HazelcastInstance`
  - Another thread is spawned from `BounceMemberRule` that executes the member bouncing logic and updates the current cluster members in `BounceMemberRule.members`

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/bounce/BounceMemberRule.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/bounce/BounceMemberRule.java
@@ -18,6 +18,8 @@ package com.hazelcast.test.bounce;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.logging.ILogger;
+import com.hazelcast.logging.Logger;
 import com.hazelcast.nio.ClassLoaderUtil;
 import com.hazelcast.test.TestHazelcastInstanceFactory;
 import com.hazelcast.test.bounce.BounceTestConfiguration.DriverType;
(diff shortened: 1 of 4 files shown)


hazelcast/src/test/java/com/hazelcast/test/bounce/BounceMemberRule.java
```diff
@@ -363,25 +388,33 @@ public void evaluate() throws Throwable {
 
     // wait until all test tasks complete or one of them throws an exception
     private void waitForFutures(Future[] futures) {
+        // do not wait more than 30 seconds
+        long deadline = currentTimeMillis() + 30 * 1000;
```
Donnerbart: I would make the timout a constant like `TIMEOUT_SECONDS` and use `TimeUnit.Seconds.toMillis(TIMEOUT_SECONDS)` here (and of course use the constant in the comment and warning log below as well).

vbekiaris: thx, fixed

vbekiaris:
- Member bouncing thread is now a daemon thread.
- Fixes timeouts in `QueryBounceTest` & `ClientQueryBounceTest` #10107 & #9870 : sometimes `HazelcastInstance.getMap` in the `QueryRunnable` constructor could take minutes to obtain a reference while members are bouncing; this time was counted against the test method's timeout but not included in the bouncing test duration, so the 3 minutes of bouncing test duration would start minutes after the `FailOnTimeoutStatement` started counting the default 5-minute `@Test` method timeout.

mdogan:
run-lab-run

vbekiaris:
thanks @mdogan @Donnerbart

Issue #10065, opened 2017-03-13
Multiple entry EntryProcessors do not respect locked entries


mmedenjak:
In case of multiple entry `EntryProcessor` (`MultipleEntryOperation` and `PartitionWideEntryOperation`), the processor will execute on the entry regardless if it is locked or not.

Currently the entire EntryProcessor is run on the partition thread so in theory we could check if all affected entries are locked and park the operation if so. On unlock & notification we re-check and again park the operation until all affected entries are unlocked. This sounds simple but would require a special `WaitNotifyKey` for a partition wide-notification and it would require some changes in the operation system since it now expects that one notification can notify a single waiting key.

With the introduction of EP `Offloadable` the situation is more complex since the EP can now be interleaved with other operations (also lock and unlock). This is why we need to check if the entry is locked before running the EP logic on every key. If the entry is locked, there are two possible approaches:

1. Park the EP operation. This requires substantial changes in the operation system since the operation isn't supposed to be parked/unparked after it started running.
2. Create a EntryOperation for the locked key and continue processing the unlocked keys. Once the EntryOperation is finished update the value on the multiple-entry operation and check if all entries have been processed. This is more in line with existing operations where parts of the work are forked onto other operations and using callbacks we notify the parent operation (e.g. in WAN - `WanSyncOperation`).

These tests demonstrate the difference introduced by adding `Offloadable` interface for EP which relies on locking keys for entry update.

```
@RunWith(HazelcastParallelClassRunner.class)
@Category({QuickTest.class, ParallelTest.class})
public class MixedEntryProcessorsTest extends HazelcastTestSupport {

    public static final String MAP_NAME = "EntryProcessorOffloadableTest";

    private HazelcastInstance[] instances;

    @Override
    public Config getConfig() {
        Config config = super.getConfig();
        MapConfig mapConfig = new MapConfig(MAP_NAME);
        mapConfig.setInMemoryFormat(OBJECT);
        mapConfig.setBackupCount(1);
        config.addMapConfig(mapConfig);
        return config;
    }

    @Before
    public void before() {
        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(3);
        instances = factory.newInstances(getConfig());
    }

    @Test
    public void withOffloading() {
        runTest(true);
    }

    @Test
    public void withoutOffloading() {
        runTest(false);
    }

    private void runTest(final boolean offload) {
        final String key = generateKeyOwnedBy(instances[0]);
[...]

PR #10187, opened 2017-03-30
Allows checking if an entry is locked within an EntryProcessor

tombujok:
Allows checking if an entry is locked within an EntryProcessor

Issues to discuss:

- don't know if the entry within EntryBackupProcessor should be LockAware - the EP may send a delta, or it may check for locks too - the only problem is if the operations get the out-of-order, but that problem exists even without this optimisation, so I would allow the user to check for locks on backup too. If no out-of-order -> then it's the same as on the main partition. If out-of-order than the problem exists anyway.
- Offloadable (and not ReadOnly) will lock the entry -> so the user has to be aware of that, otherwise they may be surprise that the `isLocked` method returns true always.

Related to #10065

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/core/IMap.java
+++ b/hazelcast/src/main/java/com/hazelcast/core/IMap.java
@@ -22,6 +22,7 @@ import com.hazelcast.map.MapInterceptor;
 import com.hazelcast.map.QueryCache;
 import com.hazelcast.map.QueryResultSizeExceededException;
 import com.hazelcast.map.impl.LegacyAsyncMap;
+import com.hazelcast.map.LockAware;
 import com.hazelcast.map.listener.MapListener;
 import com.hazelcast.map.listener.MapPartitionLostListener;
 import com.hazelcast.mapreduce.JobTracker;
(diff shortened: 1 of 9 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/LockAwareLazyMapEntry.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.hazelcast.internal.serialization.InternalSerializationService;
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.query.impl.getters.Extractors;
+
+/**
+ * Specialization of the LazyMapEntry that is LockAware. Used in EntryProcessor.
+ * If serialized will be deserialized as the LazyMapEntry, since the locked information is volatile, valid to the run only.
+ */
+public class LockAwareLazyMapEntry extends LazyMapEntry implements LockAware {
```
mmedenjak: The test failure was because `LockAwareLazyMapEntry` extends `LazyMapEntry` which is IDS :
- it has no default constructor
- it has the same type ID

tombujok: yeah, thx for looking into this failure. I'm aware of that, it was actually intended. but due to the convention testing it fails. I'll probably make the class a separate IDS.

mmedenjak:
Looks fine except for the issues with the IDS. I think the backup entry should be lock aware as well. Otherwise it would be unexpected that the user cannot cast into the `LockAwareLazyMapEntry` in the backup processor and the `BackupEntryProcessor` would just run on all entries which would more easily lead to differences in data.
The out-of-order of operations on the backup should be detected by the `InternalPartitionService#updatePartitionReplicaVersions` causing it to sync with the primary.

hazelcast/src/main/java/com/hazelcast/core/IMap.java
```diff
@@ -73,6 +74,25 @@
...
+ * <li>{ @link IMap#submitToKey(Object, EntryProcessor, ExecutionCallback) }</li>
+ * </ul>
+ * There are however following methods that run the EntryProcessor on more than one entry. These operations are not lock-aware.
+ * The EntryProcessor will process the entries no matter if they are locked or not.
+ * The user may however check if an entry is locked by casting the {@link java.util.Map.Entry} to ]
+ * {@link LockAware} and invoking the {@link LockAware#isLocked()} method.
+ * <ul>
+ * <li>{ @link IMap#executeOnEntries(EntryProcessor) }</li>
+ * <li>{ @link IMap#executeOnEntries(EntryProcessor, Predicate) } </li>
+ * <li>{ @link IMap#executeOnKeys(Set, EntryProcessor) }</li>
```
mmedenjak: Remove the spaces after `{` and before `}` for it to become links. We could also add some javadoc for the lock awareness on those methods.

hazelcast/src/main/java/com/hazelcast/map/impl/LockAwareLazyMapEntry.java
tombujok: Fixed.

hazelcast/src/main/java/com/hazelcast/core/IMap.java
tombujok: Fixed.

hazelcast/src/main/java/com/hazelcast/core/IMap.java
```diff
@@ -73,6 +74,25 @@
  * </p>
  * <p>This class does <em>not</em> allow <tt>null</tt> to be used as a key or value.</p>
  *
+ * <p>Entry Processing</p>
+ * The following operations are lock-aware, since they operate on a single key only.
+ * If the key is locked the EntryProcessor will run until it acquires the lock.
```
jerrinot: " If the key is locked the EntryProcessor will *wait* until it acquires the lock." ?

tombujok: Fixed

tombujok:
@mmedenjak Matko, could you finalize your review?

Issue #10196, opened 2017-03-31
Servers should throw exception when message type is not known


sancar:
Protocol is defined to throw exception when a client sends a client message with unknown type to server.  Throwing exception is missing in the code and should be added. 
Versions need planned be fixed
- [x] 3.9-SNAPSHOT master
- [x] 3.8.2-SNAPSHOT maintenance
- [x] 3.8.1
- [ ] 3.7.7
- [ ] 3.6.8

PR #10197, opened 2017-03-31
Servers throw exception when gets unkown message type

sancar:
fixes https://github.com/hazelcast/hazelcast/issues/10196

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/NoSuchMessageTask.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/NoSuchMessageTask.java
@@ -44,8 +44,10 @@ public class NoSuchMessageTask
 
     @Override
     protected void processMessage() {
-        logger.warning("Unrecognized client message received with type: 0x"
-                + Integer.toHexString(parameters.getMessageType()));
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/NoSuchMessageTask.java
```diff
@@ -44,8 +44,10 @@ protected ClientMessage encodeResponse(Object response) {
 
     @Override
     protected void processMessage() {
-        logger.warning("Unrecognized client message received with type: 0x"
-                + Integer.toHexString(parameters.getMessageType()));
+        String message = "Unrecognized client message received with type: 0x"
+                + Integer.toHexString(parameters.getMessageType());
+        logger.warning(message);
+        throw new UnsupportedOperationException(message);
```
mdogan: Is caller going to receive `UnsupportedOperationException` directly? Or will it be wrapped inside another exception?

sancar: Yes, caller is going to get `UnsupportedOperationException` exception. 
I have just updated the test to use same method to peel exception when used over a proxy as our users will do.

PR #10210, opened 2017-04-03
Fixed Near Cache eviction-policy XSD default value from NONE to LRU

Donnerbart:
I checked out the code of v3.0 and the default value for the Near Cache `EvictionPolicy` has always been `LRU`. This is a copy & paste error from the `MapConfig` which has `NONE` as default.

```
public final static String DEFAULT_EVICTION_POLICY = "LRU";


pveentjer:
Merging since this is just doc update

PR #10211, opened 2017-04-03
Fixed assertion on same NearCacheStats instance

Donnerbart:
We had the following spurious failure in `com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheTest.testNearCacheMaxIdleRecordsExpired[batchInvalidationEnabled:true]`:
```
we expected to have all map entries in the Near Cache or already expired (NearCacheStatsImpl{ownedEntryCount=3530, ownedEntryMemoryCost=494200, creationTime=1491229736166, hits=0, misses=5000, ratio=0.0%, evictions=0, expirations=1469, lastPersistenceTime=0, persistenceCount=0, lastPersistenceDuration=0, lastPersistenceWrittenBytes=0, lastPersistenceKeyCount=0, lastPersistenceFailure=''})
```
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/7972/testReport/junit/com.hazelcast.client.map.impl.nearcache/ClientMapNearCacheTest/testNearCacheMaxIdleRecordsExpired_batchInvalidationEnabled_true_/

I think we have two issues with this test:

* The `map.getLocalMapStats().getNearCacheStats()` always returns the same instance of the `NearCacheStatsImpl`. So it makes no sense to get a "before something" `NearCacheStats` and compare the same field with an instance which is fetched later. To fix this I've added a copy constructor to the `NearCacheStatsImpl` and fixed the tests accordingly. I've added the method `getNearCacheStatsCopy()` for convenient access.

* The other issue is that two values in the stats are never updated in an atomic operation. In the test we have eviction enabled, so at some point in time we decrease the `ownedEntryCount` and increase the `evictionCount` after each other. We also read those two values after each other for the assert. So in the given failure we had a total value of 4999 instead of the wanted 5000. To fix this I used `assertTrueEventually()` to check for the summed value.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTest.java
@@ -259,7 +259,6 @@ public class NearCacheTest extends NearCacheTestSupport {
         // make some hits
         populateNearCache(map, mapSize);
 
-        stats = getNearCacheStats(map);
         long hits = stats.getHits();
         misses = stats.getMisses();
         long hitsAndMisses = hits + misses;
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTestSupport.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTestSupport.java
@@ -33,6 +33,7 @@ import com.hazelcast.map.impl.MapServiceContext;
 import com.hazelcast.map.impl.proxy.NearCachedMapProxyImpl;
 import com.hazelcast.map.listener.EntryEvictedListener;
 import com.hazelcast.monitor.NearCacheStats;
+import com.hazelcast.monitor.impl.NearCacheStatsImpl;
 import com.hazelcast.spi.impl.NodeEngineImpl;
 import com.hazelcast.test.AssertTask;
 import com.hazelcast.test.HazelcastTestSupport;
@@ -43,6 +44,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static com.hazelcast.config.EvictionConfig.MaxSizePolicy.ENTRY_COUNT;
 import static com.hazelcast.instance.BuildInfoProvider.BUILD_INFO;
@@ -71,7 +73,7 @@ public class NearCacheTestSupport extends HazelcastTestSupport {
         // populate Near Caches
         populateNearCache(map, size);
 
-        NearCacheStats statsBeforeEviction = getNearCacheStats(map);
+        NearCacheStats statsBeforeEviction = getNearCacheStatsCopy(map);
 
         // trigger eviction via fetching the extra entry
         map.get(size);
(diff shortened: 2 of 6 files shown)


hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTest.java
```diff
@@ -259,7 +259,6 @@ public void testNearCacheStats() {
         // make some hits
         populateNearCache(map, mapSize);
 
-        stats = getNearCacheStats(map);
```
Donnerbart: We don't need to retrieve the same instance again.

hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTestSupport.java
```diff
@@ -71,7 +72,7 @@ protected void testNearCacheEviction(IMap<Integer, Integer> map, int size) {
         // populate Near Caches
         populateNearCache(map, size);
 
-        NearCacheStats statsBeforeEviction = getNearCacheStats(map);
+        NearCacheStats statsBeforeEviction = getNearCacheStatsCopy(map);
```
Donnerbart: Here we need a copy to keep the old statistic values.

hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTestSupport.java
```diff
@@ -86,16 +87,22 @@ protected void testNearCacheEviction(IMap<Integer, Integer> map, int size) {
         assertEquals("got the wrong eviction count", expectedEvictions, stats.getEvictions());
         assertEquals("got the wrong expiration count", 0, stats.getExpirations());
         assertEquals("we expect the same hits", statsBeforeEviction.getHits(), stats.getHits());
-        assertEquals("we expect the same misses", statsBeforeEviction.getMisses(), stats.getMisses());
+        assertEquals("we expect one miss more", statsBeforeEviction.getMisses() + 1, stats.getMisses());
```
Donnerbart: Now this assert makes sense :)

hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTestSupport.java
```diff
@@ -86,16 +87,22 @@ protected void testNearCacheEviction(IMap<Integer, Integer> map, int size) {
...
 
     protected void testNearCacheExpiration(final IMap<Integer, Integer> map, final int size, int expireSeconds) {
         populateMap(map, size);
         populateNearCache(map, size);
 
-        final NearCacheStats statsBeforeExpiration = getNearCacheStats(map);
-        assertTrue(format("we expected to have all map entries in the Near Cache or already expired (%s)", statsBeforeExpiration),
-                statsBeforeExpiration.getOwnedEntryCount() + statsBeforeExpiration.getExpirations() >= size);
+        final NearCacheStats statsBeforeExpiration = getNearCacheStatsCopy(map);
+        assertTrueEventually(new AssertTask() {
```
Donnerbart: This is the fix for the non-atomic issue with the ` NearCacheStats`.

hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTestSupport.java
```diff
@@ -106,13 +113,17 @@ public void run() {
                 map.get(0);
 
                 NearCacheStats stats = getNearCacheStats(map);
-                assertEquals("we expect the same hits", statsBeforeExpiration.getHits(), stats.getHits());
-                assertEquals("we expect the same misses", statsBeforeExpiration.getMisses(), stats.getMisses());
-                assertEquals("we expect all entries beside the 'trigger entry' to be deleted from the Near Cache",
+                assertEquals(format("we expect all entries beside the 'trigger entry' to be evicted (%s)", stats),
```
Donnerbart: Most of these asserts needed a little overhaul.

hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/NearCacheTestSupport.java
```diff
@@ -86,34 +88,48 @@ protected void testNearCacheEviction(IMap<Integer, Integer> map, int size) {
...
                 // map.get() triggers Near Cache eviction/expiration process,
                 // but we need to call this on every assert since the Near Cache has a cooldown for expiration cleanups
                 map.get(0);
+                long invocations = invocationCounter.incrementAndGet();
 
-                NearCacheStats stats = getNearCacheStats(map);
-                assertEquals("we expect the same hits", statsBeforeExpiration.getHits(), stats.getHits());
-                assertEquals("we expect the same misses", statsBeforeExpiration.getMisses(), stats.getMisses());
-                assertEquals("we expect all entries beside the 'trigger entry' to be deleted from the Near Cache",
+                NearCacheStats stats = getNearCacheStatsCopy(map);
```
Donnerbart: Here I get a stats copy again. The copy may suffer from the non-atomic updates, but at least the numbers don't change during the asserts. Since I print the whole stats in the assert message it could be very confusing otherwise, if one gets something like `we expect 5000 expirations, but was 5 (NearCachStatsImpl{..., expiration=6, ...})` due to ongoing evictions.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

PR #10225, opened 2017-04-05
[3.8.1][BACKPORT] Fixes unlocking in exceptional situation

tombujok:


hazelcast/src/main/java/com/hazelcast/map/impl/operation/EntryOperation.java
```diff
@@ -248,76 +251,157 @@ private void throwModificationInReadOnlyException() {
...
-                    // EntryOffloadableLockMismatchException is a marker send from the EntryOffloadableSetUnlockOperation
-                    // meaning that the whole invocation of the EntryOffloadableOperation should be retried
-                    if (t instanceof EntryOffloadableLockMismatchException) {
-                        t = new RetryableHazelcastException(t.getMessage(), t);
+            private void retry(final Operation op) {
+                setUnlockRetryCount++;
+                if (isRetryLimitReached()) {
+                    try {
+                        Object exceeded = new HazelcastException("Set & Unlock retry limit exceeded for key " + dataKey);
+                        getOperationResponseHandler().sendResponse(EntryOperation.this, exceeded);
```
mdogan: So, when retry limit exceeded, lock will remain acquired forever? In that case I think we should somehow force unlock the key.

tombujok: I was thinking about it thoroughly before, and the reasoning is as follows:
- I need to get to the partition thread the to unlock it, if I can't do it, how can I force unlock it?
I could try doing it, but I may not be on the partition thread, nor the partition may be available for me (because it got migrated away or sth similar).
- in theory, if there's no migrations and there is no hazelcast internal error this operation should eventually succeed, so maybe we should just retry indefinitely?

mdogan: Problem is, when you get a retryable response/exception (maybe apart from WrongTargetException, since operation is guaranteed to be local), you cannot give up until you get a non-retryable failure or a success response.

Suppose partition is migrating when you call unlock. So, you get retry exception and retried until you reach the cap. Then you give up and return failure response to the caller, without able to unlock. 
Then partition migration failed for some reason (target crashed etc), and migration rolled back. Now, partition is still on former node and key remains locked forever. 

Both migrations and heartbeat have timeout mechanisms. So, there should not be infinite blocking/retrying here. 

To me either there shouldn't be any retry limit or lock should be acquired with a lease (something greater than heartbeat timeout).

tombujok: OK, agreed. I removed the retry limit since it's only on retryable (and not on WrongTarget) and on Timeout, so eventually it should finish.

PR #10226, opened 2017-04-05
Added unified Near Cache tests for EntryProcessor methods

Donnerbart:
* added `EntryProcessor` related methods to `DataStructureAdapter`
* added unified Near Cache tests for invalidations via `EntryProcessor`s
* added unified Near Cache tests for population via `EntryProcessor`s

To ease the review I kept the original three commits.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/adapter/DataStructureAdapter.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/adapter/DataStructureAdapter.java
@@ -20,6 +20,9 @@ import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.monitor.LocalMapStats;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
+import javax.cache.processor.EntryProcessor;
+import javax.cache.processor.EntryProcessorException;
+import javax.cache.processor.EntryProcessorResult;
 import java.util.Map;
 import java.util.Set;
 
@@ -50,6 +53,12 @@ public interface DataStructureAdapter<K, V> {
 
     ICompletableFuture<V> removeAsync(K key);
 
+    <T> T invoke(K key, EntryProcessor<K, V, T> entryProcessor, Object... arguments) throws EntryProcessorException;
+
+    Object executeOnKey(K key, com.hazelcast.map.EntryProcessor entryProcessor);
+
+    Map<K, Object> executeOnKeys(Set<K> keys, com.hazelcast.map.EntryProcessor entryProcessor);
+
     boolean containsKey(K key);
 
     Map<K, V> getAll(Set<K> keys);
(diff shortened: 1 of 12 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
```diff
@@ -267,10 +266,10 @@ public void clear() {
         Data keyData = toData(key);
         Data epData = toData(entryProcessor);
 
-        return (T) invokeInternal(keyData, epData, arguments);
+        return (T) toObject(invokeInternal(keyData, epData, arguments));
     }
 
-    protected Object invokeInternal(Data keyData, Data epData, Object... arguments) {
+    protected Data invokeInternal(Data keyData, Data epData, Object... arguments) {
```
Donnerbart: I moved the `toObject()` to the (only) invoking method, to have the `Data` instance available in the near cached proxy.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -349,11 +349,19 @@ public void clear() {
...
+        Data valueData = null;
         try {
-            return super.invokeInternal(keyData, epData, arguments);
+            valueData = super.invokeInternal(keyData, epData, arguments);
+            return valueData;
         } finally {
-            invalidateNearCache(keyData);
+            // we just have the value as `Data` here, so it doesn't make sense to use cacheOrInvalidate(),
+            // which looks for the best candidate between `Object` and `Data` format
+            if (cacheOnUpdate && valueData != null) {
```
Donnerbart: I think we can get `null` here, but I'm still missing that test case. I just faked it an got an assertion error later, since we are not allowed to store `null`. So I'm sure that we are not allowed to safe `null` here. But I'm not 100% sure if it's correct to call `invalidateNearCache(keyData);` in that case or if I should just ignore the result.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -349,11 +349,19 @@ public void clear() {
     }
 
     @Override
-    protected Object invokeInternal(Data keyData, Data epData, Object[] arguments) {
+    protected Data invokeInternal(Data keyData, Data epData, Object[] arguments) {
+        Data valueData = null;
         try {
-            return super.invokeInternal(keyData, epData, arguments);
+            valueData = super.invokeInternal(keyData, epData, arguments);
```
ahmetmircik: Seems it can be problematic to put result of entry-processor(EP) into near cache. Because EP can return any arbitrary type of object, result type is not limited to type of value. For example, a user defined type is a valid return type. Please see [this](https://github.com/jsr107/jsr107spec/blob/master/src/main/java/javax/cache/processor/EntryProcessor.java#L151). We can be on safe side by always invalidating near cache.

Donnerbart: That's a good catch :) Then I'll remove the tests for that code path.

Issue #9153, opened 2016-10-21
Backup lost if "maxIdle" property is used


MeYoGui:
Hi,

I'm facing an issue with an IMap distributed on two different nodes.

 I'm using this IMap with :
- only one "**Async Backup**"
- **MaxIdleSeconds** set to 10 seconds.  
- **readBackupData** set to false
- Default **eviction-policy** (NONE)

In the beginning I'm putting some values in the map but after that the values are only accessed and not updated for quite a long period (like 10 or 15 minutes).

What is the problem ?

In the beginning (before 10 seconds elapsed), the backup is working great ... 

![image](https://cloud.githubusercontent.com/assets/4062869/19607392/7e4044e0-9797-11e6-9ee1-891c06ff26d3.png)

But after 10 seconds : all the backups are lost !

![image](https://cloud.githubusercontent.com/assets/4062869/19607413/ad5f069e-9797-11e6-8b07-409142533aa4.png)

_If I change the amount of seconds for the MaxIdle property to an higher value (let's say 300 seconds) then everything is working well until 300 seconds._ 

**The backups gets evicted if I use MaxIdle property, is that a normal behavior ?** 

For information : using a "sync" backup didn't solve the problem at all.
The only "workaround"is to use a TTL instead but that's not what I want because my values are not updated for a certain amount of time and I want eviction if not "accessed" for 10 seconds.

I don't have any workaround here and it's quite a severe issue to integrate Hazelcast in our system.
If you want more information, don't hesitate.

Hazelcast version : 3.7.2

ahmetmircik:
Hi @MeYoGui, if you are using `imap#get`, can you please try to get value via EntryProcessor by using `imap#executeOnKey`? You can touch backup entry with EntryBackupProcessor in this case and expiration time will be shifted.

MeYoGui:
Hi @ahmetmircik, 
Yes I'm using get()  method from IMap. 
I'll try your solution with the executeOnKey()  method as you propose. I'll let you know if it's working. 

I'm just not sure about the performance of this way of doing... Is that efficient ?  (I'm doing almost 3k get/s)

MeYoGui:
I tried via the EntryProcessor. It worked well, the backup is now touched and I haven't seen any lack of performance except that the code is less intuitive now...

You may update your documentation to tell about this behavior as this is not expected at all (and quite dangerous : backup is lost), don't you think ?

Are you still going to change the get() method to touch the backup ?

[...]

PR #10227, opened 2017-04-05
Backup expirations

ahmetmircik:
fixes https://github.com/hazelcast/hazelcast/issues/9153

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
@@ -16,26 +16,38 @@
 
 package com.hazelcast.map.impl.eviction;
 
-import com.hazelcast.config.Config;
+import com.hazelcast.internal.cluster.ClusterService;
+import com.hazelcast.internal.nearcache.impl.invalidation.InvalidationQueue;
+import com.hazelcast.map.impl.MapService;
 import com.hazelcast.map.impl.PartitionContainer;
 import com.hazelcast.map.impl.operation.ClearExpiredOperation;
+import com.hazelcast.map.impl.operation.EvictBatchBackupOperation;
+import com.hazelcast.map.impl.recordstore.ExpiredKey;
 import com.hazelcast.map.impl.recordstore.RecordStore;
 import com.hazelcast.nio.Address;
 import com.hazelcast.spi.ExecutionService;
 import com.hazelcast.spi.NodeEngine;
 import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.OperationResponseHandler;
+import com.hazelcast.spi.OperationService;
 import com.hazelcast.spi.impl.operationservice.InternalOperationService;
 import com.hazelcast.spi.partition.IPartition;
 import com.hazelcast.spi.partition.IPartitionService;
+import com.hazelcast.spi.properties.HazelcastProperties;
+import com.hazelcast.spi.properties.HazelcastProperty;
 import com.hazelcast.util.Clock;
+import com.hazelcast.version.Version;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Queue;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledFuture;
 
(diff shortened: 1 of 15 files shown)


ahmetmircik:
run-lab-run

ahmetmircik:
run-lab-run

ahmetmircik:
run-lab-run

ahmetmircik:
@jerrinot can you please have a look again?

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
```diff
@@ -293,27 +347,95 @@ private boolean notHaveAnyExpirableRecord(PartitionContainer partitionContainer)
...
+        for (RecordStore recordStore : maps.values()) {
+            sendExpiredKeysToBackups(recordStore, false);
         }
     }
 
+    public void sendExpiredKeysToBackups(RecordStore recordStore, boolean checkIfReachedBatch) {
+        InvalidationQueue invalidationQueue = recordStore.getExpiredKeys();
+
+        int size = invalidationQueue.size();
+        if (size == 0 || checkIfReachedBatch && size < MAX_EXPIRED_KEY_COUNT_IN_BATCH) {
```
jerrinot: shouldn't this also by take time-from-last-sending into consideration? 

imagine a single key expires. they I presume you want to eventually send this information to the backup even when no other key expires after that. something like the Nagle's algorithm basically.

jerrinot:
@ahmetmircik: did you consider having a (hidden?) switch to re-introduce to old behavior? perhaps there are situations where extra backup operations do more harm than good. the impl is there anyway due the Rolling Upgrade.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EvictBatchBackupOperation.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public void run() {
+        assert getNodeEngine().getClusterService().getClusterVersion().isGreaterOrEqual(Versions.V3_9);
+
+        for (Data key : expiredKeys) {
+            recordStore.evict(key, true);
+        }
+
+        // equalize backup entry count to owner entry count to have identical memory occupancy
+        int diff = recordStore.size() - ownerPartitionEntryCount;
```
jerrinot: can you please elaborate this part a bit? why is it needed?

jerrinot:
I am thinking about this scenario:
1. key `K` expires. the primary put `K` into the expiration queue (for batching)
2. someone does `map(K, foo)` and inserts the same key again. This will create a new entry on a primary  and overwrites the backup entry on the backup replica.
3. eventually the expired keys are sent from primary to the backup replica. this will remove the key `K` from the backup replica even when it should not.

is this scenario realistic? is there anything we can do against it?

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
ahmetmircik: It is already like that. Background expiration task sets `checkIfReachedBatch` to `false` and it scans queue periodically regardless of the batch size.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/EvictBatchBackupOperation.java
ahmetmircik: please see [this](https://hazelcast.atlassian.net/wiki/display/EN/Partition+Owner+Should+Be+Responsible+for+Expirations+and+Evictions+on+Replicas+-+Design?focusedCommentId=131339648#comment-131339648)

ahmetmircik:
@jerrinot yes good catch. It is possible. I merged a fix for that here.

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/ExpirationManager.java
```diff
@@ -161,6 +186,12 @@ private int getInteger(String propertyName, int defaultValue) {
         return property == null ? defaultValue : parseInt(property);
     }
 
+    private boolean getBoolean(String propertyName, boolean defaultValue) {
+        Config config = nodeEngine.getConfig();
+        String property = config.getProperty(propertyName);
+        return property == null ? defaultValue : Boolean.parseBoolean(property);
```
ahmetmircik: fixed by using `HazelcastProperty`instead

PR #10230, opened 2017-04-05
Create map proxy without triggering creation of map container

vbekiaris:
Also adds tests to verify map proxy is created without triggering map container creation.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -195,7 +195,7 @@ abstract class MapProxySupport extends AbstractDistributedObject<MapService> imp
         this.partitionService = getNodeEngine().getPartitionService();
         this.lockSupport = new LockProxySupport(new DefaultObjectNamespace(SERVICE_NAME, name),
                 LockServiceImpl.getMaxLeaseTimeInMillis(properties));
-        this.operationProvider = mapServiceContext.getMapOperationProvider(name);
+        this.operationProvider = mapServiceContext.getMapOperationProvider(mapConfig);
         this.operationService = nodeEngine.getOperationService();
         this.serializationService = nodeEngine.getSerializationService();
         this.thisAddress = nodeEngine.getClusterService().getThisAddress();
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -195,7 +195,7 @@ protected MapProxySupport(String name, MapService service, NodeEngine nodeEngine
         this.partitionService = getNodeEngine().getPartitionService();
         this.lockSupport = new LockProxySupport(new DefaultObjectNamespace(SERVICE_NAME, name),
                 LockServiceImpl.getMaxLeaseTimeInMillis(properties));
-        this.operationProvider = mapServiceContext.getMapOperationProvider(name);
+        this.operationProvider = mapServiceContext.getMapOperationProvider(mapConfig);
```
vbekiaris: This is the fix: `getMapOperationProvider(String mapName)` gets-or-creates the map container in order to obtain the config and choose the appropriate operation provider. Since the config is already available, it is passed to `getMapOperationProvider(Config config)` and skips the map container lookup.

vbekiaris:
An eager map store will still be triggered to load on map proxy creation: `MapProxyImpl` is an initializing object --> its `initialize` method will be triggered after the distributed object is created and wait for the map store to load: https://github.com/hazelcast/hazelcast/blob/ad54f033031cda63990f6af0c2cd7e8dfd1a28ef/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java#L222-L222

ahmetmircik:
Seems [TransactionalMapProxySupport](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java#L75) also needs this fix.

vbekiaris:
@ahmetmircik good finding! I created #10254 for that so we can keep this fix small & backportable and fix the `TransactionalMapProxy` which has an [explicit dependency on `MapContainer`](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/tx/TransactionalMapProxySupport.java#L73) in a separate PR.

Issue #10204, opened 2017-04-02
sustained load of large async backup crashes system


pveentjer:
When doing a simulator test with async backups and puts with values between 1 and 10kb, performance collapses after a few minutes. Probably due to accumulation of backup operations.

```
INFO  16:58:54 LongStringMapTest Running 00d 00h 00m 10s (  1.67%)
INFO  16:59:04 LongStringMapTest Running 00d 00h 00m 20s (  3.33%)     4,973,030 ops   496,845.45 ops/s        821 µs (avg)     77,135 µs (99.9th)    233,963 µs (max)
INFO  16:59:14 LongStringMapTest Running 00d 00h 00m 30s (  5.00%)     9,422,067 ops   444,591.76 ops/s        960 µs (avg)    149,291 µs (99.9th)    387,448 µs (max)
INFO  16:59:24 LongStringMapTest Running 00d 00h 00m 40s (  6.67%)    14,100,793 ops   467,565.95 ops/s        897 µs (avg)     10,108 µs (99.9th)    673,710 µs (max)
INFO  16:59:34 LongStringMapTest Running 00d 00h 00m 50s (  8.33%)    18,692,760 ops   458,873.50 ops/s        970 µs (avg)    164,364 µs (99.9th)    912,785 µs (max)
INFO  16:59:44 LongStringMapTest Running 00d 00h 01m 00s ( 10.00%)    23,331,498 ops   463,576.99 ops/s        907 µs (avg)     10,076 µs (99.9th)    965,214 µs (max)
INFO  16:59:54 LongStringMapTest Running 00d 00h 01m 10s ( 11.67%)    27,755,063 ops   442,052.53 ops/s      1,014 µs (avg)    171,704 µs (99.9th)  1,303,379 µs (max)
INFO  17:00:04 LongStringMapTest Running 00d 00h 01m 20s ( 13.33%)    32,239,548 ops   448,107.84 ops/s      1,081 µs (avg)    174,194 µs (99.9th)  1,474,297 µs (max)
INFO  17:00:14 LongStringMapTest Running 00d 00h 01m 30s ( 15.00%)    36,026,950 ops   378,463.23 ops/s      1,263 µs (avg)    200,933 µs (99.9th)  1,701,838 µs (max)
[...]

PR #10231, opened 2017-04-06
Fixes Async backup overload

pveentjer:
The backpressure mechanism for async backups has been modified. Instead of having a single counter per partition, there is a single global counter that triggers periodic syncs. The problem with the counter per partition, is that it is still possible to get a huge number of backup operations using the default configuration. By default 271 partitions, 100 async per partition, giving 27,100 pending backups.

With the new approach this is simplified to at most 100 async backups before a sync is given.

Also a logging warning is given to increase the backup timeout if back pressure is enabled and a low (default) timeout value is given.

fixes https://github.com/hazelcast/hazelcast/issues/10204

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
@@ -19,19 +19,24 @@ package com.hazelcast.spi.impl.operationservice.impl;
 import com.hazelcast.internal.util.ThreadLocalRandomProvider;
 import com.hazelcast.logging.ILogger;
 import com.hazelcast.spi.BackupAwareOperation;
-import com.hazelcast.spi.Operation;
-import com.hazelcast.spi.properties.GroupProperty;
+import com.hazelcast.spi.UrgentSystemOperation;
+import com.hazelcast.spi.impl.operationservice.impl.CallIdSequence.CallIdSequenceWithBackpressure;
+import com.hazelcast.spi.impl.operationservice.impl.CallIdSequence.CallIdSequenceWithoutBackpressure;
 import com.hazelcast.spi.properties.HazelcastProperties;
 
 import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
 
-import static com.hazelcast.nio.Bits.CACHE_LINE_LENGTH;
-import static com.hazelcast.nio.Bits.INT_SIZE_IN_BYTES;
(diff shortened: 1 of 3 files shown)


pveentjer:
run-lab-run

Donnerbart:
Whatever happens to this `WIP` should not be the final commit message ;)

pveentjer:
I'll fix it :)

jerrinot:
can you print the warning only if the backup timeout is breached?

pveentjer:
What do you mean? Currently the warning is given when the backup timeout is small. Isn't that the same?

pveentjer:
As part of a different PR we should make it less painful to increase the backup timeout. So make it quite long by default... but since there are no heartbeats; perhaps an idea that if a system is uncertain about the state of a backup (e.g. member loss) we just complete the invocation if the primary already has ack'ed

pveentjer:
@jerrinot @Donnerbart can you have another look?

hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
```diff
@@ -536,6 +536,9 @@
      * Back pressure is implemented by making asynchronous backups operations synchronous. This prevents the internal queues from
      * overflowing because the invoker will wait for the primary and for the backups to complete. The frequency of this is
      * determined by the sync-window.
+     *
+     * To deal with overloads of backups, the property 'hazelcast.operation.backup.timeout.millis' should be set to a larger
+     * value; above 60000 is recommended. Otherwise it can still happen backups accumulate.
```
Donnerbart:
```
     * To deal with overloads of backups the property {@link #OPERATION_BACKUP_TIMEOUT_MILLIS} should be set to a larger value;
     * above {@code 60000} is recommended. Otherwise it can still happen that backups accumulate.
```

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/BackpressureRegulator.java
```diff
@@ -81,27 +79,34 @@
...
-
         if (enabled) {
             logger.info("Backpressure is enabled"
                     + ", maxConcurrentInvocations:" + maxConcurrentInvocations
                     + ", syncWindow: " + syncWindow);
+
+            int backupTimeoutMillis = properties.getInteger(OPERATION_BACKUP_TIMEOUT_MILLIS);
+            if (backupTimeoutMillis < MINUTES.toMillis(1)) {
+                logger.warning(
+                        format("Back pressure is enabled, but '%s' is too small. ",
```
pveentjer: Good idea

PR #10232, opened 2017-04-06
Fixed wrong RAW_ESTIMATE number in DenseHyperLogLogEncoder

Donnerbart:
I just found the reason for the super spurious failures of the DenseHyperLogLog tests.
I had a suspicion and just compared the arrays with an older version:

```java
public class OldDenseHyperLogLogEncoderTest {

    private static final double DELTA = 0.0000000000001;

    @Test
    public void compareDenseHyperLogLogEncoderArraysWithOldRevision() {
        assertArrayEquals("THRESHOLD",
            OldDenseHyperLogLogEncoder.THRESHOLD,
            DenseHyperLogLogEncoder.THRESHOLD);
        for (int i = 0; i < OldDenseHyperLogLogEncoder.RAW_ESTIMATE_DATA.length; i++) {
            assertArrayEquals("RAW_ESTIMATE_DATA at index " + i,
                    OldDenseHyperLogLogEncoder.RAW_ESTIMATE_DATA[i],
                    DenseHyperLogLogEncoder.RAW_ESTIMATE_DATA[i],
                    DELTA);
        }
        for (int i = 0; i < OldDenseHyperLogLogEncoder.BIAS_DATA.length; i++) {
            assertArrayEquals("BIAS_DATA at index " + i,
                    OldDenseHyperLogLogEncoder.BIAS_DATA[i],
                    DenseHyperLogLogEncoder.BIAS_DATA[i],
                    DELTA);
        }
    }
}
```

```java
RAW_ESTIMATE_DATA at index 14: arrays first differed at element [162]; 
Expected :1062476.184
Actual   :62476.184
```

When I fix the number, the test passes. So it was me who accidetally deleted that digit during a reformat. I used IDEA for the reformat, so there should be no human error. But I must have pressed [DEL] at the wrong time. And since the feature and tests are non-deterministic, it slipped through.

Long story short, there is a `1` missing, and we should re-add it to the array to make our users happy. Sorry for that :(


tkountis:
Bloody hell :) - Nice catch!! :+1:

PR #10239, opened 2017-04-06
Pulled out DenseHyperLogLogConstants and added hashcode test

Donnerbart:
As discussed after our standup this PR pulls out the precious magic constants in a separate class. It also adds a test which checks the hashcodes of the constants, so it's harder to introduce accidental modifications (have to change two classes at the same time). Also changes in the hashcodes are easier to review and compare.

@tkountis Would you be so kind to calculate the hashcodes from your original source, so we can be sure there are no other copy & paste errors?


Donnerbart:
run-lab-run

PR #10246, opened 2017-04-06
Add projection and filtering support for map iterator

mmedenjak:
The added projection and filtering support reuses the query engine but
allows running the query on chunk within a single partition. This
allows iteration with predicates and projections. Currently iterating
the indexes is not supported, the predicate will be run on each entry.
Also, the projection does not support the Offloadable which means that
it is invoked on the partition thread.

Changed the type of the entry which is provided to the predicate from
CachedQueryEntry to LazyMapEntry to allow the identity projection -
previously it failed when being serialized since CachedQueryEntry is
not serializable.

Also added some more generics and javadocs.

EE part : https://github.com/hazelcast/hazelcast-enterprise/pull/1398
Client protocol : https://github.com/hazelcast/hazelcast-client-protocol/pull/66

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/AbstractMapQueryPartitionIteratorTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/AbstractMapQueryPartitionIteratorTest.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.client.map;
+
+import com.hazelcast.client.proxy.ClientMapProxy;
+import com.hazelcast.client.test.TestHazelcastFactory;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.projection.Projection;
+import com.hazelcast.query.Predicate;
+import com.hazelcast.query.TruePredicate;
+import com.hazelcast.test.HazelcastTestSupport;
+import org.junit.After;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.NoSuchElementException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public abstract class AbstractMapQueryPartitionIteratorTest extends HazelcastTestSupport {
+
+    protected TestHazelcastFactory factory;
+    protected HazelcastInstance server;
+    protected HazelcastInstance client;
+
+    @After
+    public void teardown() {
+        factory.terminateAll();
+    }
+
+
+    @Test(expected = NoSuchElementException.class)
+    public void test_next_Throws_Exception_On_EmptyPartition() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        proxy.iterator(10, 1,
+                new TestProjection(), TruePredicate.<String, String>truePredicate()).next();
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void test_null_projection_throws_exception() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        proxy.iterator(10, 1, null, TruePredicate.<String, String>truePredicate());
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void test_null_predicate_throws_exception() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        proxy.iterator(10, 1, new TestProjection(), null);
+    }
+
+    @Test
+    public void test_HasNext_Returns_False_On_EmptyPartition() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        final Iterator<String> iterator = proxy.iterator(10, 1,
+                new TestProjection(), TruePredicate.<String, String>truePredicate());
+        assertFalse(iterator.hasNext());
+    }
+
+    @Test
+    public void test_Next_Returns_Value_On_NonEmptyPartition() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        String key = generateKeyForPartition(server, 1);
+        String value = randomString();
+        proxy.put(key, value);
+
+        final Iterator<String> iterator = proxy.iterator(10, 1,
+                new GetValueProjection<String>(), TruePredicate.<String, String>truePredicate());
+        final String next = iterator.next();
+        assertEquals(value, next);
+    }
+
+    @Test
+    public void test_Next_Returns_Value_On_NonEmptyPartition_and_HasNext_Returns_False_when_Item_Consumed() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        String key = generateKeyForPartition(server, 1);
+        String value = randomString();
+        proxy.put(key, value);
+
+        final Iterator<String> iterator = proxy.iterator(10, 1,
+                new GetValueProjection<String>(), TruePredicate.<String, String>truePredicate());
+        final String next = iterator.next();
+        assertEquals(value, next);
+        boolean hasNext = iterator.hasNext();
+        assertFalse(hasNext);
+    }
+
+    @Test
+    public void test_HasNext_Returns_True_On_NonEmptyPartition() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        final String key = generateKeyForPartition(server, 1);
+        final String value = randomString();
+        proxy.put(key, value);
+
+        final Iterator<String> iterator = proxy.iterator(10, 1,
+                new TestProjection(), TruePredicate.<String, String>truePredicate());
+        assertTrue(iterator.hasNext());
+    }
+
+
+    @Test
+    public void test_with_projection_and_true_predicate() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        for (int i = 0; i < 100; i++) {
+            String key = generateKeyForPartition(server, 1);
+            proxy.put(key, randomString());
+        }
+        final Iterator<String> iterator = proxy.iterator(10, 1,
+                new TestProjection(), TruePredicate.<String, String>truePredicate());
+
+        final ArrayList<String> projected = collectAll(iterator);
+
+        final Collection<String> actualValues = proxy.values();
+        assertEquals(actualValues.size(), projected.size());
+
+        for (String value : actualValues) {
+            assertTrue(projected.contains("dummy" + value));
+        }
+    }
+
+    @Test
+    public void test_with_projection_and_predicate() throws Exception {
+        final ClientMapProxy<String, Integer> intMap = getMapProxy();
+
+        for (int i = 0; i < 100; i++) {
+            String key = generateKeyForPartition(server, 1);
+            intMap.put(key, i);
+        }
+
+        final Iterator<Map.Entry<String, Integer>> iterator = intMap.iterator(10, 1,
+                new IdentityProjection<Entry<String, Integer>>(),
+                new EvenPredicate());
+
+        final ArrayList<Map.Entry<String, Integer>> projected = collectAll(iterator);
+        for (Map.Entry<String, Integer> i : projected) {
+            assertTrue(i.getValue() % 2 == 0);
+        }
+
+
+        final Collection<Integer> actualValues = intMap.values();
+        assertEquals(actualValues.size() / 2, projected.size());
+
+
+        for (Entry<String, Integer> e : intMap.entrySet()) {
+            if (e.getValue() % 2 == 0) {
+                assertTrue(projected.contains(e));
+            }
+        }
+    }
+
+    private <T> ArrayList<T> collectAll(Iterator<T> iterator) {
+        final ArrayList<T> projected = new ArrayList<T>();
+        while (iterator.hasNext()) {
+            projected.add(iterator.next());
+        }
+        return projected;
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void test_remove_Throws_Exception() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
+        final Iterator<String> iterator = proxy.iterator(10, 1,
+                new TestProjection(), TruePredicate.<String, String>truePredicate());
+
+        iterator.remove();
+    }
+
+    @Test
+    public void test_Next_Returns_Values_When_FetchSizeExceeds_On_NonEmptyPartition() throws Exception {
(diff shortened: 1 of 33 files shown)


mmedenjak:
Needs client protocol released to turn green.

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanRunner.java
```diff
@@ -89,7 +93,7 @@ public PartitionScanRunner(MapServiceContext mapServiceContext) {
                 continue;
             }
             //we want to always use CachedQueryEntry as these are short-living objects anyway
-            QueryableEntry queryEntry = new CachedQueryEntry(serializationService, key, value, extractors);
+            QueryableEntry queryEntry = new LazyMapEntry(key, value, serializationService, extractors);
```
mmedenjak: This is so that we can use the identity projection with the query engine since `CachedQueryEntry` is not serializable.

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanRunner.java
```diff
@@ -98,6 +103,49 @@ public PartitionScanRunner(MapServiceContext mapServiceContext) {
         return getSortedSubList(resultList, pagingPredicate, nearestAnchorEntry);
     }
 
+    /**
+     * Executes the predicate on a partition chunk. The offset in the partition is defined by the {@code tableIndex}
+     * and the soft limit is defined by the {@code fetchSize}. The method returns the matched entries and an
+     * index from which new entries can be fetched which allows for efficient iteration of query results.
+     * <p>
+     * <b>NOTE</b>
+     * The reading should be done when the {@link IMap} is static. When reading from a Map, with or without
```
tombujok: Nitpicking: `Static` is a quite misleading word. I would add (so `no writes are allowed`). Also I don't quite get the rest of the javadoc, I mean I don't know what it is trying to say.

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanExecutor.java
```diff
@@ -28,4 +30,25 @@
 public interface PartitionScanExecutor {
 
     Collection<QueryableEntry> execute(String mapName, Predicate predicate, Collection<Integer> partitions);
+
```
tombujok: This could also return QueryableEntries

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanExecutor.java
```diff
@@ -28,4 +30,25 @@
...
+     *
+     * @param mapName     the map name
+     * @param predicate   the predicate which the entries must match
+     * @param partitionId the partition which is queried
+     * @param tableIndex  the index from which entries are queried
+     * @param fetchSize   the soft limit for the number of entries to fetch
+     * @return entries matching the predicate and a table index from which new entries can be fetched
+     */
+    QueryableEntriesWithCursor executeOnPartitionChunk(String mapName, Predicate predicate, int partitionId, int tableIndex,
+                                                       int fetchSize);
```
tombujok: Then this would be called 
`QueryableEntriesWithCursor execute(String mapName, Predicate predicate, int partitionId, int tableIndex, int fetchSize);`

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
```diff
@@ -68,6 +75,38 @@ public QueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer
...
+     * and the soft limit {@code fetchSize}.
+     *
+     * @param query       the query
+     * @param partitionId the partition which is queried
+     * @param tableIndex  the index at which to start querying
+     * @param fetchSize   the soft limit for the number of items to be queried
+     * @return the queryied entries along with the next {@code tableIndex} to resume querying
+     */
+    public MapEntriesWithCursor runPartitionScanQueryOnPartitionChunk(Query query, int partitionId, int tableIndex,
+                                                                      int fetchSize) {
```
tombujok: This method could just return `Result`, that could incorporate lastTableIndex. It's just a list of Entries and lastTableIndex. 
Then, the `MapEntriesWithCursor` could be populated on the receiving side, and we would have a clean interface where the QueryRunner returns instances of`Result` only, and where the projection is run in a single spot too.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
```diff
@@ -68,6 +75,38 @@ public QueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer
...
+        final Collection<QueryableEntry> entries = entriesWithCursor.getEntries();
+
+        updateStatistics(mapContainer);
+
+        final Projection projection = query.getProjection();
+        final ArrayList<Entry<Data, Data>> projected = new ArrayList<Entry<Data, Data>>(entries.size());
+
+        for (QueryableEntry entry : entries) {
+            final Data value = serializationService.toData(projection.transform(entry));
+            projected.add(new SimpleEntry<Data, Data>(entry.getKeyData(), value));
```
tombujok: If we used QueryResult as the result of the call, it would automatically apply the Projection, so there would be no need to have a couple of places where the projection is run.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
```diff
@@ -74,8 +74,33 @@
...
+     * from the last returned item by receiving the {@code tableIndex} of the last item. The index can represent a bucket
+     * with multiple items and in this case the returned object will contain all items in that bucket, regardless if we exceed
+     * the requested {@code size}.
+     *
+     * @param tableIndex           the index (position) from which to resume
+     * @param size                 the minimal count of returned items
+     * @param serializationService the serialization service
+     * @return fetched entries and the table index for entries AFTER the last returned entry
+     */
     MapEntriesWithCursor fetchEntries(int tableIndex, int size, SerializationService serializationService);
```
tombujok: I know it's not you who added the SerializationService to the signature, but it's such a horrid API.
The StorageImpl uses StorageSCHM which actually has a reference to the SerializationService, so maybe we could fix that.

hazelcast/src/test/java/com/hazelcast/map/MapPartitionIteratorWithQueryTest.java
```diff
@@ -0,0 +1,240 @@
...
+        final Collection<Integer> actualValues = intMap.values();
+        assertEquals(actualValues.size() / 2, projected.size());
+
+
+        for (Entry<String, Integer> e : intMap.entrySet()) {
+            if (e.getValue() % 2 == 0) {
+                assertTrue(projected.contains(e));
+            }
+        }
+    }
```
tombujok: Nitpicking: in such multiline tests it's quite hard to distinguish, where's the division between preparing input data, running the business logic, and asserting.

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanRunner.java
mmedenjak: Right, will try and make it more clearer.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
mmedenjak: I tried that but I reverted it to this since I wanted the predicate and projection to run (meaning the query needs to be `IterationType#VALUE`) but it to return entries because the iterator needs both values to return to the user and keys for `AbstractMapPartitionIterator#remove` (meaning the query would then be `IterationType#ENTRY`). 
It's a bit of a mess so I chose this since it didn't involve too much duplication. I would prefer to avoid it so if you have a suggestion please say so.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
mmedenjak: Got it, will try.

hazelcast/src/test/java/com/hazelcast/map/MapPartitionIteratorWithQueryTest.java
mmedenjak: Got it, will try and rearrange the code a bit.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
mmedenjak: I'll think about it. We can first try and solve this issue with projections and predicates not working when you have a query iteration type of `IterationType#ENTRY` - see below.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
tombujok: Hmm, I looked at it, but you should be fine specifying ENTRY, it's used only in the population of the QueryResult. Right now, the IterationType is not even passed to `public MapEntriesWithCursor runPartitionScanQueryOnPartitionChunk(Query query, int partitionId, int tableIndex,`, so in order to return the Keys and Values in the Cursor, just specify ENTRY and the QueryResult will take care of it. Am I missing sth?

mmedenjak: The problem with ENTRY was that the `QueryResult` didn't run a predicate and projection on that iteration type : https://github.com/hazelcast/hazelcast/blob/77ee1969666d385b1bcbc6a6294911f398b7e9fd/hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryResult.java#L112
And I didn't want to change the existing code just to be safe.
If it is ok we can add projections and predicates on ENTRY as well.

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanRunner.java
mmedenjak: This has been updated, please check.

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanRunner.java
```diff
@@ -98,6 +103,49 @@ public PartitionScanRunner(MapServiceContext mapServiceContext) {
...
+     *
+     * @param mapName     the map name
+     * @param predicate   the predicate which the entries must match
+     * @param partitionId the partition which is queried
+     * @param tableIndex  the index from which entries are queried
+     * @param fetchSize   the soft limit for the number of entries to fetch
+     * @return entries matching the predicate and a table index from which new entries can be fetched
+     */
+    public QueryableEntriesWithCursor runOnPartitionChunk(String mapName, Predicate predicate, int partitionId,
+                                                          int tableIndex, int fetchSize) {
```
mmedenjak: Now it's called `run` and it returns `QueryableEntriesSegment`. I'm open to suggestions for a different naming. The existing method return type was left unchanged. I can make a new PR to refactor them to the same return type.

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanExecutor.java
mmedenjak: Now it's `QueryableEntriesSegment execute()`

mmedenjak: Will make the refactoring in a new PR

hazelcast/src/main/java/com/hazelcast/map/impl/query/PartitionScanExecutor.java
mmedenjak: This is now called `QueryableEntriesSegment execute`

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
mmedenjak: Since now we've decided that the iterator doesn't support `remove`, I've now reused the `ResultProcessor` and the method returns `ResultSegment`.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
mmedenjak: Updated, it now uses `ResultProcessor` so it fits more with the query engine.

hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/Storage.java
mmedenjak: I've decided to make this a new PR since I've been having issues running tests. Seems like `hazelcast-client-protocol` depends on `hazelcast` which it downloads from `master` (where the method has `SerializationService`) and then I start getting `AbstractMethodError` when running tests in `hazelcast`.

hazelcast/src/test/java/com/hazelcast/map/MapPartitionIteratorWithQueryTest.java
mmedenjak: Done some improvements, please check

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxyImpl.java
```diff
@@ -880,10 +882,68 @@ protected Object invoke(Operation operation, int partitionId) throws Throwable {
...
+     * @param partitionId the partition ID which is being iterated
+     * @param projection  the projection to apply before returning the value. {@code null} value is not allowed
+     * @param predicate   the predicate which the entries must match. {@code null} value is not allowed
+     * @param <R>         the return type
+     * @return the iterator for the projected entries
+     * @throws UnsupportedOperationException when cluster version is less than {@link Versions#V3_9}
+     * @since 3.9
+     */
+    public <R> Iterator<R> iterator(int fetchSize, int partitionId, Projection<Map.Entry<K, V>, R> projection,
+                                    Predicate<K, V> predicate) {
```
ahmetmircik: this supports paging predicate? if not, better to throw exception for such case.

mmedenjak: Updated.

hazelcast-client/src/test/java/com/hazelcast/client/map/AbstractMapQueryPartitionIteratorTest.java
```diff
@@ -0,0 +1,235 @@
...
+
+    @After
+    public void teardown() {
+        factory.terminateAll();
+    }
+
+
+    @Test(expected = NoSuchElementException.class)
+    public void test_next_Throws_Exception_On_EmptyPartition() throws Exception {
+        final ClientMapProxy<String, String> proxy = getMapProxy();
```
mmedenjak: I like `final` and I add them to my code - so I should add them where missing. Shall we debate on this? :)

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapFetchWithQueryMessageTask.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public String getServiceName() {
+        return MapService.SERVICE_NAME;
+    }
+
+    @Override
+    public Permission getRequiredPermission() {
+        return null;
```
mmedenjak: Good point

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/MapFetchWithQueryMessageTask.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public String getMethodName() {
+        return "iterator";
+    }
+
+    @Override
+    public Object[] getParameters() {
+        return new Object[0];
```
mmedenjak: Got it

Issue #10252, opened 2017-04-07
Client andThen bug


pveentjer:
If invocation future for the client has completed, and the user calls 'andThen' then the callable is executed on the calling thread. The whole point of the andThen is to not obstruct the calling thread.

I can imagine that under load we would start to run the callback on the calling thread to provide some form of back pressure, but even if there is no load, the task will still be executed on the calling thread. Meaning that the andThen becomes very unreliable for latency sensitive calls that do not want their thread to be obstructed if there is no high load.

sancar:
fixed by https://github.com/hazelcast/hazelcast/pull/10270

Issue #10253, opened 2017-04-07
Client backpressure bug


pveentjer:
There is a race condition here because the check if someting is complete and making the InternalDelegatingExecutionCallback isn't atomic.

So it could happen a single invocation future leads to multiple callidsequence completion. And the consequence is that the back pressure doesn't work for the client since you can get an increasing number of invocations.

```
  @Override
    public void andThen(ExecutionCallback<ClientMessage> callback) {
        isNotNull(callback, "callback");

        if (completeCount.get() == 0) {
            try {
                callback.onResponse(get());
            } catch (Exception e) {
                callback.onFailure(peel(e));
            }
            return;
        }
        super.andThen(new InternalDelegatingExecutionCallback(callback));
    }
```

I made some modifications to demonstrate the behavior:

```
public class ClientInvocationFuture extends AbstractInvocationFuture<ClientMessage> {

    private final AtomicIntegerFieldUpdater COMPLETED_COUNT = newUpdater(ClientInvocationFuture.class, "completeCount");

    private final ClientMessage request;
    private final ClientInvocation invocation;
    private final CallIdSequence callIdSequence;

    public final static ConcurrentHashMap<Long,AtomicLong> invocations = new ConcurrentHashMap<Long, AtomicLong>();

    private volatile int completeCount = 1;

    public ClientInvocationFuture(ClientInvocation invocation, Executor internalExecutor,
                                  ClientMessage request, ILogger logger,
                                  CallIdSequence callIdSequence) {
        super(internalExecutor, logger);
        this.request = request;
        this.invocation = invocation;
        this.callIdSequence = callIdSequence;
    }

    @Override
    protected String invocationToString() {
        return request.toString();
    }

    @Override
    protected void onInterruptDetected() {
        complete(new InterruptedException());
    }

    @Override
    protected TimeoutException newTimeoutException(long timeout, TimeUnit unit) {
        return new TimeoutException();
    }

    @Override
    protected Throwable unwrap(Throwable throwable) {
        return throwable;
    }

    @Override
    protected Object resolve(Object value) {
        if (value instanceof Throwable) {
            return new ExecutionException((Throwable) value);
        }
        return value;
    }

    @Override
    public void andThen(ExecutionCallback<ClientMessage> callback) {
        isNotNull(callback, "callback");
        System.out.println(Thread.currentThread().getName()+" getting paused: "+completeCount);

        if (completeCount == 0) {
            try {
                callback.onResponse(get());
[...]

PR #10256, opened 2017-04-07
Fixes client invocationfuture callback backpressure

pveentjer:
The idea is that for every callback, an invocationslot is taken.
And for every completion of a callback, an invocation slot is released.

So if there are many pending invocations, there will be few invocation
slots. Which will reduce the pressure, and give the callbacks time to
complete instead of executor getting overloaded.

fixes
https://github.com/hazelcast/hazelcast/issues/10252
https://github.com/hazelcast/hazelcast/issues/10253

When callbacks are used there will be more contention on the head/tail of the CallIdSequence. I don't believe this will be real concern since remoting will normally dominate performance. In the future we can always add 2 sequences; 1 for the inflight invocations, and 1 for the pending callbacks.


sancar:
Merging this since 3.8.1 backport is already reviewed and merged.

PR #10257, opened 2017-04-07
Fixes back pressure bug in client invocation in case of retrying

pveentjer:


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -115,11 +116,10 @@ public ClientInvocationFuture invoke() {
         assert (clientMessage != null);
 
         try {
-            if (urgent) {
-                clientMessage.setCorrelationId(callIdSequence.renew());
-            } else {
-                clientMessage.setCorrelationId(callIdSequence.next());
+            if (callId == 0) {
```
sancar: This will cause problems because a client should not reuse correlation id according to design.

sancar: Sorry my mistake. I misunderstood.

PR #10258, opened 2017-04-07
Enabled optimized selector by default

pveentjer:
Various other projects like Netty, Agrona do this be default as well.

There is a switch to disable it, and if something fails while trying
to modify, it will fallback on the non optimized version anyway.

The optimization is that no litter is created for the sake of selection keys.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
@@ -127,7 +127,8 @@ public class NonBlockingIOThread extends Thread implements OperationHostileThrea
     private static Selector newSelector(ILogger logger) {
         try {
             Selector selector = Selector.open();
-            if (Boolean.getBoolean("tcp.optimizedselector")) {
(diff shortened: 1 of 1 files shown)


Donnerbart:
I don't get the optimization. Is it that you activate the `optimzize` by default? That's okay to me. I'm just confused why you don't use `Boolean.getBoolean()` anymore. This doesn't look like an improvement to me. You just inlined JDK code which looks litter-free to me.

Oracle 6 and 7:
```
    public static boolean getBoolean(String name) {
        boolean result = false;
        try {
            result = toBoolean(System.getProperty(name));
        } catch (IllegalArgumentException e) {
        } catch (NullPointerException e) {
        }
        return result;
    }

    private static boolean toBoolean(String name) { 
	return ((name != null) && name.equalsIgnoreCase("true"));
    }
```
Oracle 8:
```
    public static boolean getBoolean(String name) {
        boolean result = false;
        try {
            result = parseBoolean(System.getProperty(name));
        } catch (IllegalArgumentException | NullPointerException e) {
        }
        return result;
    }
```

Donnerbart:
Ah, darn, it's because you cannot parse a default value with `Boolean.getBoolean()`, isn't it?

pveentjer:
Exactly!!!!!!

PR #10263, opened 2017-04-07
Fixes back pressure bug in client invocation in case of retrying 

sancar:
This pr includes fix here #10257
and a cleanup to remove if statements and fields with preserving behaviour
Also added backpressure tests provided by @jerrinot


pveentjer:
:+1:

PR #10264, opened 2017-04-08
Tiny cleanup OperationRunnerImpl.handleResponse

pveentjer:


pveentjer:
run-lab-run

pveentjer:
run-lab-run

Donnerbart:
What is with operations which have backups, but no response? Those backups won't be made after this change, won't they? And for me it's not obvious that this is an "illegal" combination.

A little explanation why this is a valid optimization would be nice :)

pveentjer:
Good spotting. Apparently I had sh*t in my eyes. I read is at the number of backups.. so a simple 'getter'; but a lot more logic is going on

Issue #10205, opened 2017-04-03
Response thread should not remove response envelope


pveentjer:
The response thread currently removes the response envelope from a response packet.

```
public final class InboundResponseHandler implements PacketHandler, MetricsProvider {

    @Override
    public void handle(Packet packet) throws Exception {
        Response response = serializationService.toObject(packet);<----
```

This response contains response Object or Data which then needs to get deserialized by the end user thread. 

There are 2 problems:
1: the larger the response gets, the more copying the response threads needs to do and the response thread is a serial part in the system. So it will become a bottleneck. 

2: additional litter is being created. In case of a Data in the response, this Data is backed up by an on heap byte-array. So for every 10kb response you receive, first the io system creates a 10kb Packet byte-array, and then response thread creates an additional 10kb Data byte-array. This second 10kb byte-array isn't needed and will only increase pressure on the gc needlessly. 

Solution:
Instead of needing to deserialize the response packet into a Response object, we can read the data from the response-packet byte-array directly (e.g. call id). And once we find the invocation, we hand over the whole response-packet to the calling thread, and it will deserialize beginning from a certain offset (we don't care for the Response envelope; only about the payload.

Setting the packet InvocationFuture makes it very easy for the InvocationFuture to understand it needs skip the header/response envelope part of the packet and it will not get confused when dealing with a regular byte-array

![response](https://cloud.githubusercontent.com/assets/105243/24594130/b32e91ae-1832-11e7-806d-2e919767cec9.png)

The above screenshot is from a test with 8 clients, and 1..10kb payloads, 2 members and the clients being lite members. As can be seen, the response thread is taking a significant amount of time.

mmedenjak:
@pveentjer can this be closed now?

PR #10269, opened 2017-04-08
Moved deserialization of response from response thread

pveentjer:
Before this change, when a response-packet is received, it is first deserialized
to a (Normal)Response on the response thread. And then handed over to
the calling thread to do the real deserialization if needed. The problem
is that this puts a lot of pressure on the response thread and it becomes
a scalability bottleneck; especially with larger responses since there is 
only a single response thread.

This pr changes this behavior so that instead of deserializing the response,
we poke in the bytes to get call id/type of response without needing to 
deserialize and handover the Packet to the invocation future which will 
do these 2 consequtive deserializations
packet -> NormalResponse -> payload.

In case of backup ack response and call timeout, no response object is created at all. 
So we already have some litter reduction here.

In a follow up PR, this intermediate deserialization of NormalResponse
will be fully removed. Removing more unwanted litter.

See https://github.com/hazelcast/hazelcast/issues/10205


pveentjer:
Good idea. There is more work in the pipeline for this section. 

https://github.com/hazelcast/hazelcast/pull/10274

And 

https://github.com/hazelcast/hazelcast/issues/10285

PR #10271, opened 2017-04-08
Fixes client invocationfuture callback backpressure

### LINKED ISSUE #10252: Client andThen bug

sancar:
The idea is that for every callback, an invocationslot is taken.
And for every completion of a callback, an invocation slot is released.

So if there are many pending invocations, there will be few invocation
slots. Which will reduce the pressure, and give the callbacks time to
complete instead of executor getting overloaded.

fixes
#10252
#10253

When callbacks are used there will be more contention on the head/tail of the CallIdSequence. I don't believe this will be real concern since remoting will normally dominate performance. In the future we can always add 2 sequences; 1 for the inflight invocations, and 1 for the pending callbacks.

ported from https://github.com/hazelcast/hazelcast/pull/10256

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
@@ -26,10 +26,8 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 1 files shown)


sancar:
Merging this since backport is already reviewed and merged.

PR #10272, opened 2017-04-08
Litter optimization MPSCQueue

pveentjer:
In case of a single item on the queue, the item isn't wrapped in a Node
instance.

Also if there are N items on the queue, it will only take N-1 Node
instances.


pveentjer:
verify

pveentjer:
run-lab-run

mdogan:
How much gain is this change going to provide? Do you have any measurements?

Holmistr:

PR #10273, opened 2017-04-08
Deleted the AsyncAtomicReference/Long

pveentjer:
They have been deprecated long time ago; so we can delete them


pveentjer:
The code has been deprecated in 3.8. So we can remove it in 3.9 and have our 1 minor version grace period.

Donnerbart:
I know the policy. But as I said, it seems that some deprecations are kept alive until 4.0, e.g. the read-only configuration cleanup was closed, although everything was properly deprecated in 3.8.

pveentjer:
What is then the point of deprecating it :)

Donnerbart:
I'm the wrong person to ask that question.

vbekiaris:
For additional consideration: `AsyncAtomicLong` is used in `vertx-hazelcast` cluster manager as at the time [async counter support was introduced](https://github.com/vert-x3/vertx-hazelcast/commit/9f1ddde81ee8617f713b0a6912cf51aa2826fcc0), 3.7 was not yet released. I think it's quite common for vert-x users to override the Hazelcast version, so even though 3.6.3 is used in `vertx-hazelcast` pom, vert-x apps may override this with a newer version --> attempting to use a previously released`vertx-hazelcast` with 3.9 and the async option enabled will fail.
Given the proximity of 4.0, I 'd be in favor of removing `AsyncAtomicLong` in 4.0.

pveentjer:
Ok. I can live with that explanation

PR #10274, opened 2017-04-09
Invocationfuture skipping normalresponse deserialization

pveentjer:
This is the follow up for

https://github.com/hazelcast/hazelcast/pull/10269

When a NormalResponse is received, instead of deserializing the packet
into a NormalResponse and then getting the content, the NormalResponse
deserialization is completely skipped and the data accessed directly
from the packet-byte array. This removes the need to create a
NormalResponse instance.
    
If the NormalResponse contains a Data, and the caller expects the
value to be deserialized, the Data instance is skipped as well.
    
So it could be that you get either 1 or 2 instances per invocation less.


Holmistr:

PR #10279, opened 2017-04-10
Fix unwanted toByteArray to determine message size

### LINKED ISSUE #10278: Client serverside calculateDataSize  litter
To send back a response to the client, the client needs to know the size of the payload to send. For this the following method is called

```
    public static int calculateDataSize(Data response) {
            int dataSize = ClientMessage.HEADER_SIZE;
            ++dataSize;
            if(response != null) {
                dataSize += ParameterUtil.calculateDataSize(response);
            }

            return dataSize;
        }
```

Inside the ParameterUtil we find:

```
  public static int calculateDataSize(Data data) {
        return calculateDataSize(data.toByteArray());
    }
```
For onheap data this is perfectly fine since we can just return the reference to the internal byte-array.

But for offheap data, we create a byte array, fill it, determine the length of the byte array and then we toss the byte array. And then to fill the client message, the toByteArray is called again. This causes unwanted litter + unwanted data copying.

pveentjer:
I have verified that for:
- heap data the 'totalSize' is the length of the byte array.
- for offheap data the 'totalSize' is the length used to otherwise create the byte array
  for the sake of the 'toByteArray'

fix #10278

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ParameterUtil.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ParameterUtil.java
@@ -25,22 +25,28 @@ public final class ParameterUtil {
 
     static final int UTF8_MAX_BYTES_PER_CHAR = 3;
 
-    private ParameterUtil() { }
+    private ParameterUtil() {
+    }
 
     public static int calculateDataSize(String string) {
         return Bits.INT_SIZE_IN_BYTES + string.length() * UTF8_MAX_BYTES_PER_CHAR;
     }
 
     public static int calculateDataSize(Data data) {
-        return calculateDataSize(data.toByteArray());
+        return addByteArrayLengthHeader(data.totalSize());
     }
 
     public static int calculateDataSize(Map.Entry<Data, Data> entry) {
-        return calculateDataSize(entry.getKey().toByteArray()) + calculateDataSize(entry.getValue().toByteArray());
+        return addByteArrayLengthHeader(entry.getKey().totalSize())
+                + addByteArrayLengthHeader(entry.getValue().totalSize());
     }
 
     public static int calculateDataSize(byte[] bytes) {
-        return Bits.INT_SIZE_IN_BYTES + bytes.length;
+        return addByteArrayLengthHeader(bytes.length);
+    }
+
+    private static int addByteArrayLengthHeader(int length) {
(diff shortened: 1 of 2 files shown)


pveentjer:
@mdogan can you have a look at this PR.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ParameterUtil.java
```diff
@@ -32,11 +32,11 @@ public static int calculateDataSize(String string) {
     }
 
     public static int calculateDataSize(Data data) {
-        return calculateDataSize(data.toByteArray());
+        return calculateDataSize(data.totalSize());
```
sancar: I think you meant  to write :
`return data.totalSize`
right ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ParameterUtil.java
```diff
@@ -32,11 +32,11 @@ public static int calculateDataSize(String string) {
...
     }
 
     public static int calculateDataSize(Map.Entry<Data, Data> entry) {
-        return calculateDataSize(entry.getKey().toByteArray()) + calculateDataSize(entry.getValue().toByteArray());
+        return calculateDataSize(entry.getKey().totalSize()) + calculateDataSize(entry.getValue().totalSize());
```
sancar: Same typo here.
`return entry.getKey().totalSize() + entry.getValue().totalSize()`
And you need to take size field into account . You need to add Bits.INT_SIZE_IN_BYTES per data .
As we were doing in here when byte[] is used to calculate data size
```
    public static int calculateDataSize(byte[] bytes) {
        return Bits.INT_SIZE_IN_BYTES + bytes.length;
    }
```

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ParameterUtil.java
pveentjer: I made an error there. I already updated the PR.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/ParameterUtil.java
pveentjer: Good spotting. I already figured out what was going on :)

PR #10296, opened 2017-04-11
Add delta for long running durable task test

### LINKED ISSUE #10071: ScheduledExecutorServiceSlowTest.stats_longRunningTask_durable
```
java.lang.AssertionError: expected:<6> but was:<8>
	at com.hazelcast.scheduledexecutor.ScheduledExecutorServiceSlowTest.stats_longRunningTask_durable(ScheduledExecutorServiceSlowTest.java:120)
```

https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast%20Maintenance/job/Hazelcast-3.maintenance-nightly/com.hazelcast$hazelcast/1156/testReport/junit/com.hazelcast.scheduledexecutor/ScheduledExecutorServiceSlowTest/stats_longRunningTask_durable/

tkountis:
Can't see anything wrong with the test or functionality.
There is a small chance of delays + sleep time to add up to yet another schedule of the task, and cause the wrong assertion. With the delta we expect a minimum of 6 and maximum of 7.

Improved the implementation by removing the sleep and using a second latch.

Fix #10071

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceSlowTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceSlowTest.java
@@ -103,23 +103,26 @@ public class ScheduledExecutorServiceSlowTest
 
         String key = generateKeyOwnedBy(instances[1]);
 
-        ICountDownLatch latch = instances[0].getCountDownLatch("latch");
-        latch.trySetCount(6);
+        ICountDownLatch firstLatch = instances[0].getCountDownLatch("firstLatch");
+        firstLatch.trySetCount(2);
+
+        ICountDownLatch lastLatch = instances[0].getCountDownLatch("lastLatch");
+        lastLatch.trySetCount(6);
 
         IScheduledExecutorService executorService = getScheduledExecutor(instances, "s");
         IScheduledFuture future = executorService.scheduleOnKeyOwnerAtFixedRate(
-                new ICountdownLatchRunnableTask("latch"), key, 0, 10, SECONDS);
+                new ICountdownLatchRunnableTask("firstLatch", "lastLatch"), key, 0, 10, SECONDS);
 
-        Thread.sleep(12000);
+        firstLatch.await(12, SECONDS);
 
         instances[1].getLifecycleService().shutdown();
 
-        latch.await(70, SECONDS);
+        lastLatch.await(70, SECONDS);
         sleepSeconds(4); // Wait for run-cycle to finish before cancelling, in order for stats to get updated.
         future.cancel(false);
 
         ScheduledTaskStatistics stats = future.getStats();
-        assertEquals(6, stats.getTotalRuns());
+        assertEquals(6, stats.getTotalRuns(), 1);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


vbekiaris:
run-lab-run

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceSlowTest.java
```diff
@@ -103,23 +103,26 @@ public void stats_longRunningTask_durable()
...
         instances[1].getLifecycleService().shutdown();
 
-        latch.await(70, SECONDS);
+        lastLatch.await(70, SECONDS);
         sleepSeconds(4); // Wait for run-cycle to finish before cancelling, in order for stats to get updated.
         future.cancel(false);
 
         ScheduledTaskStatistics stats = future.getStats();
-        assertEquals(6, stats.getTotalRuns());
+        assertEquals(6, stats.getTotalRuns(), 1);
```
vbekiaris: Instead of adding the delta, could we perhaps remove the reliance on time completely by changing the runnable to execute exactly N `latch.countdown()`'s? I am just wondering if this delta could hide a real stats issue in the future.

tkountis: Thanks for the feedback @vbekiaris . I already cancel the task from the test, line 122. But with unfortunate timing and slow operations this is not reliable. Cancelling within the task is easier said than done, mainly because I don't have access on the future. Also gating the run with a branch wont work either because the stats get updated on a finally block on the outside `runnable` wrapper. I think without major changes, that's the best we can do, of course open to suggestions.

pveentjer: I agree with @vbekiaris ; relying on time is a royal PITA.

Which future are you referring too?

vbekiaris: @tkountis I understand you cannot eliminate completely the need for some wait for updating stats, but you can avoid the major source of uncertainty by submitting something like this task:
```
static class GatedCountdownLatchRunnable extends ICountdownLatchRunnableTask {

        final int executions;
        final String counterName;

        GatedCountdownLatchRunnable(int executions, String... runsCounterlatchNames) {
            super(runsCounterlatchNames);
            this.executions = executions;
            this.counterName = randomString();
        }

        @Override
        public void run() {
            IAtomicLong counter = instance.getAtomicLong(counterName);
            if (counter.get() < executions) {
                for (String runsCounterLatchName : runsCounterlatchNames) {
                    instance.getCountDownLatch(runsCounterLatchName).countDown();
                }
                counter.getAndIncrement();
            }
        }
    }
```

@pveentjer I was refering to the future as in wall-clock time sense, not a specific `Future` ;-)

tkountis: @vbekiaris thanks for your time on providing an approach. That was the obvious approach, but as I said above, every runnable is executed by https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledExecutorContainer.java#L375 which will wrap the run, and no matter what you do in it to prevent stats updating, it will update them through the finally block. Cancelling the Future which you nicely did, is not a deterministic action, because we can not guarantee interruption of the task, thus, the stats may or not, get updated. And in your case, where counter == executions, we do want them to get updated, so if you successfully cancel it, (again not deterministic) then the test will fail, with countdownLatch = 5. Last, keep in mind that the scheduled tasks runs _outside_ the partition threads.

tkountis:
run-lab-run

PR #10314, opened 2017-04-11
Fix typo in ReadHandler JavaDoc

tkountis:


pveentjer:
Since javadoc change, we can just merge it.

PR #10317, opened 2017-04-12
Outbound BackupAckResponse instance creation skipped

pveentjer:
So instead of creating a BackupAckResponse and recording the content in some byte-array and then making a new packet-byte array where the content gets copied to, we immediately create the packet-byte array without additional copy and skipping the BackupAckResponse.

So this optimization provides the following benefits:
- skip an intermediate data-copy
- prevents an BackupAckResponse from being created on every backup member.

The is a no difference in the data-send over the wire; it is bit for bit the same as in the old situation.

There will be a follow up PR where the NormalResponse will also be skipped. One of the big advantages of the latter optimization is that if the response value is 'Data', the content will immediately be copied into a correctly sized packet-bytearray instead of having an intermediate copy into a ByteArrayObjectDataOutput.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
@@ -25,12 +25,28 @@ import com.hazelcast.nio.ConnectionManager;
 import com.hazelcast.nio.Packet;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.OperationResponseHandler;
+import com.hazelcast.spi.impl.SpiDataSerializerHook;
 import com.hazelcast.spi.impl.operationservice.impl.responses.ErrorResponse;
 import com.hazelcast.spi.impl.operationservice.impl.responses.NormalResponse;
 import com.hazelcast.spi.impl.operationservice.impl.responses.Response;
 
+import java.nio.ByteOrder;
+
+import static com.hazelcast.internal.serialization.impl.SerializationConstants.CONSTANT_TYPE_DATA_SERIALIZABLE;
+import static com.hazelcast.nio.Bits.writeInt;
+import static com.hazelcast.nio.Bits.writeIntB;
+import static com.hazelcast.nio.Bits.writeLong;
 import static com.hazelcast.nio.Packet.FLAG_OP_RESPONSE;
 import static com.hazelcast.nio.Packet.FLAG_URGENT;
+import static com.hazelcast.nio.Packet.Type.OPERATION;
+import static com.hazelcast.spi.impl.SpiDataSerializerHook.BACKUP_ACK_RESPONSE;
+import static com.hazelcast.spi.impl.operationservice.impl.responses.BackupAckResponse.BACKUP_RESPONSE_SIZE_IN_BYTES;
+import static com.hazelcast.spi.impl.operationservice.impl.responses.Response.OFFSET_CALL_ID;
+import static com.hazelcast.spi.impl.operationservice.impl.responses.Response.OFFSET_IDENTIFIED;
+import static com.hazelcast.spi.impl.operationservice.impl.responses.Response.OFFSET_SERIALIZER_TYPE_ID;
+import static com.hazelcast.spi.impl.operationservice.impl.responses.Response.OFFSET_TYPE_FACTORY_ID;
+import static com.hazelcast.spi.impl.operationservice.impl.responses.Response.OFFSET_TYPE_ID;
+import static com.hazelcast.spi.impl.operationservice.impl.responses.Response.OFFSET_URGENT;
 import static com.hazelcast.util.Preconditions.checkNotNull;
 
 /**
@@ -42,6 +58,7 @@ public final class OutboundResponseHandler implements OperationResponseHandler {
 
     private final Address thisAddress;
     private final InternalSerializationService serializationService;
+    private final boolean useBigEndian;
     private final ILogger logger;
     // it sucks we need to pass in Node as argument; but this is due to the ConnectionManager which is created after
     // the OperationService is created.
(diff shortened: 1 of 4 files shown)


pveentjer:
run-lab-run

pveentjer:
run-lab-run

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
```diff
@@ -98,4 +114,39 @@ public boolean send(Response response, Address target) {
...
+        // identified
+        bytes[OFFSET_IDENTIFIED] = 1;
+        // factory id
+        writeInt(bytes, OFFSET_TYPE_FACTORY_ID, SpiDataSerializerHook.F_ID, useBigEndian);
+        // type id
+        writeInt(bytes, OFFSET_TYPE_ID, BACKUP_ACK_RESPONSE, useBigEndian);
+        //call id
+        writeLong(bytes, OFFSET_CALL_ID, callId, useBigEndian);
+
+        Packet packet = new Packet(bytes, -1)
```
Donnerbart: Maybe you can pull out this part as `createPacket(bytes, urgent)` so there is no code duplication in this class.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
```diff
@@ -98,4 +114,39 @@ public boolean send(Response response, Address target) {
...
+
+        if (thisAddress.equals(target)) {
+            throw new IllegalArgumentException("Target is this node! -> " + target);
+        }
+
+        byte[] bytes = new byte[BACKUP_RESPONSE_SIZE_IN_BYTES];
+
+        // the first 4 bytes contain the partition hash which are not set.
+
+        //serializer type id
```
Donnerbart: Missing space here and in line 137 `//call id`

pveentjer:
```
Maybe you can make a static methods in DataSerializableSerializer to have the shared / common IDS serialization code in one place (partition hash, serializer type id, factory id, class id etc.).
```

Partition hash is just something the serialization service does for the sake of a packet.

Donnerbart:
There is more than the partition hash. Just saying, if you do this kind of optimization in several places it would be good to have a single method to set the basic bytes (and maybe create the byte array).
```java
public static bytes[] createBytesIDS(int length, int factoryId, int classId, int typeId);
```
The offsets of those fixed fields could also be in a serializer class once, not spread into the classes you want to serialize. Then you just need to define the additional offsets for the class fields and add them later.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
pveentjer: Good idea

pveentjer:
The offsets of these fields are custom for the response. They have no meaning in other parts of the system.

pveentjer:
This is the best balance of pulling out a method which can be used by the other responses

```
public void sendBackupAck(Address target, long callId, boolean urgent) {
        checkNotNull(target, "Target is required!");

        if (thisAddress.equals(target)) {
            throw new IllegalArgumentException("Target is this node! -> " + target);
        }

        byte[] bytes = new byte[BACKUP_RESPONSE_SIZE_IN_BYTES];

        // the first 4 bytes contain the partition hash which are not set.

        // serializer type id
        setIdentifiedDataSerializableBytes(bytes,OFFSET_SERIALIZER_TYPE_ID,useBigEndian);
        // call id
        writeLong(bytes, OFFSET_CALL_ID, callId, useBigEndian);

        Packet packet = newResponsePacket(bytes, urgent);

        ConnectionManager connectionManager = node.getConnectionManager();
        Connection connection = connectionManager.getOrConnect(target);
        connectionManager.transmit(packet, connection);
    }

    private void setIdentifiedDataSerializableBytes(byte[] bytes, int typeId, boolean useBigEndian) {
        writeInt(bytes, OFFSET_SERIALIZER_TYPE_ID, CONSTANT_TYPE_DATA_SERIALIZABLE, useBigEndian);
        // identified
        bytes[OFFSET_IDENTIFIED] = 1;
        // factory id
        writeInt(bytes, OFFSET_TYPE_FACTORY_ID, SpiDataSerializerHook.F_ID, useBigEndian);
        // type id
        writeInt(bytes, OFFSET_TYPE_ID, typeId, useBigEndian);
    }
```

Donnerbart:
That's exactly what I meant :) If you plan to make this more often, I think this would be the cleaner separation (including the constants for those "fixed" fields).

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
```diff
@@ -86,16 +102,51 @@ public boolean send(Response response, Address target) {
...
+        // serializer type id
+        writeInt(bytes, OFFSET_SERIALIZER_TYPE_ID, CONSTANT_TYPE_DATA_SERIALIZABLE, useBigEndian);
+        // identified
+        bytes[OFFSET_IDENTIFIED] = 1;
+        // factory id
+        writeInt(bytes, OFFSET_TYPE_FACTORY_ID, SpiDataSerializerHook.F_ID, useBigEndian);
+        // type id
+        writeInt(bytes, OFFSET_TYPE_ID, BACKUP_ACK_RESPONSE, useBigEndian);
+        // call id
+        writeLong(bytes, OFFSET_CALL_ID, callId, useBigEndian);
```
Donnerbart: You are missing to write the urgent field at `OFFSET_URGENT` here!

pveentjer: Yes. It is already part of the other pr.

pveentjer:
I already added it to the other PR. The other pr contains tests for the backup ack response and I already saw it was missing.

pveentjer:
Thanks for the review. The code quality has improved (especially with the other response added)

Donnerbart:
run-lab-run

PR #10318, opened 2017-04-12
Outbound NormalResponse optimizations

### LINKED ISSUE #10285: Sending responses unwanted copying
When a normal response is made, it first copied into the byte-array of the data-output. And then a packet is made based on a copy of the 'active' data in the data-output. This means there is an intermediate copy.

But if you have offheap data, then you know the size of the payload, so you only need to prefix it with the response data. No intermediate copy is needed. The heapdata can be directly copied into the packet byte-array. 

Probably we can also immediately get rid of the response instances.

pveentjer:
This PR contains the following optimizations

- for an operation that doesn't require backup acks (e.g. map.get or a map.put with async backup_ the NormalResponse instance is skipped (less litter)
- if the response is data, the intermediate copy of the data into a ByteArrayObjectDataOutput is skipped. A byte array with the right size to store the whole response is created and the data is copied into this  byte-array directly. So we get a packet-bytearray with a single copy of data instead of multiple copies.

This PR is native-memory friendly; it doesn't force the NativeMemoryData to create a copy onheap. 

This is a follow up PR for 
https://github.com/hazelcast/hazelcast/pull/10317

There is no change in the data that goes over the wire. So there will not be any minor version incompatibilities.

fix #10285

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
@@ -23,16 +23,18 @@ import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
 import com.hazelcast.nio.ConnectionManager;
 import com.hazelcast.nio.Packet;
+import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.OperationResponseHandler;
 import com.hazelcast.spi.impl.SpiDataSerializerHook;
+import com.hazelcast.spi.impl.operationservice.impl.responses.CallTimeoutResponse;
 import com.hazelcast.spi.impl.operationservice.impl.responses.ErrorResponse;
 import com.hazelcast.spi.impl.operationservice.impl.responses.NormalResponse;
 import com.hazelcast.spi.impl.operationservice.impl.responses.Response;
 
-import java.nio.ByteOrder;
-
 import static com.hazelcast.internal.serialization.impl.SerializationConstants.CONSTANT_TYPE_DATA_SERIALIZABLE;
+import static com.hazelcast.internal.serialization.impl.SerializationConstants.CONSTANT_TYPE_NULL;
+import static com.hazelcast.nio.Bits.INT_SIZE_IN_BYTES;
 import static com.hazelcast.nio.Bits.writeInt;
 import static com.hazelcast.nio.Bits.writeIntB;
 import static com.hazelcast.nio.Bits.writeLong;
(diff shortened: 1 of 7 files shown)


pveentjer:
run-lab-run

pveentjer:
run-lab-run

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
```diff
@@ -87,15 +119,123 @@ public boolean send(Response response, Address target) {
...
+    private boolean sendNormalResponse(Address target, long callId, int backupAcks, boolean urgent, Object value) {
+        checkTarget(target);
+
+        Packet packet = toNormalResponsePacket(callId, (byte) backupAcks, urgent, value);
+
+        return transmit(target, packet);
+    }
+
+    Packet toNormalResponsePacket(long callId, int backupAcks, boolean urgent, Object value) {
+        byte[] bytes;
```
pveentjer: They can still be called; I'm using them even to verify if the 'optimized' serialization does the same thing as serialization NormalResponse. 

Eventually I would like to drop all the responses since they are not truly needed; but I tried this a few time and it leads to a lot of code change.

```
I like the optimisation, I'm just worrying that someone forgets that we duplicate the logic from the DataSerializableSerializer and EnterpriseDataSerializableSerializer here.
```
They can't be changed.. at least not in the 3.x timeline due to minor version compatibility.

```
Another idea - we move this into a new method on the internal serialization service?
```
I don't think we should get a dependency of something as specific as response into something generic as serialization. It would cause a cycle (tight coupling).

mmedenjak: Ok, good that we have a test meaning that it will fail if someone tries to change the serialization format and forgets to change it in all places. Looking forward to moving the responses away from IDS and more internally.

Donnerbart: Can we still remove them at all? They are part of 3.8 and with rolling upgrade we still have to support them (at least one minor version). Otherwise this should go on the 4.0 list.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
```diff
@@ -87,15 +119,123 @@ public boolean send(Response response, Address target) {
...
+        // identified
+        bytes[OFFSET_IDENTIFIED] = 1;
+        // factory id
+        writeInt(bytes, OFFSET_TYPE_FACTORY_ID, SpiDataSerializerHook.F_ID, useBigEndian);
+        // type id
+        writeInt(bytes, OFFSET_TYPE_ID, BACKUP_ACK_RESPONSE, useBigEndian);
+        //call id
+        writeLong(bytes, OFFSET_CALL_ID, callId, useBigEndian);
+        // urgent
+        bytes[OFFSET_URGENT] = (byte) (urgent ? 1 : 0);
```
pveentjer: Here it is

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OutboundResponseHandler.java
pveentjer: The actual data is going to remain the same. Only we don't need to have 'physical' response classes.

pveentjer:
run-lab-run

pveentjer:
run-lab-run

pveentjer:
disk space problems.

pveentjer:
run-lab-run

pveentjer:
verify

pveentjer:
run-lab-run

PR #10320, opened 2017-04-12
Post join operations should run regardless of cluster state

mdogan:
When executed via `OperationService`, post join ops will fail if cluster
state is passive. This is not desired. So, they should be run directly.

This flaw was introduced by https://github.com/hazelcast/hazelcast/pull/10137


mdogan:
run-lab-run

Issue #10291, opened 2017-04-10
Warning level exception when shutting down the client


alparslanavci:
When shutting down a client with a topic subscription, it logs a warning level excepiton which seems harmless. But it is not good to log a harmless exception as warning. See the stacktrace below: 

```
2017-04-06 16:15:45,715 WARN [main] com.hazelcast.client.proxy.ClientReliableTopicProxy - hz.client_1 [dev] [3.8] Terminating MessageListener TopicMessageLostTest$MyClient@6075b2d3 on topic: my-topic. Reason: Unhandled exception, message: Internal executor rejected task: com.hazelcast.spi.impl.AbstractInvocationFuture$1@160396db, because client is shutting down... 
java.util.concurrent.RejectedExecutionException: Internal executor rejected task: com.hazelcast.spi.impl.AbstractInvocationFuture$1@160396db, because client is shutting down... 
at com.hazelcast.client.spi.impl.ClientExecutionServiceImpl$2.rejectedExecution(ClientExecutionServiceImpl.java:81) ~[hazelcast-all-3.8.jar:3.8] 
at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:823) ~[?:1.8.0_102] 
at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1369) ~[?:1.8.0_102] 
at com.hazelcast.client.spi.impl.ClientExecutionServiceImpl.execute(ClientExecutionServiceImpl.java:98) ~[hazelcast-all-3.8.jar:3.8] 
at com.hazelcast.spi.impl.AbstractInvocationFuture.unblock(AbstractInvocationFuture.java:239) [hazelcast-all-3.8.jar:3.8] 
at com.hazelcast.spi.impl.AbstractInvocationFuture.unblockAll(AbstractInvocationFuture.java:225) [hazelcast-all-3.8.jar:3.8] 
at com.hazelcast.spi.impl.AbstractInvocationFuture.complete(AbstractInvocationFuture.java:362) [hazelcast-all-3.8.jar:3.8] 
at com.hazelcast.client.spi.impl.ClientInvocation.notifyException(ClientInvocation.java:159) [hazelcast-all-3.8.jar:3.8] 
at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport.shutdown(ClientInvocationServiceSupport.java:186) [hazelcast-all-3.8.jar:3.8] 
[...]

PR #10323, opened 2017-04-12
Aligning exception delivered to callback behaviour of client with server

### LINKED ISSUE #10335: ExecutionCallback.onFailure behaves differently between client and server
In hazelcast servers, exceptions are delivered to `ExecutionCallback.onFailure` directly.
In hazelcast clients, they are wrapped to ExecutionException.

sancar:
`ExecutionCallback.onFailure` registered to `ICompletableFuture.andThen`
 was wrapped to ExecutionException on client while on server exception
 is directly passed to `onFailure` without wrapping.

 Client behaviour is changed to behave same servers.

fixes #10335

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
@@ -63,19 +63,6 @@ public class ClientInvocationFuture extends AbstractInvocationFuture<ClientMessa
     }
 
     @Override
-    protected Throwable unwrap(Throwable throwable) {
-        return throwable;
-    }
-
-    @Override
-    protected Object resolve(Object value) {
-        if (value instanceof Throwable) {
-            return new ExecutionException((Throwable) value);
-        }
-        return value;
-    }
-
-    @Override
     public void andThen(ExecutionCallback<ClientMessage> callback) {
         isNotNull(callback, "callback");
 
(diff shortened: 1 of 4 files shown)


sancar:
@pveentjer 
After reading your comment I realised issue title could be misleading.  `resolve` method is not really same with server. Client `resolve` method is actually no-op after this change on client.  
I have updated description and added test to make the intention clearer.

This part come to my attention because of another issue I am working on https://github.com/hazelcast/hazelcast/issues/10291

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocationFuture.java
```diff
@@ -63,19 +63,6 @@ protected TimeoutException newTimeoutException(long timeout, TimeUnit unit) {
     }
 
     @Override
-    protected Throwable unwrap(Throwable throwable) {
-        return throwable;
-    }
-
-    @Override
-    protected Object resolve(Object value) {
```
pveentjer: This is the good part from the client: wrapping the remote exception in a generic exception. This is how it should be done in member/member in hz4.

sancar:
run-lab-run

sancar:
run-lab-run

sancar:
run-lab-run

sancar:
run-lab-run

sancar:
run-lab-run

sancar:
run-lab-run

PR #10332, opened 2017-04-12
Only queue SplitBrainJoinMessages for processing on master node

### LINKED ISSUE #10325: Split Brain messages still occupies memory after recovery
Hi,  
we have Hazelcast v3.8 cluster with around 80 nodes distributed over 4 servers (2x24,2x16)
Usually the cluster breaks down during server-by-server apps redeployment and recovers after a while.

When we analyzed our last heap dump (~97% full), we found ~66k messages of `com.hazelcast.internal.cluster.impl.SplitBrainJoinMessage` allocated in 616MB. That's a lot.

Each message occupied 9kB and most of the message (8KB) consists of `memberAddresses ArrayList<com.hazelcast.nio.Address>` of 80 items (=number of nodes).

Those messages cannot even be GarbageCollected because of a reference from thread `com.hazelcast.internal.partition.impl.MigrationThread`.

We also noticed this behavior without server restart. After 26h server uptime there were 18k SplitBrainJoinMessages taking 167MB on heap. No restart was done, only some heavy tests and there are no Hazelcast errors in logs other than "MonitorInvocationsTask/BroadcastOperationControlTask delayed" right before heap dump creation.

What can we do to somehow debug this, or cleanup those messages from heap?
If you need more information please let me know.

Thanks in advance.
Regards

Stanislav

Environment:
Hazelcast 3.8
Cluster of 80 members
0 clients, each web application acts as a cluster member.
JDK7u75
hazelcast JVM related params: -Djava.net.preferIPv4Stack=true -DscMulticastAddress=231.12.21.132 -DscMulticastPort=50000 -DscIpTtl=1
OS: Oracle Linux 6.1

vbekiaris:
The `SplitBrainMulticastListener` is registered on every member, however only master needs to actually queue `SplitBrainJoinMessage`s, as it is the only member that executes the `SplitBrainHandler`. This was (implicitly) the case before #9328 was merged, because the split brain multicast listener was only created & registered during the execution of `SplitBrainHandler`.

Fixes #10325 in `master`, a backport for 3.8.2 will be created once reviewed & merged.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastJoiner.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastJoiner.java
@@ -118,7 +118,7 @@ public class MulticastJoiner extends AbstractJoiner {
                             logger.fine("Ignoring merge join response, since " + joinInfo.getAddress()
                                     + " is already a member.");
                         }
-                        return;
+                        continue;
                     }
 
                     if (joinInfo.getMemberCount() == 1) {
@@ -144,8 +144,20 @@ public class MulticastJoiner extends AbstractJoiner {
     }
 
     @Override
+    public void reset() {
+        super.reset();
+        // since this node is going to merge with a detected cluster, clear the queued split brain join messages (if any)
+        splitBrainJoinMessages.clear();
+    }
+
+    @Override
     public String getType() {
         return "multicast";
+    }
+
+    // for tests only
+    public int getSplitBrainMessagesCount() {
+        return splitBrainJoinMessages.size();
     }
 
     void onReceivedJoinRequest(JoinRequest joinRequest) {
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastJoiner.java
```diff
@@ -118,7 +118,7 @@ public void searchForOtherClusters() {
                             logger.fine("Ignoring merge join response, since " + joinInfo.getAddress()
                                     + " is already a member.");
                         }
-                        return;
+                        continue;
```
vbekiaris: Continue processing `SplitBrainJoinMessage`s even if the last one received is ignored, so that all messages in the deque will be processed in one execution of the `SplitBrainHandler` as [originally intended](https://github.com/hazelcast/hazelcast/pull/9328#discussion_r89759399).

vbekiaris:
Thanks for the quick reviews!

PR #10333, opened 2017-04-13
Use newUnsecureUUID() for metadata UUID

jerrinot:
We only need unique IDs, but not necessary random.
Secure UUIDs cause contention.


jerrinot:
run-lab-run

jerrinot:
the failure above is from an unrelated test which is time-sensitive. I'll send an extra PR.

jerrinot:
run-lab-run

pveentjer:
run-lab-run

pveentjer:
verify

PR #10334, opened 2017-04-13
Minor ClientInvocationServiceSupport loop cleanup

pveentjer:
The doRun loop has been simplified a bit.


pveentjer:
Merging since very small change.

PR #10336, opened 2017-04-13
Alligning exception delivered to callback of client with server

### LINKED ISSUE #10335: ExecutionCallback.onFailure behaves differently between client and server

sancar:
`ExecutionCallback.onFailure` registered to `ICompletableFuture.andThen`
was wrapped to ExecutionException on client while on server exception
is directly passed to `onFailure` wihout wrapping.

Client behaviour is changed as servers.

fixes #10335


sancar:
run-lab-run

Issue #10185, opened 2017-03-30
 Add/Remove Listeners, with Cluster Member Shutdown, Hangs 


Danny-Hazelcast:
using 3.8.1

when adding and removing listeners from the Client size, while Shutting down cluster Members
and restart.  the test hangs and the clients never return from a call to adding or removing listeners.

e.g functions: 
addItemListener(new ItemCounterListener(), true);     or   q.removeItemListener(id);
can hang and never return.

the test ran on a 3 member 3 client cluster,  the clients ran the test,  while the members where
shutdown and restarted one by one.

all clients running the test Hung, did not return and timedOut.
```
listeners_add-remove_partition_map HzClient2HZ timeout=899 running bench
 listeners_add-remove_partition_map HzClient1HZ timeout=899 running bench
 listeners_add-remove_partition_map HzClient3HZ timeout=899 running bench
 listeners_add-remove_set HzClient2HZ timeout=899 running bench
 listeners_add-remove_set HzClient1HZ timeout=899 running bench
 listeners_add-remove_set HzClient3HZ timeout=899 running bench
 listeners_add-remove_repMap HzClient2HZ timeout=899 running bench
 listeners_add-remove_repMap HzClient1HZ timeout=899 running bench
 listeners_add-remove_repMap HzClient3HZ timeout=899 running bench
 listeners_add-remove_queue HzClient2HZ timeout=899 running bench
 listeners_add-remove_queue HzClient1HZ timeout=899 running bench
 listeners_add-remove_queue HzClient3HZ timeout=899 running bench
 listeners_add-remove_map HzClient2HZ timeout=899 running bench
 listeners_add-remove_map HzClient1HZ timeout=899 running bench
 listeners_add-remove_map HzClient3HZ timeout=899 running bench
 listeners_add-remove_distObj HzClient2HZ timeout=899 running bench
 listeners_add-remove_distObj HzClient1HZ timeout=899 running bench
 listeners_add-remove_distObj HzClient3HZ timeout=899 running bench
 listeners_add-remove_list HzClient2HZ timeout=899 running bench
 listeners_add-remove_list HzClient1HZ timeout=899 running bench
 listeners_add-remove_list HzClient3HZ timeout=899 running bench
```

we have jstack trace from all clients, and all members in the cluster
```
output/HZ/HzClient1HZ/timeout_stack.txt
output/HZ/HzMember1HZ/timeout_stack.txt
output/HZ/HzClient2HZ/timeout_stack.txt
[...]

PR #10338, opened 2017-04-13
Fixes deadlock on listener registration

sancar:
Deadlock occurs when listener registration failed to a member and
client trying to deregister it from other registered members.

fixes https://github.com/hazelcast/hazelcast/issues/10185

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
@@ -61,6 +61,9 @@ public class ClientSmartListenerService extends ClientListenerServiceImpl
 
     @Override
     public String registerListener(final ListenerMessageCodec codec, final EventHandler handler) {
+        //This method should not be called from registrationExecutor
+        assert (!Thread.currentThread().getName().contains("eventRegistration"));
+
         Future<String> future = registrationExecutor.submit(new Callable<String>() {
             @Override
             public String call() {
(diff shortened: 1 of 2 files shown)


sancar:
run-lab-run

sancar:
run-lab-run

sancar:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientSmartListenerService.java
```diff
@@ -74,7 +77,7 @@ public String call() {
                         invoke(registrationKey, connection);
                     }
                 } catch (Exception e) {
-                    deregisterListener(userRegistrationId);
```
ihsandemir: Understand the deadlock reason. The change introduced a behaviour difference. The registerListener call will now block until a cleanup is done (deregister finishes) in case of failure, and sometimes this cleanup may take some time since the server may not respond. Is this the behaviour we want?

sancar: `The change introduced a behaviour difference` Yes and no. 
Yes, it is a behaviour change because otherwise it was always deadlock when this method is called. 
No, it is not a behaviour change because we were already waiting for future inside `deregisterListener` . This means intention was to wait for cleanup to finish in case of failure. 

`Is this the behaviour we want?`
I think it is. We don't want to have a listener getting events after registerListener saying it is failed to register.

sancar:
run-lab-run

sancar:
run-lab-run

PR #10340, opened 2017-04-13
Increases client buffersize from 32KB to 128KB

pveentjer:
Especially with larger values being send/received this gives a significant boost
in performance and reduction in latency; also the higher percentiles.


pveentjer:
run-lab-run

pveentjer:
verify

PR #10343, opened 2017-04-13
Apply minor improvements in migration tests

metanet:
Occasionally, migration tests fail in slow build environment because of some false alarms. For instance, a migration starts and immediately fails because a new member joins to the cluster but doesn't complete the start process yet.
Changes here make the tests less fragile for similar cases.


mmedenjak:
@metanet Great job! I hope we solve this slow environment problem but also it's much better when the tests fail with some kind of "time out" rather than stuff like "AssertionError : null" or "expected x, got y".
I am thinking of adding these assertions on some common failing tests and generally lowering the partition count for tests that don't rely on it being the default.

PR #10345, opened 2017-04-13
[BACKPORT] Apply minor improvements in migration tests

metanet:
Occasionally, migration tests fail in slow build environment because of same false alarms. For instance, a migration starts and immediately fails because a new member joins to the cluster but doesn't complete the start process yet.
Changes here make the tests less fragile for similar cases.

Backport of #10343


metanet:
run-lab-run

PR #10353, opened 2017-04-17
Packet not extending HeapData any longer

pveentjer:
This optimization was added some time ago to prevent creating and
HeapData and Packet; but with the recent response optimizations
this 'ugly' design isn't needed any longer.

No Enterprise PR needed. 

Doesn't change anything that goes over the wire; so no minor upgrade incompatibility.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
@@ -28,6 +28,7 @@ import com.hazelcast.internal.metrics.MetricsProvider;
 import com.hazelcast.internal.metrics.MetricsRegistry;
 import com.hazelcast.internal.metrics.Probe;
 import com.hazelcast.internal.partition.InternalPartition;
+import com.hazelcast.internal.serialization.impl.HeapData;
 import com.hazelcast.internal.serialization.impl.SerializationServiceV1;
 import com.hazelcast.internal.util.counters.Counter;
 import com.hazelcast.logging.ILogger;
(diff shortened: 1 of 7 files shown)


pveentjer:
verify

pveentjer:
verify

hazelcast/src/test/java/com/hazelcast/nio/tcp/OperationPacketFilter.java
```diff
@@ -41,7 +42,7 @@ public final boolean allow(Packet packet, Address endpoint) {
 
     private boolean allowOperation(Packet packet, Address endpoint) {
         try {
-            ObjectDataInput input = serializationService.createObjectDataInput(packet);
+            ObjectDataInput input = serializationService.createObjectDataInput(new HeapData(packet.getPayload()));
```
Donnerbart: We are just extracting a few fields from the Packet here, don't we? Maybe we can skip the whole serialization part if the Packet would have `isIdentified()`, `isCompressed()`, `getFactoryId()` and `getType()` methods, which would directly peak into its bytes. This would also get rid of the introduced `new HeapData()`.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/OperationRunnerImpl.java
```diff
@@ -426,9 +426,9 @@ public void run(Packet packet) throws Exception {
      * <p>
      * It makes an assumption that the callId is the first long field in the serialized operation.
      */
-    private long extractOperationCallId(Data data) throws IOException {
+    private long extractOperationCallId(Packet packet) throws IOException {
         ObjectDataInput input = ((SerializationServiceV1) node.getSerializationService())
-                .initDataSerializableInputAndSkipTheHeader(data);
+                .initDataSerializableInputAndSkipTheHeader(new HeapData(packet.getPayload()));
```
Donnerbart: This looks like a good candidate for a `Packet::getCallId()` method to skip the whole serialization part. This would also get rid of the introduced `new HeapData()`.

Holmistr:

PR #10355, opened 2017-04-17
Introduced the OutboundOperationHandler

pveentjer:
The logic of converting an Operation to packet and sending it to the
connection has now been pulled into a separate class. 

The OutboundOperationHandler is comparable to the OutboundResponseHandler.


pveentjer:
verify

pveentjer:
Since it is just moving a method from one class to another, I'll merge this one.

PR #10356, opened 2017-04-17
Introduce fragmented migration and anti-entropy

mdogan:
- Introduces fragmented migrations, which allows migrating partition replicas in smaller fragments instead of copying whole replica in single packet.  See https://hazelcast.atlassian.net/wiki/display/PM/Gigantic+Cache+Migration+Enhancements requirement 5.
For detailed design, see [Gigantic Cache Migration Enhancements Design](https://hazelcast.atlassian.net/wiki/display/EN/Gigantic+Cache+Migration+Enhancements+Design)

- Introduces fragmented anti-entropy replications, which replicating inconsistent backup replicas in smaller fragments. See https://hazelcast.atlassian.net/wiki/display/PM/Fine-Grained+Anti+Entropy+Mechanism.
For detailed design, see [Fine-Grained Anti Entropy Mechanism Design](https://hazelcast.atlassian.net/wiki/display/EN/Fine-Grained+Anti+Entropy+Mechanism+Design)


mdogan:
verify

pveentjer:
What is the implication of this PR on the regular map.get/map.put?

mdogan:
@pveentjer: `map.get()` should be unaffected. `map.put()` will have an additional `HashMap` lookup both on primary and backups.

pveentjer:
Ok. I doubt that is going to make any impact.

mdogan:
Only changes in operation execution flow are in `OperationBackupHandler` and `Backup` classes.

mdogan:
Additional litter in lookup itself; no. But backup-aware operations can create a namespace object for lookup. Like;

```java
public abstract class MapOperation extends AbstractNamedOperation implements IdentifiedDataSerializable, ReplicaFragmentAware {
   [...]
 
    @Override
    public ReplicaFragmentNamespace getReplicaFragmentNamespace() {
        return new DefaultReplicaFragmentNamespace(MapService.SERVICE_NAME, name);
    }
}
```

pveentjer:
Ok.. this is additional litter on the fast path. Can't this ReplicaFragmentNamespace be stored in the MapServiceConext?

mdogan:
That part is not included in this PR. This is only to introduce infrastructure. I think it can be stored in some map context, in that case it will need an additional lookup too instead of creating an object. 
Do you prefer that?

mdogan:
So, it will look like;

```java
public abstract class MapOperation extends AbstractNamedOperation implements IdentifiedDataSerializable, ReplicaFragmentAware {
   [...]
 
    @Override
    public ReplicaFragmentNamespace getReplicaFragmentNamespace() {
        return mapServiceContext.getNamespace(name);
    }
}
```

pveentjer:
Yes. I think this is less of a problem then litter.

hazelcast/src/main/java/com/hazelcast/spi/FragmentedMigrationAwareService.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Collection;
+
+/**
+ * {@code FragmentedMigrationAwareService} is an extension to the {@link MigrationAwareService} which allows
+ * migration/replication of partition replicas in smaller fragments.
+ * <p>
+ * Each replica fragment is distinguished by a {@link ReplicaFragmentNamespace}, all replica data belonging to
+ * a specific namespace will be transferred in the same packet.
+ * <p>
+ * Backup operations created by {@code FragmentedMigrationAwareService} must implement {@link ReplicaFragmentAware} interface
```
mmedenjak: Both backup and backup aware operations

hazelcast/src/main/java/com/hazelcast/internal/partition/operation/ReplicaSyncRequest.java
```diff
@@ -124,16 +148,6 @@ private boolean preCheckReplicaSync(NodeEngineImpl nodeEngine, int partitionId,
             return false;
         }
 
-        if (currentVersion == 0) {
```
mdogan: If replica version is `0`, then service (most probably) won't create a replication operation, because it will not have any data for that replica. Even if service creates an operation, it will be just empty, won't contain any data.

hazelcast/src/main/java/com/hazelcast/internal/partition/operation/MigrationOperation.java
```diff
@@ -221,44 +168,39 @@ private void onOperationFailure(Operation op, Throwable e) {
...
+        out.writeBoolean(firstFragment);
+        out.writeBoolean(lastFragment);
     }
 
     @Override
-    public int getId() {
-        return PartitionDataSerializerHook.MIGRATION;
+    protected void readInternal(ObjectDataInput in) throws IOException {
+        super.readInternal(in);
+        fragmentMigrationState = new ReplicaFragmentMigrationState();
```
mdogan: yes, could be. but is it going to be better? we already know the concrete type here.

mmedenjak: Not necessarily better but it will be consistent with other IDS deserialisations.

hazelcast/src/main/java/com/hazelcast/internal/partition/InternalReplicaFragmentNamespace.java
```diff
@@ -0,0 +1,79 @@
...
+ * {@link com.hazelcast.spi.FragmentedMigrationAwareService} will be registered with {@code InternalReplicaFragmentNamespace}.
+ *
+ * @see com.hazelcast.spi.FragmentedMigrationAwareService
+ * @since 3.9
+ */
+public final class InternalReplicaFragmentNamespace implements ReplicaFragmentNamespace, IdentifiedDataSerializable {
+
+    public static final ReplicaFragmentNamespace INSTANCE = new InternalReplicaFragmentNamespace();
+
+    public InternalReplicaFragmentNamespace() {
```
mdogan: make sense. thanks for the tip.

mdogan:
@mmedenjak & @vbekiaris:  Thanks for the reviews.

PR #10360, opened 2017-04-17
Io balancing improvements

pveentjer:
This pr contains a few things

1: there is some undefined behavior in the NonBlockingThreadingModel where on shutdown the io threads array elements get nilled. This has been changed by just setting the whole array to null. This way the array remains effectively immutable.

2: a new diagnostics networking plugin that currently renders information about io threads. This should help to spot networking problems. This plugin works for both server and client.

2:  modifications to the non blocking io threads so it contains the total number of bytes, frames etc. This is done with a zero overhead on the fast path. So instead of having contented counters, periodically a task is written to each read/write handler to publish information to its owning io thread. This is disabled by default and only works when debug metrics level is enabled.

3: the non blocking readers/writers have a configurable mechanism for returning 'load'. So byte based,frame based, handle-method-count etc. The old behavior was always handle method count, but this doesnt' seem to give the best results. Byte based seems to be more reliable. Currently these properties are only configurable through a 'hidden' system property. In the 3.9 release we need to pick a better one and use that be default. But at least we can now experiment with the different policies.

Example output from the diagnostics plugin

```
17-4-2017 17:20:30 Networking[
                          InputThreads[
                                  hz._hzInstance_1_workers.IO.thread-in-0[
                                          frames=34.22 %
                                          priorityFrames=34.22 %
                                          bytes=33.96 %
                                          events=11.53 %
                                          handle-events=36.26 %
                                          tasks=38.66 %]
                                  hz._hzInstance_1_workers.IO.thread-in-1[
                                          frames=31.53 %
                                          priorityFrames=31.53 %
                                          bytes=31.48 %
                                          events=10.09 %
                                          handle-events=31.72 %
                                          tasks=30.16 %]
                                  hz._hzInstance_1_workers.IO.thread-in-2[
                                          frames=34.25 %
                                          priorityFrames=34.25 %
                                          bytes=34.57 %
                                          events=10.19 %
                                          handle-events=32.03 %
                                          tasks=31.18 %]]
                          OutputThreads[
                                  hz._hzInstance_1_workers.IO.thread-out-0[
                                          frames=32.46 %
                                          priorityFrames=32.46 %
                                          bytes=32.82 %
                                          events=0.00 %
                                          handle-events=33.25 %
                                          tasks=33.25 %]
                                  hz._hzInstance_1_workers.IO.thread-out-1[
                                          frames=33.85 %
                                          priorityFrames=33.85 %
                                          bytes=33.80 %
                                          events=0.00 %
                                          handle-events=33.36 %
                                          tasks=33.36 %]
                                  hz._hzInstance_1_workers.IO.thread-out-2[
                                          frames=33.68 %
                                          priorityFrames=33.68 %
                                          bytes=33.38 %
                                          events=0.00 %
                                          handle-events=33.40 %
                                          tasks=33.40 %]]]
```

The above output contains a bug in the events, which already has been addressed.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -156,6 +156,10 @@ public class ClientConnectionManagerImpl implements ClientConnectionManager {
         this.credentials = client.getCredentials();
     }
 
+    public NonBlockingIOThreadingModel getIoThreadingModel() {
+        return ioThreadingModel;
+    }
+
     protected void initIOThreads(HazelcastClientInstanceImpl client) {
         HazelcastProperties properties = client.getProperties();
         boolean directBuffer = properties.getBoolean(SOCKET_CLIENT_BUFFER_DIRECT);
--- a/hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
@@ -37,6 +37,8 @@ import java.util.Queue;
 import java.util.Random;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
+import static com.hazelcast.internal.metrics.ProbeLevel.DEBUG;
+import static com.hazelcast.internal.networking.nonblocking.SelectorMode.SELECT_NOW;
 import static com.hazelcast.internal.networking.nonblocking.SelectorOptimizer.optimize;
 import static com.hazelcast.internal.util.counters.SwCounter.newSwCounter;
 import static java.lang.Math.max;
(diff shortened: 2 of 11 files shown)


pveentjer:
verify

jerrinot:
please change commit message to before merging it

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThreadingModel.java
```diff
@@ -51,8 +55,8 @@
 public class NonBlockingIOThreadingModel
         implements IOThreadingModel {
 
-    private final NonBlockingIOThread[] inputThreads;
-    private final NonBlockingIOThread[] outputThreads;
+    private volatile NonBlockingIOThread[] inputThreads;
```
Donnerbart: I would prefer the order `final`, `normal` and `volatile` for fields (to see faster which are immutable or if they need to be thread-safe).

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -192,6 +196,8 @@ protected void initIOThreads(HazelcastClientInstanceImpl client) {
                 new ClientSocketReaderInitializer(getBufferSize(), directBuffer));
     }
 
+
```
Donnerbart: Superfluous newlines.

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThread.java
```diff
@@ -92,6 +103,7 @@
     // set to true while testing
     private boolean selectorWorkaroundTest;
 
+
```
Donnerbart: Another newline you could put back in your savings box :)

hazelcast/src/main/java/com/hazelcast/internal/networking/nonblocking/NonBlockingIOThreadingModel.java
pveentjer: Me too.I'll fix it.

Issue #9696, opened 2017-01-19
Client IOException escapes to user


sancar:
Background on the issue for anyone come across this exception:
IOException is a retryable Exception for HazelcastClient. It should only send to user when `hazelcast.client.invocation.timeout.seconds` is passed. This value is 2 minutes by default. 
From reports, we dont really know if this value is passed or not. We may need to have a log to tell if invocation timeout is violated. Issue is not reproduced so far in local.

Reports so far indicates issue appeared in 3.7 

Sample reported stack traces

		java.io.IOException: No available connection to address [REDACTED]:5702
		com.hazelcast.core.HazelcastException: java.io.IOException: No available connection to address [REDACTED]:5702
		at com.hazelcast.util.ExceptionUtil.peel(ExceptionUtil.java:73) ~[hazelcast-all-3.7.jar!/:3.7]
		at com.hazelcast.util.ExceptionUtil.peel(ExceptionUtil.java:63) ~[hazelcast-all-3.7.jar!/:3.7]
		at com.hazelcast.util.ExceptionUtil.peel(ExceptionUtil.java:52) ~[hazelcast-all-3.7.jar!/:3.7]
		at com.hazelcast.util.ExceptionUtil.rethrow(ExceptionUtil.java:83) ~[hazelcast-all-3.7.jar!/:3.7]
		at com.hazelcast.client.spi.ClientProxy.invokeOnPartition(ClientProxy.java:155) ~[hazelcast-all-3.7.jar!/:3.7]
		at com.hazelcast.client.spi.ClientProxy.invoke(ClientProxy.java:147) ~[hazelcast-all-3.7.jar!/:3.7]
		at com.hazelcast.client.proxy.ClientMapProxy.getInternal(ClientMapProxy.java:245) ~[hazelcast-all-3.7.jar!/:3.7]
		at com.hazelcast.client.proxy.ClientMapProxy.get(ClientMapProxy.java:240) ~[hazelcast-all-3.7.jar!/:3.7]

and

			com.hazelcast.core.HazelcastException: java.io.IOException: Not able to setup owner connection!
		    at com.hazelcast.util.ExceptionUtil.peel(ExceptionUtil.java:73)
		    at com.hazelcast.util.ExceptionUtil.peel(ExceptionUtil.java:63)
		    at com.hazelcast.util.ExceptionUtil.peel(ExceptionUtil.java:52)
[...]

PR #10370, opened 2017-04-18
Client invocation exception made more verbose when timed out

sancar:
We were setting last exception to user future when invocation
timed out. That was making it harder to reason about what is going
on. OperationTimeoutException is used to differentiate cases.

PR is prepared especially to understand following issue: https://github.com/hazelcast/hazelcast/issues/9696

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
@@ -26,6 +26,7 @@ import com.hazelcast.core.HazelcastInstanceNotActiveException;
 import com.hazelcast.core.HazelcastOverloadException;
 import com.hazelcast.core.LifecycleService;
 import com.hazelcast.logging.ILogger;
+import com.hazelcast.core.OperationTimeoutException;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
 import com.hazelcast.spi.exception.RetryableException;
@@ -61,7 +62,7 @@ public class ClientInvocation implements Runnable {
     private final Connection connection;
     private volatile ClientConnection sendConnection;
     private boolean bypassHeartbeatCheck;
-    private long retryTimeoutPointInMillis;
+    private long retryExpirationMillis;
     private EventHandler handler;
 
     protected ClientInvocation(HazelcastClientInstanceImpl client,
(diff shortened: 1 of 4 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -176,28 +177,34 @@ public void notifyException(Throwable exception) {
...
 
-        if ((isBindToSingleConnection() && exception instanceof IOException)
-                || System.currentTimeMillis() > retryTimeoutPointInMillis) {
+        if (isBindToSingleConnection() && exception instanceof IOException) {
             clientInvocationFuture.complete(exception);
             return;
         }
 
-        if (isRetrySafeException(exception)
+        boolean isAllowedToRetry = isRetrySafeException(exception)
```
pveentjer: perhaps just rename it to 'retry'? This also explains the intent.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -176,28 +177,34 @@ public void notifyException(Throwable exception) {
...
-                    logger.finest("Retry could not be scheduled ", e);
-                }
-                clientInvocationFuture.complete(exception);
+                || (exception instanceof TargetDisconnectedException && clientMessage.isRetryable());
+        if (!isAllowedToRetry) {
+            clientInvocationFuture.complete(exception);
+            return;
+        }
+
+        long msAfterTimeout = System.currentTimeMillis() - retryTimeoutPointInMillis;
```
pveentjer: can we rename 'retryTimeoutPointInMillis' to 'retryExpirationMillis' 

I would rename 'msAfterTimeout' to something like 'remainingMillis'. If it is a negative value, it has expired.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -176,28 +177,34 @@ public void notifyException(Throwable exception) {
...
+        if (msAfterTimeout > 0) {
+            if (logger.isFinestEnabled()) {
+                logger.finest("Exception will not be retried because invocation timed out", exception);
             }
+            clientInvocationFuture.complete(new OperationTimeoutException("Invocation timed out by " + msAfterTimeout + " ms"));
```
pveentjer: Probably it would be good to capture more information. So when has this call started, address, partition id etc etc etc. On the serverside there is a lot of information available when the infamous OperationTimeoutException is thrown.

pveentjer:
Looks good. Minor review comments.

sancar:
@pveentjer I have updated pr according to your comments

sancar:
run-lab-run

PR #10371, opened 2017-04-18
Nearcache doEvictionIfRequired key-argument

pveentjer:
A key argument has been added to the doEvictionIfRequired so for HZ enterprise so that the nearcache eviction can be done on a single segment instead of all segments.


Holmistr:

PR #10372, opened 2017-04-18
Convert Map, Cache, MultiMap and Lock to FragmantedMigrationAwareService

metanet:
Depends on #10356 
Depends on https://github.com/hazelcast/hazelcast/pull/10435

Commits can be reviewed one-by-one, each commit is specific to related service only.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
@@ -26,16 +26,18 @@ import com.hazelcast.config.EvictionConfig.MaxSizePolicy;
 import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.core.HazelcastInstanceAware;
 import com.hazelcast.internal.diagnostics.StoreLatencyPlugin;
+import com.hazelcast.internal.eviction.EvictionChecker;
 import com.hazelcast.internal.eviction.EvictionListener;
 import com.hazelcast.internal.eviction.EvictionPolicyEvaluatorProvider;
-import com.hazelcast.internal.eviction.EvictionChecker;
 import com.hazelcast.internal.eviction.impl.evaluator.EvictionPolicyEvaluator;
 import com.hazelcast.internal.eviction.impl.strategy.sampling.SamplingEvictionStrategy;
 import com.hazelcast.map.impl.MapEntries;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.spi.DistributedObjectNamespace;
 import com.hazelcast.spi.EventRegistration;
 import com.hazelcast.spi.NodeEngine;
+import com.hazelcast.spi.ObjectNamespace;
 import com.hazelcast.spi.impl.NodeEngineImpl;
 import com.hazelcast.spi.impl.eventservice.InternalEventService;
 import com.hazelcast.util.Clock;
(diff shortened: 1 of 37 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
```diff
@@ -131,4 +133,14 @@ void clearHavingLesserBackupCountThan(int backupCount) {
             }
         }
     }
+
+   public Collection<ServiceNamespace> getAllNamespaces(int replicaIndex) {
+        Collection<ServiceNamespace> namespaces = new HashSet<ServiceNamespace>();
```
pveentjer: I guess there can be duplicate names spaces?

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/AbstractCacheOperation.java
```diff
@@ -107,6 +109,16 @@ public void logError(Throwable e) {
     }
 
     @Override
+    public ObjectNamespace getServiceNamespace() {
```
pveentjer: will this cause any litter?

hazelcast/src/main/java/com/hazelcast/cache/impl/CachePartitionSegment.java
mdogan: As far as I know, no, each record-store will have a different name. We can use a list too.

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/AbstractCacheOperation.java
mdogan: No, record-store has a final namespace field, that will be returned.

hazelcast/src/main/java/com/hazelcast/cache/impl/AbstractCacheRecordStore.java
```diff
@@ -147,6 +150,7 @@ public AbstractCacheRecordStore(String name, int partitionId, NodeEngine nodeEng
         evictionChecker = createCacheEvictionChecker(evictionConfig.getSize(), evictionConfig.getMaximumSizePolicy());
         evictionPolicyEvaluator = createEvictionPolicyEvaluator(evictionConfig);
         evictionStrategy = createEvictionStrategy(evictionConfig);
+        objectNamespace = new DistributedObjectNamespace(ICacheService.SERVICE_NAME, name);
```
ahmetmircik: is it ok to use same name-space for all record stores?

mdogan: No, there's a namespace per record-store.

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CachePutAllBackupOperation.java
```diff
@@ -78,6 +80,16 @@ public void run() throws Exception {
     }
 
     @Override
+    public ObjectNamespace getServiceNamespace() {
```
ahmetmircik: may be not subject of this pr but it would be better to put this method to a single place to eliminate code duplications

hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
```diff
@@ -122,6 +124,22 @@ private RecordStore createRecordStore(String name) {
         return maps.values();
     }
 
+    public Collection<ServiceNamespace> getAllNamespaces(int replicaIndex) {
+        Collection<ServiceNamespace> namespaces = new HashSet<ServiceNamespace>();
+
+        for (RecordStore recordStore : maps.values()) {
```
ahmetmircik: seems using `mapServiceContext.getMapContainers()` would be better to eliminate excessive looping on record-stores

hazelcast/src/main/java/com/hazelcast/cache/impl/operation/CachePutAllBackupOperation.java
mdogan: yes, that's a pain point. unlike map, cache operations don't have a single parent.

hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
mdogan: why excessive? Isn't there just a single record-store per namespace in a partition?

PR #10373, opened 2017-04-18
Reduced complexity in AbstractNearCacheRecordStore

Donnerbart:
* deprecated `EvictionPolicyType`
* removed non overridden methods in `AbstractNearCacheRecordStore`
* removed `null` checks for `EvictionConfig` fields

`EvictionPolicyType` was just used internally. Originally it had less fields than `EvictionPolicy`, but they are equal now, so there cannot be an `EvictionPolicyType` of `null` anymore. So `EvictionPolicy` can be used as replacement to reduce duplication and complexity.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
@@ -200,6 +200,7 @@ public class EvictionConfig implements EvictionConfiguration, DataSerializable, 
         return this;
     }
 
+    @Override
     public EvictionPolicy getEvictionPolicy() {
         return evictionPolicy;
     }
@@ -234,18 +235,22 @@ public class EvictionConfig implements EvictionConfiguration, DataSerializable, 
         return EvictionStrategyType.DEFAULT_EVICTION_STRATEGY;
     }
 
-    @Override
+    /**
+     * @deprecated since 3.9, use {@link #getEvictionPolicy()} instead
+     */
+    @Deprecated
     public EvictionPolicyType getEvictionPolicyType() {
-        if (evictionPolicy == EvictionPolicy.LFU) {
-            return EvictionPolicyType.LFU;
-        } else if (evictionPolicy == EvictionPolicy.LRU) {
-            return EvictionPolicyType.LRU;
-        } else if (evictionPolicy == EvictionPolicy.RANDOM) {
-            return EvictionPolicyType.RANDOM;
-        } else if (evictionPolicy == EvictionPolicy.NONE) {
-            return EvictionPolicyType.NONE;
-        } else {
-            return null;
+        switch (evictionPolicy) {
+            case LFU:
+                return EvictionPolicyType.LFU;
+            case LRU:
+                return EvictionPolicyType.LRU;
+            case RANDOM:
+                return EvictionPolicyType.RANDOM;
+            case NONE:
+                return EvictionPolicyType.NONE;
+            default:
+                return null;
         }
     }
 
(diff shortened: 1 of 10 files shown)


Donnerbart:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionConfiguration.java
```diff
@@ -29,11 +31,11 @@
...
 
     /**
-     * Gets the type of eviction policy.
+     * Gets the eviction policy.
      *
-     * @return the type of eviction policy
+     * @return the eviction policy
      */
-    EvictionPolicyType getEvictionPolicyType();
+    EvictionPolicy getEvictionPolicy();
```
ahmetmircik: public api and should be kept as is? see [EvictionConfig](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java#L39)

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/store/AbstractNearCacheRecordStore.java
```diff
@@ -121,17 +122,19 @@ protected AbstractNearCacheRecordStore(NearCacheConfig nearCacheConfig, NearCach
         this.serializationService = serializationService;
         this.classLoader = classLoader;
         this.nearCacheStats = nearCacheStats;
-        this.evictionPolicyType = nearCacheConfig.getEvictionConfig().getEvictionPolicyType();
+
+        this.isEvictionDisabled = nearCacheConfig.getEvictionConfig().getEvictionPolicy() == EvictionPolicy.NONE;
```
ahmetmircik: `isEvictionDisabled` --> `evictionDisabled` ?

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -234,18 +235,22 @@ public EvictionStrategyType getEvictionStrategyType() {
...
+            case LFU:
+                return EvictionPolicyType.LFU;
+            case LRU:
+                return EvictionPolicyType.LRU;
+            case RANDOM:
+                return EvictionPolicyType.RANDOM;
+            case NONE:
+                return EvictionPolicyType.NONE;
+            default:
+                throw new IllegalArgumentException("Unknown EvictionPolicy configured! " + evictionPolicy);
```
ahmetmircik: behaviour change?
previously returning `null` but now `throwing exception`

Donnerbart: I can revert it, but it actually can never happen that `null` is returned, since both enums have the same elements.

ahmetmircik: thanks ok i can see now there should be any issue with this change.

hazelcast/src/main/java/com/hazelcast/internal/eviction/EvictionConfiguration.java
Donnerbart: I've added the old method again and deprecated it.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/store/AbstractNearCacheRecordStore.java
Donnerbart: Done

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
Donnerbart: I reverted it, just to be sure. We just use it in the `ConfigCompatibilityChecker.MapConfigChecker.check()`, where a `null` value won't hurt even if it would occur.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

PR #10376, opened 2017-04-18
Support symmetric encryption for Multicast

tkountis:
Group password leaked in the multicast datagram is considered a security flaw.
Starting from 3.8.2 (#10377) it will get deprecated from the multicast protocol, and validation will be removed allowing future releases to completely remove this piece of information from the protocol. 

In 3.9 we also stop broadcasting the value, we send instead an empty string.
Furthermore, as an additional layer of protection, when symmetric encryption is in use, datagrams will also be encrypted, to prevent any other sensitive information leaking.

Group name is not affected by this change, and also any form of Authentication relying on the password is not affected either.


tkountis:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
```diff
@@ -179,7 +171,8 @@ public int getId() {
     @Override
     public void writeData(ObjectDataOutput out) throws IOException {
         out.writeUTF(groupName);
-        out.writeUTF(groupPassword);
+        //TODO @tkountis remove in 4.0
```
pveentjer: what happened to the password check?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/EchoNoCopyByteArrayProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.internal.cluster.impl;
+
+import com.hazelcast.util.ByteArrayProcessor;
+
+public class EchoNoCopyByteArrayProcessor
```
pveentjer: The name of this class doesn't reveal its intent. I'm a but confused about Echo No Copy... but In the code I do see an array copy.

hazelcast/src/main/java/com/hazelcast/util/ByteArrayProcessor.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.hazelcast.util;
+
+/**
+ * A byte array processor used to produce a byte[] out of the given one.
+ * Example use case would be when you have an encrypted byte array and want to decrypt it and vice-versa.
+ */
+public interface ByteArrayProcessor {
+
+    byte[] process(final byte[] src);
```
pveentjer: Do we really need to have finals in interface definitions? Having finals on local variables I already find questionable due to horizontal noise.. but on interfaces...

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeExtension.java
```diff
@@ -303,6 +305,16 @@ public String createMemberUuid(Address address) {
         return UuidUtil.createMemberUuid(address);
     }
 
+    @Override
+    public ByteArrayProcessor createMulticastInputProcessor() {
+        return new EchoNoCopyByteArrayProcessor();
```
pveentjer: Should this not be a singleton?

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
```diff
@@ -179,7 +171,8 @@ public int getId() {
...
+        out.writeUTF(EMPTY_PWD);
```
mmedenjak: As mentioned on the other PR, we can send the password to 3.8.x members and not send the password to 3.9 members. This would allow rolling upgrade of members with 3.8.1 or lower.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/EchoNoCopyByteArrayProcessor.java
mmedenjak: Agree, also maybe some javadoc.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/MulticastService.java
```diff
@@ -209,8 +217,10 @@ private JoinMessage receive() {
             try {
                 final byte[] data = datagramPacketReceive.getData();
                 final int offset = datagramPacketReceive.getOffset();
-                final BufferObjectDataInput input = node.getSerializationService().createObjectDataInput(data);
-                input.position(offset);
+                final int length = datagramPacketReceive.getLength();
+
+                final BufferObjectDataInput input = node.getSerializationService()
```
mmedenjak: We should possibly do some special casing here. This way, even if the processor is no-op, it will introduce an array copy.

hazelcast/src/test/java/com/hazelcast/cluster/ClusterJoinConfigCheckTest.java
```diff
@@ -52,28 +51,6 @@ public void killAllHazelcastInstances() throws IOException {
     }
 
     @Test
-    public void tcp_whenGroupPasswordMismatch_thenNewNodeIsShutDown() {
```
mmedenjak: We can add some tests verifying that the password is empty and some tests for verifying that some custom byte array processor was called.

mmedenjak:
run-lab-run

hazelcast/src/test/java/com/hazelcast/cluster/ClusterJoinConfigCheckTest.java
tkountis: Not super-psyched about the test I put in, since its not going to be as reliable as we would want it to be, but its something.

tkountis:
Did a minor optimization to avoid copying the data when have no special processor. That should do it for now.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
Donnerbart: It's part of the transition of the security features (and explained in the according TDD), so this should be fine.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
Donnerbart: I think this could be worth a try. It shouldn't be more than `out.writeUtf(clusterService.getClusterVersion().isLessThan(Versions.V3_9) ? groupPassword : EMPTY_PWD);`, shouldn't it?

tkountis: I included some short explanation in the TDD the other day, but its worth mentioning here too.
The JoinRequest is happening _before_ a member is part of a cluster, therefore the getClusterVersion() always returns the node's codebase. 

Having that in mind, the above snippet wouldn't make sense, because it will be part of a 3.9 release, meaning that the check will always be false. Therefore, sending empty-pwd by default, and if the receiving node is >=3.8.2 it will know to ignore whatever value is in this field.

vbekiaris: Actually [`Cluster.getClusterVersion` should return `Version.UNKNOWN`](https://github.com/vbekiaris/hazelcast/blob/c08968bd5592d56fcd5590342cc0cafcfe3bf7d0/hazelcast/src/main/java/com/hazelcast/core/Cluster.java#L193-L195) until node joins, so I think the suggestion to send out the password does make sense --> upgrade from 3.8 & 3.8.1 should be still possible.

tkountis: hmm, I am probably not seeing sth, but from your comment, even if the return value is UNKNOWN (instead of the codebase one), that would mean that we never get to exercise the EMPTY_PWD path, since UNKNOWN=0 which is always less then 3.9, and since this operation happens before cluster join, then the else path will never happen. Missing something ?

vbekiaris: I think conditionally sending the group password has value because `ConfigCheck` is used in multicast messages throughout the lifetime of the member and in split-brain join messages, not just until cluster is initially joined:
* Assuming an existing 3.8 EE cluster
* Rolling upgrade each member to 3.9 EE members (`UNKNOWN < 3.9` --> 3.9 member still sends out pwd and is allowed to join the 3.8 EE cluster)
* As long as cluster version is still `3.8`, the split-brain join message sent out from a 3.9 member will include the group password and will be allowed to join back to the 3.8 cluster.
* Once all members are upgraded, upgrade cluster version to 3.9 --> from now on multicast & split-brain join messages have a proper `ConfigCheck` with an empty password.

Otherwise, if in 3.9 the pwd is dropped, one would need to upgrade first 3.8 -> 3.8.2+ and then move on to 3.9. wdyt?

tkountis: All accurate points, thank you. However, I see a downside, which is major, security wise. The most often use-case of the ConfigCheck is during the 1st minute of a Node going up, which sends rapidly datagrams, until it discovers a cluster. During that period, single node (never joined) no-matter what state the cluster is at, we leak the password with this approach.

So imagine the above scenario with a 3.8 EE cluster, which we rolling-upgrade to 3.9 EE, one by one, successfully, so now the occasional ConfigCheck payload, from the cluster to the 'wild', is properly skipping the password, but now we expand the cluster by adding a new node, which now again leaks the password, until it joins.

Of course I am ok with both solutions, but in your approach we will have to go to 4.0 totally stop leaking the pwd, whereas in my approach we can do it now, with the compromise that the client needs to first upgrade to latest 3.8 patch, which understandably is a burden. 

I guess this is a TDD decision, so do you mind adding your comment in the TDD ?

PR #10377, opened 2017-04-18
Remove group password validation during ConfigCheck

tkountis:
Group password leaked in the multicast datagram is considered a security flaw.
Starting from 3.8.2 it will get deprecated from the multicast protocol, and validation will be removed allowing future releases to completely remove this piece of information from the protocol. 

Group name is not affected by this change, and also any form of Authentication relying on the password is not affected either.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
@@ -39,6 +39,7 @@ public final class ConfigCheck implements IdentifiedDataSerializable {
 
     private String groupName;
 
+    @Deprecated
     private String groupPassword;
 
     private String joinerType;
@@ -98,7 +99,6 @@ public final class ConfigCheck implements IdentifiedDataSerializable {
             return false;
         }
 
-        verifyGroupPassword(found);
         verifyJoiner(found);
         verifyPartitionGroup(found);
         verifyPartitionCount(found);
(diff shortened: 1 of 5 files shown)


tkountis:
run-lab-run

pveentjer:
What is the reasoning behind this change? If this password check is removed; when is it done?

mmedenjak:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
```diff
@@ -208,7 +202,7 @@ public void writeData(ObjectDataOutput out) throws IOException {
     @Override
     public void readData(ObjectDataInput in) throws IOException {
```
mmedenjak: As mentioned [here](https://hazelcast.atlassian.net/wiki/display/EN/Prevent+Group+Password+Broadcast+Design?focusedCommentId=130520070#comment-130520070), the alternative to doing this in 3.8.x is making the change in 3.9 only and send the password if the cluster version is lower than 3.9:
```
if (out.getVersion().isLessOrEqual(Versions.V3_8)){
            out.writeUTF(groupPassword);
        }
```
```
if (in.getVersion().isLessOrEqual(Versions.V3_8)){
            in.readUTF();
        }
```

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
```diff
@@ -208,7 +202,7 @@ public void writeData(ObjectDataOutput out) throws IOException {
...
         groupName = in.readUTF();
-        groupPassword = in.readUTF();
+        in.readUTF();
```
mdogan: You're not reading group-password but it's still written to the stream and broadcasted. So, it'll be visible in multicast packets.

tkountis: Yes the plan is to stop broadcasting in 3.9. So for 3.8.2 we remove validation, to keep it backwards compatible and rolling upgrades compatible.

mdogan: I see, I've not realized that this is 3.8.x PR.

PR #10378, opened 2017-04-19
MulticastJoiner split handler broadcasts message on every round

mdogan:
MulticastJoiner split handler should broadcast a new split brain message
on every round, when it detects other side should join.

Otherwise a live-lock can happen, on every poll, a new message is received but
`shouldMerge()` decides that other side should join to this side. Node cannot
break the message poll loop and as a result cannot broadcast split brain message.

Fixes `ClientSplitBrainTest` failure:
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/1315/com.hazelcast$hazelcast-client/testReport/com.hazelcast.client/ClientSplitBrainTest/testClientListeners_InSplitBrain/history/


mdogan:
run-lab-run

PR #10385, opened 2017-04-19
Improve concurrent queues

mtopolnik:
Replace draining to a consumer with a predicate that can stop the draining at any step.

This change is needed for the next release of Jet.


mmedenjak:
run-lab-run

PR #10388, opened 2017-04-19
Implement exponential backoff for client backpressure

### LINKED ISSUE #8533: Configurable back-pressure strategy
When back-pressure is enabled:
- Members always use back-off. 
- Clients always throws `HazelcastOverloadedException`

This is inconsistent. Behaviour should be configurable on both members and clients.

jerrinot:
Fixes #8533
The default behaviour is still fail-fast. This is something we should
change in Hazelcast 4.


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/CallIdSequenceWithBackoff.java
```diff
@@ -0,0 +1,80 @@
...
+    public long next() {
+        if (!hasSpace()) {
+            waitForSpace();
+        }
+
+        return longs.incrementAndGet(INDEX_HEAD);
+    }
+
+    private void waitForSpace() {
+        if (backoffTimeoutMs <= 0) {
```
jerrinot: This is not exactly OOP code, but it's simple and perhaps easier to read then class hierarchies and what not.
I am open for suggestions tho.

hazelcast-client/src/main/java/com/hazelcast/client/spi/properties/ClientProperty.java
```diff
@@ -86,6 +86,20 @@
...
+     * <p/>
+     * If an invocation can't be made because there are too many pending invocations, then an exponential backoff is done
+     * to give the system time to deal with the backlog of invocations. This property controls how long an invocation is
+     * allowed to wait before getting a {@link com.hazelcast.core.HazelcastOverloadException}.
+     * <p/>
+     * <p>
+     * When set to -1 then <code>HazelcastOverloadException</code> is thrown immediately without any waiting.
+     * </p>
+     */
+    public static final HazelcastProperty INVOCATION_BACKOFF_TIMEOUT_MILLIS
```
jerrinot: I am open for suggestions here too

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/CallIdSequenceWithBackoff.java
```diff
@@ -0,0 +1,96 @@
+/*
```
pveentjer: Why don't we use the serverside logic for this? I would rather fix this problem once. 

Also we need to pull out the policy from the CallIdGenerator. This way we can offer the old behavior (fail immediately), but also provide an out of the box implementation.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/CallIdSequenceWithBackoff.java
pveentjer: OOP is highly overrated.

hazelcast-client/src/main/java/com/hazelcast/client/spi/properties/ClientProperty.java
pveentjer: This should be fully aligned with the server.

jerrinot: the question is whether to 1. align it with the server or 2. align it with the existing client properties.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/CallIdSequenceWithBackoff.java
jerrinot: I wanted to make a minimal possible change to deliver the feature. with the exception of pulling out the inner class. I agree there is a little reason for client to have its own interface for callId generation.

pveentjer: I would call this a hack because you copy paste the code to fix it a bit different in another place. Having shared code saves a huge amount of time.

hazelcast-client/src/main/java/com/hazelcast/client/spi/properties/ClientProperty.java
pveentjer: Since it is new for the client, I would reuse from the server.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/CallIdSequenceWithBackoff.java
jerrinot: ok. I just pushed WIP doing exactly that - the same sequence is used by both client and member sides

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
```diff
@@ -233,7 +236,12 @@ public HazelcastClientInstanceImpl(ClientConfig config,
         clusterService = new ClientClusterServiceImpl(this, addressProviders);
 
         int maxAllowedConcurrentInvocations = properties.getInteger(MAX_CONCURRENT_INVOCATIONS);
-        callIdSequence = new CallIdSequence.CallIdSequenceFailFast(maxAllowedConcurrentInvocations);
+        long backofftimeoutMs = properties.getLong(INVOCATION_BACKOFF_TIMEOUT_MILLIS);
+        if (maxAllowedConcurrentInvocations == Integer.MAX_VALUE) {
```
jerrinot: the client does not have any option to explicitly disable backpressure. max in-flight operation is set the MAX_VALUE by default thus it's effectively disabled.

hazelcast-client/src/main/java/com/hazelcast/client/spi/properties/ClientProperty.java
jerrinot: ok. I will push another commit

hazelcast/src/main/java/com/hazelcast/internal/sequence/CallIdSequenceWithBackpressure.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private boolean hasSpace() {
+        return longs.get(INDEX_HEAD) - longs.get(INDEX_TAIL) < maxConcurrentInvocations;
+    }
+
+    private void waitForSpace() throws TimeoutException {
+        long deadline = System.nanoTime() + backoffTimeoutNanos;
+        for (long idleCount = 0; ; idleCount++) {
+            if (System.nanoTime() >= deadline) {
```
jerrinot: fast exit when `backofftimeNanos <= 0`

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -112,7 +114,11 @@ public ClientMessage getClientMessage() {
 
     public ClientInvocationFuture invoke() {
         assert (clientMessage != null);
-        clientMessage.setCorrelationId(callIdSequence.next());
+        try {
+            clientMessage.setCorrelationId(callIdSequence.next());
+        } catch (TimeoutException e) {
+            throw new HazelcastOverloadException("Timed out trying to acquire another call ID.");
```
pveentjer: would be good to have indication of amount of time that has been spend waiting. Can't we include the timeoutexception as cause?

jerrinot: good one! I'll do that!

hazelcast/src/main/java/com/hazelcast/internal/sequence/CallIdSequence.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Generates the next unique call ID. When the implementation
+     * supports backpressure, it will not return unless the number of outstanding invocations is within the
+     * configured limit. Instead it will block until the condition is met and eventually throw a timeout exception.
+     *
+     * @return the generated call ID
+     * @throws TimeoutException if the outstanding invocation count hasn't dropped below the configured limit
+     * within the configured timeout
+     */
+    public abstract long next() throws TimeoutException;
```
pveentjer: To prevent package bloat in the internal package, I would throw this in com.hazelcast.internal.util.

hazelcast/src/main/java/com/hazelcast/internal/sequence/CallIdSequenceWithBackpressure.java
```diff
@@ -0,0 +1,111 @@
...
+                throw new TimeoutException(String.format("Timed out trying to acquire another call ID."
```
pveentjer: Would it be an idea to throw here the OverloadException instead of TimeoutException? This way you don't need to catch/convert it on the client. But perhaps the server side isn't too happy with this.. something which probably should be aligned?

hazelcast/src/main/java/com/hazelcast/internal/sequence/CallIdSequence.java
jerrinot: ok. I was also thinking about `spi.impl`

pveentjer:
Looks good. Looking forward to get this PR merged.

hazelcast/src/main/java/com/hazelcast/internal/sequence/CallIdSequence.java
pveentjer: That is also an option.

hazelcast/src/main/java/com/hazelcast/internal/sequence/CallIdSequenceWithBackpressure.java
jerrinot: the original implementation of `CallIdSequenceWithBackpressure` would throw the `HazelcastOverloadedException`. 

This was change to by https://github.com/hazelcast/hazelcast/pull/9307/files#diff-bb18d16905545de423a596ca3a59f48fR184

I assume the reasoning is the SequenceGenerator should not decide if Hazelcast is overloaded or not. It just indicates a timeout while trying to obtain a call id. It's up to the caller to deal with caller. btw: The member-side also catch the `TimeoutExcption` and wrap it inside `HazelcastOverloadedEx`

pveentjer:
Part of another PR should be to add randomization. Now the system is at risk that the thread that have requested a call id the longest time ago, are getting starved by not granting them a slot because they have the longest waits. With randomized exponential backoff fairness improves

jerrinot:
@pveentjer, @sancar: guys, can you please have another look? thanks!

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -117,7 +119,11 @@ public ClientMessage getClientMessage() {
 
     public ClientInvocationFuture invoke() {
         assert (clientMessage != null);
-        clientMessage.setCorrelationId(callIdSequence.next());
+        try {
```
pveentjer: Future note; would be nice of the call id sequence would immediately throw the right exception. Now there is catching noise.

hazelcast/src/main/java/com/hazelcast/spi/impl/operationservice/impl/InvocationRegistry.java
```diff
@@ -107,7 +108,7 @@ public boolean register(Invocation invocation) {
         final long callId;
         try {
             boolean force = invocation.op.isUrgent() || invocation.isRetryCandidate();
-            callId = callIdSequence.next(force);
+            callId = force ? callIdSequence.forceNext() : callIdSequence.next();
         } catch (TimeoutException e) {
```
pveentjer: Same exception caught and then converted.. Makes sense to just throw the right exception if possible. Perhaps the exception message is a pita.

hazelcast/src/main/java/com/hazelcast/spi/impl/sequence/CallIdSequence.java
```diff
@@ -0,0 +1,80 @@
...
+ * <p/>
+ * In the future we could add a CallIdSequence per partition or using some 'concurrency level'
+ * and do a mod based on the partition-id. The advantage is that you reduce contention and improve isolation,
+ * at the expense of:
+ * <ol>
+ * <li>increased complexity</li>
+ * <li>not always being able to fully utilize the number of invocations.</li>
+ * </ol>
+ */
+public abstract class CallIdSequence {
```
pveentjer: Why is this not an interface if everything is abstract?

hazelcast/src/main/java/com/hazelcast/spi/impl/sequence/CallIdSequenceWithBackpressure.java
```diff
@@ -0,0 +1,111 @@
...
+ *     <li>system operations</li>
+ *     <li>the racy nature of checking if space is available and getting the next sequence. </li>
+ * </ul>
+ * The latter cause is not a problem since the capacity is exceeded temporarily and it isn't sustainable.
+ * So perhaps there are a few threads that at the same time see that the there is space and do a next.
+ * But any following invocation needs to wait till there is is capacity.
+ */
+public final class CallIdSequenceWithBackpressure extends CallIdSequence {
+    static final int MAX_DELAY_MS = 500;
+    private static final IdleStrategy IDLER = new BackoffIdleStrategy(
```
pveentjer: AFAIK there is no randomization here. Randomization will prevent that invocations have waited for the longest amount of time are starved and eventually run into a OperationTimeoutException.

jerrinot: agreed, that's nice to have, but out of the scope here.

PR #10401, opened 2017-04-20
[BACKPORT] Improve concurrent queues

mtopolnik:
Replace draining to a consumer with a predicate that can stop the draining at any step.

Backport of a commit to master that Jet will depend on in its next release.


mtopolnik:
verify

PR #10402, opened 2017-04-20
Lite member promotion should not allowed when cluster is not active

mdogan:
See comments in the PR: https://github.com/hazelcast/hazelcast/pull/10142


mdogan:
run-lab-run

PR #10405, opened 2017-04-20
Do not retry TargetNotMemberException when invocation made on target

### LINKED ISSUE #3422: ClientLockWithTerminationTest.testLockOnClient_withNodeCrash
https://hazelcast-l337.ci.cloudbees.com/job/Hazelcast-3.x-pr-builder/com.hazelcast$hazelcast-client/10092/testReport/junit/com.hazelcast.client.lock/ClientLockWithTerminationTest/testLockOnClient_withNodeCrash/

```
Error Message

Lock was obtained by 2 different clients 
Stacktrace

java.lang.AssertionError: Lock was obtained by 2 different clients 
    at org.junit.Assert.assertFalse(Assert.java:64)
    at com.hazelcast.client.lock.ClientLockWithTerminationTest.testLockOnClient_withNodeCrash(ClientLockWithTerminationTest.java:87)
```

sancar:
Invocation made `ProxyManager.initialize(ClientProxy)` was on connection
to be able to prevent it from retrying. It was a workaround that does not
comply with ClientInvocation API hence failed today.

In this pr, I made the proxy creation invocation on address as intended.
And I changed invocation exception handling so that it will not be retried
when given address not a member anymore. Retrying does not make sense in
this case.

ClientLockWithTerminationTest.testLockOnClient_withNodeCrash was failing
because it was retrying creating proxy on closed member until
clientInvocationTimeout(120 seconds) hit and then throws OperationTimeoutException.

This pr https://github.com/hazelcast/hazelcast/pull/10370 made issue apparent.
It was waiting for 120 seconds unncessarily, then throwing a
retryableException(instead OperationTimeoutException). Because of retyable
exception there issue was hidden.

fixes #3422

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
@@ -20,17 +20,19 @@ import com.hazelcast.client.HazelcastClientNotActiveException;
 import com.hazelcast.client.connection.nio.ClientConnection;
 import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.client.impl.protocol.ClientMessage;
+import com.hazelcast.client.spi.ClientClusterService;
 import com.hazelcast.client.spi.ClientExecutionService;
 import com.hazelcast.client.spi.EventHandler;
 import com.hazelcast.core.HazelcastInstanceNotActiveException;
 import com.hazelcast.core.HazelcastOverloadException;
 import com.hazelcast.core.LifecycleService;
-import com.hazelcast.logging.ILogger;
 import com.hazelcast.core.OperationTimeoutException;
+import com.hazelcast.logging.ILogger;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
 import com.hazelcast.spi.exception.RetryableException;
 import com.hazelcast.spi.exception.TargetDisconnectedException;
+import com.hazelcast.spi.exception.TargetNotMemberException;
 
 import java.io.IOException;
 import java.util.concurrent.Executor;
@@ -53,6 +55,7 @@ public class ClientInvocation implements Runnable {
     private final ClientInvocationFuture clientInvocationFuture;
     private final ILogger logger;
     private final LifecycleService lifecycleService;
+    private final ClientClusterService clientClusterService;
     private final ClientInvocationServiceSupport invocationService;
     private final ClientExecutionService executionService;
     private final ClientMessage clientMessage;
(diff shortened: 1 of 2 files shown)


sancar:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -182,6 +184,11 @@ public void notifyException(Throwable exception) {
             return;
         }
 
+        if (address != null && exception instanceof TargetNotMemberException) {
```
ihsandemir: This will prevent all invocations which receive TargetNotMemberException being not retried even when it is actually retryable and it was being retried before. E.g. Imap.get will end up at invocationService.invokeOnTarget where address != null. Don't we want to retry these invocations?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
```diff
@@ -171,6 +172,7 @@ public void notify(ClientMessage clientMessage) {
         clientInvocationFuture.complete(clientMessage);
     }
 
+    @SuppressWarnings({ "checkstyle:cyclomaticcomplexity", "checkstyle:npathcomplexity" })
```
ihsandemir: is there no workaround not to introduce this suppression?

Donnerbart: The workaround would be to simplify the method, so less branches etc. (normally you can achieve this by pulling out methods).

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
sancar: My thinking was as following:
TargetNotMemberException is thrown when `address` is not in memberlist. It can be thrown from either server or client. 
If invocation intended to be send to an address `address != null`  it will retry always in same address. Address is not in memberlist, there is no point retrying. It does not matter if client message is retyable or not it will never be executed. 
Makes sense ? 

Note: `IMap.get` does not use `invokeOnTarget`,  it uses `invokeOnPartition`

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientInvocation.java
sancar: I am not in favor of pulling out methods when methods are not reused. But I have a pr cookinf up built on top of this. It will simplify these parts by introducing invocation classes for different targets. There I will remove back the suppressions.

PR #10398, opened 2017-04-20
Implemented Near Cache configuration to store its keys by-reference

Donnerbart:
* added `serializeKeys` to `NearCacheConfig`, `NearCacheConfigReadOnly`,
  `XmlConfigBuilder`, `XmlClientConfigBuilder` and XSD files
* added `isSerializeKeys()` to `NearCache`
* added `setSerializeKeys()` to `NearCacheTestUtils.createNearCacheConfig()`
* added on-demand key de-serialization to `RepairingHandler` and
  `NearCacheTestUtils`

PR #10407, opened 2017-04-20
Cleanup of NearCacheConfig and EvictionConfig

Donnerbart:
* fixed JavaDoc of `NearCacheConfig` and `EvictionConfig`
* re-ordered fields of `NearCacheConfig` according to reference manual

No fields or logic has been changed, just re-ordering.

Pulled out from https://github.com/hazelcast/hazelcast/pull/10398

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
@@ -34,6 +34,14 @@ import static com.hazelcast.util.Preconditions.checkPositive;
 /**
  * Configuration for eviction.
  * You can set a limit for number of entries or total memory cost of entries.
+ * <p/>
+ * The default values of the eviction configuration are
+ * <ul>
+ * <li>{@link EvictionPolicy#LRU} as eviction policy</li>
+ * <li>{@link EvictionConfig.MaxSizePolicy#ENTRY_COUNT} as max size policy</li>
+ * <li>{@value DEFAULT_MAX_ENTRY_COUNT_FOR_ON_HEAP_MAP} as maximum size for on-heap {@link com.hazelcast.core.IMap}</li>
+ * <li>{@value DEFAULT_MAX_ENTRY_COUNT} as maximum size for all other data structures and configurations</li>
+ * </ul>
  */
 @BinaryInterface
 public class EvictionConfig implements EvictionConfiguration, DataSerializable, Serializable {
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -75,55 +111,27 @@ public EvictionConfig() {
     }
 
     public EvictionConfig(int size, MaxSizePolicy maxSizePolicy, EvictionPolicy evictionPolicy) {
-        /**
```
Donnerbart: It's not common proxies to use setters in a constructor. So there is actually no need for a dangeling JavaDoc comment (it doesn't belong in a method body), which creates an IDEA warning. Also no other config classes beside the copy&paste `NearCachePreloaderConfig` has those comments.

hazelcast/src/main/java/com/hazelcast/config/NearCacheConfig.java
```diff
@@ -164,31 +163,32 @@ public NearCacheConfig(int timeToLiveSeconds, int maxSize, String evictionPolicy
...
-        name = config.getName();
-        evictionPolicy = config.getEvictionPolicy();
-        inMemoryFormat = config.getInMemoryFormat();
-        invalidateOnChange = config.isInvalidateOnChange();
-        maxIdleSeconds = config.getMaxIdleSeconds();
-        maxSize = config.getMaxSize();
-        timeToLiveSeconds = config.getTimeToLiveSeconds();
-        cacheLocalEntries = config.isCacheLocalEntries();
-        localUpdatePolicy = config.localUpdatePolicy;
+        name = config.name;
```
Donnerbart: Change the copy constructor to use fields instead of getters. This reduces the amount of code and doesn't change the logic.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -58,6 +66,34 @@
      */
     public static final EvictionPolicy DEFAULT_EVICTION_POLICY = EvictionPolicy.LRU;
 
+    /**
+     * Maximum Size Policy
```
pveentjer: This documentation doesn't provide anything what the class name already provided :) So can you add some extra documentation behind the purpose of this policy?

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
```diff
@@ -254,6 +261,52 @@ public EvictionPolicyType getEvictionPolicyType() {
...
+
+    /**
+     * Sets the class name of the configured {@link EvictionPolicyComparator} implementation.
+     * <p/>
+     * Only one of the {@code comparator class name} and {@code comparator} can be configured in the eviction configuration.
+     *
+     * @param comparatorClassName the class name of the configured {@link EvictionPolicyComparator} implementation
+     * @return this EvictionConfig instance
+     */
+    public EvictionConfig setComparatorClassName(String comparatorClassName) {
```
pveentjer: Probably good to document the null argument and its behavior.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
Donnerbart: Originally I didn't want to touch the `EvictionConfig` at all. I just started with it since I moved some JavaDoc from the `NearCacheConfig` over here. I'll put your comments on the TODO. The `NearCachePreloaderConfig` needs an overhaul as well.

hazelcast/src/main/java/com/hazelcast/config/EvictionConfig.java
Donnerbart: To be honest, I have to look up the code to see what this does. It's on my TODO. I think it's also not documented in the ref manual, at least not in the Near Cache chapter (maybe somewhere in the IMap or JCache eviction).

PR #10415, opened 2017-04-21
Temporarily persist joined member uuids

metanet:
When a member joins and leaves the cluster very quickly, a stale join request can be processed by master and the member is added back to the cluster by mistake.

With this change, we temporarily persist the member uuids that has joined in the 'joinable' cluster state so that stale join requests will be ignored. If the cluster is not in 'joinable' state, previous members can leave the cluster and join back with hot restart.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
@@ -29,10 +29,10 @@ import com.hazelcast.internal.cluster.impl.operations.ConfigMismatchOp;
 import com.hazelcast.internal.cluster.impl.operations.FinalizeJoinOp;
 import com.hazelcast.internal.cluster.impl.operations.GroupMismatchOp;
 import com.hazelcast.internal.cluster.impl.operations.JoinRequestOp;
-import com.hazelcast.internal.cluster.impl.operations.WhoisMasterOp;
+import com.hazelcast.internal.cluster.impl.operations.MasterResponseOp;
 import com.hazelcast.internal.cluster.impl.operations.MembersUpdateOp;
 import com.hazelcast.internal.cluster.impl.operations.PostJoinOp;
-import com.hazelcast.internal.cluster.impl.operations.MasterResponseOp;
+import com.hazelcast.internal.cluster.impl.operations.WhoisMasterOp;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
```diff
@@ -335,6 +341,18 @@ private boolean checkClusterStateBeforeJoin(Address target, String uuid) {
         return true;
     }
 
+    private void cleanupRecentlyJoinedMemberUuids() {
+        long currentTime = System.currentTimeMillis();
+        long staleJoinPreventionDuration = node.getProperties().getMillis(GroupProperty.MAX_JOIN_SECONDS);
```
mmedenjak: This can be looked up at initialisation

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterJoinManager.java
```diff
@@ -335,6 +341,18 @@ private boolean checkClusterStateBeforeJoin(Address target, String uuid) {
         return true;
     }
 
+    private void cleanupRecentlyJoinedMemberUuids() {
+        long currentTime = System.currentTimeMillis();
```
mmedenjak: `System.currentTimeMillis` or `Clock#currentTimeMillis`? Probably doesn't matter though.

hazelcast/src/test/java/com/hazelcast/internal/cluster/impl/MembershipUpdateTest.java
```diff
@@ -573,6 +574,21 @@ public void run() throws Exception {
...
     }
 
+    @Test
+    public void shouldNotProcessStaleJoinRequest() {
+        HazelcastInstance hz1 = factory.newHazelcastInstance();
+        HazelcastInstance hz2 = factory.newHazelcastInstance();
+
+        JoinRequest staleJoinReq = getNode(hz2).createJoinRequest(true);
+        hz2.shutdown();
+        assertClusterSize(1, hz1);
```
mdogan: `assertClusterSizeEventually()` ?

metanet: do we need eventual assertion? once hz2 completes shutdown, don't we guarantee that master node removes it from the member list.

Issue #9335, opened 2016-11-28
CacheSplitBrainTest Failures


hasancelik:
```
java.lang.AssertionError: CountDownLatch failed to complete within 120 seconds , count left: 1
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:820)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:813)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:805)
	at com.hazelcast.cache.merge.CacheSplitBrainTest.testPassThroughCacheMergePolicy(CacheSplitBrainTest.java:260)
```

https://hazelcast-l337.ci.cloudbees.com/view/Sonar/job/Hazelcast-3.x-sonar/1302/#showFailuresLink

vbekiaris:
Fixed by #9328

mmedenjak:
**com.hazelcast.cache.merge.CacheSplitBrainTest.testSplitBrain[format:BINARY, mergePolicy:class com.hazelcast.spi.merge.PassThroughMergePolicy]**
**com.hazelcast.cache.merge.CacheSplitBrainTest.testSplitBrain[format:BINARY, mergePolicy:class com.hazelcast.spi.merge.PutIfAbsentMergePolicy]**

```
java.lang.AssertionError: expected:<PutIfAbsentValue2> but was:<null>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at com.hazelcast.cache.merge.CacheSplitBrainTest.afterMergePutIfAbsentMergePolicy(CacheSplitBrainTest.java:362)
	at com.hazelcast.cache.merge.CacheSplitBrainTest.onAfterSplitBrainHealed(CacheSplitBrainTest.java:182)
```

Donnerbart:
**com.hazelcast.cache.merge.CacheSplitBrainTest.testSplitBrain[format:BINARY, mergePolicy:class com.hazelcast.spi.merge.PassThroughMergePolicy]**
```java
[...]

PR #10416, opened 2017-04-21
Eliminate unconditional waiting in node join/shutdown

mmedenjak:
When joining and shutting the node down there are some places where we
unconditionally wait before checking if we should wait any longer.
This has been eliminated, the check is done before the first wait as
well. 

NOTE : 
One risky change from an older version of this PR is already in master and now I see this PR as low-risk as the remaining changes are simple to reason about.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
@@ -178,20 +178,12 @@ public abstract class AbstractJoiner implements Joiner {
     }
 
     private void ensureConnectionToAllMembers() {
-        boolean allConnected = false;
         if (clusterService.isJoined()) {
             logger.fine("Waiting for all connections");
             int connectAllWaitSeconds = node.getProperties().getSeconds(GroupProperty.CONNECT_ALL_WAIT_SECONDS);
             int checkCount = 0;
-            while (checkCount++ < connectAllWaitSeconds && !allConnected) {
-                try {
-                    //noinspection BusyWait
-                    TimeUnit.SECONDS.sleep(1);
-                } catch (InterruptedException ignored) {
-                    EmptyStatement.ignore(ignored);
-                }
-
-                allConnected = true;
+            while (checkCount++ < connectAllWaitSeconds) {
+                boolean allConnected = true;
                 Collection<Member> members = clusterService.getMembers();
                 for (Member member : members) {
                     if (!member.localMember() && node.connectionManager.getOrConnect(member.getAddress()) == null) {
@@ -200,6 +192,15 @@ public abstract class AbstractJoiner implements Joiner {
                             logger.fine("Not-connected to " + member.getAddress());
                         }
                     }
+                }
+                if (allConnected) {
+                    break;
+                }
+                try {
+                    //noinspection BusyWait
+                    TimeUnit.SECONDS.sleep(1);
+                } catch (InterruptedException ignored) {
+                    EmptyStatement.ignore(ignored);
                 }
             }
         }
--- a/hazelcast/src/test/java/com/hazelcast/test/mocknetwork/MockJoiner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/mocknetwork/MockJoiner.java
@@ -50,7 +50,7 @@ class MockJoiner extends AbstractJoiner {
 
         Address previousJoinAddress = null;
         long joinAddressTimeout = 0;
-        while (shouldRetry() && (Clock.currentTimeMillis() - joinStartTime < maxJoinMillis)) {
+        while (Clock.currentTimeMillis() - joinStartTime < maxJoinMillis) {
             synchronized (registry) {
                 Address joinAddress = getJoinAddress();
                 verifyInvariant(joinAddress != null, "joinAddress should not be null");
@@ -78,6 +78,9 @@ class MockJoiner extends AbstractJoiner {
                     joinAddressTimeout = 0;
                     clusterService.setMasterAddressToJoin(null);
                 }
+            }
+            if (!shouldRetry()) {
+                break;
             }
             try {
                 Thread.sleep(500);


mmedenjak:
run-lab-run

mmedenjak:
Running failing tests on oracle JDK 50 times each : https://hazelcast-l337.ci.cloudbees.com/job/matko-test-jfr/ : 
`GracefulShutdownCorrectnessTest`, `MembershipUpdateTest`, `MockJoinerTest`, `TcpIpJoinTest`

mmedenjak:
@mdogan @metanet can we try and merge this one again?

mdogan:
@mmedenjak: we can try. let's first run in PR builder a few times.

mdogan:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
The failed runs could be because this enhancement eliminates waiting, thus revealing race conditions. We also have some tests which are already spuriously failing in master. The failed runs should be inspected.

Failed runs : 
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/10037/

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
Failed: https://github.com/hazelcast/hazelcast/issues/9335

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
run-lab-run

mmedenjak:
Without PR:
Success	#13795	39 min
Success	#13792	35 min
Success	#13791	40 min
Success	#13790	33 min
Success	#13789	34 min
Success	#13786	39 min
Success	#13782	36 min
Success	#13781	38 min
Success	#13780	35 min
Success	#13777	35 min
Success	#13776	37 min
Success	#13775	35 min
Success	#13774	34 min
Success	#13772	34 min
Success	#13769	35 min
Success	#13765	34 min
Success	#13764	35 min
Success	#13763	34 min
Success	#13761	34 min
Success	#13760	34 min
Success	#13759	37 min
Success	#13756	34 min
Success	#13755	34 min
Success	#13753	33 min
Success	#13752	33 min
Success	#13751	35 min
Success	#13750	35 min
Success	#13749	34 min
Success	#13748	35 min

With PR:
Success	#13793	31 min
Success	#13797	35 min
Success	#13800	31 min
Success	#13803	29 min

mmedenjak:
@mdogan I've come to realise that one change from an older version of this PR is already in master for a long time. So now the only two remaining changes are in eliminating some waiting in joiner implementations. I think this is now low-risk. I also ran the builder a couple of times and there were no fishy failures. Please take a look.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
```diff
@@ -201,7 +194,15 @@ private void ensureConnectionToAllMembers() {
                         }
                     }
                 }
-            }
+                if (!allConnected) {
```
Donnerbart: We could do a 
```java
if (allConnected) {
    break;
}
```
here to ease the boolean logic of `sleep` and `while`. We could even keep the old `while-do` loop with it, which is IMO a bit easier to read.

Donnerbart: We could even put the `boolean allConnected` completely in the loop:
```java
            while (checkCount++ < connectAllWaitSeconds) {
                boolean allConnected = true;
                Collection<Member> members = clusterService.getMembers();
                for (Member member : members) {
                    if (!member.localMember() && node.connectionManager.getOrConnect(member.getAddress()) == null) {
                        allConnected = false;
                        if (logger.isFineEnabled()) {
                            logger.fine("Not-connected to " + member.getAddress());
                        }
                    }
                }
                if (allConnected) {
                    break;
                }
                try {
                    //noinspection BusyWait
                    TimeUnit.SECONDS.sleep(1);
                } catch (InterruptedException ignored) {
                    EmptyStatement.ignore(ignored);
                }
            }
```

hazelcast/src/test/java/com/hazelcast/test/mocknetwork/MockJoiner.java
```diff
@@ -50,7 +50,8 @@ public void doJoin() {
 
         Address previousJoinAddress = null;
         long joinAddressTimeout = 0;
-        while (shouldRetry() && (Clock.currentTimeMillis() - joinStartTime < maxJoinMillis)) {
+        boolean shouldRetry;
```
Donnerbart: Similar thing here:
```java
while (Clock.currentTimeMillis() - joinStartTime < maxJoinMillis) {
    synchronized (registry) {
        Address joinAddress = getJoinAddress();
        verifyInvariant(joinAddress != null, "joinAddress should not be null");

        if (!joinAddress.equals(previousJoinAddress)) {
            previousJoinAddress = joinAddress;
            joinAddressTimeout = Clock.currentTimeMillis() + JOIN_ADDRESS_TIMEOUT_IN_MILLIS;
        }

        if (node.getThisAddress().equals(joinAddress)) {
            logger.fine("This node is found as master, no need to join.");
            clusterJoinManager.setThisMemberAsMaster();
            break;
        }

        logger.fine("Sending join request to " + joinAddress);
        if (!clusterJoinManager.sendJoinRequest(joinAddress, true)) {
            logger.fine("Could not send join request to " + joinAddress);
            clusterService.setMasterAddressToJoin(null);
        }

        if (Clock.currentTimeMillis() > joinAddressTimeout) {
            logger.warning("Resetting master address because join address timeout");
            previousJoinAddress = null;
            joinAddressTimeout = 0;
            clusterService.setMasterAddressToJoin(null);
        }
    }
    if (!shouldRetry()) {
        break;
    }
    try {
        Thread.sleep(100);
        e.printStackTrace();
        break;
    }
}
```

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/AbstractJoiner.java
mmedenjak: The `do-while`-> `while` looks good and seems better but it changes the behaviour in the pathological case when someone configures hazelcast with `GroupProperty#CONNECT_ALL_WAIT_SECONDS` set to 0. In this case, we don't print out to which members we are not connected to. Same thing with the `MockJoiner`, if we set `GroupProperty#MAX_JOIN_SECONDS` to 0, we wouldn't perform the logic in the loop at all.
@mdogan WDYT?

mmedenjak:
run-lab-run

mmedenjak:
@Donnerbart @mdogan addressed review, the logic is now changed to from a `do-while` to a `while` loop. Please take a look.

mmedenjak:
@Donnerbart @mdogan thanks for the reviews!

PR #10423, opened 2017-04-24
[DONT MERGE] Prototype for client class load deployment

sancar:
Makes use of listener to request class loading from clients.

ClientUserCodeDeploymentSmokeTest this test is passing on my local.
It will fail in Pr since protocol jar needs to be updated first.


sancar:
@jerrinot 
When I run `testUserCodeDeploymentIsDisabledByDefault` test, the method `fetchBytecodeFromRemote` called twice before throwing exception. Is that expected ?

PR #10426, opened 2017-04-25
Fixes query performance regression for native

tombujok:
before regression:
3.7.0, 10B, -> 19.5k ops

with regression:
3.8.0, 10B, ->  5.1k ops

after the fix:
3.9.0, 10B, -> 21.7k ops


tombujok:
Still some stuff I need to check here.

PR #10427, opened 2017-04-25
[BACKPORT] Fixes query performance degredation for native

tombujok:
Backport of #10426 (OS side)

```
before regression:
3.7.0, 10B, -> 19.5k ops

with regression:
3.8.0, 10B, -> 5.1k ops

after the fix:
3.9.0, 10B, -> 21.7k ops
```

The problem was the each partitionOperation for NATIVE was a remote one.


pveentjer:
Since this just makes a few methods accessible, I don't see a reason to add another set of eyes.

PR #10432, opened 2017-04-26
Rename DistributedObjectFuture isSet to isSetAndInitialized

mdogan:
Uninitialized means DistributedObject is not retrieved from future yet.

Fix for bug created by https://github.com/hazelcast/hazelcast/pull/10428


mdogan:
run-lab-run

PR #10434, opened 2017-04-26
Rename ReplicaFragmentNamespace to ServiceNamespace

mdogan:
ObjectNamespace now extends ServiceNamespace. DefaultObjectNamespace is deprecated
and replaced by DistributedObjectNamespace. Reason is to make it IDS.

Depends on https://github.com/hazelcast/hazelcast/pull/10356

PR #10435, opened 2017-04-26
ServiceNamespace can belong to multiple services

mdogan:
For example, locks can belong to lock service itself or map or multimap services.
Partition data belonging to a namespace should be replicated together.

FragmentedMigrationAwareService can return null namespaces.

Depends on https://github.com/hazelcast/hazelcast/pull/10434

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/partition/operation/AbstractPartitionOperation.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/partition/operation/AbstractPartitionOperation.java
@@ -30,6 +30,7 @@ import com.hazelcast.spi.impl.servicemanager.ServiceInfo;
 import java.util.ArrayList;
 import java.util.Collection;
 
+import static java.util.Collections.emptySet;
 import static java.util.Collections.singleton;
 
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/internal/partition/operation/AbstractPartitionOperation.java
```diff
@@ -67,16 +68,51 @@
...
+            Operation op = service.prepareReplicationOperation(event, singleton(ns));
+            if (op == null) {
+                continue;
+            }
+
+            op.setServiceName(serviceInfo.getName());
+
+            if (operations.isEmpty()) {
+                // generally a namespace belongs to a single service only
+                operations = singleton(op);
```
vbekiaris: minor: why not just initialize a new `ArrayList` here and remove the next condition branch?

mdogan: My reasoning is, to avoid `ArrayList` creating, if a namespace belongs to a single service. Currently only lock service shares namespaces with map and multimap.

PR #10440, opened 2017-04-26
[DO NOT MERGE] Off-heap index per partition

tombujok:


tombujok:
run-lab-run

tombujok:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
```diff
@@ -92,32 +96,89 @@ public Result runIndexOrPartitionScanQueryOnOwnedPartitions(Query query)
         if (entries != null) {
```
mmedenjak: Can you use `populateResult(query, initialPartitions, entries);`?

tkountis:
Re-based version here https://github.com/hazelcast/hazelcast/pull/10720

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryPartitionOperationFactory.java
```diff
@@ -0,0 +1,70 @@
...
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.spi.Operation;
+import com.hazelcast.spi.impl.operationservice.impl.operations.PartitionAwareOperationFactory;
+
+import java.io.IOException;
+import java.util.Collection;
+
+import static com.hazelcast.util.CollectionUtil.toIntArray;
+
+public class QueryPartitionOperationFactory extends PartitionAwareOperationFactory {
```
tombujok: Not anymore. Good catch! Deleted.

hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
```diff
@@ -111,6 +118,13 @@ private RecordStore createRecordStore(String name) {
         keyLoader.setMaxSize(getMaxSizePerNode(mapConfig.getMaxSizeConfig()));
         keyLoader.setHasBackup(mapConfig.getTotalBackupCount() > 0);
         keyLoader.setMapOperationProvider(serviceContext.getMapOperationProvider(name));
+
+        InternalSerializationService ss = (InternalSerializationService) nodeEngine.getSerializationService();
+        IndexProvider indexProvider = serviceContext.getIndexProvider(mapConfig);
+        final boolean globalIndex = false;
+        Indexes indexesForMap = new Indexes(ss, indexProvider, mapContainer.getExtractors(), globalIndex);
+        indexes.putIfAbsent(name, indexesForMap);
```
tombujok: You were right. This is fixed indirectly by other commit. Thx anyway!

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryPartitionOperation.java
```diff
@@ -41,15 +39,9 @@ public QueryPartitionOperation(Query query) {
...
+    public void run() throws Exception {
         QueryRunner queryRunner = mapServiceContext.getMapQueryRunner(getName());
-        try {
-            result = queryRunner.runPartitionScanQueryOnGivenOwnedPartition(query, getPartitionId());
-        } catch (ExecutionException e) {
-            throw new HazelcastException(e);
-        } catch (InterruptedException e) {
-            throw new HazelcastException(e);
-        }
+        result = queryRunner.runPartitionScanQueryOnGivenOwnedPartition(query, getPartitionId());
```
tombujok: They are not thrown anymore, the signatures of upper methods were wrong.

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -76,6 +80,10 @@
     protected final InterceptorRegistry interceptorRegistry = new InterceptorRegistry();
     protected final IFunction<Object, Data> toDataFunction = new ObjectToData();
     protected final ConstructorFunction<Void, RecordFactory> recordFactoryConstructor;
+    protected final boolean globalIndex;
```
tombujok: 👍

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -178,8 +194,31 @@ private PartitioningStrategy createPartitioningStrategy() {
...
+     * @return
+     */
+    public Indexes getIndexes(int partitionId) {
+        if (globalIndex) {
+            return indexes;
+        }
+        return mapServiceContext.getPartitionContainer(partitionId).getIndexes(name);
+    }
+
+    public boolean isGlobalIndex() {
```
tombujok: Not anymore :)

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MapReplicationStateHolder.java
```diff
@@ -109,6 +111,14 @@ void applyState() {
                 recordStore.reset();
                 recordStore.setPreMigrationLoadedStatus(loaded.get(mapName));
 
+                MapContainer mapContainer = recordStore.getMapContainer();
```
tombujok: It's the replication. The addIndex operation is executed on the main replica only. With global index, the indexes are in the MapContainer anyway, so each replica can access this knowledge. With index-per-partition, the replica has no way of knowing which indexes have been added in runtime.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
tombujok: Fixed 👍

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PostJoinMapOperation.java
```diff
@@ -57,14 +58,16 @@ public String getServiceName() {
...
     }
 
-    public void addMapIndex(MapContainer mapContainer) {
-        final Indexes indexes = mapContainer.getIndexes();
-        if (indexes.hasIndex()) {
-            MapIndexInfo mapIndexInfo = new MapIndexInfo(mapContainer.getName());
-            for (Index index : indexes.getIndexes()) {
-                mapIndexInfo.addIndexInfo(index.getAttributeName(), index.isOrdered());
+    public void addMapIndex(MapServiceContext mapServiceContext, MapContainer mapContainer) {
+        for (PartitionContainer partitionContainer : mapServiceContext.getPartitionContainers()) {
```
tombujok: Fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/operation/PostJoinMapOperation.java
```diff
@@ -136,10 +139,13 @@ public void run() throws Exception {
         MapService mapService = getService();
         MapServiceContext mapServiceContext = mapService.getMapServiceContext();
         for (MapIndexInfo mapIndex : indexInfoList) {
-            final MapContainer mapContainer = mapServiceContext.getMapContainer(mapIndex.mapName);
-            final Indexes indexes = mapContainer.getIndexes();
+            MapContainer mapContainer = mapServiceContext.getMapContainer(mapIndex.mapName);
             for (MapIndexInfo.IndexInfo indexInfo : mapIndex.lsIndexes) {
-                indexes.addOrGetIndex(indexInfo.attributeName, indexInfo.ordered);
+                mapContainer.addIndexDefinition(indexInfo.attributeName, indexInfo.ordered);
+                for (PartitionContainer partitionContainer : mapServiceContext.getPartitionContainers()) {
```
tombujok: Fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/MapSplitBrainHandlerService.java
```diff
@@ -82,9 +81,8 @@ public Runnable prepareMergeRunnable() {
                 }
                 // clear all records either owned or backup
                 recordStore.reset();
+                mapContainer.getIndexes(i).clearIndexes();
```
tombujok: Yeah, here I kind of thought the code is simpler like this, and the overhead is not super significant.

hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
```diff
@@ -225,4 +239,29 @@ public void setLastCleanupTimeCopy(long lastCleanupTimeCopy) {
...
+            MapServiceContext mapServiceContext = mapService.getMapServiceContext();
+            MapContainer mapContainer = mapServiceContext.getMapContainer(name);
+
+            InternalSerializationService ss = (InternalSerializationService)
+                    mapServiceContext.getNodeEngine().getSerializationService();
+            Extractors extractors = mapServiceContext.getMapContainer(name).getExtractors();
+            IndexProvider indexProvider = mapServiceContext.getIndexProvider(mapContainer.getMapConfig());
+            Indexes indexesForMap = new Indexes(ss, indexProvider, extractors, globalIndex);
+            ixs = indexes.putIfAbsent(name, indexesForMap);
+            if (ixs == null) {
```
tombujok: Right now not, since the partitioned-index is used in off-heap only (which uses partition-threading). but the implementation is more generic, and initially in this PR the on-heap indexes were partitioned too. If it wasn't for performance overhead it would stay like this. I would prefer to keep the implementation a bit more generic, since the overhead is zero (it's just the first access), but in this way it may be extended later.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
```diff
@@ -231,4 +246,26 @@ private void collectResultsFromMissingPartitions(BitSet finishedPartitions, Coll
...
+            return new QueryOperation(query);
+        }
+    }
+
+    // for rolling-upgrade compatibility, can be deleted in 4.0
+    private QueryPartitionOperation createQueryPartitionOperation(Query query, Version clusterVersion,
+                                                                  MapServiceContext mapServiceContext) {
+        boolean isVersion39orGreater = clusterVersion.isGreaterOrEqual(Versions.V3_9);
+        if (isVersion39orGreater) {
+            return (QueryPartitionOperation) mapServiceContext.getMapOperationProvider(
```
tombujok: Fixed.

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -288,14 +327,22 @@ public InterceptorRegistry getInterceptorRegistry() {
...
     public ObjectNamespace getObjectNamespace() {
         return objectNamespace;
     }
 
+    public void addIndexDefinition(String name, Boolean ordered) {
+        indexDefinitions.put(name, ordered);
+    }
+
+    public Set<Map.Entry<String, Boolean>> getIndexDefinitionsEntrySet() {
+        return Collections.unmodifiableSet(indexDefinitions.entrySet());
```
tombujok: Fixed 👍

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
```diff
@@ -68,22 +70,24 @@ public QueryRunner(MapServiceContext mapServiceContext, QueryOptimizer optimizer
...
     public Result runIndexOrPartitionScanQueryOnOwnedPartitions(Query query)
             throws ExecutionException, InterruptedException {
-
         int migrationStamp = getMigrationStamp();
         Collection<Integer> initialPartitions = mapServiceContext.getOwnedPartitions();
-
         MapContainer mapContainer = mapServiceContext.getMapContainer(query.getMapName());
 
+        // to optimize the query we need to get any index instance
+        Indexes indexes = mapContainer.getIndexes();
```
tombujok: My thinking was that we avoid allocating the iterator for the path, which is most common (and fastest).

hazelcast/src/main/java/com/hazelcast/map/impl/operation/MapReplicationStateHolder.java
mmedenjak: Ah, you mean `PostJoinMapOperation` replicates all indexes (global and per-partition) when joining and `AddIndexOperation` adds them when the cluster is stable (but only on partition owner). This class replicates them after they've been added on the partition owner to the backup replicas?

tombujok: Yep.

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
tombujok: Hmm, this one actually won't work when the index is not global. I had to change it to:
```
 Map<String, Boolean> definitions = new HashMap<String, Boolean>();
        if (isGlobalIndexEnabled()) {
            for (Index index : globalIndexes.getIndexes()) {
                definitions.put(index.getAttributeName(), index.isOrdered());
            }
        } else {
            for (PartitionContainer container : mapServiceContext.getPartitionContainers()) {
                for (Index index : container.getIndexes(name).getIndexes()) {
                    definitions.put(index.getAttributeName(), index.isOrdered());
                }
            }
        }
        return definitions;
```

PR #10463, opened 2017-05-02
Removed superfluous implements in AbstractCacheProxyBase

Donnerbart:
`PrefixedDistributedObject` is already implemented via `ICacheInternal`

This small PR helps to clean up our class proxy class hierarchy a bit by removing a superfluous line in this graph:
proxies


pveentjer:
:+1:

PR #10465, opened 2017-05-02
Moved socket configure to IOService

pveentjer:
This moves the logic of configuring a socket in a more
logical place. TcpIpConnectionManager is meant for more
high level tasks; not low level socket configuration.


pveentjer:
run-lab-run

PR #10468, opened 2017-05-02
Moved TcpIpConnectionManager.getIoBalancer to test

pveentjer:
Method has no added value otherwise; so lets add it to the single test that
uses this method.


hazelcast/src/test/java/com/hazelcast/internal/networking/nonblocking/iobalancer/IOBalancerMemoryLeakTest.java
```diff
@@ -62,18 +66,23 @@ public void testMemoryLeak_with_RestConnections() throws IOException {
...
                 int outHandlerSize = ioBalancer.getOutLoadTracker().getHandlers().size();
-                Assert.assertEquals(0, inHandlerSize);
-                Assert.assertEquals(0, outHandlerSize);
+                assertEquals(0, inHandlerSize);
+                assertEquals(0, outHandlerSize);
             }
         });
     }
 
+    private static IOBalancer getIoBalancer(TcpIpConnectionManager connectionManager) {
```
Donnerbart: I think it would match our common pattern to move the static method to the end of the class.

pveentjer: sure. NP

PR #10476, opened 2017-05-02
TcpIpAcceptor cleanup

pveentjer:
TcpIpAcceptor has now an internal thread instead of being a thread it self.

Some minor tweaks:
- renamed alive field to stop to allign with regular io threads
- trapped all throwables in the run method so we don't loose exceptions


pveentjer:
run-lab-run

pveentjer:
run-lab-run

Issue #10410, opened 2017-04-21
ClientRegressionWithMockNetworkTest failures


mmedenjak:
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/1319/

com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenSmartClientAndOwnerNodeDiesTogether

```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenClientAndOwnerNodeDiesTogether(ClientRegressionWithMockNetworkTest.java:534)
	at com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenSmartClientAndOwnerNodeDiesTogether(ClientRegressionWithMockNetworkTest.java:520)
```

com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenDummyClientAndOwnerNodeDiesTogether

```
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenClientAndOwnerNodeDiesTogether(ClientRegressionWithMockNetworkTest.java:534)
	at com.hazelcast.client.ClientRegressionWithMockNetworkTest.testLock_WhenDummyClientAndOwnerNodeDiesTogether(ClientRegressionWithMockNetworkTest.java:514)
```

mmedenjak:
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/1321

mmedenjak:
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/com.hazelcast$hazelcast-client/1322/

mmedenjak:
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-nightly/1325/

mmedenjak:
[...]

PR #10479, opened 2017-05-03
Mock network member should not use local member of another member

sancar:
Partially fixes https://github.com/hazelcast/hazelcast/issues/10410


hazelcast/src/test/java/com/hazelcast/quorum/PartitionedCluster.java
```diff
@@ -200,21 +201,21 @@ private void splitCluster() {
         cm5.block(n2.address);
         cm5.block(n3.address);
 
-        n4.clusterService.suspectMember(n1.getLocalMember(), null, true);
-        n4.clusterService.suspectMember(n2.getLocalMember(), null, true);
-        n4.clusterService.suspectMember(n3.getLocalMember(), null, true);
+        suspectMember(n4, n1);
```
mdogan: You should also update https://github.com/hazelcast/hazelcast-enterprise/blob/master/hazelcast-enterprise/src/test/java/com/hazelcast/quorum/NativeMemPartitionedCluster.java in EE repo after this gets merged.

PR #10483, opened 2017-05-03
Fixed spelling of Id, id and -id to ID

Donnerbart:
We have a very inconsistent use of ID in the comments and JavaDoc. This
 makes the existing documentation hard to read. It also makes it hard to
 decide how to write new documentation.

This commit unifies the usage to the style of the reference manual. It
has been chosen since ID and IDs is easier to distinct from it and its.

Old style:
* this id identifies
* all the ids in its chunk
* collects its ids if it's done
* `@param` id id of the interceptor

New style:
* this ID identifies
* all the IDs in its chunk
* collects its IDs if it's done
* `@param` id ID of the interceptor


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

PR #10486, opened 2017-05-04
Ordered some fields in TcpIpConnectionManager

pveentjer:


hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
```diff
@@ -121,7 +116,13 @@ public TcpIpConnectionErrorHandler createNew(Address endpoint) {
     private final ScheduledExecutorService scheduler
             = new ScheduledThreadPoolExecutor(4, new ThreadFactoryImpl("TcpIpConnectionManager-thread-"));
 
-    private TcpIpConnector tcpIpConnector;
```
pveentjer: renamed to connector, since we use acceptor and not tcpIpAcceptor.

PR #10488, opened 2017-05-04
Removed TcpIpConnection socket configure methods

pveentjer:
The SocketReader/Writer-Initializer now access the socketchannel directly to configure it.
This way the TcpIpConnection can focus on higher level functions and the localization of the
socket configuration is improved (the socket and bytebuffers are now configured in the SocketReader/Writer-Initializer)

No Enterprise PR needed.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
+++ b/hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
@@ -36,7 +36,6 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.net.SocketAddress;
-import java.net.SocketException;
 import java.nio.channels.CancelledKeyException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
```diff
@@ -187,29 +186,16 @@ public void setErrorHandler(TcpIpConnectionErrorHandler errorHandler) {
         this.errorHandler = errorHandler;
     }
 
-    public TcpIpConnectionErrorHandler getErrorHandler() {
```
pveentjer: Method isn't used.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
```diff
@@ -187,29 +186,16 @@ public void setErrorHandler(TcpIpConnectionErrorHandler errorHandler) {
...
-    }
-
-    public void setReceiveBufferSize(int size) throws SocketException {
-        socketChannel.socket().setReceiveBufferSize(size);
-    }
-
     @Override
     public boolean isClient() {
         ConnectionType t = type;
-        return (t != null) && t != ConnectionType.NONE && t.isClient();
```
pveentjer: Removed brackets since not needed.

pveentjer:
run-lab-run

PR #10495, opened 2017-05-04
Client directBuffer support.

pveentjer:
For TLS/OpenSSL client needs to support direct buffers.

There are only 2 reliances on the byte-array from the regular heap ByteBuffer.
- determine the size of the frame
- copy the payload from the socket buffer into the ClientMessage payload.

Both these have been addressed as part of this PR.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
@@ -328,7 +328,9 @@ public class ClientMessage
                 // we don't have even the frame length ready
                 return false;
             }
-            frameLength = Bits.readIntL(src.array(), src.position());
+            frameLength = Bits.readIntL(src);
+            // we need to restore the position; as if we didn't read the frame-length
+            src.position(src.position() - Bits.INT_SIZE_IN_BYTES);
             if (frameLength < HEADER_SIZE) {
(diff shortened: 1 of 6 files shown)


hazelcast/src/main/java/com/hazelcast/client/impl/protocol/ClientMessage.java
```diff
@@ -328,7 +328,9 @@ public boolean readFrom(ByteBuffer src) {
                 // we don't have even the frame length ready
                 return false;
             }
-            frameLength = Bits.readIntL(src.array(), src.position());
+            frameLength = Bits.readIntL(src);
```
asimarslan: what about directly reading from byteBuffer and get rid of `Bits.readIntL(ByteBuffer)`

```
            src.order(ByteOrder.LITTLE_ENDIAN);
            frameLength = src.getInt(src.position());
```

pveentjer: this would involve storing the older order and resetting it after the 'getInt' completes. I think that is more complex than the current approach.

asimarslan: It's the only place you read from the provided byteBuffer so you don't need to store and restore any endianness state. But if you don't like it we can keep it, it's a minor thing

PR #10496, opened 2017-05-04
[DONT MERGE] Validate & init max-size when eviction is configured

### LINKED ISSUE #10179: Eviction and max-size less then 271 cause map to never populate
This is my config
<map name="indiciescache">
        <max-size>8</max-size>
        <eviction-policy>LFU</eviction-policy>
    </map>

While I understand you CANNOT set the max-size below the partition count which is 271 by default Hazel should handle this and if max-size is lower then partition-count THEN set it to the partition count.  

What happens is nothing ever makes it into the map. This is confusing behavior.  The config is not rejected meaning no errors on startup.  Everything looks like it works but it is not.

vbekiaris:
Logs a warning and resets the `maxSize` value to be the `max(userConfiguredMaxSize, partitionCount)` when an `IMap` is configured with an `EvictionPolicy` other than `NONE`.

Fixes #10179

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/config/ConfigValidator.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/config/ConfigValidator.java
@@ -20,6 +20,8 @@ import com.hazelcast.config.EvictionConfig;
 import com.hazelcast.config.EvictionPolicy;
 import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.config.MapConfig;
+import com.hazelcast.config.MaxSizeConfig;
+import com.hazelcast.config.MaxSizeConfig.MaxSizePolicy;
 import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.config.NearCachePreloaderConfig;
 import com.hazelcast.instance.BuildInfoProvider;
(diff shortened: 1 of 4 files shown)


hazelcast/src/main/java/com/hazelcast/spi/properties/HazelcastProperties.java
```diff
@@ -234,4 +202,49 @@ public int getSeconds(HazelcastProperty property) {
...
+
+    /**
+     * Lookup a {@link HazelcastProperty}'s {@link String} value in the given {@code properties} or in
+     * {@link System#getProperties()}, as specified in this class' javadoc.
+     * @param properties    user-provided {@link Properties}. If {@code null}, a {@link NullPointerException} will
+     *                      be thrown.
+     * @param property      the {@link HazelcastProperty} whose value is looked up
+     * @return              the configured value
+     */
+    public static String getString(Properties properties, HazelcastProperty property) {
```
vbekiaris: The property lookup implementation was extracted to this `public static` method, so that properties can be looked up from contexts where `Node` is not available.

hazelcast/src/main/java/com/hazelcast/internal/config/ConfigValidator.java
```diff
@@ -144,6 +146,20 @@ public static void checkEvictionConfig(EvictionPolicy evictionPolicy,
         }
     }
 
+    public static void checkMaxSizeEvictionConfig(MapConfig mapConfig, int partitionCount) {
```
Donnerbart: I think this is the first time the `ConfigChecker` changes a config. Maybe it's cleaner to return a `boolean` value with the check result and do the `maxSizeConfig.setSize(partitionCount)` at the caller side? Of course this would need a different test setup then.

ahmetmircik:
As per my understanding the issue is about `PER_NODE` max size policy.  Translated per-partition max size from it is calculated [here](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictionChecker.java#L113). Seems fixing the size calculation in method `calculatePerNodeMaxRecordStoreSize` should be sufficient as a fix.

vbekiaris:
@ahmetmircik I agree, I also just realized that. The static config check cannot take into account the runtime cluster variables, eg on a 4-nodes cluster with `PER_NODE` policy & `maxSize = 68` , each partition should be able to hold 1 entry.

PR #10497, opened 2017-05-04
Added directBuffer parameter to SocketChannelWrapperFactory

pveentjer:
needed for tls so that the SSLSocketChannelWrapper knows when it needs to create
direct or indirect buffers.

For enterprise PR see https://github.com/hazelcast/hazelcast-enterprise/pull/1464


hazelcast/src/main/java/com/hazelcast/nio/IOService.java
```diff
@@ -82,7 +82,7 @@
 
     int getSocketSendBufferSize();
 
-    boolean isSocketBufferDirect();
```
pveentjer: A bit more sane english

PR #10506, opened 2017-05-05
Fix topic overload test whenBlock_whenNoSpace

sancar:
Test was failing because of ttl was kicking in before
assertTrueAllTheTime.

setTimeToLiveSeconds set to Integer.MAX to avoid test failures
because of unfortunate timing.

Removed the check that messages are published eventually because it was causing test
to take too long time(larger than configured ttl time, it was 30 seconds) 

fixes https://github.com/hazelcast/hazelcast/issues/10288


pveentjer:
merge once green since it just a bug fix

sancar:
run-lab-run

PR #10514, opened 2017-05-05
Cleanup of ClientProxy method usage

Donnerbart:
* removed duplicated fields and methods
* removed chained method calls which have their own method, e.g.
  `getContext().getSerializationService()` -> `getSerializationService()`
  or `getSerializationService().toData()` -> `toData()`

Follow-up PR of https://github.com/hazelcast/hazelcast/pull/10481

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
@@ -18,7 +18,6 @@ package com.hazelcast.client.cache.impl;
 
 import com.hazelcast.cache.CacheStatistics;
 import com.hazelcast.client.impl.ClientMessageDecoder;
-import com.hazelcast.client.impl.HazelcastClientInstanceImpl;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.impl.protocol.codec.CacheGetAllCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheGetCodec;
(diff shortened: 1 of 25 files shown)


Donnerbart:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/AbstractClientCacheProxy.java
```diff
@@ -98,11 +97,10 @@ protected V getSyncInternal(Data keyData, ExpiryPolicy expiryPolicy) {
     private ClientDelegatingFuture<V> getInternal(Data keyData, ExpiryPolicy expiryPolicy, boolean deserializeResponse) {
         Data expiryPolicyData = toData(expiryPolicy);
 
-        HazelcastClientInstanceImpl client = (HazelcastClientInstanceImpl) clientContext.getHazelcastInstance();
         ClientMessage request = CacheGetCodec.encodeRequest(nameWithPrefix, keyData, expiryPolicyData);
-        int partitionId = clientContext.getPartitionService().getPartitionId(keyData);
+        int partitionId = getContext().getPartitionService().getPartitionId(keyData);
```
pveentjer: The field access is replaced by a getter? What is the added value?

hazelcast-client/src/main/java/com/hazelcast/client/spi/ClientProxy.java
```diff
@@ -84,6 +85,18 @@ protected final ClientProxy setContext(ClientContext context) {
         return this;
     }
 
+    protected SerializationService getSerializationService() {
+        return serializationService != null ? serializationService : lazySerializationService;
```
pveentjer: To make thing easier to understand I always try to get rid of not.

```
return serializationService == null ? lazySerializationService: serializationService;
```

Donnerbart: Normally I do the same, but in this case I tried both variants and in the end it looked smoother to read: `return A != null ? A : B` instead of `return A == null ? B : A`

PR #10521, opened 2017-05-08
Configurable event sync timeout

pveentjer:
It was hardcoded at 5 seconds, but now it is made configurable. This property is needed
to configure the max waiting time for a 'fake synchronous event' to get the outbound
queue of a connection drained. But if the wait is short; the mechanism doesn't provide
sufficient back pressure and you still end up with an overloaded connection.

Backport of https://github.com/hazelcast/hazelcast/pull/10520

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
+++ b/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
@@ -117,6 +117,17 @@ public final class GroupProperty {
     public static final HazelcastProperty EVENT_QUEUE_TIMEOUT_MILLIS
             = new HazelcastProperty("hazelcast.event.queue.timeout.millis", 250, MILLISECONDS);
 
+    /**
+     * To prevent overload on the outbound connections, once and a while an event is made synchronous by wrapping it in a
+     * fake operation and waiting for a fake response. This causes the outbound write queue of the connection to get drained.
+     *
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
```diff
@@ -117,6 +117,17 @@
     public static final HazelcastProperty EVENT_QUEUE_TIMEOUT_MILLIS
             = new HazelcastProperty("hazelcast.event.queue.timeout.millis", 250, MILLISECONDS);
 
+    /**
+     * To prevent overload on the outbound connections, once and a while an event is made synchronous by wrapping it in a
+     * fake operation and waiting for a fake response. This cases the outbound write queue of the connection to get drained.
```
tkountis: small typo 'cases' -> 'causes'

hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java
```diff
@@ -117,6 +117,17 @@
...
+     *
+     * This timeout configures the maximum amount of waiting time for this fake response. Setting it too a too low value
```
tkountis: also 'too' -> 'to'

pveentjer:
Thanks for the reviews guys.

PR #10524, opened 2017-05-08
Added NearCacheTestContextBuilder to ease NearCacheTestContext usage

Donnerbart:
* added builder for `NearCacheTestContext` and changed all tests to use it
* created distinct configurations with and without Near Cache for the
  remaining serialization count tests

The `NearCacheTestContext` gets more and more fields which are not set in all tested scenarios. So we are getting more and more specialized constructors or `null` fields. This PR cleans this up by using the builder pattern.

I also created distinct configurations for the last tests which used the same config (or even same instance) for the `NearCacheTestContext` creation.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/LiteMemberMapNearCacheSerializationCountTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/LiteMemberMapNearCacheSerializationCountTest.java
@@ -27,6 +27,7 @@ import com.hazelcast.internal.nearcache.AbstractNearCacheSerializationCountTest;
 import com.hazelcast.internal.nearcache.NearCache;
 import com.hazelcast.internal.nearcache.NearCacheManager;
 import com.hazelcast.internal.nearcache.NearCacheTestContext;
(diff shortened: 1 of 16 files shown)


hazelcast/src/test/java/com/hazelcast/map/impl/nearcache/LiteMemberMapNearCacheSerializationCountTest.java
```diff
@@ -116,16 +117,14 @@ public void tearDown() {
...
         NearCache<Data, String> nearCache = nearCacheManager.getNearCache(DEFAULT_NEAR_CACHE_NAME);
 
-        return new NearCacheTestContext<K, V, Data, String>(
-                getSerializationService(member),
-                liteMember,
-                member,
-                new IMapDataStructureAdapter<K, V>(liteMemberMap),
-                new IMapDataStructureAdapter<K, V>(memberMap),
-                nearCacheConfig,
-                true,
```
Donnerbart: This `boolean` switch is not evaluated in the serialization count tests, but it would anyway be wrong for a LiteMember.

Donnerbart:
run-lab-run

Issue #10389, opened 2017-04-20
Near Cache size didn't reach the desired value (off-by-one)


mmedenjak:
**com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheBasicTest.whenNearCacheIsFull_thenPutOnSameKeyShouldUpdateValue_onDataAdapter[format:BINARY]**
```
java.lang.AssertionError: Near Cache size didn't reach the desired value (499 vs. 500) (NearCacheStatsImpl{ownedEntryCount=500, ownedEntryMemoryCost=74443, creationTime=1492640495276, hits=0, misses=1001, ratio=0.0%, evictions=0, expirations=0, lastPersistenceTime=0, persistenceCount=0, lastPersistenceDuration=0, lastPersistenceWrittenBytes=0, lastPersistenceKeyCount=0, lastPersistenceFailure=''}) expected:<499> but was:<500>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at com.hazelcast.internal.nearcache.NearCacheTestUtils.assertNearCacheSize(NearCacheTestUtils.java:271)
	at com.hazelcast.internal.nearcache.NearCacheTestUtils$2.run(NearCacheTestUtils.java:291)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:992)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1009)
	at com.hazelcast.internal.nearcache.NearCacheTestUtils.assertNearCacheSizeEventually(NearCacheTestUtils.java:288)
	at com.hazelcast.internal.nearcache.AbstractNearCacheBasicTest.whenNearCacheIsFull_thenPutOnSameKeyShouldUpdateValue(AbstractNearCacheBasicTest.java:1036)
	at com.hazelcast.internal.nearcache.AbstractNearCacheBasicTest.whenNearCacheIsFull_thenPutOnSameKeyShouldUpdateValue_onDataAdapter(AbstractNearCacheBasicTest.java:1008)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
[...]

PR #10525, opened 2017-05-08
Added Near Cache invalidation listeners for unified test

Donnerbart:
There was a race between the data structure population, which can create
invalidation events and the Near Cache population. If the Near Cache is
populated faster than the invalidations batches are sent or processed,
the Near Cache entries are invalidated and the `assertNearCacheSize()`
check fails.

Depends on https://github.com/hazelcast/hazelcast/pull/10524
Fixes https://github.com/hazelcast/hazelcast/issues/10389

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheNearCacheBasicTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheNearCacheBasicTest.java
@@ -128,6 +128,9 @@ public class ClientCacheNearCacheBasicTest extends AbstractNearCacheBasicTest<Da
                 .setNearCacheManager(nearCacheManager)
                 .setCacheManager(cacheManager)
                 .setMemberCacheManager(memberCacheManager)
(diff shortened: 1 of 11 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/cache/nearcache/ClientCacheNearCacheBasicTest.java
```diff
@@ -128,6 +128,9 @@ public void tearDown() {
                 .setNearCacheManager(nearCacheManager)
                 .setCacheManager(cacheManager)
                 .setMemberCacheManager(memberCacheManager)
+                // FIXME: the JCache doesn't send invalidation on CREATED entries, so this will crash some tests
```
Donnerbart: AbstractRecordStore
```java
    protected R doPutRecord(Data key, R record, String source) {
        R oldRecord = records.put(key, record);
        if (oldRecord != null) {
            invalidateEntry(key, source);
        }
        return oldRecord;
    }
```
seems to prevent invalidation events on CREATED entries, so we cannot use the listener for now to wait for invalidation events after a JCache population

Donnerbart:
Kudos to @ahmetmircik for finding the race behind this error.

PR #10527, opened 2017-05-08
Fixed missing Near Cache invalidations on ICache.put() for new entries

Donnerbart:
While fixing the race between Near Cache invalidations and Near Cache population I found out, that `JCache` doesn't send invalidation events on created entries (which `IMap` and `ReplicatedMap` do). This PR adds the missing invalidations to `JCache` and activates the assertion for the invalidations (via an invalidation listener).

It also adds the `NearCacheInvalidationListener` to the `NearCacheTestContext` and tests, so there are no race conditions in the JCache Near Cache tests.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/cache/impl/nearcache/ClientNearCacheInvalidationTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/cache/impl/nearcache/ClientNearCacheInvalidationTest.java
@@ -35,6 +35,7 @@ import com.hazelcast.config.InMemoryFormat;
 import com.hazelcast.config.NearCacheConfig;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.internal.nearcache.NearCache;
+import com.hazelcast.internal.nearcache.NearCacheInvalidationListener;
 import com.hazelcast.internal.nearcache.NearCacheManager;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.test.AssertTask;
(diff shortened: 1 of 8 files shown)


Donnerbart:
run-lab-run

Donnerbart:
I just had a look at the test failures, this change brings with it and found another reason why ICache doesn't send invalidations for created entries:
* `LocalUpdatePolicy.CACHE_ON_UPDATE` is configured
* populate the ICache from the client
* Near Cache on the client is directly populated
* member creates entries in the backing ICache data structure
* sends out invalidations
* Near Cache on the client is invalidated

This breaks some tests and is surely not intended. I don't know if there is an easy fix for this or not. As far as I know the member doesn't know about the client configs (so it cannot skip invalidations for those clients). And I don't think that the invalidation events contain info if an entry was created or updated.

So yeah, don't know if we can align the ICache and IMap behavior in this case.

Donnerbart:
This might also be a blocker to implement `CACHE_ON_UPDATE` for `IMap`.

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
I ran multiple PR builders, beside some non-related spurious failures, all race-conditions seem to be gone now. The spurious failures have been reported, so this PR is clean and shiny now.

PR #10528, opened 2017-05-08
Added ClientReplicatedMapInvalidationListener to fix race in tests

Donnerbart:
Added the same listener type like https://github.com/hazelcast/hazelcast/pull/10525 for `ReplicatedMap`.


sancar:
@Donnerbart 
Could this be related to last test failure here ? 
https://github.com/hazelcast/hazelcast/issues/9333

Donnerbart:
`whenCacheIsFull_thenPutOnSameKeyShouldUpdateValue_withUpdateOnDataAdapter` means that we have an async (batch) invalidation from a "member" to a "client with Near Cache", which is the racy scenario. Also we have a `member.put()` which triggers an invalidation and then populate the client Near Cache again, which is exactly what is racy to test.

I've already added an additional `assertNearCacheInvalidations()` to the test, but without the listener implementation from this PR the assert will just pass for the `ReplicatedMap` test.

PR #10529, opened 2017-05-09
Clean up cluster-related assertions in tests

metanet:
* Fix some usages of assertClusterSize() and assertClusterSizeEventually()
* Use var-args based cluster size assertions to reduce duplicate code

This pr does not make any change related to test logic.


pveentjer:
Since this is just simple assert cleanup, I don't see a reason waiting for a second reviewer.

metanet:
thx!

PR #10533, opened 2017-05-09
Cleanup of NearCachedMapProxyImpl

Donnerbart:
Preparation to implement the key store-by-reference feature.


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
```diff
@@ -143,41 +142,34 @@ public void onFailure(Throwable t) {
...
-
         return future;
     }
 
     @Override
     protected Data putInternal(Data key, Data value, long ttl, TimeUnit timeunit) {
-        Data data;
         try {
-            data = super.putInternal(key, value, ttl, timeunit);
+            return super.putInternal(key, value, ttl, timeunit);
```
pveentjer: +5

So much simpler to read.

PR #10534, opened 2017-05-09
Weak configuration secrets policy

tkountis:
OSS side changes needed to support EE feature.
See hazelcast/hazelcast-enterprise#1397

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
+++ b/hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
@@ -21,6 +21,7 @@ import com.hazelcast.hotrestart.HotRestartService;
 import com.hazelcast.hotrestart.InternalHotRestartService;
 import com.hazelcast.internal.cluster.impl.JoinMessage;
 import com.hazelcast.internal.cluster.impl.JoinRequest;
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/internal/management/ManagementCenterService.java
```diff
@@ -130,6 +130,10 @@ public ManagementCenterService(HazelcastInstanceImpl instance) {
         }
     }
 
+    protected TimedMemberStateFactory createTimedMemberStateFactory(HazelcastInstanceImpl instance) {
```
pveentjer: I guess this method is some kind of template method that can be overridden the EnterprisemanagementCenter?

If so; I think having a policy based approach is way cleaner.

hazelcast/src/main/java/com/hazelcast/instance/NodeExtension.java
```diff
@@ -224,4 +224,5 @@
      * @return new uuid
      */
     String createMemberUuid(Address address);
+
```
pveentjer: unwanted new line?

mmedenjak:
run-lab-run

PR #10538, opened 2017-05-10
[WIP] Channel cleanup

pveentjer:
* ChannelReader/ChannelWriter are now implementation details. All requests go through the Channel. So Connection implementation have been decoupled from channel implementation details.
* EventLoopGroup no more reliance on Connection; just a single register(Channel) method which does the job.
* SocketReader/Writer no more reliance on Connection. 
* Channel initialization has been simplified. Instead of needing to expose the reader/writer so that certain fields can be set; a result object is returned with the fields to set. It is now up the the ChannelReader/Writer to set these fields.
* ChannelReaderInitializer and ChannelWriterInitializer have been merged into the ChannelInitializer.


hazelcast-client/src/main/java/com/hazelcast/client/impl/DefaultClientExtension.java
```diff
@@ -40,7 +40,7 @@
 import com.hazelcast.map.impl.MapService;
 import com.hazelcast.nio.ClassLoaderUtil;
 import com.hazelcast.nio.SocketInterceptor;
-import com.hazelcast.nio.tcp.PlainChannelFactory;
+import com.hazelcast.internal.networking.PlainChannelFactory;
```
pveentjer: Moved PlainNioChannelFactory to com.hazelcast.internal.networking since both client and server use these classes.

hazelcast-client/src/test/java/com/hazelcast/Main.java
```diff
@@ -0,0 +1,39 @@
+package com.hazelcast;
+
```
pveentjer: will be deleted

hazelcast/src/main/java/com/hazelcast/instance/DefaultNodeContext.java
```diff
@@ -65,15 +64,12 @@ private EventLoopGroup createEventLoopGroup(Node node, NodeIOService ioService)
         boolean spinning = Boolean.getBoolean("hazelcast.io.spinning");
         LoggingServiceImpl loggingService = node.loggingService;
 
-        MemberChannelWriterInitializer socketWriterInitializer
-                = new MemberChannelWriterInitializer(loggingService.getLogger(MemberChannelWriterInitializer.class));
-        MemberChannelReaderInitializer socketReaderInitializer
-                = new MemberChannelReaderInitializer(loggingService.getLogger(MemberChannelReaderInitializer.class));
+        MemberChannelInitializer socketReaderInitializer
```
pveentjer: The initializer takes now care of both read/write initialization

hazelcast/src/main/java/com/hazelcast/internal/networking/ChannelReader.java
```diff
@@ -1,90 +0,0 @@
-/*
```
pveentjer: Implementation detail; should not be exposed as a general API. All communication of higher level systems like Connections go through Channel.

hazelcast/src/main/java/com/hazelcast/internal/networking/ChannelWriter.java
```diff
@@ -1,96 +0,0 @@
-/*
```
pveentjer: Implementation detail; should not be exposed as a general API. All communication of higher level systems like Connections go through Channel.

hazelcast/src/main/java/com/hazelcast/internal/networking/CloseListener.java
```diff
@@ -16,9 +16,7 @@
 
 package com.hazelcast.internal.networking;
 
-import com.hazelcast.nio.Connection;
+public interface CloseListener {
 
-public interface ChannelWriterInitializer<C extends Connection> {
```
pveentjer: The ChannelWriterInitializer has been merged into the ChannelInitializer.

hazelcast/src/main/java/com/hazelcast/internal/networking/OnFirstContact.java
```diff
@@ -16,13 +16,23 @@
 
 package com.hazelcast.internal.networking;
 
-import com.hazelcast.nio.Connection;
+import java.nio.ByteBuffer;
 
-public interface ChannelConnection extends Connection {
```
pveentjer: The com.hazelcast.internal.networking doesn't rely on Connection any longer.

hazelcast/src/main/java/com/hazelcast/internal/networking/nio/iobalancer/IOBalancer.java
```diff
@@ -100,12 +100,12 @@ LoadTracker getOutLoadTracker() {
         return outLoadTracker;
     }
 
-    public void connectionAdded(MigratableHandler readHandler, MigratableHandler writeHandler) {
```
pveentjer: Since we don't rely on connections any longer, but focus on channels; best to rename this to something more channel oriented.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
```diff
@@ -80,42 +63,31 @@
...
         this.connectionId = connectionId;
         this.connectionManager = connectionManager;
         this.ioService = connectionManager.getIoService();
         this.logger = ioService.getLoggingService().getLogger(TcpIpConnection.class);
         this.channel = channel;
-        this.channelWriter = eventLoopGroup.newSocketWriter(this);
-        this.channelReader = eventLoopGroup.newSocketReader(this);
-    }
-
-    @Override
```
pveentjer: All the metric stuff is responsibility of the channel.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnection.java
```diff
@@ -40,24 +35,12 @@
...
- *
- * @see EventLoopGroup
  */
 @SuppressWarnings("checkstyle:methodcount")
-public final class TcpIpConnection implements ChannelConnection, DiscardableMetricsProvider {
+public final class TcpIpConnection implements Connection {
 
     private final Channel channel;
 
-    private final ChannelReader channelReader;
```
pveentjer: No more reliance on readers/writer since they are implementation details.

PR #10542, opened 2017-05-10
Adds Channel attribute map

pveentjer:
Using the attribute map one can add data to a channel without needing to add fields to that channel.

The protocol-buffer has been removed as explicit field on the channelreaders; and is now stored
as attachment on the channel. The protocol is server-side specific; client doesn't care for
protocol-buffer; so it should not be part of the shared infrastructure.


hazelcast/src/main/java/com/hazelcast/internal/networking/ChannelReader.java
```diff
@@ -85,6 +85,4 @@
     void setInboundHandler(ChannelInboundHandler inboundHandler);
 
     Channel getChannel();
-
-    ByteBuffer getProtocolBuffer();
```
pveentjer: Is now stored in the channel attribute map

PR #10543, opened 2017-05-10
Renamed ClientConnection.getLocalSocketAddress to getLocalAddress

pveentjer:
Now in line with getRemoteAddress.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
@@ -229,7 +229,7 @@ public class ClientConnection implements ChannelConnection, DiscardableMetricsPr
         this.remoteEndpoint = remoteEndpoint;
     }
 
-    public InetSocketAddress getLocalSocketAddress() {
+    public InetSocketAddress getLocalAddress() {
         return (InetSocketAddress) channel.getLocalAddress();
     }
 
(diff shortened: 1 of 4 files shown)


sancar:
I think both getLocalAddress and getRemoteAddress should be getLocalSocketAddress and getRemoteSocketAddress. 
The methods added on Channel interface should end with SocketAddress.  

When you look at getLocalAddress expectation is for it to return Address. But it returns SocketAddress, hence naming it getLocalSocketAddress looks more intuitive. 

I also see that Connection.getRemoteSocketAddress() also renamed to getRemoteAddress(), which should not be changed because of same reasoning.

pveentjer:
The idea is that Channel and therefor Connection loose coupling with the Channel implementation; which could be a socket. In the future we want to experiment with different Channel implementations like e.g. UDP.

The naming following that of Netty:
https://netty.io/4.0/api/io/netty/channel/Channel.html#remoteAddress--

sancar:
One problem is that we already have an Address class. 

Secondly, getRemoteSocketAddress is not block for loose coupling. It just follows the name of return type. 
As long as return type is SocketAddress , `getRemoteSocketAddress` makes more sense. 

If SocketAddress can support loose coupling than name should be  `getRemoteSocketAddress` , 
if SocketAddress is not generic enough to support different implementations, then return type and method name should change together accordingly. 

Javadoc of SocketAddress related to protocol:
```
This class represents a Socket Address with no protocol attachment. As an abstract class, it is meant to be subclassed with a specific, protocol dependent, implementation.
```

pveentjer:
Makes sense. I'll change the naming on the Channel as well so that we don't get different names for the same thing.

pveentjer:
I have made a new PR using get...SocketAddress as naming

https://github.com/hazelcast/hazelcast/pull/10545

PR #10548, opened 2017-05-10
Event journal for map and cache

mmedenjak:
https://hazelcast.atlassian.net/wiki/display/PM/Core+Part+-+Stream+Source+from+Map+and+Cache

https://hazelcast.atlassian.net/wiki/display/EN/Core+Part+-+Stream+Source+from+Map+and+Cache+-+Design

Gotchas : 
- the event journal config is outside of map and cache config and is matched to a specific map or cache by name. In the case of cache, the simple cache name is used but the configuration needs to be created on all nodes before subscribing
- the event journal config is public but the subscribe/unsubscribe is not public (they are public methods on an internal class - `MapProxyImpl`, `ClientMapProxy`, `CacheProxy` and `ClientCacheProxy`)
- the journal is destroyed if the map is destroyed. This includes migrations (when a member is no longer a replica). The event journal should be colocated with the replicas. If the map is destroyed, the listeners should fail as well. Sorry, the map is no more.

Depends on : https://github.com/hazelcast/hazelcast-client-protocol/pull/80
EE : https://github.com/hazelcast/hazelcast-enterprise/pull/1501

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
@@ -167,6 +167,7 @@ public class DefaultRecordStore extends AbstractEvictableRecordStore {
     public void destroy() {
         clearPartition(false);
         storage.destroy(false);
+        eventJournal.destroy(mapContainer.getObjectNamespace(), partitionId);
     }
 
     @Override
@@ -228,6 +229,7 @@ public class DefaultRecordStore extends AbstractEvictableRecordStore {
     public void putRecord(Data key, Record record) {
         markRecordStoreExpirable(record.getTtl());
         storage.put(key, record);
+        eventJournal.writeAddEvent(mapContainer.getObjectNamespace(), partitionId, key, record.getValue());
         updateStatsOnPut(record.getHits());
     }
 
@@ -245,6 +247,7 @@ public class DefaultRecordStore extends AbstractEvictableRecordStore {
         if (record == null) {
             record = createRecord(value, ttl, now);
             storage.put(key, record);
+            eventJournal.writeAddEvent(mapContainer.getObjectNamespace(), partitionId, key, record.getValue());
         } else {
             updateRecord(key, record, value, now);
         }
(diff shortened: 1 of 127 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/recordstore/DefaultRecordStore.java
```diff
@@ -765,6 +774,7 @@ protected Object putInternal(Data key, Object value, long ttl, boolean loadFromS
         if (record == null) {
             record = createRecord(value, ttl, now);
             storage.put(key, record);
+            eventJournal.writeAddEvent(mapContainer.getObjectNamespace(), partitionId, record.getKey(), record.getValue());
```
ahmetmircik: shouldn't we add to journal before `storage#put`? For example, if we have a db backed ring buffer, it seems the true way is first to try insert into db then add to map.

mmedenjak: It's a matter of flavour - on the other hand I would not want to receive an event for something that failed to be stored in the `RecordStore`.
Currently the event journal cannot be configured with a ringbuffer store so there is no way it could store to a database. And the choice was that it should contain an event only if the call to the `RecordStore` succeeded.

ahmetmircik: ok if there will be no store, this is not an issue but if we will add store to ringbuffer later, i expect it to behave as in`mapstore` case. Currently, if something cannot be stored in `mapstore`, operations fail.

mmedenjak: I don't think we will be adding the store any time soon - it can severely slow down the partition operations, even to the point of stalling them. We would also need some way of determining if it is a backup operation so it doesn't get stored twice. 
Currently it is a simple "put in array" which can hardly fail. If and when we will add the ringbuffer store to the event journal we can easily switch places for all calls to the record store and event journal.

mmedenjak:
Split this PR into 2, closing this one :
https://github.com/hazelcast/hazelcast/pull/10684
https://github.com/hazelcast/hazelcast/pull/10685

PR #10550, opened 2017-05-10
Apply minor fixes in the new clustering design

### LINKED ISSUE #10526: MockJoinerTest.restart_multipleInParallel()
```java
java.lang.AssertionError: Cluster size is not correct. Expected: 5 Actual: 4
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSize(HazelcastTestSupport.java:870)
	at com.hazelcast.test.HazelcastTestSupport$12.run(HazelcastTestSupport.java:889)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1003)
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSizeEventually(HazelcastTestSupport.java:885)
	at com.hazelcast.test.HazelcastTestSupport.assertClusterSizeEventually(HazelcastTestSupport.java:880)
	at com.hazelcast.test.mocknetwork.MockJoinerTest.assertClusterSizeEventually(MockJoinerTest.java:145)
	at com.hazelcast.test.mocknetwork.MockJoinerTest.restart_multipleInParallel(MockJoinerTest.java:139)
```
https://hazelcast-l337.ci.cloudbees.com/job/new-lab-fast-pr/8738/testReport/junit/com.hazelcast.test.mocknetwork/MockJoinerTest/restart_multipleInParallel/

metanet:
1) A node should not perform any cluster-related tasks, such as mastership claim, if the cluster service is already reset during shutdown. Otherwise, if the current master and another member leaves the cluster simultaneously, the current master may think that it has just became the master and can start the mastership claim process by mistake.
2) If a node restarts with the same address and a new uuid, it should respond (i.e., explicit suspicions) to incoming heartbeats that include its previous uuid, even before it joins to the cluster.

Fixes #10526

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterHeartbeatManager.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterHeartbeatManager.java
@@ -160,7 +160,13 @@ public class ClusterHeartbeatManager {
 
         try {
             if (!clusterService.isJoined()) {
-                logger.fine("Ignoring heartbeat of sender: " + senderMembersViewMetadata + ", because node is not joined!");
+                if (clusterService.getThisUuid().equals(receiverUuid)) {
+                    logger.fine("Ignoring heartbeat of sender: " + senderMembersViewMetadata + ", because node is not joined!");
+                } else {
+                    logger.fine("Sending explicit suspicion to " + senderAddress + " for heartbeat " + senderMembersViewMetadata
+                            + ", because this node has received an invalid heartbeat before it joins to the cluster");
+                    clusterService.sendExplicitSuspicion(senderMembersViewMetadata);
+                }
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ClusterHeartbeatManager.java
```diff
@@ -160,7 +160,13 @@ public void handleHeartbeat(MembersViewMetadata senderMembersViewMetadata, Strin
 
         try {
             if (!clusterService.isJoined()) {
-                logger.fine("Ignoring heartbeat of sender: " + senderMembersViewMetadata + ", because node is not joined!");
+                if (clusterService.getThisUuid().equals(receiverUuid)) {
+                    logger.fine("Ignoring heartbeat of sender: " + senderMembersViewMetadata + ", because node is not joined!");
```
pveentjer: I don't know the frequency of this code being called; but if it is frequent, it is best to add a logging guard to prevent litter.

mdogan: I don't think this will be frequent. At most, this can be logged on every heartbeat (every 5 sec). But this case happens only when node is not joined yet.

metanet: Yes, it is very infrequent. That is why there is no filtering but I can add just add it. np

PR #10552, opened 2017-05-10
Fixed some FindBugs warnings in the quorum tests

Donnerbart:
All map fields are used in instance methods, so FindBugs think its suspicious that they are static. Renamed `foo` to `future`. Got rid of a lot of `throws Exception`. Changed some fields to local variables and added missing generics.

It's actually not much of a change, but due to the massive code duplication in the quorum tests, it's a big PR :(


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

PR #10553, opened 2017-05-10
Fixed some FindBugs warnings in the client tests

Donnerbart:
FindBugs finds a static volatile variable scary that is written from an instance method. An `AtomicBoolean` is easier to understand than trying to figure out how static and volatile work together.

The other changes are mostly dead local stores.


hazelcast-client/src/test/java/com/hazelcast/client/ClientMemberAttributeTest.java
```diff
@@ -120,7 +119,6 @@ public void testPresharedAttributes() throws Exception {
         JoinConfig join = c.getNetworkConfig().getJoin();
         join.getTcpIpConfig().addMember("127.0.0.1").setEnabled(true);
         join.getMulticastConfig().setEnabled(false);
-        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory(2);
```
Donnerbart: The factory is a class field by now, so this one is not used at all.

PR #10554, opened 2017-05-10
Fixed some FindBugs warnings in the cache tests

Donnerbart:
Fixed dead local stores, missing static keyword for classes, not used method return types, superfluous boxing and integer division which is assigned to a float.


Donnerbart:
run-lab-run

PR #10561, opened 2017-05-11
Prevented thread dump on assumeThat() mismatch

Donnerbart:
The motivation behind this PR is to prevent thread dumps when a test was skipped during runtime due to a assumeThat() mismatch. For example the unified Near Cache tests use this a lot to skip tests on methods, which are not available on all data structures:
```java
22:24:45,343  INFO |whenGetAsyncIsUsed_thenNearCacheShouldBePopulated[format:BINARY]| -
    [MethodAvailableMatcher] whenGetAsyncIsUsed_thenNearCacheShouldBePopulated -
    TransactionalMapDataStructureAdapter.getAsync(Object) is available: false
    ([@com.hazelcast.internal.adapter.MethodNotAvailable()])!
THREAD DUMP FOR TEST FAILURE: "got: <class com.hazelcast.internal.adapter.TransactionalMapDataStructureAdapter>,
    expected: getAsync(Object) to be available" at "whenGetAsyncIsUsed_thenNearCacheShouldBePopulated"

(...) huge thread dump
```
This leads to a thread dump, although this is no real failure. Throwing an `AssumptionViolatedException` is just the system used by JUnit to end a test on a `assumeThat()` mismatch.

With this change the size of the `com.hazelcast.map.impl.tx.TxnMapNearCacheBasicTest-output.txt` should shrink from about 19.7 MB to 952 KB.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/AbstractHazelcastClassRunner.java
@@ -29,6 +29,7 @@ import net.bytebuddy.jar.asm.Opcodes;
 import net.bytebuddy.matcher.ElementMatcher;
 import net.bytebuddy.utility.JavaModule;
 import org.junit.After;
+import org.junit.AssumptionViolatedException;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.internal.runners.statements.RunAfters;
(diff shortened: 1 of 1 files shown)


Donnerbart:
I just checked the log size of this PR build vs. an older one: 803 MB vs. 663 MB
That saves us about 17% per PR build run :)

PR #10562, opened 2017-05-11
Channel: remove coupling Connection

pveentjer:
The main theme for this PR is removing coupling from Channel to Connection. Connection relies on a channel do the actual work, but the Channel relies on Connection for all kinds of information like if it is running, error handling etc. So currently there is a cycle.

This PR removes this dependency cycle and reduces coupling with implementation details of the channel:

Another important part is making the protocol/initialization part self contained. This is required for the TLS functionality since it will rely on the same mechanisms.

* ErrorHandling is pulled out into a strategy: ChannelErrorHandler. It will also take care of OOME.. 

* ChannelReader/ChannelWriter are dropped; they are implementation details. All interaction with the ChannelReader/Writer have been replaced by Channel interaction or the interactions have been dropped completely since they expose implementation details.

* ChannelCloseListener was added for listen to a closing channel and be able to clean up resources.

* Metrics are now fully configured through the EventLoopGroup; the connection doesn't need to care about metrics from the read/writehandler (reduce coupling)

* ChannelInitializer doesn't rely on Connection; Connection is added as attachment to the Channel; but there is no API dependence.

* PlainChannel has been renamed to NioChannel. Also we have now a SpinningChannel as well. For encryption we still need to subclass the NioChannel; but this will in the near future be resolved by adding encryption through the channel pipeline.

* EventLoopGroup has no more dependence on Connection; lots of methods have been removed and replaced by a single register method. Closing of  connection ---> channel --> channel close listener. The surface area of the API has been reduced.

* Channel: removed a bunch of socket-channel specific methods. These are implementation dependent and can be executed directly on the underlying socket channel in the code that knows it is talking to a socketcannel. No need to expose this in the Channel API.

* Fixes blocking on IO threads for protocol https://github.com/hazelcast/hazelcast/issues/10577

* The protocol handling is now fully contained in the ChannelInitializer instead of it being spread over multiple classes.

* Channel exposes method to see if it is client; so the one initiating the connection (client), or accepting the connection (server).

This PR doesn't change what goes over the wire!


hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -395,18 +391,19 @@ protected ClientConnection createSocketConnection(final Address address) throws
...
+            Channel channel = channelFactory.create(socketChannel, true, directBuffer);
 
             final ClientConnection clientConnection = new ClientConnection(
-                    client, eventLoopGroup, connectionIdGen.incrementAndGet(), channel);
+                    client, connectionIdGen.incrementAndGet(), channel);
             socketChannel.configureBlocking(true);
             if (socketInterceptor != null) {
                 socketInterceptor.onConnect(socket);
             }
-            socketChannel.configureBlocking(eventLoopGroup.isBlocking());
```
pveentjer: The eventloopgroup takes care of configuring the socketchannel when the channel is registered.

hazelcast/src/main/java/com/hazelcast/internal/networking/nio/NioEventLoopGroup.java
```diff
@@ -271,31 +250,64 @@ private void shutdown(NioThread[] threads) {
...
+        final NioChannelReader reader = newChannelReader(nioChannel);
+        final NioChannelWriter writer = newChannelWriter(nioChannel);
+
+        nioChannel.setReader(reader);
+        nioChannel.setWriter(writer);
+
+        ioBalancer.channelAdded(reader, writer);
+
+        String metricsId = channel.getLocalSocketAddress() + "->" + channel.getRemoteSocketAddress();
+        metricsRegistry.scanAndRegister(writer, "tcp.connection[" + metricsId + "].out");
```
pveentjer: In the future we can remove 'tcp.connection' into a prefix variable. Currently both client/member connections are tcp.connections.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionManager.java
```diff
@@ -368,14 +364,7 @@ private TcpIpConnectionErrorHandler getErrorHandler(Address endpoint, boolean re
     public void onConnectionClose(Connection connection) {
         closedCount.inc();
 
-        if (activeConnections.remove(connection)) {
-            // this should not be needed; but some tests are using DroppingConnection which is not a TcpIpConnection.
```
pveentjer: The EventLoopGroup will listen to channel closing, and takes actions accordingly.

hazelcast/src/main/java/com/hazelcast/internal/networking/ChannelReader.java
```diff
@@ -1,82 +0,0 @@
-/*
```
pveentjer: we don't need these interfaces.

hazelcast/src/main/java/com/hazelcast/internal/networking/ChannelWriter.java
```diff
@@ -1,89 +0,0 @@
...
- *
- * Each {@link TcpIpConnection} has its own {@link ChannelWriter} instance.
- *
- * Before Hazelcast 3.6 the name of this interface was ChannelOutboundHandler.
- *
- * @see ChannelReader
- * @see ChannelInboundHandler
- * @see EventLoopGroup
- */
-public interface ChannelWriter {
```
pveentjer: we don't need this interface. All communication goes through the channel.

hazelcast/src/main/java/com/hazelcast/internal/networking/AbstractChannel.java
```diff
@@ -14,26 +14,49 @@
...
+public abstract class AbstractChannel implements Channel {
+
+    private static final AtomicIntegerFieldUpdater<AbstractChannel> CLOSED = newUpdater(AbstractChannel.class, "closed");
 
     protected final SocketChannel socketChannel;
     private final ConcurrentMap<?, ?> attributeMap = new ConcurrentHashMap<Object, Object>();
+    private final Set<ChannelCloseListener> closeListeners
+            = newSetFromMap(new ConcurrentHashMap<ChannelCloseListener, Boolean>());
+    private final boolean client;
+    private volatile int closed;
```
mdogan: why not just a bool instead of int? is there any other value other than 1 and 0?

hazelcast/src/main/java/com/hazelcast/internal/networking/Channel.java
```diff
@@ -105,12 +120,58 @@
...
+
+    /**
+     * Checks if this is the 'client' side of a Channel or 'server' side.
+     *
+     * Each channel has a client/server side. The client side is the one that initiated the channel, and the serverside
+     * is the one that 'accepts' the channel.
+     *
+     * @return
+     */
+    boolean isClient();
```
mdogan: `isClient()` name may be confusing,  like Hazelcast client and server. What about something similar to `isClientMode()`? or any other more explicit name...

hazelcast/src/main/java/com/hazelcast/internal/networking/ChannelCloseListener.java
```diff
@@ -16,13 +16,12 @@
...
-    ChannelWriter getChannelWriter();
+/**
+ * A listener called when a {@link Channel} is closed.
+ *
+ * One of the potential usages is to release resources attached to a channel e.g. deregistration of metrics.
+ */
+public interface ChannelCloseListener {
 
-    Channel getChannel();
+    void onClose();
```
mdogan: Shouldn't a listener need to know the channel being closed? `void onClose(Channel channel)`

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionChannelErrorHandler.java
```diff
@@ -0,0 +1,53 @@
...
+        if (channel == null) {
+            // todo: question is if logging is the best solution. If an exception happened without a channel, it is a pretty
+            // big event and perhaps we should shutdown the whole HZ instance.
+            logger.severe(cause);
+        } else {
+            TcpIpConnection connection = (TcpIpConnection) channel.attributeMap().get(TcpIpConnection.class);
+            if (cause instanceof EOFException) {
+                connection.close("Connection closed by the other side", cause);
+            } else {
+                connection.close("Exception in " + getClass().getSimpleName(), cause);
```
mdogan: not `getClass().getSimpleName()` anymore... 
should be `connection.close("Exception in " + channel, cause);`

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionChannelErrorHandler.java
```diff
@@ -0,0 +1,53 @@
...
+        this.logger = logger;
+    }
+
+    @Override
+    public void onError(Channel channel, Throwable cause) {
+        if (cause instanceof OutOfMemoryError) {
+            OutOfMemoryErrorDispatcher.onOutOfMemory((OutOfMemoryError) cause);
+        }
+
+        if (channel == null) {
```
mdogan: Do you know the path where `channel` is null when error is raised?

hazelcast/src/main/java/com/hazelcast/internal/networking/AbstractChannel.java
pveentjer: There is no AtomicBooleanFieldUpdater

hazelcast/src/main/java/com/hazelcast/internal/networking/Channel.java
pveentjer: If you know a better name, I'll do your dishes for a week.

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionChannelErrorHandler.java
pveentjer: Currently there is only 1. The main NIO thread run method.

```
 @Override
    public final void run() {
        // This outer loop is a bit complex but it takes care of a lot of stuff:
        // * it calls runSelectNowLoop or runSelectLoop based on selectNow enabled or not.
        // * handles backoff and retrying in case if io exception is thrown
        // * it takes care of other exception handling.
        //
        // The idea about this approach is that the runSelectNowLoop and runSelectLoop are as clean as possible and don't contain
        // any logic that isn't happening on the happy-path.
        try {
            for (; ; ) {
                try {
                    switch (selectMode) {
                        case SELECT_WITH_FIX:
                            selectLoopWithFix();
                            break;
                        case SELECT_NOW:
                            selectNowLoop();
                            break;
                        case SELECT:
                            selectLoop();
                            break;
                        default:
                            throw new IllegalArgumentException("Selector.select mode not set, use -Dhazelcast.io.selectorMode="
                                    + "{select|selectnow|selectwithfix} to explicitly specify select mode or leave empty for "
                                    + "default select mode.");
                    }
                    // break the for loop; we are done
                    break;
                } catch (IOException nonFatalException) {
                    selectorIOExceptionCount.inc();
                    logger.warning(getName() + " " + nonFatalException.toString(), nonFatalException);
                    coolDown();
                }
            }
        } catch (OutOfMemoryError e) {
            errorHandler.onError(null, e); <--------
        } catch (Throwable e) {
            logger.warning("Unhandled exception in " + getName(), e);
        } finally {
            closeSelector();
        }

        logger.finest(getName() + " finished");
    }
```

hazelcast/src/main/java/com/hazelcast/nio/tcp/TcpIpConnectionChannelErrorHandler.java
pveentjer: Good spotting. 

I'll include the threadname (current thread) and fix the channel to give a better 'toString'

hazelcast/src/main/java/com/hazelcast/internal/networking/ChannelCloseListener.java
pveentjer: Perhaps that would be a bit cleaner. Will reduce litter.

hazelcast/src/main/java/com/hazelcast/internal/networking/Channel.java
pveentjer: isClientSide?

hazelcast/src/main/java/com/hazelcast/internal/networking/AbstractChannel.java
pveentjer: To make it a bit easier to digest I added TRUE/FALSE constants.

hazelcast/src/main/java/com/hazelcast/nio/tcp/MemberChannelInitializer.java
```diff
@@ -41,67 +38,98 @@
...
+        ConcurrentMap attributeMap = channel.attributeMap();
+        ByteBuffer protocolBuffer = (ByteBuffer) attributeMap.get(PROTOCOL_BUFFER);
+        if (protocolBuffer == null) {
+            protocolBuffer = ByteBuffer.allocate(3);
+            attributeMap.put(PROTOCOL_BUFFER, protocolBuffer);
+        }
 
         int readBytes = channel.read(protocolBuffer);
 
+        System.out.println("readBytes:" + readBytes);
```
mdogan: Should be removed before merge.

hazelcast/src/main/java/com/hazelcast/nio/tcp/MemberChannelInitializer.java
```diff
@@ -110,72 +138,118 @@ public MemberChannelInitializer(ILogger logger) {
...
-        IOService ioService = connection.getConnectionManager().getIoService();
-        int sizeKb = CLUSTER.equals(protocol)
-                ? ioService.getSocketSendBufferSize()
-                : ioService.getSocketClientSendBufferSize();
+    private String outboundProtocol(Channel channel) {
+        String protocol = (String) channel.attributeMap().get(PROTOCOL);
+
+        if (protocol == null && channel.isClient()) {
+            // the other side has not yet identified itself, but we are a 'client' member, so the protocol must be CLUSTER.
+            protocol = CLUSTER;
```
mdogan: Protocol is not put into attribute map. Isn't it needed?

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
```diff
@@ -695,5 +690,23 @@ private void onAuthenticationFailed(Address target, ClientConnection connection,
...
+            } else {
+                if (cause instanceof OutOfMemoryError) {
+                    logger.severe(cause);
+                }
 
+                ClientConnection connection = (ClientConnection) channel.attributeMap().get(ClientConnection.class);
+                if (cause instanceof EOFException) {
+                    connection.close("Connection closed by the other side", cause);
+                } else {
+                    connection.close("Exception in " + getClass().getSimpleName(), cause);
```
mdogan: Similar to `TcpIpConnectionChannelErrorHandler`, shouldn't use `getClass().getSimpleName()`.

hazelcast/src/main/java/com/hazelcast/nio/tcp/MemberChannelInitializer.java
```diff
@@ -41,67 +38,98 @@
...
+        this.ioService = ioService;
     }
 
     @Override
-    public InitResult<ChannelInboundHandler> initInbound(TcpIpConnection connection, ChannelReader reader) throws IOException {
-        TcpIpConnectionManager connectionManager = connection.getConnectionManager();
-        IOService ioService = connectionManager.getIoService();
+    public InitResult<ChannelInboundHandler> initInbound(Channel channel) throws IOException {
+        String protocol = inboundProtocol(channel);
+        System.out.println("protocol received:'" + protocol + "'");
```
mdogan: Should be removed before merge.

hazelcast/src/test/java/com/hazelcast/internal/ascii/MemcacheTest.java
```diff
@@ -35,6 +35,7 @@
 import net.spy.memcached.internal.OperationFuture;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
```
mdogan: why ignored?

hazelcast/src/test/java/com/hazelcast/nio/tcp/TcpIpConnection_TransferStressBaseTest.java
```diff
@@ -112,26 +107,26 @@ private void testPackets(long verifyTimeoutInMillis) throws Exception {
...
-                        + reader.getPriorityFramesReadCounter().get());
-                logger.info("reader last read time millis  : " + reader.lastReadTimeMillis());
-
-                assertEquals(expectedNormalPackets, reader.getNormalFramesReadCounter().get());
-                assertEquals(expectedUrgentPackets, reader.getPriorityFramesReadCounter().get());
-            }
-        }, verifyTimeoutInMillis);
-        logger.info("Waiting for pending packets to be sent and received finished in "
-                + (System.currentTimeMillis() - start) + " milliseconds");
+//
```
mdogan: Is this a temporary change?

hazelcast/src/main/java/com/hazelcast/nio/tcp/MemberChannelInitializer.java
pveentjer: I can add it for completion, but it isn't needed.

hazelcast/src/test/java/com/hazelcast/internal/ascii/MemcacheTest.java
pveentjer: I need to enable it again. I was struggling with memcache due to it poking too much in the internals. This has been addressed; but didn't enable the test again.

hazelcast/src/test/java/com/hazelcast/nio/tcp/TcpIpConnection_TransferStressBaseTest.java
pveentjer: Need to enable that again as well. Good spotting.

hazelcast/src/test/java/com/hazelcast/internal/ascii/MemcacheTest.java
pveentjer: unused import.

hazelcast/src/main/java/com/hazelcast/nio/tcp/MemberChannelInitializer.java
```diff
@@ -42,69 +38,96 @@
...
+        this.ioService = ioService;
     }
 
     @Override
-    public InitResult<ChannelInboundHandler> initInbound(TcpIpConnection connection, ChannelReader reader) throws IOException {
-        TcpIpConnectionManager connectionManager = connection.getConnectionManager();
-        final IOService ioService = connectionManager.getIoService();
+    public InitResult<ChannelInboundHandler> initInbound(Channel channel) throws IOException {
+        String protocol = inboundProtocol(channel);
+        System.out.println("protocol received:'" + protocol + "'");
```
mdogan: leftover

pveentjer:
run-lab-run

PR #10563, opened 2017-05-11
Client logs all as warning on cluster connection failure

sancar:
Exceptions except Authentication exception were logged as finest.
The connection problems like SSL misconfiguration etc are not
easily detectable. All converted to WARNING.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnectionManagerImpl.java
@@ -83,7 +83,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.logging.Level;
 
 import static com.hazelcast.client.config.SocketOptions.DEFAULT_BUFFER_SIZE_BYTE;
 import static com.hazelcast.client.config.SocketOptions.KILO_BYTE;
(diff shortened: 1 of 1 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClusterListenerSupport.java
```diff
@@ -200,8 +198,7 @@ private boolean connect(Set<InetSocketAddress> triedAddresses) throws Exception
                 fireConnectionEvent(LifecycleEvent.LifecycleState.CLIENT_CONNECTED);
                 return true;
             } catch (Exception e) {
-                Level level = e instanceof AuthenticationException ? Level.WARNING : Level.FINEST;
-                logger.log(level, "Exception during initial connection to " + inetSocketAddress, e);
+                logger.warning("Exception during initial connection to " + inetSocketAddress + ", exception " + e);
```
pveentjer: Is the connection not logging the exception?

sancar: It is. Actually currently if Authentication is configured wrong. It is printed twice :
```
WARNING: hz.client_0 [PrometheusUATx] [3.9-SNAPSHOT] ClientConnection{alive=false, connectionId=1, channel=NioChannel{/127.0.0.1:59831->/127.0.0.1:5701}, remoteEndpoint=null, lastReadTime=2017-05-29 15:11:51.743, lastWriteTime=2017-05-29 15:11:51.723, closedTime=2017-05-29 15:11:51.753, lastHeartbeatRequested=never, lastHeartbeatReceived=never, connected server version=null} lost. Reason: com.hazelcast.client.AuthenticationException[Invalid credentials! Principal: null]
May 29, 2017 3:11:51 PM com.hazelcast.client.spi.impl.ClusterListenerSupport
WARNING: hz.client_0 [PrometheusUATx] [3.9-SNAPSHOT] Exception during initial connection to /127.0.0.1:5701
com.hazelcast.client.AuthenticationException: Invalid credentials! Principal: null
	at com.hazelcast.client.connection.nio.ClientConnectionManagerImpl$1.onResponse(ClientConnectionManagerImpl.java:576)
```

In ssl, it is reported that we are missing SSL exceptions here. Not sure how it is missed.
There can be other exceptions other than connection get here. Adding membership listener could fail for example. 
I don't think  printing exception twice is a problem here. The other cases should be printed

sancar:
run-lab-run

PR #10564, opened 2017-05-11
Client logs all as warning on cluster connection failure

sancar:
Exceptions except Authentication exception were logged as finest.
The connection problems like SSL misconfiguration etc are not
easily detectable. All converted to WARNING.


sancar:
run-lab-run

PR #10569, opened 2017-05-12
Added Thread leak tests to check for leftover threads from Hazelcast

Donnerbart:
Added a test which check for leftover threads of Hazelcast and if they are joinable.


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

PR #10573, opened 2017-05-12
Minor io improvements

pveentjer:
- moved OutboundFrame into com.hazelcast.internal.networking since it is part of that API
- minor renames in the Spinning logic (internal classes/fields which still had socket/handler in the name)
- added javadoc


hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
```diff
@@ -33,7 +33,7 @@
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.Connection;
 import com.hazelcast.nio.ConnectionType;
-import com.hazelcast.nio.OutboundFrame;
+import com.hazelcast.internal.networking.OutboundFrame;
```
Donnerbart: The import is out-of-order, could you please organize them? Same for a bunch of other classes.

pveentjer:
Thanks for the review @Donnerbart.  I have optimized the imports.

PR #10555, opened 2017-05-10
Implemented Near Cache key store-by-reference in member IMap proxies

Donnerbart:
Depends on #10398

PR #10579, opened 2017-05-15
Cleanup of member map proxy classes

Donnerbart:
Pulled out another bunch of cleanups from https://github.com/hazelcast/hazelcast/pull/10555 to increase the readability of that that PR.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
@@ -373,22 +373,26 @@ abstract class MapProxySupport<K, V>
     }
 
     protected Data putInternal(Data key, Data value, long ttl, TimeUnit timeunit) {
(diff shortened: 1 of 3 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/MapProxySupport.java
```diff
@@ -373,22 +373,26 @@ private Data readBackupDataOrNull(Data key) {
     }
 
     protected Data putInternal(Data key, Data value, long ttl, TimeUnit timeunit) {
-        MapOperation operation = operationProvider.createPutOperation(name, key, value, getTimeInMillis(ttl, timeunit));
+        long timeInMillis = getTimeInMillis(ttl, timeunit);
```
Donnerbart: I pulled out the `getTimeInMillis(ttl, timeunit)` calls because in the follow-up PR these lines will be longer than 130 chars.

Issue #10581, opened 2017-05-15
ClientDelegatingFuture falsely given null response 


sancar:
When two callbacks are registered as callback to ClientDelegatingFuture, 
if one is wants to deserialize the response while the other is not, 
we are falsely giving null response to one waiting a deserialized value. 

Issue can be seen with following test in 3.8 and 3.7

Nearcache internally registers a non-deserializing callback. 
Users callback wants to deserialize the object.
User ends up with null response. 

```
 Config config = new XmlConfigBuilder().build();
        Hazelcast.newHazelcastInstance(config);

        ClientConfig clientConfig = new ClientConfig();

        NearCacheConfig nearCacheConfig = new NearCacheConfig();
        nearCacheConfig.setName("map");
        EvictionConfig evictionConfig = new EvictionConfig();
        evictionConfig.setSize(10000);
        evictionConfig.setEvictionPolicy(EvictionPolicy.LRU);
        nearCacheConfig.setEvictionConfig(evictionConfig);
        nearCacheConfig.setInvalidateOnChange(false);

        clientConfig.addNearCacheConfig(nearCacheConfig);
        HazelcastInstance client = HazelcastClient.newHazelcastClient(clientConfig);

        IMap<Object, AtomicInteger> clientMap = client.getMap("map");

        for (int i = 0; i < 1000; i++) {
            clientMap.put(i, new AtomicInteger(i));

            ICompletableFuture future = clientMap.getAsync(i);
            final int finalI = i;
            future.andThen(new ExecutionCallback<Object>() {
                @Override
                public void onResponse(Object response) {
                    if (response == null) {
                        System.out.println("NULL " + finalI);
                    }
                }

                @Override
                public void onFailure(Throwable t) {
                }
            });
        }


```

Issue #8127, opened 2016-05-08
ClientDelegatingFuture remove


pveentjer:
Forward to the underlying InvocationFuture. No need for this class

pveentjer:
There are some serious problems with this class. 

```
    synchronized (mutex) {
                if (!done || !isResponseSet()) {
                    try {
                        response = resolveMessageToValue(future.get(timeout, unit));
                        if (deserializedValue == null) {
                            deserializedValue = serializationService.toObject(response);
                        }
                    } catch (InterruptedException e) {
                        error = e;
                    } catch (ExecutionException e) {
                        error = e;
                    }
                    done = true;
                }
            }
```

So even if you call with 1 ms timeout, but someone else called with a infinite timeout and it takes a long time for the future to respond, the 1ms timeout can't be obeyed since it will end up in a lock. 

The whole class is fragile due to lots of flags.

pveentjer:
Another problem is that it introduces an intrinsic lock. If there are multiple concurrent get calls or interactions with the andThen callbacks, the intrinsic lock could be promoted to a fat lock and the lock deflation issues are back.

PR #10584, opened 2017-05-15
Fix clientDelegatingFuture problems

sancar:
The old implementation was trying to deserialize only ones and every delegating future was using same deserialized value. 
With this change, every callback and get method deserializes the response again. Initial motivation is to simplify and get rid of lock. Not sure about performance implications: waiting for lock vs using more cpu but parallel work. 
AbstractInvocationFuture seems also deserializes multiple times for each callback so I followed the same logic. 

fixes https://github.com/hazelcast/hazelcast/issues/10581
fixes https://github.com/hazelcast/hazelcast/issues/8127


hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -38,20 +39,14 @@
...
-    private final Object mutex = new Object();
     private final Executor userExecutor;
-    private Throwable error;
-    private V deserializedValue;
-    /**
-     * mutex object is used as an initial NIL value
-     */
-    private volatile Object response = mutex;
-    private volatile boolean done;
+    private final AtomicReference<Object> responseMessage = new AtomicReference<Object>(NULL_OBJECT);
```
pveentjer: please use a Atomicreferencefieldupdater. 

Using an  AtomicReference for every invocation will increase litter and this will increase gc pressure.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -113,22 +108,17 @@ public boolean complete(Object value) {
...
     }
 
     @Override
     public final boolean isDone() {
-        return done ? done : future.isDone();
-    }
-
-    public Object getResponse() {
-        return response;
+        return responseMessage.get() != NULL_OBJECT || future.isDone();
```
pveentjer: why not always forward to the future?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/MessageFlyweight.java
```diff
@@ -52,8 +52,12 @@ public MessageFlyweight() {
         offset = 0;
     }
 
-    public MessageFlyweight wrap(ClientProtocolBuffer buffer, int offset) {
-        this.buffer = buffer;
+    public MessageFlyweight wrap(byte[] buffer, int offset, boolean useUnsafe) {
+        if (useUnsafe) {
```
pveentjer: This would be an excellent place to use a ternery operator so you go from 5 lines to 1.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -38,20 +39,14 @@
  */
 public class ClientDelegatingFuture<V> implements InternalCompletableFuture<V> {
 
+    private static final Object NULL_OBJECT = new Object();
```
pveentjer: I would call this VOID. Because null could afaik be a perfect legal return value.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -144,24 +134,22 @@ public V get() throws InterruptedException, ExecutionException {
...
+            Object response;
+            try {
+                response = future.get(timeout, unit);
+            } catch (InterruptedException e) {
+                response = e;
+            } catch (ExecutionException e) {
+                response = e;
             }
+
+            responseMessage.compareAndSet(NULL_OBJECT, response);
```
pveentjer: I believe this is broken in case of exceptions. So imagine 2 threads waiting for a future and one thread gets interrupted and he therefor sets the 'responseMessage' to that exception. Now the second thread, which isn't interrupted at all.. will also get this interrupted exception.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -189,87 +177,62 @@ public V join() {
...
-        if (deserializedValue != null) {
-            return deserializedValue;
-        }
-        if (deserializeResponse) {
-            // Otherwise, it is possible that received data may not be deserialized
-            // if "shouldDeserializeData" flag is not true in any of registered "DelegatingExecutionCallback".
-            // So, be sure that value is deserialized before returning to caller.
-            deserializedValue = serializationService.toObject(response);
-            return deserializedValue;
+        ClientMessage tempMessage = ClientMessage.createForDecode(clientMessage.buffer(), 0);
```
pveentjer: Is this additional litter?

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -189,87 +177,62 @@ public V join() {
...
-                        error = t;
-                        done = true;
-                    }
-                }
+            complete(t);
+        }
+
+        private void complete(Object remoteResponse) {
+            Object response;
+            if (responseMessage.compareAndSet(NULL_OBJECT, remoteResponse)) {
```
pveentjer: I would not start with a compareAndSet. First do a read; check is a not NULL_RESPONSE is set..if so. you are done. It isn't set, then you do the compareAndSet.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -189,87 +177,62 @@ public V join() {
...
-                synchronized (mutex) {
-                    if (!done) {
-                        error = t;
-                        done = true;
-                    }
-                }
+            complete(t);
+        }
+
+        private void complete(Object remoteResponse) {
```
pveentjer: I would not call this method 'complete'. Complete has a special meaning on the completable future.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
sancar: Creates a new ClientMessage object but wraps internal buffer. The buffer is not copied.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientAddressCancellableDelegatingFuture.java
```diff
@@ -62,18 +62,12 @@ public boolean cancel(boolean mayInterruptIfRunning) {
...
-                return true;
-            }
-            return false;
-        } catch (Exception e) {
-            throw rethrow(e);
-        } finally {
-            setDone();
+        if (cancelSuccessful) {
+            complete(new CancellationException());
+            cancelled = true;
```
mdogan: As far as I understand from the Future contract, regardless of result of cancellation, future should be completed with `CancellationException ` and `future.isCancelled()` return `true`.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -189,86 +142,41 @@ public V join() {
...
+    private Object resolveResponse(ClientMessage clientMessage, boolean deserialize) {
         if (defaultValue != null) {
             return defaultValue;
         }
 
-        // If value is already deserialized, use it.
-        if (deserializedValue != null) {
-            return deserializedValue;
+        ClientMessage tempMessage = ClientMessage.createForDecode(clientMessage.buffer(), 0);
+        Object decodedResponse = clientMessageDecoder.decodeClientMessage(tempMessage);
```
mdogan: `clientMessageDecoder.decodeClientMessage()` creates a copy of internal buffer each time. 

@pveentjer: Which one do you prefer; 
- allocating memory and creating temp Data object on each invocation 
or 
- guarding the method with a lock and creating a single Data object only?

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientPartitionCancellableDelegatingFuture.java
```diff
@@ -60,18 +60,12 @@ public boolean cancel(boolean mayInterruptIfRunning) {
...
-                return true;
-            }
-            return false;
-        } catch (Exception e) {
-            throw rethrow(e);
-        } finally {
-            setDone();
+        if (cancelSuccessful) {
+            complete(new CancellationException());
+            cancelled = true;
```
mdogan: See comment on `ClientAddressCancellableDelegatingFuture`.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientAddressCancellableDelegatingFuture.java
```diff
@@ -51,7 +51,7 @@ public ClientAddressCancellableDelegatingFuture(ClientInvocationFuture future,
 
     @Override
     public boolean cancel(boolean mayInterruptIfRunning) {
```
pveentjer: Why not pass through to the underlying future and not adding any logic to this method?

sancar: It is sending a request to remote to cancel not started executions.
Is there a way to achieve this with underlying future ?

hazelcast-client/src/main/java/com/hazelcast/client/proxy/IExecutorDelegatingFuture.java
```diff
@@ -0,0 +1,115 @@
...
+                    ExecutorServiceCancelOnPartitionCodec.encodeRequest(uuid, partitionId, mayInterruptIfRunning);
+            ClientInvocation clientInvocation = new ClientInvocation(client, request, partitionId);
+            ClientInvocationFuture f = clientInvocation.invoke();
+            return ExecutorServiceCancelOnPartitionCodec.decodeResponse(f.get()).response;
+        } else {
+            ClientMessage request = ExecutorServiceCancelOnAddressCodec.encodeRequest(uuid, target, mayInterruptIfRunning);
+            ClientInvocation clientInvocation = new ClientInvocation(client, request, target);
+            ClientInvocationFuture f = clientInvocation.invoke();
+            return ExecutorServiceCancelOnAddressCodec.decodeResponse(f.get()).response;
+
```
pveentjer: Extra line isn't needed

sancar:
run-lab-run

sancar:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/proxy/IExecutorDelegatingFuture.java
```diff
@@ -0,0 +1,114 @@
...
+        this.context = context;
+        this.uuid = uuid;
+        this.partitionId = partitionId;
+        this.target = null;
+
+    }
+
+    @Override
+    public boolean cancel(boolean mayInterruptIfRunning) {
+        if (isDone()) {
```
pveentjer: If the cancel method is implemented, the isCancelled should be be implemented.

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
```diff
@@ -189,86 +147,51 @@ public V join() {
...
 
-    protected void setError(Throwable error) {
-        this.error = error;
-    }
+    private Object decodeResponse(ClientMessage clientMessage) {
+        if (decodedResponse != VOID) {
+            return decodedResponse;
+        }
+        ClientMessage message = ClientMessage.createForDecode(clientMessage.buffer(), 0);
+        Object newDecodedResponse = clientMessageDecoder.decodeClientMessage(message);
```
pveentjer: What is th type of the 'newDecodedResponse'? Could this be an ordinary user object like a POJO? If so, can't we run into sharing the same instance between multiple threads?

hazelcast-client/src/main/java/com/hazelcast/client/proxy/IExecutorDelegatingFuture.java
sancar: IsCancelled will return true , because i am calling 
` complete(new CancellationException());` 
at the end of this method. 
What else I need to implement ?

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
mdogan: What's wrong with sharing future's value between multiple threads? Isn't it same with JDK executor and futures?

sancar: Not end users POJO. It is mostly Data.It can be Long, Boolean etc also. How could it be a problem anyway ?

sancar:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
pveentjer: I have mixed feelings; so I'm fine with Mehmet's comment. It is up to the end user to not shoot himself in the foot and that is how it works with a regular future. I believe this is how it should be.

But would we be breaking any code that assumes each thread will get its own instance and now we return shared instance.

hazelcast-client/src/main/java/com/hazelcast/client/proxy/IExecutorDelegatingFuture.java
pveentjer: Does the 'isCancelled' method not always return 'false'; as hardcoded value?

hazelcast-client/src/main/java/com/hazelcast/client/util/ClientDelegatingFuture.java
sancar: I don't think there is a problem here. User should not assume this to begin with. And before this pr it was always sharing the same instance, not vice versa.

PR #10592, opened 2017-05-16
Fixed key serialization in StaleReadDetectorImpl

Donnerbart:
`StaleReadDetector.isStaleRead()` is in the fast path of the Near Cache.
The single serialization to get the partition ID is already crushing our
performance. So we cache the `MetaDataContainer` in the `NearCacheRecord` to
have fast access to the actual `UUID` and `invalidationSequence` to perform
the stale read check.

This PR is needed to show the effect of Near Cache keys stored by-reference. If we don't merge that feature this PR will show not much effect and just increase the memory footage.

The memory costs for the speedup is an integer for the partition ID for a fast `MetaDataContainer` lookup per `AbstractNearCacheRecord` (so 4 bytes).

Based on the work of @ahmetmircik in https://github.com/hazelcast/hazelcast/pull/10140

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/record/AbstractNearCacheRecord.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/record/AbstractNearCacheRecord.java
@@ -42,8 +42,10 @@ public abstract class AbstractNearCacheRecord<V> implements NearCacheRecord<V> {
             AtomicLongFieldUpdater.newUpdater(AbstractNearCacheRecord.class, "recordState");
 
     protected long creationTime = TIME_NOT_SET;
-    protected long sequence;
-    protected UUID uuid;
+
+    protected volatile int partitionId;
+    protected volatile long sequence;
+    protected volatile UUID uuid;
 
     protected volatile V value;
     protected volatile long expirationTime = TIME_NOT_SET;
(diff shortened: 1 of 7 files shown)


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

ahmetmircik:
ee port?

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/record/AbstractNearCacheRecord.java
```diff
@@ -42,6 +42,7 @@
             AtomicLongFieldUpdater.newUpdater(AbstractNearCacheRecord.class, "recordState");
 
     protected long creationTime = TIME_NOT_SET;
+    protected int partitionId;
```
vbekiaris: I think `partitionId` can be accessed from different user threads so maybe should be `volatile`?

Donnerbart: Hmm, would be the same with `sequence` and `uuid` then.

vbekiaris: 👍

Donnerbart: So you think we should make all three fields `volatile`? They are just set once, but unfortunately lazily after the instance creation.

vbekiaris: Yes, that 's my understanding. Otherwise there is no guarantee that the values set `onRecordCreate` on one thread will be visible to other threads. Would it be possible to construct the `Record` in a single step and turn the fields `final`?

Donnerbart: I would like that, but I think this clashes with the actual reservation system. Maybe we can refactor this in another PR (Ahmet is also working on a fix on this, which already gets rid of some internal states etc.).

I'll make the fields volatile for now.

PR #10594, opened 2017-05-16
Fix clientDelegatingFuture problems

### LINKED ISSUE #8127: ClientDelegatingFuture remove

### LINKED ISSUE #10581: ClientDelegatingFuture falsely given null response

sancar:
The old implementation was trying to deserialize only ones and every delegating future was using same deserialized value.
With this change, every callback and get method deserializes the response again.
Initial motivation is to simplify and get rid of lock. Not sure about performance implications:
waiting for lock vs using more cpu but parallel work.
AbstractInvocationFuture seems also deserializes multiple times for each callback so I followed the same logic.

fixes #10581
fixes #8127


sancar:
run-lab-run

PR #10596, opened 2017-05-16
Added keys to Near Cache serialization count tests

Donnerbart:
The serialization count test were just measuring the (de)serialization counts of values. With this PR we also measure the key (de)serializations, so we can eventually proof that the Near Cache key store-by-reference will eliminate all serializations in the code path of a simple `put()` and `get()` of all data structures with Near Cache support.


Donnerbart:
run-lab-run

PR #10599, opened 2017-05-17
Prevent lookup of partition ID on LiteMembers with Near Cache

Donnerbart:
With the key store-by-reference feature for the Near Cache the
`NearCachedMapProxyImpl.isOwn()` method introduced an unwanted
serialization of the key to check if a key belongs to this member.
Lite Members can skip this check, since they never have local keys.

This PR is needed to show the effect of Near Cache keys stored by-reference.
If we don't merge that feature this PR will show not much effect.

Depends on https://github.com/hazelcast/hazelcast/pull/10555

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
@@ -29,7 +29,6 @@ import com.hazelcast.map.impl.MapService;
 import com.hazelcast.map.impl.nearcache.MapNearCacheManager;
 import com.hazelcast.map.impl.nearcache.invalidation.InvalidationListener;
 import com.hazelcast.map.impl.nearcache.invalidation.UuidFilter;
-import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.query.Predicate;
 import com.hazelcast.spi.EventFilter;
@@ -63,6 +62,7 @@ public class NearCachedMapProxyImpl<K, V> extends MapProxyImpl<K, V> {
 
     private final boolean cacheLocalEntries;
     private final boolean invalidateOnChange;
+    private final boolean serializeKeys;
 
     private MapNearCacheManager mapNearCacheManager;
     private NearCache<Object, Object> nearCache;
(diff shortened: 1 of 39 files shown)


hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
```diff
@@ -73,6 +75,8 @@
     public NearCachedMapProxyImpl(String name, MapService mapService, NodeEngine nodeEngine, MapConfig mapConfig) {
         super(name, mapService, nodeEngine, mapConfig);
 
+        localMember = getNodeEngine().getLocalMember();
```
mdogan: You should not cache localMember. When lite member is promoted, `localMember` field of ClusterService is updated.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
```diff
@@ -508,7 +512,7 @@ private long tryReserveForUpdate(Data key) {
     }
 
     private boolean cachingAllowedFor(Data key) {
-        return cacheLocalEntries || !isOwn(key);
+        return cacheLocalEntries || localMember.isLiteMember() || !isOwn(key);
```
mdogan: Should be `getNodeEngine().getLocalMember().isLiteMember()`.

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
mdogan: It seems `NodeEngine.getLocalMember()` javadoc is not correct anymore:

```
/**
     * Returns the local member.
     * <p/>
     * The returned value will never change and will never be null.
     *
     * @return the local member.
     */
    Member getLocalMember();
```

hazelcast/src/main/java/com/hazelcast/map/impl/proxy/NearCachedMapProxyImpl.java
Donnerbart: The `getNodeEngine()` seems to have some overhead of a lifecycle check. Also `getLocalMember()` is redirected several times. The `ClusterService` seems to be the source of truth, so I picked `clusterService.getLocalMember().isLiteMember()` and cache the `ClusterService` as field. I hope this is the most performant and correct solution :)

Donnerbart:
With the Near Cache key store-by-reference the keys can be in Object or Data format. But the `isOwn()` check needs a serialized key for the partition lookup. So there will be an on-demand serialization in that case. This PR prevents this at least for Lite Members.
screenshot from 2017-05-17 11-39-00

Donnerbart:
I've added this small fix into https://github.com/hazelcast/hazelcast/pull/10555

PR #10604, opened 2017-05-17
Fix rescheduling on stopped tasks after migrating back to origin

### LINKED ISSUE #10603: Scheduled task remains cancelled after migration
I'm trying to use hazelcast ScheduledExecutorService to execute some periodic tasks. I'm using hazelcast 3.8.1.

I start one node and then the other, and the tasks are distributed between both nodes and properly executed.

If I shutdown the first node, then the second one will start to execute the periodic tasks that were previously on the first node.

The problem is that, if I stop the second node instead of the first, then its tasks are not rescheduled to the first one. This happens even if I have more nodes. If I shutdown the last node to receive tasks, those tasks are lost.

The shutdown is always done with ctrl+c

I've created a test application, with some sample code from hazelcast examples and with some pieces of code I've found on the web. I start two instances of this app.
```
public class MasterMember {

/**
 * The constant LOG.
 */
final static Logger logger = LoggerFactory.getLogger(MasterMember.class);

    Config config = new Config();
    config.setProperty("hazelcast.logging.type", "slf4j");
    config.getScheduledExecutorConfig("scheduler").
    setPoolSize(16).setCapacity(100).setDurability(1);

    final HazelcastInstance instance = Hazelcast.newHazelcastInstance(config);

    Runtime.getRuntime().addShutdownHook(new Thread() {

        HazelcastInstance threadInstance = instance;

        @Override
        public void run() {
            logger.info("Application shutdown");

            for (int i = 0; i < 12; i++) {
                logger.info("Verifying whether it is safe to close this instance");
                boolean isSafe = getResultsForAllInstances(hzi -> {
                    if (hzi.getLifecycleService().isRunning()) {
                        return hzi.getPartitionService().forceLocalMemberToBeSafe(10, TimeUnit.SECONDS);
                    }
                    return true;
                });

                if (isSafe) {
                    logger.info("Verifying whether cluster is safe.");
                    isSafe = getResultsForAllInstances(hzi -> {
                        if (hzi.getLifecycleService().isRunning()) {
                            return hzi.getPartitionService().isClusterSafe();
                        }
                        return true;
                    });

                    if (isSafe) {
                        System.out.println("is safe.");
                        break;
                    }
                }

                try {
                    Thread.sleep(1000);
                    // TODO Auto-generated catch block
                    e.printStackTrace();
                }
            }

            threadInstance.shutdown();

        }

        private boolean getResultsForAllInstances(
                Function<HazelcastInstance, Boolean> hazelcastInstanceBooleanFunction) {

            return Hazelcast.getAllHazelcastInstances().stream().map(hazelcastInstanceBooleanFunction).reduce(true,
                    (old, next) -> old && next);
        }
    });

    new Thread(() -> {

        try {
            Thread.sleep(10000);
            // TODO Auto-generated catch block
            e.printStackTrace();
        }

        IScheduledExecutorService scheduler = instance.getScheduledExecutorService("scheduler");
        scheduler.scheduleAtFixedRate(named("1", new EchoTask("1")), 5, 10, TimeUnit.SECONDS);
        scheduler.scheduleAtFixedRate(named("2", new EchoTask("2")), 5, 10, TimeUnit.SECONDS);
        scheduler.scheduleAtFixedRate(named("3", new EchoTask("3")), 5, 10, TimeUnit.SECONDS);
        scheduler.scheduleAtFixedRate(named("4", new EchoTask("4")), 5, 10, TimeUnit.SECONDS);
        scheduler.scheduleAtFixedRate(named("5", new EchoTask("5")), 5, 10, TimeUnit.SECONDS);
        scheduler.scheduleAtFixedRate(named("6", new EchoTask("6")), 5, 10, TimeUnit.SECONDS);
    }).start();

    new Thread(() -> {

        try {
            // delays init
            Thread.sleep(20000);

            while (true) {

                IScheduledExecutorService scheduler = instance.getScheduledExecutorService("scheduler");
                final Map<Member, List<IScheduledFuture<Object>>> allScheduledFutures =
                        scheduler.getAllScheduledFutures();

                // check if the subscription already exists as a task, if so, stop it
                for (final List<IScheduledFuture<Object>> entry : allScheduledFutures.values()) {
                    for (final IScheduledFuture<Object> objectIScheduledFuture : entry) {
                        logger.info(
                                "TaskStats: name {} isDone() {} isCanceled() {} total runs {} delay (sec) {} other statistics {} ",
                                objectIScheduledFuture.getHandler().getTaskName(), objectIScheduledFuture.isDone(),
                                objectIScheduledFuture.isCancelled(),
                                objectIScheduledFuture.getStats().getTotalRuns(),
                                objectIScheduledFuture.getDelay(TimeUnit.SECONDS),
                                objectIScheduledFuture.getStats());
                    }
                }

                Thread.sleep(15000);

            }

            // TODO Auto-generated catch block
            e.printStackTrace();
        }

    }).start();

    while (true) {
        Thread.sleep(1000);
    }
    // Hazelcast.shutdownAll();
}
}
```

```
public class EchoTask implements Runnable, Serializable {

/**
 * serialVersionUID
 */
private static final long serialVersionUID = 5505122140975508363L;

final Logger logger = LoggerFactory.getLogger(EchoTask.class);

private final String msg;

public EchoTask(String msg) {
    this.msg = msg;
}

@Override
public void run() {
    logger.info("--> " + msg);
}
}
```

tkountis:
Starting a task on a single node, and then adding one more member, causes some of the tasks to migrate to partitions owned by the latter member. 
During this process, the container stops currently running task, leaving it in a `cancelled` state. If this state, doesn't get disposed (eg. cluster never gets bigger), when the latter member goes down, the tasks will migrate back to the first one. However, since their state there is marked as cancelled, they never get re-scheduled.

The fix, introduces a `stop` method, which still cancels the task, but doesn't interact with the state.
Fix #10603 

Also, minor cleanup and namings.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskStatisticsImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskStatisticsImpl.java
@@ -39,11 +39,11 @@ public class ScheduledTaskStatisticsImpl
 
     private long lastRunEnd;
 
-    private long lastIdleTime;
+    private long lastIdleDuration;
 
-    private long totalRunTime;
+    private long totalRunDuration;
 
-    private long totalIdleTime;
+    private long totalIdleDuration;
 
     public ScheduledTaskStatisticsImpl() {
     }
@@ -57,9 +57,9 @@ public class ScheduledTaskStatisticsImpl
     public ScheduledTaskStatisticsImpl(long runs, long lastIdleTimeNanos, long totalRunTimeNanos,
                                        long totalIdleTimeNanos) {
         this.runs = runs;
-        this.lastIdleTime = lastIdleTimeNanos;
-        this.totalRunTime = totalRunTimeNanos;
-        this.totalIdleTime = totalIdleTimeNanos;
+        this.lastIdleDuration = lastIdleTimeNanos;
+        this.totalRunDuration = totalRunTimeNanos;
+        this.totalIdleDuration = totalIdleTimeNanos;
     }
 
     ScheduledTaskStatisticsImpl(long createdAt, long runs, long firstRunStartNanos, long lastRunStartNanos,
@@ -70,9 +70,9 @@ public class ScheduledTaskStatisticsImpl
         this.firstRunStart = firstRunStartNanos;
         this.lastRunStart = lastRunStartNanos;
         this.lastRunEnd = lastRunEndNanos;
-        this.lastIdleTime = lastIdleTimeNanos;
-        this.totalRunTime = totalRunTimeNanos;
-        this.totalIdleTime = totalIdleTimeNanos;
+        this.lastIdleDuration = lastIdleTimeNanos;
+        this.totalRunDuration = totalRunTimeNanos;
+        this.totalIdleDuration = totalIdleTimeNanos;
     }
 
     @Override
@@ -88,17 +88,17 @@ public class ScheduledTaskStatisticsImpl
 
     @Override
     public long getLastIdleTime(TimeUnit unit) {
-        return unit.convert(lastIdleTime, MEASUREMENT_UNIT);
+        return unit.convert(lastIdleDuration, MEASUREMENT_UNIT);
     }
 
     @Override
     public long getTotalIdleTime(TimeUnit unit) {
-        return unit.convert(totalIdleTime, MEASUREMENT_UNIT);
+        return unit.convert(totalIdleDuration, MEASUREMENT_UNIT);
     }
 
     @Override
     public long getTotalRunTime(TimeUnit unit) {
-        return unit.convert(totalRunTime, MEASUREMENT_UNIT);
+        return unit.convert(totalRunDuration, MEASUREMENT_UNIT);
     }
 
 
@@ -116,18 +116,18 @@ public class ScheduledTaskStatisticsImpl
     public void writeData(ObjectDataOutput out)
             throws IOException {
         out.writeLong(runs);
-        out.writeLong(lastIdleTime);
-        out.writeLong(totalIdleTime);
-        out.writeLong(totalRunTime);
+        out.writeLong(lastIdleDuration);
+        out.writeLong(totalIdleDuration);
+        out.writeLong(totalRunDuration);
     }
 
     @Override
     public void readData(ObjectDataInput in)
             throws IOException {
         runs = in.readLong();
-        lastIdleTime = in.readLong();
-        totalIdleTime = in.readLong();
-        totalRunTime = in.readLong();
+        lastIdleDuration = in.readLong();
+        totalIdleDuration = in.readLong();
+        totalRunDuration = in.readLong();
     }
 
     @Override
@@ -139,8 +139,8 @@ public class ScheduledTaskStatisticsImpl
     public void onBeforeRun() {
         long now = System.nanoTime();
         this.lastRunStart = now;
-        this.lastIdleTime = now - (lastRunEnd != 0L ? lastRunEnd : createdAt);
-        this.totalIdleTime += lastIdleTime;
+        this.lastIdleDuration = now - (lastRunEnd != 0L ? lastRunEnd : createdAt);
+        this.totalIdleDuration += lastIdleDuration;
 
         if (this.firstRunStart == 0L) {
             this.firstRunStart = this.lastRunStart;
@@ -155,7 +155,7 @@ public class ScheduledTaskStatisticsImpl
 
         this.lastRunEnd = now;
         this.runs++;
-        this.totalRunTime += lastRunTime;
+        this.totalRunDuration += lastRunTime;
 
     }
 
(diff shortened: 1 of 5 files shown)


hazelcast/src/main/java/com/hazelcast/scheduledexecutor/impl/ScheduledTaskStatisticsImpl.java
```diff
@@ -165,8 +165,10 @@ public ScheduledTaskStatisticsImpl snapshot() {
 
     @Override
     public String toString() {
-        return "ScheduledTaskStatisticsImpl{ runs=" + runs + ", createdAt="
-                + createdAt + ", firstRunStart=" + firstRunStart + ", lastRunStart=" + lastRunStart + ", lastRunEnd=" + lastRunEnd
-                + ", lastIdleTime=" + lastIdleTime + ", totalRunTime=" + totalRunTime + ", totalIdleTime=" + totalIdleTime + '}';
+        return "ScheduledTaskStatisticsImpl{ runs=" + runs
```
Donnerbart: `ScheduledTaskStatisticsImpl{runs=` without a space matches the majority of our `toString()` methods.

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceSlowTest.java
```diff
@@ -370,4 +373,27 @@ public void cancelUninterruptedTask_waitUntilRunCompleted_killMember_checkStatus
...
+            scheduler.scheduleAtFixedRate(named(valueOf(i), new EchoTask()), 5, 10, SECONDS);
+        }
+
+        assertTrueEventually(new AllTasksRunning(scheduler));
+
+        // Start a second member
+        HazelcastInstance second = factory.newHazelcastInstance();
+        waitAllForSafeState(first, second);
+
+        // Kill the second member, tasks should now get rescheduled back in first member
```
Donnerbart: How can you be sure the task was migrated to the second member in between? `assertTrueEventually(new AllTasksRunning(scheduler));` should also pass if there was no migration at all, shouldn't it?

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceSlowTest.java
```diff
@@ -370,4 +373,27 @@ public void cancelUninterruptedTask_waitUntilRunCompleted_killMember_checkStatus
         assertTrue(future.isDone());
     }
 
+    @Test
+    public void reschedulingAfterMigration_whenCurrentNodePreviouslyOwnedTask() {
+        TestHazelcastInstanceFactory factory = createHazelcastInstanceFactory();
+        HazelcastInstance first = factory.newHazelcastInstance();
+
+        int tasksCount = 1000;
+        final IScheduledExecutorService scheduler = getScheduledExecutor(new HazelcastInstance[] {first }, "scheduler");
```
Donnerbart: I thought there would be more logic inside the `getScheduledExecutor()` method, but since it isn't I would just use `first.getScheduledExecutorService("scheduler");`

tkountis: its used in the Client tests and its slightly different there. eg. ClientScheduledExecutorServiceBasicTest

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceSlowTest.java
tkountis: Well, I am not sure, I can add an extra check to see if there are tasks in the second member, specifically, but I tried to keep the taskCount rather high so that at least some will migrate. But yes, its not deterministic. I will modify to make it so.

hazelcast/src/test/java/com/hazelcast/scheduledexecutor/ScheduledExecutorServiceSlowTest.java
Donnerbart: I don't want to remove the whole method. But this test looks like it doesn't need it, since it requires to construct an array (which looks more complex than just retrieving the scheduler).

Batter2014:
Guys, 
i'm currently getting:
java.lang.NullPointerException
	at com.hazelcast.scheduledexecutor.impl.ScheduledTaskDescriptor.stopForMigration(ScheduledTaskDescriptor.java:146) ~[hazelcast-all-3.8.3.jar:3.8.3]

When trying to schedule the Task:
IScheduledExecutorService service = hazelcastInstance.getScheduledExecutorService(executorName);
service.scheduleAtFixedRate(new HzcastTimerExchangeSender(this.getEndpoint().getEndpointUri(), executorName), endpoint.getDelay(), endpoint.getPeriod(), TimeUnit.MILLISECONDS);

I have looked to the fix you have here and it doesn't look safe and could cause issue i have.

You had:
try {
 -                        descriptor.cancel(true);
 -                        descriptor.setScheduledFuture(null);
 -                        descriptor.setTaskOwner(false);
                      } catch (Exception ex) {
                          throw rethrow(ex);
                      }

And now:
try {
 +                        descriptor.stopForMigration();
                } catch (Exception ex) {
                          throw rethrow(ex);
                      }
Where descriptor.stopForMigration() has:
void stopForMigration() {
 +        // Result is not set, allowing task to get re-scheduled, if/when needed.
 +        this.isTaskOwner = false;
 +        this.future.cancel(true); //Nullpointer here
 +        this.future = null;
 +    }

And in old code :
descriptor.cancel(true);

Which has:
boolean cancel(boolean mayInterrupt)
        if (!resultRef.compareAndSet(null, new ScheduledTaskResult(true)) || future == null) {
            return false;
        }

        return future.cancel(mayInterrupt);
    }

This look way safer. 

Nullpointer i have on 2 machines which have code you can see above.

mmedenjak:
@Batter2014 can you please submit a new issue for this?

Batter2014:
Yep, just did that: https://github.com/hazelcast/hazelcast/issues/11047

PR #10605, opened 2017-05-17
Added an invalidation parameter to the Near Cache serialization tests

Donnerbart:
If invalidations are not enabled, the `StaleReadDetectorImpl` is not enabled. This might hide key serialization we have to check for the Near Cache key store-by-reference feature.


Donnerbart:
run-lab-run

PR #10611, opened 2017-05-18
Client statistics needs to be communicated to the cluster

ihsandemir:
Implemented the client statistics. Introduces two configuration parameters: hazelcast.client.statistics.enabled and hazelcast.client.statistics.period.seconds.


hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
```diff
@@ -242,6 +244,8 @@ public HazelcastClientInstanceImpl(ClientConfig config,
         lockReferenceIdGenerator = new ClientLockReferenceIdGenerator();
         nearCacheManager = clientExtension.createNearCacheManager();
         clientExceptionFactory = initClientExceptionFactory();
+
+        statistics = new Statistics(this);
```
pveentjer: Be careful with passing this in a constructor to another object. Easy to run into a partial constructed object (one of the Effective Java anti patterns)

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+
+        staticStats.put("/enterprise", Boolean.toString(enterprise));
+        // Register for one time statistics. These are resent only when reconnected to the cluster.
+        registerStaticStatistics();
+
+        registerPeriodicMetrics();
+
+        logger.finest("Client statistics is enabled with period " + periodSeconds + " seconds.");
+
+        // Start task for sending periodic statistics
```
pveentjer: If you need a comment for the line below, perhaps it makes more sense to just rename the method to the comment :)

startTaskForSendingPeriodicStatistics.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+        scheduleStatisticsCollection(periodSeconds);
+    }
+
+    private void scheduleStatisticsCollection(long periodSeconds) {
+        client.getExecutionService().scheduleWithRepetition(new Runnable() {
+            @Override
+            public void run() {
+                final Map<String, String> stats = new HashMap<String, String>();
+
+                metricsRegistry.render(new ProbeRenderer() {
```
pveentjer: I think this approach is going to lead to problems because you have no control on the amount of probes that are going to be rendered. There could be a huge number; e.g. a client connecting to many servers; then for each connection there will easily be like 10/20 probes.

As we discussed, I believe it is best if you access the concrete probes you are looking for. Retrieve them once and recycle them, like is done in the HealthMonitor.HealthMetrics

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+                        logger.finest("Exception while processing statistics " + name + ". Error:" + e.getMessage());
+                    }
+
+                    @Override
+                    public void renderNoValue(String name) {
+                    }
+                });
+
+                // put user executor queue size
+                ThreadPoolExecutor userExecutor = (ThreadPoolExecutor) client.getClientExecutionService().getUserExecutor();
```
pveentjer: Probably better to get this apparently important value defined as a probe. Now you only get this data in case of client statistics it is not available in regular diagnostics.

hazelcast/src/main/java/com/hazelcast/client/ClientEndpoint.java
```diff
@@ -105,6 +107,17 @@ void authenticated(ClientPrincipal principal, Credentials credentials, boolean f
      *
      * @param version The version string as obtained from the environment
      */
-    void setClientVersion(String version);
+    void setClientVersion(java.lang.String version);
```
pveentjer: Are you expecting to import any other String? :) I would get rid of the java.lang

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+import com.hazelcast.spi.properties.HazelcastProperty;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ThreadPoolExecutor;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Created by ihsan on 15/05/17.
```
pveentjer: We can find the author using the git history.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+ */
+public class Statistics {
+    private final boolean enabled;
+    private final HazelcastProperties properties;
+    private final ILogger logger = Logger.getLogger(this.getClass());
+    private final HazelcastClientInstanceImpl client;
+    private final boolean enterprise;
+    public static final String PREFIX = "hazelcast.client.statistics";
```
pveentjer: Checkstyle is going to choke on this. Statics should be defined first.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+
+    public final static String NEAR_CACHE_CATEGORY_PREFIX = "/nearcache";
+
+    private final MetricsRegistry metricsRegistry = new MetricsRegistryImpl(Logger.getLogger(this.getClass()),
```
pveentjer: This seems a bit strange to me. Why does statistics create its own MetricsRegistry? Why don't you make use of the one that is available in the HazelcastClientInstanceImpl

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+            ProbeLevel.MANDATORY);
+    Map<String, String> staticStats = new HashMap<String, String>(1);
```
pveentjer: No access modifier even not final?

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+     * The period in seconds the statictics runs.
+     * <p>
+     * The StatisticsPlugin periodically sending the client statistics to the owner member
+     * <p>
+     * <p>
+     * If set to 0, the plugin is disabled.
+     */
+    public static final HazelcastProperty PERIOD_SECONDS = new HazelcastProperty(PREFIX + ".period.seconds", 60, SECONDS);
+
+    public Statistics(HazelcastClientInstanceImpl client) {
```
pveentjer: I would not inject HazelcastClientInstanceImpl, but the concrete dependencies to prevent coupling with the HazelcastClientInstanceImpl.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+
+                // addNearCacheStats
+                addNearCachStats(stats);
+
+                sendStats(stats);
+            }
+        }, 0, periodSeconds, SECONDS);
+    }
+
+    private void addNearCachStats(Map<String, String> stats) {
```
pveentjer: Rendering to a StringBuffer would safe a lot of intermediate string and support objects from being created.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    private void registerStaticStatistics() {
+        // The client may have already connected, hence, send the stat at least once at startup
+        sendStats(staticStats);
+
+        client.getLifecycleService().addLifecycleListener(new LifecycleListener() {
+            @Override
+            public void stateChanged(LifecycleEvent event) {
+                if (LifecycleEvent.LifecycleState.CLIENT_CONNECTED.equals(event.getState())) {
```
pveentjer: I normally do static import of the enum to reduce horizontal noise.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+
+    private void sendStats(Map<String, String> oneTimeStats) {
+        ClientMessage request = ClientStatisticsCodec.encodeRequest(oneTimeStats.entrySet());
+        Address ownerConnectionAddress = client.getClientClusterService().getOwnerConnectionAddress();
+        if (null != ownerConnectionAddress) {
+            try {
+                new ClientInvocation(client, request, ownerConnectionAddress).invoke().get();
+            } catch (Exception e) {
+                // suppress exception, do not print too many messages
+                logger.finest("Could not send stats " + e.getMessage());
```
pveentjer: Best to add a logging guard and to add the exception as cause. This way you can at least see the exception if somebody puts logging on finest. Discarding stacktraces makes it a pita to look for a problem

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+            }
+        }
+    }
+
+    private void registerPeriodicMetrics() {
+        RuntimeMetricSet.register(metricsRegistry);
+        OperatingSystemMetricSet.register(metricsRegistry);
```
pveentjer: This should not be needed if would not create your own MetricsRegistry

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+
+        // register user executor queue size
+        metricsRegistry
```
pveentjer: This need probe you define, you should define it on the client version so we get this new probe in general diagnostics

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,235 @@
...
+                .register("User Executor", "/userExecutor/queueSize", ProbeLevel.MANDATORY, new LongProbeFunction<String>() {
```
pveentjer: Why did you decide to use '/' as delimiter and not a  '.'

Now we have another style in the system.

hazelcast-client/src/test/java/com/hazelcast/client/statistics/ClientStatisticsTest.java
```diff
@@ -0,0 +1,248 @@
...
+        Assert.assertNull(cacheHits);
+
+        IMap<Integer, Integer> map = client.getMap(testMapName);
+
+        // Wait enough time for statistics collection
+        sleepSeconds(statsPeriodSeconds + 1);
+
+        stats = getStats(client, clientEngine);
+        mapHits = stats.get("/nearcache/" + testMapName + "/Hits");
+        Assert.assertNotNull(mapHits);
```
pveentjer: use static imports for asserts

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
ihsandemir: I initially had the explicit parameters passed but it came out to be five of them (ClientExecutionService, ClusterService, LifeCycleService, NearCacheManager, ClientProperties) which I thought would be cleaner to pass "this" instead of five parameters. Partial construction will not be a problem for this case but I understand the concern. In addition, "this" is passed HazelcastClientInstanceImpl as a parameter to a number of objects. Do you think it is still better to pass five parameters instead of "this"?

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, changed the name to schedulePeriodicStatisticsSendTask.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: Yes, I agree, we can add this to Diagnostics. It will be better to send a separate PR for this change, since changing the Diagnostics related statistics is not in the scope of this PR.

hazelcast/src/main/java/com/hazelcast/client/ClientEndpoint.java
ihsandemir: Ok, corrected. I think IDE refactoring caused it and I missed this one.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, Removed.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: The set of statistics to be collected and sent for client statistics is different from the diagnostics set of metrics, hence, i kept them as separate.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, corrected.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, made private final.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: Ok, changed to use StringBuffer.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: Ok, changed.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, changed.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: Need our own metrics set.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: Will do in another PR to add to diagnostics.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, changing it to be the same a Diagnostics usage.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
Donnerbart: If you don't need thread-safety, please use `StringBuilder` instead. `StringBuffer` is the same, but with synchronized methods.

hazelcast-client/src/test/java/com/hazelcast/client/statistics/ClientStatisticsTest.java
ihsandemir: OK, changed.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, changed.

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
pveentjer: Passing concrete dependencies is in almost all cases a better approach.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
pveentjer: Jit probably gets rid of the locks and even if it couldn't it would be very cheap because there is only a single thread needing the 'lock'. But yes.. StringBuilder is better.. somehow I keep forgetting it.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
pveentjer: Having a single shared set of metrics and just filtering out the elements you need make more sense. Duplication a pain. 

Also I would not grab all probes; but just the focus on the most important ones.

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
ihsandemir: @pveentjer I tried not using "this" but I end up needing it due to this line: https://github.com/ihsandemir/hazelcast/blob/clientStats/hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java#L196  ClientInvocation constructor needs ClientInstance, hence it seems like I can not get rid of it unless ClientInvocation is also changed. We can open a cleanup issue for this since this change affects a number of different places ("this" is used a lot in HazelcastClientInstanceImpl).

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, changed.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: Adding filtering capability to the MetricsRegistry would be a nice feature but I think we need to do it in another PR.

hazelcast/src/main/java/com/hazelcast/client/impl/ClientEngineImpl.java
```diff
@@ -588,6 +590,32 @@ public Operation getPostJoinOperation() {
         return resultMap;
     }
 
+    @Override
+    public List<Map.Entry<String, List<Map.Entry<String, String>>>> getClientStatistics() {
+        Collection<ClientEndpoint> clientEndpoints = endpointManager.getEndpoints();
+        List<Map.Entry<String, List<Map.Entry<String, String>>>> statsList = new ArrayList<Map.Entry<String, List<Map.Entry<String, String>>>>(
+                clientEndpoints.size());
+        for (final ClientEndpoint e : clientEndpoints) {
+            statsList.add(new Map.Entry<String, List<Map.Entry<String, String>>>() {
```
sancar: Instead of Entry interface use AbstractMap.SimpleEntry it also has correct hashcode and equals implementations which is missing here.

ihsandemir: OK, done.

hazelcast-client/src/main/java/com/hazelcast/client/impl/HazelcastClientInstanceImpl.java
pveentjer: I know the problem. In some parts on the serverside we also have this issue where 'this' is being passed to constructor.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,246 @@
...
+    public final static int FEATURE_SUPPORTED_SINCE_VERSION = BuildInfo.calculateVersion(FEATURE_SUPPORTED_SINCE_VERSION_STRING);
+
+    private final boolean enabled;
+    private final HazelcastProperties properties;
+    private final ILogger logger = Logger.getLogger(this.getClass());
+    private final HazelcastClientInstanceImpl client;
+
+    private final boolean enterprise;
+
+    private final MetricsRegistry metricsRegistry = new MetricsRegistryImpl(Logger.getLogger(this.getClass()),
```
pveentjer: I'm not convinced this is the right approach. Creating a new instance just for statistics smells like a design problem. Also keep in mind that a MetricsRegistry has its own thread; so by creating another instance you add another thread to the system.

So I would reuse the existing metrics and filter out all the probes that are mandatory; should be part of the PRD. Minimize this so we can add new ones in the future.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,246 @@
...
+        if (null == ownerConnectionAddress) {
+            return null;
+        }
+        ClientConnection connection = (ClientConnection) client.getConnectionManager().getConnection(ownerConnectionAddress);
+        if (null == connection) {
+            return null;
+        }
+
+        int serverVersion = connection.getConnectedServerVersion();
+        if (serverVersion < FEATURE_SUPPORTED_SINCE_VERSION) {
```
pveentjer: Could it be that this is going to be logged potentially every x seconds if someone has enabled finest?

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,246 @@
...
+                final Map<String, String> stats = new HashMap<String, String>();
+
+                stats.put("lastStatisticsCollectionTime", Long.toString(System.currentTimeMillis()));
+
+                stats.putAll(staticStats);
+
+                periodicStats.fillMetrics(stats);
+
+                ThreadPoolExecutor userExecutor = (ThreadPoolExecutor) client.getClientExecutionService().getUserExecutor();
+                stats.put("userExecutor.queueSize", Long.toString(userExecutor.getQueue().size()));
```
pveentjer: This should have been exposed as a probe so that also the regular diagnostics can read out this value. THis way the design doesn't get riddles with ad hoc fixes.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,246 @@
...
+            if (name.startsWith("/")) {
+                pathPrefix += name.substring(1);
+            } else {
+                pathPrefix += name;
+            }
+            pathPrefix += ".";
+
+            NearCacheStats nearCacheStats = nearCache.getNearCacheStats();
+
+            StringBuilder buffer = new StringBuilder();
```
pveentjer: Why is the buffer created in the loop? If you create it outside the loop, you get less litter.

ihsandemir: OK, changed.

hazelcast/src/main/java/com/hazelcast/internal/metrics/metricsets/RuntimeMetricSet.java
```diff
@@ -97,5 +106,10 @@ public long get(RuntimeMXBean runtimeMXBean) {
                     }
                 }
         );
+        registeredMetrics.add("runtime.uptime");
+    }
+
+    public static Set<String> getRegisteredMetricNames() {
```
pveentjer: I would just check out the probes as in HealthMonitor.HealthMetrics.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: yes, that is correct. Alternatives:
1. No logging: But then how can we know if this was the reason of not getting the statistics populated.
2. Do the log only once per connection: But then this would require unnecessary lookup table and such.

Do you have any other suggestion?

1.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: Which metric set do we want to put it to?

hazelcast/src/main/java/com/hazelcast/internal/metrics/metricsets/RuntimeMetricSet.java
ihsandemir: In HealthMetrics, the metric names are being duplicated. Is this what you like, duplicating the names? All I need is to get all the statistics that are registered for RuntimeMetricSet and OperatingSystemMetricSet.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
pveentjer: You can define a probe directly on the metrics system.

```
 <S> void register(S source, String name, ProbeLevel level, LongProbeFunction<S> probe);
```

hazelcast/src/main/java/com/hazelcast/internal/metrics/metricsets/RuntimeMetricSet.java
pveentjer: This duplication is fine because these are the keys of which the probes can be used; it is their official contract.

pveentjer: If you really want to be lazy; just walk over all probes and filter out the probes you don't want based on their name. If it doesn't start with e.g. 'os' then you ignore it. The probes you like you store in a hashmap.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, done.

hazelcast/src/main/java/com/hazelcast/internal/metrics/metricsets/RuntimeMetricSet.java
ihsandemir: OK, changed to use hard coded statistic names.

hazelcast/src/main/java/com/hazelcast/internal/metrics/StringGauge.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
```
pveentjer: Why is a new gauge needed? I don't think this is the right approach.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,254 @@
...
+            if (name.startsWith("/")) {
+                pathPrefix += name.substring(1);
+            } else {
+                pathPrefix += name;
+            }
+            pathPrefix += ".";
+
+            NearCacheStats nearCacheStats = nearCache.getNearCacheStats();
+
+            stats.put(buffer.append(pathPrefix).append("creationTime").toString(),
```
pveentjer: I see a lot of litter being created in this code. Lots of Strings and underlying byte-arrays.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
```diff
@@ -0,0 +1,254 @@
...
+
+        private final Map<String, StringGauge> allMetrics = new HashMap<String, StringGauge>(STATISTIC_NAMES.length);
+
+        PeriodicStatistics(final MetricsRegistry metricsRegistry) {
+            for (String name : STATISTIC_NAMES) {
+                allMetrics.put(name, metricsRegistry.newStringGauge(name));
+            }
+        }
+
+        void fillMetrics(final Map<String, String> metricsMap) {
```
pveentjer: I'm questioning the type of the 'metricsMap'.

If you would just send a formatted String to the other side, then only a single String is needed.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
```diff
@@ -103,6 +108,11 @@ public ExecutorService getUserExecutor() {
         return userExecutor;
     }
 
+    @Probe (level = MANDATORY)
```
pveentjer: Very nice. 

You are allowed to return int; queue.size will also return int.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
pveentjer: Another approach would be to limit the rate of this exception to be printed. E.g. print it at most once a minute.

hazelcast/src/main/java/com/hazelcast/internal/metrics/StringGauge.java
ihsandemir: There is no object creation after start for the gauge. It is initialized and reused, hence it should not cause any litter. The gauge is cached in Statistics.PeriodicStatistics object. 

The reason why defined the string gauge is this: in construction at PeriodicStatistics I only know the name of the gauge but do not know if the type is long or double, hence StringGauge parses the value to string correctly based on the gauge type.

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/ClientExecutionServiceImpl.java
ihsandemir: OK, changed.

hazelcast-client/src/main/java/com/hazelcast/client/impl/statistics/Statistics.java
ihsandemir: OK, cached the last connected server address and only printing the message if the last cached Address is different from the current connected server.

hazelcast-client/src/main/java/com/hazelcast/client/connection/nio/ClientConnection.java
```diff
@@ -79,6 +80,7 @@ public ClientConnection(HazelcastClientInstanceImpl client,
         channel.attributeMap().put(ClientConnection.class, this);
         this.connectionId = connectionId;
         this.logger = client.getLoggingService().getLogger(ClientConnection.class);
+        this.startTime = System.currentTimeMillis();
```
pveentjer: Flavor warning: 

I try to initialize fields when I can to reduce the size of the constructor.

ihsandemir: OK, updated.

hazelcast/src/main/java/com/hazelcast/internal/metrics/StringGauge.java
ihsandemir: OK, removed StringGauge and implemented render method in the Long and Double gauges as suggested.

ihsandemir:
run-lab-run

PR #10613, opened 2017-05-18
Small cleanup of AbstractNearCacheRecordStore methods

Donnerbart:
Made some methods `private` to ease refactoring (with `protected` you never know if there is an overridden method). Cleaned up redirection madness in `clear()` and `destroy()` methods. No EE changes are needed.


Donnerbart:
run-lab-run

PR #10616, opened 2017-05-18
NPE problems in SSLEngineFactorySupport

pveentjer:
Incorrect dealing of null keystore/truststore.


pveentjer:
merging this one since it enterprise is failing because of it.

PR #10617, opened 2017-05-18
Fixed additional key serializations in Near Cache

Donnerbart:
As already mentioned in the TDD and noticed by Ahmet in one of the other PRs: we still have some unwanted key serializations in the Near Cache proxies.

The reason is that we need the partition ID to fetch the `MetaDataContainer` deep down in the `NearCacheRecordStore` implementation. For this we need the serialized key. And also the normal proxy will serialize the key. Since we removed the centralized key serialization, the key is now serialized twice in both code branches.

---

This PR is a possible solution to this issue. The main idea is to serialize the key after it could not be found in the Near Cache (because the normal proxy will serialize anyway). Then we pass both keys to the `NearCacheRecordStore` (two keys, since the Near Cache key can be serialized or not, depending on the configuration). The main struggle is that one path which creates the `NearCacheRecord` (where the partition ID is needed) is "guarded" by a fixed `IFunction`, which cannot pass an additional parameter.

```java
protected final IFunction<K, R> reserveForUpdate = new ReserveForUpdateFunction();

    @Override
    protected R getOrCreateToReserve(K key) {
        // it's easy to get the serialized key to this method, but we cannot get it
        // through this method, unless we create a new instance per call --> litter!
        return records.applyIfAbsent(key, reserveForUpdate);
    }
```

Of course I could we could create a new `ReserveForUpdateFunction` on each call, but that would create litter. So I moved the `MetaDataContainer` part to where the record is updated (value is set, reservation flag is removed etc.). Before the reservation flag is not removed, the record is not accessible. At least the data from the `MetaDataContainer` is not used before. So this should be safe, but needs a critical code review.

Another solution would be to jump on the changes Ahmet is preparing. I saw that he changed that final `ReserveForUpdateFunction` to a new instance per call. If this is not avoidable, we should be able to add the serialized key as parameter there. Then we would not have to change so much in the `NearCacheRecordStore` API.

---

We can see in the serialization count tests, that the approach works and doesn't seem to break any other Near Cache test. The EE adaption also worked fine in local test runs.


hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -137,9 +137,10 @@ protected V getSyncInternal(Object key, ExpiryPolicy expiryPolicy) {
 
         try {
             long reservationId = nearCache.tryReserveForUpdate(key);
-            value = super.getSyncInternal(key, expiryPolicy);
+            Data keyData = toData(key);
```
Donnerbart: The key would anyway be serialized in `super.getSyncInternal()` and we also need the serialized key to create the `NearCacheRecord`. So we add the serialization here to have it in a single place again. The serialized key is then passed as additional argument to `tryPublishReserved()` which ends up in the `NearCacheRecordStore`.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/store/AbstractNearCacheRecordStore.java
```diff
@@ -314,7 +314,8 @@ public void put(K key, V value) {
         R oldRecord = null;
         try {
             record = valueToRecord(value);
-            onRecordCreate(key, record);
+            onRecordCreate(record);
+            initInvalidationMetaData(record, key, keyData);
```
Donnerbart: This is a second code path where the invalidation metadata is set. The change is uncritical, since the metadata is still set on the record creation.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/store/AbstractNearCacheRecordStore.java
```diff
@@ -411,16 +412,18 @@ public long tryReserveForUpdate(K key) {
...
     }
 
-    protected R updateReservedRecordInternal(K key, V value, R reservedRecord, long reservationId) {
+    protected R updateReservedRecordInternal(K key, Data keyData, V value, R reservedRecord, long reservationId) {
         if (!reservedRecord.casRecordState(reservationId, UPDATE_STARTED)) {
             return reservedRecord;
         }
 
         updateRecordValue(reservedRecord, value);
+        initInvalidationMetaData(reservedRecord, key, keyData);
```
Donnerbart: This is the more critical change, since the invalidation metadata is now changed on the record update, not on the record creation. But as you can see the record is not accessible (aka readable) before the `reservedRecord.casRecordState(UPDATE_STARTED, READ_PERMITTED);` is called, so it should not matter that the meta data is not set directly on the creation.

I also checked that the metadata is not used before the record is in READ_PERMITTED state:
```java
    @Override
    public V get(K key) {
        checkAvailable();

        R record = null;
        V value = null;
        try {
            record = getRecord(key);
            if (record != null) {
                if (record.getRecordState() != READ_PERMITTED) {
                    return null;
                }
                if (staleReadDetector.isStaleRead(key, record)) {
                    remove(key);
                    return null;
                }
(...)
```
So it should be okay to have the metadata set in the `updateReservedRecordInternal()` method. I also think it doesn't make a difference that the metadata itself (UUID and invaldation sequence) is now taken from the `MetaDataContainer` after the remove call, but this should be confirmed by Ahmet.

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/store/AbstractNearCacheRecordStore.java
ahmetmircik: `initInvalidationMetaData` should be done at reservation phase, doing it in publishing phase can break eventual consistency.

ahmetmircik:
@Donnerbart what about making `ReserveForUpdateFunction` thread local and initialize it for each call via keyData or partitionID whatever needed?

Donnerbart:
That could indeed be a nice solution!

Donnerbart:
run-lab-run

PR #10623, opened 2017-05-19
Added getRecord(K key) method to NearCacheRecordStore

Donnerbart:
This enabled to test `NearCacheRecord` on OS and EE without casting
madness.


Donnerbart:
run-lab-run

Issue #10312, opened 2017-04-11
Client intermediate HeapData litter


pveentjer:
When the client `map.get` returns a value (and I'm sure it will be the same thing for many other calls), the client makes an `HeapData` from the content in the client-buffer. And then this heap data is deserialized.

This intermediate HeapData object shoudn't be needed; the serialization service should make use of the `clientmessage.buffer`.

This skips one potentially large object being allocated

Perhaps we should consider introducing a new `HeapData` with an offset.

pveentjer:
This is an optimization that doesn't require an API change or wire level change. So I'm going to move this to 4.1

sancar:
@pveentjer This is fixed in 4.0. HeapData uses the payload available in ClientMessage without copying it. 
While HeapData was a copy of the data in 3.12, it is just a small wrapper in 4.0. 

Further removing HeapData does not seem possible. It means returning byte[] from the protocol instead of Data on decode. But most of the internal API(Operations, DataWriter/Reader classes, Services ) expects Data.  

Since the `large` litter object is removed(it is not large anymore), I would consider this issue is fixed in 4.0.

PR #10629, opened 2017-05-22
HeapData reuses byte[] of ClientMessage to avoid copy

sancar:
Follow up to https://github.com/hazelcast/hazelcast/pull/10584

When internal array of ClientMessage is used inside Data, Data consumes large memory. 
When ClientMessage read from server, it does not reuse internal Data. In server, less memory consumption is chosen, instead of avoiding copy. 

When Client Message read from client, all data's read from ClientMessage reuses internal byte array of ClientMessage. In the places where it needs be stored, Data is copied. `compact` method is introduced on Data for this.

Related Protocol side changes https://github.com/hazelcast/hazelcast-client-protocol/pull/79/files

fixes https://github.com/hazelcast/hazelcast/issues/10312

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/MessageFlyweight.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/MessageFlyweight.java
@@ -20,9 +20,7 @@ import com.hazelcast.internal.serialization.impl.HeapData;
 import com.hazelcast.nio.Bits;
 import com.hazelcast.nio.serialization.Data;
 
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
 import java.util.Map;
 
 /**
@@ -178,15 +176,14 @@ public class MessageFlyweight {
         return new HeapData(getByteArray());
     }
 
-    public List<Data> getDataList() {
+    public Data getDataReuse() {
         final int length = buffer.getInt(index + offset);
         index += Bits.INT_SIZE_IN_BYTES;
-        final List<Data> result = new ArrayList<Data>();
-        for (int i = 0; i < length; i++) {
-            result.add(getData());
-        }
-        return result;
+        Data data = new HeapData(buffer.byteArray(), index + offset, length);
+        index += length;
+        return data;
     }
+
     //endregion GET Overloads
 
     protected int int32Get(int index) {
--- a/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
@@ -37,51 +37,55 @@ public class HeapData implements Data {
 
     public static final int HEAP_DATA_OVERHEAD = DATA_OFFSET;
 
-    // array (12: array header, 4: length)
-    private static final int ARRAY_HEADER_SIZE_IN_BYTES = 16;
+    // array jvm (12: array header, 4: length) + fields (4: offset , 4: length)
+    private static final int ARRAY_HEADER_SIZE_IN_BYTES = 24;
 
+    protected final int offset;
     protected byte[] payload;
+    protected int length;
 
     public HeapData() {
+        this.offset = 0;
     }
 
     public HeapData(byte[] payload) {
-        if (payload != null && payload.length > 0 && payload.length < HEAP_DATA_OVERHEAD) {
+        this(payload, 0, payload == null ? 0 : payload.length);
+    }
+
+    public HeapData(byte[] payload, int offset, int length) {
+        this.payload = payload;
+        this.offset = offset;
+        this.length = payload == null ? 0 : length;
+        assert payload == null || payload.length >= length + offset;
+        if (payload != null && payload.length > 0 && length < HEAP_DATA_OVERHEAD) {
             throw new IllegalArgumentException(
                     "Data should be either empty or should contain more than " + HeapData.HEAP_DATA_OVERHEAD + " bytes! -> "
-                            + Arrays.toString(payload));
+                            + Arrays.toString(Arrays.copyOfRange(payload, offset, length)));
         }
-        this.payload = payload;
     }
 
     @Override
     public int dataSize() {
-        return Math.max(totalSize() - HEAP_DATA_OVERHEAD, 0);
+        return Math.max(length - HEAP_DATA_OVERHEAD, 0);
     }
 
     @Override
     public int totalSize() {
-        return payload != null ? payload.length : 0;
+        return length;
     }
 
     @Override
-    public void copyTo(byte[] dest, int destPos) {
-        if (totalSize() > 0) {
-            System.arraycopy(payload, 0, dest, destPos, payload.length);
+    public Data compact() {
+        if (offset == 0) {
+            return this;
         }
+        byte[] bytes = Arrays.copyOfRange(payload, offset, offset + length);
+        return new HeapData(bytes);
     }
 
     @Override
-    public int getPartitionHash() {
-        if (hasPartitionHash()) {
-            return Bits.readIntB(payload, PARTITION_HASH_OFFSET);
-        }
-        return hashCode();
-    }
-
-    @Override
-    public boolean hasPartitionHash() {
-        return payload != null && payload.length >= HEAP_DATA_OVERHEAD && Bits.readIntB(payload, PARTITION_HASH_OFFSET) != 0;
+    public int offset() {
+        return offset;
     }
 
     @Override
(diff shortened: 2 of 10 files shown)


hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/HeapData.java
```diff
@@ -39,17 +39,23 @@
...
+    protected ByteBuffer payload;
+    private final int length;
 
     public HeapData() {
+        length = 0;
     }
 
     public HeapData(byte[] payload) {
-        if (payload != null && payload.length > 0 && payload.length < HEAP_DATA_OVERHEAD) {
+        this(ByteBuffer.wrap(payload));
```
mdogan: Creating an additional `ByteBuffer` for each data object. I'm not sure about cost vs benefit.

sancar: I will try to make this only for Data's when read from ClientMessage. Plan is to not the introduce ByteBuffer for the rest.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/BaseHeapData.java
```diff
@@ -0,0 +1,184 @@
+/*
```
pveentjer: Keep in mind that by introducing a new class of heap data, you could make it much more difficult for the JIT to do optimizations.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteBufferHeapData.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.hazelcast.nio.serialization.Data;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+/**
+ * A {@link Data} implementation where the content lives on the heap, and part of a byteBuffer
+ */
+public final class ByteBufferHeapData extends com.hazelcast.internal.serialization.impl.BaseHeapData {
```
pveentjer: Why do we need a new heap data for this? Can't the underlying byte-array of the bytebuffer be pushed into the HeapData? 

I would like to get rid of Data classes completely.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/BaseHeapData.java
```diff
@@ -0,0 +1,184 @@
...
+import com.hazelcast.util.HashUtil;
+
+import java.util.Arrays;
+
+import static com.hazelcast.util.JVMUtil.REFERENCE_COST_IN_BYTES;
+
+/**
+ * A {@link Data} implementation where the content lives on the heap.
+ */
+public abstract class BaseHeapData implements Data {
```
pveentjer: Why not call this AbstractHeapData? Abstract is the recommended prefix for Abstract classes.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteBufferHeapData.java
```diff
@@ -0,0 +1,65 @@
...
+        return payload.arrayOffset();
+    }
+
+    @Override
+    protected int readIntB(int offset) {
+        return payload.getInt(offset);
+    }
+
+    @Override
+    public byte[] toByteArray() {
```
pveentjer: This will introduce litter. 

Why does the array need to be copied?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteBufferHeapData.java
```diff
@@ -0,0 +1,65 @@
...
+ * A {@link Data} implementation where the content lives on the heap, and part of a byteBuffer
+ */
+public final class ByteBufferHeapData extends com.hazelcast.internal.serialization.impl.BaseHeapData {
+
+    private final ByteBuffer payload;
+    private final int length;
+
+    public ByteBufferHeapData(byte[] bytes, int offset, int length) {
+        this.length = length;
+        this.payload = ByteBuffer.wrap(bytes, offset, length).slice();
```
pveentjer: More litter because you are creating at least one new ByteBuffer.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/MessageFlyweight.java
```diff
@@ -175,7 +175,11 @@ public String getStringUtf8() {
     }
 
     public Data getData() {
-        return new HeapData(getByteArray());
+        final int length = buffer.getInt(index + offset);
+        index += Bits.INT_SIZE_IN_BYTES;
+        Data data = new ByteBufferHeapData(buffer.byteArray(), index + offset, length);
```
mdogan: Is this `ByteBufferHeapData` going to be stored in Hazelcast data structures, IMap, ICache etc?

pveentjer:
I see one problem exchanged for new problems.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/MessageFlyweight.java
pveentjer: If that would be the case, it would potentially be very bad because of increased memory usage.

mdogan: If yes, then there's another trade-off when `ByteBufferHeapData`  is stored in map/cache.`ByteBufferHeapData` will keep a larger `byte[]` than expected and consume more memory.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteBufferHeapData.java
sancar: This method is not used in hot path. And in the places it is used, it is forcing me to copy the byte array. I can't return the internal byte array directly here, because it is byte array that contains the expected byte[] not itself. Only way to avoid copy here is to return ByteBuffer here(something with offset)

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteBufferHeapData.java
sancar: Actually I am exchanging creating a large buffer with reusing it with offset. 
ByteBuffer reuses internal byte array. 
After reading your comment here https://github.com/hazelcast/hazelcast/pull/10629#pullrequestreview-40722488
How about making our HeapData work with offset instead of using ByteBuffer ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/util/MessageFlyweight.java
sancar: Ok. I did not see this coming. I am avoiding copy but it causes more memory usage where these are stored. 
So this improvement will be beneficial everywhere Data comes as response to client. When client message going to remote, it is not beneficial since it needs to be stored on cluster. 
What I can do is to keep the improvement but actually identify the places where Data needs to be stored. And copy to a smaller buffer only needed. 
For the data coming to client, only place seems like the places we use caching. 
How does that sound ?

sancar:
@pveentjer Well, lets work on the pr, until new problems are also cleaned up.

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/BaseHeapData.java
sancar: I will probably get back to single HeapData. see https://github.com/hazelcast/hazelcast/pull/10629#discussion_r118892903

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/ByteBufferHeapData.java
sancar: I will probably get back to single HeapData. see https://github.com/hazelcast/hazelcast/pull/10629#discussion_r118892903

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/BaseHeapData.java
sancar: I will probably get back to single HeapData. see https://github.com/hazelcast/hazelcast/pull/10629#discussion_r118892903

sancar:
run-lab-run

sancar:
Parking this change for now.

Issue #10556, opened 2017-05-10
[query-cache] NPE in member when setting the same key value twice.


blur0224:
I originally found this issue because I was unknowingly setting a value twice.  After hours of debugging I was able to determine the cause of the issue.  The error should not be a NPE, it should be something that actually indicates what the problem is. 

> Hazelcast version **3.8.1**
> Cluster size 1-4 members
> Number of the clients. (2 - 6 clients)
> Version of Java. openjdk version "1.8.0_121"
> Operating system. Linux and Windows
 

MapConfiguration:
```java 
QueryCacheConfig revocationQueryCacheConfig = new QueryCacheConfig(revocationTokenCacheName);
revocationQueryCacheConfig.getPredicateConfig().setImplementation((Predicate) entry -> true);

MapConfig revocationMapConfig = new MapConfig(revocationTokenMapName);
revocationMapConfig.addQueryCacheConfig(revocationQueryCacheConfig);
revocationMapConfig.setEvictionPolicy(EvictionPolicy.LRU);
revocationMapConfig.setTimeToLiveSeconds(revocationMapTimeToLive);
revocationMapConfig.setMaxIdleSeconds(revocationMapMaxIdle);

NearCacheConfig nearCacheConfig = new NearCacheConfig();
nearCacheConfig.setTimeToLiveSeconds(revocationCacheTimeToLive);
nearCacheConfig.setMaxIdleSeconds(revocationCacheMaxIdle);
revocationMapConfig.setNearCacheConfig(nearCacheConfig);

return revocationMapConfig;
```

Code to reproduce exception:
```java 
IEnterpriseMap<String, CacheToken> map = (IEnterpriseMap) instance.getMap(revocationMapName);
CacheToken cacheToken = new CacheToken();
cacheToken.setValue("abcd123");

map.set("abcd123", cacheToken);
map.set("abcd123", cacheToken);
```

Results in the exception in the member log:
```java
java.lang.NullPointerException: null
        at com.hazelcast.query.impl.CachedQueryEntry.getTargetObject(CachedQueryEntry.java:100) ~[hazelcast-enterprise-all-3.8.1.jar:3.8.1]
        at com.hazelcast.query.impl.QueryableEntry.extractAttributeValue(QueryableEntry.java:81) ~[hazelcast-enterprise-all-3.8.1.jar:3.8.1]
[...]

PR #10632, opened 2017-05-22
Increased code coverage of CachedQueryEntry

Donnerbart:
We have this issue which triggers the NPE: https://github.com/hazelcast/hazelcast/issues/10556

I cannot get the reproducer running, so I just created test cases for the class which throws the NPE.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/query/impl/CachedQueryEntryTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/query/impl/CachedQueryEntryTest.java
@@ -18,6 +18,8 @@ package com.hazelcast.query.impl;
 
 import com.hazelcast.internal.serialization.InternalSerializationService;
 import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.nio.serialization.Portable;
+import com.hazelcast.query.SampleTestObjects.PortableEmployee;
 import com.hazelcast.query.impl.getters.Extractors;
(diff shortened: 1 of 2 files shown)


hazelcast/src/main/java/com/hazelcast/query/impl/CachedQueryEntry.java
```diff
@@ -102,7 +102,10 @@ protected Object getTargetObject(boolean key) {
             }
         } else {
             if (valueObject == null) {
-                if (valueData.isPortable()) {
+                // valueObject and valueData are null when the default constructor is used and init() is not yet called
+                if (valueData == null) {
+                    return null;
+                } else if (valueData.isPortable()) {
```
Donnerbart: This is the line which produces a NPE if you run `getTargetObject()` on an `CachedQueryEntry` which was not initialized.

ahmetmircik: Seems returning null is hiding the underlying issue. 

Passing `null` as an `oldValue` to the event after `map#set` can cause this NPE. Also it seems expected that one of valueData or valueObject should be set to a non-null value (at least for `getTargetObject` method).

Donnerbart: That's what I already thought. I changed the PR to just increase the code coverage of the class.

Issue #10637, opened 2017-05-23
No transaction rollback and data loss when master node is terminated


fsgonz:
I am using hazelcast 3.8.1. 
java version "1.8.0_131"

In case the master node is terminated in the middle of a two_phase transaction, the rollback for that transacion is not made resulting in data loss.
What I see is that the transaction rollback is attempted but a race condition arises as the new partition table has not been fetched in the new master node. This is resulting in a WrongTargetException as the node has already been removed.

To reproduce the issue consider this two nodes. One implements this logic:

            final Config hazelcastConfig = new Config().addQueueConfig(new QueueConfig("queue").setStatisticsEnabled(true).setBackupCount(1));
            final HazelcastInstance instance1 = Hazelcast.newHazelcastInstance(hazelcastConfig);
            
            
            IQueue<Integer> queue = instance1.getQueue("queue");
            
            queue.offer(1);

            try
            {
                Thread.sleep(10000);
            }
            catch (InterruptedException e)
            {
                // TODO Auto-generated catch block
                e.printStackTrace();
            }
            TransactionOptions options = new TransactionOptions()
                    .setTransactionType( TransactionType.TWO_PHASE );

            TransactionContext context = instance1.newTransactionContext( options );
            context.beginTransaction();
            TransactionalQueue queue1 = context.getQueue("queue"); 
            queue1.poll();
            queue1.offer(2);
            queue1.offer(3);
            
            while(true);

This will be the master node. A second node which should be initiated and added to the cluser before the transaction is created is the following:

        final Config hazelcastConfig = new Config().addQueueConfig(new QueueConfig("queue").setStatisticsEnabled(true).setBackupCount(1));
        final HazelcastInstance instance1 = Hazelcast.newHazelcastInstance(hazelcastConfig);

        TransactionOptions options = new TransactionOptions()
                .setTransactionType( TransactionType.TWO_PHASE );

        TransactionContext context = instance1.newTransactionContext( options );
        context.beginTransaction();
        TransactionalQueue queue1 = context.getQueue("queue"); 

        while (true)
        {
            System.out.println("queue size: " + queue1.size());
        }

If I kill the master node while the transaction is opened (during the infinite while), when the open transaction of the killed node are finalized in TransactionManagerServiceImpl.finalizeTransactionsOf a wrongtargetException is raised (when the QueueTransactionRollbackOperation is performed)

What results in the exception is the following. After killing the master node:

[...]

PR #10641, opened 2017-05-24
TransactionalServices should be able to handle rollback failures

mdogan:
During transaction rollback execution, operation may fail with wrong target exception
due to stale partition table or because of partition migration.
TransactionalServices should be able to handle that by invoking rollback on partition
instead of executing rollback operation locally.

Fixes https://github.com/hazelcast/hazelcast/issues/10637


mdogan:
run-lab-run

mdogan:
run-lab-run

PR #10638, opened 2017-05-23
Support for sampling serialized objects during test suite execution

vbekiaris:
Adds a maven profile and required test code that samples serialized objects during tests execution. When test suite is finished, the captured samples are output in a binary file accompanied by an index, suitable for testing deserialization of captured objects.

PR #10643, opened 2017-05-24
Introduces compatibility samples blob generation profile

vbekiaris:
Partial backport of #10638 to capture and persist samples of serialized objects during test suite execution in 3.8.x series

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/test/compatibility/SamplingSerializationService.java
+++ b/hazelcast/src/test/java/com/hazelcast/test/compatibility/SamplingSerializationService.java
@@ -1,0 +1,255 @@
+/*
+ * Copyright (c) 2008-2017, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hazelcast.test.compatibility;
+
+import com.hazelcast.core.ManagedContext;
+import com.hazelcast.core.PartitioningStrategy;
+import com.hazelcast.internal.serialization.InternalSerializationService;
+import com.hazelcast.internal.serialization.PortableContext;
+import com.hazelcast.nio.BufferObjectDataInput;
+import com.hazelcast.nio.BufferObjectDataOutput;
+import com.hazelcast.nio.ObjectDataInput;
+import com.hazelcast.nio.ObjectDataOutput;
+import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.nio.serialization.PortableReader;
+import com.hazelcast.test.TestEnvironment;
+
+import java.io.IOException;
+import java.nio.ByteOrder;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import static java.util.Collections.newSetFromMap;
+
+/**
+ * Serialization service that intercepts and samples serialized objects.
+ * Employed to gather samples of serialized objects used in member-to-member communication during test suite
+ * execution with mock network.
+ *
+ * @see TestEnvironment#SAMPLE_SERIALIZED_OBJECTS
+ */
+public class SamplingSerializationService implements InternalSerializationService {
+
+    static final ConcurrentMap<String, List<byte[]>> SERIALIZED_SAMPLES_PER_CLASS_NAME =
+            new ConcurrentHashMap<String, List<byte[]>>(1000);
+    // cache classes for which samples have already been captured
+    static final Set<String> SAMPLED_CLASSES = newSetFromMap(new ConcurrentHashMap<String, Boolean>(1000));
+    private static final int MAX_SERIALIZED_SAMPLES_PER_CLASS = 5;
+    // utility strings to locate test classes commonly used as user objects
+    private static final String DUMMY_CLASS_INFIX = "Dummy";
+    private static final String TEST_CLASS_INFIX = "Test";
+    private static final String TEST_PACKAGE_INFIX = ".test";
+    private static final String SAMPLE_OBJECTS_TEST_CLASS_NAME = "com.hazelcast.query.SampleObjects";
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -253,6 +253,7 @@
                         -Dhazelcast.phone.home.enabled=false
                         -Dhazelcast.mancenter.enabled=false
                         -Dhazelcast.test.use.network=false
+                        -Dhazelcast.test.record.serialized.class.names=/Users/bekiarisv/tmp/class-recorder/
```
Donnerbart: This looks wrong for a final commit :)

hazelcast/src/test/java/com/hazelcast/test/compatibility/SamplingSerializationService.java
```diff
@@ -0,0 +1,255 @@
...
+    static final ConcurrentMap<String, List<byte[]>> SERIALIZED_SAMPLES_PER_CLASS_NAME =
+            new ConcurrentHashMap<String, List<byte[]>>(1000);
+    // cache classes for which samples have already been captured
+    static final Set<String> SAMPLED_CLASSES = newSetFromMap(new ConcurrentHashMap<String, Boolean>(1000));
+    private static final int MAX_SERIALIZED_SAMPLES_PER_CLASS = 5;
+    // utility strings to locate test classes commonly used as user objects
+    private static final String DUMMY_CLASS_PREFIX = "Dummy";
+    private static final String TEST_CLASS_SUFFIX = "Test";
+    private static final String TEST_PACKAGE_INFIX = ".test";
+    private static final String SAMPLE_OBJECTS_TEST_CLASS_NAME = "com.hazelcast.query.SampleObjects";
```
Donnerbart: I guess it will lead to no error when those classes have different names between versions? They will just not be compared? This is actually a good test how this test setup will deal with renamed classes.

vbekiaris: For `IdentifiedDataSerializable`s (which make up most of our network communication classes), renaming should not affect serialization eg `FinalizeJoinOperation` is renamed in 3.9 to `FinalizeJoinOp` and is properly deserialized. The class name persisted in the index file is only used for the failure message (so actually reflects previous version's class names).
For `DataSerializable` & plain `Serializable` classes renaming will break deserialization so the static binary compatibility test should catch that.

hazelcast/src/test/java/com/hazelcast/test/compatibility/SamplingSerializationService.java
```diff
@@ -0,0 +1,255 @@
...
+    public void dispose() {
+        delegate.dispose();
+    }
+
+    // record the given object, then return it
+    private static <T> T sampleObject(T obj, byte[] serializedObject) {
+        if (obj == null)
+            return null;
+
+        if (serializedObject != null && shouldAddSerializedSample(obj)) {
```
sancar: I class name is checked here if should it be serialized or not. If 2 class names comes consecutively, second one will return false here, because of following check
```
if (SAMPLED_CLASSES.contains(className)) {
            return false;
      }
```

So how there can be a second object in the SERIALIZED_SAMPLES_PER_CLASS_NAME list corresponding to same class name ? As far as I understand only `MAX_SERIALIZED_SAMPLES_PER_CLASS` is meant to check. And the check from `SAMPLED_CLASS` should be removed, right ?

hazelcast/src/test/java/com/hazelcast/test/compatibility/SamplingSerializationService.java
```diff
@@ -0,0 +1,255 @@
...
+public class SamplingSerializationService implements InternalSerializationService {
+
+    static final ConcurrentMap<String, List<byte[]>> SERIALIZED_SAMPLES_PER_CLASS_NAME =
+            new ConcurrentHashMap<String, List<byte[]>>(1000);
+    // cache classes for which samples have already been captured
+    static final Set<String> SAMPLED_CLASSES = newSetFromMap(new ConcurrentHashMap<String, Boolean>(1000));
+    private static final int MAX_SERIALIZED_SAMPLES_PER_CLASS = 5;
+    // utility strings to locate test classes commonly used as user objects
+    private static final String DUMMY_CLASS_PREFIX = "Dummy";
+    private static final String TEST_CLASS_SUFFIX = "Test";
```
sancar: These two does not look really like `SUFFIX` and `PREFIX, since you are checking them with `contains`.
Should we rename these to `INFIX` or is the check wrong ?

vbekiaris: Yes, renaming to `INFIX`

hazelcast/src/test/java/com/hazelcast/test/compatibility/SamplingSerializationService.java
vbekiaris: The class name is only added to `SAMPLES_CLASSES` after `MAX_SERIALIZED_SAMPLES_PER_CLASS` have been already collected in [line 243 below](https://github.com/hazelcast/hazelcast/pull/10643/files#diff-a39b4ca181361e82a7963f36dfcbc196R243); as long as there are fewer than `MAX_SERIALIZED_SAMPLES_PER_CLASS` `shouldAddSerializedSample(obj)` will return true before adding the class name to  `SAMPLED_CLASSES`.

sancar: I still think there is a problem. 
If I call `sampleObject` method twice. Is not second call will skip
 `addSerializedSample(obj, serializedObject);` (line 203),
 because `SAMPLED_CLASSES.contains(className)` check will hold true in line 224 ?
 And `addSerializedSample`  is the only place we add bytes to `SERIALIZED_SAMPLES_PER_CLASS_NAME`

vbekiaris: > If I call sampleObject method twice. Is not second call will skip
addSerializedSample(obj, serializedObject); (line 203),
because SAMPLED_CLASSES.contains(className) check will hold true in line 224 ?

No, unless `MAX_SERIALIZED_SAMPLES_PER_CLASS` have already been collected. In the first invocation to `shouldAddSerializedSample` the flow will be as follows:
- L219 `if (klass.isPrimitive())`, continue execution
- L224 `if (SAMPLED_CLASSES.contains(className))` `false`, continue execution
- ...
- L237 `if (existingSamples == null)` is `true` --> returns `true` and sample if picked up
Notice, the class name was not added to  `SAMPLED_CLASSES`.

Second invocation, is same as above but:
- L237 `if (existingSamples == null)` is `false` --> continue
- L240 `if (existingSamples.size() < MAX_SERIALIZED_SAMPLES_PER_CLASS)` is `true` --> return true and sample is collected, class name still not added to `SAMPLED_CLASSES`.

Only when the number of collected classes is at least `MAX_SERIALIZED_SAMPLES_PER_CLASS` does L240 continue further to execute
```
SAMPLED_CLASSES.add(className);
return false;
```

sancar: Now, I  got it. Sorry for taking your time.

PR #10649, opened 2017-05-25
Add PromoteMemberRequest for Management Center

emre-aydin:
* A new Request is introduced for promoting a lite member to a data
member.
* We also send if the member is a lite member as part of member's
state to Management Center. This is required to know which members are
fit for promotion.


emre-aydin:
run-lab-run

hazelcast/src/main/java/com/hazelcast/monitor/TimedMemberState.java
```diff
@@ -155,6 +169,9 @@ public void fromJson(JsonObject json) {
         memberState = new MemberStateImpl();
         memberState.fromJson(jsonMemberState);
         sslEnabled = getBoolean(json, "sslEnabled", false);
+        if (memberState.getNodeState().getClusterVersion().isGreaterOrEqual(Versions.V3_9)) {
```
mesutcelik: why do we add this? we are already in 3.9 codebase?

emre-aydin: Members send messages to each other based on the cluster version, even if their codebase version is newer. This way, they can continue working together during a rolling upgrade. I think it should be the same with the Management Center related data that they send. Vassilis said something similar in the independent mancenter release plugin PRD.

However, on second though, I'm not sure if it's really required in the ManCenter context though. We will already require the users to upgrade their ManCenter version before a rolling upgrade. This will be required only when they don't upgrade ManCenter before starting out the rolling upgrade process.

@vbekiaris What do you think? Do you think there's any good reason to do this (conditional serialization based on cluster version), keeping in mind that we recommend the user to upgrade ManCenter version before starting out with rolling upgrade process?

emre-aydin: @mesutcelik Thinking about this again, I realized that a member with codebase version 3.9 but cluster version 3.8 cannot use 3.9 features, so it makes sense to only send this piece of data only if the cluster version is 3.9.

hazelcast/src/main/java/com/hazelcast/internal/management/request/PromoteMemberRequest.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package com.hazelcast.internal.management.request;
+
+import com.eclipsesource.json.JsonObject;
+import com.hazelcast.internal.management.ManagementCenterService;
+
+import java.io.IOException;
+
+public class PromoteMemberRequest implements ConsoleRequest {
```
mdogan: I would consider making an `AsyncConsoleRequest`.  `promoteLocalLiteMember()` is a blocking call, it can block `TaskPollThread` for a long time.

emre-aydin: @mdogan, makes sense, will update the PR accordingly. Thanks for the feedback.

emre-aydin:
verify

emre-aydin:
verify

PR #10654, opened 2017-05-27
Log warning when max-size PER_NODE config does not allow any data in map

### LINKED ISSUE #10179: Eviction and max-size less then 271 cause map to never populate

vbekiaris:
When using `PER_NODE` max size policy, max size must be at least `partitionCount / memberCount` in order for any data to be stored in the map. If this condition does not hold, logs a warning once per misconfigured map.

Fixes #10179

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictionChecker.java
+++ b/hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictionChecker.java
@@ -35,6 +35,7 @@ import com.hazelcast.util.MemoryInfoAccessor;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.hazelcast.cluster.memberselector.MemberSelectors.DATA_MEMBER_SELECTOR;
 import static com.hazelcast.memory.MemorySize.toPrettyString;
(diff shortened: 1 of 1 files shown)


vbekiaris:
run-lab-run

vbekiaris:
run-lab-run

hazelcast/src/main/java/com/hazelcast/map/impl/eviction/EvictionChecker.java
```diff
@@ -119,8 +120,15 @@ public double calculatePerNodeMaxRecordStoreSize(RecordStore recordStore) {
         int memberCount = nodeEngine.getClusterService().getSize(DATA_MEMBER_SELECTOR);
         int partitionCount = nodeEngine.getPartitionService().getPartitionCount();
 
-        return (1D * configuredMaxSize * memberCount / partitionCount);
-
+        final double perNodeMaxRecordStoreSize = (1D * configuredMaxSize * memberCount / partitionCount);
+        if (perNodeMaxRecordStoreSize < 1 && !misconfiguredPerNodeMaxSizeWarningLogged) {
```
ahmetmircik: Can an `AtomicBoolean` fit better to prevent possible multiple similar logs?

`if (perNodeMaxRecordStoreSize < 1 && misconfiguredPerNodeMaxSizeWarningLogged.compareAndSet(false,true))`

vbekiaris: Thanks, done.

vbekiaris:
Backported for `3.8.3`: #10821

PR #10662, opened 2017-05-29
Improved AwsCloudDiscoveryTest

Donnerbart:
This test requires the environment variables `AWS_ACCESS_KEY_ID` and
`AWS_SECRET_ACCESS_KEY` to be set. This makes it hard to execute the test
locally, since it will fail with a misleading error message.

With this change we check the values upfront by failing if they are not
set on Jenkins and just ignoring the test in other runs. This will make
it easier to get the testsuite running.

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/discovery/AwsCloudDiscoveryTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/discovery/AwsCloudDiscoveryTest.java
@@ -27,13 +27,22 @@ import com.hazelcast.core.IMap;
 import com.hazelcast.test.HazelcastParallelClassRunner;
 import com.hazelcast.test.annotation.ParallelTest;
 import com.hazelcast.test.annotation.SlowTest;
-import org.junit.Assert;
+import org.hamcrest.Matchers;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/discovery/AwsCloudDiscoveryTest.java
```diff
@@ -44,15 +53,26 @@
...
-        props.put(AwsProperties.TAG_VALUE.getDefinition().key(), "aws-tag-value-1");
-        props.put(AwsProperties.CONNECTION_TIMEOUT_SECONDS.getDefinition().key(), "10");
+        props.put(PORT.getDefinition().key(), "60000");
+        props.put(ACCESS_KEY, System.getenv("AWS_ACCESS_KEY_ID"));
+        props.put(SECRET_KEY, System.getenv("AWS_SECRET_ACCESS_KEY"));
+        props.put(TAG_KEY.getDefinition().key(), "aws-test-tag");
+        props.put(TAG_VALUE.getDefinition().key(), "aws-tag-value-1");
+        props.put(CONNECTION_TIMEOUT_SECONDS.getDefinition().key(), "10");
+
+        if (isOnJenkins()) {
```
Donnerbart: Here we distinct to either fail or ignore the test if the environment variables are not set, depending if we are running on Jenkins or not.

Donnerbart:
run-lab-run

PR #10667, opened 2017-05-29
Added whitelist to ThreadLeakTest

Donnerbart:
The `JitterThread` is a test thread which is not joinable once started.

### DIFF
--- a/hazelcast/src/test/java/classloading/ThreadLeakTest.java
+++ b/hazelcast/src/test/java/classloading/ThreadLeakTest.java
@@ -21,40 +21,58 @@ import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.test.HazelcastSerialClassRunner;
 import com.hazelcast.test.HazelcastTestSupport;
 import com.hazelcast.test.annotation.QuickTest;
+import com.hazelcast.test.jitter.JitterThread;
+import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


Donnerbart:
run-lab-run

Donnerbart:
We just had a failure of this test in the PR builder :) I had a look at the output:
```java
There are still Hazelcast threads running after shutdown!
  -> testThreadLeak (id: 72370) (group: main) (daemon: true) (alive: true) (interrupted: false) (state: RUNNABLE)
```
and it looks like it's waiting for it's own thread to join... So I removed the current thread from the diff.

hazelcast/src/test/java/classloading/ThreadLeakTest.java
```diff
@@ -21,51 +21,88 @@
...
-        Thread[] threads = new Thread[diff.size()];
-        diff.toArray(threads);
+        assertJoinable(joinableThreads);
+    }
 
-        assertJoinable(threads);
+    private static Thread[] getJoinableThreads(Set<Thread> oldThreads, Set<Thread> newThreads) {
+        Set<Thread> joinable = new HashSet<Thread>(newThreads);
+        joinable.removeAll(oldThreads);
+        joinable.remove(Thread.currentThread());
```
Donnerbart: I think this may be the true root cause of the timeout failures.

PR #10668, opened 2017-05-29
Cleanup of JitterRule

Donnerbart:
* Made some classes of the `jitter` package more private.
* Ensured that the `JitterThread` is always created as daemon thread.
* Set a name for the `JitterThread`.


Donnerbart:
run-lab-run

Issue #10593, opened 2017-05-16
com.hazelcast.client.listeners.ReplicatedMapEntryListenerOnReconnectTest


Donnerbart:
**com.hazelcast.client.listeners.ReplicatedMapEntryListenerOnReconnectTest.testListenersNonSmartRoutingMultipleServer**
```java
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest$8.run(AbstractListenersOnReconnectTest.java:488)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1037)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:1054)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.validateRegistrations(AbstractListenersOnReconnectTest.java:482)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.validateRegistrationsAndListenerFunctionality(AbstractListenersOnReconnectTest.java:472)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersTerminateOwnerNode(AbstractListenersOnReconnectTest.java:458)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersNonSmartRoutingMultipleServer(AbstractListenersOnReconnectTest.java:409)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
	at java.lang.reflect.Method.invoke(Method.java:597)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

Issue #10359, opened 2017-04-17
ListItemListenerOnReconnectTest.testListenersNonSmartRoutingTerminateRandomNode


hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 300000 milliseconds
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1037)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1328)
	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:277)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:877)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:872)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:864)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest$9.run(AbstractListenersOnReconnectTest.java:522)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:972)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:989)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.validateListenerFunctionality(AbstractListenersOnReconnectTest.java:510)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.validateRegistrationsAndListenerFunctionality(AbstractListenersOnReconnectTest.java:473)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersTerminateRandomNode(AbstractListenersOnReconnectTest.java:113)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersNonSmartRoutingTerminateRandomNode(AbstractListenersOnReconnectTest.java:77)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

Issue #10301, opened 2017-04-11
QueueItemListenerOnReconnectTest test failures


mmedenjak:
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-IbmJDK1.7/1242/

com.hazelcast.client.listeners.QueueItemListenerOnReconnectTest.testClusterReconnectDueToHeartbeatSmartRouting

```
java.lang.AssertionError: CountDownLatch failed to complete within 120 seconds, count left: 1
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:884)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:877)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:869)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersHeartbeatTimeoutToOwner(AbstractListenersOnReconnectTest.java:386)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testClusterReconnectDueToHeartbeatSmartRouting(AbstractListenersOnReconnectTest.java:328)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:95)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:56)
	at java.lang.reflect.Method.invoke(Method.java:620)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:105)
	at com.hazelcast.test.FailOnTimeoutStatement$CallableStatement.call(FailOnTimeoutStatement.java:97)
[...]

Issue #10235, opened 2017-04-06
*OnReconnectTest.testListenersNonSmartRoutingTerminateRandomNode


hasancelik:
```
org.junit.runners.model.TestTimedOutException: test timed out after 300000 milliseconds
	at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:226)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1037)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1326)
	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:281)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:877)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:872)
	at com.hazelcast.test.HazelcastTestSupport.assertOpenEventually(HazelcastTestSupport.java:864)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest$9.run(AbstractListenersOnReconnectTest.java:523)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:972)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:989)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.validateListenerFunctionality(AbstractListenersOnReconnectTest.java:510)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.validateRegistrationsAndListenerFunctionality(AbstractListenersOnReconnectTest.java:473)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersTerminateRandomNode(AbstractListenersOnReconnectTest.java:113)
	at com.hazelcast.client.listeners.AbstractListenersOnReconnectTest.testListenersNonSmartRoutingTerminateRandomNode(AbstractListenersOnReconnectTest.java:77)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
[...]

PR #10670, opened 2017-05-30
Fix for testListenersNonSmartRouting* tests

sancar:
Fix on https://github.com/hazelcast/hazelcast/pull/9865
is revised.

Verification part for events is more strict as before.
Verification for listener registrations was broken for non-smart
clients. It is fixed, so that verification of events does not need
assertTrueEventually at the top.

I could not reproduce the issue locally. Just found a reason to fail and fixing it. Closing the related issues for now. If they fail again, we will take another look.

fixes https://github.com/hazelcast/hazelcast/issues/10301
fixes https://github.com/hazelcast/hazelcast/issues/10235
fixes https://github.com/hazelcast/hazelcast/issues/10157
fixes https://github.com/hazelcast/hazelcast/issues/8744
fixes https://github.com/hazelcast/hazelcast/issues/8254
fixes https://github.com/hazelcast/hazelcast/issues/10593
fixes https://github.com/hazelcast/hazelcast/issues/10359

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientNonSmartListenerService.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientNonSmartListenerService.java
@@ -31,16 +31,19 @@ import com.hazelcast.util.UuidUtil;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 
 public class ClientNonSmartListenerService extends ClientListenerServiceImpl implements ConnectionListener {
 
-    private final Map<ClientRegistrationKey, ClientEventRegistration> registrations
+    private final Map<ClientRegistrationKey, ClientEventRegistration> activeRegistrations
             = new ConcurrentHashMap<ClientRegistrationKey, ClientEventRegistration>();
+    private final Set<ClientRegistrationKey> userRegistrations = new HashSet<ClientRegistrationKey>();
 
     public ClientNonSmartListenerService(HazelcastClientInstanceImpl client,
                                          int eventThreadCount, int eventQueueCapacity) {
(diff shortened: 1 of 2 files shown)


sancar:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientNonSmartListenerService.java
```diff
@@ -184,7 +185,7 @@ public void run() {
                             return Collections.EMPTY_LIST;
                         }
                         LinkedList<ClientEventRegistration> activeRegistrations = new LinkedList<ClientEventRegistration>();
-                        if (getEventHandler(registration.getCallId()) != null) {
+                        if (registration.isActive()) {
```
ihsandemir: Why not check existence using the eventHandlerMap.containsKey instead of introducing a new state variable active?

hazelcast-client/src/test/java/com/hazelcast/client/listeners/AbstractListenersOnReconnectTest.java
```diff
@@ -585,9 +581,6 @@ private ClientConfig getNonSmartClientConfig() {
     void onEvent(String event) {
         events.remove(event);
         eventCount.incrementAndGet();
-        if (events.isEmpty()) {
-            eventsLatch.countDown();
```
ihsandemir: When you removed this, you are only checking how many time the onEvent is triggered. What if it was triggered twice for the same event and none-for an expected event. The test will pass in that case as well. The check was ensuring that the removed events also should match the expected events one-to-one. After this change, are we missing this case?

hazelcast-client/src/main/java/com/hazelcast/client/spi/impl/listener/ClientNonSmartListenerService.java
sancar: event handler is registered before invocation is send. I need a value to check that is updated after invocation is send and successful.

hazelcast-client/src/test/java/com/hazelcast/client/listeners/AbstractListenersOnReconnectTest.java
sancar: I removed the latch because no one setting it to 1 again. In case `outer assertTrueEventually` cause the verify part run again, it was working wrong. 
You are right, I forget about its use and since I already removed `outer assertTrueEventually`  I can safely put it back.

sancar:
run-lab-run

PR #10674, opened 2017-05-31
Add List<Entry<String,byte[]>> to reference objects

mmedenjak:
ClientMessageTemplate.deployClasses uses this parameter type and the
compatibility tests are generated with a compile error. This fix adds
 the reference object with this type.

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ReferenceObjects.java
+++ b/hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ReferenceObjects.java
@@ -94,6 +94,7 @@ import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.security.AccessControlException;
 import java.util.AbstractMap;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.ConcurrentModificationException;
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ReferenceObjects.java
```diff
@@ -229,6 +229,9 @@ private static boolean isEqualStackTrace(StackTraceElement stackTraceElement1, S
     public static Xid anXid = new SerializableXID(1, aString.getBytes(), aString.getBytes());
     public static List<Map.Entry<Data, Data>> aListOfEntry = Collections.<Map.Entry<Data, Data>>singletonList(
             new AbstractMap.SimpleEntry<Data, Data>(aData, aData));
+    public static List<Map.Entry<String, byte[]>> aListOfStringToByteArrEntry
+            = Collections.<Map.Entry<String, byte[]>>singletonList(
+                    new AbstractMap.SimpleEntry<String, byte[]>(aString, new byte[0]));
```
ihsandemir: If you provide a byte array with a non-zero size, it will be more close to real use-case.

hazelcast/src/test/java/com/hazelcast/client/protocol/compatibility/ReferenceObjects.java
```diff
@@ -229,6 +229,9 @@ private static boolean isEqualStackTrace(StackTraceElement stackTraceElement1, S
     public static Xid anXid = new SerializableXID(1, aString.getBytes(), aString.getBytes());
     public static List<Map.Entry<Data, Data>> aListOfEntry = Collections.<Map.Entry<Data, Data>>singletonList(
             new AbstractMap.SimpleEntry<Data, Data>(aData, aData));
+    public static List<Map.Entry<String, byte[]>> aListOfStringToByteArrEntry
+            = Collections.<Map.Entry<String, byte[]>>singletonList(
```
ihsandemir: Why not put a second entry to the collection?

mmedenjak: I saw the same thing above with `aListOfEntry` so I figured this was enough. Is it?

mmedenjak: @ihsandemir added a two-element list of entries and non-empty byte array. Looks ok?

PR #10677, opened 2017-05-31
Fixed HazelcastInstanceFactoryTest which left rogue threads alive

Donnerbart:
The `HazelcastInstanceFactoryTest` contained a test which throws an
exception in `NodeExtension.beforeShutdown()`. This prevented a clean
shutdown of the instance, so basically all threads of that instance
were still running. This caused a failure in `ThreadLeakTest`.

After some discussion we decided that a failure of beforeShutdown() is
not an unsupported scenario (it's just used internally), so we don't
need to fix this with a try-finally-block, but remove the test.

Made `HazelcastInstanceFactoryTest` to check for thread leaks (just for
all tests together, so we can keep the `HazelcastParallelClassRunner`).

### DIFF
--- a/hazelcast/src/test/java/com/hazelcast/instance/HazelcastInstanceFactoryTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/instance/HazelcastInstanceFactoryTest.java
@@ -25,14 +25,18 @@ import com.hazelcast.test.TestHazelcastInstanceFactory;
 import com.hazelcast.test.annotation.QuickTest;
 import org.junit.After;
 import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import static classloading.ThreadLeakTestUtils.assertHazelcastThreadShutdown;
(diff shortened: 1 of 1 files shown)


hazelcast/src/test/java/com/hazelcast/instance/HazelcastInstanceFactoryTest.java
```diff
@@ -177,24 +190,6 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
         }
     }
 
-    @Test(expected = ExpectedRuntimeException.class)
-    public void test_NewInstance_failedAfterStartAndBeforeShutdown() throws Exception {
-        NodeContext context = new TestNodeContext() {
-            @Override
-            public NodeExtension createNodeExtension(Node node) {
-                NodeExtension nodeExtension = super.createNodeExtension(node);
-                doThrow(new ExpectedRuntimeException()).when(nodeExtension).afterStart();
```
Donnerbart: `NodeExtension.afterStart()` is already tested like this in `test_NewInstance_failed_afterNodeStart()`

hazelcast/src/test/java/com/hazelcast/instance/HazelcastInstanceFactoryTest.java
```diff
@@ -177,24 +190,6 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
...
-                doThrow(new ExpectedRuntimeException()).when(nodeExtension).beforeShutdown();
```
Donnerbart: This is the exception which crashes the shutdown process of the HZ instance and leaves most of its threads running.

PR #10684, opened 2017-06-01
Fragment RingbufferService and support predicates and projections

mmedenjak:
Adapt RingbufferService to hold ringbuffers for other services (e.g.
map). This means that it should implement
FragmentedMigrationAwareService and return partition namespace data
for other services when requested.

Also, the read result sets were returning only how many items were
read as well as the matching items. In case the ringbuffer is read
with a filter, not all items match and the result set may not contain
items with a contiguous set of sequences. The itemCount is the
difference between the sequence ID of the last and first returned
item and is a hack not to send all sequence IDs.

We will send the sequence IDs instead. This allows the user to store
the sequence ID and resume processing even if the processing crashes
mid-point in the result set.

The ringbuffer already has supports for IFilter but we also add support
for predicates and projections.

Depends on : 
https://github.com/hazelcast/hazelcast-client-protocol/pull/87

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
+++ b/hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
@@ -46,4 +46,26 @@ public interface ReadResultSet<E> extends Iterable<E> {
      * @throws IllegalArgumentException if index out of bounds.
      */
     E get(int index);
+
+    /**
+     * Return the sequence number for the item at the given index.
+     * The method may throw a {@link UnsupportedOperationException} if there are no
+     * sequences available. This can happen when the cluster version is
+     * {@link com.hazelcast.internal.cluster.Versions#V3_8} or lower.
+     *
+     * @param index the index
+     * @return the sequence number for the ringbuffer item
+     * @throws UnsupportedOperationException if the sequence IDs are not available
+     * @see com.hazelcast.internal.cluster.ClusterService#getClusterVersion()
+     * @since 3.9
+     */
+    long getSequence(int index);
+
+    /**
+     * Return the result set size.
+     *
+     * @return the result set size
+     * @since 3.9
+     */
+    int size();
 }
--- a/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
+++ b/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
@@ -19,13 +19,17 @@ package com.hazelcast.ringbuffer.impl;
 import com.hazelcast.core.HazelcastInstance;
 import com.hazelcast.core.HazelcastInstanceAware;
 import com.hazelcast.core.IFunction;
-import com.hazelcast.instance.HazelcastInstanceImpl;
+import com.hazelcast.internal.cluster.Versions;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.Data;
 import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
+import com.hazelcast.nio.serialization.impl.Versioned;
+import com.hazelcast.projection.Projection;
 import com.hazelcast.ringbuffer.ReadResultSet;
+import com.hazelcast.spi.impl.SerializationServiceSupport;
 import com.hazelcast.spi.serialization.SerializationService;
+import com.hazelcast.util.function.Predicate;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.io.IOException;
@@ -38,34 +42,50 @@ import static com.hazelcast.util.Preconditions.checkTrue;
 
 /**
  * A list for the {@link com.hazelcast.ringbuffer.impl.operations.ReadManyOperation}.
- *
+ * <p>
  * The problem with a regular list is that if you store Data objects, then on the receiving side you get
  * a list with data objects. If you hand this list out to the caller, you have a problem because he sees
  * data objects instead of deserialized objects.
+ * <p>
+ * The predicate, filter and projection may be {@code null} in which case all elements are returned
+ * and no projection is applied.
  *
- * @param <E>
+ * @param <O> deserialized ringbuffer type
+ * @param <E> result set type, is equal to {@code O} if the projection
+ *            is {@code null} or returns the same type as the parameter
  */
-public class ReadResultSetImpl<E> extends AbstractList<E>
-        implements IdentifiedDataSerializable, HazelcastInstanceAware, ReadResultSet<E> {
+public class ReadResultSetImpl<O, E> extends AbstractList<E>
+        implements IdentifiedDataSerializable, HazelcastInstanceAware, ReadResultSet<E>, Versioned {
 
+    protected transient SerializationService serializationService;
     private transient int minSize;
     private transient int maxSize;
-    private transient IFunction<Object, Boolean> filter;
-    private transient HazelcastInstance hz;
+    private transient IFunction<O, Boolean> filter;
+    private transient Predicate<? super O> predicate;
+    private transient Projection<? super O, E> projection;
 
     private Data[] items;
+    private long[] seqs;
     private int size;
     private int readCount;
 
     public ReadResultSetImpl() {
     }
 
-    public ReadResultSetImpl(int minSize, int maxSize, HazelcastInstance hz, IFunction<Object, Boolean> filter) {
+    public ReadResultSetImpl(int minSize, int maxSize, SerializationService serializationService, IFunction<O, Boolean> filter) {
         this.minSize = minSize;
         this.maxSize = maxSize;
         this.items = new Data[maxSize];
-        this.hz = hz;
+        this.seqs = new long[maxSize];
+        this.serializationService = serializationService;
         this.filter = filter;
+    }
+
+    public ReadResultSetImpl(int minSize, int maxSize, SerializationService serializationService,
+                             Predicate<? super O> predicate, Projection<? super O, E> projection) {
+        this(minSize, maxSize, serializationService, null);
+        this.predicate = predicate;
+        this.projection = projection;
     }
 
     public boolean isMaxSizeReached() {
@@ -88,7 +108,11 @@ public class ReadResultSetImpl<E> extends AbstractList<E>
 
     @Override
     public void setHazelcastInstance(HazelcastInstance hz) {
-        this.hz = hz;
+        setSerializationService(((SerializationServiceSupport) hz).getSerializationService());
+    }
+
+    public void setSerializationService(SerializationService serializationService) {
+        this.serializationService = serializationService;
     }
 
     @Override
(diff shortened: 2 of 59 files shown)


hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
```diff
@@ -46,4 +46,19 @@
...
      */
     E get(int index);
+
+    /**
+     * Return the sequence number for the item at the given index.
+     *
+     * @param index the index
+     * @return the sequence number for the ringbuffer item
+     */
+    long getSequence(int index);
```
mdogan: `@since 3.9`

hazelcast/src/main/java/com/hazelcast/ringbuffer/ReadResultSet.java
```diff
@@ -46,4 +46,19 @@
...
+
+    /**
+     * Return the result set size.
+     *
+     * @return the result set size
+     */
+    int size();
```
mdogan: `@since 3.9`

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
```diff
@@ -88,46 +108,65 @@ public int readCount() {
...
 
-    private SerializationService getSerializationService() {
-        HazelcastInstanceImpl impl = (HazelcastInstanceImpl) hz;
-        return impl.getSerializationService();
-    }
-
-    public void addItem(Object item) {
+    @Override
+    public long getSequence(int index) {
+        return seqs != null && seqs.length > index ? seqs[index] : -1;
```
mdogan: When does `seqs` become `null`? Is this an expected path? Can `getSequence()` be called when `clusterVersion < 3.9` ?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
```diff
@@ -38,36 +42,52 @@
...
+public class ReadResultSetImpl<O, E> extends AbstractList<E>
+        implements IdentifiedDataSerializable, HazelcastInstanceAware, ReadResultSet<E>, Versioned {
 
+    protected transient SerializationService serializationService;
     private transient int minSize;
     private transient int maxSize;
-    private transient IFunction<Object, Boolean> filter;
-    private transient HazelcastInstance hz;
+    private transient IFunction<O, Boolean> filter;
+    private transient Predicate<? super O> predicate;
```
mdogan: What is the difference between a `boolean` returning `IFunction` and a `Predicate`?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -464,8 +520,11 @@ public void writeData(ObjectDataOutput out) throws IOException {
     @Override
     @SuppressWarnings("unchecked")
     public void readData(ObjectDataInput in) throws IOException {
-        name = in.readUTF();
-        emptyRingWaitNotifyKey = new RingbufferWaitNotifyKey(name, EMPTY);
+        assert !in.getVersion().isUnknown();
```
mdogan: When `RingbufferContainer` is serialized from 3.8, version will be unknown, since `RingbufferContainer` in 3.8 is not `Versioned`.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -417,22 +440,55 @@ private long addInternal(Data dataItem) {
...
+        return ret;
     }
 
     @Override
     public void writeData(ObjectDataOutput out) throws IOException {
+        assert !out.getVersion().isUnknown();
         boolean ttlEnabled = expirationPolicy != null;
-        out.writeUTF(name);
+        if (out.getVersion().isLessThan(Versions.V3_9)) {
+            out.writeUTF(namespace.getObjectName());
```
mdogan: `namespace` is not serialized when `version >= 3.9`?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
```diff
@@ -43,12 +49,11 @@
  * it is null to save space.
  */
 @SuppressWarnings("checkstyle:methodcount")
-public class RingbufferContainer implements IdentifiedDataSerializable {
+public class RingbufferContainer<T> implements IdentifiedDataSerializable, Notifier, Versioned {
```
mdogan: As far as I know, `Notifier ` works only for operations. What does it mean here?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
```diff
@@ -19,43 +19,62 @@
...
     public static final String SERVICE_NAME = "hz:impl:ringbufferService";
-    private final ConcurrentMap<String, RingbufferContainer> containers
-            = new ConcurrentHashMap<String, RingbufferContainer>();
+
+    /**
+     * Map from namespace to actual ringbuffer containers. The namespace defines the service and object name which
+     * is the owner of the ringbuffer container.
+     */
+    private final ConcurrentMap<Integer, ConcurrentMap<ObjectNamespace, RingbufferContainer>> containers
+            = new ConcurrentHashMap<Integer, ConcurrentMap<ObjectNamespace, RingbufferContainer>>();
```
mdogan: Does inner map have to be a `ConcurrentMap`?  I guess it's the per partition map.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
```diff
@@ -101,29 +128,113 @@ public void shutdown(boolean terminate) {
...
+    /**
+     * {@inheritDoc}
+     * <p>
+     * Here we will replicate only ringbuffer containers for the ringbuffer service. Other services are responsible
+     * for migrating their ringbuffer containers on migration events. The main reason is because of fine-grained
+     * anti entropy (replication). The partition data might not be migrated in full but rather in part.
+     * For instance, the partition data for a single map might not be in sync between replicas but the ringbuffer
+     * data might be in sync. This means that the map data will be replicated but the ringbuffer data will not be replicated
+     * which can cause differences between ringbuffer data for some services.
+     * In another words, each service is responsible for replicating all of it's data, including data in other services.
```
mdogan: I don't think this is exactly true. A namespace can be shared between services and while migrating a namespace, migration system asks all participating services to replicate data for that namespace. But still each service is responsible for replicating it's own managed data.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
```diff
@@ -101,29 +128,113 @@ public void shutdown(boolean terminate) {
...
+        final Map<ObjectNamespace, RingbufferContainer> partitionContainers = containers.get(partitionId);
+        if (partitionContainers == null || partitionContainers.isEmpty()) {
+            return null;
+        }
+        final Map<String, RingbufferContainer> migrationData = new HashMap<String, RingbufferContainer>();
+        for (ServiceNamespace namespace : namespaces) {
+            final ObjectNamespace ns = (ObjectNamespace) namespace;
+            final RingbufferContainer container = partitionContainers.get(ns);
+            if (container != null && container.getConfig().getTotalBackupCount() >= event.getReplicaIndex()) {
+                migrationData.put(ns.getObjectName(), container);
```
mdogan: How is `ns.getObjectName()` going to be converted a full namespace on target, without having service name?

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
mmedenjak: None. `IFunction` is the legacy way of doing filtering. I would prefer to keep `Predicate` since it is more common, explicit and doesn't box booleans.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mmedenjak: Ah, this could probably be switched to the event journal PR.

Yes, until now only operations could notify other parked operations. Now, all sorts of map/cache and scheduled tasks such as eviction/expiration can notify parked operations since they add events to the event journal. Since the operation unpark method requires an implementation of  `Notifier` I have to provide one every time an event is added to the journal. Making the container a notifier avoids allocation of a new anonymous implementation.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
mmedenjak: 3.8 members don't send sequence IDs and `ReadResultSet` is returned to the user so yes, the user is free to call it even though the cluster version is 3.8.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mmedenjak: The `ReplicationOperation` which serialises/deserialises the container is `Versioned`. I could make the container `Versioned` as well.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mmedenjak: The `name` in is not necessary since the `ReplicationOperation` has the `name` as well and provides it in the `RingbufferContainer#init` method. 
Hence 3.9 writes it only when the cluster version is lower than 3.9 and ignores it on read.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
mmedenjak: I think the javadoc is outdated. This used to be the case until we figured out how it was supposed to work. I have to fix the javadoc.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
mmedenjak: It doesn't, will fix.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
mmedenjak: Good point, will fix. In 3.8 it can only be `RingbufferService`, now we have to serialise the whole namespace.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
mdogan: I see.

mmedenjak: Also Predicate is the de-facto way of filtering on IMap which means I would have to adapt it to IFilter, were we to keep it as the way of filtering on Ringbuffer.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
mdogan: Then shouldn't `getSequence(index)` explicitly fail when version is 3.8, similar to other 3.9 features?

mmedenjak: Actually yes, that would be a better option. Thanks

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mdogan: So, other PR will introduce the implementation for `Notifier` mechanism?

mmedenjak: Yes, initially I had one big PR which I've decided to split into two. This change leaked from the other PR. Since I'm addressing the review comments I might as well move this change into the appropriate PR

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mdogan: But serialized `byte[]` from a 3.8 member won't have `versioned` bit set. As far as I know, input version will be unknown in this case.

mmedenjak: Aha, so the version retrieved in `ObjectDataInput#getVersion` is set from the version in the serialised `byte[]`, not the `Cluster#getClusterVersion`? I thought it was the latter.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mmedenjak: Due to the comment on the service name below, this is the resolution : the name will be removed/ignored from the serialised format of RingbufferContainer. The ReplicationOperation will need to serialise the namespaces, not only the names. And it will provide the namespaces on initialisation.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferContainer.java
mdogan: That's what I know. @vbekiaris can clarify.

mmedenjak: Seems this is the case : https://github.com/hazelcast/hazelcast-enterprise/blob/488e4ea80503a19d048a44b0271a61b7acb3a6d1/hazelcast-enterprise/src/main/java/com/hazelcast/internal/serialization/impl/EnterpriseDataSerializableSerializer.java#L146.
I'll adapt the `readData` and `writeData` then.

mmedenjak:
@mdogan updated the replication mechanism. Added a check that disallows creating a ringbuffer if the cluster version is less than 3.9 and service name is not ringbuffer service. This is because we can't replicate that ringbuffer safely - service names are lost as you've noticed.
This means that you can't use the event journal until the cluster version is 3.9 or higher - since we can't replicate it without knowing the service name.
If the cluster version is 3.9 everything works fine - the ringbuffer service keeps track of namespaces, the namespaces are serialised upon replication and deserialised on the target. The event journal PR will just have to include getting the appropriate config based on the namespace - e.g. ringbuffer service namespaces get the ringbuffer config from the usual place but map and cache service namespaces have an another place for rinbuffer config.
What do you think?

mmedenjak:
run-lab-run

Issue #10249, opened 2017-04-06
Proper way of heap-data conversion 


ahmetmircik:
Current workaround is `ToHeapDataConverter` : https://github.com/hazelcast/hazelcast/pull/10245

mmedenjak:
@ahmetmircik there is now a method exposed in OS: `InternalSerializationService#toData(java.lang.Object, com.hazelcast.nio.serialization.DataType)`. Can you use this?

ahmetmircik:
@mmedenjak this issue was originally requested by @jerrinot but i'm ok with the new method `InternalSerializationService#toData(java.lang.Object, com.hazelcast.nio.serialization.DataType`

mmedenjak:
I know there are some concerns as to how it should be exposed but this variant seems fine to me - at least as a start. Perhaps it could be even better but at least you don't need to subclass/override anything in EE, instantiate that EE class and inject it instead of the OS version...

ahmetmircik:
@jerrinot wdyt, can we close this one?

burakcelebi:
Closing with @jerrinot's confirmation.

PR #10685, opened 2017-06-01
Add map and cache event journal

mmedenjak:
Add event journal for tracking mutation events on map and cache.
The events are written in a ringbuffer (event journal) and
can be read from MapProxyImpl, CacheProxy, ClientMapProxy and
ClientCacheProxy.
The reading is done by first subscribing to the stream and
then calling the read method.

Depends on:
https://github.com/hazelcast/hazelcast-client-protocol/pull/87
https://github.com/hazelcast/hazelcast-client-protocol/pull/88

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
@@ -20,11 +20,16 @@ import com.hazelcast.cache.impl.CacheEntryProcessorResult;
 import com.hazelcast.cache.impl.CacheEventListenerAdaptor;
 import com.hazelcast.cache.impl.event.CachePartitionLostEvent;
 import com.hazelcast.cache.impl.event.CachePartitionLostListener;
+import com.hazelcast.cache.impl.journal.EventJournalCacheEvent;
+import com.hazelcast.client.impl.ClientMessageDecoder;
 import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.impl.protocol.codec.CacheAddEntryListenerCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheAddPartitionLostListenerCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheContainsKeyCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheEntryProcessorCodec;
+import com.hazelcast.client.impl.protocol.codec.CacheEventJournalReadCodec;
+import com.hazelcast.client.impl.protocol.codec.CacheEventJournalSubscribeCodec;
+import com.hazelcast.client.impl.protocol.codec.CacheEventJournalSubscribeCodec.ResponseParameters;
 import com.hazelcast.client.impl.protocol.codec.CacheListenerRegistrationCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheLoadAllCodec;
 import com.hazelcast.client.impl.protocol.codec.CacheRemoveEntryListenerCodec;
@@ -32,12 +37,20 @@ import com.hazelcast.client.impl.protocol.codec.CacheRemovePartitionLostListener
 import com.hazelcast.client.spi.ClientContext;
 import com.hazelcast.client.spi.EventHandler;
 import com.hazelcast.client.spi.impl.ClientInvocation;
+import com.hazelcast.client.spi.impl.ClientInvocationFuture;
 import com.hazelcast.client.spi.impl.ListenerMessageCodec;
+import com.hazelcast.client.util.ClientDelegatingFuture;
 import com.hazelcast.config.CacheConfig;
 import com.hazelcast.core.ICompletableFuture;
 import com.hazelcast.core.Member;
+import com.hazelcast.journal.EventJournalInitialSubscriberState;
 import com.hazelcast.nio.Address;
 import com.hazelcast.nio.serialization.Data;
+import com.hazelcast.projection.Projection;
+import com.hazelcast.ringbuffer.ReadResultSet;
+import com.hazelcast.ringbuffer.impl.client.PortableReadResultSet;
+import com.hazelcast.spi.serialization.SerializationService;
+import com.hazelcast.util.function.Predicate;
 
 import javax.cache.CacheException;
 import javax.cache.CacheManager;
@@ -68,19 +81,45 @@ import static com.hazelcast.util.Preconditions.checkNotNull;
 
 /**
  * {@link com.hazelcast.cache.ICache} implementation for Hazelcast clients.
- *
+ * <p>
  * This proxy is the implementation of {@link com.hazelcast.cache.ICache} and {@link javax.cache.Cache} which is returned by
  * {@link HazelcastClientCacheManager}. Represents a cache on client.
- *
+ * <p>
  * This implementation is a thin proxy implementation using hazelcast client infrastructure.
  *
  * @param <K> key type
  * @param <V> value type
  */
+@SuppressWarnings("checkstyle:classfanoutcomplexity")
 public class ClientCacheProxy<K, V> extends AbstractClientCacheProxy<K, V> {
+    private ClientMessageDecoder eventJournalReadResponseDecoder;
+    private ClientMessageDecoder eventJournalSubscribeResponseDecoder;
+
 
     ClientCacheProxy(CacheConfig<K, V> cacheConfig, ClientContext context) {
         super(cacheConfig, context);
+    }
+
+    @Override
+    protected void onInitialize() {
+        super.onInitialize();
+        eventJournalReadResponseDecoder = new ClientMessageDecoder() {
+            @Override
+            public ReadResultSet<?> decodeClientMessage(ClientMessage message) {
+                final CacheEventJournalReadCodec.ResponseParameters params = CacheEventJournalReadCodec.decodeResponse(message);
+                final PortableReadResultSet<?> resultSet
+                        = new PortableReadResultSet<Object>(params.readCount, params.items, params.itemSeqs);
+                resultSet.setSerializationService(getSerializationService());
+                return resultSet;
+            }
+        };
+        eventJournalSubscribeResponseDecoder = new ClientMessageDecoder() {
+            @Override
+            public EventJournalInitialSubscriberState decodeClientMessage(ClientMessage message) {
+                final ResponseParameters resp = CacheEventJournalSubscribeCodec.decodeResponse(message);
+                return new EventJournalInitialSubscriberState(resp.oldestSequence, resp.newestSequence);
+            }
+        };
     }
 
     @Override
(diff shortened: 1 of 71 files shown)


hazelcast/src/main/java/com/hazelcast/cache/impl/journal/RingbufferCacheEventJournalImpl.java
```diff
@@ -0,0 +1,256 @@
...
+        addToEventRingbuffer(namespace, partitionId, CREATED, key, null, value);
+    }
+
+    @Override
+    public void writeRemoveEvent(ObjectNamespace namespace, int partitionId, Data key, Object value) {
+        addToEventRingbuffer(namespace, partitionId, REMOVED, key, value, null);
+    }
+
+    @Override
+    public void writeEvictEvent(ObjectNamespace namespace, int partitionId, Data key, Object value) {
```
ahmetmircik: It can be better to note that different entries can be evicted on primary and on backup as a result of the same call. So for example when `cache#put` returns, its possible to see that primary evicted entry-A but backup evicted entry-B. This leads to see different journals between replicas.

mmedenjak: So if the partition owner is terminated and the backup is promoted you will get different entries?

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationServiceV1.java
```diff
@@ -110,6 +111,30 @@
         registerJavaTypeSerializers();
     }
 
+    @Override
+    public <B extends Data> B toData(Object obj, DataType type) {
```
ahmetmircik: i remember @jerrinot had some objections to add a this kind of methods but not sure if he keeps them still. We opened a related issue previously: https://github.com/hazelcast/hazelcast/issues/10249

mmedenjak: I found it easier to have the same journal code for OS and EE - otherwise EE would have to extend the journal implementation class and override some `toData` method and the EE code would have to instantiate this extension class. This seemed awkward.

hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/ReadResultSetImpl.java
```diff
@@ -88,46 +108,68 @@ public int readCount() {
...
-    private SerializationService getSerializationService() {
-        HazelcastInstanceImpl impl = (HazelcastInstanceImpl) hz;
-        return impl.getSerializationService();
-    }
-
-    public void addItem(Object item) {
+    @Override
+    public long getSequence(int index) {
+        if (seqs == null) {
+            throw new UnsupportedOperationException("Sequence IDs are not available when the cluster version is lower than 3.9");
```
ahmetmircik: is unexpected cluster version the only reason to see sequences as null?

mmedenjak: 3.8 members don't send sequences - that's the only scenario

hazelcast/src/main/java/com/hazelcast/cache/impl/journal/RingbufferCacheEventJournalImpl.java
ahmetmircik: right

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationServiceV1.java
mmedenjak: @jerrinot can you check if moving this method to OS is ok?

hazelcast/src/main/java/com/hazelcast/cache/impl/journal/RingbufferCacheEventJournalImpl.java
mmedenjak: Added description on `MapEventJournal`, `CacheEventJournal` and all the member and client proxies. Please check

mmedenjak:
run-lab-run

hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationServiceV1.java
jerrinot: I am not very excited about fiddling with various data types in a regular serialization service.

mmedenjak: I don't have a strong reason to keep it this way, I just thought it would be simpler than having two EE event journal classes overriding some `toOnHeapData()` method and instantiating those classes in EE code. What do you think? Is there a better way of getting on-heap `Data` in OS?

hazelcast/src/main/java/com/hazelcast/cache/impl/journal/CacheEventJournalReadResultSetImpl.java
```diff
@@ -0,0 +1,57 @@
...
+                    @SuppressFBWarnings("BC_UNCONFIRMED_CAST")
+                    public T transform(InternalEventJournalCacheEvent input) {
+                        return projection.transform((DeserialisingEventJournalCacheEvent<K, V>) input);
+                    }
+                });
+    }
+
+    @Override
+    public void addItem(long seq, Object item) {
+        final InternalEventJournalCacheEvent e = serializationService.toObject(item);
```
mmedenjak: As discussed, the deserialisation of `InternalEventJournalCacheEvent` would happen if we allowed the ringbuffer to store the event in serialised format (or if there was an off-heap ringbuffer). But I've hardcoded the in-memory format for event journal ringbuffers to `OBJECT` - so that call is a no-op. I'll switch to a cast to make it more explicit.

mmedenjak:
run-lab-run

mmedenjak:
@cangencer addressed all comments. Also added a note in `EventJournalConfig` that it is intended to be used with jet :  https://github.com/hazelcast/hazelcast/pull/10685/files#diff-7dd13850a4da1bc206aacde20033ff8aR32

hazelcast/src/main/resources/META-INF/services/com.hazelcast.DataSerializerHook
```diff
@@ -30,3 +30,4 @@ com.hazelcast.internal.usercodedeployment.impl.UserCodeDeploymentSerializerHook
 com.hazelcast.aggregation.impl.AggregatorDataSerializerHook
 com.hazelcast.projection.impl.ProjectionDataSerializerHook
 com.hazelcast.config.ConfigDataSerializerHook
+com.hazelcast.journal.EventJournalDataSerializerHook
```
Donnerbart: Missing newline at EOF (violates some POSIX rules, which is a reason for an according CheckStyle rule, which of course doesn't check this file)

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheProxy.java
```diff
@@ -360,4 +369,61 @@ public boolean removePartitionLostListener(String id) {
...
+     * items have been read by this subscriber. It will return the subscription ID and
+     * the newest and oldest event journal sequence.
+     *
+     * @param partitionId the partition ID of the entries to which we are subscribing
+     * @return the initial subscriber state containing the subscription ID, newest and oldest event journal sequence
+     * @throws UnsupportedOperationException if the cluster version is lower than 3.9 or there is no event journal
+     *                                       configured for this cache
+     * @since 3.9
+     */
+    public EventJournalInitialSubscriberState subscribeToEventJournal(int partitionId) {
```
tkountis: Maybe async by default ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheEventJournalReadTask.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public String getDistributedObjectName() {
+        return parameters.name;
+    }
+
+    @Override
+    public String getMethodName() {
+        return null;
```
tkountis: Missed ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheEventJournalSubscribeTask.java
```diff
@@ -0,0 +1,93 @@
...
+        return CacheEventJournalSubscribeCodec.encodeResponse(state.getOldestSequence(), state.getNewestSequence());
+    }
+
+    @Override
+    public final String getServiceName() {
+        return CacheService.SERVICE_NAME;
+    }
+
+    public Permission getRequiredPermission() {
+        return new CachePermission(parameters.name, ActionConstants.ACTION_READ);
```
tkountis: Different type of action, maybe subscribe ? I believe we have something similar in the listeners

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheEventJournalSubscribeTask.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public String getDistributedObjectName() {
+        return parameters.name;
+    }
+
+    @Override
+    public String getMethodName() {
+        return null;
```
tkountis: Missed ?

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheEventJournalSubscribeTask.java
mmedenjak: Hm, ok. Currently this is a read-only task but in the future it will modify state. I'll try and put it as `SUBSCRIBE`

hazelcast/src/main/java/com/hazelcast/cache/impl/CacheProxy.java
mmedenjak: Hm, good idea.

hazelcast-client/src/test/java/com/hazelcast/client/cache/ClientBasicCacheJournalTest.java
```diff
@@ -0,0 +1,84 @@
...
+import com.hazelcast.journal.EventJournalInitialSubscriberState;
+import com.hazelcast.projection.Projection;
+import com.hazelcast.ringbuffer.ReadResultSet;
+import com.hazelcast.util.function.Predicate;
+import org.junit.After;
+
+import javax.cache.Cache;
+import javax.cache.CacheManager;
+
+public class ClientBasicCacheJournalTest extends BasicCacheJournalTest {
```
mmedenjak: You're right. I think I'll definitely need to do some work on the tests.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/ClientCacheProxy.java
```diff
@@ -472,6 +504,64 @@ public boolean removePartitionLostListener(String id) {
         return getContext().getListenerService().deregisterListener(id);
     }
 
+
+    /**
+     * Subscribe to the event journal for this cache and a specific partition ID.
+     * The method will register the subscription so that the event journal knows which
+     * items have been read by this subscriber. It will return the subscription ID and
```
mmedenjak: Very likely, comes from the fact that I write down the behaviour all over the place and then forget to update it when it is updated.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/cache/CacheEventJournalSubscribeTask.java
mmedenjak: Hm, there's `ACTION_LISTEN`. WDYT?

PR #10687, opened 2017-06-01
Removed superfluous initDefaultMaxSizeForOnHeapMaps() for default maps

Donnerbart:
Since the Near Cache implementations have been unified, we have different default values for the `EvictionConfig.getSize()`:
* `Integer.MAX_VALUE` for on-heap IMaps
* `10000` for JCache and off-heap IMaps

We use `NearCacheConfigAccessor.initDefaultMaxSizeForOnHeapMaps()` to set the correct default value for IMap and JCache, if no custom value was set.

The default maps have no `NearCacheConfig`, so there is nothing to fix. If a user modifies the default configuration, then the `getMapConfig()` call doesn't enter the branch, which creates a new `MapConfig` and tries to apply the fix.

The added tests check all combinations of changed and un-changed map and cache combinations (default and named configurations).

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/Config.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/Config.java
@@ -454,7 +454,6 @@ public class Config {
         if (defConfig == null) {
             defConfig = new MapConfig();
             defConfig.setName("default");
-            initDefaultMaxSizeForOnHeapMaps(defConfig.getNearCacheConfig());
             mapConfigs.put(defConfig.getName(), defConfig);
         }
         config = new MapConfig(defConfig);
--- a/hazelcast/src/test/java/com/hazelcast/config/ConfigTest.java
+++ b/hazelcast/src/test/java/com/hazelcast/config/ConfigTest.java
@@ -25,11 +25,19 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
+import java.io.InputStream;
+
+import static com.hazelcast.config.EvictionConfig.DEFAULT_MAX_ENTRY_COUNT;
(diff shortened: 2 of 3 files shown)


Donnerbart:
After some general discussion with Jaromir about `findFooConfig()` vs. `getFooConfig()` I think this PR might be wrong.

The `getFooConfig()` methods are for the users to setup a `Config` before starting Hazelcast. So you should be able to call `getMapConfig()` multiple times, before I apply the default value initialization. The fix should just be applied in `findMapConfig()`, which returns the read-only variant for runtime usage.

```java
MapConfig mapConfig = config.getMapConfig("myMap); // here the fix would already be applied -> size = Integer.MAX_VALUE
mapConfig.setInMemoryFormat(NATIVE); // ups, now the fix isn't correct anymore
```
So it was correct to omit the fix on that line. I'll adapt the tests though to check that `findMapConfig()` will return the fixed configuration.

pveentjer:
@Donnerbart can you rebase so we can merge this PR?

mdogan:
> @Donnerbart can you rebase so we can merge this PR?

👆

mmedenjak:
@Donnerbart is this PR still valid? Can we merge it?

Donnerbart:
I have rebased, added more tests and fixed the commit message + PR description.

The removed line does nothing, since a `new MapConfig()` has no `NearCacheConfig`, so the `initDefaultMaxSizeForOnHeapMaps()` cannot change the default values.

The `getXXXConfig()` and `findXXXConfig()` methods seem to work as expected. The get methods don't change anything, the find methods apply the fix.

jerrinot:
moving to 3.11.

Donnerbart:
This should better be fixed with a Config API cleanup of the next major release.

PR #10697, opened 2017-06-05
Added a configurable idle strategy for the partition threads

pveentjer:
By default a partition thread blocks and needs to be notified (e.g. io thread) if a pending operation arrives. In EC2 notification apparently is a big problem. When everything
is put into blocking mode (io,response,partition-thread); throughput at some atomiclong.get
benchmark was 290k op/s. When setting io/response to spinning; performance increased to 500K op/s;
when reducing the number of partition threads and setting them to backoff, performance jumped to
750K op/s.

The impact of this optimization is very very test/environment specific. In some cases it doesn't provide any gain.


pveentjer:
Has no performance impact for default situation; default is to block and it remains that way.

Issue #10400, opened 2017-04-20
ClientMapIssueTest


Donnerbart:
**com.hazelcast.client.map.ClientMapIssueTest.testNoOperationTimeoutException_whenUserCodeLongRunning**
```java
com.hazelcast.core.OperationTimeoutException: EntryOperation invocation failed to complete due to operation-heartbeat-timeout. Current time: 2017-04-20 12:41:18.212. Start time: 2017-04-20 12:41:08.208. Total elapsed time: 10004 ms. Last operation heartbeat: 2017-04-20 12:41:11.224. Last operation heartbeat from member: 2017-04-20 12:41:17.222. Invocation{op=com.hazelcast.map.impl.operation.EntryOperation{serviceName='hz:impl:mapService', identityHash=134585789, partitionId=232, replicaIndex=0, callId=-4, invocationTime=1492692068208 (2017-04-20 12:41:08.208), waitTimeout=-1, callTimeout=2000, name=b9c91e06-643c-4078-876b-028128190d0b}, tryCount=250, tryPauseMillis=500, invokeCount=1, callTimeoutMillis=2000, firstInvocationTimeMs=1492692068208, firstInvocationTime='2017-04-20 12:41:08.208', lastHeartbeatMillis=1492692071224, lastHeartbeatTime='2017-04-20 12:41:11.224', target=[127.0.0.1]:5003, pendingResponse={VOID}, backupsAcksExpected=0, backupsAcksReceived=0, connection=null}
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.newOperationTimeoutException(InvocationFuture.java:157)
	at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.resolve(InvocationFuture.java:104)
	at com.hazelcast.spi.impl.AbstractInvocationFuture$1.run(AbstractInvocationFuture.java:243)
	at com.hazelcast.client.impl.protocol.task.AbstractPartitionMessageTask.execute(AbstractPartitionMessageTask.java:78)
	at com.hazelcast.spi.impl.AbstractInvocationFuture.unblock(AbstractInvocationFuture.java:239)
	at com.hazelcast.spi.impl.AbstractInvocationFuture.andThen(AbstractInvocationFuture.java:215)
	at com.hazelcast.client.impl.protocol.task.AbstractPartitionMessageTask.processMessage(AbstractPartitionMessageTask.java:69)
[...]

PR #10701, opened 2017-06-05
Test testNoOperationTimeoutException_whenUserCodeLongRunning may sometime fail - issue #10400

ihsandemir:
Increased the insufficient timeout time for the call timeout in the testNoOperationTimeoutException_whenUserCodeLongRunning test and marked the test as a NightlyTest. Assuming that 10 seconds will hopefully be enough even when there is gc pause. Total test duration is 2 * 10 = 20 seconds.

Fixes issue https://github.com/hazelcast/hazelcast/issues/10400


sancar:
@ihsandemir Do we have test in maintenance ? If yes, don't forget to fix in maintenance.

PR #10707, opened 2017-06-06
Utility methods for cluster version conditions in distributed objects

vbekiaris:
Adds cluster version condition checks in `AbstractDistributedObject`, useful for gating user calls depending on current cluster version.


vbekiaris:
run-lab-run

PR #10713, opened 2017-06-07
Fix ClientMapNearCacheTest to start the instances before the client listener registration

### LINKED ISSUE #10024: ClientMapNearCacheTest.receives_one_clearEvent_after_mapEvictAll_call_from_member
```
java.lang.AssertionError: Expecting only 1 clear event expected:<1> but was:<0>
	at com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheTest$29.run(ClientMapNearCacheTest.java:1053)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:967)
	at com.hazelcast.test.HazelcastTestSupport.assertTrueEventually(HazelcastTestSupport.java:984)
	at com.hazelcast.client.map.impl.nearcache.ClientMapNearCacheTest.receives_one_clearEvent_after_mapEvictAll_call_from_member(ClientMapNearCacheTest.java:1050)
```
https://hazelcast-l337.ci.cloudbees.com/view/Hazelcast/job/Hazelcast-3.x-freestyle/134/testReport/junit/com.hazelcast.client.map.impl.nearcache/ClientMapNearCacheTest/receives_one_clearEvent_after_mapEvictAll_call_from_member_batchInvalidationEnabled_false_/

possibly related to #9940

ihsandemir:
Initialises the members before the client initialisation and invalidation listener registration so that the test does not fail due to the scenario described at https://github.com/hazelcast/hazelcast/issues/10024#issuecomment-306776701 .

Also, changed some tests to run with two instances instead of one so that it tests better the case where only one invalidation event is received.

fixes #10024

### DIFF
--- a/hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
+++ b/hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
@@ -84,9 +84,6 @@ public class ClientMapNearCacheTest extends NearCacheTestSupport {
 
     protected final TestHazelcastFactory hazelcastFactory = new TestHazelcastFactory();
 
-    private static final Random rand = new Random();
-    private static final int NIGHTLY_TEST_WAIT_DURATION = 30;
-
     @After
     public void tearDown() {
         hazelcastFactory.shutdownAll();
@@ -911,7 +908,7 @@ public class ClientMapNearCacheTest extends NearCacheTestSupport {
             public void run() {
                 assertEquals("Expecting only 1 clear event", 1, handler.getClearEventCount());
             }
-        }, getRandomTime());
+        }, 10);
     }
 
     @Test
@@ -962,7 +959,7 @@ public class ClientMapNearCacheTest extends NearCacheTestSupport {
             public void run() {
                 assertEquals("Expecting only 1 clear event", 1, handler.getClearEventCount());
             }
-        }, getRandomTime());
+        }, 10);
     }
 
     private void mapClearFromMember(final ClearEventCounterEventHandler handler) {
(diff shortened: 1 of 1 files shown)


hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
```diff
@@ -911,48 +911,50 @@ public void receives_one_clearEvent_after_mapClear_call_from_client() {
         IMap<Object, Object> anotherClientMap = anotherClient.getMap(clientMap.getName());
         anotherClientMap.clear();
 
-        // sleep for a while to see there is another clear event coming
-        sleepSeconds(2);
-
         assertTrueEventually(new AssertTask() {
```
ahmetmircik: seems this assertion is redundant

sancar: Could not assertTrueAllTheTime fail without this check ?

ahmetmircik: Two assertions makes it hard to aware what is being tested. We can introduce 2 test instead if two assertions are needed. 

Also you are right, `assertTrueAllTheTime` was confused me.

ihsandemir: OK, added nightly versions of the tests as well.

hazelcast-client/src/test/java/com/hazelcast/client/map/impl/nearcache/ClientMapNearCacheTest.java
```diff
@@ -83,6 +84,9 @@
 
     protected final TestHazelcastFactory hazelcastFactory = new TestHazelcastFactory();
 
+    private static final Random rand = new Random();
```
Donnerbart: `Random` should not be static (and if an UPPERCASE name should be used). Would be better to make it a normal final field, so each test gets its own random generator.

ihsandemir: Ok, removed Random :)

PR #10714, opened 2017-06-07
Fixed additional key serializations in Near Cache

Donnerbart:
We need to have a new "centralized" key serialization in the Near Cache proxies for all methods, which create a `NearCacheRecord`. Otherwise the normal proxy and the `NearCacheRecordStore` will serialize the key independently. Beside the `proxy.get()` the `proxy.getAll()` methods are the main candidates which needed a fix.

I unified the `getAll()` method to this schema:
```java
@Override
protected void getAllInternal(Set<K> keys, List<Data> dataKeys,
                              List<Object> resultingKeyValuePairs) {
    // we have a key map for a fast lookup of the serialized keys
    Map<Object, Data> keyMap = null;
    // if the Near Cache uses serialized keys, we directly serialize them
    if (serializeKeys) {
        keyMap = toDataKeys(keys, dataKeys);
    }
    // we chose the wanted key collection
    Collection ncKeys = serializeKeys ? dataKeys : keys;

    // we populate the result from the Near Cache,
    // removing cached keys from the collection
    populateResultFromNearCache(ncKeys, resultingKeyValuePairs);
    // if the collection is empty, we are done
    // this shortcut is new, but the same as in get()
    if (ncKeys.isEmpty()) {
        return;
    }

    // if keys were not serialized before
    // we have to do it here to prevent double serialization
    if (!serializeKeys) {
        keyMap = toDataKeys(ncKeys, dataKeys);
    }

    // in the reservation method we need the serialized key, so we pass the lookup map
    Map<Object, Long> reservations = getNearCacheReservations(ncKeys, keyMap);
    try {
        int oldSize = resultingKeyValuePairs.size();
        // we also pass the serialized keys to the normal proxy
        super.getAllInternal(keys, dataKeys, resultingKeyValuePairs);
        populateResultFromRemote(oldSize, resultingKeyValuePairs, reservations);
    } finally {
        releaseReservedKeys(reservations);
    }
}
```
The `get()` methods were also modified in the same way (but they are easier to understand).

---

One known issue is that
```java
protected final IFunction<K, R> reserveForUpdate = new ReserveForUpdateFunction();
```
had to be replaced with a new instance per call, since we need to pass the serialized key as additional parameter at its usage. We could try to use a `ThreadLocal` here, but I'm unsure since the thread is a user thread. So a proper cleanup could be impossible, leading to a new `ThreadLocal` leakage. Nevertheless this should be done in another PR (so we can measure the performance impact of new instance vs. `ThreadLocal` separately).

---

Depends on https://github.com/hazelcast/hazelcast/pull/10723


Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

Donnerbart:
run-lab-run

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -219,123 +219,140 @@ protected void onPutIfAbsentAsyncInternal(K key, V value, Data keyData, Data val
...
+    }
+
+    private void populateResultFromRemote(int currentSize, List<Object> resultingKeyValuePairs, Map<Object, Long> reservations) {
+        for (int i = currentSize; i < resultingKeyValuePairs.size(); i += 2) {
+            Data keyData = (Data) resultingKeyValuePairs.get(i);
+            Data valueData = (Data) resultingKeyValuePairs.get(i + 1);
+
+            Object ncKey = serializeKeys ? keyData : toObject(keyData);
+            if (!serializeKeys) {
+                resultingKeyValuePairs.set(i, ncKey);
```
Donnerbart: If we had to deserialize the key here (for the reservation lookup), we replace the Data key in the result with it, to prevent a second deserialization in the final result map population.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -219,123 +219,140 @@ protected void onPutIfAbsentAsyncInternal(K key, V value, Data keyData, Data val
...
         if (serializeKeys) {
             for (int partitionId = 0; partitionId < entriesPerPartition.length; partitionId++) {
                 List<Map.Entry<Data, Data>> entries = entriesPerPartition[partitionId];
                 if (entries != null) {
                     for (Map.Entry<Data, Data> entry : entries) {
+                        Data key = entry.getKey();
                         if (isCacheOrInvalidate) {
-                            cacheOrInvalidate(entry.getKey(), null, entry.getValue());
+                            // FIXME: the null value produces an unwanted deserialization if Near Cache in-memory format is OBJECT
+                            cacheOrInvalidate(key, key, null, entry.getValue());
```
Donnerbart: This is the remaining code part, where we eventually deserialize the value twice in a distinct use case.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
```diff
@@ -219,123 +218,139 @@ protected void onPutIfAbsentAsyncInternal(K key, V value, Data keyData, Data val
...
+        }
+        return reservations;
+    }
+
+    private void populateResultFromRemote(int currentSize, List<Object> resultingKeyValuePairs, Map<Object, Long> reservations) {
+        for (int i = currentSize; i < resultingKeyValuePairs.size(); i += 2) {
+            Data keyData = (Data) resultingKeyValuePairs.get(i);
+            Data valueData = (Data) resultingKeyValuePairs.get(i + 1);
+
+            Object ncKey = serializeKeys ? keyData : toObject(keyData);
```
mmedenjak: Could we eliminate this deserialisation by having a bidirectional `object-data` map? Do you think it is worth it?

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -224,7 +224,7 @@ private int loadKeySet(BufferingInputStream bis, DataStructureAdapter<Object, ?>
                 break;
             }
             Data key = new HeapData(payload);
-            builder.add(serializeKeys ? key : serializationService.toObject(key));
+            builder.add(serializationService.toObject(key));
```
mmedenjak: Too bad the keys are about to be serialised again soon. If only we had some lazy serialising/deserialising wrapper like `CachedQueryEntry` to avoid serialisations and deserialisations all through the call chain

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
```diff
@@ -224,7 +224,7 @@ private int loadKeySet(BufferingInputStream bis, DataStructureAdapter<Object, ?>
                 break;
             }
             Data key = new HeapData(payload);
-            builder.add(serializeKeys ? key : serializationService.toObject(key));
```
Donnerbart: Good catch!

hazelcast/src/main/java/com/hazelcast/internal/nearcache/impl/preloader/NearCachePreloader.java
Donnerbart: I thought about adding code to the `getAll()` method to deal with serialized keys as input as well. But this would make the proxy code even more complex. Since the preloader works in its own thread I went this way for now. Maybe we can have a small (internal) method which don't even return the results, just populates the Near Cache and can handle serialized keys.

hazelcast-client/src/main/java/com/hazelcast/client/cache/impl/NearCachedClientCacheProxy.java
Donnerbart: I had the same thought here. It's not just the deserialization cost, it's also the issue with the OBJECT key not being the same instance as the original one. So we rely on `equals()`/`hashCode()` here to have `Long reservationId = reservations.get(ncKey);` working properly. This is actually a real issue, since it will fail silently otherwise and never populate the Near Cache -> super hard to debug when a customer reports a "Near Cache is not working". A bi-directional map would solve this bigger issue as well.

Issue #10698, opened 2017-06-05
EntryListenerOnReconnectTest.testClusterReconnectMultipleServersDueToHeartbeatSmartRouting & testTemporaryBlockedNoDisconnectionMultipleServerSmartRouting 


hasancelik:
```
com.hazelcast.core.HazelcastException: Listener can not be added 
	at com.hazelcast.client.spi.impl.listener.ClientSmartListenerService$1.call(ClientSmartListenerService.java:81)
	at com.hazelcast.client.spi.impl.listener.ClientSmartListenerService$1.call(ClientSmartListenerService.java:67)
	at java.util.concurrent.FutureTask$Sync.innerRun(FutureTask.java:303)
	at java.util.concurrent.FutureTask.run(FutureTask.java:138)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:98)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:206)
	at java.util.concurrent.ThreadPoolExecutor$Worker.runTask(ThreadPoolExecutor.java:895)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:918)
	at java.lang.Thread.run(Thread.java:662)
	at com.hazelcast.util.executor.HazelcastManagedThread.executeRun(HazelcastManagedThread.java:64)
	at com.hazelcast.util.executor.HazelcastManagedThread.run(HazelcastManagedThread.java:80)
Caused by: java.io.IOException: Packet not send to [127.0.0.1]:5429
	at com.hazelcast.client.spi.impl.ClientInvocationServiceSupport.send(ClientInvocationServiceSupport.java:114)
	at com.hazelcast.client.spi.impl.ClientSmartInvocationServiceImpl.invokeOnConnection(ClientSmartInvocationServiceImpl.java:79)
	at com.hazelcast.client.spi.impl.ClientInvocation.invokeOnSelection(ClientInvocation.java:141)
	at com.hazelcast.client.spi.impl.ClientInvocation.invokeUrgent(ClientInvocation.java:134)
	at com.hazelcast.client.spi.impl.listener.ClientSmartListenerService.invoke(ClientSmartListenerService.java:112)
[...]

PR #10718, opened 2017-06-08
Increased heartbeat timeout to 20 sec in listener tests

sancar:
Test failures reported in #10698 are caused by heartbeat timeout
to unexpected members. Timeout value is increased to 20 sec to
avoid false alarms.

testTemporaryBlockedNoDisconnectionMultipleServerSmartRouting
is cleaned up with preserving the behaviour.

fixes https://github.com/hazelcast/hazelcast/issues/10698


Donnerbart:
Note: If tests are important for code coverage, make them `SlowTest`, not `NightlyTest`, otherwise they will be executed in the wrong Jenkins job and not considered for coverage measurement.

sancar:
run-lab-run

PR #10720, opened 2017-06-08
Support for HD memory Indexes

tkountis:
Re-based version of the original PR #10440 
------------------------------------------------------

Commit has modified author - pointing to @tombujok

EE PR: hazelcast/hazelcast-enterprise#1518

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
+++ b/hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
@@ -21,8 +21,10 @@ import com.hazelcast.client.impl.protocol.ClientMessage;
 import com.hazelcast.client.impl.protocol.task.AbstractCallableMessageTask;
 import com.hazelcast.core.Member;
 import com.hazelcast.instance.Node;
+import com.hazelcast.internal.cluster.Versions;
 import com.hazelcast.map.QueryResultSizeExceededException;
 import com.hazelcast.map.impl.MapService;
+import com.hazelcast.map.impl.MapServiceContext;
 import com.hazelcast.map.impl.query.Query;
 import com.hazelcast.map.impl.query.QueryOperation;
 import com.hazelcast.map.impl.query.QueryPartitionOperation;
@@ -32,9 +34,11 @@ import com.hazelcast.projection.Projection;
 import com.hazelcast.query.Predicate;
 import com.hazelcast.security.permission.ActionConstants;
 import com.hazelcast.security.permission.MapPermission;
+import com.hazelcast.spi.Operation;
 import com.hazelcast.spi.impl.operationservice.InternalOperationService;
 import com.hazelcast.util.BitSetUtils;
 import com.hazelcast.util.IterationType;
+import com.hazelcast.version.Version;
 
 import java.security.Permission;
 import java.util.ArrayList;
@@ -116,10 +120,16 @@ public abstract class AbstractMapQueryMessageTask<P, QueryResult extends Result,
         List<Future> futures = new ArrayList<Future>(members.size());
         final InternalOperationService operationService = nodeEngine.getOperationService();
         final Query query = buildQuery(predicate);
+
+        Version clusterVersion = nodeEngine.getClusterService().getClusterVersion();
+        MapService mapService = nodeEngine.getService(getServiceName());
+        MapServiceContext mapServiceContext = mapService.getMapServiceContext();
+
         for (Member member : members) {
             try {
                 Future future = operationService.createInvocationBuilder(SERVICE_NAME,
-                        new QueryOperation(query), member.getAddress())
+                        createQueryOperation(query, clusterVersion, mapServiceContext),
+                        member.getAddress())
                         .invoke();
                 futures.add(future);
             } catch (Throwable t) {
@@ -139,9 +149,9 @@ public abstract class AbstractMapQueryMessageTask<P, QueryResult extends Result,
 
     private Query buildQuery(Predicate predicate) {
         Query.QueryBuilder builder =
-            Query.of().mapName(getDistributedObjectName())
-                      .predicate(predicate)
-                      .iterationType(getIterationType());
+                Query.of().mapName(getDistributedObjectName())
+                        .predicate(predicate)
+                        .iterationType(getIterationType());
 
         if (getAggregator() != null) {
             builder = builder.aggregator(getAggregator());
@@ -205,9 +215,14 @@ public abstract class AbstractMapQueryMessageTask<P, QueryResult extends Result,
                                                        Predicate predicate) {
 
         final InternalOperationService operationService = nodeEngine.getOperationService();
+        Version clusterVersion = nodeEngine.getClusterService().getClusterVersion();
+        MapService mapService = nodeEngine.getService(getServiceName());
+        MapServiceContext mapServiceContext = mapService.getMapServiceContext();
+
         Query query = buildQuery(predicate);
         for (Integer partitionId : missingPartitionsList) {
-            QueryPartitionOperation queryPartitionOperation = new QueryPartitionOperation(query);
+            QueryPartitionOperation queryPartitionOperation = createQueryPartitionOperation(
+                    query, clusterVersion, mapServiceContext);
             queryPartitionOperation.setPartitionId(partitionId);
             try {
                 Future future = operationService.invokeOnPartition(SERVICE_NAME,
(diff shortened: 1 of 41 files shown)


hazelcast/src/main/java/com/hazelcast/query/impl/Indexes.java
```diff
@@ -93,6 +100,10 @@ public void saveEntryIndex(QueryableEntry queryableEntry, Object oldValue) throw
         }
     }
 
+    public boolean isGlobal() {
```
ahmetmircik: can you please add some javaDoc on this method as an answer to questions what is `global` & what is it used for?

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryDispatcher.java
```diff
@@ -125,6 +137,16 @@ protected QueryDispatcher(MapServiceContext mapServiceContext) {
         }
     }
 
+    // for rolling-upgrade compatibility, can be deleted in 4.0
```
ahmetmircik: seems not only for rolling upgrade compatibility but also about creating an enterprise counterpart

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
```diff
@@ -92,32 +96,89 @@ public Result runIndexOrPartitionScanQueryOnOwnedPartitions(Query query)
...
+        } else {
+            // else: if fallback to full table scan also failed to return any results due to migrations,
+            // then return empty result set without any partition IDs set (so that it is ignored by callers).
+            return populateEmptyResult(query, initialPartitions);
+        }
+    }
+
+    // MIGRATION UNSAFE QUERYING - MIGRATION STAMPTS ARE NOT VALIDATED, so assumes a run on partition-thread
+    // for a single partition. If the index is global it won't be asked
+    public Result runPartitionIndexOrPartitionScanQueryOnGivenOwnedPartition(Query query, int partitionId) {
```
ahmetmircik: maybe an assertion like this:  `assert nodeEngine.getPartitionService().isPartitionOwner(partitionId)` can be added to get some early notification against potential development hazards

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
```diff
@@ -92,32 +96,89 @@ public Result runIndexOrPartitionScanQueryOnOwnedPartitions(Query query)
...
             // else: if fallback to full table scan also failed to return any results due to migrations,
             // then return empty result set without any partition IDs set (so that it is ignored by callers).
-            return resultProcessorRegistry.get(query.getResultType()).populateResult(query,
-                    queryResultSizeLimiter.getNodeResultLimit(initialPartitions.size()));
+            return populateEmptyResult(query, partitions);
         }
     }
 
-    private Result populateTheResult(Query query, Collection<QueryableEntry> entries, Collection<Integer> initialPartitions) {
+    Result runPartitionScanQueryOnGivenOwnedPartition(Query query, int partitionId) {
```
ahmetmircik: same here.

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -108,7 +116,15 @@ public MapContainer(final String name, final Config config, final MapServiceCont
         this.objectNamespace = new DistributedObjectNamespace(MapService.SERVICE_NAME, name);
         initWanReplication(nodeEngine);
         this.extractors = new Extractors(mapConfig.getMapAttributeConfigs(), config.getClassLoader());
-        this.indexes = new Indexes((InternalSerializationService) serializationService, extractors);
+        if (!mapConfig.getInMemoryFormat().equals(NATIVE)) {
```
ahmetmircik: we have an `EnterpriseMapContainer`, adding this if clause there can be better.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
```diff
@@ -231,4 +246,26 @@ private void collectResultsFromMissingPartitions(BitSet finishedPartitions, Coll
             }
         }
     }
+
+    // for rolling-upgrade compatibility, can be deleted in 4.0
+    private Operation createQueryOperation(Query query, Version clusterVersion, MapServiceContext mapServiceContext) {
```
ahmetmircik: cant we put this method in DefaultMapOperationProvider to get rid of some code duplication?

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
```diff
@@ -76,6 +80,10 @@
     protected final InterceptorRegistry interceptorRegistry = new InterceptorRegistry();
     protected final IFunction<Object, Data> toDataFunction = new ObjectToData();
     protected final ConstructorFunction<Void, RecordFactory> recordFactoryConstructor;
+    protected final boolean globalIndex;
+    protected final Indexes indexes;
+    protected final ConcurrentHashMap<String, Boolean> indexDefinitions = new ConcurrentHashMap<String, Boolean>();
```
ahmetmircik: can you please put some short javaDoc to explain `indexDefinitions` and its usage?

hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
```diff
@@ -225,4 +239,29 @@ public void setLastCleanupTimeCopy(long lastCleanupTimeCopy) {
         this.lastCleanupTimeCopy = lastCleanupTimeCopy;
     }
 
+    // -------------------------------------------------------------------
+    // WATCH OUT: never use directly! use MapContainer.getIndex() instead
```
ahmetmircik: can you please also add the reason for not using directly? 
Also is it possible to throw some exception in case of misuse?

hazelcast/src/main/java/com/hazelcast/nio/VersionSettable.java
```diff
@@ -0,0 +1,37 @@
...
+import com.hazelcast.version.Version;
+
+/**
+ * Enables setting the version on the implementing object.
+ * It may be any Version (cluster version, node version, custom version, etc.) It's up to the implementer.
+ *
+ * @since 3.9
+ */
+
+public interface VersionSettable {
```
ahmetmircik: I think it is better to introduce `VersionSettable` in a separate PR to make change more visible or maybe in a separate commit, now i'm questioning the relation of it with this PR. Or maybe i'm still missing something

tombujok: I removed it. It was a left-over that was needed earlier but not anymore.

hazelcast/src/main/java/com/hazelcast/map/impl/PartitionContainer.java
tombujok: Done.

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
tombujok: It's a non-native case (there's a not operator there), so I need this check here.
I removed the else clause, that was doing nothing anyway, so it's less confusing.

hazelcast/src/main/java/com/hazelcast/query/impl/Indexes.java
tombujok: Done.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryDispatcher.java
tombujok: I updated the comment, to make it more accurate :)

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
tombujok: Done.

hazelcast/src/main/java/com/hazelcast/map/impl/query/QueryRunner.java
tombujok: Done.

hazelcast/src/main/java/com/hazelcast/map/impl/MapContainer.java
tombujok: Done.

hazelcast/src/main/java/com/hazelcast/client/impl/protocol/task/map/AbstractMapQueryMessageTask.java
tombujok: This sounds legit, but then I'll have the same duplication in the MapOperationProvider and HDMapOperationProvider, so it's no good...
I'm planning to get rid of the duplication of the query logic in the AbstractMapQueryMessageTask, and use the MapQueryEngine there -> then I'll get rid of this duplication implicitly.

PR #10723, opened 2017-06-08
Added getAll() method to Near Cache serialization count tests

Donnerbart:
We had no check for the `getAll()` method.

This also adds the shortcut to leave the `getAllInternal()` methods if all entries were fetched from the Near Cache (same as in the `get()` call). Without this we even do a remote call if all entries are in the Near Cache. This is super inefficient.

I also had to clone the collection of user keys, since we use `iterator.remove()` on them, to just fetch remote keys which are not in the Near Cache. If a user submits an modifiable collection the `iterator.remove()` fails with an `UnsupportedOperationException`.

### DIFF
--- a/hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
+++ b/hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
@@ -51,6 +51,7 @@ import com.hazelcast.util.executor.CompletedFuture;
 
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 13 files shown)


hazelcast-client/src/main/java/com/hazelcast/client/proxy/NearCachedClientMapProxy.java
```diff
@@ -391,18 +392,29 @@ protected void loadAllInternal(boolean replaceExistingValues, Collection<?> keys
...
     @Override
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"checkstyle:npathcomplexity", "unchecked"})
     protected List<MapGetAllCodec.ResponseParameters> getAllInternal(Set<K> keys, Map<Integer, List<Data>> partitionToKeyData,
                                                                      Map<K, V> result) {
         Map<Object, Long> reservations = new HashMap<Object, Long>();
         try {
             if (serializeKeys) {
                 fillPartitionToKeyData(keys, partitionToKeyData);
+                boolean allEmpty = true;
```
Donnerbart: This is just a temporary fix to prevent the remote calls if all values are served from the Near Cache. A proper fix is made in https://github.com/hazelcast/hazelcast/pull/10714

PR #10724, opened 2017-06-09
MetricsRegistryImpl should not hold on to ClassLoader instances

cangencer:
MetricsRegistryImpl currently holds on to ClassLoader instances
because of an internal cache. The cache has been changed to use
ConcurrentReferenceHashMap instead which uses WeakReference for keys.


hazelcast/src/main/java/com/hazelcast/internal/metrics/impl/MetricsRegistryImpl.java
```diff
@@ -65,7 +66,7 @@ public int compare(ProbeInstance o1, ProbeInstance o2) {
 
     private final ConcurrentMap<String, ProbeInstance> probeInstances = new ConcurrentHashMap<String, ProbeInstance>();
     private final ConcurrentMap<Class<?>, SourceMetadata> metadataMap
-            = new ConcurrentHashMap<Class<?>, SourceMetadata>();
+            = new ConcurrentReferenceHashMap<Class<?>, SourceMetadata>();
```
pveentjer: Can you add a line of comment why this is done?

PR #10726, opened 2017-06-10
Keep broadcasting group password to support RU

tkountis:
Relax previous feature to prevent group password leak. Original implementation here (https://github.com/hazelcast/hazelcast/pull/10376)
We now still send the password when cluster version is unknown or less than 3.9, allowing clusters to be formed between 3.8, 3.8.1 and >= 3.9 members. (>=3.8.2 were already working due to a patch)

This will eventually be removed in 4.0 altogether.
Fixing hazelcast/hazelcast-enterprise#1492

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
+++ b/hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
@@ -19,10 +19,12 @@ package com.hazelcast.internal.cluster.impl;
 import com.hazelcast.config.Config;
 import com.hazelcast.config.GroupConfig;
 import com.hazelcast.config.PartitionGroupConfig;
+import com.hazelcast.internal.cluster.Versions;
 import com.hazelcast.nio.ObjectDataInput;
 import com.hazelcast.nio.ObjectDataOutput;
 import com.hazelcast.nio.serialization.IdentifiedDataSerializable;
 import com.hazelcast.util.EmptyStatement;
+import com.hazelcast.version.Version;
 
 import java.io.IOException;
 import java.util.HashMap;
@@ -41,9 +43,13 @@ public final class ConfigCheck implements IdentifiedDataSerializable {
 
     private String groupName;
 
+    private String groupPassword;
+
     private String joinerType;
 
     private boolean partitionGroupEnabled;
+
+    private Version clusterVersion = Version.UNKNOWN;
 
     private PartitionGroupConfig.MemberGroupType memberGroupType;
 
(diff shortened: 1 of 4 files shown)


hazelcast/src/test/java/com/hazelcast/cluster/ConfigCheckTest.java
```diff
@@ -50,6 +61,86 @@ public void whenGroupNameDifferent_thenFalse() {
...
+        final Config config = new Config();
+        config.getNetworkConfig().getJoin().getMulticastConfig()
+               .setEnabled(true).setMulticastTimeoutSeconds(3);
+        config.getNetworkConfig().getJoin().getTcpIpConfig().setEnabled(false);
+
+        final AtomicBoolean leaked = new AtomicBoolean(false);
+
+        ObjectDataOutput odo = mock(ObjectDataOutput.class);
+
+        try {
```
pveentjer: I think you can remove all these try/catches. Just let the exception propagate; the test will fails as well.

Donnerbart:
Just a reminder: if we allow 3.8 and 3.8.1 to join a 3.9 cluster again, we should adapt the compatibility tests again. We removed those versions there, since the join failed.

tkountis:
@Donnerbart I recall only removing compatibility checks related to password validation, two nodes with different passwords were compatible before. Did you spot something I missed ?

Donnerbart:
I'm talking about the "rolling upgrade" compatibility tests with the `HazelcastStarter` magic factory. After your first change 3.8 and 3.8.1 stopped working together with 3.9. Just 3.8.2 could join the cluster with 3.9, due to the ignored group password check.

If we revert that now (3.9 sending the group password again), so all 3.8.x versions are able to join, we have to add those versions again (we removed them to get the compatibility tests running). See `CompatibilityTestHazelcastInstanceFactory.RELEASED_VERSIONS` in EE code.

tkountis:
Thanks @Donnerbart didn't think of that.
PR with the versions restored https://github.com/hazelcast/hazelcast-enterprise/pull/1527. Tested a couple of compatibility classes locally, and got a pass.

hazelcast/src/main/java/com/hazelcast/internal/cluster/impl/ConfigCheck.java
```diff
@@ -172,7 +184,13 @@ public int getId() {
     public void writeData(ObjectDataOutput out) throws IOException {
         out.writeUTF(groupName);
         //TODO @tkountis remove in 4.0
-        out.writeUTF(EMPTY_PWD);
+        if (clusterVersion.isLessThan(Versions.V3_9)) {
```
vbekiaris: I think this should be changed to `isUnknownOrLessThan` to explicitly indicate that the group password should be sent out when cluster version is not yet determined.

tkountis:
@vbekiaris done - thx

PR #10735, opened 2017-06-12
Add missing Versioned markers

mdogan:
Normally on OSS side, input/output `getVersion()` returns codebase version
regardless of class being marked as `Versioned`.
But on EE, `getVersion()` returns cluster version for `Versioned` classes,
returns `UNKNOWN` otherwise.

That's why this issue didn't happen on OSS tests.


mdogan:
run-lab-run

PR #10738, opened 2017-06-12
Added EXECUTE_ON_ENTRIES Near Cache tests and fixed Near Cache map proxy

Donnerbart:
The `NearCachedMapProxyImpl` used the wrong key type to invalidate keys when `serializeKeys` is `false`.


Donnerbart:
run-lab-run

PR #10739, opened 2017-06-12
Fixed RingbufferService to fetch read-only RingBufferConfig

Donnerbart:
We should always use `findFooConfig()` to retrieve the read-only version of the config, to prevent user config changes having an effect during runtime. The `RingBufferService` was not doing this.

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
+++ b/hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
@@ -29,7 +29,6 @@ import com.hazelcast.spi.PartitionReplicationEvent;
 import com.hazelcast.spi.RemoteService;
 import com.hazelcast.spi.impl.NodeEngineImpl;
 import com.hazelcast.spi.partition.IPartitionService;
-import com.hazelcast.spi.serialization.SerializationService;
 
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


hazelcast/src/main/java/com/hazelcast/ringbuffer/impl/RingbufferService.java
```diff
@@ -164,33 +165,30 @@ public RingbufferContainer getContainer(String name) {
...
+                getRingbufferConfig(name),
+                nodeEngine.getSerializationService(),
                 nodeEngine.getConfigClassLoader());
         ringbuffer.getStore().instrument(nodeEngine);
         containers.put(name, ringbuffer);
     }
+
+    private RingbufferConfig getRingbufferConfig(String name) {
+        Config config = nodeEngine.getConfig();
+        return config.findRingbufferConfig(getConfigName(name));
```
Donnerbart: This is the important change :)

mmedenjak: Looks fine to me. The only thing I've noticed now, not so much related to this change, is that there is a slight difference in using `find*Config` and `get*Config`. If there is no config that matches by pattern, `get*Config` will add a new config with the requested name but based on the default config (and add the default config if necessary). On the other hand `find*Config` will not add this new config with the requested name. Usually there's no repercussion but I need to think about it in the context of dynamic configs.

Donnerbart: That's true, but all other code uses `find*Config()` to get the read-only config. So the RingBuffer (and actually the Quorum) are "vulnerable" to runtime config changes, without dynamic configs being supported so far.

PR #10742, opened 2017-06-12
Allow use of Key & Trust stores for Management Center config

tkountis:
Support for custom properties in management center config, to allow end-users to provide key & trust stores when needed for mutual-auth between members and MC.

This change makes Management Center config more complex, and some additional logic was put in the Config builder from the XML to handle both old/new scenarios. Please keep that in mind during review.

Example of old:
```<management-center enabled="true" update-interval="2">http://localhost:8080/mancenter</management-center>```

Example of new:
```
<management-center enabled="true" update-interval="8">
    <url>http://foomybar.ber</url>
    <mutual-auth enabled="true">
        <properties>
            <property name="keyStore">/tmp/foo_keystore</property>
            <property name="trustStorePassword">****</property>
            <property name="trustStore">/tmp/foo_truststore</property>
            <property name="keyStorePassword">****</property>
        </properties>
    </mutual-auth>
</management-center>
```

Also, SPI `SSLContextFactory`, originally deprecated for 3.9, is now restored back to full duty to support the change. /cc @pveentjer 

Relevant MC PR: https://github.com/hazelcast/management-center/pull/457
EE: https://github.com/hazelcast/hazelcast-enterprise/pull/1525

### DIFF
--- a/hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
+++ b/hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
@@ -99,12 +99,8 @@ public class ConfigXmlGenerator {
                 .node("license-key", MASK_FOR_SESITIVE_DATA)
                 .node("instance-name", config.getInstanceName());
 
-        if (config.getManagementCenterConfig() != null) {
-            final ManagementCenterConfig mcConfig = config.getManagementCenterConfig();
-            gen.node("management-center", mcConfig.getUrl(),
-                    "enabled", mcConfig.isEnabled(),
-                    "update-interval", mcConfig.getUpdateInterval());
-        }
+
+        manCenterXmlGenerator(gen, config);
         gen.appendProperties(config.getProperties());
         wanReplicationXmlGenerator(gen, config);
         networkConfigXmlGenerator(gen, config);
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: hazelcast/src/main/java/com/hazelcast/nio/ssl/SSLContextFactory.java [hunk 1]
```diff
@@ -21,8 +21,6 @@
 
 /**
  * Factory class for creating {@link javax.net.ssl.SSLContext}
- *
- * @deprecated since Hazelcast 3.9. Use the {@link SSLEngineFactory} instead.
  */
```
pveentjer: Why is this deprecated removed? The SSLEngineFactory should be used.

pveentjer:
Looks good. The only thing I don't understand is the removal of the deprecated section. This interface will be removed in HZ 4

hazelcast/src/main/java/com/hazelcast/config/ConfigXmlGenerator.java
```diff
@@ -868,6 +882,29 @@ private static void sslConfigXmlGenerator(XmlGenerator gen, NetworkConfig netCfg
         gen.close();
     }
 
+    private static void mcMutualAuthConfigXmlGenerator(XmlGenerator gen, ManagementCenterConfig mcConfig) {
+        final MCMutualAuthConfig mutualAuthConfig = mcConfig.getMutualAuthConfig();
+        gen.open("mutual-auth", "enabled", mutualAuthConfig != null && mutualAuthConfig.isEnabled());
+        if (mutualAuthConfig != null) {
```
tkountis: Short answer, no. Long one, if the config is not null (regardless if enabled or not) it might contain properties, so if i check for the enable flag, then the properties will never get saved in the generated XML. We follow the same logic elsewhere in the generator.

PR #10729, opened 2017-06-12
Async client start and reconnect (Client Connection Strategy)

asimarslan:

PR #10747, opened 2017-06-13
Need a helper to track client lifecycle states

ihsandemir:
We needed a helper utility class to track the connected and disconnected states of the client with the introduction of  the new async client connection strategy. The user needs to instantiate the client with the same client config which is used to instantiate the helper ClientStateListener to use it correctly.


asimarslan:
This PR is related to new async client feature #10729

ihsandemir:
verify

ihsandemir:
run-lab-run

PR #10748, opened 2017-06-13
Apply join fixes

metanet:
* After mastership claim is done, check if local uuid is still the same. It might be changed because of force start or split brain merge
* Send explicit suspicion on heartbeat when not joined if sender is known to be version 3.9

Fixes https://github.com/hazelcast/hazelcast-enterprise/issues/1155


metanet:
run-lab-run

